PR #2411, opened 2020-08-13
[CBRD-23606] An error occurs when the binding value is larger than 256 chars in merge query

byungwook-kim:
http://jira.cubrid.org/browse/CBRD-23606

This is a bug not to process a string with 255 or more characters as a character type in case of MERGE INTO statement.
The bug is from the 'long_string_skipped' flag of parse tree node (at pt_print_value function), the flag is set when query statement text has a long string with 255 or more characters.

In do_merge function, to process the long string value in the server the function calls update_objs_for_list_file with list_id.
The update_objs_for_list_file try to assign object (with oid) to string value and it results domain conflict.

I think the long_string_skipped flag might set to prevent long string transferring from cas to server or viseversa. However, the related routines might have some bugs but I couldn't find the bugs.

If anyone find the bugs, please report me.


src/query/query_executor.c
```diff
@@ -1160,6 +1160,21 @@ qexec_end_one_iteration (THREAD_ENTRY * thread_p, XASL_NODE * xasl, XASL_STATE *
...
+	    {
+	      if (xasl->outptr_list->valptrp)
+		{
+		  REGU_VARIABLE_LIST valptrp;
+		  for (valptrp = xasl->outptr_list->valptrp; valptrp; valptrp = valptrp->next)
+		    {
+		      valptrp->value.flags |= REGU_VARIABLE_UPD_INS_LIST;
+		    }
+		}
+	    }
```
hgryoo, line 1175: how about extracting it into a function? The same code exists in the changes.

shparkcubrid:
I commented bug of related routines at jira.

src/query/query_executor.c
byungwook-kim: I modified it as a common routine.

src/query/query_executor.c
```diff
@@ -1160,6 +1182,11 @@ qexec_end_one_iteration (THREAD_ENTRY * thread_p, XASL_NODE * xasl, XASL_STATE *
 	  GOTO_EXIT_ON_ERROR;
 	}
 
+      if (XASL_IS_FLAGED (xasl, XASL_IS_MERGE_QUERY))
+	{
+	  qexec_set_upd_ins_flag (xasl);
+	}
```
shparkcubrid, line 1188: I disagree with the source code location. This seems to be a routine that should be in the XASL generator such as pt_to_merge_xasl(). If it is in end_one_iteration, it will run every row.

shparkcubrid:
How about adding only the code below to the patch for this bug?
```
if (parser->long_string_skipped || parser->print_type_ambiguity)
{
  statement->cannot_prepare = 1;
+  statement->info.merge.flags &= ~PT_MERGE_INFO_SERVER_OP;
  goto cleanup;
}
```
I will check the modifications related to the REGU_VARIABLE_UPD_INS_LIST flag and make a new PR.
By the way, why didn't you include me in the CBRD-23679 issue reviewer? It would have been better if I checked in advance.

shparkcubrid:
I recommend the two changes below for allow_truncated_string.
--allow_truncated_string for client side execution
https://github.com/shparkcubrid/cubrid/commit/b18bf1e3794aea48adb3d2df312776daa3fd8061

--allow_truncated_string for server side execution
https://github.com/shparkcubrid/cubrid/commit/f0e58e2022e3bc1ed829d63acb3924fdad3faa0b

shparkcubrid (approved):

hgryoo (approved):

PR #2412, opened 2020-08-13
[CBRD-23685] coredump while insert something into column of type 'char (1073741823)'

byungwook-kim:
http://jira.cubrid.org/browse/CBRD-23685

the max length for char type is defined at dbtype_def.h as below:

#define DB_MAX_STRING_LENGTH 0x3fffffff
#define DB_MAX_CHAR_PRECISION (DB_MAX_STRING_LENGTH/4)

I modified the length of char type to 0x10000000 (256MB) as above.

csql> create table tb_max (a int, b char(300000000));

In line 4, column 36,

ERROR: before ' )); '
Maximum byte string length is 268435456 bytes.

0 command(s) successfully processed.


src/compat/dbtype_def.h
```diff
@@ -529,7 +529,7 @@ extern "C"
 #define DB_NUMERIC_UNDERFLOW_LIMIT 1e-38
 
 /* The maximum precision that can be specified for a CHAR(n) domain. */
-#define DB_MAX_CHAR_PRECISION DB_MAX_STRING_LENGTH
+#define DB_MAX_CHAR_PRECISION (DB_MAX_STRING_LENGTH/4)
```
hgryoo, line 532: please add a blank before and after a binary operator
```suggestion
#define DB_MAX_CHAR_PRECISION (DB_MAX_STRING_LENGTH / 4)
```

hgryoo: To improve the readability of code, how about using INTL_UTF8_MAX_CHAR_SIZE macro instead of 4?

byungwook-kim: I referred the below definition to keep similar style:

#define DB_MAX_NCHAR_PRECISION (DB_MAX_STRING_LENGTH/2)
/* The maximum precision that can be specified for a NATIONAL CHARACTER VARYING domain. This probably isn't restrictive enough.	
 * We may need to define this functionally as the maximum precision will depend on the size multiplier of the codeset.	 
*/

PR #2413, opened 2020-08-18
[CBRD-23714] change an limit value of max_client

mhoh3963:
modified to occur error if max_clients over than 4000.


cubridci:
Build comment file: 
--------------
#### File: src/base/system_parameter.c (1 incorrect indentation(s) found)
```
  2894    (void *) &prm_css_max_clients_upper,
```
--------------
**Please run 'indent -l120 -lc120' to correct the indentation**

--------------

byungwook-kim:
this setting might not be needed because the global variable prm_css_max_clients_upper is used instead of prm_Def[...]
at prm_tune_parameters(...) function, the following code refer to the prm_css_max_clients_upper directly.

     max_clients = MIN (prm_css_max_clients_upper, css_get_max_socket_fds ());
     if (PRM_GET_INT (max_clients_prm->value) > max_clients)
	  {
	    sprintf (newval, "%d", max_clients);
	    (void) prm_set (max_clients_prm, newval, false);
	  }

mhoh3963:
Thank you for comment. I already knew about this code.
But QA team want to occur error if max_clients over than 4000, so I    decide to modify prm_Def[...] for occurring error.

PR #2425, opened 2020-08-28
[CBRD-23677] fixes bug reported in regression test.

hornetmj:
http://jira.cubrid.org/browse/CBRD-23677

When executing cubrid hb start without setting ha_mode=on in cubrid.conf, the following incorrect message is displayed.

> Invalid Parameter. ha_db_list
Could not load one or more mandatory properties in configuration files.
++ cubrid heartbeat start: fail

so, it was modified to display the correct message as follows.

> The server was not configured for HA.
++ cubrid heartbeat start: fail

### DIFF
--- a/src/executables/util_service.c
+++ b/src/executables/util_service.c
@@ -235,6 +235,7 @@ static UTIL_SERVICE_PROPERTY_T us_Property_map[] = {
 
 
 static const char **Argv;
+static int ha_mode_in_common;
 
 static int util_get_service_option_mask (int util_type);
 static int util_get_command_option_mask (int command_type);
@@ -567,6 +568,8 @@ main (int argc, char *argv[])
 
       return EXIT_FAILURE;
     }
+
+  ha_mode_in_common = prm_get_integer_value (PRM_ID_HA_MODE_FOR_SA_UTILS_ONLY);
 
   if (util_type == ADMIN)
     {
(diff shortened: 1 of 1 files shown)


src/executables/util_service.c
```diff
@@ -4255,6 +4258,14 @@ process_heartbeat (int command_type, int argc, const char **argv)
 #if !defined(WINDOWS)
   HA_CONF ha_conf;
 
+  if (ha_mode_in_common == HA_MODE_OFF)
```
wrlawodms, line 4261: How about using util_get_ha_mode_for_sa_utils () instead of the global variable 'ha_mode_in_common'?
It seems to be used to check this case in other codes like 'process_server()'.
`if ( util_get_ha_mode_for_sa_utils () == HA_MODE_OFF) 
{
...
`

hornetmj: @wrlawodms In my opinion, there is a design level problem with the logic that handles the configuration. the return value of the util_get_ha_mode_for_sa_utils() function at this point becomes HA_MODE_OFF when executing the 'cubrid service start' command after configuring cubrid.conf as follows

>[common]
ha_mode=on
...
[@db1]
ha_mode=on
...
[@db2]
ha_mode=off

this is because properties that can be set for each section are handled by using one global variable.

wrlawodms (approved):

PR #2430, opened 2020-08-31
[CBRD-23733] Add update warning popup when os is Window 7

airnet73:
http://jira.cubrid.org/browse/CBRD-23733

* There are cases where dll cannot be installed in windows7, so Update warning popup is added during installation.


kisoo-han (approved):
Its good for me.

PR #2432, opened 2020-09-01
[CBRD-23737] Leaves a log to trace the cause when a failover is occurred due to a server hang.

hornetmj:
http://jira.cubrid.org/browse/CBRD-23737

The master process performs failover when a server hang occurs. there are two conditions for determining it.

1. no network response.
2. no logs were added for a certain period.

so, leaves a message to distinguish two conditions which are causing the failover.

### DIFF
--- a/src/executables/master_heartbeat.c
+++ b/src/executables/master_heartbeat.c
@@ -4197,11 +4197,11 @@ hb_resource_check_server_log_grow (void)
 	  if (proc->is_curr_eof_received)
 	    {
 #if !defined(WINDOWS)
-	      syslog (LOG_ALERT, "[CUBRID] no change to eof [%lld|%lld] received from (pid:%d)", proc->curr_eof.pageid,
-		      proc->curr_eof.offset, proc->pid);
+	      syslog (LOG_ALERT, "[CUBRID] no change to eof [%lld|%d] received from (pid:%d)",
+		      LSA_AS_ARGS (&proc->curr_eof), proc->pid);
 #endif
-	      MASTER_ER_LOG_DEBUG (ARG_FILE_LINE, "no change to eof [%lld|%lld] received from (pid:%d)\n",
-				   proc->curr_eof.pageid, proc->curr_eof.offset, proc->pid);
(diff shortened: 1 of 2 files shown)


mhoh3963 (approved):

src/executables/master_heartbeat.c
```diff
@@ -4191,6 +4193,23 @@ hb_resource_check_server_log_grow (void)
 	{
 	  proc->server_hang = true;
 	  dead_cnt++;
+
+	  if (proc->is_curr_eof_received)
+	    {
+#if !defined(WINDOWS)
+	      syslog (LOG_ALERT, "[CUBRID] no change to eof [%lld|%lld] received from (pid:%d)", proc->curr_eof.pageid,
```
wrlawodms, line 4200: How about using `%lld|%ld` as format specifiers and `LSA_AS_ARGS (proc_curr_eof)` as the parameter for LOG_LSA. I found out it from other codes to print LSA. It seems to be used for flexbility.

hornetmj: @wrlawodms that's a good idea. however, I got this format from the MASTER_ER_LOG_DEBUG macro at the hb_resource_receive_get_eof() function in this file. the format specifiers are [%lld,%d] if you look at the log_dump_header() and log_dump_record_header_to_string() etc. overall, it is in an inconsistent state. what is the best choice? 1. no change 2. apply your idea to the master_heartbeat.c file only 3. different options.

wrlawodms: `offset` uses only 16 bit, so I think %d is enough. When it comes to `LSA_AS_ARGS`, it is added in the commit `cc368329d0de047` which is about redesigning the File manager. `log_dump_record_header_to_string()` is older code than the commit and it seems not to be applied to use LSA_AS_ARGS by mistake, the way I see it. After all, I think using `LSA_AS_ARGS` is better. What do you think? I respect your decision.

hornetmj: @wrlawodms okay. then I will reflect your feedback in the scope of the master_heartbeat.c file.

wrlawodms (approved):
It looks good to me, thank you!

PR #2444, opened 2020-09-09
[CBRD-23730] remove -u and -p option at tranlist and -p option only work with -i or --kill option on killtran utility

mhoh3963:
If a user uses the wrong password and ignores it, the user can recognize it as the correct password. So I am going to revise as follows.

* remove -u and -p options at tranlist utility
* the -p option of killtran utility only work with killing option as like -i, otherwise occurs error
* modify msg files


cubridci:
Build comment file: 
--------------
#### File: src/executables/util_cs.c (3 incorrect indentation(s) found)
```
  1908 	{
  1909 	  goto print_killtran_usage;
  1910 	}
```
--------------
**Please run 'indent -l120 -lc120' to correct the indentation**

--------------

src/executables/util_cs.c
```diff
@@ -1898,6 +1904,10 @@ killtran (UTIL_FUNCTION_ARG * arg)
   /* disable password, if don't use kill option */
   if (isbatch == 0)
     {
+      if (dba_password != NULL)
+	{
+	  goto print_killtran_usage;
+	}
       dba_password = NULL;
```
kisoo-han, line 1911: This statement is not required. The value of **dba_password** is already NULL.
If it was not NULL, it would have jumped to the **print_killtran_usage** label.

mhoh3963: OK, I will remove this code.

kisoo-han (approved):

PR #2454, opened 2020-09-17
[CBRD-23665] fix to avoid compilation warning

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-23665

fix to avoid compilation warning. (remove int == null condition)


hgryoo (commented):
When calling qdata_free_hscan_key (), it seems val_count always get from HASH_SCAN_KEY.

shparkcubrid:
> When calling qdata_free_hscan_key (), it seems val_count always get from HASH_SCAN_KEY.

No. There is a function call that directly inputs the number while handling exceptions
qdata_copy_hscan_key()
'''
...
qdata_free_hscan_key (thread_p, new_key, i);
'''

hgryoo (approved):

byungwook-kim (approved):

PR #2463, opened 2020-09-28
[CBRD-23607] When updating a column itself using Java SP and there is a WHERE clause condition, if you do not use an index scan, an 'Accessing deleted object' error occurs

hgryoo:
http://jira.cubrid.org/browse/CBRD-23607

When getting a reference attribute node and append to the predicate nodes, the found reference node from `mq_get_references ()` has a link to a node which possibly unrelated with predicates by the `next` pointer. And `pred_nodes = parser_append_node (ref_node, pred_nodes);` just append to the tail of predicate node list (`pred_nodes`). Because of it, When attr nodes split to regu_list_pred and regu_list_rest by pt_find_attribute(), not predicate attribute nodes are classified wrongly as the regu_list_pred.

### **The following printed by XASL_DUMP_DEBUG**

**Problem case:** With predicate, Without index
`UPDATE t1 SET c2 = hello(c2) WHERE c3 = 'Y';`
```
...
-->access spec: class,sequential
        hfid:vfid(3520;0):3521oid[196 4 0]
        regu_list_key:
        regu_list_pred:[TYPE_OID][type:OBJECT];
 [TYPE_ATTR_ID:0x1b1da20]attr_id:1|db_type:VARCHAR; 
 [TYPE_ATTR_ID:0x1b1db18]attr_id:0|db_type:VARCHAR;
        regu_list_rest:
      access pred:[TYPE:NULL][TYPE_ATTR_ID:0x1b1d798]attr_id:0|db_type:VARCHAR = [TYPE_POS_VALUE:0x1b1d810][type:CHAR],single fetch=0
...
```

**Expected**
```
...
-->access spec: class,sequential
        hfid:vfid(3520;0):3521oid[196 4 0]
        regu_list_key:
        regu_list_pred:[TYPE_ATTR_ID:0x2490cb8]attr_id:0|db_type:VARCHAR;
        regu_list_rest:[TYPE_OID][type:OBJECT]; [TYPE_ATTR_ID:0x2490ea8]attr_id:1|db_type:VARCHAR;
      access pred:[TYPE:NULL][TYPE_ATTR_ID:0x2490b28]attr_id:0|db_type:VARCHAR = [TYPE_POS_VALUE:0x2490ba0][type:CHAR],single fetch=0
-->next access spec:
...
```

Comparing with the problem case, you can refer the following two cases
1) Without predicate and Without INDEX
`UPDATE t1 SET c2 = hello(c2);`
```
...
-->access spec: class,sequential
        hfid:vfid(3520;0):3521oid[196 4 0]
        regu_list_key:
        regu_list_pred:
        regu_list_rest:[TYPE_OID][type:OBJECT];
[TYPE_ATTR_ID:0x6182308]attr_id:1|db_type:VARCHAR;
  grouped scan=0,fixed scan=0,single fetch=0
...
```
2) With predicate and With INDEX
`CREATE INDEX idx_t1_001 ON t1 (c3);`
`UPDATE t1 SET c2 = hello(c2) WHERE c3 = 'Y';`
```
-->access spec: class,index<index id:<Btree:(3584;0;3585)>><R_KEY>key info:<key cnt:1>key ranges:<EQ>[[TYPE_POS_VALUE:0x1b11558][type:CHAR]][NIL]<is constant:1><is user given keylimit:0><reset:0> key limit: [NIL][NIL]>
        hfid:vfid(3520;0):3521oid[196 4 0]
        regu_list_key:
        regu_list_pred:
        regu_list_rest:[TYPE_OID][type:OBJECT]; 
[TYPE_ATTR_ID:0x1b10f38]attr_id:1|db_type:VARCHAR; 
[TYPE_ATTR_ID:0x1b11030]attr_id:0|db_type:VARCHAR;
      access range:[TYPE:NULL][TYPE_ATTR_ID:0x1b108b0]attr_id:0|db_type:VARCHAR = [TYPE_POS_VALUE:0x1b10928][type:CHAR],single fetch=0
```


byungwook-kim (approved):

src/parser/xasl_generation.c
```diff
@@ -2744,7 +2744,8 @@ pt_split_attrs (PARSER_CONTEXT * parser, TABLE_INFO * table_info, PT_NODE * pred
 	      node->next = NULL;
 
 	      ref_node = mq_get_references (parser, node, table_info->class_spec);
-	      pred_nodes = parser_append_node (ref_node, pred_nodes);
+	      ref_node_copy = parser_copy_tree (parser, ref_node);
+	      pred_nodes = parser_append_node (ref_node_copy, pred_nodes);
```
shparkcubrid, line 2748: It looks like 'test_sql' wasn't performing properly. After testing, you should check the results.

In my opinion, the result value of mq_get_references(), ref_node, can result in multiple nodes.
``` a.col1 + a.col2 = 3 ```
In the predicate above, two nodes a.col1 and a.col2 should be returned.

The addition of unnecessary columns occurs in the code below. I am checking more.
mq_reset_ids_and_references_helper()
```
  if (get_spec_referenced_attr)
     {
       (void) parser_walk_tree (parser, spec, mq_get_references_node, spec, pt_continue_walk, NULL);
     }
```

**Comment by shparkcubrid** [comment]
How about modifying it like below?

pt_split_attrs()
```
ref_node = mq_get_references (parser, node, table_info->class_spec);
==> ref_nodes = mq_get_references_helper (parser, assignments, spec, false);
```

It is necessary to check whether there are any problems in CTC.

shparkcubrid (approved):

PR #2465, opened 2020-09-29
[CBRD-23726] Add DDL Audit Log

airnet73:
http://jira.cubrid.org/browse/CBRD-23726

* If sql text is ddl, log is recorded.
   * cas ddl audit log file: **<broker_name>_<app_server_num>_ddl.log**
   * csql ddl audit log file: **csql_ddl.log**
* Log item
   * ip, db name, user name, driver version(only cas), sql text(ddl), execution status etc.
* Add ddl audit log paramer to system parameter
   * ddl_audit_log: The default value is **OFF**.
   * ddl_audit_log_dir: The default value is **log/ddl_audit**.
   * ddl_audit_log_size: The default is **10M** and the maximum is **2G**.


kisoo-han:
You proposed use of new 3 system parameters for **ddl_audit**:
1. **ddl_audit_log**: {yes|no}, default **no**
2. **ddl_audit_lgo_dir**: default $CUBRID/log/ddl_audit
3. **ddl_audit_log_size**: upper limit, the minimum value should be 10M and the maximum value should be 2G.

I think use of many system parameters could provide flexibility but users may confused, so I propose the above three system parameters as one as follow:

**ddl_audit_file_size**
  * 0: means **no ddl logging**
  * any natural number: **enable ddl logging**, it is the maximum size of the individual ddl audit file. 
   (allow capacity indication symbol such as KB, MB, GB, TB)

**ddl_audit_lgo_dir**
not used, $CUBRID/log/ddl_audit is mandatory

@mhoh3963 , please let us know your opinion.

PR #2421, opened 2020-08-27
[CBRD-23731] In data comparing, no ignoring to trailing spaces at the end of the data.

byungwook-kim:
http://jira.cubrid.org/browse/CBRD-23731

the following collations are added as default collation:

- iso88591_bin_ts
- utf8_bin_ts
- iso88591_en_cs_ts
- iso88591_en_ci_ts
- utf8_en_cs_ts
- utf8_en_ci_ts
- utf8_tr_cs_ts
- utf8_ko_cs_ts
- euckr_bin_ts 

trail _ts means "trailing space sensitive".

the following macro will be modified to support "ts" in default:

#define LANG_GET_BINARY_COLLATION(c) (((c) == INTL_CODESET_UTF8) \
  ? LANG_COLL_UTF8_BINARY_TS :                                   \
  (((c) == INTL_CODESET_KSC5601_EUC) ? LANG_COLL_EUCKR_BINARY_TS :  \
  (((c) == INTL_CODESET_ISO88591) ? LANG_COLL_ISO_BINARY_TS :    \
    LANG_COLL_BINARY)))

PR #2468, opened 2020-10-12
[CBRD-23731] revised: In data comparing, no ignoring to trailing spaces at the end of the data.

byungwook-kim:

this is revised version for #2421 

the following collations are added as new collation:

iso88591_bin_ti
utf8_bin_ti
iso88591_en_cs_ti
iso88591_en_ci_ti
utf8_en_cs_ti
utf8_en_ci_ti
utf8_tr_cs_ti
utf8_ko_cs_ti
euckr_bin_ti
trail _ti means "trailing space insensitive".

the following macro will be reverted to set TS(trailing space sensitive) as default collation:

#define LANG_GET_BINARY_COLLATION(c) (((c) == INTL_CODESET_UTF8)
? LANG_COLL_UTF8_BINARY:
(((c) == INTL_CODESET_KSC5601_EUC) ? LANG_COLL_EUCKR_BINARY :
(((c) == INTL_CODESET_ISO88591) ? LANG_COLL_ISO_BINARY :
LANG_COLL_BINARY)))


src/optimizer/query_planner.c
```diff
@@ -1128,7 +1128,7 @@ qo_plan_print_sort_spec_helper (PT_NODE * list, bool is_iscan_asc, FILE * f, int
 
       if (TP_TYPE_HAS_COLLATION (TP_DOMAIN_TYPE (list->info.sort_spec.pos_descr.dom))
 	  && TP_DOMAIN_COLLATION (list->info.sort_spec.pos_descr.dom) != LANG_SYS_COLLATION
-	  && TP_DOMAIN_COLLATION (list->info.sort_spec.pos_descr.dom) != (LANG_SYS_COLLATION + COLL_TS)
+	  && TP_DOMAIN_COLLATION (list->info.sort_spec.pos_descr.dom) != (LANG_SYS_COLLATION + COLL_TI)
```
shparkcubrid, line 1131: How about deleting this condition?
I know this condition is for the unusual 'char' case of the previous modification.

src/query/string_opfunc.c
```diff
@@ -581,7 +581,7 @@ db_string_unique_prefix (const DB_VALUE * db_string1, const DB_VALUE * db_string
 
       intl_pad_char (codeset, pad, &pad_size);
 
-      if (collation_id < LANG_COLL_ISO_BINARY_TS)
+      if (collation_id >= LANG_COLL_ISO_BINARY_TI && collation_id < LANG_MAX_BUILTIN_COLLATIONS)
```
shparkcubrid, line 584: Is this condition valid even if collation is added by user?

src/parser/type_checking.c
```diff
@@ -22979,11 +22980,18 @@ pt_common_collation (PT_COLL_INFER * arg1_coll_infer, PT_COLL_INFER * arg2_coll_
...
-      *common_coll = arg2_coll_infer->coll_id;
+      if (arg2_coll_infer->coll_id > arg1_coll_infer->coll_id)
+	{
+	  *common_coll = arg1_coll_infer->coll_id;
+	}
+      else
+	{
+	  *common_coll = arg2_coll_infer->coll_id;
+	}
       *common_cs = arg2_coll_infer->codeset;
```
shparkcubrid, line 22991: Common_coll and common_cs are being allocated from different data(arg1, arg2). is it okay?

src/parser/type_checking.c
```diff
@@ -22979,11 +22980,18 @@ pt_common_collation (PT_COLL_INFER * arg1_coll_infer, PT_COLL_INFER * arg2_coll_
...
 
       /* check arg3 */
-      if (op_has_3_args && arg3_coll_infer->coll_id != *common_coll)
+      if (op_has_3_args && !IS_SAME_VARIATION (arg3_coll_infer->coll_id, *common_coll))
```
shparkcubrid, line 22994: Why should TS and TI be considered the same collation?

src/parser/type_checking.c
byungwook-kim: I fixed it.

src/optimizer/query_planner.c
byungwook-kim: the user can specify ti-collation for char type, so this condition is needed.

src/query/string_opfunc.c
byungwook-kim: this checks the key's collation is ti-type of built-in collation or not.

src/parser/type_checking.c
byungwook-kim: if it's not checked, it cause the collation compatibility fail.
for example, select * from t1, t2 where t1.a = t2.a; which t1.a is ti-collation and t2.a is ts-collation.

hgryoo:
I have some suggestions possibly related to design.

To know a collation ignores trailing spaces or not, comparing with COLL_TI macro is used in this implementation.
The first suggestion is that Instead of using COLL_TI, adding a new boolean variable in `COLL_DATA` (language_support.h) whether cubrid rule at the following is applied or not would be better.
```
<cubridrules>
    <set>
        <scp>20</scp>
        <w>[0.0.0.0]</w>
    </set>
</cubridrules> 
```
From this design, initializing COLL_DATA including `weight vector(array)` and `next alpha array `might be more general. When we load from ldml data, I expected that it is possible to check the cubrid rule is applied or not. then now, we have information from the COLL_DATA whether a collation ignores trailing spaces or not.

The second suggestion is making a variation table explicitly. Instead of using the following code; `(LANG_SYS_COLLATION + COLL_TI)`, How about creating LANG_MAX_COLLATIONS  (# of max collations) size of array and refer the table from IS_SAME_VARIATION macro and LANG_SYS_COLLATION_VARIATION.

```
#define LANG_VARIATION_COLL(coll_id)        lang_variation (coll_id)
#define COLL_IS_SAME_VARIATION(c1,c2)       (c1 == c2 || c1 == LANG_VARIATION_COLL(c1))

static int lang_Variations[LANG_MAX_COLLATIONS] = { 0 };

// lang_Variations [LANG_COLL_UTF8_BINARY] = LANG_COLL_UTF8_BINARY_TI;
// lang_Variations [LANG_COLL_UTF8_BINARY_TI] = LANG_COLL_UTF8_BINARY;

int
lang_variation (const int coll_id)
{
  assert (coll_id >= 0 && coll_id < LANG_MAX_COLLATIONS);

  return lang_Variation[coll_id];
}
```

src/optimizer/query_planner.c
hgryoo, line 1131: How about the following?
1) define LANG_GET_VARIATION_COLLATION (coll)
2) LANG_SYS_COLLATION + COLL_TI => LANG_GET_VARIATION_COLLATION (LANG_SYS_CODESET)
3) remove COLL_TI

byungwook-kim: I add a macro "LANG_SYS_VARIATION" by defined as below:

#define LANG_SYS_VARIATION (LANG_SYS_COLLATION + COLL_TI)

hgryoo: it seems better removing `#define COLL_TI 15`.

byungwook-kim: I removed the macro COLL_TI

src/base/language_support.h
```diff
@@ -144,6 +138,7 @@ enum
 
   /* collation and charset do be used by system : */
 #define LANG_SYS_COLLATION  (LANG_GET_BINARY_COLLATION(lang_charset()))
+#define LANG_SYS_VARIATION  (LANG_SYS_COLLATION + LANG_COLL_ISO_BINARY_TI)
```
hgryoo, line 141:
```suggestion
#define LANG_SYS_VARIATION  (lang_get_collation(LANG_SYS_COLLATION)->coll_variation)
```

src/cci/cci_network.c
```diff
@@ -305,11 +305,10 @@ net_connect_srv (T_CON_HANDLE * con_handle, int host_id, T_CCI_ERROR * err_buf,
     {
       MUTEX_LOCK (create_ssl_mutex);
 
-      ret_value = ssl_session_init (con_handle, srv_sock_fd);
-      if (ret_value < 0)
+      if (ssl_session_init (con_handle, srv_sock_fd) < 0)
 	{
 	  MUTEX_UNLOCK (create_ssl_mutex);
-	  err_code = ret_value;
+	  err_code = CCI_ER_COMMUNICATION;
```
hgryoo, line 311: it seems unrelated change with this PR. please try the following command before the merging.
```
git fetch upstream
git merge upstream/develop
```

src/parser/type_checking.c
```diff
@@ -23014,8 +23021,11 @@ pt_common_collation (PT_COLL_INFER * arg1_coll_infer, PT_COLL_INFER * arg2_coll_
 		  goto error;
 		}
 
-	      *common_coll = arg3_coll_infer->coll_id;
-	      *common_cs = arg3_coll_infer->codeset;
+	      if (*common_coll < arg3_coll_infer->coll_id)
+		{
+		  *common_coll = arg3_coll_infer->coll_id;
+		  *common_cs = arg3_coll_infer->codeset;
+		}
```
hgryoo, line 23028: same with the review above

src/parser/type_checking.c
```diff
@@ -23026,27 +23036,35 @@ pt_common_collation (PT_COLL_INFER * arg1_coll_infer, PT_COLL_INFER * arg2_coll_
...
+      if ((arg2_coll_infer->coll_id < arg1_coll_infer->coll_id))
+	{
+	  *common_coll = arg1_coll_infer->coll_id;
+	  *common_cs = arg1_coll_infer->codeset;
+	}
+      else
+	{
+	  *common_coll = arg2_coll_infer->coll_id;
+	  *common_cs = arg2_coll_infer->codeset;
+	}
```
hgryoo, line 23064: same with the review above

src/parser/type_checking.c
```diff
@@ -23077,8 +23095,11 @@ pt_common_collation (PT_COLL_INFER * arg1_coll_infer, PT_COLL_INFER * arg2_coll_
 		  goto error;
 		}
 
-	      *common_coll = arg3_coll_infer->coll_id;
-	      *common_cs = arg3_coll_infer->codeset;
+	      if (*common_coll < arg3_coll_infer->coll_id)
+		{
+		  *common_coll = arg3_coll_infer->coll_id;
+		  *common_cs = arg3_coll_infer->codeset;
+		}
```
hgryoo, line 23102: same with the review above

src/parser/type_checking.c
```diff
@@ -22979,11 +22978,19 @@ pt_common_collation (PT_COLL_INFER * arg1_coll_infer, PT_COLL_INFER * arg2_coll_
...
+      if (arg2_coll_infer->coll_id < arg1_coll_infer->coll_id)
+	{
+	  *common_coll = arg1_coll_infer->coll_id;
+	  *common_cs = arg1_coll_infer->codeset;
+	}
+      else
+	{
+	  *common_coll = arg2_coll_infer->coll_id;
+	  *common_cs = arg2_coll_infer->codeset;
+	}
```
hgryoo, line 22990: coll_id has no meaning to compare. it is just an index of lang_Collations array.

src/base/language_support.h
byungwook-kim: Ok, I accept it.

src/parser/type_checking.c
byungwook-kim: ok, I fixed it.

src/optimizer/query_planner.c
byungwook-kim: I fixed it, this code is removed.

PR #2475, opened 2020-10-21
[CBRD-23754] Outer join is incorrectly changed to inner join in queries that contain OR operation.

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-23754

The outer join converts to an inner join in the qo_rewrite_outerjoin().
It converts under the following conditions.

```
1. Excluding the join condition, there is a predicate that refers to the outer join table
2. a operation of the predicate must not allow nulls.
```

The problem is that spec check and nullable check work together.
qo_rewrite_outerjoin()
```
 (void) parser_walk_leaves (parser, expr, qo_get_name_by_spec_id, &info, qo_check_nullable_expr, &nullable_cnt);
```
qo_get_name_by_spec_id() and qo_check_nullable_expr() are called as pre function and post function, and when finding a matching spec in qo_get_name_by_spec_id(), the tree stop walking. This is the cause of the problem.

I fix to perform checking nullable and checking equivalent spec respectively.
```
(void) parser_walk_tree (parser, expr, NULL, NULL, qo_check_nullable_expr_with_spec, &info);
(void) parser_walk_tree (parser, expr, qo_get_name_by_spec_id, &info_spec, NULL, NULL);
```


hgryoo (approved):

src/optimizer/query_rewrite.c
```diff
@@ -5457,26 +5501,31 @@ qo_rewrite_outerjoin (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int *c
 	    {
 	      if (spec->info.spec.join_type == PT_JOIN_LEFT_OUTER)
 		{
-		  info.id = spec->info.spec.id;
+		  info.id = info_spec.id = spec->info.spec.id;
 		}
 	      else if (prev_spec != NULL)
 		{
-		  info.id = prev_spec->info.spec.id;
+		  info.id = info_spec.id = prev_spec->info.spec.id;
```
byungwook-kim, line 5508: for else clause, not needed to set info.id?

shparkcubrid: I fixed it

byungwook-kim (approved):

PR #2476, opened 2020-10-21
[CBRD-23731] revised (no collation add): no ignoring to trailing spaces at the end of the data

byungwook-kim:

This is a revision for #2468 

All added ti-collations are removed.
System parameter ignore_trailing_space is added (default value is false).
if the value is set true, the way to deal the trailing space is not change like prior version.


locales/data/ldml/cubrid_es_ES.xml
```diff
@@ -114,4 +120,4 @@
 		</alphabet>
 	</alphabets>
 	<consoleconversion type="ISO88591" windows_codepage="28591" linux_charset="iso88591,ISO_8859-1,ISO8859-1,ISO-8859-1" />		
-</ldml>
+</ldml>
```
shparkcubrid, line 123: NO newline at end of file

shparkcubrid (approved):

hgryoo (approved):

PR #2478, opened 2020-10-26
[CBRD-23756] [Regression] [sql] Core dumped in lang_get_w_first_el

byungwook-kim:
http://jira.cubrid.org/browse/CBRD-23756

This a bug from #2476 (http://jira.cubrid.org/browse/CBRD-23731)

This is a bug from computing the weight of utf8 code set character.
the weights array for a collation consists as two arrays, weights[] and weights_ti (for trailing ignore).
but some collations from lib consists single array only, weights[].
So, we have to change the logic to compute the weight for trailing ignore.

### DIFF
--- a/src/base/language_support.c
+++ b/src/base/language_support.c
@@ -3026,7 +3026,7 @@ lang_strmatch_utf8_w_contr (const LANG_COLLATION * lang_coll, bool is_match, con
   unsigned int cp1, cp2, w_cp1, w_cp2;
   const COLL_DATA *coll = &(lang_coll->coll);
   const int alpha_cnt = coll->w_count;
-  const unsigned int *weight_ptr = (ignore_trailing_space) ? lang_coll->coll.weights_ti : lang_coll->coll.weights;
+  const unsigned int *weight_ptr = lang_coll->coll.weights;
 
   bool is_str1_contr = false;
   bool is_str2_contr = false;
@@ -3768,7 +3768,7 @@ lang_strmatch_utf8_uca_w_level (const COLL_DATA * coll_data, const int level, bo
 		}
 	      else
 		{
-		  if (is_match || !ignore_trailing_space)
+		  if (is_match)
 		    {
 		      result = -1;
 		      goto exit;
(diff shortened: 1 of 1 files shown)


hgryoo:
From the description, it seems that this regression may occur not only `lang_get_w_first_el()` resolved here but the functions at the following where weights_ti is used.
```
lang_strmatch_utf8()
lang_strmatch_utf8_w_contr()
lang_str_utf8_trail_zero_weights()
lang_split_key_iso()
lang_split_key_byte()
lang_split_key_euckr()
lang_fastcmp_byte()
lang_strmatch_byte()
lang_fastcmp_ko()
```

byungwook-kim:
> From the description, it seems that this regression may occur not only `lang_get_w_first_el()` resolved here but the functions at the following where weights_ti is used.
> 
> ```
> lang_strmatch_utf8()
> lang_strmatch_utf8_w_contr()
> lang_str_utf8_trail_zero_weights()
> lang_split_key_iso()
> lang_split_key_byte()
> lang_split_key_euckr()
> lang_fastcmp_byte()
> lang_strmatch_byte()
> lang_fastcmp_ko()
> ```

I will modify the 'lang_strmatch_utf8' only to set weight_ptr at built-in collation.
I will modify the 'lang_strmatch_utf8_w_contr' not to set weight_ptr for 'coll.weights_ti'
I will modify the 'lang_str_utf8_trail_zero_weights' only to set weight_ptr at built-in collation.
And for the other collations, it's already set only for built-in collation.

src/base/language_support.c
```diff
@@ -5830,8 +5839,6 @@ lang_init_coll_Utf8_tr_cs (LANG_COLLATION * lang_coll)
       lang_Next_alpha_char_TR_ti[i] = i + 1;
     }
 
-  assert (DIM (special_lower_cp) == DIM (special_upper_cp));
```
hgryoo, line 5829: why this assert is removed?

byungwook-kim: It's a wrong copied code. It exists at the middle of the function. I copied the code wrongly.

hgryoo: There is the same line at the release/10.2 branch. It seems not the wrong copied code.

byungwook-kim: Aha, I made a mistake.

hgryoo (approved):

shparkcubrid (approved):

PR #2479, opened 2020-10-26
[CBRD-23759] [Regression][ha_repl_debug] Core dumped in btree_get_prefix_separator

byungwook-kim:
http://jira.cubrid.org/browse/CBRD-23759

This a bug from #2476 (http://jira.cubrid.org/browse/CBRD-23731)
This is a bug from btree index prefix processing under char-type index key.

the prefix is handled as varchar in comparing const key value, but the comparing function for char-type mr_cmpval_char handles two comparing argument as ti-style only if both of two arguments type are DB_TYPE_CHAR.
So, we have to change the mr_cmpval_char to process specially in case that one of argument type is DB_TYPE_CHAR under the btree-comparing. I have an idea that cmpval's argument do_coercion has value 2 in btree-compare and mr_cmpval_char do process the case in special.

### DIFF
--- a/src/object/object_primitive.c
+++ b/src/object/object_primitive.c
@@ -11990,7 +11990,14 @@ mr_cmpval_char (DB_VALUE * value1, DB_VALUE * value2, int do_coercion, int total
 
   if (!ignore_trailing_space)
     {
-      ti = (value1->domain.char_info.type == DB_TYPE_CHAR && value2->domain.char_info.type == DB_TYPE_CHAR);
+      if (do_coercion == 2)
+	{
+	  ti = (value1->domain.char_info.type == DB_TYPE_CHAR || value2->domain.char_info.type == DB_TYPE_CHAR);
+	}
+      else
+	{
+	  ti = (value1->domain.char_info.type == DB_TYPE_CHAR && value2->domain.char_info.type == DB_TYPE_CHAR);
+	}
     }
   strc = QSTR_CHAR_COMPARE (collation, string1, (int) db_get_string_size (value1), string2,
 			    (int) db_get_string_size (value2), ti);
@@ -12884,10 +12891,19 @@ mr_cmpval_nchar (DB_VALUE * value1, DB_VALUE * value2, int do_coercion, int tota
       return DB_UNK;
     }
 
+
   if (!ignore_trailing_space)
     {
-      ti = (value1->domain.char_info.type == DB_TYPE_NCHAR && value2->domain.char_info.type == DB_TYPE_NCHAR);
+      if (do_coercion == 2)
+	{
+	  ti = (value1->domain.char_info.type == DB_TYPE_NCHAR || value2->domain.char_info.type == DB_TYPE_NCHAR);
+	}
+      else
+	{
+	  ti = (value1->domain.char_info.type == DB_TYPE_NCHAR && value2->domain.char_info.type == DB_TYPE_NCHAR);
+	}
     }
+
   strc = QSTR_NCHAR_COMPARE (collation, string1, (int) db_get_string_size (value1), string2,
 			     (int) db_get_string_size (value2), db_get_string_codeset (value2), ti);
   c = MR_CMP_RETURN_CODE (strc);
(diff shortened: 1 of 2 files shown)


src/storage/btree.c
```diff
@@ -18721,7 +18721,16 @@ btree_compare_key (DB_VALUE * key1, DB_VALUE * key2, TP_DOMAIN * key_domain, int
 
       if (are_types_comparable)
 	{
-	  c = key_domain->type->cmpval (key1, key2, do_coercion, total_order, NULL, key_domain->collation_id);
+	  bool ignore_trailing_space = prm_get_bool_value (PRM_ID_IGNORE_TRAILING_SPACE);
```
shparkcubrid, line 18724: Since btree comparisons is invoked very often, it doesn't seem like a good way to put it in this.

src/storage/btree.c
```diff
@@ -18721,7 +18721,16 @@ btree_compare_key (DB_VALUE * key1, DB_VALUE * key2, TP_DOMAIN * key_domain, int
...
+
+	  if (!ignore_trailing_space && TP_IS_STRING_TYPE (key1_type) && TP_IS_STRING_TYPE (key2_type))
```
shparkcubrid, line 18726: Is it correct to modify the entire case in btree_compare_key()? Could there be any other case?
Core dumped is generated by btree_get_prefix_separator().

byungwook-kim: I revised the code to apply only _btree_get_prefix_separator_ for modified _btree_compare_key_.

src/storage/btree.c
byungwook-kim: This code was excluded at the next version

shparkcubrid (approved):

hgryoo (commented):
do_coercion is used like boolean in tp_set_compare().
```
tp_set_compare()
...
     if (vtype1 != vtype2)
	{
	  if (!do_coercion)
	    {
	      /* types are not comparable */
	      return DB_NE;
	    }
	  else
	    {
...
```
Is it safe that do_coercion can be assigned as 2? If so, we need to define a new enumeration to clearly know the meaning of 0, 1, and 2 values.

byungwook-kim:
> do_coercion is used like boolean in tp_set_compare().
> 
> ```
> tp_set_compare()
> ...
>      if (vtype1 != vtype2)
> 	{
> 	  if (!do_coercion)
> 	    {
> 	      /* types are not comparable */
> 	      return DB_NE;
> 	    }
> 	  else
> 	    {
> ...
> ```
> 
> Is it safe that do_coercion can be assigned as 2? If so, we need to define a new enumeration to clearly know the meaning of 0, 1, and 2 values.

Yes, your are right. I will define the macro as "#define COERCE_AS_STRING 2", "#define CORECE_AS_DEFAULT 1".

src/object/object_primitive.c
```diff
@@ -11990,7 +11990,14 @@ mr_cmpval_char (DB_VALUE * value1, DB_VALUE * value2, int do_coercion, int total
...
     {
-      ti = (value1->domain.char_info.type == DB_TYPE_CHAR && value2->domain.char_info.type == DB_TYPE_CHAR);
+      if (do_coercion == 2)
+	{
+	  ti = (value1->domain.char_info.type == DB_TYPE_CHAR || value2->domain.char_info.type == DB_TYPE_CHAR);
+	}
+      else
+	{
+	  ti = (value1->domain.char_info.type == DB_TYPE_CHAR && value2->domain.char_info.type == DB_TYPE_CHAR);
+	}
```
hgryoo, line 12000: Can I know why both ignore_trailing_space (from getting parameter) and do_coercion need to be checked here? Are there any cases that do_coercion is given 2 but ignore_trailing_space is set to true at the parameter?

byungwook-kim: At the "btree_get_prefix_separator", the btree_compare_key (finally called as char-type comparing) is called with mixed string argument (CHAR-type and VARCHAR-type).
If the key is composed of CHAR-type, the mixed comparison argument mis-operated as varchar-type-compare because the char-type comparison is operated only if both of two arguments are char-type.

hgryoo (commented):
IMO, comments are needed about the hacky way using do_coercion for the different behavior of routine which uses the mr_cmpval_char().

hgryoo (approved):

PR #2482, opened 2020-11-03
[CBRD-23767] Index scan cannot be used in hierarchical queries when there is a value predicate like '1=1'

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-23767

'1 = 1' is value node. index scan cannot be used because of !PT_IS_VALUE_NODE() condition.
'PT_IS_VALUE_NODE()' condition is unnecessary because value predicate split into filter part, NOT join part in pt_split_join_preds().
so I remove it.


hgryoo:
It seems that '1 = 1' constant folded and be 'true'. So the term becomes the value node. but I wonder why it is not cut-offed in the middle of the rewriting phase.

byungwook-kim (approved):

shparkcubrid:
> It seems that '1 = 1' constant folded and be 'true'. So the term becomes the value node. but I wonder why it is not cut-offed in the middle of the rewriting phase.

Since it works as below, it is not cut-offed in qo_optimize_queries().
```
1. cut-off value node (pt_eval_type() )
2. rewrite '1=1' to 'true' value node.
3. rewrite process related to HQ. <== here is source code i fixed
4. re-eval type. cut-off value node.
```
value node is normally cut-offed in 4.

hgryoo (approved):

hgryoo:
When splitting join predicates, the function (pt_split_join_preds()) assumes that the given predicates are CNF. a value is also a CNF so the function should accept it. It makes sense to me. approved

PR #2487, opened 2020-11-06
[CBRD-23770] Get explicitly specified libjvm path from JVM_PATH env variable

hgryoo:
http://jira.cubrid.org/browse/CBRD-23770

Check JVM_PATH environment variable
- Revised loading libjvm routine from Java SP
1) get JVM_PATH and If it is specified, try to load libjvm at the path
2) If libjvm is not loaded with the JVM_PATH, get JAVA_HOME path
2-1) check the path for JDK 8 or lower version ($JAVA_HOME/jre/lib/.../server/)
2-2) check the path for JDK 11 or upper version ($JAVA_HOME/lib/server/)


shparkcubrid (approved):

src/jsp/jsp_sr.c
```diff
@@ -269,11 +282,30 @@ delay_load_hook (unsigned dliNotify, PDelayLoadInfo pdli)
...
-	void *libVM;
+	char *java_home = NULL, *jvm_path = NULL, *tmp = NULL, *tail;
+	void *libVM = NULL;
 
+	jvm_path = getenv ("JVM_PATH");
 	java_home = getenv ("JAVA_HOME");
+
+	if (jvm_path)
+	  {
+	    err_msgs.append ("\n\tFailed to load libjvm from 'JVM_PATH' envirnment variable: ");
```
mhoh3963, line 293: (1) envirnment -> environment
(2) is it right that error message append to err_msgs, if jvm_path is not null ?

src/jsp/jsp_sr.c
```diff
@@ -290,13 +322,37 @@ delay_load_hook (unsigned dliNotify, PDelayLoadInfo pdli)
 
 	if (java_home)
 	  {
+	    err_msgs.append ("\n\tFailed to load libjvm from 'JAVA_HOME' environment variable: ");
```
mhoh3963, line 325: is it right that error message append to err_msgs, if java_home is not null ?

src/jsp/jsp_sr.c
```diff
@@ -348,38 +404,68 @@ delay_load_dll_exception_filter (PEXCEPTION_POINTERS pep)
...
+      err_msgs.append ("\n\tFailed to get 'JVM_PATH' environment variable");
+    }
 
-      java_home = getenv ("JAVA_HOME");
-      if (java_home != NULL)
+  char *java_home = getenv ("JAVA_HOME");
+  if (java_home != NULL)
+    {
+      char jvm_library_path[PATH_MAX];
+      err_msgs.append ("\n\tFailed to load libjvm from 'JAVA_HOME' environment variable: ");
```
mhoh3963, line 435: is it right that error message append to err_msgs, if jvm_home is not null ?

src/executables/util_service.c
```diff
@@ -2443,19 +2443,23 @@ process_javasp_start (const char *db_name, bool process_window_service)
...
+	      if (javasp_status == JAVASP_SERVER_STOPPED)
 		{
-		  /* invalid database name or failed to open info file, wait upto 3 seconds */
-		  break;
+		  util_log_write_errstr ("Waiting for javasp server to start... (%d / %d)\n", waited_secs,
+					 wait_timeout);
+		}
+	      else
+		{
+		  if (waited_secs > 3)
```
mhoh3963, line 2458: is it correct that if wait_timeout is 30, waited_secs is 3 not 30 ?

src/jsp/jsp_sr.c
hgryoo: (1) Thank you for catching the typo.
(2) If it succeeds to load libjvm, err_msgs will not be used. But, for readability, It would better to move the code. I'll fix it.

src/executables/util_service.c
hgryoo: When starting the JVM, is_javasp_running () can return STOPPED and ERROR.
This is to avoid waiting too long for the user to repeatedly encounter errors when running the JVM.
In the case of ERROR, the reason for setting 3 seconds without exiting immediately is that when the JVM is executed for the first time, Sometimes, ERROR is returned at first for some abnormal reason and then changed to the STOPPED status. And the JVM succeeds to be executed after a certain period of time. This is a kind of heuristic approach.

src/jsp/jsp_sr.c
hgryoo: On windows, If an error occurred in the middle of calling LoadLibrary at the next, it jumps to the exception handler. To gather the appropriate error message, append the error first here.

src/executables/util_service.c
mhoh3963: OK, i got it.

mhoh3963 (approved):

src/jsp/jsp_sr.c
hgryoo: The same reason with above

src/jsp/jsp_sr.c
hgryoo: The same reason with above

byungwook-kim (approved):

PR #2488, opened 2020-11-11
[CBRD-23731] regression-revised: In data comparing, no ignoring to trailing spaces at the end of the data.

byungwook-kim:

This is a revised version for #2483 

The below bug is patched:
- Turkey character set, comparison with ignoring trailing space.
- French (utf8_fr_exp_ab) character set, comparison with ignoring trailing space.
- two fixed length char-type db-string is compared with ignoring trailing space only if both type of them is char-type.
- issue fixed the string concatenation result in sub-query

### DIFF
--- a/src/query/string_opfunc.c
+++ b/src/query/string_opfunc.c
@@ -9302,11 +9302,11 @@ qstr_concatenate (const unsigned char *s1, int s1_length, int s1_precision, DB_T
 
       if (QSTR_IS_NATIONAL_CHAR (s1_type))
 	{
-	  *result_type = DB_TYPE_NCHAR;
+	  *result_type = DB_TYPE_VARNCHAR;
 	}
       else
 	{
-	  *result_type = DB_TYPE_CHAR;
+	  *result_type = DB_TYPE_VARCHAR;
 	}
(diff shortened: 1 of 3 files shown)


src/query/string_opfunc.c
```diff
@@ -423,7 +423,7 @@ db_string_compare (const DB_VALUE * string1, const DB_VALUE * string2, DB_VALUE
 
 	  if (!ignore_trailing_space)
 	    {
-	      ti = (string2_category == QSTR_CHAR || string2_category == QSTR_NATIONAL_CHAR);
+	      ti = (QSTR_IS_FIXED_LENGTH (str1_type) && QSTR_IS_FIXED_LENGTH (str2_type));
```
shparkcubrid, line 426: There is the same case for db_string_compare().
It also looks good to use QSTR_IS_FIXED_LENGTH() 
It seems to be checked overall.

byungwook-kim: I fixed it.

shparkcubrid (approved):

hgryoo (approved):

PR #2491, opened 2020-11-16
[CBRD-23783] Fix incorrect domain definition is printed for SET of arbitrarily type

hgryoo:
http://jira.cubrid.org/browse/CBRD-23783

### DIFF
--- a/src/parser/parse_tree_cl.c
+++ b/src/parser/parse_tree_cl.c
@@ -8577,7 +8577,7 @@ pt_print_datatype (PARSER_CONTEXT * parser, PT_NODE * p)
       q = pt_append_nulstring (parser, q, pt_show_type_enum (p->type_enum));
 
       /* not to print data_type node for SET data types with empty domain */
-      if (p->data_type && p->data_type.info.data_type.precision != TP_FLOATING_PRECISION_VALUE)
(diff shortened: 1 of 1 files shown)


hgryoo:
For the following example,
```
create class tb(
        col1 char(20) primary key,
        col12 set
);
insert /*+ USE_SBR */ into tb  values('aaa', {'abc','def'});
```
The following debugging message is shown at the slave node.
```
...
update_query_execute : insert /*+ USE_SBR */ into tb (col1, col12) values ( cast('bbb' as char(20)),  cast({'abc', 'def'} as set))
...
```

And for a example which is not empty domain (there exists concrete type), 
```
create class tb2_with_domain (
        col1 char(20) primary key,
        col12 set(char(10))
);
insert /*+ USE_SBR */ into tb2_with_domain values('aaa', {'abc','def'});
```
The domain definition is printed as follow
```
update_query_execute : insert /*+ USE_SBR */ into tb2 (col1, col12) values ( cast('bbbb' as char(10)),  cast({'abc', 'def'} as set(char(10))))
```

byungwook-kim (approved):

shparkcubrid (approved):

hornetmj (approved):

PR #2496, opened 2020-11-22
[CBRD-23792] Refactoring get cpu clock

hornetmj:
http://jira.cubrid.org/browse/CBRD-23792

Please refer to Jira's comments.


src/base/get_clockfreq.c
```diff
@@ -43,62 +44,84 @@ get_clockfreq (void)
...
+
+  while ((dest - dp) < 6)
+    {
+      *dest++ = '0';
+    }
+
+  *dest = '\0';
+
+  clock_freq = strtoull (hz, NULL, 10);
+  if (clock_freq == 0)
```
mhoh3963, line 112: It look that if and goto statement is unneeded code.

mhoh3963 (commented):
please, change comment of comment and license as cubrid code's comment.

src/base/get_clockfreq.c
hornetmj: @mhoh3963 Even though the clock_freq value is the same, it indicates that an error is being considered.

mhoh3963 (approved):

PR #2502, opened 2020-11-25
[CBRD-23793] [Regression] Core dumped in db_string_unique_prefix

byungwook-kim:
http://jira.cubrid.org/browse/CBRD-23793
http://jira.cubrid.org/browse/CBRD-23794
related from http://jira.cubrid.org/browse/CBRD-23731

In db_string_compare, ignore-trailing-space comparing should be considered because that db_string_unique_prefix calls the db_string_compare use char-type and varchar-type as arguments. The b-tree node uses char-type can be separated at some point and the prefix of the key is used for doing this separation. Therefore, the prefix of key (varchar) is compared with origin key value including trailing space and the comparing.

I will fix it by adding the comparing function for index prefix.

### DIFF
--- a/src/query/string_opfunc.c
+++ b/src/query/string_opfunc.c
@@ -326,24 +326,25 @@ static int parse_tzd (const char *str, const int max_expect_len);
  *  Public Functions for Strings - Bit and Character
  */
 
-
 /*
  * db_string_compare () -
  *
  * Arguments:
  *                string1: Left side of compare.
  *                string2: Right side of compare
- *                 result: Integer result of comparison.
- *            data_status: Status of errors.
+ *                result: Integer result of comparison.
  *
  * Returns: int
  *
  * Errors:
- *      ER_QSTR_INVALID_DATA_TYPE   :
+ *    ER_QSTR_INVALID_DATA_TYPE   :
  *        <string1> or <string2> are not character strings.
  *
  *    ER_QSTR_INCOMPATIBLE_CODE_SETS:
  *        <string1> and <string2> have differing character code sets.
+ *
+ *    ER_QSTR_INCOMPATIBLE_COLLATIONS
+ *        <string1> and <string2> have incompatible collations.
  *
  */
 
@@ -435,7 +436,7 @@ db_string_compare (const DB_VALUE * string1, const DB_VALUE * string2, DB_VALUE 
 				       DB_GET_UCHAR (string2), (int) db_get_string_size (string2));
 	  break;
 	default:		/* QSTR_UNKNOWN */
-	  break;
+	  assert (false);
 	}
     }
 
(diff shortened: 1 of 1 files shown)


src/query/string_opfunc.c
```diff
@@ -3678,6 +3680,130 @@ db_string_trim (const MISC_OPERAND tr_operand, const DB_VALUE * trim_charset, co
...
+	    }
+	  cmp_result = QSTR_COMPARE (coll_id, DB_GET_UCHAR (string1), (int) db_get_string_size (string1),
+				     DB_GET_UCHAR (string2), (int) db_get_string_size (string2), ti);
+	  break;
+	case QSTR_BIT:
+	  cmp_result = varbit_compare (DB_GET_UCHAR (string1), (int) db_get_string_size (string1),
+				       DB_GET_UCHAR (string2), (int) db_get_string_size (string2));
+	  break;
+	default:		/* QSTR_UNKNOWN */
+	  break;
```
hgryoo, line 3790: "assert" may be needed.
```suggestion
	default:		/* QSTR_UNKNOWN */
	  assert (false);
	  break;
```

src/query/string_opfunc.c
```diff
@@ -325,6 +326,7 @@ static int parse_tzd (const char *str, const int max_expect_len);
  *  Public Functions for Strings - Bit and Character
  */
 
+
```
hgryoo, line 329: redundant newline

src/query/string_opfunc.c
```diff
@@ -3678,6 +3680,130 @@ db_string_trim (const MISC_OPERAND tr_operand, const DB_VALUE * trim_charset, co
...
 
+/*
+ * db_string_prefix_compare () - this function is similar with db_string_compare.
+ * 				 but if one of 2 string arguments char-type
+ * 				 they are compared by the ignore-trailing-space rule.
+ * Arguments:
+ *                string1: Left side of compare.
+ *                string2: Right side of compare
+ *                 result: Integer result of comparison.
+ *            data_status: Status of errors.
```
hgryoo, line 3691: there is no data_status argument

src/query/string_opfunc.c
```diff
@@ -3678,6 +3680,130 @@ db_string_trim (const MISC_OPERAND tr_operand, const DB_VALUE * trim_charset, co
...
+ *
+ * Returns: int
+ *
+ * Errors:
+ *      ER_QSTR_INVALID_DATA_TYPE   :
+ *        <string1> or <string2> are not character strings.
+ *
+ *    ER_QSTR_INCOMPATIBLE_CODE_SETS:
+ *        <string1> and <string2> have differing character code sets.
+ */
```
hgryoo, line 3701: ER_QSTR_INCOMPATIBLE_COLLATIONS can also be returned.

shparkcubrid (approved):

src/query/string_opfunc.c
byungwook-kim: I fixed. thanks.

src/query/string_opfunc.c
byungwook-kim: fixed also db_string_compare.

src/query/string_opfunc.c
byungwook-kim: ok

src/query/string_opfunc.c
byungwook-kim: OK, I fixed it.

hgryoo (approved):

PR #2505, opened 2020-11-27
[CBRD-23798] [Regression] some collation has wrong checked the null value

byungwook-kim:
http://jira.cubrid.org/browse/CBRD-23798

This is related to the issue http://jira.cubrid.org/browse/CBRD-23731 processing ignore trailing space.

The trailing space should be ignore in System Parameter ignore_trailing_space=no, but in some collations the ignoring code is missed.

lang_fastcmp_byte is used at iso8859 collation and lang_strmatch_utf8_uca_w_level is used at uf8_uca collations should be reconsidered because ignoring space is not completed.

First, in lang_fastcmp_byte trailing space is processed well but 0-weight character is not, 0-weight character is ASCII code value 0 and it doesn't mean null.

In lang_strmatch_utf8_uca_w_level, modified point is as below.
  **is_match and ignore_trailing_space follows the same string compare rule to ignore trailing space.**

### DIFF
--- a/src/base/language_support.c
+++ b/src/base/language_support.c
@@ -3785,11 +3785,17 @@ lang_strmatch_utf8_uca_w_level (const COLL_DATA * coll_data, const int level, bo
 	      goto exit;
 	    }
 
+	  if (!ignore_trailing_space)
+	    {
+	      result = -1;
+	      goto exit;
+	    }
+
 	  /* consume any remaining zero-weight values (skip them) from str2 */
 	  do
 	    {
 	      w2 = GET_UCA_WEIGHT (level, ce_index2, uca_w_l13_2, uca_w_l4_2);
-	      if (w2 != 0 || !ignore_trailing_space)
+	      if (w2 != 0)
 		{
 		  /* non-zero weight : strings are not equal */
 		  result = -1;
--- a/src/base/language_support.c
+++ b/src/base/language_support.c
@@ -3768,7 +3768,7 @@ lang_strmatch_utf8_uca_w_level (const COLL_DATA * coll_data, const int level, bo
 		}
 	      else
 		{
-		  if (is_match)
+		  if (is_match || !ignore_trailing_space)
 		    {
 		      result = -1;
 		      goto exit;
@@ -3789,7 +3789,7 @@ lang_strmatch_utf8_uca_w_level (const COLL_DATA * coll_data, const int level, bo
 	  do
 	    {
 	      w2 = GET_UCA_WEIGHT (level, ce_index2, uca_w_l13_2, uca_w_l4_2);
-	      if (w2 != 0)
+	      if (w2 != 0 || !ignore_trailing_space)
 		{
 		  /* non-zero weight : strings are not equal */
 		  result = -1;
@@ -3810,6 +3810,13 @@ lang_strmatch_utf8_uca_w_level (const COLL_DATA * coll_data, const int level, bo
 	      assert (result == 0);
 	      goto exit;
 	    }
+
+	  if (!ignore_trailing_space)
+	    {
+	      result = 1;
+	      goto exit;
+	    }
+
 	  /* consume any remaining zero-weight values (skip them) from str1 */
 	  while (num_ce1 > 0)
 	    {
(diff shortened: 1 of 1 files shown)


src/base/language_support.c
```diff
@@ -5522,6 +5529,11 @@ lang_fastcmp_byte (const LANG_COLLATION * lang_coll, const unsigned char *string
       return cmp;
     }
 
+  if (!ignore_trailing_space && size1 != size2)
```
shparkcubrid, line 5532: I
```suggestion
  else if (!ignore_trailing_space)
```

byungwook-kim: aha, good point

src/base/language_support.c
```diff
@@ -5522,6 +5529,11 @@ lang_fastcmp_byte (const LANG_COLLATION * lang_coll, const unsigned char *string
...
+    {
+      return size1 - size2;
+    }
+
   c1 = c2 = ZERO;
```
shparkcubrid, line 5537: if ignore_trailing_space is set to yes then doesn't the problem occur when c1 or c2 == SPACE?
```
if (c1 == SPACE)
{
  c1 = ZERO;
}
```

byungwook-kim: the SPACE's weight is always 0 from "weight" array, therefore the comparing result is always 0 while the c1 or c2 is SPACE.

src/base/language_support.c
```diff
@@ -3789,7 +3789,7 @@ lang_strmatch_utf8_uca_w_level (const COLL_DATA * coll_data, const int level, bo
 	  do
 	    {
 	      w2 = GET_UCA_WEIGHT (level, ce_index2, uca_w_l13_2, uca_w_l4_2);
-	      if (w2 != 0)
+	      if (w2 != 0 || !ignore_trailing_space)
```
hgryoo, line 3792: why `!ignore_trailing_space` is added? `w2 != 0` is enough to check for non-zero weight and return it strings are not equal.

shparkcubrid, line 3792: Is it correct to result = -1 regardless of the value of w2?
If so, I don't think the condition needs to be here.

byungwook-kim: although the weight is zero, check it is ignore_trailing_space is false because trailing space (zero weight) should not be ignored.

byungwook-kim: because the string1's length is already less than string2's

hgryoo: There may exist a case that `!ignore_trailing_space` is true and w2 is 0 but str2 is not space.

byungwook-kim: actually the code is prior to the "do... " as like below
```
if (!ignore_trailing_space)
  {
     return -1;
  }
```
I move the code to simplify the code line.

byungwook-kim: because the string1's length is already less than string2's, so the comparing is done as string1 < string2.

hgryoo: That means there are remaining chars in string2, this routine consumes trailing zero-weight values.
"space" is zero-weight. but I'm not sure that all zero-weight values are from "space".

src/base/language_support.c
shparkcubrid: if so, why is it written like this below in this function?
```
      if (c2 == SPACE)
	{
	  c2 = ZERO;
	}
      else
	{
	  c2 = weight[c2];
	}
```

src/base/language_support.c
byungwook-kim: Once the trailing character exists, whatever it is, the comparing is done although the weight is zero but not SPACE.
The trailing space's weight zero and the null character is zero weight but the string length is more matter than its weight.

src/base/language_support.c
byungwook-kim: in comparing step, space character 0x20 should be weight 0, it is simply for sorting regardless ignoring or not ignoring trailing space; however in comparing trailing space, the length of a string is more important than remain characters' weight if the comparing option is 'no ignoring space' regardless the weight is 0 or not.

shparkcubrid: I thought the weight[SPACE] is not ZERO because there is the code below on line 5505 in this function.
```
        if (c1 == SPACE)
	{
	  c1 = ZERO;
	}
      else
	{
	  c1 = weight[c1];
	}
```
if weight[SPACE] is ZERO then the above code can be removed. Otherwise above code is need when ignore_trailing_space  is yes.

byungwook-kim: you're write but when ignore_trailing_space setting is no, weight[SPACE] is not zero, so the code is needed.

hgryoo (approved):

shparkcubrid (approved):

PR #2443, opened 2020-09-09
[CBRD-23602] Statement/Result cache to improve performance for frequently repeated queries

byungwook-kim:
http://jira.cubrid.org/browse/CBRD-23602

Environment:
Frequently repeated queries can be used in many applications. These queries are executed repeatedly in a loop of an application routine for processing batch data.

Purpose:
In these circumstances, the performance of the query execution can be improved by caching the statements/results of the repeated queries.

PR #2512, opened 2020-12-02
[CBRD-23801] [Regression] In JDBC, the cursor forward and backward doesn't work.

byungwook-kim:
http://jira.cubrid.org/browse/CBRD-23801

This is for regression test #2443 

After cursor forward or backward jumped to the position not fetched yet, the tuples is not read properly.

### DIFF
--- a/src/jdbc/cubrid/jdbc/jci/UConnection.java
+++ b/src/jdbc/cubrid/jdbc/jci/UConnection.java
@@ -1517,6 +1517,10 @@ public abstract class UConnection {
 	public boolean getOracleStyleEmpltyString() {
 		return connectionProperties.getOracleStyleEmptyString();
 	}
+	
+	public int getClientCacheSize() { /* unit = MByte */
+		return connectionProperties.getClientCacheSize()*1024*1024;
+	}
 
 	public void setCasIp (String casIp) {
 		this.casIp = casIp;
@@ -1881,6 +1885,7 @@ public abstract class UConnection {
 		if (url_cache == null) {
 			UUrlHostKey key = new UUrlHostKey(casIp, casPort, dbname, user);
 			url_cache = UJCIManager.getUrlCache(key);
+			url_cache.setLimit(getClientCacheSize());
 		}
 		return url_cache;
 	}
--- a/src/jdbc/cubrid/jdbc/jci/UConnection.java
+++ b/src/jdbc/cubrid/jdbc/jci/UConnection.java
@@ -1519,7 +1519,7 @@ public abstract class UConnection {
 	}
 	
 	public int getClientCacheSize() { /* unit = MByte */
-		return connectionProperties.getClientCacheSize()*1024*1024;
+		return connectionProperties.getClientCacheSize() * 1024 * 1024;
 	}
 
 	public void setCasIp (String casIp) {
--- a/src/jdbc/cubrid/jdbc/jci/UResCache.java
+++ b/src/jdbc/cubrid/jdbc/jci/UResCache.java
@@ -34,22 +34,30 @@ package cubrid.jdbc.jci;
 public class UResCache {
 	UBindKey key;
 
-	private boolean used;
+	private boolean inUse;
 	private UStatementCacheData cache_data;
 
 	public UResCache(UBindKey key) {
 		this.key = key;
 
 		cache_data = null;
-		used = true;
+		inUse = true;
 	}
 
 	public UStatementCacheData getCacheData() {
-		used = true;
+		inUse = true;
 
 		return (new UStatementCacheData(cache_data));
 	}
 
+	public long getCacheTime() {
+		return cache_data.srvCacheTime;
+	}
+
+	public int getCacheSize() {
+		return cache_data.size;
+	}
+	
 	public void saveCacheData(UStatementCacheData cd) {
 		if (cd.srvCacheTime <= 0)
 			return;
@@ -62,12 +70,17 @@ public class UResCache {
 	}
 
 	public void setExpire() {
-		cache_data.srvCacheTime = 0;
+		inUse = false;
 	}
 	
 	boolean isExpired(long checkTime) {
-		if (cache_data != null && used == false) {
-			return true;
+		if (cache_data != null && !inUse) {
+			if (checkTime > cache_data.srvCacheTime) {
+				return true;
+			}
+			else {
+				return false;
+			}
 		} else {
 			return false;
 		}
--- a/src/jdbc/cubrid/jdbc/jci/UStatement.java
+++ b/src/jdbc/cubrid/jdbc/jci/UStatement.java
@@ -913,7 +913,7 @@ public class UStatement {
 			executeInternal(maxRow, maxField, isScrollable, queryTimeout, cacheData);
 			//jdbc cache feature
 			if (cacheData != null && resultInfo.length == 1) {
-				cacheData.setCacheData(totalTupleNumber, tuples, resultInfo);
+				cacheData.setCacheData(totalTupleNumber, tuples, resultInfo, currentFirstCursor, fetchedTupleNumber);
 			}
 			else if (resultInfo.length > 1) {
 				result_cacheable = false;
@@ -2030,8 +2030,8 @@ public class UStatement {
 		 */
 		Object obj;
 		if ((tuples == null)
-		        || (tuples[cursorPosition] == null)
-		        || ((obj = tuples[cursorPosition]
+		        || (tuples[cursorPosition - currentFirstCursor] == null)
+		        || ((obj = tuples[cursorPosition - currentFirstCursor]
 		                .getAttribute(index)) == null)) {
 			errorHandler.setErrorCode(UErrorCode.ER_WAS_NULL);
 			return null;
@@ -2044,11 +2044,9 @@ public class UStatement {
 		if ((currentFirstCursor < 0)
 		        || (cursorPosition >= 0 && ((cursorPosition < currentFirstCursor) || (cursorPosition > currentFirstCursor
 		                + fetchedTupleNumber - 1)))) {
-			if (tuples[cursorPosition] == null) {
-				fetch();
-				if (errorHandler.getErrorCode() != UErrorCode.ER_NO_ERROR)
-					return false;
-			}
+			fetch();
+			if (errorHandler.getErrorCode() != UErrorCode.ER_NO_ERROR)
+				return false;
 		}
 		return true;
 	}
@@ -2279,16 +2277,7 @@ public class UStatement {
 			fetchedTupleNumber = 0;
 		}
 
-		if (tuples == null) {
-			if (totalTupleNumber > 0) {
-				if (totalTupleNumber > fetchedTupleNumber) {
-					tuples = new UResultTuple[totalTupleNumber];
-				}
-				else {
-					tuples = new UResultTuple[fetchedTupleNumber]; /* for maxRow */
-				}
-			}
-		}
+		tuples = new UResultTuple[fetchedTupleNumber];
 
 		for (int i = 0; i < fetchedTupleNumber; i++) {
 			readATuple(i, inBuffer);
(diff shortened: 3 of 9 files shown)


src/jdbc/cubrid/jdbc/jci/UStatement.java
```diff
@@ -2044,9 +2044,11 @@ private boolean checkReFetch() {
 		if ((currentFirstCursor < 0)
 		        || (cursorPosition >= 0 && ((cursorPosition < currentFirstCursor) || (cursorPosition > currentFirstCursor
 		                + fetchedTupleNumber - 1)))) {
-			fetch();
-			if (errorHandler.getErrorCode() != UErrorCode.ER_NO_ERROR)
-				return false;
+			if (tuples[cursorPosition] == null) {
```
shparkcubrid, line 2047: I don't know exactly this routine. However, one concern is whether it is correct not to fetch again if tuples[cursorPosition]  is not null in updatable cursor.

byungwook-kim: for updatable cursor, the fetch is processed unconditionally through reFech( ).
this routine for cursor moving, and by result-cache the fetch can be minimized.

shparkcubrid (approved):

src/jdbc/cubrid/jdbc/jci/UStatement.java
```diff
@@ -2317,16 +2317,20 @@ private void readATupleByOid(CUBRIDOID oid, UInputBuffer inBuffer)
 
 	private void readATuple(int index, UInputBuffer inBuffer)
 	        throws UJciException {
-		tuples[index] = new UResultTuple(inBuffer.readInt(), columnNumber);
-		tuples[index].setOid(inBuffer.readOID(relatedConnection.getCUBRIDConnection()));
+
+		UResultTuple tuple = new UResultTuple(inBuffer.readInt(), columnNumber);
+		int n = tuple.tupleNumber() - 1;
+		tuples[n] = tuple;
```
hgryoo, line 2323: this line can throw the OutOfIndexException. the `tuples` is allocated as the size of `fetchSize`, and the `tupleNumber()` can be over the `fetchSize`.

hgryoo (changes requested):

src/jdbc/cubrid/jdbc/jci/UStatement.java
byungwook-kim: tuples is allocated with TotalTupleNumber which is the number of whole result tuples not of a fetched tuples.

hgryoo: It seems to slip in #2443.
```
// read_fetch_data ()
		if (tuples == null) {
			if (totalTupleNumber > 0) {
				tuples = new UResultTuple[totalTupleNumber];
			}
		}
```
It may cause OutOfMemoryError if the number of whole result's row is big enough.

byungwook-kim: for result cache the whole results should be cached, however tuples is allocated first empty array not any data assigned.
in allocation step, OutOfMemoryError may not occur.

hgryoo: read_fetch_data () is not only for result cache, it is used in fetching resultset normally. and even data is not assigned, OutOfMemoryError could occur in the allocation step too.

And this change seems to break the main concept: JDBC only receives and processes `fetch_size` results in a single request.
-Allocates an array to hold the result (UResultTuple) as much as `fetch_size`.
-Reads (parsing) the contents of the requested buffer and makes it UResultTuple.
-After reading all `fetch_size`, JDBC will re-request and reallocate the result array.

If `the tuples` is not reallocated, UResultTuple will not be deleted by GC and then OutOfMemoryError will occur eventually.

byungwook-kim: When the  cursor is closed or statement is closed, the tuples array is free.

If we allocate the tuples as fetch size only, is there any way how to keep the result cache? and the performance of result-cache can be degraded. If we wouldn't find a way to keep the cache, we may not use the result cache as policy.

src/jdbc/cubrid/jdbc/jci/UStatement.java
```diff
@@ -841,9 +863,12 @@ private void executeInternal(int maxRow, int maxField,
...
 		if (cacheData != null && cache_reusable == (byte) 1) {
-                	setCacheData(cacheData);
-                	return;
-        	}
+			/* get data from cache */
+			getCacheData(cacheData);
+			stmt_cache_data = cacheData;
+			tuples = cacheData.tuples.get(0);
+			return;
+        }
```
hgryoo: indent

src/jdbc/cubrid/jdbc/jci/UStatement.java
```diff
@@ -811,19 +813,39 @@ private void readResultMeta(UInputBuffer inBuffer) throws UJciException {
...
 		        && totalTupleNumber > 0) {
 			inBuffer.readInt(); // fetch_rescode
 			read_fetch_data(inBuffer, UFunctionCode.FETCH);
 		}
+		
+		//jdbc cache feature
+		if (stmt_cache_data != null) {
+			int total_size = fetchedSize + relatedConnection.getUrlCache().getCacheSize();
+			if (total_size < relatedConnection.getUrlCache().getLimit()) {
+				stmt_cache_data.setCacheData(tuples, currentFirstCursor, fetchedTupleNumber, fetchedSize);
```
hgryoo, line 844: `addCacheData` would be better for readability

src/jdbc/cubrid/jdbc/jci/UUrlCache.java
```diff
@@ -36,10 +36,29 @@
 public class UUrlCache {
 	private Hashtable<String, UStmtCache> stmt_cache_table;
 	private ArrayList<UStmtCache> stmt_cache_remove_list;
+	private int max_size;
+	private int cache_size;
```
hgryoo, line 40: cache_size should be an AtomicInteger.

src/jdbc/cubrid/jdbc/jci/UConnection.java
```diff
@@ -1517,6 +1517,10 @@ public boolean getUseOldBooleanValue() {
 	public boolean getOracleStyleEmpltyString() {
 		return connectionProperties.getOracleStyleEmptyString();
 	}
+	
+	public int getClientCacheSize() { /* unit = MByte */
+		return connectionProperties.getClientCacheSize()*1024*1024;
```
hgryoo, line 1522: indentation for binary operator
```suggestion
		return connectionProperties.getClientCacheSize() * 1024 * 1024;
```

src/jdbc/cubrid/jdbc/jci/UResCache.java
```diff
@@ -34,22 +34,30 @@
 public class UResCache {
 	UBindKey key;
 
-	private boolean used;
+	private boolean inUse;
```
hgryoo, line 37: naming; isInUsed

src/jdbc/cubrid/jdbc/jci/UStatement.java
```diff
@@ -2443,15 +2493,13 @@ public boolean isReturnable() {
...
+	public void getCacheData(UStatementCacheData cache_data) {
 		totalTupleNumber = cache_data.tuple_count;
-		tuples = cache_data.tuples;
 		resultInfo = cache_data.resultInfo;
-
-		cursorPosition = currentFirstCursor = 0;
-		fetchedTupleNumber = totalTupleNumber;
+		currentFirstCursor = cache_data.first.get(0);
+		cursorPosition = 0;
+		fetchedTupleNumber = cache_data.fetched.get(0);
```
hgryoo, line 2501:
```suggestion
		currentFirstCursor = cache_data.getFirstCursor(0);
		cursorPosition = 0;
		fetchedTupleNumber = cache_data.getFetchNumber(0);
```

src/jdbc/cubrid/jdbc/jci/UStmtCache.java
```diff
@@ -81,24 +81,32 @@ void clear() {
...
+			synchronized (res_cache_remove_list) {
+				res_cache_table.remove(victim.key);
+				UResCache lastObj = res_cache_remove_list.remove(res_cache_remove_list.size() - 1);
+				if (v_idx < res_cache_remove_list.size()) {
+					res_cache_remove_list.set(v_idx, lastObj);
 				}
 			}
 		}
+
 		return res_cache_remove_list.size();
```
hgryoo, line 110: Why only the last selected victim is removed?

src/jdbc/cubrid/jdbc/jci/UStatement.java
```diff
@@ -843,9 +863,12 @@ private void executeInternal(int maxRow, int maxField,
 		// cache reusable
 		byte cache_reusable = inBuffer.readByte();
 		if (cacheData != null && cache_reusable == (byte) 1) {
-                	setCacheData(cacheData);
-                	return;
-        	}
+			/* get data from cache */
+			getCacheData(cacheData);
+			stmt_cache_data = cacheData;
+			tuples = cacheData.tuples.get(0);
```
hgryoo, line 869:
```suggestion
			tuples = cacheData.getTuples(0);
```

src/jdbc/cubrid/jdbc/jci/UStatement.java
byungwook-kim: Ok

src/jdbc/cubrid/jdbc/jci/UStatement.java
```diff
@@ -2121,9 +2163,24 @@ public void closeResult() {
...
 			tuples = null;
-			readTupleNumber = 0;
+		}
+	}
+	
+	public void closeResult() {
+		if (stmt_cache_data != null) {
 			getResCache().setExpire();
+			for (int n = 0; n < stmt_cache_data.tuples.size(); n++) {
+				if (tuples == stmt_cache_data.tuples.get(n)) {
```
hgryoo, line 2173: How about testing from the last index of tuples of cached data?
```suggestion
			for (int n = stmt_cache_data.tuples.size(); n > 0; n--) {
				if (tuples == stmt_cache_data.tuples.get(n - 1)) {
```

byungwook-kim: Ok

src/jdbc/cubrid/jdbc/jci/UUrlCache.java
byungwook-kim: another way:
use synchronize at updating?

src/jdbc/cubrid/jdbc/jci/UStmtCache.java
byungwook-kim: for simplifying LRU, for each statement cache check victims and remove a victim because the result cache is distributed to several statement cache.

src/jdbc/cubrid/jdbc/jci/UResCache.java
byungwook-kim: I think "inUse" is better than your suggestion.

src/jdbc/cubrid/jdbc/jci/UStatement.java
byungwook-kim: Ok,

src/jdbc/cubrid/jdbc/jci/UUrlCache.java
hgryoo: `AtomicInteger` is significantly faster than `synchronized` block.

src/jdbc/cubrid/jdbc/jci/UStatement.java
byungwook-kim: Ok

src/jdbc/cubrid/jdbc/jci/UStmtCache.java
hgryoo: Do you mean that the last index in res_cache_remove_list is the least recently used?

src/jdbc/cubrid/jdbc/jci/UResCache.java
hgryoo: ok.

src/jdbc/cubrid/jdbc/jci/UStmtCache.java
byungwook-kim: by the victim selection algorithm, it picks the least serverTime of result-caches.

hgryoo: Ok, I undetstood.

src/jdbc/cubrid/jdbc/jci/UUrlCache.java
byungwook-kim: show me a partial code for updating.

hgryoo (approved):

PR #2521, opened 2020-12-08
[CBRD-23808] Core dumped in csql_execute_statements

airnet73:
http://jira.cubrid.org/browse/CBRD-23808

In case of FILE_INPUT type in csql, sql_user_text of the statement is null. Therefore, add null check to sql_user_text

### DIFF
--- a/src/executables/csql.c
+++ b/src/executables/csql.c
@@ -1862,7 +1862,8 @@ csql_execute_statements (const CSQL_ARGUMENT * csql_arg, int type, const void *s
       if (session->statements != NULL)
 	{
 	  statement = session->statements[num_stmts];
-	  if (statement != NULL && strlen (statement->sql_user_text) >= statement->sql_user_text_len)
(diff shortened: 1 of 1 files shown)


src/executables/csql.c
```diff
@@ -1862,7 +1862,8 @@ csql_execute_statements (const CSQL_ARGUMENT * csql_arg, int type, const void *s
       if (session->statements != NULL)
 	{
 	  statement = session->statements[num_stmts];
-	  if (statement != NULL && strlen (statement->sql_user_text) >= statement->sql_user_text_len)
+	  if (statement != NULL && statement->sql_user_text != NULL
```
kisoo-han, line 1865: In CUBRID, we code this type with the following style as well (not important):
` if (statement  && statement->sql_user_text`

kisoo-han (approved):

src/executables/csql.c
airnet73: Thank you, i will change code as your recommend.

PR #2524, opened 2020-12-08
[CBRD-23731] Regression: in prepare containing + '%' query cast as char which expected cast as varchar.

byungwook-kim:

in prepare statement containing + ''%'''  query cast as char which '' expected cast as varchar.

### DIFF
--- a/src/parser/type_checking.c
+++ b/src/parser/type_checking.c
@@ -24378,7 +24378,7 @@ static PT_TYPE_ENUM
 pt_wrap_type_for_collation (const PT_NODE * arg1, const PT_NODE * arg2, const PT_NODE * arg3,
 			    PT_TYPE_ENUM * wrap_type_collection)
 {
-  PT_TYPE_ENUM common_type = PT_TYPE_CHAR;
+  PT_TYPE_ENUM common_type = (PT_IS_COLLECTION_TYPE (arg1->type_enum))?PT_TYPE_CHAR:PT_TYPE_VARCHAR;
   PT_TYPE_ENUM arg1_type = PT_TYPE_NONE, arg2_type = PT_TYPE_NONE, arg3_type = PT_TYPE_NONE;
 
   if (arg1)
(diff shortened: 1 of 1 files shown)


shparkcubrid:
What test cases is this PR for? It would be better to specify.

src/parser/type_checking.c
```diff
@@ -24378,7 +24378,7 @@ static PT_TYPE_ENUM
 pt_wrap_type_for_collation (const PT_NODE * arg1, const PT_NODE * arg2, const PT_NODE * arg3,
 			    PT_TYPE_ENUM * wrap_type_collection)
 {
-  PT_TYPE_ENUM common_type = PT_TYPE_CHAR;
+  PT_TYPE_ENUM common_type = (PT_IS_COLLECTION_TYPE (arg1->type_enum))?PT_TYPE_CHAR:PT_TYPE_VARCHAR;
```
shparkcubrid, line 24381: Why is it a char type only when it is a collection type?

hgryoo: This line is changed to PT_TYPE_CHAR from PT_TYPE_VARCHAR at #2483.
It would better that reverting this line to the previous implementation and resolve common_type which starts from type_checking:24425.

byungwook-kim: in #2483, for collection type it's changed to CHAR but host-variable is also set to CHAR, so I modified to set CHAR for collection type only

shparkcubrid: For what test case did you change to CHAR in #2483? In my opinion, it would be better to specify the test case for history management.

shparkcubrid (approved):

src/parser/type_checking.c
hgryoo, line 24381: ok, I got it.
please fix indent before merging.
```suggestion
  PT_TYPE_ENUM common_type = (PT_IS_COLLECTION_TYPE (arg1->type_enum)) ? PT_TYPE_CHAR : PT_TYPE_VARCHAR;
```

byungwook-kim: for #2483
medium/_02_xtests/cases/settyp3.sql
create class set_attr(
se_object set(object),
se_obj_char set( set_attr, string));
insert into set_attr values({},{}) to s2;

byungwook-kim: Ok,

hgryoo (approved):

PR #2525, opened 2020-12-09
[CBRD-23810] Improvement of DML and transaction processing for result-cache

byungwook-kim:
http://jira.cubrid.org/browse/CBRD-23810

Improvement of DML and transaction processing for result-cache

The update query affects the result-cache by cache invalidation for related tables.
We should minimize the affect for performance and the transaction processing should be also dealt in detail.
- The invalidation for DML does not need to start for 0 cache entry.
- The invalidation for DML does not need to start for no-cache mode.
- query caches "IN Transaction" must not refer.

### DIFF
--- a/src/query/query_manager.c
+++ b/src/query/query_manager.c
@@ -1531,10 +1531,19 @@ xqmgr_execute_query (THREAD_ENTRY * thread_p, const XASL_ID * xasl_id_p, QUERY_I
 	    }
 
 	  /* update the cache entry for the result associated with the used parameter values (DB_VALUE array) if there
-	   * is, or make new one */
-	  list_cache_entry_p =
-	    qfile_update_list_cache_entry (thread_p, &xasl_cache_entry_p->list_ht_no, &params, list_id_p,
-					   xasl_cache_entry_p);
+	   * is, or make new one
+	   * in case list_ht_no is less than 0,
+	   *     the cache is not found and should be newly added (surely list_cache_entry_p is null)
+	   * in case list_ht_no is not less than 0 and list_cache_entry_p is not null
+	   *     the cache entry is found but the entry is used by other transaction
+	   */
+	  if (list_cache_entry_p || xasl_cache_entry_p->list_ht_no < 0)
+	    {
+	      list_cache_entry_p =
+		qfile_update_list_cache_entry (thread_p, &xasl_cache_entry_p->list_ht_no, &params, list_id_p,
+					       xasl_cache_entry_p);
+	    }
+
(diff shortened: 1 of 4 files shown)


src/query/query_manager.c
```diff
@@ -1531,10 +1531,19 @@ xqmgr_execute_query (THREAD_ENTRY * thread_p, const XASL_ID * xasl_id_p, QUERY_I
 	    }
 
 	  /* update the cache entry for the result associated with the used parameter values (DB_VALUE array) if there
-	   * is, or make new one */
-	  list_cache_entry_p =
-	    qfile_update_list_cache_entry (thread_p, &xasl_cache_entry_p->list_ht_no, &params, list_id_p,
-					   xasl_cache_entry_p);
+	   * is, or make new one
+	   * in case list_ht_no is less than 0,
+	   *     the cache is not found and should be newly added (surely list_cache_entry_p is null)
```
hgryoo, line 1536: please add `assert (xasl_cache_entry_p->list_ht_no < 0 && list_cache_entry_p == null)`

src/query/query_manager.c
```diff
@@ -1531,10 +1531,19 @@ xqmgr_execute_query (THREAD_ENTRY * thread_p, const XASL_ID * xasl_id_p, QUERY_I
...
+	   * in case list_ht_no is not less than 0 and list_cache_entry_p is not null
+	   *     the cache entry is found but the entry is used by other transaction
```
hgryoo, line 1538: please add assert for the cases described here in the following if-block

byungwook-kim: the comment is error, should be as follow
in case list_ht_no is not less than 0 and list_cache_entry_p is null

shparkcubrid (approved):

src/query/query_manager.c
hgryoo: list_ht_no >= 0 && list_cache_entry_p != null: the routine doesn't reaches inside of the following if-block

Other cases for list_ht_no and list_cache_entry_p will reach inside of the following if-block.
1) **list_ht_no < 0** && list_cache_entry_p == null
2) list_ht_no >= 0 && **list_cache_entry_p != null**
3) **list_ht_no < 0** && **list_cache_entry_p != null**

The if-condition includes all of the above cases, but only 1) is described in the comment. In order to guarantee that there are no other cases 2) and 3), assert is required.

byungwook-kim: Ok, I add assert for case "list_ht_no < 0 && list_cache_entry_p is not null"

hgryoo (approved):

PR #2529, opened 2020-12-10
[CBRD-23629] Fix slip of javasp utility

hgryoo:
http://jira.cubrid.org/browse/CBRD-23629

This PR fixes the following regression. It caused because an error is not cleared before starting JVM in javasp_start_server (). errors created by er_set() in javasp_is_running () was detected by jsp_start_server() and was handled incorrectly. it should be cleared.

**Repro.**  
```
cubrid javasp start demodb
ps -ef | grep cub_javasp
kill -9 <pid>
cubrid javasp start demodb
```

**Expected**
```
@ cubrid javasp start: demodb
++ cubrid javasp start: success
```

**Actual** 
```
@ cubrid javasp start: demodb
Can't connect Java VM: connect()
++ cubrid javasp start: fail
```

And I fixed some trivial slip.

### DIFF
--- a/src/executables/javasp.cpp
+++ b/src/executables/javasp.cpp
@@ -211,7 +211,7 @@ main (int argc, char *argv[])
 	    goto exit;
 	  }
 
-	status = javasp_start_server (jsp_info, db_name, pathname.c_str ());
+	status = javasp_start_server (jsp_info, db_name, pathname);
 	if (status == NO_ERROR)
 	  {
 	    while (true)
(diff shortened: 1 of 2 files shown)


src/executables/javasp.cpp
```diff
@@ -211,7 +211,7 @@ main (int argc, char *argv[])
 	    goto exit;
 	  }
 
-	status = javasp_start_server (jsp_info, db_name, pathname.c_str ());
+	status = javasp_start_server (jsp_info, db_name, pathname);
```
byungwook-kim, line 214: what is different pathname.c_str() and pathname?

hgryoo: c_str() returns char* of std::string. the type of an argument of javasp_start_server is std::string&.
it is just a slip that causes creating new std::string from char* by the constructor.

byungwook-kim: please check the windows build fail

shparkcubrid (approved):

byungwook-kim (approved):

src/executables/javasp.cpp
hgryoo: I fixed it, thank you.

PR #2530, opened 2020-12-11
[CBRD-23810] Regression, DML and transaction processing for result-cache

byungwook-kim:
http://jira.cubrid.org/browse/CBRD-23810

The update query affects the result-cache by cache invalidation for related tables.

We should minimize the affect for performance and the transaction processing should be also dealt in detail.

- The invalidation for DML does not need to start for 0 cache entry.
- The invalidation for DML does not need to start for no-cache mode.
- query caches "IN Transaction" must not refer.

I fixed some bugs related to transaction processing result-cached.


src/query/query_manager.c
```diff
@@ -2299,11 +2275,12 @@ qmgr_add_modified_class (THREAD_ENTRY * thread_p, const OID * class_oid_p)
...
   tran_index = LOG_FIND_THREAD_TRAN_INDEX (thread_p);
   tran_entry_p = &qmgr_Query_table.tran_entries_p[tran_index];
-
-  if (tran_entry_p->modified_classes_p == NULL
-      && (tran_entry_p->modified_classes_p = qmgr_allocate_oid_block (thread_p)) == NULL)
+  if (tran_entry_p->modified_classes_p == NULL)
     {
-      return;
+      tran_entry_p->modified_classes_p = (OID_BLOCK_LIST *) malloc (sizeof (OID_BLOCK_LIST));
+      assert (tran_entry_p->modified_classes_p != NULL);
```
hgryoo, line 2281: How about error handling as below?
```suggestion
      if (tran_entry_p->modified_classes_p == NULL)
    {
	  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, sizeof (OID_BLOCK_LIST));
	  return;
    }
```

src/base/system_parameter.c
```diff
@@ -3623,7 +3623,7 @@ static SYSPRM_PARAM prm_Def[] = {
    (DUP_PRM_FUNC) NULL},
   {PRM_ID_LIST_MAX_QUERY_CACHE_ENTRIES,
    PRM_NAME_LIST_MAX_QUERY_CACHE_ENTRIES,
-   (PRM_FOR_SERVER),
+   (PRM_FOR_SERVER | PRM_FORCE_SERVER),
```
hgryoo, line 3626: PRM_ID_LIST_MAX_QUERY_CACHE_ENTRIES and PRM_ID_LIST_MAX_QUERY_CACHE_PAGES should apply PRM_HIDDEN same like PRM_ID_LIST_QUERY_CACHE_MODE?

src/base/system_parameter.c
```diff
@@ -3635,7 +3635,7 @@ static SYSPRM_PARAM prm_Def[] = {
    (DUP_PRM_FUNC) NULL},
   {PRM_ID_LIST_MAX_QUERY_CACHE_PAGES,
    PRM_NAME_LIST_MAX_QUERY_CACHE_PAGES,
-   (PRM_FOR_SERVER),
+   (PRM_FOR_SERVER | PRM_FORCE_SERVER),
```
hgryoo, line 3638: PRM_ID_LIST_MAX_QUERY_CACHE_ENTRIES and PRM_ID_LIST_MAX_QUERY_CACHE_PAGES should apply PRM_HIDDEN same like PRM_ID_LIST_QUERY_CACHE_MODE?

src/query/list_file.c
```diff
@@ -6002,6 +5978,14 @@ qfile_update_list_cache_entry (THREAD_ENTRY * thread_p, int *list_ht_no_ptr, con
 	  break;
 	}
 
+#if defined(SERVER_MODE)
+      /* check in-use by other transaction */
+      if ((int) lent->last_ta_idx > 0);
+      {
+	csect_exit (thread_p, CSECT_QPROC_LIST_CACHE);
+	return lent;
+      }
```
hgryoo, line 5987: checking indentation is required.

src/query/xasl_cache.c
```diff
@@ -2223,9 +2223,10 @@ xcache_cleanup (THREAD_ENTRY * thread_p)
       while ((xcache_entry = iter.iterate ()) != NULL && count < xcache_Soft_capacity)
 	{
 	  candidate.xid = xcache_entry->xasl_id;
-	  candidate.xcache = xcache_entry;
-	  if (candidate.xid.cache_flag > 0 || (candidate.xid.cache_flag & XCACHE_ENTRY_FLAGS_MASK)
```
shparkcubrid, line 2227: candidate.xid.cache_flag > 0 means !XCACHE_ENTRY_MARK_DELETED and XCACHE_ENTRY_FIX_COUNT_MASK > 0.
can it be deleted?

src/base/system_parameter.c
byungwook-kim: PRM_ID_LIST_MAX_QUERY_CACHE_ENTRIES and PRM_ID_LIST_MAX_QUERY_CACHE_PAGES should be allowed to disable the "query cache mode". It is a policy.

src/query/xasl_cache.c
byungwook-kim: I reverted to 10.2. The original code of 10.2 did not check cache_flag, so I modified the code but I reverted it because it might affect performance (ex. YCSB). I think this code does not affect the performance and I will revert again.

src/query/query_manager.c
byungwook-kim: It's not recoverable. The system is corrupted something.

shparkcubrid (approved):

src/base/system_parameter.c
hgryoo: ok, I understand.

hgryoo (approved):

PR #2531, opened 2020-12-16
[CBRD-23804][Regression] Core dumped in btree_key_find_and_insert_delete_mvccid

byungwook-kim:
http://jira.cubrid.org/browse/CBRD-23804
http://jira.cubrid.org/browse/CBRD-23814

Core dumped in btree_key_find_and_insert_delete_mvccid

The b-tree key compare is not same as general compare when the char-type and varchar-type is mixed.
The mixed type case is commonly made at splitting b-tree node  to check a split position of the tree nodes.

I modified the codes related to b-tree compare to process mixed char and varchar-type.
The coerce 2 means tow comparing arguments may have char-type and varchar-type each other.

### DIFF
--- a/src/storage/btree.c
+++ b/src/storage/btree.c
@@ -11767,19 +11767,13 @@ btree_get_prefix_separator (const DB_VALUE * key1, const DB_VALUE * key2, DB_VAL
 {
   int c;
   int err = NO_ERROR;
-  static bool ignore_trailing_space = prm_get_bool_value (PRM_ID_IGNORE_TRAILING_SPACE);
-
-  /* for coerce = 2, we need to process key comparing as char-type
-   * in case that one of two arguments has varchar-type
-   * if the other argument has char-type */
-  static int coerce = (ignore_trailing_space) ? 1 : 2;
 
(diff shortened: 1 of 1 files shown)


src/storage/btree.c
```diff
@@ -11767,19 +11767,13 @@ btree_get_prefix_separator (const DB_VALUE * key1, const DB_VALUE * key2, DB_VAL
 {
   int c;
   int err = NO_ERROR;
-  static bool ignore_trailing_space = prm_get_bool_value (PRM_ID_IGNORE_TRAILING_SPACE);
```
shparkcubrid, line 11770: Is it not necessary to change the behavior according to the system parameters anymore?

shparkcubrid:
What is the cause of the core? What problem has this fix solved?

src/storage/btree.c
byungwook-kim: at btree_compare_key **do_coercion** is set to 2 and at **cmp_val ignore_trailing_space** is also checked.

byungwook-kim:
> What is the cause of the core? What problem has this fix solved?
In prior version, the key compare with coersion=2 is only done at split key for splitting node but the comparison can cause to mislead the btree key search in varchar and char-type mixed.

shparkcubrid (approved):

hgryoo (approved):

PR #2535, opened 2021-01-04
[CBRD-23812] [Regression] The results of two queries that scan different indexes in the same transaction are different.

byungwook-kim:
http://jira.cubrid.org/browse/CBRD-23812
related to http://jira.cubrid.org/browse/CBRD-23731

Environment:
1. Create a table by increasing columns from 2 to 30.
2. Insert/update/select sequentially with 10 threads targeting each table.
3. 10,000 times per thread.

Fail case:
When comparing each column value by executing the same select query that scans different indexes in the same transaction, in the select result comparison step, the result of the first query is failing with 0 tuple.

I found a weak point to compare with "midxkey" in b-tree index processing. The point is related to coercion at the different components consist of multi-column key. The comparing is performed by pr_midxkey_compare_element  through the pr_midxkey_compare function. I modified the coercion of pr_midxkey_compare to 2 to char-type compare in char-type and varchar-type mixed case. In modified version, the fail case is not occurred.

### DIFF
--- a/src/object/object_domain.c
+++ b/src/object/object_domain.c
@@ -4110,7 +4110,7 @@ tp_domain_select (const TP_DOMAIN * domain_list, const DB_VALUE * value, int all
 
   best = NULL;
 
-  static bool ti = true;
+  bool ti = true;
   static bool ignore_trailing_space = prm_get_bool_value (PRM_ID_IGNORE_TRAILING_SPACE);
 
   /*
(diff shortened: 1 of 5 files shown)


hgryoo (approved):

shparkcubrid:
There is a problem with ambiguous results.
In Oracle, there is no data in both cases.

--repro
drop table tmp;
create table tmp(col1 char(10), col2 varchar(10));
insert into tmp values('a ','a ');
-- heap scan
select * from tmp where col1 = 'a' and col2 = 'a';
```
There are no results.
```
create index idx on tmp(col1,col2);
--index scan
select * from tmp where col1 = 'a' and col2 = 'a';
```
  col1                  col2
  'a         '          'a '
```

shparkcubrid (approved):

PR #2537, opened 2021-01-06
[CBRD-23828] Support the use of OID of temp file in HASH LIST SCAN.

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-23828

The HASH LIST SCAN operates with the following priorities. (check_hash_list_scan() )
```
1. in-memory hash if all data can be stored (hash value : tuple)
2. Hybrid in-memory hash if all OID of data can be stored (hash value : OID)
3. LIST SCAN if all data cannot be stored (After CBRD-23816, use HASH TEMP FILE.)
```

**Hash table for HASH LIST SCAN**
In order to minimize the size of the hash entry, a hash table for HASH LIST SCAN is created separately. It has the following features.
```
1. lru, act is not used. (remove variable related to lru, act)
2. key comparison is performed in executor. (remove key of hash entry)
3. put data orderly. (add tail pointer for it)
4. Since hash size is fixed, rehashing the hash table is not necessary.
```
A key of entry of hash table is removed. Previously, a data filter related to key comparison has been performed redundantly in the executor. If hash collision occur, this is a way to improve performance. However, it is rather the cause of performance degradation in situations where hash collisions do not occur. I believe that deleting duplicate key comparison will be more advantageous in all cases. And to reduce the size of the hash entry, the key is needed to remove.

**EXECUTOR**
OIDs of temp file are stored when building hash table. (scan_build_hash_list_scan() )
When probing hash table, data is retrieved from temp file using OIDs. (scan_hash_probe_next() )

### DIFF
--- a/src/base/memory_hash.c
+++ b/src/base/memory_hash.c
@@ -1568,7 +1568,10 @@ mht_get_hls (const MHT_HLS_TABLE * ht, const void *key, void **last)
 
   if (hentry != NULL)
     {
-      *((HENTRY_HLS_PTR *) last) = hentry;
+      if (last != NULL)
+	{
+	  *((HENTRY_HLS_PTR *) last) = hentry;
+	}
       return hentry->data;
     }
   return NULL;
--- a/src/base/memory_hash.c
+++ b/src/base/memory_hash.c
@@ -105,7 +105,7 @@ static int mht_rehash (MHT_TABLE * ht);
 
 static const void *mht_put_internal (MHT_TABLE * ht, const void *key, void *data, MHT_PUT_OPT opt);
 static const void *mht_put2_internal (MHT_TABLE * ht, const void *key, void *data, MHT_PUT_OPT opt);
-static const void *mht_put_orderly_internal (MHT_TABLE * ht, const void *key, void *data, MHT_PUT_OPT opt);
+static const void *mht_put_hls_internal (MHT_HLS_TABLE * ht, const void *key, void *data, MHT_PUT_OPT opt);
 
 static unsigned int mht_get_shiftmult32 (unsigned int key, const unsigned int ht_size);
 #if defined (ENABLE_UNUSED_FUNCTION)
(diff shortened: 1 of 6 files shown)


src/parser/type_checking.c
```diff
@@ -5530,6 +5530,7 @@ pt_coerce_range_expr_arguments (PARSER_CONTEXT * parser, PT_NODE * expr, PT_NODE
 	      if (PT_IS_STRING_TYPE (common_type) && PT_IS_STRING_TYPE (temp->type_enum))
 		{
 		  /* A bigger codesets's number can represent more characters. */
+		  /* to_do : check to use functions pt_common_collation() or pt_make_cast_with_compatble_info(). */
```
byungwook-kim, line 5533: unrelated code is included. it looks merged from the other issue.

src/query/query_hash_scan.h
```diff
@@ -29,11 +29,31 @@
...
+  int offset;			/* Tuple offset inside the page */
+};
+
 /* hash scan value */
 typedef struct hash_scan_value HASH_SCAN_VALUE;
 struct hash_scan_value
 {
-  QFILE_TUPLE tuple;		/* tuple */
+  void *data;
 };
```
hgryoo, line 57: How about using a union?
```
typedef union hash_scan_value HASH_SCAN_VALUE;
union hash_scan_value
{
  QFILE_TUPLE_SIMPLE_POS *pos;
  QFILE_TUPLE *tuple;
};
```

src/query/scan_manager.c
```diff
@@ -7987,7 +8056,7 @@ scan_hash_probe_next (THREAD_ENTRY * thread_p, SCAN_ID * scan_id, QFILE_TUPLE *
 
 /*
  * check_hash_list_scan () - Check if hash list scan is possible
- *   return: bool
+ *   return: int  1: in-memory 2: hybrid in-memory
```
hgryoo, line 8059: trivial; Defining it with enum type would better for readability

src/query/query_hash_scan.h
shparkcubrid: okay. I'll fix it

src/query/scan_manager.c
shparkcubrid: okay. I'll fix it

src/query/query_hash_scan.h
```diff
@@ -41,6 +41,16 @@
     } \
   while (0)
 
+/* kind of hash list scan method */
+enum hash_method
+{
+  NOT_USE = 0,
+  IN_MEMORY = 1,
+  HYBRID_IN_MEMORY = 2,
+  HASH_TEMP_FILE = 3 /* not used */
```
hgryoo, line 50: To recognize the enum value is related to the hash method, how about adding a prefix for example HASH_METH_IN_MEM?

shparkcubrid: okay.

src/base/memory_hash.c
```diff
@@ -1194,6 +1350,63 @@ mht_dump (THREAD_ENTRY * thread_p, FILE * out_fp, const MHT_TABLE * ht, const in
...
+  if (out_fp == NULL)
+    {
+      out_fp = stdout;
+    }
+
+  fprintf (out_fp,
+	   "HTABLE NAME = %s, SIZE = %d,\n" "NENTRIES = %d, NPREALLOC = %d, NCOLLISIONS = %d\n\n",
+	   ht->name, ht->size, ht->nentries, ht->nprealloc_entries, ht->ncollisions);
+
+  if (print_id_opt)
```
byungwook-kim, line 1389: "if (print_id_opt == -1)" in **mht_dump** is not necessary?

src/base/memory_hash.c
```diff
@@ -1323,6 +1536,44 @@ mht_get2 (const MHT_TABLE * ht, const void *key, void **last)
...
+ *
+ * NOTE: This call does not affect the LRU list.
+ */
+void *
+mht_get_hls (const MHT_HLS_TABLE * ht, const void *key, void **last)
+{
+  unsigned int hash;
+  HENTRY_HLS_PTR hentry;
+
+  assert (ht != NULL && key != NULL);
```
byungwook-kim, line 1555: is not necessary to check the **last** variable ?

src/query/query_hash_scan.c
```diff
@@ -244,13 +243,12 @@ qdata_build_hscan_key (THREAD_ENTRY * thread_p, val_descr * vd, REGU_VARIABLE_LI
  *   args(in)   :
  */
 int
-qdata_print_hash_scan_entry (THREAD_ENTRY * thread_p, FILE * fp, const void *key, void *data, void *args)
+qdata_print_hash_scan_entry (THREAD_ENTRY * thread_p, FILE * fp, const void *data, void *args)
 {
   HASH_SCAN_VALUE *data2 = (HASH_SCAN_VALUE *) data;
-  HASH_SCAN_KEY *key2 = (HASH_SCAN_KEY *) key;
-  QFILE_TUPLE_VALUE_TYPE_LIST *list = (QFILE_TUPLE_VALUE_TYPE_LIST *) args;
+  int hash_list_scan_yn = *((int *) args);
```
byungwook-kim, line 249: is not necessary to check if **args** is null ?

src/query/query_hash_scan.c
```diff
@@ -342,7 +342,56 @@ qdata_copy_hscan_key (cubthread::entry * thread_p, HASH_SCAN_KEY * key, REGU_VAR
...
+qdata_copy_hscan_key_without_alloc (cubthread::entry * thread_p, HASH_SCAN_KEY * key, REGU_VARIABLE_LIST probe_regu_list,
+				    HASH_SCAN_KEY * new_key)
+{
+  DB_TYPE vtype1, vtype2;
+  TP_DOMAIN_STATUS status = DOMAIN_COMPATIBLE;
+
+  if (new_key)
+    {
+      /* copy values */
+      new_key->val_count = key->val_count;
```
byungwook-kim, line 360: is not necessary to check if **key** is null ?

src/query/query_hash_scan.h
```diff
@@ -29,11 +29,43 @@
...
+    } \
+  while (0)
+
+/* kind of hash list scan method */
+enum hash_method
+{
+  HASH_METH_NOT_USE = 0,
+  HASH_METH_IN_MEM = 1,
+  HASH_METH_HYBRID = 2,
+  HASH_METH_HASH_FILE = 3 /* not used */
```
byungwook-kim, line 50: not used? or not implement?

src/query/scan_manager.c
```diff
@@ -8041,20 +8110,41 @@ check_hash_list_scan (LLIST_SCAN_ID * llsidp, int *val_cnt, int hash_list_scan_y
...
+      /* HENTRY_HLS = pointer(8bytes) * 3 = 24 bytes*/
+      /* SIMPLE_POS = pageid(4bytes) + voldid(2bytes) + padding(2bytes) + offset(4bytes) = 12 bytes*/
+      return HASH_METH_HYBRID;
+    }
+  else
+    {
+      return HASH_METH_NOT_USE;
+    }
+
+  return HASH_METH_NOT_USE;
```
byungwook-kim, line 8149: this code is not reached, no need "else", simply "return HASH_METH_NOT_USE"

shparkcubrid:
Test case error
sql/_13_issues/_20_2h/cases/cbrd_23665.sql-linux_sql_64bit_release

In the comparison between varchar and char, data was not searched, but in this patch data is searched. 
It is normal for data to be searched, and found that there is a problem when coerce varchar type in char.

reproduction
```
drop table if exists tmp2;
create table tmp2 (col1 varchar(5), col2 char(5));
insert into tmp2 values('12','12');
select * from tmp2 where col1 = col2;
select * from tmp2 where col1 = cast( col2 as varchar(5));
```
The results of the two queries are different.
In the hash list scan, this problem does not occur by deleting the key comparison.
I think it could be a potential problem if there is a case that implicitly coerce a char type to varchar in another case.

src/base/memory_hash.c
shparkcubrid: The act related routine has been removed, so there is no need for it.

src/base/memory_hash.c
shparkcubrid: okay. i'll fix it

src/query/query_hash_scan.c
shparkcubrid: I'll fix it.

src/query/query_hash_scan.c
shparkcubrid: I'll fix it

src/query/query_hash_scan.h
shparkcubrid: 'not implemented yet'. However, since no one knows the future, I just keep it 'not used'.

src/query/scan_manager.c
shparkcubrid: I wanted to make the if-else more clear and put the return 'NOT_USE' at the end as defense code. If it doesn't matter, I'll keep it.

byungwook-kim (approved):

hgryoo (approved):

PR #2542, opened 2021-01-08
[CBRD-23812] [Regression] The results of two queries that scan different indexes in the same transaction are different.

byungwook-kim:
http://jira.cubrid.org/browse/CBRD-23812

Environment:

Create a table by increasing columns from 2 to 30.
Insert/update/select sequentially with 10 threads targeting each table.
10,000 times per thread.
Fail case:
When comparing each column value by executing the same select query that scans different indexes in the same transaction, in the select result comparison step, the result of the first query is failing with 0 tuple.

I changed the default type to VARCHAR for constant string.
And I also modified comparing routine mr_cmpval_string and mr_cmpval_char.
In the modified comparing routine, varchar-type data comparing always no ignore trailing spaces and char-type data comparing always ignore trailing spaces.

NVL and CONCAT function is fixed for return to origin column's type

### DIFF
--- a/src/compat/db_vdb.c
+++ b/src/compat/db_vdb.c
@@ -2566,7 +2566,7 @@ do_cast_host_variables_to_expected_domain (DB_SESSION * session)
 
       if (TP_IS_CHAR_TYPE (hv_dom->type->id))
         {
-          if (typ == DB_TYPE_VARCHAR || typ == DB_TYPE_VARNCHAR)
+          if (hv_dom->type->id != typ && (typ == DB_TYPE_VARCHAR || typ == DB_TYPE_VARNCHAR))
             {
               db_value_domain_init (hv, typ, prec, 0);
             }
--- a/src/compat/db_vdb.c
+++ b/src/compat/db_vdb.c
@@ -2541,7 +2541,12 @@ do_cast_host_variables_to_expected_domain (DB_SESSION * session)
 
   for (i = 0; i < hv_count; i++)
     {
+      int prec;
+      DB_TYPE typ;
+
       hv = &host_vars[i];
+      typ = db_value_type (hv);
+      prec = db_value_precision (hv);
       hv_dom = expected_domains[i];
       if (TP_DOMAIN_TYPE (hv_dom) == DB_TYPE_UNKNOWN || hv_dom->type->id == DB_TYPE_ENUMERATION)
 	{
@@ -2557,6 +2562,14 @@ do_cast_host_variables_to_expected_domain (DB_SESSION * session)
 	  pt_report_to_ersys (session->parser, PT_EXECUTION);
 	  pt_reset_error (session->parser);
 	  return ER_PT_EXECUTE;
+	}
+
+      if (TP_IS_CHAR_TYPE (hv_dom->type->id))
+	{
+	  if (typ == DB_TYPE_VARCHAR || typ == DB_TYPE_VARNCHAR)
+	    {
+	      db_value_domain_init (hv, typ, prec, 0);
+	    }
 	}
     }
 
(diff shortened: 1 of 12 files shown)


src/parser/csql_grammar.y
```diff
@@ -27409,10 +27411,31 @@ pt_create_char_string_literal (PARSER_CONTEXT *parser, const PT_TYPE_ENUM char_t
...
+                if (char_type == PT_TYPE_CHAR)
+	          {
+		    node->type_enum = PT_TYPE_VARCHAR;
+	          }
+	        else if (char_type == PT_TYPE_NCHAR)
+	          {
+		    node->type_enum = PT_TYPE_VARNCHAR;
+	          }
+              }
+	  }
```
hgryoo, line 27437: indent

hgryoo (changes requested):
For host variables, it is necessary to consider the type handling of constant values modified in this PR.

src/parser/csql_grammar.y
```diff
@@ -18708,7 +18708,7 @@ predicate_expr_sub
 			PT_NODE *v, *lhs, *rhs, *subq;
 			bool found_match = false;
 			bool found_paren_set_expr = false;
-
+
```
shparkcubrid, line 18711: indent

src/parser/csql_grammar.y
```diff
@@ -23381,10 +23382,10 @@ char_string
 			    force = true;
 			  }
 
-			node = pt_create_char_string_literal (this_parser,
+		   	node = pt_create_char_string_literal (this_parser,
 							      PT_TYPE_CHAR,
 							      $1, charset);
-
+
```
shparkcubrid, line 23388: indent

src/parser/csql_grammar.y
byungwook-kim: I fixed it.

src/parser/csql_grammar.y
byungwook-kim: I fixed it.

src/parser/csql_grammar.y
byungwook-kim: I fixed it.

src/compat/db_vdb.c
```diff
@@ -2558,6 +2563,14 @@ do_cast_host_variables_to_expected_domain (DB_SESSION * session)
...
 	}
+
+      if (TP_IS_CHAR_TYPE (hv_dom->type->id))
+        {
+          if (hv_dom->type->id != typ && (typ == DB_TYPE_VARCHAR || typ == DB_TYPE_VARNCHAR))
+            {
+              db_value_domain_init (hv, typ, prec, 0);
+            }
+        }
     }
```
hgryoo, line 2574: indent

src/object/object_primitive.c
```diff
@@ -11139,6 +11144,32 @@ mr_cmpval_string (DB_VALUE * value1, DB_VALUE * value2, int do_coercion, int tot
...
+
+      if (type1 == DB_TYPE_CHAR || type1 == DB_TYPE_NCHAR)
+        {
+          for (i = size1; i > 1; i--)
+            {
+              if (string1[i - 1] != 0x20)
+                {
+                  break;
+                }
+            }
```
hgryoo, line 11157: indent

src/object/object_primitive.c
```diff
@@ -11139,6 +11144,32 @@ mr_cmpval_string (DB_VALUE * value1, DB_VALUE * value2, int do_coercion, int tot
...
+        {
+          for (i = size2; i > 1; i--)
+            {
+              if (string2[i - 1] != 0x20)
+                {
+                  break;
+                }
+            }
+          size2 = i;
+        }
```
hgryoo, line 11171: indent

shparkcubrid (approved):

hgryoo (approved):

PR #2549, opened 2021-01-14
[CBRD-23831] Some collection-related (has VARCHAR elements) query returns incorrect result

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-23831

A codeset of collection type is not considered when implicit cast.
I add code set which is bigger among set's elements.
I also fix wrong inequality sign to avoid potential problems.

### DIFF
--- a/src/parser/type_checking.c
+++ b/src/parser/type_checking.c
@@ -5446,6 +5446,8 @@ pt_coerce_range_expr_arguments (PARSER_CONTEXT * parser, PT_NODE * expr, PT_NODE
 	{
 	  PT_NODE *temp = NULL;
 	  int precision = 0, scale = 0;
+	  int units = data_type->info.data_type.units; /* code set */
+	  int collation_id = data_type->info.data_type.collation_id; /* collation_id */
 	  bool keep_searching = true;
 	  for (temp = arg2->data_type; temp != NULL && keep_searching; temp = temp->next)
 	    {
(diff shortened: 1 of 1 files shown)


src/parser/type_checking.c
```diff
@@ -5446,6 +5446,8 @@ pt_coerce_range_expr_arguments (PARSER_CONTEXT * parser, PT_NODE * expr, PT_NODE
 	{
 	  PT_NODE *temp = NULL;
 	  int precision = 0, scale = 0;
+	  int units = data_type->info.data_type.units; /* code set */
+	  int collation_id = data_type->info.data_type.collation_id; /* collation_id */
```
hgryoo, line 5450: `data_type` was newly created from line 5437. So these values may be initialized with a garbage value.
```suggestion
	  int units = LANG_SYS_CODESET; /* code set */
	  int collation_id = LANG_SYS_COLLATION; /* collation_id */
```

shparkcubrid: I agree.

src/parser/type_checking.c
```diff
@@ -5526,9 +5528,21 @@ pt_coerce_range_expr_arguments (PARSER_CONTEXT * parser, PT_NODE * expr, PT_NODE
...
 		}
+
+	      if (PT_IS_STRING_TYPE (common_type) && PT_IS_STRING_TYPE (temp->type_enum))
+		{
+		  /* A bigger codesets's number can represent more characters. */
+		  if (units < temp->info.data_type.units)
+		    {
+		      units = temp->info.data_type.units;
+		      collation_id = temp->info.data_type.collation_id;
+		    }
```
byungwook-kim, line 5539: In the case of mixed code set, which of them should be selected ?
Currently ISO88591's code set = 3, EUCKR = 4, UTF8 = 5, therefore the selection policy you thought seems good.
The selection by code set value is dangerous a little when the code set value is changed. But it's the best in current.

byungwook-kim (approved):

hgryoo (approved):

shparkcubrid:
It is necessary to check later if it can be improved by using functions such as pt_common_collation() or pt_make_cast_with_compatble_info(). This will clarify the common codeset and collation.

PR #2558, opened 2021-02-04
[LETS-5] Add server_type system parameter

radu-vtv:
http://jira.cubrid.org/browse/LETS-5

Add a system parameter for configuring whether the server starts as a page- or a transaction-server (default when parameter is missing in `cubrid.conf`).

I had to include <algorithm> in `src/loaddb/load_error_handler.cpp`, otherwise it didn't build and also fixed some `unused variable` warnings in `boot_sr.c`

### DIFF
--- a/src/base/server_type.cpp
+++ b/src/base/server_type.cpp
@@ -20,15 +20,15 @@
 
 #include "system_parameter.h"
 
-static SERVER_TYPE server_type;
+static SERVER_TYPE g_server_type;
 
 void init_server_type ()
 {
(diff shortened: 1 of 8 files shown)


src/base/server_type.cpp
```diff
@@ -0,0 +1,34 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+
+#include "server_type.hpp"
+
+#include "system_parameter.h"
+
+static SERVER_TYPE server_type;
```
ribram, line 23: Rename to make global variable obvious, by adding 'g_' or 'global_' prefix or by naming as server_Type.

ribram (approved):

PR #2559, opened 2021-02-05
[CBRD-23856] Adding mutex for result-cache to process multi-threaded transaction

byungwook-kim:
http://jira.cubrid.org/browse/CBRD-23856

The result-cache should work on multi-threaded transaction contenting environment, however the xasl cache and list file cache (result-cache entry) is not applied to mutex mechanism.
We should apply the mutex for xasl cache and list file cache to handle properly on the heavy contenting transactions.

**result-cache spec. modified as below:**

- if there were modified classes, not use the result cache until committed.

```
select /*+ query_cache */ * from cache_t;
update cache_t set col2 = 99 where col1 = 3;
--stopping lookup result-cache for modified classes
select /*+ query_cache */ * from cache_t where col1 = 3;
-- result-cache is not used and the result is also not added to the cache.
commit;
-- starting lookup result-cache
select /*+ query_cache */ * from cache_t where col1 = 3;
-- result-cache is added to the other transaction use.
```

- transaction isolation: REPETABLE READ might not be available.
- remove-marked result-cache (by modified) is removed at lookup time.
- list_cache_mutex is added for thread-safe.
- pool_mutex for list cache entry alloc. and free is added for thread-safe.
- qfile_list_cache_Cleanup_flag is added for list cache clean up

result-cache final spec.

- transaction isolation: REPETABLE READ might not be available.
- remove-marked result-cache (by modified) is removed at lookup time.
- list_cache_mutex is removed.
     because critical section CSECT_QPROC_LIST_CACHE already roles a mutex.
     `csect_enter (thread_p, CSECT_QPROC_LIST_CACHE, INF_WAIT) /* it's a kind of a mutex. */`
- pool_mutex for list cache entry alloc. and free is added for thread-safe.
     it's removed as the same as the reason of the list_cache_mutex
- qfile_list_cache_Cleanup_flag is not needed for list cache clean up
    the cleanup processing is already protected by CSECT_QPROC_LIST_CACHE
- qfile_delete_list_cache_entry
    no need to check if the deletion_marker is set because this routine is called from list-cache-entry related routines, with only when the cache entry should be deleted at no transaction in use.


ctshim (approved):

src/query/list_file.c
```diff
@@ -5577,12 +5445,7 @@ qfile_delete_list_cache_entry (THREAD_ENTRY * thread_p, void *data)
       return ER_FAILED;
     }
 
-  lent->deletion_marker = true;
-#if defined(SERVER_MODE)
-  if (lent->deletion_marker && lent->last_ta_idx == 0)
-#else /* not SERVER_MODE */
   if (lent->deletion_marker)
```
shparkcubrid, line 5448: Can 'lent->last_ta_idx == 0' be omitted? Isn't there a transaction using the cache?

shparkcubrid:
The functions related to CSECT_QPROC_LIST_CACHE are as follows.
```
qfile_initialize_list_cache()
qfile_finalize_list_cache()
qfile_clear_list_cache()
qfile_dump_list_cache_internal()
qfile_clear_uncommited_list_cache_entry()
qfile_lookup_list_cache_entry()
qfile_update_list_cache_entry()
qfile_end_use_of_list_cache_entry()
```

src/query/list_file.c
byungwook-kim: because the transaction isolation level "REPEATABLE READ" is not need to keep.

byungwook-kim:
> The functions related to CSECT_QPROC_LIST_CACHE are as follows.
> 
> ```
> qfile_initialize_list_cache()
> qfile_finalize_list_cache()
> qfile_clear_list_cache()
> qfile_dump_list_cache_internal()
> qfile_clear_uncommited_list_cache_entry()
> qfile_lookup_list_cache_entry()
> qfile_update_list_cache_entry()
> qfile_end_use_of_list_cache_entry()
> ```

the result-cache mutex seems will not be needed if the critical section is enough to control accessing the resources.

src/query/list_file.c
```diff
@@ -154,20 +151,21 @@ struct qfile_list_cache_entry_pool
   QFILE_POOLED_LIST_CACHE_ENTRY *pool;	/* pre-allocated array */
   int n_entries;		/* number of entries in the pool */
   int free_list;		/* the head(first entry) of the free list */
+  pthread_mutex_t pool_mutex;	/* for thread safe */
```
hgryoo, line 154: This mutex seems to manage the qfile_List_cache_entry_pool
- get an entry from the entry pool
- put newly allocated entry to entry pool

We seem have a lock-free concept data structures in base/lockfree::*
lockfree::circular_queue could be used to avoid the cost of mutex locking.

src/query/list_file.c
```diff
@@ -5229,8 +5176,13 @@ qfile_allocate_list_cache_entry (int req_size)
 
       (void) memset ((void *) &pent->s.entry, 0, sizeof (QFILE_LIST_CACHE_ENTRY));
     }
+  else
+    {
+      return NULL;
```
hgryoo, line 5181: The following pthread_mutex_init will not be called

hgryoo:
To check code style and static analysis by cppcheck, please merge with develop in upstream
```
git fetch upstream
git merge upstream/develop
```

src/query/list_file.c
```diff
@@ -5686,64 +5554,46 @@ qfile_lookup_list_cache_entry (THREAD_ENTRY * thread_p, int list_ht_no, const DB
...
   if (lent)
     {
+      unsigned int i;
+
       /* check if it is marked to be deleted */
-      if (lent->deletion_marker)
-	{
-	  (void) qfile_delete_list_cache_entry (thread_p, lent);
-	  lent = NULL;
-	}
```
hgryoo, line 5696: removing the comment is missed

src/query/list_file.c
byungwook-kim: I found the mutex **pool_mutex** is not needed to control the pool because the rountine **qfile_allocate_list_cache_entry** and qfile_free_list_cache_entry is called with critical section **CSECT_QPROC_LIST_CACHE** from the routines related to cache entry managing. So, I will remove the mutex.

src/query/list_file.c
byungwook-kim: this code will be fixed not to the mutex.

src/query/list_file.c
byungwook-kim: Ok

src/query/list_file.c
```diff
@@ -5686,64 +5542,44 @@ qfile_lookup_list_cache_entry (THREAD_ENTRY * thread_p, int list_ht_no, const DB
   /* look up the hash table with the key */
   lent = (QFILE_LIST_CACHE_ENTRY *) mht_get (qfile_List_cache.list_hts[list_ht_no], params);
   qfile_List_cache.lookup_counter++;	/* counter */
+
   if (lent)
     {
+      unsigned int i;
+
+#if defined(SERVER_MODE)
```
shparkcubrid, line 5550: The cache is deleted only for server mode?

src/query/list_file.c
```diff
@@ -5686,64 +5542,44 @@ qfile_lookup_list_cache_entry (THREAD_ENTRY * thread_p, int list_ht_no, const DB
...
+#endif
 
+#if defined(SERVER_MODE)
       /* finally, we found an useful cache entry to reuse */
-      if (lent)
+      if (new_tran)
 	{
 	  /* record my transaction id into the entry and adjust timestamp and reference counter */
-#if defined(SERVER_MODE)
+	  lent->uncommitted_marker = true;
```
shparkcubrid, line 5576: The logic related to uncommitted_marker has been removed, but it is still used.

byungwook-kim: uncommitted_marker is used at printing the status of result-cache.
for example, at csql
```
csql> ;info qcache
....
uncommitted_marker = true
...
```

src/query/list_file.c
byungwook-kim: Ok, I will fix it.

ctshim (approved):

src/query/list_file.c
shparkcubrid: I recommend deleting the uncommitted_marker. uncommitted_marker is meaningless.

byungwook-kim: I removed it from result-cache dump (;info qcache)

shparkcubrid (approved):

youngjinj (approved):

src/query/xasl_cache.c
```diff
@@ -1768,6 +1772,8 @@ xcache_invalidate_entries (THREAD_ENTRY * thread_p, bool (*invalidate_check) (XA
        */
       while (true)
 	{
+	  int num_entries;
```
hgryoo, line 1775: this variable seems not used

src/query/xasl_cache.c
```diff
@@ -399,13 +399,18 @@ xasl_cache_ent::init_clone_cache ()
 static void *
 xcache_entry_alloc (void)
 {
+  static int xcache_nentries = 0;
```
hgryoo, line 402: I'm not sure it is safe

hgryoo (approved):

src/query/xasl_cache.c
byungwook-kim: it's called from xcache initialize routine at boot time.
so, the data is safe.

src/query/xasl_cache.c
byungwook-kim: Ok, I fixed.

PR #2562, opened 2021-02-05
[LETS-15] page_server_hosts system parameter

radu-vtv:
http://jira.cubrid.org/browse/LETS-15

Add a system parameter for configuring the page server host(s) (for now only one page server, to be extended at a later time to more).

This only has effect if the server is started as `transaction`-server.


src/base/server_type.cpp
```diff
@@ -18,17 +18,63 @@
...
+    {
+      er_set (ER_WARNING_SEVERITY, ARG_FILE_LINE, ER_HOST_PORT_PARAMETER, 2, prm_get_name (PRM_ID_PAGE_SERVER_HOSTS),
+	      hosts.c_str ());
+      return;
+    }
+
+  long port = -1;
+  try
+    {
+      port = std::stol (hosts.substr (col_pos+1));
```
hornetmj, line 64: Is a string like 3000a treated as 3000?

hornetmj (approved):

src/base/server_type.cpp
radu-vtv: Yes, trailing letters are ignored by `std::stol`. I can't imagine how this could lead to errors tough. Maybe someone typing `300O` instead of `3000`...

Of course, I can parse the string if you think it's important.

ribram:
> Is a string like 3000a treated as 3000?
(@ribram Can I ask you questions anytime? I'm participating as a viewer, so if it bothers you, I'll ask you through another route.)

Yes, please, ask questions and add reviews at any time, we will answer and fix.

ribram (approved):

PR #2569, opened 2021-02-15
[CBRD-23859] ROUND()'s result of decimal data type is wrong

byungwook-kim:
http://jira.cubrid.org/browse/CBRD-23859

In the case of a division operation in the process of calculating a double type in the linux gcc environment, 0 may change to 1 at the 16th digit below the decimal point.
If we change this operation to a multiplication operation, it is calculated correctly.


ctshim (approved):
Is there any reason not to use the round() function?

mhoh3963 (approved):

PR #2571, opened 2021-02-16
[CBRD-23854] When using /*+ USE_SBR */, an error occurs on the slave node.

byungwook-kim:
http://jira.cubrid.org/browse/CBRD-23854

When using /*+ USE_SBR */, an error occurs on the slave node.
We can use sql_user_text of parser to replicate query for SBR.

The host variables are replaced the values using string parsing in sql_user_text and host variables.

### DIFF
--- a/src/query/execute_statement.c
+++ b/src/query/execute_statement.c
@@ -14634,9 +14634,11 @@ do_replicate_statement (PARSER_CONTEXT * parser, PT_NODE * statement)
   REPL_INFO repl_info;
   REPL_INFO_SBR repl_stmt;
   PARSER_VARCHAR *name = NULL;
+  PARSER_VARCHAR **host_val = NULL;
   static const char *unknown_name = "-";
   char stmt_separator;
   char *stmt_end = NULL;
+  char *sbr_text = NULL;
 
   if (log_does_allow_replication () == false)
     {
@@ -14807,8 +14809,7 @@ do_replicate_statement (PARSER_CONTEXT * parser, PT_NODE * statement)
     }
   else
     {
-      PARSER_VARCHAR **host_val;
-      char *sbr_text, *sql_text;
+      char *sql_text;
       int i, n, nth;
       int sql_len = strlen (parser->context.sql_user_text);
       int var_len = 0;
@@ -14829,8 +14830,8 @@ do_replicate_statement (PARSER_CONTEXT * parser, PT_NODE * statement)
       sbr_text = (char *) malloc (sql_len + var_len);
       if (sbr_text == NULL)
 	{
-	  free (host_val);
-	  return ER_OUT_OF_VIRTUAL_MEMORY;
+	  error = ER_OUT_OF_VIRTUAL_MEMORY;
+	  goto end;
 	}
 
       n = nth = 0;
@@ -14859,9 +14860,8 @@ do_replicate_statement (PARSER_CONTEXT * parser, PT_NODE * statement)
 		}
 	      else
 		{
-		  free (host_val);
-		  free (sbr_text);
-		  return ER_IT_UNKNOWN_VARIABLE;
+		  error = ER_IT_UNKNOWN_VARIABLE;
+		  goto end;
 		}
 	    }
 	  else
@@ -14872,8 +14872,6 @@ do_replicate_statement (PARSER_CONTEXT * parser, PT_NODE * statement)
 
       sbr_text[n] = 0;
       repl_stmt.stmt_text = sbr_text;
-
-      free (host_val);
     }
 
   repl_stmt.db_user = db_get_user_name ();
@@ -14899,6 +14897,18 @@ do_replicate_statement (PARSER_CONTEXT * parser, PT_NODE * statement)
     }
 
   db_string_free (repl_stmt.db_user);
+
+end:
+  if (sbr_text)
+    {
+      free (sbr_text);
+    }
+
+  if (host_val)
+    {
+      free (host_val);
+    }
+
   if (repl_stmt.sys_prm_context)
     {
       free (repl_stmt.sys_prm_context);
--- a/src/query/execute_statement.c
+++ b/src/query/execute_statement.c
@@ -14809,9 +14809,15 @@ do_replicate_statement (PARSER_CONTEXT * parser, PT_NODE * statement)
     }
   else
     {
-      char *sql_text;
+      /*
+       * if the query string includes the host variables, while processing the variable holder '?'
+       * the values of the host variables can be replaced into the user's original query string
+       * the pt_print_db_value(...) returns the value string and its length.
+       * the length includes quotes in case of the char string.
+       */
+      char *sql_text = statement->sql_user_text;
+      int sql_len = statement->sql_user_text_len;
       int i, n, nth;
-      int sql_len = strlen (parser->context.sql_user_text);
       int var_len = 0;
       bool begin_quote = false;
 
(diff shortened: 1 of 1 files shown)


src/query/execute_statement.c
```diff
@@ -14809,12 +14809,67 @@ do_replicate_statement (PARSER_CONTEXT * parser, PT_NODE * statement)
...
+      for (i = 0; i < parser->host_var_count; i++)
+	{
+	  host_val[i] = pt_print_db_value (parser, &parser->host_variables[i]);
+	  var_len += host_val[i]->length;
+	}
+
+      sbr_text = (char *) malloc (sql_len + var_len);
+      if (sbr_text == NULL)
+	{
+	  return ER_OUT_OF_VIRTUAL_MEMORY;
```
mhoh3963, line 14834: if sbr_text is NULL, do you need free(host_val) ?

src/query/execute_statement.c
```diff
@@ -14809,12 +14809,67 @@ do_replicate_statement (PARSER_CONTEXT * parser, PT_NODE * statement)
...
+	}
+
+      for (i = 0; i < parser->host_var_count; i++)
+	{
+	  host_val[i] = pt_print_db_value (parser, &parser->host_variables[i]);
+	  var_len += host_val[i]->length;
+	}
+
+      sbr_text = (char *) malloc (sql_len + var_len);
```
mhoh3963, line 14831: When allocing memory, I think, you will change memory size to sql_len+var_len+1 for adding a null character into the last character.

src/query/execute_statement.c
```diff
@@ -14809,12 +14809,67 @@ do_replicate_statement (PARSER_CONTEXT * parser, PT_NODE * statement)
...
+	    {
+	      strncpy (&sbr_text[n], (char *) host_val[nth]->bytes, host_val[nth]->length);
+	      n += host_val[nth++]->length;
+	    }
+	  else
+	    {
+	      sbr_text[n++] = sql_text[i];
+	    }
+	}
+
```
mhoh3963, line 14866: I think, you will add a code for checking error as the following.

if (parser->host_var_count != nth) error....

mhoh3963 (commented):
If there is host variable, please check to be able to remove old code about parser.

src/query/execute_statement.c
byungwook-kim: '?' roles 1 byte memory, so the sql_len + var_len is enough,

src/query/execute_statement.c
byungwook-kim: Ok, I will fix it.

src/query/execute_statement.c
byungwook-kim: Ok, I will fix it.

src/query/execute_statement.c
```diff
@@ -14809,12 +14809,67 @@ do_replicate_statement (PARSER_CONTEXT * parser, PT_NODE * statement)
     {
       PT_PRINT_VALUE_FUNC saved_func = parser->print_db_value;
       int saved_custom_print = parser->custom_print;
+      PARSER_VARCHAR **host_val;
```
hornetmj, line 14812: How about introducing a new function that performs this processing?

src/query/execute_statement.c
```diff
@@ -14809,12 +14809,67 @@ do_replicate_statement (PARSER_CONTEXT * parser, PT_NODE * statement)
     {
```
hornetmj, line 14809: If you don't need codes that already exist in else part, why don't you remove them?

src/query/execute_statement.c
```diff
@@ -14809,12 +14809,67 @@ do_replicate_statement (PARSER_CONTEXT * parser, PT_NODE * statement)
...
+	    {
+	      if (!begin_quote)
+		{
+		  begin_quote = true;
+		}
+	      else
+		{
+		  begin_quote = false;
+		}
+	    }
```
hornetmj, line 14855: I think it would be better to add '\n' under the '}'.

src/query/execute_statement.c
```diff
@@ -14809,12 +14809,67 @@ do_replicate_statement (PARSER_CONTEXT * parser, PT_NODE * statement)
...
+	      sbr_text[n++] = sql_text[i];
+	    }
+	}
+
+      sbr_text[n] = 0;
+      repl_stmt.stmt_text = sbr_text;
       parser->print_db_value = saved_func;
       parser->custom_print = saved_custom_print;
+
+      free (host_val);
```
hornetmj, line 14872: Where is the sbr_text free?

byungwook-kim: sbr_text should be free after the SBR's query string is used, so it must not be free.

src/query/execute_statement.c
byungwook-kim: I don't understand what you say.

src/query/execute_statement.c
byungwook-kim: I removed the unused codes.

src/query/execute_statement.c
hornetmj: Do you mean the sbr_text is deallocation somewhere? I can't find where it is deallocation.

src/query/execute_statement.c
hornetmj: Oh! sorry about this comment. I saw it wrong. There is already '\n' character between 14851 line and 14853 line.

hornetmj: Oh, it's not wrong. It has been modified from the db01c5b commit that you modified after this comment. ^0^

src/query/execute_statement.c
mhoh3963: It seems that the sql_len and var_len values ​​do not contain values ​​for handling null strings in sbr_text.

src/query/execute_statement.c
byungwook-kim: I fixed to free all of allocated memories at the end of the function do_replicate_statement.

src/query/execute_statement.c
byungwook-kim: the value of host variable does not need to include a null byte because the value is not the end of the query string.
if the sql is "insert into sbr_t values (?, ?)" and host variables are 'abcd' and 'cdef' each other, the final query string is "insert into sbr_t values ('abcd', 'cdef')". the final query string's length is strlen(sql) + strlen('abcd') + strlen('cdef'). it's enough to include all of host values.

src/query/execute_statement.c
byungwook-kim: how about the below comments?
     /*
       * if the query string includes the host variables, while processing the variable holder '?'
       * the values of the host variables can be replaced into the user's original query string
       * the pt_print_db_value(...) returns the value string and its length.
       * the length includes quotes in case of the char string.
       */

hornetmj: I prefer making it a function, but I think the decision is up to you.

src/query/execute_statement.c
mhoh3963: OK, I got it.

mhoh3963 (approved):

src/query/execute_statement.c
```diff
@@ -14807,14 +14809,74 @@ do_replicate_statement (PARSER_CONTEXT * parser, PT_NODE * statement)
...
+	  goto end;
+	}
+
+      n = nth = 0;
+
+      for (i = 0; i < sql_len; i++)
+	{
+	  if (sql_text[i] == '\'')
+	    {
+	      if (!begin_quote)
```
hornetmj, line 14849: As far as I know, CUBRID uses '' instead of \' to represent quotation marks, so this shouldn't be a problem. I am just worried that there may be a problem related to the quotation marks in the string. So I leave a comment hoping you will confirm. (I don't find any bad case for this.)

hornetmj (approved):

src/query/execute_statement.c
byungwook-kim: I checked the case which you said as your comment but it was no problem. thanks.

PR #2575, opened 2021-02-18
[CBRD-23850] coredump occurs when heap_get_visible_version_from_log() function is called after executing backupdb with -r option.

hornetmj:
http://jira.cubrid.org/browse/CBRD-23850

When executing backupdb with the -r option, the archive log volume referenced by the transaction(mvcc) is not considered. So, coredump occurs when the transaction reads the old image of mvcc after performing the backupdb. There is a test scenario for this in the comments of the related issue. Looking at that scenario will help you understand. (I defined this problem as PROBLEM-1 in the jira issue.)

### DIFF
--- a/src/transaction/log_page_buffer.c
+++ b/src/transaction/log_page_buffer.c
@@ -6130,6 +6130,8 @@ logpb_remove_archive_logs (THREAD_ENTRY * thread_p, const char *info_reason)
   LOG_LSA newflush_upto_lsa;	/* Next to be flush */
   int first_deleted_arv_num;
   int last_deleted_arv_num;
+  int min_arv_required_for_vacuum;
+  LOG_PAGEID vacuum_first_pageid;
 
   assert (LOG_CS_OWN_WRITE_MODE (thread_p));
 
@@ -6177,6 +6179,14 @@ logpb_remove_archive_logs (THREAD_ENTRY * thread_p, const char *info_reason)
 
   last_deleted_arv_num--;
 
+  vacuum_first_pageid = vacuum_min_log_pageid_to_keep (thread_p);
+  if (logpb_is_page_in_archive (vacuum_first_pageid))
+    {
+      min_arv_required_for_vacuum = logpb_get_archive_number (thread_p, vacuum_first_pageid);
+      min_arv_required_for_vacuum--;
+      last_deleted_arv_num = MIN (last_deleted_arv_num, min_arv_required_for_vacuum);
+    }
+
   if (log_Gl.hdr.last_deleted_arv_num + 1 > last_deleted_arv_num)
     {
       /* Nothing to remove */
@@ -6190,6 +6200,9 @@ logpb_remove_archive_logs (THREAD_ENTRY * thread_p, const char *info_reason)
 
       log_Gl.hdr.last_deleted_arv_num = last_deleted_arv_num;
       logpb_flush_header (thread_p);	/* to get rid of archives */
+
+      _er_log_debug (ARG_FILE_LINE, "first_deleted_arv_num = %d, last_deleted_arv_num = %d", first_deleted_arv_num,
+		     last_deleted_arv_num);
     }
 }
 
--- a/src/transaction/log_page_buffer.c
+++ b/src/transaction/log_page_buffer.c
@@ -6114,12 +6114,10 @@ logpb_remove_archive_logs_exceed_limit (THREAD_ENTRY * thread_p, int max_count)
  *
  *   info_reason(in):
  *
- * NOTE: Archive that are not needed for system crashes are removed.
+ * NOTE: Archive that are not needed for system crashes and vacuum are removed.
  *       That these archives may be needed for media crash recovery.
  *       Therefore, it is important that the user copy these archives
  *       to tape. Check the log information file.
- *
- * TODO: Make sure the removed logs have been processed by vacuum.
  */
 void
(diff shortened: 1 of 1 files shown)


JoohoK (approved):

src/transaction/log_page_buffer.c
```diff
@@ -6177,6 +6179,14 @@ logpb_remove_archive_logs (THREAD_ENTRY * thread_p, const char *info_reason)
 
   last_deleted_arv_num--;
 
+  vacuum_first_pageid = vacuum_min_log_pageid_to_keep (thread_p);
```
wrlawodms, line 6182: Could you let me know why _vacuum_is_safe_to_remove_archives()_ is not needed here? It is different from _logpb_remove_archive_logs_exceed_limit()_ .

src/transaction/log_page_buffer.c
```diff
@@ -6177,6 +6179,14 @@ logpb_remove_archive_logs (THREAD_ENTRY * thread_p, const char *info_reason)
...
+  if (logpb_is_page_in_archive (vacuum_first_pageid))
```
wrlawodms, line 6183: I think ```vacuum_first_pageid != NULL_PAGEID``` should be checked before this condition.

wrlawodms (changes requested):

wrlawodms:
1. The comment of the function should be modified.
```
 * NOTE: Archive that are not needed for system crashes are removed.
 *       That these archives may be needed for media crash recovery.
 *       Therefore, it is important that the user copy these archives
 *       to tape. Check the log information file.
 *
 * TODO: Make sure the removed logs have been processed by vacuum.
```
2. Is there no need to consider the minimum copied pageid by copylogdb here?

src/transaction/log_page_buffer.c
hornetmj: You are right. I will fix it.

src/transaction/log_page_buffer.c
hornetmj: After vacuum_boot() is called, I think there is no problem with calling vacuum_min_log_pageid_to_keep(). However, for code consistency, I will call vacuum_is_safe_to_remove_archives() to make sure that the deletion of archive log volumes is safe.

hornetmj:
> 1. The comment for this function should be modified along.
> 
> ```
>  * NOTE: Archive that are not needed for system crashes are removed.
>  *       That these archives may be needed for media crash recovery.
>  *       Therefore, it is important that the user copy these archives
>  *       to tape. Check the log information file.
>  *
>  * TODO: Make sure the removed logs have been processed by vacuum.
> ```
> 
> 1. Could you let me know the reason there's no need to consider the minimum copied page id by copylogdb here?



> 1. The comment for this function should be modified along.
> 
> ```
>  * NOTE: Archive that are not needed for system crashes are removed.
>  *       That these archives may be needed for media crash recovery.
>  *       Therefore, it is important that the user copy these archives
>  *       to tape. Check the log information file.
>  *
>  * TODO: Make sure the removed logs have been processed by vacuum.
> ```
> 
> 1. Could you let me know the reason there's no need to consider the minimum copied page id by copylogdb here?

1. I'll fix the comment.
2. Your comments are not included in the purpose of this PR. If necessary, I will process it with another PR.

src/transaction/log_page_buffer.c
wrlawodms: 1. `logpb_remove_archive_logs_exceed_limit()` seems to be also called after vacuume_boot().
2. In my opinion, if it is missed in `logpb_remove_archive_logs()`. This function has to be always called when it is guaranteed that `vacuum_Data.is_archive_removal_safe` is set to true. It seems more stable to check it. Or you can `assert()` instead of my suggestion if the condition is pre-condition.

hornetmj: * The logpb_remove_archive_logs_exceed_limit() function is called from logpb_archive_active_log().
* So it can be called even before vacuum_boot().
* Therefore, I think vacuum_is_safe_to_remove_archives() must be called in order to operate after vacuum_boot().
* However, the logpb_remove_archive_logs() function does not seem to be a problem because it is a function that is called at runtime after server boot. What do you think?

mhoh3963: How about replacing all codes (line 6172 ~ 6193)  that remove archive log to logpb_remove_archive_logs_exceed_limit() function ?
I think the function is safe, because it have routine such as checking  "force remove archive" option, "vacuum min log" and "system crash min log" before removing archive logs.

hornetmj: @mhoh3963 Maybe it will work fine I think. However, running 'backupdb -r' will depend on the "log_max_archives" and "force_remove_log_archives" system properties. So, I think this is not a good choice. Nevertheless, if I need to replace it with the logpb_remove_archive_logs_exceed_limit() function, I think we need to reconsider the definition of the -r option. What do other reviewers(@JoohoK @wrlawodms) think?

hornetmj: I think vacuum_is_safe_to_remove_archives() is better than prm_get_bool_value (PRM_ID_DISABLE_VACUUM) for code consistency and readability. So I used it. Please refer to the review. In addition, the vacuum_disable value is being checked by logpb_is_page_in_archive(), so I don't think any issues we discussed will occur. However, I want to call the function vacuum_is_safe_to_remove_archives() for explicit processing.

mhoh3963 (approved):

wrlawodms (approved):

PR #2581, opened 2021-02-23
[LETS-28] fix buffer overflow in LOG_ZIP.log_data

cristiarg:
http://jira.cubrid.org/browse/LETS-28

- by default, LOG_ZIP.log_data is `malloc`ed with 16k
- when it needs to unzip larger than 16k buffers, `log_data` is allocated a large enough buffer on-the-fly
- extract a function `log_zip_realloc_if_needed` out of already existing code - with the single modification that `realloc` is used instead of `free`/`malloc`
- call this realloc in `log_rv_get_unzip_log_data` where the `LOG_ZIP` struct is _also_ used as a temporary buffer to pass data to the calling function

Remarks:
An alternative to not hijacking the `LOG_ZIP` data buffer for - what is, in essence - a different purpose that for what it is actually passed downstream from `log_recovery_redo`, is to have such a pointer on the `LOG_RCV` structure. It makes the most sense to have it on the `LOG_RCV` structure, because already that structure's purpose is to provide the mechanism by which to pass data into the undo/redo apply functions and thus, it is already guarantees that the `LOG_RCV` instances already have the corresponding lifetime. Similarly to what happens currently, when the data to be applied is small (i.e.: fits within one page), the log pointer will be passed directly towards the apply functions. Otherwise, a buffer will be allocated  on the `LOG_RCV` structure, will be used, and will be deallocated when `LOG_RCV` goes out of scope.

### DIFF
--- a/src/transaction/log_compress.c
+++ b/src/transaction/log_compress.c
@@ -62,7 +62,10 @@ log_zip (LOG_ZIP * log_zip, LOG_ZIP_SIZE_T length, const void *data)
 
   buf_size = LOG_ZIP_BUF_SIZE (length);
 
-  log_zip_realloc_if_needed (*log_zip, buf_size);
+  if (!log_zip_realloc_if_needed (*log_zip, buf_size))
+    {
+      return false;
+    }
 
 #if defined (SERVER_MODE) || defined (SA_MODE)
   PERF_UTIME_TRACKER_START (NULL, &time_track);
@@ -123,7 +126,10 @@ log_unzip (LOG_ZIP * log_unzip, LOG_ZIP_SIZE_T length, const void *data)
 
   length -= sizeof (LOG_ZIP_SIZE_T);
 
-  log_zip_realloc_if_needed (*log_unzip, buf_size);
+  if (!log_zip_realloc_if_needed (*log_unzip, buf_size))
+    {
+      return false;
+    }
 
 #if defined (SERVER_MODE) || defined (SA_MODE)
   PERF_UTIME_TRACKER_START (NULL, &time_track);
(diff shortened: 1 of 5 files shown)


src/transaction/log_compress.c
```diff
@@ -61,27 +61,8 @@ log_zip (LOG_ZIP * log_zip, LOG_ZIP_SIZE_T length, const void *data)
...
-      log_zip->buf_size = buf_size;
-    }
-
-  if (log_zip->log_data == NULL)
-    {
-      log_zip->data_length = 0;
-      log_zip->buf_size = 0;
-      return false;
-    }
+  log_zip_realloc_if_needed (*log_zip, buf_size);
```
ribram, line 65:
```suggestion
  if (!log_zip_realloc_if_needed (*log_zip, buf_size))
    {
      return false;
    }
```

src/transaction/log_compress.c
```diff
@@ -142,27 +123,7 @@ log_unzip (LOG_ZIP * log_unzip, LOG_ZIP_SIZE_T length, const void *data)
...
-      log_unzip->buf_size = buf_size;
-    }
-
-  if (log_unzip->log_data == NULL)
-    {
-      log_unzip->data_length = 0;
-      log_unzip->buf_size = 0;
-      return false;
-    }
+  log_zip_realloc_if_needed (*log_unzip, buf_size);
```
ribram, line 126:
```suggestion
  if (!log_zip_realloc_if_needed (*log_unzip, buf_size))
    {
      return false;
    }
```

ribram (approved):

src/transaction/log_compress.c
cristiarg: Done

src/transaction/log_compress.c
cristiarg: Done

PR #2583, opened 2021-02-26
[LETS-29] Serialize and deserialize log prior node lists 

ribram:
http://jira.cubrid.org/browse/LETS-29

Add two functions in the log_append files:

- prior_list_serialize: serialize entire log_prior_node list into an std::string
- prior_list_deseriaze: deserialize list from an std::string

**Implementation**

- Serialize

  - Calculate serialized size, reserve enough memory and serialize nodes one by one.

- Deserialize:

  - Allocate necessary nodes, data_header, udata and rdata using malloc, same as prior_lsa_alloc_and_copy_data does.
  - Read node by node and append to list, until all string is consumed

**Unit Test**

- Add serialize/deserialize to new test_log_transfer unit test project. The project will be shared with other log transfer tests.
- Fix all log_append dependencies.
- Test serialize and deserialize with different log_prior_node field values, with and without undo & redo data.
- Deserialized list has to match original list.

**Out of scope**
- Changes to test_request_client_server: fix license header and option name.


adrianbrab (approved):

src/transaction/log_append.cpp
```diff
@@ -1569,6 +1571,157 @@ prior_is_tde_encrypted (const log_prior_node *node)
...
+    {
+      total_size += prior_node_serialized_size (*nodep);
+    }
+  return total_size;
+}
+
+static void
+prior_node_serialize (const log_prior_node &node, std::string &serialized)
+{
+#define copy_mem(mem, memsz) if ((memsz) > 0) serialized.append (mem, memsz)
```
cristiarg, line 1609: Nitpicking, couldn't these macros be some [templated, if needed] inline functions?

src/transaction/log_append.cpp
```diff
@@ -1569,6 +1571,157 @@ prior_is_tde_encrypted (const log_prior_node *node)
...
+  copy_mem (node.rdata, node.rlength);
+
+#undef copy_mem
+#undef copy
+}
+
+static const char *
+prior_node_deserialize (const char *ptr, log_prior_node &node)
+{
+#define copy_mem(mem, memsz) std::memcpy (mem, ptr, memsz); ptr += (memsz)
```
cristiarg, line 1630: Nitpicking, couldn't these macros be some [templated, if needed] inlined functions?

src/transaction/log_append.cpp
```diff
@@ -1569,6 +1571,157 @@ prior_is_tde_encrypted (const log_prior_node *node)
...
+  serialized.reserve (size);
+
+  for (const log_prior_node *nodep = head; nodep != nullptr; nodep = nodep->next)
+    {
+      prior_node_serialize (*nodep, serialized);
+    }
+
+  assert (serialized.size () == size);
+
+  return std::move (serialized);
```
cristiarg, line 1685: I've seen this in many places.
I'm not so knowledgeable into this either but maybe we should investigate how to do it correctly - https://stackoverflow.com/questions/14856344/when-should-stdmove-be-used-on-a-function-return-value - https://duckduckgo.com/?q=c%2B%2B+%22return+std%3A%3Amove%22&t=canonical&ia=web

unit_tests/CMakeLists.txt
```diff
@@ -103,8 +104,13 @@ if (UNIT_TESTS OR UNIT_TEST_MONITOR)
   add_subdirectory(monitor)
 endif(UNIT_TESTS OR UNIT_TEST_MONITOR)
 
-if (UNIT_TESTS OR UNIT_TEST_SERVER_SERVER)
+if (UNIT_TESTS OR UNIT_TEST_REQUEST_CS)
```
cristiarg, line 107: Good. :)
I have this change locally for 2 weeks so I could compile ;)

unit_tests/log_transfer/test_main_prior_list_serialize.cpp
```diff
@@ -0,0 +1,489 @@
...
+  const char *leftp = memleft;
+  const char *ritep = memrite;
+  for ( ; leftp < memleft + memsz; ++leftp, ++ritep)
+    {
+      REQUIRE (*leftp == *ritep);
+    }
+}
+
+//
+// Definitions of CUBRID stuff that is not used, but is needed for compile
```
cristiarg, line 292: technically, these are needed by the linker :)

cristiarg (changes requested):

src/transaction/log_append.cpp
```diff
@@ -1569,6 +1571,157 @@ prior_is_tde_encrypted (const log_prior_node *node)
   return node->tde_encrypted;
 }
 
+static size_t
+prior_node_serialized_size (const log_prior_node &node)
+{
+  size_t total_size = 0;
+
+  // fixed data
+  total_size += sizeof (node.log_header)
```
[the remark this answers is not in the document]
ribram: done

src/transaction/log_append.cpp
```diff
@@ -1569,6 +1571,157 @@ prior_is_tde_encrypted (const log_prior_node *node)
...
+  copy (node.tde_encrypted);
+  copy (node.data_header_length);
+  copy (node.ulength);
+  copy (node.rlength);
+
+  copy_mem (node.data_header, node.data_header_length);
+  copy_mem (node.udata, node.ulength);
+  copy_mem (node.rdata, node.rlength);
+
+#undef copy_mem
```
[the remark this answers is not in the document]
ribram: Not sure it matters. But I reversed it regardless :)

src/transaction/log_append.cpp
ribram: You're right. I removed the std::move.

unit_tests/log_transfer/test_main_prior_list_serialize.cpp
ribram: right, comment changed

src/transaction/log_append.cpp
ribram: I considered it... But macros were preferred since there are very specific needs here (append mem/object to string) and they allowed the lazy me to use less arguments :)

I also considered using lambdas, but I am not sure the compiler will know to inline.

Do you think the inlined/template functions will improve readability?

cristiarg (approved):

src/transaction/log_append.cpp
```diff
@@ -1569,6 +1571,157 @@ prior_is_tde_encrypted (const log_prior_node *node)
   return node->tde_encrypted;
 }
 
+static size_t
+prior_node_serialized_size (const log_prior_node &node)
+{
+  size_t total_size = 0;
+
+  // fixed data
+  constexpr size_t total_size = sizeof (node.log_header)
```
cristiarg, line 1580: Compilation error here. `constexpr size_t total_size` needs another name.

src/transaction/log_append.cpp
cristiarg: Lambdas with this complexity would be inlined - so they say - but generic (ie: auto argument) lambdas are`c++14`.
Templates would add explicitness but also a lot of verbosity.
Sticking to the [not] good ol' macros...

cristiarg (approved):
compilation error in `log_append.cpp - prior_node_serialized_size`

src/transaction/log_append.cpp
ribram: fixed

PR #2585, opened 2021-03-03
[CBRD-23870] occurred core dump when getting the specific system parameter at csql

mhoh3963:
- add code to check if the value of system parameter is NULL

### DIFF
--- a/src/base/system_parameter.c
+++ b/src/base/system_parameter.c
@@ -8589,6 +8589,11 @@ sysprm_obtain_parameters (char *data, SYSPRM_ASSIGN_VALUE ** prm_values_ptr)
 	  error = PRM_ERR_UNKNOWN_PARAM;
 	  break;
 	}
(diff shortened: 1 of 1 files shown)


src/base/system_parameter.c
```diff
@@ -8589,6 +8589,11 @@ sysprm_obtain_parameters (char *data, SYSPRM_ASSIGN_VALUE ** prm_values_ptr)
 	  error = PRM_ERR_UNKNOWN_PARAM;
 	  break;
 	}
+      else if (prm->value == NULL)
+        {
+	  error = PRM_ERR_NO_VALUE;
```
byungwook-kim, line 8594: please make indentation.

mhoh3963: OK, I fixed.

byungwook-kim (approved):

PR #2586, opened 2021-03-03
[LETS-8] Disable writing (and log creation) on the page server

adrianbrab:
http://jira.cubrid.org/browse/LETS-8

Queries and log creation should be disabled on the page server. Other tools that operate changes on the database like addvoldb, compactdb, loaddb, optimizedb, synccoldb, are also disabled. New error messages will appear when trying to run changes on the page server, or when the server is appending a log record.

### DIFF
--- a/src/transaction/log_append.cpp
+++ b/src/transaction/log_append.cpp
@@ -27,6 +27,7 @@
 #include "perf_monitor.h"
 #include "thread_entry.hpp"
 #include "thread_manager.hpp"
+#include "server_type.hpp"
 #include "vacuum.h"
 
 static bool log_Zip_support = false;
@@ -37,6 +38,8 @@ static LOG_ZIP *log_zip_redo = NULL;
 static char *log_data_ptr = NULL;
 static int log_data_length = 0;
 #endif
+
+#define assertm(exp, msg) assert(((void)msg, exp))
 
 size_t
 LOG_PRIOR_LSA_LAST_APPEND_OFFSET ()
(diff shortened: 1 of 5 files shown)


src/transaction/log_append.cpp
```diff
@@ -1578,6 +1583,9 @@ prior_is_tde_encrypted (const log_prior_node *node)
 static void
 prior_lsa_start_append (THREAD_ENTRY *thread_p, LOG_PRIOR_NODE *node, LOG_TDES *tdes)
 {
+#ifndef CS_MODE
+  assertm(get_server_type() == SERVER_TYPE_TRANSACTION, "Log append can be executed only on transaction server");
+#endif
```
ribram, line 1588: - log_append is never used in CS_MODE
- an empty space is required between function/macro name and parenthesis.
```suggestion
  assertm (get_server_type() == SERVER_TYPE_TRANSACTION, "Log append can be executed only on transaction server");
```

src/transaction/log_page_buffer.c
```diff
@@ -6727,6 +6728,9 @@ logpb_checkpoint_topops (THREAD_ENTRY * thread_p, LOG_INFO_CHKPT_SYSOP * &chkpt_
 LOG_PAGEID
 logpb_checkpoint (THREAD_ENTRY * thread_p)
 {
+  if (get_server_type () == SERVER_TYPE_PAGE)
+    return NULL_LOG_PAGEID;
```
ribram, line 6732: - The preferred style is to make scopes obvious.
- Add a comment to explain this is temporary
```suggestion
  if (get_server_type () == SERVER_TYPE_PAGE)
    {
      // todo: reactivate once checkpoint is changed without log records
      return NULL_LOG_PAGEID;
    }
```

ribram (approved):

PR #2593, opened 2021-03-05
[CBRD-23874] The function IFNULL returns incorrect data type at the query with numeric host-variables

byungwook-kim:
http://jira.cubrid.org/browse/CBRD-23874

The following query expects 12.34568 but query returns 12.00000. The function NVL also returns 12.00000 as like IFNULL.

```
CREATE TABLE dec_tbl (a decimal(10,5));
PREPARE stmt from 'INSERT INTO dec_tbl VALUES (IFNULL(?,0))';
EXECUTE stmt USING 12.3456789;

SELECT a FROM dec_tbl;
```
**Expected result:**

```
  a
======================
  12.34568 
```
**Actual result:**

```
  a
======================
  12.00000 
```

### DIFF
--- a/src/parser/type_checking.c
+++ b/src/parser/type_checking.c
@@ -3273,7 +3273,7 @@ pt_get_expression_definition (const PT_OP_TYPE op, EXPRESSION_DEFINITION * def)
 
       def->overloads_count = num;
       break;
-
+#if 0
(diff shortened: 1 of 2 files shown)


byungwook-kim:
it cannot be a solution and it will be reopened in the future.

PR #2609, opened 2021-03-11
[LETS-19] Add active_tran_server/page_server and complete the log prior transfer

ribram:
http://jira.cubrid.org/browse/LETS-19

Complete the last pieces of transferring the prior list from the active transaction server to the page server. The active_tran_server and page_server classes have been created to oversee the connection between the active transaction server and the page server.

Added a new source folder - server - for server type related files.

**active_tran_server**

The active transaction server currently oversees the connection with one page server. Should be extended in the future.

  - The functions init_page_server_hosts and connect_to_page_server have been moved from from server_type.cpp to active_tran_server.
  - Added function to disconnect the page server.
  - Added a function to push request to the page server. The request is pushed into a request queue that gets sent by request_autosend.
  - ats_Gl global instance is used to access active_tran_server functionality

**page_server**

 The page server currently oversees the connection with the active transaction server. Should be extended in the future.

  - Connection management functions (set connection, disconnect, is connected)
  - A request_client_server is used to process the requests coming from active transaction server.
  - ps_Gl global instance is used to access page_server functionality.

**ats_ps_request**

Two enums with the requests in each direction: active tran server to page server and page server to active tran server. Currently only SEND_LOG_PRIOR_LIST request exists.

**Other**

  - Added final_server_type() to disconnect active transaction server and page server during shutdown.
  - Renamed CONNECT_TRANSACTION_SERVER as CONNECT_ACTIVE_TRAN_TO_PAGE_SERVER, because we can expect more connection types from transaction server.
  - Replaced css_process_new_transaction_server with ps_Gl.set_active_tran_server_connection
  - Added prior receiver to log_Gl to be used by page server. SERVER_MODE only (page server can never run in SA_MODE).
  - Update prior list serialize unit to log_Gl changes.


src/base/server_type.hpp
```diff
@@ -25,7 +25,8 @@ typedef enum
   SERVER_TYPE_PAGE,
 } SERVER_TYPE;
 
-void init_server_type (const char* db_name);
+void init_server_type (const char *db_name);
+void final_server_type ();
```
cristiarg, line 29: The name of this function. Maybe `finalize`? Or some other verb?

src/server/page_server.cpp
```diff
@@ -0,0 +1,85 @@
...
+					std::bind (&page_server::receive_log_prior_list, std::ref (*this),
+					    std::placeholders::_1));
+
+  m_ats_conn->start_thread ();
+}
+
+void
+page_server::disconnect_active_tran_server ()
+{
+  delete m_ats_conn;
```
cristiarg, line 61: test if already connected? or if not already disconnected? or assert otherwise?

src/server/page_server.hpp
```diff
@@ -0,0 +1,51 @@
...
+#include "ats_ps_request.hpp"
+#include "request_client_server.hpp"
+
+// forward declaration
+namespace cubpacking
+{
+  class unpacker;
+}
+
+class page_server
```
cristiarg, line 31: would be good to mark all classes that are not supposed to be derived from - or at least which are not supposed to be derived from for now - with `final`

src/transaction/log_impl.h
```diff
@@ -644,6 +647,9 @@ struct log_global
 
   // *INDENT-OFF*
   cublog::prior_sender m_prior_sender;
+#if defined (SERVER_MODE)
+  cublog::prior_recver m_prior_recver;
```
cristiarg, line 651: why is only `prior_recver` guarded by `SERVER_MODE` and `prior_sender` is not?

cristiarg (approved):

src/server/page_server.hpp
ribram: Not sure.
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c139-use-final-on-classes-sparingly

src/server/page_server.cpp
ribram: May be both. If ats si connected, then it is deleted. If it is not connected, then m_ats_conn is nullptr and delete has no effect.

src/base/server_type.hpp
ribram: Renamed as finalize.

src/server/page_server.cpp
cristiarg: I know it has no effect if already nullptr. Testing for nullptr before actual deletion clearly suggests intent. But it's just pedantic. :)

src/server/page_server.hpp
cristiarg: You have a point. I was not aware.

PR #2611, opened 2021-03-15
[CBRD-23877] To CREATE a table as SELECT query, the alias of attributes should be checked.

byungwook-kim:
http://jira.cubrid.org/browse/CBRD-23877

```
csql> CREATE TABLE nvl_tbl3 AS SELECT SUM(1),MAX(1),COUNT(*),NVL(1,'');
Execute OK. (0.006011 sec) Committed.

csql> ;sc nvl_tbl3
=== <Help: Schema of a Class> ===
<Class Name>nvl_tbl3<Attributes>sum(1) INTEGER
max(1) INTEGER
count(*) INTEGER
nvl(1, '') CHARACTER VARYING(1073741823)

csql> DESC nvl_tbl3;
Field Type Null Key Default Extra
===================================================================================
'sum(1)' 'INTEGER' 'YES' '' NULL ''
'max(1)' 'INTEGER' 'YES' '' NULL ''
'count(*)' 'INTEGER' 'YES' '' NULL ''
'nvl(1, '')' 'VARCHAR(1073741823)' 'YES' '' NULL '' 
```
**Expected**
`ERROR: Missing an expected attribute name for sum(1). `

Unlike the CREATE TABLE statement, the CREATE VIEW statement checks the alias of the column attributes. This issue can be solved by checking the alias of the column attribute in the CREATE TABLE statement and treating a non-aliased column.

However, aliases are required only in SELECT statements that include functions such as SUM, COUNT, and AVG, or that contain calculation expressions, so alias checks for all columns are not required.

### DIFF
--- a/src/parser/semantic_check.c
+++ b/src/parser/semantic_check.c
@@ -8512,14 +8512,12 @@ pt_check_create_entity (PARSER_CONTEXT * parser, PT_NODE * node)
 	{
 	  for (crt_attr = qspec_attr; crt_attr != NULL; crt_attr = crt_attr->next)
 	    {
-	      if (crt_attr->alias_print || crt_attr->node_type == PT_NAME || crt_attr->node_type == PT_DOT_)
+	      if (crt_attr->alias_print == NULL && crt_attr->node_type != PT_NAME && crt_attr->node_type != PT_DOT_)
 		{
-                  continue;
+	          PT_ERRORmf (parser, qry_specs, MSGCAT_SET_PARSER_SEMANTIC, MSGCAT_SEMANTIC_MISSING_ATTR_NAME,
+		          pt_short_print (parser, crt_attr));
+	          return;
 		}
-
-	      PT_ERRORmf (parser, qry_specs, MSGCAT_SET_PARSER_SEMANTIC, MSGCAT_SEMANTIC_MISSING_ATTR_NAME,
-		      pt_short_print (parser, crt_attr));
-	      return;
 	    }
 
 	  if (select->info.query.with != NULL)
--- a/src/parser/semantic_check.c
+++ b/src/parser/semantic_check.c
@@ -8514,9 +8514,9 @@ pt_check_create_entity (PARSER_CONTEXT * parser, PT_NODE * node)
 	    {
 	      if (crt_attr->alias_print == NULL && crt_attr->node_type != PT_NAME && crt_attr->node_type != PT_DOT_)
 		{
-	          PT_ERRORmf (parser, qry_specs, MSGCAT_SET_PARSER_SEMANTIC, MSGCAT_SEMANTIC_MISSING_ATTR_NAME,
-		          pt_short_print (parser, crt_attr));
-	          return;
+		  PT_ERRORmf (parser, qry_specs, MSGCAT_SET_PARSER_SEMANTIC, MSGCAT_SEMANTIC_MISSING_ATTR_NAME,
+			      pt_short_print (parser, crt_attr));
+		  return;
 		}
 	    }
 
--- a/src/parser/semantic_check.c
+++ b/src/parser/semantic_check.c
@@ -8497,7 +8497,10 @@ pt_check_create_entity (PARSER_CONTEXT * parser, PT_NODE * node)
   qry_specs = node->info.create_entity.as_query_list;
   if (node->info.create_entity.entity_type == PT_CLASS)
     {
-      PT_NODE *select = NULL;
+      PT_NODE *select = node->info.create_entity.create_select;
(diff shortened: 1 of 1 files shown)


ctshim (commented):
When the for loop first runs, if the if() statement on line 8517 is false, an error message is returned.
However, in the loop condition after that, if the if() statement is false, it seems that the error message will not be processed.

ctshim (approved):

youngjinj (approved):

src/parser/semantic_check.c
```diff
@@ -8497,17 +8497,36 @@ pt_check_create_entity (PARSER_CONTEXT * parser, PT_NODE * node)
...
+	  for (crt_attr = qspec_attr; crt_attr != NULL; crt_attr = crt_attr->next)
+	    {
+	      PT_NODE *s_attr = NULL;
+
+	      if (crt_attr->node_type == PT_NAME || crt_attr->alias_print)
+		{
+		  s_attr = crt_attr;
+		}
+
+	      if (s_attr == NULL)
```
shparkcubrid, line 8522: how about?
```	      
if (crt_attr->node_type != PT_NAME && crt_attr->alias_print  == NULL)
```

shparkcubrid:
test_sql error : sql/_19_apricot/_10_enum_type/cases/trac_343_01.sql - linux_sql_64bit_release
```
create table t2(e enum('a', 'b') default 'b') as select i % 2 + 1 from t1;
```
In this case, it should proceed normally.

src/parser/semantic_check.c
byungwook-kim: Ok, it seems that's better readability.

shparkcubrid: How about checking PT_EXPR? I think PT_VALUE is okay.
```
if (crt_attr->node_type == PT_EXPR && crt_attr->alias_print  == NULL)
```

shparkcubrid:
you should check sql/_13_issues/_10_1h/cases/bug_2887.sql - linux_sql_64bit_release
```
create table c as select a.x from a;
```

src/parser/semantic_check.c
byungwook-kim: PT_DOT_ is also OK.

byungwook-kim: I think it should be excluded PT_NAME and PT_DOT_ only.

shparkcubrid: OK. I agree only PT_NAME.

src/parser/semantic_check.c
```diff
@@ -8497,17 +8497,31 @@ pt_check_create_entity (PARSER_CONTEXT * parser, PT_NODE * node)
...
 	  PT_ERRORm (parser, qry_specs, MSGCAT_SET_PARSER_SEMANTIC, MSGCAT_SEMANTIC_CLASS_NO_QUERY_SPEC);
 	  return;
 	}
       /* user variables are not allowed in select list for CREATE ... AS SELECT ... */
-      select = node->info.create_entity.create_select;
       if (select != NULL)
 	{
+	  for (crt_attr = qspec_attr; crt_attr != NULL; crt_attr = crt_attr->next)
+	    {
+	      if (crt_attr->alias_print || crt_attr->node_type == PT_NAME || crt_attr->node_type == PT_DOT_)
```
shparkcubrid, line 8515: I think It is clearer to use the case where the error occurs.
```
if (crt_attr->alias_print == NULL && crt_attr->node_type != PT_NAME && crt_attr->node_type != PT_DOT_)
```

byungwook-kim: OK, I agree.

shparkcubrid (approved):

hgryoo (approved):

src/parser/semantic_check.c
```diff
@@ -8497,17 +8497,29 @@ pt_check_create_entity (PARSER_CONTEXT * parser, PT_NODE * node)
...
       /* user variables are not allowed in select list for CREATE ... AS SELECT ... */
-      select = node->info.create_entity.create_select;
       if (select != NULL)
 	{
+	  for (crt_attr = qspec_attr; crt_attr != NULL; crt_attr = crt_attr->next)
+	    {
+	      if (crt_attr->alias_print == NULL && crt_attr->node_type != PT_NAME && crt_attr->node_type != PT_DOT_)
+		{
+	          PT_ERRORmf (parser, qry_specs, MSGCAT_SET_PARSER_SEMANTIC, MSGCAT_SEMANTIC_MISSING_ATTR_NAME,
+		          pt_short_print (parser, crt_attr));
```
hgryoo, line 8518: please correct indentation

PR #2612, opened 2021-03-15
[CBRD-23878] To prevent setting the precision to -1 for STRING type as derived attribute

byungwook-kim:
http://jira.cubrid.org/browse/CBRD-23878

```
csql> CREATE TABLE nvl_tbl1 AS SELECT NVL(x,111) AS x, NVL(y,'aaa') AS y, NVL(z,sysdatetime) AS z FROM tbl1;
Execute OK. (0.004824 sec) Committed.1 

command(s) successfully processed.

csql> ;sc nvL_tbl1
=== <Help: Schema of a Class> ===
<Class Name> 
nvl_tbl1

<Attributes>
x INTEGER
y STRING
z DATETIME
```

**Expected:**
```
<Attributes>
x INTEGER 
y VARCHAR(1073741823) 
z DATETIME
```

This is an issue that occurs because the data value of the string type is copied and the precision of the domain is not copied in the process of creating a column through pt_make_regu_arith with the CREATE TABLE AS SELECT statement. For string type domain, the default precision is -1, and this value should be copied from the precision of the data value

### DIFF
--- a/src/object/object_domain.h
+++ b/src/object/object_domain.h
@@ -242,6 +242,12 @@ typedef enum tp_match
   (((typeid) == DB_TYPE_VARCHAR)  || ((typeid) == DB_TYPE_CHAR) || \
    ((typeid) == DB_TYPE_VARNCHAR) || ((typeid) == DB_TYPE_NCHAR))
 
+#define TP_IS_FIXED_LEN_CHAR_TYPE(typeid) \
+  (((typeid) == DB_TYPE_CHAR) || ((typeid) == DB_TYPE_NCHAR))
+
+#define TP_IS_VAR_LEN_CHAR_TYPE(typeid) \
+    (((typeid) == DB_TYPE_VARCHAR) || ((typeid) == DB_TYPE_VARNCHAR))
+
 /*
  * TP_IS_CHAR_BIT_TYPE
  *    Tests to see if a type is one of the character or bit types.
--- a/src/query/execute_schema.c
+++ b/src/query/execute_schema.c
@@ -8617,16 +8617,25 @@ do_create_entity (PARSER_CONTEXT * parser, PT_NODE * node)
       /* check for mis-creating string type with -1 precision */
       for (column = query_columns; column != NULL; column = db_query_format_next (column))
         {
-          if (column->domain->precision == DB_DEFAULT_PRECISION)
+          switch (column->domain->type->id)
             {
-              switch (column->domain->type->id)
-                {
-                case DB_TYPE_VARCHAR:
-                  column->domain->precision = DB_MAX_VARCHAR_PRECISION;
-                  break;
+              case DB_TYPE_VARCHAR:
+                if (column->domain->precision == DB_DEFAULT_PRECISION)
+                  {
+                    column->domain->precision = DB_MAX_VARCHAR_PRECISION;
+                  }
+                break;
+              case DB_TYPE_VARNCHAR:
+                if (column->domain->precision == DB_DEFAULT_PRECISION)
+                  {
+                    column->domain->precision = DB_MAX_VARNCHAR_PRECISION;
+                  }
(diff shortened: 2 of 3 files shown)


ctshim (approved):

hgryoo:
This change may cause a bug in the following case
```
CREATE TABLE tbl2 (x integer, y varchar(5555), z DATETIME);
CREATE TABLE nvl_tbl2 AS SELECT NVL(x,111) AS x, NVL(y,'aaa') AS y, NVL(z,sysdatetime) AS z FROM tbl2;

;sc nvl_tbl2
```

Expected:
```
 <Class Name>

     nvl_tbl2

 <Attributes>

     x                    INTEGER
     y                    CHARACTER VARYING(5555)
     z                    DATETIME
```
Actual:
```
 <Class Name>

     nvl_tbl2

 <Attributes>

     x                    INTEGER
     y                    CHARACTER VARYING(1073741823)
     z                    DATETIME
```

src/object/object_domain.h
```diff
@@ -242,6 +242,12 @@ typedef enum tp_match
   (((typeid) == DB_TYPE_VARCHAR)  || ((typeid) == DB_TYPE_CHAR) || \
    ((typeid) == DB_TYPE_VARNCHAR) || ((typeid) == DB_TYPE_NCHAR))
 
+#define TP_IS_FIXED_LEN_CHAR_TYPE(typeid) \
+  (((typeid) == DB_TYPE_CHAR) || ((typeid) == DB_TYPE_NCHAR))
+
+#define TP_IS_VAR_LEN_CHAR_TYPE(typeid) \
+    (((typeid) == DB_TYPE_VARCHAR) || ((typeid) == DB_TYPE_VARNCHAR))
+
```
shparkcubrid, line 250: It can be removed.

src/query/execute_schema.c
```diff
@@ -8606,11 +8606,38 @@ do_create_entity (PARSER_CONTEXT * parser, PT_NODE * node)
...
       error = pt_get_select_query_columns (parser, create_select, &query_columns);
       if (error != NO_ERROR)
 	{
 	  goto error_exit;
 	}
+
+      /* check for mis-creating string type with -1 precision */
+      for (column = query_columns; column != NULL; column = db_query_format_next (column))
+        {
+          switch (column->domain->type->id)
```
shparkcubrid, line 8620: How about?
```
if (TP_IS_VAR_LEN_CHAR_TYPE(column->domain->type->id) &&  ( column->domain->precision == DB_DEFAULT_PRECISION || column->domain->precision > DB_MAX_VARNCHAR_PRECISION)

src/object/object_domain.h
byungwook-kim: it's better to remain for use in the future.

src/query/execute_schema.c
byungwook-kim: I think the switch-case style is better for readability.

src/object/object_domain.h
shparkcubrid: Okay. I just said so that because it seemed like you miss deleting it.

src/query/execute_schema.c
shparkcubrid: Is there any reason the logic of varchar and varnchar are different?
If it's not different, the if statement I'm suggesting looks short and clear.

byungwook-kim: the max length of VARCHAR and VARNCHAR is different and it is assigned DB_MAX_VARCHAR_PRECISION instead of DB_MAX_VARNCHAR_PRECISION at VARNCHAR type in some test cases. So I separated two cases VARCHAR and VARNCHAR.

shparkcubrid: Okay.

shparkcubrid (approved):

PR #2614, opened 2021-03-16
[LETS-33] new system parameter for recovery redo parallelization

cristiarg:

New log recovery redo integer system parameter with the semantics:
- 0 - do not use parallel log recovery redo
- value greater than 0 - use parallel log recovery redo with a number of async tasks specified by the parameter
- internal hardcoded default value: 8
- maximum allowed value (internally hardcoded): 32

Since the tasks that the parameter will govern are heavily dependent on I/O-bound work (ie: retrieving/fixing/unfixing page buffer pages) it is not really of much value to bind the value of the parameter on actual number of cores that the host system has.


src/base/system_parameter.c
```diff
@@ -6126,6 +6133,18 @@ static SYSPRM_PARAM prm_Def[] = {
    (char *) NULL,
    (DUP_PRM_FUNC) NULL,
    (DUP_PRM_FUNC) NULL},
+  {PRM_ID_LOG_RECOVERY_REDO_WORKER_COUNT,
+   PRM_NAME_LOG_RECOVERY_REDO_WORKER_COUNT,
+   (PRM_FOR_SERVER | PRM_HIDDEN),
```
ribram, line 6138: Should be visible to users.
```suggestion
   (PRM_FOR_SERVER),
```

src/base/system_parameter.c
```diff
@@ -695,6 +695,7 @@ static const char sysprm_ha_conf_file_name[] = "cubrid_ha.conf";
 
 #define PRM_NAME_PAGE_SERVER_HOSTS "page_server_hosts"
 #define PRM_NAME_SERVER_TYPE "server_type"
+#define PRM_NAME_LOG_RECOVERY_REDO_WORKER_COUNT "log_recovery_redo_worker_count"
```
ribram, line 698: Maybe a shorter name?
```suggestion
#define PRM_NAME_LOG_RECOVERY_REDO_WORKER_COUNT "recovery_parallel_count"
```

ribram (commented):
Please change the title of the PR to offer a better description of the changes. Maybe "Add configuration for recovery parallelization"

src/base/system_parameter.c

src/base/system_parameter.c

ribram (approved):

PR #2615, opened 2021-03-16
[LETS-33] log compress code reuse

cristiarg:
introduce a new function `log_zip_free_data` to be used to free data for stack allocated `struct log_zip` instances; some code reuse changes


src/transaction/log_compress.h
```diff
@@ -55,10 +55,20 @@ struct log_zip
...
   char *log_data;		/* compressed/uncompressed log_zip data (used as data buffer) */
+
+    log_zip ():data_length (0), buf_size (0), log_data (nullptr)
+  {
+  }
+  log_zip (const log_zip &) = delete;
+  log_zip (log_zip &&) = delete;
+
+  log_zip & operator= (const log_zip &) = delete;
+  log_zip & operator= (log_zip &&) = delete;
```
ribram, line 66: For default initialization, the best option is to initialize members directly :).

Being a .h file, indent needs to be disabled.
```suggestion
  LOG_ZIP_SIZE_T data_length = 0;	/* length of stored (compressed/uncompressed)log_zip data */
  LOG_ZIP_SIZE_T buf_size = 0;	/* size of log_zip data buffer */
  char *log_data = nullptr;		/* compressed/uncompressed log_zip data (used as data buffer) */

  // *INDENT-OFF*
  log_zip () = default
  log_zip (const log_zip &) = delete;
  log_zip (log_zip &&) = delete;

  log_zip & operator= (const log_zip &) = delete;
  log_zip & operator= (log_zip &&) = delete;
  // *INDENT-ON*
```

src/transaction/log_compress.c
```diff
@@ -198,15 +198,18 @@ log_zip_realloc_if_needed (LOG_ZIP & log_zip, LOG_ZIP_SIZE_T new_size)
 
   if (new_size > 0 && new_size > log_zip.buf_size)
     {
-      log_zip.log_data = (char *) realloc (log_zip.log_data, new_size);
+      const LOG_ZIP_SIZE_T buf_size = LOG_ZIP_BUF_SIZE (new_size);
+      assert (buf_size <= LZ4_MAX_INPUT_SIZE);
```
ribram, line 202: Are you sure this assert is required? Existing code only checks size before `LOG_ZIP_BUF_SIZE()` transformation

cristiarg: @ribram
I thought that it makes sense, since `LOG_ZIP_BUF_SIZE` increases the input value. However, it is an unattainable corner case.

src/transaction/log_compress.h
cristiarg: Done

ribram (approved):

PR #2616, opened 2021-03-16
[LETS-33] re-arrange sequential recovery redo code in preparation for the addition of parallel recovery redo infrastructure

cristiarg:

This is a preparatory change to prepare for the addition of parallel recovery redo infrastructure.
Functionally, there is no change, only code reorganization.

log_recovery_redo.hpp:
- renamed header to `log_recovery_redo.hpp' because it contains all code related to [sequential/synchronous] recovery redo

log_recovery.h:
- added a new header and moved all functions implemented in the corresponding `.c` file and which need to be exposed for other areas
- adapted includes where needed

log_recovery.c:
- renamed function `log_rv_redo_record_sync_or_dispatch_parallel` to `log_rv_redo_record_sync_or_dispatch_async`; in a future change this function will become the point where the decision is take to either apply a log redo synchronously or dispatch it asynchronously
- out of function `log_rv_redo_record_sync_or_dispatch_async` extracted all code that will be invoked both sync and async in a new function `log_rv_redo_record_sync` which has been moved to the `log_recovery_redo.hpp` header
- moved templated functions that will be called both sequentially (sync) and async to the `log_recovery_redo.hpp` header

Note:
- file `log_recovery_redo_log_rec.hpp` has been properly `git mv`-ed to `log_recovery_redo.hpp` but it seems that GitHub does not detect, or does not use this. So, to see changes, these files would have to be manually diffed locally with smth like this: `git difftool scalability_dev LETS-33_sequential_recovery_redo -M01 -- src/transaction/log_recovery_redo_log_rec.hpp src/transaction/log_recovery_redo.hpp`


src/transaction/log_impl.h
```diff
@@ -950,7 +944,7 @@ extern void logtb_set_to_system_tran_index (THREAD_ENTRY * thread_p);
 extern LOG_LSA *logtb_find_largest_lsa (THREAD_ENTRY * thread_p);
 #endif
 extern int logtb_set_num_loose_end_trans (THREAD_ENTRY * thread_p);
-extern void log_find_unilaterally_largest_undo_lsa (THREAD_ENTRY * thread_p, LOG_LSA & max_undo_lsa);
+extern void log_find_unilaterally_largest_undo_lsa (THREAD_ENTRY * thread_p, LOG_LSA & max_undo_lsa); // TODO: move header and impl to log_recovery
```
cristiarg, line 947: this one is implemented in `log_tran_table` but only used in `log_recovery`

src/transaction/log_impl.h
```diff
@@ -1012,14 +1006,8 @@ extern int logtb_delete_global_unique_stats (THREAD_ENTRY * thread_p, BTID * bti
...
-extern int log_rv_undoredo_record_partial_changes (THREAD_ENTRY * thread_p, char *rcv_data, int rcv_data_length,
-						   RECDES * record, bool is_undo);
-extern int log_rv_redo_record_modify (THREAD_ENTRY * thread_p, LOG_RCV * rcv);
-extern int log_rv_undo_record_modify (THREAD_ENTRY * thread_p, LOG_RCV * rcv);
-extern char *log_rv_pack_redo_record_changes (char *ptr, int offset_to_data, int old_data_size, int new_data_size,
-					      char *new_data);
-extern char *log_rv_pack_undo_record_changes (char *ptr, int offset_to_data, int old_data_size, int new_data_size,
-					      char *old_data);
+extern int log_rv_redo_record_modify (THREAD_ENTRY * thread_p, LOG_RCV * rcv); // TODO: this and dependent function not used at all
+extern int log_rv_undo_record_modify (THREAD_ENTRY * thread_p, LOG_RCV * rcv); // TODO: this and dependent function not used at all
```
cristiarg, line 1010: these two do not seem to be used

src/transaction/log_recovery_redo.hpp
```diff
@@ -0,0 +1,594 @@
...
+//  static std::map<std::pair<short, int32_t>, log_lsa> map_proper_vpid_lsa;
+//  std::lock_guard<std::mutex> lck (mtx);
+//  std::pair<short, int32_t> key {rcv_vpid.volid, rcv_vpid.pageid};
+//  const auto key_it =  map_proper_vpid_lsa.find (key);
+//  if (key_it != map_proper_vpid_lsa.end())
+//    {
+//      assert ((*key_it).second < rcv_lsa);
+//    }
+//  map_proper_vpid_lsa.emplace(key, rcv_lsa);
+//#endif
```
ribram, line 525: This check seems that it will be useful to detect any inconsistencies that may come after additional changes in the future. It is better if it is active in debug mode to immediately detect inconsistencies.

The map should be cleared after recovery. It might be useful to move the code in a self-contained class. The class may be defined in a log_recovery_redo.cpp file.

Thoughts?

src/transaction/log_impl.h
ribram, line 947: How do you want to fix the todo? Imo it could be fixed immediately.

src/transaction/log_impl.h
```diff
@@ -1012,14 +1006,8 @@ extern int logtb_delete_global_unique_stats (THREAD_ENTRY * thread_p, BTID * bti
...
 
-extern int log_rv_undoredo_record_partial_changes (THREAD_ENTRY * thread_p, char *rcv_data, int rcv_data_length,
-						   RECDES * record, bool is_undo);
-extern int log_rv_redo_record_modify (THREAD_ENTRY * thread_p, LOG_RCV * rcv);
-extern int log_rv_undo_record_modify (THREAD_ENTRY * thread_p, LOG_RCV * rcv);
-extern char *log_rv_pack_redo_record_changes (char *ptr, int offset_to_data, int old_data_size, int new_data_size,
-					      char *new_data);
-extern char *log_rv_pack_undo_record_changes (char *ptr, int offset_to_data, int old_data_size, int new_data_size,
-					      char *old_data);
+extern int log_rv_redo_record_modify (THREAD_ENTRY * thread_p, LOG_RCV * rcv); // TODO: this and dependent function not used at all
```
ribram, line 1009: Move these declarations to log_recovery.h too. They are not used yet, but they are generic functions that may be used in the future for adding new recovery operations.

src/transaction/log_impl.h
ribram: The implementation too should be moved to log_recovery.c.

src/transaction/log_recovery_redo.hpp
cristiarg: Done

src/transaction/log_impl.h
cristiarg: Done

cristiarg: Done

src/transaction/log_impl.h
cristiarg: Done

ribram (approved):

PR #2619, opened 2021-03-17
[LETS-33] parallel log recovery redo

cristiarg:

This is the concluding change-set that introduces the mechanism to apply log recovery redo in parallel.

The mechanism itself is implemented in `log_recovery_redo_parallel` files and is made up of a set of classes that together implement a M-to-N prod-consumer async mechanism. On the prod side, there is only one thread - the main thread supplying redo jobs (`redo_job_impl`). On the consumer side there is a number of tasks that consume the redo jobs defined by the system parameter `PRM_ID_RECOVERY_PARALLEL_COUNT`.
The tasks consume (ie: apply) log redo records abiding by the following rules:
- within one page (defined by a VPID) log redo records are applied in the same order they appear in the log;
- at a certain moment, there is only one task that applies one log redo record on a page

To use this infrastructure, an encapsulating facade class - `redo_parallel` - is used.

Integration:
- function `log_recovery_redo`: depending on build type and system parameter, instantiate `redo_parallel` with the number of workers; this creates the async threads (tasks)
- function `log_rv_redo_record_sync_or_dispatch_async`: if the log redo record entry has a non-null VPID, a task is created and dispatched which applies the log redo record on that page;
- a number of exceptions are in place which are always applied synchronously:
  - all null VPID log redo records
  - all types specified by the function `log_rv_need_sync_redo`
- `RVDK_UNRESERVE_SECTORS`: To apply RVDK_UNRESERVE_SECTORS, one must first wait for all changes in this sector to be redone. Otherwise, asynchronous jobs skip redoing changes in this sector's pages because they are seen as deallocated. When sector is reserved again, redo is resumed in the sector's pages, but the pages are not in a consistent state.
The current workaround is to wait for all changes to be finished, including changes in the unreserved sector.
- after all log redo records have been processed (or dispatched async), the logic waits for the async ones to be applied before terminating the redo phase (end of `log_recovery_redo` function)


src/storage/page_buffer.c
```diff
@@ -1834,7 +1834,8 @@ pgbuf_fix_release (THREAD_ENTRY * thread_p, const VPID * vpid, PAGE_FETCH_MODE f
     {
       /* Make sure that the page has been allocated (i.e., is a valid page) */
       /* Suppress errors if fetch mode is OLD_PAGE_IF_IN_BUFFER. */
-      if (pgbuf_is_valid_page (thread_p, vpid, fetch_mode == OLD_PAGE_IF_IN_BUFFER, NULL, NULL) != DISK_VALID)
+      const bool no_error_when {fetch_mode == OLD_PAGE_IF_IN_BUFFER || fetch_mode == OLD_PAGE_IF_IN_BUFFER};
```
cristiarg, line 1837: @ribram 
I think changes in this file are not needed anymore. I have started a test suite without these changes just to validate this assumption.

src/transaction/log_recovery.c
```diff
@@ -572,18 +592,38 @@ template <typename T>
...
+  const bool need_sync_redo = log_rv_need_sync_redo (log_data.rcvindex);
+  assert(log_data.rcvindex != RVDK_UNRESERVE_SECTORS || need_sync_redo);
 
   // once vpid is extracted (or not), and depending on parameters, either dispatch the applying of
   // log redo asynchronously, or invoke synchronously
+  if (parallel_recovery_redo == nullptr || VPID_ISNULL (&rcv_vpid) || need_sync_redo)
+    {
+      // TODO: workaround, for few functions, make sure all dispatched actions have finished to prevent
+      // data race conditions
+      if (parallel_recovery_redo != nullptr && log_data.rcvindex == RVDK_UNRESERVE_SECTORS)
```
cristiarg, line 611: @ribram 
I need a more semantically-strong explanation for this exception (see PR description).

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -0,0 +1,397 @@
...
+	    finished = true;
+	  }
+	else
+	  {
+	    if (job == nullptr)
+	      {
+		// TODO: if needed, check if requested to finish ourselves
+
+		// expecting more data, sleep and check again
+		std::this_thread::sleep_for (std::chrono::milliseconds (WAIT_AND_CHECK_MILLIS));
```
cristiarg, line 275: I will see whether there's a better solution than this `sleep`.

src/transaction/log_recovery_redo_parallel.hpp
```diff
@@ -0,0 +1,352 @@
...
+
+      bool get_is_to_be_waited_for() const
+      {
+	return vpid.volid == NULL_VOLID || vpid.pageid == NULL_PAGEID;
+      }
+
+    private:
+      const VPID vpid;
+  };
+  using ux_redo_job_base = std::unique_ptr<redo_job_base>;
```
ribram, line 83: Please don't use aliases for std::unique_ptr in header. Better to use the full name, or use aliases in smaller scopes (unit files or inside a class/function)

src/transaction/log_recovery_redo_parallel.hpp
```diff
@@ -0,0 +1,352 @@
...
+
+      redo_job_impl ( redo_job_impl const &) = delete;
+      redo_job_impl ( redo_job_impl &&) = delete;
+
+      ~redo_job_impl() override = default;
+
+      redo_job_impl &operator = ( redo_job_impl const &) = delete;
+      redo_job_impl &operator = ( redo_job_impl &&) = delete;
+
+      int execute (THREAD_ENTRY *thread_p, log_reader &log_pgptr_reader,
```
ribram, line 111: Please move all definitions to the .cpp file

src/storage/page_buffer.c
ribram: @cristiarg me too.

src/transaction/log_recovery.c
```diff
@@ -73,7 +74,8 @@ template <typename T>
 static void log_rv_redo_record_sync_or_dispatch_async (THREAD_ENTRY *thread_p, log_reader &log_pgptr_reader,
 						       const T &log_rec, const log_lsa &rcv_lsa,
 						       const LOG_LSA *end_redo_lsa, LOG_RECTYPE log_rtype,
-						       LOG_ZIP &undo_unzip_support, LOG_ZIP &redo_unzip_support);
+						       LOG_ZIP &undo_unzip_support, LOG_ZIP &redo_unzip_support,
+						       cublog::ux_redo_parallel& parallel_recovery_redo);
```
ribram, line 78:
```suggestion
						       cublog::ux_redo_parallel &parallel_recovery_redo);
```

src/transaction/log_recovery.c
ribram: > To apply RVDK_UNRESERVE_SECTORS, one must first wait for all changes in this sector to be redone.
Otherwise, asynchronous jobs skip redoing changes in this sector's pages because they are seen as deallocated. When sector is reserved again, redo is resumed in the sector's pages, but the pages are not in a consistent state.
The current workaround is to wait for all changes to be finished, including changes in the unreserved sector.

src/transaction/log_recovery_redo_parallel.cpp
ribram: In a separate patch, the sleep loop may be replaced with a conditional wait_for.

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -0,0 +1,397 @@
...
+   **********************/
+
+  redo_parallel::redo_parallel(int a_worker_count)
+    : thread_manager (nullptr), worker_pool (nullptr), waited_for_termination (false)
+  {
+    assert(a_worker_count > 0);
+    if (a_worker_count <= 0)
+      {
+        a_worker_count = std::thread::hardware_concurrency ();
+      }
```
ribram, line 306: Not needed.
```suggestion
```

ribram (commented):
This was a partial review. I need to go through log_recovery_redo_parallel.hpp file more thoroughly.

src/transaction/log_recovery_redo_parallel.cpp
cristiarg: I added it as a "failsafe of the failsafe".

src/storage/page_buffer.c
cristiarg: Done.

src/transaction/log_recovery.c
cristiarg: Thanks

src/transaction/log_recovery_redo_parallel.cpp
cristiarg: Added http://jira.cubrid.org/browse/LETS-60

src/compat/dbtype_def.hpp
```diff
@@ -0,0 +1,28 @@
+#ifndef DBTYPE_DEF_HPP
+#define DBTYPE_DEF_HPP
+
+#include "dbtype_def.h"
+
+/* C++ extensions for strict-C structures defined in corresponding '.h' header
+ */
+
+inline constexpr bool operator== (const VPID &_left, const VPID &_rite)
```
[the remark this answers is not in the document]
cristiarg: Done.

src/thread/thread_manager.cpp
```diff
@@ -85,6 +85,9 @@ namespace cubthread
     return;
 #else // not SA_MODE = SERVER_MODE
 
+    assert(m_all_entries == nullptr);
+    assert(m_entry_dispatcher == nullptr);
```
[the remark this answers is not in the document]
cristiarg: Done.

src/transaction/log_reader.hpp
```diff
@@ -31,6 +31,8 @@
  * NOTE: further functionality:
  *  - introduce an internal buffer to be used for copying and serving
  *    data from the log of arbitrary size; currently this is done manually outside this class
+ *
+ * TODO: introduce THREAD_ENTRY as argument to all functions or as ctor argument
```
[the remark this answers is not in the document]
cristiarg: Done

src/transaction/log_recovery.c
```diff
@@ -572,18 +592,38 @@ template <typename T>
...
-                                                LOG_ZIP & undo_unzip_support, LOG_ZIP & redo_unzip_support)
+                                                LOG_ZIP & undo_unzip_support, LOG_ZIP & redo_unzip_support,
+                                                cublog::ux_redo_parallel& parallel_recovery_redo)
 {
   const VPID rcv_vpid = log_rv_get_log_rec_vpid<T> (log_rec);
   // at this point, vpid can either be valid or not
 
   const LOG_DATA &log_data = log_rv_get_log_rec_data<T> (log_rec);
+  const bool need_sync_redo = log_rv_need_sync_redo (log_data.rcvindex);
+  assert(log_data.rcvindex != RVDK_UNRESERVE_SECTORS || need_sync_redo);
```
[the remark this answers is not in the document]
cristiarg: Done

src/transaction/log_recovery.c
```diff
@@ -572,18 +592,38 @@ template <typename T>
 void log_rv_redo_record_sync_or_dispatch_async (THREAD_ENTRY * thread_p, log_reader & log_pgptr_reader,
                                                 const T & log_rec, const log_lsa & rcv_lsa,
                                                 const LOG_LSA * end_redo_lsa, LOG_RECTYPE log_rtype,
-                                                LOG_ZIP & undo_unzip_support, LOG_ZIP & redo_unzip_support)
+                                                LOG_ZIP & undo_unzip_support, LOG_ZIP & redo_unzip_support,
+                                                cublog::ux_redo_parallel& parallel_recovery_redo)
```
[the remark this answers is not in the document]
cristiarg: Done

src/transaction/log_recovery_redo_parallel.hpp
```diff
@@ -0,0 +1,352 @@
...
+	: vpid (a_vpid)
+      {
+	// the logic implemented below, in the queue and task, does allow for null-vpid type of
+	// entries to be executed - they are called "to be waited for" or "synched" operations;
+	// but, for now, do not allow them to be dispatched to be executed asynchronously
+	assert (!VPID_ISNULL (&vpid));
+      }
+      redo_job_base () = delete;
+      redo_job_base ( redo_job_base const &) = delete;
+      redo_job_base ( redo_job_base &&) = delete;
```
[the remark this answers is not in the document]
cristiarg: Solved

src/transaction/log_recovery_redo_parallel.hpp
```diff
@@ -0,0 +1,352 @@
...
+      {
+	return vpid;
+      }
+
+      virtual int execute (THREAD_ENTRY *thread_p, log_reader &log_pgptr_reader,
+			   LOG_ZIP &undo_unzip_support, LOG_ZIP &redo_unzip_support) = 0;
+
+      bool get_is_to_be_waited_for() const
+      {
+	return vpid.volid == NULL_VOLID || vpid.pageid == NULL_PAGEID;
```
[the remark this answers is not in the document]
cristiarg: Removed.
Mechanism for "is to be waited for" jobs has been added as part of the PoC. In the end, it turns out we do not need it because all such jobs (mainly: those which do not have a non-null VPID) are in fact executed on the main thread (ie: synchronously) which achieves the same effect as was implemented by this mechanism.

src/transaction/log_recovery_redo_parallel.hpp
cristiarg: Done. Kept `typedef`s only as part of the internal implementation.

src/transaction/log_recovery.c
cristiarg: Done

src/transaction/log_recovery_redo_parallel.cpp
cristiarg: Removed. Made param `unsigned`.

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -0,0 +1,386 @@
...
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+
+#include "vpid.hpp"
+#include "log_recovery_redo_parallel.hpp"
```
ribram, line 20: Own header first, then project includes.
```suggestion
#include "log_recovery_redo_parallel.hpp"

#include "vpid.hpp"
```

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -0,0 +1,386 @@
...
+
+namespace cublog
+{
+  /*********************************************************************
+   * redo_parallel::redo_job_queue
+   *********************************************************************/
+
+  redo_parallel::redo_job_queue::redo_job_queue ()
+    : m_produce_queue ( new ux_redo_job_deque () )
+    , m_consume_queue ( new ux_redo_job_deque () )
```
ribram, line 30:
```suggestion
    : m_produce_queue (new ux_redo_job_deque ())
    , m_consume_queue (new ux_redo_job_deque ())
```

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -0,0 +1,386 @@
...
+  bool redo_parallel::redo_job_queue::get_adding_finished () const
+  {
+    return m_adding_finished.load ();
+  }
+
+  redo_parallel::redo_job_queue::ux_redo_job_base redo_parallel::redo_job_queue::locked_pop (bool &out_adding_finished)
+  {
+    std::unique_lock<std::mutex> consume_queue_lock (m_consume_queue_mutex);
+
+    out_adding_finished = false;
```
ribram, line 70: May be loaded here directly.
```suggestion
    out_adding_finished = m_adding_finished.load ();
```

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -0,0 +1,386 @@
...
+
+    if (consume_queue_it != m_consume_queue->end ())
+      {
+	job = std::move (*consume_queue_it);
+	m_consume_queue->erase (consume_queue_it);
+      }
+    else
+      {
+	// consumer will have to spin-wait
+      }
```
ribram, line 177:
```suggestion
    for (auto queue_it = m_consume_queue->begin (); queue_it != m_consume_queue->end (); ++queue_it)
      {
	const VPID it_vpid = (*queue_it)->get_vpid ();
	if (m_in_progress_vpids.find ((it_vpid)) == m_in_progress_vpids.cend ())
	  {
	    job = std::move (*queue_it);
	    m_consume_queue->erase (queue_it);
	    break;
	  }
      }
```

src/transaction/log_recovery_redo_parallel.hpp
```diff
@@ -0,0 +1,274 @@
...
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+
+#ifndef _LOG_RECOVERY_REDO_PARALLEL_HPP_
+#define _LOG_RECOVERY_REDO_PARALLEL_HPP_
+
+#include "log_compress.h"
```
ribram, line 22: I think all headers are only SERVER_MODE.

Some headers may be replaced with forward declarations (thread related ones)

<bitset> is probably unused anymore.

src/transaction/log_recovery_redo_parallel.hpp
```diff
@@ -0,0 +1,274 @@
...
+	log_pgptr_reader.add_align (sizeof (LOG_RECORD_HEADER));
+	log_pgptr_reader.advance_when_does_not_fit (sizeof (log_rec_t));
+	const log_rec_t log_rec
+	  = log_pgptr_reader.reinterpret_copy_and_add_align<log_rec_t> ();
+
+	const auto &rcv_vpid = get_vpid ();
+	log_rv_redo_record_sync<log_rec_t> (thread_p, log_pgptr_reader, log_rec, rcv_vpid, m_rcv_lsa, m_end_redo_lsa,
+					    m_log_rtype, undo_unzip_support, redo_unzip_support);
+	return NO_ERROR;
+      }
```
ribram, line 258: Move definition to .cpp file.

src/transaction/log_recovery_redo_parallel.hpp
```diff
@@ -0,0 +1,274 @@
...
+
+      ~redo_job_impl () override = default;
+
+      redo_job_impl &operator = (redo_job_impl const &) = delete;
+      redo_job_impl &operator = (redo_job_impl &&) = delete;
+
+      int execute (THREAD_ENTRY *thread_p, log_reader &log_pgptr_reader,
+		   LOG_ZIP &undo_unzip_support, LOG_ZIP &redo_unzip_support) override
+      {
+	const int err_set_lsa_and_fetch_page =  log_pgptr_reader.set_lsa_and_fetch_page (m_rcv_lsa);
```
ribram, line 244:
```suggestion
	const int err_set_lsa_and_fetch_page = log_pgptr_reader.set_lsa_and_fetch_page (m_rcv_lsa);
```

src/transaction/log_recovery.c
```diff
@@ -3170,6 +3217,23 @@ log_recovery_redo (THREAD_ENTRY * thread_p, const LOG_LSA * start_redolsa, const
   LOG_ZIP *redo_unzip_ptr = NULL;
   bool is_mvcc_op = false;
 
+  /* depending on compilation mode and on a system parameter, initialize the
+   * infrastructure for parallel log recovery;
+   * if infrastructure is not initialized dependent code below works sequentially
+   */
+  LOG_CS_EXIT (thread_p);
+  std::unique_ptr < cublog::redo_parallel > parallel_recovery_redo;
```
ribram, line 3225:
```suggestion
  // *INDENT-OFF*
  std::unique_ptr<cublog::redo_parallel> parallel_recovery_redo;
  // *INDENT-ON*
```

src/transaction/log_recovery_redo_parallel.cpp
cristiarg: Done

src/transaction/log_recovery_redo_parallel.cpp
cristiarg: Done

src/transaction/log_recovery_redo_parallel.cpp
cristiarg: I did it like this, conservatively, on purpose. The out param is set only when it is needed to be checked. There's a sort of hierarchy between the return value and the out param value:
- the return value has higher prio
- the out param is to be checked only when the return value is null
Ideally, it should be refactored into a semantically stronger return type.
I can do it if needed.

src/transaction/log_recovery_redo_parallel.hpp
cristiarg: Yes, better to guard all of them `SERVER_MODE` just to be on the safe side.

src/transaction/log_recovery_redo_parallel.hpp
cristiarg: Tried.
Linker error because the template is instantiated externally in `log_recovery.c`:
```      
    using redo_job_impl_t = cublog::redo_job_impl<T>;
    std::unique_ptr<redo_job_impl_t> job{ new redo_job_impl_t (rcv_vpid, rcv_lsa, end_redo_lsa, log_rtype) };
```

src/transaction/log_recovery_redo_parallel.hpp
cristiarg: Done

src/transaction/log_recovery.c
cristiarg: Done

src/transaction/log_recovery_redo_parallel.cpp
cristiarg: Later edit.
I've added a proper explanation why it is loaded only in one situation:
```
	      // specifically leave the 'out_adding_finished' on false:
	      //  - even if adding has been finished, the produce queue might still have jobs to be consumed
	      //  - if so, setting the out param to true here, will cause consuming tasks to finish
	      //  - thus allowing for a corner case where the jobs are not fully processed and all
	      //    tasks have finished executing
```

src/transaction/log_recovery_redo_parallel.cpp

ribram (approved):

PR #2624, opened 2021-03-21
[CBRD-23887] bash autocompletion scripts for cubrid and csql utilities

cristiarg:
http://jira.cubrid.org/browse/CBRD-23887

I wrote these for my own use and thought that they might benefit others as well.
Bash autocomplete for `cubrid` utility:
  - distinguishes betweend the service's utilities and administrator's utilities
  - is aware of the context and, for each utlity, list that utility's options (the long form only, as is customary for bash completion)
  - is aware of the context and either lists - or not - the existing database names as well by parsing the names from the `$CUBRID_DATABASES/databases.txt` file
  - caveat: uses a hardcoded list of both service's and administrator's utilities
Bash autocomplete script for `csql` utility:
  - lists the name of the existing databases by parsing the `$CUBRID_DATABASES/databases.txt` file

If there are other utilities for which bash autocompletion would be useful, I can write functions for those as well.

### DIFF
--- a/contrib/bash/csql
+++ b/contrib/bash/csql
@@ -1,0 +1,52 @@
+#
+# Cubrid CSQL client utility bash autocompletion script
+# Usage:
+#   - copy to /etc/bash_completion_d/ and it should be picked up by bash
+#
+# TODO:
+#   - in case of '--input-file' list local [sql] files
+#   - for user, list database users if these can be extracted somehow
+
+_csql_bash_completion() 
+{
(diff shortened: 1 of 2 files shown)


hornetmj:
@cristiarg I register @kisoo-han as a reviewer. In my opinion, he is a better reviewer than me.

mhoh3963 (approved):
Sounds good to me.

PR #2625, opened 2021-03-22
[CBRD-23879] Outer dependency is set improperly when rewriting SQL from outer join to inner join

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-23879

AS-is: When written in ANSI SQL syntax, the order of scan of tables cannot be changed.
```
select * from t1 inner join t2 on t1.a = t2.a left outer join t3 on t1.a = t3.a

t1->t2->t3 join order only can be selected
actually t2->t1->t3, t1->t3->t2 can be generated
```

To-be: Outer dependency is set only for outer joined tables.

Dependency is set based on the following two criteria.
1. Left outer join: The preceding table is set as a dependency.
2. Right outer join: The all preceding tables written in the ANSI SQL syntax are set as dependency.
                        (The right dep set is used to check the preceding tables of the right outer join)

### DIFF
--- a/src/optimizer/query_graph.h
+++ b/src/optimizer/query_graph.h
@@ -371,10 +371,10 @@ struct qo_node
   /* NULL if no USING INDEX clause in the query */
 
   BITSET outer_dep_set;		/* outer join dependency; to preserve join sequence */
+  BITSET right_dep_set;		/* outer join dependency for right outer join; to preserve join sequence */
   PT_HINT_ENUM hint;		/* hint comment contained in given */
   bool sargable;		/* whether sargs are applicable to this node */
   bool sort_limit_candidate;	/* whether this node is a candidate for a SORT_LIMIT plan */
-  BITSET right_dep_set;		/* outer join dependency for right outer join; to preserve join sequence */
 };
 
 #define QO_NODE_ENV(node)		(node)->env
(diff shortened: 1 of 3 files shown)


ctshim (approved):

src/optimizer/query_planner.c
```diff
@@ -6743,9 +6744,14 @@ planner_visit_node (QO_PLANNER * planner, QO_PARTITION * partition, PT_HINT_ENUM
 	    edge_cnt++;
 
 	    /* set join type */
-	    if (join_type == NO_JOIN)
+	    if (join_type == NO_JOIN || is_dummy_term)
```
byungwook-kim, line 6747: I thank it needs a comment to replace as-is comment explains why **is_dummy_term** is applied.

src/optimizer/query_graph.h
```diff
@@ -374,6 +374,7 @@ struct qo_node
   PT_HINT_ENUM hint;		/* hint comment contained in given */
   bool sargable;		/* whether sargs are applicable to this node */
   bool sort_limit_candidate;	/* whether this node is a candidate for a SORT_LIMIT plan */
+  BITSET right_dep_set;		/* outer join dependency for right outer join; to preserve join sequence */
```
byungwook-kim, line 377: it's better **right_dep_set** to move below the field "BITSET outer_dep_set"

youngjinj (approved):

src/optimizer/query_graph.h
shparkcubrid: Okay.

src/optimizer/query_planner.c
shparkcubrid: Okay. I'll fix it

hgryoo (approved):
Looks good to me. It would better to describe right_dep_set and to make macro function in setting right_dept_set in qo_add_dummy_join_term() and qo_analyze_term() for readability.

byungwook-kim (approved):

PR #2628, opened 2021-03-22
[LETS-10] Do not persist info about temporary volumes

cristiarg:
http://jira.cubrid.org/browse/LETS-10

The creation of temporary data volumes is not meant to be replicated on page server. Each server has its own temporary data.
As a consequence, temporary data creation cannot be logged and information on temporary data cannot be saved to disk.

Implementation:
- removed temporary volumes info from `boot_dbparm` struct and replaced with dummy same-alignment dummy data in order to maintain backwards compatibility
- added a new struct `boot_temp_info` which is used to store info about temporary volumes and is used to remove temporary volumes upon server clean-close
- in case of server abnormal termination - and when temporary volumes are present - the function `boot_remove_unknown_temp_volumes` has been adapted to work without any additional data: it checks for the presence of temporary volumes starting from `LOG_MAX_DBVOLID` downwards and removes them in ascending order
- adapted function `boot_dbparm_save_volume`  to only log undo data for other than `DB_TEMPORARY_VOLTYPE` volume type

Manual tests:
- connect with `csql` console with user `dba` to a running database instance (preferably one where the volume size is small, eg: `cubrid createdb sdb --db-volume-size=20m --log-volume-size=20m en_US`)
- execute, in at least one or more consoles, statements that lead to creation of temporary volumes:
```
select 'a very long string' from _db_class a, _db_class b, _db_class c, _db_class d;
```
- check that temporary volumes are being created; be sure that every time the statement is executed, a different string is used (to prevent results reuse); the more consoles such a statement is executed in, the more temporary volume files are created
- close the server normally, observe that temporary volume files are removed
- repeat the test but kill the `cub_server` process while temporary volume files are still on disk, restart the server and observe that the temporary volume files are removed upon start

Automatic tests:
- a full suite of shell tests has been executed


src/transaction/boot_sr.c
```diff
@@ -988,34 +994,50 @@ boot_make_temp_volume_fullname (char *temp_vol_fullname, VOLID temp_volid)
...
       boot_make_temp_volume_fullname (temp_vol_fullname, temp_volid);
       if (!fileio_is_volume_exist (temp_vol_fullname))
 	{
+	  /* the last found volume id */
+	  /* if no temporary volumes are found, and this is LOG_MAX_DBVOLID, incrementing here
+	   * will not overflow and will invalidate the loop below as well */
+	  ++temp_volid;
 	  break;
 	}
       er_set (ER_NOTIFICATION_SEVERITY, ARG_FILE_LINE, ER_BO_UNKNOWN_VOLUME, 1, temp_vol_fullname);
```
ribram, line 1027: ER_BO_UNKNOWN_VOLUME is also logged in the next loop. Remove a duplicate.
```suggestion
```

src/transaction/boot_sr.c
```diff
@@ -6121,28 +6143,35 @@ boot_dbparm_save_volume (THREAD_ENTRY * thread_p, DB_VOLTYPE voltype, VOLID voli
...
 	  goto exit;
 	}
-      boot_Db_parm->temp_nvols++;
-      boot_Db_parm->temp_last_volid = volid;
+      boot_Temp_info.temp_nvols++;
+      boot_Temp_info.temp_last_volid = volid;
     }
 
   /* todo: is flush needed? */
   VPID_GET_FROM_OID (&vpid_boot_bp_parm, boot_Db_parm_oid);
```
ribram, line 6159: All this part: logging, update boot_db_parm, heap_flush & fileio_synchronize can be moved to `if (voltype == DB_PERMANENT_VOLTYPE)`

src/transaction/boot_sr.c
cristiarg: Done

src/transaction/boot_sr.c
cristiarg: Done

src/transaction/boot_sr.c
```diff
@@ -989,34 +995,49 @@ boot_make_temp_volume_fullname (char *temp_vol_fullname, VOLID temp_volid)
...
-    }
+  assert (!BO_IS_SERVER_RESTARTED ());
+  /* logic below needs this invariant */
+  static_assert (LOG_MAX_DBVOLID < VOLID_MAX, "LOG_MAX_DBVOLID is expected to be less than VOLID_MAX");
 
-  for (; temp_volid > boot_Db_parm->last_volid; temp_volid--)
+  /* inspect volumes in descending order */
+  for (; temp_volid > LOG_DBFIRST_VOLID; temp_volid--)
     {
       boot_make_temp_volume_fullname (temp_vol_fullname, temp_volid);
```
cristiarg, line 1019: Does it make sense to memoize the results of `boot_make_temp_volume_fullname` in a stack-like struct and just pop them below? I noticed that this function might do some I/O when checking for `pwd`.

ribram: @cristiarg I think it will have insignificant impact compared to the overall time to boot the database.

ribram (approved):

PR #2629, opened 2021-03-23
[CBRD-23881] Add a new option to the vacuumdb utility to output the smallest log page ID and log volume name referenced by vacuum

hornetmj:
http://jira.cubrid.org/browse/CBRD-23881

Add the following two options to the vacuumdb utility.
- --dump
  - Prints the following two pieces of information
    - smallest log page ID referenced by vacuum
    - log volume name including smallest log page ID
- -o, --output-file=FILE

**Usage**
```
$ cubrid vacuumdb
vacuumdb: Vacuums dead records from database.
usage: cubrid vacuumdb [OPTION] database-name

valid options:
  -S, --SA-mode                stand-alone mode execution
  -C, --CS-mode                client-server mode execution
      --dump                   dump current state of vacuum
  -o, --output-file=FILE       redirect output messages to FILE; default: none
```

**result of --dump option**
- --CS-mode
```
$ cubrid vacuumdb --CS-mode --dump demodb

*** Vacuum Dump ***
First log page ID referenced = 93 (in demodb_lgat)
```
- --SA-mode
```
$ cubrid vacuumdb --SA-mode --dump demodb

*** Vacuum Dump ***
First log page ID referenced = 93 (in demodb_lgat)
```

PR #2640, opened 2021-03-29
[CBRD-23881] Add a new option to the vacuumdb utility to output the smallest log page ID and log volume name referenced by vacuum (#2629)

hornetmj:
http://jira.cubrid.org/browse/CBRD-23881

backport #2629

The following is different compared to the develop branch modification.
- When using the --dump option in SA mode, VACUUM should not be performed. For this, I used the code committed in CBRD-23712 when modifying the develop branch. Currently, the modifications of CBRD-23712 are not merged in version 10.1, so only necessary parts have been applied.
- One of the messages is changed in the xvacuum_dump() function. Because, unlike the develop branch, the return value of vacuum_min_log_pageid_to_keep() may be NULL_PAGEID even after vacuum_initialize() in version 10.1. (The code is a little different.)
   - AS-IS
     - "vacuum did not boot properly."
   - TO-BE
     - "There are no log pages referenced by vacuum."
- The vacuum_Is_booted global variable is added.
- Other processing parts for SA_MODE have been slightly changed.

### DIFF
--- a/msg/de_DE.utf8/utils.msg
+++ b/msg/de_DE.utf8/utils.msg
@@ -1156,13 +1156,16 @@ gültigen Optionen:\n\
 $set 55 MSGCAT_UTIL_SET_VACUUMDB
 20 vacuumdb vorübergehend nicht verfügbar ist für Client-Server-Modus\n
 21 Der Vakuumbetrieb gescheitert\n
+22 Ausgangsdatei '%1$s' konnte nicht geöffnet werden \n
 60 \
 vacuumdb: Vakuum Toten Datensätze aus der Datenbank.\n\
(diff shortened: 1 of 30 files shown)


mhoh3963 (approved):

wrlawodms (approved):
I think we should get the error message to be the same later when we want to provide showing vacuum dump officially. It might be into MSGCAT then.

JoohoK (approved):

hornetmj:
> I think we should get the error message to be the same later when we want to provide dumping vacuum stats officially. It might be into MSGCAT then.

I agree with you. But there is a slight difference in the code base between 10.1 and develop. I hope to be able to provide consistent error messages defined in msg files.

PR #2642, opened 2021-03-29
[LETS-67] Fix -Wregister warnings

adrianbrab:
http://jira.cubrid.org/browse/LETS-67

Fix all warnings about register specifier

Example:

 ```
warning: ISO C++17 does not allow 'register' storage class specifier [-Wregister]
 253 | register int k = 0
```


ribram (approved):
@hornetmj please consider this for backporting to develop.

hornetmj (approved):

PR #2646, opened 2021-03-30
[LETS-68] Fix -Wdeprecated-declarations caused by use of ftime

adrianbrab:
http://jira.cubrid.org/browse/LETS-68

‘int ftime(timeb*)’ is deprecated and should be replaced with the usage of functions from std::chrono

### DIFF
--- a/src/base/tz_support.c
+++ b/src/base/tz_support.c
@@ -5470,6 +5470,7 @@ exit:
   return err_status;
 }
 
+//TODO: make tz_get_offset_in_mins get into account DST
 /*
  * tz_get_offset_in_mins () - time zone offset in minutes from GMT
  */
@@ -5487,12 +5488,6 @@ tz_get_offset_in_mins ()
 
   // Get offset in minutes from UTC
   double offsetFromUTC = difftime (utc, local) / 60;
-
-  // Adjust for DST
-  if (timeinfo->tm_isdst)
-    {
-      offsetFromUTC -= 60;
-    }
 
   return offsetFromUTC;
 }
--- a/src/base/tz_support.c
+++ b/src/base/tz_support.c
@@ -5487,7 +5487,7 @@ tz_get_offset_in_mins ()
   time_t local = mktime (timeinfo);
 
   // Get offset in minutes from UTC
-  double offsetFromUTC = difftime (utc, local) / 60;
+  int offsetFromUTC = difftime (utc, local) / 60;
 
   return offsetFromUTC;
 }
--- a/src/base/tz_support.c
+++ b/src/base/tz_support.c
@@ -84,7 +84,7 @@ typedef enum ds_search_direction DS_SEARCH_DIRECTION;
 
 
 #define FULL_DATE(jul_date, time_sec) ((full_date_t) jul_date * 86400ll \
-				       + (full_date_t) time_sec)
+                                       + (full_date_t) time_sec)
 #define TIME_OFFSET(is_utc, offset) \
   ((is_utc) ? (-offset) : (offset))
 #define ABS(i) ((i) >= 0 ? (i) : -(i))
@@ -188,9 +188,9 @@ static int tz_get_iana_zone_id_by_windows_zone (const char *windows_zone_name);
     v = (SYM_TYPE) TZ_GET_SYM_ADDR (lh, SYM_NAME);			    \
     if (v == NULL)							    \
       {									    \
-	strncpy (sym_name, (SYM_NAME), sizeof (sym_name) - 1);		    \
-	sym_name[sizeof (sym_name) - 1] = '\0';				    \
-	goto error_loading_symbol;					    \
+        strncpy (sym_name, (SYM_NAME), sizeof (sym_name) - 1);		    \
+        sym_name[sizeof (sym_name) - 1] = '\0';				    \
+        goto error_loading_symbol;					    \
       }									    \
   } while (0)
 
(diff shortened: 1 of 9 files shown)


ribram (approved):
@hornetmj this PR should be backported to develop too.

src/base/tz_support.c
```diff
@@ -5469,3 +5469,16 @@ conv_tz (void *p_out, const void *p_in, DB_TYPE type)
...
+ * tz_get_offset_in_mins () - time zone offset in minutes from GMT
+ - gmtime(0) returns the GMT time of the second ZERO of the epoch
+ - mktime() simply translate the above moment into equivalent local time
+ - the diff between these 2 values gives the timezone offset in seconds (but keep in mind that one of them is ZERO)
+ */
+int
+tz_get_offset_in_mins ()
+{
+  time_t tt0 = 0;
+  return (int) (mktime (gmtime (&tt0)) / -60);
```
razvanradulescu, line 5483: This is used in db_sys_timezone to return current timezone (as in moment of call).
The  code of tz_get_offset_in_mins returns the timezone as it was in 1970.
We may find examples (locale, and current time) were the tow values are not the same.

razvanradulescu: https://arstechnica.com/civis/viewtopic.php?t=136625

adrianbrab: As a fun fact there are in fact multiple countries who have changed their timezone since 1970 , Venezuela comes to mind. So I changed the implementation.

src/base/tz_support.c
```diff
@@ -5469,3 +5469,30 @@ conv_tz (void *p_out, const void *p_in, DB_TYPE type)
...
+  timeinfo = localtime (&currtime);
+  time_t local = mktime (timeinfo);
+
+  // Get offset in minutes from UTC
+  double offsetFromUTC = difftime (utc, local) / 60;
+
+  // Adjust for DST
+  if (timeinfo->tm_isdst)
+    {
+      offsetFromUTC -= 60;
```
razvanradulescu, line 5494: I looked more into the code and I think best option is to change the implementation of TZ_OFFSET (https://www.cubrid.org/manual/en/11.0/sql/function/datetime_fn.html#tz-offset)
which is the only context for using db_sys_timezone function.
We should prefer using our TZ implementation to using system functions.
This should be a concern for another issue.

For this PR, I propose to not adjust for DST:
1) Old version of db_sys_timezone uses ftime
I didn't found a clear description of expected value of ftime->timezone regarding if it already includes dst offset or not; in the previous code, we don't take ftime->dstflag into account.
I suggest make a small test with ftime and check if timezone field already includes DST or not.
We should return the same value as by using ftime (have no test regression)

2) DST offset is not always -60 mins.
For instance :
https://github.com/CUBRID/cubrid/blob/develop/timezones/tzdata/asia#L1982
(20 minutes)

src/base/tz_support.c
```diff
@@ -5469,3 +5469,25 @@ conv_tz (void *p_out, const void *p_in, DB_TYPE type)
...
+  time (&currtime);
+  timeinfo = gmtime (&currtime);
+  time_t utc = mktime (timeinfo);
+  timeinfo = localtime (&currtime);
+  time_t local = mktime (timeinfo);
+
+  // Get offset in minutes from UTC
+  double offsetFromUTC = difftime (utc, local) / 60;
+
+  return offsetFromUTC;
```
hornetmj, line 5492: How about '(int) offsetFromUTC'?

src/base/tz_support.c
```diff
@@ -5469,3 +5469,25 @@ conv_tz (void *p_out, const void *p_in, DB_TYPE type)
...
+  struct tm *timeinfo;
+
+  time (&currtime);
+  timeinfo = gmtime (&currtime);
+  time_t utc = mktime (timeinfo);
+  timeinfo = localtime (&currtime);
+  time_t local = mktime (timeinfo);
+
+  // Get offset in minutes from UTC
+  double offsetFromUTC = difftime (utc, local) / 60;
```
hornetmj, line 5490: I am a little confused. But isn't it correct to call difftime (local, utc)?

src/base/tz_support.h
```diff
@@ -228,6 +228,7 @@ extern "C"
   extern int tz_create_datetimetz_from_parts (const int m, const int d, const int y, const int h, const int mi,
 					      const int s, const int ms, const TZ_ID * tz_id, DB_DATETIMETZ * dt_tz);
   extern int conv_tz (void *, const void *, DB_TYPE);
+  int tz_get_offset_in_mins ();	//time zone offset in minutes from GMT
```
hornetmj, line 231: Don't need the extern keyword?

hornetmj (approved):

src/broker/broker_util.c
```diff
@@ -430,12 +431,12 @@ ut_time_string (char *buf, struct timeval *time_val)
...
-
-      /* current time */
-      (void) ftime (&tb);
-      sec = tb.time;
-      millisec = tb.millitm;
+      timespec ts = { };
+      timespec_get (&ts, TIME_UTC);
+      sec = ts.tv_sec;
+      // *INDENT-OFF*
+      millisec = std::chrono::duration_cast<std::chrono::milliseconds> (std::chrono::nanoseconds (ts.tv_nsec)).count();
```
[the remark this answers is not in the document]
adrianbrab: I corrected the mistake

src/query/string_opfunc.c
```diff
@@ -13205,31 +13207,34 @@ db_sys_datetime (DB_VALUE * result_datetime)
...
+  timespec tloc = { };
   struct tm *c_time_struct, tm_val;
 
   assert (result_datetime != (DB_VALUE *) NULL);
 
   /* now return null */
   db_value_domain_init (result_datetime, DB_TYPE_DATETIME, DB_DEFAULT_PRECISION, DB_DEFAULT_SCALE);
 
-  if (ftime (&tloc) != 0)
+  if (timespec_get (&tloc, TIME_UTC) != TIME_UTC)
```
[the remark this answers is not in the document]
adrianbrab: Changed to using 0, TIME_UTC is defined as being 1 so the error was not set when necessary

src/base/tz_support.c
adrianbrab: I changed the type, the return type of the function was already int

src/base/tz_support.c
adrianbrab: I think timezones are defined as (gmt - local)

src/base/tz_support.h
adrianbrab: I don't think is necessary in this situation

PR #2659, opened 2021-04-01
[LETS-52] Fix cubrid service/server stop hang when cub_server remains zombie

ribram:
http://jira.cubrid.org/browse/LETS-52

A consequence of the design of starting server process causes the process to become defunct after it stops. Normally, child processes should be waited by their parent processes.
Linux kernels know how to handle orphan zombie processes and init process reaps them and properly removes them from process table. However, init of docker containers misses cub_server process.
As consequence, the cubrid service/server stop commands are blocked until cub_server pid is no longer found, which never happens.
No proper solution to reap the zombie process was found. The fix is a workaround to force the commands to ignore zombie processes. The zombie state of a process can be determined based on the flag inside /proc/[pid]/stat file.


adrianbrab (approved):

adrianbrab:
Could a way to kill the zombie process exist, given the fact that the pid is known? or is it not desired?

cristiarg:
https://www.howtogeek.com/701971/how-to-kill-zombie-processes-on-linux/
According to this `The only reliable solution is to kill the parent process.` 
So, might this not be the proper solution?
[ Just asking. I don't know the proper answer. ]

ribram:
> https://www.howtogeek.com/701971/how-to-kill-zombie-processes-on-linux/
> According to this `The only reliable solution is to kill the parent process.`
> So, might this not be the proper solution?
> [ Just asking. I don't know the proper answer. ]

Yeah, we don't want to kill cub_master, the parent process :). And I don't know another way to get rid of the zombie process.

Not getting rid is not such a big deal because of the next reasons:

  - the issue was never experienced it any linux systems. the process is always reaped.
  - the issue was experienced in some docker containers, but not in all
  - if the cub_master process is stopped, then the zombies disappear. `cubrid service stop` command also stops cub_master, so the only problematic cases remain a) `cubrid server stop` and b) server crashes.

cristiarg (approved):

PR #2666, opened 2021-04-05
[CBRD-23903] Make COUNT function's return type as BIGINT

hgryoo:
http://jira.cubrid.org/browse/CBRD-23903

NOTE: Please refer to the JIRA issue on how to use _the count_debug system_ parameter to test counting from the value you set.
It is just for doing code review easily and this system parameter will be removed before merging.


hgryoo:
Failed cases in test_sql occurs according to the following
1) Data type is changed from INTEGER to BIGINT at _DESCRIBE_ query.
2) type conversion;
    ```select foo.a, foo.b, count(foo.c) from foo foo group by foo.a, foo.b having foo.a= cast(foo.b as bigint)+count(foo.c)```

src/base/system_parameter.c
```diff
@@ -2344,6 +2346,12 @@ static UINT64 prm_ddl_audit_log_size_lower = 10485760ULL;	/* 10M */
 static UINT64 prm_ddl_audit_log_size_upper = 2147483648ULL;	/* 2G */
 static unsigned int prm_ddl_audit_log_size_flag = 0;
 
+UINT64 PRM_COUNT_DEBUG = false;
+static UINT64 prm_count_debug_default = 0;	/* 10M */
+static UINT64 prm_count_debug_lower = 0;	/* 10M */
+static UINT64 prm_count_debug_upper = UINT64_MAX;	/* 2G */
```
hgryoo, line 2352: please ignore the comment :-)

src/parser/type_checking.c
```diff
@@ -5444,8 +5447,8 @@ pt_coerce_range_expr_arguments (PARSER_CONTEXT * parser, PT_NODE * expr, PT_NODE
 	{
 	  PT_NODE *temp = NULL;
 	  int precision = 0, scale = 0;
-	  int units = LANG_SYS_CODESET; /* code set */
-	  int collation_id = LANG_SYS_COLLATION; /* collation_id */
+	  int units = LANG_SYS_CODESET;	/* code set */
+	  int collation_id = LANG_SYS_COLLATION;	/* collation_id */
```
hgryoo, line 5451: it is an unrelated change caused by broken code formatting at any previous commit.

src/base/system_parameter.c
ctshim: What does the 10M, 2G comment mean?

hgryoo: There is no meaning. It is just copied from another code. This line will be erased before merging.

src/parser/type_checking.c
```diff
@@ -19964,7 +20006,6 @@ pt_fold_const_function (PARSER_CONTEXT * parser, PT_NODE * func)
 	      func->info.function.arg_list = NULL;
 	    }
 	}
-      func->type_enum = PT_TYPE_INTEGER;
```
ctshim, line 19967: Is it not necessary to replace PT_TYPE_INTEGER with PT_TYPE_BIGINT?

hgryoo: There are two routines to evaluate the (arg, return) type for the function node. I migrated _PT_COUNT_ and _PT_COUNT_STAR_ into the new routine. The new routine (_pt_eval_function_type_new ()_) gets type of arguments and return type information from **func_all_signatures** struct defined at func_type.cpp file. From now, the count and count star function will not reach to here. That's why I put the _assert(false)_ here.

hgryoo: And when doing a constant folding for spacial cases (COUNT(1), COUNT(:x)... described at the comment above), there is no reason to set a return type here If there is nothing changed. Even this line is out of if-statement (it always set a return type).
So It seems a bug, I removed it.

ctshim: Thank you for your kind.

ctshim (approved):

src/query/query_analytic.cpp
```diff
@@ -782,7 +786,7 @@ qdata_finalize_analytic_func (cubthread::entry *thread_p, ANALYTIC_TYPE *func_p,
   /* set count-star aggregate values */
   if (func_p->function == PT_COUNT_STAR)
     {
-      db_make_int (func_p->value, func_p->curr_cnt);
+      db_make_bigint (func_p->value, func_p->curr_cnt + (int64_t) prm_get_bigint_value (PRM_ID_COUNT_DEBUG));
```
shparkcubrid, line 789: How about using 'defined(NDEBUG)'?
```
#if defined(NDEBUG)
db_make_bigint (func_p->value, func_p->curr_cnt);
#else 
db_make_bigint (func_p->value, func_p->curr_cnt + (int64_t) prm_get_bigint_value (PRM_ID_COUNT_DEBUG));
#endif 
```

src/query/query_aggregate.cpp
```diff
@@ -1219,7 +1219,8 @@ qdata_finalize_aggregate_list (cubthread::entry *thread_p, cubxasl::aggregate_li
       /* set count-star aggregate values */
       if (agg_p->function == PT_COUNT_STAR)
 	{
-	  db_make_int (agg_p->accumulator.value, agg_p->accumulator.curr_cnt);
+	  db_make_bigint (agg_p->accumulator.value,
+			  agg_p->accumulator.curr_cnt + (int64_t) prm_get_bigint_value (PRM_ID_COUNT_DEBUG));
```
byungwook-kim, line 1223: it's final routine for count, is it needed to accumulate debug count?

src/query/query_analytic.cpp
```diff
@@ -392,11 +396,11 @@ qdata_evaluate_analytic_func (cubthread::entry *thread_p, ANALYTIC_TYPE *func_p,
     case PT_COUNT:
       if (func_p->curr_cnt < 1)
 	{
-	  db_make_int (func_p->value, 1);
+	  db_make_bigint (func_p->value, 1);
```
byungwook-kim, line 399:
```suggestion
	  /* first value */
	 db_make_bigint (func_p->value, (int64_t) 1 + (int64_t) prm_get_bigint_value (PRM_ID_COUNT_DEBUG));
```

src/query/query_analytic.cpp
byungwook-kim: good point.

hgryoo: Thank you! I'll change it as you suggest

src/query/query_analytic.cpp
hgryoo: Thank you.

byungwook-kim (approved):

src/query/query_aggregate.cpp
```diff
@@ -999,11 +999,11 @@ qdata_evaluate_aggregate_optimize (cubthread::entry *thread_p, cubxasl::aggregat
     case PT_COUNT:
       if (agg_p->option == Q_ALL)
 	{
-	  db_make_int (agg_p->accumulator.value, oid_count - null_count);
+	  db_make_bigint (agg_p->accumulator.value, oid_count - null_count);
 	}
       else
 	{
-	  db_make_int (agg_p->accumulator.value, key_count);
+	  db_make_bigint (agg_p->accumulator.value, key_count);
```
shparkcubrid, line 1006: Oid_count, null_count, key_count are integer.
I think the aggregate_optimize function seems to be deprecated after version 10, so it is okay to keep this.

shparkcubrid:
Please attach a test case of this issue in jira.
Among SQL errors, what about checking the cause of the following errors cast by integer.

sql/_23_apricot_qa/_02_performance/_01_filtered_index/cases/filtered_index_data_type_basic_view.sql - linux_sql_64bit_release
+(select count(v_t?_a?c.b), count(v_t?_a?c.c), ( **cast(count(v_t?_a?c.a) as integer)**) from t? v_t?_a?c where (v_t?_a?c.a> ?:? ) and (v_t?_a?c.d> ?:? ) using index v_t?_a?c.i_t?_a?c(+))

youngjinj:
If the distinct count exceeds the INT_MAX value, 0 is returned as a result.
----
csql> select count(*) from t1;
              count(*)
======================
            2232889346
----
csql> select count(distinct c1) from t1;
    count(distinct c1)
======================
                     0
----
csql> select count(distinct c1) from (select c1 from t1 limit 1000);
    count(distinct c1)
======================
                  1000

hgryoo:
@youngjinj Thank you for the test. I'm going to look into the problem.

youngjinj:
> @youngjinj Thank you for the test. I'm going to look into the problem.

youngjinj:
Closing is my mistake. Sorry.

The issue of returning 0 as a result when the distinct count exceeds the INT_MAX value can be reproduced in the following way.

```
create table t1 (c1 bigint);
insert into t1 values (1), (2);
select count(distinct c1) from t1;

/* src/query/list_file.c, line 1499 - qfile_add_tuple_to_list_id */
gdb> b list_file.c:1501 if list_id_p->tuple_cnt == 1
gdb> set list_id_p->tuple_cnt = 2147483647
gdb> c
```

ctshim (approved):

shparkcubrid (approved):

youngjinj (approved):

PR #2668, opened 2021-04-05
[LETS-77] Do not use slotted pages' saved space on page server.

cristiarg:
http://jira.cubrid.org/browse/LETS-77

Extend condition to skip slotted page saved space to include also the situation when the server is in page server mode because this - slotted page saved space provisioning - has already been performed on the transaction server.

This occurred in stress testing of sequential replication executing on the page server.

### DIFF
--- a/src/storage/slotted_page.c
+++ b/src/storage/slotted_page.c
@@ -50,6 +50,7 @@
 #include "thread_entry.hpp"
 #include "thread_lockfree_hash_map.hpp"
 #include "thread_manager.hpp"	// for thread_get_thread_entry_info
+#include "server_type.hpp"
 
 #if !defined(SERVER_MODE)
 #define pthread_mutex_init(a, b)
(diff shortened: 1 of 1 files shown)


src/storage/slotted_page.c
```diff
@@ -233,6 +234,7 @@ static int spage_put_helper (THREAD_ENTRY * thread_p, PAGE_PTR pgptr, PGSLOTID s
 			     const RECDES * recdes, bool is_append);
 static void spage_add_contiguous_free_space (PAGE_PTR pgptr, int space);
 static void spage_reduce_contiguous_free_space (PAGE_PTR pgptr, int space);
+static bool spage_skip_save_space_altogether ();
```
ribram, line 237: Nitpicking: if you also agree a name like spage_is_save_space_enabled/disabled, please rename/update the code.

ribram (approved):

src/storage/slotted_page.c
cristiarg: Not nitpicking. It was a bad name.
Fixed.

PR #2672, opened 2021-04-07
[LETS-57] Page Server Parallel Replication

cristiarg:
http://jira.cubrid.org/browse/LETS-57

Implementation:
- introduce a new parameter `page_server_replication_parallel_count` that drives whether replication is to performed in parallel or sequentially; by default it is parallel with `8` tasks; setting the value of the parameter to `0` will render replication working sequentially;
- the parallel log apply infrastructure has been added in `cublog::replicator` (`log_replication.*`) where it essentially is chained to the `cublog::replicator` daemon task; it comes online before the task and is waited-for-to-finish and stop-destroyed after the daemon has been stopped; additionally, a soft-synch-point has been added on the retire handler of the daemon task (it may be that this is gratuitous and could work without it, but, can't hurt);
- for the dispatch of jobs, the same rules (read: workarounds) as for parallel log recovery have been added; so, the logic in `replicator::read_and_redo_record` is similar to the one in `log_recovery - log_rv_redo_record_sync_or_dispatch_async`
- `cublog::redo_job_impl`:
  - introduced a new `bool force_each_page_fetch` parameter that allows to drive `log_reader` to always fetch a log page; normally, if the `vpid` is the same, the page is assumed as being already "there" (i.e. up to date) and not fetched again; however, in the recovery case, it is needed to fetch the page with each use; this is, most likely, a workaround that needs a better solution as each fetch is a page-sized `memcpy` that affects performance;
  - in `redo_parallel::redo_job_base` introduced a `log_lsa` - needed for future development in tracking the progress of the asynchronous log replication
  - adapted existing unit tests for the above changes

Testing:
- manual stress-testing within a container enviroment using a custom testing script
- including start/stop/restart
- verification of data consistency by means of comparing the result of select queries executed  in stand-alone mode on the transaction server (TS) database and on the page server (PS) database

Diffing:
- diff for `src/transaction/log_replication.cpp` can be viewed locally with a command such as:
```
git difftool --dir-diff <upstream remote>/scalability_dev cristiarg/lets-57_page_server_replicate_parallel
```
  provided the local repo is up to date with all remotes


src/transaction/log_replication.cpp
```diff
@@ -1,173 +1,264 @@
...
+	    break;
+	  }
+
+	{
+	  std::unique_lock<std::mutex> lock (m_redo_lsa_mutex);
+	  m_redo_lsa = header.forw_lsa;
+	}
+
+        // TODO: why is this notification inside this loop? where this function is
+        // called, can we be sure that 'm_redo_lsa == nxio_lsa' ?
```
cristiarg, line 181: Either I did not understand or this notify is too early.

src/transaction/log_replication.cpp
```diff
@@ -1,173 +1,264 @@
...
+    // been dispatched for async replication;
+    // cannot decide if NO OTHER DATA will ever be fed because the daemon is still executing;
+    // will be able to tell that no data will ever be fed again for async replication
+    // only when the daemon is not running anymore;
+    // however, at this point introduce a fuzzy syncronization point by waiting all
+    // fed data to be effectively consumed/applied
+    if (m_parallel_replication_redo != nullptr)
+      {
+        m_parallel_replication_redo->wait_for_idle ();
+      }
```
cristiarg, line 262: I do not like the semantics of this function.
I added the bits for the async tasks but there is nothing in place that stops the daemon/async-tasks infrastructure resuming the consumption/application of data.
Actually the only place where we can be sure that the consumption has stopped is in the dtor where the daemon is stopped and the async tasks are waited to terminate work.

src/transaction/log_replication.cpp
```diff
@@ -1,173 +1,264 @@
...
+        m_daemon_task.reset (
+          new cubthread::entry_callable_task (std::move (func_exec), std::move (func_retire)));
+      }
+    else
+      {
+        // do not move task ownershit to itself - aka, do not let the task delete itself;
+        // it can be done, but prefer to do this locally for uniformity with the 'if' case
+        m_daemon_task.reset (
+          new cubthread::entry_callable_task (std::move (func_exec), false));
+      }
```
ribram, line 72: `wait_parallel_replication_idle()` can be called during retire in synchronous case too, and for now it would do nothing. It would help to add an else clause to the `if (m_parallel_replication_redo != nullptr)` condition and a comment explaining that in sync case nothing is required. Of course, the name of wait_parallel_replication_idle() 
 would have to be changed too :).
```suggestion
    auto func_retire = std::bind (&replicator::wait_parallel_replication_idle, std::ref (*this));
    m_daemon_task.reset (new cubthread::entry_callable_task (std::move (func_exec), std::move (func_retire)));
```

src/transaction/log_replication.cpp
ribram: Are you suggesting to move the notification to redo_upto_nxio_lsa? I think that would work too.

src/transaction/log_replication.cpp
```diff
@@ -1,173 +1,264 @@
...
+    // To allow reads on the page server, make sure that all changes are visible.
+    // Having log_Gl.hdr.mvcc_next_id higher than all MVCCID's in the database is a requirement.
+    MVCCID mvccid = log_rv_get_log_rec_mvccid (log_rec);
+    if (mvccid != MVCCID_NULL && !MVCC_ID_PRECEDES (mvccid, log_Gl.hdr.mvcc_next_id))
+      {
+	log_Gl.hdr.mvcc_next_id = mvccid;
+	MVCCID_FORWARD (log_Gl.hdr.mvcc_next_id);
+      }
+
+    const VPID rec_vpid = log_rv_get_log_rec_vpid<T> (log_rec);
```
ribram, line 207: Isn't the next part identical to log_rv_redo_record_sync_or_dispatch_async?

src/transaction/log_replication.cpp
ribram: In theory, the function is called after all log records have been generated. nxio_lsa should not increase anymore, so if m_redo_lsa catches up, that's it, it is done. In this context, the current function is enough.

It is not meant to be called while log records are still being generated. Stronger guarantees would be possible, by forcing daemon somehow not pickup anymore records.

Or maybe it is better to make the function purpose more clear, by renaming as wait_replication_finish_during_shutdown(). Add a comment inside that nxio_lsa is no longer expected to grow.

src/transaction/log_replication.hpp
```diff
@@ -52,16 +59,20 @@ namespace cublog
       void redo_upto (cubthread::entry &thread_entry, const log_lsa &end_redo_lsa);
       template <typename T>
       void read_and_redo_record (cubthread::entry &thread_entry, LOG_RECTYPE rectype, const log_lsa &rec_lsa);
+      void wait_parallel_replication_idle ();
 
+      std::unique_ptr <cubthread::entry_task> m_daemon_task;
```
ribram, line 64:
```suggestion
      std::unique_ptr<cubthread::entry_task> m_daemon_task;
```

src/transaction/log_replication.hpp
```diff
@@ -52,16 +59,20 @@ namespace cublog
...
       cubthread::daemon *m_daemon = nullptr;
 
       log_lsa m_redo_lsa = NULL_LSA;
-      mutable std::mutex m_redo_mutex;
+      mutable std::mutex m_redo_lsa_mutex;
       mutable std::condition_variable m_redo_condvar;
```
ribram, line 69: Then maybe this should be m_redo_lsa_condvar too.

src/transaction/log_replication.hpp
```diff
@@ -52,16 +59,20 @@ namespace cublog
...
 
       log_reader m_reader;
       LOG_ZIP m_undo_unzip;
       LOG_ZIP m_redo_unzip;
+
+      std::unique_ptr <cublog::redo_parallel> m_parallel_replication_redo;
```
ribram, line 75:
```suggestion
      std::unique_ptr<cublog::redo_parallel> m_parallel_replication_redo;
```

unit_tests/recovery/ut_database.cpp
```diff
@@ -18,14 +18,17 @@
...
+
 /*******************************************************
  * ut_database_values_generator
  *******************************************************/
 
 ut_database_values_generator::ut_database_values_generator (const ut_database_config &a_database_config)
   : m_database_config (a_database_config)
-  , m_lsa_log_id (10100) // just starts from an arbitrary value
   , m_gen (m_rd ())
+  //, m_log_lsa () // the log_lsa ctor actually does not set any value
```
ribram, line 30: About log_lsa ctor not actually setting any value, what do you think of making it set?

```
struct log_lsa
{
  std::int64_t pageid:48 = 0;		/* Log page identifier : 6 bytes length */
  std::int64_t offset:16 = 0;		/* Offset in page : 2 bytes length.
                                              offset == 'area offset' */
```

unit_tests/recovery/ut_database.cpp
```diff
@@ -34,11 +37,37 @@ ut_database_values_generator::ut_database_values_generator (const ut_database_co
...
+const log_lsa &ut_database_values_generator::increment_and_get_lsa_log ()
 {
-  return ++m_lsa_log_id;
+  const auto if_false_increment_page_id_else_increment_offset = m_rand_log_lsa_dist (m_gen);
+  if (if_false_increment_page_id_else_increment_offset == false)
+    {
+      // increment page id
+      assert (m_log_lsa.pageid < MAX_LOG_LSA_PAGEID);
+      ++m_log_lsa.pageid;
+      // just for fun, don't restart with 0-offset
```
ribram, line 54: Fun is fun, but it may confuse next reader :). If the database behavior is any indication, the offset should be reset to 0 when advancing to next page.
```suggestion
      ++m_log_lsa.pageid;
      // just for fun, don't restart with 0-offset
```

src/transaction/log_lsa.hpp
```diff
@@ -64,6 +64,9 @@ constexpr std::int64_t NULL_LOG_PAGEID = -1;
 constexpr std::int16_t NULL_LOG_OFFSET = -1;
 constexpr log_lsa NULL_LSA { NULL_LOG_PAGEID, NULL_LOG_OFFSET };
 
+constexpr std::int64_t MAX_LOG_LSA_PAGEID = (static_cast<std::int64_t> (1u) << (48 - 1)) - 1;
+constexpr std::int16_t MAX_LOG_LSA_OFFSET = (static_cast<std::int16_t> (1u) << (16 - 1)) - 1;
```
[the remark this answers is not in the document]
cristiarg: - 16 times it does overflow; 15 times it does not;
- `SHRT_MAX` has the same value as the one generated by the shift expression;
- nothing against changing for `MAX_LOG_LSA_OFFSET` but for `MAX_LOG_LSA_PAGEID` it would have to remain explicit as there is no built in expression
- I thought that having them laid out explicitly is nicer and more explicit;
- I will add a comment though

src/base/system_parameter.c
```diff
@@ -698,13 +698,15 @@ static const char sysprm_ha_conf_file_name[] = "cubrid_ha.conf";
...
 #define PRM_NAME_SERVER_TYPE "server_type"
-#define PRM_NAME_RECOVERY_PARALLEL_COUNT "recovery_parallel_count"
+#define PRM_NAME_PAGE_SERVER_REPLICATION_PARALLEL_COUNT "page_server_replication_parallel_count"
 
 #define PRM_NAME_ER_LOG_PRIOR_TRANSFER "er_log_prior_transfer"
 #define PRM_NAME_ER_LOG_COMM_REQUEST "er_log_comm_request"
 #define PRM_NAME_ER_LOG_COMM_CHANNEL "er_log_comm_channel"
 #define PRM_NAME_ER_LOG_COMMIT_CONFIRM "er_log_commit_confirm"
 
+#define PRM_NAME_RECOVERY_PARALLEL_COUNT "recovery_parallel_count"
```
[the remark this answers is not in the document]
cristiarg: Kept

src/base/system_parameter.c
```diff
@@ -698,13 +698,15 @@ static const char sysprm_ha_conf_file_name[] = "cubrid_ha.conf";
 
 #define PRM_NAME_PAGE_SERVER_HOSTS "page_server_hosts"
 #define PRM_NAME_SERVER_TYPE "server_type"
-#define PRM_NAME_RECOVERY_PARALLEL_COUNT "recovery_parallel_count"
+#define PRM_NAME_PAGE_SERVER_REPLICATION_PARALLEL_COUNT "page_server_replication_parallel_count"
```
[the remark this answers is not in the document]
cristiarg: Done

src/transaction/log_recovery.c
```diff
@@ -627,7 +627,8 @@ void log_rv_redo_record_sync_or_dispatch_async (THREAD_ENTRY *thread_p, log_read
     {
       // dispatch async
       using redo_job_impl_t = cublog::redo_job_impl<T>;
-      std::unique_ptr<redo_job_impl_t> job{ new redo_job_impl_t (rcv_vpid, rcv_lsa, end_redo_lsa, log_rtype) };
+      std::unique_ptr<redo_job_impl_t> job {
+        new redo_job_impl_t (rcv_vpid, rcv_lsa, end_redo_lsa, log_rtype, false) };
```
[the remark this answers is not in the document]
cristiarg: Done

src/transaction/log_replication.cpp
cristiarg: Initially it was like you suggest. Then I thought it is more explicit to only initialize what is needed when is needed.
Changed as suggested.

src/transaction/log_replication.cpp
cristiarg: It is. Except for the `#ifdef (SERVER_MODE)` bit

src/transaction/log_recovery.c
```diff
@@ -551,6 +550,7 @@ log_rv_fix_page_and_check_redo_is_needed (THREAD_ENTRY * thread_p, const VPID &
 }
 
 /* log_rv_need_sync_redo - force some of the redo records to be applied synchronously
+ *          depending on the function to be applied
```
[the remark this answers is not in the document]
cristiarg: Changed and also introduced the `VPID` check inside the function.
I still need help with a semantically strong&correct explanation for why those rcv-indices need to be executed in-sync.

src/transaction/log_replication.hpp
cristiarg: Done

src/transaction/log_replication.hpp
cristiarg: Done

src/transaction/log_replication.hpp
cristiarg: Done

unit_tests/recovery/ut_database.cpp
cristiarg: Nothing against. I was thinking conservatively to not alter a class whose behavior is key in many contexts.

unit_tests/recovery/ut_database.cpp
cristiarg: Corrected.

src/transaction/log_replication.cpp
cristiarg: Done.

src/transaction/log_replication.cpp
cristiarg: Moved.

unit_tests/recovery/ut_database.cpp
```diff
@@ -18,14 +18,17 @@
...
+
 /*******************************************************
  * ut_database_values_generator
  *******************************************************/
 
 ut_database_values_generator::ut_database_values_generator (const ut_database_config &a_database_config)
   : m_database_config (a_database_config)
-  , m_lsa_log_id (10100) // just starts from an arbitrary value
   , m_gen (m_rd ())
+    //, m_log_lsa () // the log_lsa ctor actually does not set any value
```
ribram, line 30: The original [discussion ](https://github.com/CUBRID/cubrid/pull/2672#discussion_r609110744)was deemed outdated by Github and it does not allow me to reply there.

 > constexpr std::int64_t NULL_LOG_PAGEID = -1;
constexpr std::int16_t NULL_LOG_OFFSET = -1;
constexpr log_lsa NULL_LSA { NULL_LOG_PAGEID, NULL_LOG_OFFSET };
I think we must use these instead as the starting values.

I agree.

src/server/page_server.cpp
```diff
@@ -174,7 +174,7 @@ page_server::finish_replication (cubthread::entry &thread_entry)
   assert (m_replicator != nullptr);
 
   logpb_force_flush_pages (&thread_entry);
-  m_replicator->wait_replication_finish ();
+  m_replicator->wait_replication_finish_during_shutdown ();
```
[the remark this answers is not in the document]
cristiarg: Done

src/transaction/log_recovery.c
```diff
@@ -550,22 +549,33 @@ log_rv_fix_page_and_check_redo_is_needed (THREAD_ENTRY * thread_p, const VPID &
...
+          case RVDK_FORMAT:
+          case RVDK_INITMAP:
+          case RVDK_RESERVE_SECTORS:
+          case RVDK_UNRESERVE_SECTORS:
+            return true;
+          default:
+            return false;
+          }
+      } ();
+// *INDENT-ON*
```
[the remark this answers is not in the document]
cristiarg: Done

unit_tests/recovery/ut_database.cpp
cristiarg: 1. It is not possible to initialize bitfield data members (apparently, it's a c++20 level feature).
2. If a default ctor is added, because `log_lsa` is included in other non-POD/trivial structs, many more cascading modifications are needed - explicit default ctors are needed. I am not sure that adding so many changes will not break smth.
Let's take this as a separate issue.

src/transaction/log_recovery_redo_parallel.hpp
```diff
@@ -319,4 +346,77 @@ namespace cublog
...
+				  undo_unzip_support, redo_unzip_support);
+#if defined(SERVER_MODE)
+    }
+  else
+    {
+      // dispatch async
+      using redo_job_impl_t = cublog::redo_job_impl<T>;
+      std::unique_ptr<redo_job_impl_t> job
+      {
+	new redo_job_impl_t (rcv_vpid, rcv_lsa, end_redo_lsa, log_rtype, force_each_log_page_fetch /*false*/)
```
ribram, line 415:
```suggestion
	new redo_job_impl_t (rcv_vpid, rcv_lsa, end_redo_lsa, log_rtype, force_each_log_page_fetch)
```

unit_tests/recovery/ut_database.cpp
ribram: Agreed.

ribram (approved):

src/transaction/log_recovery_redo_parallel.hpp

PR #2675, opened 2021-04-08
[CBRD-23908] Upgrade cppcheck version to 2.4.1 in Github Check

wrlawodms:
http://jira.cubrid.org/browse/CBRD-23908

I can't find an apt-repo to be able to install cppcheck over 1.9.0 version, so I've chosen to download the source and build manually.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -112,7 +112,21 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Install packages
-        run: sudo apt-get install -yq cppcheck 
+        run: |
+          sudo apt-get install -yq wget build-essential
+
+          # Download cppcheck-2.4.1 source
+          wd=`pwd`
+          cd ..
+          wget https://github.com/danmar/cppcheck/archive/2.4.1.tar.gz
+          tar xf 2.4.1.tar.gz
+
+          # build cppcheck-2.4.1
+          cd cppcheck-2.4.1
+          make -j4 SRCDIR=build CFGDIR=cfg CXXFLAGS="-O2 -DNDEBUG -Wall -Wno-sign-compare -Wno-unused-function"
+          echo "CPPCHECK_PATH=`pwd`" >> $GITHUB_ENV
+
+          cd $wd
       - uses: actions/checkout@v2
       - uses: jitterbit/get-changed-files@v1
         id: files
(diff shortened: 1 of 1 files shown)


.github/workflows/check.yml
```diff
@@ -112,7 +112,21 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Install packages
-        run: sudo apt-get install -yq cppcheck 
+        run: |
+          sudo apt-get install -yq wget build-essential
+
+          # Download cppcheck-2.4.1 source
+          wd=`pwd`
+          cd ..
```
wrlawodms, line 120: Because the checkout action deletes the working directory, it should be located in the other directory.

.github/workflows/check.yml
```diff
@@ -137,7 +151,7 @@ jobs:
             *) continue ;; # skip others
             esac
 
-            cppcheck --force -j4 --suppressions-list=.github/workflows/cppcheck-spr-list --quiet --enable=warning -iheaplayers $f 2>>$result_file
+            ${{ env.CPPCHECK_PATH }}/cppcheck --force -j4 --suppressions-list=.github/workflows/cppcheck-spr-list --quiet --enable=warning -iheaplayers $f 2>>$result_file
```
wrlawodms, line 154: I've run across some errors when "make install" cppcehck, so I use the local executable file which is built.

hgryoo (approved):

JoohoK (approved):

.github/workflows/check.yml
cristiarg: If using the local executable, please note that cppcheck also needs the `cfg` directory from the cppcheck source tree. If the directory is copied near the newly compiled executable, it works.

wrlawodms: Thank you for your advice. When I installed cppcheck by using "make install", the error you mentioned occurs, which says  "std.cfg not found". So, I just tried to use the local executable like this, and it seems to work. 

> If the directory is copied near the newly compiled executable
I don't quite follow you. Could you elaborate on it?

cristiarg: in the cppcheck source tree, there is a `cfg` directory with - I presume - default settings.
I found on SO a mention that the simplest to make it work (without a `make install`) is to just copy the `cfg` directory near the newly build binary:
```
~/tmp/cppcheck_build/bin$ tree
.
├── cfg
│   ├── avr.cfg
│   │   ...
│   ├── std.cfg
│   │   ...
│   └── zlib.cfg
└── cppcheck
```
This might not be relevant if the script does a `make install`.

wrlawodms: After building cppcheck in this script, the executable is created in the cppcheck source directory which includes the `cfg`. The location of the cppcheck source directory is assigned to `${{ env.CPPCHECK_PATH }}`. So, there seems to be no problem. What do you think?

cristiarg: Well, I understand now.
So, it worked because the build was done in-tree.
In that case, please disregard my comment.

wrlawodms: OK. Thank you for the review :).

PR #2677, opened 2021-04-09
[CBRD-23912] Display the build type correctly when it executes cubrid_rel

hgryoo:
http://jira.cubrid.org/browse/CBRD-23912


sjkim777 (approved):
Thanks for your brilliant problem-​solving ability.

PR #2678, opened 2021-04-09
[LETS-31] Add a class for all checkpoint information

adrianbrab:
http://jira.cubrid.org/browse/LETS-31

Checkpoint information is saved to disk as end checkpoint log record. The information must be moved to another file. A step to improve handling checkpoint is to create a class that holds all the required information and all required operations.

As the first part of the changes, a new class has been added along with the its own serialization method.


ribram (approved):

ribram (changes requested):
I accidentally approved. Please make the changes in the review first :).

ribram (approved):

PR #2679, opened 2021-04-09
[CBRD-23913] Allow the inline suppressions of cppcheck

wrlawodms:
http://jira.cubrid.org/browse/CBRD-23913

ㅉe are going to suppress each reported line if a developer assignee and reviewers agree an error on the line is a false positive. To do that, `--inline-suppr` option has to be turned.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -151,7 +151,7 @@ jobs:
             *) continue ;; # skip others
             esac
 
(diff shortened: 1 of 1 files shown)


wrlawodms:
I've tested it on my local repo (https://github.com/wrlawodms/cubrid/pull/23).

hgryoo (approved):

PR #2682, opened 2021-04-13
[CBRD-22334] Cursor Holdability on stmt, get wrong results by rs.absolute(num)

airnet73:
http://jira.cubrid.org/browse/CBRD-22334

Purpose
In order to transmit the query result to the client, the error of copying the data of page0 to page1 when reading data from the page in the following order has been corrected.
page0 -> page1 -> page0 

Implementation
N/A

Remarks
N/A

### DIFF
--- a/src/query/cursor.c
+++ b/src/query/cursor.c
@@ -996,11 +996,6 @@ cursor_buffer_last_page (CURSOR_ID * cursor_id_p, VPID * vpid_p)
 
   if (cursor_id_p->list_id.last_pgptr && VPID_EQ (&(cursor_id_p->list_id.first_vpid), vpid_p))
     {
-      if (cursor_id_p->buffer == NULL)
-	{
-	  return ER_FAILED;
-	}
-
       cursor_id_p->buffer = cursor_id_p->list_id.last_pgptr;
     }
   else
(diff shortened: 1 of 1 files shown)


src/query/cursor.c
```diff
@@ -1001,7 +1001,7 @@ cursor_buffer_last_page (CURSOR_ID * cursor_id_p, VPID * vpid_p)
 	  return ER_FAILED;
 	}
```
kisoo-han, line 1003: A direct pointer of the memory allocation for CURSOR ID is stored in **buffer_area**. Memory allocation/release is executed in association with pointer variable **buffer_area**. I think the pointer **buffer** is just a pointer variable to hold some CURSOR id buffer_area and perform cursor operations for it. Therefore, memcpy and free for the pointer **buffer** are not necessary and should not be done. It seems you got a great point on the bug fix. By the way, the above if statement that checks whether **buffer** is NULL doesn't seem meaningful and could be removed (It seems that someone is concerned that memcpy could fail).

kisoo-han (approved):

src/query/cursor.c
airnet73: Thank you for the review

byungwook-kim (approved):

PR #2683, opened 2021-04-13
[Lets-42] ATS waits confirmation commit from PS

danilailie:
http://jira.cubrid.org/browse/LETS-42

Added max_ps_flushed_lsa in log global to account for the latest lsa that's been committed by the page server.
Now all the threads that tries to flush lsa will also wait until the page server has flushed an lsa that's >= than this current thread's lsa.


src/transaction/log_impl.h
```diff
@@ -652,10 +654,16 @@ struct log_global
 #endif // SERVER_MODE = !SA_MODE
   // *INDENT-ON*
 
+  std::mutex ps_lsa_mutex;
+  std::condition_variable ps_lsa_cv;
+  LOG_LSA max_ps_flushed_lsa;
+
   // *INDENT-OFF*
```
ribram, line 661: All namespaced variables need to have indent disabled. Add m_ prefix to the names of member variables.
```suggestion
  // *INDENT-OFF*
  std::mutex m_ps_lsa_mutex;
  std::condition_variable m_ps_lsa_cv;
  LOG_LSA m_max_ps_flushed_lsa;

```

src/transaction/log_page_buffer.c
```diff
@@ -4099,6 +4099,17 @@ logpb_flush_pages (THREAD_ENTRY * thread_p, LOG_LSA * flush_lsa)
...
+    if (ats_Gl.is_page_server_connected ())
+        {
+            std::unique_lock<std::mutex> lock (log_Gl.ps_lsa_mutex);
+            log_Gl.ps_lsa_cv.wait (lock, [flush_lsa] { return log_Gl.max_ps_flushed_lsa >= *flush_lsa; });
+
+            if (prm_get_bool_value (PRM_ID_ER_LOG_COMMIT_CONFIRM))
+            {
+                _er_log_debug (ARG_FILE_LINE, "Page server committed LSA = %lld|%d.\n", LSA_AS_ARGS (&log_Gl.max_ps_flushed_lsa));
+            }
+        }
```
ribram, line 4112: - Indentantion still looks broken.
- Namespace & lambda syntax are not recognized by indent tool. Disable indent.
- Nitpicking: move the two lines (mutex lock guard and conditional wait) to a log_global function (e.g. `log_globla::wait_flushed_lsa (const log_lsa &flush_lsa)`);

src/transaction/log_global.c
```diff
@@ -103,4 +103,12 @@ log_global::~log_global ()
 {
   delete writer_info;
 }
+
+void
+log_global::update_max_ps_flushed_lsa (const LOG_LSA &lsa)
+{
+  max_ps_flushed_lsa = lsa;
+  std::unique_lock<std::mutex> lock(ps_lsa_mutex);
+  ps_lsa_cv.notify_all();
```
ribram, line 112: [condition_variable](https://en.cppreference.com/w/cpp/thread/condition_variable) documentation says to:

>The thread that intends to modify the shared variable has to
> 1. acquire a std::mutex (typically via std::lock_guard)
> 2. perform the modification while the lock is held
> 3. execute notify_one or notify_all on the std::condition_variable (the lock does not need to be held for notification)

```suggestion
  // *INDENT-OFF*
  std::unique_lock<std::mutex> lock (ps_lsa_mutex);
  // *INDENT-ON*
  max_ps_flushed_lsa = lsa;
  lock.unlock ();
  ps_lsa_cv.notify_all();
```

hornetmj (approved):

src/transaction/log_global.c
```diff
@@ -103,4 +103,18 @@ log_global::~log_global ()
...
+log_global::update_max_ps_flushed_lsa (const LOG_LSA &lsa)
+{
+  m_max_ps_flushed_lsa = lsa;
+  m_ps_lsa_cv.notify_all ();
+}
+
+void
+log_global::wait_flushed_lsa (const log_lsa &flush_lsa)
+{
+  std::unique_lock<std::mutex> lock (m_ps_lsa_mutex);
```
ribram, line 117: Is it ok to add a small optimization here, to check the condition first before locking:
```suggestion
  if (m_max_ps_flushed_lsa >= flush_lsa)
    {
      // already flushed
      return;
    }
  std::unique_lock<std::mutex> lock (m_ps_lsa_mutex);
```

ribram (approved):

PR #2688, opened 2021-04-14
[LETS-58] Page Server Replication Delay Statistics

cristiarg:
http://jira.cubrid.org/browse/LETS-58

Add a statistic for replication delay based on `at_time` logged on LOG_COMMIT, LOG_ABORT and LOG_DUMMY_HA_SERVER_STATE log records.

Implementation:
  - a new "PEEK" statistic has been added - `Time_scalability_replication_delay`
  - for better resolution of the delay values the interpretation/values of the `at_time` field of the `log_rec_donetime` (used for LOG_COMMIT/ABORT) and  `log_rec_ha_server_state` structures has been changed from seconds to milliseconds
  - `log_rec_donetime` function has been added to extract the current time in millseconds
  - in order not to alter existing usage of the values, where appropriate, the values have been truncated back to seconds semantics
  - the milliseconds semantic is only used for replication delay calculation - `eplicator::redo_upto` and `log_rpl_calculate_replication_delay`
  - when replication is executing in synchronous mode, the delay is calculated and logged synchronously
  - when replication is executing in async/parallel mode, a new "fake" job is created and dispatched (`redo_job_replication_delay_impl`) much like the actual workhorse `redo_job_impl`) and the delay is calculated and logged when the job is picked up for execution by the async mechanism; this gives an accurate estimation of the time it takes for the log to transferred and applied from the active transaction server (ATS) to the page server (PS)
  - added asserts and an explicit verification on the `perf_monitor` module as to the order of the statistics definitions in the `pstat_Metadata` array because dump functions non-explicitly imply that all "complex"-type statistics must appear after all non-"complex" statistics definitions

Caveats:
  - the newly introduced statistic is generated on the page server (PS), since it is not possible to connect to the PS in the regular way to dump statistics, the only way testing (evaluation) of the actual delay can be done is either by logging the values or by enabling the `stats_on` parameter and consulting the server log upon PS normal termination


src/base/perf_monitor.c
```diff
@@ -3011,6 +3022,22 @@ perfmon_initialize (int num_trans)
     }
 #endif
 
+  /* check proper ordering of statistics definitions:
+   *  - non-complex definitions come before the complex ones
+   *  - complex definitions all come at the end, just before PSTAT_COUNT
+   * needed because dump to file/buffer functions depend on this
+   * NOTE: this should actually be a unit test */
+  for (idx = 0; idx < PSTAT_COUNT && pstat_Metadata[idx].valtype != PSTAT_COMPLEX_VALUE; ++idx);
```
hornetmj, line 3030: What is the purpose of this for loop? I think this 'for' statement ends on this single line because of ';'. Is this correct what you intended? If it is, Plz explain more detail to me.

hornetmj: Sorry. I understand your intention.

src/transaction/log_replication.cpp
```diff
@@ -204,6 +229,32 @@ namespace cublog
...
+	std::unique_ptr<redo_job_replication_delay_impl_t> replication_delay_job
+	{
+	  new redo_job_replication_delay_impl_t (m_redo_lsa, start_time_msec, a_source)
+	};
+	m_parallel_replication_redo->add (std::move (replication_delay_job));
+      }
+    else
+      {
+	// calculate the time difference synchronously
+	log_rpl_calculate_replication_delay (&thread_entry, start_time_msec, a_source);
```
hornetmj, line 254: Even in this case, shouldn't LOG_COMMIT and LOG_ABORT log records are processed?

hornetmj (approved):

src/transaction/log_replication.cpp
cristiarg: I do not think I understant your question. Can you please elaborate?
Some explanations though:
The else statement performs the replication delay calculation & perfstat reporting when the replication is working in the synchronous mode (ie: when the param `PRM_ID_REPLICATION_PARALLEL_COUNT` is set to `0`). We decided to leave this infrastructure able to also work - with full functionality - in synchronous mode as well.

src/base/perf_monitor.c
cristiarg: Yes, it works because `idx` variable's scope is outside the loop and, thus, the compiler sees side effect and does not optimize it away (not even in release/optimized compilation mode).

src/transaction/log_replication.cpp
cristiarg: @hornetmj
After 3 days I finally undestand what you asked. They were not processed before I added handling to calculate the delay, so I suppose they do not need to be. I'm not knowledgeable in this matter but I suppose the following happens:
- the transactions are handled on the active transaction server and also introduce other log entries that are actually affect data which are properly processed;
- the COMMIT and ABORT log entries are only introduced such that certain bookkeeping is also performed (by analogy with the handling of `LOG_COMMIT` `LOG_ABORT` in log_recovery - `log_recovery_redo` function) and this bookkeeping is not needed on the page server;

@ribram
Please confirm this.

src/base/perf_monitor.c
ribram: My suggestion is to move the check to a separate function and make it obvious that it is meant for debug by using preprocessor directive `if !defined (NDEBUG)`

src/base/util_func.c
```diff
@@ -808,6 +809,35 @@ util_bsearch (const void *key, const void *base, int n_elems, unsigned int sizeo
...
+  gettimeofday (&now, NULL);
+
+  const int64_t msec_from_sec = now.tv_sec * 1000LL;
+  const int64_t msec_from_usec = now.tv_usec / 1000LL;
+  // *INDENT-OFF*
+  // overflow check
+  assert (std::numeric_limits<int64_t>::max () - msec_from_usec >= msec_from_sec);
+  // *INDENT-ON*
+  const int64_t msec = msec_from_sec + msec_from_usec;
+  return msec;
```
ribram, line 830: Suggestion: use <chrono> 
```suggestion
  // *INDENT-OFF*
  auto now_timept = std::chrono::system_clock::now ();
  return std::chrono::duration_cast<std::chrono::milliseconds> (now_timept .time_since_epoch ()).count ();
  // *INDENT-ON*
```

src/base/util_func.c
```diff
@@ -808,6 +809,35 @@ util_bsearch (const void *key, const void *base, int n_elems, unsigned int sizeo
...
+}
+
+/* util_msec_to_sec - truncate milliseconds to seconds
+ */
+time_t
+util_msec_to_sec (int64_t msec)
+{
+  return static_cast < time_t > (msec / 1000LL);
```
ribram, line 838:
```suggestion
  // *INDENT-OFF*
  return static_cast<time_t> (msec / 1000LL);
  // *INDENT-ON*
```

src/transaction/log_record.hpp
```diff
@@ -230,11 +230,16 @@ struct log_rec_replication
...
+
 /* Log the time of termination of transaction */
 typedef struct log_rec_donetime LOG_REC_DONETIME;
 struct log_rec_donetime
 {
-  INT64 at_time;		/* Database creation time. For safety reasons */
+  time_msec_t at_time;		/* Database creation time. For safety reasons .. LE: what?
+                                 * NOTE: stored as milliseconds with the value as accurate
+                                 * as possible as provided via system structs such as
+                                 * 'timeval' or 'timespec' */
```
ribram, line 242: The original comment seems outdated, replace it. Please keep comments brief.
```suggestion
  time_msec_t at_time;		/* Transaction commit time */
```

src/transaction/log_record.hpp
```diff
@@ -244,7 +249,10 @@ struct log_rec_ha_server_state
   int state;			/* ha_Server_state */
   int dummy;			/* dummy for alignment */
 
-  INT64 at_time;		/* time recorded by active server */
+  time_msec_t at_time;		/* time recorded by active server
+                                 * NOTE: stored as milliseconds with the value as accurate
+                                 * as possible as provided via system structs such as
+                                 * 'timeval' or 'timespec' */
```
ribram, line 255: I think the comment explanation is not useful for the future reader. The type `time_msec_t` is self-explanatory.
```suggestion
  time_msec_t at_time;		/* time recorded by active server */
```

src/transaction/log_recovery_redo_parallel.hpp
```diff
@@ -299,6 +315,43 @@ namespace cublog
       const log_reader::fetch_mode m_log_reader_page_fetch_mode;
   };
 
+
+  /* job implementation that performs log replication delay calculation
+   * using log records that register creation time
+   */
+  template <typename TYPE_LOG_REC> // TODO: does not need to be a template
```
ribram, line 322: Please fix the TODO :).

src/transaction/log_recovery_redo_parallel.hpp
```diff
@@ -299,6 +315,43 @@ namespace cublog
...
+  {
+      using log_rec_t = TYPE_LOG_REC;
+
+      /* sentinel VPID value needed for the internal mechanics of the parallel log recovery/replication
+       * internally, such a VPID is needed to maintain absolute order of the processing
+       * of the log records with respect to their order in the global log record
+       */
+      static constexpr short SENTINEL_VOLID = -2;
+      static constexpr int32_t SENTINEL_PAGEID = -2;
+      static constexpr vpid SENTINEL_VPID = { SENTINEL_PAGEID, SENTINEL_VOLID };
```
ribram, line 333: Better shorter? :D
```suggestion
      static constexpr vpid SENTINEL_VPID = { -2, -2 };
```

src/transaction/log_recovery_redo_parallel.hpp
```diff
@@ -337,6 +390,30 @@ namespace cublog
...
+  {
+  }
+
+  template <typename TYPE_LOG_REC>
+  int  redo_job_replication_delay_impl<TYPE_LOG_REC>::execute (THREAD_ENTRY *thread_p, log_reader &,
+      LOG_ZIP &, LOG_ZIP &)
+  {
+    const int res = log_rpl_calculate_replication_delay (thread_p, m_start_time_msec, m_source);
+    return res;
+  }
```
ribram, line 415: Move definition to .cpp after removing template.

src/transaction/log_recovery_redo_parallel.hpp
```diff
@@ -299,6 +315,43 @@ namespace cublog
...
+
+      redo_job_replication_delay_impl &operator = (redo_job_replication_delay_impl const &) = delete;
+      redo_job_replication_delay_impl &operator = (redo_job_replication_delay_impl &&) = delete;
+
+      int execute (THREAD_ENTRY *thread_p, log_reader &log_pgptr_reader,
+		   LOG_ZIP &undo_unzip_support, LOG_ZIP &redo_unzip_support) override;
+
+    private:
+      const time_t m_start_time_msec;
+      const char *const m_source;
```
ribram, line 351: Is source relevant? Can it be removed?
```suggestion
```

src/transaction/log_replication.cpp
ribram: Short answer, I confirm what @cristiarg says.

Long answer, as of now, the page server only replicated changes with effects on disk. The transaction table is not updated. Recovery also tracks transaction table changes, so it has some extra operations.

src/transaction/log_replication.cpp
```diff
@@ -225,4 +276,41 @@ namespace cublog
...
+   *
+   *  a_start_time_source: the name of the source where this
+   */
+  int
+  log_rpl_calculate_replication_delay (THREAD_ENTRY *thread_p, time_t a_start_time_msec,
+				       const char *a_start_time_source)
+  {
+    // skip calculation if bogus input (sometimes, it is -1);
+    // TODO: fix bogus input at the source if at all possible
+    if (a_start_time_msec > 0)
```
ribram, line 292: Does it happen to get bogus start time?

src/transaction/log_replication.hpp
```diff
@@ -77,6 +81,9 @@ namespace cublog
 
       std::unique_ptr<cublog::redo_parallel> m_parallel_replication_redo;
   };
+
+  extern int log_rpl_calculate_replication_delay (THREAD_ENTRY *thread_p, time_t a_start_time_msec,
```
ribram, line 85:
```suggestion
  extern int log_rpl_calculate_replication_delay (THREAD_ENTRY *thread_p, time_msec_t a_start_time_msec,
```

src/transaction/log_replication.cpp
```diff
@@ -225,4 +276,41 @@ namespace cublog
...
+      {
+	const int64_t end_time_msec = util_gettime_msec ();
+	const double time_diff_msec = difftime (end_time_msec, a_start_time_msec);
+	assert (time_diff_msec > .0);
+	perfmon_set_stat (thread_p, PSTAT_SC_REPL_DELAY, time_diff_msec, PSTAT_SC_REPL_DELAY_ALWAYS_COLLECT);
```
ribram, line 297:
```suggestion
	perfmon_set_stat (thread_p, PSTAT_SC_REPL_DELAY, time_diff_msec, false);

```

src/transaction/log_replication.cpp
```diff
@@ -225,4 +276,41 @@ namespace cublog
...
+
+#if !defined(NDEBUG)
+	const double time_diff_sec = time_diff_msec / 1000.0;
+	er_log_debug (ARG_FILE_LINE, "repl_delay(%s): %9g sec, %9.2f msec",
+		      a_start_time_source, time_diff_sec, time_diff_msec);
```
ribram, line 302: Suggestion for a separate patch: add a system parameter for replication error logging.

src/transaction/log_replication.cpp
```diff
@@ -225,4 +276,41 @@ namespace cublog
...
+  int
+  log_rpl_calculate_replication_delay (THREAD_ENTRY *thread_p, time_t a_start_time_msec,
+				       const char *a_start_time_source)
+  {
+    // skip calculation if bogus input (sometimes, it is -1);
+    // TODO: fix bogus input at the source if at all possible
+    if (a_start_time_msec > 0)
+      {
+	const int64_t end_time_msec = util_gettime_msec ();
+	const double time_diff_msec = difftime (end_time_msec, a_start_time_msec);
```
ribram, line 295: The perf statistics are UINT64. Better to keep directly the value in msec, instead of seconds represented as double.
```suggestion
	const UINT64 time_diff_msec = end_time_msec - a_start_time_msec;
```

src/base/perf_monitor.c
cristiarg: Done

src/base/util_func.c
cristiarg: Done

src/base/util_func.c
cristiarg: Done

src/transaction/log_record.hpp
cristiarg: Done

src/transaction/log_record.hpp
cristiarg: Removed

src/transaction/log_recovery_redo_parallel.hpp
cristiarg: Done

src/transaction/log_recovery_redo_parallel.hpp
cristiarg: Better.

src/transaction/log_recovery_redo_parallel.hpp
cristiarg: Obviously

src/transaction/log_recovery_redo_parallel.hpp
cristiarg: I think it is.
I noticed that the COMMIT/ABORT delays are quite homogenous. But the HA_REPL one is wildly varying - sometimes similar to COMMIT/ABORT, othertimes much higher values, in the seconds.
TODO: I'll provide a logging snapshot.

src/transaction/log_replication.cpp
cristiarg: Sometimes yes, that is one of the two reasons I added the `m_source`.
TODO: add logging snapshot

src/transaction/log_replication.hpp
cristiarg: Done

cristiarg: Done

src/transaction/log_replication.cpp
cristiarg: Removed

src/transaction/log_replication.cpp
cristiarg: Done

src/transaction/log_replication.cpp
cristiarg: Done

src/base/perf_monitor.c
```diff
@@ -563,6 +565,9 @@ PSTAT_METADATA pstat_Metadata[] = {
   PSTAT_METADATA_INIT_SINGLE_PEEK (PSTAT_PB_AVOID_DEALLOC_CNT, "Num_data_page_avoid_dealloc"),
   PSTAT_METADATA_INIT_SINGLE_PEEK (PSTAT_PB_AVOID_VICTIM_CNT, "Num_data_page_avoid_victim"),
 
+  /* Scalability statistics */
+  PSTAT_METADATA_INIT_SINGLE_PEEK (PSTAT_SC_REPL_DELAY, "Time_scalability_replication_delay"),
```
ribram, line 569:
```suggestion
  PSTAT_METADATA_INIT_SINGLE_PEEK (PSTAT_SC_REPL_DELAY, "Time_scalability_replication_delay_msec"),
```

src/transaction/log_replication.cpp
```diff
@@ -26,20 +26,25 @@
...
     : m_redo_lsa { start_redo_lsa }
   {
     log_zip_realloc_if_needed (m_undo_unzip, LOGAREA_SIZE);
     log_zip_realloc_if_needed (m_redo_unzip, LOGAREA_SIZE);
 
+    // before initializing any threads, init stats
+    cubthread::entry *const thread_entry = &cubthread::get_entry ();
+    // initialization is always performed
+    perfmon_set_stat (thread_entry, PSTAT_SC_REPL_DELAY, 0, false);
+
```
ribram, line 47: All stats are initialized as 0, not required.
```suggestion
```

src/transaction/log_replication.cpp
```diff
@@ -164,13 +169,25 @@ namespace cublog
...
+	    const log_rec_dbout_redo dbout_redo = m_reader.reinterpret_copy_and_add_align<log_rec_dbout_redo> ();
 	    log_rcv rcv;
 	    rcv.length = dbout_redo.length;
 	    log_rv_redo_record (&thread_entry, m_reader, RV_fun[dbout_redo.rcvindex].redofun, &rcv, &m_redo_lsa, 0,
 				nullptr, m_redo_unzip);
+	    break;
 	  }
-	  break;
+	  case LOG_COMMIT:
+	    calculate_replication_delay_or_dispatch_async<struct log_rec_donetime> (
```
ribram, line 180: Using struct before type name was required by old C, no longer required for C++.
```suggestion
	    calculate_replication_delay_or_dispatch_async<log_rec_donetime> (
```

src/transaction/log_replication.cpp
```diff
@@ -164,13 +169,25 @@ namespace cublog
...
+		    thread_entry, m_redo_lsa);
+	    break;
+	  case LOG_ABORT:
+	    calculate_replication_delay_or_dispatch_async<struct log_rec_donetime> (
```
ribram, line 184:
```suggestion
	    calculate_replication_delay_or_dispatch_async<log_rec_donetime> (
```

src/transaction/log_replication.cpp
```diff
@@ -164,13 +169,25 @@ namespace cublog
...
+		    thread_entry, m_redo_lsa);
+	    break;
+	  case LOG_DUMMY_HA_SERVER_STATE:
+	    calculate_replication_delay_or_dispatch_async<struct log_rec_ha_server_state> (
```
ribram, line 188:
```suggestion
	    calculate_replication_delay_or_dispatch_async<log_rec_ha_server_state> (
```

src/transaction/log_replication.cpp
ribram: If this is an outlier event, I think we can just ignore it. I don't see anything that can be done with this error.

src/transaction/log_replication.hpp
```diff
@@ -77,6 +81,8 @@ namespace cublog
 
       std::unique_ptr<cublog::redo_parallel> m_parallel_replication_redo;
   };
+
+  extern int log_rpl_calculate_replication_delay (THREAD_ENTRY *thread_p, time_t a_start_time_msec);
```
ribram, line 85: extern not needed.
```suggestion
  int log_rpl_calculate_replication_delay (THREAD_ENTRY *thread_p, time_t a_start_time_msec);
```

src/transaction/log_replication.cpp
```diff
@@ -225,4 +267,33 @@ namespace cublog
...
+  {
+    // skip calculation if bogus input (sometimes, it is -1);
+    // TODO: fix bogus input at the source if at all possible (debugging revealed that
+    // it happens for LOG_COMMIT messages only)
+    if (a_start_time_msec > 0)
+      {
+	const int64_t end_time_msec = util_get_time_as_ms_since_epoch ();
+	const int64_t time_diff_msec = end_time_msec - a_start_time_msec;
+
+	perfmon_set_stat (thread_p, PSTAT_SC_REPL_DELAY, static_cast<int> (time_diff_msec), false);
```
ribram, line 286:
```suggestion
	perfmon_set_stat (thread_p, PSTAT_SC_REPL_DELAY, static_cast<UINT64> (time_diff_msec), false);

```

src/base/perf_monitor.c
cristiarg: Done

src/transaction/log_replication.cpp
cristiarg: Done

src/transaction/log_replication.cpp
cristiarg: Done

src/transaction/log_replication.hpp
cristiarg: Done

src/transaction/log_replication.cpp
cristiarg: but
```
STATIC_INLINE void
perfmon_set_stat (THREAD_ENTRY * thread_p, PERF_STAT_ID psid, int statval, bool always_collect)
111

ribram: ah, you're right.

ribram (approved):

PR #2695, opened 2021-04-19
[CBRD-23911] Implement supplemental logging

JoohoK:

**Purpose**
Add two information in the log record, in order to support CDC interfaces. 
1. Transaction user
2. undo image for delete log records

**Implementation**
1. Transaction user
* Add new log record type : LOG_REC_SUPPLEMENT_TRAN_USER 
* User info is logged at the begin and the end of the transaction
- begin :  
-- before head_lsa is updated at first from NULL. (prior_lsa_next_record()) 
-- transaction with DDL at first , is started logging with log_append_savepoint() 
-- transaction with DML at first, is started logging with log_append_undo/redo/undoredo_crumbs()
- end : 
-- right before log_commit or log_abort

2. Undo in delete log
* Add argument _RECDES * undo_recdes_ to heap_mvcc_log_delete () 

**Remarks**
N/A


JohnNasse:
cpp check errors will be fixed when http://jira.cubrid.org/browse/CBRD-23898 is merged

JoohoK:
cpp check errors will be fixed when http://jira.cubrid.org/browse/CBRD-23898 is merged

src/transaction/log_manager.c
```diff
@@ -3380,6 +3395,11 @@ log_append_savepoint (THREAD_ENTRY * thread_p, const char *savept_name)
       return NULL;
     }
 
+  if (prm_get_bool_value (PRM_ID_SUPPLEMENTAL_LOG) == true)
+    {
+      log_append_supplement_user_at_start (thread_p, tdes, LOG_PRIOR_LSA_WITHOUT_LOCK);
```
wrlawodms, line 3400: Could you let me know why is it needed in `log_append_savepoint ()`? If this is because you want to catch all the points `head_lsa` is initialized, it seems that you need to put it into all other `append_log_...()` function like log_append_ha_server_state(), log_append_empty_record(), log_append_postpone(), and so on.

wrlawodms: Technically, all `log_append_..()` functions that eventually call `prior_lsa_next_record_internal()`.

src/storage/heap_file.c
```diff
@@ -15868,7 +15869,15 @@ heap_mvcc_log_delete (THREAD_ENTRY * thread_p, LOG_DATA_ADDR * p_addr, LOG_RCVIN
     }
   else
     {
-      log_append_undoredo_data (thread_p, rcvindex, p_addr, 0, redo_data_size, NULL, redo_data_p);
+      if (undo_recdes != NULL)
+	{
```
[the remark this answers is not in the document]
JoohoK: It sounds good, thanks I'll apply it

src/transaction/log_manager.c
JoohoK: DDL statements like 'CREATE', and 'ALTER' tables, indexes calls log_append_savepoint() at first. 
( ex) do_alter() -> tran-system_savepoint() -> log_append_savepoint() )
Other DML statements calls log_append_undo/redo/undoredo_crumbs. 
I just want to catch the moment that transaction starts with DDL or DML.

src/transaction/log_manager.c
```diff
@@ -4738,9 +4761,56 @@ log_append_commit_log_with_lock (THREAD_ENTRY * thread_p, LOG_TDES * tdes, LOG_L
...
+{
+  LOG_REC_SUPPLEMENT_TRAN_USER *spplmnt_usr;
+  LOG_PRIOR_NODE *node;
+
+  assert (prm_get_bool_value (PRM_ID_SUPPLEMENTAL_LOG) == true);
+
+  node = prior_lsa_alloc_and_copy_data (thread_p, LOG_SUPPLEMENT_TRAN_USER, RV_NOT_DEFINED, NULL, 0, NULL, 0, NULL);
+  if (node == NULL)
+    {
+      return;
```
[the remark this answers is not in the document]
JoohoK: It sounds good, thanks

src/transaction/log_manager.c
```diff
@@ -4738,9 +4761,56 @@ log_append_commit_log_with_lock (THREAD_ENTRY * thread_p, LOG_TDES * tdes, LOG_L
...
+    }
+
+  spplmnt_usr = (LOG_REC_SUPPLEMENT_TRAN_USER *) node->data_header;
+  memcpy (spplmnt_usr->user_name, tdes->client.get_db_user (), DB_MAX_USER_LENGTH);
+  if (with_lock == LOG_PRIOR_LSA_WITH_LOCK)
```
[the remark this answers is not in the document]
JoohoK: LOG_PRIOR_LSA_WITH_LOCK was intended acquired when supplement log is appended while replication log is appending using prior_lsa_mutex. (in log_append_repl_info_and_commit_log())
```
prior_lsa_mutex.lock
log_append_repl_info_with_lock()
log_append_supplement_user()
log_append_commit_log_with_lock()
prior_lsa_mutex.unlock()
```
However, it is not necessary to add supplement log like above code. supplement log will be added before replication logs are appended. So I will remove the case (LOG_PRIOR_LSA_WITH_LOCK == true) and argument. 
Sorry for less explanation in previous meeting.

src/transaction/log_manager.c
```diff
@@ -4710,6 +4729,10 @@ log_change_tran_as_completed (THREAD_ENTRY * thread_p, LOG_TDES * tdes, LOG_RECT
 static void
 log_append_commit_log (THREAD_ENTRY * thread_p, LOG_TDES * tdes, LOG_LSA * commit_lsa)
 {
+  if (prm_get_bool_value (PRM_ID_SUPPLEMENTAL_LOG) == true)
+    {
+      log_append_supplement_user (thread_p, tdes, LOG_PRIOR_LSA_WITHOUT_LOCK);
```
wrlawodms, line 4734: It seems to be put into `log_append_donetime_internal()` with this in `log_append_abort_log()`.

src/transaction/log_manager.c
wrlawodms: What if a HA status log is appended before starting DDL or DML? It seems to be unsafe to assume all DDL will do Savepoint before starting DDL operation. Savepoint is just one of the interfaces given to a user by SM, and It might not be used. We should assume some operations don't use savepoint although now all use it. The way I see it, you're making a restriction all the DDL has to use Savepoint.

wrlawodms: What if a user do other jobs that doesn't call log_append_savepoint() or `log_append_undo/redo..()` and then do DDL or DML operation in a transaction?

src/transaction/log_manager.c
JoohoK: OK, Thanks. It could be replaced.

src/transaction/log_manager.c
wrlawodms: 1. OK
2. No, the USER I said is not a database user, but all clients that can call SM interface like CAS.

wrlawodms: Plus, even though all DDLs use Savepoint when starting DDL, I think it's just clients (CAS, csql and so on) use them together now and SM should assume they are called independently.

JoohoK: OK, I understand what you mean. I also need to consider that one. thanks

src/storage/heap_file.c
```diff
@@ -21412,7 +21421,14 @@ heap_delete_home (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * context, boo
 	  rec_address.pgptr = context->home_page_watcher_p->pgptr;
 	  rec_address.vfid = &context->hfid.vfid;
 	  rec_address.offset = context->oid.slotid;
-	  heap_mvcc_log_delete (thread_p, &rec_address, RVHF_MVCC_DELETE_REC_HOME);
+	  if (prm_get_bool_value (PRM_ID_SUPPLEMENTAL_LOG) == true)
+	    {
+	      heap_mvcc_log_delete (thread_p, &rec_address, RVHF_MVCC_DELETE_REC_HOME, &context->home_recdes);
```
hornetmj, line 21426: The heap_mvcc_log_delete() function is also called in case of context->record_type == REC_ASSIGN_ADDRESS. In this case, how to deal with it?

src/transaction/log_manager.c
```diff
@@ -4738,9 +4761,56 @@ log_append_commit_log_with_lock (THREAD_ENTRY * thread_p, LOG_TDES * tdes, LOG_L
...
+      (void) prior_lsa_next_record_with_lock (thread_p, node, tdes);
+    }
+  else
+    {
+      (void) prior_lsa_next_record (thread_p, node, tdes);
+    }
+}
+
+static void
+log_append_supplement_user_at_start (THREAD_ENTRY * thread_p, LOG_TDES * tdes, int with_lock)
```
hornetmj, line 4806: I can't find the use case of 'with_lock != LOG_PRIOR_LSA_WITHOUT_LOCK'. So, is it necessary to use this parameter?

hornetmj, line 4806: comment; How about removing the '_at_start' part in the function name?

src/transaction/log_manager.c
```diff
@@ -4738,9 +4761,56 @@ log_append_commit_log_with_lock (THREAD_ENTRY * thread_p, LOG_TDES * tdes, LOG_L
...
+    }
   log_append_donetime_internal (thread_p, tdes, abort_lsa, LOG_ABORT, LOG_PRIOR_LSA_WITHOUT_LOCK);
 }
 
+/*
+ * log_append_supplement_user - append supplement user log record 
+ *
+ * return: nothing
+ *
+ *   tdes(in/out): State structure of transaction containing transaction user.
```
hornetmj, line 4776: It needs an explanation of other parameters.

src/transaction/log_manager.c
hornetmj: I think that this approach(check it every log_append_* function call) is needed to reconsider. Please, find a better place than now.

PR #2696, opened 2021-04-19
[CBRD-23912] Revise printing build type string and Fix slip for utility

hgryoo:
http://jira.cubrid.org/browse/CBRD-23912

1) QA requests "Coverage" and "Profile" prints like "coverage debug" and "profile debug"
2) Print "RelWithDebInfo" as "release"
3) Print the build type string when starting utility (e.g. createdb)


sjkim777 (approved):
I have checked
- release ver on Windows, linux. 
- debug ver on linux.
- coverage ver on linux. 
All looks good.

PR #2697, opened 2021-04-20
[LETS-86] Replace timespec_get with chrono functions

ribram:
http://jira.cubrid.org/browse/LETS-86

On CentOS devtoolset-8 system compile fails because timespec_get and TIME_UTC are not found, even though the C++17 documentation includes them.

Quick fix by using std::chrono::now() and building the seconds/millisecond parts that were initially generated using the deprecated function ftime.

Some of the changes in db_vdb.c made by #2646 have been reverted.

### DIFF
--- a/src/base/util_func.h
+++ b/src/base/util_func.h
@@ -27,12 +27,9 @@
 
 #ident "$Id$"
 
-#if defined __cplusplus
-#include <chrono>
-#include <ctime>
-#endif
 #include <sys/types.h>
 #include <math.h>
+#include <time.h>
 
 #define UTIL_PID_ENVVAR_NAME         "UTIL_PID"
 #define UTIL_infinity()     (HUGE_VAL)
@@ -77,12 +74,6 @@ extern int util_log_write_command (int argc, char *argv[]);
 extern int util_bsearch (const void *key, const void *base, int n_elems, unsigned int sizeof_elem,
 			 int (*func_compare) (const void *, const void *), bool * out_found);
 
-// *INDENT-OFF*
-#if defined __cplusplus
-template <typename Duration>
-void util_get_seconds_and_rest_since_epoch (std::chrono::seconds &secs, Duration &rest);
-void util_get_second_and_ms_since_epoch (time_t &secs, int &msec);
-#endif
-// *INDENT-ON*
+extern void util_get_second_and_ms_since_epoch (time_t * secs, int *msec);
 
 #endif /* _UTIL_FUNC_H_ */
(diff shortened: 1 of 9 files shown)


src/base/util_func.h
```diff
@@ -73,4 +77,12 @@ extern int util_log_write_command (int argc, char *argv[]);
 extern int util_bsearch (const void *key, const void *base, int n_elems, unsigned int sizeof_elem,
 			 int (*func_compare) (const void *, const void *), bool * out_found);
 
+// *INDENT-OFF*
+#if defined __cplusplus
+template <typename Duration>
+void util_get_seconds_and_rest_since_epoch (std::chrono::seconds &secs, Duration &rest);
+void util_get_second_and_ms_since_epoch (time_t &secs, int &msec);
```
cristiarg, line 84: `util_get_second_and_ms_since_epoch` function should not be guarded by `__cplusplus`, I think.
`util_get_seconds_and_rest_since_epoch` maybe should not be be exposed at all since not used externally. And the header to remain pure C?

adrianbrab (approved):

src/compat/db_vdb.c
```diff
@@ -398,6 +399,21 @@ db_open_file_name (const char *name)
...
+// *INDENT-OFF
+static void
+db_calculate_current_time (struct timeb &tb)
+{
+  // use chrono functions to populate timeb
+  time_t sec;
+  int millisec;
+  util_get_second_and_ms_since_epoch (sec, millisec);
+  tb =
+  {
```
cristiarg, line 411: alignment

src/base/util_func.h
ribram: To make it pure C, util_get_second_and_ms_since_epoch signature should be changed to `(time_t * secs, int * msec)`. Should be ok though :)

src/compat/db_vdb.c
ribram: done

src/base/util_func.h
ribram: done

cristiarg (approved):
I was not "dead" serious with "C-only".
And I was not aware that C does not allow for address-of params.
Maybe I should have written "template-free" iso "C-only".
Sorry for the confusion.
Feel free to change the signature either way you consider better.

ribram:
> I was not "dead" serious with "C-only".
> And I was not aware that C does not allow for address-of params.
> Maybe I should have written "template-free" iso "C-only".
> Sorry for the confusion.
> Feel free to change the signature either way you consider better.

Since the header is exposed to files compiled with C, I went with your first suggestion :).

PR #2699, opened 2021-04-20
[LETS-38] secont parameter of rvfun.fun_t should be a pointer to const

adrianbrab:
http://jira.cubrid.org/browse/LETS-38

Right now, the parameter is a pointer to non-const and some of the applying functions might be tempted to - and some do - modify the data (be it because it is easier or more performant than first copying and then modifying it).

After the changes, all the usages have been changed to const and a copy was created where necessary.
The changed to const also revealed a bug in an assertion that has been fixed now.


src/storage/disk_manager.c
```diff
@@ -3855,15 +3855,17 @@ disk_rv_reserve_sectors (THREAD_ENTRY * thread_p, LOG_RCV * rcv)
       /* disk check is in progress. unfix page, get critical section again and then fix page and make the changes */
       VPID vpid;
       pgbuf_get_vpid (rcv->pgptr, &vpid);
-      pgbuf_unfix_and_init (thread_p, rcv->pgptr);
+      pgbuf_unfix (thread_p, rcv->pgptr);
+      memset (rcv->pgptr, 0, sizeof (char *));
```
ribram, line 3859: - Please check pgbug_unfix_and_init again. It doesn't reset the data in page, it only resets the page pointer as NULL.
- What this function does, unfixing & fixing page again, is quite a tricky work-around. But we have no choice at this time but to keep it. Add a comment and strip the const.
```suggestion
      // to avoid a deadlock between CSECT_DISK_CHECK and page fix, we are forced to unfix the page here.
      // make an exception to the const rule.
      // *INDENT-OFF*
      pgbuf_unfix (thread_p, (const_cast<LOG_RCV *> (rcv))->pgptr);
      // *INDENT-ON*
```

src/storage/disk_manager.c
```diff
@@ -3855,15 +3855,17 @@ disk_rv_reserve_sectors (THREAD_ENTRY * thread_p, LOG_RCV * rcv)
...
       error_code = csect_enter_as_reader (thread_p, CSECT_DISK_CHECK, INF_WAIT);
       if (error_code != NO_ERROR)
 	{
 	  assert_release (false);
 	  return ER_FAILED;
 	}
       /* fix page again */
-      rcv->pgptr = pgbuf_fix (thread_p, &vpid, OLD_PAGE, PGBUF_LATCH_WRITE, PGBUF_UNCONDITIONAL_LATCH);
+      memcpy (rcv->pgptr, pgbuf_fix (thread_p, &vpid, OLD_PAGE, PGBUF_LATCH_WRITE, PGBUF_UNCONDITIONAL_LATCH),
+	      sizeof (char *));
```
ribram, line 3868:
```suggestion
      // strip const again
      // *INDENT-OFF*
      (const_cast<LOG_RCV *> (rcv))->pgptr = pgbuf_fix (thread_p, &vpid, OLD_PAGE, PGBUF_LATCH_WRITE, PGBUF_UNCONDITIONAL_LATCH);
      // *INDENT-ON*
```

src/storage/disk_manager.c
```diff
@@ -3936,15 +3938,17 @@ disk_rv_unreserve_sectors (THREAD_ENTRY * thread_p, LOG_RCV * rcv)
       /* disk check is in progress. unfix page, get critical section again and then fix page and make the changes */
       VPID vpid;
       pgbuf_get_vpid (rcv->pgptr, &vpid);
-      pgbuf_unfix_and_init (thread_p, rcv->pgptr);
+      pgbuf_unfix (thread_p, rcv->pgptr);
```
ribram, line 3941: Same as disk_rv_reserve_sectors

ribram (changes requested):

ribram:
@adrianbrab make sure code is updated with latest scalability_dev.

ribram (approved):

cristiarg (approved):

PR #2700, opened 2021-04-20
[LETS-32] Add new log volume to store meta log

ribram:
http://jira.cubrid.org/browse/LETS-32

Add a new volume for the meta log (checkpoint info). In this patch, the meta log volume is empty. The actual writing of meta log (and checkpoint info) will be done in a subsequent task.

The new file is named dbname_lgmeta, similarly to other log files and it is stored in the configured log directory. The file is created during createdb and deleted during deletedb. It is copied during copydb and is backed-up/restored.

The meta log is loaded from file when log is initialized. Not finding the meta log file is acceptable when emergency initialization is called (e.g. deletedb).

Did manual testing for deletedb/createdb, shutdown. A shell_debug test is underway.


CMakeLists.txt
```diff
@@ -314,7 +314,7 @@ else(UNIX)
   set(CMAKE_SHARED_LINKER_FLAGS_RELWITHDEBINFO "/INCREMENTAL:NO ${replaceIncFlag}" )
 
   #set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /WX /wd4820 /showIncludes" )
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /W3 /wd4820" )
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /W3 /wd4820 /std:c++17")
```
adrianbrab, line 317: This change is already present on cubrid/scalabilty_dev, maybe a merge will make it not show anymore and fix the conflict

adrianbrab (approved):

danilailie (approved):

PR #2719, opened 2021-04-22
[Lets 65] Transaction server fetches log page from page server

danilailie:
http://jira.cubrid.org/browse/LETS-65

Created async_log_page_receiver that receives log pages from page server and keeps them, so the workers will receive them from this class asynchronously.
Added a small optimization - same page cannot be requested multiple times now.
Added a unit test.


src/server/active_tran_server.cpp
```diff
@@ -153,6 +153,22 @@ bool active_tran_server::is_page_server_connected () const
...
+
+void active_tran_server::finalize_log_page_receiver ()
+{
+  m_async_log_page_receiver.reset ();
+}
+
+cublog::async_log_page_receiver &
+active_tran_server::get_log_page_receiver ()
+{
+  return *m_async_log_page_receiver;
```
ribram, line 169:
```suggestion
  assert (m_async_log_page_receiver != nullptr);
  return *m_async_log_page_receiver;
```

src/transaction/async_log_page_receiver.hpp
```diff
@@ -0,0 +1,55 @@
...
+#include <unordered_set>
+
+namespace cublog
+{
+
+  class async_log_page_receiver
+  {
+    public:
+      async_log_page_receiver ();
+      ~async_log_page_receiver ();
```
ribram, line 38: If ctor/dtor does not do anything, better to add = default. Or let them be implicitly defined.
```suggestion
      async_log_page_receiver () = default;
      ~async_log_page_receiver () = default;
```

src/transaction/async_log_page_receiver.hpp
```diff
@@ -0,0 +1,55 @@
...
+
+      void set_page_requested (LOG_PAGEID log_pageid);
+      bool is_page_requested (LOG_PAGEID log_pageid);
+      std::shared_ptr<LOG_PAGE> wait_for_page (LOG_PAGEID log_pageid);
+      void set_page (std::shared_ptr<LOG_PAGE> log_page);
+    private:
```
ribram, line 44: For readability, add an empty line between sections. In this case, before private section.
```suggestion
      void set_page (std::shared_ptr<LOG_PAGE> log_page);

    private:
```

src/transaction/async_log_page_receiver.hpp
```diff
@@ -0,0 +1,55 @@
...
+  class async_log_page_receiver
+  {
+    public:
+      async_log_page_receiver ();
+      ~async_log_page_receiver ();
+
+      void set_page_requested (LOG_PAGEID log_pageid);
+      bool is_page_requested (LOG_PAGEID log_pageid);
+      std::shared_ptr<LOG_PAGE> wait_for_page (LOG_PAGEID log_pageid);
+      void set_page (std::shared_ptr<LOG_PAGE> log_page);
```
ribram, line 43: What do you think of moving the shared_ptr to async_log_page_receiver?
```suggestion
      void set_page (std::shared_ptr<LOG_PAGE> &&log_page);
```

src/transaction/async_log_page_receiver.hpp
```diff
@@ -0,0 +1,55 @@
...
+    private:
+      std::mutex m_log_pages_mutex;
+      std::condition_variable m_pages_cv;
+      std::unordered_map<LOG_PAGEID, int> m_requested_page_id_count;
+      std::unordered_map<LOG_PAGEID, std::shared_ptr<LOG_PAGE>> m_received_log_pages;
+  };
+
+
```
ribram, line 51: One empty line is enough.
```suggestion

```

src/transaction/async_log_page_receiver.hpp
```diff
@@ -0,0 +1,55 @@
...
+} // namespace cublog
+
+
```
ribram, line 54:
```suggestion

```

src/transaction/async_log_page_receiver.hpp
```diff
@@ -0,0 +1,55 @@
...
+      void set_page_requested (LOG_PAGEID log_pageid);
+      bool is_page_requested (LOG_PAGEID log_pageid);
+      std::shared_ptr<LOG_PAGE> wait_for_page (LOG_PAGEID log_pageid);
+      void set_page (std::shared_ptr<LOG_PAGE> log_page);
+    private:
+      std::mutex m_log_pages_mutex;
+      std::condition_variable m_pages_cv;
+      std::unordered_map<LOG_PAGEID, int> m_requested_page_id_count;
+      std::unordered_map<LOG_PAGEID, std::shared_ptr<LOG_PAGE>> m_received_log_pages;
```
ribram, line 48: What do you think of merging these two maps into a single one. For each requested LOG_PAGEID, there would be a pair of requested count and page.
```suggestion
      std::unordered_map<LOG_PAGEID, int> m_requested_page_id_count;
      std::unordered_map<LOG_PAGEID, std::shared_ptr<LOG_PAGE>> m_received_log_pages;
```

src/transaction/log_page_buffer.c
```diff
@@ -1988,6 +1995,31 @@ logpb_copy_page (THREAD_ENTRY * thread_p, LOG_PAGEID pageid, LOG_CS_ACCESS_MODE
...
+    {
+      ats_Gl.push_request (ats_to_ps_request::SEND_LOG_PAGE_FETCH, std::move (message));
+
+      if (prm_get_bool_value (PRM_ID_ER_LOG_READ_LOG_PAGE))
+        {
+          _er_log_debug (ARG_FILE_LINE, "Sent request for log to Page Server. Page ID: %lld \n", log_pageid);
+        }
+    }
+  
+  ats_Gl.get_log_page_receiver ().set_page_requested (log_pageid);
```
ribram, line 2018: This is not entirely safe because it allows the possibility of multiple requests for the same log_pageid. Take the next scenario with thread T1 and T2, making a request for log page LP.

- receiver has no request for LP initially
- T1 checks is_page_requested (LP) and gets false.
- T2 checks is_page_requested (LP) and gets false.
- T1 & T2 push separate requests for same LP
- T1 & T2 register a single entry in log page receiver, with two requesters.
- Receiver will get two copies of LP from the page server, because two requests are sent. After receiving the first copy, it serves them to both T1 & T2 and erases the map entries for LP.
- The second copy is received, but there is no requester left to serve.

There are some possible fixes for this case (e.g. discard the second copy).

src/transaction/log_storage.hpp
```diff
@@ -62,6 +62,14 @@ const size_t LOGPB_BUFFER_NPAGES_LOWER = 128;
 typedef struct log_hdrpage LOG_HDRPAGE;
 struct log_hdrpage
 {
+  bool operator== (const log_hdrpage &other)
```
ribram, line 65: Move definition to log_storage.cpp

unit_tests/log_transfer/test_main_async_log_page_receiver.cpp
```diff
@@ -0,0 +1,125 @@
...
+}
+
+void
+test_env::run_test ()
+{
+  for (int i = 0; i <= m_log_pages_count - count_pages_per_thread; i += count_skip_pages)
+    {
+      std::thread worker (&test_env::request_and_consume_log_pages, this, i, count_pages_per_thread);
+      m_threads.push_back (std::move (worker));
+    }
```
ribram, line 79: 1. Please add short comments explaining the purpose of the test/functions.
2. I am not sure the current test gives enough opportunity for the threads to overlap requests. Although the page ranges that each thread do overlap, I don't know if they will really overlap. I think a new case that gives them more opportunity to overlap is required.

src/transaction/async_log_page_receiver.hpp
danilailie: Makes very little difference.

danilailie: And this is not really in line with how the shared_ptr should be used.

src/transaction/log_page_buffer.c
danilailie: I will ignore the log pages received from PS if there's no request ongoing for them (IE the requests have been fulfilled).

src/transaction/async_log_page_receiver.hpp
danilailie: This makes the code harder to read and contribute to, with little (if any) gain.

unit_tests/log_transfer/CMakeLists.txt
```diff
@@ -140,3 +140,37 @@ target_link_libraries(test_async_log_page_fetcher fake_thread_utils)
...
+
+set(TEST_ASYNC_LOG_PAGE_RECEIVER_SOURCES
+  ${TRANSACTION_DIR}/async_log_page_receiver.cpp
+  test_main_async_log_page_receiver.cpp)
+
+SET_SOURCE_FILES_PROPERTIES(
+  ${TEST_ASYNC_LOG_PAGE_RECEIVER_SOURCES}
+  PROPERTIES LANGUAGE CXX)
+
+add_executable(test_async_log_page_receiver
```
ribram, line 156: Add test_async_log_page_receiver to unit test running script.

src/server/active_tran_server.cpp
```diff
@@ -173,9 +190,14 @@ void active_tran_server::receive_log_page (cubpacking::unpacker &upk)
...
     {
-      char log_page_buffer[IO_MAX_PAGE_SIZE + MAX_ALIGNMENT];
-      LOG_PAGE *log_page = reinterpret_cast<LOG_PAGE *> (PTR_ALIGN (log_page_buffer, MAX_ALIGNMENT));
+
+      char *log_page_buffer = new char [db_log_page_size ()];
+      LOG_PAGE *log_page = reinterpret_cast<LOG_PAGE *> (log_page_buffer);
       std::memcpy (log_page, message.c_str () + sizeof (error_code), db_log_page_size ());
+
+      std::shared_ptr<LOG_PAGE> shared_log_page (log_page);
+      m_async_log_page_receiver->set_page (std::move (shared_log_page));
```
ribram, line 199: Setting char[] to std::shared_ptr<LOG_PAGE> may not be safe. Please find a way to make sure memory allocator/deallocator match.

ribram (commented):
Do you know the reason for the submodule apparent changes?

src/transaction/async_log_page_receiver.hpp
danilailie: Github is acting funny. There is only one empty line at the end of the file. On my honor!

src/transaction/log_storage.hpp
```diff
@@ -81,10 +83,27 @@ typedef struct log_page LOG_PAGE;
...
+    explicit log_page_wrapper (const char *buffer);
+    ~log_page_wrapper ();
+
+    bool operator== (const log_page_wrapper &other);
+    bool operator== (const LOG_PAGE &other);
+    const LOG_HDRPAGE &get_header () const;
+
+  private:
+    char *m_buffer;
+    LOG_PAGE *m_log_page;
```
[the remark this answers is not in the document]
danilailie: Tried this change and it gave a really tough time to implement the == operator and also since I'm not really sure that the start address of a std::string is the same as it's char array...the memcpy might be more difficult to implement. .c_str() doesn't work because it's a const.

src/server/active_tran_server.hpp
```diff
@@ -45,6 +46,11 @@ class active_tran_server
     void disconnect_page_server ();
     bool is_page_server_connected () const;
 
+    void init_page_broker ();
+    void finalize_page_broker ();
+
+    cublog::page_broker &get_page_broker ();
```
ribram, line 52:
```suggestion
    void init_log_page_broker ();
    void finalize_log_page_broker ();

    cublog::page_broker &get_log_page_broker ();
```

src/server/active_tran_server.hpp
```diff
@@ -59,6 +65,7 @@ class active_tran_server
     std::unique_ptr<page_server_conn> m_ps_conn;
     std::unique_ptr<page_server_request_queue> m_ps_request_queue;
     std::unique_ptr<page_server_request_autosend> m_ps_request_autosend;
+    std::unique_ptr<cublog::page_broker> m_page_broker;
```
ribram, line 68:
```suggestion
    std::unique_ptr<cublog::page_broker> m_log_page_broker;
```

src/transaction/log_page_broker.hpp
```diff
@@ -0,0 +1,60 @@
...
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+
+#ifndef _ASYNC_LOG_PAGE_RECEIVER_HPP_
+#define _ASYNC_LOG_PAGE_RECEIVER_HPP_
```
ribram, line 20:
```suggestion
#ifndef _LOG_PAGE_BROKER_HPP_
#define _LOG_PAGE_BROKER_HPP_
```

src/transaction/log_page_broker.hpp
```diff
@@ -0,0 +1,60 @@
...
+    private:
+      std::mutex m_log_pages_mutex;
+      std::condition_variable m_pages_cv;
+      std::unordered_map<LOG_PAGEID, int> m_requested_page_id_count;
+      std::unordered_map<LOG_PAGEID, std::shared_ptr<log_page_owner>> m_received_log_pages;
+  };
+
+} // namespace cublog
+
+#endif // _ASYNC_LOG_PAGE_RECEIVER_HPP_
```
ribram, line 60:
```suggestion
#endif // _LOG_PAGE_BROKER_HPP_
```

src/transaction/log_page_buffer.c
```diff
@@ -1989,6 +1996,29 @@ logpb_copy_page (THREAD_ENTRY * thread_p, LOG_PAGEID pageid, LOG_CS_ACCESS_MODE
...
+#if defined(SERVER_MODE)
+  // *INDENT-OFF*
+  constexpr size_t BIG_INT_SIZE = 8;
+  char buffer[BIG_INT_SIZE];
+  std::memcpy (buffer, &log_pageid, sizeof (log_pageid));
+  std::string message (buffer, BIG_INT_SIZE);
+
+  if (ats_Gl.get_page_broker ().register_entry (log_pageid) == cublog::page_broker::ADDED_ENTRY)
+    {
+      ats_Gl.push_request (ats_to_ps_request::SEND_LOG_PAGE_FETCH, std::move (message));
```
ribram, line 2011:
```suggestion
      // First to add an entry must also sent the request to the page server
      ats_Gl.push_request (ats_to_ps_request::SEND_LOG_PAGE_FETCH, std::move (message));
```

unit_tests/log_transfer/CMakeLists.txt
```diff
@@ -108,35 +108,70 @@ add_dependencies(test_prior_sendrecv
...
+)
+
+#
+# test_log_page_broker
+# 
+
+set(TEST_LOG_PAGE_BROKER_SOURCES
+  ${TRANSACTION_DIR}/log_storage.cpp
+  ${TRANSACTION_DIR}/log_page_broker.cpp
+  test_main_async_log_page_broker.cpp)
```
ribram, line 151:
```suggestion
  test_main_log_page_broker.cpp)
```

ribram (approved):

PR #2722, opened 2021-04-23
[CBRD-23898] Fix for formatted print

hgryoo:
http://jira.cubrid.org/browse/CBRD-23898

the NULL checking for the _buf_ should be performed only for unformatted print.
it caused by #2673. Sorry for missing it while the code review.

### DIFF
--- a/src/loaddb/load_object.c
+++ b/src/loaddb/load_object.c
@@ -486,11 +486,11 @@ start:
       va_end (ap);
     }
 
-  if (nbytes > 0 && buf != NULL)
+  if (nbytes > 0)
     {
       if (nbytes < size)
 	{			/* OK */
-	  if (buflen > 0)
+	  if (buflen > 0 && buf != NULL)
 	    {			/* unformatted print */
 	      memcpy (tout->ptr, buf, buflen);
 	      *(tout->ptr + buflen) = '\0';	/* Null terminate */
--- a/src/loaddb/load_object.c
+++ b/src/loaddb/load_object.c
@@ -490,8 +490,9 @@ start:
     {
       if (nbytes < size)
 	{			/* OK */
-	  if (buflen > 0 && buf != NULL)
+	  if (buflen > 0)
 	    {			/* unformatted print */
+	      assert (buf != NULL);
 	      memcpy (tout->ptr, buf, buflen);
 	      *(tout->ptr + buflen) = '\0';	/* Null terminate */
 	    }


src/loaddb/load_object.c
```diff
@@ -486,11 +486,11 @@ text_print (TEXT_OUTPUT * tout, const char *buf, int buflen, char const *fmt, ..
...
 
-  if (nbytes > 0 && buf != NULL)
+  if (nbytes > 0)
     {
       if (nbytes < size)
 	{			/* OK */
-	  if (buflen > 0)
+	  if (buflen > 0 && buf != NULL)
 	    {			/* unformatted print */
 	      memcpy (tout->ptr, buf, buflen);
```
wrlawodms, line 495: It seems to change the existing semantic. The changed code skips `memcpy()` if buf  is NULL, but it just assumed `buf` is not NULL if `buflen > 0` in the previous code. 
```suggestion
	  if (buflen > 0)
	  {
	     assert (buf != NULL)
```

src/loaddb/load_object.c
```diff
@@ -486,11 +486,11 @@ text_print (TEXT_OUTPUT * tout, const char *buf, int buflen, char const *fmt, ..
...
     }
 
-  if (nbytes > 0 && buf != NULL)
+  if (nbytes > 0)
     {
       if (nbytes < size)
 	{			/* OK */
-	  if (buflen > 0)
+	  if (buflen > 0 && buf != NULL)
 	    {			/* unformatted print */
```
wrlawodms, line 494: It seems to change the existing semantic. The changed code skips `memcpy()` if buf  is NULL, but it just assumed `buf` is not NULL if `buflen > 0` in the previous code. How about this below:
```suggestion
	  if (buflen > 0)
	    {			/* unformatted print */
	      assert (buf != NULL);
```

wrlawodms: I'm not sure if assert() is better than just null point exception, and it can pass cppcheck.

hgryoo: I think passing a null pointer to `memcpy()` is unsafe according to the following reference.

https://en.cppreference.com/w/cpp/string/byte/memcpy
> If either dest or src is an invalid or null pointer, the behavior is undefined, even if count is zero.

Here is another discussion about passing the null pointer argument to `memcpy ()`.
https://stackoverflow.com/questions/5243012/is-it-guaranteed-to-be-safe-to-perform-memcpy0-0-0

wrlawodms: Yes. It has to be avoided, so assert it. If it is skipped because `buf` is null, it may behave unexpectedly.

wrlawodms (approved):

src/loaddb/load_object.c
cristiarg: Sorry for not spotting it either.
This function has an input invariant that looks like the folowing:
```
assert((buf == NULL && buflen ==0) || (buf != NULL && buflen > 0));
```

Also, I would also add a new indirection for this function in the form (pseudocode):
```
int
text_print_fixed_buffer (TEXT_OUTPUT * tout, const char *buf, char const *fmt, ...)
{
  assert(buf != NULL);
  return (tout, buf, strlen(buf), fmt, ...);
}

```
because there are some issues.
For example there is one usage which I think is wrong (~1070):
```
	  if (i < len - 1)
	    {
	      CHECK_PRINT_ERROR (text_print (tout, ",\n ", 2, NULL));
	    }
```
and usage of the type:
```
	  CHECK_PRINT_ERROR (text_print
			     (tout, (left_nbytes <= 0) ? "'\'+\n \'" : "'", (left_nbytes <= 0) ? 6 : 1, NULL));
```
is error prone.
But maybe this is not for this PR.

cristiarg (approved):

PR #2733, opened 2021-04-26
[LETS-93] Proper transaction index for async log page fetcher threads

cristiarg:
http://jira.cubrid.org/browse/LETS-93

Fix for, when perfmon logging is activated, the tasks dispatched by the `async_page_fetcher` fail upon perfmon calls because of null `tran_index`.

### DIFF
--- a/src/thread/thread_entry_task.hpp
+++ b/src/thread/thread_entry_task.hpp
@@ -27,15 +27,14 @@
 #error Wrong module
 #endif // not SERVER_MODE and not SA_MODE
 
+#include "thread_entry.hpp"
 #include "thread_task.hpp"
+#include "transaction_global.hpp"
 
 #include <cassert>
 
 namespace cubthread
 {
-
-  // forward definition
-  class entry;
 
   // cubthread::entry_task
   //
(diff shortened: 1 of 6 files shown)


src/transaction/log_page_async_fetcher.cpp
```diff
@@ -49,11 +49,15 @@ namespace cublog
...
   async_page_fetcher::async_page_fetcher ()
   {
-    cubthread::manager *thread_manager = cubthread::get_manager ();
+    cubthread::manager *const thread_manager = cubthread::get_manager ();
 
     const auto thread_count = std::thread::hardware_concurrency ();
-    m_threads = thread_manager->create_worker_pool (thread_count, thread_count, "async_page_fetcher_worker_pool",
-		nullptr, thread_count, false /*debug_logging*/);
+    m_worker_pool_context_manager.reset (
+	    new cubthread::system_worker_entry_manager (TT_WORKER));
```
ribram, line 56: Although the name may not be fortunate, TT_WORKER is the type of threads that execute USER operations. For system operations, the default type would be TT_DAEMON.

It is probably better to add a new thread type: TT_SYSTEM_WORKER.

cristiarg: Added.
I pasted in the explanations above as comments.
But the semantics & separation is not really clear and could be better.

src/thread/thread_entry_task.hpp
```diff
@@ -133,6 +132,31 @@ namespace cubthread
...
+  //      thread type and transaction index
+  //    - useful in scenarios where there is no actual transaction and perf logging
+  //      still needs a non-null transaction index
+  //
+  class system_worker_entry_manager : public entry_manager
+  {
+    public:
+      system_worker_entry_manager (thread_type a_thread_type);
+      system_worker_entry_manager (const system_worker_entry_manager &) = delete;
+      system_worker_entry_manager (system_worker_entry_manager &&) = delete;
```
danilailie, line 148: Have you considered to make these move-able? Like std::thread?

cristiarg: There's no need currently to have them moveable. So, instead they are explicitly restricted.
The usage pattern is that the `cubthread::manager::create_worker_pool` take a raw pointer wo ownership (it does have an internal `m_entry_manager` which it owns) and thus the - currently - single use case does not mandate move semantics.
To make it more streamlined, changes would need to start in `cubthread::manager`.
Did I get your comment straight?

danilailie: Yes, you got it right.

src/transaction/log_page_async_fetcher.cpp
ribram: True, it could, but I think your comments are enough for now :).

ribram (approved):

PR #2737, opened 2021-04-27
[LETS-78] Measure the performance of page server replication

cristiarg:
http://jira.cubrid.org/browse/LETS-78

The replication system performance is critical. It has to keep up with everything active transaction throws at it. The system performance needs to be measured and monitored, to find problems and optimize.

Implementation:
- added two helper classes specifically for counter timer statistisc value type: `perfmon_manual_tracker_counter_timer` and `perfmon_raii_tracker_counter_timer`
- added 3 new counter timers statistics:
- `PSTAT_SCAL_REPL_LOG_REDO_SYNC`: perf data for processing log redo on the page server:
  - it covers the synchronous part;
  - if the infrastructure to apply recovery log redo in parallel is used, it does not include the calling of the redo function as that part will be included in the 'PSTAT_LOG_REDO_ASYNC' couterpart logging
  - if the log redo is applied synchronously, these values will include the effective calling of the redo function (which is also caught by `PSTAT_LOG_REDO_FUNC_EXEC` and, thus, a reader would need to take this fact into account);
- `PSTAT_LOG_REDO_ASYNC`: perf data for processing log redo asynchronously:
  - recorded also during log crash recovery
  - recorded on the page server, when replication is executing in the asynchronous mode
  -  in both cases, it does include the part that effectively calls the redo function, so, for accurate evaluation the part that effectively executes the redo function - `PSTAT_LOG_REDO_FUNC_EXEC` - must be subtracted from this one
- `PSTAT_LOG_REDO_FUNC_EXEC`: perf data for actually applying the log redo; it is relevant in two contexts:
  - log recovery redo after a crash (either synchronously or using the parallel infrastructure)
  - log replication on the page server; both applying the replication redo synchronously or in parallel will log to this entry, so interpreting this must be done in conjunction with the way replication has been performed
- added 2 more thread types: `TT_RECOVERY` and `TT_REPLICATION` which are used to identify thread pool threads (via a `entry_manager` implementation supplied to the thread pool); used dependent of context in `redo_parallel`

Unrelated change:
- `request_server::register_request_handler`: do not allow adding the same handler more than once; also removed double entry in `page_server::set_active_tran_server_connection`


src/transaction/log_page_async_fetcher.cpp
```diff
@@ -38,10 +38,6 @@ namespace cublog
       async_page_fetcher::callback_func_type m_callback;
   };
 
-  /*********************************************************************
```
ribram: Any reason that this comment was removed?

src/base/perf_monitor.h
```diff
@@ -634,8 +635,11 @@ typedef enum
   PSTAT_PB_AVOID_DEALLOC_CNT,
   PSTAT_PB_AVOID_VICTIM_CNT,
 
-  /* Scalability statistics */
-  PSTAT_SC_REPL_DELAY,
+  /* Scalability (and recovery) related statistics */
+  PSTAT_SCAL_REPL_DELAY,
+  PSTAT_SCAL_REPL_LOG_REDO_SYNC,
```
ribram, line 640: I think a better naming here would be "Redo replication" instead of scalability. Replication is just one part of a bigger mechanism that provides the scalable property (albeit an important part).
```suggestion
  /* Redo recovery and replication statistics */
  PSTAT_REDO_REPL_DELAY,
  PSTAT_REDO_REPL_LOG_REDO_SYNC,
```

src/base/perf_monitor_trackers.hpp
```diff
@@ -0,0 +1,75 @@
...
+#define PERF_MONITOR_HPP
+
+#include "perf_monitor.h"
+
+/* manual perf stat monitoring utility for counter timer values
+ *
+ * NOTE: the global 'perfmon_is_perf_tracking' is only checked in the ctor
+ * and used for the rest of the lifetime as is
+ */
+class perfmon_manual_tracker_counter_timer
```
ribram, line 33: Can it be named without "manual"? Also, "perfmon counter timer tracker" sounds better (more natural in English)
```suggestion
class perfmon_counter_timer_tracker (PERF_STAT_ID a_stat_id);
```

Shouldn't default constructor be deleted?

src/base/perf_monitor_trackers.hpp
```diff
@@ -0,0 +1,75 @@
...
+    TSC_TICKS m_start_tick;
+};
+
+
+/* RAII perf stat monitoring utility for counter timer values
+ *
+ * NOTE: the global 'perfmon_is_perf_tracking' is only checked in the ctor
+ * and used for the rest of the lifetime as is
+ */
+class perfmon_raii_tracker_counter_timer final : private perfmon_manual_tracker_counter_timer
```
ribram, line 68: - English natural wording would be perfmon counter timer RAII tracker
- Use composition when you can, private inheritance when you have to ([isocpp guideline](https://isocpp.org/wiki/faq/private-inheritance#priv-inherit-vs-compos))
- Delete default constructor

src/thread/thread_entry.cpp
```diff
@@ -707,8 +707,10 @@ thread_type_to_string (thread_type type)
       return "VACUUM_MASTER";
     case TT_VACUUM_WORKER:
       return "VACUUM_WORKER";
-    case TT_SYSTEM_WORKER:
-      return "SYSTEM_WORKER";
+    case TT_RECOVERY:
+      return "RECOVERY";
+    case TT_REPLICATION:
+      return "REPLICATION";
```
ribram, line 713:
```suggestion
    case TT_RECOVERY:
      return "RECOVERY";
    case TT_REPLICATION:
      return "REPLICATION";
    case TT_SYSTEM_WORKER:
      return "SYSTEM_WORKER";
```

src/transaction/log_recovery_redo_parallel.hpp
```diff
@@ -52,6 +52,10 @@ namespace cublog
       class redo_task;
 
     public:
+      /* - worker_count: the number of parallel tasks to spin that consume jobs
+       * - pool_context_manager: can be used to control the identity of the parallel
+       *    tasks that consume jobs
+       */
```
ribram, line 58:
```suggestion
      /* - worker_count: the number of parallel tasks to spin that consume jobs
       */
```

src/transaction/log_replication.cpp
```diff
@@ -36,10 +36,16 @@ namespace cublog
...
     : m_redo_lsa { start_redo_lsa }
+    , m_perfmon_log_processing { PSTAT_SCAL_REPL_LOG_REDO_SYNC }
   {
     log_zip_realloc_if_needed (m_undo_unzip, LOGAREA_SIZE);
     log_zip_realloc_if_needed (m_redo_unzip, LOGAREA_SIZE);
 
+    // in case replication is applied synchronously, the perf subsystem needs
+    // to have a non-negative transation index on all threads
+    cubthread::entry &thread_entry = cubthread::get_entry ();
+    thread_entry.tran_index = LOG_SYSTEM_TRAN_INDEX;
```
ribram, line 47: thread_entry.tran_index is not supposed to be set here. This is the thread that creates the replicator, probably the main thread during boot, or another thread.

The thread_entry that needs to be set to LOG_SYSTEM_TRAN_INDEX is the thread_entry of m_daemon. Give m_daemon a context manager that sets tran_index to LOG_SYSTEM_TRAN_INDEX.

Also, imo the replicator daemon thread should also be a TT_REPLICATION.

src/transaction/log_replication.cpp
```diff
@@ -130,9 +136,12 @@ namespace cublog
 
     // redo all records from current position (m_redo_lsa) until end_redo_lsa
 
+    m_perfmon_log_processing.start ();
```
ribram, line 139: - Rename m_perfmon_log_processing => m_perfmon_redo_sync?
- Why not start & track only once per redo_upto call? Imo the important information is here is the total time spent on the sync execution. It can inform us whether the sync operations are a bottleneck.

ribram (commented):
After going through the code, I think one more stat may be useful: the overall time spent in log_rv_redo_record_sync. I suggest a new PSTAT & a raii tracker there :).

cs/CMakeLists.txt
```diff
@@ -150,6 +150,7 @@ set(BASE_HEADERS
   ${BASE_DIR}/packable_object.hpp
   ${BASE_DIR}/packer.hpp
   ${BASE_DIR}/perf.hpp
+  ${BASE_DIR}/perf_monitor.h
   ${BASE_DIR}/perf_def.hpp
```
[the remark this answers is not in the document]
cristiarg: Done

cristiarg:
> After going through the code, I think one more stat may be useful: the overall time spent in log_rv_redo_record_sync. I suggest a new PSTAT & a raii tracker there :).

I am thinking that everything that happens in-synch should be accounted for "under the same roof" - served by `PSTAT_SCAL_REPL_LOG_REDO_SYNC`. The function `log_rv_redo_record_sync` is just a chunk of that.

What would be really useful would be to have the following as completely disjoint pieces (now we have an amalgamation of this, which is not really easy to "see through"):
1) in main thread: the time spent preparing to execute the redo function in sync
2) in main thread: the time spent preparing to dispatch the redo function async (the same code as above, but executed for a different purpose)
3) in async task: the time spent to prepare for redo function execution (esentially part of the same actions as for 1) - and would be useful to see whether too much resources are spent twice)
4) in main thread: the time spent executing the redo function (continuation of 1) )
5) in async task: the time spent executing the redo function (continuation of 3) )
1) and 2) might prove too coarse and, for them, finer grained pieces could be logged as well (eg: `pgptr` fixing)

Doing this requires passing some context info between the involved code.
Shall I attempt this?

src/base/perf_monitor.c
```diff
@@ -566,7 +566,31 @@ PSTAT_METADATA pstat_Metadata[] = {
   PSTAT_METADATA_INIT_SINGLE_PEEK (PSTAT_PB_AVOID_VICTIM_CNT, "Num_data_page_avoid_victim"),
 
   /* Scalability statistics */
-  PSTAT_METADATA_INIT_SINGLE_PEEK (PSTAT_SC_REPL_DELAY, "Time_scalability_replication_delay_msec"),
+  /* Peek time in msec taken for the log to be transferred, processed an applied on the page server */
+  PSTAT_METADATA_INIT_SINGLE_PEEK (PSTAT_SCAL_REPL_DELAY, "Time_scal_replication_delay_msec"),
+  /* perf data for processing log redo on the page server - the synchronous part
```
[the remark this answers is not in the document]
cristiarg: Done. Moved the comment at the usage location.

ribram:
> > After going through the code, I think one more stat may be useful: the overall time spent in log_rv_redo_record_sync. I suggest a new PSTAT & a raii tracker there :).
> 
> I am thinking that everything that happens in-synch should be accounted for "under the same roof" - served by `PSTAT_SCAL_REPL_LOG_REDO_SYNC`. The function `log_rv_redo_record_sync` is just a chunk of that.
> 
> What would be really useful would be to have the following as completely disjoint pieces (now we have an amalgamation of this, which is not really easy to "see through"):
> 
> 1. in main thread: the time spent preparing to execute the redo function in sync
> 2. in main thread: the time spent preparing to dispatch the redo function async (the same code as above, but executed for a different purpose)
> 3. in async task: the time spent to prepare for redo function execution (esentially part of the same actions as for 1) - and would be useful to see whether too much resources are spent twice)
> 4. in main thread: the time spent executing the redo function (continuation of 1) )
> 5. in async task: the time spent executing the redo function (continuation of 3) )
> 6. and 2) might prove too coarse and, for them, finer grained pieces could be logged as well (eg: `pgptr` fixing)
> 
> Doing this requires passing some context info between the involved code.
> Shall I attempt this?

What do you think about putting this in a Jira issue and attempt to make the change later?

src/transaction/log_page_async_fetcher.cpp
cristiarg: I added it in this PR, then I removed it as part of one of the PR's disjoined from this one.

src/base/perf_monitor.h
cristiarg: Agreed. Done

src/thread/thread_entry.cpp
cristiarg: Done

src/transaction/log_recovery_redo_parallel.hpp
cristiarg: Done

src/base/perf_monitor_trackers.hpp
cristiarg: Renamed. I think the default ctor is automatically dtor once there is a user defined one. I deleted the other ctors and ops.

src/base/perf_monitor_trackers.hpp
cristiarg: Renamed. Changed to composition. Thanks for noticing as the base class did not have a virtual dtor and slicing could happen. No need to delete default ctor. I deleted the other unneeded ctors and ops.

src/transaction/log_replication.cpp
cristiarg: You are so right. How did I miss that?

src/transaction/log_replication.cpp
cristiarg: I was thinking more to log perf at the level of a single entry in the log record. Because - more or less - the other entries are also logged at the level of processing a single entry.

cristiarg:
> What do you think about putting this in a Jira issue and attempt to make the change later?

@ribram To be honest, that was the purpose of this PR. But I will add another one.

ribram (approved):

ribram:
> @ribram To be honest, that was the purpose of this PR. But, added http://jira.cubrid.org/browse/LETS-97.

Please let me present a few arguments for the separation of proposed changes :).

I think that the statistics added by this patch represent a good starting point to get valuable information about the performance of replication in the future tests. There is certainly room for refining this information and adding more stats about parts that are not measured yet.

As you pointed, getting the fined grained information requires passing additional context. Making these changes in this patch will add more code to the current 280 new lines of code, that brought various changes and fixes. It will open new threads of discussion and it will add a few more days of development and reviewing.

In a separate patch, the focus of development and review will be just on refining these statistics. The discussion will be cleaner and the review easier. Also, I hope that in the meantime we do some testing and we get a view on what the current statistics tell us and what they don't tell us. We will have a better idea of what else will need to be measured.

PR #2742, opened 2021-04-27
[CBRD-23914] Fix -Wsign-compare

adrianbrab:

Comparison between signed and unsigned variables generate warnings.
To fix this the variable type was changed where possible and cast were used where not.

### DIFF
--- a/src/base/ddl_log.c
+++ b/src/base/ddl_log.c
@@ -781,7 +781,7 @@ logddl_write ()
 	}
 
       len = logddl_create_log_msg (buf);
-      if (len < 0 || fwrite (buf, sizeof (char), len, fp) != len)
+      if (len < 0 || fwrite (buf, sizeof (char), len, fp) != (size_t) len)
 	{
 	  goto write_error;
 	}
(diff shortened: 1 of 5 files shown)


ribram (approved):

mhoh3963 (commented):
@adrianbrab 
I think that it is needed to review by related developer, so I add reviewer @kisoo-han.

mhoh3963 (approved):

src/base/ddl_log.c
```diff
@@ -888,7 +888,7 @@ logddl_write_tran_str (const char *fmt, ...)
 	  len = DDL_LOG_BUFFER_SIZE;
 	}
 
-      if (len < 0 || fwrite (msg, sizeof (char), len, fp) != len)
+      if (fwrite (msg, sizeof (char), len, fp) != len)
```
kisoo-han, line 891: I want to leave len as an **int** not a **size_t** and change it like this (same as line #970):
`if (len < 0 || fwrite (buf, sizeof (char), len, fp) != (size_t) len)`

kisoo-han (approved):

airnet73 (approved):

hornetmj (approved):

src/base/ddl_log.c
adrianbrab: Changed back

mhoh3963:
@adrianbrab 
It was occurred code-style error, so you will run .github/workflows/codestyle.sh and send PR again.

mhoh3963 (approved):

PR #2744, opened 2021-04-27
[CBRD-23914] Fix -Wredundant-decls warnings

adrianbrab:

The project contains duplicate declarations for functions and macros

### DIFF
--- a/src/executables/csql_launcher.c
+++ b/src/executables/csql_launcher.c
@@ -429,12 +429,7 @@ main (int argc, char *argv[])
     }
   else if (!csql_arg.sa_mode && csql_arg.skip_vacuum)
     {
-      /* Don't allow to skip vacuum on CS mode
-         goto print_usage;
-         }
-         else if (explicit_single_line && csql_arg.single_line_execution == false)
-         {
-         /* Don't allow both at once. */
+      /* Don't allow both at once. */
       goto print_usage;
     }
   else if (csql_arg.sa_mode)
--- a/src/executables/csql_launcher.c
+++ b/src/executables/csql_launcher.c
@@ -429,6 +429,11 @@ main (int argc, char *argv[])
     }
   else if (!csql_arg.sa_mode && csql_arg.skip_vacuum)
     {
+      /* Don't allow to skip vacuum on CS mode */
+      goto print_usage;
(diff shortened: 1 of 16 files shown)


ribram (approved):

src/executables/csql_launcher.c
```diff
@@ -429,12 +429,7 @@ main (int argc, char *argv[])
     }
   else if (!csql_arg.sa_mode && csql_arg.skip_vacuum)
     {
-      /* Don't allow to skip vacuum on CS mode
-         goto print_usage;
-         }
-         else if (explicit_single_line && csql_arg.single_line_execution == false)
-         {
-         /* Don't allow both at once. */
+      /* Don't allow both at once. */
```
mhoh3963, line 432: The comments and indentation are wrong, so please recover your code like this:

/* Don't allow to skip vacuum on CS mode */
         goto print_usage;
  }
  else if (explicit_single_line && csql_arg.single_line_execution == false)
  {
         /* Don't allow both at once. */

hornetmj: @mhoh3963 This removed part is already in the comment. So, I think if we need to re-insert this part, then how about making another PR for it? Because this PR is for fixing the compile error.

hornetmj (approved):

src/executables/csql_launcher.c
adrianbrab: OK, i will open another PR for the change to keep fixes separated by type.

hornetmj: @adrianbrab We going to open a new CBRD issue for handling it. Please revert this.

mhoh3963: @hornetmj 
OK, I think good your suggestion.

@adrianbrab 
This is bug occurred with wrong comment made by me. So, please, don't modify this code. I will create new issue and fix it.

mhoh3963: @adrianbrab 
I already fixed codes in csql_launcher.c, so please revert the modified code and send PR again,

adrianbrab: merged with develop

mhoh3963 (approved):

PR #2761, opened 2021-05-03
[CBRD-23941] Remove the global type declaration, unordered_oid_set

wrlawodms:
http://jira.cubrid.org/browse/CBRD-23941

For more information, see the Jira issue.

As a side note,  I've removed a function for "< (lt)" and added one for "!=". The former seems better to add when it is needed, and the latter is for consistency with "==" defined.

### DIFF
--- a/src/object/schema_manager.c
+++ b/src/object/schema_manager.c
@@ -232,10 +232,6 @@ static const char *method_file_extension = ".o";
 #include <nlist.h>
 #endif /* !WINDOWS */
 
-// *INDENT-OFF*
-using unordered_oid_set = std::unordered_set<OID*, decltype(oid_pseudo_key)*, decltype(oid_eq)*>;
-// *INDENT-ON*
-
 #if defined (ENABLE_UNUSED_FUNCTION)	/* to disable TEXT */
 const char TEXT_CONSTRAINT_PREFIX[] = "#text_";
 #endif /* ENABLE_UNUSED_FUNCTION */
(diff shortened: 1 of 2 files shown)


src/storage/oid.h
```diff
@@ -133,24 +136,28 @@
...
-  return OID_EQ (oidp1, oidp2);
+  return OID_EQ (&oid1, &oid2);
 }
 
-inline bool
-oid_pseudo_key (const OID * oidp)
+inline bool operator!=(const OID& oid1, const OID& oid2)
 {
-  return OID_PSEUDO_KEY (oidp);
+  return !(oid1 == oid2);
```
hornetmj, line 157: How about '!OID_EQ (&oid1, &oid2)'?

src/object/schema_manager.c
```diff
@@ -422,7 +418,7 @@ static bool sm_is_possible_to_recreate_constraint (MOP class_mop, const SM_CLASS
 static bool sm_filter_index_pred_have_invalid_attrs (SM_CLASS_CONSTRAINT * constraint, char *class_name,
 						     SM_ATTRIBUTE * old_atts, SM_ATTRIBUTE * new_atts);
 
-static int sm_collect_truncatable_classes (MOP class_mop, unordered_oid_set & trun_classes, bool is_cascade);
+static int sm_collect_truncatable_classes (MOP class_mop, std::unordered_set < OID > &trun_classes, bool is_cascade);
```
hornetmj, line 421: Personal curious:

Why do you change the typename T from OID* to OID? Is there any benefit?

hornetmj (approved):

src/object/schema_manager.c
wrlawodms: I think value is much safer and simpler than pointer, and OID* and OID have the same size in 64bit.

src/storage/oid.h
wrlawodms: Could you let me know why you think it's better?

hornetmj: I just think it is more readable. And It is consistent because the !OID_EQ() is used in many parts. Is there any benefit to your style? Are there any standards or rules to check by me?

src/object/schema_manager.c
hornetmj: I got it.

src/storage/oid.h
wrlawodms: The rule is just that "When defining ==, also define != and make sure that a!=b means !(a==b)".
 I thought it would be easy to keep this rule if I define `==` and write `!=` like `!(oid1 == oid2)` so that if needed to change, I can just modify `==`.

hornetmj: It is not an important issue. So it depends on you.

hornetmj (approved):

src/storage/oid.h
wrlawodms: I think your suggestion also makes sense. I'll follow it.

JoohoK (approved):

PR #2762, opened 2021-05-03
[CBRD-23936] Make hfid-cache allow the hfid of a class to be changed

wrlawodms:
http://jira.cubrid.org/browse/CBRD-23936

See the Jira issue for more information.

### DIFF
--- a/src/query/vacuum.c
+++ b/src/query/vacuum.c
@@ -2461,6 +2461,7 @@ vacuum_heap_get_hfid_and_file_type (THREAD_ENTRY * thread_p, VACUUM_HEAP_HELPER 
 {
   int error_code = NO_ERROR;	/* Error code. */
   OID class_oid = OID_INITIALIZER;	/* Class OID. */
+  FILE_DESCRIPTORS file_descriptor;
   FILE_TYPE ftype;
 
   assert (helper != NULL);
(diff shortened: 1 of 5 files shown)


src/query/vacuum.c
```diff
@@ -6031,14 +6021,6 @@ vacuum_log_add_dropped_file (THREAD_ENTRY * thread_p, const VFID * vfid, const O
 
   /* Initialize recovery data */
   VFID_COPY (&rcv_data.vfid, vfid);
-  if (class_oid != NULL)
-    {
-      COPY_OID (&rcv_data.class_oid, class_oid);
-    }
-  else
-    {
-      OID_SET_NULL (&rcv_data.class_oid);
```
hornetmj, line 6040: This initialization code is necessary. The rcv_data.class_oid will be referenced by REDO/UNDO functions. If there is an initialization value, then unexpected behavior will occur.

wrlawodms: I've removed the usage of the `class_oid`, please check `vacuum_rv_notify_dropped_file()` below

src/communication/network_interface_cl.c
```diff
@@ -1686,7 +1686,7 @@ heap_destroy (const HFID * hfid)
 
   THREAD_ENTRY *thread_p = enter_server ();
 
-  success = xheap_destroy (thread_p, hfid, NULL);
+  success = xheap_destroy (thread_p, hfid);
```
hornetmj: When is the hfid removed from the cache in this case?
- previous:
  - the heap_delete_hfid_from_cache() is called by the vacuum_rv_notify_dropped_file() if the class_oid in not null
- now:
  - the heap_delete_hfid_from_cache() is called by the xheap_destroy_newly_created() only
 
Doesn't it need to remove from the cache?

src/query/vacuum.c
hornetmj: I got it. However, I think it is safe to set it to null even if it is unused. (But, it depends on you. ^^)

wrlawodms:
I've integrated checking and getting for the cache. `heap_hfid_is_cached()` has been removed and `heap_get_hfid_if_cached()` has been created.

hornetmj (approved):

PR #2763, opened 2021-05-03
[CBRD-23937] Recreating heap file while TRUNCATE if REUSE_OID	

wrlawodms:
http://jira.cubrid.org/browse/CBRD-23937

See the Jira issue for more information.

As a side note, I've removed "log_does_allow_replication()" condition checking in the code that was commented out. According to my analysis and test, TRUNCATE is logged as one of Statement Based Logging, and it looks working well with destroying heap even if the PK is not dropped. I guess it is a kind of legacy code.

### DIFF
--- a/src/object/schema_manager.c
+++ b/src/object/schema_manager.c
@@ -15780,17 +15780,6 @@ sm_truncate_class_internal (MOP class_mop)
       goto error_exit;
     }
 
-  c = classobj_find_cons_primary_key (class_->constraints);
-  if (c != NULL && classobj_is_pk_referred (class_mop, c->fk_info, false, NULL))
-    {
-      /* Fk-referred PK can't be dropped. so we just remove the instances of the current class. */
-      error = locator_remove_class_from_index (ws_oid (class_mop), &c->index_btid, &class_->header.ch_heap);
-      if (error != NO_ERROR)
-	{
-	  goto error_exit;
-	}
-    }
-
   /* collect index information */
   for (c = class_->constraints; c; c = c->next)
     {
@@ -15800,13 +15789,23 @@ sm_truncate_class_internal (MOP class_mop)
 	  continue;
 	}
 
-      if (c->type == SM_CONSTRAINT_PRIMARY_KEY)
+      if ((c->type == SM_CONSTRAINT_PRIMARY_KEY && classobj_is_pk_referred (class_mop, c->fk_info, false, NULL))
+	  || !sm_is_possible_to_recreate_constraint (class_mop, class_, c))
 	{
-	  /* Do not save PK referred by FK. the PK can't be dropped. */
-	  continue;
+	  /*
+	   * 1. the PK referred to by a FK.
+	   * 2. the B+tree which contains OIDs of other classes in the inheritance hierarchy.
+	   *
+	   * In these cases, We cannot drop and recreate the index as this might be
+	   * too costly, so we just remove the instances of the current class.
+	   */
+	  error = locator_remove_class_from_index (ws_oid (class_mop), &c->index_btid, &class_->header.ch_heap);
+	  if (error != NO_ERROR)
+	    {
+	      goto error_exit;
+	    }
 	}
-
-      if (sm_is_possible_to_recreate_constraint (class_mop, class_, c))
+      else
 	{
 	  /* All the OIDs in the index should belong to the current class, so it is safe to drop and create the
 	   * constraint again. We save the information required to recreate the constraint. */
@@ -15831,18 +15830,6 @@ sm_truncate_class_internal (MOP class_mop)
 		{
 		  goto error_exit;
 		}
-	    }
-	}
-      else
-	{
-	  /* The B+tree might contain OIDs of other classes in the inheritance
-	   * hierarchy. We cannot drop and recreate the index as this might be
-	   * too costly, so we just remove the instances of the current class.
-	   */
-	  error = locator_remove_class_from_index (ws_oid (class_mop), &c->index_btid, &class_->header.ch_heap);
-	  if (error != NO_ERROR)
-	    {
-	      goto error_exit;
 	    }
 	}
     }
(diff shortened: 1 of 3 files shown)


src/object/schema_manager.c
```diff
@@ -15883,26 +15881,18 @@ sm_truncate_class_internal (MOP class_mop)
...
-  else
+  if (sm_is_reuse_oid_class (class_mop))
     {
       error = sm_truncate_using_destroy_heap (class_mop);
     }
-  if (error != NO_ERROR)
+  else
     {
-      goto error_exit;
+      error = sm_truncate_using_delete (class_mop);
```
hornetmj, line 15890: In this case, The locator_remove_class_from_index() can be called. Why is it different from before?

src/object/schema_manager.c
```diff
@@ -15797,13 +15785,23 @@ sm_truncate_class_internal (MOP class_mop)
...
-	  /* Do not save PK referred by FK. the PK can't be dropped. */
-	  continue;
+	  /*
+	   * 1. the PK referred to by a FK.
+	   * 2. the B+tree which contains OIDs of other classes in the inheritance hierarchy.
+	   *
+	   * In these cases, We cannot drop and recreate the index as this might be
+	   * too costly, so we just remove the instances of the current class.
+	   */
+	  error = locator_remove_class_from_index (ws_oid (class_mop), &c->index_btid, &class_->header.ch_heap);
```
hornetmj, line 15798: How about moving this line out of for loop? I think it is more readable and It seems to serve the purpose of this for loop better.

src/object/schema_manager.c
```diff
@@ -15797,13 +15785,23 @@ sm_truncate_class_internal (MOP class_mop)
...
 	}
 
-      if (keep_pk == true && c->type == SM_CONSTRAINT_PRIMARY_KEY)
+      if ((c->type == SM_CONSTRAINT_PRIMARY_KEY && classobj_is_pk_referred (class_mop, c->fk_info, false, NULL))
+	  || !sm_is_possible_to_recreate_constraint (class_mop, class_, c))
 	{
-	  /* Do not save PK referred by FK. the PK can't be dropped. */
-	  continue;
+	  /*
+	   * 1. the PK referred to by a FK.
```
hornetmj, line 15792: Please explain more detail about this condition and leave it to related Jira issue. Because it is thought to be the most commonly used case. (I remember you shared the reason with me. But, I ask you once again)

JoohoK (approved):

hornetmj (approved):

hornetmj:
Plz check and reply to my comment before you merge it. I made the mistake to select 'approve' or 'comment' when I left my comment.

src/object/schema_manager.c
wrlawodms: Do you mean like the original code?

src/object/schema_manager.c
wrlawodms: No matter an index is droppable or not, the heap records in a heap with DONT_REUSE_OID have to be deleted record by record. I'll be modified when dealing with the type of heap file.

src/object/schema_manager.c
wrlawodms: Ok, I've added an overview of these undroppable indexes on the CBRD-23931. When it comes to details, do you mean the description of CBRD-23938 is not enough? To give a precise reason for this need more code analysis.

src/storage/heap_file.c
```diff
@@ -6427,7 +6427,7 @@ heap_ovf_find_vfid (THREAD_ENTRY * thread_p, const HFID * hfid, VFID * ovf_vfid,
 	  log_append_redo_data (thread_p, RVHF_STATS, &addr_hdr, sizeof (*heap_hdr), heap_hdr);
 	  pgbuf_set_dirty (thread_p, addr_hdr.pgptr, DONT_FREE);
 
-	  log_sysop_commit (thread_p);
+	  log_sysop_attach_to_outer (thread_p);
```
wrlawodms, line 6430: Without this change, when ROLLBACK after TRUNCATE, it generates a dangling overflow file referring to a dropped heap file. This sysop only makes sense when the heap file for a class is never changed and even had an advantage then because it created an overflow heap file only once even with ROLLBACK.

wrlawodms: I'll be addressed in another issue because it is a more general bug.

src/object/schema_manager.c
hornetmj: It's enough. (CBRD-23931)

src/object/schema_manager.c
wrlawodms: But, if done as you said, it'll still need a condition checking like this to skip a PK referred some FKs. Don't you think it looks redundant?

src/storage/system_catalog.c
```diff
@@ -4499,7 +4500,9 @@ catalog_update (THREAD_ENTRY * thread_p, RECDES * record_p, OID * class_oid_p)
 
   assert (OID_EQ (&rep_dir, &(class_info_p->ci_rep_dir)));
 
-  if (HFID_IS_NULL (&class_info_p->ci_hfid))
+  or_class_hfid (record_p, &(class_hfid));
+
+  if (HFID_IS_NULL (&class_info_p->ci_hfid) || !HFID_EQ (&class_info_p->ci_hfid, &class_hfid))
     {
       or_class_hfid (record_p, &(class_info_p->ci_hfid));
```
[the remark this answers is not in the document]
hornetmj: @wrlawodms I'm in favor of your opinion If it's guaranteed that a record_p always has a hfid.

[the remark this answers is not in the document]
wrlawodms: `record_p` could be NULL, but then it will be passed in the following `if`. It looks the same logic as the current code.

src/object/schema_manager.c
hornetmj: I thought that a dummy variable which is the SM_CONSTRAINT_INFO type is used to collect constraints matched in these conditions. It looks a little in-efficient, but more readable. I think the comment just before for loop might be changed or removed if you don't want to change this processing(/* collect index information */).

wrlawodms: I see, I'll revise the comment.

PR #2766, opened 2021-05-04
[CBRD-23942] When executing sql query in csql, the actual execution time and the time displayed on the screen are different.

airnet73:
http://jira.cubrid.org/browse/CBRD-23942

Purpose
When autocommit is true in csql, the time that does not include the commit time is displayed on the screen when the sql query is executed.
Therefore, the actual sql query execution time and the time displayed on the screen are different.
Added commited transaction time to avoid this confusion.

if the query execution time is 0.005703 sec and time for committing that transaction is 9.0 sec.
Execute OK. (0.005703 sec) Committed. **(9.000000 sec)**

Implementation
N/A

Remarks
N/A


kisoo-han:
Please change csql as follow when it is in auto-commit mode:
if the query execution time is 0.204657 sec and time for committing that transaction is 9.0 sec.
`Execute OK. (0.204657 sec) Committed. (9.000000 sec)`

kisoo-han (changes requested):
Please refer comment for csql output in auto-commit mode, and change time.

kisoo-han (approved):

PR #2768, opened 2021-05-04
[LETS-98] Transaction server crashes during recovery if not connected to page server

danilailie:
http://jira.cubrid.org/browse/LETS-98
Transaction server crashes during recovery if not connected to page server

Made sure there is no log page request sent to page server if there is no page server connected.

* [LETS-98] Do not fetch log page if there is no PS connection


ribram (approved):
Please set the name of PR & commit message `[LETS-98] Do not fetch log page if there is no PS connection`

PR #2769, opened 2021-05-05
[LETS-11] Rewrite checkpoint using meta log checkpoint information.

ribram:
http://jira.cubrid.org/browse/LETS-11

Rewrite checkpoint handling (in logpb_checkpoint and recovery analysis implementation) to use meta log checkpoint info map instead of log records. As a result checkpoint becomes usable on the page server.

**logpb_checkpoint**

- Reactivate checkpoint for page server
- Use checkpoint_info functions and remove all code that added log records and loaded transaction table information into the end checkpoint log record.

  - Instead of using LOG_START_CHKPT start_lsa as checkpoint LSA, use log_Gl.prior_info.prev_lsa, the LSA of last appended log record. LSA of last record was chosen instead of end of log LSA because checkpoint LSA has to be less than nxio_lsa.
  - Replaced tmp_chkpt.redo_lsa with oldest_unflushed_lsa & new_chkpt_redo_lsa. oldest_unflushed_lsa is obtained by pgbuf_flush_checkpoint, new_chkpt_redo_lsa is the minimum between checkpoint LSA and oldest_unflushed_lsa
  - Flush checkpoint info do disk before flushing log header.
  - Remove old checkpoint info after checkpoint is finished.

- Renamed chkpt_lsa/chkpt_redo_lsa as prev_chkpt_lsa/prev_chkpt_redo_lsa to clarify their meaning.
- Renamed smallest_lsa as smallest_tran_lsa
- Remove obsolete functions logpb_checkpoint_trans and logpb_checkpoint_topops

**log_recovery_analysis**

- Get checkpoint information from meta log instead of log records and update the code.

  - Identify the checkpoint info with the start_lsa. For normal recovery, start_lsa matches checkpoint LSA and should find a checkpoint info entry. For restore, no checkpoint info is found or used.
  - Apply recovery analysis based on checkpoint information before analyzing the record at snapshot LSA. This represents the same point in time when the transaction table was loaded.
  - Apply recovery_2pc_analysys at the end.

- Remove debug verification related to checkpoint LSA. It was not an useful verification up until now and it requires passing information that is not directly related to run postpone analysis.
- Update log_rv_analysis_record signature without checkpoint related arguments.
- Remove obsolete functions log_rv_analysis_start_checkpoint and log_rv_analysis_end_checkpoint.

**Checkpoint info**

- m_has_2pc setting moved from recovery_analysis() to load_trantable_snapshot() to allow calling recovery_analysis() on a const instance.
- added bodies for get_snapshot_lsa() and get_start_redo_lsa()
- added get_transaction_count() and get_sysop_count(). the first one is used in logpb_checkpoint().
- Fix a warning regarding signed/unsigned compare

**Other**

- Expose log_write_metalog_to_file to log_manager.h to be called by logpb_checkpoint.

**Tests**

- Tested configuration with a transaction server without page server

  - Manually tested forced crashes and checkpoint.
  - Full shell_debug with no regressions related to checkpoint.

### DIFF
--- a/src/transaction/log_recovery.c
+++ b/src/transaction/log_recovery.c
@@ -26,9 +26,11 @@
 
 #include "boot_sr.h"
 #include "locator_sr.h"
+#include "log_checkpoint_info.hpp"
 #include "log_impl.h"
 #include "log_lsa.hpp"
 #include "log_manager.h"
+#include "log_meta.hpp"
 #include "log_reader.hpp"
 #include "log_recovery_redo_parallel.hpp"
 #include "log_system_tran.hpp"
@@ -52,8 +54,8 @@ static bool log_rv_find_checkpoint (THREAD_ENTRY * thread_p, VOLID volid, LOG_LS
 static int log_rv_analysis_undo_redo (THREAD_ENTRY * thread_p, int tran_id, LOG_LSA * log_lsa);
 static int log_rv_analysis_dummy_head_postpone (THREAD_ENTRY * thread_p, int tran_id, LOG_LSA * log_lsa);
 static int log_rv_analysis_postpone (THREAD_ENTRY * thread_p, int tran_id, LOG_LSA * log_lsa);
-static int log_rv_analysis_run_postpone (THREAD_ENTRY * thread_p, int tran_id, LOG_LSA * log_lsa, LOG_PAGE * log_page_p,
-					 LOG_LSA * check_point);
+static int log_rv_analysis_run_postpone (THREAD_ENTRY * thread_p, int tran_id, LOG_LSA * log_lsa,
+					 LOG_PAGE * log_page_p);
 static int log_rv_analysis_compensate (THREAD_ENTRY * thread_p, int tran_id, LOG_LSA * log_lsa, LOG_PAGE * log_page_p);
 static int log_rv_analysis_will_commit (THREAD_ENTRY * thread_p, int tran_id, LOG_LSA * log_lsa);
 static int log_rv_analysis_commit_with_postpone (THREAD_ENTRY * thread_p, int tran_id, LOG_LSA * log_lsa,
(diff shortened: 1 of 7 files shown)


src/transaction/log_recovery.c
```diff
@@ -2497,7 +2188,6 @@ log_recovery_analysis (THREAD_ENTRY * thread_p, LOG_LSA * start_lsa, LOG_LSA * s
 		       LOG_LSA * end_redo_lsa, bool is_media_crash, time_t * stop_at, bool * did_incom_recovery,
```
cristiarg, line 2188: Maybe the comments/notes to these functions whose internal mechanics has been changed should be revised too?

Eg:
```
since the last checkpoint record reflected in the log and the data volumes
```

src/transaction/log_recovery.c
```diff
@@ -2915,9 +2603,16 @@ log_recovery_analysis (THREAD_ENTRY * thread_p, LOG_LSA * start_lsa, LOG_LSA * s
...
 	    }
 
-	  log_rv_analysis_record (thread_p, log_rtype, tran_id, &log_lsa, log_page_p, &checkpoint_lsa, &prev_lsa,
-				  start_lsa, start_redo_lsa, is_media_crash, stop_at, did_incom_recovery,
-				  &may_use_checkpoint, &may_need_synch_checkpoint_2pc);
+	  if (chkpt_infop != nullptr && chkpt_infop->get_snapshot_lsa () == log_lsa)
+	    {
+	      // The transaction table snapshot was taken before the next log record was logged.
+	      // Rebuild the transaction table image based on checkpoint information
+	      assert (start_redo_lsa != nullptr);
```
cristiarg, line 2610: already asserted before in a `LSA_COPY`

cristiarg (approved):

adrianbrab (approved):

src/transaction/log_recovery.c
```diff
@@ -2351,15 +2051,6 @@ log_rv_analysis_record (THREAD_ENTRY * thread_p, LOG_RECTYPE log_type, int tran_
       log_rv_analysis_sysop_end (thread_p, tran_id, log_lsa, log_page_p);
       break;
 
-    case LOG_START_CHKPT:
```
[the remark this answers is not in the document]
ribram: This will be part of LETS-100, a refactoring task to remove the obsolete code. It was separated to keep this patch small.

src/transaction/log_recovery.c
ribram: The comment does not look outdated to me. The quoted mentions is still true because the last checkpoint is still reflected into the log & data volumes. An implementation detail has changed, but in my opinion it is not useful to mention it in the description comment of log_recovery_analysis.

src/transaction/log_recovery.c
ribram: removed the assert.

PR #2770, opened 2021-05-05
[LETS-88] add debug logging to replication delay calculation

cristiarg:
http://jira.cubrid.org/browse/LETS-88

- add new parameter `log_calculate_replication_delay`
- add logging in function log_rpl_calculate_replication_delay


src/transaction/log_replication.cpp
```diff
@@ -280,15 +280,21 @@ namespace cublog
...
 	const int64_t end_time_msec = util_get_time_as_ms_since_epoch ();
 	const int64_t time_diff_msec = end_time_msec - a_start_time_msec;
-	assert (time_diff_msec > 0);
+	assert (time_diff_msec >= 0);
 
 	perfmon_set_stat (thread_p, PSTAT_REDO_REPL_DELAY, static_cast<int> (time_diff_msec), false);
 
+	if (prm_get_bool_value (PRM_ID_LOG_CALC_REPL_DELAY))
+	  {
+	    er_log_debug (ARG_FILE_LINE, "[CALC_REPL_DELAY]: %9lld msec", time_diff_msec);
```
hornetmj, line 295: I think It is more helpful to you to use '_er_ log_debug()' than 'er log_debug()'. It is just a comment and you can choose a better one.

hornetmj (approved):

src/base/system_parameter.h
```diff
@@ -463,6 +463,8 @@ enum param_id
   PRM_ID_ER_LOG_READ_LOG_PAGE,
   PRM_ID_ER_LOG_READ_DATA_PAGE,
 
+  PRM_ID_LOG_CALC_REPL_DELAY,
```
ribram, line 466: To be in sync with the other debug logging parameters.
```suggestion
  PRM_ID_ER_LOG_CALC_REPL_DELAY,
```

src/transaction/log_replication.cpp
```diff
@@ -280,15 +280,21 @@ namespace cublog
   {
     // skip calculation if bogus input (sometimes, it is -1);
     // TODO: fix bogus input at the source if at all possible (debugging revealed that
-    // it happens for LOG_COMMIT messages only)
+    // it happens for LOG_COMMIT messages only and there is no point at the source where the 'at_time'
+    // is not filled in; the only other possibility is a reading misalignment that remains below the radar)
```
ribram, line 284: I think it's best to keep unconfirmed theories out of the code :P
```suggestion
    // it happens for LOG_COMMIT messages only and there is no point at the source where the 'at_time'
    // is not filled in
```

src/transaction/log_replication.cpp
ribram: er_log_debug requires PRM_ID_ER_LOG_DEBUG to be true. It is the generic debug logging function. Since this debug message already requires PRM_ID_LOG_CALC_REPL_DELAY to be true, \_er_log_debug may be called directly.

ribram (approved):

src/base/system_parameter.h
cristiarg: Done. I thought 'er' was for 'error' and this one is not about any error.

src/transaction/log_replication.cpp
cristiarg: Done

src/base/system_parameter.c
```diff
@@ -2040,6 +2041,10 @@ bool PRM_ER_LOG_READ_DATA_PAGE = true;
 static bool prm_er_log_read_data_page_default = true;
 static unsigned int prm_er_log_read_data_page_flag = 0;
 
+bool PRM_ER_LOG_CALC_REPL_DELAY = true;
+static bool prm_er_log_calc_repl_delay_default = true;
```
ribram, line 2045: I should have payed more attention. By default the debug logging parameters should be false.

An issue was open to change the default value for prm_er_log_read_log_page_default and prm_er_log_read_log_page_default too.

ribram (approved):

PR #2773, opened 2021-05-06
[CBRD-23914] Fix tmpnam is dangerous warning by replacing with mkstemp

adrianbrab:

The usage of tmpnam is considered dangerous, this patch replaces usages of the function with std variants and adds the usage of filesystem library.

### DIFF
--- a/src/base/filesys_temp.cpp
+++ b/src/base/filesys_temp.cpp
@@ -20,6 +20,7 @@
  * file_sys.cpp: File System namespace & functionality
  */
 #include "filesys_temp.hpp"
+#include <filesystem>
 #include <stdlib.h>
 
 #ifdef LINUX
@@ -35,7 +36,10 @@ namespace
   std::string unique_tmp_filename (const char *prefix="cub_") //generates an unique filename in tmp folder
   {
 #ifdef LINUX
-    std::string filename = std::string ("/tmp/") + prefix + "XXXXXX"; //used with mkstemp()
+    std::string filename = std::filesystem::temp_directory_path ();
+    filename += "/";
+    filename += prefix;
+    filename += "XXXXXX"; //used with mkstemp()
     //TBD (not necessary yet)
 #elif WINDOWS
     char buf[L_tmpnam] = {};
--- a/src/base/filesys_temp.cpp
+++ b/src/base/filesys_temp.cpp
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2008 Search Solution Corporation
+ * Copyright 2016 CUBRID Corporation
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+
+/*
+ * file_sys.cpp: File System namespace & functionality
+ */
+#include "filesys_temp.hpp"
+#include <stdlib.h>
+
+#ifdef LINUX
+#include "porting.h"
+#elif WINDOWS
+#include <cstdio>
+#include <fcntl.h>
+#include <io.h>
+#endif
+
+namespace
+{
+  std::string unique_tmp_filename (const char *prefix="cub_") //generates an unique filename in tmp folder
+  {
+#ifdef LINUX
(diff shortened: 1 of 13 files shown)


ribram (approved):

src/base/filesys_temp.cpp
```diff
@@ -0,0 +1,77 @@
...
+#include <fcntl.h>
+#include <io.h>
+#endif
+
+namespace
+{
+  std::string unique_tmp_filename (const char *prefix="cub_") //generates an unique filename in tmp folder
+  {
+#ifdef LINUX
+    std::string filename = std::string ("/tmp/") + prefix + "XXXXXX"; //used with mkstemp()
```
mhoh3963, line 38: "/tmp" directory may not be available for security reasons.
So, It would be better to use the $TMPDIR as the tmpname() function instead of fixing it to "/tmp" directory.
If $TMPDIR doesn't exist, you will use "/tmp" directory.

adrianbrab: I changed to std::filesystem::temp_directory_path () to better handle the path

mhoh3963: OK. it sounds good.

mhoh3963 (approved):

src/executables/csql_support.c
```diff
@@ -249,84 +251,61 @@ csql_invoke_system (const char *command)
...
+  auto[filename, fileptr] = filesys::open_temp_file ("csql_");
+  if (fileptr == NULL)
     {
       csql_Error_code = CSQL_ERR_OS_ERROR;
-      goto error;
+      nonscr_display_error (csql_Scratch_text, SCRATCH_TEXT_LEN);
+      return CSQL_FAILURE;
     }
+  filesys::auto_delete_file file_del (filename.c_str ());	//deletes file at scope end
+  filesys::auto_close_file file (fileptr);	//closes file at scope end (before the above file deleter); forget about fp from now on
```
hornetmj, line 270: How is 'before the above file deleter' guaranteed? Is it using the nature of the stack to which local variables are allocated? If so, there is a dependency that must adhere to the order of declaration when using the two variables.

src/connection/connection_cl.c
```diff
@@ -953,35 +959,36 @@ css_connect_to_master_server (int master_port_id, const char *server_name, int n
...
 	  er_set_with_oserror (ER_ERROR_SEVERITY, ARG_FILE_LINE, ERR_CSS_ERROR_DURING_SERVER_CONNECT, 1, server_name);
 	  goto fail_end;
 	}
+      if (!css_tcp_listen_server_datagram (socket_fd, &datagram_fd))
+	{
+	  (void) unlink (pname.c_str ());
+	  css_free_conn (conn);
+	  close (socket_fd);
+	  er_set_with_oserror (ER_ERROR_SEVERITY, ARG_FILE_LINE, ERR_CSS_ERROR_DURING_SERVER_CONNECT, 1, server_name);
+	  return NULL;
```
hornetmj: How about using 'goto fail_end;' like the previous error case?

src/connection/connection_sr.c
```diff
@@ -1157,38 +1163,36 @@ css_connect_to_master_server (int master_port_id, const char *server_name, int n
 #else /* WINDOWS */
       /* send the "pathname" for the datagram */
       /* be sure to open the datagram first.  */
+      pname = std::filesystem::temp_directory_path ();
+      pname += "/cubrid_tcp_setup_server" + std::to_string (getpid ());
```
hornetmj, line 1167: Why didn't you use the newly created mechanism?

hornetmj (approved):

src/executables/file_hash.c
```diff
@@ -215,7 +216,9 @@ fh_create (const char *name, int est_size, int page_size, int cached_pages, cons
   /* Open the hash file */
   if (!hash_filename || hash_filename[0] == '\0')
     {
-      ht->hash_filename = strdup (tmpnam (NULL));
+      auto[filename, filedes] = filesys::open_temp_filedes ("fhash_");
```
hornetmj, line 219: Personally, it would be better to use the same prefix for all temporary files created in CUBRID. However, the opinion of @mhoh3963  is the most important. I will follow him.

PS. Is there any reason to use a different prefix like "fhash_"?

src/connection/connection_sr.c
hornetmj: Come to think of it, it can't be used. If filesys_temp.hpp has a function that can do this, it seems more consistent.

src/connection/connection_cl.c
adrianbrab: changed to goto fail_end for more consistency

src/executables/file_hash.c
adrianbrab: I think the only reason is to better differentiate the temp files between them, to be easier to find, in case that is necessary.

src/connection/connection_sr.c
adrianbrab: there is a task opened to refactor the code from connection_cl and connection_sr in the connect functions as is already 80% identical.

PR #2774, opened 2021-05-06
[LETS-81] Packing functions for log_lsa

danilailie:
http://jira.cubrid.org/browse/LETS-81

Added utils to help pack and unpack log_lsa. 
Added a new constructor for log_lsa that receives int64_t.
Added a new conversion operator to int64_t from log_lsa.

Added unit test.


src/transaction/log_lsa.hpp
```diff
@@ -29,12 +29,18 @@
...
+const int page_offset_bit_count = 16;
+
 struct log_lsa
 {
-  std::int64_t pageid:48;		/* Log page identifier : 6 bytes length */
-  std::int64_t offset:16;		/* Offset in page : 2 bytes length.
+std::int64_t pageid:
+  page_id_bit_count;		/* Log page identifier : 6 bytes length */
+std::int64_t offset:
+  page_offset_bit_count;		/* Offset in page : 2 bytes length.
```
ribram, line 43: I find previous code more readable. It was easily understandable that pageid+offset take 8 bytes.

Usually it is best to give literals names to help other devs understand the literal meaning. But this is a bit_field and it is expected to see the literals. Please notice that the code in the standard [bit_field](https://en.cppreference.com/w/cpp/language/bit_field) manual does the same.

Later edit: I see now that the constants are used for the cast operators. The same kind of result can be obtained with the help of std::memcpy. In my opinion, std::memcpy version will be friendlier with the future reader.

src/transaction/log_lsa.hpp
```diff
@@ -44,6 +50,8 @@ struct log_lsa
     , offset (log_offset)
   {
   }
+
+  inline log_lsa (const int64_t value);
```
ribram, line 54:
```suggestion
  inline log_lsa (const std::int64_t value);
```

src/transaction/log_lsa.hpp
```diff
@@ -56,6 +64,7 @@ struct log_lsa
   inline bool operator<= (const log_lsa &olsa) const;
   inline bool operator> (const log_lsa &olsa) const;
   inline bool operator>= (const log_lsa &olsa) const;
+  inline operator int64_t () const;
```
ribram, line 67:
```suggestion
  inline operator std::int64_t () const;
```

src/transaction/log_lsa.hpp
```diff
@@ -91,6 +100,12 @@ inline bool LSA_GT (const log_lsa *plsa1, const log_lsa *plsa2);
 // inline/template implementation
 //////////////////////////////////////////////////////////////////////////
 
+log_lsa::log_lsa (const int64_t value)
```
ribram, line 103:
```suggestion
log_lsa::log_lsa (const std::int64_t value)
```

src/transaction/log_lsa_utils.cpp
```diff
@@ -0,0 +1,41 @@
...
+#include "packer.hpp"
+
+#include <cstring>
+
+namespace cublog::lsa_utils
+{
+  void
+  pack (const log_lsa &lsa, cubpacking::packer &serializer)
+  {
+    serializer.pack_bigint (lsa);
```
ribram, line 31: Be specific that a cast is involved.
```suggestion
    serializer.pack_bigint (static_cast<std::int64_t> (lsa));
```

unit_tests/log_transfer/test_main_log_lsa_utils.cpp
```diff
@@ -0,0 +1,99 @@
...
+void do_test (test_env &env)
+{
+  env.run_test ();
+}
+
+TEST_CASE ("First test", "")
+{
+  test_env env;
+  do_test (env);
+}
```
ribram, line 76: Having a test_env for this simple test seems overkill. 
```suggestion
TEST_CASE ("First test", "")
{
  log_lsa lsa;
  lsa.pageid = 4;
  lsa.offset = 8;
  char buffer[8];

  cubpacking::packer packer (buffer, 8);
  cublog::lsa_utils::pack (lsa, packer);

  cubpacking::unpacker unpacker (buffer, 8);
  log_lsa unpacked_lsa;
  cublog::lsa_utils::unpack (unpacker, unpacked_lsa);

  REQUIRE (lsa.pageid == unpacked_lsa.pageid);
  REQUIRE (lsa.offset == unpacked_lsa.offset);

  int64_t big_int = lsa;
  log_lsa new_lsa (big_int);

  REQUIRE (lsa == new_lsa);
}
```

src/transaction/log_lsa.hpp
danilailie: is there another int64_t besides the standard one? I feel like it only makes the code line longer...

ribram: consider this just a suggestion and do how you feel is better. in most cases standard functions, classes and types should be used with the std:: namespace. There are exceptions like size_t. int64_t could be one such exception too.

src/transaction/log_lsa.hpp
```diff
@@ -91,6 +98,11 @@ inline bool LSA_GT (const log_lsa *plsa1, const log_lsa *plsa2);
...
+{
+  std::memcpy (this, &value, sizeof (value));
```
ribram, line 103:
```suggestion
  static_assert (sizeof (int64_t) == sizeof (log_lsa), "Same size is required");
  std::memcpy (this, &value, sizeof (value));
```

src/transaction/log_lsa.hpp
```diff
@@ -141,6 +153,13 @@ log_lsa::operator>= (const log_lsa &olsa) const
   return !operator< (olsa);
 }
 
+log_lsa::operator int64_t () const
+{
+  int64_t result = 0;
```
ribram, line 158:
```suggestion
  static_assert (sizeof (int64_t) == sizeof (log_lsa), "Same size is required");
  int64_t result = 0;
```

src/transaction/log_lsa.hpp
```diff
@@ -29,8 +29,12 @@
 
 #include <cassert>
 #include <cinttypes>
+#include <cstring>
 #include <cstddef>
 
+const int page_id_bit_count = 48;
+const int page_offset_bit_count = 16;
+
```
ribram, line 37: No longer required.
```suggestion
```

ribram (approved):

PR #2781, opened 2021-05-11
[CBRD-23816] Support HASH TEMP FILE

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-23816

**SCAN MANAGER**
 - scan_open_list_scan: add fhs_create()
 - scan_build_hash_list_scan: add fhs_insert()
 - scan_hash_probe_next: add fhs_search(), fhs_search_next()
 - scan_close_scan: add fhs_destroy()

**FILE HASH STRUCTURE**
Separate related logic from extendible_hash.c. There are the following changes.
 1. Remove routines related to concurrency (lock, undo log)
 2. duplicate data can be stored.
  ==> By adding a duplicate key bucket, the duplicate data is stored in the chaining bucket.
3. Search in inserted order for backwards compatibility.
4. Remove routines related to deleting data.


src/query/query_hash_scan.h
```diff
@@ -77,17 +90,41 @@ struct hash_scan_key
...
+      HENTRY_HLS_PTR curr_hash_entry;	/* current hash entry */
+    } memory;
+    struct
+    {
+      FHSID *hash_table;	/* extendible hash table file */
+      OID curr_oid;		/* current bucket oid */
+      bool is_dk_bucket;	/* is current bucket dk? */
+    } file;
+  };
   int hash_list_scan_yn;	/* Is hash list scan possible? */
```
beyondykk9:
```suggestion
  int hash_list_scan_type;	/* MEM, HYBRID, or TEMP_FILE */
```

src/query/query_hash_scan.c
```diff
@@ -506,12 +666,2483 @@ int
...
+fhs_write_oid_to_record (char *record_p, OID * oid_p)
+{
+  *(PAGEID *) record_p = oid_p->pageid;
+  record_p += sizeof (PAGEID);
+
+  *(VOLID *) record_p = oid_p->volid;
+  record_p += sizeof (VOLID);
+
+  *(PGSLOTID *) record_p = oid_p->slotid;
+  record_p += sizeof (PGSLOTID);
```
ctshim: I think memcpy() is faster and safer than repeating type casting.
It is also free from alignment issues.

memcpy(record_p, oid_p, sizeof(OID));

If it is modified, it is better to modify the corresponding fhs_read*() series functions as well.

src/query/query_hash_scan.c
```diff
@@ -506,12 +666,2483 @@ int
...
+	}
+      else if (u1 > u2)
+	{
+	  compare_result = 1;
+	}
+      else
+	{
+	  compare_result = -1;
+	}
+      break;
```
ctshim: Since the value of out_compare_result_p will be judged only as positive, negative, or 0, there is no need to substitute 0, 1, -1, right?
How about modifying it like below?

```
*out_compare_result_p = *((unsigned int *) key_p) - *((unsigned int *) bucket_record_p);
return;
```

src/query/query_hash_scan.c
```diff
@@ -506,12 +666,2483 @@ int
...
+
+static FHS_HASH_KEY
+fhs_hash_four_bytes_type (char *key_p)
+{
+  FHS_HASH_KEY hash_key = 0;
+  unsigned int i;
+  unsigned short Short, *Short2;
+  char Char;
+
+  Short = ntohs (*(unsigned short *) key_p);	/* Get the left half of the int */
```
ctshim: Is there any reason to use the ntohs() function?

ctshim:
```
fhs_search (thread_p, **llsidp->hlsid.file**.hash_table, &hash_key, &oid, &**llsidp->hlsid.file**.curr_oid,
			&**llsidp->hlsid.file**.is_dk_bucket);
fhs_search_next (thread_p, `llsidp->hlsid`.file.hash_table, &`llsidp->hlsid`.curr_hash_key, &oid,
			     &`llsidp->hlsid`.file.curr_oid, llsidp->hlsid.file.is_dk_bucket);

```
			
When the above two functions are called, multiple parameters that pass members belonging to a common structure can be processed so that only one parameter is passed by passing the structure.
If you have a lot of calls, this might be more advantageous.

==> I fixed it.

ctshim:
fhs_hash(), fhs_compare_key() The data passed to this function is always int.
Do I need to check it by passing it along with key_type?
==> I fixed it. thanks.

src/query/scan_manager.c
```diff
@@ -8117,7 +8225,7 @@ check_hash_list_scan (LLIST_SCAN_ID * llsidp, int *val_cnt, int hash_list_scan_y
   UINT64 mem_limit = prm_get_bigint_value (PRM_ID_MAX_HASH_LIST_SCAN_SIZE);
 
   /* no_hash_list_scan sql hint check */
-  if (hash_list_scan_yn == 0)
+  if (hash_list_scan_yn == HASH_METH_NOT_USE)
```
hgryoo: The function signature should be like this?
from
```cpp
check_hash_list_scan (..., int hash_list_scan_yn)
```
to
```cpp
check_hash_list_scan (..., HASH_METHOD hash_list_scan_yn)
```

src/query/query_hash_scan.c
```diff
@@ -506,12 +666,2483 @@ int
...
+  else
+    {
+      result = fhs_insert_bucket_after_extend_if_need (thread_p, fhsid_p, &bucket_vpid, key_p, hash_key, value_p);
+      if (result == FHS_ERROR_OCCURRED)
+	{
+	  return NULL;
+	}
+      else if (result == FHS_BUCKET_FULL)
+	{
+	  return fhs_insert (thread_p, fhsid_p, key_p, value_p);
```
hgryoo, line 1992: If bucket is full, bucket is extended in `fhs_insert_bucket_after_extend_if_need()`. Can I know why fhs_insert should be called recursively in this case?

src/query/query_hash_scan.h
```diff
@@ -77,17 +90,41 @@ struct hash_scan_key
...
 };
 
+typedef struct file_hash_scan_id FHSID;
+struct file_hash_scan_id
+{
+  /* Fields should be ordered according to their sizes */
+  EHID ehid;			/* dir file identifier */
+  VFID bucket_file;		/* bucket file identifier */
+  DB_TYPE key_type;		/* type of the keys */
+  short depth;			/* global depth of the directory */
```
beyondykk9: how about unsigned short or unsigned char ?

src/query/scan_manager.c
```diff
@@ -7852,28 +7877,53 @@ scan_build_hash_list_scan (THREAD_ENTRY * thread_p, SCAN_ID * scan_id)
 	  new_key = key;
 	}
 
+      /* make hash key */
+      hash_key = qdata_hash_scan_key (new_key, UINT_MAX);
+
       /* create new value */
       if (llsidp->hlsid.hash_list_scan_yn == HASH_METH_IN_MEM)
```
beyondykk9: how about switch case statement?

src/query/scan_manager.c
```diff
@@ -8167,20 +8275,24 @@ check_hash_list_scan (LLIST_SCAN_ID * llsidp, int *val_cnt, int hash_list_scan_y
...
     {
       return HASH_METH_IN_MEM;
     }
   else if ((UINT64) llsidp->list_id->tuple_cnt * (sizeof (HENTRY_HLS) + sizeof (QFILE_TUPLE_SIMPLE_POS)) <= mem_limit)
     {
-      /* bytes of 1 row = sizeof(HENTRY_HLS) + sizeof(QFILE_TUPLE_SIMPLE_POS) = 36 bytes (64bit) */
-      /* HENTRY_HLS = pointer(8bytes) * 3 = 24 bytes */
+      /* bytes of 1 row = sizeof(HENTRY_HLS) + sizeof(QFILE_TUPLE_SIMPLE_POS) = 44 bytes (64bit) */
+      /* HENTRY_HLS = pointer(8bytes) * 4 = 32 bytes */
       /* SIMPLE_POS = pageid(4bytes) + voldid(2bytes) + padding(2bytes) + offset(4bytes) = 12 bytes */
```
beyondykk9: slip:
voldid --> volid

src/query/scan_manager.c
```diff
@@ -7852,28 +7877,53 @@ scan_build_hash_list_scan (THREAD_ENTRY * thread_p, SCAN_ID * scan_id)
...
+	      return S_ERROR;
+	    }
+	}
+      else if (llsidp->hlsid.hash_list_scan_yn == HASH_METH_HASH_FILE)
+	{
+	  /* curr_offset is int and slotid is short. */
+	  /* In fact, the offset is a position within a page(16K), so it can be stored as a short type. */
+	  oid.volid = llsidp->lsid.curr_vpid.volid;
+	  oid.pageid = llsidp->lsid.curr_vpid.pageid;
+	  oid.slotid = llsidp->lsid.curr_offset;
```
hgryoo: How about defining a new structure for hash ID instead of resuing OID with dummy `slotid`.
It could make confusion with the real OID.

```cpp
// for example
struct HFSID
{
  int volid;
  int pageid;
  int offset;
}
```

src/query/query_hash_scan.c
```diff
@@ -30,10 +40,160 @@
...
+
+#define FHS_HEADER_SLOT_ID	0
+#define FHS_FIRST_SLOT_ID	1
+
+/* key */
+typedef unsigned int FHS_HASH_KEY;
+#define FHS_KEY_TYPE DB_TYPE_INTEGER
+#define FHS_KEY_SIZE (sizeof (FHS_HASH_KEY))
+#define FHS_ALIGNMENT ((char) FHS_KEY_SIZE)
+#define FHS_MAX_DUP_KEY 100	/* 10% of MAXNUM (PAGE 16K / RECORD 14 bytes) */
```
ctshim, line 76: It would be nice to be able to change it to a variable and adjust the setting value later.
The waste of pages for DK can be extreme.

src/query/query_hash_scan.c
```diff
@@ -506,12 +666,2483 @@ int
...
+		  er_set (ER_FATAL_ERROR_SEVERITY, ARG_FILE_LINE, ER_EH_CORRUPTED, 0);
+		  return false;
+		}
+	    }
+	  return true;
+	}
+
+      if (compare_result < 0)
+	{
+	  high = middle - 1;
```
ctshim, line 1923: If the flag is positive, it means that the data in front of it is also not in the search range.
The following method seems to be more efficient when encountering duplicate data.

high = (flag > 0) ?   middle - flag  :  middle -1;

src/query/query_hash_scan.h
shparkcubrid: I fixed it.

src/query/scan_manager.c
shparkcubrid: Because of 'XASL to STREAM', it is better to keep hash_list_scan_yn as int.
In executor, the hash_list_scan_type is 'HASH_METHOD'.

src/query/query_hash_scan.c
shparkcubrid: This is to split the bucket again.
This is occurred when data is inserted into only one of the split buckets.

src/query/scan_manager.c
shparkcubrid: ok. I'll fix it

src/query/scan_manager.c
hgryoo: Then, how about adding assert()? to check the `hash_list_scan_yn` value is out of HASH_METHOD values possibly.

shparkcubrid: OK. I'll add assert().

src/query/query_hash_scan.h
shparkcubrid: I fixed it.

src/query/scan_manager.c
shparkcubrid: I fixed it.

byungwook-kim (approved):

src/query/scan_manager.c
shparkcubrid: I fixed it.

hgryoo (approved):

youngjinj (approved):

ctshim (changes requested):

src/query/query_hash_scan.c
shparkcubrid: The overflow occurs because out_compare_result_p is integer type.
I think now is better than changing to double type.

ctshim: Okay, I See.

src/query/query_hash_scan.c
shparkcubrid: Since the in front of bit is used first, it needs to be reversed.

ctshim: OK,

src/query/query_hash_scan.c
shparkcubrid: I fixed it.

src/query/query_hash_scan.c
shparkcubrid: I fixed it.

src/query/query_hash_scan.c
shparkcubrid: I disagree with making it a system parameter. Users will not be able to judge an appropriate that value. Also, since this depends on the data, there will be multiple cases on one system.
The reason I set it to 10% is that it is more important to prevent unnecessary splitting than to waste space in the DK bucket. The performance degradation caused by unnecessary splitting can occur on small data sets.

ctshim: I mean a variable that you want the query optimizer or plan stage to set. I thought it might be possible to adjust it using statistical information in the future.
Also, spliting and moving to the DK page seem to have similar workloads.

However, it does not necessarily need to be corrected, so you can proceed as you wish.

ctshim (approved):

src/query/query_hash_scan.c
shparkcubrid: thanks. I added comment for later work.

PR #2785, opened 2021-05-12
[CBRD-23925] When used as a function parameter, the original pointer variable is not initialized to NULL.

youngjinj:
http://jira.cubrid.org/browse/CBRD-23925

Purpose
- When initializing a pointer variable used as a parameter of a function to NULL, modify the original pointer variable to be initialized to NULL.

Implementation
- I changed the following two functions to macro functions.
  * src/query/cursor.c: cursor_free_list_id(), cursor_free_self_list_id()

Remarks
- N/A


byungwook-kim (approved):

ctshim (commented):
If you just modify the function's parameter declaration as shown below, it seems that you can get the same result as the current modification without modifying any other code at all.

```
void
cursor_free_list_id (QFILE_LIST_ID*  &list_id_p, bool self)

void
cursor_free_self_list_id (QFILE_LIST_ID* &list_id)
```

shparkcubrid (approved):

ctshim (commented):
Consider also changing the cursor_free_self_list_id() function to a macro like the one below.
I think I can minimize code changes this way
```

#define cursor_free_self_list_id(p) do {        \
        if((p)) {                               \
                cursor_free_list_id (list_id, true);  \
                list_id = NULL;                 \
        }                                       \
    } while(0)

```

ctshim (approved):
There seems to be no problem with the code apart from the comments I gave.

hgryoo (approved):

ctshim:
When trying to process using & in the parameter, I found that if type casting is needed, you can use reinterpret_cast.
See fnxx( **_reinterpret_cast<int*&>(pv)_** ); in the code below.

```
void fnxx(int* & pi)
{
        if(pi)
            free(pi);

        pi = (int*) malloc(32);
        pi[0] = 7;
}

void test()
{
        void* pv = (void*) malloc(64);
  
        printf("%08x\n", pv);
        fnxx( reinterpret_cast<int*&>(pv) );
        printf("%08x   %d\n",  pv, ((int*)pv)[0]);
}
```

Result)
```
00607010
00607060   7
```

youngjinj:
@ctshim

As you commented on, I want to change it to a macro function.

Both functions are not complex functions.
I agree that it is better to avoid unnecessary function calls and casting to double pointer variables.

However, it is different from the original PR content, so I will ask other people to review it again.

shparkcubrid (approved):

hgryoo (approved):

ctshim (approved):

byungwook-kim (approved):

PR #2789, opened 2021-05-14
[CBRD-23641] Collation compatible check between string parameters of expression/function does not work properly

byungwook-kim:
http://jira.cubrid.org/browse/CBRD-23641

COLLATE modifier does not convert a charset of expressions without using CAST(). Therefore, if collation compatibility of parameters should be checked for expression or function and the codeset of the string parameters is clearly different, incompatible collation error should occur regardless of using the collate modifier.

at **pt_common_collation**, it should be checked if corec_level or arguments are **PT_COLLATION_NOT_COERC**

### DIFF
--- a/src/parser/type_checking.c
+++ b/src/parser/type_checking.c
@@ -22969,10 +22969,13 @@ pt_common_collation (PT_COLL_INFER * arg1_coll_infer, PT_COLL_INFER * arg2_coll_
 {
 #define MORE_COERCIBLE(arg1_coll_infer, arg2_coll_infer)		     \
   ((((arg1_coll_infer)->can_force_cs) && !((arg2_coll_infer)->can_force_cs)) \
-   || ((arg1_coll_infer)->coerc_level > (arg2_coll_infer)->coerc_level	     \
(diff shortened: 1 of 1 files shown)


src/parser/type_checking.c
```diff
@@ -5444,8 +5444,8 @@ pt_coerce_range_expr_arguments (PARSER_CONTEXT * parser, PT_NODE * expr, PT_NODE
 	{
 	  PT_NODE *temp = NULL;
 	  int precision = 0, scale = 0;
-	  int units = LANG_SYS_CODESET; /* code set */
-	  int collation_id = LANG_SYS_COLLATION; /* collation_id */
+	  int units = LANG_SYS_CODESET;	/* code set */
+	  int collation_id = LANG_SYS_COLLATION;	/* collation_id */
```
beyondykk9, line 5448: it is an unrelated change caused by broken code formatting at any previous commit.

beyondykk9:
This issue is not a bug, but it will be at the next version with refactoring for character set and collation handling.

PR #2790, opened 2021-05-14
[CBRD-23952] Deal with partitioning table when TRUNCATE using heap destruction

wrlawodms:
http://jira.cubrid.org/browse/CBRD-23952

See the issue for more information.

### DIFF
--- a/src/object/schema_manager.c
+++ b/src/object/schema_manager.c
@@ -15808,7 +15808,7 @@ sm_truncate_class_internal (MOP class_mop)
       assert (class_->users);
       for (subs = class_->users; subs; subs = subs->next)
 	{
-	  error = au_fetch_class (subs->op, &subclass, AU_FETCH_READ, AU_SELECT);
+	  error = au_fetch_class (subs->op, &subclass, AU_FETCH_READ, DB_AUTH_ALTER);
 	  if (error != NO_ERROR)
 	    {
 	      goto error_exit;
--- a/src/object/schema_manager.c
+++ b/src/object/schema_manager.c
@@ -15651,7 +15651,7 @@ sm_collect_truncatable_classes (MOP class_mop, std::unordered_set < OID > &trun_
   if (partition_type == DB_PARTITION_CLASS)
     {
       /*
-       * No need to be added:
+       * No need to be added to trun_classes:
        *  Partitioning classes will be truncated while truncating partitioned class of them.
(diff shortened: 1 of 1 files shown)


src/object/schema_manager.c
```diff
@@ -15641,21 +15645,58 @@ sm_collect_truncatable_classes (MOP class_mop, std::unordered_set < OID > &trun_
...
+	  if (error != NO_ERROR)
+	    {
+	      return error;
+	    }
+	  if (subclass->partition)
+	    {
+	      trun_classes.emplace (*ws_oid (subs->op));
+	    }
+	}
+      //yes, fallthrough
```
hornetmj: I think the 'if statement' is more readable because of this.

src/object/schema_manager.c
```diff
@@ -15641,21 +15645,58 @@ sm_collect_truncatable_classes (MOP class_mop, std::unordered_set < OID > &trun_
...
+      /* Find partition classes if exists */
+      assert (class_->users);
+      for (subs = class_->users; subs; subs = subs->next)
+	{
+	  error = au_fetch_class (subs->op, &subclass, AU_FETCH_READ, DB_AUTH_ALTER);
+	  if (error != NO_ERROR)
+	    {
+	      return error;
+	    }
+	  if (subclass->partition)
```
hornetmj: Is the value of 'subclass->partition' always not null in this case? If it is, how about use the assert function?

hornetmj (approved):

src/object/schema_manager.c
wrlawodms: Yes, you're right. It is a pre-condition rather than an if-condition. I'll change it to assertion.

src/object/schema_manager.c
wrlawodms: OK, I agree

src/object/schema_manager.c
wrlawodms: I mimicked some code in `do_drop_partitioned_class()`, and it seems to assume that partitioned class can have child classes which are not partition class. However, it is impossible to inherit a partitioned table. Assuming it is a legacy code, I've decided not to consider that case.

wrlawodms: Or we can be conservative that just leave it like before with the assumption that another feature but partitioning uses inheritance hierarchy.

hornetmj (approved):

PR #2791, opened 2021-05-14
[CBRD-23930] Fix to check fk info even if the fk-child is already checked

wrlawodms:
http://jira.cubrid.org/browse/CBRD-23930

The `fk_info` has to be checked even if fk child class is already checked. Even if the PK of fk-child is checked, FKs of fk-child is not yet checked. Imaging that you do TRUNCATE A, and there is A->B->A FK-cycle (->: refers to by a FK). 
1. the FK of A checked, and A is put into `trun_classes`.
2. then FK of B is checked, and B is put into `trun_claases`.
3. Although A is already in `trun_claasses`, the cascade action of B->A has to be checked.


wrlawodms:
Some new failures in test_sql are due to partitioning tables, I'll test and merge this PR after #2790 is merged.

hornetmj (approved):

PR #2792, opened 2021-05-14
[LETS-100] Remove start/end checkpoint log records related code

ribram:
http://jira.cubrid.org/browse/LETS-100

**Code cleanup**
- Remove all LOG_START_CHKPT and LOG_END_CHKPT usages
- Replace log_info_chkpt_trans and log_info_chkpt_sysop with checkpoint_info::tran_info and checkpoint_info::sysop_info (and move to log_checkpoint_info.hpp)
- Remove prior_lsa_gen_end_chkpt_record
- Remove log_dump_record_checkpoint, logpb_dump_checkpoint_trans, log_dump_checkpoint_topops

**Other**
- Fix bad indentations in log_2pc.c
- Remove const_cast in log_2pc_read_prepare.
- log_checkpoint_info.cpp: Use const references when iterating instead of copies.


src/transaction/log_checkpoint_info.cpp
```diff
@@ -167,7 +168,7 @@ namespace cublog
       }
 
     size += serializator.get_packed_bigint_size (start_offset + size);
-    for (const checkpoint_sysop_info sysop_info : m_sysops)
+    for (const auto &sysop_info : m_sysops)
```
cristiarg, line 171: seems like there are more copies being done:
~295
```
for (auto chkpt : m_trans)
```
~344
```
for (auto sysop : m_sysops)
```
~394
```
for (auto chkpt : m_trans)
```
or am I mistaken?

src/transaction/log_2pc.c
```diff
@@ -1422,18 +1425,16 @@ log_2pc_read_prepare (THREAD_ENTRY * thread_p, int acquire_locks, log_tdes * tde
...
+  const LOG_REC_2PC_PREPCOMMIT *prepared;	/* A 2PC prepare to commit log record */
   LK_ACQUIRED_LOCKS acq_locks;	/* List of acquired locks before the system crash */
   int size;
 
   log_pgptr_reader.advance_when_does_not_fit (sizeof (*prepared));
-
-  // *INDENT-OFF*
-  prepared = const_cast<LOG_REC_2PC_PREPCOMMIT*> (log_pgptr_reader.reinterpret_cptr<LOG_REC_2PC_PREPCOMMIT> ());
-  // *INDENT-ON*
+  prepared = log_pgptr_reader.reinterpret_cptr<LOG_REC_2PC_PREPCOMMIT> ();
```
cristiarg, line 1437: could even be `const ... *const prepared ...` but given it's a pure C file, maybe stick to C conventions

src/transaction/log_checkpoint_info.cpp
ribram: No, you're right. I'll push a fix.

src/transaction/log_2pc.c
ribram: Could be, but I'd rather keep the current code. Imo, making a variable that doesn't change const is optional, unless it is meant not to change under any circumstances :). The value of the pointer here is not meant to be changed under any circumstances, the pointer itself could be changed.

cristiarg (approved):

adrianbrab (approved):

PR #2793, opened 2021-05-18
[CBRD-23933] Simplify the bit property variables in the structure by making it a separate new structure.

ctshim:
http://jira.cubrid.org/browse/CBRD-23933

Purpose
 In the three structures below, member variables with bit properties are grouped into a struct and the expression method is changed.

- PT_NODE
- PT_QUERY_INFO
- PARSER_CONTEXT

This is an improvement because it is difficult to see multiple pieces of information at the same time because there are many lists in the structure in tasks such as debugging.
There is no function change.

Implementation
N/A

Remarks
N/A


shparkcubrid (approved):

hgryoo:
Can we consider a more meaningful variable name instead of `bs`?
ex) `opt`, `hint`..

hgryoo (approved):

youngjinj (approved):

beyondykk9 (approved):

ctshim:
The reason why the "github-checks / code-style" item was not resolved is not this modification, but because the indent did not match when reflecting the existing code.

youngjinj:
image

The use of spaces and tabs is different.

PR #2794, opened 2021-05-18
[LETS-84] Wraper class for request client + request send queue + queue autosend

cristiarg:
http://jira.cubrid.org/browse/LETS-84

**Description**
Request sending between servers will often be handled by the next trio:
- `request_client_server` that handles requests, but is not thread safe
- `request_sync_send_queue` that synchronizes request sending
- `request_queue_autosend` that automatically sends everything that is added to a `request_sync_send_queue`

This kind of trio is already present in the active transaction server (for the connection with the page server) and on the page server (for the connection with the active tran server).
New such trios will be required for the additional connections on the ATS with multiple PS.

**Implementation**
- new class `cubcomm::request_sync_client_server` that wraps `request_client_server`, `request_sync_send_queue` and `request_queue_autosend`
- class is templatized on the message types that `request_client_server` will send and receive
- another template argument is the payload that `request_sync_send_queue` will send
- a state-machine like interface is exposed with following sequences
    - initialization: ctor, `init`, `register_request_handler`  and `connect` - these are supposed to be called in order
    - execution: `push`
    - termination: `disconnect` (to be explicitly called before destruction), dtor
- replace with new wrapper class in `request_sync_client_server` in `page_server` and `active_tran_server`

**Testing**
- manual tests
- a unit test in `unit_tests/request_client_server`

PR #2795, opened 2021-05-19
[LETS-115] Reset nxio_lsa to append_lsa during boot/recovery

ribram:
http://jira.cubrid.org/browse/LETS-115

Pages are requested during boot/recovery, before any log being flushed to disk and thus, before setting any value to nxio_lsa. These requests are sent with a null target replication LSA and trigger an assert on the page server.

The pages need to be sent with a valid target replication LSA. E.g. if page server replication is a couple of seconds behind, and if the transaction server restarts, it needs to wait until all the changes in the requested pages are replicated.

nxio_lsa is the LSA after the last log record written to disk. During boot/recovery, append LSA is reset to the same value with the  
LOG_RESET_APPEND_LSA function. Same function is used to also set the value of nxio_lsa.

EDIT: Some pages (e.g. data volume headers) are requested before the log module is initialized. The transaction server has no idea at this point what target LSA to set for its data page requests. For now, the wait for replication is disabled when target_lsa is null. In the future, the transaction server should set a good target from the start (probably with the help of page server, by obtaining some information from there).

### DIFF
--- a/src/storage/async_page_fetcher.cpp
+++ b/src/storage/async_page_fetcher.cpp
@@ -68,7 +68,7 @@ namespace cublog
 
   void data_page_fetch_task::execute (context_type &context)
   {
-    if (m_lsa.is_null ())
+    if (!m_lsa.is_null ())
       {
 	// TODO: FIXME
 	// The transaction server boots and reads pages before initializing its log module and before knowing a safe target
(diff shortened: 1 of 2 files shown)


danilailie (approved):

cristiarg (approved):
Tested with bash/csql script against changes from #2794. Works.

danilailie (approved):

PR #2800, opened 2021-05-21
[CBRD-23963] Parallel make for openssl external library project

hgryoo:
http://jira.cubrid.org/browse/CBRD-23963

The following warning is shown when building libopenssl with externalproject_add().
```
make[4]: warning: jobserver unavailable: using -j1.  Add '+' to parent make rule.
```
It causes a bottleneck in build. This warning can be fixed by setting the BUILD_COMMAND to `$(MAKE)` instead of `make`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1172,7 +1172,6 @@ if(WITH_JDBC AND EXISTS ${CMAKE_SOURCE_DIR}/cubrid-jdbc/src)
   )
 endif()
 
-
 # include subdirectories
 add_subdirectory(cas)
 add_subdirectory(sa)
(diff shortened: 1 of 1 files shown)


wrlawodms:
GOOOD. Have you considered BUILD_COMMANDs for other libraries like LIBJANSSON, LIBEDIT, and so on? It would not a bottleneck regarding that you only modify it for OpenSSL, but it seems better to change other BUILD_COMMAND to use `(MAKE)`. What do you say?

hgryoo:
@wrlawodms 
I'm going to proceed with them on other issues. But now, the build of other libraries seems not a bottleneck.

wrlawodms (approved):
@hgryoo 
Ok, we can deal with other libraries when it is needed later.

kisoo-han (approved):
Thank you for your efforts to improve the development environment. The build time has been reduced a lot.

mhoh3963 (approved):

PR #2801, opened 2021-05-21
[CBRD-23938] Drop the PK referred to by a FK while TRUNCATE

wrlawodms:
http://jira.cubrid.org/browse/CBRD-23938

The goal of this issue is to allow the PK referred by FKs can be dropped, but the main work of this PR is to make TRUNCATE in a vertical way and refactoring for it. See the Jira issue for more information.


hgryoo:
How about extracting the refactored C++ code into a separate file? 
The code lines of **schema_manager.c** are too long enough. And skipping indent for whole classes and member function by `// *INDENT-OFF*` comment will make it hard for you to keep the code format whenever you modify them.

wrlawodms:
@hgryoo Your advice is right. Plus, I looked around other .cpp files in CUBRID, and confirm that usually one class has its own file. So, I split the code into a separate .cpp file. Thank you.

During extracting the cpp code, I've also extracted the mediating processing into a class that contains collecting truncating classes and doing truncate all classes. I've named it `cubschema::class_truncator` and renamed the original `sm_class_truncator` to `cubschema::class_truncate_context`.

src/object/schema_class_truncator.hpp
```diff
@@ -0,0 +1,48 @@
...
+
+#include <unordered_set>
+
+namespace cubschema
+{
+  /*
+   * In charge of truncating a class.
+   * During that, it can truncate several classes related to the class with FK or partitioning.
+   */
+  class class_truncator
```
hornetmj, line 33: I think the prefix "class_" is not necessary.

src/object/schema_class_truncator.cpp
```diff
@@ -0,0 +1,634 @@
...
+
+	    MOP fk_child_mop = ws_mop (&fk_ref->self_oid, NULL);
+	    if (fk_child_mop == NULL)
+	      {
+		assert (er_errid () != NO_ERROR);
+		return er_errid ();
+	      }
+
+	    int partition_type = DB_NOT_PARTITIONED_CLASS;
+	    error = sm_partitioned_class_type (fk_child_mop, &partition_type, NULL, NULL);
```
hornetmj, line 121: I think you need to check the truncate target table when collect_trun_classes() is first called.

eg.) truncate table [target table: partitioned table];

It looks like checking only the fk-child tables. And I think it makes confuse to code reader even If it is not acceptable caused for an architectural reason. If so, I request you to leave a comment about it.

src/object/schema_class_truncator.cpp
```diff
@@ -0,0 +1,634 @@
...
+	    if (error != NO_ERROR)
+	      {
+		return error;
+	      }
+
+	    if (partition_type == DB_PARTITION_CLASS)
```
hornetmj, line 127: "m_trun_classes.find (fk_ref->self_oid)" has already been run and passed. It means there is 'DB_PARTITION_CLASS' class.

src/object/schema_class_truncator.cpp
wrlawodms: If you mean the case in which the target table is  '**partition** table', it is already filtered out in a caller, `do_truncate()`. If it is what you talked about, yes, you're right, I'll add a comment for it.

src/object/schema_class_truncator.cpp
wrlawodms: Here, `m_trun_classes.find (fk_ref->self_oid) == m_trun_classes.end ()` is true, which means the `fk_child_mop` has not yet been in `m_trun_classes`. And, if it is a partition class, it is skipped to add. Could you elaborate on what I'm missing?

src/object/schema_class_truncator.hpp
wrlawodms: In my opinion, it'd be better to keep the prefix. Schema Manager deals with not only Class, but also Domain, Constraint, and so on. 'truncate' seems quite obviously an operation for a class, but considering other operations like `sm_rename_class()` and `sm_delete_class_mop()`, it seems necessary to specify the target for the future.

src/object/schema_class_truncator.cpp
```diff
@@ -0,0 +1,634 @@
...
+	};
+
+	error = context.save_constraints_or_clear (cons_predicate);
+	if (error != NO_ERROR)
+	  {
+	    return error;
+	  }
+
+	/* FK must be dropped earlier than PK, because of referencing */
+	auto saved_cons_predicate = [] (const SM_CONSTRAINT_INFO& cons_info) -> bool
```
hornetmj: Is this lambda function necessary? I think it is enough to check whether or not m_*_info is NULL. Is there anything else I have to re-consider?

src/object/schema_class_truncator.cpp
```diff
@@ -0,0 +1,634 @@
...
+    return error;
+  }
+
+  /*
+   * truncate_heap () - truncate the heap of the class.
+   *
+   *   return: error code or NO_ERROR
+   */
+  int
+  class_truncate_context::truncate_heap ()
```
hornetmj, line 500: This function looks like it always executes destroying the heap because of the name of this function. Do you have any plan to reconsider when you run the sub-task for DONT_REUSE_OID?

wrlawodms: I don't think that "truncate" always means "destroying the heap of a class".

src/object/schema_class_truncator.cpp
wrlawodms: You're right. It is not necessary, but I thought it is more readable, and otherwise, we have to write this like: 
```
auto saved_cons_predicate = [] (const SM_CONSTRAINT_INFO& cons_info) -> bool
{
    return true;
};
```
However, while considering your review, I've found out that `std::function` can receive `nullptr`. With this, we can make these kinds of functions can be called without a predicate. It looks better to me. What do you think?

src/object/schema_class_truncator.hpp
hornetmj: @wrlawodms For now, I can't imagine that we will make domain_truncator or constraint_truncator because the behavior of truncate is related to the purpose of truncator class obviously. There is a similar example I think. We don't need to class_replicator and serial_replicator and so on. We just make replicator class and it deals with all kinds of replication. This conversation related to the class design and it can be different by person.

src/object/schema_class_truncator.cpp
hornetmj: @wrlawodms Thanks. But, I reconsider after leaving this comment. And I understand the mechanism more precisely. There are no need changes and comments.

wrlawodms: OK!

src/object/schema_class_truncator.hpp
wrlawodms: Me neither. I can't imagine `doamin_truncator` and `constraint_truncator`, too. But, I can imagine `class_deleter` as an alternative of `sm_delete_class_mop()`, and I think we can't name it just `deleter`. Considering other operations for the class, I think that we should use the prefix.

src/object/schema_class_truncator.cpp
wrlawodms: What I mentioned is: [cb66b4d](https://github.com/CUBRID/cubrid/pull/2801/commits/cb66b4d9b6520ac81ae6b323a29000f55fad29dc). We can revert it.

src/object/schema_class_truncator.hpp
hornetmj: @wrlawodms I don't want to design(make) a class just related to some kind of existing functions. If you make a 'deleter' as a class, then it can extend(include) to class_deleter and index_deleter, and so on. But, I don't want to push you to follow my thought and If you don't agree with me, then you can use the 'class_' prefix without any hesitation. I just want to share my though(opinion) with you and vice versa.

src/object/schema_class_truncator.cpp
hornetmj: @wrlawodms Yes, I mean exactly what you say. But, don't you think the name of this function 'truncate_heap' makes us think "destroying the heap of a class"? How about "truncate_class ()"?

src/object/schema_class_truncator.cpp
hornetmj: Sorry, I think someone punched my head in the morning (maybe my 3-rd son). It is a wrong comment that should be thrown into the trash.

hornetmj (approved):

src/object/schema_class_truncator.cpp
wrlawodms: Regarding the callees in this function `sm_truncate_using_destroy_heap()` and `sm_truncate_using_delete()`, it doesn't seem to include the context of destroying heap.

PR #2802, opened 2021-05-21
[CBRD-23955] Speed improvement of pt_find_keyword() function

ctshim:
http://jira.cubrid.org/browse/CBRD-23955

Purpose
   -Improves the speed of pt_find_keyword() for checking reserved keywords.

Implementation
   - We changed the sorting policy for bipartite search.
   - Previously, only keyword string comparison was performed,
   - Use the structure changed to (Hash Value, Keyword String).
   - We narrowed our search by adding an array to manage the scope of the search.
  -  By using the intl_identifier_upper() function, the part that was converted to upper case was removed and changed to upper case directly.

 -  The "value" variable in struct keyword_record is meaningful only when ENABLE_UNUSED_FUNCTION is declared.
 - ENABLE_UNUSED_FUNCTION will not be declared.
 - To minimize code changes, the existing structure value initialization part is kept as it is (because the types are compatible).

Remarks
  -N/A

### DIFF
--- a/src/parser/keyword.c
+++ b/src/parser/keyword.c
@@ -526,22 +526,26 @@ static KEYWORD_RECORD keywords[] = {
 static KEYWORD_RECORD *pt_find_keyword (const char *text);
 static int keyword_cmp (const void *k1, const void *k2);
 
-#define GET_KEYWORD_HASH_VALUE(h,s)  do {       \
-  unsigned char* p = (unsigned char*)(s);       \
-  for((h) = 5381;  *p; p++ )                    \
-    {                                           \
-      (h) = (((h) << 5) + (h)) + *p; /* hash * 33 + c */ \
-    }                                           \
-}while(0)
+#define GET_KEYWORD_HASH_VALUE(h,s)             \
+  do {                                          \
+      unsigned char* p = (unsigned char*)(s);   \
+      for((h) = 5381;  *p; p++ )                \
+        {                                       \
+             (h) = (((h) << 5) + (h)) + *p; /* hash * 33 + c */ \
+        }                                       \
+  }while(0)
 
 static int
 keyword_cmp (const void *k1, const void *k2)
 {
   if (((KEYWORD_RECORD *) k1)->hash_value > ((KEYWORD_RECORD *) k2)->hash_value)
-    return 1;
+    {
+      return 1;
+    }
   if (((KEYWORD_RECORD *) k1)->hash_value < ((KEYWORD_RECORD *) k2)->hash_value)
-    return -1;
-
+    {
+      return -1;
+    }
   return strcmp (((KEYWORD_RECORD *) k1)->keyword, ((KEYWORD_RECORD *) k2)->keyword);
 }
 
@@ -568,9 +572,13 @@ pt_find_keyword (const char *text)
 	{
 	  len = strlen (keywords[i].keyword);
 	  if (len < keyword_min_len)
-	    keyword_min_len = len;
+	    {
+	      keyword_min_len = len;
+	    }
 	  if (len > keyword_max_len)
-	    keyword_max_len = len;
+	    {
+	      keyword_max_len = len;
+	    }
 
 	  GET_KEYWORD_HASH_VALUE (keywords[i].hash_value, keywords[i].keyword);
 	}
@@ -611,12 +619,18 @@ pt_find_keyword (const char *text)
   for (p = (unsigned char *) text; *p; p++, s++)
     {
       if (*p >= 0x80)
-	return NULL;
+	{
+	  return NULL;
+	}
 
       if (*p >= 'a' && *p <= 'z')
-	*s = *p + ('A' - 'a');
+	{
+	  *s = *p + ('A' - 'a');
+	}
       else if (*p < 0x80)
-	*s = *p;
+	{
+	  *s = *p;
+	}
     }
   *s = 0x00;
 #else
@@ -629,16 +643,24 @@ pt_find_keyword (const char *text)
     {
       for (len = start_pos[i]; len < start_pos[i + 1]; len++)
 	{
-	  if (dummy.hash_value < keywords[len].hash_value)
-	    continue;
-	  else if (dummy.hash_value > keywords[len].hash_value)
-	    return NULL;
+	  if (dummy.hash_value > keywords[len].hash_value)
+	    {
+	      continue;
+	    }
+	  else if (dummy.hash_value < keywords[len].hash_value)
+	    {
+	      return NULL;
+	    }
 
 	  cmp = strcmp (dummy.keyword, keywords[len].keyword);
 	  if (cmp < 0)
-	    continue;
+	    {
+	      continue;
+	    }
 	  else if (cmp > 0)
-	    return NULL;
+	    {
+	      return NULL;
+	    }
 
 	  return keywords + len;
 	}
--- a/src/parser/keyword.c
+++ b/src/parser/keyword.c
@@ -541,14 +541,13 @@ static int keyword_cmp (const void *k1, const void *k2);
 static int
 keyword_cmp (const void *k1, const void *k2)
 {
-  if (((KEYWORD_RECORD *) k1)->hash_value > ((KEYWORD_RECORD *) k2)->hash_value)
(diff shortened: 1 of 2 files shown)


src/parser/keyword.c
```diff
@@ -526,10 +526,22 @@ static KEYWORD_RECORD keywords[] = {
 static KEYWORD_RECORD *pt_find_keyword (const char *text);
 static int keyword_cmp (const void *k1, const void *k2);
 
+#define GET_KEYWORD_HASH_VALUE(h,s)  do {       \
+  unsigned char* p = (unsigned char*)(s);       \
+  for((h) = 5381;  *p; p++ )                    \
+    {                                           \
+      (h) = (((h) << 5) + (h)) + *p; /* hash * 33 + c */ \
+    }                                           \
+}while(0)
```
hgryoo: For code consistency, `do { } while (0)` format as follow.
```suggestion
#define GET_KEYWORD_HASH_VALUE(h,s) \
  do { \
    unsigned char* p = (unsigned char*)(s); \
    for((h) = 5381; *p; p++ ) \
      { \
        (h) = (((h) << 5) + (h)) + *p; /* hash * 33 + c */ \
      } \
} while(0);
```

What is the meaning of 5381?

src/parser/keyword.c
```diff
@@ -526,10 +526,22 @@ static KEYWORD_RECORD keywords[] = {
...
 
 static int
 keyword_cmp (const void *k1, const void *k2)
 {
+  if (((KEYWORD_RECORD *) k1)->hash_value > ((KEYWORD_RECORD *) k2)->hash_value)
+    return 1;
+  if (((KEYWORD_RECORD *) k1)->hash_value < ((KEYWORD_RECORD *) k2)->hash_value)
+    return -1;
```
hgryoo: brace `{`, `}` needs even for single statement

src/parser/keyword.c
```diff
@@ -556,18 +598,55 @@ pt_find_keyword (const char *text)
...
+  s = (unsigned char *) dummy.keyword;
+  for (p = (unsigned char *) text; *p; p++, s++)
+    {
+      if (*p >= 0x80)
+	return NULL;
+
+      if (*p >= 'a' && *p <= 'z')
+	*s = *p + ('A' - 'a');
+      else if (*p < 0x80)
+	*s = *p;
```
hgryoo: converting to uppercase already implemented at `char_toupper() in chartype.c`

hgryoo (commented):
Could you elaborate new searching method using hash?

src/parser/keyword.c
ctshim: That function is slow.
Also, here we try to return NULL immediately if it is not an ASCII character.

src/parser/keyword.c
ctshim: OK, I do.

src/parser/keyword.c
ctshim: I referenced the existing string_hash().
It is known that there have been research results showing that the distribution efficiency of hash values is good when the algorithm starts with that value when the initial value is 5381.

src/parser/keyword.c
hgryoo: Returning NULL after checking a character is not related to the function, `char_toupper()`. It would better reuse it to remove duplicated code.

hgryoo:
If `intl_identifier_upper()` is going to removed, `#include "intl_support.h"` also should be removed too

src/parser/keyword.c
```diff
@@ -33,6 +33,7 @@
 #include "intl_support.h"
 #include "dbtype.h"
 #include "string_opfunc.h"
+#include "../base/chartype.h"
```
hgryoo:
```suggestion
#include "chartype.h"
```

src/parser/keyword.c
ctshim: OK,  I will fix it.

src/parser/keyword.c
ctshim: OK, thank you!

src/parser/keyword.c
hgryoo: 1) we need a reference and description about this value. please add comment
2) please keep the code format for macro function same with others
`}while(0)` -> `} while(0)`
```
#define GET_KEYWORD_HASH_VALUE(h,s)  do {       \

to

#define GET_KEYWORD_HASH_VALUE(h,s) \  
    do { \
```

src/parser/parse_tree.h
```diff
@@ -3515,7 +3515,11 @@ struct execution_state_values
 typedef struct keyword_record KEYWORD_RECORD;
 struct keyword_record
 {
+#if defined(ENABLE_UNUSED_FUNCTION)
```
beyondykk9, line 3518: this macro should be only at the function to be used sometime in the future. I think this macro is not needed.

src/parser/keyword.c
ctshim: When defining a macro, it is common not to add';' at the end of the macro for safety reasons.

hgryoo: Sorry for typo. I changed the example above. It's about formatting

src/parser/keyword.c
```diff
@@ -527,9 +527,28 @@ static KEYWORD_RECORD *pt_find_keyword (const char *text);
...
 keyword_cmp (const void *k1, const void *k2)
 {
+  if (((KEYWORD_RECORD *) k1)->hash_value > ((KEYWORD_RECORD *) k2)->hash_value)
+    {
+      return 1;
+    }
+  if (((KEYWORD_RECORD *) k1)->hash_value < ((KEYWORD_RECORD *) k2)->hash_value)
+    {
+      return -1;
+    }
```
beyondykk9, line 550:
```suggestion
    int cmp = ((KEYWORD_RECORD *) k1)->hash_value - ((KEYWORD_RECORD *) k2)->hash_value;
    
    if (cmp == 0)
      {
        cmp = strcmp (((KEYWORD_RECORD *) k1)->keyword, ((KEYWORD_RECORD *) k2)->keyword);
      }
      
    return cmp;
```

src/parser/keyword.c
ctshim: I don't know the exact meaning of the numbers 5381 and 33.
It is a number included in the djb2 algorithm, and I think you can get some related information from the link below.

https://stackoverflow.com/questions/10696223/reason-for-5381-number-in-djb-hash-function/13809282#13809282

src/parser/keyword.c
ctshim: I will change it.
Thanks for the comments.

src/parser/parse_tree.h
ctshim: This will remain as it is for now.

src/parser/keyword.c
ctshim: @hgryoo  OK, I will change indent on macro.

src/parser/keyword.c
```diff
@@ -527,9 +527,28 @@ static KEYWORD_RECORD *pt_find_keyword (const char *text);
...
 
+/* The GET_KEYWORD_HASH_VALUE() macro is the definition of the djb2 algorithm as a macro.
+ * Refer to the string_hash() function implemented in the libcubmemc.c file.
+ */
+#define GET_KEYWORD_HASH_VALUE(h,s) \
+  do { \
+      unsigned char* p = (unsigned char*)(s); \
+      for((h) = 5381;  *p; p++ ) \
+        { \
+             (h) = (((h) << 5) + (h)) + *p; /* hash * 33 + c */ \
```
youngjinj, line 538: When using the djb2 hash function, use the unsigned long type.

However, the hash_value variable of the KEYWORD_RECORD struct uses an unsigned short type.

Is there any possibility of overflow while repeating h * 33?

youngjinj: I looked for content that overflows often occur when using a hash function. But isn't the unsigned short type small enough to cause frequent collisions?

Since the keyword is not long, is it okay if it is small?

ctshim: The hash value does not matter if overflow occurs.
Even if overflow occurs, all you need to do is show an even distribution.

youngjinj (approved):

shparkcubrid (approved):

beyondykk9 (approved):

hgryoo (approved):

PR #2808, opened 2021-05-24
[CBRD-23964] Drop constraints of Partitioning table while TRUNCATE

wrlawodms:
http://jira.cubrid.org/browse/CBRD-23964

### DIFF
--- a/src/object/schema_manager.c
+++ b/src/object/schema_manager.c
@@ -15025,7 +15025,20 @@ sm_is_possible_to_recreate_constraint (MOP class_mop, const SM_CLASS * const cla
 
   if (class_->users != NULL)
     {
-      return false;
+      if (class_->partition != NULL)
+	{
+	  /*
+	   * partitioned class
+	   *
+	   * if there is a child class, it can be shared,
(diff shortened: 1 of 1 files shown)


src/object/schema_manager.c
```diff
@@ -15025,7 +15025,20 @@ sm_is_possible_to_recreate_constraint (MOP class_mop, const SM_CLASS * const cla
 
   if (class_->users != NULL)
     {
-      return false;
+      if (class_->partition != NULL)
```
hornetmj, line 15028: Not related to this line.

Can you explain why other "constraint->type"s (eg. SM_CONSTRAINT_PRIMARY_KEY) are impossible to recreate?
- I want to know why it is impossible conceptually.
- and I want to know the exact criterion between possible and impossible.

wrlawodms: 1. Basically, indexes on a class in a hierarchy can share an index. Do you want me to elaborate on the cases using a shared index?
2. the `SM_CONSTRAINT_PRIMARY_KEY` case doesn't have to do with the Shared index. This seems to be conceptually possible to drop if FKs referring to it drops in advance or with ignorance of them.

PR #2810, opened 2021-05-24
[CBRD-23953] Rewrite the initialization function of PT_STATEMENT_INFO…

ctshim:
http://jira.cubrid.org/browse/CBRD-23953

Purpose
  Changed initialization processing for structures included in pt_init_f.
  Changed from the method of initializing individual variables to the form using memset().

Implementation
   The part that was coded using pt_init_func_array in the pt_init_node() function has been rewritten using pt_init_f for consistency.

Remarks
  N/A

### DIFF
--- a/src/parser/parse_tree_cl.c
+++ b/src/parser/parse_tree_cl.c
@@ -118,9 +118,9 @@ struct pt_tree_copy_info
   PT_CTE_COPY_INFO *cte_structures_list;
 };
 
-PARSER_INIT_NODE_FUNC *pt_init_f = NULL;
-PARSER_PRINT_NODE_FUNC *pt_print_f = NULL;
-PARSER_APPLY_NODE_FUNC *pt_apply_f = NULL;
+static PARSER_INIT_NODE_FUNC *pt_init_f = NULL;
+static PARSER_PRINT_NODE_FUNC *pt_print_f = NULL;
+static PARSER_APPLY_NODE_FUNC *pt_apply_f = NULL;
 PARSER_CONTEXT *parent_parser = NULL;
 
 static void strcat_with_realloc (PT_STRING_BLOCK * sb, const char *tail);
@@ -5217,6 +5217,7 @@ pt_init_node (PT_NODE * node, PT_NODE_TYPE node_type)
       return;
     }
 
+  memset (&(node->info), 0x00, sizeof (node->info));
   (pt_init_f[node_type]) (node);
   node->node_type = node_type;
 }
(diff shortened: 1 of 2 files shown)


src/parser/parse_tree_cl.c
```diff
@@ -5209,13 +5209,15 @@ pt_init_print_f (void)
 void
 pt_init_node (PT_NODE * node, PT_NODE_TYPE node_type)
 {
-  PARSER_INIT_NODE_FUNC parser_init_node_func = pt_init_func_array[node_type];
-  if (!node || !parser_init_node_func)
+  assert (node_type < PT_LAST_NODE_NUMBER);
+  assert (pt_init_f != NULL);
```
beyondykk9, line 5213: It seems not to be needed because pt_init_f is initialized at parser creating.
I wonder why you use **pt_init_f**  instead of **parser_init_node_func**

ctshim: In either case, if pt_init_f is non-null, the parser_init_node_func array will match the appropriate functions.
However, if pt_init_f is null, the parser_init_node_func array contains pointers to meaningless garbage values. That is, it has not been initialized.
So I think pt_init_f is more appropriate.
And there is only one place here that uses parser_init_node_func

beyondykk9: but by assert (pt_init_f != NULL) it is proved that pt_init_f is not null.

shparkcubrid (approved):

youngjinj (approved):

beyondykk9 (approved):

src/parser/parse_tree_cl.c
```diff
@@ -6392,10 +6394,7 @@ pt_apply_alter_trigger (PARSER_CONTEXT * parser, PT_NODE * p, PT_NODE_FUNCTION g
 static PT_NODE *
 pt_init_alter_trigger (PT_NODE * p)
 {
-  p->info.alter_trigger.trigger_owner = NULL;
-  p->info.alter_trigger.trigger_priority = NULL;
-  p->info.alter_trigger.trigger_spec_list = NULL;
-  p->info.alter_trigger.comment = NULL;
+  memset (&(p->info.alter_trigger), 0x00, sizeof (p->info.alter_trigger));
```
hgryoo: Instead of writing the same code for each `init_*` function, how about applying `memset()` for `PT_STATEMENT_INFO union`? (Specifically, `p->info`)

hgryoo: `memset (node->info, 0x00, sizeof (p->info));` could be added before `(pt_init_f[node_type]) (node);` in `pt_init_node()` function.

ctshim: Not all init_*() functions were modified, so it was not possible to perform memset() in batches in the pt_init_node() function.
Instead, it might be possible to do memset(node->info) inside each init_*() function.
However, it is not sensitive, so I will try to keep it as it is.

beyondykk9 (approved):

hgryoo (approved):

shparkcubrid (approved):

youngjinj (approved):

PR #2814, opened 2021-05-25
[CBRD-23874] The function IFNULL returns incorrect data type at the query with numeric host-variables

beyondykk9:
http://jira.cubrid.org/browse/CBRD-23874

The following query expects 12.34568 but query returns 12.00000. The function NVL also returns 12.00000 as like IFNULL.
```

CREATE TABLE dec_tbl (a decimal(10,5));
PREPARE stmt from 'INSERT INTO dec_tbl VALUES (IFNULL(?,0))';
EXECUTE stmt USING 12.3456789;

SELECT a FROM dec_tbl;
```
Expected Result
```
 a
======================
  12.34568 
```
Actual Result
```
  a
======================
  12.00000 
```
At the semantic checking, the type cast from PT_TYPE_MAYBE makes the precision and scale as their default values. So, the precision and scale is set 15 and 0 respectively.
I modified the **pt_is_compatible_without_cast** not to return 'false' at the above case.

### DIFF
--- a/src/parser/semantic_check.c
+++ b/src/parser/semantic_check.c
@@ -11467,7 +11467,7 @@ pt_assignment_compatible (PARSER_CONTEXT * parser, PT_NODE * lhs, PT_NODE * rhs)
 			      // TODO: It should be considered for parameterized types, refer to PT_IS_PARAMETERIZED_TYPE()
 			      if (lhs->type_enum == PT_TYPE_NUMERIC && lhs->data_type != NULL)
 				{
-				  tp_domain_resolve (lhs_dbtype, NULL, sci.prec, sci.scale, NULL, 0);
+				  d = tp_domain_resolve (lhs_dbtype, NULL, sci.prec, sci.scale, NULL, 0);
 				}
 			      else
 				{
--- a/src/parser/semantic_check.c
+++ b/src/parser/semantic_check.c
@@ -11424,6 +11424,7 @@ pt_assignment_compatible (PARSER_CONTEXT * parser, PT_NODE * lhs, PT_NODE * rhs)
 		  if (rhs->node_type != PT_HOST_VAR)
 		    {
 		      d = tp_domain_resolve_default (lhs_dbtype);
+		      // TODO: It should be considered for parameterized types, refer to PT_IS_PARAMETERIZED_TYPE()
 		      if (lhs->type_enum == PT_TYPE_NUMERIC && lhs->data_type != NULL)
 			{
 			  d = tp_domain_copy (d, false);
@@ -11464,6 +11465,7 @@ pt_assignment_compatible (PARSER_CONTEXT * parser, PT_NODE * lhs, PT_NODE * rhs)
 			  else
 			    {
 			      d = tp_domain_resolve_default (lhs_dbtype);
+			      // TODO: It should be considered for parameterized types, refer to PT_IS_PARAMETERIZED_TYPE()
 			      if (lhs->type_enum == PT_TYPE_NUMERIC && lhs->data_type != NULL)
 				{
 				  d = tp_domain_copy (d, false);
--- a/src/parser/semantic_check.c
+++ b/src/parser/semantic_check.c
@@ -11423,14 +11423,14 @@ pt_assignment_compatible (PARSER_CONTEXT * parser, PT_NODE * lhs, PT_NODE * rhs)
 
 		  if (rhs->node_type != PT_HOST_VAR)
 		    {
-		      d = tp_domain_resolve_default (lhs_dbtype);
 		      // TODO: It should be considered for parameterized types, refer to PT_IS_PARAMETERIZED_TYPE()
 		      if (lhs->type_enum == PT_TYPE_NUMERIC && lhs->data_type != NULL)
 			{
-			  d = tp_domain_copy (d, false);
-			  d->precision = sci.prec;
-			  d->scale = sci.scale;
-			  d = tp_domain_cache (d);
+			  d = tp_domain_resolve (lhs_dbtype, NULL, sci.prec, sci.scale, NULL, 0);
+			}
+		      else
+			{
+			  d = tp_domain_resolve_default (lhs_dbtype);
 			}
 		    }
 		  else
@@ -11464,14 +11464,14 @@ pt_assignment_compatible (PARSER_CONTEXT * parser, PT_NODE * lhs, PT_NODE * rhs)
 			    }
 			  else
 			    {
-			      d = tp_domain_resolve_default (lhs_dbtype);
 			      // TODO: It should be considered for parameterized types, refer to PT_IS_PARAMETERIZED_TYPE()
 			      if (lhs->type_enum == PT_TYPE_NUMERIC && lhs->data_type != NULL)
 				{
-				  d = tp_domain_copy (d, false);
-				  d->precision = sci.prec;
-				  d->scale = sci.scale;
-				  d = tp_domain_cache (d);
+				  tp_domain_resolve (lhs_dbtype, NULL, sci.prec, sci.scale, NULL, 0);
+				}
+			      else
+				{
+				  d = tp_domain_resolve_default (lhs_dbtype);
 				}
 			    }
 
--- a/src/parser/semantic_check.c
+++ b/src/parser/semantic_check.c
@@ -11423,14 +11423,12 @@ pt_assignment_compatible (PARSER_CONTEXT * parser, PT_NODE * lhs, PT_NODE * rhs)
 
 		  if (rhs->node_type != PT_HOST_VAR)
 		    {
-		      if (lhs->data_type != NULL)
+		      d = tp_domain_resolve_default (lhs_dbtype);
+		      if (lhs->type_enum == PT_TYPE_NUMERIC && lhs->data_type != NULL)
 			{
-			  d = pt_data_type_to_db_domain (parser, lhs->data_type, NULL);
-			  d = tp_domain_cache (d);
-			}
-		      else
-			{
-			  d = tp_domain_resolve_default (lhs_dbtype);
+			  d = tp_domain_copy (d, false);
+			  d->precision = lhs->data_type->info.data_type.precision;
+			  d->scale = lhs->data_type->info.data_type.dec_precision;
 			}
 		    }
 		  else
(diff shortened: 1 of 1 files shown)


src/parser/semantic_check.c
```diff
@@ -2298,6 +2298,10 @@ pt_is_compatible_without_cast (PARSER_CONTEXT * parser, SEMAN_COMPATIBLE_INFO *
 
   if (dest_sci->type_enum != src->type_enum)
     {
+      if (src->type_enum == PT_TYPE_MAYBE && dest_sci->type_enum == PT_TYPE_NUMERIC)
```
ctshim: If it is called through pt_values_query_to_compatible_cast() or pt_to_compatible_cast(), can it never happen if the following conditions are true?
(src->type_enum == PT_TYPE_MAYBE && dest_sci->type_enum == PT_TYPE_NUMERIC)

beyondykk9: I think, in general, PT_TYPE_MAYBE to PT_TYPE_NUMERIC is compatible without cast. In any similar case of this, the cast is not needed at compile time. It will be automatically done by the coercing instead of casting at execution time.

hgryoo:
I'm not sure the context that MAYBE and NUMERIC are compatible without casting.
Could you explain it in detail?

In pt_assignment_compatbile() function, I found the following after calling pt_is_compatible_without_cast().
IMO, pt_assignment_compatbile() seems the proper place to fix assigning the data type (domain) right.
```
pt_assignment_compatbile (...)
{
...
      if (pt_is_compatible_without_cast (parser, &sci, rhs, &is_cast_allowed))
	{
	  return rhs;
	}
...

      if (rhs->type_enum != PT_TYPE_NULL)
	{
	  if (rhs->type_enum == PT_TYPE_MAYBE)
          {
            ...
          }
          else
         {
            if (rhs->node_type != PT_HOST_VAR)
            {
               d = tp_domain_resolve_default (lhs_dbtype); // <-------- here to fix? 
            }
         }
         // tp_domain_resolve_default () appears below too,
```

beyondykk9:
> I'm not sure the context that MAYBE and NUMERIC are compatible without casting.
> Could you explain it in detail?
> 
> In pt_assignment_compatbile() function, I found the following after calling pt_is_compatible_without_cast().
> IMO, pt_assignment_compatbile() seems the proper place to fix assigning the data type (domain) right.
> 
> ```
> pt_assignment_compatbile (...)
> {
> ...
>       if (pt_is_compatible_without_cast (parser, &sci, rhs, &is_cast_allowed))
> 	{
> 	  return rhs;
> 	}
> ...
> 
>       if (rhs->type_enum != PT_TYPE_NULL)
> 	{
> 	  if (rhs->type_enum == PT_TYPE_MAYBE)
>           {
>             ...
>           }
>           else
>          {
>             if (rhs->node_type != PT_HOST_VAR)
>             {
>                d = tp_domain_resolve_default (lhs_dbtype); // <-------- here to fix? 
>             }
>          }
>          // tp_domain_resolve_default () appears below too,
> ```

In execution, by **tp_value_coerce** the value is cast automatically. The tp_value_coerce returns an error for invalid cast, for example, "String to Numeric".
But I will try to modify code as you suggested.

src/parser/semantic_check.c
beyondykk9: The code is reverted and I modified the function pt_assignment_compatible as @hgryoo suggested.

ctshim (approved):

hgryoo:
How about adding the following code from line 11474 to cover types that have precision and scale?
```
if (PT_IS_PARAMETERIZED_TYPE (lhs->type_enum))
{
  d = tp_domain_copy (d, false);
  d->precision= sci.prec;
  d->scale = sci.scale;
  d = tp_domain_cache (d);
}
```

At line 11474, Assigning collation from SEMAN_COMPATIBLE_INFO is implemented already. 
It seems the same pattern.
```
if (PT_HAS_COLLATION (lhs->type_enum))
{
  d = tp_domain_copy (d, false);
  d->codeset = sci.coll_infer.codeset;
  d->collation_id = sci.coll_infer.coll_id;
  d = tp_domain_cache (d);
}
```

src/parser/semantic_check.c
```diff
@@ -11424,6 +11424,12 @@ pt_assignment_compatible (PARSER_CONTEXT * parser, PT_NODE * lhs, PT_NODE * rhs)
 		  if (rhs->node_type != PT_HOST_VAR)
 		    {
 		      d = tp_domain_resolve_default (lhs_dbtype);
+		      if (lhs->type_enum == PT_TYPE_NUMERIC && lhs->data_type != NULL)
+			{
+			  d = tp_domain_copy (d, false);
+			  d->precision = lhs->data_type->info.data_type.precision;
+			  d->scale = lhs->data_type->info.data_type.dec_precision;
```
shparkcubrid: I recommend adding a tp_domain_cache ().
```
d = tp_domain_cache (d);
```

src/parser/semantic_check.c
```diff
@@ -11424,7 +11424,7 @@ pt_assignment_compatible (PARSER_CONTEXT * parser, PT_NODE * lhs, PT_NODE * rhs)
 		  if (rhs->node_type != PT_HOST_VAR)
 		    {
 		      d = tp_domain_resolve_default (lhs_dbtype);
-		      if (PT_IS_PARAMETERIZED_TYPE (lhs->type_enum))
```
hgryoo: checking `lhs->data_type != NULL` is not needed. `lhs->data_type` is not accessed in the if-block.

shparkcubrid (approved):

src/parser/semantic_check.c
```diff
@@ -11424,6 +11424,13 @@ pt_assignment_compatible (PARSER_CONTEXT * parser, PT_NODE * lhs, PT_NODE * rhs)
 		  if (rhs->node_type != PT_HOST_VAR)
 		    {
 		      d = tp_domain_resolve_default (lhs_dbtype);
+		      if (lhs->type_enum == PT_TYPE_NUMERIC && lhs->data_type != NULL)
```
hgryoo, line 11427: checking lhs->data_type != NULL is not needed. lhs->data_type is not accessed in the if-block.

src/parser/semantic_check.c
```diff
@@ -11457,6 +11464,13 @@ pt_assignment_compatible (PARSER_CONTEXT * parser, PT_NODE * lhs, PT_NODE * rhs)
 			  else
 			    {
 			      d = tp_domain_resolve_default (lhs_dbtype);
+			      if (lhs->type_enum == PT_TYPE_NUMERIC && lhs->data_type != NULL)
```
hgryoo, line 11468: checking lhs->data_type != NULL is not needed. lhs->data_type is not accessed in the if-block.

hgryoo (commented):
Could you add TODO comment?
// TODO: It should be considered for parameterized types , refer to PT_IS_PARAMETERIZED_TYPE()

src/parser/semantic_check.c
beyondykk9: The following code snippet is from the earlier part of this routine, the **lhs->data_type** is checked if it's null. I think the checking is needed.
```
if (lhs->data_type)
	{
	  sci.prec = lhs->data_type->info.data_type.precision;
	  sci.scale = lhs->data_type->info.data_type.dec_precision;
	  if (lhs->type_enum == PT_TYPE_JSON && lhs->data_type->info.data_type.json_schema != NULL)
	    {
	      sci.force_cast = true;
	    }
	}
```

src/parser/semantic_check.c
beyondykk9: Ok, I will comment it as you suggested.

hgryoo (approved):

src/parser/semantic_check.c
hgryoo: Ok, I agree

src/parser/semantic_check.c
```diff
@@ -11424,6 +11424,14 @@ pt_assignment_compatible (PARSER_CONTEXT * parser, PT_NODE * lhs, PT_NODE * rhs)
 		  if (rhs->node_type != PT_HOST_VAR)
 		    {
 		      d = tp_domain_resolve_default (lhs_dbtype);
```
youngjinj: If resolving a domain by default is a problem, how about using a function that doesn't resolve by default?

```suggestion
		      d = tp_domain_resolve(lhs_dbtype, NULL, sci.prec, sci.scale, NULL, sci.coll_infer.coll_id);
```

src/parser/semantic_check.c
```diff
@@ -11457,6 +11465,14 @@ pt_assignment_compatible (PARSER_CONTEXT * parser, PT_NODE * lhs, PT_NODE * rhs)
 			  else
 			    {
 			      d = tp_domain_resolve_default (lhs_dbtype);
```
youngjinj: If resolving a domain by default is a problem, how about using a function that doesn't resolve by default?

```suggestion
		      d = tp_domain_resolve(lhs_dbtype, NULL, sci.prec, sci.scale, NULL, sci.coll_infer.coll_id);
```

src/parser/semantic_check.c
beyondykk9: Good point, thanks.
`d = tp_domain_resolve (lhs_dbtype, NULL, sci.prec, sci.scale, NULL, 0);`
the collation looks not needed for numeric type.

youngjinj (approved):

src/parser/semantic_check.c
```diff
@@ -11456,7 +11464,15 @@ pt_assignment_compatible (PARSER_CONTEXT * parser, PT_NODE * lhs, PT_NODE * rhs)
 			    }
 			  else
 			    {
-			      d = tp_domain_resolve_default (lhs_dbtype);
+			      // TODO: It should be considered for parameterized types, refer to PT_IS_PARAMETERIZED_TYPE()
+			      if (lhs->type_enum == PT_TYPE_NUMERIC && lhs->data_type != NULL)
+				{
+				  tp_domain_resolve (lhs_dbtype, NULL, sci.prec, sci.scale, NULL, 0);
```
youngjinj: 'd = ' is missing

beyondykk9: Ok, thanks.

PR #2819, opened 2021-05-26
[CBRD-23957] Implement packing/unpacking on Java SP Server compatible with CUBRID packer

hgryoo:
http://jira.cubrid.org/browse/CBRD-23957

**Purpose**
New packing and unpacking helper classes are introduced. data send to/receive from Java SP Server at CAS is compatible with cubrid packing/unpacking behavior.

**Implementations**

Overview of how CAS and ExecuteThread in Java SP Server communicate each other for a single stored procedure call
```
csql> select testplussp (1, 2);
====
3
```
```
[CAS/CSQL]                                                       [ExecuteThread in Java SP Server]

// assumes connection is already done, argument received from DB Server

<jsp_cl.c>                                                        <ExecuteThread.java>
jsp_send_call_request (SP_ARGS) ---------------------------------> processStoredProcedure ()
   - SP_ARGS is packed to cubmem::extensible_block                   | buffer.put (input.readFully ())
                                                                     | procedure = makeStoredProcedure ()
                                                                     | Object result = procedure.getMethod().invoke()
                                                                     | Value resolvedResult = procedure.makeReturnValue(result);
jsp_receive_response ()     <-------------------------------------   | sendResult(resolvedResult, procedure);
  | jsp_alloc_response ()
  | jsp_receive_result () or jsp_receive_error ()
    - buffer is unpacked to DB_VALUE

// the result DB_VALUE is going to send to DB Server
```

C/C++: A client to call stored procedure from Java SP Server. This routine runs on CAS
- **jsp_packer.[cpp|hpp]**: defines classes inherits from cubpacking::packable_object implements packing/unpacking through `cubpacking::packer/unpacker` in src/base/packer.hpp
  - **sp_args class**
    - It was SP_ARG struct at the previous version. It includes method signature string, argument values, argument meta infos and return value.
    - packing/unpacking for these member variables are implemented
  - **sp_header class**: it is just helper class to send command and size of buffer
- **jsp_cl.c**: There are functions to send arguments and receive result value to/from Java SP Server. It refactored to use new packing/unpacking classes
  - `db_arg_list` and `SP_ARGS` are moved into **jsp_struct.hpp**
  - `SP_ARGS` inherits `cubpacking::packable_object`
  - legacy packing/unpacking routines are removed
  - `jsp_pack_*_argument ()`, `jsp_unpack_*_value()`
  - `cubmem::extensible_block` is used as buffer to store packed argument bytes or received value buffer.

Java: A server to receive command to call Java method from CAS. Each call perform in each thread
- **DataUtilites.java**: a Utility class implements static methods which aligns memory, checks range to put and read data for buffer and defines constants related to data format.
- **CUBRIDPacker.java**, **CUBRIDUnpacker.java**
  - They perform packing/unpacking based on `ByteBuffer` class.
- **ExecuteThread.java**: dispatches command from CAS and Java SP Utility. processing stored procedure call is only refactored. The others are kept it is.
  - the legacy packing methods (`getLengthtoSend()`, `packAndSendRawString()`...) are moved to `DataUtilities` class
  - the packing/unpacking routines by `DataInputStream` and `DataOutputStream` are removed
  - reads buffer fully for the received buffer size at `processStoredProcedure()`
  - interprets stored procedure calling signatures and arguments through `CUBRIDUnpacker`
  - a result value to send is packed through `CUBRIDPacker`

**Remarks**
The Java SP demo examples are added too for easy code review. please refer to the `demo/*`
`select testplussp (1, 2);` can run without registering procedure if you test with **demodb**


src/base/packer.cpp
```diff
@@ -922,8 +922,8 @@ namespace cubpacking
   packer::delegate_to_or_buf (const size_t size, or_buf &buf)
   {
     check_range (m_ptr, m_end_ptr, size);
-    m_ptr += size;
     OR_BUF_INIT (buf, m_ptr, size);
+    m_ptr += size;
```
ctshim, line 926: The position to move m_ptr has been moved after OR_BUF_INIT() is specified. 
Does that mean the code before the change was a bug?

src/jsp/com/cubrid/jsp/data/CUBRIDPacker.java
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    public void packString(String value, String charset) throws UnsupportedEncodingException {
+        packCString(value.getBytes(charset));
+    }
+
+    public void packCString(byte[] value) {
+        int len = value.length;
+        if (len < DataUtilities.MAX_SMALL_STRING_SIZE) {
+            ensureSpace(value.length + 1); // str + len
```
ctshim, line 103: Is the len information resolved to 1 byte? Maybe 4 bytes are needed?
ensureSpace(value.length + 4);

src/jsp/com/cubrid/jsp/data/DBType.java
```diff
@@ -0,0 +1,54 @@
...
+    public static final int DB_TIMESTAMP = 11;
+    public static final int DB_DATE = 12;
+    public static final int DB_MONETARY = 13;
+    public static final int DB_SHORT = 18;
+    public static final int DB_NUMERIC = 22;
+    public static final int DB_CHAR = 25;
+    public static final int DB_RESULTSET = 28;
+    public static final int DB_BIGINT = 31;
+    public static final int DB_DATETIME = 32;
+}
```
ctshim, line 54: Is  JSON type not supported?

hgryoo: As far as I know, JSON is considered as String type.

src/base/packer.cpp
hgryoo: Yes, this function (`packer::delegate_to_or_buf`) is defined to migrate with legacy OR code. But the function was not used anywhere. After testing it for un-implemented float and double types, I found a bug. The ptr should be move after OR_BUF is initialized.

src/jsp/jsp_cl.c
```diff
@@ -2872,7 +1699,7 @@ jsp_do_call_stored_procedure (DB_VALUE * returnval, DB_ARG_LIST * args, const ch
 
   db_make_null (&method);
   db_make_null (&param);
-  memset (&sp_args, 0, sizeof (SP_ARGS));
+  //memset (&sp_args, 0, sizeof (SP_ARGS));
```
beyondykk9: why **memset** not needed ?

hgryoo: SP_ARGS is now a C++ struct, and member variables are initialized in the constructor. Also, This line occurs SEGFAULT.

src/jsp/com/cubrid/jsp/value/ValueUtilities.java
```diff
@@ -0,0 +1,94 @@
...
+                resolvedResult = value.toObjectArray();
+                break;
+            case DBType.DB_TIME:
+                resolvedResult = value.toTime();
+                break;
+            case DBType.DB_DATE:
+                resolvedResult = value.toDate();
+                break;
+            case DBType.DB_TIMESTAMP:
+                resolvedResult = value.toTimestamp();
```
beyondykk9, line 72: I think DATETIMETZ and DATETIMELTZ should be considered as JDBC client driver.
Consider also TIMESTAMPTZ, TIMESTAMPLTZ.

hgryoo: Since this PR is just refactoring change, the types you mentioned are not considered before and I didn't cover them. I'm going to test for the unsupported types on another issue.

src/jsp/jsp_struct_value.cpp
```diff
@@ -0,0 +1,718 @@
...
+      case DB_TYPE_DATE:
+      case DB_TYPE_TIME:
+	size += serializator.get_packed_int_size (size);
+	size += serializator.get_packed_int_size (size);
+	size += serializator.get_packed_int_size (size);
+	size += serializator.get_packed_int_size (size);
+	break;
+
+      case DB_TYPE_TIMESTAMP:
+	size += serializator.get_packed_int_size (size);
```
ctshim: If they are arranged in the same type, direct calculation is possible as shown below.
size += (OR_INT_SIZE * 6);
If the same type, not variable length, is used repeatedly, only the first of them is moved by align, and after that it seems to just fit.

beyondykk9: add comments for readability
ex) size += serializator.get_packed_int_size (size); // year

hgryoo: @ctshim 
I'd like to keep this pattern as it is. because of the following reasons
1) consistency how to use this serializer
2) To not care about how the aligning is done
3) There is no particular impact on performance.

@byungwook-kim 
Good point. I'm going to add comments

ctshim: OK, Thank you for your kindness.

ctshim (approved):

beyondykk9 (approved):

shparkcubrid (approved):

youngjinj (approved):

src/jsp/jsp_struct.cpp
```diff
@@ -0,0 +1,134 @@
...
+namespace cubprocedure
+{
+
+//////////////////////////////////////////////////////////////////////////
+// header
+//////////////////////////////////////////////////////////////////////////
+
+  sp_header::sp_header ()
+    : command ()
+    , size ()
```
youngjinj: It would be better to explicitly put 0.

hgryoo: Thank you. I'm gonna fix it

PR #2821, opened 2021-05-27
[LETS-34] Wrap recovery information into log_recovery_context class.

ribram:

This a refactoring task that groups all variables that are passed around different recovery phases and that have an impact the behavior of recovery.

The grouped information is:

  - Restore related: if recovery code is executed for restore rather than recovering from a crash, there are some changes in analysis and recovery redo parts. Restore may have a time_t recovery and may execute recovery incompletely (stops before end of log).
  - Recovery LSA's: checkpoint and start/end redo LSA's. Checkpoint LSA is the starting point of recovery analysis. Start/end redo LSA's are limits for recovery redo. Recovery analysis set these values.
  - m_is_page_server: for now the flag is initialized, but not used. It was added as an indicator of how the page server will be handled. With the help of this recovery context flag, parts of recovery will be disabled on the page server.

There are no behavior changes in this patch. Several variables have been replaced with a context instance and as consequence several function signatures have been updated. Small parts of code have been moved from analysis functions to log_recovery_context member functions (init_for_restore, force_stop_restore_at).

log_rv_find_checkpoint() function was rewritten as get_disk_checkpoint_min_lsa().


src/transaction/log_recovery.c
```diff
@@ -2222,18 +2159,14 @@ log_recovery_analysis (THREAD_ENTRY * thread_p, LOG_LSA * start_lsa, LOG_LSA * s
...
-  LSA_COPY (&lsa, start_lsa);
+  lsa = context.get_checkpoint_lsa ();
 
   // *INDENT-OFF*
   // If the recovery start matches a checkpoint, use the checkpoint information.
-  const cublog::checkpoint_info *chkpt_infop = log_Gl.m_metainfo.get_checkpoint_info (*start_lsa);
+  const cublog::checkpoint_info *chkpt_infop = log_Gl.m_metainfo.get_checkpoint_info (lsa);
   // *INDENT-ON*
-
-  LSA_COPY (start_redo_lsa, &lsa);
```
cristiarg, line 2232: Shouldn't `set_start_redo_lsa` be called here?

src/transaction/log_recovery_context.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+    log_recovery_context (const log_recovery_context &) = delete;
+    log_recovery_context (log_recovery_context &&) = delete;
+
+    // Accessors
+    const log_lsa &get_checkpoint_lsa () const;
+    const log_lsa &get_start_redo_lsa () const;
+    const log_lsa &get_end_redo_lsa () const;
+    void set_end_redo_lsa (const log_lsa &end_redo_lsa);
+    void set_start_redo_lsa (const log_lsa &start_redo_lsa);
```
cristiarg: Pedantic, order:
```suggestion
    const log_lsa &get_start_redo_lsa () const;
    const log_lsa &get_end_redo_lsa () const;
    void set_start_redo_lsa (const log_lsa &start_redo_lsa);
    void set_end_redo_lsa (const log_lsa &end_redo_lsa);
```

src/transaction/log_recovery.c
```diff
@@ -725,14 +694,12 @@ log_recovery (THREAD_ENTRY * thread_p, int ismedia_crash, time_t * stopat)
...
   LOG_RECORD_HEADER *eof;	/* End of the log record */
-  LOG_LSA rcv_lsa;		/* Where to start the recovery */
-  LOG_LSA start_redolsa;	/* Where to start redo phase */
-  LOG_LSA end_redo_lsa;		/* Where to stop the redo phase */
-  bool did_incom_recovery;
   int tran_index;
   INT64 num_redo_log_records;
   int error_code = NO_ERROR;
 
+  log_recovery_context context;
```
cristiarg, line 729: Maybe, make the context ctor empty/default and, to initialize, add an explicit function that copies from `log_Gl.hdr.chkpt_lsa` because I think it makes the code more explicit.
There is already an `init_for_restore`.
Or at least make the ctor take `log_Gl.hdr.chkpt_lsa` as explicit param even if global.

src/transaction/log_recovery.c
```diff
@@ -2958,13 +2885,9 @@ log_recovery_redo (THREAD_ENTRY * thread_p, const LOG_LSA * start_redolsa, const
...
                 // *INDENT-OFF*
-                log_rv_redo_record_sync_or_dispatch_async<LOG_REC_MVCC_UNDOREDO> (thread_p, log_pgptr_reader,
-                                                                                  log_rec_mvcc_undoredo,
-                                                                                  rcv_lsa, end_redo_lsa,
-                                                                                  log_rtype, *undo_unzip_ptr,
-                                                                                  *redo_unzip_ptr,
-                                                                                  parallel_recovery_redo,
-                                                                                  force_each_log_page_fetch);
+                log_rv_redo_record_sync_or_dispatch_async<LOG_REC_MVCC_UNDOREDO>
+		  (thread_p, log_pgptr_reader, log_rec_mvcc_undoredo, rcv_lsa, &context.get_end_redo_lsa (), log_rtype,
```
cristiarg, line 2926: These indentations went from one extreme to the other. :)

src/transaction/log_recovery_context.cpp
```diff
@@ -0,0 +1,134 @@
...
+log_recovery_context::init_for_restore (const time_t *stopat_p)
+{
+  m_is_restore_from_backup = true;
+  if (stopat_p)
+    {
+      m_restore_stop_point = *stopat_p;
+    }
+
+  /* we may have to start from an older checkpoint... */
+  (void) fileio_map_mounted (&cubthread::get_entry (), get_disk_checkpoint_min_lsa, &m_checkpoint_lsa);
```
cristiarg: I find it a bit odd to bury this aspect here from its prominent place in `log_recovery`. This class is supposed to provide some bookkeeping container for recovery and add as little business logic as possible.

src/transaction/log_recovery_context.hpp
ribram: Done

src/transaction/log_recovery_context.cpp
ribram: You make a valid point :). The logic of getting min LSA from disk was moved back to log_recovery().

src/transaction/log_recovery.c
ribram: It is already set to the same LSA by the init functions.

src/transaction/log_recovery.c
ribram: Good point, init_for_recovery() was added.

src/transaction/log_recovery.c
ribram: No midway I could find.

cristiarg (approved):

ribram:
Made a fix for the next bug in recovery redo.

- Restore hits a page that cannot be fetched from disk.
- Incomplete stop is forced and the time upto restore is set to -1 (actually, it is set to the value of `last_at_time` in log_recovery_analysis that is always -1; I missed this fact).
- During recovery redo, the commit times are compared to the time of restore. log_recovery_context::does_restore_stop_before_time could have skipped if the restore time was zero, but its value was -1. The very first commit was already "sooner" that restore time and recovery stopped almost immediately. Almost no changes were applied and hell broke loose.

The case was corrected and the redo is applied up to the unfetchable page.

- Since last_at_time always remains -1, it is not useful. It is removed completely.
- force_stop_restore_at looses the time argument and is renamed set_forced_restore_stop()
- m_restore_stop_point is set to 0
- log_recovery_context::does_restore_stop_before_time also skips the check if m_restore_stop_point is negative (unnecessary, but better).

ribram:
During the investigations of the bug above, another bug was discovered: prev_prev_lsa value was always the same as prev_lsa value, but it not supposed to.

Additional changes were applied:

 - Correctly set prev_prev_lsa to prev_lsa before updating prev_lsa.
 - Moved "is_log_page_broken" handling to a function, to make things a bit more clear. When log_is_page_of_record_broken() returns true, the new function is called instead of breaking the nested loop and relying on that "is_log_page_broken" block in the outer loop.

cristiarg (approved):

PR #2822, opened 2021-05-27
[LETS-102] Use log_lsa packing function for checkpoint_information

adrianbrab:
http://jira.cubrid.org/browse/LETS-102

Changed the pack/unpack functions in the checkpoint_information class to use the newly added log_lsa serialization functions instead fo the local ones. 

Also added a get_pack_size function for log_lsa.

### DIFF
--- a/src/transaction/log_lsa_utils.cpp
+++ b/src/transaction/log_lsa_utils.cpp
@@ -40,8 +40,8 @@ namespace cublog::lsa_utils
   }
 
   std::size_t
-  get_packed_size (cubpacking::packer &serializator, std::size_t start_offset, std::size_t size)
+  get_packed_size (cubpacking::packer &serializator, std::size_t start_offset)
   {
-    return serializator.get_packed_bigint_size (start_offset + size);
+    return serializator.get_packed_bigint_size (start_offset);
   }
 }
--- a/src/transaction/log_lsa_utils.cpp
+++ b/src/transaction/log_lsa_utils.cpp
@@ -39,7 +39,7 @@ namespace cublog::lsa_utils
     lsa = big_int;
   }
 
-  std::size_t
+  size_t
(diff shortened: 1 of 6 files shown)


src/transaction/log_lsa_utils.cpp
```diff
@@ -38,4 +38,10 @@ namespace cublog::lsa_utils
     deserializer.unpack_bigint (big_int);
     lsa = big_int;
   }
+
+  std::size_t
+  get_packed_size (cubpacking::packer &serializator, std::size_t start_offset, std::size_t size)
+  {
+    return serializator.get_packed_bigint_size (start_offset + size);
+  }
```
ribram: The signature for get_packed_size() functions is serializator & start_offset. The start_offset is the offset to the packed object. Remove the size argument. Caller has to do the sum between another starting point and the offset of the packed log_lsa.
```suggestion
  std::size_t
  get_packed_size (cubpacking::packer &serializator, std::size_t start_offset)
  {
    return serializator.get_packed_bigint_size (start_offset);
  }
```

unit_tests/log/test_main_chkpt_info.cpp
```diff
@@ -960,6 +961,26 @@ mvcc_snapshot::mvcc_snapshot ()
 {
 }
 
+void
+lsa_utils::pack (const log_lsa &lsa, cubpacking::packer &serializer)
```
ribram: Nitpicking: an alternative is to include log_lsa_utils.cpp in the project and remove duplicated functions.

adrianbrab: That might be better, it would have saved me some time debugging. I will change it

ribram (approved):

danilailie (approved):

PR #2823, opened 2021-05-27
[LETS-99] Transaction server received data pages from page server

danilailie:
http://jira.cubrid.org/browse/LETS-99

* Implemented template class for page_broker.
* Updated unit test for (log)page_broker.


src/transaction/log_page_broker.tpp
```diff
@@ -0,0 +1,86 @@
+/*
```
cristiarg: I am not sure whether is ok to introduce a new extension.
Indenting tools?
CMake? (even if it is only, technically, an include file)
One way to do it is to just make the extension `.cpp.hpp` to still mark that it is smth different than a header and still allow indenting scripts to do their stuff.

src/transaction/log_page_broker.tpp
```diff
@@ -0,0 +1,86 @@
...
+
+  return result;
+}
+
+template <typename PageT>
+void
+page_broker<PageT>::set_page (typename map_type<PageT>::key id, typename map_type<PageT>::value &&page)
+{
+  {
+    std::unique_lock<std::mutex> lock (m_pages_mutex);
```
cristiarg: `lock_guard` iso `unique_lock`

src/transaction/log_page_broker.hpp
```diff
@@ -24,37 +24,69 @@
...
+    typename map_type<PageT>::value wait_for_page (typename map_type<PageT>::key id);
+    void set_page (typename map_type<PageT>::key id, typename map_type<PageT>::value &&page);
+
+  private:
+    std::mutex m_pages_mutex;
+
+    std::condition_variable m_pages_cv;
+
+    std::map<typename map_type<PageT>::key, int> m_requested_page_id_count;
+    std::map<typename map_type<PageT>::key, typename map_type<PageT>::value> m_received_pages;
```
cristiarg: awkward to have these two parallel maps;
why not make it a single map whose value contains all that is needed;
suppose some new bookkeeping is needed: either add a third map, or refactor then

src/server/active_tran_server.cpp
```diff
@@ -329,7 +341,7 @@ active_tran_server::receive_log_page (cubpacking::unpacker &upk)
   if (error_code == NO_ERROR)
     {
       auto shared_log_page = std::make_shared<log_page_owner> (message.c_str () + sizeof (error_code));
-      m_log_page_broker->set_page (std::move (shared_log_page));
+      m_log_page_broker->set_page (shared_log_page->get_id (), std::move (shared_log_page));
 
       if (prm_get_bool_value (PRM_ID_ER_LOG_READ_LOG_PAGE))
 	{
```
cristiarg, line 352: after `shared_log_page` is `std::moved`'d, it cannot be used at ~349
.. unless `std::move` actually does not do its job

src/transaction/log_page_broker.hpp
```diff
@@ -24,37 +24,69 @@
...
+  ADDED_ENTRY,
+  EXISTING_ENTRY
+};
+
+template <typename PageT>
+class page_broker
+{
+  public:
+    page_broker () = default;
+    ~page_broker () = default;
```
cristiarg: explicitly delete copy/move ctors/operators if not needed

cristiarg: dtor should actually assert that all requests have been fulfilled; unless so by design

src/transaction/log_page_broker.hpp
```diff
@@ -24,37 +24,69 @@
...
+
+    entry_state register_entry (typename map_type<PageT>::key id);
+    size_t get_requests_count ();
+    size_t get_pages_count ();
```
cristiarg:
```suggestion
    size_t get_requests_count () const;
    size_t get_pages_count () const;
```
Later edit: I see that there's a need to lock internally. In such situations it's better to keep the API of the class consistent and easy to use (aka: `const` correct) and use `mutable` member mutexes internally.

src/transaction/log_page_broker.tpp
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+  return result;
+}
+
+template <typename PageT>
+size_t
+page_broker<PageT>::get_requests_count ()
+{
+  std::unique_lock<std::mutex> lock (m_pages_mutex);
```
cristiarg:
```suggestion
  std::lock_guard<std::mutex> lock (m_pages_mutex);
```

src/transaction/log_page_broker.tpp
```diff
@@ -0,0 +1,86 @@
...
+  return m_requested_page_id_count.size ();
+}
+
+template <typename PageT>
+std::size_t
+page_broker<PageT>::get_pages_count ()
+{
+  std::unique_lock<std::mutex> lock (m_pages_mutex);
```
cristiarg: `lock_guard`

src/transaction/log_page_broker.tpp
```diff
@@ -0,0 +1,86 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+
+template <typename PageT>
+entry_state
+page_broker<PageT>::register_entry (typename map_type<PageT>::key id)
+{
+  std::unique_lock<std::mutex> lock (m_pages_mutex);
```
cristiarg: `lock_guard`

src/transaction/log_page_broker.tpp
```diff
@@ -0,0 +1,86 @@
...
+}
+
+template <typename PageT>
+typename map_type<PageT>::value
+page_broker<PageT>::wait_for_page (typename map_type<PageT>::key id)
+{
+  std::unique_lock<std::mutex> lock (m_pages_mutex);
+  m_pages_cv.wait (lock, [this, id]
+  {
+    return m_received_pages.find (id) != m_received_pages.end ();
```
cristiarg:
```suggestion
    return m_received_pages.find (id) != m_received_pages.cend ();
```

unit_tests/log/test_main_chkpt_info.cpp
```diff
@@ -720,7 +720,7 @@ logtb_rv_find_allocate_tran_index (THREAD_ENTRY *thread_p, TRANID trid, const LO
       if (log_Gl.trantable.all_tdes[i]->trid != NULL_TRANID && log_Gl.trantable.all_tdes[i]->trid == trid)
 	{
 	  std::map<TRANID, log_tdes *>::const_iterator inside_tdes = tran_map.find (trid);
-	  if (inside_tdes != tran_map.end())
+	  if (inside_tdes != tran_map.end ())
```
cristiarg, line 723:
```suggestion
	  if (inside_tdes != tran_map.cend ())
```

src/transaction/log_page_broker.tpp
```diff
@@ -0,0 +1,86 @@
...
+ */
+
+template <typename PageT>
+entry_state
+page_broker<PageT>::register_entry (typename map_type<PageT>::key id)
+{
+  std::unique_lock<std::mutex> lock (m_pages_mutex);
+
+  entry_state result = EXISTING_ENTRY;
+  if (m_requested_page_id_count.find (id) != m_requested_page_id_count.end ())
```
cristiarg: If using `find`, keep the result non-const interator and use it instead of calling `operator []` will do another, gratuitous, search again.

src/transaction/log_page_buffer.c
```diff
@@ -1944,7 +1944,7 @@ logpb_copy_page (THREAD_ENTRY * thread_p, LOG_PAGEID pageid, LOG_CS_ACCESS_MODE
   if (get_server_type () == SERVER_TYPE_TRANSACTION && ats_Gl.is_page_server_connected ())
     {
       // wait for answer.
-      auto log_page = ats_Gl.get_log_page_broker ().wait_for_page (pageid);
+      auto log_page = ats_Gl.get_page_broker<log_page_type> ().wait_for_page (pageid);
 
       // Sould be the same.
```
cristiarg: Obvious comment is obvious. :)

src/server/active_tran_server.hpp
```diff
@@ -51,10 +51,11 @@ class active_tran_server
...
     bool is_page_server_connected () const;
 
-    void init_log_page_broker ();
-    void finalize_log_page_broker ();
+    void init_page_brokers ();
+    void finalize_page_brokers ();
 
-    cublog::page_broker &get_log_page_broker ();
+    template <typename PageT>
+    page_broker<PageT> &get_page_broker ();
```
cristiarg: It makes not sense to have one template function unless it has a generic implementation, but since there are already two explicit implementations ... make'em two functions.

src/server/active_tran_server.cpp
```diff
@@ -281,25 +281,37 @@ active_tran_server::is_page_server_connected () const
...
+active_tran_server::init_page_brokers ()
 {
-  m_log_page_broker.reset (new cublog::page_broker ());
+  m_log_page_broker.reset (new page_broker<log_page_type> ());
+  m_data_page_broker.reset (new page_broker<data_page_type> ());
 }
 
 void
-active_tran_server::finalize_log_page_broker ()
+active_tran_server::finalize_page_brokers ()
```
cristiarg, line 299: Some assert in dtor that these have been explicitely finalized?

src/server/page_server.cpp
```diff
@@ -166,13 +166,18 @@ void page_server::on_log_page_read_result (const LOG_PAGE *log_page, int error_c
 void page_server::on_data_page_read_result (const FILEIO_PAGE *io_page, int error_code)
 {
   char buffer[sizeof (int) + IO_MAX_PAGE_SIZE];
```
cristiarg: Actually ... unused variable.

src/storage/file_io.h
```diff
@@ -180,6 +180,11 @@ struct fileio_page_reserved
   INT32 p_reserve_1;
   INT32 p_reserve_2;		/* unused - Reserved field */
   INT64 tde_nonce;		/* tde nonce. atomic counter for temp pages, lsa for perm pages */
+
+  inline bool operator== (const fileio_page_reserved & other) const
```
cristiarg, line 185: This operator does not seem to be complete.
If it is needed to compare two `fileio_page_reserved` incompletely in a specific context, better implement an out-of-type comparison function than pollute the type definition with an incomplete implementation.

src/server/active_tran_server.cpp
danilailie: good catch

src/transaction/log_page_buffer.c
danilailie: Comment was there before the code...so I know what code to write :)

src/server/active_tran_server.hpp
danilailie: Open up to the beauty of templates :)

src/transaction/log_page_broker.tpp
danilailie: the code is full of unique_lock...

src/transaction/log_page_broker.hpp
danilailie: one map instead of these two would make the code extremely hard to follow and modify safely for anyone.

src/transaction/log_page_broker.hpp
danilailie: they are needed. we just need the default version

src/storage/file_io.h
danilailie: This is what Remzi considered enough.

cubrid/CMakeLists.txt
```diff
@@ -354,6 +352,8 @@ set(TRANSACTION_HEADERS
   ${TRANSACTION_DIR}/mvcc.h
   ${TRANSACTION_DIR}/mvcc_active_tran.hpp
   ${TRANSACTION_DIR}/mvcc_table.hpp
+  ${TRANSACTION_DIR}/page_broker.hpp
```
ribram: The STORAGE_DIR is a better fit for a generic page broker.

src/server/active_tran_server.cpp
```diff
@@ -281,25 +281,43 @@ active_tran_server::is_page_server_connected () const
...
+template<>
+page_broker<data_page_type> &
+active_tran_server::get_page_broker ()
+{
+  if (!m_data_page_broker)
+    {
+      m_data_page_broker.reset (new page_broker<data_page_type> ());
+    }
+  return *m_data_page_broker;
+}
```
ribram: There is debatable benefit in templetizing get_page_broker function, because there are different definitions. Having two functions, get_log_page_broker and get_data_page_broker, is a more straight-forward approach.

src/server/active_tran_server.cpp
```diff
@@ -281,25 +281,43 @@ active_tran_server::is_page_server_connected () const
...
-active_tran_server::get_log_page_broker ()
+template<>
+page_broker<log_page_type> &
+active_tran_server::get_page_broker ()
 {
-  assert (m_log_page_broker);
+  if (!m_log_page_broker)
+    {
+      m_log_page_broker.reset (new page_broker<log_page_type> ());
+    }
```
ribram: The page brokers should be already initialized before they're first needed. Best to use a safe-guard and to avoid hiding an initialization error.

```suggestion
  assert (m_log_page_broker != nullptr);
```

src/server/active_tran_server.cpp
```diff
@@ -351,23 +369,34 @@ void active_tran_server::receive_data_page (cubpacking::unpacker &upk)
   std::string message;
   upk.unpack_string (message);
 
-  int error_code;
-  std::memcpy (&error_code, message.c_str (), sizeof (error_code));
+  if (message.size () == sizeof (int))
```
ribram, line 373: A comment could help
```suggestion
  // The message is either an error code or the content of the data page
  if (message.size () == sizeof (int))
```

src/server/active_tran_server.cpp
```diff
@@ -351,23 +369,34 @@ void active_tran_server::receive_data_page (cubpacking::unpacker &upk)
...
+
+      m_data_page_broker->set_page (id, std::move (shared_data_page));
+
+      if (prm_get_bool_value (PRM_ID_ER_LOG_READ_DATA_PAGE))
 	{
-	  char buf[IO_MAX_PAGE_SIZE + MAX_ALIGNMENT];
-	  FILEIO_PAGE *io_page = (FILEIO_PAGE *) PTR_ALIGN (buf, MAX_ALIGNMENT);
-	  std::memcpy (io_page, message.c_str () + sizeof (error_code), db_io_page_size ());
 	  _er_log_debug (ARG_FILE_LINE, "Received data page message from Page Server. LSA: %lld|%d, Page ID: %ld, Volid: %d",
 			 LSA_AS_ARGS (&io_page->prv.lsa), io_page->prv.pageid, io_page->prv.volid);
```
ribram, line 400: Move the declaration of io_page to the scope where it is used.
```suggestion
	  auto io_page = reinterpret_cast<const FILEIO_PAGE *> (shared_data_page->c_str ());
	  _er_log_debug (ARG_FILE_LINE, "Received data page message from Page Server. LSA: %lld|%d, Page ID: %ld, Volid: %d",
			 LSA_AS_ARGS (&io_page->prv.lsa), io_page->prv.pageid, io_page->prv.volid);
```

src/server/active_tran_server.cpp
```diff
@@ -351,23 +369,34 @@ void active_tran_server::receive_data_page (cubpacking::unpacker &upk)
...
-  if (prm_get_bool_value (PRM_ID_ER_LOG_READ_DATA_PAGE))
+      if (prm_get_bool_value (PRM_ID_ER_LOG_READ_DATA_PAGE))
+	{
+	  _er_log_debug (ARG_FILE_LINE, "Received data page message from Page Server. Error: %d \n", error_code);
+	}
+    }
+  else
     {
-      if (error_code == NO_ERROR)
+      // We have a page.
```
ribram, line 387: Add safe-guard that the message size matches the log page size.
```suggestion
      assert (message.size () == db_io_page_size ());

      // We have a page.
```

src/server/active_tran_server.cpp
```diff
@@ -351,23 +369,34 @@ void active_tran_server::receive_data_page (cubpacking::unpacker &upk)
...
+      auto shared_data_page = std::make_shared<std::string> (message);
```
ribram:
```suggestion
      auto shared_data_page = std::make_shared<std::string> (std::move (message));
```

src/server/page_server.cpp
ribram:
```suggestion
```

src/storage/page_buffer.c
```diff
@@ -7754,6 +7754,22 @@ pgbuf_claim_bcb_for_fix (THREAD_ENTRY * thread_p, const VPID * vpid, PAGE_FETCH_
...
+	if (get_server_type () == SERVER_TYPE_TRANSACTION)
+	  {
+	    auto data_page = ats_Gl.get_page_broker<data_page_type> ().wait_for_page (*vpid);
+	
+	    char buf[IO_MAX_PAGE_SIZE + MAX_ALIGNMENT];
+	    FILEIO_PAGE *io_page = (FILEIO_PAGE *) PTR_ALIGN (buf, MAX_ALIGNMENT);
+	    std::memcpy (io_page, data_page->c_str (), db_io_page_size ());
+
+	    assert (bufptr->iopage_buffer->iopage.prv == io_page->prv);
+    	}
```
ribram: Indentation of closing bracket looks off.

src/transaction/page_broker.cpp.hpp
```diff
@@ -0,0 +1,88 @@
...
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+
+#include "page_broker.hpp"  // for convenience - code completion.
```
ribram: The convention used so far is to include the implementation of inline/template functions in the same header that they're declared.

Please move the content of page_broker.cpp.hpp at the end of page_broker.hpp (a comment may explain where the implementation part starts).

src/transaction/page_broker.hpp
```diff
@@ -0,0 +1,92 @@
...
+  ADDED_ENTRY,
+  EXISTING_ENTRY
+};
+
+template <typename PageT>
+class page_broker
+{
+  public:
+    page_broker () = default;
+    ~page_broker () = default;
```
ribram: As @cristiarg pointed, the default constructor/destructor are implicitly declared in the absence of any constructors/destructors. Therefore declaring them here is unnecessary.
```suggestion
```

src/transaction/page_broker.hpp
```diff
@@ -0,0 +1,92 @@
...
+{
+  typedef vpid key;
+  typedef std::shared_ptr<std::string> value;
+};
+
+enum entry_state
+{
+  ADDED_ENTRY,
+  EXISTING_ENTRY
+};
```
ribram: Move entry_state enum inside page_broker class to avoid name collisions.

src/transaction/page_broker.hpp
```diff
@@ -0,0 +1,92 @@
...
+namespace std
+{
+  template<> struct less<VPID>
+  {
+    bool operator () (const VPID &lhs, const VPID &rhs) const
+    {
+      return lhs.pageid < rhs.pageid ? true : lhs.pageid > rhs.pageid ? false : lhs.volid < rhs.volid; // debug this!
+    }
+  };
+}
```
ribram: Why must this be part of namespace std?

Note: This should be moved to vpid_utils, after #2842 is merged.

src/transaction/page_broker.cpp.hpp
```diff
@@ -0,0 +1,88 @@
...
+page_broker<PageT>::register_entry (typename map_type<PageT>::key id)
+{
+  std::unique_lock<std::mutex> lock (m_pages_mutex);
+
+  entry_state result = EXISTING_ENTRY;
+  if (m_requested_page_id_count.find (id) != m_requested_page_id_count.end ())
+    {
+      m_requested_page_id_count[id]++;
+    }
+  else
```
ribram: As @cristiarg pointed out too, when ID is found, it is actually searched twice. Once by find() and once by operator[]. Save the iterator from find() to avoid the double lookup.

src/transaction/page_broker.cpp.hpp
```diff
@@ -0,0 +1,88 @@
...
+  return result;
+}
+
+template <typename PageT>
+void
+page_broker<PageT>::set_page (typename map_type<PageT>::key id, typename map_type<PageT>::value &&page)
+{
+  {
+    std::unique_lock<std::mutex> lock (m_pages_mutex);
+    m_received_pages.insert (std::make_pair (id, page));
```
ribram: Is it possible to set same page multiple times?

```suggestion
    assert (m_received_pages.find (id) == m_received_pages.cend ());
    m_received_pages.insert (std::make_pair (id, page));
```

src/transaction/page_broker.hpp
danilailie: That will templatize them.. I would rather leave them as they are. There was a good reason to take that enum outside of the class in the first place.

src/transaction/page_broker.hpp
danilailie: A: because the template struct less is part of std namespace.

src/transaction/page_broker.cpp.hpp
danilailie: A: no, because there are no multiple requests sent for the same page, so there won't be multiple responses for the same page.

src/storage/file_io.h
```diff
@@ -180,6 +180,11 @@ struct fileio_page_reserved
...
+  {
+    return lsa == other.lsa && pageid == other.pageid && volid == other.volid;
+  }
```
ribram, line 188: I think indent should be disabled. It may be confused by the member function.

src/server/active_tran_server.hpp
```diff
@@ -19,7 +19,7 @@
 #ifndef _ACTIVE_TRAN_SERVER_HPP_
 #define _ACTIVE_TRAN_SERVER_HPP_
 
-#include "log_page_broker.hpp"
+#include "page_broker.hpp"
```
ribram: Order the header includes alphabetically.

src/storage/page_broker.hpp
```diff
@@ -0,0 +1,162 @@
...
+namespace std
+{
+  template<> struct less<VPID>
+  {
+    bool operator () (const VPID &lhs, const VPID &rhs) const
+    {
+      return lhs.pageid < rhs.pageid ? true : lhs.pageid > rhs.pageid ? false : lhs.volid < rhs.volid; // debug this!
+    }
+  };
+}
```
ribram: Nitpicking, this could also be written as:
```suggestion
template<>
struct std::less<VPID>
{
  bool operator () (const VPID &lhs, const VPID &rhs) const
  {
    return lhs.pageid < rhs.pageid ? true : lhs.pageid > rhs.pageid ? false : lhs.volid < rhs.volid; // debug this!
  }
};
```
Please choose whichever version you like :).

src/storage/page_broker.hpp
```diff
@@ -0,0 +1,162 @@
...
+{
+  typedef vpid key;
+  typedef std::shared_ptr<std::string> value;
+};
+
+enum entry_state
+{
+  ADDED_ENTRY,
+  EXISTING_ENTRY
+};
```
ribram: I don't think there is really a problem of having the entry_state in the class and using it with the template. But if leaving the enum outside is the preferred choice, the name should include the "page_broker" suffix to be specific enough and to avoid naming collisions (e.g. `enum page_broker_entry_state` or `enum page_broker_registration_response`)

ribram:
There is a build error regarding std::less<VPID>, please fix it.

src/storage/vpid_utilities.hpp
```diff
@@ -19,6 +19,9 @@
 #ifndef _VPID_UTILS_HPP_
 #define _VPID_UTILS_HPP_
 
+#include "dbtype_def.h"
+
+#include <algorithm>
 #include <cstdint>
 #include <cstdio>
```
ribram, line 25: I think \<functional\> has to be included (instead \<algorithm\>).

danilailie: done

ribram (approved):

PR #2824, opened 2021-05-28
[CBRD-23968] dblink: type casting and proper error handling

beyondykk9:
http://jira.cubrid.org/browse/CBRD-23968

**Data Coerce (Type Casting):**

DBLink support only the following data types:

CHAR STRING NCHAR VARNCHAR
BIT VARBIT
NUMERIC INT BIGINT SHORT MONETARY FLOAT DOUBLE
DATE TIME TIMESTAMP DATETIME DATETIMETZ DATETIMELTZ TIMESTAMPTZ  TIMESTAMPLTZ

The above data types should be coerced to other data types if it needed as below example:

```
SELECT rt.col1, rt.col2, rt.col3
FROM dblink('URL=192.168.1.8:55300:demodb USER=dba PASSWORD=',
      "SELECT col1, col2_varchar, col3_big FROM dblink WHERE col4_int > 0")
      as rt(col1 float, col2 string, col3 double), dblink2 as loc
WHERE loc.col1_loc = rt.col1;
```
If the col1 has INTEGER type at remote-db, the result value should be coerced to FLOAT value.

**Error Handling:**

We need to handle the error properly for dblink. The error can be at connection processing as well as at data processing. We may need a new error category for dblink.
For example, the below query should print error message:
```
SELECT /*+ recompile ordered */ rt.col1, rt.col2, rt.col3
FROM dblink('URL=cci:CUBRID:192.168.1.8:55300:demodb::: USER=dba PASSWORD=',
           "SELECT col1, col2_varchar, col3_big FROM dblink WHERE col4_int > 0")
                       as rt(col1 int, col2 int, col3 string), dblink2 as loc
WHERE loc.col1_loc = rt.col1;

ERROR: Cannot coerce value of domain "character varying" to domain "integer".
```

I fixed the function do_create_entity for processing error while creating a table from dblink as follows:
```
create table dblink4
    as select t.col1, t.col2, t.col4 
	   from dblink('URL=cci:CUBRID:192.168.1.8:55300:demodb::: USER=dba PASSWORD=', 
	               "select col1, col2_varchar, col3_big from dblink") 
				    as t(col1 int, col2 varchar(20), col4 biginit);
```
the above query raise segmentation fault (core dumped), so I fixed the code in this issue because the error is from dblink-related.


src/base/error_code.h
```diff
@@ -1643,7 +1643,10 @@
 #define ER_SCALABILITY_DEV_RESERVED_ERROR19         -1282
 #define ER_SCALABILITY_DEV_RESERVED_ERROR20         -1283
 
-#define ER_LAST_ERROR                               -1284
+#define ER_DBLINK		    		    -1284
+#define ER_DBLINK_INVALID_COLUMNS_SPECIFIED         -1285
```
hgryoo: trivial
```suggestion
#define ER_DBLINK		    		      -1284
#define ER_DBLINK_INVALID_COLUMNS_SPECIFIED         -1285
```

src/query/dblink_scan.c
```diff
@@ -120,39 +112,101 @@ dblink_get_basic_utype (T_CCI_U_EXT_TYPE u_ext_type)
...
+}
+
+static int
 dblink_make_date_time (T_CCI_U_TYPE utype, DB_VALUE * value_p, T_CCI_DATE * date_time)
 {
   DB_TIME t_time;
   DB_DATE t_date;
   DB_DATETIME t_datetime;
   DB_TIMESTAMP t_timestamp;
+  int error;
```
hgryoo: it should be initialized
```suggestion
  int error = NO_ERROR;
```

src/query/dblink_scan.c
```diff
@@ -439,13 +605,9 @@ dblink_scan_next (DBLINK_SCAN_INFO * scan_info, val_list_node * val_list)
...
 
-  if (scan_info->conn_handle >= 0 && scan_info->stmt_handle >= 0)
-    {
-      scan_info->cursor = CCI_CURSOR_FIRST;
-      return S_SUCCESS;
-    }
+  scan_info->cursor = CCI_CURSOR_FIRST;
 
-  return S_ERROR;
+  return S_SUCCESS;
```
ctshim, line 615: The function always returns S_SUCCESS so how about altering the function return type as void?

src/query/dblink_scan.c
hgryoo: `value_p` also need to be initialized to null
```
db_make_null (value_p);
```

src/query/dblink_scan.c
```diff
@@ -161,40 +215,88 @@ dblink_make_date_time_tz (T_CCI_U_TYPE utype, DB_VALUE * value_p, T_CCI_DATE_TZ
   DB_DATETIMETZ tz_datetime;
   DB_TIMESTAMPTZ tz_timestamp;
   TZ_REGION region;
+  int error;
```
hgryoo: the error should be initialized here too
```suggestion
  int error = NO_ERROR;
```

src/query/dblink_scan.c
```diff
@@ -161,40 +215,88 @@ dblink_make_date_time_tz (T_CCI_U_TYPE utype, DB_VALUE * value_p, T_CCI_DATE_TZ
...
+      error =
+	tz_create_datetimetz (&t_datetime, date_time_tz->tz, strlen (date_time_tz->tz), &region, &tz_datetime, NULL);
+      if (error != NO_ERROR)
+	{
+	  break;
+	}
+      error = db_make_datetimeltz (value_p, &tz_datetime.datetime);
       break;
     default:
+      assert (false);
```
hgryoo, line 297:
```suggestion
      error = ER_FAILED; // proper error, I just set dummy one for code review.
      assert (false);
```

src/query/dblink_scan.c
```diff
@@ -120,39 +112,101 @@ dblink_get_basic_utype (T_CCI_U_EXT_TYPE u_ext_type)
...
+	}
+      error = db_timestamp_encode (&t_timestamp, &t_date, &t_time);
+      if (error != NO_ERROR)
+	{
+	  break;
+	}
+      error = db_make_timestamp (value_p, t_timestamp);
       break;
     default:
+      assert (false);
```
hgryoo, line 203:
```suggestion
      error = ER_FAILED; // proper error, I just set dummy one for code review.
      assert (false);
```

src/query/dblink_scan.c
```diff
@@ -275,7 +392,7 @@ SCAN_CODE
 dblink_scan_next (DBLINK_SCAN_INFO * scan_info, val_list_node * val_list)
 {
   T_CCI_ERROR err_buf;
-  int col_no, col_cnt, ind, error;
+  int col_no, col_cnt, ind, error = S_SUCCESS;
```
hgryoo:
```suggestion
  int col_no, col_cnt, ind; 
  SCAN_CODE error = S_SUCCESS;
```

src/query/dblink_scan.c
```diff
@@ -249,17 +355,28 @@ dblink_open_scan (DBLINK_SCAN_INFO * scan_info, char *conn_url, char *user_name,
...
+  /*  note: return NO_ERROR even though the connection or stmt handle is not valid */
+
+  if (scan_info->stmt_handle >= 0)
     {
-      return error;
+      if ((error = cci_close_req_handle (scan_info->stmt_handle)) < 0)
+	{
+	  cci_get_err_msg (error, err_buf.err_msg, sizeof (err_buf.err_msg));
+	  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_DBLINK, 1, err_buf.err_msg);
+	  return S_ERROR;
```
ctshim, line 372: Even if the close of the stmt handle fails,
Closing conn_handle would be better if possible.

src/query/dblink_scan.c
beyondykk9: Ok, thanks

src/query/dblink_scan.c
beyondykk9: it must not be reached because this routine is called with only the proper date/time type from dblink_scan_next.
it should return an error, we need a proper error number.

src/query/dblink_scan.c
beyondykk9: Ok, thanks

src/query/dblink_scan.c
beyondykk9: it must not be reached because this routine is called with only the proper date/time zone type from dblink_scan_next.
it should return an error, we need a proper error number.

src/query/dblink_scan.c
beyondykk9: Ok, thanks

src/query/dblink_scan.c
beyondykk9: Ok, thanks, but the error is used only for cci API function.
the error is returned via S_ERROR or S_SUCCESS not by error.

src/query/dblink_scan.c
beyondykk9: scan_reset_scan_block calls this routine and it request error code. finally scan_reset_scan_block returns a status (error code)

src/query/dblink_scan.c
hgryoo: Then it should be initialized to NO_ERROR.
NO_ERROR is 0 and S_SUCCESS is 1.

beyondykk9: Thanks, I fixed it.

src/query/execute_schema.c
```diff
@@ -8616,28 +8616,39 @@ do_create_entity (PARSER_CONTEXT * parser, PT_NODE * node)
...
-        }
+	{
+	  if (column->domain == NULL)
+	    {
+	      /*
+	       * this might be from dblink which has errors for column definition
+	       * the error code is not need to set at this point
+	       * because the error code is already set from dblink
+	       */
+	      error = -1;
```
shparkcubrid: I suggest below.
```
error = er_errid ();
OR
error =ER_FAILED
```

shparkcubrid: How about generating an error when column->domain creation fails?

beyondykk9: Ok, thanks.

shparkcubrid (approved):

ctshim (approved):

hgryoo (approved):

youngjinj (approved):

PR #2825, opened 2021-05-31
[CBRD-23893] Change the expression of <dblink_conn>

ctshim:
http://jira.cubrid.org/browse/CBRD-23893

Purpose
-    <dblink_conn> Change of connection information expression format
-    Validation of access information
-   Correcting and adding messages in case of errors
-   Remove unnecessary comments from existing code

Implementation
N/A

Remarks
AS IS:
```
SELECT /*+ recompile ordered */ rt.col1, rt.col2, rt.col3
FROM dblink('URL=cci:CUBRID:192.168.1.8:55300:demodb::: USER=dba PASSWORD=', 
		"SELECT col1, col2_varchar, col3_big FROM dblink WHERE col4_int > 0") 
		as rt(col1 int, col2 string, col3 bigint), dblink2 as loc
WHERE loc.col1_loc = rt.col1;
```

TO BE:
```
SELECT /*+ recompile ordered */ rt.col1, rt.col2, rt.col3
FROM dblink('HOST=192.168.1.8 PORT=55300 DBNAME=demodb USER=dba PASSWORD=', 
		'SELECT col1, col2_varchar, col3_big FROM dblink WHERE col4_int > 0') 
		as rt(col1 int, col2 string, col3 bigint), dblink2 as loc
WHERE loc.col1_loc = rt.col1;
```


hgryoo:
Could you consider placing static functions in separate files? csql_grammer.y is too long (almost 30000 lines) and this file is not managed on the code formatting tool.
Some parts of this file use functions declared at `parser.h` and the implementation is in `*.c` file such as parser_tree.c, etc. 
Also, validating ipv4 could be used in common for a parser later. so the scope doesn't have to have in the csql_grammer file.

ctshim:
@hgryoo OK,, I agree.

shparkcubrid (approved):

beyondykk9:
Please check the printed line number at error case. I encountered incorrect line number at error case as like "In line 943075630, column 942551342,"

beyondykk9:
enum type should be included in "not supported type"

src/parser/csql_grammar.y
```diff
@@ -530,12 +530,8 @@ static PT_NODE * pt_create_json_value (PARSER_CONTEXT *parser,
 				       const char *str);
 static void pt_jt_append_column_or_nested_node (PT_NODE * jt_node, PT_NODE * jt_col_or_nested);
 
-typedef struct {
-        char* pUrl;
-        char* pUser;
-        char* pPwd;
-} dblink_conn_info;
-static bool pt_ct_check_fill_connection_info (char* pIn, dblink_conn_info* pInfo);
+#define  DBLINK_CONN_PARAM_CNT   (5)
```
hgryoo:
```suggestion
#define DBLINK_CONN_PARAM_CNT   (5)
```

ctshim: OK,

src/parser/csql_grammar.y
```diff
@@ -24764,26 +24760,29 @@ dblink_expr
                         PT_NODE_PRINT_VALUE_TO_TEXT (this_parser, val);
 		   }
 
-                if( $1->node_type == PT_NAME )
+                if( $1 )
                 {
-                        ct->info.dblink_table.is_name = true;
-                        ct->info.dblink_table.conn = $1;       
+                        if( $1->node_type == PT_NAME )
```
youngjinj, line 24765: When processing the 'alter_stmt' rule, PT_NODE is first created, and the necessary processing is performed in the sub-rule using the parser_save_alter_node() and parser_get_alter_node() functions.

Similarly, how about creating PT_NODE *ct first, and then using the parser_save_ct_node() and parser_get_ct_node() functions to perform the necessary processing according to PT_NAME and PT_VALUE in dblink_conn, a sub-rule of dblink_expr?

ctshim: As you said, it may be possible to modify it, but for now, we will maintain the status quo.

youngjinj (approved):

hgryoo (approved):

beyondykk9 (approved):

PR #2826, opened 2021-06-01
[CBRD-23971] Reduce meaningless pt_walk_private() calls

ctshim:
http://jira.cubrid.org/browse/CBRD-23971

Purpose
Reduce meaningless pt_walk_private() calls.

Implementation
pt_walk_private() is modified so that it is not called when node is null.
According to the above operation, remove PT_NODE_FUNCTION g among the parameters of the pt_apply_xxxx() series functions.
Before calling pt_private_walk() in the pt_apply_xxxx() family of functions, check if it is null.
This is done using macros.

Remarks
N/A

### DIFF
--- a/src/parser/parse_tree_cl.c
+++ b/src/parser/parse_tree_cl.c
@@ -68,6 +68,8 @@
 #define MAX_STRING_SEGMENT_LENGTH 254
 #define DONT_PRT_LONG_STRING_LENGTH 256
 
+#define PT_APPLY_CHECK_ACTION(parser, ptr, arg)  if((ptr)) (ptr) = g ((parser), (ptr), (arg))
+
 typedef struct pt_lambda_arg PT_LAMBDA_ARG;
 struct pt_lambda_arg
 {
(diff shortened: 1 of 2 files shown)


src/parser/parse_tree_cl.c
```diff
@@ -68,6 +68,8 @@
 #define MAX_STRING_SEGMENT_LENGTH 254
 #define DONT_PRT_LONG_STRING_LENGTH 256
 
+#define PT_APPLY_CHECK_ACTION(parser, ptr, arg)  if((ptr)) (ptr) = pt_walk_private ((parser), (ptr), (arg))
```
beyondykk9: I think, PT_APPLY_CHECK_ACTION, macro name is hard to understand. How about PT_APPLY_WALK?

hgryoo: For multi-line macro function, how about `do { } while (0);`?

ctshim: @byungwook-kim I'll take a good suggestion and fix it.

@hgryoo I had the same thought. However, since it is a simple syntax, it is not difficult to understand, and I did not want to wrap it in an empty loop for this. However, because of byungwook-kim's suggestion, I will edit it anyway, so I will edit it at this time.

shparkcubrid (approved):

beyondykk9 (approved):

hgryoo (approved):

youngjinj (approved):

PR #2829, opened 2021-06-02
[CBRD-23978] execute command returns error in prepared statement includes DECODE

beyondykk9:
http://jira.cubrid.org/browse/CBRD-23978

The csql result is different from execution by prepared statement and direct execution.
In case of the same result expected, execution by prepared statement raise error but the direct execution runs successfully.
It fails only if the prepared statement has DECODE function with host variable as follows:
```
PREPARE pst FROM 'SELECT DECODE(?,'''',NULL,?)';
EXECUTE pst using '1', '1'; 

ERROR: Cannot coerce value of domain "character" to domain "*NULL*".
```
but the following statement runs successfully.
```
SELECT DECODE('1', '', NULL, '1');

decode('1', '', null, '1')
======================
'1' 
```

### DIFF
--- a/src/parser/type_checking.c
+++ b/src/parser/type_checking.c
@@ -5444,8 +5444,8 @@ pt_coerce_range_expr_arguments (PARSER_CONTEXT * parser, PT_NODE * expr, PT_NODE
 	{
 	  PT_NODE *temp = NULL;
 	  int precision = 0, scale = 0;
-	  int units = LANG_SYS_CODESET; /* code set */
-	  int collation_id = LANG_SYS_COLLATION; /* collation_id */
+	  int units = LANG_SYS_CODESET;	/* code set */
+	  int collation_id = LANG_SYS_COLLATION;	/* collation_id */
 	  bool keep_searching = true;
 	  for (temp = arg2->data_type; temp != NULL && keep_searching; temp = temp->next)
 	    {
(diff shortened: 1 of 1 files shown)


beyondykk9:
The following case also raise error.
```
prepare st from 'select  case when a=? then NULL else ? end';
execute st using  1,2;

ERROR: Cannot coerce value of domain "integer" to domain "*NULL*".
```
This error is appeared at the sql test case in this PR. The test case should be modified to return successful results.

ctshim:
There is one that test_sql has failed. 
Does the QA team need to change the correct answer by this modification?

beyondykk9:
> There is one that test_sql has failed.
> Does the QA team need to change the correct answer by this modification?

Yes, I already commented for this error.

ctshim (approved):

src/parser/type_checking.c
```diff
@@ -10172,7 +10172,7 @@ pt_eval_expr_type (PARSER_CONTEXT * parser, PT_NODE * node)
 	  common_type = (arg1_type == PT_TYPE_MAYBE && arg2_type != PT_TYPE_NULL) ? arg2_type : arg1_type;
```
shparkcubrid: This issue is related to CUBRIDSUS-9134. https://github.com/CUBRID/cubrid/commit/f892ae1212f3da880371289356bf27371334ffa5

I think PT_TYPE_MAYBE should be kept if arg1 or arg2 are PT_TYPE_NULL.
I recommend the following changes.
```
if (common_type == PT_TYPE_MAYBE && arg1_type != PT_TYPE_NULL && arg2_type != PT_TYPE_NULL)
{
common_type = (arg1_type == PT_TYPE_MAYBE) ? arg2_type : arg1_type;
}
```

beyondykk9: Ok, I modified.

src/parser/type_checking.c
```diff
@@ -5444,8 +5444,8 @@ pt_coerce_range_expr_arguments (PARSER_CONTEXT * parser, PT_NODE * expr, PT_NODE
 	{
 	  PT_NODE *temp = NULL;
 	  int precision = 0, scale = 0;
-	  int units = LANG_SYS_CODESET; /* code set */
-	  int collation_id = LANG_SYS_COLLATION; /* collation_id */
+	  int units = LANG_SYS_CODESET;	/* code set */
+	  int collation_id = LANG_SYS_COLLATION;	/* collation_id */
 	  bool keep_searching = true;
```
beyondykk9, line 5449: it's not related code, it's just for indentation.

shparkcubrid (approved):

hgryoo (approved):

youngjinj (approved):

PR #2830, opened 2021-06-02
[CBRD-23972] Modify pt_init_xxxx() series functions

ctshim:
http://jira.cubrid.org/browse/CBRD-23972

Purpose
 - Since memset() is guaranteed to be called before entering the pt_init_xxxx() family of functions, there is no meaning in the code that sets values such as NULL, 0, and false inside.

Implementation
  - N/A

Remarks
  - N/A

### DIFF
--- a/src/parser/parse_tree_cl.c
+++ b/src/parser/parse_tree_cl.c
@@ -5538,6 +5538,8 @@ pt_apply_alter (PARSER_CONTEXT * parser, PT_NODE * p, PT_NODE_FUNCTION g, void *
 static PT_NODE *
 pt_init_alter (PT_NODE * p)
 {
+  p->info.alter.constraint_list = NULL;
+  p->info.alter.create_index = NULL;
   return p;
 }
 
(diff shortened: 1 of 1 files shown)


beyondykk9 (approved):

youngjinj (approved):

hgryoo:
There are many empty functions. How about removing them and set `NULL` to function pointers in `pt_init_init_f ()`?
```
// pt_init_alter is empty, do nothing
static PT_NODE *
pt_init_alter (PT_NODE * p)
{
  return p;
}

// for example
static void
pt_init_init_f (void)
{
  pt_init_func_array[PT_ALTER] = NULL;
  ...
}
```

Checking `parser_init_node_func` is NULL already done in `pt_init_node()`
```
void
pt_init_node (PT_NODE * node, PT_NODE_TYPE node_type)
{
  PARSER_INIT_NODE_FUNC parser_init_node_func = pt_init_func_array[node_type];
  if (!node || !parser_init_node_func)
    {
      return;
    }

  parser_init_node_func (node);
  node->node_type = node_type;
}
```

And It would better to add a comment in `pt_init_init_f()` or some proper place: "To initialize for any member variable of PT_STATEMENT_INFO to a non-zero value, define a init function and assign a value for the member variable"

shparkcubrid (approved):

hgryoo (approved):

youngjinj (approved):

shparkcubrid (approved):

beyondykk9 (approved):

PR #2831, opened 2021-06-02
[LETS-34] Recovery analysis navigation refactoring

ribram:

There are two LSA variables that are used in log_recovery_analysis to navigate the log: `lsa` and `log_lsa`. `lsa` is supposed to be the address of current/next log record and `log_lsa` is the address of current offset in page. Both have the same value, at least in the first part of the nested loop and they are both used without a very clear rule.

The purpose of this pull request is to make the log_recovery_analysis a bit more clear, by renaming navigation LSA's and using them consistently. The while loops are merged into a single loop.

**Merging loops**

log_recovery_analysis contained two loops that basically do the same thing: iterate through log records. The outer loop advances to the next record into a different page, while the inner loop advances to the next record in the same log page. Merge the loops into a single one and:

  - Check if page has changed in the beginning of the loop by comparing the record navigation LSA with the log navigation LSA. Log navigation LSA always points to the current page, therefore if record LSA pageid is different, the log page must be re-fetched.
  - Remove all `LSA_SET_NULL (&lsa)` operations with the purpose of breaking the outer loop. A break operation is now enough.
  
**Navigation LSA's**

Replace lsa/log_lsa with four different values:

- `record_nav_lsa`: navigation LSA through log records. Sets the value of current log record at the beginning of the loop. Replaces all `lsa` references until a valid value is found (until offset is not NULL_OFFSET)
- `log_nav_lsa`: navigation through log data. acts like a cursor in the log and should always point to the current log page. Is updated whenever the page is changed and by `log_rv_analysis_record` while processing through the data of the log record.
- `const crt_record_lsa`: once a valid `record_nav_lsa` value is set, it is copied to the const value `crt_record_lsa` and this is used onward for all references to the current record LSA. It replaces many of the `lsa` and `log_lsa` references that are meant as current record LSA (all references in the first part of the loop, until `next_record_lsa` is calculated and `log_nav_lsa` is used by `log_rv_analysis_record`).
- `next_record_lsa`: copied from log_rec->forw_lsa and going through corrections when offset is NULL. replaces `lsa` references in the second part of the loop.

**Other**

- Reorganized local variables declaration/initialization
- prev_lsa values starts as null instead of checkpoint LSA. No real difference for the observable behavior.

**Tests**

- Manual tests for simple server start, create table, kill & recovery.
- shell_debug.


ribram:
@reviewers set "Hide whitespace changes" to review

cristiarg (approved):

PR #2833, opened 2021-06-02
[LETS-83] Transaction server connects to all page server

adrianbrab:
http://jira.cubrid.org/browse/LETS-83

Transaction server connects to all the page server in page_server_hosts parameter.
Create connections with all page servers in the page_server_hosts parameter. Add server error log message with all successful and failed connections.

Keep only first successful connection.


src/server/active_tran_server.cpp
```diff
@@ -183,21 +185,25 @@ active_tran_server::init_page_server_hosts (const char *db_name)
       const int local_exit_code = connect_to_page_server (node, db_name);
```
ribram, line 183: Keep all connections by converting m_page_server_conn into a vector of connections. Add all successful connections to the vector.

When requests are pushed, use the first connection in the vector.

adrianbrab: Done

src/server/active_tran_server.cpp
```diff
@@ -174,6 +174,8 @@ active_tran_server::init_page_server_hosts (const char *db_name)
       er_clear ();
     }
   exit_code = NO_ERROR;
+  cubcomm::node first_valid_connection;
```
cristiarg: `first_valid_connection` seems to not be used;

src/server/active_tran_server.cpp
```diff
@@ -183,17 +185,22 @@ active_tran_server::init_page_server_hosts (const char *db_name)
...
+	  if (!connected)
+	    {
+	      first_valid_connection = node;
+	    }
+	  //found valid host clear the errors rom the bad ones
 	  er_clear ();
-	  // successfully connected to a page server. stop now.
-	  break;
+	  connected = true;
+	  continue;
```
cristiarg: remove this `continue` by moving the `er_log_debug..` on the else branch below this comment

src/server/active_tran_server.cpp
```diff
@@ -174,6 +174,8 @@ active_tran_server::init_page_server_hosts (const char *db_name)
       er_clear ();
     }
   exit_code = NO_ERROR;
+  cubcomm::node first_valid_connection;
+  bool connected = false;
```
cristiarg: `connected` and `valid_connection_count` serve the same purpose; could make do with just one of them for the check at the end; potentially with the count which offers more "info"

src/server/active_tran_server.cpp
```diff
@@ -241,8 +248,7 @@ active_tran_server::connect_to_page_server (const cubcomm::node &node, const cha
   er_log_debug (ARG_FILE_LINE, "Transaction server successfully connected to the page server. Channel id: %s.\n",
 		srv_chn.get_channel_id ().c_str ());
 
-  assert (m_page_server_conn == nullptr);
```
cristiarg, line 244: this assert was here for the purpose to prevent all this code from being called twice (albeit buried in the implementation); the purpose of this assert should be served by another check somewhere more to the beginning of the connections loop

src/server/active_tran_server.cpp
```diff
@@ -269,15 +275,15 @@ active_tran_server::disconnect_page_server ()
...
+  m_page_server_conn.clear();
 }
 
 bool
 active_tran_server::is_page_server_connected () const
 {
   assert_is_active_tran_server ();
 
-  return m_page_server_conn != nullptr;
+  return m_page_server_conn.empty();
```
cristiarg:
```suggestion
  return false == m_page_server_conn.empty();
```
?

src/server/active_tran_server.hpp
```diff
@@ -78,7 +78,7 @@ class active_tran_server
     std::string m_ps_hostname;
     int m_ps_port = -1;
 
-    std::unique_ptr<page_server_conn_t> m_page_server_conn;
+    std::vector<std::unique_ptr<page_server_conn_t>> m_page_server_conn;
```
cristiarg: I would change the name of the member to smth that suggests it is a collection and not a single connection:
```suggestion
    std::vector<std::unique_ptr<page_server_conn_t>> m_page_server_conn_vec;
```
is my preference. Yours might be different. :)

cristiarg:
I see that the issue specifies:
```
Keep only first successful connection.
```
So only the first connection should be kept and the other dropped?
I don't actually know what's best because the next issue is to keep and send data to all page servers.
So I guess it is ok to just keep'em all.

src/server/active_tran_server.cpp
adrianbrab: done. was old code relics

src/server/active_tran_server.cpp
adrianbrab: removed connected, the count had other purposes.

src/server/active_tran_server.cpp
adrianbrab: should ahve been ```!m_page_server_conn.empty()```, changed now

src/server/active_tran_server.hpp
adrianbrab: a proper name for a proper field.

src/server/active_tran_server.cpp
```diff
@@ -208,15 +206,15 @@ active_tran_server::init_page_server_hosts (const char *db_name)
   er_log_debug (ARG_FILE_LINE, "Transaction server runs on %s storage.",
 		m_uses_remote_storage ? "remote" : "local");
 
-  assert (exit_code == NO_ERROR);
-  return exit_code;
+  // failed to connect to any page server
+  assert (valid_connection_count);
+  return valid_connection_count ? NO_ERROR : exit_code;
```
cristiarg: According to the table at the start of the function, not having any valid connection, and if remote storage is not specified, is OK.

But, to prevent early out and also to only log successful message when actually successfull:
```
  if (valid_connection_count == 0 && m_uses_remote_storage)
    {
      assert (exit_code != NO_ERROR);
      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_NO_PAGE_SERVER_CONNECTION, 0);
    }
else if (valid_connection_count == 0)
  {
      // failed to connect to any page server
      assert (exit_code != NO_ERROR);
      er_clear ();
    //  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, exit_code, 0);
      exit_code = NO_ERROR;
      er_log_debug (ARG_FILE_LINE, "Transaction server runs on %s storage.",
		m_uses_remote_storage ? "remote" : "local");
  }
  return exit_code;
```

Note:
regarding
```
assert (valid_connection_count);
```
Don't force an integer into a bool. Be explicit with `> 0`. On the pedantic side, `assert(-1)` will also pass even if this is not what we want from a `count`.

src/server/active_tran_server.cpp
```diff
@@ -183,16 +182,15 @@ active_tran_server::init_page_server_hosts (const char *db_name)
       const int local_exit_code = connect_to_page_server (node, db_name);
       if (local_exit_code == NO_ERROR)
 	{
-	  // found valid host clear the errors from the bad ones
 	  ++valid_connection_count;
 	  exit_code = NO_ERROR;
+	  //found valid host clear the errors rom the bad ones
```
cristiarg:
```suggestion
	  // found valid host; clear the error possibly remaining from previous failing ones
```
What if the last connection attempt is actually a failing one, but there is at least one valid one. Don't we want to `er_clear` in that case?

src/server/active_tran_server.hpp
```diff
@@ -78,7 +78,7 @@ class active_tran_server
     std::string m_ps_hostname;
     int m_ps_port = -1;
 
-    std::unique_ptr<page_server_conn_t> m_page_server_conn;
+    std::vector<std::unique_ptr<page_server_conn_t>> m_page_server_conn_vec;
```
cristiarg, line 81: Even if not used anywhere, I would prefer a private typedef for `unique_ptr<page_server_conn_t>`. Makes code easier to read.
But I know @ribram does not.
Your choice. :)

src/server/active_tran_server.cpp
adrianbrab: moved er_clear so it is only called once

src/server/active_tran_server.hpp
adrianbrab: kept it the same way for now

src/server/active_tran_server.cpp
adrianbrab: done

src/server/active_tran_server.cpp
```diff
@@ -204,19 +207,30 @@ active_tran_server::init_page_server_hosts (const char *db_name)
...
+      er_clear ();
+      exit_code = ER_NO_PAGE_SERVER_CONNECTION;
+      er_log_debug (ARG_FILE_LINE, "Transaction server runs on %s storage.",
+		    m_uses_remote_storage ? "remote" : "local");
+    }
+  else
+    {
+      er_log_debug (ARG_FILE_LINE, "Transaction server runs on %s storage.",
+		    m_uses_remote_storage ? "remote" : "local");
+    }
```
cristiarg:
```suggestion
  else if (valid_connection_count == 0)
    {
      // failed to connect to any page server
      assert (exit_code != NO_ERROR);
      er_clear ();
      exit_code = NO_ERROR;
    }
    er_log_debug (ARG_FILE_LINE, "Transaction server runs on %s storage.",
		    m_uses_remote_storage ? "remote" : "local");

```

src/server/active_tran_server.cpp
```diff
@@ -204,19 +207,30 @@ active_tran_server::init_page_server_hosts (const char *db_name)
...
 
-  er_log_debug (ARG_FILE_LINE, "Transaction server runs on %s storage.",
-		m_uses_remote_storage ? "remote" : "local");
-
-  assert (exit_code == NO_ERROR);
-  return exit_code;
+  // failed to connect to any page server
+  assert (valid_connection_count > 0);
```
cristiarg:
```suggestion
  assert (valid_connection_count > 0 || !m_uses_remote_storage);
```

src/server/active_tran_server.cpp
```diff
@@ -204,19 +207,30 @@ active_tran_server::init_page_server_hosts (const char *db_name)
...
+  return valid_connection_count > 0 ? NO_ERROR : exit_code;
```
cristiarg:
```suggestion
  return exit_code;
```

cristiarg (approved):

src/server/active_tran_server.cpp
adrianbrab: changed now

src/server/active_tran_server.cpp
adrianbrab: done

src/server/active_tran_server.cpp
adrianbrab: fixed

danilailie (approved):

PR #2834, opened 2021-06-02
[LETS-107] prevent some tools to be executed when 'remote_storage' param is set

cristiarg:
http://jira.cubrid.org/browse/LETS-107

`copydb`, `backupdb`, `restoredb` and `emergency_patchlog` tools run on the disk database files. The tool is not applicable to a transaction server running on remote storage. An error is returned.

**Implementation**
- added new error code & message
- prevented tools to be executed by patching corresponding `xboot_*` functions

**Tests**
- manual tests


src/base/server_type.cpp
```diff
@@ -83,8 +83,18 @@ void finalize_server_type ()
 
 int init_server_type (const char *)
 {
-  g_server_type = SERVER_TYPE_TRANSACTION;
-  return NO_ERROR;
+  int err_code = NO_ERROR;
+
+  g_server_type = (SERVER_TYPE) prm_get_integer_value (PRM_ID_SERVER_TYPE);
```
ribram: SA_MODE still won't act as a page server. Better to leave it a transaction server, regardless of the system parameter. Later on, when transaction server will lose local storage, maybe there will be different server types - transaction and SA. To be determined.
```suggestion
  g_server_type = SERVER_TYPE_TRANSACTION;
```

msg/de_DE.utf8/cubrid.msg
```diff
@@ -1338,7 +1338,7 @@ $ LOADDB
 1267 Cannot copy volume %s to %s. Error: %s.
 1268 At least one valid page server host is necessary, but no page server found in configuration.
 1269 At least one valid page server host is necessary, but no page server could be connected.
-1270 Error reserved for scalability development.
+1270 This tool (%1$s) cannot be executed on a server with remote storage enabled.
```
ribram:
```suggestion
1270 %1$s cannot be executed on a server with remote storage enabled.
```

cristiarg: Done

src/base/server_type.cpp
cristiarg: Done

src/transaction/boot_sr.c
```diff
@@ -5448,6 +5464,16 @@ xboot_emergency_patch (const char *db_name, bool recreate_log, DKNPAGES log_npag
       goto error_exit;
     }
 
+  {
+    const bool uses_remote_storage = prm_get_bool_value (PRM_ID_REMOTE_STORAGE);
+    if (uses_remote_storage)
+      {
+	er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_TOOL_INVALID_WITH_REMOTE_STORAGE, 1, "'emergency_patchlog'");
```
adrianbrab, line 5471: the indentation for this if block looks strange, is it git problem? If not I think it should be changed

adrianbrab (approved):

src/transaction/boot_sr.c
cristiarg, line 5471: It's github's rendering of tabs at an arbitrary width.

PR #2835, opened 2021-06-03
[CBRD-23973] Improvements to the keyword_offset() function in csql_grammar.y

ctshim:
http://jira.cubrid.org/browse/CBRD-23973

Purpose
   Since the configuration of the function or source is consistent with what was covered in the "CBRD-23955 Speed improvement of pt_find_keyword() function" issue, it can be improved in the same way.

Implementation
  Rename keyword_offset() to pt_find_function_name().
 The function name list table and functions for searching for this table are moved to the keyword.c file.
 Change the lookup method in the same way as the pt_find_keyword() function.

Remarks
  N/A

### DIFF
--- a/src/parser/keyword.c
+++ b/src/parser/keyword.c
@@ -38,6 +38,7 @@
  * will be sorted when needed. See pt_find_keyword.
  */
 
+/* Keyword names must be written in uppercase */
 static KEYWORD_RECORD keywords[] = {
   {ABSOLUTE_, "ABSOLUTE", 0},
   {ACCESS, "ACCESS", 1},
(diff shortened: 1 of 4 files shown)


shparkcubrid (approved):

beyondykk9 (approved):

hgryoo:
If you don't mind, how about refactoring this change to C++ style?
`function_keyword_cmp` and `keyword_cmp` are totally same comparator except the type.
They compare `hash_value` and if they are same, compare `keyword`.
In my opinion, `std::sort` and `generic comparator` make code much shorter and readable.

I suggest the following refactoring lists

1) define common comparator
```
template<typename T>
class keyword_hash_comparator {
    bool operator()(const T &lhs, const T &rhs) const {
        return (lhs.hash_value < rhs.hash_value
                || (lhs.hash_value  == rhs.hash_value
                    && strcmp (lhs.keyword, rhs.keyword))
    }
};
```

2) call `std::sort` instead of `qsort`
```
std::sort (std::begin (functions), std::end(functions), keyword_hash_comparator);
```

Too many kinds of routines are included in `pt_find_function_name()`. and I'm not sure this function name clearly describes them.

3) extract initializing routine setting hash value into a new function
```
  // keyword.c:949~ : initializing routine and run only once
  if (function_keyword_sorted == false)
    {
      for (i = 0; i < functions_cnt; i++)
	{
	  len = strlen (functions[i].keyword);
   ...
```

4) The routine searches on bounded array elements under `MAGIC_NUM_BI_SEQ` value can be extracted too.
```
 // keyword.c:1014 ~
if (len <= MAGIC_NUM_BI_SEQ)
    {
      for (len = function_start_pos[i]; len < function_start_pos[i + 1]; len++)
	{
	  if (dummy.hash_value > functions[len].hash_value)
	    {
	      continue;
	    }
	  else if (dummy.hash_value < functions[len].hash_value)
	    {
	      return NULL;
```

My suggestion could proceed on another issue. The decision is yours.

ctshim:
@hgryoo 

Thanks for the good comments.
I also like to unify functions and break them down into smaller ones.
I will make corrections based on your comments.

ctshim:
The two functions, pt_find_function_name() and pt_find_keyword(), which differ only in the structure used but have almost identical execution contents, are reorganized using a template.

The pt_get_keyword_rec() function, which is not used anywhere, is
It is bound to be activated only when ENABLE_UNUSED_FUNCTION is declared.

Added function verify_test() for self-verification when NDEBUG is not declared.

@hgryoo
   The comparison function of the std::sort() function returns a bool.
   I kept the old qsort() to keep only one comparison function that returns an int. This comparison function is also used in bsearch().

src/parser/keyword.c
```diff
@@ -539,129 +695,192 @@ static int keyword_cmp (const void *k1, const void *k2);
...
+  short start_pos[257];		// (0x00 ~ 0xFF) + 1  
+} KEYWORDS_TABLE_SRCH_INFO;
+
+template < typename T > static int
+keyword_hash_comparator (const void *lhs, const void *rhs)
+{
+  int cmp = ((T *) lhs)->hash_value - ((T *) rhs)->hash_value;
 
-  return strcmp (((KEYWORD_RECORD *) k1)->keyword, ((KEYWORD_RECORD *) k2)->keyword);
+  return cmp ? cmp : strcmp (((T *) lhs)->keyword, ((T *) rhs)->keyword);
```
hgryoo, line 714: `cmp` can be a negative value?
```suggestion
  return (cmp != 0) ? cmp : strcmp (((T *) lhs)->keyword, ((T *) rhs)->keyword);
```

src/parser/keyword.c
```diff
@@ -539,129 +695,192 @@ static int keyword_cmp (const void *k1, const void *k2);
...
-      return NULL;
+  if (keyword_sorted == false)
+    {
+      kinfo.min_len = MAX_KEYWORD_SIZE;
+      kinfo.max_len = 0;
+      kinfo.cnt = sizeof (keywords) / sizeof (keywords[0]);
+      memset (kinfo.start_pos, 0x00, sizeof (kinfo.start_pos));
+      kinfo.func_char_convert = char_toupper;
+
+      init_keyword_tables (keywords, kinfo, keyword_hash_comparator < KEYWORD_RECORD >);
```
hgryoo, line 873: How about extracting it to a new function, `pt_init_keyword ()`?

hgryoo:
@ctshim, `std::binary_search()` can be used instead of `bsearch()`. `std::binary_search()` the return value of its comparator is convertible to bool

> The comparison function of the std::sort() function returns a bool.
> I kept the old qsort() to keep only one comparison function that returns an int. This comparison function is also used in bsearch().

src/parser/keyword.c
beyondykk9: the meaning of the code is not same?
" if (cmp) " means cmp is not zero.

ctshim:
@hgryoo 
       I was trying to use that std::binary_search() function.
       The std::binary_search() function returns a bool that only tells if it exists or not.
       So we don't know what the found items are.

src/parser/keyword.c
ctshim: After reading the comments, it seems that it should be moved into the init_keyword_tables() function.

src/parser/keyword.c
ctshim: In the case of "(int) = (unsigned short) <small> - (unsigned short) <big>", the type is calculated according to the large (int) and a negative number is generated.

Byungwook-kim's comment is correct.
The two lines have the same meaning, and if there is no confusion in the interpretation, I try not to do one additional operation.

ctshim:
The keywords parameter of these two functions is the same as the global variable name, so there could be confusion, so we changed it.
find_keyword_tables()
init_keyword_tables()

hgryoo:
> @hgryoo
> I was trying to use that std::binary_search() function.
> The std::binary_search() function returns a bool that only tells if it exists or not.
> So we don't know what the found items are.

To get a searched element and index, `std::lower_bound` can be used.
Please refer to the following example
```
// from https://en.cppreference.com/w/cpp/algorithm/lower_bound
auto lower = std::lower_bound(data.begin(), data.end(), I);
if (lower != data.end())
{
  std::cout << *lower << " at index " << std::distance(data.begin(), lower) << std::endl;
}
else
{
  std::cout << "not found" << std::end;
}
```

src/parser/keyword.c
hgryoo: Yes, if-condition checks the condition is true. a non-zero value in the condition converted to true and a zero value is converted to false.

If someone later sees this part for the first time, it may be difficult to understand the meaning of subtracting the two hash values and the resulting `cmp` value, so I suggest adding it to make more clear.

ctshim:
@hgryoo
    " std::lower_bound"
When I went to the page you mentioned, the internal implementation is in the form of binary search,
I will take note.
However, this time, we want to keep it as is, without further modifications. Would that be okay?

hgryoo:
> @hgryoo
> " std::lower_bound"
> When I went to the page you mentioned, the internal implementation is in the form of binary search,
> I will take note.
> However, this time, we want to keep it as is, without further modifications. Would that be okay?

okay. :-)

hgryoo (approved):

shparkcubrid (approved):

youngjinj (approved):

src/parser/keyword.c
beyondykk9: @hgryoo, it looks more clear. agree or disagree is up to @ctshim.

beyondykk9 (approved):

PR #2840, opened 2021-06-08
[LETS-112] Remove disk dependency of vacuum dump

cristiarg:
http://jira.cubrid.org/browse/LETS-112

On a transaction server with `remote_storage` enabled, disable displaying the archive file name of the minimum required log page.

Testing:
- tested on a transaction server with/without `remote_storage` enabled (`cubrid -C --dump <database name>`)
- tested in stand-alone mode (`cubrid -S --dump <database name>`)

Remarks:
- testing on a page server with `cubrid -C --dump <database name>` returns error message:
```
vacuum did not boot properly.
```

### DIFF
--- a/src/query/vacuum.c
+++ b/src/query/vacuum.c
@@ -1132,10 +1132,18 @@ xvacuum_dump (THREAD_ENTRY * thread_p, FILE * outfp)
 	{
 	  LOG_CS_ENTER_READ_MODE (thread_p);
 	  archive_number = logpb_get_archive_number (thread_p, min_log_pageid);
-	  if (archive_number >= 0)
+	  if (archive_number < 0)
 	    {
-	      fprintf (outfp, "(in %s%s%03d)", log_Prefix, FILEIO_SUFFIX_LOGARCHIVE, archive_number);
+	      /* this is an assertion case but ignore. */
 	    }
+	  else
+	    {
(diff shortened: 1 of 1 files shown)


src/query/vacuum.c
```diff
@@ -1124,27 +1125,29 @@ xvacuum_dump (THREAD_ENTRY * thread_p, FILE * outfp)
 
   fprintf (outfp, "\n");
   fprintf (outfp, "*** Vacuum Dump ***\n");
-  fprintf (outfp, "First log page ID referenced = %lld ", min_log_pageid);
-
-  if (logpb_is_page_in_archive (min_log_pageid))
+  fprintf (outfp, "First log page ID referenced = %lld", (long long int) min_log_pageid);
```
danilailie, line 1128: are we still using fprintf? the previous fprintf's can also be switched to ofstream...

danilailie (approved):

src/query/vacuum.c
cristiarg: given it is existing code, I preferred not to touch it too much;
I did the `long long int` explicit cast to remove a compilation warning; I tried to use the `PRId64` macro but, for a reason I did not understand, CircleCI compilation failed - https://circleci.com/gh/CUBRID/cubrid/24547?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link - (although the macro is used in few other areas of the code)

PR #2841, opened 2021-06-10
[CBRD-23975] Add new system catalog class for 'CREATE SERVER'

ctshim:
http://jira.cubrid.org/browse/CBRD-23975

Purpose
   For dblink support, create a catalog table for "CREATE SERVER" syntax information.
   
   Also, add "conn_server" column to _db_class.
   This is to add in advance in case syntax support for creating VIEW for DBLINK will be added later.

Implementation
   N/A

Remarks
   N/A


src/object/transform.h
```diff
@@ -163,6 +164,7 @@ typedef struct tf_ct_class
 #define CTV_PARTITION_NAME         "db_partition"
 #define CTV_DB_COLLATION_NAME      "db_collation"
 #define CTV_DB_CHARSET_NAME	   "db_charset"
+#define CTV_DB_SERVER_NAME         "db_server"
```
shparkcubrid, line 167: How about 'db_link_server'?

youngjinj: Since it is a system table to support the sql standard create server syntax, and it can be used as a Federated Table concept, doesn't it feel like the scope of use is narrowed to use the prefix with db_link ?

src/object/schema_manager.c
```diff
@@ -2828,7 +2828,8 @@ sm_mark_system_class_for_catalog (void)
     CTV_METHARG_NAME, CTV_METHARG_SD_NAME, CTV_METHFILE_NAME,
     CTV_INDEX_NAME, CTV_INDEXKEY_NAME, CTV_AUTH_NAME,
     CTV_TRIGGER_NAME, CTV_STORED_PROC_NAME, CTV_STORED_PROC_ARGS_NAME,
-    CTV_PARTITION_NAME, CT_COLLATION_NAME, NULL
+    CTV_PARTITION_NAME, CT_COLLATION_NAME, CT_DB_SERVER_NAME,
+    CTV_DB_SERVER_NAME, NULL
```
youngjinj: How about changing it to list only one name per line with this change? (Same opinion about the classes array in the boot_destroy_catalog_classes() function.)

src/transaction/boot_cl.c
```diff
@@ -4013,6 +4014,105 @@ boot_define_dual (MOP class_mop)
...
+ *
+ *   class(IN) :
+ */
+static int
+boot_define_db_server (MOP class_mop)
+{
+  SM_TEMPLATE *def;
+  char args_string[64];
+  int error_code = NO_ERROR;
+  const char *index_col_names[2] = { "link_name", NULL };
```
youngjinj, line 4030: This is a similar discussion to whether the system table name should be 'db_server' or 'db_link_server'. Since server information can be used more extensively, how about using 'name' rather than 'link_name'?

src/transaction/boot_cl.c
```diff
@@ -5477,7 +5647,8 @@ catcls_vclass_install (void)
     {"CTV_STORED_PROC_NAME", boot_define_view_stored_procedure},
     {"CTV_STORED_PROC_ARGS_NAME", boot_define_view_stored_procedure_arguments},
     {"CTV_DB_COLLATION_NAME", boot_define_view_db_collation},
-    {"CTV_DB_CHARSET_NAME", boot_define_view_db_charset}
+    {"CTV_DB_CHARSET_NAME", boot_define_view_db_charset},
+    {CTV_DB_SERVER_NAME, boot_define_view_db_server}
```
youngjinj: In the catcls_class_install() function, the macro variable value is used as the name value in the clist array. How about using the same macro variable for code unity?

ctshim: Unlike the catcls_class_install() function, the catcls_vclass_install() function does not use the name item of the clist structure.
However, semantically, if you see it as the same concept as catcls_class_install(), you need to remove the quotes from the quoted parts like "CTV_CLASS_NAME" in the existing code.
It can be seen as an error in the existing code.

youngjinj:
Fails in "ci/circleci - test_medium, test_sql" are caused by creating test data using db_class or _db_class.

If we create test data in the same way, whenever a change occurs in the system table, we will see an error for several test cases that are not related to the behavior of the changed code. I think this is inefficient.

I would like to avoid using system tables when creating test data. I think that this inefficiency can be reduced by making the test data purely separately.

src/object/transform.h
ctshim: For now, the name will remain as it is.

youngjinj (approved):

src/object/schema_manager.c
ctshim: OK, I agree.

src/transaction/boot_cl.c
ctshim: For now, the name will remain as it is.

shparkcubrid (approved):

src/transaction/boot_cl.c
```diff
@@ -5455,29 +5631,29 @@ catcls_vclass_install (void)
...
+    { boot_define_view_index}, /* CTV_INDEX_NAME */
+    { boot_define_view_index_key}, /* CTV_INDEXKEY_NAME */
+    { boot_define_view_authorization}, /* CTV_AUTH_NAME */
+    { boot_define_view_trigger}, /* CTV_TRIGGER_NAME */
+    { boot_define_view_partition}, /* CTV_PARTITION_NAME */
+    { boot_define_view_stored_procedure}, /* CTV_STORED_PROC_NAME */
+    { boot_define_view_stored_procedure_arguments}, /* CTV_STORED_PROC_ARGS_NAME */
+    { boot_define_view_db_collation}, /* CTV_DB_COLLATION_NAME */
+    { boot_define_view_db_charset}, /* CTV_DB_CHARSET_NAME */
+    { boot_define_view_db_server} /* CTV_DB_SERVER_NAME */
```
hgryoo: This refactoring seems not the scope of this PR.

hgryoo (changes requested):

src/transaction/boot_cl.c
ctshim: The original code was as below.
{**"CTV_CLASS_NAME"**, boot_define_view_class},

If corrected correctly, the quotes should be removed as shown below.
     { **CTV_CLASS_NAME**, boot_define_view_class},
However, the name member variable of the catcls_function structure was never used.

Even if I modify it correctly, it only takes up unnecessary memory, so I modified it because there was an opinion that it would be better to remove it altogether.

Since there are no other influences, I don't think there is a need to make it a separate issue. Would it be better to treat it as a separate issue though?

hgryoo: I couldn't recognize that this structure's scope is in the `catcls_vclass_install()` function. yes, It seems no impact.

hgryoo (approved):

src/transaction/boot_cl.c
hgryoo: Because of `// *INDENT OFF*` hint, formatting seems broken.
```suggestion
    {boot_define_view_class},  /* CTV_CLASS_NAME */
    {boot_define_view_super_class}, /* CTV_SUPER_CLASS_NAME */
    {boot_define_view_vclass}, /* CTV_VCLASS_NAME */
    {boot_define_view_attribute}, /* CTV_ATTRIBUTE_NAME */
    {boot_define_view_attribute_set_domain}, /* CTV_ATTR_SD_NAME */
    {boot_define_view_method}, /* CTV_METHOD_NAME */
    {boot_define_view_method_argument}, /* CTV_METHARG_NAME */
    {boot_define_view_method_argument_set_domain}, /* CTV_METHARG_SD_NAME */
    {boot_define_view_method_file}, /* CTV_METHFILE_NAME */
    {boot_define_view_index}, /* CTV_INDEX_NAME */
    {boot_define_view_index_key}, /* CTV_INDEXKEY_NAME */
    {boot_define_view_authorization}, /* CTV_AUTH_NAME */
    {boot_define_view_trigger}, /* CTV_TRIGGER_NAME */
    {boot_define_view_partition}, /* CTV_PARTITION_NAME */
    {boot_define_view_stored_procedure}, /* CTV_STORED_PROC_NAME */
    {boot_define_view_stored_procedure_arguments}, /* CTV_STORED_PROC_ARGS_NAME */
    {boot_define_view_db_collation}, /* CTV_DB_COLLATION_NAME */
    {boot_define_view_db_charset}, /* CTV_DB_CHARSET_NAME */
    {boot_define_view_db_server} /* CTV_DB_SERVER_NAME */
```

ctshim: thank you, I will clear white space.

beyondykk9 (approved):

src/transaction/boot_cl.c
```diff
@@ -1963,6 +1964,12 @@ boot_define_class (MOP class_mop)
       return error_code;
     }
 
+  error_code = smt_add_attribute (def, "conn_server", CT_DB_SERVER_NAME, NULL);
```
youngjinj: I think we need more discussion about adding that column.

We would appreciate it if we could discuss it through another PR after sufficient discussion about whether it is necessary to modify _db_class and db_class while adding _db_server and db_server.

ctshim: Yes, we will cancel the modification of the _db_class catalog and decide whether to reflect it after further discussion.

PR #2844, opened 2021-06-14
[CBRD-23976] Added CREATE SERVER/DROP SERVER syntax

ctshim:
http://jira.cubrid.org/browse/CBRD-23976

Purpose
  Add CREATE SERVER/DROP SERVER syntax.

Implementation

-    Add New Syntax

```
            <create_server> ::=  CREATE SERVER <server_name> '(' <connect_info>  ')'
            <server_name> ::= identifier        
            <connect_info> ::=  <connect_info> ',' <connect_item>
            <connect_item> ::=
                        HOST '=' identifier 
                      | HOST '=' IPV4_ADDRESS 
                      | PORT '=' UNSIGNED_INTEGER 
                      | DBNAME '=' identifier 
                      | USER '=' identifier 
                      | PASSWORD '=' [ CHAR_STRING ] 
                      | PROPERTIES '=' [ CHAR_STRING ]
                      | COMMENT '=' [ char_string ]

           <drop_server> ::= DROP SERVER [IF EXISTS ]  <server_name>

```

- Add some error code & message
```
#define ER_DBLINK_DB_SERVER_NOT_FOUND            -1286
#define ER_DBLINK_SERVER_NOT_FOUND               -1287 
#define ER_DBLINK_SERVER_ALREADY_EXIST           -1288
#define ER_DBLINK_CANNOT_UPDATE_SERVER              -1289

#define MSGCAT_SEMANTIC_SERVER_ALREADY_EXIST            MSGCAT_SEMANTIC_NO(306) 
#define MSGCAT_RUNTIME_RT_SERVER_NOT_DEFINED            MSGCAT_RUNTIME_NO(83) 
#define MSGCAT_RUNTIME_RT_SERVER_ALTER_NOT_ALLOWED      MSGCAT_RUNTIME_NO(84)  

./msg/en_US/cubrid.msg
```

Remarks
N/A


src/base/error_code.h
```diff
@@ -1645,8 +1645,11 @@
 
 #define ER_DBLINK		    		      -1284
 #define ER_DBLINK_INVALID_COLUMNS_SPECIFIED         -1285
+#define ER_QPROC_DB_SERVER_NOT_FOUND		    -1286
```
beyondykk9:
```suggestion
#define ER_DBLINK_CATALOG_DB_SERVER_NOT_FOUND		    -1286
```

src/base/error_code.h
```diff
@@ -1645,8 +1645,11 @@
...
+#define ER_QPROC_SERVER_NOT_FOUND                   -1287
```
beyondykk9:
```suggestion
#define ER_DBLINK_SERVER_NOT_FOUND                   -1287
```

src/base/error_code.h
```diff
@@ -1645,8 +1645,11 @@
...
+#define ER_QPROC_SERVER_ALREADY_EXIST               -1288
```
beyondykk9:
```suggestion
#define ER_DBLINK_SERVER_ALREADY_EXISTS               -1288
```

src/query/execute_statement.c
```diff
@@ -17724,3 +17758,361 @@ do_insert_checks (PARSER_CONTEXT * parser, PT_NODE * statement, PT_NODE ** class
...
+  int error = NO_ERROR;
+  int save;
+  bool au_disable_flag = false;
+
+  PT_CREATE_SERVER_INFO *si = &statement->info.create_server;
+
+  CHECK_MODIFICATION_ERROR ();
+
+  db_make_null (&port_no);
+  db_make_int (&port_no, 0);
```
beyondykk9, line 17929: Why don't you assign the type of **port_no** to "unsigned short" or "unsigned integer" ?

msg/en_US/cubrid.msg
```diff
@@ -1356,8 +1356,11 @@ Check the path of the key file (_keys) and make sure that it includes proper key
 
 1284 dblink - %1$s
 1285 dblink - invalid number of columns specified.
+1286 _db_server class not found.
```
beyondykk9:
```suggestion
1286 dblink - catalog _db_server class not found.
```

msg/en_US/cubrid.msg
```diff
@@ -1356,8 +1356,11 @@ Check the path of the key file (_keys) and make sure that it includes proper key
...
+1287 Server "%1$s" not found.
```
beyondykk9:
```suggestion
1287 dblink - server "%1$s" not found.
```

msg/en_US/cubrid.msg
```diff
@@ -1356,8 +1356,11 @@ Check the path of the key file (_keys) and make sure that it includes proper key
...
+1288 Server "%1$s" already exist.
```
beyondykk9:
```suggestion
1288 dblink - server "%1$s" already exists.
```

msg/en_US/cubrid.msg
```diff
@@ -1822,6 +1825,7 @@ $set 8 MSGCAT_SET_PARSER_SEMANTIC
 303 Unexpected number of arguments %1$d.
 304 Could not find compatible signature for function %1$s()
 305 Signature %1$s is not compatible with function arguments.
+306 Server '%1$s' already exist.
```
beyondykk9:
```suggestion
306 The server '%1$s' already exists.
```

src/parser/parse_tree_cl.c
```diff
@@ -18384,3 +18400,102 @@ pt_print_dblink_table (PARSER_CONTEXT * parser, PT_NODE * p)
...
+  /*
+     PT_NODE *server_name;
+     PT_NODE *host;
+     PT_NODE *port;  
+     PT_NODE *dbname;  
+     PT_NODE *user;
+     PT_NODE *pwd;
+     PT_NODE *prop;
+     PT_NODE *comment;
+   */
```
beyondykk9: I don't understand what's the meaning of this comment.

src/parser/parse_tree_cl.c
```diff
@@ -18384,3 +18400,102 @@ pt_print_dblink_table (PARSER_CONTEXT * parser, PT_NODE * p)
...
+    }
+
+  if (si->prop != NULL)
+    {
+      q = pt_append_nulstring (parser, q, ", PROPERTIES=");
+      r = pt_print_bytes (parser, si->prop);
+      q = pt_append_varchar (parser, q, r);
+    }
+#if 0
+  if (si->comment != NULL)
```
beyondykk9, line 18456: Why is supposed to "#if 0" ?

src/parser/parser_message.h
```diff
@@ -473,6 +473,7 @@
 #define MSGCAT_SEMANTIC_WRONG_ARGS_COUNT                MSGCAT_SEMANTIC_NO(303)
 #define MSGCAT_SEMANTIC_NO_VALID_FUNCTION_SIGNATURE     MSGCAT_SEMANTIC_NO(304)
 #define MSGCAT_SEMANTIC_INCOMPATIBLE_SIGNATURE          MSGCAT_SEMANTIC_NO(305)
+#define MSGCAT_SEMANTIC_SERVER_ALREADY_EXIST            MSGCAT_SEMANTIC_NO(306)
```
beyondykk9:
```suggestion
#define MSGCAT_SEMANTIC_SERVER_ALREADY_EXISTS            MSGCAT_SEMANTIC_NO(306)
```

msg/en_US/cubrid.msg
```diff
@@ -1905,6 +1909,8 @@ $set 9 MSGCAT_SET_PARSER_RUNTIME
 80 Isolation level value in MVCC must be 'read committed', 'repeatable read' or 'serializable'.
 81 '%1$s' is an 'auto_increment' object, which is not allowed to be altered as a serial.
 82 Invalid JSON: '%1$s'.
+83 The server '%1$s' is not defined yet.
```
beyondykk9:
```suggestion
83 The server '%1$s' is not defined.
```

src/parser/csql_grammar.y
```diff
@@ -4195,6 +4263,20 @@ drop_stmt
 			PARSER_SAVE_ERR_CONTEXT ($$, @$.buffer_pos)
 
 		DBG_PRINT}}
+	| DROP SERVER opt_if_exists identifier
+		{{
+                        PT_NODE *node = parser_new_node (this_parser, PT_DROP_SERVER);
+
+			if (node)
+			  {
+                            node->info.serial.if_exists = $3;
```
beyondykk9: serial ?
it looks confused with "drop_server".

msg/en_US/cubrid.msg
ctshim: I'll fix it as per your suggestion.

msg/en_US/cubrid.msg
ctshim: I'll fix it as per your suggestion.

msg/en_US/cubrid.msg
ctshim: I'll fix it as per your suggestion.

msg/en_US/cubrid.msg
ctshim: I'll fix it as per your suggestion.

msg/en_US/cubrid.msg
ctshim: I'll fix it as per your suggestion.

src/parser/parse_tree_cl.c
ctshim: I wasn't sure there was a need to print the comments.

src/parser/parser_message.h
ctshim: I'll fix it as per your suggestion.

src/base/error_code.h
ctshim: I'll fix it as per your suggestion.

src/base/error_code.h
ctshim: I'll fix it as per your suggestion.

src/base/error_code.h
ctshim: I'll fix it as per your suggestion.

src/parser/csql_grammar.y
ctshim: There was a mistake in the copy & paste process.
Thanks for finding the bug.

src/query/execute_statement.c
ctshim: The type of the port column of _db_server is INTEGER.
Actually, it will not have a value beyond the range of the port number because valid value check is carried out.

src/parser/parse_tree_cl.c
ctshim: It is a list of variables belonging to create_server, and it was written for reference.
I will delete it.

src/query/execute_statement.c
```diff
@@ -17724,3 +17758,361 @@ do_insert_checks (PARSER_CONTEXT * parser, PT_NODE * statement, PT_NODE ** class
...
+	  return NO_ERROR;
+	}
+      error = ER_DBLINK_SERVER_NOT_FOUND;
+      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, error, 1, name);
+      PT_ERRORmf (parser, statement, MSGCAT_SET_PARSER_RUNTIME, MSGCAT_RUNTIME_RT_SERVER_NOT_DEFINED, name);
+      goto end;
+    }
+
+  /* check if user is creator or DBA  */
+  error = au_check_serial_authorization (server_object);
```
beyondykk9: looks copy and paste fault

src/parser/parse_tree_cl.c
ctshim: The comment content has been modified to be printed as well.

src/query/execute_statement.c
ctshim: Regardless of the function name, what is being processed is correct.
However, there is a risk of confusion, so I will separate the function name and rewrite it.

src/query/execute_statement.c
ctshim: In accordance with your seperated  comments, we have simplified the processing parameters in the function.

src/object/authenticate.c
```diff
@@ -8656,20 +8656,20 @@ au_get_dba_user (void)
...
+ * au_check_row_authorization - check whether the current user is able to
+ *                                 modify row object or not
  *   return: NO_ERROR if available, otherwise error code
- *   serial_object(in): serial object pointer
+ *   row_object(in): row object pointer
  */
-int
-au_check_serial_authorization (MOP serial_object)
+static int
+au_check_row_authorization (MOP row_object)
```
beyondykk9: how about **au_check_user_authorization** ?

ctshim: This is a function to check if I have permission to access a specific row and process it.
Whether the emphasis is on row or user.
If it is a user, I feel like I have to check it while passing the user information.

I'm trying to keep the current name unless it's a particularly sensitive issue, is that ok?

beyondykk9: I think, in common, the word "row" is not used as a function name. It would be better to define it to have a more clear meaning. If you can't find a better name, keep the name.

ctshim: How about au_check_ownership_authorization()?

beyondykk9: **au_check_ownership_authorization** is not bad. and how about **au_check_owner** more shortly ?

ctshim: OK, then I'll change it to au_check_owner.

shparkcubrid: I think `au_check_user_authorization `or `au_check_authorization` are good.

src/object/authenticate.c
```diff
@@ -8656,20 +8656,20 @@ au_get_dba_user (void)
...
   DB_VALUE creator_val;
   MOP creator;
   DB_SET *groups;
   int ret_val;
 
-  ret_val = db_get (serial_object, "owner", &creator_val);
+  ret_val = db_get (row_object, "owner", &creator_val);
 
   if (ret_val != NO_ERROR || DB_IS_NULL (&creator_val))
     {
```
shparkcubrid: `ret_val = ER_QPROC_CANNOT_UPDATE_SERIAL;`
It should be modified. I think separating functions would be nice.

ctshim: Thanks for the thorough check.
Let's fix the error code to be distinct.

src/query/execute_statement.c
```diff
@@ -17724,3 +17758,305 @@ do_insert_checks (PARSER_CONTEXT * parser, PT_NODE * statement, PT_NODE ** class
...
+{
+  return do_get_obj_id (server_obj_id, server_class_mop, server_name, "link_name");
+}
+
+int
+do_drop_server (PARSER_CONTEXT * parser, PT_NODE * statement)
+{
+  DB_OBJECT *server_class = NULL, *server_object = NULL;
+  DB_IDENTIFIER server_obj_id;
+  DB_VALUE class_name_val;
```
shparkcubrid: I think `class_name_val` can be removed.

ctshim: I agree. I will remove it.

beyondykk9 (approved):

shparkcubrid (approved):

src/object/authenticate.c
```diff
@@ -8656,20 +8656,20 @@ au_get_dba_user (void)
...
- *                                 modify serial object or not
+ * au_check_owner - check whether the current user is able to
+ *                                 modify row object or not
  *   return: NO_ERROR if available, otherwise error code
- *   serial_object(in): serial object pointer
+ *   record_object(in): row object pointer
  */
 int
-au_check_serial_authorization (MOP serial_object)
+au_check_owner (MOP record_object, int err_code)
```
hgryoo: Instead of adding err_code parameter, how about call this function like the following pattern?
```
int
au_check_serial_authorization (MOP record_object)
{
  if (au_check_owner (record_object) != NO_ERROR)
  {
    return ER_QPROC_CANNOT_UPDATE_SERIAL;
  }
  else
  {
    return NO_ERROR;
  }
}
```

hgryoo: For checking server authorization, the same pattern could be applied.

ctshim: If the return value of the au_check_owner() function is not NO_ERROR, and it is an error value that occurred in db_get(), it must be returned as it is.

Instead, if you add a wrapper function to handle it, it is possible without a parameter for the default error code.
Instead, the function call step will be incremented by one.
Even so, would it feel easy?

ctshim: Let's split the au_check_owner function into two.

hgryoo (approved):

youngjinj (approved):

PR #2847, opened 2021-06-15
[LETS-34] Deactivate record analysis/undo/finish postpone in page server recovery

ribram:

Recovery on the page server must only apply unflushed change in the data pages. It must not generate any log records and it does not have to recovery the state of transaction table. As consequence, the next operations are disabled:

  - Almost all log record analysis. Log record analysis is usually run to rebuild the transaction table state before the server died or to find the end point for partial recovery. The only analysis required on the page server is on the last log record (end of log) to reset append LSA.
  - No undo of system operations, no finishing postpones and no undo of transactions.

Another necessary change is to disable replication verification in logpb_checkpoint, if it is run during recovery. The replicator is not initialized yet.


src/transaction/log_recovery_context.cpp
```diff
@@ -116,3 +116,9 @@ log_recovery_context::get_end_redo_lsa () const
 {
   return m_end_redo_lsa;
 }
+
+bool
+log_recovery_context::is_page_server () const
+{
+  return m_is_page_server;
```
cristiarg, line 123: `m_is_page_server` can be `const`'ed.

src/transaction/log_recovery_analysis.cpp
```diff
@@ -404,11 +404,23 @@ log_recovery_analysis (THREAD_ENTRY *thread_p, INT64 *num_redo_log_records, log_
...
+	  if (context.is_restore_incomplete ())
+	    {
+	      break;
+	    }
+	}
+      else
+	{
+	  if (log_rtype == LOG_END_OF_LOG)
+	    {
+	      log_rv_analysis_log_end (tran_id, &log_nav_lsa);
```
cristiarg: Can analysis bail out after this? Or is this the last entry in the log record anyway?

src/transaction/log_recovery_analysis.cpp
```diff
@@ -404,11 +404,23 @@ log_recovery_analysis (THREAD_ENTRY *thread_p, INT64 *num_redo_log_records, log_
 	  context.set_start_redo_lsa (start_redo_lsa);
 	}
 
-      log_rv_analysis_record (thread_p, log_rtype, tran_id, &log_nav_lsa, log_page_p, &prev_lsa, context);
-      if (context.is_restore_incomplete ())
+      if (!context.is_page_server ())
```
cristiarg: Extract a new - forking function - `log_rv_analysis_record_fork_tran_page` - that does this bit? 
The old forking function can be renaned to `..._tran`. The new page server specific bit - `..._page`.
.. just a thought.

cristiarg (approved):

src/transaction/log_recovery_context.cpp
ribram: its value is initialized during construction.

```
log_recovery_context.cpp(26,3): error C2166: l-value specifies const object
```

src/transaction/log_recovery_analysis.cpp
ribram: It is the last anyway.

src/transaction/log_recovery_analysis.cpp
ribram: Done. Forked `log_rv_analysis_record` into `log_rv_analysis_record_on_page` and `log_rv_analysis_record_on_tran`.

src/transaction/log_recovery_context.cpp
cristiarg: Yes, I knew. Should have been moved to initializer list after `const`. :)

PR #2851, opened 2021-06-17
[CBRD-23984] Recreate a heap file with DONT_REUSE_OID while TRUNCATE

wrlawodms:
http://jira.cubrid.org/browse/CBRD-23984

### DIFF
--- a/src/communication/network_interface_cl.h
+++ b/src/communication/network_interface_cl.h
@@ -119,7 +119,7 @@ extern int heap_create (HFID * hfid, const OID * class_oid, bool reuse_oid);
 #if defined(ENABLE_UNUSED_FUNCTION)
 extern int heap_destroy (const HFID * hfid);
 #endif
-extern int heap_destroy_newly_created (const HFID * hfid, const OID * class_oid, const int force = 0);
+extern int heap_destroy_newly_created (const HFID * hfid, const OID * class_oid, const bool force = false);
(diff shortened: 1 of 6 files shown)


src/communication/network_interface_cl.h
```diff
@@ -119,7 +119,7 @@ extern int heap_create (HFID * hfid, const OID * class_oid, bool reuse_oid);
 #if defined(ENABLE_UNUSED_FUNCTION)
 extern int heap_destroy (const HFID * hfid);
 #endif
-extern int heap_destroy_newly_created (const HFID * hfid, const OID * class_oid);
+extern int heap_destroy_newly_created (const HFID * hfid, const OID * class_oid, const int force = 0);
```
hornetmj: how about using bool type instead of int?

hornetmj (approved):

src/communication/network_interface_cl.h
wrlawodms: At first, I did as you said. However, this is a network interface function, and that interfaces use int for a boolean type. It seems because there is no `or_pack_bool()` function. It is for code consistency.

wrlawodms: Oh, I've found out `heap_create` is the same sort of function, but it uses boolean as a parameter. It turns out case by case. I'll follow your advice.

hornetmj (approved):

PR #2852, opened 2021-06-17
[LETS-118]alterdbhost does not mount/dismount active log

adrianbrab:
http://jira.cubrid.org/browse/LETS-118

alterdbhost changes the hostname set in databases.txt.

The implementation also mounts/dismounts the active log file, probably as a check that the database really exists. It is not mandatory for the mission of alterdbhost utilitary and it won't work with transaction servers running on remote storage (and it won't work with any transaction servers in the end).

Changed the code for alterdbhost to not mount/unmount the log active file.
The functionality of alterdbhost should not be altered by the change.

### DIFF
--- a/src/executables/util_sa.c
+++ b/src/executables/util_sa.c
@@ -1899,12 +1899,10 @@ alterdbhost (UTIL_FUNCTION_ARG * arg)
   const char *db_name;
   const char *host_name;
   int dbtxt_vdes = NULL_VOLDES;
-  int log_vdes = NULL_VOLDES;
   char dbtxt_label[PATH_MAX];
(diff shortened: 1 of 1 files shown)


src/executables/util_sa.c
```diff
@@ -1967,40 +1965,19 @@ alterdbhost (UTIL_FUNCTION_ARG * arg)
 	}
     }
   db = cfg_find_db_list (dir, db_name);
+
   if (db == NULL)
```
ribram:
```suggestion
  if (db == NULL)
```

ribram (approved):

src/executables/util_sa.c
adrianbrab: Removed the empty line.

PR #2853, opened 2021-06-18
[CBRD-23985] Don’t load indexes during recreating index while TRUNCATE

wrlawodms:
http://jira.cubrid.org/browse/CBRD-23985

### DIFF
--- a/src/object/class_object.h
+++ b/src/object/class_object.h
@@ -773,7 +773,7 @@ struct sm_class
   unsigned has_active_triggers:1;	/* set if trigger processing is required */
   unsigned dont_decache_constraints_or_flush:1;	/* prevent decaching class constraint and flushing. */
   unsigned recache_constraints:1;	/* class constraints need recache. */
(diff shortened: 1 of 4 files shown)


src/object/class_object.h
```diff
@@ -773,6 +773,7 @@ struct sm_class
   unsigned has_active_triggers:1;	/* set if trigger processing is required */
   unsigned dont_decache_constraints_or_flush:1;	/* prevent decaching class constraint and flushing. */
   unsigned recache_constraints:1;	/* class constraints need recache. */
+  unsigned dont_load_index_from_heap:1;	/* don't load index, but create a new empty one when allocating an index. e.g. TRUNCATE */
```
hornetmj: How about using load_index_from_heap instead of dont_load_index_from_heap?

hornetmj (approved):

src/object/class_object.h
wrlawodms: I agree that the negative name could be confusing, but `load_index_from_heap=true` sounds like it always loads index from a heap. How about "force_empty_index_when_alloc"?
Or do you have any idea?

PR #2854, opened 2021-06-23
[CBRD-23990] Redesign of query cache management to handle the plan cache overflown

beyondykk9:
http://jira.cubrid.org/browse/CBRD-23990

A query cache is linked to an XASL cache entry by hash table number (list_ht_no of XASL_CACHE_ENTRY).

The qfile_List_cache has hash tables for query and result cache management. The number of hash tables is the same as max number of query plan (XASL_CACHE_MAX_ENTRIES). To minimize use of mutex, whenever allocation of xasl cache entry the list_ht_no is set to the number of current entries linked to qfile_List_cache's hash table index. See the routine xcache_entry_alloc.

But the xasl cache entries can be overflown to XASL_CACHE_MAX_ENTRIES, the list_ht_no of the entry can be out of range because the qfile_List_cache's hash table entries are fixed to XASL_CACHE_MAX_ENTRIES + 10.

I modified the routine xcache_entry_alloc not to assign the number of current entries. The list_ht_no is from qcache_get_new_ht_no ( ) instead. And the qfile_List_cache's structure is modified to get free hash table index (qcache_get_ht_no).

free_list and next field are newly added. qcache_get_new_ht_no and qcache_free_ht_no are also added for allocation and free for hash table index.

```
-- cubrid.conf
max_query_cache_entries = 2
query_cache_size_in_pages = 10000
max_plan_cache_entries = 2
max_plan_cache_clones = 2
 
create table tmp (col1 int, col2 int, col3 int);
select /*+ query_cache recompile */ * from tmp a, tmp b, tmp c where a.col1 = b.col1 and b.col1 = c.col1;
select /*+ query_cache recompile */ * from tmp a, tmp b where a.col1 = b.col1 and a.col1 = 1;
select /*+ query_cache recompile */ * from tmp a where col1 = 1;
```

the above execution should be successfully done.

### DIFF
--- a/src/query/list_file.c
+++ b/src/query/list_file.c
@@ -5081,7 +5081,8 @@ qfile_finalize_list_cache (THREAD_ENTRY * thread_p)
       for (i = 0; i < qfile_List_cache.n_hts; i++)
 	{
 	  bool invalidate = true;
-	  (void) mht_map_no_key (thread_p, qfile_List_cache.list_hts[i], qfile_end_use_of_list_cache_entry_local, &invalidate);
+	  (void) mht_map_no_key (thread_p, qfile_List_cache.list_hts[i], qfile_end_use_of_list_cache_entry_local,
+				 &invalidate);
 	  mht_destroy (qfile_List_cache.list_hts[i]);
 	}
       free_and_init (qfile_List_cache.list_hts);
--- a/src/query/list_file.c
+++ b/src/query/list_file.c
@@ -5094,19 +5094,16 @@ qfile_clear_list_cache (THREAD_ENTRY * thread_p, XASL_CACHE_ENTRY * xcache_entry
   if (csect_enter (thread_p, CSECT_QPROC_LIST_CACHE, INF_WAIT) != NO_ERROR)
     {
       goto end;
-      //return ER_FAILED;
     }
 
   if (QFILE_IS_LIST_CACHE_DISABLED || xcache_entry->list_ht_no < 0)
     {
       goto end;
-      //return ER_FAILED;
     }
 
   if (qfile_List_cache.n_hts == 0)
     {
       goto end;
-      //return ER_FAILED;
     }
 
   list_ht_no = xcache_entry->list_ht_no;
--- a/src/query/list_file.c
+++ b/src/query/list_file.c
@@ -5568,7 +5568,10 @@ qfile_lookup_list_cache_entry (THREAD_ENTRY * thread_p, XASL_CACHE_ENTRY * xasl,
 
   if (xasl->list_ht_no < 0)
     {
-      xasl->list_ht_no = qcache_get_new_ht_no (thread_p);
+      if ((xasl->list_ht_no = qcache_get_new_ht_no (thread_p)) < 0)
+	{
+	  goto end;
+	}
     }
 
   tran_index = LOG_FIND_THREAD_TRAN_INDEX (thread_p);
--- a/src/query/list_file.c
+++ b/src/query/list_file.c
@@ -96,6 +96,8 @@ typedef struct qfile_list_cache QFILE_LIST_CACHE;
 struct qfile_list_cache
 {
   MHT_TABLE **list_hts;		/* array of memory hash tables for list cache; pool for list_ht of XASL_CACHE_ENTRY */
+  int *free_list;		/* array of freed hash tables */
+  int next;			/* the next freed hash table number */
   unsigned int n_hts;		/* number of elements of list_hts */
   int n_entries;		/* total number of cache entries */
   int n_pages;			/* total number of pages used by the cache */
@@ -158,7 +160,7 @@ struct qfile_list_cache_entry_pool
  */
 
 /* list cache and related information */
-static QFILE_LIST_CACHE qfile_List_cache = { NULL, 0, 0, 0, 0, 0, 0, 0 };
+static QFILE_LIST_CACHE qfile_List_cache = { NULL, NULL, 0, 0, 0, 0, 0, 0, 0, 0 };
 
 /* information of candidates to be removed from XASL cache */
 static QFILE_LIST_CACHE_CANDIDATE qfile_List_cache_candidate = { 0, 0, 0, 0, NULL, NULL, NULL, 0, 0, false };
@@ -391,6 +393,55 @@ qfile_list_cache_cleanup (THREAD_ENTRY * thread_p)
   return NO_ERROR;
 }
 #endif
+
+int
+qcache_get_new_ht_no (THREAD_ENTRY * thread_p)
+{
+  int ht_no = -1;
+
+  if (QFILE_IS_LIST_CACHE_DISABLED)
+    {
+      return NO_ERROR;
+    }
+
+  if (csect_enter (thread_p, CSECT_QPROC_LIST_CACHE, INF_WAIT) != NO_ERROR)
+    {
+      return ER_FAILED;
+    }
+
+  if (qfile_List_cache.next >= 0)
+    {
+      ht_no = qfile_List_cache.next;
+      qfile_List_cache.next = qfile_List_cache.free_list[qfile_List_cache.next];
+    }
+
+  csect_exit (thread_p, CSECT_QPROC_LIST_CACHE);
+
+  return ht_no;
+}
+
+void
+qcache_free_ht_no (THREAD_ENTRY * thread_p, int ht_no)
+{
+  if (QFILE_IS_LIST_CACHE_DISABLED)
+    {
+      return;
+    }
+
+  if (csect_enter (thread_p, CSECT_QPROC_LIST_CACHE, INF_WAIT) != NO_ERROR)
+    {
+      return;
+    }
+
+  if (ht_no >= 0)
+    {
+      (void) mht_clear (qfile_List_cache.list_hts[ht_no], NULL, NULL);
+      qfile_List_cache.free_list[ht_no] = qfile_List_cache.next;
+      qfile_List_cache.next = ht_no;
+    }
+
+  csect_exit (thread_p, CSECT_QPROC_LIST_CACHE);
+}
 
 /* qfile_modify_type_list () -
  *   return:
@@ -4922,8 +4973,9 @@ qfile_initialize_list_cache (THREAD_ENTRY * thread_p)
   else
     {
       /* create */
-      qfile_List_cache.n_hts = prm_get_integer_value (PRM_ID_XASL_CACHE_MAX_ENTRIES) + 10;
+      qfile_List_cache.n_hts = prm_get_integer_value (PRM_ID_XASL_CACHE_MAX_ENTRIES);
       qfile_List_cache.list_hts = (MHT_TABLE **) calloc (qfile_List_cache.n_hts, sizeof (MHT_TABLE *));
+      qfile_List_cache.free_list = (int *) calloc (qfile_List_cache.n_hts, sizeof (int));
       if (qfile_List_cache.list_hts == NULL)
 	{
 	  goto error;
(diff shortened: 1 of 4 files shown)


src/query/list_file.c
```diff
@@ -392,6 +394,55 @@ qfile_list_cache_cleanup (THREAD_ENTRY * thread_p)
...
+
+  csect_exit (thread_p, CSECT_QPROC_LIST_CACHE);
+
+  return ht_no;
+}
+
+void
+qcache_free_ht_no (THREAD_ENTRY * thread_p, int ht_no)
+{
+  if (QFILE_IS_LIST_CACHE_DISABLED)
```
ctshim: If you modify the first if as below, you don't need to go into the critical section and validate ht_no.

`if (QFILE_IS_LIST_CACHE_DISABLED || ht_no < 0)`

src/query/xasl_cache.c
```diff
@@ -1718,7 +1719,10 @@ xcache_invalidate_qcaches (THREAD_ENTRY * thread_p, const OID * oid)
 	      finished = true;
 	      break;
 	    }
-
+	  if (xcache_entry->list_ht_no < 0)
```
ctshim, line 1721: Does it make sense to call xcache_Hashmap.erase() even if (xcache_entry->list_ht_no < 0)?

src/query/list_file.c
```diff
@@ -4924,6 +4975,7 @@ qfile_initialize_list_cache (THREAD_ENTRY * thread_p)
       /* create */
       qfile_List_cache.n_hts = prm_get_integer_value (PRM_ID_XASL_CACHE_MAX_ENTRIES) + 10;
       qfile_List_cache.list_hts = (MHT_TABLE **) calloc (qfile_List_cache.n_hts, sizeof (MHT_TABLE *));
+      qfile_List_cache.free_list = (int *) calloc (qfile_List_cache.n_hts, sizeof (int));
```
ctshim: When do you free() the free_list allocated here? 
I couldn't find that code.

beyondykk9: I missed the codes to free at the finalize routine. I fixed. Thanks.

src/query/xasl_cache.c
beyondykk9: xcache_entry->list_ht_no = -1, it means that the query cache is not available. And xcache_Hashmap.erase() is never called in this routine. I think of you confuse the routine xcache_invalidate_entries.

src/query/list_file.c
beyondykk9: This routine is always called with "ht_no >= 0" so it's not necessary.

ctshim: yes, then a few lines down
The "if (ht_no >= 0)" statement is unnecessary.

src/query/xasl_cache.c
ctshim: I was confused.

src/query/list_file.c
```diff
@@ -96,6 +96,8 @@ typedef struct qfile_list_cache QFILE_LIST_CACHE;
 struct qfile_list_cache
 {
   MHT_TABLE **list_hts;		/* array of memory hash tables for list cache; pool for list_ht of XASL_CACHE_ENTRY */
+  int *free_list;		/* array of freed hash tables */
+  int next;			/* the next freed hash table number */
```
hgryoo: It seems queue structure. how about using `std::deque`?
```
#include <deque>

std::deque<int> free_ht_idx;

// insert
free_ht_idx.embrace_front (ht_no);

// get value
free_ht_idx.front();
free_ht_idx.back();

// remove
free_ht_idx.pop_front(); // like stack
free_ht_idx.pop_back(); // like queue
```

Since thread safety is guaranteed by the critical section, it can be safely used.

shparkcubrid: I recommend below for readability.
```
int *free_ht_list;
int next_ht_id;
```

src/query/list_file.c
```diff
@@ -392,6 +394,55 @@ qfile_list_cache_cleanup (THREAD_ENTRY * thread_p)
...
+    {
+      return NO_ERROR;
+    }
+
+  if (csect_enter (thread_p, CSECT_QPROC_LIST_CACHE, INF_WAIT) != NO_ERROR)
+    {
+      return ER_FAILED;
+    }
+
+  if (qfile_List_cache.next >= 0)
```
shparkcubrid: It looks good to write 'if (qfile_List_cache.next >= 0)' before csect_enter() .

src/query/list_file.c
```diff
@@ -392,6 +394,55 @@ qfile_list_cache_cleanup (THREAD_ENTRY * thread_p)
...
+    {
+      return;
+    }
+
+  if (csect_enter (thread_p, CSECT_QPROC_LIST_CACHE, INF_WAIT) != NO_ERROR)
+    {
+      return;
+    }
+
+  if (ht_no >= 0)
```
shparkcubrid: It looks good to write 'if (ht_no >= 0)' before csect_enter() .

src/query/list_file.c
shparkcubrid: add free() in this function where error:

src/query/query_manager.c
```diff
@@ -1391,14 +1391,23 @@ xqmgr_execute_query (THREAD_ENTRY * thread_p, const XASL_ID * xasl_id_p, QUERY_I
 
       if (do_not_cache == false)
 	{
-	  /* lookup the list cache with the parameter values (DB_VALUE array) */
-	  list_cache_entry_p =
-	    qfile_lookup_list_cache_entry (thread_p, xasl_cache_entry_p->list_ht_no, &params, &cached_result);
-	  /* If we've got the cached result, return it. */
-	  if (cached_result)
+	  /* lookup the hash table number for qfile list cache */
+	  if (xasl_cache_entry_p->list_ht_no < 0)
```
shparkcubrid: When ht_no is newly allocated, it is not necessary to call qfile_lookup_list_cache_entry(). This code can be moved back several lines.

beyondykk9: very good point. thanks.

src/query/list_file.c
beyondykk9: Ok, I fixed it.

src/query/list_file.c
beyondykk9: It should be protected because the **qfile_List_cache.next** is shared variable through whole threads

src/query/list_file.c
beyondykk9: your suggestion is also good, but the current routine is also so simple. I'd like to keep this codes.

src/query/list_file.c
beyondykk9: Ok, I fixed it.

src/query/list_file.c
beyondykk9: Ok, I agree your opinion. I changed the name to "free_ht_list" and I renamed the "next" to "next_ht_no"

src/query/list_file.c
beyondykk9: Ok, I fixed it.

src/query/list_file.c
```diff
@@ -392,6 +394,40 @@ qfile_list_cache_cleanup (THREAD_ENTRY * thread_p)
...
+      qfile_List_cache.next_ht_no = qfile_List_cache.free_ht_list[qfile_List_cache.next_ht_no];
+    }
+
+  return ht_no;
+}
+
+void
+qcache_free_ht_no (THREAD_ENTRY * thread_p, int ht_no)
+{
+  if (QFILE_IS_LIST_CACHE_DISABLED || ht_no < 0)
```
beyondykk9: this free routine is called from only qfile_end_use_of_list_cache_entry, so critical section is not necessary. I'm going to remove the critical section.

shparkcubrid:
There is case where ht_no is not free.
```
--cubrid.conf
max_query_cache_entries = 3
query_cache_size_in_pages = 10000
max_plan_cache_entries = 3
max_plan_cache_clones = 3

drop table tmp;
create table tmp (col1 int, col2 int, col3 int);
insert into tmp values(2,1,2);

select /*+ query_cache */ count(*) from tmp a, tmp b where a.col1 = 1 and b.col1 = 1 and b.col2 = 1;   --1932
select /*+ query_cache */ count(*) from tmp a, tmp b where a.col1 = b.col1 and a.col1 = 1;
select /*+ query_cache */ count(*) from tmp a where col1 = 1;

;info qcache

update tmp set col1 =2;
;info qcache

select /*+ query_cache */ count(*) from tmp a, tmp b where a.col1 = 1 and b.col1 = 1 and b.col2 = 1;   --1932
select /*+ query_cache */ count(*) from tmp a, tmp b where a.col1 = b.col1 and a.col1 = 1;
select /*+ query_cache */ count(*) from tmp a where col1 = 1;

;info qcache
```

beyondykk9:
...
> select /*+ query_cache */ count(*) from tmp a, tmp b where a.col1 = 1 and b.col1 = 1 and b.col2 = 1;   --1932
> select /*+ query_cache */ count(*) from tmp a, tmp b where a.col1 = b.col1 and a.col1 = 1;
> select /*+ query_cache */ count(*) from tmp a where col1 = 1;
> 
> ;info qcache
> ```

this bug is fixed by commit # 071e7a7

src/query/list_file.c
```diff
@@ -363,20 +366,31 @@ qfile_list_cache_cleanup (THREAD_ENTRY * thread_p)
...
-  for (candidate_index = 0; candidate_index < bh->element_count; candidate_index++)
+  /* traverse in reverse for weight ordering, from light weight to heavy weight */
+  for (candidate_index = bh->element_count - 1; candidate_index >= 0; candidate_index--)
     {
       bh_element_at (bh, candidate_index, &candidate);
+
+      invalidate = candidate.qcache->invalidate;
+      ht_no = candidate.qcache->list_ht_no;
       qfile_delete_list_cache_entry (thread_p, candidate.qcache);
+      if (invalidate && qfile_get_list_cache_number_of_entries (ht_no) == 0)
```
shparkcubrid: Deletion of qcache entries proceeds without checking last_ta_idx, deletion_marker, invalidate, CSECT_QPROC_LIST_CACHE.
qfile_end_use_of_list_cache_entry() checks the following.
```
  if (lent->last_ta_idx == 0)
    {
      /* to check if it's the last transaction using the lent */
      if (marker || lent->deletion_marker)
	{
	  invalidate = lent->invalidate;
	  qfile_delete_list_cache_entry (thread_p, lent);
	}
    }
```

beyondykk9: lent->last_ta_idx is already checked at victim selection.

src/query/list_file.c
```diff
@@ -5085,31 +5085,38 @@ qfile_finalize_list_cache (THREAD_ENTRY * thread_p)
...
     }
 
+  list_ht_no = xcache_entry->list_ht_no;
+
   if (qfile_get_list_cache_number_of_entries (list_ht_no) == 0)
     {
       /* if no entries, to invalidate free the entry here */
       if (invalidate)
 	{
+	  xcache_entry->list_ht_no = -1;
```
shparkcubrid, line 5116: I think It looks good to write it inside qcache_free_ht_no().

shparkcubrid (approved):

ctshim (approved):

src/query/list_file.c
```diff
@@ -5050,32 +5085,49 @@ qfile_finalize_list_cache (THREAD_ENTRY * thread_p)
...
-  bool del = true;
   int cnt;
+  int list_ht_no;
 
-  if (QFILE_IS_LIST_CACHE_DISABLED)
+  if (csect_enter (thread_p, CSECT_QPROC_LIST_CACHE, INF_WAIT) != NO_ERROR)
     {
-      return ER_FAILED;
+      goto end;
+      //return ER_FAILED;
```
hgryoo: Is it the redundant comment? or Is error handling necessary here?

hgryoo (approved):

youngjinj (approved):

src/query/list_file.c
beyondykk9: Ok, I'll remove it.

PR #2856, opened 2021-06-23
[CBRD-23991] Clear the session information including holdable cursor when CAS connection is down

beyondykk9:
http://jira.cubrid.org/browse/CBRD-23991

While holdable cursor processing, even though CAS is down the cursor-related information is not cleared because the server keep the information for holdable cursor only by query-id.

The transaction-id is already remove by commit and the only way to remove the information related to holdable cursor is query-id from CAS. The query-id is only available from query_end by CAS but if the CAS is down the query-id is not available any more. At this time, the server should clear the session information including holdable cursor to prevent memory leak (i.e. dangling session information).

```
public static Connection connect() {
      Connection conn = null;
      try { 
           Class.forName("cubrid.jdbc.driver.CUBRIDDriver");
           conn = DriverManager.getConnection("jdbc:cubrid:192.168.1.3:55300:demodb:::","dba","");
           conn.setAutoCommit (false) ;
           conn.setHoldability(ResultSet.HOLD_CURSORS_OVER_COMMIT);
      } catch ( Exception e ) {
           System.err.println("SQLException : " + e.getMessage());
      }      return conn;
   } 

public static void main(String[] args) throws Exception {
    conn = connect();

    PreparedStatement stmt = conn.prepareStatement("select * from game where host_year = ?");
    stmt.setInt(1, 1988);
    rs = stmt.executeQuery();
    rs.next();

    System.out.println("1 " + rs.getInt(1));
    System.out.println("1 " + rs.getString(5));

    conn.commit();

    while (rs.next()) {
       System.out.println("2 " + rs.getInt(1));
       System.out.println("2 " + rs.getString(5));
       Thread.sleep(10000);
    }
}
```
- kill the CAS process at the server side.
- check stat by "cubrid statdump -s cursor -i 2 demodb"

**Expected:**
 *** SERVER EXECUTION STATISTICS ***
Num_query_holdable_cursors = 0

**Actual:**
 *** SERVER EXECUTION STATISTICS ***
Num_query_holdable_cursors = 1

### DIFF
--- a/src/session/session.c
+++ b/src/session/session.c
@@ -2394,7 +2394,7 @@ session_preserve_temporary_files (THREAD_ENTRY * thread_p, SESSION_QUERY_ENTRY *
 	    {
 	      if (!tfile_vfid_p->preserved)
 		{
-	          file_temp_preserve (thread_p, &tfile_vfid_p->temp_vfid);
+		  file_temp_preserve (thread_p, &tfile_vfid_p->temp_vfid);
 		  tfile_vfid_p->preserved = true;
 		}
 	    }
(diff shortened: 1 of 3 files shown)


src/session/session.c
```diff
@@ -2394,7 +2394,7 @@ session_preserve_temporary_files (THREAD_ENTRY * thread_p, SESSION_QUERY_ENTRY *
 	    {
 	      if (!tfile_vfid_p->preserved)
 		{
-	          file_temp_preserve (thread_p, &tfile_vfid_p->temp_vfid);
+		  file_temp_preserve (thread_p, &tfile_vfid_p->temp_vfid);
```
beyondykk9, line 2397: it is not related to this issue. it is just for indentation.

ctshim (approved):

hgryoo (approved):

shparkcubrid (approved):

youngjinj (approved):

PR #2859, opened 2021-06-25
[CBRD-23993] Implementation of predicate push down and bind param for dblink

beyondykk9:
http://jira.cubrid.org/browse/CBRD-23993

**Background:**

The dblinked-query is a type of sub-query surrounded by main query. The dblinked-query should be merged or assembled with predicate from main query. We call this processing as "predicate push". This should be done for performance enhancement. The below query is an example needs to push the predicate down to dblinked-query.

```
SELECT loc.col1_loc, substring(r.col2, 1, 3), loc.col3_big
FROM dblink( '192.168.1.8:55300:demodb:dba::?queryTimeout=5000',  
                       'SELECT col1, col2_varchar, col3_big, col4_int from [dblink] WHERE col4_int > 0' ) 
           AS r(col1 int, col2 varchar(16), col3 bigint, col4 int), 
           dblink2 loc
WHERE r.col1 + r.col4 > r.col3 AND r.col1 = loc.col1_loc;
```
In the above query, the dblinked query can be rewritten to the following:

```
SELECT col1, col2_varchar, col3_big, col4_int 
FROM [dblink]
WHERE col4_int > 0 AND col1 + col4_int > col3_big
```
The "col1 + col4_int > col3_big" predicate is pushed to the dblinked query.

**Implementation:**

The predicate push down is already implemented as mq_copypush_sargable_terms_helper but this function is applied to only sub-query (called as "IN LINE VIEW"). We need it extend to derived table (as DBLINK).


src/query/dblink_scan.c
```diff
@@ -301,6 +333,147 @@ dblink_make_date_time_tz (T_CCI_U_TYPE utype, DB_VALUE * value_p, T_CCI_DATE_TZ
...
+	default:
+	  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_DBLINK, 1, "bind: not supported type");
+	  return S_ERROR;
+	}
+      ret = cci_bind_param (scan_info->stmt_handle, i + 1, a_type, value, u_type, 0);
+      if (ret < 0)
+	{
+	  if (ret == CCI_ER_BIND_INDEX)
+	    {
+	      break;
```
ctshim: S_SUCCESS is returned when exiting from here with a break.
Is your intention correct?

src/query/dblink_scan.c
```diff
@@ -301,6 +333,147 @@ dblink_make_date_time_tz (T_CCI_U_TYPE utype, DB_VALUE * value_p, T_CCI_DATE_TZ
...
+	{
+	case DB_TYPE_INTEGER:
+	  a_type = CCI_A_TYPE_INT;
+	  u_type = CCI_U_TYPE_BIGINT;
+	  break;
+	case DB_TYPE_BIGINT:
+	  a_type = CCI_A_TYPE_BIGINT;
+	  u_type = CCI_U_TYPE_BIGINT;
+	  break;
+	case DB_TYPE_NUMERIC:
```
ctshim, line 372: If you modify it as follows, the type variable is not needed.

```
	case DB_TYPE_NUMERIC:
           {
	      numeric_coerce_num_to_double (db_locate_numeric (&vd->dbval_ptr[i]),
					    vd->dbval_ptr[i].domain.numeric_info.scale, &adouble);
	      value = &adouble;
	    }
	case DB_TYPE_DOUBLE:
	case DB_TYPE_FLOAT:
	  a_type = CCI_A_TYPE_DOUBLE;
	  u_type = CCI_U_TYPE_DOUBLE;	  
	  break;
```

src/parser/parse_tree.h
```diff
@@ -114,6 +114,7 @@ struct json_t;
 
 #define PT_IS_QUERY_NODE_TYPE(x) \
     (  (x) == PT_SELECT     || (x) == PT_UNION \
+    || (x) == PT_DBLINK_TABLE \
```
ctshim: There are so many places where PT_IS_QUERY_NODE_TYPE() is used that it looks like you need to be careful.

How about modifying it like the one below where it needs to be processed?
`if (PT_IS_QUERY_NODE_TYPE (node->node_type) || node->node_type == PT_DBLINK_TABLE)`

In a similar form, there is a part expressed as below in the go_optimize_queries() function.
`if (PT_IS_QUERY_NODE_TYPE (node->node_type) || node->node_type == PT_UPDATE || node->node_type == PT_DELETE)`

src/parser/view_transform.c
```diff
@@ -3147,6 +3151,45 @@ pt_copypush_terms (PARSER_CONTEXT * parser, PT_NODE * spec, PT_NODE * query, PT_
...
+	  rewritten = pt_append_varchar (parser, rewritten, pushed_pred);
+	}
+
+      query->info.dblink_table.rewritten = rewritten;
+
+      parser->custom_print = save_custom;
+
+#if !defined (NDEBUG)
+      printf ("===> rewriting query = %s\n", (char *) rewritten->bytes);
+#endif
```
ctshim, line 3196: Is this the code you want to remove before generating the answer sheet for QA later?
How about removing it if it doesn't make much sense?

src/parser/view_transform.c
```diff
@@ -4647,7 +4695,8 @@ mq_check_rewrite_select (PARSER_CONTEXT * parser, PT_NODE * select_statement)
 	    {
 	      from->next = mq_rewrite_vclass_spec_as_derived (parser, select_statement, from->next, NULL);
 	    }
-	  else if (from->next->info.spec.derived_table_type == PT_IS_SUBQUERY)
+	  else if (from->next->info.spec.derived_table_type == PT_IS_SUBQUERY
+		   || from->info.spec.derived_table_type == PT_DERIVED_DBLINK_TABLE)
```
ctshim, line 4759: Is it not necessary to modify the code below on line 4726?

`if (is_union_translation == false && from && from->info.spec.derived_table_type == PT_IS_SUBQUERY)`

src/query/dblink_scan.c
beyondykk9: yes, return S_SUCCESS at the outside of this loop.

src/query/dblink_scan.c
beyondykk9: ok, I agree.

src/parser/parse_tree.h
beyondykk9: Ok, I agree.

src/parser/view_transform.c
beyondykk9: I will remove this code after debugging done.

src/parser/view_transform.c
beyondykk9: Ok, I fixed it.

src/parser/view_transform.c
```diff
@@ -2755,6 +2755,7 @@ pt_pushable_query_in_pos (PARSER_CONTEXT * parser, PT_NODE * query, int pos)
   switch (query->node_type)
     {
     case PT_SELECT:
+    case PT_DBLINK_TABLE:
```
shparkcubrid, line 2759: Is PT_DBLINK_TABLE the same structure as PT_SELECT? 
Does the PT_DBLINK_TABLE node have query->info.query.q.select.list?

src/parser/view_transform.c
```diff
@@ -4610,7 +4656,9 @@ mq_check_rewrite_select (PARSER_CONTEXT * parser, PT_NODE * select_statement)
     }
 
   from = select_statement->info.query.q.select.from;
-  if (from && (from->next || pt_has_aggregate (parser, select_statement)))
+  if (from
+      && (from->next || (from->info.spec.derived_table_type == PT_DERIVED_DBLINK_TABLE)
```
shparkcubrid: This condition doesn't seem to be necessary.

ctshim (approved):

src/parser/view_transform.c
beyondykk9: info.query.q.select.list is null on the node PT_DBLINK_TABLE, but the node should be dealt as a pushable node in order to return "pushable = true"

src/parser/view_transform.c
beyondykk9: but wihout this condition the pushable term is not produced for dblink query.

src/parser/view_transform.c
shparkcubrid: query->info.query.q.select.list may not be null.
PT_DBLINK_TABLE should be separated from existing routine.

beyondykk9: How about modifying as below:
```
case PT_DBLINK_TABLE:
  pushable = true;
  break;
```

shparkcubrid: I can't comment because I haven't analyzed routines related to view. I'm analyzing the relevant routines as a whole, which I think will take quite some time.

src/query/query_executor.c
```diff
@@ -6718,14 +6719,15 @@ qexec_open_scan (THREAD_ENTRY * thread_p, ACCESS_SPEC_TYPE * curr_spec, VAL_LIST
...
+      host_vars.index = curr_spec->s.dblink_node.host_var_index;
       error_code =
 	scan_open_dblink_scan (thread_p, s_id,
 			       curr_spec->single_fetch, scan_op_type,
 			       curr_spec->s.dblink_node.conn_url,
 			       curr_spec->s.dblink_node.conn_user,
 			       curr_spec->s.dblink_node.conn_password,
-			       curr_spec->s.dblink_node.conn_sql,
-			       vd, val_list, curr_spec->s.dblink_node.dblink_regu_list_pred, curr_spec->where_pred);
+			       curr_spec->s.dblink_node.conn_sql, vd, val_list, &host_vars, curr_spec->where_pred);
```
hgryoo: how about just passing `curr_spec->s.dblink_node` as an argument of `scan_open_dblink_scan()`?

src/parser/parse_tree.h
```diff
@@ -3267,6 +3275,9 @@ typedef struct pt_dblink_info
   PT_NODE *pwd;
   PT_NODE *qstr;		/* query string */
   PT_NODE *cols;		/* column definition  */
+  PT_NODE *pushed_pred;		/* pushed predicate from main query */
```
ctshim, line 3278: Add the following to the pt_apply_dblink_table() function
`PT_APPLY_WALK(parser, p->info.dblink_table.pushed_pred, arg);`

For variables other than dblink_table.cols, actions will be taken on other variables that were already there through CBRD-23977.

src/query/query_executor.c
beyondykk9: Ok I modified.

src/parser/view_transform.c
shparkcubrid: There is same routine at below else statement.
Can't it be handled with that routine?

src/parser/view_transform.c
shparkcubrid: The logic to check pushable in the mq_copypush_sargable_terms function occupies a large part. However, I don't think the check is necessary in this case because the predicate is added to the dblink clause as shown below.
```
select * from ( subquery) where 'pushed predicate'
```
so real predicate push is performed at remote server.

I think adding dblink-related logic inside the mq_copypush_sargable_terms function will result in a lot of unintended behavior.
I would recommend completely separating the relevant logic from mq_copypush_sargable_terms ().

src/parser/view_transform.c
beyondykk9: The dblink node might not be a single and the multiple dblink should be processed by while loop with "from->next".

shparkcubrid: If it's multiple dblinks, isn't from->next true?

src/parser/view_transform.c
beyondykk9: mq_copypush_sargable_terms_dblink is added for separating routiine.

src/parser/parse_tree.h
beyondykk9: It looks not necessary because the pushed_pred is walk done.

src/parser/view_transform.c
beyondykk9: Ok, I understand what you say. I removed a redundant code.

src/parser/view_transform.c
```diff
@@ -3162,6 +3207,68 @@ pt_copypush_terms (PARSER_CONTEXT * parser, PT_NODE * spec, PT_NODE * query, PT_
...
+  PT_NODE *save_next;
+
+  /* init */
+  push_term_list = NULL;
+  push_cnt = 0;
+  copy_cnt = -1;
+
+  for (term = statement->info.query.q.select.where; term; term = term->next)
+    {
+      if (pt_sargable_term (parser, term, infop) && PT_PUSHABLE_TERM (infop))
```
shparkcubrid, line 3236: pt_sargable_term() should also be separated. In dblink, only the attr of the same spec should be checked.

shparkcubrid: When I checked it again, it seems that it is okay to use pt_sargable_term().
I recommend that various tests will be performed with this routine.

beyondykk9: Ok, I agree. Thanks.

src/parser/view_transform.c
shparkcubrid: This is a routine that examines the select list of a sub query.
this routine is not needed in dblink.
```
case PT_DBLINK_TABLE:
  pushable = true;
  break;
```

src/parser/view_transform.c
```diff
@@ -3162,6 +3207,68 @@ pt_copypush_terms (PARSER_CONTEXT * parser, PT_NODE * spec, PT_NODE * query, PT_
...
+	{
+	  /* copy term */
+	  new_term = parser_copy_tree (parser, term);
+
+	  /* for term, mark as copy-pushed term */
+	  if (term->node_type == PT_EXPR)
+	    {
+	      if (copy_cnt == -1)	/* very the first time */
+		{
+		  copy_cnt = pt_check_copypush_subquery (parser, new_query);
```
shparkcubrid: There is nowhere to use copy_cnt. You need to confirm that it can be deleted.

src/parser/view_transform.c
```diff
@@ -3162,6 +3207,68 @@ pt_copypush_terms (PARSER_CONTEXT * parser, PT_NODE * spec, PT_NODE * query, PT_
...
+		}
+
+	      if (copy_cnt == 0)	/* not found not-pushable query */
+		{
+		  PT_EXPR_INFO_SET_FLAG (term, PT_EXPR_INFO_COPYPUSH);
```
shparkcubrid: There is nowhere to use PT_EXPR_INFO_COPYPUSH.  it is cleared immediately after setting it.

src/parser/view_transform.c
beyondykk9: the next squence is to check if copy_cnt is zero. you can see the code " if (copy_cnt == 0)" at the next line.

src/parser/view_transform.c
beyondykk9: PT_EXPR_INFO_SET_FLAG  is used after push down done for removing the term node (which is already pushed to the dblink).
You can see the code at **qo_optimize_queries_post**.

src/parser/view_transform.c
beyondykk9: Yes, the dblink is necessary to separate form the case but the select list is always null so result in no loop to examine.

src/parser/view_transform.c
```diff
@@ -3301,8 +3408,11 @@ mq_copypush_sargable_terms (PARSER_CONTEXT * parser, PT_NODE * statement, PT_NOD
 
   if (statement->node_type == PT_SELECT
       /* never do copy-push optimization for a hierarchical query */
-      && statement->info.query.q.select.connect_by == NULL && spec->info.spec.derived_table_type == PT_IS_SUBQUERY
-      && (derived_table = spec->info.spec.derived_table) && PT_IS_QUERY (derived_table)
+      && statement->info.query.q.select.connect_by == NULL
+      && (spec->info.spec.derived_table_type == PT_IS_SUBQUERY
+	  || spec->info.spec.derived_table_type == PT_DERIVED_DBLINK_TABLE)
+      && (derived_table = spec->info.spec.derived_table)
+      && (PT_IS_QUERY (derived_table) || derived_table->node_type == PT_DBLINK_TABLE)
```
shparkcubrid: How about fixing it like below.
```
if (statement->node_type == PT_SELECT
   && statement->info.query.q.select.connect_by == NULL
   && (derived_table = spec->info.spec.derived_table) 
{
  if ( PT_IS_QUERY (derived_table)
        && (spec->info.spec.derived_table_type == PT_IS_SUBQUERY )
     { 
        mq_copypush_sargable_terms_helper ()
     }
  else if (derived_table->node_type == PT_DBLINK_TABLE
           spec->info.spec.derived_table_type == PT_DERIVED_DBLINK_TABLE)
    {
      mq_copypush_sargable_terms_dblink ()
    }
}
```

src/parser/view_transform.c
```diff
@@ -3147,6 +3153,45 @@ pt_copypush_terms (PARSER_CONTEXT * parser, PT_NODE * spec, PT_NODE * query, PT_
...
+      parser->custom_print |= PT_CONVERT_RANGE | PT_SUPPRESS_RESOLVED | PT_PRINT_NO_HOST_VAR_INDEX;
+      pushed_pred = pt_print_and_list (parser, query->info.dblink_table.pushed_pred);
+
+      /* wrapped query SELECT * FROM */
+      rewritten = pt_append_bytes (parser, rewritten, "SELECT * FROM (", 15);
+      query_str = query->info.dblink_table.qstr->info.value.data_value.str;
+      rewritten = pt_append_varchar (parser, rewritten, query_str);
+
+      /* alias name : '_r' */
+      rewritten = pt_append_bytes (parser, rewritten, ") AS _r", 7);
```
hgryoo, line 3172: alias name `_r` possibly conflict?

beyondykk9: it's no conflict because the sub-query is not depend on the main query. it's used just as remote query at the remote server.

src/parser/view_transform.c
shparkcubrid: info is union. query->info.query.q.select.list may not be null.

src/parser/view_transform.c
beyondykk9: the below condition should be also added:
  && !PT_SELECT_INFO_IS_FLAGED (statement, PT_SELECT_INFO_IS_MERGE_QUERY) at the top as top of the condition.

and it looks not necessary "spec->info.spec.derived_table_type == PT_DERIVED_DBLINK_TABLE" because the the condition is always true when the node type is PT_BLINK_TABLE.

src/parser/view_transform.c
shparkcubrid: Should copy_cnt be checked in dblink?

src/parser/view_transform.c
shparkcubrid: I will continue the discussion in the comments below.

src/parser/view_transform.c
shparkcubrid: I said it for readability reasons.

src/parser/view_transform.c
byungwook-kim: the copy_cnt is used to set the flag PT_EXPR_INFO_COPYPUSH, the flag is used to check the term is pushed down and to remove the node.

shparkcubrid: The call of pt_check_copypush_subquery () is not needed in dblink. The copy_cnt is always 0.

beyondykk9: copy_cnt is always 0 in dblink, so the below code looks better.

	  /* for term, mark as copy-pushed term */
	  if (term->node_type == PT_EXPR)
	    {
	      PT_EXPR_INFO_SET_FLAG (term, PT_EXPR_INFO_COPYPUSH);
	      PT_EXPR_INFO_CLEAR_FLAG (new_term, PT_EXPR_INFO_COPYPUSH);
	    }

src/parser/view_transform.c
beyondykk9: Ok, good point for readability.

src/parser/view_transform.c
beyondykk9: Ok, I separate the dblink from the case.

src/parser/view_transform.c
shparkcubrid: 'PT_EXPR_INFO_CLEAR_FLAG (new_term, PT_EXPR_INFO_COPYPUSH);' is not needed in dblink.

In my opinion it is better not to delete predicates in dblink. This helps when predicate push is not working properly. This is my opinion, the decision is yours.

beyondykk9: Ok, I agree you.

shparkcubrid (approved):

hgryoo (approved):

youngjinj (approved):

PR #2864, opened 2021-06-25
[LETS-114] Do not write log to disk

ribram:
http://jira.cubrid.org/browse/LETS-114

On a transaction server with remote storage:

- Instead of writing the appended log pages to disk, set nxio_lsa and reset flush page count directly.
- Overwrite dirty old log pages in the log page buffer with new pages.
- Do not create an archive from the active log.

**Tests**

Tested the same scenario (many inserts) with both transaction server on remote and local storage. Transaction server with remote storage does not create any archives and does not write any pages to disk, while the transaction server with local storage does.

### DIFF
--- a/src/transaction/log_page_buffer.c
+++ b/src/transaction/log_page_buffer.c
@@ -351,6 +351,8 @@ static int logpb_add_archive_page_info (THREAD_ENTRY * thread_p, int arv_num, LO
 static int logpb_get_archive_num_from_info_table (THREAD_ENTRY * thread_p, LOG_PAGEID page_id);
 
 static int logpb_flush_all_append_pages (THREAD_ENTRY * thread_p);
+static int logpb_write_append_pages_to_disk (THREAD_ENTRY * thread_p);
+static void logpb_skip_flush_append_pages ();
 static int logpb_append_next_record (THREAD_ENTRY * thread_p, LOG_PRIOR_NODE * ndoe);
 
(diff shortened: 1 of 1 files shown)


src/transaction/log_page_buffer.c
```diff
@@ -3307,6 +3310,20 @@ logpb_flush_all_append_pages (THREAD_ENTRY * thread_p)
   gettimeofday (&start_time, NULL);
 #endif /* CUBRID_DEBUG */
 
+  if (is_tran_server_with_remote_storage ())
+    {
+      // Do not write on local disk
+      rv = pthread_mutex_lock (&flush_info->flush_mutex);
```
cristiarg: `rv` is not used anywhere, only assigned

src/transaction/log_page_buffer.c
```diff
@@ -3307,6 +3310,20 @@ logpb_flush_all_append_pages (THREAD_ENTRY * thread_p)
   gettimeofday (&start_time, NULL);
 #endif /* CUBRID_DEBUG */
 
+  if (is_tran_server_with_remote_storage ())
```
cristiarg: This is already a huge function and this new bit is an 'early out'. Maybe fork with two new functions? - fork function, remote storage function, local storage function (current function, only renamed).
Just saying...

cristiarg (approved):

PR #2868, opened 2021-06-29
[LETS-136] Disconnect transaction server from page server

adrianbrab:
http://jira.cubrid.org/browse/LETS-136

The transaction server disconnects from the page server when it shuts down. All resources involved in the connection, on the page server, are destroyed.

The transaction server reconnects to the page server and the connection is re-created.

### DIFF
--- a/src/base/server_type.cpp
+++ b/src/base/server_type.cpp
@@ -86,6 +86,11 @@ void finalize_server_type ()
     {
       ats_Gl.disconnect_page_server ();
     }
+  else
+    {
+      assert (get_server_type () == SERVER_TYPE_PAGE);
+      ps_Gl.disconnect_active_tran_server ();
+    }
 }
 
 bool is_tran_server_with_remote_storage ()
--- a/src/base/server_type.cpp
+++ b/src/base/server_type.cpp
@@ -86,11 +86,6 @@ void finalize_server_type ()
     {
       ats_Gl.disconnect_page_server ();
     }
-  else
-    {
-      assert (get_server_type () == SERVER_TYPE_PAGE);
-      ps_Gl.disconnect_active_tran_server ();
-    }
 }
 
 bool is_tran_server_with_remote_storage ()
--- a/src/server/active_tran_server.cpp
+++ b/src/server/active_tran_server.cpp
@@ -276,7 +276,8 @@ void
 active_tran_server::disconnect_page_server ()
 {
   assert_is_active_tran_server ();
-
+  std::string msg = "" + static_cast<int> (cubcomm::server_server::CONNECT_ACTIVE_TRAN_TO_PAGE_SERVER);
+  push_request (ats_to_ps_request::SEND_DISCONNECT_MSG, std::move (msg));
   m_page_server_conn_vec.clear ();
 }
 
--- a/src/server/active_tran_server.cpp
+++ b/src/server/active_tran_server.cpp
@@ -266,7 +266,7 @@ active_tran_server::connect_to_page_server (const cubcomm::node &node, const cha
     },
   }));
 
-  log_Gl.m_prior_sender.add_sink (std::bind (&active_tran_server::push_request, std::ref (*this), 0,
+  log_Gl.m_prior_sender.add_sink (std::bind (&active_tran_server::push_request, std::ref (*this),
 				  ats_to_ps_request::SEND_LOG_PRIOR_LIST, std::placeholders::_1));
 
   return NO_ERROR;
@@ -276,7 +276,7 @@ void
 active_tran_server::disconnect_page_server ()
 {
   assert_is_active_tran_server ();
-  const int payload = cubcomm::server_server::CONNECT_ACTIVE_TRAN_TO_PAGE_SERVER;
+  const int payload = static_cast<int> (cubcomm::server_server::CONNECT_ACTIVE_TRAN_TO_PAGE_SERVER);
   std::string msg (reinterpret_cast<const char *> (&payload), sizeof (payload));
   er_log_debug (ARG_FILE_LINE, "Transaction server starts disconnecting from the page servers.");
 
@@ -284,7 +284,7 @@ active_tran_server::disconnect_page_server ()
     {
       er_log_debug (ARG_FILE_LINE, "Transaction server disconnected from page server with channel id: %s.\n",
 		    m_page_server_conn_vec[i]->get_underlying_channel_id ());
-      m_page_server_conn_vec[i]->push (ats_to_ps_request::SEND_DISCONNECT_MSG, msg);
+      m_page_server_conn_vec[i]->push (ats_to_ps_request::SEND_DISCONNECT_MSG, std::move (msg));
     }
   m_page_server_conn_vec.clear ();
   er_log_debug (ARG_FILE_LINE, "Transaction server disconnected from all page servers.");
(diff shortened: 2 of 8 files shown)


src/base/server_type.cpp
```diff
@@ -86,11 +86,6 @@ void finalize_server_type ()
     {
       ats_Gl.disconnect_page_server ();
     }
-  else
-    {
-      assert (get_server_type () == SERVER_TYPE_PAGE);
-      ps_Gl.disconnect_active_tran_server ();
-    }
```
ribram: This might still have to be done. If the page server is shutting down, it will have to remove the connection with the transaction server.

What is the reasoning for removing the code?

src/server/active_tran_server.cpp
```diff
@@ -276,8 +276,17 @@ void
 active_tran_server::disconnect_page_server ()
 {
   assert_is_active_tran_server ();
+  std::string msg = "" + static_cast<int> (cubcomm::server_server::CONNECT_ACTIVE_TRAN_TO_PAGE_SERVER);
```
ribram: Send the memory reprezentation of cubcomm::server_server::CONNECT_ACTIVE_TRAN_TO_PAGE_SERVER.
```suggestion
  const int payload = cubcomm::server_server::CONNECT_ACTIVE_TRAN_TO_PAGE_SERVER;
  std::string msg (reinterpret_cast<const char *> (&payload), sizeof (payload));
```

src/server/active_tran_server.cpp
```diff
@@ -276,8 +276,17 @@ void
...
+  er_log_debug (ARG_FILE_LINE, "Transaction server starts disconnecting from the page servers.");
 
+  for (int i = 0; i < m_page_server_conn_vec.size(); i++)
```
ribram:
```suggestion
  for (size_t i = 0; i < m_page_server_conn_vec.size(); i++)
```

src/server/active_tran_server.cpp
```diff
@@ -276,8 +276,17 @@ void
...
+    {
+      er_log_debug (ARG_FILE_LINE, "Transaction server disconnected from page server with channel id: %s.\n",
+		    m_page_server_conn_vec[i]->get_underlying_channel_id ());
+      push_request (i, ats_to_ps_request::SEND_DISCONNECT_MSG, std::move (msg));
```
ribram: - Do not change push_request() function. Access m_page_server_conn_vec directly.
- If there are multiple page servers, moving the message on the first request can cause invalid memory access in the other requests. Copy it instead.
```suggestion
      m_page_server_conn_vec[i]->push (ats_to_ps_request::SEND_DISCONNECT_MSG, msg);
```

src/server/active_tran_server.hpp
```diff
@@ -59,7 +59,7 @@ class active_tran_server
 
     bool uses_remote_storage () const;
 
-    void push_request (ats_to_ps_request reqid, std::string &&payload);
+    void push_request (int page_server_index, ats_to_ps_request reqid, std::string &&payload);
```
ribram:
```suggestion
    void push_request (ats_to_ps_request reqid, std::string &&payload);
```

src/server/ats_ps_request.hpp
```diff
@@ -23,14 +23,16 @@ enum class ats_to_ps_request
...
+  SEND_DISCONNECT_MSG
 };
 
 enum class ps_to_ats_request
 {
   SEND_SAVED_LSA,
   SEND_LOG_PAGE,
-  SEND_DATA_PAGE
+  SEND_DATA_PAGE,
+  SEND_DISCONNECT_RSP
```
ribram: Not used
```suggestion
  SEND_DATA_PAGE
```

src/server/page_server.cpp
```diff
@@ -124,6 +131,12 @@ void page_server::receive_data_page_fetch (cubpacking::unpacker &upk)
 				   std::placeholders::_2));
 }
 
+void page_server::receive_disconnect_request (cubpacking::unpacker &upk)
+{
+  //start a thread to destroy the ATS to PS connection object
+  std::thread thread (&page_server::disconnect_active_tran_server, std::ref (*this));
```
cristiarg: Use another name for the variable - `disconnect_thread`. Thread should be detached or joined.

src/base/server_type.cpp
adrianbrab: If one PS shuts down shouldn't the ATS be notified? The connection, from the ATS point of view, is still up

src/server/active_tran_server.cpp
adrianbrab: Changed

cristiarg (approved):

src/server/active_tran_server.cpp
adrianbrab: Done.

src/server/ats_ps_request.hpp
adrianbrab: Removed now

src/server/page_server.cpp
adrianbrab: Used detach. Is it really necessary in this context?

ribram: https://en.cppreference.com/w/cpp/thread/thread/~thread

This proves @cristiarg is right and it must be detached.

cristiarg: Either `join` or `detach` is needed in any context. If merely to express intent. If process decides to close (not crash) in that instant, and the thread did not do its job, process remains hanging (even for a tiny amount of time).

src/base/server_type.cpp
ribram: It should, and `disconnect_active_tran_server ()` will have to incorporate that functionality. It is not however in the scope of this pull request. and it will have to be updated later. In the meantime, my opinions is that it is better to have it called.

src/server/page_server.cpp
cristiarg: Actually I was wrong. Without `detach` the thread might not get to do what was intended, and its dtor will cause a core.

src/server/page_server.cpp
```diff
@@ -124,6 +131,13 @@ void page_server::receive_data_page_fetch (cubpacking::unpacker &upk)
 				   std::placeholders::_2));
 }
 
+void page_server::receive_disconnect_request (cubpacking::unpacker &upk)
+{
+  //start a thread to destroy the ATS to PS connection object
+  std::thread disconnect_thread (&page_server::disconnect_active_tran_server, std::ref (*this));
+  disconnect_thread.detach();
```
ribram:
```suggestion
  disconnect_thread.detach ();
```

src/server/active_tran_server.cpp
```diff
@@ -276,8 +276,18 @@ void
...
   assert_is_active_tran_server ();
+  const int payload = static_cast<int> (cubcomm::server_server::CONNECT_ACTIVE_TRAN_TO_PAGE_SERVER);
+  std::string msg (reinterpret_cast<const char *> (&payload), sizeof (payload));
+  er_log_debug (ARG_FILE_LINE, "Transaction server starts disconnecting from the page servers.");
 
+  for (size_t i = 0; i < m_page_server_conn_vec.size(); i++)
+    {
+      er_log_debug (ARG_FILE_LINE, "Transaction server disconnected from page server with channel id: %s.\n",
+		    m_page_server_conn_vec[i]->get_underlying_channel_id ());
+      m_page_server_conn_vec[i]->push (ats_to_ps_request::SEND_DISCONNECT_MSG, std::move (msg));
```
ribram:
```suggestion
      m_page_server_conn_vec[i]->push (ats_to_ps_request::SEND_DISCONNECT_MSG, std::move (std::string (msg)));
```

ribram (approved):
Please make the request changes and merge.

src/server/page_server.cpp
adrianbrab: Fixed now.

src/server/active_tran_server.cpp
adrianbrab: Done.

PR #2869, opened 2021-06-29
[LETS-126] - Include unit tests in the appveyor build

danilailie:
http://jira.cubrid.org/browse/LETS-126

* Fixed unit tests compile errors for windows.


unit_tests/log/dummy_page_server.cpp
```diff
@@ -0,0 +1,56 @@
+#include "log_replication.hpp"
```
ribram, line 19: Add license header.

unit_tests/log/CMakeLists.txt
```diff
@@ -273,42 +279,43 @@ add_dependencies(test_data_page_fetcher
...
-  )
-target_compile_definitions(test_log_recovery_parallel PRIVATE
-  ${COMMON_DEFS}
-  SERVER_MODE
-  )
-target_include_directories(test_log_recovery_parallel PRIVATE
-  ${TEST_INCLUDES}
-  )
-#target_link_libraries(test_log_recovery_parallel PRIVATE test_common)
 if(UNIX)
```
ribram, line 315: Why are these unit tests limited to linux?

unit_tests/packing/CMakeLists.txt
```diff
@@ -44,15 +44,15 @@ target_include_directories(test_packing PRIVATE
...
+    test_common
     cubrid
     )
 elseif(WIN32)
-	target_link_libraries(test_packing PRIVATE
+	target_link_libraries(test_packing PRIVATE 
+    test_common
     cubrid-win-lib
     )
 else()
```
ribram: Is this change required? I think test_packing does not need fake_thread_utils.

unit_tests/log/CMakeLists.txt
```diff
@@ -231,7 +236,7 @@ target_compile_definitions(test_log_lsa_utils PRIVATE
 target_include_directories(test_log_lsa_utils PRIVATE
   ${TEST_INCLUDES}
   )
-# target_link_libraries(test_log_lsa_utils fake_thread_utils)
+target_link_libraries(test_log_lsa_utils fake_thread_utils)
```
ribram: Does test_log_lsa_utils depend on fake_thread_util? It doesn't like it does. If it is does not, better to remove the line completely.

ribram (changes requested):
The issue (and PR title) mentions including building unit tests in the appveyor build. Update .appveyor.yml.

unit_tests/log/ut_database.hpp
```diff
@@ -97,10 +97,10 @@ class ut_database_values_generator
     std::uniform_real_distribution<double> m_duration_in_millis_dist;
 
     // discrete distribution for adding new volumes or updating existing volumes
-    std::discrete_distribution<int8_t> m_add_or_update_volume_dist;
+    std::discrete_distribution<int32_t> m_add_or_update_volume_dist;
 
     // discrete distribution for adding new pages or updating existing pages
-    std::discrete_distribution<int8_t> m_add_or_update_page_dist;
+    std::discrete_distribution<int32_t> m_add_or_update_page_dist;
```
ribram, line 103: Why this change? Please explain.

ribram:
@danilailie the appveyor build is still missing -DUNIT_TESTS option that triggers unit test compilation.

unit_tests/log/ut_database.hpp
danilailie: On vs compiler it complains that the template instantiation with an 8 bit type is illegal.

ribram: Right, the standard documentation for [discrete_distribution](https://en.cppreference.com/w/cpp/numeric/random/discrete_distribution) mentions:

<html>
<body>
<h3 style="color: rgb(0, 0, 0); background: none rgb(255, 255, 255); font-weight: bold; margin: 0px 0px 0.3em; overflow: hidden; padding: 1.1em 0px 0.2em 0.75em; border-bottom: none; width: auto; font-size: 16.896px; font-family: DejaVuSans, &quot;DejaVu Sans&quot;, arial, sans-serif; font-style: normal; font-variant-ligatures: no-common-ligatures; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><span class="mw-headline" id="Template_parameters">Template parameters</span></h3>

IntType | - | The result type generated by the generator. The effect is undefined if this is not one of short, int, long, long long, unsigned short, unsigned int, unsigned long, or unsigned long long.
-- | -- | --



</body>
</html>

Shouldn't m_add_or_update_page_dist & m_add_or_update_page_dist usages be updated too, as consequence of this change?

ribram:
Please check appveyor error:
```
CMake Error: Could not create named generator Visual Studio 16 2019 Win64
CMake Error: CMAKE_GENERATOR was set but the specified generator doesn't exist. Using CMake default.
Generators
* Visual Studio 16 2019        = Generates Visual Studio 2019 project files.
                                 Use -A option to specify architecture.
  Visual Studio 15 2017 [arch] = Generates Visual Studio 2017 project files.
                                 Optional [arch] can be "Win64" or "ARM".
```
The fix should be adding `-A` before Win64.

unit_tests/log/ut_database.hpp
danilailie: It's done

ribram (approved):

PR #2865, opened 2021-06-28
[CBRD-23903] Set tuple_count as INT64 type in list file

hgryoo:
http://jira.cubrid.org/browse/CBRD-23903

At the previous PR(#2666), The return type of count queries is changed to BIGINT type. However, counting row count should be performed in real.
```
All kinds of queries to count # of rows are gathered into list_file.tuple_count

                      Temp file, including result (list_file)
                                       ^
    -----------------------------------|----------------------------------------------
    |                                  |                   |                         |
B-tree(index)              Temp file (list_file)      Heap file              Aggregate/Analysis

```

1) list_file->tuple_cnt is changed to INT64
- db_query_count(),  db_query_seek_tuple(), insert/delete/update routines in execute_statement.c take list_file->tuple_cnt value. therefore they are changed.
- packing/unpacking at object_representation.c is changed. (aligning is considered)
2) stats that counting # of rows/keys in scan_stats are changed to print correctly for `;trace on`

About client cursor (cursor.c)
- trivial change in cursor.c, typo; when I investigating the code, I found a typo.
- tuple number and tuple count in cursor id don't have to change. They get the number from the page's # of tuples, not the whole list_file's count.

I've tested for the following count queries and showed the trace
```
;sc t1
====
 <Class Name>

     t1

 <Attributes>

     c1                   BIGINT NOT NULL

 <Constraints>

     PRIMARY KEY pk_t1 ON t1 (c1)
```
```
SELECT count(*) from t1;
=== <Result of SELECT Command in Line 3> ===

              count(*)
======================
            2232889346

1 row selected. (1126.773044 sec) Committed.

=== Auto Trace ===

Query Plan:
  TABLE SCAN (t1)

  rewritten query: select count(*) from t1 t1

Trace Statistics:
  SELECT (time: 1126689, fetch: 4885872, ioread: 4885871)
    SCAN (table: t1), (heap time: 896044, fetch: 4885870, ioread: 4885870, readrows: 2232889346, rows: 2232889346)
```
```
> SELECT count(c1) from t1;
=== <Result of SELECT Command in Line 5> ===

             count(c1)
======================
            2232889346

1 row selected. (1800.630586 sec) Committed.

=== Auto Trace ===

Query Plan:
  TABLE SCAN (t1)

  rewritten query: select count(t1.c1) from t1 t1

Trace Statistics:
  SELECT (time: 1800615, fetch: 4885872, ioread: 4885870)
[...]

PR #2876, opened 2021-07-05
[CBRD-23903] Improve COUNT query to return a BIGINT value

hgryoo:
http://jira.cubrid.org/browse/CBRD-23903

It is a merge commit, #2666, #2865 are included.
The `count_debug` parameter, which was added temporarily for code review, has been removed.


hornetmj:
@ssihil @mhoh3963 The 'ci/circleci: test_sql' fails cause to https://github.com/CUBRID/cubrid/pull/2876. How can we handle it? Do we wait for fixing the answer files? Or do you merge this PR forcefully?

PR #2877, opened 2021-07-06
[CBRD-23977] Interoperability support of the catalog tables for the dblink in the syntax DBLINK

ctshim:
http://jira.cubrid.org/browse/CBRD-23977

Purpose
  Support to be executed in connection with server information in db_server table when executing DBLINK query.

Implementation
-  remove pt_init_dblink_table() function.
-  Fix to apply functions (pt_apply_dblink_table(),  pt_apply_create_server(), pt_apply_drop_server())
-  Fix pt_print_dblink_table()
- Fix Authorization issue when running drop server
- Check if dblink query statement is SELECT.
   (SELECT with WITH clause is not supported.)
Remarks
  - Password encryption is not yet supported.


src/parser/name_resolution.c
```diff
@@ -10234,3 +10234,80 @@ pt_bind_name_to_spec (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int *c
...
+  DB_VALUE values[3];
+  PT_DBLINK_INFO *dblink_table = &node->info.dblink_table;
+  int i;
+
+  assert (dblink_table->conn && dblink_table->conn->node_type == PT_NAME);
+
+  db_make_null (&(values[0]));
+  db_make_null (&(values[1]));
+  db_make_null (&(values[2]));
+  i = get_dblink_info_from_dbserver (parser, dblink_table->conn->info.name.original, values);
```
beyondykk9: variable "error" looks better than "i"

src/parser/name_resolution.c
```diff
@@ -10234,3 +10234,80 @@ pt_bind_name_to_spec (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int *c
...
+      if (i == ER_DBLINK_SERVER_NOT_FOUND)
+	{
+	  PT_ERRORmf (parser, node, MSGCAT_SET_PARSER_RUNTIME, MSGCAT_RUNTIME_RT_SERVER_NOT_DEFINED,
+		      dblink_table->conn->info.name.original);
+	}
+      else if (i == ER_DBLINK_CANNOT_UPDATE_SERVER)
+	{
+	  PT_ERRORmf (parser, node, MSGCAT_SET_PARSER_RUNTIME, MSGCAT_RUNTIME_RT_SERVER_ALTER_NOT_ALLOWED, 0);
+	}
+      else if (!pt_has_error (parser))
```
beyondykk9, line 10264: "if (pt_has_error (parser))" is right?
!pt_has_error means no error occurred.

ctshim: Yes, that's right.
If an error occurs but an error message is not set, you need to set an error message, so this is what you are asking.
In the returned function, it checks the necessity of adding it because it uses pt_has_error (parser) to check the error.

src/parser/name_resolution.c
ctshim: It's a reuse of a variable, I'll  fix it for readability.

src/parser/name_resolution.c
```diff
@@ -241,6 +241,7 @@ static PT_NODE *pt_get_attr_list_of_derived_table (PARSER_CONTEXT * parser, PT_M
 static void pt_set_attr_list_types (PARSER_CONTEXT * parser, PT_NODE * as_attr_list, PT_MISC_TYPE derived_table_type,
 				    PT_NODE * derived_table, PT_NODE * parent_spec);
 static PT_NODE *pt_count_with_clauses (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int *continue_walk);
+static void pt_resolve_dblink_server_name (PARSER_CONTEXT * parser, PT_NODE * node);
```
hgryoo: how about returning an error value?

src/parser/name_resolution.c
```diff
@@ -999,6 +1000,12 @@ pt_bind_scope (PARSER_CONTEXT * parser, PT_BIND_NAMES_ARG * bind_arg)
 	  else if (table->node_type == PT_DBLINK_TABLE)
 	    {
 	      assert (spec->info.spec.derived_table_type == PT_DERIVED_DBLINK_TABLE);
+	      if (table->info.dblink_table.is_name)
+		{
+		  pt_resolve_dblink_server_name (parser, table);
+		  if (pt_has_error (parser))
+		    return;
```
hgryoo: brace `{`, `}` for if-statement

src/parser/name_resolution.c
```diff
@@ -999,6 +1000,12 @@ pt_bind_scope (PARSER_CONTEXT * parser, PT_BIND_NAMES_ARG * bind_arg)
 	  else if (table->node_type == PT_DBLINK_TABLE)
 	    {
 	      assert (spec->info.spec.derived_table_type == PT_DERIVED_DBLINK_TABLE);
+	      if (table->info.dblink_table.is_name)
+		{
+		  pt_resolve_dblink_server_name (parser, table);
```
hgryoo, line 1005: Is it possible that pt_resolve_dblink_server_name() can be failed? If not, assert (pt_has_error (parser)); could be enough

src/parser/name_resolution.c
ctshim: OK, I will fix it. thank you.

src/parser/name_resolution.c
ctshim: OK, I agree.

src/parser/name_resolution.c
shparkcubrid: If i is not in ('ER_DBLINK_SERVER_NOT_FOUND', 'ER_DBLINK_CANNOT_UPDATE_SERVER') then error is not handled.

src/parser/view_transform.c
```diff
@@ -8856,21 +8856,14 @@ mq_class_lambda (PARSER_CONTEXT * parser, PT_NODE * statement, PT_NODE * class_,
 	  spec->info.spec.path_entities = NULL;
 	  if (newspec)
 	    {
-	      if (newspec->info.spec.derived_table_type == PT_DERIVED_JSON_TABLE)
+	      if ((newspec->info.spec.derived_table_type == PT_DERIVED_JSON_TABLE)
+		  || (newspec->info.spec.derived_table_type == PT_DERIVED_DBLINK_TABLE))
```
shparkcubrid: is  this routine necessary? Does removing it cause any problems?

src/query/execute_statement.c
```diff
@@ -18056,3 +18056,114 @@ do_create_server (PARSER_CONTEXT * parser, PT_NODE * statement)
 
   return error;
 }
+
+int
+get_dblink_info_from_dbserver (PARSER_CONTEXT * parser, const char *server, DB_VALUE * out_val)
```
hgryoo, line 18061: error-handling routine is not clear for me. It would better that either return the error directly or return it with a goto statement.

src/parser/name_resolution.c
ctshim: It is not a problem because the following functions are error-checking using pt_has_error (parser).
For this, I wrote PT_ERROR (parser, node, "Failed to obtain server information");

src/parser/view_transform.c
ctshim: I'm not even sure if that code must be there.
However, the code was written by imitating the JSON_TABLE processing routine.
We need to look a little more closely at what cases that code is handled in.

src/query/execute_statement.c
ctshim: It is unified by using the goto statement without separating the functions.

src/parser/name_resolution.c
```diff
@@ -10234,3 +10235,82 @@ pt_bind_name_to_spec (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int *c
...
+
+  assert (dblink_table->conn && dblink_table->conn->node_type == PT_NAME);
+
+  db_make_null (&(values[0]));
+  db_make_null (&(values[1]));
+  db_make_null (&(values[2]));
+  error = get_dblink_info_from_dbserver (parser, dblink_table->conn->info.name.original, values);
+  if (error != NO_ERROR)
+    {
+      if (i == ER_DBLINK_SERVER_NOT_FOUND)
```
beyondykk9:
```suggestion
      if (error == ER_DBLINK_SERVER_NOT_FOUND)
```

src/parser/name_resolution.c
```diff
@@ -10234,3 +10235,82 @@ pt_bind_name_to_spec (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int *c
...
+	{
+	  PT_ERRORmf (parser, node, MSGCAT_SET_PARSER_RUNTIME, MSGCAT_RUNTIME_RT_SERVER_NOT_DEFINED,
+		      dblink_table->conn->info.name.original);
+	}
+      else if (i == ER_DBLINK_CANNOT_UPDATE_SERVER)
```
beyondykk9:
```suggestion
      else if (error == ER_DBLINK_CANNOT_UPDATE_SERVER)
```

src/parser/view_transform.c
beyondykk9: I think that the condition is not necessary. the "newspec->info.spec.derived_table_type == PT_DERIVED_JSON_TABLE" condition is enough for JSON table only because the dblink table information exists at the remote server.

src/parser/name_resolution.c
shparkcubrid: I understand what you said.

src/parser/view_transform.c
ctshim: As a result of testing, it seems that there is no need to separate and process PT_DERIVED_DBLINK_TABLE.
I'll  fix the code.

src/query/execute_statement.c
```diff
@@ -18056,3 +18056,103 @@ do_create_server (PARSER_CONTEXT * parser, PT_NODE * statement)
...
+  server_object = do_get_server_obj_id (&server_obj_id, server_class, server_name);
+  if (server_object == NULL)
+    {
+      error = ER_DBLINK_SERVER_NOT_FOUND;
+      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, error, 1, server_name);
+      goto error_end;
+    }
+
+  AU_DISABLE (au_save);
+  is_au_disable = true;
```
hgryoo: It seems redundant. AU_ENABLE should be called anyway at the end of this function

ctshim: I didn't understand which part seems to be redundant.
If there is no error, ENABLE is processed in advance at line 18126.
This is to reduce the time to leave the DISABLE state if possible.

hgryoo: `is_au_disable` is redundant. just calling AU_ENABLE (au_save) is enough after `error_end:` label.

ctshim: Appropriate "au_save" values are required when performing AU_ENABLE().
au_save is obtained by performing AU_DISABLE().
However, in the code before AU_DISABLE (au_save) is executed, a "goto error_end" syntax may be encountered, but the value of au_save is not guaranteed in this case.
So we need a variable that indicates that the "au_save" value is guaranteed.

It is possible if the scope of leaving DISABLE state is expanded to the entire get_dblink_info_from_dbserver() function area.
I'll fix it like that (because it will run as a single process).

beyondykk9 (approved):

src/parser/name_resolution.c
ctshim: Fix already through 66aba5d.

src/parser/name_resolution.c
ctshim: Fix already through 66aba5d.

src/parser/csql_grammar.y
```diff
@@ -28382,4 +28390,25 @@ pt_fill_conn_info_container(PARSER_CONTEXT *parser,  int buffer_pos, container_1
...
+        if( strncasecmp(p, "SELECT", 6) == 0 )
+        {
+                if( p[6] == ' ' || p[6] == '\t' || p[6] == '\r' || p[6] == '\n' )
+                      return true;
+        }
+   }
+
+   sprintf(perr_msg, "Only SELECT statements are supported.");
+   return false;
 }
```
shparkcubrid: no new line at end of file.

shparkcubrid (dismissed):

src/parser/csql_grammar.y
```diff
@@ -28382,4 +28390,25 @@ pt_fill_conn_info_container(PARSER_CONTEXT *parser,  int buffer_pos, container_1
...
+{  
+   perr_msg[0] = 0x00;
+   while (*p == ' ' || *p == '(' || *p == '\t' || *p == '\r' || *p == '\n')
+     {
+        p++;
+     }
+
+   if(*p)
+   {
+        if( strncasecmp(p, "SELECT", 6) == 0 )
```
shparkcubrid, line 28405: This could be a problem with the sql below.
```
WITH table AS
(
  SELECT ...
)
SELECT * FROM table;
```
It is necessary to check the various sql usage.

src/parser/csql_grammar.y
ctshim: I will fix it. thank you

src/parser/csql_grammar.y
ctshim: First of all, only the case of starting with SELECT is supported.
The WITH clause is not supported.

shparkcubrid (approved):

hgryoo (approved):

youngjinj (approved):

PR #2879, opened 2021-07-06
[LETS-59] Recovery speedup by caching log pages in log page buffer

cristiarg:
http://jira.cubrid.org/browse/LETS-59

Function `logpb_copy_page` function first checks if the page is in log page buffer (`pageid` must match with the corresponding slot). If this does not match, then the page is read from disk (or, in the scalability setup, from the page server).
After reading the page from storage (disk or page server), logic attempts to save the page in the log page buffer using a "highest page id" criteria.

### DIFF
--- a/src/transaction/log_page_buffer.c
+++ b/src/transaction/log_page_buffer.c
@@ -754,14 +754,16 @@ logpb_invalidate_pool (THREAD_ENTRY * thread_p)
 
   /*
    * Flush any append dirty buffers at this moment.
-   * Then, invalidate any buffer that it is not fixed and dirty
    */
   logpb_flush_pages_direct (thread_p);
 
+  /*
+   * Invalidate any buffer that is not fixed.
+   */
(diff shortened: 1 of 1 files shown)


src/transaction/log_page_buffer.c
```diff
@@ -754,14 +754,16 @@ logpb_invalidate_pool (THREAD_ENTRY * thread_p)
 
   /*
    * Flush any append dirty buffers at this moment.
-   * Then, invalidate any buffer that it is not fixed and dirty
    */
   logpb_flush_pages_direct (thread_p);
 
+  /*
+   * Invalidate any buffer that is not fixed.
```
cristiarg: I am not sure about the "that is not fixed" part of the comment.

ribram: Yeah, the comment can be removed. There is no "fixing".

ribram (approved):

PR #2882, opened 2021-07-06
[CBRD-24003] When creating the view table, the IN operation is violated because the subquery's order by clause column is automatically added to select list.

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-24003

I fixed to set do_not_replace_orderby flag to 1 in the subquery using parser_walk_tree ()

### DIFF
--- a/src/parser/name_resolution.c
+++ b/src/parser/name_resolution.c
@@ -1104,6 +1104,7 @@ pt_mark_location (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int *conti
 PT_NODE *
 pt_set_is_view_spec (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int *continue_walk)
 {
+  bool do_not_replace_orderby = (bool *) arg ? *((bool *) arg) : false;
   if (!node)
     {
       return node;
@@ -1114,6 +1115,11 @@ pt_set_is_view_spec (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int *co
       /* Reset query id # */
       node->info.query.id = (UINTPTR) node;
       node->info.query.is_view_spec = 1;
+
+      if (do_not_replace_orderby)
+	{
+	  node->flag.do_not_replace_orderby = 1;
+	}
     }
 
   return node;
--- a/src/parser/name_resolution.c
+++ b/src/parser/name_resolution.c
@@ -1136,11 +1136,6 @@ pt_set_is_view_spec (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int *co
 PT_NODE *
 pt_set_do_not_replace_orderby (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int *continue_walk)
 {
-  if (!node)
-    {
-      return node;
-    }
-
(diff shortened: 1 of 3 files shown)


beyondykk9 (approved):

src/parser/name_resolution.c
```diff
@@ -1114,11 +1115,40 @@ pt_set_is_view_spec (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int *co
...
+{
+  if (!node)
+    {
+      return node;
+    }
+
+  if (pt_is_query (node))
+    {
+      node->flag.do_not_replace_orderby = 1;
+    }
```
hgryoo: `pt_is_query()` checks node is NULL.

```suggestion
  if (pt_is_query (node))
    {
      node->flag.do_not_replace_orderby = 1;
    }
```

hgryoo (approved):

src/parser/name_resolution.c
shparkcubrid: Okay. thanks.

youngjinj (approved):

src/parser/semantic_check.c
```diff
@@ -7984,7 +7985,9 @@ pt_check_create_view (PARSER_CONTEXT * parser, PT_NODE * stmt)
 
       /* TODO This seems to flag too many queries as view specs because it also traverses the tree to subqueries. It
        * might need a pre_function that returns PT_STOP_WALK for subqueries. */
-      result_stmt = parser_walk_tree (parser, crt_qry, pt_set_is_view_spec, NULL, NULL, NULL);
+      do_not_replace_orderby = true;
+      result_stmt =
+	parser_walk_tree (parser, crt_qry, pt_set_is_view_spec, (void *) &do_not_replace_orderby, NULL, NULL);
```
ctshim, line 7989: Before calling pt_semantic_check() at the bottom instead of the modified contents here
Can't we just add the code as if we modified the pt_check_vclass_query_spec() function?
In this way, it seems that the modification of the pt_set_is_view_spec() function is not necessary.

```
**(void) parser_walk_tree(parser, crt_qry, pt_set_do_not_replace_orderby, NULL, NULL, NULL);**
result_stmt = pt_semantic_check(parser, crt_qry);
```

shparkcubrid: To reduce calls to parser_walk_tree (), I used pt_set_is_view_spec.

ctshim: Yes, I understand.
Then, how about removing the pt_set_do_not_replace_orderby() function and modifying the code on line 13494 in the pt_validate_query_spec() function as follows?

s = parser_walk_tree(parser, s, pt_set_is_view_spec, NULL, NULL, NULL);
==> 
bool do_not_replace_orderby = true;
s = parser_walk_tree(parser, s, pt_set_is_view_spec, (void *) &do_not_replace_orderby, NULL, NULL);

ctshim (approved):

src/parser/semantic_check.c
shparkcubrid: I fixed it. thanks.

PR #2883, opened 2021-07-07
[LETS-135] Transaction server thinks it is connected to a page server that does not exist

adrianbrab:
http://jira.cubrid.org/browse/LETS-135
The communication setup protocol between TS and PS has a flaw. If cub_master on PS node works, but the page server doesn't, the transaction server wrongfully thinks it has a valid connection. The send_int that is supposed to confirm the connection does not actually confirm that the message is received.

Fixed the connection protocol by adding a confirmation message from the page server side. After the ATS sends cubcomm::server_server::CONNECT_ACTIVE_TRAN_TO_PAGE_SERVER to the page server it will wait to receive the same value as a confirmation. If the same value is not received the ATS-PS connection is not created.


src/connection/server_support.c
```diff
@@ -2645,6 +2645,11 @@ css_process_server_server_connect (SOCKET master_fd)
   switch (STATIC_CAST (cubcomm::server_server, request))
     {
     case cubcomm::server_server::CONNECT_ACTIVE_TRAN_TO_PAGE_SERVER:
+      if (!chn.send_int (static_cast < int >(cubcomm::server_server::CONNECT_ACTIVE_TRAN_TO_PAGE_SERVER)))
+	{
+	  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_NET_PAGESERVER_CONNECTION, 1, chn.get_channel_id ());
+	  break;
+	}
```
ribram: - Disable indent
- The error message is not a good fit. The message says: "Error connecting with the page server at $host". This is an error of the page server connection with the active transaction server. Create a new message (e.g. "Error connecting with the active transaction server").

```suggestion
      // *INDENT-OFF*
      if (!chn.send_int (static_cast<int> (cubcomm::server_server::CONNECT_ACTIVE_TRAN_TO_PAGE_SERVER)))
	{
	  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_NET_PAGESERVER_CONNECTION, 1, chn.get_channel_id ());
	  break;
	}
      // *INDENT-ON*
```

src/server/active_tran_server.cpp
```diff
@@ -247,6 +247,15 @@ active_tran_server::connect_to_page_server (const cubcomm::node &node, const cha
       return ER_NET_PAGESERVER_CONNECTION;
     }
 
+  int exit_code;
```
ribram: - "exit_code" is usually the name used for a value that is returned by the function or by the program. Use a more intuitive name (e.g. received_code or received_int).
- I think the second condition of failure is "code is not equal CONNECT_ACTIVE_TRAN_TO_PAGE_SERVER.
- Avoid using in the same conditional statement a value that is output by a function.
- To avoid repeating the same sequence of code that sets the error, add a lambda function (use it everywhere possible)
- er_log_debug after er_set is overkill :).

```c
auto ps_conn_error_lambda = [&node] ()
  {
    er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_NET_PAGESERVER_CONNECTION, 1, node.get_host ().c_str ());
    return ER_NET_PAGESERVER_CONNECTION;
  };
  
int returned_code;
if (srv_chn.recv_int (returned_code)!= css_error_code::NO_ERRORS)
  {
    return ps_conn_error_lambda ();
  }
if (returned_code != static_cast<int> (cubcomm::server_server::CONNECT_ACTIVE_TRAN_TO_PAGE_SERVER))
  {
    return ps_conn_error_lambda ();
  }
```

src/connection/server_support.c
adrianbrab: Done

src/connection/server_support.c
```diff
@@ -2645,6 +2645,13 @@ css_process_server_server_connect (SOCKET master_fd)
   switch (STATIC_CAST (cubcomm::server_server, request))
     {
     case cubcomm::server_server::CONNECT_ACTIVE_TRAN_TO_PAGE_SERVER:
+      // *INDENT-OFF*
+      if (!chn.send_int (static_cast < int >(cubcomm::server_server::CONNECT_ACTIVE_TRAN_TO_PAGE_SERVER)))
+	{
+	  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_NET_TRANSACTIONSERVER_CONNECTION, 1, chn.get_channel_id ());
```
ribram: Printing channel ID does not help inform the user. The one information that could be relevant is the transaction server hostname, but for now it is not known. The server_channel connection protocol may have to be modified to include the hostname of the connecter.

I see two possible options here:
  - Do not include "at %s" part in the error message and do not print get_channel_id().
  - Let "at %s" part in the error message, print "unknown" for now and leave a todo to be replaced by the hostname of the connecter host.

src/server/active_tran_server.cpp
```diff
@@ -247,6 +247,22 @@ active_tran_server::connect_to_page_server (const cubcomm::node &node, const cha
       return ER_NET_PAGESERVER_CONNECTION;
     }
 
+  auto ps_conn_error_lambda = [&node] ()
```
ribram: Move `ps_conn_error_lambda` definition to the top of the function and use in the two above error cases too.

src/base/error_code.h
```diff
@@ -1630,7 +1630,7 @@
 #define ER_NO_PAGE_SERVER_CONNECTION                -1269
 #define ER_TOOL_INVALID_WITH_REMOTE_STORAGE         -1270
 #define ER_COMMAND_INVALID_WITH_REMOTE_STORAGE      -1271
-#define ER_SCALABILITY_DEV_RESERVED_ERROR09         -1272
+#define ER_NET_TRANSACTIONSERVER_CONNECTION         -1272
```
ribram: `TRANSACTIONSERVER` is too long and unreadable.
```suggestion
#define ER_NET_TRANSACTION_SERVER_CONNECTION         -1272
```

adrianbrab: added the missing _

src/server/active_tran_server.cpp
adrianbrab: done

src/connection/server_support.c
adrianbrab: I'll leave out the %s for now.

src/connection/server_support.c
```diff
@@ -2645,6 +2645,13 @@ css_process_server_server_connect (SOCKET master_fd)
   switch (STATIC_CAST (cubcomm::server_server, request))
     {
     case cubcomm::server_server::CONNECT_ACTIVE_TRAN_TO_PAGE_SERVER:
+      // *INDENT-OFF*
+      if (!chn.send_int (static_cast < int >(cubcomm::server_server::CONNECT_ACTIVE_TRAN_TO_PAGE_SERVER)))
+	{
+	  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_NET_TRANSACTION_SERVER_CONNECTION, 1, chn.get_channel_id ());
```
ribram: Argument no longer needed.
```suggestion
	  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_NET_TRANSACTION_SERVER_CONNECTION, 0);
```

src/connection/server_support.c
```diff
@@ -2645,6 +2645,13 @@ css_process_server_server_connect (SOCKET master_fd)
   switch (STATIC_CAST (cubcomm::server_server, request))
     {
     case cubcomm::server_server::CONNECT_ACTIVE_TRAN_TO_PAGE_SERVER:
+      // *INDENT-OFF*
+      if (!chn.send_int (static_cast < int >(cubcomm::server_server::CONNECT_ACTIVE_TRAN_TO_PAGE_SERVER)))
```
ribram:
```suggestion
      if (!chn.send_int (static_cast<int> (cubcomm::server_server::CONNECT_ACTIVE_TRAN_TO_PAGE_SERVER)))
```

ribram (approved):
Please fix indentation and er_set argument & merge.

PR #2885, opened 2021-07-08
[LETS-158] Do not update minimum lsa monitoring during recovery

cristiarg:
http://jira.cubrid.org/browse/LETS-158

The mechanism to monitor the recovery/replication progress via monitoring the minimum unreplicated `log_lsa` is only of use for the replication scenario. Thus, dead weight, on the recovery scenario.

NOTE: this is preliminary modification that just disables it; a refactoring will be needed when the internal implementation of `redo_parallel::redo_job_queue` will change with the next PR;

### DIFF
--- a/src/transaction/log_recovery.c
+++ b/src/transaction/log_recovery.c
@@ -991,7 +991,6 @@ log_recovery_redo (THREAD_ENTRY * thread_p, log_recovery_context & context)
    */
   LOG_CS_EXIT (thread_p);
   // *INDENT-OFF*
-  std::unique_ptr<cublog::minimum_log_lsa_monitor> minimum_log_lsa;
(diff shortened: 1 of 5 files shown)


ribram (approved):

ribram:
m_in_progress_lsas should be disabled too. in this patch or in a separate patch, how do you prefer?

cristiarg:
> m_in_progress_lsas should be disabled too. in this patch or in a separate patch, how do you prefer?

@ribram You're very much right. My bad. Moreso that perf revealed that the underlying red-black tree implementation of the `set<log_lsa>` pops un quite a lot in the tops. HoweverI will do it as part of [LETS-157](http://jira.cubrid.org/browse/LETS-157).

PR #2886, opened 2021-07-08
[CBRD-23940] An invalid expression of assert function exists in the vacuum_rv_set_next_page_dropped_files function.

eido5:
http://jira.cubrid.org/browse/CBRD-23940

**Purpose**
A invalid expression of assert function should be changed a valid one.

**Implementation**
vacuum_rv_set_next_page_dropped_files() in vacuum.c

**Remarks**
N/A

### DIFF
--- a/src/query/vacuum.c
+++ b/src/query/vacuum.c
@@ -6737,7 +6737,7 @@ vacuum_rv_set_next_page_dropped_files (THREAD_ENTRY * thread_p, LOG_RCV * rcv)
   VPID_COPY (&page->next_page, (VPID *) rcv->data);
 
   /* Check recovery data is as expected */
-  assert (rcv->length = sizeof (VPID));
+  assert (rcv->length == sizeof (VPID));
 
   vacuum_er_log (VACUUM_ER_LOG_RECOVERY, "Set link for dropped files from page %d|%d to page %d|%d.",
 		 pgbuf_get_vpid_ptr (rcv->pgptr)->pageid, pgbuf_get_vpid_ptr (rcv->pgptr)->volid, page->next_page.volid,


wrlawodms (approved):
Good. How about putting the assertion above `VPID_COPY (&page->next_page, (VPID *) rcv->data);` though?. It looks more natural for a pre-condition check to be before using it.

joohok:
> Good. How about putting the assertion above `VPID_COPY (&page->next_page, (VPID *) rcv->data);` tough?. It looks more natural for a pre-condition check to be before using it.

I agree with it. Checking the condition before VPID_COPY is more reasonable.

joohok (approved):

eido5:
> Good. How about putting the assertion above `VPID_COPY (&page->next_page, (VPID *) rcv->data);` though?. It looks more natural for a pre-condition check to be before using it.

Thank you for comments!

I also think it looks like natural intuitively. But it does not make me understand reasonable. Shall we understand the reason why it should be put above the `VPID_COPY (&page->next_page, (VPID *) rcv->data);` . 

I think I cannot understand correctly about this issue, so assertion may be the reasonable position to check the condition after the  `VPID_COPY (&page->next_page, (VPID *) rcv->data);` ; below the `VPID_COPY (&page->next_page, (VPID *) rcv->data);`

wrlawodms:
@eido5  `rcv->length` is the size of `rcv->data`, and they are put into when the log record was written. The assertion checks whether the log record includes expected input or not while recovery. I think the checking should be done before using it. By the way, the is not related to this issue directly and not serious. You don't have to do it in this PR if you don't follow it.

eido5:
@wrlawodms Thank you for the explaining what I wonder!
At this time, we should not fix it right now.

hornetmj (approved):

PR #2889, opened 2021-07-09
[LETS-148] Replicate unique index statistics

ribram:
http://jira.cubrid.org/browse/LETS-148

Changing statistics in the b-tree root page on every operation is very expensive. The transaction server is optimized to update these statistics only in-memory and log its changes. The statistics are flushed to b-tree root page during checkpoint.

The redo recovery follows the same system. Statistics are updated only in memory and they are flushed during checkpoint (recovery always ends with a checkpoint).

On the page server however, the replication has to diverge from the redo recovery system. B-tree root pages may be fetched at any time and the statistics should be there. Therefore, the replication has to replicate statistics changes before the page is fetched.

The easiest option is to replicate the changes immediately, when the log record is found.

**Implementation**

- Add special handling for RVBT_LOG_GLOBAL_UNIQUE_STATS_COMMIT (replicator::read_and_redo_btree_stats). It gets btid & stats from redo data and applies the changes.
- Add redo_job_btree_stats to parallelize b-tree stats replication. The VPID is obtained from the btid read from redo data.
- Add replicate_btree_stats, to apply the stats change and set the new page LSA.

Other:

- Move redo_job_replication_delay_impl to log_replication.cpp
- Rewrite btree_reflect_global_unique_statistics. Remove unused only_active_tran argument. Extract the code that changes stats in the b-tree root header to btree_root_update_stats
- Expose log_rv_redo_fix_page for log_replication.cpp

**Tests**

- Passed shell_debug
- Manual test on replication


src/storage/btree.c
```diff
@@ -14048,92 +14072,50 @@ btree_update (THREAD_ENTRY * thread_p, BTID * btid, DB_VALUE * old_key, DB_VALUE
...
-btree_reflect_global_unique_statistics (THREAD_ENTRY * thread_p, GLOBAL_UNIQUE_STATS * unique_stat_info,
-					bool only_active_tran)
+btree_reflect_global_unique_statistics (THREAD_ENTRY * thread_p, GLOBAL_UNIQUE_STATS * unique_stat_info)
 {
-  VPID root_vpid;
-  PAGE_PTR root = NULL;
-  BTREE_ROOT_HEADER *root_header = NULL;
-  int ret = NO_ERROR;
-  LOG_LSA *page_lsa = NULL;
+  assert (unique_stat_info != NULL);
```
cristiarg, line 14077: Old code returned an error code if null. New code will crash.

src/transaction/log_replication.cpp
```diff
@@ -1,332 +1,498 @@
...
+	return ER_FAILED;
+      }
+  }
+
+  /*********************************************************************
+   * replication b-tree unique statistics - declaration
+   *********************************************************************/
+
+  void
+  replicate_btree_stats (cubthread::entry &thread_entry, const VPID &root_vpid, const log_unique_stats &stats,
```
cristiarg, line 463: Function is not specific to replication, maybe it should be moved to btree module with a proper name.

src/transaction/log_replication.cpp
```diff
@@ -1,332 +1,498 @@
...
+    const char *datap = rcv.data;
+    OR_GET_BTID (datap, &btid);
+    datap += OR_BTID_ALIGNED_SIZE;
+    stats.num_keys = OR_GET_INT (datap);
+    datap += OR_INT_SIZE;
+    stats.num_oids = OR_GET_INT (datap);
+    datap += OR_INT_SIZE;
+    stats.num_nulls = OR_GET_INT (datap);
+    datap += OR_INT_SIZE;
+    VPID root_vpid = { btid.root_pageid, btid.vfid.volid };
```
cristiarg, line 296: A template function should be extracted up to here and moved to `log_recovery_redo.hpp`

cristiarg (approved):

src/storage/btree.c
ribram: It's intentional. Actually the function never gets a null pointer.

src/transaction/log_replication.cpp
ribram: Please make the suggested change more specific. What is the signature of the template function and what is its purpose (in what context should it be reused)?

src/transaction/log_replication.cpp
ribram: The function is specific to replication, because it does some stuff differently:

- A redo specific function is used to fix the page - `log_rv_redo_fix_page`; b-tree normally uses pgbuf_fix with OLD_PAGE fetch mode.
- The page LSA is set after applying the change. Normally, the page LSA is auto-updated when the log record is generated.

You may argue the function is specific to both replication and b-tree and you'd be right :). But between the two, I would choose replication as destination for this function because it is better scoped than btree.c, where everything even remotely linked to b-tree was dumped.

src/transaction/log_replication.cpp
cristiarg: It looks to me that the section of code that extracts `log_unique_stats` and `BTID` is very much on the same functional level as the rest of the `log_rv_get_` functions in `log_recovery_redo.hpp` (particularly `log_rv_get_log_rec_redo_data`).
Although it does not have any reusability bound to it, it makes the code structure easier to reason about.
So (pseudocode):
```
template<typename T>
log_unique_stats
log_rv_get_btree_unique_stats(THREAD_ENTRY *thread_p, log_reader &log_pgptr_reader, const T &log_rec,
				  log_rcv &rcv, log_rectype log_rtype, struct log_zip &undo_unzip_support, struct log_zip &redo_unzip_support, log_unique_stats &out_log_unique_stats, BTID &out_btid);

```

src/transaction/log_replication.cpp
cristiarg: Fair.

cristiarg (approved):

src/transaction/log_replication.cpp
ribram: I see the template functions in log_recovery_redo.hpp as generally applicable to all log records of a given type. The log_rv_get_btree_unique_stats function is applicable only to a specific recovery index and only to replication, not redo.

One possible improvement would be to group the redo data reading part into a function:
```c
void
btree_rv_read_stats_data (const char *data, BTID &btid, log_unique_stats &stats)
{
    BTID btid;
    log_unique_stats stats;
    const char *datap = data;
    OR_GET_BTID (datap, &btid);
    datap += OR_BTID_ALIGNED_SIZE;
    stats.num_keys = OR_GET_INT (datap);
    datap += OR_INT_SIZE;
    stats.num_oids = OR_GET_INT (datap);
    datap += OR_INT_SIZE;
    stats.num_nulls = OR_GET_INT (datap);
    datap += OR_INT_SIZE;
}
```
This chunk of code is repeated in three other functions (btree_rv_update_tran_stats, btree_rv_undo_global_unique_stats_commit and btree_rv_redo_global_unique_stats_commit).

My suggestion is to add btree_rv_read_stats_data and use it everywhere applicabe. read_and_redo_btree_stats becomes:
```c
    // Get redo data and read it
    LOG_RCV rcv;
    rcv.length = log_rv_get_log_rec_redo_length<T> (log_rec);
    if (log_rv_get_log_rec_redo_data (&thread_entry, m_reader, log_rec, rcv, rectype, m_undo_unzip, m_redo_unzip)
	!= NO_ERROR)
      {
	logpb_fatal_error (&thread_entry, true, ARG_FILE_LINE, "replicator::read_and_redo_btree_stats");
	return;
      }
    
    BTID btid;
    log_unique_stats stats;
    btree_rv_read_stats_data (rcv.data, btid, stats);
    VPID root_vpid = { btid.root_pageid, btid.vfid.volid };
    
    // Create a job or apply the change immediately
    if (m_parallel_replication_redo)
      {
	auto job = std::make_unique<redo_job_btree_stats> (root_vpid, rec_lsa, stats);
	m_parallel_replication_redo->add (std::move (job));
      }
    else
      {
	replicate_btree_stats (thread_entry, root_vpid, stats, rec_lsa);
      }
```

What do you think?

PR #2890, opened 2021-07-09
[LETS-160] Redoing unreserve sector doesn't wait for redoing other log records

ribram:
http://jira.cubrid.org/browse/LETS-160

As a consequence of the fact that redoing log records is no longer skipped if the page belongs to a not-reserved sector, un-reserving a sector does not have to wait for the log records to be applied.

**Implementation**
- Remove the `wait_for_idle()` for RVDK_UNRESERVE_SECTORS
- Remove safe-guard and update comment because reserve/unreserve are no longer required to by applied on sync; left it as sync because it is offers better control and possibly better performance.
- Don't do valid check when pages are fetched for recovery; they may be unreserved when redo is applied.
- Re-written SERVER_MODE/SA_MODE separation to make each flow more readable.

### DIFF
--- a/src/transaction/log_recovery.c
+++ b/src/transaction/log_recovery.c
@@ -507,8 +507,7 @@ log_rv_need_sync_redo (const vpid & a_rcv_vpid, LOG_RCVINDEX a_rcvindex)
       return true;
     case RVDK_RESERVE_SECTORS:
     case RVDK_UNRESERVE_SECTORS:
-      // When sectors are unreserved, redo ops on these pages are not applied.
-      // Sector reservation is handled synchronously for better control
+      // Sector reservation is handled synchronously for better control; may be changed to async
       return true;
     default:
       return false;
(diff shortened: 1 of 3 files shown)


src/transaction/log_recovery.c
```diff
@@ -507,8 +507,7 @@ log_rv_need_sync_redo (const vpid & a_rcv_vpid, LOG_RCVINDEX a_rcvindex)
       return true;
     case RVDK_RESERVE_SECTORS:
     case RVDK_UNRESERVE_SECTORS:
-      // When sectors are unreserved, redo ops on these pages are not applied.
-      // Sector reservation is handled synchronously for better control
+      // Sector reservation is handled synchronously for better control; may be changed to async
```
cristiarg, line 510: Should applying `RVDK_UNRESERVE_SECTORS` in parallel be tested?

cristiarg (approved):

src/transaction/log_recovery.c
ribram: I tested it.

PR #2891, opened 2021-07-09
[LETS-157] Parallel recovery performance: refactor raw queues with maps of vpids

cristiarg:
http://jira.cubrid.org/browse/LETS-157

In the parallel recovery redo infrastructure, jobs (aka: log record entries) are applied (consumed) one by one. This leads to high contention on the consume side.
Refactor such that jobs are consumed with less contention. The solution adopted by this patch is to consume in bundles, per vpid (aka: per page) without any affinity between vpid and the executing thread.

Implementation:
- replace raw deques with maps of vpids and such deques;
- refactor bookkeeping of minimum unreplicated log lsa with the help of an indirection map that keeps the minimum lsa corresponding to a vpid; this map also has the role of allowing to always dispatch for execution the vpid whose lsa is the smallest; the end goal is to avoid the currently kept log lsa from jumping back and forth which would compromise the bookkeeping on the replication side;
- minimum unreplicated log lsa is only computed if requested


src/storage/vpid_utilities.hpp
```diff
@@ -56,6 +56,18 @@ namespace std
...
+  template<>
+  struct hash<VPID>
+  {
+    inline std::size_t operator () (const VPID &val) const noexcept
+    {
+      std::size_t res = 0;
+      res ^= std::hash<short> {} (val.volid) + 0x9e3779b9 + (res << 6) + (res >> 2);
+      res ^= std::hash<int32_t> {} (val.pageid) + 0x9e3779b9 + (res << 6) + (res >> 2);
+      return res;
+    }
```
ribram: Please explain this hash formula.

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -268,21 +289,25 @@ namespace cublog
...
-	  std::swap (m_produce_queue, m_consume_queue);
-	  // if both queues empty, notify the, possibly waiting, producing thread
-	  m_queues_empty = m_produce_queue->empty () && m_consume_queue->empty ();
-	  notify_queues_empty = m_queues_empty;
+	  std::lock_guard<std::mutex> produce_lockg (m_produce_mutex);
+	  std::swap (m_produce, m_consume);
+	  m_produce_min_lsa_map.swap (m_consume_min_lsa_map);
 
-	  if (m_minimum_log_lsa != nullptr)
+	  assert (m_produce->size () == m_produce_min_lsa_map.size () && m_produce->empty ());
```
```suggestion
	  assert (m_produce->empty () && m_produce_min_lsa_map.empty ());
```

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -291,92 +316,119 @@ namespace cublog
...
+	    m_minimum_log_lsa->set_for_consume_and_in_progress (
+		    consume_minimum_log_lsa, in_progress_minimum_log_lsa);
 	  }
+
+	return ret_job_deq;
       }
 
     // consumer task will have to spin-wait
-    return nullptr;
+    return ux_redo_job_deque ();
```
ribram: If the min LSA vpid is in progress, all threads except the one redo-ing the changes on that vpid will spin. Is this the desired behavior?

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -291,92 +316,119 @@ namespace cublog
...
+	  assert (m_in_progress_vpids.size () == m_in_progress_lsas.size ());
+	  assert ((vpid_set_empty && lsa_set_empty) || (!vpid_set_empty && !lsa_set_empty));
 	}
+      lsa_set_empty = m_in_progress_lsas.empty ();
 
-      assert (m_in_progress_vpids.size () == m_in_progress_lsas.size ());
     }
-    if (set_empty)
+
+    if (vpid_set_empty && lsa_set_empty)
```
ribram: if `m_monitor_minimum_log_lsa == nullptr`, lsa_set_empty is always true. If `m_monitor_minimum_log_lsa != nullptr`, lsa_set_empty is always equial to vpid_set_empty. It seems this conditional does not need it. Is this right?

src/transaction/log_recovery_redo_parallel.hpp
```diff
@@ -158,8 +159,10 @@ namespace cublog
       {
 	  using ux_redo_job_base = std::unique_ptr<redo_job_base>;
 	  using ux_redo_job_deque = std::deque<ux_redo_job_base>;
+	  using vpid_ux_redo_job_deque_map_t = std::unordered_map<vpid, ux_redo_job_deque, std::hash<VPID>>;
```
ribram: I think `ux_` was used for unique_pointer right? Probably the type names need an update as `redo_job_deque` and `vpid_redo_job_deque_map_t`?

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -540,15 +597,15 @@ namespace cublog
     for (; !finished ;)
       {
 	bool adding_finished = false;
-	std::unique_ptr<redo_job_base> job = m_queue.pop_job (adding_finished);
+	std::deque<std::unique_ptr<redo_job_base>> job_deque = m_queue.pop_job (adding_finished);
```
```suggestion
	redo_job_queue::ux_redo_job_deque job_deque = m_queue.pop_job (adding_finished);
```

src/transaction/log_recovery_redo_parallel.hpp
```diff
@@ -216,25 +219,38 @@ namespace cublog
...
 	private:
-	  /* two queues are internally managed and take turns at being either
+	  /* two maps of queues are internally managed and take turns at being either
 	   * on the producing or on the consumption side
 	   */
-	  ux_redo_job_deque *m_produce_queue;
-	  mutable std::mutex m_produce_queue_mutex;
-	  ux_redo_job_deque *m_consume_queue;
-	  std::mutex m_consume_queue_mutex;
+	  vpid_ux_redo_job_deque_map_t *m_produce;
```
```suggestion
	  vpid_ux_redo_job_deque_map_t *m_produce_jobs;
```

src/transaction/log_recovery_redo_parallel.hpp
```diff
@@ -216,25 +219,38 @@ namespace cublog
...
+	  /* contains, for each entry in 'm_produce', the log_lsa
+	   * of the first job (aka, the minimum), thus maintaining a parallel bookeeping
+	   * that will allow, on the consume side, to estimate, not precisely, but
+	   * conservatively correct, where consumption is at
+	   */
+	  log_lsa_vpid_map_t m_produce_min_lsa_map;
+	  mutable std::mutex m_produce_mutex;
+
+	  vpid_ux_redo_job_deque_map_t *m_consume;
```
```suggestion
	  vpid_ux_redo_job_deque_map_t *m_consume_jobs;
```

src/transaction/log_recovery_redo_parallel.cpp
cristiarg: In a way, yes, all threads that come to 'ask' for work and get into this situation will have to spin-wait until the min-lsa vpid is done.
The reason is explained at the beginning of the function.
There might be a better way to handle this corner case. I did not do any profiling to see how often this occurs in the particular load scenario used currently.

src/transaction/log_recovery_redo_parallel.cpp
cristiarg: yes, too defensive; removed

src/transaction/log_recovery_redo_parallel.hpp

src/transaction/log_recovery_redo_parallel.hpp
cristiarg: :) I simplified it because in one refactoring it is a `deque` in another one it will be a `map`, aso.

src/transaction/log_recovery_redo_parallel.hpp

src/transaction/log_recovery_redo_parallel.hpp

src/storage/vpid_utilities.hpp
cristiarg: Added a comment. It is based on Boost `hash_combine` implementation.

cristiarg:
Changes in this PR are no longer relevant due to low performance..

PR #2893, opened 2021-07-13
[LETS-140] Extend cub_server arguments with server type argument

adrianbrab:
http://jira.cubrid.org/browse/LETS-140

Add -t or --type cub_server optional argument for server type. "transaction" and "page" are the valid values.
Usage: 
cub_server -t transaction dbname
            cub_server --type transaction dbname
            cub_server dbname

This overwrites the type specified in cubrid.conf.


src/base/server_type.cpp
```diff
@@ -104,6 +107,37 @@ bool is_tran_server_with_remote_storage ()
...
+      return;
+    }
+
+  int arg = 0;
+  //check for server type argument
+  for (size_t i = 2; i < argc; i++)
+    {
+      if (arg)
+	{
+	  if (!strcmp (argv[i], "transaction"))
```
cristiarg:
```suggestion
	  if (strcmp (argv[i], "transaction") == 0)
```

src/base/server_type.cpp
```diff
@@ -104,6 +107,37 @@ bool is_tran_server_with_remote_storage ()
...
+	    {
+	      g_server_type = SERVER_TYPE_TRANSACTION;
+	      return;
+	    }
+	  if (!strcmp (argv[i], "page"))
```
cristiarg:
```suggestion
	  if (strcmp (argv[i], "page") == 0)
```
Does lowercase comparison conform to cubrid conventions regarding argument casing? I checked and `cubrid sErViCe StarT` works (`cubrid server start SCALDB` does not).

src/base/server_type.cpp
```diff
@@ -104,6 +107,37 @@ bool is_tran_server_with_remote_storage ()
...
+  int arg = 0;
+  //check for server type argument
+  for (size_t i = 2; i < argc; i++)
+    {
+      if (arg)
+	{
+	  if (!strcmp (argv[i], "transaction"))
+	    {
+	      g_server_type = SERVER_TYPE_TRANSACTION;
+	      return;
```
cristiarg:
```suggestion
	      break;
```

src/base/server_type.cpp
```diff
@@ -104,6 +107,37 @@ bool is_tran_server_with_remote_storage ()
...
+	    }
+	  if (!strcmp (argv[i], "page"))
+	    {
+	      g_server_type = SERVER_TYPE_PAGE;
+	      return;
```
cristiarg:
```suggestion
	      break;
```

src/base/server_type.cpp
```diff
@@ -104,6 +107,37 @@ bool is_tran_server_with_remote_storage ()
...
 }
 
+void set_server_type_from_arg (int argc, char **argv)
+{
+  if (argc < 4)
+    {
+      return;
+    }
+
+  int arg = 0;
```
cristiarg: Bool if variable is a flag.
```suggestion
  bool server_type_mnemonic_found = false;
```

src/base/server_type.cpp
```diff
@@ -104,6 +107,37 @@ bool is_tran_server_with_remote_storage ()
...
+	      g_server_type = SERVER_TYPE_TRANSACTION;
+	      return;
+	    }
+	  if (!strcmp (argv[i], "page"))
+	    {
+	      g_server_type = SERVER_TYPE_PAGE;
+	      return;
+	    }
+	}
+      if (!strcmp (argv[i], COMMDB_SERVER_TYPE_S) || !strcmp (argv[i], COMMDB_SERVER_TYPE_L))
```
cristiarg: Casing?

src/base/server_type.hpp
```diff
@@ -19,16 +19,21 @@
...
 #define _SERVER_TYPE_H_
 
+#define COMMDB_SERVER_TYPE_S              "-t"
+#define COMMDB_SERVER_TYPE_L              "--type"
+
 typedef enum
 {
   SERVER_TYPE_TRANSACTION,
   SERVER_TYPE_PAGE,
+  UNKNOWN,
```
cristiarg: It is more established that the neutral value be placed first.

src/base/server_type.hpp
```diff
@@ -19,16 +19,21 @@
...
 } SERVER_TYPE;
 
 int init_server_type (const char *db_name);
 void finalize_server_type ();
 SERVER_TYPE get_server_type ();
 
 bool is_tran_server_with_remote_storage ();
+void set_server_type_from_arg (int argc, char **argv);
```
cristiarg: If there's no intention of changing the argv values:
```suggestion
void set_server_type_from_arg (int argc, const char *const *const argv);
```

src/transaction/log_append.cpp
```diff
@@ -1735,7 +1735,7 @@ prior_list_deserialize (const std::string &str, log_prior_node *&head, log_prior
 static void
 prior_lsa_start_append (THREAD_ENTRY *thread_p, LOG_PRIOR_NODE *node, LOG_TDES *tdes)
 {
-  assertm (get_server_type () == SERVER_TYPE_TRANSACTION, "Log append can be executed only on transaction server");
+  assertm (get_server_type () != SERVER_TYPE_PAGE, "Log append can be executed only on transaction server");
```
cristiarg: I do not understand the context in which this function is called, but at least the text should be updated.

src/executables/server.c
```diff
@@ -348,6 +349,7 @@ main (int argc, char **argv)
     (void) envvar_bindir_file (executable_path, PATH_MAX, binary_name);
     /* save database name */
     database_name = argv[1];
+    set_server_type_from_arg (argc, argv);
```
cristiarg: There should be some error handling here as per:
```
cub_server accepts new argument -t or --type followed by one of the values "transaction" or "page". any other value will result in an error.
```

src/transaction/log_append.cpp
adrianbrab: I changed the logic so it would work on servers that start as UNKNOWN, the text can be changed to `Log append can not be executed only on page server`

src/executables/server.c
```diff
@@ -65,6 +67,7 @@ static void register_fatal_signal_handler (int signo);
 static void crash_handler (int signo, siginfo_t * siginfo, void *dummyp);
 #if !defined (NDEBUG)
 static void abort_handler (int signo, siginfo_t * siginfo, void *dummyp);
+static int argument_handler (int argc, char **argv);
```
ribram: The arguments are the same for debug & release and for Linux & windows. `argument_handler` function should be declared outside `!defined (NDEBUG)` and outside `defined (WINDOWS)`.

src/executables/server.c
```diff
@@ -305,6 +308,69 @@ abort_handler (int signo, siginfo_t * siginfo, void *dummyp)
...
+      database_name = argv[optind];
+    }
+  else if (argc > optind)
+    {
+      er_log_debug(ARG_FILE_LINE, "Argument '%s' is not needed for cub_server", argv[optind + 1]); // exit
+    }
+  else
+    {
+      er_log_debug(ARG_FILE_LINE, "A database-name is missing."); // exit
+    }
```
ribram: A generic invalid argument message may be used here:
```suggestion
  else
    {
      util_log_write_errid (MSGCAT_UTIL_GENERIC_INVALID_ARGUMENT);
      return ER_FAILED;
    }
```

ribram (approved):

PR #2900, opened 2021-07-15
[CBRD-23939] The system parameter 'read_only' should be the hidden pa…

eido5:
http://jira.cubrid.org/browse/CBRD-23939

**Purpose**
The system parameter 'read_only' is not a hidden property. So it should be a hidden property. 

**Implementation**
read_only system parameter in system_parameter.c 
I added PRM_HIDDEN parameter to hide "read_only" property

**Remarks**
N/A


eido5:
I did this mistake again. This PR must be closed

PR #2903, opened 2021-07-15
[CBRD-23977] Encryption processing for DBLINK password.

ctshim:
http://jira.cubrid.org/browse/CBRD-23977

Purpose
Encryption processing for DBLINK password.

Implementation
  I refer to the two functions below
 ` tde_encrypt_internal(), tde_decrypt_internal()`
 
Remarks
N/A


src/query/execute_statement.c
```diff
@@ -18135,7 +18168,7 @@ get_dblink_info_from_dbserver (PARSER_CONTEXT * parser, const char *server, DB_V
 	  sprintf (dblink_url, dblink_url_fmt_none, host, port_no, dbname);
 	}
 
-      error = db_make_string (&(out_val[0]), dblink_url);
+      error = db_make_string_copy (&(out_val[0]), dblink_url);
```
ctshim, line 18221: This is a bug fix, not a feature change.
There was a problem that it was not guaranteed when exiting the function when using a local array.

src/query/execute_statement.c
```diff
@@ -18148,3 +18195,311 @@ get_dblink_info_from_dbserver (PARSER_CONTEXT * parser, const char *server, DB_V
...
+	}
+    }
+  while (++pwdlen % 3);
+
+  return NO_ERROR;
+}
+
+#ifndef NDEBUG
+void
+ciper_func_test (PARSER_CONTEXT * parser)
```
ctshim: ciper_func_test() is a temporary test function and will be deleted in the future.
The code that calls this function will also be deleted.

src/transaction/boot_cl.c
```diff
@@ -5590,8 +5590,16 @@ boot_define_view_db_server (void)
...
+	   "SELECT [ds].[link_name], [ds].[host], [ds].[port], [ds].[db_name], [ds].[user_name], [ds].[properties],"
+	   "       [ds].[owner].[name], [ds].[comment]"
+	   " FROM [%s] [ds] WHERE CURRENT_USER = 'DBA' "
+	   "  OR {[ds].[owner].[name]} SUBSETEQ (SELECT SET{CURRENT_USER} + COALESCE(SUM(SET{[t].[g].[name]}), SET{})"
+	   "        FROM [%s] [u], TABLE([groups]) AS [t]([g]) WHERE [u].[name] = CURRENT_USER)"
+	   "  OR {[ds]} SUBSETEQ ( SELECT SUM(SET{[au].[class_of]}) FROM [%s] [au] WHERE {[au].[grantee].[name]} "
+	   "        SUBSETEQ ( SELECT SET{CURRENT_USER} + COALESCE(SUM(SET{[t].[g].[name]}), SET{}) "
+	   "                   FROM [%s] [u], TABLE([groups]) AS [t]([g])"
+	   "                   WHERE [u].[name] = CURRENT_USER) AND [au].[auth_type] = 'SELECT')", CT_DB_SERVER_NAME,
+	   CT_USER_NAME, CT_CLASSAUTH_NAME, CT_USER_NAME);
```
ctshim, line 5602: Change the creation syntax of DB_SERVER VIEW
Changed so that members of the same group can also access it.

beyondykk9:
```
# csql -u dba demodb
csql> create server rem_svr (
    host=locahost,
    port=33000,
    dbname=somedb,
    user=public,
    password='abcdef'
);

Execute OK. (0.001232 sec) Committed.
csql> ;ex

# cubrid server stop demodb

csql S -u dba demodb
csql> select * from dblink (rem_svr, 'SELECT col1 FROM t1 ') as r(col1 int);

In the command from line 13,

ERROR: Semantic: The checksum of the encrypted password does not match.
```
the above query sequences raise a semantic error.
vice-versa also raise the same error.

src/compat/db_admin.c
```diff
@@ -951,7 +951,21 @@ db_restart_ex (const char *program, const char *db_name, const char *db_user, co
       db_set_preferred_hosts (preferred_hosts);
     }
 
+#if !defined(CS_MODE)
   return db_restart (program, false, db_name);
+#else
+  tde_Cipher.is_loaded = false;
+  if ((retval = db_restart (program, false, db_name)) == NO_ERROR)
+    {
+      // Even if tde_get_data_keys() fails, it is executed normally.
```
beyondykk9: What does this comment mean?

src/query/crypt_opfunc.c
```diff
@@ -697,3 +699,359 @@ crypt_generate_random_bytes (char *dest, int length)
...
+      extern int tde_get_data_keys ();	// declared in "network_interface_cl.c"
+      if ((err = tde_get_data_keys ()) != NO_ERROR)
+	{
+	  return err;
+	}
+      tde_Cipher.is_loaded = true;
+    }
+  memcpy (evp_cipher.master_key, tde_Cipher.data_keys.log_key, TDE_MASTER_KEY_LENGTH);
+#endif
+
```
beyondykk9, line 737:
```suggestion
memcpy (evp_cipher.master_key, tde_Cipher.data_keys.log_key, TDE_MASTER_KEY_LENGTH);
```
it can be separated from CS_MODE definition together with prior line 772

src/communication/network_interface_cl.c
```diff
@@ -1840,7 +1840,7 @@ file_apply_tde_to_class_files (const OID * class_oid)
 #endif /* !CS_MODE */
 }
 
-#ifdef UNSTABLE_TDE_FOR_REPLICATION_LOG
+#if defined(UNSTABLE_TDE_FOR_REPLICATION_LOG) || defined(CS_MODE)
```
hgryoo: If `defined(CS_MODE)` should be added, we can remove this line.

src/compat/db_admin.c
ctshim: Even if the TDE value is not obtained normally, it means that the error handling is not done to ensure that the basic function can be performed.

See below page
"https://www.cubrid.org/manual/en/11.0/security.html#tde"


`When TDE is unavailable
In the following cases, the TDE feature cannot be used, and an error occurs because the TDE module cannot be loaded correctly.

When the valid key file cannot be found

When the key set on the database cannot be found in the key file

Even if the TDE module is not loaded, the server can start normally, and users can access unencrypted tables. This means that all DML and DDL such as SELECT and INSERT only for TDE-encrypted tables cannot be executed.

However, the case log data has been encrypted is different. If the log data is encrypted when the TDE module is not loaded and the log is accessed by recovery, HA, VACUUM, etc., the system cannot be properly executed, and the entire server has no option but to stop running the server.`

src/query/crypt_opfunc.c
ctshim: In case of CS_MODE, the value is copied every time,
In case of SA_MODE(= not CS_MODE), it is copied only once initially.

src/communication/network_interface_cl.c
ctshim: Do you mean "#if defined(UNSTABLE_TDE_FOR_REPLICATION_LOG) || defined(CS_MODE)"  same as  "#if 1" ?

shparkcubrid: Is this a function that hasn't been used before? 
Why is UNSTABLE_TDE_FOR_REPLICATION_LOG written?

src/compat/db_admin.c
shparkcubrid: Should the case when the TDE module is not loaded also be handled in this issue?

src/compat/db_admin.c
```diff
@@ -965,6 +979,10 @@ db_shutdown (void)
 {
   int error = NO_ERROR;
 
+#if defined(CS_MODE)
+  tde_Cipher.is_loaded = false;
```
shparkcubrid: Is this a bug fix in TDE?

src/query/crypt_opfunc.c
beyondykk9: Ok, I understood.

src/communication/network_interface_cl.c
ctshim: @shparkcubrid 
UNSTABLE_TDE_FOR_REPLICATION_LOG is a declaration to keep the code developed by the 2-Team.
It has not been completed, and so far, there are no plans to activate it in the future.

src/compat/db_admin.c
ctshim: @shparkcubrid
If the TDE module is not loaded, DBLINK's encryption processing cannot work. However, it is intended to enable behavior other than DBLINK.

src/compat/db_admin.c
ctshim: No,  This is only required for DBLINK.

src/query/execute_statement.c
```diff
@@ -181,6 +182,15 @@ static void init_compile_context (PARSER_CONTEXT * parser);
 
 static int do_select_internal (PARSER_CONTEXT * parser, PT_NODE * statement, bool for_ins_upd);
 
+static int get_dblink_password_encrypt (PARSER_CONTEXT * parser, const char *passwd, DB_VALUE * encrypt_val,
+					bool is_parser);
+static int get_dblink_password_decrypt (PARSER_CONTEXT * parser, const char *passwd_cipher, DB_VALUE * decrypt_val);
+static int shake_4_encrypt (const char *passwd, char *confused, struct timeval *chk_time);
+static int reverse_shake_4_decrypt (char *confused, char *passwd);
```
hgryoo: I don't think these functions are placed in the proper file. The parser context is only used for throwing errors and there seems not related to executing statements.

src/storage/tde.c
```diff
@@ -65,6 +65,11 @@
   } while (0)
 #define restore_signals(old_mask) sigprocmask(SIG_SETMASK, &(old_mask), NULL)
 
+#if defined(CS_MODE)
+DBLINK_CHPHER_KEY dblink_Cipher_key = { false, {0x00,}
+};
```
hgryoo: GNU indent broke indentation
```suggestion
// *INDENT-OFF*
DBLINK_CHPHER_KEY dblink_Cipher_key = {false, {0x00,}};
// *INDENT-ON*
```

src/communication/network_interface_cl.h
```diff
@@ -125,6 +125,7 @@ extern int file_apply_tde_to_class_files (const OID * class_oid);
 #ifdef UNSTABLE_TDE_FOR_REPLICATION_LOG
 extern int tde_get_data_keys ();
 #endif /* UNSTABLE_TDE_FOR_REPLICATION_LOG */
+extern int get_dblink_chpher_master_key ();
```
hgryoo: how about adding a prefix to this function's name? other functions have prefixes to easily know which module calls it
e.g) `dblink_get_cipher_master_key();`

src/query/execute_statement.c
```diff
@@ -18063,10 +18110,14 @@ get_dblink_info_from_dbserver (PARSER_CONTEXT * parser, const char *server, DB_V
   char *server_name, *t;
   DB_OBJECT *server_object, *server_class;
   DB_IDENTIFIER server_obj_id;
-  DB_VALUE values[4];
+  DB_VALUE values[4], pwd_val;
```
hgryoo, line 18128: clearing for `pwd_val` is required

src/communication/network_interface_cl.c
```diff
@@ -1885,6 +1885,46 @@ tde_get_data_keys ()
...
+      ptr = or_unpack_int (reply, &area_size);
+      ptr = or_unpack_int (ptr, &error);
+      if (area_size > 0)
+	{
+	  ptr = or_unpack_stream (area, (char *) data_keys.perm_key, TDE_DATA_KEY_LENGTH);
+	  ptr = or_unpack_stream (ptr, (char *) data_keys.temp_key, TDE_DATA_KEY_LENGTH);
+	  ptr = or_unpack_stream (ptr, (char *) dblink_Cipher_key.master_key, TDE_DATA_KEY_LENGTH);
+	  dblink_Cipher_key.is_loaded = true;
+	}
+      free_and_init (area);
```
beyondykk9, line 1921: this should be moved into prior "if" clause for safety.

src/query/execute_statement.c
```diff
@@ -18148,3 +18212,562 @@ get_dblink_info_from_dbserver (PARSER_CONTEXT * parser, const char *server, DB_V
...
+	  sprintf ((char *) buf1, "%s", pt);
+
+	  db_make_null (&encrypt_val[0]);
+	  err = get_dblink_password_encrypt (parser, pt, &encrypt_val[0], false);
+	  if (err == NO_ERROR)
+	    {
+	      gettimeofday (&startTime[1], NULL);
+
+	      db_make_null (&decrypt_val[0]);
+              pt2 = (char *) db_get_string (&encrypt_val[0]);
```
hgryoo: indent

src/storage/tde.c
ctshim: OK, I will fix it.

src/query/execute_statement.c
ctshim: thank you,

src/communication/network_interface_cl.c
ctshim: Which part is not safe?

src/query/execute_statement.c
ctshim: OK, thank you

src/communication/network_interface_cl.h
ctshim: I will accept your comments and fix it.

src/parser/view_transform.c
```diff
@@ -3426,8 +3426,8 @@ mq_copypush_sargable_terms (PARSER_CONTEXT * parser, PT_NODE * statement, PT_NOD
       info.in.attr_list = spec->info.spec.as_attr_list;
       info.in.query_list = derived_table;
 
-      sub = derived_table->info.query.q.select.from->info.spec.derived_table;
-      if (sub && sub->node_type == PT_DBLINK_TABLE)
+      sub = derived_table->info.query.q.select.from;
+      if (sub && sub->info.spec.derived_table_type == PT_DERIVED_DBLINK_TABLE)
```
shparkcubrid, line 3430: @byungwook-kim 
It does not work properly in the following cases.
```
select .. from (select .. from db_link() a, table b ... )
```
It should be pushed like this:
```
select ..
from ( select ..
               from db_link() <== second push
             where col1 = 1 <== first push
          )
where col1 =1
```
The old source code was written to work like this.

beyondykk9: I tested the query as which you wrote. The rewritten query was the same as you wrote.

src/query/execute_statement.c
```diff
@@ -17998,23 +18029,37 @@ do_create_server (PARSER_CONTEXT * parser, PT_NODE * statement)
...
+      goto end;
+    }
+  error = get_dblink_password_encrypt (parser, (char *) db_get_string (&tmp_passwd), &passwd, false);
+  if (error != NO_ERROR)
+    {
+      if (!pt_has_error (parser))
 	{
-	  error = ER_FAILED;
-	  goto end;
+	  PT_ERRORf (parser, statement, "Failed to encryption password. error=%d", error);
```
hgryoo, line 18066: The error message id related to db link server, password encryption should be defined

src/query/execute_statement.c
```diff
@@ -18148,3 +18212,562 @@ get_dblink_info_from_dbserver (PARSER_CONTEXT * parser, const char *server, DB_V
...
+      return ER_FAILED;
+    }
+
+  if (passwd && *passwd)
+    {
+      length = strlen (passwd);
+    }
+
+  db_make_null (&val);
+  if (length <= DBLINK_PASSWORD_MAX_LENGTH)
```
shparkcubrid, line 18299: When are raw password and encrypted password entered respectively? Can't the function be separated?

ctshim: The very first user input to perform the CREATE SERVER syntax raw Password,
Subsequent HA logs or passwords created by the unloaddb tool are encrypted.

src/communication/network_interface_cl.c
beyondykk9: I confused that the area variable is null if the error returns and free_and_init will fail. However I was aware the free_and_init checks the null point.

src/query/execute_statement.c
ctshim: An error code has been added, but the error message-related processing will be done in the subsequent work.

src/query/execute_statement.c
ctshim: The below two functions have been moved to crypt_opfunc.c.
  shake_4_encrypt(), reverse_shake_4_decrypt()
  
The two functions below remove the parser parameter and add an error code.
  get_dblink_password_encrypt (), get_dblink_password_decrypt ()

src/query/execute_statement.c
```diff
@@ -18256,6 +18271,22 @@ pt_check_dblink_password (PARSER_CONTEXT * parser, const char *passwd, char *cip
...
+	  ciper_func_test (parser, false);
+	  return ER_FAILED;
+	}
+      else if (strcmp (passwd, "test_time") == 0)
+	{
+	  ciper_func_test (parser, true);
+	  return ER_FAILED;
+	}
+    }
+#endif
```
ctshim, line 18292: This code and ciper_func_test() function don't need PR check.
The code will be deleted after the DBLINK function is complete.

src/query/execute_statement.c
```diff
@@ -18421,180 +18448,11 @@ get_dblink_password_decrypt (PARSER_CONTEXT * parser, const char *passwd_cipher,
...
-    }
-  while (++pwdlen % 3);
-
-  return NO_ERROR;
-}
 
 #ifndef NDEBUG
 void
-ciper_func_test (PARSER_CONTEXT * parser)
+ciper_func_test (PARSER_CONTEXT * parser, bool is_timecheck)
```
ctshim, line 18460: This ciper_func_test() function don't need PR check.
The code will be deleted after the DBLINK function is complete.

src/query/crypt_opfunc.c
```diff
@@ -697,3 +700,590 @@ crypt_generate_random_bytes (char *dest, int length)
...
+    {
+      return err;
+    }
+  if (pk[0] == 0x00)
+    {
+      key = evp_cipher.master_key;
+    }
+  else
+    {
+      strcpy ((char *) master_key, (char *) pk);
```
shparkcubrid: How about using strncpy()?

src/query/execute_statement.c
```diff
@@ -17998,23 +18027,50 @@ do_create_server (PARSER_CONTEXT * parser, PT_NODE * statement)
...
+  if (pwd == NULL)
     {
-      assert (si->pwd->node_type == PT_VALUE);
-      attr_val[4] = (char *) PT_VALUE_GET_BYTES (si->pwd);
-      if (attr_val[4] == NULL)
+      error = ER_FAILED;
+      goto end;
+    }
+
+  error = get_dblink_password_decrypt (pwd, &tmp_passwd);
```
shparkcubrid, line 18039: Why decrypt and encrypt?

src/query/execute_statement.c
```diff
@@ -18135,16 +18214,410 @@ get_dblink_info_from_dbserver (PARSER_CONTEXT * parser, const char *server, DB_V
...
+	    {
+	      strcpy (cipher_buf, str);
+	      err = NO_ERROR;
+	    }
+	}
+    }
+  else if (length == max_len)
+    {
+      // A encrypted password from the raw password.      
+      err = get_dblink_password_decrypt (passwd, &val);
```
shparkcubrid: If it is an encrypted password, why is it decrypted?
I'm curious as it's the opposite of encrypting the raw password.

ctshim: The length comparison only confirms that it is likely an encrypted password.
It is necessary to check whether the decryption process is successful.

src/query/crypt_opfunc.c
ctshim: In the current code situation, the two codes below are the same, so there seems to be no meaning of change.
```
strncpy ((char *) master_key, (char *) pk, strlen(pk));
strcpy ((char *) master_key, (char *) pk);
```

The code below is a little different, in this case we need to ensure that the pk's garbage area is clean.
`strncpy((char *) master_key, (char *) pk, DE_MASTER_KEY_LENGTH);`

In order to keep even the garbage area clean, it is initialized with the code below, and after strcpy() is executed, the back part will be filled with nulls.
`unsigned char master_key[TDE_MASTER_KEY_LENGTH] = { 0, };`

src/query/execute_statement.c
shparkcubrid: There is not 'err != NO_ERROR'.
And the decryption will proceed in do_create_server() or ...bind(), does it have to be checked here?

src/query/crypt_opfunc.c
shparkcubrid: How about checking length of pk?

ctshim: Is it to be prepared in case the pk exceeds the buffer size when copying?

pk will come as a string of up to 20 bytes or 14 bytes.
It won't overflow the buffer, but I'll check it when I'm in debug mode.

src/query/execute_statement.c
ctshim: Returns err if 'err != NO_ERROR'.

This function verifies the validity of the PASSWORD item read in the parsing stage.
When the user enters a manipulated value, an error can be raised without going to do_create_server().

shparkcubrid: Does it have to be checked in parser?
Is it a problem if it is checked in do_create_server()?

ctshim: It's just a time difference, but it shouldn't be a problem.
If there is no problem in the ALTER SERVER statement, it may be a good idea to remove it because it is a duplicate decryption operation.

src/base/error_code.h
```diff
@@ -1645,14 +1645,17 @@
 
 #define ER_DBLINK                                   -1284
 #define ER_DBLINK_INVALID_COLUMNS_SPECIFIED         -1285
-#define ER_DBLINK_CATALOG_DB_SERVER_NOT_FOUND	      -1286
+#define ER_DBLINK_CATALOG_DB_SERVER_NOT_FOUND	    -1286
```
hgryoo:
```suggestion
#define ER_DBLINK_CATALOG_DB_SERVER_NOT_FOUND	      -1286
```

ctshim: I wil fix it. thank you

shparkcubrid (approved):

beyondykk9 (approved):

src/storage/tde.h
```diff
@@ -101,6 +101,15 @@ typedef struct tde_mk_file_item
...
 #endif /* UNSTABLE_TDE_FOR_REPLICATION_LOG */
 
+#if defined(CS_MODE)
+typedef struct dblink_cipher
+{
+  bool is_loaded;
+  unsigned char master_key[TDE_DATA_KEY_LENGTH];
+} DBLINK_CHPHER_KEY;
+extern DBLINK_CHPHER_KEY dblink_Cipher_key;
+#endif
```
hgryoo: This structure doesn't have to be here. It is not related to TDE itself. Also, I think this file should be dealt with as a server's module (Only SERVER_MODE and SA_MODE can include this file).
How about adding authors (@wrlawodms) as a reviewer?

ctshim: tde.h is already included in includ CS_MODE.
I think it's natural for the declaration to be in tde.h. Otherwise the declaration of TDE_DATA_KEY_LENGTH should be moved to a file that can be referenced in two different places simultaneously for the same buffer size.
 
However, I think there may be disagreements over whether the body should be in tde.c.

First, let's hear from @wrlawodms.

ctshim: @wrlawodms
     You can only look at the TDE related part.
    The relevant sources are the files below.
`    tde.h, tde.c, network_interface_cl.c, crypt_opfunc.c`

   You can find it in the source with the "dblink_Cipher_key" variable name.

src/query/crypt_opfunc.c
```diff
@@ -697,3 +700,592 @@ crypt_generate_random_bytes (char *dest, int length)
...
+
+#if !defined(CS_MODE)
+      memcpy (evp_cipher.master_key, tde_Cipher.data_keys.log_key, TDE_DATA_KEY_LENGTH);
+#else
+      memcpy (evp_cipher.master_key, dblink_Cipher_key.master_key, TDE_DATA_KEY_LENGTH);
+#endif
+      is_init_done = 1;
+    }
+
+#if defined(CS_MODE)
```
wrlawodms, line 730: Where does it get the master key if not CS_MODE?

src/storage/tde.h
wrlawodms: I took a look at the code. To begin with, I'd like to ask you for the overall architecture for the encryption.
1. Please elaborate on the data to be encrypted. Are they decrypted and encrypted only in the same database instance? How about their lifecycle? They will be gone after the server shutdown?
2. Which value did you choose as NONCE? they must be unique for every message.
3. Which values can be used as the master key? the key must not be hard-corded or retrievable.
- BTW, I don't think the name "master key" is good. I used it in TDE because of 2-level encryption. You seem that you need only one key.

wrlawodms: And, generally, I don't recommend using constants and interfaces for TDE.  They are defined for the TDE's requirements and designs. They can be changed when we revise the TDE design or deal with issues on TDE. Please decouple it.

ctshim: @wrlawodms
We are adding a new feature called DBLINK.
A catalog called db_server is added to support this feature.
A password is included to store access information to the remote DB in this catalog, and this password must be able to be decrypted.
It uses the AES algorithm, and for this purpose, it was written by referring to the processing code of the TDE-related source.
There's more, but I'll leave out the parts that are beyond the scope of what wrlawodms will review.
1. The string itself is encrypted with the password of the remote DB and stored as a column value in the db_server catalog. Since AES comes in binary, it is saved as a base64 readable string.

2. The nonce is 16 bytes in size, same as the tde source,
The value is filled in using the code below.
`memset(evp_cipher.nonce, 0x07, sizeof (evp_cipher.nonce));`

3. Obtain the tde_Cipher.data_keys.log_key value generated from the tde-related code and use it as it is. What is the master key here?
It means the value passed to the key parameter when calling the EVP_EncryptInit_ex() function.



Here, we do not use the TDE function, but just use "tde_Cipher.data_keys.log_key" that is used for encryption processing in the TDE function. This is because there is an expectation that, without exposing the master key, it will always be possible to obtain a value that does not change in the same DB.

wrlawodms: Thank you for answering!
2. Basically, NONCE (Number used ONCE) has to be different for each data encrypted, but you are using the hard-coded number, 0x070707. Is it ok?
3. "Master key" is a concept for TDE key management, which is irrelevant to general encryption. In TDE, master keys mean keys that encrypts data keys.

One more question: TDE assumes the master key for TDE could be lost, which in turn means you can't access the `log_key`. In this case, DB just works without encrypted data. If you use `log_key` to encrypt the password, the DB link feature will be dependant on the TDE Key management policy. Is it on purpose?

I'll leave other reviews on each code line.

ctshim: @wrlawodms
    I refer to the code below inside the tde_dk_nonce() function
```
  case TDE_DATA_KEY_TYPE_LOG:
       memset(dk_nonce, 2, TDE_DK_NONCE_LENGTH);
```
I don't know what the exact role of the nonce is, but I think that using the nonce value as a fixed value will not cause any problems.

If the TDE key (<dbname>_keys file) is lost, the TDE function will be limited and DBLINK's encryption processing will not be possible.
However, other functions except dBLINK can be used as they are now.

What I am concerned about is this.
It should be guaranteed that the value of tde_Cipher.data_keys.log_key cannot be changed in the DB that has already been created (except for loss).

wrlawodms: > I don't know what the exact role of the nonce is, but I think that using the nonce value as a fixed value will not cause any problems.

In the code you referred to, you can see each data key is encrypted with a difference NONCE. Please refer to `tde_encrpyt_data_page()` which use a different NONCE for each encrypted data. In your case, I think NONCE has to be different per each data (maybe per each user password). In my opinion, if they are stored in a catalog class, you can use each OID of the record as the NONCE for it. From what I know, shortly, this is a security requirement and if NONCE of two encrypted data is the same, an attacker can guess whether two data are the same or not. 

> If the TDE key (_keys file) is lost, the TDE function will be limited and DBLINK's encryption processing will not be possible.
> However, other functions except dBLINK can be used as they are now.

I don't really know how dBLINK works. If it can work without the encrypted data (maybe the user password for a remote server), it doesn't matter. 

> What I am concerned about is this.
> It should be guaranteed that the value of tde_Cipher.data_keys.log_key cannot be changed in the DB that has already been created (except for loss).

Yes, the data key is guaranteed not to be changed.

wrlawodms: Oh, It doesn't seem enough to use the record OID for a Nonce for it. With this, if a user changed the password with the same word, an attacker can know it. I think you should figure out another unique nonce for each data if you want to meet the requirement strictly.

src/storage/tde.h
```diff
@@ -92,7 +92,7 @@ typedef struct tde_data_key_set
 typedef struct tde_mk_file_item
 {
   time_t created_time;		/* If it is -1, it is invalid and avaliable for a new key */
-  unsigned char master_key[TDE_MASTER_KEY_LENGTH];
+  unsigned char master_key[TDE_DATA_KEY_LENGTH];
```
wrlawodms: Could you tell me why you should change this?

src/storage/tde.h
```diff
@@ -101,6 +101,15 @@ typedef struct tde_mk_file_item
 #endif /* CS_MODE */
 #endif /* UNSTABLE_TDE_FOR_REPLICATION_LOG */
 
+#if defined(CS_MODE)
+typedef struct dblink_cipher
```
wrlawodms: I think this structure should move to a file for dblink and use its own definition for the key length for the key for dblink.

src/communication/network_interface_cl.c
```diff
@@ -1885,6 +1885,46 @@ tde_get_data_keys ()
...
+  char *ptr;
+  OR_ALIGNED_BUF (OR_INT_SIZE + OR_INT_SIZE) a_reply;
+  char *reply, *area;
+
+  dblink_Cipher_key.is_loaded = false;
+
+  reply = OR_ALIGNED_BUF_START (a_reply);
+
+  req_error =
+    net_client_request2 (NET_SERVER_TDE_GET_DATA_KEYS, NULL, 0, reply,
```
wrlawodms: I don't think using `NET_SERVER_TDE_GET_DATA_KEYS` directly is a good idea even if you're going to use the tde key as the key for dblink. This interface can be changed along with the design of TDE. I recommend creating its own interface for getting a key that returns the key. For the same reason, I think you should define its own constants, not using  `TDE_DATA_KEY_LENGTH`.

src/query/crypt_opfunc.c
```diff
@@ -697,3 +700,595 @@ crypt_generate_random_bytes (char *dest, int length)
...
+
+// 
+// *INDENT-OFF*
+#if defined(CS_MODE)
+DBLINK_CHPHER_KEY dblink_Cipher_key = {false, {0x00,}};
+#endif
+static struct
+{
+  unsigned char nonce[TDE_DK_NONCE_LENGTH]; // See TDE_DK_NONCE_LENGTH in tde.h
+  unsigned char master_key[TDE_DATA_KEY_LENGTH];
```
wrlawodms: For the same reason as the comment I wrote in network_interface_cl.c, please define and use its own constants.

src/query/crypt_opfunc.c
```diff
@@ -697,3 +700,595 @@ crypt_generate_random_bytes (char *dest, int length)
...
+init_dblink_cipher (EVP_CIPHER_CTX ** ctx, const EVP_CIPHER ** cipher_type, bool is_aes_algorithm)
+{
+  static int is_init_done = 0;
+
+  if (is_init_done == 0)
+    {
+      memset (evp_cipher.nonce, 0x07, sizeof (evp_cipher.nonce));
+
+#if !defined(CS_MODE)
+      memcpy (evp_cipher.master_key, tde_Cipher.data_keys.log_key, TDE_DATA_KEY_LENGTH);
```
wrlawodms: For the same reason as the comment I wrote in network_interface_cl.c, please don't use tde_Cipher directly here.

wrlawodms (commented):
TDE-related comments.

src/storage/tde.h
ctshim: Currently TDE_MASTER_KEY_LENGTH and TDE_DATA_KEY_LENGTH are the same size.
The value to be included in the master_key is the log_key of the TDE_DATA_KEY_SET structure, so I set the define representing the size of the array.

src/storage/tde.h
ctshim: I will review your comments positively.

src/communication/network_interface_cl.c
ctshim: The point of what you are saying is that it would be good to have the key value independent of the TDE policy?
I understand that the <db-name>_keys file is not used by anything other than TDE.
Do you mean to write a separate routine for accessing that information?
Or is it better to get the key-related files in another way?

wrlawodms: _keys file is only for master keys, not the data key that you use.

src/storage/tde.h
wrlawodms: Yes, they are the same size now but could be different later. `tde_mk_file_item` is for master keys, not data keys so that I think we have to keep `TDE_MASTER_KEY_LENGTH`.

hornetmj:
@byungwook-kim If the @dev3 requests a review with a code related to our team, please register me as a reviewer too.

src/communication/network_interface_cl.c
```diff
@@ -1885,6 +1885,68 @@ tde_get_data_keys ()
 }
 #endif /* UNSTABLE_TDE_FOR_REPLICATION_LOG */
 
+
+int
+dblink_get_cipher_master_key ()
+{
+#if defined(CS_MODE)
+  TDE_DATA_KEY_SET data_keys;
```
```suggestion
```

src/storage/tde.h
wrlawodms: This conversation was resolved without a consensus on it. This should not be changed.

ctshim: Oops, My mistake, I will revert to the old code.

src/communication/network_interface_cl.c
ctshim: Removed it. thank you.

wrlawodms (approved):
I think all TDE-related problems are solved. There still seems to be a kind of security problem like non-unique NONCE mentioned above, but I think it depends on the requirements and design of dBlink. So, I'll leave it to dev3 and approve it.

hornetmj (approved):

src/base/error_code.h
```diff
@@ -1645,14 +1645,17 @@
...
 #define ER_DBLINK_SERVER_ALREADY_EXISTS             -1288
 #define ER_DBLINK_CANNOT_UPDATE_SERVER              -1289
 #define ER_DBLINK_UNSUPPORTED_TYPE                  -1290
 #define ER_DBLINK_INVALID_BIND_PARAM                -1291
+#define ER_DBLINK_PASSWORD_OVER_MAX_LENGTH          -1292
+#define ER_DBLINK_PASSWORD_INVALID_LENGTH           -1293
+#define ER_DBLINK_PASSWORD_CHECKSUM                 -1294
 
-#define ER_LAST_ERROR                               -1292
+#define ER_LAST_ERROR                               -1294
```
beyondykk9: ER_LAST_ERROR, it should be -1295

shparkcubrid (approved):

src/base/error_code.h
ctshim: Fixed it,  thank you.

beyondykk9 (approved):

youngjinj (approved):

hgryoo (approved):

PR #2907, opened 2021-07-16
[LETS-116] Disable log archiving on transaction server with remote storage

ribram:
http://jira.cubrid.org/browse/LETS-116

Disable log archiving related code on the transaction server with remote storage because it not used since log is not written to disk.

- Do not start log archive removing daemon
- Disable background archiving
- Add safe-guard that archiving related functions are never reached on TS with remote storage

Temporarily disabled the "recovery" path on the transaction server with remote storage. It remains to be done later on.

### DIFF
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -1332,22 +1332,16 @@ log_initialize_internal (THREAD_ENTRY * thread_p, const char *db_fullname, const
   /*
    * Was the database system shut down or was it involved in a crash ?
    */
-  if (init_emergency == false && (log_Gl.hdr.is_shutdown == false || ismedia_crash == true))
+  if (init_emergency == false && (log_Gl.hdr.is_shutdown == false || ismedia_crash == true)
+      && !is_tran_server_with_remote_storage ())
     {
       /*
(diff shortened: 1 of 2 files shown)


src/transaction/log_manager.c
```diff
@@ -1332,13 +1332,16 @@ log_initialize_internal (THREAD_ENTRY * thread_p, const char *db_fullname, const
...
+  if (init_emergency == false && (log_Gl.hdr.is_shutdown == false || ismedia_crash == true)
+      && !is_tran_server_with_remote_storage ())
     {
       /*
        * System was involved in a crash.
        * Execute the recovery process
        */
       log_recovery (thread_p, ismedia_crash, stopat);
+
+      // todo: TS with remote storage recovery
```
cristiarg, line 1344: Does this todo apply to the else branch, or, in the final version, some log recovery action is still needed even on a TS with remote storage?

cristiarg (approved):

src/transaction/log_manager.c
ribram: Some recovery action may be needed even on a TS with remote storage.

PR #2908, opened 2021-07-19
[CBRD-23911] Add system parameter for supplemental logging

joohok:

System parameter for supplemental log is added.

Default value is 0 which means no supplemental logging.

### DIFF
--- a/src/base/system_parameter.c
+++ b/src/base/system_parameter.c
@@ -692,6 +692,8 @@ static const char sysprm_ha_conf_file_name[] = "cubrid_ha.conf";
 #define PRM_NAME_DDL_AUDIT_LOG "ddl_audit_log"
 #define PRM_NAME_DDL_AUDIT_LOG_SIZE "ddl_audit_log_size"
 
+#define PRM_NAME_SUPPLEMENTAL_LOG "supplemental_log"
+
 #define PRM_VALUE_DEFAULT "DEFAULT"
 #define PRM_VALUE_MAX "MAX"
 #define PRM_VALUE_MIN "MIN"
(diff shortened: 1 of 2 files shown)


src/base/system_parameter.c
```diff
@@ -6036,6 +6042,18 @@ static SYSPRM_PARAM prm_Def[] = {
...
    (DUP_PRM_FUNC) NULL,
+   (DUP_PRM_FUNC) NULL},
+  {PRM_ID_SUPPLEMENTAL_LOG,
+   PRM_NAME_SUPPLEMENTAL_LOG,
+   (PRM_FOR_SERVER | PRM_FOR_CLIENT),
+   PRM_INTEGER,
+   &prm_supplemental_log_flag,
+   (void *) &prm_supplemental_log_default,
+   (void *) &PRM_SUPPLEMENTAL_LOG,
+   (void *) NULL,
```
wrlawodms, line 6053: Isn't it needed to define the upper and lower bound of this?

joohok: No, It is required to be defined. I'll fix it after the level of supplemental log is defined.

wrlawodms (approved):

src/base/system_parameter.c
```diff
@@ -6036,6 +6042,18 @@ static SYSPRM_PARAM prm_Def[] = {
    (void *) &prm_ddl_audit_log_size_lower,
    (char *) NULL,
    (DUP_PRM_FUNC) NULL,
+   (DUP_PRM_FUNC) NULL},
+  {PRM_ID_SUPPLEMENTAL_LOG,
+   PRM_NAME_SUPPLEMENTAL_LOG,
+   (PRM_FOR_SERVER | PRM_FOR_CLIENT),
```
hornetmj, line 6048: How to handle the following case?
- client is set 'true'.
- server is set 'false'.

or vise versa

PR #2909, opened 2021-07-19
[CBRD-23911] Add hint for not appending supplemental logs

joohok:

This PR is for not appending supplemental logs. 

Some supplemental logs for DML are able to be appended redundantly by the DDL supplemental logs.
Therefore, supplemental logs for DML are required to be blocked if the logs are generated during DDL execution. 

To block the DML supplemental log, /*+ NO_SUPPLEMENTAL_LOG */ hint is added, and it implemented as /*+ NO_LOGGING */ hint.

### DIFF
--- a/src/object/schema_manager.c
+++ b/src/object/schema_manager.c
@@ -15513,16 +15513,7 @@ sm_truncate_using_delete (MOP class_mop)
   /* We will run a DELETE statement with triggers disabled. */
   save_tr_state = tr_set_execution_state (false);
 
-  if (prm_get_integer_value (PRM_ID_SUPPLEMENTAL_LOG) > 0)
-    {
-      (void) snprintf (delete_query, sizeof (delete_query), "DELETE /*+ RECOMPILE NO_SUPPLEMENTAL_LOG */ FROM [%s];",
-		       class_name);
-    }
-  else
-    {
-    
-    }
-
+  (void) snprintf (delete_query, sizeof (delete_query), "DELETE /*+ RECOMPILE NO_SUPPLEMENTAL_LOG */ FROM [%s];", class_name);
 
   session = db_open_buffer (delete_query);
   if (session == NULL)
--- a/src/object/schema_manager.c
+++ b/src/object/schema_manager.c
@@ -15520,7 +15520,7 @@ sm_truncate_using_delete (MOP class_mop)
     }
   else
     {
-      (void) snprintf (delete_query, sizeof (delete_query), "DELETE /*+ RECOMPILE */ FROM [%s];", class_name);
+    
     }
 
 
--- a/src/object/schema_manager.c
+++ b/src/object/schema_manager.c
@@ -15513,7 +15513,16 @@ sm_truncate_using_delete (MOP class_mop)
   /* We will run a DELETE statement with triggers disabled. */
   save_tr_state = tr_set_execution_state (false);
 
-  (void) snprintf (delete_query, sizeof (delete_query), "DELETE /*+ RECOMPILE */ FROM [%s];", class_name);
+  if (prm_get_integer_value (PRM_ID_SUPPLEMENTAL_LOG) > 0)
+    {
+      (void) snprintf (delete_query, sizeof (delete_query), "DELETE /*+ RECOMPILE NO_SUPPLEMENTAL_LOG */ FROM [%s];",
+		       class_name);
+    }
+  else
+    {
+      (void) snprintf (delete_query, sizeof (delete_query), "DELETE /*+ RECOMPILE */ FROM [%s];", class_name);
+    }
+
 
   session = db_open_buffer (delete_query);
   if (session == NULL)
--- a/src/parser/parse_tree.h
+++ b/src/parser/parse_tree.h
@@ -1201,7 +1201,8 @@ typedef enum
   /* SELECT page header information from heap file instead of record data */
   PT_HINT_SELECT_KEY_INFO = 0x80000000,	/* 1000 0000 0000 0000 0000 0000 0000 0000 */
   /* SELECT key information from index b-tree instead of table record data */
-  PT_HINT_NO_SUPPLEMENTAL_LOG = 0x100000000
+  PT_HINT_NO_SUPPLEMENTAL_LOG = 0x100000000 /* 0001 0000 0000 0000 0000 0000 0000 0000 0000 */
+  /* Used in DML (insert, update, delete) to avoid adding DML supplemental logs that may be duplicated by DDL */
 } PT_HINT_ENUM;
 
 /* Codes for error messages */
--- a/src/parser/parse_tree.h
+++ b/src/parser/parse_tree.h
@@ -1199,8 +1199,9 @@ typedef enum
   /* SELECT record info from tuple header instead of data */
   PT_HINT_SELECT_PAGE_INFO = 0x40000000,	/* 0100 0000 0000 0000 0000 0000 0000 0000 */
   /* SELECT page header information from heap file instead of record data */
-  PT_HINT_SELECT_KEY_INFO = 0x80000000	/* 1000 0000 0000 0000 0000 0000 0000 0000 */
-    /* SELECT key information from index b-tree instead of table record data */
+  PT_HINT_SELECT_KEY_INFO = 0x80000000,	/* 1000 0000 0000 0000 0000 0000 0000 0000 */
+  /* SELECT key information from index b-tree instead of table record data */
+  PT_HINT_NO_SUPPLEMENTAL_LOG = 0x100000000
 } PT_HINT_ENUM;
 
 /* Codes for error messages */
--- a/src/parser/parse_tree.h
+++ b/src/parser/parse_tree.h
@@ -1201,8 +1201,8 @@ typedef enum
   /* SELECT page header information from heap file instead of record data */
   PT_HINT_SELECT_KEY_INFO = 0x80000000,	/* 1000 0000 0000 0000 0000 0000 0000 0000 */
   /* SELECT key information from index b-tree instead of table record data */
-  PT_HINT_NO_SUPPLEMENTAL_LOG = 0x100000000 /* 0001 0000 0000 0000 0000 0000 0000 0000 0000 */
-  /* Used in DML (insert, update, delete) to avoid adding DML supplemental logs that may be duplicated by DDL */
+  PT_HINT_NO_SUPPLEMENTAL_LOG = 0x100000000	/* 0001 0000 0000 0000 0000 0000 0000 0000 0000 */
+    /* Used in DML (insert, update, delete) to avoid adding DML supplemental logs that may be duplicated by DDL */
 } PT_HINT_ENUM;
 
 /* Codes for error messages */
--- a/src/parser/parse_tree_cl.c
+++ b/src/parser/parse_tree_cl.c
@@ -8506,6 +8506,11 @@ pt_print_delete (PARSER_CONTEXT * parser, PT_NODE * p)
 	  q = pt_append_nulstring (parser, q, " USE_SBR ");
 	}
 
+      if (p->info.delete_.hint & PT_HINT_NO_SUPPLEMENTAL_LOG)
+	{
+	  q = pt_append_nulstring (parser, q, " NO_SUPPLEMENTAL_LOG");
+	}
+
       q = pt_append_nulstring (parser, q, " */");
     }
   if (r1)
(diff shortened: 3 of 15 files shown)


joohok:
code-style check failed due to previous PR (https://github.com/CUBRID/cubrid/pull/2612)

src/parser/parse_tree.h
```diff
@@ -1199,8 +1199,9 @@ typedef enum
   /* SELECT record info from tuple header instead of data */
   PT_HINT_SELECT_PAGE_INFO = 0x40000000,	/* 0100 0000 0000 0000 0000 0000 0000 0000 */
   /* SELECT page header information from heap file instead of record data */
-  PT_HINT_SELECT_KEY_INFO = 0x80000000	/* 1000 0000 0000 0000 0000 0000 0000 0000 */
-    /* SELECT key information from index b-tree instead of table record data */
+  PT_HINT_SELECT_KEY_INFO = 0x80000000,	/* 1000 0000 0000 0000 0000 0000 0000 0000 */
+  /* SELECT key information from index b-tree instead of table record data */
+  PT_HINT_NO_SUPPLEMENTAL_LOG = 0x100000000
```
wrlawodms: Please add a comment for this as above

src/parser/parse_tree_cl.c
```diff
@@ -8506,6 +8506,11 @@ pt_print_delete (PARSER_CONTEXT * parser, PT_NODE * p)
 	  q = pt_append_nulstring (parser, q, " USE_SBR ");
 	}
 
+      if (p->info.delete_.hint & PT_HINT_NO_SUPPLEMENTAL_LOG)
+	{
+	  q = pt_append_nulstring (parser, q, " NO_SUPPLEMENTAL_LOG");
```
wrlawodms:
```suggestion
	  q = pt_append_nulstring (parser, q, " NO_SUPPLEMENTAL_LOG ");
```

src/parser/parse_tree_cl.c
```diff
@@ -14986,6 +14996,11 @@ pt_print_update (PARSER_CONTEXT * parser, PT_NODE * p)
 	  b = pt_append_nulstring (parser, b, " USE_SBR ");
 	}
 
+      if (p->info.update.hint & PT_HINT_NO_SUPPLEMENTAL_LOG)
+	{
+	  b = pt_append_nulstring (parser, b, " NO_SUPPLEMENTAL_LOG");
```
wrlawodms:
```suggestion
	  b = pt_append_nulstring (parser, b, " NO_SUPPLEMENTAL_LOG ");
```

src/object/schema_manager.c
```diff
@@ -15513,7 +15513,16 @@ sm_truncate_using_delete (MOP class_mop)
...
 
-  (void) snprintf (delete_query, sizeof (delete_query), "DELETE /*+ RECOMPILE */ FROM [%s];", class_name);
+  if (prm_get_integer_value (PRM_ID_SUPPLEMENTAL_LOG) > 0)
+    {
+      (void) snprintf (delete_query, sizeof (delete_query), "DELETE /*+ RECOMPILE NO_SUPPLEMENTAL_LOG */ FROM [%s];",
+		       class_name);
+    }
+  else
+    {
+      (void) snprintf (delete_query, sizeof (delete_query), "DELETE /*+ RECOMPILE */ FROM [%s];", class_name);
```
wrlawodms: How about always using the hint in this sort of context (DDL)? The value of `PRM_ID_SUPPLEMENTAL_LOG` will be checked in the code later where the supplement log is appended. 
```suggestion
    
```

wrlawodms (commented):
With these, I have some questions.

1. Just to make it clear, did we decide to apply this hint only for DML? and for all DML including update, insert, and delete?
2. `pt_to_merge_update_xasl()` and `pt_to_merge_insert_xasl()` doesn't need to consider this hint? they consider `no_logging` hint.
3. From what I remember, there are more DML queries that use statements as TRUNCATE does. We saw it by using `grep, didn't we? They don't need to use this hint?

src/query/execute_schema.c
```diff
@@ -4656,7 +4656,15 @@ do_redistribute_partitions_data (const char *classname, const char *keyname, cha
 	  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, query_size + 1);
 	  return ER_FAILED;
 	}
-      sprintf (query_buf, "UPDATE [%s] SET [%s]=[%s];", classname, keyname, keyname);
+
+      if (prm_get_integer_value (PRM_ID_SUPPLEMENTAL_LOG) > 0)
+	{
+	  sprintf (query_buf, "UPDATE /*+ NO_SUPPLEMENTAL_LOG */ [%s] SET [%s]=[%s];", classname, keyname, keyname);
```
wrlawodms: Where have you handled the case where DDL does not use a statement but uses a locator function directly?

joohok: It was omitted in the process of writing a new PR. I'll add that part.

src/parser/parse_tree_cl.c
joohok: Thanks

src/parser/parse_tree_cl.c
joohok: Thanks

src/parser/parse_tree.h
joohok: OK, I'll add a comment

src/object/schema_manager.c
joohok: OK, It may reduce the cost of checking conditions.

joohok:
> With these, I have some questions.
> 
> 1. Just to make it clear, did we decide to apply this hint only for DML? and for all DML including update, insert, and delete?

- Yes. This hint is to avoid adding DML supplemental logs that may be duplicated by DDL.

> 2. `pt_to_merge_update_xasl()` and `pt_to_merge_insert_xasl()` doesn't need to consider this hint? they consider `no_logging` hint.

- No, I didn't consider Merge clause. I just consider DMLs that can be used in the process of DDL.
 
> 3. From what I remember, there are more DML queries that use statements as TRUNCATE does. We saw it by using `grep, didn't we? They don't need to use this hint?

- Actually, I remember that we can not cover all the DDL cases that use DML queries, so it can be covered during regression test. However, I found only _one case_ that uses DML query in DDL processing. Most of cases are in log_applier. So, I will add that case. thank you 
-- Oh, I mistake. The hint is required to be append to the DML queries. And other locator functions will be covered during regression test.

wrlawodms:
> * Yes. This hint is to avoid adding DML supplemental logs that may be duplicated by DDL. 
> * No, I didn't consider Merge clause. I just consider DMLs that can be used in the process of DDL.

Yes, but I remember that we just made a conclusion that it depends on whether the hint can be hidden or not and you would check. 

> * Actually, I remember that we can not cover all the DDL cases that use DML queries, so it can be covered during regression test. However, I found only _one case_ that uses DML query in DDL processing. Most of cases are in log_applier. So, I will add that case. thank you
>   -- Oh, I mistake. The hint is required to be append to the DML queries. And other locator functions will be covered during regression test.

From what I've understood, we decided to add the hint to all DDL by using DML statements and handle DDL using locator() after the regression test except for the partitioning reorganization case. This is what you're saying?

src/parser/scanner_support.c
```diff
@@ -510,6 +510,23 @@ pt_get_hint (const char *text, PT_HINT hint_table[], PT_NODE * node)
...
+	      else if (node->node_type == PT_DELETE)
+		{
+		  node->info.delete_.hint = (PT_HINT_ENUM) (node->info.delete_.hint | hint_table[i].hint);
+		}
+	      else if (node->node_type == PT_INSERT)
+		{
+		  node->info.insert.hint = (PT_HINT_ENUM) (node->info.insert.hint | hint_table[i].hint);
+		}
+
+	      hint_table[i].arg_list = NULL;
```
wrlawodms: Could you tell me why this is necessary? some hints initialize it and some don't.

src/transaction/locator_sr.c
```diff
@@ -12791,6 +12791,9 @@ redistribute_partition_data (THREAD_ENTRY * thread_p, OID * class_oid, int no_oi
 		  goto exit;
 		}
 
+              /* No supplemental log for insert is appended due to DDL statement */
+              thread_p->no_supplemental_log = true;
```
wrlawodms: You have to turn this off after a locator function using a heap logical function.

wrlawodms (changes requested):

src/parser/scanner_support.c
joohok: I just follow the NO_LOGGING hint, and I found that this function(pt_get_hint) are not called anywhere. So I will discard this change.

src/transaction/locator_sr.c
joohok: Ok, I will turn off it after the locator function

joohok:
> Yes, but I remember that we just made a conclusion that it depends on whether the hint can be hidden or not and you would check.
> 

Yes, that's right. I check that there is no way to hide the hint. So, I only add a hint to DELETE and UPDATE. 
The changes related to INSERT will be discarded.  

> From what I've understood, we decided to add the hint to all DDL by using DML statements and handle DDL using locator() after the regression test except for the partitioning reorganization case. This is what you're saying?

Yes, that's what I mean. So I added some more DDL cases that use DML statements inside.

src/parser/scanner_support.c
wrlawodms: I doubt it's good to ignore `pt_get_hint()`. Although it is not used now, this seems a utility function to be able to be used later. Plus, it is not surrounded by macro `ENABLE_UNUSED_FUNCTION`. I'd say it should be considered if it is not that picky to handle.

src/transaction/locator_sr.c
```diff
@@ -12804,6 +12804,8 @@ redistribute_partition_data (THREAD_ENTRY * thread_p, OID * class_oid, int no_oi
 		{
 		  goto exit;
 		}
+
+	      thread_p->no_supplemental_log = false;
```
wrlawodms, line 12808: What if `locator_insert_force()` fails?

src/parser/parse_tree.h
```diff
@@ -1199,8 +1199,10 @@ typedef enum
   /* SELECT record info from tuple header instead of data */
   PT_HINT_SELECT_PAGE_INFO = 0x40000000,	/* 0100 0000 0000 0000 0000 0000 0000 0000 */
   /* SELECT page header information from heap file instead of record data */
-  PT_HINT_SELECT_KEY_INFO = 0x80000000	/* 1000 0000 0000 0000 0000 0000 0000 0000 */
-    /* SELECT key information from index b-tree instead of table record data */
+  PT_HINT_SELECT_KEY_INFO = 0x80000000,	/* 1000 0000 0000 0000 0000 0000 0000 0000 */
+  /* SELECT key information from index b-tree instead of table record data */
+  PT_HINT_NO_SUPPLEMENTAL_LOG = 0x100000000	/* 0001 0000 0000 0000 0000 0000 0000 0000 0000 */
+    /* Used in DML (insert, update, delete) to avoid adding DML supplemental logs that may be duplicated by DDL */
```
hornetmj, line 1205: How about informing someone who read this comment that this hint currently works only update and delete which are in the code level? So, it needs more considerations if someone wants to use this hint generally.

src/query/execute_schema.c
```diff
@@ -13112,7 +13112,9 @@ do_run_update_query_for_new_notnull_fields (PARSER_CONTEXT * parser, PT_NODE * a
 
   /* Using UPDATE ALL to update the current class and all its children. */
 
-  n = snprintf (q, remaining, "UPDATE ALL [%s] SET ", alter->info.alter.entity_name->info.name.original);
+  n =
```
hornetmj, line 13115: For me, share the scenario when this update is executed.

src/query/execute_schema.c
```diff
@@ -13200,7 +13202,9 @@ do_run_update_query_for_new_default_expression_fields (PARSER_CONTEXT * parser,
   query[0] = 0;
 
   /* Using UPDATE ALL to update the current class and all its children. */
-  n = snprintf (q, remaining, "UPDATE ALL [%s] SET ", alter->info.alter.entity_name->info.name.original);
+  n =
```
hornetmj, line 13205: For me, share the scenario when this update is executed.

src/query/stream_to_xasl.c
```diff
@@ -3652,6 +3654,7 @@ stx_build_insert_proc (THREAD_ENTRY * thread_p, char *ptr, INSERT_PROC_NODE * in
   ptr = or_unpack_int (ptr, &insert_info->has_uniques);
   ptr = or_unpack_int (ptr, &insert_info->wait_msecs);
   ptr = or_unpack_int (ptr, &insert_info->no_logging);
+  ptr = or_unpack_int (ptr, &insert_info->no_supplemental_log);
```
hornetmj, line 3657: Does it need currently?

PR #2911, opened 2021-07-19
[CBRD-23911] Implement log append functions for supplemental log

joohok:

Implemented log_append_supplemental_*
Ignore supplemental logs when recovery processed

### DIFF
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -4639,10 +4639,6 @@ log_append_donetime_internal (THREAD_ENTRY * thread_p, LOG_TDES * tdes, LOG_LSA 
     }
 
   LSA_COPY (eot_lsa, &lsa);
-#if !defined(NDEBUG) && 1	//JOOHOK
-  _er_log_debug (ARG_FILE_LINE, "LOG_TYPE : %s , time : %lld, LSA : %lld|%d \n ", log_to_string (iscommitted),
-		 donetime->at_time, LSA_AS_ARGS (eot_lsa));
-#endif
 }
 
 /*
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -451,7 +451,8 @@ log_to_string (LOG_RECTYPE type)
       return "LOG_DUMMY_OVF_RECORD";
     case LOG_DUMMY_GENERIC:
       return "LOG_DUMMY_GENERIC";
-
+    case LOG_SUPPLEMENTAL_INFO:
+      return "LOG_SUPPLEMENTAL_INFO";
     case LOG_SMALLER_LOGREC_TYPE:
     case LOG_LARGER_LOGREC_TYPE:
       break;
@@ -4638,6 +4639,10 @@ log_append_donetime_internal (THREAD_ENTRY * thread_p, LOG_TDES * tdes, LOG_LSA 
     }
 
   LSA_COPY (eot_lsa, &lsa);
+#if !defined(NDEBUG) && 1	//JOOHOK
+  _er_log_debug (ARG_FILE_LINE, "LOG_TYPE : %s , time : %lld, LSA : %lld|%d \n ", log_to_string (iscommitted),
+		 donetime->at_time, LSA_AS_ARGS (eot_lsa));
+#endif
 }
 
 /*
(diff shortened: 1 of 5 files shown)


src/transaction/log_manager.c
```diff
@@ -4741,6 +4742,126 @@ log_append_abort_log (THREAD_ENTRY * thread_p, LOG_TDES * tdes, LOG_LSA * abort_
...
+ * log_append_supplemental_log - append supplemental log record 
+ *
+ * return: nothing
+ *
+ *   rec_type (in): type of supplemental log record .
+ *   length (in) : length of supplemental data length.
+ *   data (in) : supplemental data
+ *   
+ */
+
```
wrlawodms, line 4755:
```suggestion
```

src/transaction/log_manager.c
```diff
@@ -4741,6 +4742,126 @@ log_append_abort_log (THREAD_ENTRY * thread_p, LOG_TDES * tdes, LOG_LSA * abort_
...
+int
+log_append_supplemental_undo_record (THREAD_ENTRY * thread_p, RECDES * undo_recdes)
+{
+  assert (undo_recdes != NULL);
+  assert (prm_get_integer_value (PRM_ID_SUPPLEMENTAL_LOG) > 0);
+
+  LOG_ZIP *zip_undo = NULL;
+  int length = undo_recdes->length;
+  void *data = undo_recdes->data;
+  int min_size_to_compress = 255;
```
wrlawodms, line 4840: It seems that it'd be better to use `log_Zip_min_size_to_compress` if you intended to use this value.

src/transaction/log_manager.c
```diff
@@ -4741,6 +4742,126 @@ log_append_abort_log (THREAD_ENTRY * thread_p, LOG_TDES * tdes, LOG_LSA * abort_
...
+	{
+	  return ER_OUT_OF_VIRTUAL_MEMORY;
+	}
+
+      log_zip (zip_undo, length, data);
+      length = zip_undo->data_length;
+      data = zip_undo->log_data;
+    }
+
+  log_append_supplemental_log (thread_p, LOG_SUPPLEMENT_UNDO_RECORD, length, data);
```
wrlawodms, line 4855: We should consider TDE-encryption to `LOG_SUPPLEMENT_UNDO_RECORD`.

src/transaction/log_record.hpp
```diff
@@ -133,6 +133,8 @@ enum log_rectype
   LOG_DUMMY_GENERIC,		/* used for flush for now. it is ridiculous to create dummy log records for every single
 				 * case. we should find a different approach */
 
+  LOG_SUPPLEMENTAL_INFO,
```
wrlawodms, line 136: Please add a comment for it as above.

src/transaction/log_record.hpp
```diff
@@ -403,6 +405,28 @@ struct log_rec_2pc_particp_ack
   int particp_index;		/* Index of the acknowledging participant */
 };
 
+typedef enum supplement_rec_type
+{
+  LOG_SUPPLEMENT_TRAN_USER,
+  LOG_SUPPLEMENT_UNDO_RECORD, /*CONTAINS undo raw record */
```
hornetmj, line 411:
```suggestion
  LOG_SUPPLEMENT_UNDO_RECORD, /* Contains undo raw record */
```

src/transaction/log_record.hpp
```diff
@@ -403,6 +405,28 @@ struct log_rec_2pc_particp_ack
...
+  LOG_SUPPLEMENT_DDL,
+  /* CONTAINS LSA of logs which contain undo, redo raw record
```
hornetmj, line 413:
```suggestion
  /* Contains LSA of logs which contain undo, redo raw record
```

src/transaction/log_record.hpp
```diff
@@ -403,6 +405,28 @@ struct log_rec_2pc_particp_ack
...
+   * | LOG_REC_HEADER | SUPPLEMENT_REC_TYPE | LENGTH | CLASS OID |  UNDO LSA (sizeof LOG_LSA) | REDO LSA | */
+  LOG_SUPPLEMENT_UPDATE,
+  LOG_SUPPLEMENT_DELETE,
+  LOG_SUPPLEMENT_INSERT,
+  LOG_SUPPLEMENT_CLASS_OID,
+  LOG_SUPPLEMENT_STATEMENT,
```
hornetmj, line 419: Is there any difference between LOG_SUPPLEMENT_DDL and LOG_SUPPLEMENT_STATEMENT?

src/transaction/log_manager.c
```diff
@@ -4741,6 +4742,126 @@ log_append_abort_log (THREAD_ENTRY * thread_p, LOG_TDES * tdes, LOG_LSA * abort_
...
+
+  LOG_TDES *tdes;
+  int tran_index;
+
+  assert (prm_get_integer_value (PRM_ID_SUPPLEMENTAL_LOG) > 0);
+
+  tran_index = LOG_FIND_THREAD_TRAN_INDEX (thread_p);
+  tdes = LOG_FIND_TDES (tran_index);
+
+  /*supplement data will be stored at undo data */
```
hornetmj, line 4770:
```suggestion
  /* supplement data will be stored at undo data */
```

src/transaction/log_manager.h
```diff
@@ -166,6 +166,13 @@ extern int log_get_next_nested_top (THREAD_ENTRY * thread_p, LOG_TDES * tdes, LO
 				    LOG_TOPOP_RANGE ** out_nxtop_range_stack);
 extern void log_append_repl_info (THREAD_ENTRY * thread_p, LOG_TDES * tdes, bool is_commit);
 
+extern void log_append_supplemental_log (THREAD_ENTRY * thread_p, SUPPLEMENT_REC_TYPE rec_type, int length,
+					 const void *data);
+extern int log_append_supplemental_lsa (THREAD_ENTRY * thread_p, SUPPLEMENT_REC_TYPE rec_type, OID * classoid,
+					LOG_LSA * undo_lsa, LOG_LSA * redo_lsa);
+
+extern int log_append_supplemental_undo_record (THREAD_ENTRY * thread_p, RECDES * undo_recdes);
```
hornetmj, line 174: How about providing an interface like log_append_supplemental_info() instead of supporting several interfaces?

src/transaction/log_manager.c
```diff
@@ -4741,6 +4742,126 @@ log_append_abort_log (THREAD_ENTRY * thread_p, LOG_TDES * tdes, LOG_LSA * abort_
...
+  char data[OR_OID_SIZE + OR_LOG_LSA_SIZE + OR_LOG_LSA_SIZE];
+
+  assert (prm_get_integer_value (PRM_ID_SUPPLEMENTAL_LOG) > 0);
+
+  switch (rec_type)
+    {
+    case LOG_SUPPLEMENT_INSERT:
+      assert (redo_lsa != NULL);
+
+      size = sizeof (OID) + sizeof (LOG_LSA);
```
hornetmj, line 4800: uses static const variable.

src/transaction/log_manager.c
```diff
@@ -4741,6 +4742,126 @@ log_append_abort_log (THREAD_ENTRY * thread_p, LOG_TDES * tdes, LOG_LSA * abort_
...
+  assert (prm_get_integer_value (PRM_ID_SUPPLEMENTAL_LOG) > 0);
+
+  LOG_ZIP *zip_undo = NULL;
+  int length = undo_recdes->length;
+  void *data = undo_recdes->data;
+  int min_size_to_compress = 255;
+
+  if (length >= min_size_to_compress)
+    {
+      zip_undo = log_zip_alloc (IO_PAGESIZE);
```
hornetmj, line 4844: Do you have any chance to use the log_append_get_zip_undo() function?

hornetmj: And this routine should be in the log_append_supplemental_log() and the compression should be executed when the data size of the supplemental log is over the limitation(min_size_to_compress). I know there are no cases except this undo record currently.

src/transaction/log_recovery.c
```diff
@@ -3998,6 +3999,7 @@ log_recovery_redo (THREAD_ENTRY * thread_p, const LOG_LSA * start_redolsa, const
 	    case LOG_DUMMY_HA_SERVER_STATE:
 	    case LOG_DUMMY_OVF_RECORD:
 	    case LOG_DUMMY_GENERIC:
+	    case LOG_SUPPLEMENTAL_INFO:
```
hornetmj, line 4002: I was wondering if applylogdb should also handle this log record type, so I checked the la_apply_log_file() and la_log_record_process() functions. But, I think we don't need to handle it. With this, do you have any opinion?

PR #2912, opened 2021-07-19
[CBRD-23911] Supplemental logs for DML, User

joohok:

1. DML : class OID, undo, redo LSA are recorded in supplemental logs
   - when DML (insert, update, delete) logs are appended. 
2. User : transaction user is recorded in supplemental logs
   - when supplemental log is first appended in the transaction 
   - when commit log is appended

### DIFF
--- a/src/storage/heap_file.c
+++ b/src/storage/heap_file.c
@@ -20548,6 +20548,10 @@ heap_delete_bigone (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * context, b
   OID overflow_oid;
   int rc;
 
+  LOG_LSA undo_lsa;
+  bool is_supplement = false;
+  LOG_TDES *tdes = NULL;
+
   /* check input */
   assert (context != NULL);
   assert (context->type == HEAP_OPERATION_DELETE);
@@ -20556,6 +20560,13 @@ heap_delete_bigone (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * context, b
   assert (context->home_page_watcher_p->pgptr != NULL);
   assert (context->overflow_page_watcher_p != NULL);
   assert (context->overflow_page_watcher_p->pgptr == NULL);
+
+  if (prm_get_integer_value (PRM_ID_SUPPLEMENTAL_LOG) > 0 && !thread_p->no_supplemental_log)
+    {
+      is_supplement = true;
+      tdes = LOG_FIND_CURRENT_TDES (thread_p);
+      LSA_SET_NULL (&undo_lsa);
+    }
 
   /* MVCC info is in overflow page, we only keep and OID in home */
   overflow_oid = *((OID *) context->home_recdes.data);
@@ -20602,6 +20613,21 @@ heap_delete_bigone (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * context, b
       log_addr.pgptr = context->overflow_page_watcher_p->pgptr;
       log_addr.vfid = &context->hfid.vfid;
       log_addr.offset = overflow_oid.slotid;
+      if (is_supplement)
+	{
+	  RECDES ovf_recdes = RECDES_INITIALIZER;
+	  if ((rc =
+	       heap_get_bigone_content (thread_p, context->scan_cache_p, COPY, &overflow_oid,
+					&ovf_recdes)) != S_SUCCESS)
+	    {
+	      return rc;	/*supplemental log affects.. existing system..? */
+	    }
+
+	  log_append_supplemental_undo_record (thread_p, &ovf_recdes);
+/*SUPPLEMENT_DELETE UNDO LSA */
+	  LSA_COPY (&undo_lsa, &tdes->tail_lsa);
+	}
+
       heap_mvcc_log_delete (thread_p, &log_addr, RVHF_MVCC_DELETE_OVERFLOW);
 
       HEAP_PERF_TRACK_LOGGING (thread_p, context);
@@ -20663,6 +20689,21 @@ heap_delete_bigone (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * context, b
 	    }
 	}
 
+      if (is_supplement)
+	{
+	  RECDES ovf_recdes = RECDES_INITIALIZER;
+	  if ((rc =
+	       heap_get_bigone_content (thread_p, context->scan_cache_p, COPY, &overflow_oid,
+					&ovf_recdes)) != S_SUCCESS)
+	    {
+	      return rc;	/*supplemental log affects.. existing system..? */
+	    }
+
+	  log_append_supplemental_undo_record (thread_p, &ovf_recdes);
+/*SUPPLEMENT_DELETE UNDO LSA */
+	  LSA_COPY (&undo_lsa, &tdes->tail_lsa);
+	}
+
       /* log operation */
       heap_log_delete_physical (thread_p, context->home_page_watcher_p->pgptr, &context->hfid.vfid, &context->oid,
 				&context->home_recdes, is_reusable, NULL);
@@ -20687,6 +20728,11 @@ heap_delete_bigone (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * context, b
       perfmon_inc_stat (thread_p, PSTAT_HEAP_BIG_DELETES);
     }
 
+  if (is_supplement)
+    {
+      log_append_supplemental_lsa (thread_p, LOG_SUPPLEMENT_DELETE, &context->class_oid, &undo_lsa, NULL);
+    }
+
   /* all ok */
   return NO_ERROR;
 }
@@ -20705,6 +20751,10 @@ heap_delete_relocation (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * contex
   OID forward_oid;
   int rc;
 
+  LOG_LSA undo_lsa;
+  bool is_supplement = false;
+  LOG_TDES *tdes = NULL;
+
   /* check input */
   assert (context != NULL);
   assert (context->type == HEAP_OPERATION_DELETE);
@@ -20712,6 +20762,13 @@ heap_delete_relocation (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * contex
   assert (context->home_page_watcher_p != NULL);
   assert (context->home_page_watcher_p->pgptr != NULL);
   assert (context->forward_page_watcher_p != NULL);
+
+  if (prm_get_integer_value (PRM_ID_SUPPLEMENTAL_LOG) > 0 && !thread_p->no_supplemental_log)
+    {
+      is_supplement = true;
+      tdes = LOG_FIND_CURRENT_TDES (thread_p);
+      LSA_SET_NULL (&undo_lsa);
+    }
 
   /* get forward oid */
   forward_oid = *((OID *) context->home_recdes.data);
@@ -21027,6 +21084,15 @@ heap_delete_relocation (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * contex
 	  forward_addr.vfid = &context->hfid.vfid;
 	  forward_addr.pgptr = context->forward_page_watcher_p->pgptr;
 	  forward_addr.offset = forward_oid.slotid;
+
+	  if (is_supplement)
+	    {
+	      /* relocation -> relocation case does not have any undo image to refer */
+	      log_append_supplemental_undo_record (thread_p, &new_forward_recdes);
+/*SUPPLEMENT_DELETE UNDO LSA */
+	      LSA_COPY (&undo_lsa, &tdes->tail_lsa);
+	    }
+
 	  heap_mvcc_log_delete (thread_p, &forward_addr, RVHF_MVCC_DELETE_REC_NEWHOME);
 
 	  HEAP_PERF_TRACK_LOGGING (thread_p, context);
@@ -21067,6 +21133,12 @@ heap_delete_relocation (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * contex
 	  forward_addr.offset = forward_oid.slotid;
 
 	  log_append_undoredo_recdes (thread_p, RVHF_DELETE, &forward_addr, &forward_recdes, NULL);
+/*SUPPLEMENT_DELETE UNDO LSA */
+	  if (is_supplement)
+	    {
+	      LSA_COPY (&undo_lsa, &tdes->tail_lsa);
+	    }
+
 	  if (heap_is_reusable_oid (context->file_type))
 	    {
 	      log_append_postpone (thread_p, RVHF_MARK_REUSABLE_SLOT, &forward_addr, 0, NULL);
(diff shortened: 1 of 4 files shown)


wrlawodms:
Supplemental log for DDL is still in progress?

joohok:
> Supplemental log for DDL is still in progress?

- yes, almost done. I just wait for reply from R2B solution to define which OID they want. I will cover the DDL supplemental logs in other PR.

wrlawodms:
> > Supplemental log for DDL is still in progress?
> 
> * yes, almost done. I just wait for reply from R2B solution to define which OID they want. I will cover the DDL supplemental logs in other PR.

OK!

src/storage/heap_file.c
```diff
@@ -20602,6 +20613,21 @@ heap_delete_bigone (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * context, b
       log_addr.pgptr = context->overflow_page_watcher_p->pgptr;
       log_addr.vfid = &context->hfid.vfid;
       log_addr.offset = overflow_oid.slotid;
+      if (is_supplement)
+	{
+	  RECDES ovf_recdes = RECDES_INITIALIZER;
+	  if ((rc =
+	       heap_get_bigone_content (thread_p, context->scan_cache_p, COPY, &overflow_oid,
```
wrlawodms, line 20620:
```suggestion
	       heap_get_bigone_content (thread_p, context->scan_cache_p, PEEK, &overflow_oid,
```
it seems that `PEEK` is enough, and this is much faster.

wrlawodms: So do other `heap_get_*()`s.

src/storage/heap_file.c
```diff
@@ -20548,6 +20548,10 @@ heap_delete_bigone (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * context, b
   OID overflow_oid;
   int rc;
 
+  LOG_LSA undo_lsa;
```
wrlawodms, line 20551: It seems to be better to use a name related to the supplemental log like `suppl_image_lsa` or `suppl_undo_lsa`.

src/storage/heap_file.c
```diff
@@ -20548,6 +20548,10 @@ heap_delete_bigone (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * context, b
...
+  bool is_supplement = false;
```
wrlawodms, line 20552: `need` isn't clearer than `is`?

src/storage/heap_file.c
```diff
@@ -20663,6 +20689,21 @@ heap_delete_bigone (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * context, b
 	    }
 	}
 
+      if (is_supplement)
```
wrlawodms, line 20692: Could you tell me why you do it for mvcc and non-mvcc respectively?

src/transaction/log_manager.c
```diff
@@ -4738,6 +4738,14 @@ log_append_commit_log_with_lock (THREAD_ENTRY * thread_p, LOG_TDES * tdes, LOG_L
 static void
 log_append_abort_log (THREAD_ENTRY * thread_p, LOG_TDES * tdes, LOG_LSA * abort_lsa)
 {
+  if (prm_get_integer_value (PRM_ID_SUPPLEMENTAL_LOG) > 0)
+    {
+      if (tdes->has_supplemental_log)
```
wrlawodms, line 4743: Is there a case in which `tdes->has_supplemental_log == true` but `prm_get_integer_value (PRM_ID_SUPPLEMENTAL_LOG) > 0`?

src/storage/heap_file.c
```diff
@@ -21410,6 +21515,13 @@ heap_delete_home (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * context, boo
 	  rec_address.pgptr = context->home_page_watcher_p->pgptr;
 	  rec_address.vfid = &context->hfid.vfid;
 	  rec_address.offset = context->oid.slotid;
+	  if (is_supplement)
```
wrlawodms, line 21518: this if-block has to be here between the initialization of rec_address and logging it?

src/storage/heap_file.c
```diff
@@ -21388,12 +21482,23 @@ heap_delete_home (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * context, boo
 		}
 	    }
 
+	  if (is_supplement)
+	    {
+	      log_append_supplemental_undo_record (thread_p, &built_recdes);
```
wrlawodms, line 21487: supl. undo record is written but not used?

src/storage/heap_file.c
```diff
@@ -21442,6 +21554,12 @@ heap_delete_home (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * context, boo
       heap_log_delete_physical (thread_p, context->home_page_watcher_p->pgptr, &context->hfid.vfid, &context->oid,
 				&context->home_recdes, is_reusable, NULL);
 
+/*SUPPLEMENT_DELETE UNDO LSA */
+      if (is_supplement)
+	{
+	  LSA_COPY (&undo_lsa, &tdes->tail_lsa);
```
wrlawodms, line 21560: Have you seen the `heap_log_delete_physical()`? It seems that it can leave another log at the last depending on `is_reusable`. Or, you're planning to traverse to find a log record containing the undo data?

src/storage/heap_file.c
```diff
@@ -21570,13 +21692,25 @@ heap_update_bigone (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * context, b
...
   assert (context->recdes_p != NULL);
   assert (context->home_page_watcher_p != NULL);
   assert (context->home_page_watcher_p->pgptr != NULL);
   assert (context->overflow_page_watcher_p != NULL);
 
+  if (prm_get_integer_value (PRM_ID_SUPPLEMENTAL_LOG) > 0 && !thread_p->no_supplemental_log)
+    {
+      is_supplement = true;
+      tdes = LOG_FIND_CURRENT_TDES (thread_p);
+      LSA_SET_NULL (&undo_lsa);
```
wrlawodms, line 21710: It would be better to use `LOG_LSA undo_lsa = LSA_INITIALIZER;` or `LOG_LSA undo_lsa = NULL_LSA` when declaring it above instead of setting null here, because it is just initialization.

src/storage/heap_file.c
```diff
@@ -20602,6 +20613,21 @@ heap_delete_bigone (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * context, b
       log_addr.pgptr = context->overflow_page_watcher_p->pgptr;
       log_addr.vfid = &context->hfid.vfid;
       log_addr.offset = overflow_oid.slotid;
+      if (is_supplement)
```
wrlawodms, line 20616: Does this if-block have to be here between the initialization of log_addr and logging it?

src/storage/heap_file.c
```diff
@@ -21630,6 +21771,10 @@ heap_update_bigone (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * context, b
       or_mvcc_set_log_lsa_to_record (context->recdes_p, logtb_find_current_tran_lsa (thread_p));
     }
 
+/* SUPPLEMENT_UPDATE UNDO LSA
```
wrlawodms, line 21774: Which code is this comment for?

src/storage/heap_file.c
```diff
@@ -21641,6 +21786,18 @@ heap_update_bigone (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * context, b
...
 	}
+
+/*SUPPLEMENT_UPDATE REDO LSA (mvcc), UNDO,REDO (non-mvcc) bigone to bigone */
+      if (is_supplement)
+	{
+	  LSA_COPY (&redo_lsa, &tdes->tail_lsa);
+
+	  if (!is_mvcc_op)
+	    {
+	      LSA_COPY (&undo_lsa, &tdes->tail_lsa);
```
wrlawodms, line 21797: same lsa to redo_lsa. You're going to traverse from the same lsa?

src/transaction/log_manager.c
```diff
@@ -4738,6 +4738,14 @@ log_append_commit_log_with_lock (THREAD_ENTRY * thread_p, LOG_TDES * tdes, LOG_L
 static void
 log_append_abort_log (THREAD_ENTRY * thread_p, LOG_TDES * tdes, LOG_LSA * abort_lsa)
 {
+  if (prm_get_integer_value (PRM_ID_SUPPLEMENTAL_LOG) > 0)
```
wrlawodms, line 4741: I think you put this in `log_abort_local()` to be consistent with the commit.

wrlawodms (commented):
With these, I have some opinions.

1. You should consider that system classes such as catalog classes, system catalog, and root class use these heap operations.
2. While INSERT writes the supplemental log in heap_insert_logical(), UPDATE, DELETE write it in the inner functions. For readability, they should be on the same level or put comments and I recommend using `HEAP_OPERATION_CONTEXT` to do it.
3. You put a logging of the user name in log_commit_local() and log_append_abort_log() to catch the end of transaction. It would be better to put it in the same level of functions like log_commit_local() and log_commit_commit() or log_append_commit_log() and log_append_abort_log(). Plus, it isn't logged when using 2PC. Is it okay?

src/storage/heap_file.c
joohok: Ok, that's more clear

src/storage/heap_file.c
joohok: OK, I will change it. it seems more clear.

src/storage/heap_file.c
joohok: Thanks, I will change it.

src/storage/heap_file.c
joohok: Oh thanks. I'll move this part above 'if (is_mvcc_op)' so only do it once, whether mvcc or non-mvcc

src/storage/heap_file.c
joohok: I'll fix it. thanks

src/storage/heap_file.c
joohok: No, I understand what you mean. it can reduce the readability, so I'll move it to end of the else-block (after the heap_mvcc_log_delete).

src/storage/heap_file.c
joohok: OK, I will add the initialization

src/storage/heap_file.c
joohok: Yes. for non-mvcc update (bigone to bigone), I have to traverse from same lsa. because the logs are appended inside the heap_ovf_update(). Traverse from the lsa, and find the REDO or UNDO log records until it meets the LOG_DUMMY_OVF_RECORD.

src/transaction/log_manager.c
joohok: Ok, I will fix it.

src/transaction/log_manager.c
joohok: you mean, there is no case that 
- tdes->has_supplemental_log == true 
- prm_get_integer_value (PRM_ID_SUPPLEMENTAL_LOG) == 0 ? 

No, So checking only 'tdes->has_supplemental_log == true' will be enough.

joohok:
> With these, I have some opinions.
> 
> 1. You should consider that system classes such as catalog classes, system catalog, and root class use these heap operations.

- I will use the macro that checks the system classes, and make a exception for system class not to append supplemental logs. 

> 2. While INSERT writes the supplemental log in heap_insert_logical(), UPDATE, DELETE write it in the inner functions. For readability, they should be on the same level or put comments and I recommend using `HEAP_OPERATION_CONTEXT` to do it.

OK, I will add LOG_LSA *supp_undo_lsa, *supp_redo_lsa into HEAP_OPERATION_CONTEXT, and move the  'log_append_supplemental_lsa()' to heap_insert/update/delete_logical(). 

> 3. You put a logging of the user name in log_commit_local() and log_append_abort_log() to catch the end of transaction. It would be better to put it in the same level of functions like log_commit_local() and log_commit_commit() or log_append_commit_log() and log_append_abort_log(). Plus, it isn't logged when using 2PC. Is it okay?

Ok, I will put those logging in the same level. 
Actually, I didn't consider 2PC. 
It will be better to move logging function to log_append_commit/abort_log to handle 2PC.

src/storage/heap_file.c
joohok: Oh.. this is just trash comment during implementation. I will remove it

src/storage/heap_file.c
joohok: Oh, Thanks. 
heap_log_delete_physical() has an output argument (LOG_LSA * undo_lsa (out)), which stores  lsa before appending logs for reusable slot. I'll apply your review using this argument.

src/storage/heap_file.c
hornetmj: Is there any reason to append the supplemental undo record in advance? I think it is OK to append it just before the supplemental LSA.

src/storage/heap_file.c
```diff
@@ -22428,12 +22694,31 @@ heap_insert_logical (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * context,
...
+  bool is_supplement = false;
+  LOG_TDES *tdes = NULL;
+
   /* check required input */
   assert (context != NULL);
   assert (context->type == HEAP_OPERATION_INSERT);
   assert (context->recdes_p != NULL);
   assert (!HFID_IS_NULL (&context->hfid));
 
+  if (prm_get_integer_value (PRM_ID_SUPPLEMENTAL_LOG) > 0 && !thread_p->no_supplemental_log)
```
hornetmj, line 22707: How about making an inline function or a macro instead of this line?

I think it is more readable if you leave a comment about the meaning of prm_get_integer_value (PRM_ID_SUPPLEMENTAL_LOG) and thread_p->no_supplemental_log in the inline function or macro.

src/storage/heap_file.c
```diff
@@ -22428,12 +22694,31 @@ heap_insert_logical (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * context,
...
+    {
+      is_supplement = true;
+      tdes = LOG_FIND_CURRENT_TDES (thread_p);
+      LSA_SET_NULL (&redo_lsa);
+
+      if (!tdes->has_supplemental_log)
+	{
+	  const char *user = tdes->client.get_db_user ();
+	  int length = strlen (user);
+	  log_append_supplemental_log (thread_p, LOG_SUPPLEMENT_TRAN_USER, length, user);
```
hornetmj, line 22717:
```suggestion
	  log_append_supplemental_log (thread_p, LOG_SUPPLEMENT_TRAN_USER, strlen (tdes->client.get_db_user ()), tdes->client.get_db_user ());
```

src/storage/heap_file.c
```diff
@@ -22486,6 +22771,12 @@ heap_insert_logical (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * context,
       goto error;
     }
 
+/*SUPPLEMENT_INSERT REDO LSA, INSERT REC_BIGONE  */
```
hornetmj, line 22774: Please, check all comments you wrote.

1. /*comment */ -> /\* comment */
2. there is one more space in the middle of the comments.
3. there is not enough explanation. So I was confused about whether it is needed or not.
4. there are some capital letters in the middle of the comments.

I hope you only leave comments that you really need and they include enough information.

src/storage/heap_file.c
```diff
@@ -22486,6 +22771,12 @@ heap_insert_logical (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * context,
...
+  if (is_supplement && heap_is_big_length (context->recdes_p->length))
+    {
+      LSA_COPY (&redo_lsa, &tdes->tail_lsa);
```
hornetmj, line 22777: This information(redo_lsa) is overwritten by the next one.

src/storage/heap_file.c
joohok: I agree with your opinion. I will make an inline function for better readability.

PR #2917, opened 2021-07-22
[LETS-161] Proper default identity for thread entry

cristiarg:
http://jira.cubrid.org/browse/LETS-161

By default, a thread in the thread pool is created with the `TT_WORKER` identity.
This is misleading when analyzing perf results because the perf module identifies `TT_WORKER` as belonging to the `PERF_MODULE_USER` module type - as per `perfmon_get_module_type` function.
Construct a thread entry with `TT_NONE` and initialize with `TT_MASTER`.
Thus, perf collected on a default constructed thread entry will be assigned to the `PERF_MODULE_SYSTEM` module type.

### DIFF
--- a/src/thread/thread_entry.cpp
+++ b/src/thread/thread_entry.cpp
@@ -77,7 +77,7 @@ namespace cubthread
   entry::entry ()
   // public:
     : index (-1)
-    , type (TT_WORKER)
+    , type (TT_MASTER)
     , emulate_tid ()
     , client_id (-1)
     , tran_index (NULL_TRAN_INDEX)
--- a/src/thread/thread_entry.cpp
+++ b/src/thread/thread_entry.cpp
@@ -77,7 +77,7 @@ namespace cubthread
   entry::entry ()
   // public:
     : index (-1)
-    , type (TT_MASTER)
+    , type (TT_NONE)
     , emulate_tid ()
     , client_id (-1)
     , tran_index (NULL_TRAN_INDEX)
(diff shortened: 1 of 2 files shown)


src/thread/thread_entry.cpp
```diff
@@ -77,7 +77,7 @@ namespace cubthread
   entry::entry ()
   // public:
     : index (-1)
-    , type (TT_WORKER)
+    , type (TT_MASTER)
```
cristiarg: @ribram Somehow I feel this change is either incomplete or incorrect.
May be completed/corrected with:
- change `TT_MASTER` to `TT_MAIN`, in order to not confuse with `cub_master` related functionality?
- add an explicit `thread_type` to the `cubthread::entry` ctor and explicitly change all over - or request everyone to explicitly set a thread type
With this change, from now on, all non-initialized thread entries in thread manager will be `TT_MASTER` whereas they were `TT_WORKER` (more proper imho).

ribram: > change TT_MASTER to TT_MAIN, in order to not confuse with cub_master related functionality?

It does not matter, the difference is in the context. This is the master thread of cub_server process, the cub_master is entirely another process.

>add an explicit thread_type to the cubthread::entry ctor and explicitly change all over - or request everyone to explicitly set a thread type
With this change, from now on, all non-initialized thread entries in thread manager will be TT_MASTER whereas they were TT_WORKER (more proper imho).

What are your thoughts on initializing as TT_NONE in the cubthread::entry constructor and setting TT_MASTER type in `cubthread::initialize()`?

ribram (approved):

PR #2920, opened 2021-07-22
[LETS-117] Transaction server with remote storage does not flush log header

ribram:
http://jira.cubrid.org/browse/LETS-117

Never write log header page to the local disk of transaction server with remote storage.


cristiarg (commented):
The call to `logpb_flush_header` in `logpb_next_append_page`  is not guarded as specified in the issue.
Also I do not seem to find anything related to "Ignore recovery references (transaction server recovery is out of scope)".

ribram:
@cristiarg The call to logpb_flush_header in logpb_next_append_page was already guarded in one of the previous patches.

There are some log header flushes in the recovery code; handling recovery is however left for later and therefore these flushes are not guarded (e.g. in log_recovery_redo, log_recovery_undo).

cristiarg:
> @cristiarg The call to logpb_flush_header in logpb_next_append_page was already guarded in one of the previous patches.
> 
> There are some log header flushes in the recovery code; handling recovery is however left for later and therefore these flushes are not guarded (e.g. in log_recovery_redo, log_recovery_undo).

@ribram
My bad about `logpb_next_append_page`, I see it now.
Ok about the

cristiarg (approved):

PR #2921, opened 2021-07-26
[CBRD-24010] Predicate push dose not work in nested subqueries.

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-24010
http://jira.cubrid.org/browse/CBRD-24001

This change solves the problem below.
1. nested subquery problem (CBRD-24010)
2. indirect predicate problem (CBRD-24001)
3. Perform predicate push after changing outer join to inner join (newly discovered problem)

**CBRD-24010**
The predicate push should be performed from the outermost to the inner subquery.
```
select ... 
   from ( select ... 
               from (select ...
                        from 
                        c.col1 =1  <== second push
                      ) c
           c.col1 =1  <== first push
           ) b      
  where b.col1 =1
```
When walking the tree, mq_push_path () is invoked in post order. So, it is executed from the innermost subquery.

mq_translate_helper()
```
node = parser_walk_tree (parser, node, NULL, NULL, mq_push_paths, NULL);

mq_push_paths()
  mq_copypush_sargable_terms() <== predicate push
```
I fix that mq_copypush_sargable_terms() is invoked in pre-order.

**CBRD-24001**
A predicate push is performed before a transitive predicate generation. This causes problems.
I move the location of mq_copypush_sargable_terms() in mq_optimize().
```
mq_translate
mq_translate_helper
  mq_push_paths
    mq_check_rewrite_select
      mq_copypush_sargable_terms  <== old location
  mq_optimize  
    qo_optimize_queries
      qo_reduce_equality_terms <== Transitive Predicate Generation
      qo_rewrite_outerjoin <== rewrite outerjoin to innerjoin
      mq_copypush_sargable_terms  <== new location, invoked in pre-order
```

### DIFF
--- a/src/parser/view_transform.c
+++ b/src/parser/view_transform.c
@@ -3198,6 +3198,14 @@ mq_copypush_sargable_terms_helper (PARSER_CONTEXT * parser, PT_NODE * statement,
 
   for (term = statement->info.query.q.select.where; term; term = term->next)
     {
+      /* check for on_cond term */
+      assert (term->node_type == PT_EXPR || term->node_type == PT_VALUE);
+      if ((term->node_type == PT_EXPR && term->info.expr.location > 0)
+	  || (term->node_type == PT_VALUE && term->info.expr.location > 0))
+	{
+	  continue;		/* do not copy-push on_cond-term */
+	}
+
       /* check for nullable-term */
       if (term->node_type == PT_EXPR)
 	{
--- a/src/parser/view_transform.c
+++ b/src/parser/view_transform.c
@@ -213,7 +213,6 @@ static void pt_copypush_terms (PARSER_CONTEXT * parser, PT_NODE * spec, PT_NODE 
 			       FIND_ID_TYPE type);
 static int mq_copypush_sargable_terms_helper (PARSER_CONTEXT * parser, PT_NODE * statement, PT_NODE * spec,
 					      PT_NODE * new_query, FIND_ID_INFO * infop);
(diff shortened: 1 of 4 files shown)


shparkcubrid:
ci/circleci: test_sql fail
sql/_13_issues/_20_2h/cases/cbrd_23665.sql - linux_sql_64bit_release

```
max(b.b)>'cubrid' ==> (max(b.b)>'cubrid')
```
This occurs when the operation of expression is changed from PT_GT to PT_RANGE. It's not a bug. It seems better to change the answer sheet than to fix the print of PT_RANGE. If the print of PT_RANGE is fixed , the more test cases is needed to edit.

src/parser/view_transform.c
```diff
@@ -3199,6 +3198,14 @@ mq_copypush_sargable_terms_helper (PARSER_CONTEXT * parser, PT_NODE * statement,
 
   for (term = statement->info.query.q.select.where; term; term = term->next)
     {
+      /* check for on_cond term */
+      assert (term->node_type == PT_EXPR || term->node_type == PT_VALUE);
+      if ((term->node_type == PT_EXPR && term->info.expr.location > 0)
+	  || (term->node_type == PT_VALUE && term->info.expr.location > 0))
+	{
```
beyondykk9, line 3205: If the node_type is PT_VALUE the info.expr is not valid but th info.value.... is available. Please check if it's correct.

shparkcubrid: thanks. I fixed it.

beyondykk9 (approved):

hgryoo (approved):

youngjinj (approved):

ctshim (approved):

PR #2923, opened 2021-07-28
[LETS-164] Fix access on null active tran server connection

adrianbrab:
http://jira.cubrid.org/browse/LETS-164

When page server stops and tries to disconnect from active tran server, it logs a message that accesses the connection. If the connection is null, it crashes.
Fixed by adding a null check with a different error log message.


ribram (approved):

src/server/page_server.cpp
```diff
@@ -75,6 +75,11 @@ void page_server::set_active_tran_server_connection (cubcomm::channel &&chn)
 
 void page_server::disconnect_active_tran_server ()
 {
+  if (m_active_tran_server_conn == NULL)
```
hornetmj:
```suggestion
  if (m_active_tran_server_conn == nullptr)
```

adrianbrab: changed now

PR #2924, opened 2021-07-28
[LETS-165] Finalize "unknown" server type

adrianbrab:
http://jira.cubrid.org/browse/LETS-165

If server boot has an error before server type is initialized, the global server type may remain unknown. When server type is finalized, this case does not require any disconnect function call.


src/base/server_type.cpp
```diff
@@ -94,7 +94,7 @@ void finalize_server_type ()
     {
       ats_Gl.disconnect_page_server ();
     }
-  else
+  if (get_server_type() == SERVER_TYPE_PAGE)
```
ribram:
```suggestion
  else if (get_server_type() == SERVER_TYPE_PAGE)
```

ribram (approved):

src/base/server_type.cpp
```diff
@@ -94,7 +94,7 @@ void finalize_server_type ()
     {
       ats_Gl.disconnect_page_server ();
     }
-  else
+  else if (get_server_type() == SERVER_TYPE_PAGE)
     {
       assert (get_server_type () == SERVER_TYPE_PAGE);
```
hornetmj:
```suggestion
```

adrianbrab: removed now

src/base/server_type.cpp
```diff
@@ -99,9 +99,8 @@ void finalize_server_type ()
     {
       ats_Gl.disconnect_page_server ();
     }
-  else
+  else if (get_server_type() == SERVER_TYPE_PAGE)
     {
-      assert (get_server_type () == SERVER_TYPE_PAGE);
       ps_Gl.disconnect_active_tran_server ();
     }
```
ribram, line 105: Nitpicking:
```suggestion
    }
  else
    {
      assert (get_server_type () == SERVER_TYPE_UNKNOWN);
    }
```

adrianbrab: added now

PR #2925, opened 2021-07-29
[CBRD-24012] Optimizing dblink: rewriting as a subquery

beyondykk9:
http://jira.cubrid.org/browse/CBRD-24012

The dblink query's performance is afftected by data transfering from dblink's host.

Commonly, JOIN operation is performed by scan loop and the dblink query is also done as the way.

While scan looping, the data from remote host might be repeatedly transferred if the dblink is joined as inner loop. And the join more than 3 may be degraded the performance by this reason.

We need to rewrite the query includes dblink as a subquery to get the whole data from remote host at once. By doing this, the connection of dblink can be more safe.

### DIFF
--- a/src/optimizer/query_rewrite.c
+++ b/src/optimizer/query_rewrite.c
@@ -6704,6 +6704,37 @@ qo_move_on_clause_of_explicit_join_to_where_clause (PARSER_CONTEXT * parser, PT_
 }
 
 /*
+ * qo_rewrite_dblink_as_subquery () - rewrite dblink as a subquery
+ *   return: PT_NODE *
+ *   parser(in): parser environment
+ *   node(in): possible dblink query
+ */
+static void
+qo_rewrite_dblink_as_subquery (PARSER_CONTEXT * parser, PT_NODE * node)
+{
+  PT_NODE *spec, *derived;
+  PT_NODE *derived_table;
+
+  for (spec = node->info.query.q.select.from; spec; spec = spec->next)
+    {
+      if ((derived_table = spec->info.spec.derived_table)
+	  && spec->info.spec.derived_table_type == PT_DERIVED_DBLINK_TABLE
+	  && node->info.query.is_subquery != PT_IS_DBLINK_SUBQUERY)
+	{
+	  derived = mq_rewrite_dblink_as_derived (parser, derived_table);
+	  if (derived == NULL)
+	    {
+	      break;
+	    }
+	  /* for avoiding endless recursive call, set the is_subquery flag */
+	  derived->info.query.is_subquery = PT_IS_DBLINK_SUBQUERY;
+	  spec->info.spec.derived_table = derived;
+	  spec->info.spec.derived_table_type = PT_IS_SUBQUERY;
+	}
+    }
+}
+
+/*
  * qo_optimize_queries () - checks all subqueries for rewrite optimizations
  *   return: PT_NODE *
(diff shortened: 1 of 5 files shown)


src/optimizer/query_rewrite.c
```diff
@@ -6703,6 +6703,37 @@ qo_move_on_clause_of_explicit_join_to_where_clause (PARSER_CONTEXT * parser, PT_
...
+	  && spec->info.spec.derived_table_type == PT_DERIVED_DBLINK_TABLE
+	  && node->info.query.is_subquery != PT_IS_DBLINK_SUBQUERY)
+	{
+	  derived = mq_rewrite_dblink_as_derived (parser, derived_table);
+	  if (derived == NULL)
+	    {
+	      break;
+	    }
+	  /* for avoiding endless recursive call, set the is_subquery flag */
+	  derived->info.query.is_subquery = PT_IS_DBLINK_SUBQUERY;
```
shparkcubrid: because of pre-order, endless recursive call occur. 
PT_IS_DBLINK_SUBQUERY is unnessary if it is post-order.

beyondykk9: Ok, I fixed it.

src/parser/view_transform.c
```diff
@@ -4849,6 +4959,44 @@ mq_push_paths (PARSER_CONTEXT * parser, PT_NODE * statement, void *void_arg, int
...
+    {
+      return node;
+    }
+
+  for (spec = node->info.query.q.select.from; spec; spec = spec->next)
+    {
+      if ((derived_table = spec->info.spec.derived_table)
+	  && spec->info.spec.derived_table_type == PT_DERIVED_DBLINK_TABLE)
+	{
+	  derived = mq_rewrite_dblink_as_derived (parser, derived_table);
```
ctshim, line 4980: How about removing the derived_table variable and doing something like below?
Evaluation through substitution statements like the one below can be misleading.
(derived_table = spec->info.spec.derived_table)

```
if (spec->info.spec.derived_table && 
     spec->info.spec.derived_table_type == PT_DERIVED_DBLINK_TABLE)
{ 
   derived = mq_rewrite_dblink_as_derived (parser, spec->info.spec.derived_table);

```

beyondykk9: I agree with you but the readability is not good. I'd like to keep the style.

ctshim (approved):

src/parser/view_transform.c
```diff
@@ -3610,6 +3610,113 @@ mq_rewrite_vclass_spec_as_derived (PARSER_CONTEXT * parser, PT_NODE * statement,
...
+  /* construct new spec We are now copying the query and updating the spec_id references */
+  spec = parser_new_node (parser, PT_SPEC);
+  if (spec == NULL)
+    {
+      PT_INTERNAL_ERROR (parser, "allocate new node");
+      goto exit_on_error;
+    }
+
+  derived = query;
+  derived = mq_reset_ids_in_statement (parser, derived);
```
shparkcubrid: I think mq_reset_ids_in_statement () is not needed for dblink node.

beyondykk9: Ok, I aggre with you.

shparkcubrid (approved):

hgryoo (approved):

youngjinj (approved):

PR #2928, opened 2021-07-29
[LETS-162] Inline log reading functionality

ribram:
http://jira.cubrid.org/browse/LETS-162

Log reading functions are often called during performance sensitive operations. Inline all the operations reading from log to guarantee optimum results.

Changes:

- Moved the implementation from log_reader.cpp to log_reader.hpp. All functions are inlined.
- log_reader.cpp is deleted and removed from cubrid/cubridsa projects.
- Removed include log_reader.hpp from log_impl.h; added log_reader.hpp to all files where it's used.
- Updated page fetcher unit tests.


src/transaction/log_reader.hpp
```diff
@@ -86,54 +86,56 @@ class log_reader final
...
+    LOG_PAGE *log_pgptr);
 
 
 /* implementation
  */
+#include "log_impl.h"
+#include "thread_manager.hpp"
 
 template <typename T>
 const typename std::remove_reference<T>::type *log_reader::reinterpret_cptr () const
```
cristiarg: all log_reader implementations should be in the same order as the declarations

src/transaction/log_replication.cpp
```diff
@@ -20,6 +20,7 @@
 
 #include "btree_load.h"
 #include "log_impl.h"
+#include "log_reader.hpp"
```
cristiarg, line 23: is this needed as long as `log_reader` is included in `log_recovery_redo_parallel` header? or is it just for explicitness?

unit_tests/log/test_main_data_page_fetcher.cpp
```diff
@@ -176,19 +176,16 @@ delete_page (PAGE_PTR page_ptr)
...
 }
 
-// Mock some of the functionality
-log_reader::log_reader () = default; // needed by log_page_fetch_task::execute
-
 int
-log_reader::set_lsa_and_fetch_page (const log_lsa &lsa, fetch_mode fetch_page_mode)
+logpb_fetch_page (THREAD_ENTRY *, const LOG_LSA *, LOG_CS_ACCESS_MODE, LOG_PAGE *)
 {
   return NO_ERROR;
```
cristiarg, line 183: why no `assert (false)` here?

unit_tests/log/test_main_log_page_fetcher.cpp
```diff
@@ -138,18 +138,17 @@ void test_env::on_receive_log_page (LOG_PAGEID page_id, const LOG_PAGE *log_page
...
-  delete log_page;
 }
 
-int log_reader::set_lsa_and_fetch_page (const log_lsa &lsa, fetch_mode fetch_page_mode)
+int
+logpb_fetch_page (THREAD_ENTRY *, const LOG_LSA *log_lsa, LOG_CS_ACCESS_MODE, LOG_PAGE *log_pgptr)
 {
-  std::unique_lock<std::mutex> lock (g_log_page_fetcher_test_data.map_mutex);
-  if (g_log_page_fetcher_test_data.page_ids_requested[lsa.pageid].require_log_page_valid)
+  if (g_log_page_fetcher_test_data.page_ids_requested[log_lsa->pageid].require_log_page_valid)
```
cristiarg, line 149: shouldn't there be a
```
std::lock_guard<std::mutex> lock (g_log_page_fetcher_test_data.map_mutex);
```
here - as it was in the `log_reader::get_page` mock implementation?

src/transaction/log_replication.cpp
ribram: it is not necessary, just for explicitness

src/transaction/log_reader.hpp
ribram: keeping a mirrored implements and declaration is a great effort with little benefit. devs are usually relying on IDE navigation to go from declarations to the corresponding implementations.

but the template functions are moved at the end, because it is usual to start the implementation with the constructor/destructor functions.

unit_tests/log/test_main_data_page_fetcher.cpp
ribram: added

unit_tests/log/test_main_log_page_fetcher.cpp
ribram: done

cristiarg (approved):

PR #2929, opened 2021-07-29
[LETS-166] refactor redo_job_impl as a non-template class

cristiarg:
http://jira.cubrid.org/browse/LETS-166

Profiling showed that the overhead incurred by the allocator constantly constructing (and, probably, destructing) a large amount of objects (in particular `redo_job_impl` instances) is considerable.

Reimplement the `redo_job_impl` as a non-template class. This is needed for further refactoring when `redo_job_impl` will be used in a reusable manner (sic!) in order to minimize the overhead incurred by the memory allocator constantly constructing/destructing a large number of such objects.

### DIFF
--- a/src/transaction/log_recovery_redo_parallel.cpp
+++ b/src/transaction/log_recovery_redo_parallel.cpp
@@ -673,4 +673,116 @@ namespace cublog
 	m_worker_pool->execute (task);
       }
   }
+
+
+  /*********************************************************************
+   * redo_job_impl - definition
+   *********************************************************************/
+
+  redo_job_impl::redo_job_impl (const log_lsa *a_end_redo_lsa, bool force_each_page_fetch)
+    : redo_parallel::redo_job_base (VPID_INITIALIZER, NULL_LSA)
+    , m_end_redo_lsa { a_end_redo_lsa }
+    , m_log_reader_page_fetch_mode (force_each_page_fetch
+				    ? log_reader::fetch_mode::FORCE
+				    : log_reader::fetch_mode::NORMAL)
+    , m_log_rtype { LOG_SMALLER_LOGREC_TYPE }
+  {
+  }
+
+  void redo_job_impl::reset (VPID a_vpid, const log_lsa &a_rcv_lsa, LOG_RECTYPE a_log_rtype)
+  {
+    set_vpid (a_vpid);
+    set_log_lsa (a_rcv_lsa);
+    assert (a_log_rtype != LOG_SMALLER_LOGREC_TYPE && a_log_rtype != LOG_LARGER_LOGREC_TYPE);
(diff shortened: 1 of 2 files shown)


src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -673,4 +673,116 @@ namespace cublog
...
+      default:
+	assert (false);
+	break;
+      }
+
+    return NO_ERROR;
+  }
+
+
+
```
```suggestion
```

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -673,4 +673,116 @@ namespace cublog
...
+     */
+    perfmon_counter_timer_raii_tracker perfmon { PSTAT_LOG_REDO_ASYNC };
+
+    const auto &rcv_lsa = get_log_lsa ();
+    const int err_set_lsa_and_fetch_page
+      = log_pgptr_reader.set_lsa_and_fetch_page (rcv_lsa, m_log_reader_page_fetch_mode);
+    if (err_set_lsa_and_fetch_page != NO_ERROR)
+      {
+	return err_set_lsa_and_fetch_page;
+      }
```
ribram: Nitpicking: pick shorter names.
```suggestion
    const int err_fetch = log_pgptr_reader.set_lsa_and_fetch_page (rcv_lsa, m_log_reader_page_fetch_mode);
    if (err_fetch != NO_ERROR)
      {
	return err_fetch;
      }
```

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -673,4 +673,116 @@ namespace cublog
...
+    const auto &rcv_vpid = get_vpid ();
+    switch (m_log_rtype)
+      {
+      case LOG_REDO_DATA:
+      {
+	log_pgptr_reader.advance_when_does_not_fit (sizeof (log_rec_redo));
+	const log_rec_redo log_rec = log_pgptr_reader.reinterpret_copy_and_add_align<log_rec_redo> ();
+	log_rv_redo_record_sync<log_rec_redo> (thread_p, log_pgptr_reader, log_rec, rcv_vpid, rcv_lsa,
+					       m_end_redo_lsa, m_log_rtype,
+					       undo_unzip_support, redo_unzip_support);
```
ribram: This sequence of three rows can be thrown into a template helper function.

```c
template <typename T>
void
redo_job_impl::read_record_and_redo (...)
{
  log_pgptr_reader.advance_when_does_not_fit (sizeof (T));
  const T log_rec = log_pgptr_reader.reinterpret_copy_and_add_align<T> ();
  log_rv_redo_record_sync<T> (...);
}
```

ribram (approved):
Looks good. The only comments are for small improvements.

PR #2934, opened 2021-07-30
[CBRD-24014] Problem that nullable predicate is not pushable.

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-24014

fix that only in case of outer join spec, nullable predicate cannot be pushed.

### DIFF
--- a/src/parser/view_transform.c
+++ b/src/parser/view_transform.c
@@ -3198,6 +3198,14 @@ mq_copypush_sargable_terms_helper (PARSER_CONTEXT * parser, PT_NODE * statement,
 
   for (term = statement->info.query.q.select.where; term; term = term->next)
     {
+      /* check for on_cond term */
+      assert (term->node_type == PT_EXPR || term->node_type == PT_VALUE);
+      if ((term->node_type == PT_EXPR && term->info.expr.location > 0)
+	  || (term->node_type == PT_VALUE && term->info.expr.location > 0))
+	{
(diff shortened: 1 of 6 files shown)


shparkcubrid:
Commits prior to 89e6b42 are not relevant to this PR.

src/parser/view_transform.c
```diff
@@ -3226,7 +3228,7 @@ mq_copypush_sargable_terms_helper (PARSER_CONTEXT * parser, PT_NODE * statement,
 
 	  if (nullable_cnt)
 	    {
-	      continue;		/* do not copy-push nullable-term */
+	      continue;		/* do not copy-push nullable-term of outer spec */
```
beyondykk9, line 3231: Is guaranteed that the nullable term is related to outer join?
Some term might not be related to the outer join and the term needs to be pushed.
So, I think the only nullable term not related to any outer join should be pushed.

shparkcubrid: No. but If the spec is an outer join type, nullable terms no longer need to be checked for pushability.
Other specs( terms) will proceed in another mq_copypush_sargable_terms() call.

beyondykk9: Ok, I understood.

beyondykk9 (approved):

youngjinj (approved):

ctshim (approved):

hgryoo (approved):

PR #2935, opened 2021-07-30
[LETS-131] Adapt disk volume add/expand to transaction server with remote storage

ribram:
http://jira.cubrid.org/browse/LETS-131

Do not create/extend the physical file of a volume on the transaction server with remote storage disk. Disable all log and data flushing to disk.

Assume all volume ID's are consecutive when getting the ID of previous volume.

disk_is_page_sector_reserved_with_debug_crash does not look for volume descriptor on TSRS. It was also rewritten with a unified exit routine.

### DIFF
--- a/src/storage/disk_manager.c
+++ b/src/storage/disk_manager.c
@@ -567,9 +567,12 @@ disk_format (THREAD_ENTRY * thread_p, const char *dbname, VOLID volid, DBDEF_VOL
     }
   fault_inject_random_crash ();
 
-  /* this log must be flushed. */
-  logpb_force_flush_pages (thread_p);	//todo
-  fault_inject_random_crash ();
+  if (!is_tran_server_with_remote_storage ())
+    {
+      /* this log must be flushed. */
+      logpb_force_flush_pages (thread_p);
+      fault_inject_random_crash ();
+    }
 
   if (!(is_tran_server_with_remote_storage () && ext_info->voltype == DB_PERMANENT_VOLTYPE))
     {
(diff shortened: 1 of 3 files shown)


src/storage/disk_manager.c
```diff
@@ -565,17 +567,24 @@ disk_format (THREAD_ENTRY * thread_p, const char *dbname, VOLID volid, DBDEF_VOL
...
-  /* create and initialize the volume. recovery information is initialized in every page. */
-  vdes = fileio_format (thread_p, dbname, vol_fullname, volid, extend_npages, vol_purpose == DB_PERMANENT_DATA_PURPOSE,
-			false, false, IO_PAGESIZE, kbytes_to_be_written_per_sec, false);
-  if (vdes == NULL_VOLDES)
+  if (!(is_tran_server_with_remote_storage () && ext_info->voltype == DB_PERMANENT_VOLTYPE))
     {
-      ASSERT_ERROR_AND_SET (error_code);
-      return error_code;
+      /* create and initialize the volume. recovery information is initialized in every page. */
+      vdes =
```
cristiarg: alignment here?

src/storage/disk_manager.c
```diff
@@ -4126,62 +4144,62 @@ DISK_ISVALID
...
 
-  old_check_interrupt = logtb_set_check_interrupt (thread_p, false);
-  old_wait_msecs = xlogtb_reset_wait_msecs (thread_p, LK_INFINITE_WAIT);
+  bool old_check_interrupt = logtb_set_check_interrupt (thread_p, false);
+  int old_wait_msecs = xlogtb_reset_wait_msecs (thread_p, LK_INFINITE_WAIT);
 
-  if (fileio_get_volume_descriptor (volid) == NULL_VOLDES || pageid < 0)
+  auto exit_routine =
+    [&thread_p, &debug_crash, &page_volheader, &old_check_interrupt, &old_wait_msecs] (DISK_ISVALID isvalid) {
+    assert (isvalid != DISK_INVALID || !debug_crash);
```
cristiarg: A bit convoluted with this assert and the subsequent condition (even though function `disk_is_sector_reserved`'s return value does conform to it). It would be clearer if let to the original form + lambda function call.

cristiarg (approved):

PR #2936, opened 2021-08-02
[LETS-174] Quick find tran_index for tranid during log_recovery_redo

ribram:
http://jira.cubrid.org/browse/LETS-174

When LOG_COMMIT/LOG_ABORT records are found during log_recovery_redo, the transaction descriptors with the same tranid are freed. The transaction descriptor is searched by tranid in the transaction table. When tranid is not found (very likely), the entire transaction table is iterated. Depending on the configuration, the transaction table may have many thousands entries.

A quick optimization is to avoid looking up when tranid's are not in the transaction table. In most cases, these tranid's preceed the tranid's in the transaction table. The fix is to save the smallest tranid in the transaction table and compare the committed tranid with this. If the committed tranid is smaller, then searching in the transaction table is no longer necessary.

### DIFF
--- a/src/transaction/log_recovery.c
+++ b/src/transaction/log_recovery.c
@@ -75,6 +75,7 @@ static int log_rv_undoredo_partial_changes_recursive (THREAD_ENTRY * thread_p, O
 static void log_rv_simulate_runtime_worker (THREAD_ENTRY * thread_p, LOG_TDES * tdes);
 static void log_rv_end_simulation (THREAD_ENTRY * thread_p);
 static void log_find_unilaterally_largest_undo_lsa (THREAD_ENTRY * thread_p, LOG_LSA & max_undo_lsa);
+static TRANID log_rv_get_min_trantable_tranid ();
 
 /*
  * CRASH RECOVERY PROCESS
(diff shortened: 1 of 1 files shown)


src/transaction/log_recovery.c
```diff
@@ -981,6 +1000,7 @@ log_recovery_redo (THREAD_ENTRY * thread_p, log_recovery_context & context)
   LOG_ZIP *redo_unzip_ptr = NULL;
   bool is_mvcc_op = false;
   const bool force_each_log_page_fetch = false;
+  TRANID min_trantable_tranid = log_rv_get_min_trantable_tranid ();
```
cristiarg:
```suggestion
  const TRANID min_trantable_tranid = log_rv_get_min_trantable_tranid ();
```

src/transaction/log_recovery.c
```diff
@@ -940,6 +941,24 @@ log_recovery_needs_skip_logical_redo (THREAD_ENTRY * thread_p, TRANID tran_id, L
   return false;
 }
 
+TRANID
+log_rv_get_min_trantable_tranid ()
+{
+  TRANID min_tranid = NULL_TRANID;
+  for (int tran_index = 1; tran_index < log_Gl.trantable.num_total_indices; ++tran_index)
+    {
+      TRANID tranid = log_Gl.trantable.all_tdes[tran_index]->trid;
```
cristiarg:
```suggestion
      const TRANID tranid = log_Gl.trantable.all_tdes[tran_index]->trid;
```

cristiarg (approved):

PR #2939, opened 2021-08-03
[LETS-167] implement synchronized reusable job container

cristiarg:
http://jira.cubrid.org/browse/LETS-167

Profiling showed that the overhead incurred by the allocator constantly constructing (and, probably, destructing) a large amount of objects (in particular `redo_job_impl` instances) is considerable.
Implement a synchronized container that allows obtaining, and then releasing back, pre-constructed instances of `redo_job_impl`. The main thread, instead of each time constructing a new instance, would ask the container for one instance. On the other end of applying the recovery redo jobs, the worker threads, instead of deleting such job instances, they would release the instances back to the container.

Implementation:
- new container `reusable_jobs_stack` is instantiated both for recovery redo (`log_recovery.c`) and for replication (`log_replication.cpp`)
- internally it has two containers of raw job pointers, one from which the main thread `pop`s objects, and another one into which all worker threads `push` back objects; containers are swapped when the `pop` one becomes empty; the `pop` is blocking - it will block indefinitely until a new job becomes available (returned back by the worker threads)
- recovery main thread pops objects from the reusable container (`log_rv_redo_record_sync_or_dispatch_async`) re-initializes them and dispatches them for processing
- replication worker thread, similarly, pops objects (`replicator::calculate_replication_delay_or_dispatch_async`) from the reusable container and dispatches them for processing
- after the worker threads process each job, they have to `retire` it which will, either, return the job to the reusable container, or delete it straight away (for special implementations of jobs that are not executed using the `redo_parallel` infrastructure - `redo_job_replication_delay_impl`, `redo_job_btree_stats`)

Tests:
- unit tests have been adapted
- recovery and replication have been tested manually


src/transaction/log_recovery.c
```diff
@@ -988,15 +988,17 @@ log_recovery_redo (THREAD_ENTRY * thread_p, log_recovery_context & context)
...
   {
     const int log_recovery_redo_parallel_count = prm_get_integer_value (PRM_ID_RECOVERY_PARALLEL_COUNT);
     assert (log_recovery_redo_parallel_count >= 0);
     if (log_recovery_redo_parallel_count > 0)
       {
-	parallel_recovery_redo.reset (
-	      new cublog::redo_parallel (log_recovery_redo_parallel_count, nullptr));
+        reusable_jobs.initialize(cublog::PARALLEL_RECOVERY_REDO_TUNING_REUSABLE_JOBS_STACK_SIZE,
+                                 &context.get_end_redo_lsa (), force_each_log_page_fetch);
+	parallel_recovery_redo.reset (new cublog::redo_parallel (log_recovery_redo_parallel_count, nullptr));
```
ribram, line 1002: Indentation looks off (may be just a Github glitch).

src/transaction/log_recovery_redo_parallel.hpp
```diff
@@ -38,6 +38,10 @@
 namespace cublog
 {
 #if defined(SERVER_MODE)
+  static constexpr size_t PARALLEL_RECOVERY_REDO_TUNING_REUSABLE_JOBS_STACK_SIZE = 100 * ONE_K;
```
ribram: Keep the name to the shortest meaningful version.
```suggestion
  static constexpr size_t PARALLEL_REDO_JOBS_STACK_SIZE = 100 * ONE_K;
```

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -752,16 +756,122 @@ namespace cublog
...
+		return pop_job;
+	      }
+	    else
+	      {
+		std::unique_lock<std::mutex> ulock { m_mutex };
+		m_jobs_available_on_push_stack_cv.wait (ulock, [this] ()
+		{
+		  return !m_push_stack.empty ();
+		});
+	      }
```
ribram: The code becomes shorter if conditional wait for push stack not being empty is checked from the beginning. If push stack is already not empty, the conditional wait will return immediately. If the push stack is empty, it waits until it is not empty and the stacks are not swapped unnecessarily.
```suggestion
	    std::unique_lock<std::mutex> ulock { m_mutex };
	    m_jobs_available_on_push_stack_cv.wait (ulock, [this] ()
	    {
		return !m_push_stack.empty ();
	    });
	    m_pop_stack.swap (m_push_stack);
	    assert (!m_pop_stack.empty ());
	    redo_parallel::redo_job_base *const pop_job = m_pop_stack.top ();
	    m_pop_stack.pop ();
	    return pop_job;
```

EDIT: The while loop also becomes unnecessary.

src/transaction/log_recovery_redo_parallel.hpp
```diff
@@ -97,6 +101,7 @@ namespace cublog
       std::condition_variable m_wait_for_target_value_cv;
   };
 
+
```
```suggestion
```

src/transaction/log_recovery_redo_parallel.hpp
```diff
@@ -408,16 +411,63 @@ namespace cublog
...
+      void initialize (std::size_t a_stack_size,
+		       std::function<redo_parallel::redo_job_base * ()> a_job_factory);
+
+      std::size_t size () const
+      {
+	return m_stack_size;
+      }
+
+      redo_parallel::redo_job_base *blocking_pop ();
+      void push (redo_parallel::redo_job_base *a_job);
```
ribram: Change the type of jobs in reusable_jobs_stack from base to impl and avoid the dynamic_cast.
```suggestion
      redo_parallel::redo_job_impl *blocking_pop ();
      void push (redo_parallel::redo_job_impl *a_job);
```

If this is a problem for the unit test, better to leave reusable_jobs_stack out of the current unit test.

src/transaction/log_recovery_redo_parallel.hpp
```diff
@@ -472,13 +523,20 @@ log_rv_redo_record_sync_or_dispatch_async (
...
     {
       // dispatch async
-      std::unique_ptr<cublog::redo_job_impl> job
-      {
-	new cublog::redo_job_impl (end_redo_lsa, force_each_log_page_fetch)
-      };
-      job->reset (rcv_vpid, rcv_lsa, log_rtype);
-      parallel_recovery_redo->add (std::move (job));
-      a_rcv_redo_perf_stat.time_and_increment (PERF_STAT_ID_REDO_OR_PUSH_DO_ASYNC);
+      while (true)
```
ribram: Remove while loop.

src/transaction/log_recovery_redo_perf.hpp
```diff
@@ -121,9 +124,8 @@ class log_recovery_redo_perf_stat
 	  perf_stat_ss << "Log Recovery Redo statistics:" << std::endl;
 	  for (std::size_t perf_stat_idx = 0; perf_stat_idx < m_definition.get_value_count (); ++perf_stat_idx)
 	    {
-	      perf_stat_ss.imbue (std::locale (""));
 	      perf_stat_ss << '\t' << m_definition.get_value_name (perf_stat_idx)
-			   << " μs: " << perf_stat_results[perf_stat_idx] << std::endl;
+			   << ": " << perf_stat_results[perf_stat_idx] << std::endl;
```
ribram, line 128: Wasn't it better when the time unit was added here?

src/transaction/log_replication.cpp
```diff
@@ -114,7 +116,12 @@ namespace cublog
 	m_minimum_log_lsa.reset (new cublog::minimum_log_lsa_monitor ());
 	// no need to reset with start redo lsa
 
-	m_parallel_replication_redo.reset (new cublog::redo_parallel (replication_parallel, m_minimum_log_lsa.get ()));
+	const bool force_each_log_page_fetch { true };
```
ribram: It is odd to see a primitive type value initialized with initializer list.
```suggestion
	const bool force_each_log_page_fetch = true;
```

src/transaction/log_recovery_redo_parallel.hpp
```diff
@@ -408,16 +411,63 @@ namespace cublog
...
+       */
+      std::stack<redo_parallel::redo_job_base *> m_pop_stack;
+      /* push happens, synchronized, on this stack
+       */
+      std::stack<redo_parallel::redo_job_base *> m_push_stack;
+      std::mutex m_mutex;
+      /* used to wait for jobs to be available on the push stack
+       * in case no jobs are found on the pop stack
+       */
+      std::condition_variable_any m_jobs_available_on_push_stack_cv;
```
ribram: Quote from [std::condition_variable_any](https://en.cppreference.com/w/cpp/thread/condition_variable_any):

> If the lock is std::unique_lock\<std::mutex\>, std::condition_variable may provide better performance.

```suggestion
      std::condition_variable m_jobs_available_on_push_stack_cv;
```

cristiarg: Autocompletion mistake I guess. Fixed.

src/transaction/log_recovery_redo_perf.hpp
cristiarg: `μs` was moved to label; because the entries are actually interspersed in the array: one is for the count, one for time

src/transaction/log_recovery_redo_parallel.cpp

src/transaction/log_recovery_redo_parallel.hpp

src/transaction/log_recovery_redo_parallel.hpp
cristiarg: Done. Silly mistake.

ribram (approved):

PR #2940, opened 2021-08-03
[LETS-170] Change CS access mode of log reader to safe reader.

ribram:
http://jira.cubrid.org/browse/LETS-170

Added configurable LOG_CS_ACCESS_MODE to the log_reader class and to LOG_READ functions.

Use LOG_CS_SAFE_READER access mode in all current log_reader usages.


cristiarg (approved):
The name of the function `fetch_page_force_use` is a bit misleading in the new context. Given it only has 2 uses, it could be eliminated and inlined.

ribram:
@cristiarg you're right about `fetch_page_force_use` being outdated and misleading. It was renamed as `fetch_page`.

PR #2941, opened 2021-08-03
[CBRD-24017] The problem that predicate push is not performed without changing the outer join to the inner join.

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-24017

info_spec.appears, info->nullable is not initialized properly. and I fixed it to handle right outer join effectively using mq_is_outer_join_spec().

### DIFF
--- a/src/optimizer/query_rewrite.c
+++ b/src/optimizer/query_rewrite.c
@@ -7351,6 +7351,17 @@ qo_optimize_queries (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int *co
 	    {
 	      return node;	/* give up */
 	    }
+
+	  /* predicate push */
+	  spec = node->info.query.q.select.from;
+	  while (spec)
+	    {
+	      if (spec->info.spec.derived_table_type == PT_IS_SUBQUERY)
+		{
+		  (void) mq_copypush_sargable_terms (parser, node, spec);
+		}
+	      spec = spec->next;
+	    }
 	}
 
       /* auto-parameterization is safe when it is done as the last step of rewrite optimization */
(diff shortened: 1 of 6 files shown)


shparkcubrid:
8a4c98a bfda1c9 are related to this PR.

shparkcubrid:
sql/_13_issues/_15_1h/cases/bug_bts_13571.sql - linux_sql_64bit_release
sql/_13_issues/_15_1h/cases/bug_bts_7547_1.sql - linux_sql_64bit_release
This is a case that an outer join is changed to an inner join.
In my opinion, it is right to change. Please check it.

src/optimizer/query_rewrite.c
```diff
@@ -5495,27 +5495,21 @@ qo_rewrite_outerjoin (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int *c
       prev_spec = NULL;
       for (spec = node->info.query.q.select.from; spec; prev_spec = spec, spec = spec->next)
 	{
-	  if (spec->info.spec.join_type == PT_JOIN_LEFT_OUTER
-	      || (spec->info.spec.join_type == PT_JOIN_RIGHT_OUTER && prev_spec))
+	  /* check outer join spec. */
```
beyondykk9, line 5497: Currently, prev_spec is necessary ? it looks not useless and the prev_spec was a condition to check OUTER JOIN.
it's no problem that the condtion is changed to check OUTER JOIN without prev_spec ?

shparkcubrid: I removed prev_spec.
Using mq_is_outer_join_spec(), right outer join is checked more effectively.

beyondykk9: OK.

beyondykk9 (approved):

ctshim (approved):

hgryoo (approved):

youngjinj (approved):

PR #2937, opened 2021-08-03
[CBRD-23917] Server side log processing - DML, DCL, Timer

joohok:
http://jira.cubrid.org/browse/CBRD-23917

This PR is for, 
1. Traversing Log records 
2. Pick interpretable log record
3. Get information which is required to make kinds of log item (heap record for DML, time info for TIMER, ..)
4. Make log info (prior stage of log item, this log info is used to generate log item at client side)

generating DDL log info will be added soon. 
and queue mechanism and thread handling will be covered in next PR.

PR #2943, opened 2021-08-03
[CBRD-23917] Daemon Thread Initialization

joohok:
http://jira.cubrid.org/browse/CBRD-23917

This PR is for initialization of CDC daemons

cdc log producer : Traverses log records and make log item (implemented in #2937) 
cdc thread checker : Check conditions to wakeup cdc log producer and consumer 

Managing producer, checker, consumer threads and queue handling are not covered in this PR.

### DIFF
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -10214,6 +10214,36 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
   return error_code;
 }
 
+static cubthread::daemon *cdc_Producer_daemon = NULL;
+
+void cdc_producer_daemon_init()
+{
+  assert (cdc_Producer_daemon == NULL);
+  
+  cubthread::looper looper = cubthread::looper (cdc_get_interval);
+  cubthread::entry_callable_task *daemon_task = new cubthread::entry_callable_task (cdc_log_producer_execute);
+
+  cdc_Producer_daemon = cubthread::get_manager ()->create_daemon (looper, daemon_task, "cdc_log_producer"); 
+}
+
+#if defined (SERVER_MODE)
+static void
+cdc_log_producer_execute (cubthread::entry & thread_ref)
+{
+  if (!BO_IS_SERVER_RESTARTED ())
+  {
+    return;
+  }
+  
+  cdc_log_producer (&thread_ref);
+}
+#endif
+
+void cdc_log_producer (THREAD_ENTRY *thread_p)
+{
+
+}
+
 //
 // log critical section
 //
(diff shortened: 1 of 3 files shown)


src/transaction/log_manager.h
```diff
@@ -216,6 +216,11 @@ extern void log_flush_daemon_get_stats (UINT64 * statsp);
 
 extern void log_update_global_btid_online_index_stats (THREAD_ENTRY * thread_p);
 
+#if defined (SERVER_MODE)
```
hornetmj, line 219: makes a header file for the cdc and use it.

src/transaction/log_manager.c
```diff
@@ -321,6 +321,9 @@ static cubthread::daemon *log_Check_ha_delay_info_daemon = NULL;
 
 static cubthread::daemon *log_Flush_daemon = NULL;
 static std::atomic_bool log_Flush_has_been_requested = {false};
+
+static cubthread::daemon *cdc_log_Producer_daemon = NULL;
+static cubthread::daemon *cdc_Thread_checker_daemon = NULL;
```
wrlawodms, line 326: Could you tell me the purpose of this daemon? I wasn't told it in a design review?

src/transaction/log_manager.c
```diff
@@ -10214,6 +10217,129 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+/* *INDENT-OFF* */
+static void
+cdc_log_producer_execute (cubthread::entry & thread_ref)
+{
+  if (!BO_IS_SERVER_RESTARTED ())
+  {
+    return;
+  }
+  
+  cdc_log_producer (&thread_ref);
```
wrlawodms, line 10245: Could you tell me why this internal function is needed?

wrlawodms (commented):
I'd like to know which daemon you referred to before further review.

src/transaction/log_manager.h
hornetmj: The use of 'cdc_' prefix needs to discuss (in detail design level). And if you have any plan to use this mechanism at the flashback utility or etc, then I think the 'cdc_' prefix is not suitable for that purpose.

src/transaction/log_manager.c
```diff
@@ -10214,6 +10217,129 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+/* *INDENT-ON* */
+
+/* *INDENT-OFF* */
+void
+cdc_log_producer_interval (bool & is_timed_wait, cubthread::delta_time & period)
+{
+
+  int cdc_log_producer_interval_sec = 0; /* for future fix to decrease server load */
+
+  if (cdc_log_producer_interval_sec > 0)
```
hornetmj, line 10267: This is an unused code.

The 'log_producer' thread wakes up periodically and can check the condition of 'next_lsa' <= 'nxio_lsa'. Is it right?
If it is, then I think you can remove the 'thread_checker' thread.

PR #2947, opened 2021-08-05
[CBRD-23994] fix coredump when calling vacuum_get_log_blockid() if vacuum_disable is set to true.

hornetmj:
http://jira.cubrid.org/browse/CBRD-23994

**Purpose**
- fix coredump
- if the hidden system parameter 'vacuum_disable' is set to true, then the coredump is occurred whenever calling the vacuum_get_log_blockid() function. Because the vacuum_Data.log_block_npages is always set to 0 and it make a 'divide by 0' problem'

**Implementation**
N/A

**Remarks**
N/A

### DIFF
--- a/src/query/vacuum.c
+++ b/src/query/vacuum.c
@@ -5611,7 +5611,14 @@ vacuum_recover_lost_block_data (THREAD_ENTRY * thread_p)
 VACUUM_LOG_BLOCKID
 vacuum_get_log_blockid (LOG_PAGEID pageid)
 {
-  return ((pageid == NULL_PAGEID) ? VACUUM_NULL_LOG_BLOCKID : (pageid / vacuum_Data.log_block_npages));
+  if (prm_get_bool_value (PRM_ID_DISABLE_VACUUM) || pageid == NULL_PAGEID)
+    {
+      return VACUUM_NULL_LOG_BLOCKID;
+    }
+
+  assert (vacuum_Data.log_block_npages != 0);
+
+  return pageid / vacuum_Data.log_block_npages;
 }
 
 /*


wrlawodms:
This PR looks good enough to prevent the coredump.

By the way, It seems that we need more changes to make `vacuum_disable` work properly. With this PR, `vacuum_get_log_blockid()` will always return NULL, which in turn makes `prior_lsa_next_record_internal()` not to produce a vacuum data block. It means the log records that are supposed to belong to the data block won't be vacuumed later when the vacuum becomes enabled. This is based on my rough code analysis so that it could be wrong, but I want you to consider it.

hornetmj:
> This PR looks good enough to prevent the coredump.
> 
> By the way, It seems that we need more changes to make `vacuum_disable` work properly. With this PR, `vacuum_get_log_blockid()` will always return NULL, which in turn makes `prior_lsa_next_record_internal()` not to produce a vacuum data block. It means the log records that are supposed to belong to the data block won't be vacuumed later when the vacuum becomes enabled. This is based on my rough code analysis so that it could be wrong, but I want you to consider it.

Thanks for your opinion. It is a very good point and I agree with you. But, vacuum_produce_log_block_data() function also just return if vacuum_disable is set. And this system parameter has the HIDDEN property. So, I currently focus on fixing the coredump.

If we want to remove the HIDDEN property of it, then we need to consider more in (I hope) another issue. I think I(we) have no chance to analysis all related code(log, mvcc, vacuum) until 11.2 release. So, I hope to process one by one next time.

PS. If you are serious about improving the overall vacuum_disable behavior, please open a related issue.

src/query/vacuum.c
```diff
@@ -5611,7 +5611,14 @@ vacuum_recover_lost_block_data (THREAD_ENTRY * thread_p)
 VACUUM_LOG_BLOCKID
 vacuum_get_log_blockid (LOG_PAGEID pageid)
 {
-  return ((pageid == NULL_PAGEID) ? VACUUM_NULL_LOG_BLOCKID : (pageid / vacuum_Data.log_block_npages));
+  if (prm_get_bool_value (PRM_ID_DISABLE_VACUUM) || pageid == NULL_PAGEID)
```
wrlawodms, line 5614: How about using `vacuum_Is_booted` instead of checking the system parameter? because this function is called very frequently, especially for every DML.

wrlawodms: or just checking `vacuum_Data.log_block_npages == NULL`

wrlawodms (approved):
@hornetmj Thank you for considering. One more suggestion.

src/query/vacuum.c
hornetmj: @wrlawodms I checked several API function of vacuum and I saw that they all use 'prm_get_bool_value (PRM_ID_DISABLE_VACUUM)' instead of 'vacuum_Is_booted' or 'vacuum_Data.log_block_npages == NULL'. So I choose it as I thought it fit for purpose. However, your suggestion is good for me. So I want to check other reviewer's opinions. What do you(@JoohoK and @eido5) think about it?

wrlawodms: @hornetmj What you're saying also makes sense. I think we can leave it as you do and see the update performance test from QA system. We can improve it later if needed. Thank you.

hornetmj: @wrlawodms OK. I will check the performance test from QA. Thanks to leave your opinion.

eido5 (approved):

hornetmj:
@mhoh3963 We finished the review. Please merge this PR.

mhoh3963:
@hornetmj I merged it.

PR #2950, opened 2021-08-05
[CBRD-23763] fix coredump that occurs when executing the copydb utility with -d option

hornetmj:
http://jira.cubrid.org/browse/CBRD-23763

### Purpose
- fix coredump that occurs when executing the copydb utility with -d option because it uses a deallocated global variable after calling tp_final function.
- tp_domain_free(...) -> area_free (tp_Domain_area, dom) -> assert (tp_Domain_area != NULL)
  - tp_Domain_area is null value because tp_final() has already called by boot_server_all_finalize() with BOOT_SHUTDOWN_ALL_MODULES shutdown mode.
  - So, I change the shutdown mode from BOOT_SHUTDOWN_ALL_MODULES to BOOT_SHUTDOWN_EXCEPT_COMMON_MODULES
### Implementation
N/A
### Remarks
- -d (--delete-source) option deletes the source database.
  - [copydb manual](https://www.cubrid.org/manual/en/11.0/admin/admin_utils.html?highlight=delete%20source#cmdoption-copydb-d)
- copydb -d operates like copydb + deletedb

### DIFF
--- a/src/communication/network_interface_cl.c
+++ b/src/communication/network_interface_cl.c
@@ -4197,7 +4197,7 @@ boot_delete (const char *db_name, bool force_delete)
 
   enter_server_no_thread_entry ();
 
-  error_code = xboot_delete (db_name, force_delete, BOOT_SHUTDOWN_ALL_MODULES);
+  error_code = xboot_delete (db_name, force_delete, BOOT_SHUTDOWN_EXCEPT_COMMON_MODULES);
 
   exit_server_no_thread_entry ();
 


wrlawodms:
With this PR, the `cubrid deletedb` utility can't finalize modules excluded. Is it OK?

hornetmj:
> With this PR, the `cubrid deletedb` utility can't finalize modules excluded. Is it OK?

I think we don't worry about that.
The boot_restart_server() or boot_restart_client() function initializes modules you said and boot_server_all_finalize() or boot_client_all_finalize() finalizes them. But, when executing deletedb utility, they are never called(both pair of them). And when executing copydb with -d option, boot_client_all_finalize() is called from boot_shutdown_client_at_exit() -> boot_shutdown_client().

wrlawodms (approved):
Ok, thank you. It looks good to me.

kisoo-han (approved):

hornetmj:
@mhoh3963 We finished the review and we want to merge this PR. But the circleci complains of some testcase failure that may not be related to this PR. Could you please check it and merge it?

mhoh3963:
@hornetmj i merged it.

PR #2952, opened 2021-08-06
[LETS-179] Parallel recovery performance: reusable job container: use vector with reserve instead of stack

cristiarg:
http://jira.cubrid.org/browse/LETS-179

Better synchronization is needed in `reusable_jobs_stack` in order to minimize contention both between worker threads and also between main thread and worker threads and allow main thread to work as much as possible in a lock-free way.

Perf revealed that `std::stack` spends quite some time [re]allocating internal memory.

Implementation:
- replace two stage transfer from worker threads towards main thread by a three stage transfer
- in the first stage (starting from the worker threads' side) each worker thread pushes one job at a time in a lock-free way in its own internal container
- from time to time, configurable by a variable, each worker thread pushes - synchronized - already collected jobs in an intermediary container
- main thread consumes jobs from its own container in a lock-free way, from time to time, when this container becomes empty, it will swap - synchronized - its empty container for the intermediary container
- replace `std::stack` with pre-reserved `std::vectors` as internal containers
- all reusable jobs are constructed in-place in a contiguous memory array for cache optimization

Other changes:
- remove redundant perf stats recording
- `log_reader`: obtain thread entry only once from the thread local variable

### DIFF
--- a/src/transaction/log_recovery_redo_parallel.cpp
+++ b/src/transaction/log_recovery_redo_parallel.cpp
@@ -833,7 +833,7 @@ namespace cublog
 
   reusable_jobs_stack::~reusable_jobs_stack ()
   {
-    // consistency check
+    // consistency check that all job instances have been 'retuned to the source'
     assert ([this] ()
     {
       const std::size_t pop_size = m_pop_jobs.size ();
@@ -847,6 +847,23 @@ namespace cublog
       return true;
     }
     ());
+
+    // formally invoke dtor  on all in-place constructed objects
+    for (auto &job : m_pop_jobs)
+      {
+        job->~redo_job_impl();
+      }
+    for (auto &job : m_push_jobs)
+      {
+        job->~redo_job_impl();
+      }
+    for (auto &push_job_container: m_per_task_push_jobs_vec)
+      {
+        for (auto &job : push_job_container)
+          {
+            job->~redo_job_impl();
+          }
+      }
 
     free (m_jobs_arr);
     m_jobs_arr = nullptr;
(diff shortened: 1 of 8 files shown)


src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -779,75 +781,119 @@ namespace cublog
...
+
+    assert (a_job_count > 0);
+    assert (a_push_task_count > 0);
+    assert (a_flush_push_at_count > 0);
+
+    m_job_count = a_job_count;
+    m_push_task_count = a_push_task_count;
+    m_flush_push_at_count = a_flush_push_at_count;
+
+    m_jobs_arr = static_cast<unsigned char *> (malloc (sizeof (redo_job_impl) * m_job_count));
```
ribram, line 814:
```suggestion
    m_jobs_arr = new redo_job_impl[m_job_count];
```

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -779,75 +781,119 @@ namespace cublog
...
-	m_push_stack.pop ();
-	delete push_job;
-      }
-    while (!m_pop_stack.empty ())
-      {
-	redo_job_impl *const pop_job = m_pop_stack.top ();
-	m_pop_stack.pop ();
-	delete pop_job;
-      }
+    free (m_jobs_arr);
```
ribram, line 868: delete [] m_jobs_arr;

src/transaction/log_recovery_redo_parallel.cpp
cristiarg: `redo_job_impl` does not have a default ctor; via its ctor a part of the params are initialized; and then, at each reuse, another part of the params are initialized via `reinitialize`; that is the reason of the malloc + placement new + free

PR #2953, opened 2021-08-09
[CBRD-23987] Move slots at once when shifting slots up or down.

wrlawodms:
http://jira.cubrid.org/browse/CBRD-23987
 
Move [start_slot_p, end_slot_p) to the `end_slot_p` direction.
Slots are arranged physically in revese. e.g. |end_slot|..|..|start_slot|

### DIFF
--- a/src/storage/slotted_page.c
+++ b/src/storage/slotted_page.c
@@ -185,7 +185,7 @@ static int spage_find_empty_slot (THREAD_ENTRY * thread_p, PAGE_PTR pgptr, int l
 				  int *space, PGSLOTID * slotid);
 static int spage_find_empty_slot_at (THREAD_ENTRY * thread_p, PAGE_PTR pgptr, PGSLOTID slotid, int length, INT16 type,
 				     SPAGE_SLOT ** sptr);
-static void spage_move_slots (SPAGE_SLOT * start_slot_p, int cnt, int move_offset);
+static void spage_move_slots (SPAGE_SLOT * start_slot_p, SPAGE_SLOT * end_slot_p, int offset);
 static void spage_shift_slot_up (PAGE_PTR page_p, SPAGE_HEADER * page_header_p, SPAGE_SLOT * slot_p);
 static void spage_shift_slot_down (PAGE_PTR page_p, SPAGE_HEADER * page_header_p, SPAGE_SLOT * slot_p);
 static int spage_add_new_slot (THREAD_ENTRY * thread_p, PAGE_PTR page_p, SPAGE_HEADER * page_header_p,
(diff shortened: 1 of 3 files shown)


src/storage/slotted_page.c
```diff
@@ -1489,6 +1490,32 @@ spage_find_empty_slot (THREAD_ENTRY * thread_p, PAGE_PTR page_p, int record_leng
...
+ * spage_move_slots() -
+ *   return:
+ *
+ *   start_slot_p(in): Pointer to slotted page pointer array
+ *   end_slot_p(in): Pointer to slotted page pointer array
+ *   offset(in): How much to move. Positive means go up in a slot page, and negative down.
+ *
+ *   Move [start_slot_p, end_slot_p) to the end_slot_p direction.
+ */
+static void
```
hornetmj: Do we have a chance to move several offsets?
In aspects of code generalization, it is very good. But, how about just wrapping memmove() by spage_move_slots() with an inline keyword? Or how about just use memmove() right away?

wrlawodms: I thought It would be messy and unreadable due to the address arithmetic, but I've figured out it's not that complicated and even could be more readable. I accept your review and thank you.

src/storage/slotted_page.c
```diff
@@ -1550,13 +1545,9 @@ spage_shift_slot_down (PAGE_PTR page_p, SPAGE_HEADER * page_header_p, SPAGE_SLOT
...
     }
-  else
+  else if (page_header_p->num_slots > 1)
     {
-      for (; slot_p > last_slot_p; slot_p--)
-	{
-	  spage_set_slot (slot_p, (slot_p - 1)->offset_to_record, (slot_p - 1)->record_length,
-			  (slot_p - 1)->record_type);
-	}
+      memmove (last_slot_p + 1, last_slot_p, sizeof (SPAGE_SLOT) * (slot_p - last_slot_p));
```
hornetmj:
```suggestion
      if (last_slot_p < slot_p)
        {
          memmove (last_slot_p + 1, last_slot_p, sizeof (SPAGE_SLOT) * (slot_p - last_slot_p));
        }
```

src/storage/slotted_page.c
```diff
@@ -1550,13 +1545,9 @@ spage_shift_slot_down (PAGE_PTR page_p, SPAGE_HEADER * page_header_p, SPAGE_SLOT
     {
       spage_set_slot (slot_p, last_slot_p->offset_to_record, last_slot_p->record_length, last_slot_p->record_type);
     }
-  else
+  else if (page_header_p->num_slots > 1)
```
hornetmj:
```suggestion
  else
```

src/storage/slotted_page.c
```diff
@@ -1514,12 +1514,7 @@ spage_shift_slot_up (PAGE_PTR page_p, SPAGE_HEADER * page_header_p, SPAGE_SLOT *
     }
   else
     {
-      for (; last_slot_p < slot_p; last_slot_p++)
-	{
-	  spage_set_slot (last_slot_p, (last_slot_p + 1)->offset_to_record, (last_slot_p + 1)->record_length,
-			  (last_slot_p + 1)->record_type);
-	}
-      assert (last_slot_p == slot_p);
+      memmove (last_slot_p, last_slot_p + 1, sizeof (SPAGE_SLOT) * (slot_p - last_slot_p));
```
hornetmj:
```suggestion
      if (last_slot_p < slot_p)
        {
          memmove (last_slot_p, last_slot_p + 1, sizeof (SPAGE_SLOT) * (slot_p - last_slot_p));
        }
```

wrlawodms: Do you think it needs to do although It will be always true?

hornetmj: If the caller passes the invalid pointer(slot_p), then I think we have no chance to catch it in the scope of function. So I suggested it. But, if you think it is always true, then how about use assert? (I think you need to consider last_slot_p == slot_p case when you make a condition of if or assert)

wrlawodms: Good point, but if it gets the slot_p from spag_* interfaces, it should be larger than last_slot_p from `spage_find_slot (page_p, page_header_p, page_header_p->num_slots, false);` because slots are within [0, page_header_p->num_slots). However, I'll accept your reviews so as to just keep the original logic for safety. I guess that's what you're worried about.

hornetmj (approved):

eido5 (approved):

PR #2954, opened 2021-08-09
[CBRD-23954] Use BIGINT to get the result of SELECT COUNT

wrlawodms:
http://jira.cubrid.org/browse/CBRD-23954

[regression] Change the data type for the result of select count which has been changed in #2876.

### DIFF
--- a/src/object/schema_class_truncator.cpp
+++ b/src/object/schema_class_truncator.cpp
@@ -595,7 +595,7 @@ namespace cubschema
 	    return error;
 	  }
 
-	cnt_refers = db_get_int (&value);
+	cnt_refers = db_get_bigint (&value);
 
 	db_query_end (result);
 	db_close_session (session);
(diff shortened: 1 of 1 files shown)


joohok (approved):

eido5 (approved):

src/object/schema_class_truncator.cpp
```diff
@@ -595,7 +595,7 @@ namespace cubschema
 	    return error;
 	  }
 
-	cnt_refers = db_get_int (&value);
+	cnt_refers = db_get_bigint (&value);
```
hornetmj, line 598: Please check the type of return value.

hornetmj: After consideration of it, you can merge this PR whether or not confirm by me.

wrlawodms: Thank you, I've changed it from `int` to `DB_BIGINT`

hornetmj (approved):

PR #2955, opened 2021-08-09
[CBRD-24021] Put libncurses6 symbolic link to CUBRID/lib as libncurses5

kisoo-han:
http://jira.cubrid.org/browse/CBRD-24021

Purpose

* In CUBRID Distribution packages (sh, rpm) for Redhat Linux (including 8.4)
  * Add symbolic link of ncurses6 as ncurses5 in $CUBRID/lib
  * Same works as CBRD-23587

Implementation
N/A

Remarks

### DIFF
--- a/cmake/CPack.STGZ_Header.sh.in
+++ b/cmake/CPack.STGZ_Header.sh.in
@@ -197,14 +197,7 @@ elif [ -f /etc/os-release ];then
 fi
 
 case $OS in
-	fedoraproject)
-		if [ ! -h /lib64/libncurses.so.5 ] && [ ! -h $LIB/libncurses.so.5 ];then
-			ln -s /lib64/libncurses.so.6 $LIB/libncurses.so.5
-			ln -s /lib64/libform.so.6 $LIB/libform.so.5
-			ln -s /lib64/libtinfo.so.6 $LIB/libtinfo.so.5
-		fi
-		;;
-	centos)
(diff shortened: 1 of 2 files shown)


mhoh3963:
How about merge ubuntu routine and debian routine into "ubuntu | debian)" ?
And remark in the commit seems to be different from this issue, so please remove or fix it.

mhoh3963 (approved):

kisoo-han:
Unfortunately, **ubuntu** and **debian** cannot be grouped in the same case because the path of **libform.so.6** is different.
I will check the library path again

kisoo-han:
In '**ubuntu**', /lib is a symbolic link to 'usr/lib' but in '**debian**' '/lib' is a directory.
So, /lib/x86_64-linux-gnu/ and /usr/lib/x86_64-linux-gnu/ are same directory in '**ubuntu**, and all 3 ncurses libraries are located in it.
But they are different directories in '**debian**', and two libraries, libncurses.so.6 and libtinfo.so.6, are located in /lib/x86_64-linux-gnu/ and libform.so.6 is located in /usr/lib/x86_64-linux-gnu/.
Due to this reason '**ubuntu**' and '**debian**' cannot be grouped into one.

mhoh3963:
@kisoo-han
OK, I get it.

airnet73 (approved):

PR #2944, opened 2021-08-04
[CBRD-23917] Server-side client request processing

joohok:
http://jira.cubrid.org/browse/CBRD-23917

This PR is for handling client request

1. set configuration
2. get LSA by time 
3. get log item info 
4. get log item list
5. finalize 

In order to handle request for  log item (3, 4), consumer (cdc_get_logitem_info()) is required to be implemented. However, it is more related to thread handling, so cdc_get_logitem_info() will be covered in next PR about thread and queue management.

PR #2958, opened 2021-08-10
[CBRD-23920] Implement APIs for CDC

joohok:
http://jira.cubrid.org/browse/CBRD-23920

This PR is for CDC API

- memory optimization is still in progress
- error codes when css_* function failed are not defined
- protocol names will be modified after (#2944)

### DIFF
--- a/src/api/cubrid_log.c
+++ b/src/api/cubrid_log.c
@@ -1,0 +1,1524 @@
+/*
+ * Copyright 2008 Search Solution Corporation
+ * Copyright 2016 CUBRID Corporation
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+
+/*
+ * cubrid_log.c -
+ */
+
+#ident "$Id$"
+
+#if defined(CS_MODE)
+
+#include <arpa/inet.h>
+#include <assert.h>
+#include <limits.h>
+
+#include "connection_cl.h"
+#include "connection_list_cl.h"
+#include "cubrid_log.h"
+#include "log_lsa.hpp"
+#include "network.h"
+#include "object_representation.h"
+
+#define CUBRID_LOG_ERROR_HANDLING(e, v) \
+  (err_code) = (e); \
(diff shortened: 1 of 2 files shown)


src/api/cubrid_log.c
```diff
@@ -0,0 +1,1525 @@
...
+ *
+ */
+
+/*
+ * cubrid_log.c -
+ */
+
+#ident "$Id$"
+
+#if defined(CS_MODE)
```
wrlawodms, line 25: If all the code works only in CS_MODE, is this necessary? We could just not include cubrid_log.c and cubrid_log.h in CMAKEFILE for SA mode.

src/api/cubrid_log.c
```diff
@@ -0,0 +1,1525 @@
...
+  CUBRID_LOG_STAGE_FINALIZATION
+} CUBRID_LOG_STAGE;
+
+typedef enum
+{
+  CREATE = 0,
+  ALTER,
+  DROP,
+  RENAME,
+  TRUNCATE
```
wrlawodms, line 59: It may be going to DML?

src/api/cubrid_log.c
```diff
@@ -0,0 +1,1525 @@
...
+
+typedef enum
+{
+  DATA_ITEM_TYPE_DDL = 0,
+  DATA_ITEM_TYPE_DML,
+  DATA_ITEM_TYPE_DCL,
+  DATA_ITEM_TYPE_TIMER
+} DATA_ITEM_TYPE;
+
+CUBRID_LOG_STAGE g_stage = CUBRID_LOG_STAGE_CONFIGURATION;
```
wrlawodms, line 94: it seems to need the global variable convention and more meaningful name.

src/api/cubrid_log.c
```diff
@@ -0,0 +1,1525 @@
...
+int g_trace_log_level = 0;
+int64_t g_trace_log_filesize = 10 * 1024 * 1024;	/* 10 MB */
+
+LOG_LSA g_next_lsa = LSA_INITIALIZER;
+
+char *g_log_infos = NULL;
+int g_log_infos_size = 0;
+
+CUBRID_LOG_ITEM *g_log_items = NULL;
+int g_log_items_count = 0;
```
wrlawodms, line 120: They could be combined into one structure.

src/api/cubrid_log.c
```diff
@@ -0,0 +1,1525 @@
...
+}
+
+/*
+ * cubrid_log_set_extraction_table () -
+ *   return:
+ *   classoid_arr(in):
+ *   arr_size(in):
+ */
+int
+cubrid_log_set_extraction_table (uint64_t * classoid_arr, int arr_size)
```
wrlawodms, line 290: I thought there is a kind of table for extraction. It seems to need a clearer name.

wrlawodms (commented):
I skipped my review for naming and file structure. Most of the codes look good. Plus, I don't know well about the network interfaces, but I think it could be checked in test cases easily.

wrlawodms:
It would be good if you add comments for each variable in structures.

PR #2959, opened 2021-08-10
[CBRD-23993] merge with predicated push down

beyondykk9:
http://jira.cubrid.org/browse/CBRD-23993

The predicated push down have to include nullable terms. So, it needs to merge with recent develop branch's CBRD-24014 issue.


beyondykk9:
It has a fault for commit. I will re-requet later.

PR #2960, opened 2021-08-10
[LETS-180] Recovery redo performance: rework perf stats recording in log recovery redo applying threads

cristiarg:
http://jira.cubrid.org/browse/LETS-180

Performance statistics recording on the worker threads - `perfmon_counter_timer_tracker` and `perfmon_counter_timer_raii_tracker` - is based on the legacy performance statistics infrastructure in `perf_monitor`.
Because the overhead is non-negligible, rework the implementation in terms of performance statistics implementation in `perf.hpp`.

Implementation:
- rename log_recovery_redo_perf_stat into cublog::perf_stats and modified to be parameterized by a cubperf::statset_definition instance
- this is instantiated with two statset definitions based on and existing anonymous enum for main thread; and a new anonymous enum for each of the async recovery redo applying threads
- added some utility functions for perf stats activation at runtime based on perfmon activation flags
- added new perfmon activation flag - PERFMON_ACTIVATION_FLAG_LOG_RECOVERY_REDO_ASYNC
- change the way how instances of redo_parallel::redo_task are handled:
  - previously: ownewship of task instances went to the worker pool which self destructed them after execution via cubthread::task::retire function
  - currently: ownership remains with `redo_recovery` (`retire` function is overridden to `nop`)
- task execution is divided into a few sections which are measured: pop job, idle/sleep, execution proper, job retire
- perf stats results are logged both on a per-task basis and also averaged across all tasks

Manual tests:
- recovery tests
- TS/PS tests


src/base/perf.hpp
```diff
@@ -140,14 +139,23 @@ namespace cubperf
   //
   // see how to use in file description comment
   //
-  class statset_definition
+  class statset_definition final
   {
     public:
-      // no default constructor
+      using stat_definition_init_list_t = std::initializer_list<stat_definition>;
```
ribram: If it is part of statset_definition, repeating the same suffix is not necessary.
```suggestion
      using init_list_t = std::initializer_list<stat_definition>;
```

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -673,10 +704,41 @@ namespace cublog
...
+    for (auto &redo_task: m_redo_tasks)
+      {
+	redo_task->accumulate_perf_stats (accum_perf_stat_results.data (), accum_perf_stat_results.size ());
+	redo_task->log_perf_stats ();
       }
+
+    // average
+    const std::size_t task_count = m_redo_tasks.size ();
+    const std::size_t value_count = accum_perf_stat_results.size ();
+    std::vector<double> avg_perf_stat_results;
```
ribram: Having double is not really necessary, cubperf::stat_value should be good enough. As a result `log_perf_stats_values_with_definition()` may become a simple function, without template.
```suggestion
    std::vector<cubperf::stat_value> avg_perf_stat_results;
```

src/transaction/log_recovery_redo_perf.hpp
```diff
@@ -25,113 +25,205 @@
...
+      {
+	m_definition.time_and_increment (*m_stats_set, a_stat_id);
+      }
+  }
+
+  void perf_stats::log (const char *a_title) const
+  {
+    if (m_stats_set != nullptr)
+      {
+	std::vector < cubperf::stat_value > perf_stat_values;
```
ribram:
```suggestion
	std::vector<cubperf::stat_value> perf_stat_values;
```

PR #2961, opened 2021-08-11
[CBRD-23842] Support CDC interface 

joohok:

This PR is for CDC interface.


joohok:
Sorry for commit 'apply review for supplement-dml(#2912)' that contains synchronize code. 
in order to test the code, I inserted some code.

Please review only two files 
- src/storage/heap_file.c
- src/storage/heap_file.h

I applied the reviews in https://github.com/CUBRID/cubrid/pull/2912 

thanks

src/storage/heap_file.c
```diff
@@ -881,6 +881,8 @@ STATIC_INLINE int heap_copy_chain (THREAD_ENTRY * thread_p, PAGE_PTR page_heap,
 STATIC_INLINE int heap_get_last_vpid (THREAD_ENTRY * thread_p, const HFID * hfid, VPID * last_vpid)
   __attribute__ ((ALWAYS_INLINE));
 
+STATIC_INLINE bool check_supplemental_log (THREAD_ENTRY * thread_p, OID * classoid) __attribute__ ((ALWAYS_INLINE));
```
wrlawodms, line 884: seems to need a prefix.

src/storage/heap_file.c
```diff
@@ -22473,6 +22705,25 @@ heap_insert_logical (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * context,
...
+					tdes->client.get_db_user ());
+	  tdes->has_supplemental_log = true;
+	}
+
+      context->do_supplemental_log = true;
+      LSA_SET_NULL (&context->supp_redo_lsa);
+    }
+  else
+    {
+      context->do_supplemental_log = false;
```
wrlawodms, line 22712: You should initialize this value in heap_clear_operation_context(), and in turn, you don't have to set false here.

wrlawodms: It could be applied to other heap operations too.

src/storage/heap_file.c
```diff
@@ -22708,6 +22976,24 @@ heap_delete_logical (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * context)
...
+	}
+
+      context->do_supplemental_log = true;
+      LSA_SET_NULL (&context->supp_undo_lsa);
+    }
+  else
+    {
+      context->do_supplemental_log = false;
+    }
+
```
wrlawodms: How about creating a function for this repeated work?

src/storage/heap_file.c
```diff
@@ -22529,6 +22786,12 @@ heap_insert_logical (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * context,
     {
       heap_log_insert_physical (thread_p, context->home_page_watcher_p->pgptr, &context->hfid.vfid, &context->res_oid,
 				context->recdes_p, is_mvcc_op, context->is_redistribute_insert_with_delid);
+
+      /* redo lsa for SUPPLEMENT_INSERT log */
+      if (context->do_supplemental_log && !heap_is_big_length (context->recdes_p->length))
+	{
+	  LSA_COPY (&context->supp_redo_lsa, &tdes->tail_lsa);
+	}
```
wrlawodms: How about wrapping these into an inline function or macro?

src/storage/heap_file.c
```diff
@@ -22529,6 +22786,12 @@ heap_insert_logical (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * context,
     {
       heap_log_insert_physical (thread_p, context->home_page_watcher_p->pgptr, &context->hfid.vfid, &context->res_oid,
 				context->recdes_p, is_mvcc_op, context->is_redistribute_insert_with_delid);
+
+      /* redo lsa for SUPPLEMENT_INSERT log */
+      if (context->do_supplemental_log && !heap_is_big_length (context->recdes_p->length))
```
wrlawodms: I don't think `heap_is_big_length()` works as you expect. See the `heap_insert_handle_multipage_record()` in which `context->recdes_p->length` became sizeof OID.

src/storage/heap_file.c
```diff
@@ -22486,6 +22737,12 @@ heap_insert_logical (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * context,
       goto error;
     }
 
+/*SUPPLEMENT_INSERT REDO LSA, INSERT REC_BIGONE  */
+  if (context->do_supplemental_log && heap_is_big_length (context->recdes_p->length))
+    {
+      LSA_COPY (&context->supp_redo_lsa, &tdes->tail_lsa);
+    }
+
```
wrlawodms: How about putting it into `heap_insert_handle_multipage_record()`?

src/storage/heap_file.c
```diff
@@ -21410,6 +21503,7 @@ heap_delete_home (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * context, boo
 	  rec_address.pgptr = context->home_page_watcher_p->pgptr;
 	  rec_address.vfid = &context->hfid.vfid;
 	  rec_address.offset = context->oid.slotid;
+
```
wrlawodms, line 21502:
```suggestion
```

src/storage/heap_file.c
```diff
@@ -22486,6 +22737,12 @@ heap_insert_logical (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * context,
       goto error;
     }
 
+/*SUPPLEMENT_INSERT REDO LSA, INSERT REC_BIGONE  */
+  if (context->do_supplemental_log && heap_is_big_length (context->recdes_p->length))
+    {
+      LSA_COPY (&context->supp_redo_lsa, &tdes->tail_lsa);
```
wrlawodms: I recommend you using `logtb_find_current_tran_lsa()` instead of directly accessing `tdes->tail_lsa`. It also make you to be able to get rid of `tdes = LOG_FIND_CURRENT_TDES (thread_p);`s. This could be applied to all heap operations.

src/storage/heap_file.c
```diff
@@ -21438,9 +21540,19 @@ heap_delete_home (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * context, boo
...
+	   * So, current log lsa after heap_log_delete_physical can points unexpected log, other than delete log. */
+	  heap_log_delete_physical (thread_p, context->home_page_watcher_p->pgptr, &context->hfid.vfid, &context->oid,
+				    &context->home_recdes, is_reusable, &context->supp_undo_lsa);
+	}
+      else
+	{
+	  /* log operation */
+	  heap_log_delete_physical (thread_p, context->home_page_watcher_p->pgptr, &context->hfid.vfid, &context->oid,
+				    &context->home_recdes, is_reusable, NULL);
+	}
```
```suggestion
	  /* log operation */
	  heap_log_delete_physical (thread_p, context->home_page_watcher_p->pgptr, &context->hfid.vfid, &context->oid,
				    &context->home_recdes, is_reusable, context->do_supplemental_log ? &context->supp_undo_lsa : NULL);
```

src/storage/heap_file.c
wrlawodms: refer to `heap_log_delete_physical()`

src/storage/heap_file.c
```diff
@@ -21141,8 +21211,18 @@ heap_delete_relocation (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * contex
        * of the heap type (reusable OID or not) as the relocated record
        * should not be referenced anywhere in the database.
        */
-      heap_log_delete_physical (thread_p, context->forward_page_watcher_p->pgptr, &context->hfid.vfid, &forward_oid,
-				&forward_recdes, true, NULL);
+      if (context->do_supplemental_log)
+	{
+	  /* if reusable slot is deleted, then postponed log (RVHF_MARK_REUSABLE_SLOT) will appended last. 
+	   * So, current log lsa after heap_log_delete_physical can points unexpected log, other than delete log. */
+	  heap_log_delete_physical (thread_p, context->forward_page_watcher_p->pgptr, &context->hfid.vfid, &forward_oid,
```
wrlawodms: same opinion to `heap_delete_home()`

src/storage/heap_file.c
```diff
@@ -4160,6 +4162,28 @@ heap_copy_chain (THREAD_ENTRY * thread_p, PAGE_PTR page_heap, HEAP_CHAIN * chain
...
+   */
+  if (prm_get_integer_value (PRM_ID_SUPPLEMENTAL_LOG) > 0 && !thread_p->no_supplemental_log
+      && !oid_is_system_class (classoid))
+    {
+      return true;
+    }
+  else
+    {
+      return false;
+    }
```
```suggestion
  return prm_get_integer_value (PRM_ID_SUPPLEMENTAL_LOG) > 0 
             && !thread_p->no_supplemental_log
             && !oid_is_system_class (classoid));
```

src/transaction/log_manager.c
```diff
@@ -321,6 +369,9 @@ static cubthread::daemon *log_Check_ha_delay_info_daemon = NULL;
 
 static cubthread::daemon *log_Flush_daemon = NULL;
 static std::atomic_bool log_Flush_has_been_requested = {false};
+
+static cubthread::daemon *cdc_log_Producer_daemon = NULL;
```
```suggestion
static cubthread::daemon *cdc_Log_producer_daemon = NULL;
```

src/transaction/log_manager.c
```diff
@@ -10214,6 +10408,3430 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
 
+static int
+cdc_log_producer (THREAD_ENTRY * thread_p)
+{
+  LOG_LSA cur_log_rec_lsa, next_log_rec_lsa;
+  LOG_LSA process_lsa;
+  LOG_LSA lsa;
+
+  LOG_PAGE *log_page_p = NULL;
+  char log_pgbuf[IO_MAX_PAGE_SIZE + MAX_ALIGNMENT];
```
wrlawodms, line 10504: IO_MAX_PAGE_SIZE -> LOG_PAGESIZE?

src/transaction/log_manager.c
```diff
@@ -10214,6 +10408,3430 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+  LOG_RECTYPE log_type;
+
+  if (LSA_ISNULL (&cdc_Gl.next_lsa))
+    {
+      pthread_mutex_lock (&cdc_Gl.lsa_init_lock);
+      pthread_cond_wait (&cdc_Gl.lsa_init_cond, &cdc_Gl.lsa_init_lock);
+      pthread_mutex_unlock (&cdc_Gl.lsa_init_lock);
+    }
+
+  temp_logbuf = (CDC_TEMP_LOGBUF *) malloc (2 * sizeof (CDC_TEMP_LOGBUF));
```
wrlawodms: Where is this released? Why do you allocate it dynamically, not on the stack memory?

src/transaction/log_manager.c
```diff
@@ -10214,6 +10408,3430 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+
+  /* *INDENT-OFF* */
+  std::unordered_map<TRANID, int > tran_ignore;
+  /* *INDENT-ON* */
+  char *tran_user;
+
+  CDC_LOGINFO_ENTRY *log_info_entry;
+
+  /*temp_logbuf consists of 2 entry, every time fetching new page, page will be stored in temp_logbuf[pageid%2] */
+  CDC_TEMP_LOGBUF *temp_logbuf = NULL;
```
wrlawodms: One buffer always seems to store the page `log_page_p` refers to. Is it necessary? we can define a macro like `LOG_READ_DOES_FIT` macro.

src/transaction/log_manager.c
```diff
@@ -10214,6 +10408,3430 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+
+  /*fetch log page */
+  if (logpb_fetch_page (thread_p, &process_lsa, LOG_CS_SAFE_READER, log_page_p) != NO_ERROR)
+    {
+      return ER_FAILED;
+    }
+
+  tmpbuf_index = process_lsa.pageid % 2;
+  temp_logbuf[tmpbuf_index].pageid = process_lsa.pageid;
+  memcpy (temp_logbuf[tmpbuf_index].log_page_p, log_page_p, IO_MAX_PAGE_SIZE);
```
wrlawodms: They seem to be initialized always together. It can be a function or you can define an assignment operator of CDC_TEMP_LOGBUF for safety and simplicity.

src/transaction/log_manager.c
```diff
@@ -10214,6 +10408,3430 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+
+  log_info_entry = (CDC_LOGINFO_ENTRY *) malloc (sizeof (CDC_LOGINFO_ENTRY));
```
wrlawodms: Where is this released? Why do you allocate it dynamically, not on the stack memory?

src/transaction/log_manager.c
```diff
@@ -10214,6 +10408,3430 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+      if (log_info_entry->length != 0)
+	{
+	  LSA_COPY (&log_info_entry->start_lsa, &next_log_rec_lsa);
+          /* *INDENT-OFF* */
+	  cdc_logInfo_queue->produce (log_info_entry);
+          /* *INDENT-ON* */
+
+	  LSA_COPY (&cdc_Gl.last_produced_lsa, &cur_log_rec_lsa);
+
+	  log_info_entry = (CDC_LOGINFO_ENTRY *) malloc (sizeof (CDC_LOGINFO_ENTRY));
```
wrlawodms: You seem to allocate it above, aren't you?

wrlawodms: And, if it weren't, ` if (log_info_entry->length != 0)` above looks cause a segfault.

src/transaction/log_manager.c
```diff
@@ -170,6 +170,33 @@ struct archive_log_header_scan_context
   LOG_ARV_HEADER header;
 };
 
+/*It will be moved to the CDC module and will define overflow.h in that module  */
+
+struct overflow_first_part
```
wrlawodms: I'd like to ask the reason for defining it again. It's because this is used on the client-side?

src/transaction/log_manager.c
```diff
@@ -170,6 +170,33 @@ struct archive_log_header_scan_context
...
+{
+  VPID next_vpid;
+  char data[1];			/* Really more than one */
+};
+
+typedef struct overflow_rest_part OVERFLOW_REST_PART;
+typedef struct overflow_first_part OVERFLOW_FIRST_PART;
+
+/* *INDENT-OFF* */
+static std::unordered_map <TRANID, char *> cdc_tran_user; /*to clear when log producer ends suddenly */
```
```suggestion
static std::unordered_map <TRANID, char *> cdc_Tran_user; /*to clear when log producer ends suddenly */
```

wrlawodms: Why is it a global variable? It seems that it is used only in `cdc_log_producer()`.

src/transaction/log_manager.c
```diff
@@ -10214,6 +10418,3430 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+		}
+
+	      LOG_READ_ADD_ALIGN (thread_p, sizeof (*supplement), &process_lsa, log_page_p);
+	      data = (char *) log_page_p->area + process_lsa.offset;
+	      memcpy (*user, data, supplement->length);
+	      *user[supplement->length] = '\0';
+	      return NO_ERROR;
+	    }
+	}
+      else if (log_rec_hdr->type == LOG_ABORT && log_rec_hdr->trid == trid)
```
wrlawodms, line 12660: What if the trx is committed? It seems that it searches to the end of the log file.

src/transaction/log_manager.c
```diff
@@ -10214,6 +10418,3430 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+		  (thread_p, temp_logbuf[tmpbuf_index].log_page_p, cur_log_rec_lsa, trid, &tran_user) == NO_ERROR)
+		{
+		  cdc_tran_user.insert (std::make_pair (trid, tran_user));
+		}
+	      else
+		{
+		  /* can not find user. It meets abort log. So, ignore the logs from this transaction */
+		  tran_ignore.insert (std::make_pair (trid, 1));
+		}
+	    }
```
wrlawodms: LOG_COMMIT or LOG_ABORT isn't the end of log of the trx? You can search a user name after either record?

src/transaction/log_manager.c
```diff
@@ -10214,6 +10418,3430 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+
+	  if (is_zip_supplement && supplement_length != 0)
+	    {
+	      supp_zip = log_append_get_zip_undo (thread_p);
+
+	      is_unzip_supplement = log_unzip (supp_zip, supplement_length, supplement_data);
+	    }
+
+	  if (is_zip_supplement && is_unzip_supplement)
```
wrlawodms: `is_zip_supplement ` means the log record is zipped, and `is_unzip_supplement` means that it succeeds to unzip. If it is right, the name of variables should be distinguishable.

src/transaction/log_manager.c
```diff
@@ -10214,6 +10418,3430 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+
+	      logpb_copy_from_log (thread_p, supplement_data, supplement_length, &process_lsa, log_page_p);
+	      is_alloced = true;
+	    }
+
+	  if (is_zip_supplement && supplement_length != 0)
+	    {
+	      supp_zip = log_append_get_zip_undo (thread_p);
+
+	      is_unzip_supplement = log_unzip (supp_zip, supplement_length, supplement_data);
```
wrlawodms: Needs error handling when log_unzip() fails

src/transaction/log_manager.c
```diff
@@ -10214,6 +10418,3430 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+	  if (is_zip_supplement && is_unzip_supplement)
+	    {
+	      supplement_length = (int) supp_zip->data_length;
+	      if (is_alloced)
+		{
+		  free_and_init (supplement_data);
+		}
+
+	      supplement_data = supp_zip->log_data;
+	    }
```
wrlawodms: This seems to be able to put into the upper if-condition.

src/transaction/log_manager.c
```diff
@@ -10214,6 +10418,3430 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+		  /* JOOHOK : error handling when user not found */
+		  if (cdc_find_user
+		      (thread_p, temp_logbuf[tmpbuf_index].log_page_p, cur_log_rec_lsa, trid, &tran_user) == NO_ERROR)
+		    {
+		      cdc_tran_user.insert (std::make_pair (trid, tran_user));
+		    }
+		  else
+		    {
+		      /* can not find user. It meets abort log. So, ignore the logs from this transaction */
+		      tran_ignore.insert (std::make_pair (trid, 1));
```
wrlawodms: After this, the sppl. log which is being processed seems to has to be skipped.

src/transaction/log_manager.c
```diff
@@ -10214,6 +10418,3430 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+
+		  break;
+		}
+	    case LOG_SUPPLEMENT_INSERT:
+	      memcpy (&classoid, supplement_data, sizeof (OID));
+	      memcpy (&redo_lsa, supplement_data + sizeof (OID), sizeof (LOG_LSA));
+
+	      if (cdc_get_recdes (thread_p, temp_logbuf, NULL, NULL, &redo_lsa, &redo_recdes) != NO_ERROR)
+		{
+		  break;
```
wrlawodms: Just skip this record?

src/transaction/log_manager.c
```diff
@@ -10214,6 +10418,3430 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+		}
+
+	      cdc_make_dml_loginfo (thread_p, trid, tran_user, INSERT, classoid, NULL, &redo_recdes, log_info_entry);
```
wrlawodms: error handling

src/transaction/log_manager.c
```diff
@@ -10214,6 +10418,3430 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+	      break;
+	    case LOG_SUPPLEMENT_UPDATE:
+	      memcpy (&classoid, supplement_data, sizeof (OID));
+	      memcpy (&undo_lsa, supplement_data + sizeof (OID), sizeof (LOG_LSA));
+	      memcpy (&redo_lsa, supplement_data + sizeof (OID) + sizeof (LOG_LSA), sizeof (LOG_LSA));
+
+	      cdc_get_recdes (thread_p, temp_logbuf, &undo_lsa, &undo_recdes, &redo_lsa, &redo_recdes);
```
wrlawodms: undo_recdes->data can be malloced in `cdc_get_recdes()`, but I can't find `free()` for it.

src/transaction/log_manager.c
```diff
@@ -10214,6 +10418,3430 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+
+      LSA_COPY (&process_lsa, undo_lsa);
+
+      log_rec_hdr = LOG_GET_LOG_RECORD_HEADER (log_page_p, &process_lsa);
+
+      LOG_READ_ADD_ALIGN (thread_p, sizeof (*log_rec_hdr), &process_lsa, log_page_p);
+
+      switch (log_rec_hdr->type)
+	{
+	case LOG_SUPPLEMENTAL_INFO:
```
wrlawodms, line 11097: Question: Is this only for LOG_SUPPLEMENT_UNDO_RECORD?

src/transaction/log_manager.c
```diff
@@ -10214,6 +10418,3430 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+	      goto end;
+	    }
+	}
+
+      LSA_COPY (&process_lsa, undo_lsa);
+
+      log_rec_hdr = LOG_GET_LOG_RECORD_HEADER (log_page_p, &process_lsa);
+
+      LOG_READ_ADD_ALIGN (thread_p, sizeof (*log_rec_hdr), &process_lsa, log_page_p);
+
```
wrlawodms, line 11094: Is there no case in which we should traverse back to find a undo/redo data except for log records for overflow file?

src/transaction/log_manager.c
```diff
@@ -10214,6 +10418,3430 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+		    is_undo_alloced = true;
+		  }
+
+
+		if (is_zipped_undo && undo_length != 0)
+		  {
+		    undo_zip_ptr = log_zip_alloc (IO_PAGESIZE);
+		    if (undo_zip_ptr == NULL)
+		      {
+			return ER_OUT_OF_VIRTUAL_MEMORY;
```
```suggestion
			return er_errid();
```

src/transaction/log_manager.c
```diff
@@ -10214,6 +10418,3430 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+		else
+		  {
+		    undo_data = (char *) malloc (undo_length);
+		    if (undo_data == NULL)
+		      {
+			error_code = ER_OUT_OF_VIRTUAL_MEMORY;
+			goto end;
+		      }
+
+		    logpb_copy_from_log (thread_p, undo_data, undo_length, &process_lsa, log_page_p);
```
wrlawodms: In this function, log_page_p can go forward several pages. We don't need to consider it for `temp_logbuf`?

src/transaction/log_manager.c
```diff
@@ -10214,6 +10418,3430 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+		error_code = ER_OUT_OF_VIRTUAL_MEMORY;
+		goto end;
+	      }
+
+	    memcpy (undo_recdes->data, (char *) undo_data, undo_length);
+
+	    break;
+	  }
+	case LOG_MVCC_DIFF_UNDOREDO_DATA:
+	case LOG_MVCC_UNDOREDO_DATA:
```
wrlawodms, line 11108: Could you tell me the reason you make this routine on your own, not using `log_get_undo_record()`?

src/transaction/log_manager.c
```diff
@@ -10214,6 +10418,3430 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+	      {
+		/* GET OVF UNDO IMAGE */
+		if ((error_code =
+		     cdc_get_overflow_recdes (thread_p, log_page_p, undo_recdes, *undo_lsa, rcvindex,
+					      false)) != NO_ERROR)
+		  {
+		    goto end;
+		  }
+	      }
+
```
wrlawodms, line 11217:
```suggestion
	      }
	      else 
	      {
	       assert (false);
	      }

```

src/transaction/log_manager.c
```diff
@@ -10214,6 +10418,3430 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+	    mvcc_undoredo = (LOG_REC_MVCC_UNDOREDO *) (log_page_p->area + process_lsa.offset);
+	    rcvindex = mvcc_undoredo->undoredo.data.rcvindex;
+	    redo_length = mvcc_undoredo->undoredo.rlength;
+	    undo_length = mvcc_undoredo->undoredo.ulength;
+
+	    if (LOG_IS_DIFF_UNDOREDO_TYPE (log_rec_hdr->type) == true)
+	      {
+		is_diff = true;
+	      }
```
wrlawodms, line 11283:
```suggestion
		is_diff = (OG_IS_DIFF_UNDOREDO_TYPE (log_rec_hdr->type);
```

src/transaction/log_manager.c
```diff
@@ -10214,6 +10418,3430 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+
+		if (ZIP_CHECK (redo_length))
+		  {
+		    redo_length = (int) GET_ZIP_LEN (redo_length);
+		    is_zipped_redo = true;
+		  }
+
+		if (process_lsa.offset + redo_length < (int) LOGAREA_SIZE)
+		  {
+		    redo_data = (char *) (log_page_p->area + process_lsa.offset);
```
wrlawodms, line 11299: Is it OK no to skip the undo part?

wrlawodms: Even if the undo part can be skipped, how about checking the checking and skipping. This code assumes the undo part of RVHF_MVCC_INSERT is 0, but it may be changed later.

src/transaction/log_manager.c
```diff
@@ -10214,6 +10418,3430 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+
+      if (is_undo_alloced)
+	{
+	  free_and_init (undo_data);
+	  is_undo_alloced = false;
+	}
+    }
+
+/* Get REDO RECDES from redo lsa */
+  if (redo_lsa != NULL)
```
wrlawodms, line 11227: Have you considered the generalized way to extract a redo log record like log_recovery_redo()? They don't deal with log records rcvindex by rcvindex. I guess you see that code already. please tell me why you should do it case by case?

src/transaction/log_manager.c
```diff
@@ -10214,6 +10418,3430 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+	{
+	  error_code = ER_OUT_OF_VIRTUAL_MEMORY;
+	  goto end;
+	}
+
+      for (i = 0; i < attr_info.num_values; i++)
+	{
+	  heap_value = &attr_info.values[i];
+	  oldval_deforder = heap_value->read_attrepr->def_order;
+	  memcpy (&old_values[oldval_deforder], &heap_value->dbvalue, sizeof (DB_VALUE));
```
wrlawodms: How about just using old_attr_info and new_attr_info ?

src/transaction/log_manager.c
```diff
@@ -10214,6 +10418,3430 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+	  error_code = ER_FAILED;
+	  goto end;
+	}
+    }
+  else
+    {
+      has_pk = 0;
+    }
+
+  if (record_length > 0)
```
wrlawodms: can be <= 0?

src/transaction/log_manager.c
```diff
@@ -10214,6 +10418,3430 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+
+	  for (i = 0; i < num_change_col; i++)
+	    {
+	      if ((error_code = cdc_put_value_to_loginfo (&new_values[i], &ptr)) != NO_ERROR)
+		{
+		  goto end;
+		}
+	    }
+
+	  ptr = or_pack_int (ptr, num_cond_col);
```
wrlawodms: not num_change_col ?

wrlawodms (commented):
I took a look at it in a hurry because of the deadline. Please re-request review later when it's ready.

src/communication/network_interface_sr.c
```diff
@@ -10240,7 +10240,7 @@ scdc_get_lsa (THREAD_ENTRY * thread_p, unsigned int rid, char *request, int reql
 
   ptr = or_unpack_int64 (request, &input_time);
 
-  error = cdc_get_lsa (thread_p, input_time, &start_lsa);
+  error = cdc_get_lsa (thread_p, &input_time, &start_lsa);
```
hornetmj: The un-used variable 'b_start_lsa' should be removed.

src/parser/parse_tree.h
```diff
@@ -1199,8 +1199,9 @@ typedef enum
   /* SELECT record info from tuple header instead of data */
   PT_HINT_SELECT_PAGE_INFO = 0x40000000,	/* 0100 0000 0000 0000 0000 0000 0000 0000 */
   /* SELECT page header information from heap file instead of record data */
-  PT_HINT_SELECT_KEY_INFO = 0x80000000	/* 1000 0000 0000 0000 0000 0000 0000 0000 */
-    /* SELECT key information from index b-tree instead of table record data */
+  PT_HINT_SELECT_KEY_INFO = 0x80000000,	/* 1000 0000 0000 0000 0000 0000 0000 0000 */
+  /* SELECT key information from index b-tree instead of table record data */
+  PT_HINT_NO_SUPPLEMENTAL_LOG = 0x100000000
```
hornetmj: Adds comment in the same manner. And I think you need to leave the purpose of this hint and descriptions.

src/api/cubrid_log.c
```diff
@@ -0,0 +1,1533 @@
...
+ */
+int
+cubrid_log_set_extraction_table (uint64_t * classoid_arr, int arr_size)
+{
+  if (g_stage != CUBRID_LOG_STAGE_CONFIGURATION)
+    {
+      return CUBRID_LOG_INVALID_FUNC_CALL_STAGE;
+    }
+
+  if ((classoid_arr == NULL && arr_size != 0) || (classoid_arr != NULL && arr_size == 0))
```
hornetmj, line 271:
```suggestion
  if (classoid_arr == NULL || arr_size == 0)
```

src/api/cubrid_log.c
```diff
@@ -0,0 +1,1533 @@
...
+    {
+      return CUBRID_LOG_INVALID_CLASSOID_ARR_SIZE;
```
hornetmj, line 273: If classoid_arr == NULL, then the error code is not correct. So, it needs to be reconsidered.

src/api/cubrid_log.c
```diff
@@ -0,0 +1,1533 @@
...
+cubrid_log_set_extraction_user (char **user_arr, int arr_size)
+{
+  int i;
+
+  if (g_stage != CUBRID_LOG_STAGE_CONFIGURATION)
+    {
+      return CUBRID_LOG_INVALID_FUNC_CALL_STAGE;
+    }
+
+  if ((user_arr == NULL && arr_size != 0) || (user_arr != NULL && arr_size == 0))
```
hornetmj, line 304:
```suggestion
  if (user_arr == NULL || arr_size == 0)
```

src/api/cubrid_log.c
```diff
@@ -0,0 +1,1533 @@
...
+
+#include "connection_cl.h"
+#include "connection_list_cl.h"
+#include "cubrid_log.h"
+#include "log_lsa.hpp"
+#include "network.h"
+#include "object_representation.h"
+
+#define CUBRID_LOG_ERROR_HANDLING(e, v) \
+  (err_code) = (e); \
```
hornetmj, line 46: check the alignment.

src/api/cubrid_log.c
```diff
@@ -0,0 +1,1533 @@
...
+{
+  int64_t curr_size = 0;
+  int64_t tracelog_size = 0;
+
+  curr_size = ftell (g_trace_log);
+  tracelog_size = strlen (buf);
+
+  if (curr_size + tracelog_size > g_trace_log_filesize)
+    {
+      fseek (g_trace_log, 0, SEEK_SET);
```
hornetmj, line 158: In this case, the loggings wrote previously exist in the trace log file. It needs to handle.

src/api/cubrid_log.c
```diff
@@ -0,0 +1,1533 @@
...
+    {
+      CUBRID_LOG_ERROR_HANDLING (CUBRID_LOG_INVALID_DBNAME, trace_errbuf);
+    }
+
+  if (id == NULL)
+    {
+      CUBRID_LOG_ERROR_HANDLING (CUBRID_LOG_INVALID_ID, trace_errbuf);
+    }
+
+  if (password == NULL)
```
hornetmj, line 564: The default 'PUBLIC' and 'DBA' users have no passwords. Do you have any idea how to handle it?

src/api/cubrid_log.c
```diff
@@ -0,0 +1,1533 @@
...
+  int err_code;
+
+  char trace_errbuf[1024];
+
+  if (g_stage != CUBRID_LOG_STAGE_PREPARATION)
+    {
+      CUBRID_LOG_ERROR_HANDLING (CUBRID_LOG_INVALID_FUNC_CALL_STAGE, trace_errbuf);
+    }
+
+  if (timestamp <= 0)
```
hornetmj, line 698: The timestamp variable is a pointer type. So, this condition check is invalid.

Anyway, I think there is no reason that the timestamp variable is a pointer type. How about reconsider changing it to timestamp type?

src/api/cubrid_log.c
```diff
@@ -0,0 +1,1533 @@
...
+    {
+      free_and_init (recv_data);
+    }
+
+  queue_entry = css_find_queue_entry (g_conn_entry->buffer_queue, rid);
+  if (queue_entry != NULL)
+    {
+      queue_entry->buffer = NULL;
+      css_queue_remove_header_entry_ptr (&g_conn_entry->buffer_queue, queue_entry);
+    }
```
hornetmj, line 675: This routine is repeated in several functions. So, I think it seems nice to make a macro or inline function.

src/api/cubrid_log.c
```diff
@@ -0,0 +1,1533 @@
...
+
+  if (recv_data == NULL || recv_data_size != OR_INT_SIZE + OR_LOG_LSA_ALIGNED_SIZE + OR_INT64_SIZE)
+    {
+      CUBRID_LOG_ERROR_HANDLING (CUBRID_LOG_FAILED_CONNECT, trace_errbuf);
+    }
+
+  ptr = or_unpack_int (recv_data, &reply_code);
+
+  if (reply_code == NO_ERROR || reply_code == ER_CDC_LSA_NOT_FOUND)
```
hornetmj: Isn't the ER_CDC_LSA_NOT_FOUND error? Does it have the LSA and timestamp information?

If it is, Is it OK to return just CUBRID_LOG_SUCCESS?

src/api/cubrid_log.c
```diff
@@ -0,0 +1,1533 @@
...
+    {
+      CUBRID_LOG_ERROR_HANDLING (CUBRID_LOG_INVALID_FUNC_CALL_STAGE, trace_errbuf);
+    }
+
+  if (lsa == NULL || log_item_list == NULL || list_size == NULL)
+    {
+      CUBRID_LOG_ERROR_HANDLING (CUBRID_LOG_INVALID_OUT_PARAM, trace_errbuf);
+    }
+
+  memcpy (&g_next_lsa, lsa, sizeof (LOG_LSA));
```
hornetmj, line 1287: The 'g_next_lsa' and 'lsa' should be equals in the normal processing. So, this condition should be checked together when the function argument 'lsa' is checked.

src/api/cubrid_log.c
```diff
@@ -0,0 +1,1533 @@
...
+
+  rc = cubrid_log_extract_internal (&g_next_lsa, &num_infos, &total_length);
+
+  if (rc != CUBRID_LOG_SUCCESS)
```
hornetmj, line 1291: The return value 'CUBRID_LOG_SUCCESS_WITH_NO_LOGITEM' should be considered.

src/api/cubrid_log.c
```diff
@@ -0,0 +1,1533 @@
...
+
+  ptr = *data_info;
+
+  ptr = or_unpack_int (ptr, &dml->dml_type);
+  ptr = or_unpack_int64 (ptr, (INT64 *) & dml->classoid);
+  ptr = or_unpack_int (ptr, &dml->num_changed_column);
+
+  if (dml->num_changed_column)
+    {
+      // now, just malloc for validation and later, optimize it.
```
hornetmj, line 903: I think it is unnecessary to malloc/free every time. So, I request you to make a simple data buffer. (after releasing the second test version)

src/api/cubrid_log.c
```diff
@@ -0,0 +1,1533 @@
...
+  int rc;
+
+  char trace_errbuf[1024];
+
+  if (g_stage != CUBRID_LOG_STAGE_EXTRACTION)
+    {
+      CUBRID_LOG_ERROR_HANDLING (CUBRID_LOG_INVALID_FUNC_CALL_STAGE, trace_errbuf);
+    }
+
+  if (log_item_list == NULL)
```
hornetmj, line 1372: The address of the log_item_list is equaled to g_log_items. So, It needs to check.

hornetmj (commented):
I reviewed the client(API + lib) first.

src/communication/network.h
```diff
@@ -248,6 +248,14 @@ enum net_server_request
   NET_SERVER_LD_UPDATE_STATS,
 
   NET_SERVER_VACUUM_DUMP,
+
+  /*FOR CDC */
```
hornetmj, line 254:
```suggestion
  /* For CDC */
```

hornetmj: Or remove this comment.

src/communication/network_interface_sr.h
```diff
@@ -185,6 +185,7 @@ extern void srepl_set_info (THREAD_ENTRY * thread_p, unsigned int rid, char *req
 extern void srepl_log_get_append_lsa (THREAD_ENTRY * thread_p, unsigned int rid, char *request, int reqlen);
 extern void slocator_check_fk_validity (THREAD_ENTRY * thread_p, unsigned int rid, char *request, int reqlen);
 extern void slogwr_get_log_pages (THREAD_ENTRY * thread_p, unsigned int rid, char *request, int reqlen);
+
```
hornetmj, line 188:
```suggestion
```

src/communication/network_interface_sr.h
```diff
@@ -238,4 +239,11 @@ extern void sloaddb_destroy (THREAD_ENTRY * thread_p, unsigned int rid, char *re
 extern void sloaddb_interrupt (THREAD_ENTRY * thread_p, unsigned int rid, char *request, int reqlen);
 extern void sloaddb_update_stats (THREAD_ENTRY * thread_p, unsigned int rid, char *request, int reqlen);
 extern void ssession_stop_attached_threads (void *session);
+
+/*for CDC*/
```
hornetmj, line 243:
```suggestion
/* For CDC */
```

hornetmj: Or remove this comment.

src/communication/network_interface_sr.c
```diff
@@ -8019,6 +8020,8 @@ slogwr_get_log_pages (THREAD_ENTRY * thread_p, unsigned int rid, char *request,
   return;
 }
 
+
+
```
hornetmj, line 8091:
```suggestion
```

src/communication/network_interface_sr.c
```diff
@@ -10223,3 +10226,209 @@ ssession_stop_attached_threads (void *session)
...
+      extraction_user = (char **) malloc (sizeof (char *) * num_extraction_user);
+      if (extraction_user == NULL)
+	{
+	  error = ER_OUT_OF_VIRTUAL_MEMORY;
+	  goto end;
+	}
+
+      for (int i = 0; i < num_extraction_user; i++)
+	{
+	  ptr = or_unpack_string (ptr, &dummy_user);
```
hornetmj:
```suggestion
	  ptr = or_unpack_string_nocopy (ptr, &dummy_user);
```

src/communication/network_interface_sr.c
```diff
@@ -10223,3 +10226,209 @@ ssession_stop_attached_threads (void *session)
...
+	  extraction_user_len = strlen (dummy_user);
+	  extraction_user[i] = (char *) malloc (extraction_user_len + 1);
+
+	  if (extraction_user[i] == NULL)
+	    {
+	      for (int j = 0; j <= i; j++)
```
hornetmj:
```suggestion
	      for (int j = 0; j < i; j++)
```

src/communication/network_interface_sr.c
```diff
@@ -10223,3 +10226,209 @@ ssession_stop_attached_threads (void *session)
...
+		  free (extraction_user[j]);
+		}
+	      free (extraction_user);
+	      db_private_free_and_init (NULL, dummy_user);
+
+	      error = ER_OUT_OF_VIRTUAL_MEMORY;
+	      goto end;
+	    }
+
+	  memcpy (extraction_user[i], dummy_user, extraction_user_len);
```
hornetmj: How about use strncpy?

src/communication/network_interface_sr.c
```diff
@@ -10223,3 +10226,209 @@ ssession_stop_attached_threads (void *session)
...
+	{
+	  if (num_extraction_user > 0)
+	    {
+	      for (int i = 0; i < num_extraction_user; i++)
+		{
+		  free (extraction_user[i]);
+		}
+
+	      free (extraction_user);
+	    }
```
hornetmj:
```suggestion
```

src/communication/network_interface_sr.c
```diff
@@ -10223,3 +10226,209 @@ ssession_stop_attached_threads (void *session)
...
+
+	  error = ER_OUT_OF_VIRTUAL_MEMORY;
+	  goto end;
+	}
+      for (int i = 0; i < num_extraction_class; i++)
```
hornetmj, line 10354:
```suggestion

      for (int i = 0; i < num_extraction_class; i++)
```

src/communication/network_interface_sr.c
```diff
@@ -10223,3 +10226,209 @@ ssession_stop_attached_threads (void *session)
...
+	{
+	  ptr = or_unpack_int64 (ptr, (INT64 *) & extraction_classoids[i]);
+	}
+    }
+
+  error =
+    cdc_set_configuration (max_log_item, timeout, all_in_cond, extraction_user, num_extraction_user,
```
hornetmj:
```suggestion
    cdc_set_configuration (max_log_item, extraction_timeout, all_in_cond, extraction_user, num_extraction_user,
```

hornetmj: And you might remove the unused variable 'timeout'.

src/communication/network_interface_sr.c
```diff
@@ -10223,3 +10226,209 @@ ssession_stop_attached_threads (void *session)
...
+
+  error = cdc_initialize ();
+  if (error != NO_ERROR)
+    {
+      goto end;
+    }
+
+end:
+
+  ptr = or_pack_int (reply, error);
```
hornetmj: How about using the error_code iso error? I think it is consistent compare to the other function in this file.

src/communication/network_interface_sr.c
```diff
@@ -10223,3 +10226,209 @@ ssession_stop_attached_threads (void *session)
...
+  uint64_t b_start_lsa;
+  int error;
+
+  ptr = or_unpack_int64 (request, &input_time);
+
+  error = cdc_find_lsa (thread_p, &input_time, &start_lsa);
+
+  ptr = or_pack_int (reply, error);
+  ptr = or_pack_log_lsa (ptr, &start_lsa);
+  or_pack_int64 (ptr, input_time);
```
hornetmj: My guess is that input_time is semantically incorrect. Therefore, it is recommended to modify the variable name of input_time according to the meaning.

hornetmj: Do we need to send additional information(start_lsa, imput_time) in case of an error?

hornetmj (commented):
I reviewed the system parameter and network communications. But I will review the scdc_get_logitem_info and scdc_get_logitem functions again when I review the cdc_ function series because I need to understand how to work the cdc_Server_comm.is_sent flag.

src/api/cubrid_log.h
```diff
@@ -0,0 +1,151 @@
...
+  extern int cubrid_log_set_connection_timeout (int timeout);
+  extern int cubrid_log_set_extraction_timeout (int timeout);
+  extern int cubrid_log_set_tracelog (char *path, int level, int filesize);
+  extern int cubrid_log_set_max_log_item (int max_log_item);
+  extern int cubrid_log_set_all_in_cond (int retrieve_all);
+  extern int cubrid_log_set_extraction_table (uint64_t * classoid_arr, int arr_size);
+  extern int cubrid_log_set_extraction_user (char **user_arr, int arr_size);
+
+/* API for the preparation step */
+  extern int cubrid_log_connect_server (char *host, int port, char *dbname, char *id, char *password);
```
hornetmj, line 141:
```suggestion
  extern int cubrid_log_connect_server (char *host, int port, char *dbname, char *user, char *password);
```

src/transaction/log_impl.h
```diff
@@ -759,6 +767,100 @@ typedef struct log_logging_stat
...
+{
+  INSERT = 0,
+  UPDATE,
+  DELETE,
+  TRUNCATE
+} CDC_DML_TYPE;
+
+typedef struct cdc_server_comm
+{
+  char *log_items;		/* log item list. it is used as buffer to send to client */
```
hornetmj: Do you want to call it log_item? I remembered you said it log_info.

src/transaction/log_impl.h
```diff
@@ -788,6 +890,14 @@ extern char log_Name_volinfo[];
 extern char log_Name_bg_archive[];
 extern char log_Name_removed_archive[];
 
+/*CDC global variables */
+extern CDC_GLOBAL_INFO cdc_Gl;
+
+/* *INDENT-OFF* */
+extern lockfree::circular_queue<CDC_LOGINFO_ENTRY *> *cdc_logInfo_queue;
+/* *INDENT-ON* */
+extern CDC_SERVER_COMM cdc_Server_comm;
```
hornetmj: How about combining the global structures related to the CDC?

I think it is not good that the mutex related to the cdc_logInfo_queue exists in other global variable cdc_Gl.init_queue_lock.

hornetmj: How about initializing the members of a global variable at the same time?
(e.g. cdc_Gl.init_queue_lock is initialized in the cdc_log_producer_daemon_init() when booting but, others in the cdc_set_configuration())

IMHO, I don't think good readable code even though it works well.

src/transaction/log_impl.h
```diff
@@ -759,6 +767,100 @@ typedef struct log_logging_stat
...
+
+  pthread_cond_t lsa_init_cond;
+  pthread_cond_t nxio_lsa_cond;
+  pthread_cond_t queue_produce_cond;
+  pthread_cond_t queue_consume_cond;
+  pthread_cond_t finalize_cond;
+  pthread_cond_t queue_init_cond;
+
+  uint64_t queue_size;
+  bool is_queue_initialized;
```
hornetmj: How about replacing to cdc_logInfo_queue != NULL check?

src/transaction/log_manager.c
```diff
@@ -10214,6 +10426,3742 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+      cdc_logInfo_queue = new lockfree::circular_queue <CDC_LOGINFO_ENTRY *> (MAX_CDC_LOGINFO_QUEUE_ENTRY);
+      /* *INDENT-ON* */
+
+      cdc_Gl.is_queue_initialized = true;
+      cdc_Gl.queue_size = 0;
+      LSA_SET_NULL (&cdc_Gl.last_produced_lsa);
+      LSA_SET_NULL (&cdc_Gl.last_consumed_lsa);
+
+      pthread_cond_signal (&cdc_Gl.queue_init_cond);
+      pthread_mutex_unlock (&cdc_Gl.init_queue_lock);
```
```suggestion
      pthread_mutex_unlock (&cdc_Gl.init_queue_lock);
      pthread_cond_signal (&cdc_Gl.queue_init_cond);
```

src/transaction/log_manager.c
```diff
@@ -10214,6 +10426,3742 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+}
+
+/*
+ * arv_num (in) : archive log volume number to traverse. If it is -1, then traverse active log volume. 
+ * ret_lsa (out) : lsa of the first log which contains time info 
+ * time (out) : time of the first log which contains time info  
+ */
+
+static int
+get_start_point_from_file (THREAD_ENTRY * thread_p, int arv_num, LOG_LSA * ret_lsa, time_t * time)
```
hornetmj: prefix.

src/transaction/log_manager.c
```diff
@@ -10214,6 +10426,3742 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+  LOG_CS_EXIT (thread_p);
+
+  if ((error_code = logpb_fetch_page (thread_p, &process_lsa, LOG_CS_SAFE_READER, log_pgptr)) != NO_ERROR)
+    {
+      return error_code;
+    }
+
+  while (!LSA_ISNULL (&process_lsa))
+    {
+      log_rec_header = LOG_GET_LOG_RECORD_HEADER (log_pgptr, &process_lsa);
```
hornetmj, line 13556: At the first loop, the process_lsa.offset is 0.

I think that using the LOG_PAGE.hdr.offset is more correct.

src/transaction/log_manager.c
```diff
@@ -10214,6 +10426,3742 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+		  *time = 0;	/* can not know time of latest log */
+		  is_found = true;
+		  error = ER_CDC_LSA_NOT_FOUND;
+
+		}
+	    }
+	}
+    }
+
+  if (is_found)
```
hornetmj, line 13378: is_found is always true at this point.

src/transaction/log_manager.c
```diff
@@ -10214,6 +10426,3742 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+  /*AT first, compare the time in active log volume. */
+  error = get_start_point_from_file (thread_p, -1, &ret_lsa, &active_start_time);
+  if (error == ER_FAILED || error == ER_LOG_READ)
+    {
+      return error;
+    }
+  else
+    {
+      /* NO ERROR */
+      if (active_start_time != 0 && active_start_time <= *time)
```
hornetmj, line 13290: if active_start_time == *time, then just use ret_lsa as start_lsa.

And I think it is nice if a user could set the time as -1 (or something) then, the server returns a LSA from default action(e.g. the first LSA of the active log volume). It is my suggestion for usability.

src/transaction/log_manager.c
```diff
@@ -10214,6 +10426,3742 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+  return ER_CDC_LSA_NOT_FOUND;
+}
+
+/*
+ * time (in/out) : Time to compare (in) and actual time of log for start_lsa (out)
+ * start_lsa (in/out) : start point (in) and lsa of LOG which is found (out)  
+ */
+
+static int
+get_lsa_with_start_point (THREAD_ENTRY * thread_p, time_t * time, LOG_LSA * start_lsa)
```
hornetmj: prefix.

src/transaction/log_manager.c
```diff
@@ -10214,6 +10426,3742 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+  log_page_p->hdr.logical_pageid = NULL_PAGEID;
+  log_page_p->hdr.offset = NULL_OFFSET;
+  bool is_active = false;
+
+  int error = NO_ERROR;
+
+  if (LSA_ISNULL (start_lsa))
+    {
+      is_active = true;
+    }
```
hornetmj, line 13632:
```suggestion
```

hornetmj: It may be the assertion case.

src/transaction/log_manager.c
```diff
@@ -10214,6 +10426,3742 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+      LOG_READ_ADD_ALIGN (thread_p, sizeof (*log_rec_header), &process_lsa, log_page_p);
+
+      if (log_rec_header->type == LOG_COMMIT || log_rec_header->type == LOG_ABORT)
+	{
+	  LOG_READ_ADVANCE_WHEN_DOESNT_FIT (thread_p, sizeof (*donetime), &process_lsa, log_page_p);
+	  donetime = (LOG_REC_DONETIME *) (log_page_p->area + process_lsa.offset);
+	  if (donetime->at_time >= *time)
+	    {
+	      *time = donetime->at_time;
+	      LSA_COPY (start_lsa, &log_rec_header->forw_lsa);
```
```suggestion
	      LSA_COPY (start_lsa, &forw_lsa);
```

src/transaction/log_manager.c
```diff
@@ -10214,6 +10426,3742 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+{
+  LOG_LSA process_lsa;
+
+  LOG_RECORD_HEADER *log_rec_header;
+  LOG_PAGE *log_page_p = NULL;
+  char *log_pgbuf[IO_MAX_PAGE_SIZE + MAX_ALIGNMENT];
+
+  LOG_REC_DONETIME *donetime;
+  LOG_REC_HA_SERVER_STATE *dummy;
+  time_t at_time;
```
hornetmj, line 13618:
```suggestion
```

src/transaction/log_manager.c
```diff
@@ -10214,6 +10426,3742 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+      aligned_hdr_pgbuf = PTR_ALIGN (hdr_pgbuf, MAX_ALIGNMENT);
+
+      hdr_pgptr = (LOG_PAGE *) aligned_hdr_pgbuf;
+
+      fileio_make_log_archive_name (arv_name, log_Archive_path, log_Prefix, arv_num);
+
+      if (fileio_is_volume_exist (arv_name) == true)
+	{
+	  vdes = fileio_mount (thread_p, log_Db_fullname, arv_name, LOG_DBLOG_ARCHIVE_VOLID, false, false);
+	  if (vdes != NULL_VOLDES)
```
hornetmj, line 13507: error handling if vdes == NULL_VOLDES.

src/transaction/log_manager.c
```diff
@@ -10214,6 +10426,3742 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+#endif
+      LSA_COPY (&cdc_Gl.next_lsa, start_lsa);
+      /* pthread_cond_signal () to cdc_log_producer and queue re-initialization */
+      if (cdc_Gl.is_queue_initialized == false)
+	{
+	  cdc_reinitialize_queue (start_lsa);
+	}
+      pthread_mutex_lock (&cdc_Gl.next_lsa_lock);
+      pthread_cond_signal (&cdc_Gl.lsa_init_cond);
+      pthread_mutex_unlock (&cdc_Gl.next_lsa_lock);
```
```suggestion
      pthread_cond_signal (&cdc_Gl.lsa_init_cond);
```

src/transaction/log_manager.c
```diff
@@ -10214,6 +10426,3742 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+}
+
+void
+cdc_daemons_init ()
+{
+#if 1				//JOOHOK
+  _er_log_debug (ARG_FILE_LINE, "cdc_daemons_init ");
+#endif
+  cdc_log_producer_daemon_init ();
+  cdc_thread_checker_daemon_init ();
```
hornetmj: Is this daemon thread absolutely necessary?

src/transaction/log_manager.c
```diff
@@ -10214,6 +10426,3742 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+#if defined (SERVER_MODE)
+static void
+cdc_thread_checker (THREAD_ENTRY * thread_p)
+{
+  LOG_LSA nxio_lsa = log_Gl.append.get_nxio_lsa ();
+  if (LSA_LE (&cdc_Gl.next_lsa, &nxio_lsa))
+    {
+      pthread_mutex_lock (&cdc_Gl.next_lsa_lock);
+      pthread_cond_signal (&cdc_Gl.nxio_lsa_cond);
+      pthread_mutex_unlock (&cdc_Gl.next_lsa_lock);
```
```suggestion
      pthread_cond_signal (&cdc_Gl.nxio_lsa_cond);
```

src/transaction/log_manager.c
```diff
@@ -10214,6 +10426,3742 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+#if 1				//JOOHOK
+  _er_log_debug (ARG_FILE_LINE, "cdc_log_producer : next_lsa_lock");
+#endif
+  if (LSA_ISNULL (&cdc_Gl.next_lsa))
+    {
+      if (cdc_Gl.is_finalize == true)
+	{
+	  pthread_mutex_unlock (&cdc_Gl.next_lsa_lock);
+	  goto end;
+	}
```
hornetmj: I think this condition never occurs.

src/transaction/log_manager.c
```diff
@@ -10214,6 +10426,3742 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+      if (cdc_Gl.is_finalize == true)
+	{
+	  pthread_mutex_unlock (&cdc_Gl.queue_consume_lock);
+	  goto end;
+	}
+      pthread_cond_wait (&cdc_Gl.queue_consume_cond, &cdc_Gl.queue_consume_lock);
+    }
+  pthread_mutex_unlock (&cdc_Gl.queue_consume_lock);
+
+  pthread_mutex_lock (&cdc_Gl.is_finalize_lock);
```
hornetmj: The previous 'if (cdc_Gl.is_finalize == true)' checking doesn't get the 'cdc_Gl.is_finalize_lock' mutex.

Anyway, I think we should reduce unnecessary condition variables and mutexes.

src/transaction/log_manager.c
hornetmj: Plus, you can remove the memcpy() if you calculate this before calling the logpb_fetch_page() and pass one of the temp_logbuf to it.

src/transaction/log_manager.c
```diff
@@ -10214,6 +10426,3742 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
...
+	case LOG_ABORT:
+	  LOG_REC_DONETIME * donetime;
+
+	  LOG_READ_ADVANCE_WHEN_DOESNT_FIT (thread_p, sizeof (*donetime), &process_lsa, log_page_p);
+	  if ((process_lsa.pageid % 2) != tmpbuf_index)
+	    {
+	      tmpbuf_index = (tmpbuf_index + 1) % 2;
+	      temp_logbuf[tmpbuf_index].pageid = process_lsa.pageid;
+	      memcpy (temp_logbuf[tmpbuf_index].log_page_p, log_page_p, IO_MAX_PAGE_SIZE);
+	    }
```
hornetmj: How about introducing a new macro that covers this execution similar to the LOG_READ_ADVANCE_WHEN_DOESNT_FIT macro. It is nice if we can remove memcpy function call.

joohok:
code style failure is not from this PR.

PR #2963, opened 2021-08-11
[CBRD-24022] The problem of using reserved words of HQ as data filters in hierarchical queries

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-24022

Exclude predicates with reserved words for HQ from data filter.

### DIFF
--- a/src/parser/xasl_generation.c
+++ b/src/parser/xasl_generation.c
@@ -491,6 +491,8 @@ static int pt_split_hash_attrs (PARSER_CONTEXT * parser, TABLE_INFO * table_info
 static int pt_split_hash_attrs_for_HQ (PARSER_CONTEXT * parser, PT_NODE * pred, PT_NODE ** build_attrs,
 				       PT_NODE ** probe_attrs, PT_NODE ** pred_without_HQ);
 
+static int pt_split_pred_for_HQ (PARSER_CONTEXT * parser, PT_NODE * pred, PT_NODE ** pred_without_HQ, PT_NODE ** pred_with_HQ);
+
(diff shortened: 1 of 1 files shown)


beyondykk9 (approved):

src/parser/xasl_generation.c
```diff
@@ -702,7 +706,13 @@ pt_make_connect_by_proc (PARSER_CONTEXT * parser, PT_NODE * select_node, XASL_NO
...
 
       save_where = select_node->info.query.q.select.where;
-      select_node->info.query.q.select.where = select_node->info.query.q.select.connect_by;
+      where = select_node->info.query.q.select.connect_by;
+
+      /* Separate the predicate related to the HQ. */
+      /* pred_with_HQ is appended to if_pred instead of data filter. */
+      pt_split_pred_for_HQ (parser, where, &pred_without_HQ, &pred_with_HQ);
+
+      select_node->info.query.q.select.where = pred_with_HQ ? pred_without_HQ : where;
```
ctshim, line 715: If pred_with_HQ is not null, pred_without_HQ should be null. 
If so, does it have the same meaning as the code below?
`select_node->info.query.q.select.where = pred_with_HQ ? NULL : where`

shparkcubrid: No. If pred_with_HQ is null, it means there is no change from the existing where.
If pred_with_HQ is not null, pred_without_HQ is null or not null.

ctshim: O, I didn't see the for loop properly in pt_split_pred_for_HQ()
thank you~

ctshim (approved):

youngjinj (approved):

hgryoo (approved):

PR #2964, opened 2021-08-11
[LETS-145] Add single_node configuration

adrianbrab:
http://jira.cubrid.org/browse/LETS-145

Add a new server_type system parameter value named "single_node". It will be recognized by CUBRID processes that both type of servers run on same node.

When single_node is configured, the transaction server always runs on remote_storage and connects to the page server on the same node.

* If server type parameter is single_node, and cub_server is not provided a type argument, throw an error that the server does not have a type.
* If server type parameter is single_node and cub_server is started as transaction, the next parameters will be forced:

    * remote_storage=true

    * page_server_hosts=localhost:master_port

Output examples:

```
(base) adrian@Sparrow:~/cubridScalabil/cubrid$ cub_server scaldb3

This may take a long time depending on the amount of recovery works to do.
Unable to restart/initialize the database server. Invalid server option "Single node server must have type specified as argument".
```
```
(base) adrian@Sparrow:~/cubridScalabil/cubrid$ cub_server -t transaction scaldb3

This may take a long time depending on the amount of recovery works to do.
Unable to restart/initialize the database server. At least one valid page server host is necessary, but no page server could be connected.

```


src/base/server_type.cpp
```diff
@@ -57,15 +57,24 @@ void set_server_type (SERVER_TYPE type)
...
   if (g_server_type == SERVER_TYPE_UNKNOWN)
     {
-      g_server_type = (SERVER_TYPE) prm_get_integer_value (PRM_ID_SERVER_TYPE);
+      if (parameter_value == SERVER_TYPE_SINGLE_NODE)
+	{
+	  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_INVALID_SERVER_OPTION, 1,
+		  "Single node server must have type specified as argument");
+	  return ER_INVALID_SERVER_OPTION;
+	}
+      g_server_type = parameter_value;
```
hornetmj: Currently, SERVER_TYPE_TRANSACTION is the default value. So, I think the following if statement is not necessary.

src/base/server_type.cpp
```diff
@@ -93,6 +102,21 @@ int init_server_type (const char *db_name)
...
 
+void configure_single_node_server (SERVER_TYPE parameter_value)
+{
+  if (g_server_type != SERVER_TYPE_TRANSACTION || parameter_value != SERVER_TYPE_SINGLE_NODE)
+    {
+      return;
+    }
+
+  char *page_hosts_new_value;
+  page_hosts_new_value = (char *) malloc (22 * sizeof (char));
```
hornetmj: Why do you choose the number '22'? Can I ask you a reason?

src/base/server_type.cpp
ribram: A safe-guard should be enough:
```c
assert (g_server_type != SERVER_TYPE_UNKNOWN);
```

src/base/server_type.cpp
```diff
@@ -57,15 +57,24 @@ void set_server_type (SERVER_TYPE type)
...
+	}
+      g_server_type = parameter_value;
       //if no parameter value is provided use transaction as the default type
       if (g_server_type == SERVER_TYPE_UNKNOWN)
 	{
 	  g_server_type = SERVER_TYPE_TRANSACTION;
 	}
     }
+
+  configure_single_node_server (parameter_value);
```
ribram: It would be more clear if the function is renamed and the conditions are moved here:

```suggestion
      //if no parameter value is provided use transaction as the default type
      if (g_server_type == SERVER_TYPE_UNKNOWN)
	{
	  g_server_type = SERVER_TYPE_TRANSACTION;
	}
    }

  if (parameter_value == SERVER_TYPE_SINGLE_NODE && g_server_type == SERVER_TYPE_TRANSACTION)
    {
      setup_tran_server_params_on_single_node_config ();
    }
```

src/base/server_type.cpp
ribram: If a "big enough" size is chosen, go for a power of two (e.g. 32) and give the number literal a name.
```c
  constexpr size_t PAGE_HOSTS_BUFSIZE = 32;
  char *page_hosts_new_value = (char *) malloc (PAGE_HOSTS_BUFSIZE); // will be freed by sysprm_final()
```

For a precise size, snprintf may be used to calculate the size (see [documentation](https://en.cppreference.com/w/c/io/fprintf))
```c
  const char *format = "localhost:%d";
  int portid = prm_get_master_port_id ();
  size_t bufsize = snprintf (NULL, 0, format, portid) + 1;
  char *page_hosts_new_value = (char *) malloc (bufsize); // will be freed by sysprm_final()
```

ribram (commented):
Using SERVER_TYPE enum for SINGLE_NODE configuration is a stretch. Actually, I think we overdid it SERVER_TYPE_ANY too, but that's a separate discussion, not in the scope of this patch.

I think it is better to have two separate enums: SERVER_TYPE (the current one, used for g_server_type and other usages) and server_type_config for the system parameter values.

Make server_type_config an enum class and add TRANSACTION, PAGE and SINGLE_NODE values.
Value conversion:
server_type_config::TRANSACTION -> g_server_type = SERVER_TYPE_TRANSACTION
server_type_config::PAGE -> g_server_type = SERVER_TYPE_PAGE
server_type_config::SINGLE_NODE -> g_server_type must come from cub_server argument.

Laster we'll need to do a SERVER_TYPE enum refactoring.

src/base/server_type.cpp
adrianbrab: changed

src/base/server_type.cpp
adrianbrab: Changed to a fix sized of 32, 16 will probably be enough as the portid is most likely a 4 digit number, but that will cause a  overflow warning from de compiler.

src/base/server_type.cpp
```diff
@@ -38,6 +38,24 @@ SERVER_TYPE get_server_type ()
...
+    {
+    case TRANSACTION:
+      return SERVER_TYPE_TRANSACTION;
+      break;
+    case PAGE:
+      return SERVER_TYPE_PAGE;
+      break;
+    case SINGLE_NODE:
+      return SERVER_TYPE_SINGLE_NODE;
+      break;
```
ribram: single_node should not be expected here.
```suggestion
```

src/base/server_type.cpp
```diff
@@ -57,15 +75,28 @@ void set_server_type (SERVER_TYPE type)
...
+		  "Single node server must have type specified as argument");
+	  return ER_INVALID_SERVER_OPTION;
+	}
+
       //if no parameter value is provided use transaction as the default type
+      g_server_type = get_value_from_config (parameter_value);
       if (g_server_type == SERVER_TYPE_UNKNOWN)
 	{
 	  g_server_type = SERVER_TYPE_TRANSACTION;
 	}
```
ribram: The check is not needed anymore.
```suggestion
```

src/base/server_type_enum.hpp
```diff
@@ -25,6 +25,13 @@ typedef enum
   SERVER_TYPE_TRANSACTION,
   SERVER_TYPE_PAGE,
   SERVER_TYPE_ANY,
+  SERVER_TYPE_SINGLE_NODE,
```
ribram: Single node is not a server type, is just a configuration
```suggestion
```

src/base/server_type_enum.hpp
```diff
@@ -25,6 +25,13 @@ typedef enum
...
 } SERVER_TYPE;
 
+typedef enum
+{
+  TRANSACTION,
+  PAGE,
+  SINGLE_NODE,
+} SERVER_TYPE_CONFIG;
```
ribram: Use enum class to force scoping values and avoid naming conflicts. Values will be accessed as `server_type_config::TRANSACTION` instead of simply `TRANSACTION`.
```suggestion
enum class
{
  TRANSACTION,
  PAGE,
  SINGLE_NODE,
} server_type_config;
```

src/base/server_type.cpp
```diff
@@ -93,6 +124,17 @@ int init_server_type (const char *db_name)
   return er_code;
 }
 
+void setup_tran_server_params_on_single_node_config ()
+{
+  char *page_hosts_new_value;
+  constexpr size_t PAGE_HOSTS_BUFSIZE = 32;
+  page_hosts_new_value = (char *) malloc (PAGE_HOSTS_BUFSIZE);
```
ribram: Explain why the newly allocated memory is not freed in a comment.
```suggestion
  page_hosts_new_value = (char *) malloc (PAGE_HOSTS_BUFSIZE); // free is called by sysprm_final()
```

src/base/server_type.hpp
```diff
@@ -24,7 +24,9 @@
 int init_server_type (const char *db_name);
 void finalize_server_type ();
 SERVER_TYPE get_server_type ();
-void set_server_type (SERVER_TYPE type);
+SERVER_TYPE get_value_from_config (server_type_config parameter_value);
 bool is_tran_server_with_remote_storage ();
+void set_server_type (SERVER_TYPE type);
+void setup_tran_server_params_on_single_node_config ();
```
ribram: get_value_from_config & setup_tran_server_params_on_single_node_config may be declared in server_type.cpp; they don't have to be exposed with the header to other files.

ribram (approved):

PR #2965, opened 2021-08-12
[LETS-176] Core dumped in argument_handler at src/executables/server.c:355

adrianbrab:
http://jira.cubrid.org/browse/LETS-176

Fixed core caused by the `cubrid server start -a` command.
The problem was due to the late initialization of the error manager, the init code was moved to an earlier location.

Tested by running the failing shell tests again, output is now OK.

### DIFF
--- a/src/communication/network.h
+++ b/src/communication/network.h
@@ -280,7 +280,7 @@ typedef enum
 } QUERY_SERVER_REQUEST;
 
 /* Server startup */
-extern int net_server_start (const char *name, THREAD_ENTRY * thread_p);
+extern int net_server_start (THREAD_ENTRY * thread_p, const char *server_name);
 extern const char *net_server_request_name (int request);
 
 #endif /* _NETWORK_H_ */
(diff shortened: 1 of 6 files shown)


src/communication/network.h
```diff
@@ -280,7 +280,7 @@ typedef enum
 } QUERY_SERVER_REQUEST;
 
 /* Server startup */
-extern int net_server_start (const char *name);
+extern int net_server_start (const char *name, THREAD_ENTRY * thread_p);
```
cristiarg: `THREAD_ENTRY`, when it appears, it is usually the first function argument.

cristiarg (approved):

src/executables/server.c
```diff
@@ -46,6 +46,7 @@
 #include "connection_error.h"
 #include "network.h"
 #include "environment_variable.h"
+#include "internal_tasks_worker_pool.hpp"
```
ribram: `cubthread::initialize` function is in `thread_manager.hpp`
```suggestion
#include "thread_manager.hpp"
```

ribram (commented):
Did windows compile pass? If it was not checked, we need to wait for appveyor check too.

ribram (approved):

PR #2967, opened 2021-08-16
[LETS-190] incorrect threadp pointer handling in relaying to calling function

cristiarg:
http://jira.cubrid.org/browse/LETS-190

- pass pointer by reference to surface its value in the calling function 
- incorrect condition for initialization of dependent parameters when single node parameter is read from config

### DIFF
--- a/src/thread/thread_manager.cpp
+++ b/src/thread/thread_manager.cpp
@@ -658,7 +658,7 @@ namespace cubthread
 
 } // namespace cubthread
 
-void thread_initialize_manager (THREAD_ENTRY *thread_p)
+void thread_initialize_manager (THREAD_ENTRY *&thread_p)
 {
   cubthread::initialize (thread_p);
 }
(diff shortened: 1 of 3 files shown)


src/thread/thread_manager.cpp
```diff
@@ -658,7 +658,7 @@ namespace cubthread
 
 } // namespace cubthread
 
-void thread_initialize_manager (THREAD_ENTRY *thread_p)
+void thread_initialize_manager (THREAD_ENTRY *&thread_p)
```
cristiarg, line 661: Is there really a reason for this function to exist? Can't just `cubthread::initialize (thread_p)` be called directly?

ribram: Yes, there is an issue on Windows exporting the function from libcubrid.dll. Usually, these functions are added to the .def file, but namespaces are a problem. This was preferred to alternative solutions.

ribram (approved):

PR #2968, opened 2021-08-16
[LETS-175] Refactoring of redo function arguments

ribram:
http://jira.cubrid.org/browse/LETS-175

There are several redo functions with many arguments, 4-8 or more. Changes in the inner workings of the redo parallelization often have to modify many of these functions by modifying an argument or by adding a new one.

Some of these arguments are either resources or information used for/describing the redo process. These have been grouped into `log_rv_redo_context` structure:

- log reader, and reader fetch mode
- redo & undo zip structure
- end redo LSA

Some of these arguments are information about the log record being redone; these have been grouped into `log_rv_redo_rec_info` structure:

- The log record start LSA
- The record type
- The log record itself as it is saved in the log data.

The new structures are used starting with log_recovery function, and passed down to all other functions involved in redo recovery. The signature of functions using one or more of above resources/information have been updated to use the new structures.

As side effect of the above changes, redo_task, redo_job hierarchy and cublog::replicator classes have been changed.

`reusable_jobs_stack::initialize` have suffered a few changes too. malloc/free is avoided and the jobs array is replaced with a vector. Unnecessary members have been removed. Redo jobs now support copy/move constructors and operator= (required for `vector::resize()`).


src/transaction/log_recovery.c
```diff
@@ -1172,23 +1162,34 @@ log_recovery_redo (THREAD_ENTRY * thread_p, log_recovery_context & context)
...
 	  }
 
+	  /* Read the log record information from log and then if redo is required apply it */
+
+	  const LOG_LSA rcv_lsa = redo_context.m_reader.get_lsa ();	/* Address of redo log record */
+	  /* skip log record header HEADER */
+	  redo_context.m_reader.add_align (sizeof (LOG_RECORD_HEADER));
+
+// *INDENT-OFF*
+#define BUILD_RECORD_INFO(TEMPLATE_TYPE) \
```
ribram, line 1162: I would have liked to use lambda functions instead, but I found no solution.

src/transaction/log_recovery.c
```diff
@@ -1172,23 +1162,34 @@ log_recovery_redo (THREAD_ENTRY * thread_p, log_recovery_context & context)
 	      }
 	  }
 
+	  /* Read the log record information from log and then if redo is required apply it */
+
+	  const LOG_LSA rcv_lsa = redo_context.m_reader.get_lsa ();	/* Address of redo log record */
```
ribram, line 1157: Moved here from each case; please check for mistakes.

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -731,7 +718,7 @@ namespace cublog
     const std::size_t task_count = m_redo_tasks.size ();
     const std::size_t value_count = accum_perf_stat_results.size ();
     std::vector<cubperf::stat_value> avg_perf_stat_results;
-    avg_perf_stat_results.resize (value_count, 0.0);
+    avg_perf_stat_results.resize (value_count, 0);
```
ribram, line 724: A warning fix

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -868,33 +832,28 @@ namespace cublog
     assert (a_push_task_count > 0);
     assert (a_flush_push_at_count > 0);
 
-    m_job_count = a_job_count;
-    m_push_task_count = a_push_task_count;
     m_flush_push_at_count = a_flush_push_at_count;
 
-    m_jobs_arr = static_cast<unsigned char *> (malloc (sizeof (redo_job_impl) * m_job_count));
+    m_job_pool.resize (a_job_count, redo_job_impl (this));
```
ribram, line 840: `a_job_count` copies of `redo_job_impl (this)`

src/transaction/log_recovery.c
cristiarg: I would do away with the macros.
I do not think it justifies the extra cognitive load to just replace a two-liner with a one-liner. Especially since the function call statement has been much reduced.

src/transaction/log_recovery_redo.hpp
```diff
@@ -28,6 +30,29 @@
...
+  log_rv_redo_context () = default;
+  ~log_rv_redo_context ();
+};
+
+template <typename T>
+struct log_rv_redo_rec_info
+{
+  LOG_LSA m_start_lsa = NULL_LSA;
+  LOG_RECTYPE m_type = LOG_SMALLER_LOGREC_TYPE;
+  T m_logrec;
```
cristiarg:
```suggestion
  const LOG_LSA m_start_lsa = NULL_LSA;
  const LOG_RECTYPE m_type = LOG_SMALLER_LOGREC_TYPE;
  const T m_logrec;
```

src/transaction/log_recovery_redo.hpp
```diff
@@ -28,6 +30,29 @@
...
+
+  log_rv_redo_rec_info () = default;
+  log_rv_redo_rec_info (const log_lsa lsa, LOG_RECTYPE type, const T &t);
```
cristiarg, line 60:
```suggestion
  log_rv_redo_rec_info (const log_lsa lsa, LOG_RECTYPE type, const T &t);

  log_rv_redo_rec_info (const log_rv_redo_rec_info &) = delete;
  log_rv_redo_rec_info (log_rv_redo_rec_info &&) = delete;

  log_rv_redo_rec_info &operator = (const log_rv_redo_rec_info &) = delete;
  log_rv_redo_rec_info &operator = (log_rv_redo_rec_info &&) = delete;
```

src/transaction/log_recovery_redo.hpp
```diff
@@ -28,6 +30,29 @@
...
+struct log_rv_redo_context
+{
+  log_reader m_reader { LOG_CS_SAFE_READER };
+  LOG_ZIP m_redo_zip;
+  LOG_ZIP m_undo_zip;
+  LOG_LSA m_end_redo_lsa = NULL_LSA;
+  log_reader::fetch_mode m_reader_fetch_page_mode = log_reader::fetch_mode::FORCE;
+
+  log_rv_redo_context () = default;
+  ~log_rv_redo_context ();
```
cristiarg, line 43: if `LOG_ZIP`'s are de-allocated in dtor, they can be allocated in ctor; their initialization does not seem to be dependent on any local context:
```
  log_zip_realloc_if_needed (redo_context.m_redo_zip, LOGAREA_SIZE);
  log_zip_realloc_if_needed (redo_context.m_undo_zip, LOGAREA_SIZE);
```
I actually think that there should only be one ctor that takes the place of the `log_rv_init_redo_context` function:
```
log_rv_redo_context(const log_recovery_context &);
```
Review everywhere where `log_zip_free_data` and `log_zip_realloc_if_needed` is used on public members `m_redo_zip` and `m_undo_zip`.
Then, some of the members can become `const`:
```
  const LOG_LSA m_end_redo_lsa = NULL_LSA;
  const log_reader::fetch_mode m_reader_fetch_page_mode = log_reader::fetch_mode::FORCE;
```

src/transaction/log_recovery_redo_parallel.cpp
cristiarg: Thanks. The downsides of not using an IDE integrated with compilation.

src/transaction/log_recovery.c
```diff
@@ -76,6 +76,7 @@ static void log_rv_simulate_runtime_worker (THREAD_ENTRY * thread_p, LOG_TDES *
 static void log_rv_end_simulation (THREAD_ENTRY * thread_p);
 static void log_find_unilaterally_largest_undo_lsa (THREAD_ENTRY * thread_p, LOG_LSA & max_undo_lsa);
 static TRANID log_rv_get_min_trantable_tranid ();
+static void log_rv_init_redo_context (const log_recovery_context & context, log_rv_redo_context & redo_context);
```
cristiarg: Remove function.

cristiarg: Maybe rename all 3 context structs for consistency: `log_recovery_context`, `log_rv_redo_context` and `log_rv_redo_rec_info`?

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -925,9 +884,6 @@ namespace cublog
 	    job->~redo_job_impl ();
```
cristiarg: There is no need to explicitly call any of these dtors anymore. They were being called because of the placement ctor call (which, might have allocated extra memory as part of the objects inner workings - not the case in this situation, but added for completitude).

src/transaction/log_replication.cpp
```diff
@@ -104,16 +102,15 @@ namespace cublog
     , m_perfmon_redo_sync { PSTAT_REDO_REPL_LOG_REDO_SYNC }
     , m_perf_stat_idle { cublog::perf_stats::do_not_record_t {} }
   {
-    log_zip_realloc_if_needed (m_undo_unzip, LOGAREA_SIZE);
-    log_zip_realloc_if_needed (m_redo_unzip, LOGAREA_SIZE);
+    log_zip_realloc_if_needed (m_redo_context.m_undo_zip, LOGAREA_SIZE);
```
cristiarg: Remove lines after including function calls in context ctor.

src/transaction/log_recovery_redo.hpp
cristiarg: Also, add following to `log_recovery_context`:
```
    log_recovery_context& operator = (const log_recovery_context &) = delete;
    log_recovery_context& operator = (log_recovery_context &&) = delete;
```

src/transaction/log_recovery.c
```diff
@@ -1050,29 +1057,12 @@ log_recovery_redo (THREAD_ENTRY * thread_p, log_recovery_context & context)
       lsa.offset = NULL_OFFSET;
     }
 
-  undo_unzip_ptr = log_zip_alloc (LOGAREA_SIZE);
-  redo_unzip_ptr = log_zip_alloc (LOGAREA_SIZE);
-
-  if (undo_unzip_ptr == NULL || redo_unzip_ptr == NULL)
```
cristiarg, line 1056: This check can be included in `log_rv_redo_context` ctor.

src/transaction/log_recovery.c
cristiarg: `get_lsa` is not used for `2PC` branches, neither for ABORT/COMMIT.

src/transaction/log_recovery.c
```diff
@@ -1172,23 +1162,34 @@ log_recovery_redo (THREAD_ENTRY * thread_p, log_recovery_context & context)
 	      }
 	  }
 
+	  /* Read the log record information from log and then if redo is required apply it */
+
+	  const LOG_LSA rcv_lsa = redo_context.m_reader.get_lsa ();	/* Address of redo log record */
+	  /* skip log record header HEADER */
+	  redo_context.m_reader.add_align (sizeof (LOG_RECORD_HEADER));
```
cristiarg, line 1159: `add_align` was used conditionally for COMMIT/ABORT branch. Is moving this above the conditional scope safe?

src/transaction/log_recovery.c
ribram: You mean by renaming `log_recovery_context` as `log_rv_context`?

src/transaction/log_recovery.c
ribram: What check? Do you mean adding log_zip_alloc's in the ctor? Done that part.

src/transaction/log_recovery.c
ribram: True, but it does not matter much. It simplifies the code in most cases; for the remaining cases it may do a few extra operations that will not have a sensible impact.

src/transaction/log_recovery.c
ribram: It is safe. After completing the COMMIT/ABORT branch, the log cursor would be advanced to the next log record.

src/transaction/log_recovery.c
ribram: The purpose of the macros is to avoid so much repeated code & indent-off/on blocks. I would keep them :D

src/transaction/log_recovery_redo.hpp
ribram: Done. Added two constructor: one with end redo LSA and log page fetch mode; and a copy constructor.

src/transaction/log_recovery_redo.hpp
ribram: done

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -521,28 +517,21 @@ namespace cublog
 
   constexpr unsigned short redo_parallel::redo_task::WAIT_AND_CHECK_MILLIS;
 
-  redo_parallel::redo_task::redo_task (std::size_t a_task_idx
-				       , redo_parallel::task_active_state_bookkeeping &task_state_bookkeeping
-				       , redo_parallel::redo_job_queue &a_queue)
+  redo_parallel::redo_task::redo_task (std::size_t a_task_idx,
+				       redo_parallel::task_active_state_bookkeeping &task_state_bookkeeping,
+				       redo_parallel::redo_job_queue &a_queue,
+				       const log_rv_redo_context &copy_context)
```
ribram, line 523: Also added `const log_rv_redo_context &copy_context`. Task's redo context has to be similar to the main thread's redo context.

src/transaction/log_replication.cpp
ribram: Done

src/transaction/log_recovery_redo_parallel.cpp
ribram: done

src/transaction/log_recovery.c
cristiarg: I mean't the check for failing to allocate the memory for them, the part under:
```
if (undo_unzip_ptr == NULL || redo_unzip_ptr == NULL
```

src/transaction/log_recovery.c
cristiarg: Either change `.._recovery_..` into `.._rv_..` or the other way around, as you see fit.

src/transaction/log_recovery.c
```diff
@@ -1003,8 +993,9 @@ log_recovery_redo (THREAD_ENTRY * thread_p, log_recovery_context & context)
 {
   LOG_LSA lsa;			/* LSA of log record to redo */
 
-  log_rv_redo_context redo_context;
-  log_rv_init_redo_context (context, redo_context);
+  log_rv_redo_context redo_context
+  {
+  context.get_end_redo_lsa (), log_reader::fetch_mode::NORMAL};
```
cristiarg: Alignment. I think the indenting tools settle if they are served:
```
TYPE variable
{
  initalization ....
};
```

cristiarg (approved):

PR #2969, opened 2021-08-17
[LETS-191] crash when recovery is executed in parallel mode without any extended perfmon activation flags enabled

cristiarg:
https://jira.cubrid.org/browse/LETS-191

Guard execution of perf stats reporting function with the same function used to activated perf stats logging in the first place.

### DIFF
--- a/src/transaction/log_recovery_redo_parallel.cpp
+++ b/src/transaction/log_recovery_redo_parallel.cpp
@@ -713,32 +713,35 @@ namespace cublog
   void
   redo_parallel::log_perf_stats () const
   {
-    const cubperf::statset_definition definition
-    {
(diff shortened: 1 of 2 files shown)


src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -713,32 +713,35 @@ namespace cublog
   void
   redo_parallel::log_perf_stats () const
   {
-    const cubperf::statset_definition definition
-    {
-      perf_stats_async_definition_init_list
-    };
+    if ( perf_stats_is_active_for_async () )
```
ribram: it is easier to do an early out if `!perf_stats_is_active_for_async ()`

ribram (approved):

PR #2971, opened 2021-08-18
[CBRD-23622] [Regression][shell] Able to skip TRUNCATE according to the optimization level

wrlawodms:
http://jira.cubrid.org/browse/CBRD-23622

Basically, statements can be skipped according to the optimization level like opt-level 2. TRUNCATE was skipped during executing DELETE, but now it may not use DELETE. DDLs are skipped in  do_statement() and do_execute_statement(), but by now TRUNCATE was not regarded as DDL. So, we need to deal with it independently.

Now, we have decided to consider TRUNCATE as DDL so that we can just add PT_TRUNCATE to `pt_is_ddl_statement()`, but I'm not sure it is safe because I don't really know about `sysprm_print_parameters_for_ha_repl()` which is executed if `pt_is_ddl_statement()` is true in `do_replicate_statement()`. I checked others and make sure I can easily modify them safely. So, I want to have your opinion about it, @hornetmj.

With this change, TRUNCATE can be blocked with the system parameter `block_ddl_statement` as a side effect, but I think it's OK and even better if we regard TRUNCATE as a DDL as we discussed.

### DIFF
--- a/src/query/execute_statement.c
+++ b/src/query/execute_statement.c
@@ -2956,7 +2956,7 @@ do_statement (PARSER_CONTEXT * parser, PT_NODE * statement)
   if (statement)
     {
       /* skip ddl execution in case of parameter or opt. level */
-      if (pt_is_ddl_statement (statement) != 0)
+      if (pt_is_ddl_statement (statement) != 0 || statement->node_type == PT_TRUNCATE)
 	{
 	  if (prm_get_bool_value (PRM_ID_BLOCK_DDL_STATEMENT))
 	    {
@@ -3443,7 +3443,7 @@ do_execute_statement (PARSER_CONTEXT * parser, PT_NODE * statement)
   SET_HOST_VARIABLES_IF_INTERNAL_STATEMENT (parser);
 
   /* skip ddl execution in case of parameter or opt. level */
-  if (pt_is_ddl_statement (statement) != 0)
+  if (pt_is_ddl_statement (statement) != 0 || statement->node_type == PT_TRUNCATE)
     {
       if (prm_get_bool_value (PRM_ID_BLOCK_DDL_STATEMENT))
 	{
--- a/src/query/execute_statement.c
+++ b/src/query/execute_statement.c
@@ -368,7 +368,6 @@ is_stmt_based_repl_type (const PT_NODE * node)
   switch (node->node_type)
     {
     case PT_DROP_VARIABLE:
-    case PT_TRUNCATE:
       return true;
     case PT_INSERT:
       if (node->info.insert.hint & PT_HINT_USE_SBR)
@@ -2956,7 +2955,7 @@ do_statement (PARSER_CONTEXT * parser, PT_NODE * statement)
   if (statement)
     {
       /* skip ddl execution in case of parameter or opt. level */
-      if (pt_is_ddl_statement (statement) != 0 || statement->node_type == PT_TRUNCATE)
+      if (pt_is_ddl_statement (statement) != 0)
 	{
 	  if (prm_get_bool_value (PRM_ID_BLOCK_DDL_STATEMENT))
 	    {
@@ -3443,7 +3442,7 @@ do_execute_statement (PARSER_CONTEXT * parser, PT_NODE * statement)
   SET_HOST_VARIABLES_IF_INTERNAL_STATEMENT (parser);
 
   /* skip ddl execution in case of parameter or opt. level */
-  if (pt_is_ddl_statement (statement) != 0 || statement->node_type == PT_TRUNCATE)
+  if (pt_is_ddl_statement (statement) != 0)
     {
       if (prm_get_bool_value (PRM_ID_BLOCK_DDL_STATEMENT))
 	{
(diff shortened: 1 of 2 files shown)


src/query/execute_statement.c
```diff
@@ -2956,7 +2956,7 @@ do_statement (PARSER_CONTEXT * parser, PT_NODE * statement)
   if (statement)
     {
       /* skip ddl execution in case of parameter or opt. level */
-      if (pt_is_ddl_statement (statement) != 0)
+      if (pt_is_ddl_statement (statement) != 0 || statement->node_type == PT_TRUNCATE)
```
hornetmj: How about handling the 'statement->node_type == PT_TRUNCATE' into the pt_is_ddl_statement function?

wrlawodms: > Now, we have decided to consider TRUNCATE as DDL so that we can just add PT_TRUNCATE to `pt_is_ddl_statement()`, but I'm not sure it is safe because I don't really know about `sysprm_print_parameters_for_ha_repl()` which is executed if `pt_is_ddl_statement()` is true in `do_replicate_statement()`. I checked others and make sure I can easily modify them safely. So, I want to have your opinion about it, @hornetmj.

That's where I asked for your opinion. Regarding `sysprm_print_parameters_for_ha_repl()`, I want to make sure it's safe. Do you think it is?

hornetmj: The sysprm_print_parameters_for_ha_repl() function makes a string of HA related system parameters that should be known to the slave node(applylogdb). For example, if the master node sets to create_table_reuseoid=no and the slave node sets it to yes, then the create table statement could make a different reuse_oid type table in the slave. Please see the item->ha_sys_prm(equals to repl_stmt.sys_prm_context) and db_set_system_parameters_for_ha_repl() in the la_apply_statement_log().

This is my rough feedback for your request. I will check it more deeply tomorrow. I want to go home now.

hornetmj: You can handle it in the pt_is_ddl_statement function. And following tests show the truth related to the sysprm_print_parameters_for_ha_repl function.

1. Master: create_table_reuseoid=no, Slave: default (create_table_reuseoid=yes)
```
# Master
csql> ;get create_table_reuseoid

=== Get Param Input ===

create_table_reuseoid=n

csql> create table test (c1 int);
csql> select class_name, is_reuse_oid_class from db_class where class_name = 'test';

  class_name            is_reuse_oid_class
  'test'                'NO'

# Slave
csql> ;get create_table_reuseoid

=== Get Param Input ===

create_table_reuseoid=y
csql> select class_name, is_reuse_oid_class from db_class where class_name = 'test';

  class_name            is_reuse_oid_class
  'test'                'NO' 
```
2. Master: create_table_reuseoid=yes in the csql(no -> yes only session), Slave: default (create_table_reuseoid=yes)
```
# Master
csql> ;set create_table_reuseoid=yes

=== Set Param Input ===

create_table_reuseoid=yes

csql> ;get create_table_reuseoid

=== Get Param Input ===

create_table_reuseoid=y

csql> create table test2 (c1 int);
Execute OK. (0.007183 sec) Committed.

csql> select class_name, is_reuse_oid_class from db_class where class_name = 'test2';

  class_name            is_reuse_oid_class
  'test2'               'YES'

# Slave
csql> ;get create_table_reuseoid

=== Get Param Input ===

create_table_reuseoid=y
csql> select class_name, is_reuse_oid_class from db_class where class_name = 'test2';

  class_name            is_reuse_oid_class
  'test2'               'YES'
```
3. Master: default (create_table_reuseoid=yes), Slave: create_table_reuseoid=no
```
# Master
csql> ;get create_table_reuseoid

=== Get Param Input ===

create_table_reuseoid=y
csql> create table test3 (c1 int);
Execute OK. (0.007556 sec) Committed.

csql> select class_name, is_reuse_oid_class from db_class where class_name = 'test3';

  class_name            is_reuse_oid_class
  'test3'               'YES'

# Slave
csql> ;get create_table_reuseoid

=== Get Param Input ===

create_table_reuseoid=n

csql> select class_name, is_reuse_oid_class from db_class where class_name = 'test3';

  class_name            is_reuse_oid_class
  'test3'               'NO'
```

The result of 3rd is not expected. But only the changed system parameter value compared to the default value is sent to the slave node for the applylogdb by the sysprm_print_parameters_for_ha_repl function.

wrlawodms: Thanks to your explanation in detail, I understand how it works. In summary, you mean if the value of system parameters with `PRM_FOR_HA_CONTEXT` is different from the default value, it is copied and applied to slave nodes.

I looked into system parameters with `PRM_FOR_HA_CONTEXT`:
1. default_week_format
2. string_max_size_bytes
3. add_column_update_hard_default
4. return_null_on_function_errors
5. alter_table_change_type_strict
6. compat_numeric_division_scale
7. compat_primary_key
8. intl_number_lang
9. intl_date_lang
10. intl_collation
11. timezone
12. allow_truncated_string
13. create_table_reuseoid
 
And those don't seem to affect the behavior of TRUNCATE operation. So, I think we could put it into `pt_is_ddl_statement()`. What do you think?

hornetmj: @wrlawodms I agree with you.

hornetmj (approved):
Please merging this PR after reflecting on the comment.

eido5 (approved):

joohok (approved):

PR #2972, opened 2021-08-18
[LETS-133] Do not mount data volumes on the transaction server with remote storage

ribram:
http://jira.cubrid.org/browse/LETS-133

**Do not mount any data volumes** during the boot of a transaction server with remote storage. Disable `boot_mount()` calls on TSRS.

As consequence, a **new system to boot disk cache** is required. The volids used to be obtained from the mounted file cache. The existing volids have to be obtained from the page server instead:

  - Add ATS<->PS requests to transfer boot information. For now, the only information passed from PS is the permanent volume count.
  - On `init_server_type()` call `ats_Gl.boot()` (initializes page server connections and also obtains boot information).
  - `active_tran_server::get_boot_info_from_page_server()` sends the request to get boot information from page server and then blocks until the information is obtained (with the help of mutex, condition variable and `m_is_boot_info_received` boolean).
  
When disk cache is loaded, map `disk_cache_load_volume()` on all volids in the range [0, permanent volume count) instead of mapping on fileio cache volumes.

**Other changes:**

  - Update pgbuf_is_valid_page(); getting volume label on TSRS is not possible.
  - Fix missing PGBUF_BCB_FLUSH_NOT_NEEDED_FLAG when fetch_mode is NEW_PAGE.

### DIFF
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -18,6 +18,7 @@
 
 #include "page_server.hpp"
 
+#include "disk_manager.h"
 #include "error_manager.h"
 #include "log_impl.h"
 #include "log_lsa_utils.hpp"
@@ -141,6 +142,17 @@ void page_server::receive_disconnect_request (cubpacking::unpacker &upk)
   //start a thread to destroy the ATS to PS connection object
   std::thread disconnect_thread (&page_server::disconnect_active_tran_server, std::ref (*this));
   disconnect_thread.detach ();
+}
+
+void page_server::receive_boot_info_request (cubpacking::unpacker &upk)
+{
+  DKNVOLS nvols_perm = disk_get_perm_volume_count ();
+
+  std::string response_message;
+  response_message.reserve (sizeof (nvols_perm));
+  response_message.append (reinterpret_cast<const char *> (&nvols_perm), sizeof (nvols_perm));
+
+  m_active_tran_server_conn->push (ps_to_ats_request::SEND_BOOT_INFO, std::move (response_message));
 }
 
 void page_server::push_request_to_active_tran_server (ps_to_ats_request reqid, std::string &&payload)
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -56,6 +56,10 @@ void page_server::set_active_tran_server_connection (cubcomm::channel &&chn)
   m_active_tran_server_conn.reset (new active_tran_server_conn_t (std::move (chn),
   {
     {
+      ats_to_ps_request::GET_BOOT_INFO,
+      std::bind (&page_server::receive_boot_info_request, std::ref (*this), std::placeholders::_1)
+    },
+    {
       ats_to_ps_request::SEND_LOG_PRIOR_LIST,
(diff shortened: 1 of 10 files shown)


src/storage/disk_manager.c
```diff
@@ -6786,6 +6804,23 @@ disk_sectors_to_extend_npages (const int num_pages)
...
+}
+
+void
+disk_set_page_server_perm_volume_count (DKNVOLS nvols)
+{
+  assert (is_tran_server_with_remote_storage ());
+
+  // Set the number of permanent volumes for transaction server with remote storage.
+  // Disk manager is not initialized yet, so save the number to be used later when disk cache is loaded.
+  disk_Page_server_perm_volume_count = nvols;
```
cristiarg, line 6822: assert that `disk_Page_server_perm_volume_count` is 0 at this point?

src/server/page_server.cpp
```diff
@@ -143,6 +148,17 @@ void page_server::receive_disconnect_request (cubpacking::unpacker &upk)
   disconnect_thread.detach ();
 }
 
+void page_server::receive_boot_info_request (cubpacking::unpacker &upk)
```
cristiarg, line 151: it would have been better to maintain a certain naming convention for these functions; if the request is `ats_to_ps_request::GET_BOOT_INFO`, then the handler should have something derived from the name of the constant (ie: `on_get_goot_info` or `on_ats_to_ps_request_get_boot_info`);

src/storage/page_buffer.c
```diff
@@ -10441,8 +10452,9 @@ pgbuf_is_valid_page (THREAD_ENTRY * thread_p, const VPID * vpid, bool no_error,
     {
       if (valid != DISK_ERROR && !no_error)
 	{
+	  const char *vlabel = fileio_get_volume_label (vpid->volid, PEEK);
```
cristiarg, line 10455: shouldn't this forking be done inside `fileio_get_volume_label`? there are many calls (both direct and indirect) to this function

src/storage/disk_manager.c
ribram: done

src/server/page_server.cpp
ribram: the new functions follow the existing naming convention. maybe suggested naming convention is better, but it must be done in a separate refactoring task, because other functions should be renamed too.

src/storage/page_buffer.c
ribram: it cannot be forked inside because fileio_get_volume_label may be called on temporary volumes too. if called on permanent volumes, it will cause errors or (debug) crashes and that's fine.

cristiarg: But is it correct to say that, on a TSRS, there are potentially more places where the call would return NULL? In total there seem to be ~40 calls to this function.
If yes, shouldn't those be amended as well?

ribram: You are right, some of those 40 calls should be amended as well. It requires careful examination, so I think it may be handled as a separate issue and pull request. What do you think?

cristiarg (approved):
A new issue needs to be added for the investigation of the function `fileio_get_volume_label` when running in TSRS (transaction server - remote storage) context.

PR #2974, opened 2021-08-19
[CBRD-24025] Speedup redo recovery by avoiding "is reserved" check

cristiarg:
https://jira.cubrid.org/browse/CBRD-24025

During recovery, there is a step that checks if page is "reserved". The check fixes sector reservation table page and searches for the bit corresponding to the page's sector. If it is not reserved, applying redo is skipped. As a result, if a file is destroyed before crash (drop index, drop table), many redo changes may be skipped. Skipping these redo record is not a strict requirement; it may have been an optimization and it may pass some automatic page buffer fix checks.

However, the performance impact of the check is serious. Measurements in some artificial tests showed an overall 2x improvement of the recovery redo part of the recovery sequence without the checks.

The check is completely avoidable by applying redo on pages regardless of its reserved status. In the worst case scenario, some redo changes may be applied unnecessarily before the page is unreserved. Considering the impact of the check though, it is likely to get better performance by applying the redo without doing any checks instead of doing the check and skipping the redo.

Implementation:
- add a new page fetch mode, for recovery, that accepts any page status: new page, de-allocated page, or normal page.
- use new page fetch mode during redo recovery/replication
- no "is page sector reserved" check (update log_rv_redo_page_fix)

### DIFF
--- a/src/storage/page_buffer.c
+++ b/src/storage/page_buffer.c
@@ -4803,7 +4803,6 @@ pgbuf_set_lsa_as_temporary (THREAD_ENTRY * thread_p, PAGE_PTR pgptr)
  *   return: void
  *   bufptr(in): pointer to buffer page
  *
- * Note: This function is used for debugging.
  */
 STATIC_INLINE void
 pgbuf_set_bcb_page_vpid (PGBUF_BCB * bufptr)
@@ -4836,11 +4835,6 @@ pgbuf_set_bcb_page_vpid (PGBUF_BCB * bufptr)
 	  /* values not reset upon page deallocation */
 	  assert (bufptr->iopage_buffer->iopage.prv.volid == bufptr->vpid.volid);
 	  assert (bufptr->iopage_buffer->iopage.prv.pageid == bufptr->vpid.pageid);
-
-	  /* values reset upon page deallocation */
-	  assert (bufptr->iopage_buffer->iopage.prv.ptype == PAGE_UNKNOWN);
-	  /* only compare the encryption part of the flag, the rest is, as of now, unused */
-	  assert ((bufptr->iopage_buffer->iopage.prv.pflag & FILEIO_PAGE_FLAG_ENCRYPTED_MASK) == '\0');
 	}
     }
 }
--- a/src/storage/page_buffer.c
+++ b/src/storage/page_buffer.c
@@ -4831,6 +4831,17 @@ pgbuf_set_bcb_page_vpid (PGBUF_BCB * bufptr)
 	  bufptr->iopage_buffer->iopage.prv.p_reserve_2 = 0;
 	  bufptr->iopage_buffer->iopage.prv.tde_nonce = 0;
 	}
+      else
+	{
+	  /* values not reset upon page deallocation */
+	  assert (bufptr->iopage_buffer->iopage.prv.volid == bufptr->vpid.volid);
+	  assert (bufptr->iopage_buffer->iopage.prv.pageid == bufptr->vpid.pageid);
+
+	  /* values reset upon page deallocation */
+	  assert (bufptr->iopage_buffer->iopage.prv.ptype == PAGE_UNKNOWN);
+	  /* only compare the encryption part of the flag, the rest is, as of now, unused */
+	  assert ((bufptr->iopage_buffer->iopage.prv.pflag & FILEIO_PAGE_FLAG_ENCRYPTED_MASK) == '\0');
+	}
     }
 }
 
@@ -5008,7 +5019,7 @@ pgbuf_initialize_bcb_table (void)
       ioptr->iopage.prv.pageid = -1;
       ioptr->iopage.prv.volid = -1;
 
-      ioptr->iopage.prv.ptype = '\0';
+      ioptr->iopage.prv.ptype = (unsigned char) PAGE_UNKNOWN;
       ioptr->iopage.prv.pflag = '\0';
       ioptr->iopage.prv.p_reserve_1 = 0;
       ioptr->iopage.prv.p_reserve_2 = 0;
@@ -10461,7 +10472,7 @@ pgbuf_scramble (FILEIO_PAGE * iopage)
   iopage->prv.pageid = -1;
   iopage->prv.volid = -1;
 
-  iopage->prv.ptype = '\0';
+  iopage->prv.ptype = (unsigned char) PAGE_UNKNOWN;
   iopage->prv.pflag = '\0';
   iopage->prv.p_reserve_1 = 0;
   iopage->prv.p_reserve_2 = 0;
@@ -14012,7 +14023,7 @@ pgbuf_dealloc_page (THREAD_ENTRY * thread_p, PAGE_PTR page_dealloc)
 #endif /* !NDEBUG */
 
   /* set unknown type */
-  bcb->iopage_buffer->iopage.prv.ptype = (char) PAGE_UNKNOWN;
+  bcb->iopage_buffer->iopage.prv.ptype = (unsigned char) PAGE_UNKNOWN;
   /* clear page flags (now only tde algorithm) */
   bcb->iopage_buffer->iopage.prv.pflag = (unsigned char) 0;
 
--- a/src/storage/page_buffer.c
+++ b/src/storage/page_buffer.c
@@ -1077,7 +1077,7 @@ STATIC_INLINE void pgbuf_wakeup_reader_writer (THREAD_ENTRY * thread_p, PGBUF_BC
 
 STATIC_INLINE bool pgbuf_get_check_page_validation_level (int page_validation_level) __attribute__ ((ALWAYS_INLINE));
 static bool pgbuf_is_valid_page_ptr (const PAGE_PTR pgptr);
-STATIC_INLINE void pgbuf_set_bcb_page_vpid (PGBUF_BCB * bufptr, bool force_set_vpid) __attribute__ ((ALWAYS_INLINE));
+STATIC_INLINE void pgbuf_set_bcb_page_vpid (PGBUF_BCB * bufptr) __attribute__ ((ALWAYS_INLINE));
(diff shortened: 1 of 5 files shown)


ribram (approved):

src/storage/page_buffer.c
```diff
@@ -1950,8 +1950,7 @@ pgbuf_fix_release (THREAD_ENTRY * thread_p, const VPID * vpid, PAGE_FETCH_MODE f
 
   /* Set Page identifier if needed */
   // Redo recovery may find an immature page which should be set.
-  force_set_vpid = (fetch_mode == NEW_PAGE && log_is_in_crash_recovery_and_not_yet_completes_redo ());
```
wrlawodms, line 1953: During recovery, if the page was fixed with a recovery index of `RCV_IS_NEW_PAGE_INIT`, `iopage.prv` was initialized forcefully. Could you tell us why now we can skip it?

src/transaction/log_recovery.c
```diff
@@ -6389,85 +6389,32 @@ log_rv_pack_undo_record_changes (char *ptr, int offset_to_data, int old_data_siz
...
+  // sector reservation table are applied in parallel with the changes in pages, at times the page may appear to be
+  // deallocated (part of an unreserved sector). but the changes were done while the sector was reserved and must be
+  // re-applied to get a correct end result.
+  // 
+  // moreover, the sector reservation check is very expensive. running this check on every page fix costs much more
+  // than any time gained by skipping redoing changes on deallocated pages.
+  //
+  // therefore, fix page using RECOVERY_PAGE mode. pgbuf_fix will know to accept even new or deallocated pages.
+  //
+  page = pgbuf_fix (thread_p, vpid_rcv, RECOVERY_PAGE, PGBUF_LATCH_WRITE, PGBUF_UNCONDITIONAL_LATCH);
```
wrlawodms, line 6406: In `pgbuf_claim_bcb_for_fix()` called by `pgbuf_fix()`, if the `fetch_mode` is not `NEW_PAGE`, it tries to read the page from a disk volume. From what I've understood, It means that it would read a data page from a volume even though the recovery index is `RCV_IS_NEW_PAGE_INIT` here. It seems that it is unacceptable. Could I have your opinion about it?

src/storage/page_buffer.c
cristiarg: There are two situations that a page can find itself during recovery:
- either it is allocated for the first time, in which case:
  - the initialization of `prv` in `pgbuf_set_bcb_page_vpid` is performed
- or it is allocated after it has been previously de-allocated, in which case:
  - the initialization of `prv` in `pgbuf_set_bcb_page_vpid` is skipped, because:
    - `pageid` and `volid` are persistent, once initialized the first time,
      nothing will change them and the value is invariant regardless of how
      many times they are [un/re]allocated
    - `ptype` would have been reset upon the previous deallocation and should
      already be `PAGE_UNKNOWN`
    - `tde_nonce` is written unconditionally by tde routines in
      `tde_encrypt_data_page` and `tde_decrypt_data_page`

Therefore, the [re]initialization of `iopage.prv` is not needed.
At the very least, the change could be to add an else branch with some asserts:
```
  if (bufptr->iopage_buffer->iopage.prv.pageid == NULL_PAGEID && bufptr->iopage_buffer->iopage.prv.volid == NULL_VOLID)
    {
      // init as modified
    }
  else
    {
      // - assert that voldid pageid are the same as those in bufptr
      // - assert that ptype is PAGE_UNKNOWN
      // - no assertion about tde_nonce, unless one can be made in conjunction with another state
      //    somewhere else regarding tde state of the page
    }
```

src/transaction/log_recovery.c
cristiarg: It should read the page from the volume (disk) regardless of recovery index.
Upon reading the page from disk, there are subsequent checks (ie: in `log_rv_fix_page_and_check_redo_is_needed`) to decide whether it needs to apply the log record or not, based on the LSA of the page and the LSA of the log record.

Log record with recovery indexes in `RCV_IS_NEW_PAGE_INIT` have no guarantee that the state of the page on disk is de-allocated. But the LSA check covers these log records too:
- If the page is never allocated, it has null LSA;
- If the page is de-allocated, its LSA must be smaller than the log record's; as a result, the record is redone
- If the page is not de-allocated, then its LSA must be equal or higher and the log record must not be redone

Note that previous code also read the page from disk by calling `pgbuf_fix_if_not_deallocated` first. If the page was not allocated, the pgptr remained NULL and pgbuf_fix was called again with `NEW_PAGE` flag. Although the approach was different, the end result was the same.

src/transaction/log_recovery.c
```diff
@@ -6389,85 +6389,32 @@ log_rv_pack_undo_record_changes (char *ptr, int offset_to_data, int old_data_siz
...
+  if (page == NULL)
     {
-      ASSERT_ERROR ();
+      // this is terrible, because it makes recovery impossible
+      assert_release (false);
       return NULL;
     }
-
-  if (page == NULL && RCV_IS_NEW_PAGE_INIT (rcvindex))
```
cristiarg, line 6430: this was the single use of `RCV_IS_NEW_PAGE_INIT` and the macro can be removed

src/storage/page_buffer.c
wrlawodms: Thank you for clarifying and explaining in detail. It is very helpful. I also tried to find out the case it was needed, but I couldn't. 
Just so I'm clear, you mean that there is no case in which VPID of `prv` is not NULL, but `ptype` and `nonce` have to be initialized. And this is true even without this "is reserved check" patch, but you remove it during dealing with this patch.

wrlawodms: And I think the assertion you suggested seems good for safety because we are removing a case.

src/transaction/log_recovery.c
wrlawodms: Oh, I missed it reads a data page even with recovery indexes in `RCV_IS_NEW_PAGE_INIT` through `pgbuf_fix_if_not_deallocated()`. So, no additional I/O read. Thank you.

I think the code is correct, but have a little different thought about it. So, to understand correctly, I'd love it if you point out what I misunderstand.

> * If the page is de-allocated, its LSA must be smaller than the log record's; as a result, the record is redone

It seems that the LSA of a page could be larger than log recods even if the page is de-allocated. Imagine operations: (1) allocate - (2) deallocate - (3) allocate - (4) deallocate. If the redo recovery meets the log record of (1), but the LSA of the page is (4), it can be larger.

> * If the page is not de-allocated, then its LSA must be equal or higher and the log record must not be redone

As same above, the LSA could be smaller.

The way I see it, we can skip reading a data page and apply log records if its recovery index is in `RCV_IS_NEW_PAGE_INIT`. Even if the LSA of a data page was larger than the LSA of the log record, the subsequent redo operations will make it catch up on the latest state again. However, we can skip this redundant redo with your way. 

Please let me have your opinion.

cristiarg: I think you are right that the examples given above in the second and third bullets are incomplete (or, maybe, plain wrong).
I do agree with your remark that a page can find itself in any position before, in-between or after with respect to the states (1)-(2)-(3)-(4) described above.
However, the verification:
```
		  if (LSA_LE (&rcv_lsa, rcv_page_lsaptr))
		    {
		      /* It is already done */
		      pgbuf_unfix (thread_p, rcv.pgptr);
		      break;
		    }
```
which is done after each call to `log_rv_redo_fix_page` (and which, on the branch `scalability_dev` from where we back-ported this change, is abstracted away in an explicit function call - `log_rv_fix_page_and_check_redo_is_needed` at https://github.com/CUBRID/cubrid/blob/85101805863dfc12471e97632f2f944177c4830d/src/transaction/log_recovery.c#L430) covers all the cases that you describe.
Another check to avoid applying the log record would only be justified if it would improve performance even further, which, at this point is unknown.

src/storage/page_buffer.c
cristiarg: I added the change on the else branch with asserts for `volid` and `pageid` and performed some tests.
As for checking `ptype` a higher level context info is needed and it is already done - see code after the call to `pgbuf_set_bcb_page_vpid` in caller functions `pgbuf_set_page_ptype` and `pgbuf_fix`.

src/transaction/log_recovery.c
wrlawodms: I agree that we don't need the additional check necessarily. I just wanted to check if I understand correctly. Thank you.

src/transaction/log_recovery.c
```diff
@@ -3297,7 +3297,7 @@ log_recovery_redo (THREAD_ENTRY * thread_p, const LOG_LSA * start_redolsa, const
 	      /* If the page does not exit, there is nothing to redo */
 	      if (rcv_vpid.pageid != NULL_PAGEID && rcv_vpid.volid != NULL_VOLID)
 		{
-		  rcv.pgptr = log_rv_redo_fix_page (thread_p, &rcv_vpid, rcvindex);
+		  rcv.pgptr = log_rv_redo_fix_page (thread_p, &rcv_vpid);
 		  if (rcv.pgptr == NULL)
 		    {
 		      /* deallocated */
```
```suggestion
```

wrlawodms (approved):
It would be better to get rid of comments `/* deallocated */` after calling `log_rv_redo_fix_page()` because we do not longer return NULL now when it is deallocated.
I appreciate your contribution and explanation.

src/transaction/log_recovery.c
cristiarg: Done

cristiarg:
Executed shell_debug after 7db0332f7c4cddfe836a53bcfd8a6434b26b5d4c without any cores.

wrlawodms (approved):

hornetmj (approved):

mhoh3963 (approved):

PR #2975, opened 2021-08-19
[CBRD-24026] Speedup redo recovery by reducing time spent on logtb_find_tran_index

cristiarg:
https://jira.cubrid.org/browse/CBRD-24026

Investigation, phase 1:
- during `log_recovery_redo` it was found (via perf) that `logtb_find_tran_index` take a lot of time because it does a sequential search to find the index of a trid across the entirety of the transactions found in the log
- only to, afterwards, free that transaction index.

Investigation, phase 2:
- this freeing of a transaction index in `log_recovery_redo` looked suspect
- subsequent investigation revealed that the checkpoint mechanism considers, at the point it is executed all transactions that it finds in the transaction table regardless of their state
- thus, the checkpoint ends up picking transactions which are actually committed (ie: the commit/abort log record has been added) but have not yet been cleaned-up from the transaction table (note that threre might be a race condition in this area) and adds them to the `LOG_END_CHKPT` log entry that it adds
- at recovery analysis, the following log entries for a given transaction are found (an example): `LOG_MVCC_UNDOREDO_DATA`, `LOG_MVCC_DIFF_UNDOREDO_DATA`, `LOG_COMMIT`, `LOG_END_CHKPT`
- the last but one of these will clear the transaction from the transaction table
- the last one will actually re-add an entry in the transaction table for the same transaction (via the succession of calls: `log_recovery_analysis` - `log_rv_analysis_record` - `log_rv_analysis_end_checkpoint` - `logtb_rv_find_allocate_tran_index`)
- this entry is the one that was being cleared in `log_recovery_redo` on the `LOG_COMMIT/LOG_ABORT` switch branch.

Implementation:
- make the checkpoint mechanism not pick up transactions which have had their commit/abort log record entry added
- a new `log_tdes.commit_abort_lsa` flag is added, initialized to `NULL_LSA` and set to the LSA of the commit/abort log record once that is added - this is done under `log_prior` lock in `prior_lsa_next_record_internal`
- make the checkpoint mechanism not pick up transactions which have had their commit/abort added, based on the flag - in `logpb_checkpoint_trans`
- in `log_recovery_redo` only assert that there are no `LOG_COMMIT/ABORT` log records, unless required to `stop_at` a certain time (restore)


ribram (approved):

src/transaction/log_recovery.c
```diff
@@ -3910,7 +3938,11 @@ log_recovery_redo (THREAD_ENTRY * thread_p, const LOG_LSA * start_redolsa, const
 	      {
 		bool free_tran = false;
 
-		tran_index = logtb_find_tran_index (thread_p, tran_id);
+		tran_index = NULL_TRANID;
```
```suggestion
		tran_index = NULL_TRAN_INDEX;
```

src/transaction/log_recovery.c
```diff
@@ -3910,7 +3938,11 @@ log_recovery_redo (THREAD_ENTRY * thread_p, const LOG_LSA * start_redolsa, const
...
+		if (min_trantable_tranid != NULL_TRANID && tran_id >= min_trantable_tranid)
+		  {
+		    tran_index = logtb_find_tran_index (thread_p, tran_id);
```
wrlawodms, line 3910: I'd like to ask something. It seems that all completed transactions have been already removed from the transaction table in `log_rv_analysis_complete()` during the analysis phase. Only commit/abort records after `stop_at` (if given) are left and the transactions of them are not removed here too. I wonder what's the point of trying to get and remove completed transactions here in the original code.

cristiarg: You might be right.
I will investigate the feasibility of this suggestion (aka: make the change, run a few tests) and come back with an answer.
Thanks.

cristiarg: After running a shell_debug test suite, one test failed - shell/_28_features_844/issue_10755_checkdb/_07_checkdb_kill_server at the assert added at line ~3951.
Investigation revealed that, for a given transaction, the following sequence of log records appeared in the log:
- `LOG_MVCC_UNDOREDO_DATA`
- `LOG_MVCC_DIFF_UNDOREDO_DATA`
- `LOG_COMMIT` - this cleared the tdes from the trantable
- `LOG_END_CHKPT` - in our example, this has 3 transactions in the state `TRAN_UNACTIVE_COMMITTED` (one of which is the transaction we were following) and has the effect of allocating again the entry for those specific transactions in the trantable (this happens in the sequence of functions `log_recovery_analysis` - `log_rv_analysis_record` - `log_rv_analysis_end_checkpoint` - `logtb_rv_find_allocate_tran_index`).

So, with this change, we might have actually discovered an actual race condition between adding commit log records for a transaction and marking that transaction as commited (on one side) vs starting a checkpoint (on the other side).
There are probably 2 solutions:
- we try to fix the race condition from the checkpoint logic by taking into consideration that a transaction in the `TRAN_UNACTIVE_COMMITTED` state can be considered committed; this still requires careful investigation and also a fix on the ABORT side where a similar race condition might be possible
- we leave the code as is - with the call to `logtb_free_tran_index` and the new optimization in place as well - while documenting why it is actually needed

ribram: One possible way to handle the race condition is to ensure that if the commit/abort log record has been appended, then the transaction is not included in the checkpoint list.

To ensure that, `log_tdes` may also track the `complete_lsa` that starts as null and that is set to the LSA of commit/abort log record while prior mutex is held and the log record is being appended. Transaction with `complete_lsa` different are ignored by checkpoint (the same way transactions with null `tail_lsa` are ignored).

Then there would be no ambiguous cases:
- If a transaction commits/aborts first, it is not appended in the checkpoint list of transactions. Recovery analysis frees the transaction descriptor and it is not resurrected by the analysis of checkpoint log record.
- If a transaction commits/aborts after checkpoint saves it, then recovery analysis will process the checkpoint first and then the completion log record that frees the transaction descriptor.

hornetmj: @ribram LGTM.

cristiarg:
@wrlawodms , @hornetmj , @ribram 
Implemented suggestion in comment https://github.com/CUBRID/cubrid/pull/2975#discussion_r702715095 , executed shell_debug suite a few times without incidents related to this modification.
Please re-review.

src/transaction/log_tran_table.c
```diff
@@ -6140,6 +6142,7 @@ log_tdes::on_sysop_start ()
       LSA_SET_NULL (&tail_lsa);
       LSA_SET_NULL (&undo_nxlsa);
       LSA_SET_NULL (&tail_topresult_lsa);
+      LSA_SET_NULL (&commit_abort_lsa);
```
ribram: I don't think this is necessary. System transactions never commit and never make use of `commit_abort_lsa`.
```suggestion
```

src/transaction/log_tran_table.c
```diff
@@ -6157,6 +6160,7 @@ log_tdes::on_sysop_end ()
...
```
ribram:
```suggestion
```

ribram (approved):

src/transaction/log_tran_table.c
cristiarg: Replace with an assert then?

ribram: Replacing with assert is a valid option.

wrlawodms (approved):
Looks good to me 👍

hornetmj (approved):

mhoh3963 (approved):

PR #2979, opened 2021-08-20
[CBRD-24019] Refactor routines related to predicate push.

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-24019

Added restrictions
 - subquery having GROUPBYNUM()

Removed restrictions 
 - aggregate function without group_by (CUBRIDSUS-1158)
 - merge query (CUBRIDSUS-7584)

Delete duplicate check.
 - remove xxxnum_found and analytic_found in term check
 - remove pt_check_copypush_subquery() (Check in subquery check)
 - remove routines related to predicate push at view transformation.

Write the checklist as the comments.


shparkcubrid:
SQL test error
**sql/_18_index_enhancement_qa/_02_full_test/_02_using_index/cases/_t111_13_keylimit_view.sql - linux_sql_64bit_release
sql/_28_features_930/issue_9024_loose_index_scan/cases/_02_enum.sql - linux_sql_64bit_release**
(v.a is not null ) ==> v.a is not null, fix to not add parentheses. It doesn't matter if there are parentheses or not. Answer sheets need to be changed.

**sql/_13_issues/_14_1h/cases/bug_bts_13715.sql - linux_sql_64bit_release**
predicate cannot be pushed in sub query with order by+limit. The previous answer sheets are wrong. Answer sheets need to be changed

src/parser/parser_support.c
```diff
@@ -1201,6 +1201,40 @@ pt_is_inst_or_orderby_num_node (PARSER_CONTEXT * parser, PT_NODE * tree, void *a
...
+  if (PT_IS_INSTNUM (tree) || PT_IS_ORDERBYNUM (tree) || PT_IS_GROUPBYNUM (tree))
+    {
+      *has_inst_orderby_num = true;
+    }
+
+  if (*has_inst_orderby_num)
+    {
+      *continue_walk = PT_STOP_WALK;
+    }
+  else if (PT_IS_QUERY_NODE_TYPE (tree->node_type))
```
beyondykk9, line 1226: if the has_inst_orderby_num is already true then the prior if clauses are not necessary.
```
if (*has_inst_orderby_num)
    {
      return tree;
    }
```
it is enough for in case of  *has_inst_orderby_num == true

src/parser/parser_support.c
```diff
@@ -3027,6 +3070,27 @@ pt_has_inst_or_orderby_num_in_where (PARSER_CONTEXT * parser, PT_NODE * node)
...
+bool
+pt_has_nullable_term (PARSER_CONTEXT * parser, PT_NODE * node)
+{
+  int has_nullable_term = 0;
+  PT_NODE *next;
+
+  next = node->next;
+  node->next = NULL;
+  (void) parser_walk_tree (parser, node, NULL, NULL, qo_check_nullable_expr, &has_nullable_term);
+  node->next = next;
```
beyondykk9, line 3089: if it checks just a single node, it looks better to call **qo_check_nullable_expr** directly without setting next temporarily.

src/parser/view_transform.c
```diff
@@ -2758,75 +2757,36 @@ pt_pushable_query_in_pos (PARSER_CONTEXT * parser, PT_NODE * query, int pos)
...
 		break;
 
 	      default:		/* do traverse */
 		parser_walk_leaves (parser, list, pt_check_pushable, &cinfo, NULL, NULL);
 		break;
 	      }			/* switch (list->node_type) */
 
-	    /* check for subquery, method, rownum, inst_num(), orderby_num(), groupby_num(): does not pushable if we
-	     * find these in corresponding item in select_list of query */
+	    /* check for subquery, method does not pushable in select_list of subquery */
```
beyondykk9, line 2846: check your comment, inst_num and orderby_num is not included for "not pushable" ?

src/parser/parser_support.c
shparkcubrid: If has_inst_orderby_num is true, it is not reached to 'else if'.
Did I understand your comment correctly?

src/parser/view_transform.c
shparkcubrid: Yes. It is checked in pt_has_inst_in_where_and_select_list ().
I didn't fix pt_check_pushable() because it is called elsewhere. I'll separate this function.

src/parser/parser_support.c
shparkcubrid: The node is a predicate which can have a next.

src/parser/parser_support.c
```diff
@@ -1201,6 +1201,40 @@ pt_is_inst_or_orderby_num_node (PARSER_CONTEXT * parser, PT_NODE * tree, void *a
...
+
+  if (PT_IS_INSTNUM (tree) || PT_IS_ORDERBYNUM (tree) || PT_IS_GROUPBYNUM (tree))
+    {
+      *has_inst_orderby_num = true;
+    }
+
+  if (*has_inst_orderby_num)
+    {
+      *continue_walk = PT_STOP_WALK;
+    }
```
ctshim, line 1225: It seems that it can be made as follows by removing the has_inst_orderby_num variable and reducing the substitution statement.

```
if ( (bool *) arg || PT_IS_INSTNUM (tree) || PT_IS_ORDERBYNUM (tree) || PT_IS_GROUPBYNUM (tree))
    {
        return tree;
    }

```

src/parser/view_transform.c
```diff
@@ -2758,75 +2757,36 @@ pt_pushable_query_in_pos (PARSER_CONTEXT * parser, PT_NODE * query, int pos)
...
 	    if (cinfo.query_found || cinfo.method_found || cinfo.xxxnum_found || cinfo.analytic_found)
 	      {
 		break;		/* not pushable */
```
ctshim: Since it is a structure that only works in the case of  (i == pos) , this loop can reach this point only once.

```
if (cinfo.query_found || cinfo.method_found || cinfo.xxxnum_found || cinfo.analytic_found)
    {
		break;		/* not pushable */
    }
else 
   {
      list = NULL;
      break;
```

src/parser/parser_support.c
beyondykk9: Refer to ctshim's comment.

src/parser/parser_support.c
beyondykk9: Ok, I see.

src/parser/parser_support.c
shparkcubrid: 1. has_inst_orderby_num is return value.
2. The reason why the else if statement below is needed is to not walking for subqueries included in the node.
```
else if (PT_IS_QUERY_NODE_TYPE (tree->node_type))
```

src/parser/view_transform.c
shparkcubrid: Okay. I'll fix it.

src/parser/parser_support.c
ctshim: OK,

ctshim (approved):

beyondykk9 (approved):

src/parser/view_transform.c
```diff
@@ -3290,25 +3266,18 @@ mq_copypush_sargable_terms_helper (PARSER_CONTEXT * parser, PT_NODE * statement,
...
-      && statement->info.query.q.select.connect_by == NULL && spec->info.spec.derived_table_type == PT_IS_SUBQUERY
-      && (derived_table = spec->info.spec.derived_table) && PT_IS_QUERY (derived_table)
-      && !PT_SELECT_INFO_IS_FLAGED (statement, PT_SELECT_INFO_IS_MERGE_QUERY))
-    {
-      info.type = FIND_ID_INLINE_VIEW;	/* inline view */
-      /* init input section */
-      info.in.spec = spec;
-      info.in.others_spec_list = statement->info.query.q.select.from;
-      info.in.attr_list = spec->info.spec.as_attr_list;
-      info.in.query_list = derived_table;
```
hgryoo, line 3308: info.in.query_list was set, but it is not set at the new change.
Is it not used anymore?

shparkcubrid: The query_list was changed to subquery.

hgryoo (approved):

youngjinj (approved):

PR #2980, opened 2021-08-20
[LETS-188] Transmit server type when server connection with master is re-established

ribram:
http://jira.cubrid.org/browse/LETS-188

If the connection with master is lost and then re-established, the server type is not transmitted. The first character in the server name is wrongfully interpreted as server type and the name gets trimmed.

Fix by always transmitting the server type first. css_pack_server_name was modified to append server type as suffix. The function name was changed to css_pack_message_to_master to be self-explanatory.

Other variables have been renamed to include message_to_master in the name.

### DIFF
--- a/src/connection/server_support.c
+++ b/src/connection/server_support.c
@@ -28,6 +28,7 @@
 #include "communication_server_channel.hpp"
 #include "load_worker_manager.hpp"
 #include "log_append.hpp"
+#include "server_type.hpp"
 #include "session.h"
 #include "thread_entry_task.hpp"
 #include "thread_entry.hpp"
@@ -944,8 +945,8 @@ css_reestablish_connection_to_master (void)
 {
   CSS_CONN_ENTRY *conn;
   static int i = CSS_WAIT_COUNT;
-  char *packed_server_name;
-  int name_length;
+  char *message_to_master;
+  int message_to_master_length;
 
   if (i-- > 0)
     {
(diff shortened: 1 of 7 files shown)


adrianbrab (approved):

src/connection/server_support.c
```diff
@@ -1365,20 +1365,13 @@ css_init (THREAD_ENTRY * thread_p, char *server_name, int name_length, int port_
...
 
-  /*
-   * The packing of the server name and type is done by setting the first
-   * character to the numer related to the server type enum value, the rest of the
-   * buffer space is used to copy the server name.
-   */
-  buffer = (char) (get_server_type () + '0');
-  buffer.append (server_name, name_length);
-  conn = css_connect_to_master_server (port_id, buffer.c_str (), name_length + 1);
+  conn = css_connect_to_master_server (port_id, message_to_master, message_to_master_length);
```
cristiarg, line 1368: As far as I saw, the signature of the function `css_connect_to_master_server` is still
```
extern CSS_CONN_ENTRY *css_connect_to_master_server (int master_port_id, const char *server_name, int name_length);
```
and the function is used to perform extra checks using the said argument;
Maybe it works now by virtue of the fact that the message contains first the server name and then the appended server type value, but at least the name of the parameter should specify that it is the entire "message_to_master", not just the server_name.
Instead of the already packed message, pass around a structure containing
```
{
  server name,
  type
}
```
(and, possibly, extra info is needed as per `css_pack_server_name`) and only pack when dispatching to the wire.
Even before this patch, the argument passed to `css_init` is already the "packed_server_name", not the actual server name.

ribram: You're right, server_name & name_length arguments are outdated. Actually, they were outdated even before the changes of this patch. The arguments were renamed as message_to_master & message_to_master_length.

Since css_connect_to_master_server is also used by the client and its implementation is very similar, I preferred to just change the argument name, instead of passing server name & type.

cristiarg: Fair enough.

cristiarg (approved):

PR #2987, opened 2021-08-26
[CBRD-24030] The db_get_ha_server_state API function is expected to return the current ha state of the sever.

hornetmj:
https://jira.cubrid.org/browse/CBRD-24030

### Purpose
The db_get_ha_server_state API function is expected to return the current ha state of the sever.

### Implementation
The db_get_ha_server_state function calls the css_ha_server_state function instead of boot_get_ha_server_state function.

### Remarks
N/A

### DIFF
--- a/src/transaction/boot_cl.c
+++ b/src/transaction/boot_cl.c
@@ -5694,6 +5694,14 @@ boot_get_host_connected (void)
   return boot_Host_connected;
 }
 
+#if defined (ENABLE_UNUSED_FUNCTION)
+HA_SERVER_STATE
+boot_get_ha_server_state (void)
+{
+  return boot_Server_credential.ha_server_state;
+}
+#endif /* ENABLE_UNUSED_FUNCTION */
+
 /*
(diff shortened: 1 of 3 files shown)


src/transaction/boot_cl.c
```diff
@@ -5694,12 +5694,6 @@ boot_get_host_connected (void)
   return boot_Host_connected;
 }
 
-HA_SERVER_STATE
```
mhoh3963, line 5697: how about remain this function, not remove ?

hornetmj: @mhoh3963 Can I ask a reason? There are no usages of this function.

mhoh3963: we can use it later.
So, I think that you are better to put code into #if defined (ENABLE_UNUSED_FUNCTION) ~ #end.

hornetmj: @mhoh3963 OK. I accept your recommendation.

mhoh3963 (approved):

PR #2989, opened 2021-08-27
[CBRD-24033] Changed to not allow DOT(.) in identifier exception rules.

youngjinj:
http://jira.cubrid.org/browse/CBRD-24033

Purpose
- Changed to not allow DOT(.) in identifier exception rules.

Implementation
- Change src/parser/csql_grammar.y
- An error occurs if the name contains DOT(.) for the objects below.
- Check when creating.
  * table, view, user, trigger, serial
- Check when rename.
  * table, view, trigger

```sql
/* AS-IS: SUCCESS */
csql> CREATE TABLE [a.b.c] (c1 int);
csql> CREATE TABLE "a.b.c" (c1 int);
csql> CREATE TABLE `a.b.c` (c1 int);

/* TO-BE: SYNTAX ERROR */
csql> CREATE TABLE [a.b.c] (c1 int);
csql> CREATE TABLE "a.b.c" (c1 int);
csql> CREATE TABLE `a.b.c` (c1 int);
ERROR: before ' ; '
Table name a.b.c not allowed. It cannot contain DOT(.).
```

Remarks
- N/A


src/parser/csql_grammar.y
```diff
@@ -8930,8 +9108,17 @@ unique_constraint
...
+				    if (node->info.index.index_name != NULL)
 				      {
+					/* Check if it contains DOT(.) */
+					const char *constraint_name = node->info.index.index_name->info.name.original;
+					if (constraint_name != NULL && strstr(constraint_name, ".") != NULL)
+					  {
+					    PT_ERRORf (this_parser, node,
+						       "Constraint name %s not allowed. It cannot contain DOT(.).",
+						       constraint_name);
+					  }
```
hgryoo: It seems typo: `index_name` instead of `constraint_name`?

hgryoo (approved):

src/parser/csql_grammar.y
youngjinj: When looking at index creation only, index_name is correct.
However, index_name use UNIQUE constraint name.
I used constraint_name because it is a name given together to create UNIQUE constraint.

Any additional comments are welcome.

src/parser/csql_grammar.y
```diff
@@ -2463,11 +2463,24 @@ create_stmt
...
 			  }
-
+			  	
 			if (qc)
 			  {
 			    qc->info.create_entity.if_not_exists = $5;
 			    qc->info.create_entity.entity_name = $6;
+
+			    /* Check if it contains DOT(.) */
+			    if (qc->info.create_entity.entity_name != NULL)
```
shparkcubrid: How about making function like pt_check_contain_dot ()?

youngjinj: I wanted to show different error messages for each query.

But if code changes happen, it's better to use functions.

shparkcubrid:
I hope that PR without additional change will be reviewed along with test cases.

src/parser/csql_grammar.y
shparkcubrid: If so, how about changing like ' if (pt_check_contain_dot()) ....'.
I think it is better readability.

youngjinj:
If the pt_check_contain_dot() function does not generate an error
and generates an error by checking the return value,
it seems that the code becomes more complex and the readability decreases.

To check the return value and raise an error, I would code something like this:

```c
if (pt_check_contain_dot(qc->info.create_entity.entity_name) == true)
  {
      const char * entity_name = NULL;
      if (qc->info.create_entity.entity_name != NULL)
        {
          entity_name = qc->info.create_entity.entity_name->info.name.original;
        }
      
      PT_ERRORf (this_parser, qc,
                 "Table name %s not allowed. It cannot contain DOT(.).",
                 entity_name);
  }
````

Even if the possibility is small, it is a problem if qc->info.create_entity.entity_name is NULL.

Then there is no need to create a pt_check_contain_dot() function.
Rather, the first code looks more readable.

```c
/* Check if it contains DOT(.) */
if (qc->info.create_entity.entity_name != NULL)
  {
      const char *entity_name = qc->info.create_entity.entity_name->info.name.original;
      if (entity_name != NULL && strstr(entity_name, ".") != NULL)
        {
          PT_ERRORf (this_parser, qc,
                     "Table name %s not allowed. It cannot contain DOT(.).",
                     entity_name);
        }
  }
```

The work of the pt_check_contain_dot() function is done by the strstr() function,
so there is no need to make it a function.

==> SE PARK
I don't think so. null checking can be written in pt_check_dot (). but it is not important things.

src/parser/name_resolution.c
```diff
@@ -3901,6 +3901,33 @@ pt_check_unique_names (PARSER_CONTEXT * parser, const PT_NODE * p)
...
+ * this is to prevent difficulties in distinguishing.
+ */
+void
+pt_check_contain_dot (PARSER_CONTEXT * parser, PT_NODE * p)
+{
+  if (p != NULL)
+    {
+      const char *name = p->info.name.original;
+
+      if (name != NULL && strstr (name, ".") != NULL)
```
ctshim: It would be better to use the strchr() function instead of strstr().

ctshim:
In csql_grammar.y, the information corresponding to the table name is expressed as class_name or identifier.

Instead of calling pt_check_contain_dot() in each statement, please consider modifying the following

class_name
    : identifier_without_dot
        {{
            $$ = $1
         }}

identifier_without_dot
      :  identifier
         {{
            pt_check_contain_dot($1);
            $$ = $1
         }}

You can write identifier_without_dot instead of identifier in the place of the table name expressed as identifier elsewhere.
If you modify it in this way, you will be able to check the same in the drop statement.

```
csql> create tabel "t.x" ( id int);

In line 1, column 8,

ERROR: invalid create statement


0 command(s) successfully processed.
csql> drop table "t.x";

In line 1, column 12,

ERROR: before ' ; '
Unknown class "t.x".


0 command(s) successfully processed.
```

shparkcubrid (approved):

src/parser/name_resolution.c
youngjinj: Thank you. I'll change it to the strchr() function.

beyondykk9 (approved):

youngjinj:
Added test cases to CBRD-24033. ([cbrd-24033_test-case_20210831.sql](http://jira.cubrid.org/secure/attachment/1014870/cbrd-24033_test-case_20210831.sql))

youngjinj:
A lot has changed from the first PR code.
Please review again.

shparkcubrid (approved):

beyondykk9 (approved):

hgryoo (approved):

youngjinj:
The reason for the failing test case was commented on issue [CBRD-24033](http://jira.cubrid.org/browse/CBRD-24033?focusedCommentId=4760490&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-4760490).

Cases that failed while using user_specified_name in the delete_name rule succeed.

src/parser/csql_grammar.y
```diff
@@ -5185,45 +5192,125 @@ only_all_class_spec
...
+	: identifier DOT identifier
+		{{
+
+			PT_NODE *user = $1;
+			PT_NODE *name = $3;
+
+			if (user
+			    && name
+			    && user->node_type == PT_NAME
+			    && name->node_type == PT_NAME)
```
ctshim: Isn't there a case where it is an identifier but not PT_NAME?

src/parser/csql_grammar.y
```diff
@@ -5185,45 +5192,125 @@ only_all_class_spec
...
```
ctshim: Isn't there a case where it is an identifier but not PT_NAME?

src/parser/csql_grammar.y
```diff
@@ -5185,45 +5192,125 @@ only_all_class_spec
...
 
 		DBG_PRINT}}
 	| identifier
 		{{
 
-			$$ = $1;
+			PT_NODE *name = $1;
+
+			if (name
+			    && name->node_type == PT_NAME)
```
ctshim: Isn't there a case where it is an identifier but not PT_NAME?

src/parser/csql_grammar.y
```diff
@@ -21628,6 +21715,32 @@ simple_path_id_list
 		DBG_PRINT}}
 	;
 
+identifier_without_dot
+	: identifier
+		{{
+
+			PT_NODE *p = $1;
+
+			if (p && p->node_type == PT_NAME)
```
ctshim: identifier is always PT_NAME.

src/parser/csql_grammar.y
```diff
@@ -5185,45 +5192,125 @@ only_all_class_spec
...
-			PT_NODE *user_node = $1;
-			PT_NODE *name_node = $3;
+			PT_NODE *user = $1;
+			PT_NODE *name = $3;
 
-			if (name_node != NULL && user_node != NULL)
+			if (user
+			    && name
+			    && user->node_type == PT_NAME
+			    && name->node_type == PT_NAME)
```
ctshim: identifier_without_dot is always PT_NAME.

src/parser/csql_grammar.y
```diff
@@ -5185,45 +5192,125 @@ only_all_class_spec
...
+	| user_specified_name
+		{{
+
+			$$ = $1;
+			PARSER_SAVE_ERR_CONTEXT ($$, @$.buffer_pos)
+
+		DBG_PRINT}}
+	;
+
+object_name
```
ctshim: What is the difference between user_specified_name and object_name?
It looks identical.
If there is no reason to divide the replacement for the existing class_name into two, it is better to unify it into one object_name.

Or wouldn't it be better to keep class_name as below?
{quote}
user_specified_name_without_dot --> user_specified_class_name
user_specified_name  --> class_name
user_specified_name_list  --> class_name_list
{quote}

src/parser/csql_grammar.y
youngjinj: Okay. I'll just check for NULL.

src/parser/csql_grammar.y
youngjinj: Okay. I'll just check for NULL.

src/parser/csql_grammar.y
youngjinj: Okay. I'll just check for NULL.

src/parser/csql_grammar.y
youngjinj: Okay. I'll just check for NULL.

src/parser/csql_grammar.y
youngjinj: Okay. I'll just check for NULL.

src/parser/csql_grammar.y
youngjinj: user_specified_name sets the PT_NAME_INFO_USER_SPECIFIED flag to change the original name to user_name.object_name before semantic check.
```
PT_NAME_INFO_SET_FLAG(name, PT_NAME_INFO_USER_SPECIFIED);
```
```
class_name → user_specified_name_without_dot (PT_NAME_INFO_USER_SPECIFIED)
class_name → user_specified_name (PT_NAME_INFO_USER_SPECIFIED)
class_name_list → user_specified_name_list (PT_NAME_INFO_USER_SPECIFIED)
class_name → object_name
```
Code related to flags will be covered in another PR.

ctshim: OK

beyondykk9:
Please check the fail of Window build.

youngjinj:
After rerunning CircleCI, the Windows build was successful, the test case result was the same.

beyondykk9 (approved):

ctshim (approved):

src/parser/csql_grammar.y
```diff
@@ -5185,24 +5192,74 @@ only_all_class_spec
...
+			$$ = name;
+			PARSER_SAVE_ERR_CONTEXT ($$, @$.buffer_pos)
+
+		DBG_PRINT}}
+	| identifier_without_dot
+		{{
+
+			PT_NODE *name = $1;
+
+			if (name)
```
hyunikn, line 5254: This if block should be removed.

hyunikn (approved):

src/parser/csql_grammar.y
youngjinj: The code for the PT_NAME_INFO_USER_SPECIFIED flag didn't come with it, so the code below is commented out.
```
PT_NAME_INFO_SET_FLAG (name, PT_NAME_INFO_USER_SPECIFIED);
```
Since it is a feature branch, the code for the flag will be included through another PR. So keep the block.

src/parser/csql_grammar.y
```diff
@@ -3617,7 +3624,7 @@ alter_stmt
 	  INDEX				/* 2 */
 	  identifier			/* 3 */
 	  ON_				/* 4 */
-	  class_name			/* 5 */
+	  user_specified_name		/* 5 */
```
shparkcubrid: class_name and user_specified_name are too different. Is user_specified_name appropriate?

src/parser/csql_grammar.y
```diff
@@ -3495,22 +3502,22 @@ alter_stmt
...
-	  only_class_name				/* 9 */
-	  opt_index_column_name_list			/* 10 */
-	  opt_where_clause				/* 11 */
+	  opt_hint_list				/* 3 */
+	  opt_reverse				/* 4 */
+	  opt_unique				/* 5 */
+	  INDEX					/* 6 */
+	  identifier				/* 7 */
+	  ON_					/* 8 */
+	  only_class_name			/* 9 */
```
shparkcubrid, line 3524: From the next review, please resolve the indentation issue before requesting a source code review.

hgryoo (approved):

src/parser/csql_grammar.y
youngjinj: How about using user_specified_name in the class_name rule?
In the same way, other object_name is also changed.
- serial_name, trigger_name
```
class_name
        : user_specified_name
        ;
```

src/parser/csql_grammar.y
youngjinj: Would it be a problem to turn on the hide whitespace option and review?

youngjinj: Highlighting the whole code context is a problem.

PR #2994, opened 2021-08-30
[CBRD-24036] predicate push does not work in correlated subquery.

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-24036

The modification of correlation level was not applied to union clause. I fix that union queries are also applied.

Added restrictions
  - OID column
  - correlated column with aggregation in predicate (restricted by performance issues)


shparkcubrid:
From commit 1f72225 it's related to this PR

src/parser/parser_support.c
```diff
@@ -9699,6 +9729,11 @@ pt_has_non_groupby_column_node (PARSER_CONTEXT * parser, PT_NODE * node, void *a
       return node;
     }
 
+  if (node->info.name.correlation_level != 0)
```
beyondykk9: it looks better "node->info.name.correlation_level > 0" for readability.

ctshim (approved):

beyondykk9 (approved):

hgryoo (approved):

youngjinj (approved):

src/parser/parser_support.c
shparkcubrid: Okay. I fixed it.

PR #2997, opened 2021-09-01
[CBRD-24038] Modify the specifications for column property on ALTER TABLE clause

beyondykk9:
http://jira.cubrid.org/browse/CBRD-24038

AS-IS:
When the data type is changed with the ALTER CHANGE/MODIFY ...' query, the properties are dropped if the entire column properties are not listed.
There is an example where it is trying to change a column of INT data type to BIGINT data type, but COMMENT disappears because COMMENT part is not written.

Issue:
When the data type is changed with the 'ALTER CHANGE/MODIFY ...' query, the COMMENT disappears, the NOT NULL constraint, and the DEFAULT value disappear.
If there is no schema backed up, it is difficult to restore it.
It looks it's the user's mistake, but if there is a problem at the site that is providing technical support and an engineer tries to fix it, it takes a lot of effort.

TO-BE:
We will improve the function so that the column properties not be changed without specifying the properties as like the CBRD-23718 issue.

### DIFF
--- a/src/query/execute_schema.c
+++ b/src/query/execute_schema.c
@@ -11065,6 +11065,11 @@ build_attr_change_map (PARSER_CONTEXT * parser, DB_CTMPL * ctemplate, PT_NODE * 
       {
 	int *const p = &(attr_chg_properties->p[i]);
 
+	if (*p & ATT_CHG_PROPERTY_LOST)
+	  {
+	    continue;
+	  }
+
 	if (*p & ATT_CHG_PROPERTY_PRESENT_OLD)
 	  {
 	    if (*p & ATT_CHG_PROPERTY_PRESENT_NEW)
--- a/src/query/execute_schema.c
+++ b/src/query/execute_schema.c
@@ -8616,28 +8616,28 @@ do_create_entity (PARSER_CONTEXT * parser, PT_NODE * node)
 
       /* check for mis-creating string type with -1 precision */
       for (column = query_columns; column != NULL; column = db_query_format_next (column))
-        {
-          switch (column->domain->type->id)
-            {
-              case DB_TYPE_VARCHAR:
-                if (column->domain->precision == DB_DEFAULT_PRECISION)
-                  {
-                    column->domain->precision = DB_MAX_VARCHAR_PRECISION;
-                  }
-                break;
-              case DB_TYPE_VARNCHAR:
-                if (column->domain->precision == DB_DEFAULT_PRECISION)
-                  {
-                    column->domain->precision = DB_MAX_VARNCHAR_PRECISION;
-                  }
-                else if (column->domain->precision > DB_MAX_VARNCHAR_PRECISION)
-                  {
-                    column->domain->precision = DB_MAX_VARNCHAR_PRECISION;
-                  }
-                default:
-                break;
-            }
-        }
+	{
+	  switch (column->domain->type->id)
+	    {
+	    case DB_TYPE_VARCHAR:
+	      if (column->domain->precision == DB_DEFAULT_PRECISION)
+		{
+		  column->domain->precision = DB_MAX_VARCHAR_PRECISION;
+		}
+	      break;
+	    case DB_TYPE_VARNCHAR:
+	      if (column->domain->precision == DB_DEFAULT_PRECISION)
+		{
+		  column->domain->precision = DB_MAX_VARNCHAR_PRECISION;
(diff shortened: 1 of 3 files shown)


src/query/execute_statement.c
```diff
@@ -2043,7 +2043,6 @@ do_update_maxvalue_of_auto_increment_serial (PARSER_CONTEXT * parser, MOP * seri
 
   numeric_coerce_string_to_num (DB_SERIAL_MAX, strlen (DB_SERIAL_MAX), INTL_CODESET_ISO88591, &e38);
 
-  assert_release (att->info.attr_def.auto_increment != NULL);
```
ctshim, line 2046: Is "att->info.attr_def.auto_increment" changed to allow NULL?

ctshim (approved):

src/query/execute_statement.c
beyondykk9: auto_increment is nullable by calling this routine.

beyondykk9:
The sql failed cases should be modified according to the modified specification for  property changing by ALTER command.

src/query/execute_statement.c
shparkcubrid: Why did you remove assertion?

src/query/execute_schema.c
```diff
@@ -10982,6 +10981,8 @@ build_attr_change_map (PARSER_CONTEXT * parser, DB_CTMPL * ctemplate, PT_NODE *
 	  chg_prop_idx = P_S_CONSTR_UNI;
 	  save_pt_costraint = true;
 	  break;
+	case PT_CONSTRAIN_NULL:
+	  attr_chg_properties->p[P_NOT_NULL] = ATT_CHG_PROPERTY_LOST;
```
shparkcubrid, line 10985: Can 'break' be omitted?

beyondykk9: it ommited break clause intentionally for applying common routine.

src/query/execute_statement.c
beyondykk9: Actually, auto_increment is not used in this routine.

src/query/execute_schema.c
```diff
@@ -10610,7 +10610,6 @@ do_change_att_schema_only (PARSER_CONTEXT * parser, DB_CTMPL * ctemplate, PT_NOD
     {
       MOP auto_increment_obj = NULL;
 
-      assert (attribute->info.attr_def.auto_increment != NULL);
```
shparkcubrid, line 10613: Can assert be deleted? What test cases can I check this with?

src/query/execute_schema.c
```diff
@@ -10572,7 +10572,7 @@ do_change_att_schema_only (PARSER_CONTEXT * parser, DB_CTMPL * ctemplate, PT_NOD
 
   /* the serial property has not changed, we are only dealing with renaming */
   if (is_att_prop_set (attr_chg_prop->p[P_NAME], ATT_CHG_PROPERTY_DIFF)
-      && attribute->info.attr_def.auto_increment != NULL
+      && is_att_prop_set (attr_chg_prop->p[P_AUTO_INCR], ATT_CHG_PROPERTY_PRESENT_OLD)
```
shparkcubrid, line 10575: I think this routines to change the serial name. Why should it be done when old serial is present?

beyondykk9: the property auto_increment has no name. the auto_increment is set once it can not be dropped.
The serial name related to the auto_increment should be also renamed as column name changed.

src/query/execute_schema.c
beyondykk9: it can raise segement fault at debug mode if the auto_increment is not figured. So, I remove the assert.

src/query/execute_schema.c
shparkcubrid: I got it.

src/query/execute_schema.c
byungwook-kim: The serial name related to the auto_increment should be also renamed.

shparkcubrid: I heard from kim as a routine that changes the auto increment name when the column name is changed.
I got it.

shparkcubrid (approved):

hgryoo (approved):

youngjinj (approved):

PR #2999, opened 2021-09-01
[LETS-200] Fix processing new HA mode server on cub_master

ribram:
http://jira.cubrid.org/browse/LETS-200

When a server is configured for HA-mode, the first character sent in the connection request to master is not server type. It is the '#' character.

When the '#' character is received first, automatically consider the server a transaction server (page servers do not boot in HA mode).

### DIFF
--- a/src/executables/master.c
+++ b/src/executables/master.c
@@ -442,10 +442,22 @@ receive_server_info (CSS_CONN_ENTRY * conn, unsigned short rid, std::string & db
   int exit_code = css_receive_data (conn, rid, &buffer, &buffer_length, -1);
   if (exit_code == NO_ERRORS)
     {
-      // *INDENT-OFF*
-      type = static_cast<SERVER_TYPE> (buffer[0] - '0');
-      // *INDENT-ON*
-      dbname = std::string (buffer + 1, buffer_length - 1);
+      const char first_char = buffer[0];
+      if (first_char == '#')
+	{
+	  // HA mode on; only transaction servers connect.
+	  type = SERVER_TYPE_TRANSACTION;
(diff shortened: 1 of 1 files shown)


adrianbrab (approved):

src/executables/master.c
```diff
@@ -442,12 +442,24 @@ receive_server_info (CSS_CONN_ENTRY * conn, unsigned short rid, std::string & db
...
+	{
+	  // First character represents server type
+	  // *INDENT-OFF*
+	  type = static_cast<SERVER_TYPE> (buffer[0] - '0');
+	  // *INDENT-ON*
+	  dbname = std::string (buffer + 1, buffer_length - 1);
+	}
 
-      MASTER_ER_LOG_DEBUG (ARG_FILE_LINE, "A server with database:'%s' of type:'%s' wants to connect to cub_master.",
+      MASTER_ER_LOG_DEBUG (ARG_FILE_LINE, "The %s server of type %s wants to connect to cub_master.",
```
cristiarg, line 462: Why not have a translation function from server type to string?
I know that if it is not a page server, all other are assumed to actually be transaction but we can have something like:
```
  SERVER_TYPE_UNKNOWN, --> "unknown" | "unknown (transaction)"
  SERVER_TYPE_TRANSACTION, --> "transaction"
  SERVER_TYPE_PAGE, --> "page"
  SERVER_TYPE_ANY, --> "any (transaction)"
```
It would help in debugging.

cristiarg (approved):

src/executables/master.c
ribram: I'll make a separate issue.

PR #3000, opened 2021-09-01
[LETS-154] Improve performance of parallel recovery/replication

cristiarg:
http://jira.cubrid.org/browse/LETS-154

Improve performance of parallel recovery/replication.

Patch is made up of two main changes:
1.  refactoring:
    - the functionality provided by `minimum_log_lsa_monitor` has been integrated into member class `redo_parallel::min_unapplied_log_lsa_monitoring`
    - `redo_job_queue` became superfluous and has also been integrated partly in `redo_parallel` and partly in `redo_parallel::redo_task`
    - class `redo_parallel::redo_task` has expanded in scope to also handle internally the consumption of jobs (previously handled by `redo_job_queue`)
2. implementation change:
    - async tasks are currently created and managed by `redo_parallel`
    - a pushed job is distributed upfront to a certain async task based on its `VPID` with help from a hashing function (`hash<VPID>`)
    - `redo_task` will manage internally the jobs' consumption as needed using the approach of two swapped containers, one on the "push" (produce) side - `m_produce_vec`, and one on the "pop" (consume) side - `redo_task::execute` - `jobs_vec`
    - after each job has been executed, it is `retire`d at which point its ownership goes back to the reusable job container
    - updated unit-tests to test changes:
      - some of the tests have been deleted because the corresponding tested entities became internal implementation details
      - the core unit-test functor `execute_test` has been adapted to fully cover `redo_parallell` surface
    - `min_unapplied_log_lsa_monitoring` performs calculation pro-actively via an internal thread by interrogating the async tasks but will also be triggered from the `wait_past_target_log_lsa` function when the latter is called

Other:
- extended reported time spend in `log_recovery_redo` with the time needed for "setting up" (this includes pre-populating reusable job container with desired number of jobs, allocating memory for internals and starting up infrastructure)
- added a `log_lsa::is_max` function
- added a `hash<VPID>` implementation based on examples in Boost libraries


src/transaction/log_recovery_redo_parallel.hpp
```diff
@@ -311,9 +156,30 @@ namespace cublog
...
+      std::hash<VPID> m_vpid_hash;
+
+      /* the not-applied set from the outside by the main thread (for recovery, the main thread
+       * is the thread running the analysis redo and undo; for replication, the main thread is the
+       * recovery applying thread);
+       * the reason the outer log_lsa is needed is because some of the log entries are being applied
+       * synchronously by the main thread and the calculation of the minimum not-applied log_lsa needs
+       * to be accurate
+       */
+      std::atomic<log_lsa> m_outer_not_applied_log_lsa;
```
cristiarg: The `_not_applied_log_lsa` members and associated functions can be abstracted away in an implementation detail class.

src/transaction/log_recovery_redo_parallel.hpp
```diff
@@ -120,7 +65,7 @@ namespace cublog
     public:
```
cristiarg, line 64: at least class `redo_job_base` should be moved out of `redo_parallel`;
maybe also `redo_task`

src/transaction/log_replication.cpp
```diff
@@ -173,11 +170,12 @@ namespace cublog
   void
   replicator::conclude_task_execution ()
```
cristiarg: to be removed

src/transaction/log_lsa.hpp
```diff
@@ -104,12 +105,19 @@ log_lsa::log_lsa (const int64_t value)
...
   return pageid == NULL_LOG_PAGEID;
 }
 
+constexpr bool
+log_lsa::is_max () const
+{
+  return *this == MAX_LSA;
+}
+
+
```
ribram:
```suggestion

```

src/transaction/log_recovery.c
```diff
@@ -1014,15 +1016,19 @@ log_recovery_redo (THREAD_ENTRY * thread_p, log_recovery_context & context)
...
   }
 #endif
 
-  const auto time_start = std::chrono::system_clock::now ();
+  /* 'setting_up' measures the time taken to initialize parallel log recovery redo infrstructure */
+  // *INDENT-OFF*
+  const auto time_dur_setting_up_ms = std::chrono::duration_cast <std::chrono::milliseconds>(
+        std::chrono::system_clock::now () - time_start_setting_up);
+  // *INDENT-ON*
+  const auto time_start_main = std::chrono::system_clock::now ();
```
ribram: - A single `std::chrono::system_clock::now ()` call, under disabled indentation
- Avoid breaking after parenthesis if possible
- Nitpicking: remove `time_` from `time_dur_` to shorten the name.
```suggestion
  // *INDENT-OFF*
  const auto time_start_main = std::chrono::system_clock::now ();
  const auto dur_setting_up_ms =
    std::chrono::duration_cast <std::chrono::milliseconds> (time_start_main - time_start_setting_up);
  // *INDENT-ON*
```

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -500,15 +99,67 @@ namespace cublog
...
+
       redo_parallel::task_active_state_bookkeeping &m_task_state_bookkeeping;
-      redo_parallel::redo_job_queue &m_queue;
 
       cubperf::statset_definition m_perf_stats_definition;
       perf_stats m_perf_stats;
 
       log_rv_redo_context m_redo_context;
+
+      std::unique_ptr<redo_job_vector_t> m_produce_vec;
```
ribram: Why unique pointer to vector?
```suggestion
      redo_job_vector_t m_produce_vec;
```

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -517,135 +168,287 @@ namespace cublog
...
+  redo_parallel::redo_task::~redo_task ()
+  {
+    assert (is_idle ());
   }
 
   void
   redo_parallel::redo_task::execute (context_type &context)
   {
-    bool finished = false;
+    std::unique_ptr<redo_job_vector_t> jobs_vec { new redo_job_vector_t () };
```
ribram: Why unique pointer to vector?
```suggestion
    redo_job_vector_t jobs_vec;
```

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -517,135 +168,287 @@ namespace cublog
...
+      {
+	const log_lsa &job_log_lsa = a_job->get_log_lsa ();
+	set_not_applied_log_lsa_from_push_side (job_log_lsa);
+      }
+
+    m_produce_vec->push_back (a_job);
+  }
+
+  inline void
+  redo_parallel::redo_task::pop_jobs (std::unique_ptr<redo_parallel::redo_task::redo_job_vector_t> &a_out_job_vec)
```
```suggestion
  redo_parallel::redo_task::pop_jobs (redo_job_vector_t &a_out_job_vec)
```

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -517,135 +168,287 @@ namespace cublog
...
-	bool adding_finished = false;
-	redo_parallel::redo_job_base *const job = m_queue.pop_job (adding_finished);
+	pop_jobs (jobs_vec);
+	const bool jobs_popped { !jobs_vec->empty ()};
 	m_perf_stats.time_and_increment (cublog::PERF_STAT_ID_PARALLEL_POP);
-
-	if (job == nullptr && adding_finished)
+	if (!jobs_popped && m_adding_finished.load ())
 	  {
 	    finished = true;
```
ribram: In situations like these, it is usual to use `while (true)` loop and break it when it's finished.
```suggestion
	    break;
```

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -517,135 +168,287 @@ namespace cublog
...
+		jobs_vec->clear ();
+		m_perf_stats.time_and_increment (cublog::PERF_STAT_ID_PARALLEL_EXECUTE_AND_RETIRE);
 	      }
 	  }
       }
 
+    assert (jobs_vec->empty ());
+
     m_task_state_bookkeeping.set_inactive ();
+    m_in_execution.store (false);
```
ribram: `m_in_execution` seems to do the same thing as `m_task_state_bookkeeping`. Is it necessary?

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -517,135 +168,287 @@ namespace cublog
...
 
+  bool
+  redo_parallel::redo_task::is_idle () const
+  {
+    bool is_vector_empty = false;
+    {
+      std::scoped_lock<std::mutex> slock { m_produce_vec_mtx };
+      is_vector_empty = m_produce_vec->empty ();
+    }
+    return is_vector_empty && !m_in_execution.load ();
```
ribram: Too complicated code for a simple check.
```suggestion
    std::scoped_lock<std::mutex> slock { m_produce_vec_mtx };
    return m_produce_vec->empty () && !m_in_execution.load ();
```

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -517,135 +168,287 @@ namespace cublog
...
+
+    log_lsa expected_log_lsa;
+    do
+      {
+	expected_log_lsa = m_not_applied_log_lsa.load ();
+	// -or-equal because push side can fill in a minimum log_lsa, while the task (ie: pop side) is
+	// still empty, which is then found and tested-only by the pop side
+	assert (expected_log_lsa <= a_log_lsa);
+      }
+    while (!m_not_applied_log_lsa.compare_exchange_strong (expected_log_lsa, a_log_lsa));
```
ribram: There is no race here to need `compare_exchange_strong`. There is only one situation when main thread changes `m_not_applied_log_lsa`, when both queues are empty and calling `set_not_applied_log_lsa_from_pop_side` is not possible.

The worker thread may simply set the new value:
```suggestion
    assert (m_do_monitor_not_applied_log_lsa);
    assert (m_not_applied_log_lsa <= a_log_lsa);

    m_not_applied_log_lsa = a_log_lsa;
```

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -517,135 +168,287 @@ namespace cublog
...
+
+    {
+      const log_lsa snapshot_not_applied_log_lsa { m_not_applied_log_lsa.load () };
+      // strict comparison because jobs have ever-increasing log_lsa's
+      assert (snapshot_not_applied_log_lsa == MAX_LSA || snapshot_not_applied_log_lsa < a_log_lsa);
+    }
+
+    log_lsa expected_not_applied_log_lsa { MAX_LSA };
+    m_not_applied_log_lsa.compare_exchange_strong (expected_not_applied_log_lsa, a_log_lsa);
+    // TODO: any check for the return value?
```
ribram: `m_not_applied_log_lsa = MAX_LSA` is set only with the protection of m_produce_vec_mtx and once set, it may not be changed back. The `compare_exchange_strong` may not be necessary.
```suggestion
    assert (m_not_applied_log_lsa == MAX_LSA || m_not_applied_log_lsa < a_log_lsa);
    if (m_not_applied_log_lsa == MAX_LSA)
      {
        m_not_applied_log_lsa = MAX_LSA;
      }
```

ribram (commented):
Only partially reviewed; I will continue to review later.

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -517,135 +168,287 @@ namespace cublog
...
+
+  inline void
+  redo_parallel::redo_task::push_job (redo_parallel::redo_job_base *a_job)
+  {
+    std::scoped_lock<std::mutex> slock { m_produce_vec_mtx };
+
+    if (m_do_monitor_not_applied_log_lsa && m_produce_vec->empty ())
+      {
+	const log_lsa &job_log_lsa = a_job->get_log_lsa ();
+	set_not_applied_log_lsa_from_push_side (job_log_lsa);
```
ribram: Is there a reason to prefer an extra variable and two-line code, instead of no variable and one line code?
```suggestion
	set_not_applied_log_lsa_from_push_side (a_job->get_log_lsa ());
```

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -736,6 +523,107 @@ namespace cublog
...
+
+    return calculated_minimum_not_applied_log_lsa;
+  }
+
+  log_lsa
+  redo_parallel::calculate_minimum_log_lsa ()
+  {
+    assert (m_do_monitor_minimum_log_lsa);
+
+    log_lsa minimum_not_applied_log_lsa { m_outer_not_applied_log_lsa.load () };
```
ribram: Add a comment explaining that the calculus has to start with the LSA of main thread (outer).

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -736,6 +523,107 @@ namespace cublog
...
+    assert (a_target_lsa != NULL_LSA);
+
+    log_lsa calculated_minimum_not_applied_log_lsa { MAX_LSA };
+
+    {
+      // avoid gratuitously notifying the calculating internal thread if the condition is already satisfied
+      std::scoped_lock<std::mutex> slock { m_calculate_minimum_not_applied_log_lsa_mtx };
+      calculated_minimum_not_applied_log_lsa = m_calculated_minimum_not_applied_log_lsa;
+    }
+
```
ribram: No lock is required to do the early out.
```suggestion
    if (a_target_lsa < m_calculated_minimum_not_applied_log_lsa)
      {
        // Already past
        return;
      }
```

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -517,135 +168,287 @@ namespace cublog
...
+      std::scoped_lock<std::mutex> slock { m_produce_vec_mtx };
+      is_vector_empty = m_produce_vec->empty ();
+    }
+    return is_vector_empty && !m_in_execution.load ();
+  }
+
+  inline void
+  redo_parallel::redo_task::push_job (redo_parallel::redo_job_base *a_job)
+  {
+    std::scoped_lock<std::mutex> slock { m_produce_vec_mtx };
```
ribram: This is my first encounter with scoped_lock so I did a bit of reading to understand the differences between various locks. I tend to agree with the conclusions of this [response](https://stackoverflow.com/a/60172828):

> So my advice is to use the simplest tool for the job:
>  1. lock_guard if you need to lock exactly 1 mutex for an entire scope.
>  2. scoped_lock if you need to lock a number of mutexes that is not exactly 1.
>  3. unique_lock if you need to unlock within the scope of the block (which includes use with a condition_variable).

src/transaction/log_recovery_redo_parallel.hpp
ribram: Do you think the class may be nested into redo_parallel?

src/transaction/log_recovery_redo_parallel.cpp
ribram: Add assert is is not MAX_LSA.
```suggestion
    log_lsa minimum_not_applied_log_lsa { m_outer_not_applied_log_lsa.load () };
    assert (minimum_not_applied_log_lsa < MAX_LSA);
```

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -517,135 +168,287 @@ namespace cublog
...
-
-		job->retire (m_task_idx);
-		m_perf_stats.time_and_increment (cublog::PERF_STAT_ID_PARALLEL_RETIRE);
+		for (auto &job : *jobs_vec)
+		  {
+		    if (m_do_monitor_not_applied_log_lsa)
+		      {
+			set_not_applied_log_lsa_from_pop_side (job->get_log_lsa ());
+		      }
+		    job->execute (thread_entry, m_redo_context);
```
ribram: Is the time separation between execution and retirement irrelevant?

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -517,135 +168,287 @@ namespace cublog
...
   void
   redo_parallel::add (redo_job_base *a_job)
   {
     assert (false == m_waited_for_termination);
-    assert (false == m_job_queue.get_adding_finished ());
+    assert (false == m_adding_finished.load ());
+
+    const vpid &job_vpid = a_job->get_vpid ();
+    const std::size_t vpid_hash = m_vpid_hash (job_vpid);
+    const std::size_t task_index = vpid_hash % m_task_count;
```
ribram: The successions of operations is easier to be read in a single line:
```suggestion
    const std::size_t task_index = m_vpid_hash (a_job->get_vpid ()) % m_task_count;
```

src/transaction/log_recovery_redo_parallel.hpp
```diff
@@ -311,9 +156,30 @@ namespace cublog
...
+       * the reason the outer log_lsa is needed is because some of the log entries are being applied
+       * synchronously by the main thread and the calculation of the minimum not-applied log_lsa needs
+       * to be accurate
+       */
+      std::atomic<log_lsa> m_outer_not_applied_log_lsa;
+
+      /* following members control the pro-active calculation of the minimum not-applied log_lsa as well
+       * as means to actually trigger and wait for the calculation asynchronously
+       */
+      log_lsa m_calculated_minimum_not_applied_log_lsa;
```
ribram: To reduce the name size, `minimum_not_applied_log_lsa` may be replaced by `min_unapplied_lsa`.

src/transaction/log_recovery_redo_parallel.hpp
ribram: Can "outer" be replaced with something more suggestive? "main thread" maybe? Or something that we can use to reference that main execution thread everywhere?

unit_tests/log/test_main_log_recovery_parallel.cpp
```diff
@@ -64,43 +67,104 @@ void execute_test (const log_recovery_test_config &a_test_config,
...
   ut_database_values_generator global_values{ a_database_config };
   for (size_t idx = 0u; idx < a_test_config.redo_job_count; ++idx)
     {
       ux_ut_redo_job_impl job = db_online->generate_changes (*db_recovery, global_values);
 
       if (job->is_volume_creation () || job->is_page_creation ())
 	{
+	  if (a_test_config.wait_past_previous_log_lsa)
+	    {
+	      log_redo_parallel.set_outer_not_applied_log_lsa (job->get_log_lsa ());
```
ribram: Normally `set_outer_not_applied_log_lsa()` should be called for each log record, regardless if it is applied by the main therad or it will be applied by a worker thread.

src/transaction/log_recovery_redo_parallel.cpp
cristiarg: Evolution: raw pointer -> unique pointer -> vector.

unit_tests/log/test_main_log_recovery_parallel.cpp
cristiarg: Done.

src/transaction/log_recovery_redo_parallel.cpp
cristiarg: Fixed

src/transaction/log_recovery_redo_parallel.cpp
cristiarg: Replaced with an expanded task state bookkeeping which also helps get rid of `m_in_execution`.

src/transaction/log_recovery_redo_parallel.cpp
cristiarg: Replaced

src/transaction/log_recovery_redo_parallel.cpp

src/transaction/log_recovery_redo_parallel.cpp

src/transaction/log_recovery_redo_parallel.cpp

src/transaction/log_recovery_redo_parallel.cpp

src/transaction/log_recovery_redo_parallel.cpp
cristiarg: Added an initial `log_lsa` as a ctor parameter and asserted prerequisite in the ctor.

src/transaction/log_recovery_redo_parallel.hpp

src/transaction/log_recovery_redo_parallel.cpp

src/transaction/log_recovery_redo_parallel.cpp

src/transaction/log_recovery_redo_parallel.cpp
cristiarg: Removed.

src/transaction/log_recovery_redo_parallel.hpp

src/transaction/log_recovery.c
```diff
@@ -997,13 +997,15 @@ log_recovery_redo (THREAD_ENTRY * thread_p, log_recovery_context & context)
 
   volatile TRANID tran_id;
   volatile LOG_RECTYPE log_rtype;
+  const auto time_start_setting_up = std::chrono::system_clock::now ();
```
ribram, line 1001:
```suggestion
  // *INDENT-OFF*
  const auto time_start_setting_up = std::chrono::system_clock::now ();
  // *INDENT-ON*
```

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -24,652 +24,620 @@
...
-    std::lock_guard<std::mutex> consume_lockg (m_consume_queue_mutex);
+    m_perf_stats.accumulate (a_output_stats, a_output_stats_size);
+  }
 
-    out_adding_finished = false;
+  bool
+  redo_parallel::redo_task::is_idle () const
+  {
+    std::lock_guard<std::mutex> lockg { m_produce_vec_mtx };
+    return m_produce_vec.empty () && !m_task_state_bookkeeping.is_active (m_task_idx);
```
ribram: Doesn't `!m_task_state_bookkeeping.is_active (m_task_idx)` imply `m_produce_vec.empty ()`?

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -24,652 +24,620 @@
...
-	// it is ensured that indeed no more data exists and that no more data will be produced
-	out_adding_finished = m_adding_finished.load ();
-
-	// if no more data will be produced (signalled by the flag), the
-	// consumer will just need to terminate; otherwise, consumer is expected to
-	// spin-wait and try again
-	return nullptr;
+	set_unapplied_log_lsa_from_push (a_job->get_log_lsa (), lockg);
+
+	m_produce_vec_cv.notify_one ();
```
ribram, line 313: Notifying before unlocking `m_produce_vec_mtx` may wake-up thread only to get blocked on the mutex. Unlock the mutex before the notification.

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -24,652 +24,620 @@
...
+  redo_parallel::redo_task::notify_adding_finished ()
   {
-    // if consumption of everything in consume queue finished; see whether there's some more in the other one
-    if (m_consume_queue->size () == 0)
-      {
-	// TODO: a second barrier such that, consumption threads do not 'busy wait' by
-	// constantly swapping two empty containers; works in combination with a notification
-	// dispatched after new work items are added to the produce queue; this, in effect means that
-	// this function will semantically become 'blocking_pop'
+    m_produce_vec_cv.notify_one ();
```
ribram, line 328: I think a mutex lock/unlock is required before calling `m_produce_vec_cv.notify_one ()`. Otherwise this is possible:

- Worker calls `pop_jobs`, loads m_adding_finished false.
- Main thread sets adding finished to true immediately after and tries to notify the worker
- The workers fails to pass condition and starts waiting on the condition variable right after missing the notification.

By forcing a mutex lock/unlock, the worker can no longer miss the notification; it is either already waiting on the condition variable or it will check m_adding_finished after it is modified.

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -24,652 +24,620 @@
...
+    jobs_vec.reserve (PARALLEL_REDO_JOB_VECTOR_RESERVE_SIZE);
+
+    for ( ; ; )
       {
-	m_minimum_log_lsa->set_for_produce (a_job->get_log_lsa ());
+	pop_jobs (jobs_vec);
+	m_perf_stats.time_and_increment (cublog::PERF_STAT_ID_PARALLEL_POP);
+	if (jobs_vec.empty () && m_adding_finished.load ())
+	  {
+	    break;
```
```suggestion
	if (m_adding_finished.load ())
	  {
	    assert (jobs_vec.empty ());
	    break;
```

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -736,6 +705,26 @@ namespace cublog
...
+  redo_parallel::set_main_thread_unapplied_log_lsa (const log_lsa &a_log_lsa)
+  {
+    m_min_unapplied_log_lsa_calculation.set_main_thread_unapplied_log_lsa (a_log_lsa);
+  }
+
+  log_lsa
+  redo_parallel::get_calculated_min_unapplied_log_lsa ()
+  {
+    return m_min_unapplied_log_lsa_calculation.get_calculated_log_lsa ();
+  }
```
ribram: `redo_parallel::get_calculated_min_unapplied_log_lsa ()` and `min_unapplied_log_lsa_calculation::get_calculated_log_lsa ()` are only used in unit test; they don't seem to be useful for the engine. My suggestion is to remove them.

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -24,652 +24,617 @@
...
+	}
+
+	// there might be more than one waiting thread
+	m_calculate_cv.notify_all ();
+
+	{
+	  std::unique_lock<std::mutex> ulock { m_calculate_mtx };
+	  // wait might be interrupted by an outside notify and this is expected
+	  m_calculate_cv.wait_for (ulock, std::chrono::milliseconds (1000));
+	}
```
ribram: Wait time of 1 second is too large for a system that has only one active transaction; if notification occurs right before waiting, the transaction will be blocked for one second. Reduce the wait time to 10 msec.

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -24,652 +24,617 @@
...
-   * redo_parallel::redo_task - definition
-   *********************************************************************/
+    // notify thread performing calculation
+    m_calculate_cv.notify_one ();
 
-  constexpr unsigned short redo_parallel::redo_task::WAIT_AND_CHECK_MILLIS;
+    std::unique_lock<std::mutex> ulock { m_calculate_mtx };
+    m_calculate_cv.wait (ulock, [this, &a_target_lsa] ()
+    {
+      //assert (!m_adding_finished.load ());
```
```suggestion
```

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -24,652 +24,617 @@
...
   redo_parallel::wait_for_termination_and_stop_execution ()
   {
-    assert (false == m_waited_for_termination);
+    assert (false == m_adding_finished.load ());
+
+    m_adding_finished.store (true);
+    for (auto &redo_task: m_redo_tasks)
+      {
+	redo_task->notify_adding_finished ();
+      }
```
ribram: Since every task has to be notified, an atomic shared m_adding_finished might not be useful; maybe each task should get its own local m_adding_finished that gets set when notify_adding_finished is called.

src/transaction/log_recovery_redo_parallel.hpp
```diff
@@ -135,170 +79,124 @@ namespace cublog
...
        */
       void wait_for_termination_and_stop_execution ();
 
+      inline bool get_waited_for_termination () const
+      {
+	// all steps from the wait_for_termination.. blocking function must have been done
+	return m_adding_finished.load ()
+	       && !m_task_state_bookkeeping.is_any_active ()
+	       && (m_worker_pool == nullptr);
+      }
```
ribram: This function is incomprehensible; there is no such thing as being waited for termination :P. The code may be moved directly where the check is needed.

src/transaction/log_recovery_redo_parallel.hpp
```diff
@@ -135,170 +79,124 @@ namespace cublog
...
+	   * synchronously by the main thread and the calculation of the minimum not-applied log_lsa needs
+	   * to be accurate
 	   */
-	  inline void set_active ();
+	  std::atomic<log_lsa> m_main_thread_unapplied_log_lsa;
 
-	  /* decrement the internal number of active tasks
-	   */
-	  inline void set_inactive ();
+	  const std::vector<std::unique_ptr<redo_task>> *m_redo_tasks;
```
```suggestion
	  const std::vector<std::unique_ptr<redo_task>> &m_redo_tasks;
```

src/transaction/log_recovery_redo_parallel.cpp
cristiarg: it does

src/transaction/log_recovery_redo_parallel.cpp
cristiarg: done

src/transaction/log_recovery_redo_parallel.cpp
cristiarg: done

src/transaction/log_recovery_redo_parallel.cpp
cristiarg: The target is to consume all available jobs, not to finish when signaled.
Even if the flag is set to true from the main thread, a task may still have one more batch of jobs to process. See the wait lambda in `pop_jobs` which has two exit paths:
- either jobs are still to be processed (in which case the value of the flag is irrelevant, and might be true which would invalidate the proposed new assert)
- or there are no jobs, in which case the flag becomes the stop condition

src/transaction/log_recovery_redo_parallel.cpp
cristiarg: noticed that too; done already

src/transaction/log_recovery_redo_parallel.hpp
cristiarg: removed, asserts were already part of the dtor

src/transaction/log_recovery_redo_parallel.cpp
ribram: Right, disregard my suggestion.

src/transaction/log_recovery_redo_parallel.hpp
cristiarg: Done. But having it as a pointer allowed to spot a race condition which might have slipped otherwise.

src/transaction/log_recovery_redo_parallel.cpp
cristiarg: done

src/transaction/log_recovery.c
cristiarg: really needed? indent does not seem to be broken in anyway

src/transaction/log_recovery_redo_parallel.cpp
cristiarg: Done. And it does not need to be atomic.

src/transaction/log_recovery_redo_parallel.cpp
cristiarg: removed

src/transaction/log_recovery.c
ribram: Right now it is not broken, but it becomes unpredictable. It may break after other changes, with no apparent relations.

src/transaction/log_recovery_redo_parallel.cpp
ribram: it does not look done, the wait time is still 1000 msec.

cristiarg: Sorry. My mistake.

ribram (approved):
Please fix https://github.com/CUBRID/cubrid/pull/3000#discussion_r704313688 and merge.

PR #3002, opened 2021-09-02
[LETS-199] do not start dwb daemons

cristiarg:
http://jira.cubrid.org/browse/LETS-199

- make daemon name mandatory; adapt calls
- do not start double write buffer daemons on a transaction server with remote storage

### DIFF
--- a/src/transaction/boot_sr.c
+++ b/src/transaction/boot_sr.c
@@ -2528,12 +2528,14 @@ boot_restart_server (THREAD_ENTRY * thread_p, bool print_restart, const char *db
   oid_set_root (&boot_Db_parm->rootclass_oid);
 
   /* Load and recover data pages before log recovery */
-  // TODO: only when !is_tran_server_with_remote_storage ?
-  error_code = dwb_load_and_recover_pages (thread_p, log_path, log_prefix);
-  if (error_code != NO_ERROR)
+  if (!is_tran_server_with_remote_storage ())
     {
-      ASSERT_ERROR ();
-      goto error;
+      error_code = dwb_load_and_recover_pages (thread_p, log_path, log_prefix);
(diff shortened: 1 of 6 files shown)


src/transaction/boot_sr.c
```diff
@@ -5066,6 +5076,7 @@ boot_create_all_volumes (THREAD_ENTRY * thread_p, const BOOT_CLIENT_CREDENTIAL *
   /* Create double write buffer if not already created. DWB creation must be done before first volume.
    * DWB file is created on log_path.
    */
+  // TODO: !is_tran_server_with_remote_storage ?
```
ribram: Not necessary to check here. When database is created for remote storage, no disk files will be created. This should not be reached.

src/transaction/boot_sr.c
```diff
@@ -2528,6 +2528,7 @@ boot_restart_server (THREAD_ENTRY * thread_p, bool print_restart, const char *db
   oid_set_root (&boot_Db_parm->rootclass_oid);
 
   /* Load and recover data pages before log recovery */
+  // TODO: only when !is_tran_server_with_remote_storage ?
   error_code = dwb_load_and_recover_pages (thread_p, log_path, log_prefix);
```
ribram: It is already earlied out inside `dwb_load_and_recover_pages()`. It the check is moved here, it should be removed from `dwb_load_and_recover_pages()`

cristiarg: I moved the check upper and removed checks from `double_write_buffer.c`. I think it is more appropriate to concentrate these checks in fewer places/functions.

src/transaction/boot_sr.c
cristiarg: Ok

ribram (approved):

PR #3004, opened 2021-09-03
[CBRD-22529] [CTE extension] delete with CTEs about round function failed:Attribute "" was not found.

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-22529

I fix to remove meaningless errors that occur when searching for correlated columns.

### DIFF
--- a/src/parser/name_resolution.c
+++ b/src/parser/name_resolution.c
@@ -717,6 +717,11 @@ pt_bind_name_or_path_in_scope (PARSER_CONTEXT * parser, PT_BIND_NAMES_ARG * bind
 	    }
 	  level++;
 	}
+      if (node && er_errid () == ER_OBJ_INVALID_ATTRIBUTE)
+	{
+	  /* An error is meaningful only when it cannot be found in all scopes. */
+	  er_clear ();
+	}
     }
   else
     {


src/parser/name_resolution.c
```diff
@@ -717,6 +717,11 @@ pt_bind_name_or_path_in_scope (PARSER_CONTEXT * parser, PT_BIND_NAMES_ARG * bind
 	    }
 	  level++;
 	}
+      if (node && er_errid () == ER_OBJ_INVALID_ATTRIBUTE)
+	{
+	  /* An error is meaningful only when it cannot be found in all scopes. */
+	  er_clear ();
+	}
```
ctshim, line 724: After calling pt_expand_external_path() on line 710, it seems more readable to check as follows.

```
      if (er_errid () == ER_OBJ_INVALID_ATTRIBUTE)
	{
	  /* An error is meaningful only when it cannot be found in all scopes. */
	  er_clear ();
	}
```
However, there are no errors, so look at the whole context and decide.

ctshim (approved):

hgryoo (approved):

beyondykk9 (approved):

youngjinj (approved):

src/parser/name_resolution.c
shparkcubrid: I think checking outside the 'for loop' looks good too. I'll keep this source code.

PR #3008, opened 2021-09-07
[CBRD-24031] Change the _db_server catalog lookup method

ctshim:
http://jira.cubrid.org/browse/CBRD-24031

Purpose
   Allow different users to create the same server name. For this, the index of the _db_server catalog is changed and the search method is also changed.

Implementation
  Change the spec so that the server name cannot contain the period character in any case.
  Change the specification to allow the server name to include owner information.
  Change the index of the _db_server catalog to "link_name + owner"
  Change the _db_server catalog lookup method.
  Change the server name designation in dblink syntax from "IdName" to "identifier".
  Change HOST information from < identifier > to <CHAR_STRING>

Remarks
N/A


beyondykk9 (approved):

src/query/execute_statement.c
```diff
@@ -18835,3 +18807,175 @@ get_dblink_password_decrypt (const char *passwd_cipher, DB_VALUE * decrypt_val)
...
+      if (!owner_name)
+	{
+	  return NULL;
+	}
+    }
+
+  if (owner_name)
+    {
+      sprintf (query,
+	       "SELECT [_db_server], [owner] FROM [_db_server] WHERE [link_name] = '%s' AND [owner].[name] = '%s'",
```
shparkcubrid, line 18860: '[owner].[name]' <== Does this work?

shparkcubrid (approved):

src/parser/csql_grammar.y
```diff
@@ -3159,6 +3159,32 @@ create_stmt
...
+                                   PT_NODE *val = parser_new_node (this_parser, PT_VALUE);
+                                   if (val)                    
+                                     {
+                                        int     err;
+                                        char    cipher[512];
+
+                                        val->type_enum = PT_TYPE_CHAR;
+                                        val->info.value.string_type = ' ';
+
+                                        err = pt_check_dblink_password(this_parser, NULL, cipher, sizeof(cipher));
```
hgryoo: It returns error code by `err` value, but the following if-statement only handles an error with `pt_has_error()`.

youngjinj (approved):

src/parser/csql_grammar.y
ctshim: Modified to set the necessary error message inside the pt_check_dblink_password() function.
Thanks for the careful check.

hgryoo (approved):

PR #3010, opened 2021-09-08
[CBRD-24040] Prevent from drop and alter of system tables by sysadm

beyondykk9:
http://jira.cubrid.org/browse/CBRD-24040

The ALTER and DROP for system catalog tables should be prevented for proper system operation. Currently, the system catalog tables starts with named "db_" and "db" can be easily dropped or altered by system admin user called "sysadm" as follows:

```
csql -u dba -S --sysadm demodb

sysadm> drop table db_ha_apply_info;
Execute OK. (0.008472 sec) Committed
```

The DROP, ALTER, and RENAME operations cannot be permitted for system catalog tables even though it's sysadm mode.

To test the queries, please refer to the last comment of CBRD-24040

### DIFF
--- a/src/object/authenticate.c
+++ b/src/object/authenticate.c
@@ -307,6 +307,7 @@ MOP Au_root = NULL;
  * use the AU_DISABLE, AU_ENABLE macros instead.
  */
 int Au_disable = 1;
+bool Au_sysadm = false;
 
 /*
  * Au_ignore_passwords
@@ -5860,7 +5861,7 @@ check_authorization (MOP classobj, SM_CLASS * sm_class, DB_AUTH type)
    * Callers generally check Au_disable already to avoid the function call.
    * Check it again to be safe, at this point, it isn't going to add anything.
    */
-  if (Au_disable && !(sm_class->flags & SM_CLASSFLAG_SYSTEM))
+  if (Au_disable && (!Au_sysadm || !(sm_class->flags & SM_CLASSFLAG_SYSTEM)))
     {
       return NO_ERROR;
     }
@@ -8624,6 +8625,19 @@ au_disable (void)
 }
 
 /*
+ * au_sysadm_disable - set Au_disable for sysadm
+ *   return: original Au_disable value
+ */
+int
+au_sysadm_disable (void)
+{
+  int save = Au_disable;
+  Au_disable = 1;
+  Au_sysadm = true;
+  return save;
+}
+
+/*
  * au_enable - restore Au_disable
  *   return:
  *   save(in): original Au_disable value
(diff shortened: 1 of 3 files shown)


src/object/authenticate.c
```diff
@@ -6150,7 +6150,7 @@ au_fetch_class_internal (MOP op, SM_CLASS ** class_ptr, AU_FETCHMODE fetchmode,
 	}
     }
 
-  if (Au_disable || !(error = check_authorization (classmop, class_, type)))
+  if ((Au_disable && type != DB_AUTH_ALTER) || !(error = check_authorization (classmop, class_, type)))
```
ctshim, line 6154: Here is the code for the au_fetch_instance()function.

```
  if (Au_disable || !(error = check_authorization (classmop, class_, type)))
    {
      error = fetch_instance (op, obj_ptr, mode, fetch_version_type);
    }
```
    
Don't you need to fix it in the same context as the edit here?

src/object/authenticate.c
```diff
@@ -5860,7 +5860,7 @@ check_authorization (MOP classobj, SM_CLASS * sm_class, DB_AUTH type)
    * Callers generally check Au_disable already to avoid the function call.
    * Check it again to be safe, at this point, it isn't going to add anything.
    */
-  if (Au_disable)
+  if (Au_disable && !(sm_class->flags & SM_CLASSFLAG_SYSTEM))
```
ctshim: Is there any problem in executing db_drop_class() in the au_install() function when executing createdb?

src/object/authenticate.c
beyondykk9: au_fetch_instance is not called at ALTERing a table or a view. Authenticate check needs only at ALTERing.
And in check_authorization I added the code to check if current use is sysadm.

src/object/authenticate.c
beyondykk9: I fixed it to createdb not to check autorization. Now authorization just check excpets sysadm's altering system class.

ctshim (approved):

src/object/authenticate.c
shparkcubrid: How about below for readability?
```
au_fetch_class_internal ()
if ((Au_disable && !Au_sysadm) || !(error = check_authorization (classmop, class_, type)))

check_authorization ()
if (Au_disable && !(type == DB_AUTH_ALTER && (sm_class->flags & SM_CLASSFLAG_SYSTEM)) )
```

shparkcubrid (approved):

hgryoo (approved):

youngjinj (approved):

src/object/authenticate.c
beyondykk9: Ok, thanks, but I will keep the style.

PR #3015, opened 2021-09-09
[LETS-188] Fix server name after connection with master is re-established

ribram:
http://jira.cubrid.org/browse/LETS-188

The name registered on restarted cub_master after the server reconnects starts with the number character equivalent to the server type. The explanation is that css_Master_server_name global used to include the server type.

Fix by setting css_Master_server_name value to server name instead of the message sent to cub_master. Message packing was pushed down to css_init as a consequence.

### DIFF
--- a/src/connection/server_support.c
+++ b/src/connection/server_support.c
@@ -50,6 +50,7 @@
 #include <netinet/in.h>
 #endif /* !WINDOWS */
 #include <assert.h>
+#include <string>
 
 #include "porting.h"
 #include "memory_alloc.h"
@@ -262,6 +263,7 @@ static cubthread::wait_seconds css_get_connection_thread_timeout_configuration (
 static bool css_get_server_request_thread_pooling_configuration (void);
 static cubthread::wait_seconds css_get_server_request_thread_timeout_configuration (void);
 static void css_start_all_threads (void);
+static std::string css_pack_message_to_master (const char *server_name);
 // *INDENT-ON*
 
 #if defined (SERVER_MODE)
@@ -945,8 +947,6 @@ css_reestablish_connection_to_master (void)
 {
   CSS_CONN_ENTRY *conn;
   static int i = CSS_WAIT_COUNT;
-  char *message_to_master;
-  int message_to_master_length;
 
   if (i-- > 0)
     {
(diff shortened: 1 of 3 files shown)


adrianbrab (approved):

src/connection/server_support.c
```diff
@@ -1449,6 +1454,11 @@ css_init (THREAD_ENTRY * thread_p, char *message_to_master, int message_to_maste
   css_windows_shutdown ();
 #endif /* WINDOWS */
 
+  if (message_to_master != NULL)
+    {
+      free (message_to_master);
```
cristiarg: `free_and_init` macro ?

src/connection/server_support.c
```diff
@@ -1309,16 +1309,21 @@ css_start_shutdown_server ()
  *       css_initialize_server_interfaces before calling this function.
  */
 int
-css_init (THREAD_ENTRY * thread_p, char *message_to_master, int message_to_master_length, int port_id)
+css_init (THREAD_ENTRY * thread_p, const char *server_name, int port_id)
```
cristiarg, line 1306: update comments above

src/connection/server_support.c
```diff
@@ -1309,16 +1309,21 @@ css_start_shutdown_server ()
...
+  char *message_to_master = NULL;
 
-  if (message_to_master == NULL || port_id <= 0)
+  if (server_name == NULL || port_id <= 0)
     {
       return ER_FAILED;
     }
 
+  message_to_master = css_pack_message_to_master (server_name, &message_to_master_length);
+  assert (message_to_master != NULL);
```
cristiarg: log entry and/or `return ER_FAILED` besides assert?
if NULL, it will fallthrough to `shutdown`(being that `css_connect_to_master_server` will also return NULL), but there should be a log entry at least explaining why the server went down

src/connection/server_support.c
ribram: free is no longer required, `message_to_master` is changed to `std::string`

src/connection/server_support.c
ribram: done

src/connection/server_support.c
ribram: changed `message_to_master` type to std::string; the null case was not really necessary. packing the message should never fail and `server_name` should never be null.

cristiarg (approved):

PR #3019, opened 2021-09-13
[LETS-146] cubrid server start/stop command on single_node configuration

adrianbrab:
http://jira.cubrid.org/browse/LETS-146

Changed the `cubrid server start` command to start both a page and a transaction server when single_node is specified in the configuration file.
Also `cubrid server stop` was modified to stop both servers on a single node configuration.


src/executables/util_service.c
```diff
@@ -1668,16 +1669,16 @@ process_server (int command_type, int argc, char **argv, bool show_usage, bool c
...
-		  if (status != NO_ERROR)
-		    {
-		      util_log_write_errid (MSGCAT_UTIL_GENERIC_SERVICE_PROPERTY_FAIL);
-		      print_result (PRINT_SERVER_NAME, status, command_type);
-		      break;
-		    }
+		  util_log_write_errid (MSGCAT_UTIL_GENERIC_SERVICE_PROPERTY_FAIL);
+		  print_result (PRINT_SERVER_NAME, status, command_type);
+		  break;
+		}
```
ribram: Move `#if !defined(WINDOWS)` after `sysprm_load_and_init()`. The single_node configuration must be known on windows too.

src/executables/util_service.c
```diff
@@ -1699,14 +1700,39 @@ process_server (int command_type, int argc, char **argv, bool show_usage, bool c
...
 		  int pid;
-		  const char *args[] = { UTIL_CUBRID_NAME, token, NULL };
-		  status = proc_execute (UTIL_CUBRID_NAME, args, false, false, false, &pid);
 
-		  if (status == NO_ERROR && !is_server_running (CHECK_SERVER, token, pid))
+		  if (server_type_config::SINGLE_NODE ==
+		      (server_type_config) prm_get_integer_value (PRM_ID_SERVER_TYPE))
 		    {
-		      status = ER_GENERIC_ERROR;
+		      const char *args_page[] = { UTIL_CUBRID_NAME, "-t", "page", token, NULL };
```
ribram: Add a comment explaining the purpose of the code.
```suggestion
		      // start both page server and transaction server
		      // page server must be started before transaction server
		      const char *args_page[] = { UTIL_CUBRID_NAME, "-t", "page", token, NULL };
```

The whole scope may be moved to a separate function (possible name `process_server_start_single_node ()`)

src/executables/utility.h
```diff
@@ -1438,6 +1441,8 @@ typedef struct _ha_config
 #define COMMDB_HA_ADMIN_INFO_L                  "admin-info"
 #define COMMDB_HA_START_UTIL_PROCESS_S          't'
 #define COMMDB_HA_START_UTIL_PROCESS_L          "start-ha-util-process"
+#define COMMDB_STOP_SINGLE_NODE_S               'n'
+#define COMMDB_STOP_SINGLE_NODE_L               "single_node"
```
ribram:
```suggestion
#define COMMDB_STOP_SINGLE_NODE_S               12116
#define COMMDB_STOP_SINGLE_NODE_L               "single-node-stop"
```

src/executables/utility.h
```diff
@@ -880,6 +882,7 @@ typedef struct _ha_config
 #define CHECK_TRANSACTION_SERVER  "Transaction-Server"
 
 #define COMMDB_SERVER_STOP      "-S"
+#define COMMDB_SINGLE_NODE_STOP "-n"
```
ribram: Make the commdb command self-explanatory by using only the long version and naming it `single-node-stop`. See also the comment on commdb arguments.
```suggestion
#define COMMDB_SINGLE_NODE_STOP "--single-node-stop"
```

src/executables/commdb.c
```diff
@@ -1283,6 +1312,9 @@ main (int argc, char **argv)
 	    }
 	  commdb_Arg_server_name = strdup (optarg);
 	  break;
+	case 'n':
```
ribram:
```suggestion
	case COMMDB_STOP_SINGLE_NODE_S:
```

src/executables/util_service.c
adrianbrab: moved it now

src/executables/util_service.c
adrianbrab: i will move the code to a new function.

src/executables/util_service.c
```diff
@@ -1836,6 +1855,38 @@ process_server (int command_type, int argc, char **argv, bool show_usage, bool c
...
+  int status, pid;
+
+  const char *args_page[] = { UTIL_CUBRID_NAME, "-t", "page", server_name, NULL };
+  status = proc_execute (UTIL_CUBRID_NAME, args_page, false, false, false, &pid);
+
+  if (status == NO_ERROR && !is_server_running (CHECK_PAGE_SERVER, server_name, pid))
+    {
+      status = ER_GENERIC_ERROR;
+    }
+  print_result (PRINT_PAGE_SERVER_NAME, status, command_type);
```
ribram, line 1866: Stop if page server could not be started.
```suggestion
  print_result (PRINT_PAGE_SERVER_NAME, status, command_type);
  if (status != NO_ERROR)
    {
      return status;
    }
```

src/executables/util_service.c
```diff
@@ -1836,6 +1855,38 @@ process_server (int command_type, int argc, char **argv, bool show_usage, bool c
   return status;
 }
 
+/*
+ * process_server_start_single_node - start both a page and a transaction server,
+ *                                     the page server must be started before transaction server
+ *
+ * return:
```
ribram: Remove comment lines that offer no information.
```suggestion
 *                                     the page server must be started before transaction server
```

src/executables/util_service.c
```diff
@@ -1836,6 +1855,38 @@ process_server (int command_type, int argc, char **argv, bool show_usage, bool c
...
+ */
+static int
+process_server_start_single_node (char *server_name, int command_type)
+{
+  int status, pid;
```
ribram: Separate each variable on its own line and initialize.
```suggestion
  int status = NO_ERROR;
  int pid = 0;
```

ribram (approved):

src/executables/commdb.c
```diff
@@ -1075,8 +1091,20 @@ process_batch_command (CSS_CONN_ENTRY * conn)
 
   if ((commdb_Arg_server_name) && (!commdb_Arg_halt_shutdown))
     {
-      pid = process_server_info_pid (conn, (char *) commdb_Arg_server_name, COMM_SERVER);
-      process_slave_kill (conn, (char *) commdb_Arg_server_name, commdb_Arg_shutdown_time, pid);
+      if (!commdb_Arg_single_node)
+	{
+	  pid = process_server_info_pid (conn, (char *) commdb_Arg_server_name, COMM_SERVER, SERVER_TYPE_ANY);
+	  process_slave_kill (conn, (char *) commdb_Arg_server_name, commdb_Arg_shutdown_time, pid);
```
cristiarg, line 1096: I know it's historic, but this function does not `kill`. It commands a graceful stop.

src/executables/commdb.c
```diff
@@ -1075,8 +1091,20 @@ process_batch_command (CSS_CONN_ENTRY * conn)
...
+	}
+      else
+	{
+	  // kill the transaction server first
+	  pid = process_server_info_pid (conn, (char *) commdb_Arg_server_name, COMM_SERVER, SERVER_TYPE_TRANSACTION);
```
cristiarg: Same comment as on the start side. The system should be able to recover from any situation when only one of the servers is currently alive. It should also not try to stop a server which is running - on the same port - but is not in the 'single node' config (as in the situation where: server was started in single not config, config was changed, an now it is trying to stop servers in "single node" config).

src/executables/util_service.c
```diff
@@ -1836,6 +1855,42 @@ process_server (int command_type, int argc, char **argv, bool show_usage, bool c
...
+    {
+      return status;
+    }
+
+  const char *args_transaction[] = { UTIL_CUBRID_NAME, "-t", "transaction", server_name, NULL };
+  status = proc_execute (UTIL_CUBRID_NAME, args_transaction, false, false, false, &pid);
+
+  if (status == NO_ERROR && !is_server_running (CHECK_TRANSACTION_SERVER, server_name, pid))
+    {
+      status = ER_GENERIC_ERROR;
```
cristiarg, line 1878: If only the PS can be started, shouldn't the PS be closed at this point?
What I want to say is that the system should be easy to use and can recover itself from all scenarios.
Say, upon a call to `process_server_start_single_node`, the situation can be in any of the states:
- both servers are already running
- only PS is already running
- only TS is already running
- no server is already running
It seems to be that the function only deals with the last situation but without any upfront check.
Also some error messages should be logged in case of any failure, for easier troubleshooting (both during devel/QA as well as later by DBA's).
Another aspect is the inconsistent/incorrect switch between configs where a server might already be alive - on the same port - that was not started in the "single node" config (but with a normal config).

src/executables/commdb.c
ribram: True, and it is not only stopping the slave. However, there is an entire set of names that has the same problem and it would have to be renamed together. I'll make a low priority renaming task for this.

src/executables/commdb.c
ribram: For the case that only one server is started when single_node configuration stop command is issued, to try and stop both servers in the right order is the best approach. A small change required here, pid must be verified; if it is equal to zero, calling process_slave_kill is not necessary.

The other case could be considered a problem:

- Both servers are started within single_node configuration
- Configuration is changed to something else.
- `cubrid server stop` command is issued. it will only try to stop one server.

The easiest approach may be to try and stop both type of servers regardless of configuration. In consequence, the only required change is how server stop is handled on commdb; the process_server stop changes and the --single-node-stop commdb option would become unnecessary.

Do you guys agree with this?

src/executables/util_service.c
ribram: The single_node configuration is not for regular usage. It may be useful for testing and for trying out the CUBRID engine. Changing from single_node to other configuration or vice-versa is not a particularly interesting case.

src/executables/commdb.c
ribram: As I mentioned in the starting servers discussion thread, handling the configuration changes is not really important. More important about the suggested change is that it simplifies this patch.

adrianbrab: I will change so that it always looks for both a transaction and a page server and only stops the ones with valid pids.

cristiarg: @adrianbrab The suggestion was to always try to stop 3 processes: one regular server, two ATS/PS servers. Since they can only be started in the respective disjoint configurations, the stopping of 3 possible processes is a "catch all" option. I think it's ok. (please note that I do not know if stopping  an "any" server will also stop either a TS or PS whichever is found or will not stop anything; I am not familiar with that part).

adrianbrab: @cristiarg  wouldn't a search for both a TS and a PS always stop all the present servers? As an unknown server will default to be a TS.

cristiarg: As said in my last remark of my previous comment, I am not familiar with that part.
If it suffices to search two times to stop any combination of:
- default cub_server started
- only TS started
- only PS started
- both TS and PS started

then that is what needs to be done.
Just make sure to have the scenario tested.

src/executables/commdb.c
cristiarg: Don't. As it would needlessly diverge even more from develop. :)

src/executables/commdb.c
ribram: Default cub_server is a transaction server. Stopping one transaction server and one page server should cover all mentioned cases.

PR #3020, opened 2021-09-13
[LETS-204] Add server type to string function

adrianbrab:
http://jira.cubrid.org/browse/LETS-204

Added `server_type_to_string` and `server_type_config_to_string` to replace hard coded usages of `"page"` and `"transaction"` and group the already used logic into functions.

### DIFF
--- a/src/base/server_type_enum.hpp
+++ b/src/base/server_type_enum.hpp
@@ -33,4 +33,30 @@ enum class server_type_config
   PAGE,
   SINGLE_NODE,
 };
+
+inline const char *
+server_type_to_string (SERVER_TYPE type)
+{
+  switch (type)
+    {
+    case SERVER_TYPE_PAGE:
+      return "page";
+    default:
+      return "transaction";
+    }
+}
+
+inline const char *
+server_type_config_to_string (server_type_config type)
+{
+  switch (type)
+    {
+    case server_type_config::PAGE:
+      return "page";
+    case server_type_config::TRANSACTION:
+      return "transaction";
+    case server_type_config::SINGLE_NODE:
+      return "single_node";
+    }
+}
 #endif // _SERVER_TYPE_ENUM_H_
--- a/src/base/server_type_enum.hpp
+++ b/src/base/server_type_enum.hpp
@@ -18,6 +18,7 @@
 
 #ifndef _SERVER_TYPE_ENUM_H_
 #define _SERVER_TYPE_ENUM_H_
+#include <assert.h>
 
 typedef enum
 {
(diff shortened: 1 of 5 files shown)


adrianbrab:
Should the usages of `"page"` and `"transaction"`, present in `argument_handler` in server.c, be replaced?


```C
	  if (std::strcmp (optarg, "transaction") == 0)
	    {
	      set_server_type (SERVER_TYPE_TRANSACTION);
	    }
	  else if (std::strcmp (optarg, "page") == 0)
	    {
	      set_server_type (SERVER_TYPE_PAGE);
	    }
```

src/base/server_type_enum.hpp
```diff
@@ -33,4 +33,30 @@ enum class server_type_config
...
 };
+
+inline const char *
+server_type_to_string (SERVER_TYPE type)
+{
+  switch (type)
+    {
+    case SERVER_TYPE_PAGE:
+      return "page";
+    default:
```
ribram, line 51: Fill in all types. Add missing types "unknown" and "any".

ribram (commented):
- Also replace `"page"` and `"transaction"` references in `server.c:argument_handler()` with `server_type_to_string()` calls.

ribram (approved):

src/base/server_type_enum.hpp
```diff
@@ -33,4 +33,34 @@ enum class server_type_config
...
+  switch (type)
+    {
+    case SERVER_TYPE_UNKNOWN:
+      return  "unknown";
+    case SERVER_TYPE_TRANSACTION:
+      return "transaction";
+    case SERVER_TYPE_PAGE:
+      return "page";
+    case SERVER_TYPE_ANY:
+      return "any";
```
cristiarg, line 50:
```suggestion
      return "any";
    default:
      assert (false);
```
to guard that future modifications of the enum do not leave this code behind (without `default` would still be signalled by a compiler warning but those do tend to be overlooked)

src/base/server_type_enum.hpp
```diff
@@ -33,4 +33,34 @@ enum class server_type_config
...
   SINGLE_NODE,
 };
+
+inline const char *
+server_type_to_string (SERVER_TYPE type)
+{
+  switch (type)
+    {
+    case SERVER_TYPE_UNKNOWN:
+      return  "unknown";
```
cristiarg:
```suggestion
      return "unknown";
```

src/base/server_type_enum.hpp
```diff
@@ -33,4 +33,34 @@ enum class server_type_config
...
+server_type_config_to_string (server_type_config type)
+{
+  switch (type)
+    {
+    case server_type_config::PAGE:
+      return "page";
+    case server_type_config::TRANSACTION:
+      return "transaction";
+    case server_type_config::SINGLE_NODE:
+      return "single_node";
```
cristiarg, line 66: any, default?

src/base/server_type_enum.hpp
```diff
@@ -33,4 +33,34 @@ enum class server_type_config
...
+    case SERVER_TYPE_TRANSACTION:
+      return "transaction";
+    case SERVER_TYPE_PAGE:
+      return "page";
+    case SERVER_TYPE_ANY:
+      return "any";
+    }
+}
+
+inline const char *
```
cristiarg: Not sure, but I think both functions can be made `constexpr` which would help the compiler directly replace the call to function with the result at compile time. There are a few usages where the parameter is a `constexpr`-like expression already. `constexpr` implies `inline const`.

cristiarg (approved):

PR #3031, opened 2021-09-22
[LETS-224] transaction server shutdown flag

cristiarg:
http://jira.cubrid.org/browse/LETS-224

In order to implement a recovery mechanism on TSRS, it is needed to know whether the TSRS (or any server, for that matter) has been clean-closed or abnormally.

- because TSRS has no other local storage aside from meta log - `cublob::meta` - a boolean flag is added to `cublog::meta` with the same semantics as `log_Gl.hdr.is_shutdown`
- read from file and initialize in `log_initialize_internal` - done by the call to `log_read_metalog_from_file`
- once server booting has started, set the flag to false 
- in `log_final` set the flag to `true` and persist metalog once more
- add a temporary log in `log_initialize_internal` to specify whether the server has been clean closed or crashed

### DIFF
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -1847,6 +1847,8 @@ log_final (THREAD_ENTRY * thread_p)
 
   free_and_init (log_Gl.loghdr_pgptr);
 
+  log_Log_header_initialized = false;
+
   LOG_CS_EXIT (thread_p);
 }
 
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -1410,14 +1410,10 @@ log_initialize_internal (THREAD_ENTRY * thread_p, const char *db_fullname, const
   log_Gl.rcv_phase = LOG_RESTARTED;
 
   // PREREQ: metalog is loaded at this point
-  if (is_tran_server_with_remote_storage ())
+  if (log_Gl.m_metainfo.get_clean_shutdown () && is_tran_server_with_remote_storage ())
     {
-      er_log_debug (ARG_FILE_LINE,
-		    "TEMP: log_initialize_internal: transaction server with remote storage clean shutdown: %s\n",
-		    log_Gl.m_metainfo.get_is_tsrs_shutdown ()? "yes" : "no");
-
-      // mark that transaction server with remote storage is running and save the meta log to persist the value
-      log_Gl.m_metainfo.set_is_tsrs_shutdown (false);
+      // mark, if needed, that transaction server with remote storage is running and persist the value
+      log_Gl.m_metainfo.set_clean_shutdown (false);
 
       const int res_metalog_to_file = log_write_metalog_to_file ();
       if (res_metalog_to_file != NO_ERROR)
@@ -1750,7 +1746,7 @@ log_final (THREAD_ENTRY * thread_p)
   // mark and persist that transaction server with remote storage has been correctly closed
   if (is_tran_server_with_remote_storage ())
     {
-      log_Gl.m_metainfo.set_is_tsrs_shutdown (true);
+      log_Gl.m_metainfo.set_clean_shutdown (true);
 
       const int res_metalog_to_file = log_write_metalog_to_file ();
       if (res_metalog_to_file != NO_ERROR)
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -335,6 +335,8 @@ static void log_daemons_destroy ();
 extern int catcls_get_apply_info_log_record_time (THREAD_ENTRY * thread_p, time_t * log_record_time);
 #endif /* SERVER_MODE */
 
+static bool log_Log_header_initialized = false;
+
 /*
  * log_rectype_string - RETURN TYPE OF LOG RECORD IN STRING FORMAT
  *
@@ -1084,19 +1086,21 @@ log_initialize_internal (THREAD_ENTRY * thread_p, const char *db_fullname, const
   /* Make sure that the log is a valid one */
   LOG_SET_CURRENT_TRAN_INDEX (thread_p, LOG_SYSTEM_TRAN_INDEX);
 
-  LOG_CS_ENTER (thread_p);
-
-  if (log_Gl.trantable.area != NULL)
-    {
-      log_final (thread_p);
-    }
-
   /* Initialize log name for log volumes */
   error_code = logpb_initialize_log_names (thread_p, db_fullname, logpath, prefix_logname);
   if (error_code != NO_ERROR)
     {
-      logpb_fatal_error (thread_p, !init_emergency, ARG_FILE_LINE, "log_xinit");
+      logpb_fatal_error (thread_p, !init_emergency, ARG_FILE_LINE,
+			 "log_initialize_internal: logpb_initialize_log_names");
       goto error;
+    }
+
+  LOG_CS_ENTER (thread_p);
+
+  /* Make sure that we are starting from a clean state */
+  if (log_Gl.trantable.area != NULL)
+    {
+      log_final (thread_p);
     }
 
   error_code = log_read_metalog_from_file ();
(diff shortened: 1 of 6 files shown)


src/transaction/log_meta.hpp
```diff
@@ -35,11 +35,18 @@ namespace cublog
       ~meta () = default;
 
       void load_from_file (std::FILE *stream);       // load meta from meta log file
+      // TODO: function can be called from different threads (main, daemon); must be synched ???
```
cristiarg: Is this remark worthwhile?
The code is definitely called from at least 2 different threads, but I am not sure about the timing.
From the "main" thread the code is mainly called in the boot-up/down sequences.
While from the daemon, it is called during regular operation of the server.
There might not be overlap between the two.

src/transaction/log_meta.hpp
```diff
@@ -50,6 +57,14 @@ namespace cublog
     private:
       using checkpoint_container_t = std::map<log_lsa, checkpoint_info>;
 
+    private:
+      /* flag parallel to 'log global header is shutdown':
+       *  - false: server has not been clean shut down
+       *  - true: has been clean shut down
+       * after start, flag is set to 'false' and the meta log is saved
+       */
+      bool m_is_tsrs_shutdown = false;
```
cristiarg: Naming is not ideal. It is intended to be different than `log_Gl.hdr.is_shutdown`.

src/transaction/log_page_buffer.c
```diff
@@ -7359,26 +7362,26 @@ logpb_checkpoint_trantable (THREAD_ENTRY * const thread_p)
 	_er_log_debug (ARG_FILE_LINE, "checkpoint_trantable: failed; writing metalog to file\n");
 	return res_metalog_to_file;
       }
+  }
 
-    logpb_flush_pages (thread_p, &trantable_checkpoint_lsa);
+  logpb_flush_pages (thread_p, &trantable_checkpoint_lsa);
```
cristiarg: I explained in the PR text about this change.
Is it safe?
If not, why?
If it is, why?

src/transaction/log_manager.c
```diff
@@ -1729,6 +1747,19 @@ log_final (THREAD_ENTRY * thread_p)
 
   save_tran_index = LOG_FIND_THREAD_TRAN_INDEX (thread_p);
 
+  // mark and persist that transaction server with remote storage has been correctly closed
+  if (is_tran_server_with_remote_storage ())
```
cristiarg: Added section here because, on a TSRS, the functions, early-outs upon ~1770:
```
if (log_Gl.append.vdes == NULL_VOLDES)
```
It might be needed/better to add a TSRS-specific `log_final` function and fork in the caller.

src/transaction/log_page_buffer.c
```diff
@@ -7323,6 +7323,9 @@ logpb_checkpoint_trantable (THREAD_ENTRY * const thread_p)
       return ER_FAILED;
     }
 
+  // *INDENT-OFF*
+  log_lsa trantable_checkpoint_lsa { NULL_LSA };
+  // *INDENT-ON*
```
ribram: Doing an assignment instead of initializer avoids disabled indentation.
```suggestion
  log_lsa trantable_checkpoint_lsa = NULL_LSA;
```

src/transaction/log_meta.hpp
ribram: My suggestion is to remove `_tsrs` from the name. Although it is used in the context of transaction server with remote storage for now, the flag may replace `log_Gl.hdr.is_shutdown` completely. The name of a variable should generally reflect its function, not the context that it is used in.
```suggestion
      bool m_is_shutdown = false;
```

src/transaction/log_page_buffer.c
ribram: LOG_CS should be locked again after logpb_flush_pages. Please fix it in a separate patch.

src/transaction/log_manager.c
```diff
@@ -1729,6 +1747,19 @@ log_final (THREAD_ENTRY * thread_p)
...
+    {
+      log_Gl.m_metainfo.set_is_tsrs_shutdown (true);
+
+      const int res_metalog_to_file = log_write_metalog_to_file ();
+      if (res_metalog_to_file != NO_ERROR)
+	{
+	  ASSERT_ERROR ();
+	  _er_log_debug (ARG_FILE_LINE, "log_final: failed writing metalog to file\n");
+	}
```
ribram: This is a point where nothing can be done if the meta log cannot be written to disk. How to accept that is_shutdown is not reflected back to disk?

An error may be expected and handled during database creation (e.g. folder path does not exist or permission is denied).

My suggestion is to change `log_write_metalog_to_file ()` by adding `assert (false);` if the file could not be open and to return void.

src/transaction/log_manager.c
```diff
@@ -1409,6 +1409,24 @@ log_initialize_internal (THREAD_ENTRY * thread_p, const char *db_fullname, const
     }
   log_Gl.rcv_phase = LOG_RESTARTED;
 
+  // PREREQ: metalog is loaded at this point
+  if (is_tran_server_with_remote_storage ())
+    {
+      er_log_debug (ARG_FILE_LINE,
+		    "TEMP: log_initialize_internal: transaction server with remote storage clean shutdown: %s\n",
+		    log_Gl.m_metainfo.get_is_tsrs_shutdown ()? "yes" : "no");
+
```
ribram: Remove this message; if shutdown was not properly executed and recovery is done, a message should already be logged.
```suggestion
```

src/transaction/log_manager.c
```diff
@@ -1409,6 +1409,24 @@ log_initialize_internal (THREAD_ENTRY * thread_p, const char *db_fullname, const
     }
   log_Gl.rcv_phase = LOG_RESTARTED;
 
+  // PREREQ: metalog is loaded at this point
+  if (is_tran_server_with_remote_storage ())
```
```suggestion
  if (!log_Gl.m_metainfo.is_tsrs_shutdown () && is_tran_server_with_remote_storage ())
```

src/transaction/log_manager.c

src/transaction/log_manager.c
cristiarg: Added [LETS-226](http://jira.cubrid.org/browse/LETS-226) for above comment.

src/transaction/log_manager.c
cristiarg: Isn't it even better as:
```suggestion
  if (log_Gl.m_metainfo.is_tsrs_shutdown () && is_tran_server_with_remote_storage ())
```
..in which case the writing of the metalog would be avoided if flag is already `false`?

src/transaction/log_manager.c
cristiarg: `log_write_metalog_to_file` is used also, as of now at least, by the TSRS checkpoint trantable daemon.
There very well might appear transient errors and not all of them are fatal (see second call to write metalog in function `logpb_checkpoint_trantable`).
Even in this situation, if the flag is not properly written due to a transient error, the worst that can happen upon restart is a speedy recovery being executed.

ribram: What kind of transient errors could prevent opening a file? If there are such errors, maybe they have to be treated.

If the metalog cannot be written at any moment, for instance in the first call in `logpb_checkpoint_trantable`, it could have more serious consequences. I think an assert or a fatal error is appropriate, even if in some circumstances, like writing the is_clean_shutdown flag, it could be accepted. It also simplifies things a bit, because the returned error no longer has to be handled (how can it be handled anyway in other way than ignoring it?).

src/transaction/log_manager.c
```diff
@@ -1736,7 +1750,7 @@ log_final (THREAD_ENTRY * thread_p)
       return;
     }
 
-  if (log_Gl.append.vdes == NULL_VOLDES)
+  if (!log_Log_header_initialized)
```
cristiarg, line 1753: I don't like, nor do I fully understand this but it works.
Tested:
- createdb, deletedb
- normal mode server execution
- TSRS/PS execution

ribram (approved):

PR #3032, opened 2021-09-22
[LETS-197] Amend fileio_get_volume_label usages for TSRS

ribram:
http://jira.cubrid.org/browse/LETS-197

Categories of `fileio_get_volume_label ()` usages:

1. Logging error and messages to server error log file
2. Inaccessible on TSRS because they're part of disabled operations.
3. Need to be skipped on TSRS.

Most changes in this patch fall into the first category. A wrapper functions was added on top `fileio_get_volume_label ()` to change null pointer into the "UNKNOWN" string.

Only one unhandled case was found in the third category - `boot_check_permanent_volumes ()` - and it was disabled on TSRS.

All remaining unchanged cases fall into the second category.


src/storage/file_io.c
```diff
@@ -6323,6 +6323,13 @@ fileio_get_volume_label_by_fd (int vol_fd, bool is_peek)
   return fileio_get_volume_label (fileio_get_volume_id (vol_fd), is_peek);
 }
 
+const char *
+fileio_get_volume_label_with_unknown (VOLID volid)
+{
+  static constexpr char *UNKNOWN_VLABEL = "(UNKNOWN)";
+  const char *vlabel = fileio_get_volume_label (volid, PEEK);
```
cristiarg, line 6330: I would argue that the extra param is not needed. There are two usecases where the opposite of `PEEK` - `ALLOC_COPY` - is used, but they are not needed since the value is copied again where needed anyway (`logpb_delete` and `log_unformat_ahead_volumes`).
Therefore it can be removed. But perhaps not in the scope of this PR.
Also, `fileio_get_volume_label` can return `const char *`.

src/transaction/boot_sr.c
```diff
@@ -3825,9 +3827,12 @@ xboot_check_db_consistency (THREAD_ENTRY * thread_p, int check_flag, OID * oids,
   bool repair = check_flag & CHECKDB_REPAIR;
   int error_code = NO_ERROR;
 
-  disk_lock_extend ();
-  error_code = boot_check_permanent_volumes (thread_p);
-  disk_unlock_extend ();
+  if (!is_tran_server_with_remote_storage ())
```
cristiarg, line 3849: Should this function be called in TSRS context?

cristiarg (approved):

src/storage/file_io.c
ribram: Yeah, I thought of it too, but it was not in the scope of this PR. Without removing the `ALLOC_COPY` argument, `fileio_get_volume_label` cannot return `const char *`.

src/transaction/boot_sr.c
ribram: Yes, it should be called in TSRS context. TSRS cannot do `boot_check_permanent_volumes ()` because this function verifies information in database is matching the information in file cache; but TSRS does not keep any files in the file cache.

The other checks verify information in the database is consistent (e.g. data in heap files is consistent with data in indexes). These checks can be done only on TS, because PS is clueless and cannot interpret the information in its data pages.

cristiarg: So the question was clueless. My bad.

PR #3033, opened 2021-09-22
[LETS-194] logpb_flush_pages needs to be called in critical section-free context

cristiarg:
http://jira.cubrid.org/browse/LETS-194

Follow-up patch for LETS-194.
Function `logpb_flush_pages` specifically needs to be called in critical section-free context.
Exit CS right before calling the function and re-acquire afterwards.

### DIFF
--- a/src/transaction/log_page_buffer.c
+++ b/src/transaction/log_page_buffer.c
@@ -7323,9 +7323,7 @@ logpb_checkpoint_trantable (THREAD_ENTRY * const thread_p)
       return ER_FAILED;
     }
 
-  // *INDENT-OFF*
-  log_lsa trantable_checkpoint_lsa { NULL_LSA };
-  // *INDENT-ON*
+  log_lsa trantable_checkpoint_lsa = NULL_LSA;
   {
     LOG_CS_ENTER (thread_p);
     // *INDENT-OFF*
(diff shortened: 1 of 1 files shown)


src/transaction/log_page_buffer.c
```diff
@@ -7323,6 +7323,9 @@ logpb_checkpoint_trantable (THREAD_ENTRY * const thread_p)
       return ER_FAILED;
     }
 
+  // *INDENT-OFF*
+  log_lsa trantable_checkpoint_lsa { NULL_LSA };
+  // *INDENT-ON*
```
ribram:
```suggestion
  log_lsa trantable_checkpoint_lsa = NULL_LSA;
```

src/transaction/log_page_buffer.c
```diff
@@ -7359,15 +7362,26 @@ logpb_checkpoint_trantable (THREAD_ENTRY * const thread_p)
 	_er_log_debug (ARG_FILE_LINE, "checkpoint_trantable: failed; writing metalog to file\n");
 	return res_metalog_to_file;
       }
+  }
 
-    logpb_flush_pages (thread_p, &trantable_checkpoint_lsa);
+  // function explicitely needs to be called in critical section-free context
+  logpb_flush_pages (thread_p, &trantable_checkpoint_lsa);
```
ribram: It would be easier to:
```suggestion
  LOG_CS_EXIT (thread_p);
  logpb_flush_pages (thread_p, &trantable_checkpoint_lsa);
  LOG_CS_ENTER (thread_p);
```
Then making a new scope and repeating scope_exit code would become unnecessary.

src/transaction/log_page_buffer.c
cristiarg: Done

PR #3034, opened 2021-09-23
[CBRD-24061] The problem that the TO_CHAR() function is removed even though there is a second argument

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-24061

If the TO_CHAR () function has a second argument, it is modified so that it is not deleted even if the first argument is a STRING type.

### DIFF
--- a/src/parser/type_checking.c
+++ b/src/parser/type_checking.c
@@ -9041,7 +9041,7 @@ pt_eval_expr_type (PARSER_CONTEXT * parser, PT_NODE * node)
       break;
 
     case PT_TO_CHAR:
-      if (PT_IS_CHAR_STRING_TYPE (arg1_type))
+      if (PT_IS_CHAR_STRING_TYPE (arg1_type) && PT_IS_NULL_NODE (arg2))
 	{
 	  arg1->line_number = node->line_number;
 	  arg1->column_number = node->column_number;


ctshim (approved):

beyondykk9:
`select /*+ recompile */ to_char(nvl(d_date,''), 'yyyymmddhh24miss') from t_date;
`
==> transformed
`select  to_char( cast(nvl(t_date.d_date, '') as double), 'yyyymmddhh24miss') from t_date t_date;
`
is correct transform?
==>
shparkcubrid
yes.
The first argument of to_char() must be DATE type or NUMBER type. If the first argument is STRING type, it tries to convert to NUMBER type as double.

beyondykk9 (approved):

hgryoo:
In test_sql CI, this simple case seems not to return an error.
```
** Failed query #1 (in bug_bts_5593.sql): 
select to_char('1999-11-11','YYYY-MM-DD'), 'abc';
** Difference between Expected(-) and Actual(+) results: 
@@ -1,4 +1,2 @@ 
===================================================
-to_char('1999-11-11', 'YYYY-MM-DD') 'abc' 
-1999-11-11 abc 
-
+Error:-494
```
==>
shparkcubrid
The first argument of to_char() must be DATE or NUMBER .
  ```
to_char('1999-11-11','YYYY-MM-DD')
```
The above statement also throws an error in Oracle. In the above case, to_date() function is needed to use.

hgryoo:
> In test_sql CI, this simple case seems not to return an error.
> 
> ```
> ** Failed query #1 (in bug_bts_5593.sql): 
> select to_char('1999-11-11','YYYY-MM-DD'), 'abc';
> ** Difference between Expected(-) and Actual(+) results: 
> @@ -1,4 +1,2 @@ 
> ===================================================
> -to_char('1999-11-11', 'YYYY-MM-DD') 'abc' 
> -1999-11-11 abc 
> -
> +Error:-494
> ```

I've just understood this regression from the JIRA issue.

hgryoo (approved):

youngjinj (approved):

PR #3041, opened 2021-09-27
[LETS-229] Fix start redo LSA from checkpoint info

ribram:
http://jira.cubrid.org/browse/LETS-229

`checkpoint_info::recovery_analysis ()` does not updated start_redo_lsa argument. The arguments remains uninitialized and redo recovery crashes.

Fix by setting start_redo_lsa properly.

NOTE: a recent fix surfaced this hidden issue; `recovery_analysis` used to be skipped completely.

### DIFF
--- a/src/transaction/log_checkpoint_info.cpp
+++ b/src/transaction/log_checkpoint_info.cpp
@@ -265,6 +265,8 @@ namespace cublog
   {
     LOG_TDES *tdes = nullptr;
 
+    start_redo_lsa = m_start_redo_lsa;
+
     /* Add the transactions to the transaction table */
     for (const auto &chkpt : m_trans)
       {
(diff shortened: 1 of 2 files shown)


src/transaction/log_checkpoint_info.cpp
```diff
@@ -265,6 +265,8 @@ namespace cublog
   {
     LOG_TDES *tdes = nullptr;
 
+    start_redo_lsa = m_start_redo_lsa;
```
cristiarg, line 268: an assert can be added to `get_start_redo_lsa` as well:
```
  log_lsa
  checkpoint_info::get_start_redo_lsa () const
  {
    assert (!m_start_redo_lsa.is_null ());

    return m_start_redo_lsa;
  }
```
unit tests will pass

cristiarg (approved):

PR #3042, opened 2021-09-28
[CBRD-23842] Fix the bug that can not see the name of auto increment serial

joohok:

when update in serial that is made with auto increment record, there is a bug that name of the serial can not be seen. 
The serial name is generated at first insert statement, and it is recorded in the db_serial record.
So it also need to be set serial name when second insert statement has done for putting serial name in the supplemental log.

### DIFF
--- a/src/storage/heap_file.c
+++ b/src/storage/heap_file.c
@@ -16598,6 +16598,7 @@ heap_set_autoincrement_value (THREAD_ENTRY * thread_p, HEAP_CACHE_ATTRINFO * att
 		}
 
 	      SET_AUTO_INCREMENT_SERIAL_NAME (serial_name, classname, attr_name);
+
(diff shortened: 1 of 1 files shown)


src/storage/heap_file.c
```diff
@@ -16598,61 +16598,63 @@ heap_set_autoincrement_value (THREAD_ENTRY * thread_p, HEAP_CACHE_ATTRINFO * att
 		}
 
 	      SET_AUTO_INCREMENT_SERIAL_NAME (serial_name, classname, attr_name);
-
-	      if (string != NULL && alloced_string == 1)
+	      if (OID_ISNULL (&serial_obj_oid))
```
hornetmj, line 16602:
```suggestion

	      if (OID_ISNULL (&serial_obj_oid))
```

hornetmj (approved):

PR #3043, opened 2021-09-28
[CBRD-23842] cubrid_log_connect_server() after cubrid_log_finalize() doesn't work

joohok:

when cubrid_log_connect_server() called right after cubrid_log_finalize(), 
segfault occurs because of the finalization of trace log file is not done properly.


hornetmj (approved):
Please, do the test of the reported test scenario before merging your code.

PR #3044, opened 2021-09-28
[CBRD-24059] Add new system catalog class for SYNONYM

youngjinj:
https://jira.cubrid.org/browse/CBRD-24059

Purpose
- Add system class/view for synonym management.
- Implements the length of the ALTER/CREATE/DROP/RENAME SYNONYM statement.
- Synonym names can be used instead of class/vclass names.
  - Do not use in statements that can manipulate the target table.
  - Use is prohibited in the "ALTER/CREATE INDEX/DROP/TRUNCATE" statement.
- For synonyms, feature/user_schema must be merged first.
- PUBLIC SYNONYM is not supported.
- Users cannot own private synonym names and class/vclass names of the same name.

Implementation
- Add system class/view for synonym management.
  - In the system class, synonyms are managed using unique_name.
    - unique_name is "owner_name.synonym_name".
  - _db_synonym
  - db_synonym
- Execute on ALTER/CREATE/DROP/RENAME SYNONYM statement.
  - Before semantic check, unique_name is created in - pt_set_user_specified_name() function.
  - Semantic check is performed in the function below.
```
/* src/parser/semantic_check.c */

static void pt_check_alter_synonym (PARSER_CONTEXT * parser, PT_NODE * node);
static void pt_check_create_synonym (PARSER_CONTEXT * parser, PT_NODE * node);
static void pt_check_drop_synonym (PARSER_CONTEXT * parser, PT_NODE * node);
static void pt_check_rename_synonym (PARSER_CONTEXT * parser, PT_NODE * node);
```
  - Apply to the system class in the function below.
```
/* src/query/execute_statement.c */

static int do_alter_synonym_internal (const char *synonym_name, const char *target_name, DB_OBJECT * target_owner, const char *comment, const int is_public_synonym);
static int do_create_synonym_internal (const char *synonym_name, DB_OBJECT * synonym_owner, const char *target_name, DB_OBJECT * target_owner, const char *comment, const int is_public_synonym, const int or_replace);
static int do_drop_synonym_internal (const char *synonym_name, const int is_public_synonym, const int if_exists, DB_OBJECT * synonym_class_obj, DB_OBJECT * synonym_obj);
static int do_rename_synonym_internal (const char *old_synonym_name, const char *new_synonym_name);
```
- Synonym names can be used instead of class/vclass names.
  - If it is a synonym name in the semantic check of the "ALTER/CREATE INDEX/DROP/RENAME/TRUNCATE" statement, execution is stopped.
- Users cannot own private synonym names and class/vclass names of the same name.
  - When creating class/vclass, if there is a synonym name with the same name, it stops.

Remarks
- N/A


youngjinj:
### Analysis of ci/circleci test results
* As a result of creating _db_syonym and db_synonym, the data of _db_class, db_class, and db_auth was changed.
* So, 1 test_medium and 28 test_sql failed.
* The answer in the test case should change.

The list of test cases is commented on CBRD-24059.

shparkcubrid (approved):

src/object/transform.c
```diff
@@ -413,6 +413,15 @@ static CT_ATTR ct_partition_atts[] = {
   {"comment", NULL_ATTRID, DB_TYPE_VARCHAR}
 };
 
+static CT_ATTR ct_synonym_atts[] = {
+  {"synonym_name", NULL_ATTRID, DB_TYPE_VARCHAR},
+  {"synonym_owner_name", NULL_ATTRID, DB_TYPE_VARCHAR},
```
ctshim: Wouldn't it be better to specify db_user instead of a string?
`{"synonym_owner_name", NULL_ATTRID, DB_TYPE_OBJECT},`

When creating it, you can do it like the code below.
`error_code = smt_add_attribute(def, "synonym_owner_name", AU_USER_CLASS_NAME, NULL);`

beyondykk9 (approved):

hyunikn (approved):

src/object/transform.c
youngjinj: I wanted to avoid using the db_user type because it is OID dependent.
Even if data is managed by OID internally,
I want to use the data type used by the sql standard for the value stored in the table.

If the synonym_owner_name column type becomes db_user type,
the user must write a subquery for db_user and use the where condition.

Even if it is not an OID value, the owner_name is also a unique value,
so it is not necessary to set it as a db_user type.

youngjinj: It seems that DB_TYPE_OBJECT type is used to have the same effect on columns that require foreign key setting in system tables.

This is because when a foreign key is set,
the order of system table creation and deletion depends on the foreign key.

If so, it seems appropriate to use the DB_TYPE_OBJECT type.

youngjinj: It seems that DB_TYPE_OBJECT type is used to have the same effect on columns that require foreign key setting in system tables.

This is because when a foreign key is set,
The order of system table creation and deletion is affected by foreign keys.

hgryoo (approved):

src/transaction/boot_cl.c
```diff
@@ -5444,6 +5532,77 @@ boot_define_view_db_charset (void)
...
+	  return error_code;
+	}
+    }
+
+  /* Use ORDER BY to get PRIVATE SYNONYM first. */
+  sprintf (stmt, "SELECT [s].[name], CAST([s].[owner].[name] AS VARCHAR(255)), "
+	   "CASE WHEN [s].[is_public] = 1 THEN 'YES' ELSE 'NO' END, "
+	   "[s].[target_name], CAST([s].[target_owner].[name] AS VARCHAR(255)), [s].[comment] "
+	   "FROM [%s] [s] "
+	   "WHERE (CURRENT_USER = 'DBA') "
```
ctshim: Shouldn't DBA group members be able to see it too?
Please refer to the code when creating a view related to a trigger or method.

ctshim (approved):

src/query/execute_statement.c
```diff
@@ -17246,6 +17262,301 @@ do_set_timezone (PARSER_CONTEXT * parser, PT_NODE * statement)
...
+	  target_owner = synonym_owner;
+	}
+
+      /* is_public_synonym
+       *  - Only DBA users and members of the DBA group can create public synonyms.
+       *  - Otherwise, an error occurs.
+       */
+      if (statement->info.create_synonym.access_modifier == PT_PUBLIC)
+	{
+	  if (au_is_dba_group_member (db_get_user ()))	/* current user */
```
hyunikn: It seems better to use synonym_owner instead of calling db_get_user() again.

hyunikn (approved):

src/query/execute_statement.c
youngjinj: Thank you. I will change it as you said.

src/transaction/boot_cl.c
youngjinj: Thank you. I will change it as you said.

src/object/schema_manager.c
```diff
@@ -3008,6 +3008,124 @@ sm_is_system_class (MOP op)
...
+    CTV_STORED_PROC_ARGS_NAME,	// "db_stored_procedure_args"
+    CTV_DB_COLLATION_NAME,	// "db_collation"
+    CTV_DB_CHARSET_NAME,	// "db_charset"
+    CTV_SYNONYM_NAME,		// "db_synonym"
+
+    NULL
+  };
+
+  /* Rootclass */
+  if (strncmp (name, ROOTCLASS_NAME, SM_MAX_IDENTIFIER_LENGTH) == 0)
```
beyondykk9: This looks not necessary because it is checked at the below while loop.

src/object/schema_manager.c
```diff
@@ -3008,6 +3008,124 @@ sm_is_system_class (MOP op)
...
+ *    the same as the system class name.
+ * return: bool
+ * name(in): class name
+ */
+bool
+sm_is_system_class_by_name (const char *name)
+{
+  const char **ptr = NULL;
+
+  const char *system_classes[] = {
```
beyondykk9: this const variable will be allocated and initialized at stack area by every calling, it looks better to move to the static global variable.

src/object/authenticate.c
```diff
@@ -8717,6 +8718,49 @@ au_check_serial_authorization (MOP serial_object)
...
+  MOP creator;
+  DB_SET *groups;
+  int ret_val;
+
+  CHECK_1ARG_ERROR (synonym_object);
+
+  ret_val = db_get (synonym_object, "owner", &creator_val);
+
+  if (ret_val != NO_ERROR || DB_IS_NULL (&creator_val))
+    {
```
beyondykk9, line 9295: it looks necessary "er_set" as like below er_set (ER_WARNING_SEVERITY, ARG_FILE_LINE, ret_val, 0);

src/query/execute_statement.c
```diff
@@ -17246,6 +17262,651 @@ do_set_timezone (PARSER_CONTEXT * parser, PT_NODE * statement)
...
+      sm_downcase_name (synonym_name, synonym_downcase_name, SM_MAX_IDENTIFIER_LENGTH);
+
+      /* cannot use catalog class names as synonym names. */
+      if (db_is_system_class_by_name (synonym_downcase_name))
+	{
+	  error = ER_QPROC_CANNOT_USE_CATALOG_NAME;
+	  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, error, 0);
+	  goto end;
+	}
+    }
```
beyondykk9, line 17762: it looks necessary "else" for producing error.

src/query/execute_statement.c
```diff
@@ -17246,6 +17262,651 @@ do_set_timezone (PARSER_CONTEXT * parser, PT_NODE * statement)
...
+
+  /* synonym_owner */
+  if (statement->info.create_synonym.synonym_owner_name)
+    {
+      assert (statement->info.create_synonym.synonym_owner_name->node_type == PT_NAME);
+
+      synonym_owner_name = statement->info.create_synonym.synonym_owner_name->info.name.original;
+      synonym_owner = au_find_user (synonym_owner_name);
+    }
```
beyondykk9, line 17782: it looks necessary "else" for producing error as prior case.

src/query/execute_statement.c
```diff
@@ -17246,6 +17262,651 @@ do_set_timezone (PARSER_CONTEXT * parser, PT_NODE * statement)
...
+	      goto end;
+	    }
+	  else
+	    {
+	      error = ER_QPROC_PRIVATE_SYNONYM_ALREADY_EXIST;
+	      er_set (ER_WARNING_SEVERITY, ARG_FILE_LINE, error, 2, synonym_owner_name, synonym_downcase_name);
+	      goto end;
+	    }
+	}
+    }
```
beyondykk9, line 17849: it looks necessary "else" clause for NULL handling

src/query/execute_statement.c
```diff
@@ -17246,6 +17262,651 @@ do_set_timezone (PARSER_CONTEXT * parser, PT_NODE * statement)
...
+
+  /* target_name */
+  if (statement->info.create_synonym.target_name)
+    {
+      assert (statement->info.create_synonym.target_name->node_type == PT_NAME);
+
+      target_name = statement->info.create_synonym.target_name->info.name.original;
+    }
```
beyondykk9: target_name can not be NULL?

src/query/execute_statement.h
```diff
@@ -159,6 +159,19 @@ extern int do_execute_merge (PARSER_CONTEXT * parser, PT_NODE * statement);
...
+extern int do_alter_synonym (PARSER_CONTEXT * parser, PT_NODE * statement);
+extern int do_create_synonym (PARSER_CONTEXT * parser, PT_NODE * statement);
+extern int do_drop_synonym (PARSER_CONTEXT * parser, PT_NODE * statement);
+extern int do_rename_synonym (PARSER_CONTEXT * parser, PT_NODE * statement);
+
+static int do_create_synonym_internal (const char *synonym_name, DB_OBJECT * synonym_owner,
+				       const int is_public_synonym, const char *target_name,
+				       DB_OBJECT * target_owner, const char *comment, const bool has_already);
+static int do_drop_synonym_internal (const char *synonym_name, const char *synonym_owner, const int is_public_synonym,
+				     const bool if_exists);
```
hgryoo: please move static function declarations to .c file.

src/query/execute_statement.c
```diff
@@ -17246,6 +17262,651 @@ do_set_timezone (PARSER_CONTEXT * parser, PT_NODE * statement)
...
+	  synonym_object = db_get_object (&value);
+	}
+    }
+  else
+    {
+      if (is_public)
+	{
+	  er_set (ER_WARNING_SEVERITY, ARG_FILE_LINE, ER_QPROC_PUBLIC_SYNONYM_NOT_FOUND, 2, owner_name, name);
+	  synonym_object = NULL;
+	  goto end;
```
beyondykk9: "synonym_object = NULL;  goto end" clause is common routine, it can be move to more outer.

src/query/execute_statement.c
```diff
@@ -17246,6 +17262,651 @@ do_set_timezone (PARSER_CONTEXT * parser, PT_NODE * statement)
...
+	}
+      else
+	{
+	  er_set (ER_WARNING_SEVERITY, ARG_FILE_LINE, ER_QPROC_PRIVATE_SYNONYM_NOT_FOUND, 2, owner_name, name);
+	  synonym_object = NULL;
+	  goto end;
+	}
```
hgryoo: suggested shorten version. The decision is yours.
```suggestion
      error = is_public ? ER_QPROC_PUBLIC_SYNONYM_NOT_FOUND : ER_QPROC_PRIVATE_SYNONYM_NOT_FOUND;
      er_set (ER_WARNING_SEVERITY, ARG_FILE_LINE, error, 2, owner_name, name);
	synonym_object = NULL;
	goto end;
```

src/query/execute_statement.c
```diff
@@ -17246,6 +17262,651 @@ do_set_timezone (PARSER_CONTEXT * parser, PT_NODE * statement)
...
+	{
+	  goto end;
+	}
+    }
+
+  error = do_create_synonym_internal (synonym_downcase_name, synonym_owner, is_public_synonym, target_name,
+				      target_owner, comment, has_already);
+
+end:
+  return error;
```
hgryoo: `end:` label seems meaningless. how about returning immediately above?

src/query/execute_statement.c
```diff
@@ -17246,6 +17262,651 @@ do_set_timezone (PARSER_CONTEXT * parser, PT_NODE * statement)
...
+	      error = ER_QPROC_PUBLIC_SYNONYM_ALREADY_EXIST;
+	      er_set (ER_WARNING_SEVERITY, ARG_FILE_LINE, error, 2, synonym_owner_name, synonym_downcase_name);
+	      goto end;
+	    }
+	  else
+	    {
+	      error = ER_QPROC_PRIVATE_SYNONYM_ALREADY_EXIST;
+	      er_set (ER_WARNING_SEVERITY, ARG_FILE_LINE, error, 2, synonym_owner_name, synonym_downcase_name);
+	      goto end;
+	    }
```
hgryoo: suggested shorten version. The decision is yours.

```suggestion
	  error = is_public_synonym ? ER_QPROC_PUBLIC_SYNONYM_ALREADY_EXIST : ER_QPROC_PRIVATE_SYNONYM_ALREADY_EXIST;
	  er_set (ER_WARNING_SEVERITY, ARG_FILE_LINE, error, 2, synonym_owner_name, synonym_downcase_name);
	  goto end;
```

src/object/schema_manager.c
youngjinj: Thank you.
It would be better to change it to a global variable.

I will change it.

src/object/schema_manager.c
youngjinj: Thank you.
The check for ROOTCLASS_NAME is not included in the while loop below.
But ROOTCLASS_NAME looks good to check in while loop.

I will change it.

src/query/execute_statement.h
youngjinj: Thank you.

I will change it

src/query/execute_statement.c
youngjinj: Thank you.
Thank you.
Thank you.

I didn't think of the triple.
I will change it.

src/query/execute_statement.c
youngjinj: Thank you.

I will change it

src/query/execute_statement.c
youngjinj: I was trying to have a similar structure while coding several functions,
so I put end: .
It seems efficient to return immediately.

I will change it.

src/query/execute_statement.c
youngjinj: This is the part i agreed to change in the previous comment.

I will change it

src/object/authenticate.c
youngjinj: Thank you.

I will change it

src/query/execute_statement.c
youngjinj: Thank you.

I will change it

src/query/execute_statement.c
youngjinj: Thank you.

I will change it

src/query/execute_statement.c
youngjinj: This is the part that checks if it already exists before creating a synonym.
If the synonym does not exist, go ahead and create a new synonym.

Is there any necessary handling for the else?

src/query/execute_statement.c
youngjinj: Yes target_name cannot be NULL.

I'll add handling for the else.

src/query/execute_statement.c
```diff
@@ -17246,6 +17262,651 @@ do_set_timezone (PARSER_CONTEXT * parser, PT_NODE * statement)
...
+      if (target_owner == NULL)
+	{
+	  assert (er_errid () != NO_ERROR);
+	  error = er_errid ();
+	  goto end;
+	}
+    }
+  else
+    {
+      if (db_is_system_class_by_name (target_name))
```
ctshim: The parameter of the db_is_system_class_by_name() function must be guaranteed to be lowercase.
Use sm_downcase_name().

src/object/schema_manager.c
```diff
@@ -3008,6 +3008,124 @@ sm_is_system_class (MOP op)
...
+      return true;
+    }
+
+  /* dual */
+  if (strncmp (name, CT_DUAL_NAME, SM_MAX_IDENTIFIER_LENGTH) == 0)
+    {
+      return true;
+    }
+
+  if (strncmp (name, "_db_", sizeof ("_db_") - 1) != 0 && strncmp (name, "db_", sizeof ("db_") - 1) != 0)
```
ctshim: It's the same content, but I think it's easier to express it like the one below.

`if (strncmp (name, "_db_", 4) != 0 && strncmp (name, "db_", 3) != 0)`

src/query/execute_statement.c
beyondykk9: you can use assert for NULL if you assure it's not null.

src/query/execute_statement.c
beyondykk9: you can also use assert for NULL if you assure it's not null.

shparkcubrid (approved):

src/object/schema_manager.c
hyunikn: It seems better to declare it as 'static' at the current position without moving it to a global scope
because no other functions need to see 'system_classes' array. 
Static variables declared inside a function are also initialized just once.

src/query/execute_statement.c
```diff
@@ -17246,6 +17262,651 @@ do_set_timezone (PARSER_CONTEXT * parser, PT_NODE * statement)
...
+    {
+      assert (statement && statement->info.create_synonym.access_modifier == PT_PRIVATE);
+
+      if (synonym_owner == NULL)
+	{
+	  synonym_owner = current_user;
+	}
+
+      /* creating private synonym for other user can only be performed by DBA or a DBA group member. */
+      if (ws_is_same_object (current_user, synonym_owner) == false)
```
hyunikn: According to the above comment line, we need to also check here if the current user is a DBA or DBA group member.
If the current user is a DBA or DBA group member, then it is not an error, isn't it?

ctshim (approved):

src/object/authenticate.c
hyunikn: If ( ret_val == NO_ERROR && DB_IS_NULL(&creator_val) ) is true, then the following return statement returns NO_ERROR, which seems to be a bug.

src/query/execute_statement.c
```diff
@@ -17246,6 +17262,651 @@ do_set_timezone (PARSER_CONTEXT * parser, PT_NODE * statement)
...
+      synonym_owner = au_find_user (synonym_owner_name);
+    }
+
+  /* is_public_synonym */
+  if (statement->info.create_synonym.access_modifier == PT_PUBLIC)
+    {
+      /* cannot create public synonym for user. */
+      if (synonym_owner)
+	{
+	  error = ER_QPROC_CANNOT_EXIST_PUBLIC_SYNONYM;
```
hyunikn: sysnonym_owner is always set non-null at this point, and creating a public synonym always causes an error.
Maybe, this is not what you intend.

hyunikn: ER_QPROC_CANNOT_CREATE_PUBLIC_SYNONYM seems to be a more suitable name of the error code.

src/query/execute_statement.c
hyunikn: [Q] Does the else part returns an error or set the synonym_owner to the current user? Or, anything else?

src/query/execute_statement.c
```diff
@@ -17246,6 +17262,651 @@ do_set_timezone (PARSER_CONTEXT * parser, PT_NODE * statement)
...
+	{
+	  error = do_drop_synonym_internal (synonym_downcase_name, synonym_owner_name, is_public_synonym, if_exists);
+	  if (error < NO_ERROR)
+	    {
+	      goto end;
+	    }
+	}
+      else
+	{
+	  error = ER_QPROC_CANNOT_UPDATE_SYNONYM;
```
hyunikn: Returning the specific reason (authorization problem in this case) seems to be better.

src/query/execute_statement.c
```diff
@@ -17246,6 +17262,651 @@ do_set_timezone (PARSER_CONTEXT * parser, PT_NODE * statement)
...
+  int save = 0;
+
+  memset (&query_buf, 0, sizeof (char) * QUERY_BUF_SIZE);
+  memset (&query_error, 0, sizeof (DB_QUERY_ERROR));
+
+  query = "SELECT [%s] FROM [%s] WHERE [name] = '%s' AND [owner].[name] = '%s' AND [is_public] = %d LIMIT 1";
+  error = sprintf (query_buf, query, CT_SYNONYM_NAME, CT_SYNONYM_NAME, name, owner_name, is_public);
+  if (error < NO_ERROR)
+    {
+      er_set (ER_WARNING_SEVERITY, ARG_FILE_LINE, error, 0);
```
hyunikn: Is it OK to use er_set for an error code of sprintf?

src/query/execute_statement.c
```diff
@@ -17246,6 +17262,651 @@ do_set_timezone (PARSER_CONTEXT * parser, PT_NODE * statement)
...
+      goto end;
+    }
+
+  AU_DISABLE (save);
+
+  error = db_compile_and_execute_local (query_buf, &query_result, &query_error);
+
+  AU_DISABLE (save);
```
hyunikn: AU_DISABLE is done two lines above. Maybe, you mean AU_ENABLE at this line.

src/query/execute_statement.h
```diff
@@ -159,6 +159,19 @@ extern int do_execute_merge (PARSER_CONTEXT * parser, PT_NODE * statement);
 extern int do_set_names (PARSER_CONTEXT * parser, PT_NODE * statement);
 extern int do_set_timezone (PARSER_CONTEXT * parser, PT_NODE * statement);
 
+extern int do_alter_synonym (PARSER_CONTEXT * parser, PT_NODE * statement);
+extern int do_create_synonym (PARSER_CONTEXT * parser, PT_NODE * statement);
+extern int do_drop_synonym (PARSER_CONTEXT * parser, PT_NODE * statement);
+extern int do_rename_synonym (PARSER_CONTEXT * parser, PT_NODE * statement);
+
+static int do_create_synonym_internal (const char *synonym_name, DB_OBJECT * synonym_owner,
```
hyunikn: Static declarations should be moved to .c files, shouldn't they?

hyunikn (approved):

src/object/schema_manager.c
youngjinj: Names without prefix (\_db_, db_) are most likely not system classes.
In order to check only the prefix first, the system class name (Rootclass, dual) that does not have a prefix should be checked first.

So I will keep it as it is.

src/object/schema_manager.c
youngjinj: Thank you.
I'll put a description in the comments and use constants.

src/object/schema_manager.c
youngjinj: Thank you.
Its purpose was to make it initialize only once.
Using static looks better.

src/query/execute_statement.h
youngjinj: Yes, that's right. :sob:

I'll change it.

src/query/execute_statement.c
youngjinj: sprintf and snprintf may also cause errors.
So I do er_set() .

Is it excessive?

src/query/execute_statement.c
youngjinj: It's a mistake.
Thank you.

src/query/execute_statement.c
youngjinj: Thank you. I agree.

I'll change it.

src/query/execute_statement.c
hyunikn: I wonder whether er_set (and Cubrid error handling mechanism) understands sprintf's error code or not. (I guess not)
For example, if sprintf returns -255 and we call er_set with this code as the above code, then er_set regards it as ER_SM_ATTRIBUTE_NOT_VARIABLE and prints the erorr log corresponding to ER_SM_ATTRIBUTE_NOT_VARIABLE.
Is it OK to use an error code which does not appear in src/base/error_code.h when calling er_set?

src/query/execute_statement.c
youngjinj: Thank you.
No code to check if user is DBA or member of DBA group.

I'll change it.

src/object/authenticate.c
youngjinj: Call er_set inside the db_get function.
So, I will add only the code that checks the error code.

There seems to be no case where creator_val is NULL and NO_ERROR.

src/query/execute_statement.c
youngjinj: statement->info.create_synonym.synonym_owner_name can be NULL.
If it is NULL, it is set as the current user below.
No error handling is required.

src/query/execute_statement.c
youngjinj: Let's change the error message to be more explicit.

src/query/execute_statement.c
youngjinj: statement->info.create_synonym.synonym_name cannot be NULL.
So I put the code below.
```
assert (false);
```

src/query/execute_statement.c
youngjinj: Added the au_is_user_group_member function to check whether the user is an owner or a member of a group.

src/query/execute_statement.c
youngjinj: If the synonym does not exist, it is created, and if it already exists, or_replace is checked and has_already is set to true.

Even if it returns NULL, it is normal.

src/query/execute_statement.c
youngjinj: Yes, I will add the code below.
```
assert (false);
```

src/query/execute_statement.c
youngjinj: For now, I'm not going to do error handling for snprintf .
I will reflect on error handling after thinking more about it.

src/object/authenticate.c
```diff
@@ -3394,7 +3395,7 @@ au_drop_user (MOP user)
      * drop user command can be called only by DBA group,
      * so we can use query for _db_class directly
      */
-    "_db_class", "db_trigger", "db_serial", "_db_server", NULL
+    "_db_class", "db_trigger", "db_serial", "_db_server", "_db_synonym", NULL
```
hyunikn, line 3442: It seems that the 'class_name' array that stores these strings can be static. In other words, the array can be initialized just once.

src/object/authenticate.c
hyunikn: Then, the condition should be just "ret_val != NO_ERROR". The null check consumes unnecessary CPU clocks, though minor, and gives a wrong impression to the code reviewers that there are cases where ret_val is NO_ERROR and creator_val is NULL.

src/transaction/boot_cl.c
```diff
@@ -4014,6 +4016,91 @@ boot_define_dual (MOP class_mop)
   return NO_ERROR;
 }
 
+static int
+boot_define_synonym (MOP class_mop)
+{
+  SM_TEMPLATE *def;
+  int error_code = NO_ERROR;
+  const char *primary_key_col_names[] = { "name", "owner", "is_public", NULL };
```
hyunikn: This can be static and initialized just once.

hyunikn (approved):

src/object/authenticate.c
youngjinj: Okay. It looks good to add the static keyword.

src/query/execute_statement.c
youngjinj: The error message is more likely to change.
I also see the need to separate semantic check and execute code.
So the error message will be reviewed again later.

src/object/authenticate.c
youngjinj: It seems that there should be only the condition "ret_val != NO_ERROR".
However, other functions that start with "au_*" use the condition "ret_val == NO_ERROR && DB_IS_NULL(&creator_val)". 
This may need to be improved later.

src/transaction/boot_cl.c
youngjinj: The "boot_define_\*" functions are only called once.
So the static keyword is not needed.
For consistency with other "boot_define_\*" functions, I won't add the static keyword.

hyunikn: Aha!. I see.

beyondykk9 (approved):

src/parser/compile.c
```diff
@@ -746,7 +763,8 @@ pt_add_lock_class (PARSER_CONTEXT * parser, PT_CLASS_LOCKS * lcks, PT_NODE * spe
       return MSGCAT_RUNTIME_OUT_OF_MEMORY;
     }
 
-  sm_downcase_name (spec->info.spec.entity_name->info.name.original, lcks->classes[lcks->num_classes], len + 1);
+  memcpy (lcks->classes[lcks->num_classes], realname, len + 1);
+  lcks->classes[lcks->num_classes][len] = '\0';
```
ctshim: This code can be omitted as it is included when copying the length of len+1 in memcpy().

src/parser/parse_tree.h
```diff
@@ -3502,10 +3504,7 @@ union pt_statement_info
   PT_TRACE_INFO trace;
   PT_KILLSTMT_INFO killstmt;
   PT_WITH_CLAUSE_INFO with_clause;
-  PT_ALTER_SYNONYM_INFO alter_synonym;
-  PT_CREATE_SYNONYM_INFO create_synonym;
-  PT_DROP_SYNONYM_INFO drop_synonym;
-  PT_RENAME_SYNONYM_INFO rename_synonym;
+  PT_SYNONYM_INFO synonym;
```
ctshim: It is recommended to move the location after the PT_SPEC_INFO spec.
Existing contents are sorted in ascending order by name.

src/parser/semantic_check.c
```diff
@@ -8806,7 +9103,8 @@ pt_check_drop (PARSER_CONTEXT * parser, PT_NODE * node)
 	  /* check if class name exists. if not, we remove the corresponding node from spec_list. */
 	  if ((name = free_node->info.spec.entity_name) != NULL && name->node_type == PT_NAME
 	      && (cls_name = name->info.name.original) != NULL
-	      && (db_obj = db_find_class_with_purpose (cls_name, true)) == NULL)
+	      && ((db_obj = db_find_synonym (cls_name)) != NULL
```
hyunikn, line 9124: db_obj at this and the next lines are defined but not used

hyunikn (approved):

src/query/execute_statement.c
```diff
@@ -18712,6 +19381,91 @@ do_find_trigger_by_query (const char *name, char *buf, int buf_size)
...
+
+  if (name == NULL || name[0] == '\0')
+    {
+      ERROR_SET_WARNING (error, ER_OBJ_INVALID_ARGUMENTS);
+      return error;
+    }
+
+  assert (buf != NULL);
+
+  query = "SELECT LOWER ([target_owner].[name] || '.' || [target_name]) FROM [%s] WHERE [unique_name] = '%s'";
```
ctshim: I know the LOWER function isn't perfect for collation. Is there any problem?

src/parser/parse_tree.h
youngjinj: Changed the order.

src/parser/semantic_check.c
youngjinj: I'll keep it for consistency in the style of checking within the function.

src/query/execute_statement.c
youngjinj: I'll use target_unique_name .

src/parser/compile.c
youngjinj: I will remove the below.
```
lcks->classes[lcks->num_classes][len] = '\0';
```

shparkcubrid (approved):

hgryoo (approved):

PR #3045, opened 2021-09-28
[LETS-230] Allocate a transaction to make TDE changes after restore

ribram:
http://jira.cubrid.org/browse/LETS-230

`boot_reset_mk_after_restart_from_backup ()` operates some changes on the database and then executes commit without allocating a transaction. The commit log record is registered on the system transaction ID.

On the second restore, during analysis the system transaction descriptor is freed when the commit log record is analyzed and then recreated on the next system log record (e.g. from vacuum).

During recovery redo, an assert on commit log record fails because it expects that the transaction descriptor is already freed during analysis.

The problem is committing on system transaction descriptor. All changes that end with commit should be allocated a dedicated transaction.


ribram:
@hornetmj I think this fix must be backported to develop.

cristiarg (approved):

PR #3048, opened 2021-09-29
[CBRD-23842] Make log info for update when undo, redo has different representation

joohok:

when undo or redo record doesn't contain former representation,  cdc made empty log info to client. 

However, the update case when redo has more recent representation than undo record, it is required to make a log info.

ex) 
create table (a int) -> insert -> alter table add column b int-> update set b = ? where a = ?


hornetmj:
@joohok Please close this PR. We categorized this as a design issue. If we are ready to solve this problem, then we could discuss it in the other PR.

PR #3050, opened 2021-09-29
[CBRD-23842] DML on table partition returns OID of partition not a root class OID

joohok:

When CDC user extract a dml log item and make a sql statement with it, 
the schema information extracted with class OID is not able to be received with portition OID. 
So, if DML is operated on partition class, it returns root class(partitioned class) OID not a partition class OID.


src/transaction/log_manager.c
```diff
@@ -12245,6 +12245,21 @@ cdc_make_dml_loginfo (THREAD_ENTRY * thread_p, int trid, char *user, CDC_DML_TYP
   int record_length = 0;
 
   char *loginfo_buf = NULL;
+  OID partitioned_classoid;
```
```suggestion
  OID partitioned_classoid = OID_INITIALIZER;
```

hornetmj (approved):

PR #3056, opened 2021-10-01
[CBRD-24073] Problem that predicate push does not work when there is a correlated constant value in another subquery.

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-24073

The constant value of the select_list of the subquery is replaced with the equality term of the main query. Therefore, qo_reduce_equality_terms() must be called in post order.
```
select ...
 from (select col ==> 1 (1st replace)
          from ...
         where col = 1) a
 where a.col = b.col1 ==> 1 = b.col1 (2nd replace)
```
Only SELECT cases apply post order calls. For other cases, make changes later if necessary.

And the constant value of the select list is stored separately. This is used for substitution.


hyunikn (approved):

src/optimizer/query_rewrite.c
```diff
@@ -1307,6 +1307,16 @@ qo_reduce_equality_terms (PARSER_CONTEXT * parser, PT_NODE * node, PT_NODE ** wh
   bool copy_arg2;
   PT_NODE *dt1, *dt2;
 
+  /* check whether the function has already been called. */
+  if (node->flag.do_reduce_equality_terms)
+    {
+      return;
+    }
```
hgryoo: Instead of this, how about defining a macro function to avoid unnecessary function calls?
```
#define QO_CHECK_AND_REDUCE_EQUALITY_TERMS(parser, node, where)
  do {
      if (!node->flag.do_reduce_equality_terms)
      {
          qo_reduce_equality_terms (parser, node, where);
      }
  } while (0)
```

hgryoo (approved):

src/parser/parse_tree.h
```diff
@@ -3478,6 +3479,7 @@ struct parser_node
     unsigned is_wrapped_res_for_coll:1;	/* is a result node wrapped with CAST by collation inference */
     unsigned is_system_generated_stmt:1;	/* is internally generated by system */
     unsigned use_auto_commit:1;	/* use autocommit */
+    unsigned do_reduce_equality_terms:1;	/* reduce_equality_terms() is already called */
```
beyondykk9: how about **done_reduce_equality_terms** instead of **do_reduce_equality_terms** to inform already done  ?

beyondykk9 (approved):

youngjinj (approved):

ctshim (approved):

src/optimizer/query_rewrite.c
shparkcubrid: Okay. I'll fix it.

src/parser/parse_tree.h
shparkcubrid: Okay. I'll fix it.

PR #3069, opened 2021-10-05
[LETS-90] Move CSS connect functions to connection_support

adrianbrab:
http://jira.cubrid.org/browse/LETS-90

Combined the 2 variants of `css_common_connect` and `css_connect_to_master_server` present in `connection_cl.c` and `connection_sr.c` into a single variant that combines the functionality. 

Moved the new `css_common_connect` and `css_connect_to_master_server` to `connection_support.c`


src/connection/connection_support.c
```diff
@@ -2945,3 +2957,213 @@ css_conn_entry::init_pending_request ()
...
+  CSS_CONN_ENTRY *conn;
+  unsigned short rid;
+  int response, response_buff;
+  int server_port_id;
+  int connection_protocol;
+#if !defined(WINDOWS)
+  std::string pname;
+  int datagram_fd, socket_fd;
+#endif
+  css_Service_id = master_port_id;
```
ribram, line 2975:
```suggestion

  css_Service_id = master_port_id;
```

src/connection/connection_support.c
```diff
@@ -2945,3 +2957,213 @@ css_conn_entry::init_pending_request ()
...
+      // Linux and Unix
+      connection_protocol = SERVER_REQUEST;
+    }
+
+  if (css_common_connect
+      (conn, &rid, hname, connection_protocol, message_to_master, message_to_master_length, master_port_id, 0,
+       true) == NULL)
+    {
+      goto fail_end;
+
```
ribram:
```suggestion
```

src/connection/connection_support.c
```diff
@@ -2945,3 +2957,213 @@ css_conn_entry::init_pending_request ()
...
+#else /* WINDOWS */
+      /* send the "pathname" for the datagram */
+      /* be sure to open the datagram first.  */
+      pname = std::filesystem::temp_directory_path ();
+
+#if !defined (SERVER_MODE)
+      pname += "/csql_tcp_setup_server" + std::to_string (getpid ());
+#else // SERVER_MODE
+      pname += "/cubrid_tcp_setup_server" + std::to_string (getpid ());
+#endif // SERVER_MODE
```
ribram: More accurate names for these files would be "client_master" and "server_master"
```suggestion
#if !defined (SERVER_MODE)
      pname += "/client_master_tcp_setup_server" + std::to_string (getpid ());
#else // SERVER_MODE
      pname += "/server_master_tcp_setup_server" + std::to_string (getpid ());
#endif // SERVER_MODE
```

src/connection/connection_support.c
```diff
@@ -2945,3 +2957,213 @@ css_conn_entry::init_pending_request ()
...
+    {
+      fd = css_tcp_client_open_with_retry (host_name, port, true);
+    }
+#else /* !WINDOWS */
+  fd = css_tcp_client_open_with_retry (host_name, port, true);
+#endif /* WINDOWS */
+
+#else // SERVER_MODE
+  fd = css_tcp_client_open ((char *) host_name, port);
+#endif // SERVER_MODE
```
ribram: The next code is equivalent, but fewer lines:
```suggestion
#if !defined (WINDOWS) && !defined (SERVER_MODE)
  if (timeout > 0)
    {
      /* timeout in milli-seconds in css_tcp_client_open_with_timeout() */
      fd = css_tcp_client_open_with_timeout (host_name, port, timeout * 1000);
    }
  else
#else /* WINDOWS || SERVER_MODE */
    {
      fd = css_tcp_client_open (host_name, port);
    }
#endif /* WINDOWS || SERVER_MODE */
```

src/connection/connection_support.c
```diff
@@ -2945,3 +2957,213 @@ css_conn_entry::init_pending_request ()
...
+#if !defined (WINDOWS)
+      if (errno == ETIMEDOUT)
+	{
+	  er_set_with_oserror (ER_ERROR_SEVERITY, ARG_FILE_LINE, ERR_CSS_TCP_CONNECT_TIMEDOUT, 2, host_name, timeout);
+	}
+#endif /* !WINDOWS */
+      if (errno != ETIMEDOUT)
+	{
+	  er_set_with_oserror (ER_ERROR_SEVERITY, ARG_FILE_LINE, ERR_CSS_TCP_CANNOT_CONNECT_TO_MASTER, 1, host_name);
+	}
```
ribram: Unnecessary `#if !defined (WINDOWS)`; errno check is enough.
```suggestion
      if (errno == ETIMEDOUT)
	{
	  er_set_with_oserror (ER_ERROR_SEVERITY, ARG_FILE_LINE, ERR_CSS_TCP_CONNECT_TIMEDOUT, 2, host_name, timeout);
	}
      else
	{
	  er_set_with_oserror (ER_ERROR_SEVERITY, ARG_FILE_LINE, ERR_CSS_TCP_CANNOT_CONNECT_TO_MASTER, 1, host_name);
	}
```

ribram (approved):

src/connection/connection_support.c
```diff
@@ -2945,3 +2957,204 @@ css_conn_entry::init_pending_request ()
...
+	  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ERR_CSS_SERVER_ALREADY_EXISTS, 1,
+		  GET_REAL_MASTER_CONN_NAME (message_to_master));
+	}
+      else
+	{
+	  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ERR_CSS_SERVER_ALREADY_EXISTS, 1, message_to_master);
+	}
+#else // SERVER_MODE
+      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ERR_CSS_SERVER_ALREADY_EXISTS, 1, message_to_master);
+#endif // SERVER_MODE
```
cristiarg, line 3040: shorter, but more ambiguous :)
```suggestion
      else
#endif // !SERVER_MODE
	{
	  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ERR_CSS_SERVER_ALREADY_EXISTS, 1, message_to_master);
	}
```

src/connection/connection_support.c
```diff
@@ -2945,3 +2957,204 @@ css_conn_entry::init_pending_request ()
...
+    }
+
+  response = ntohl (response_buff);
+
+  TRACE ("css_connect_to_master_server received %d as response from master\n", response);
+
+  switch (response)
+    {
+    case SERVER_ALREADY_EXISTS:
+#if !defined (SERVER_MODE)
```
cristiarg, line 3018: Is `CS_MODE` equivalent with `!SERVER_MODE`?
@ribram can you please explain?
I do not know so it would be better for future to know where this equivalence results from.

src/connection/connection_support.c
adrianbrab: Changed. Now is even shorter and very clear

src/connection/connection_support.c
ribram: There are three definitions: `CS_MODE`, `SERVER_MODE` and `SA_MODE`. `!SERVER_MODE` usually means either `CS_MODE` or `SA_MODE`. In the context of connecting to master, I think `SA_MODE` is irrelevant (it does not connect to master).

src/connection/connection_support.c
```diff
@@ -125,6 +127,16 @@ static char css_Vector_buffer[CSS_VECTOR_SIZE];
 #endif /* SERVER_MODE */
 #endif /* WINDOWS */
 
+#ifdef PACKET_TRACE
+#define TRACE(string, arg1)        \
```
cristiarg, line 131: The `TRACE` macro defined in `connection_sr.c` was only used in the function `css_connect_to_master_server` which was removed from there. The one in `connection_cl.c` has more uses. Maybe only one can be kept in `connection_support.h` although it is arguable whether such a generically named macro should be exposed in a header? Better to transform it into an inline function in `connection_support.h` with a more connection related name.

adrianbrab: I removed the TRACE macro from `connection_sr.c`. I think we can leave TRACE as a macro in `connection_support.c`, @ribram what do you think?

ribram: I agree letting it as is. Cristi's suggestion is also valid, but it is not in the purpose of this patch. Please choose as you see fit.

adrianbrab: I think in the scope of this patch it should remain as it is.

cristiarg (approved):

PR #3072, opened 2021-10-06
[CBRD-24070] Make the recovery information logging more verbose

wrlawodms:
http://jira.cubrid.org/browse/CBRD-24070


wrlawodms:
I'll add the multi-language support after the code review.

src/transaction/log_recovery.c
```diff
@@ -6452,3 +6495,20 @@ log_rv_end_simulation (THREAD_ENTRY * thread_p)
...
 }
+
+static UINT64
+log_cnt_pages_containing_lsa (const log_lsa * from_lsa, const log_lsa * to_lsa)
+{
+  if (*from_lsa == *to_lsa)
+    {
+      return 0;
+    }
+  else if (from_lsa->offset == to_lsa->offset)
```
hornetmj: This condition means that only page id is different. I think this case also returns 'from_lsa->pageid - to_lsa->pageid + 1'.

src/transaction/log_recovery.c
```diff
@@ -4585,6 +4606,26 @@ log_recovery_undo (THREAD_ENTRY * thread_p)
   /* Find the largest LSA to undo */
   log_find_unilaterally_largest_undo_lsa (thread_p, max_undo_lsa);
 
+  /* Print undo recovery information */
+  // *INDENT-OFF*
+  auto count_func =[&cnt_trans_to_undo] (LOG_TDES & tdes) {
+    cnt_trans_to_undo++;
+  };
+  log_system_tdes::map_all_tdes (count_func);
```
hornetmj: Don't you need to check log_Gl.trantable.all_tdes?

wrlawodms: I missed it. I'll add it.

src/transaction/log_recovery.c
wrlawodms: Yes, you're right

eido5 (approved):

wrlawodms:
@hornetmj  I have taken all of your comments into account. Plus, there are some changes.

1.  Replcae `log_find_unilaterally_largest_undo_lsa()` with  the new function `logtb_rv_read_only_map_undo_tdes()` and a lambda fnction.
=> to generalize iterating undo transactions.
2.  To get the minimum head lsa and the number of undo trxs, the `logtb_rv_read_only_map_undo_tdes()` is used.
=> to iterate undo trasactions precisely. It iterated possibly a different set of transactions. It's the reason for doing (1).
3. I found out that log_system_tdes can have NULL head_lsa, so I have fixed it.

src/transaction/log_tran_table.c
```diff
@@ -4199,12 +4185,13 @@ log_find_unilaterally_largest_undo_lsa (THREAD_ENTRY * thread_p, LOG_LSA & max_u
...
 	    {
-	      max_undo_lsa_func (*tdes);
+	      map_func (*tdes);
 	    }
 	}
     }
   /* Check system worker transactions. */
-  log_system_tdes::map_all_tdes (max_undo_lsa_func);
+  // *INDENT-OFF*
+  log_system_tdes::map_all_tdes (map_func);
```
hornetmj, line 4192: Plz, move // \*INDENT-ON\* after this line.

src/transaction/log_tran_table.c
```diff
@@ -1059,8 +1059,10 @@ logtb_rv_find_allocate_tran_index (THREAD_ENTRY * thread_p, TRANID trid, const L
   if (logtb_is_system_worker_tranid (trid))
     {
       // *INDENT-OFF*
-      return log_system_tdes::rv_get_or_alloc_tdes (trid);
+      tdes = log_system_tdes::rv_get_or_alloc_tdes (trid);
       // *INDENT-ON*
+      LSA_COPY (&tdes->head_lsa, log_lsa);
```
hornetmj: Why is this line added?

wrlawodms: > I found out that log_system_tdes can have NULL head_lsa, so I have fixed it.

This is where I mentioned I fixed it. When a log record is ran into during the analysis phase, normal tdes or sys tdes is allocated. In the case of sys tdes , head_lsa is not copied (so, NULL), so we can't get the minimum lsa of the UNDO phase.

src/transaction/log_tran_table.c
wrlawodms: Thanks

src/transaction/log_tran_table.c
hornetmj: @wrlawodms Sorry, I didn't notice your comments.

This line looks like always updating the tdes->head_lsa regardless of whether the tdes already exists or is newly created. And now I'm not sure if the log_lsa passed as an argument is the first log address of this system transaction. Can you explain in more detail about this?

wrlawodms: @hornetmj Very thank you. I was careless. See the change: [db1333f](https://github.com/CUBRID/cubrid/pull/3072/commits/db1333f1f0a2536393bb358724766505cd34ea65)

src/transaction/log_system_tran.cpp
```diff
@@ -228,14 +228,15 @@ log_system_tdes::rv_get_tdes (TRANID trid)
...
-log_system_tdes::rv_get_or_alloc_tdes (TRANID trid)
+log_system_tdes::rv_get_or_alloc_tdes (TRANID trid, const LOG_LSA &log_lsa)
 {
   log_tdes *tdes = rv_get_tdes (trid);
   if (tdes == NULL)
     {
       log_tdes *tdes = systdes_create_tdes ();
       tdes->state = TRAN_UNACTIVE_UNILATERALLY_ABORTED;
       tdes->trid = trid;
+      tdes->head_lsa = log_lsa;
```
hornetmj, line 239:
```suggestion
      tdes->head_lsa = tdes->tail_lsa = log_lsa;
```

src/transaction/log_tran_table.c
```diff
@@ -1059,8 +1059,9 @@ logtb_rv_find_allocate_tran_index (THREAD_ENTRY * thread_p, TRANID trid, const L
   if (logtb_is_system_worker_tranid (trid))
     {
       // *INDENT-OFF*
-      return log_system_tdes::rv_get_or_alloc_tdes (trid);
+      tdes = log_system_tdes::rv_get_or_alloc_tdes (trid, *log_lsa);
```
hornetmj:
```suggestion
      return log_system_tdes::rv_get_or_alloc_tdes (trid, *log_lsa);
```

src/transaction/log_tran_table.c
```diff
@@ -1059,8 +1059,9 @@ logtb_rv_find_allocate_tran_index (THREAD_ENTRY * thread_p, TRANID trid, const L
...
       // *INDENT-ON*
+      return tdes;
```
hornetmj:
```suggestion
```

hornetmj (approved):

src/transaction/log_system_tran.cpp
wrlawodms: Why do you think we should initialize tail_lsa here together? The head_lsa of a normal transaction also isn't initialized when allocating. This is changed during the analysis phase.

hornetmj: Do you mean the initialization of the tail_lsa of a normal transaction?

This is proposed because it is natural to set tail lsa to the same value when head lsa is set. It is strange that head lsa has a value, but tail lsa is null. There may not be any bugs right now, but it is thought that initializing the tail lsa together will prevent problems that may arise.

wrlawodms: > Do you mean the initialization of the tail_lsa of a normal transaction?

Yes, this function, `rv_get_or_alloc_tdes()`, is called in the `logtb_rv_find_allocate_tran_index()`, in which a normal transaction also can be allocated, but their head_lsa is only initialized here. I agree what you point out is strange, but I think it is more weird that a tdes returned from `logtb_rv_find_allocate_tran_index` can have either null tail_lsa or not null tail_lsa.

hornetmj: Good point. I checked all caller of logtb_rv_find_allocate_tran_index() and I found that they changed the tail_lsa except of  log_rv_analysis_compensate() and log_rv_analysis_end_checkpoint(). These exceptions tell us why a tail_lsa did not set in the logtb_rv_find_allocate_tran_index().

Thanks for your explanations.

wrlawodms:
@eido5 @hornetmj  I have added multi-lingual support. Please review it.

msg/ko_KR.utf8/cubrid.msg
```diff
@@ -1354,8 +1354,13 @@ $ LOADDB
 1283 Error reserved for scalability development.
 1284 %1$s 의 기본키는 외래키 참조로 인해 truncate될 수 없습니다.
 1285 외래키 (%1$s) 의 ON DELETE action이 CASCADE가 아니기 때문에 truncate할 수 없습니다.
+1296 Log recovery: ANALYSIS 페이즈가 시작되었습니다.
```
hornetmj: How about changing '페이즈' to '단계'?

msg/tr_TR.utf8/cubrid.msg
```diff
@@ -1346,16 +1346,21 @@ Anahtar dosyasının (_keys) yolunu kontrol edin ve uygun anahtarı içerdiğind
 1275 Error reserved for scalability development.
 1276 Error reserved for scalability development.
 1277 Error reserved for scalability development.
-1278 Error reserved for scalability development.
+12t8 Error reserved for scalability development.
```
hornetmj:
```suggestion
1278 Error reserved for scalability development.
```

msg/ko_KR.utf8/cubrid.msg
wrlawodms: OK

msg/tr_TR.utf8/cubrid.msg
wrlawodms: Oh my god

PR #3082, opened 2021-10-08
[LETS-237] There is no error message displayed when server type is misspelt

adrianbrab:
jira.cubrid.org/browse/LETS-237

Printed the misspellt server type in terminal before closing the server.
Changed `ER_INVALID_SERVER_TYPE_ARGUMENT` error message to include the misspellt server type.


ribram (approved):

src/executables/server.c
```diff
@@ -348,7 +348,7 @@ argument_handler (int argc, char **argv)
           // *INDENT-OFF*
           else
             {
-              er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_INVALID_SERVER_TYPE_ARGUMENT, 0);	// error that the type is not valid
+              er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_INVALID_SERVER_TYPE_ARGUMENT, 1, optarg);	// error that the type is not valid
```
cristiarg: alignment, line exceeds 120 chars

src/executables/server.c
```diff
@@ -422,6 +422,8 @@ main (int argc, char **argv)
     ret_val = argument_handler (argc, argv);
     if (ret_val != NO_ERROR)
       {
+        PRINT_AND_LOG_ERR_MSG ("%s\n", er_msg ());
+        fflush (stderr);
```
cristiarg: This might not be the best approach.
The error should be printed in the scope where it appeared.
If there is more than one error that should be printed, only the last one will be printed.
Or if the function `argument_handler` is modified with the addition of another argument.

src/executables/server.c
```diff
@@ -348,7 +348,7 @@ argument_handler (int argc, char **argv)
           // *INDENT-OFF*
```
cristiarg, line 339: I suppose it is first indent `OFF` then `ON`?

adrianbrab: Fixed now.

src/executables/server.c
adrianbrab: moved the comment before er_set

src/executables/server.c
adrianbrab: Changed now to print inside argument_handler, but the function will always show just one error msg, the first one encountered.

ribram: The purpose is to print to console an error, not all. If there are more than one errors, they will be logged in the error log file. The approach of printing the error is taken from `net_server_start ()` function.

 My suggestion is a different one: add the error printing as scope_exit to `server.c:main()` function, and remove it from `net_server_start()` too. Printing the error to console in the end will also cover adding new function calls.
 
 @cristiarg @adrianbrab what do you think?

adrianbrab: I think this is the best approach to catch all errors, and also write future proof code.

cristiarg (approved):

src/executables/server.c
```diff
@@ -403,26 +406,32 @@ main (int argc, char **argv)
...
       }
     thread_initialize_manager (thread_p);
     fprintf (stdout, "\nThis may take a long time depending on the amount " "of recovery works to do.\n");
-
+    // *INDENT-OFF*
+    scope_exit <std::function<void (void)>> print_on_exit ([thread_p] ()
+    {
+      PRINT_AND_LOG_ERR_MSG ("%s\n", er_msg ());
+      fflush (stderr);
+    });
```
ribram: - Error should be printed only if there was an error, reflected by ret_val's value.
- No point on capturing `thread_p`, but a reference to `ret_val` has to be captured.
```suggestion
    scope_exit <std::function<void (void)>> print_err_on_exit ([&ret_val] ()
    {
      if (ret_val != 0)
        {
          PRINT_AND_LOG_ERR_MSG ("%s\n", er_msg ());
          fflush (stderr);
        }
    });
```

ribram (changes requested):

src/executables/server.c
```diff
@@ -47,6 +47,7 @@
 #include "network.h"
 #include "environment_variable.h"
 #include "boot_sr.h"
+#include "scope_exit.hpp"
```
ribram: No longer needed
```suggestion
```

ribram (approved):

src/executables/server.c
```diff
@@ -403,27 +406,20 @@ main (int argc, char **argv)
   {				/* to make indent happy */
     if (argc < 2)
       {
-        PRINT_AND_LOG_ERR_MSG ("Usage: server databasename\n");
-        return 1;
+	PRINT_AND_LOG_ERR_MSG ("Usage: server databasename\n");
+	return 1;
```
cristiarg, line 409: Is there a reason to return 1/0?
```suggestion
	return ER_FAILED;
```
and `NO_ERROR` at the end

cristiarg (approved):

src/executables/server.c
ribram: The main function returns an exit code, not the internal database error code. Other main functions return `EXIT_SUCCESS` or `EXIT_FAILURE`. They may be used here too to replace 1/0.

PR #3084, opened 2021-10-11
[LETS-242] Fix replication of heap records undo with vacuum cases

ribram:
http://jira.cubrid.org/browse/LETS-242

This is a fix of a replication issue in a very particular scenario. Sometimes when applying undo on heap records, the record is also vacuumed. However, the compensation log record does not reflect the vacuuming and the page server cannot and does not do the vacuum by itself.

The fix is to add a compensation log record that reflects the vacuumed heap log record.

**Details**

Intially, `vacuum_rv_check_at_undo` was updated to reflect all these changes, but it became an unmanageable monster. To make things more clear, the function was removed and the three functions - `heap_rv_undo_delete`, `heap_rv_undo_update` and `heap_rv_undo_ovf_update` have been rewritten.

A very detailed description of these three scenarios and why they need vacuuming and manual log compensation handling can be found at `heap_file.c:16214`.

`heap_rv_undo_ovf_update` is mostly the same as `vacuum_rv_check_at_undo` in the `REC_BIGONE` case.

There are some differences in `heap_rv_undo_delete` and `heap_rv_undo_update`. They used to apply the heap record change first, then read the modified record to verify if vacuum is required and modified the page again if it was. This pull requests cleans the record descriptor first and then applies the change into the heap page.

Both functions are very similar and the whole implementation can be found in `heap_rv_undo_with_vacuum_internal ()`. The only differences are:

  - Undo delete does not have to vacuum when the record type is REC_HOME. REC_HOME is never physically deleted during an MVCC heap update operation.
  - To undo the delete (and redo its compensation), `spage_insert_for_recovery` needs to be applied. To undo the update, `heap_update_physical` must be applied.

New recovery indexes have been added for new compensation cases: to compensate with an insert, with an update and with no operation.


alexstanro:
According to the scenario, VACUUM remove wrong insert MVCCID - caused by concurrent update. The real MVCCID - target of VACUUM - is inserted at rollback of concurrent update. Why not simply the VACUUM can't wait for an active transaction to finish (abort/commit)? This would simplify the logic.

ribram:
> According to the scenario, VACUUM remove wrong insert MVCCID - caused by concurrent update. The real MVCCID - target of VACUUM - is inserted at rollback of concurrent update. Why not simply the VACUUM can't wait for an active transaction to finish (abort/commit)? This would simplify the logic.

To clarify, the removed MVCCID is an old MVCCID that would be vacuumed if the record was not updated.

To change how VACUUM works and wait for transaction abort/commit sounds dangerous :). The purpose of this patch is only to correct the existing mechanism to make replication work.

alexstanro (approved):
Generally, with this idea, it looks correctly to fix the reported issue. However, I need much time to dig inside for possible issues. I will approve, to not block the team, even I do not agree with the idea, since there is another much simply and less risky fix. 

For me the fix ideea is completely wrong. Here are my argues:
- it breaks VACUUM job. Now transaction does VACCUUM. BAD!
- is too complicate  for programmers to understand the differences between different undo.
- is very complicate to understand the future issue that may appear after such change. Our experiences shows that log modification requires intensive testing.
- it complicate the log, instead of simplifying it
- we have to reproduce this complicate logic when move VACUUM into memory

A much simple solution, I already mentioned, VACUUM has to wait for concurrent transaction to finish. This would not mix VACUUM with transaction. Also would keep the old logic - ALREADY TESTED INTENSIVELY.

My conclusion: If this patch would be committed, probably would be reverted later and a simple and clear fix would replace the current one.

cristiarg (approved):

PR #3086, opened 2021-10-12
[LETS-232] transaction server and page server can have different volume paths

cristiarg:
http://jira.cubrid.org/browse/LETS-232

The data volume absolute paths are dictated by transaction server and its configuration.
If the page server has different paths configured, it must be able to replicated the changes from the transaction server.

Implementation:
- `RVDK_NEWVOL` is the log record that creates a new physical volume
- `RVDK_LINK_PERM_VOLEXT` is the log record that edits a volume to link it to a next (presumably recently added) volume (in a single linked list manner)
- the redo functions for both have been adapted to strip the prepending full path specific to TSRS and, instead, prepend the full path specific to PS taken from the global `boot_Db_full_name`

Other:
- function `disk_dump_goodvol_all` forwards its file pointer argument if supplied, otherwise, as before, `stdout`


src/storage/disk_manager.c
```diff
@@ -1407,6 +1442,20 @@ disk_rv_redo_format (THREAD_ENTRY * thread_p, const LOG_RCV * rcv)
       /* there, I fixed it */
     }
 
+  const char *const other_config_vol_fullname = disk_vhdr_get_vol_fullname (volheader);
+  char buf_new_full_name[PATH_MAX];
+  disk_adapt_other_config_volume_full_name_to_local_config (buf_new_full_name, other_config_vol_fullname);
+  (void) disk_vhdr_set_vol_fullname (volheader, buf_new_full_name);
+
+  // might be useless; there is a separate log record that performs link between volumes
+  const char *const other_config_next_vol_fullname = disk_vhdr_get_next_vol_fullname (volheader);
```
cristiarg, line 1429: This might not be needed as it seems to always be actually specified by the `RVDK_LINK_PERM_VOLEXT` log record.

src/storage/disk_manager.c
```diff
@@ -1531,7 +1579,9 @@ disk_rv_undoredo_link (THREAD_ENTRY * thread_p, const LOG_RCV * rcv)
   link = (DISK_RECV_LINK_PERM_VOLUME *) rcv->data;
 
   vhdr->next_volid = link->next_volid;
-  ret = disk_vhdr_set_next_vol_fullname (vhdr, link->next_vol_fullname);
+  char buf_new_full_name[PATH_MAX];
+  disk_adapt_other_config_volume_full_name_to_local_config (buf_new_full_name, link->next_vol_fullname);
```
cristiarg: The function `disk_rv_undoredo_link` is now common to both redo and undo. Must it be separated in two different flavors?

src/storage/file_io.c
```diff
@@ -6326,7 +6326,7 @@ fileio_get_volume_label_by_fd (int vol_fd, bool is_peek)
 const char *
 fileio_get_volume_label_with_unknown (VOLID volid)
 {
-  static constexpr char *UNKNOWN_VLABEL = "(UNKNOWN)";
+  static constexpr char UNKNOWN_VLABEL[] = "(UNKNOWN)";
```
cristiarg, line 6329: gcc warning

src/storage/disk_manager.c
```diff
@@ -1205,24 +1207,57 @@ disk_get_link (THREAD_ENTRY * thread_p, INT16 volid, INT16 * next_volid, char *n
...
+      local_config_full_name[0] = '\0';
+    }
+  else
+    {
+      // and the rest of the path from the current settings
+      char buf_temp_path[PATH_MAX];
+      const char *const vol_base_path = fileio_get_directory_path (buf_temp_path, boot_db_full_name ());
+
+      fileio_make_volume_ext_given_name (local_config_full_name, vol_base_path, vol_filename);
+    }
```
ribram: With the assert above, this can be rewritten as:
```suggestion
  assert (vol_filename != nullptr && vol_filename[0] != '\0');
  // and the rest of the path from the current settings
  char buf_temp_path[PATH_MAX];
  const char *const vol_base_path = fileio_get_directory_path (buf_temp_path, boot_db_full_name ());

  fileio_make_volume_ext_given_name (local_config_full_name, vol_base_path, vol_filename);
```

src/storage/disk_manager.c
ribram: Remove it then.

src/storage/disk_manager.c
```diff
@@ -5921,12 +5969,13 @@ disk_dump_all (THREAD_ENTRY * thread_p, FILE * fp)
...
  *   volid(in): Permanent volume identifier
- *   ignore(in):
+ *   fp(in): file pointer
  */
 static bool
-disk_dump_goodvol_all (THREAD_ENTRY * thread_p, INT16 volid, void *ignore)
+disk_dump_goodvol_all (THREAD_ENTRY * thread_p, INT16 volid, void *fp)
 {
-  (void) disk_dump_volume_system_info (thread_p, stdout, volid);
+  FILE *const fp_or_stdout = (fp != nullptr) ? (FILE *) fp : stdout;
```
ribram: The fix for file output of diagdb tool should be separated and considered for develop too. Right? @cristiarg @hornetmj

src/storage/disk_manager.c
ribram: It must not :)

src/storage/disk_manager.c
```diff
@@ -1205,24 +1207,57 @@ disk_get_link (THREAD_ENTRY * thread_p, INT16 volid, INT16 * next_volid, char *n
...
+  const char *const vol_filename = fileio_get_base_file_name (other_config_full_name);
+  if (vol_filename == nullptr || vol_filename[0] == '\0')
+    {
+      local_config_full_name[0] = '\0';
+    }
+  else
+    {
+      // and the rest of the path from the current settings
+      char buf_temp_path[PATH_MAX];
+      const char *const vol_base_path = fileio_get_directory_path (buf_temp_path, boot_db_full_name ());
```
ribram: Can we save this database path in a global variable and make it easily accessible?

src/storage/disk_manager.c
```diff
@@ -1205,24 +1207,57 @@ disk_get_link (THREAD_ENTRY * thread_p, INT16 volid, INT16 * next_volid, char *n
...
+/*
+ * disk_adapt_other_config_volume_full_name_to_local_config - adapt the full volume path and name
+ *                      coming, possibly, from an old or different config to the current config
+ */
+void
+disk_adapt_other_config_volume_full_name_to_local_config (char *local_config_full_name,
+							  const char *other_config_full_name)
+{
+  assert (local_config_full_name != nullptr);
+
```
ribram, line 1204: Is other_config_full_name allowed to be null? Maybe it is better to assert that it is not.

src/storage/disk_manager.c
```diff
@@ -1205,24 +1207,57 @@ disk_get_link (THREAD_ENTRY * thread_p, INT16 volid, INT16 * next_volid, char *n
   return next_volext_fullname;
 }
 
+/*
+ * disk_adapt_other_config_volume_full_name_to_local_config - adapt the full volume path and name
+ *                      coming, possibly, from an old or different config to the current config
+ */
+void
+disk_adapt_other_config_volume_full_name_to_local_config (char *local_config_full_name,
```
ribram: Make the name a bit shorter by dropping "other_config". It is implied and make the function easier to read/use.
```suggestion
disk_adapt_volume_full_name_to_local_config (char *local_config_full_name,
```

src/storage/disk_manager.c
```diff
@@ -5375,7 +5423,7 @@ disk_vhdr_set_next_vol_fullname (DISK_VOLUME_HEADER * vhdr, const char *next_vol
 
   (void) memcpy (disk_vhdr_get_next_vol_fullname (vhdr), next_vol_fullname, next_vol_fullname_size);
 
-  return ret;
+  return NO_ERROR;
```
ribram: Change return to void if it never fails. It makes it easier to call the function without having to handle the error cases.

src/storage/disk_manager.c
```diff
@@ -5328,7 +5377,7 @@ disk_vhdr_set_vol_fullname (DISK_VOLUME_HEADER * vhdr, const char *vol_fullname)
 
   (void) memcpy (disk_vhdr_get_vol_fullname (vhdr), vol_fullname,
 		 MIN ((ssize_t) strlen (vol_fullname) + 1, DB_MAX_PATH_LENGTH));
-  return ret;
+  return NO_ERROR;
```
ribram: Change return to void if it never fails. It makes it easier to call the function without having to handle the error cases.

src/storage/disk_manager.c
cristiarg: Function argument `other_config_full_name` cannot be null (because it comes from the volume header `var_fields` which are initialized to empty string) but it can be empty. Asserted accordingly.

src/storage/disk_manager.c
cristiarg: I think it's minor, but I can do it separately.

src/storage/disk_manager.c
cristiarg: Added a global `boot_Db_directory_path` next to `boot_Db_directory_path` and initialized together with it.
Adapted this patch's use of it.
However, there are many usages of the type:
```
      temp_path = fileio_get_directory_path (alloc_tempath, boot_Db_full_name);
```
which would benefit, but are beyond the scope.

src/storage/disk_manager.c
cristiarg: It is not allowed to be null. Its source is in the volume header's `var_fields`. Asserted.

src/storage/disk_manager.c
cristiarg: Done

src/storage/disk_manager.c
cristiarg: Done

src/storage/disk_manager.c
cristiarg: Done

ribram (approved):

PR #3087, opened 2021-10-13
[CBRD-24056] Refactor the handling of hint statements

ctshim:
http://jira.cubrid.org/browse/CBRD-24056

**Purpose**
Solves the problem of different keyword comparison methods between pt_check_hint() and pt_get_hint() and reduces the number of strstr() comparison attempts.
Improves hint keyword-only extraction and argument extraction.

**Implementation**
N/A

**Remarks**
The two sqls that failed in "ci/circleci: test_sql" are because the problem in the existing code where the USE_IDX hint is extracted from /*+ recompile USE_IDX_DESC */ has been fixed. 
I suspect you were trying to describe USE_DESC_IDX.


shparkcubrid:
Please attach a test case.

src/parser/scanner_support.c
```diff
@@ -531,214 +638,424 @@ pt_get_hint (const char *text, PT_HINT hint_table[], PT_NODE * node)
...
  * pt_check_hint () - search hint comment from hint table
  *   text(in): hint comment
  *   hint_table(in): hint info structure list
  *   result_hint(in): found result
- *   prev_is_white_char(in): flag indicates prev char
+ *   prev_is_white_char(in): -
  */
 void
-pt_check_hint (const char *text, PT_HINT hint_table[], PT_HINT_ENUM * result_hint, bool prev_is_white_char)
+pt_check_hint (const char *text, PT_HINT hint_table[], PT_HINT_ENUM * result_hint, bool)
```
hgryoo:
```suggestion
pt_check_hint (const char *text, PT_HINT hint_table[], PT_HINT_ENUM * result_hint, bool prev_is_white_char)
```
typo? or intended to remove the last argument?

src/parser/scanner_support.c
```diff
@@ -531,214 +638,424 @@ pt_get_hint (const char *text, PT_HINT hint_table[], PT_NODE * node)
...
 
-	}			/* while (hint_p) */
-    }				/* for (i = ... ) */
+  px = hint_table->arg_list;
+  do
+    {
+      if (px->node_type == PT_NAME)
+	{
+	  if (px->info.name.resolved)
+	    printf ("[%s].[%s]", px->info.name.resolved, px->info.name.original);
```
hgryoo: please use `fprintf()` and add an `FILE *` argument at this function.
It will make it easy to write in a log file too.

src/parser/scanner_support.c
ctshim: prev_is_white_char is a meaningless parameter, so it is omitted.
Let's remove the parameter itself.

src/parser/scanner_support.c
ctshim: debug_hit_hint_print() is for reference when someone needs to work on hint-related code again later or need to debug.
I will express it in the form of fprintf(stdout, ~~).

src/parser/scanner_support.c
```diff
@@ -126,6 +168,65 @@ pt_nextchar (void)
...
+      hint_table_lead_offset[i] = sum;
+      sum += tCnt;
+    }
+
+  // Copy for lower character
+  for (i = 'A'; i <= 'Z'; i++)
+    {
+      hint_table_lead_offset[i + 32 /*('a'-'A') */ ] = hint_table_lead_offset[i];
+    }
+
```
beyondykk9, line 291:
```suggestion
unsigned char token_char = 0;
int hint_tbl_len = sizeof (hint_table[0]);
memset (hint_table_lead_offset, hint_tbl_len - 1, sizeof (hint_table_lead_offset));

// ordering by asc 
qsort (hint_table, i, hint_tbl_len, &hint_token_cmp);

for (i = 0; hint_table[i].tokens; i++)
  {
    hint_table[i].is_hit = false;
    hint_table[i].length = (int) strlen (hint_table[i].tokens);
    if (token_char != (unsigned char)hint_table[i].tokens[0])
	  {
	    token_char = (unsigned char)hint_table[i].tokens[0];
	    hint_table_lead_offset[token_char] = i;
	    hint_table_lead_offset[token_char + 32] = i;
	  }
  }
```

it looks shorter.

ctshim: If there is an empty character in the middle, the concept of cumulative distribution may collapse.
We will keep the current code.

src/parser/scanner_support.c
```diff
@@ -531,214 +638,428 @@ pt_get_hint (const char *text, PT_HINT hint_table[], PT_NODE * node)
...
+  char cBk;
+  int cmp;
+
+  if (!string || string[0] == '\0')
+    {
+      return -1;
+    }
+
+  matched_idx = 0;
+  for (i = hint_table_lead_offset[*string]; hint_table[i].tokens; i++)
```
beyondykk9:
```suggestion
  for (i = hint_table_lead_offset[*string]; hint_table[i].tokens && *string == hint_table[i].tokens[0]; i++)
```

src/parser/scanner_support.c
```diff
@@ -531,214 +638,428 @@ pt_get_hint (const char *text, PT_HINT hint_table[], PT_NODE * node)
...
+      in++;
+    }
+
+  /* illegal hint expression */
+  if (is_first_hit)
+    {
+      parser_free_node (this_parser, hint_table[hint_idx].arg_list);
+      hint_table[hint_idx].arg_list = NULL;
+
+#if 1
```
beyondykk9, line 985: "#if 1 ", is it necessary?
does it mean temporary?

src/parser/scanner_support.c
ctshim: (*string == hint_table[i].tokens[0]) Why do you think the condition is necessary?
The comparison may not be appropriate because string can also be lowercase.
And when find_hint_token() is called, the first character is guaranteed to match.

src/parser/scanner_support.c
ctshim: That case is strictly a hint syntax error.
It's a bit ambiguous how to handle the case where the parentheses are not closed, but I've made it the same as in the old code.
If it is defined as handling as an error, it is a part that should be deleted.

src/parser/scanner_support.c
```diff
@@ -531,214 +638,428 @@ pt_get_hint (const char *text, PT_HINT hint_table[], PT_NODE * node)
...
+
+
+static unsigned char *
+read_hint_args (unsigned char *instr, PT_HINT hint_table[], int hint_idx, PT_HINT_ENUM * result_hint)
+{
+  unsigned char *sp, delimeter;
+  unsigned char *in = instr;
+  unsigned char *dot_ptr = 0x00;
+
+#if 0				//
```
beyondykk9: remove useless codes.

src/parser/scanner_support.c
beyondykk9: Then, compare with upper(*string), leading offset should be used to make less comparing count.

src/parser/scanner_support.c
ctshim: OK, I will remove it.

src/parser/scanner_support.c
```diff
@@ -41,13 +41,25 @@
 
 #define IS_WHITE_CHAR(c) \
                     ((c) == ' ' || (c) == '\t' || (c) == '\n' || (c) == '\0')
-
+#define IS_WHITE_SPACE(c) ((c) == ' ' || (c) == '\t' || (c) == '\n' || (c) == '\r')
```
youngjinj, line 44: #define IS_WHITE_SPACE(c) ((c) <= 32)

How is it?

src/parser/scanner_support.c
ctshim: Your comment seems to be intended to run the last one of the for loop.
Instead of using upper(), doing something like the following would do what you intended.

`for (i = hint_table_lead_offset[*string]; i < hint_table_lead_offset[*string + 1]; i++)`

I'll fix it like this.

src/parser/scanner_support.c
```diff
@@ -531,214 +638,428 @@ pt_get_hint (const char *text, PT_HINT hint_table[], PT_NODE * node)
...
+    {
+      return -1;
+    }
+
+  matched_idx = 0;
+  for (i = hint_table_lead_offset[*string]; hint_table[i].tokens; i++)
+    {
+      if (matched_idx > 0)
+	{
+	  if (memcmp (hint_table[i - 1].tokens, hint_table[i].tokens, matched_idx) != 0)
```
youngjinj, line 836: Previously, it matches up to matched_idx .
After that, there is no match until matched_idx.

Doesn't that mean there's no matching hint after that?
How about exiting the loop when the condition is true?

youngjinj (approved):

ctshim:
@shparkcubrid

New cases to be added will be discussed with the QA team.

beyondykk9 (approved):

src/parser/scanner_support.c
ctshim: I've applied your comments to the code. thank you.

src/parser/scanner_support.c
ctshim: That's a good idea, but I'll stick with the classic way.

shparkcubrid (approved):

hyunikn (approved):

hgryoo (approved):

beyondykk9 (approved):

hgryoo (approved):

PR #3088, opened 2021-10-13
[LETS-245] er_manager is already closed when calling er_msg()

adrianbrab:
http://jira.cubrid.org/browse/LETS-245

In `server.c:main()` when calling `er_msg()` after `net_server_start()` has ended the er_manager is already closed.

Moved `er_final (ER_ALL_FINAL);` at the end of the main function.

### DIFF
--- a/src/executables/server.c
+++ b/src/executables/server.c
@@ -438,8 +438,8 @@ main (int argc, char **argv)
 	PRINT_AND_LOG_ERR_MSG ("%s\n", er_msg ());
 	fflush (stderr);
       }
+    er_final (ER_ALL_FINAL);
   }
-  er_final (ER_ALL_FINAL);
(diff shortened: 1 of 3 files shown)


src/executables/server.c
```diff
@@ -439,6 +439,7 @@ main (int argc, char **argv)
 	fflush (stderr);
       }
   }
+  er_final (ER_ALL_FINAL);
```
ribram: `er_final` should be called in the same scope that `er_init` is called.
```suggestion
    er_final (ER_ALL_FINAL);
  }
```

adrianbrab: Moved it now.

ribram (approved):

cristiarg (approved):

PR #3091, opened 2021-10-14
[CBRD-24063] Update submodule (cubrid-jdbc)

hgryoo:
http://jira.cubrid.org/browse/CBRD-24063


beyondykk9:
why google-java-format version 1.7 ? recently 1.11.0 is released, how about 1.11.0?

hgryoo:
> why google-java-format version 1.7 ? recently 1.11.0 is released, how about 1.11.0?

Since the cubrid-jdbc is based on JDK 1.6, google-java-format 1.7 should be used.

beyondykk9 (approved):

PR #3092, opened 2021-10-14
[LETS-238] Duplicate error message when configuration parameter is misspelt

adrianbrab:
http://jira.cubrid.org/browse/LETS-233

On `SERVER_MODE` `sysprm_load_and_init` is called twice before the `cub_server` process is stopped in case of an error in the conf file.

The first call is in `net_server_start` and the second in `boot_restart_server`. For minimal impact in the code the exit code of the first `sysprm_load_and_init` is now used to determine if the server boot should continue;

### DIFF
--- a/src/communication/network_sr.c
+++ b/src/communication/network_sr.c
@@ -1347,7 +1347,11 @@ net_server_start (THREAD_ENTRY * thread_p, const char *server_name)
       goto end;
     }
 
-  sysprm_load_and_init (NULL, NULL, SYSPRM_LOAD_ALL);
+  if (sysprm_load_and_init (NULL, NULL, SYSPRM_LOAD_ALL) != NO_ERROR)
+    {
+      status = -1;
+      goto end;
+    }
   sysprm_set_er_log_file (server_name);
 
   if (sync_initialize_sync_stats () != NO_ERROR)


ribram (approved):

cristiarg:
The check is good and in line with other existing checks.

However, the state is as follows with regard to `sysprm_load_and_init` and `boot_restart_server`:
```
net_server_start
  - sysprm_load_and_init
  - boot_restart_server
    - SERVER_MODE: sysprm_load_and_init called again 

xboot_restart_from_backup
  - sysprm_load_and_init
  - boot_restart_server
    - SERVER_MODE: sysprm_load_and_init called again

xboot_copy
  - ** missing sysprm_load_and_init **
  - boot_restart_server
    - SERVER_MODE: sysprm_load_and_init called here
```

Wouldn't it be better to remove `sysprm_load_and_init` from `boot_restart_server` and add it - if needed - before the call in `xboot_copy`. Or, if `xboot_copy` is never run in `SERVER_MODE`, not add it at all and just remove from `boot_restart_server` ?
@ribram

ribram:
@cristiarg I think some `sysprm_load_and_init()` calls may be removed, after a careful analysis. But I don't think this is the time to do it.

ribram (approved):

cristiarg (approved):

PR #3094, opened 2021-10-14
[LETS-213] Transaction server base class

ribram:
http://jira.cubrid.org/browse/LETS-213

Extract the functions that are common to passive transaction server from active_tran_server into the tran_server base class.

The differences between active_tran_server and the passive transaction server:

  - Passive transaction server is always running on remote storage.
  - Only active transaction server sends log to page server and receives confirmation for the log saved on disk.

In consequence, some functionality regarding these two differences remains in active_tran_server. The rest is moved to the tran_server class. Some virtual functions can be used to implement the different parts:

  - `uses_remote_storage` & `get_remote_storage_config`
  - `get_request_handlers ()`

### DIFF
--- a/src/server/active_tran_server.hpp
+++ b/src/server/active_tran_server.hpp
@@ -23,8 +23,8 @@
 #include "page_broker.hpp"
 #include "request_sync_client_server.hpp"
 #include "server_type.hpp"
+#include "tran_page_requests.hpp"
 #include "tran_server.hpp"
-#include "ts_ps_request.hpp"
 
 #include <memory>
 #include <string>
--- a/src/server/active_tran_server.hpp
+++ b/src/server/active_tran_server.hpp
@@ -24,6 +24,10 @@
 class active_tran_server : public tran_server
 {
   public:
+    active_tran_server () : tran_server (cubcomm::server_server::CONNECT_ACTIVE_TRAN_TO_PAGE_SERVER)
+    {
+    }
+
     bool uses_remote_storage () const final;
 
   private:
--- a/src/server/active_tran_server.hpp
+++ b/src/server/active_tran_server.hpp
@@ -19,11 +19,12 @@
 #ifndef _ACTIVE_TRAN_SERVER_HPP_
 #define _ACTIVE_TRAN_SERVER_HPP_
 
-#include "ats_ps_request.hpp"
 #include "communication_node.hpp"
 #include "page_broker.hpp"
 #include "request_sync_client_server.hpp"
 #include "server_type.hpp"
+#include "tran_server.hpp"
+#include "ts_ps_request.hpp"
 
 #include <memory>
 #include <string>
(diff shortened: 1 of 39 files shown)


src/server/tran_server.hpp
```diff
@@ -0,0 +1,131 @@
...
+// Transaction server class hierarchies:
+//
+//                            tran_server
+//                               /   \
+//                              /     \
+//                             /       \
+//             active_tran_server     passive_tran_server
+//
+//
+// tran_server_abstract:
```
cristiarg: comment is not up to date

src/server/active_tran_server.hpp
```diff
@@ -19,80 +19,26 @@
...
-    void push_request (tran_to_page_request reqid, std::string &&payload);
-
-  private:
-    using page_server_conn_t =
-	    cubcomm::request_sync_client_server<tran_to_page_request, page_to_tran_request, std::string>;
+    bool uses_remote_storage () const final;
 
   private:
-    int connect_to_page_server (const cubcomm::node &node, const char *db_name);
+    void on_boot () final;
```
cristiarg: `override` for functions that override

src/server/tran_server.cpp
```diff
@@ -0,0 +1,456 @@
...
+  if (!hosts.length ())
+    {
+      if (uses_remote_storage)
+	{
+	  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_EMPTY_PAGE_SERVER_HOSTS_CONFIG, 0);
+	  return ER_EMPTY_PAGE_SERVER_HOSTS_CONFIG;
+	}
+      else
+	{
+	  // no page server, local storage
```
cristiarg, line 170: assert active tran server?

src/server/tran_server.cpp
```diff
@@ -0,0 +1,456 @@
...
+{
+  return false;
+}
+
+void
+tran_server::push_request (tran_to_page_request reqid, std::string &&payload)
+{
+  if (!is_page_server_connected ())
+    {
+      return;
```
cristiarg, line 360: Does this ever happen and is it valid if so? The message would simply be lost?

src/server/tran_server.cpp
```diff
@@ -0,0 +1,456 @@
...
+      assert (exit_code != NO_ERROR);
+      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_NO_PAGE_SERVER_CONNECTION, 0);
+      exit_code = ER_NO_PAGE_SERVER_CONNECTION;
+    }
+  else if (valid_connection_count == 0)
+    {
+      // failed to connect to any page server
+      assert (exit_code != NO_ERROR);
+      er_clear ();
+      exit_code = NO_ERROR;
```
cristiarg, line 229: This will also not play with passive transaction server.
But will be adapted when needed.

cristiarg (approved):

src/server/tran_server.cpp
ribram: It only happens when transaction server runs with local storage and pages are fetched. The pages are only fetched from local disk.

PR #3095, opened 2021-10-15
[APIS-878] Add charset member to T_CCI_COL_INFO to know column character set type

airnet73:
http://jira.cubrid.org/browse/APIS-878

Purpose
A charset is added to T_CCI_COL_INFO so that the character set of the column can be known.

Implementation
Added T_CCI_COL_INFO to the charset type received from CAS.

Remarks
N/A


kisoo-han (approved):
Since cci is still operated in two sets, this change should be applied to the CUBRID/cubrid-cci repo as well.

PR #3098, opened 2021-10-15
[LETS-233] separate metalog for transaction server and page server

cristiarg:
http://jira.cubrid.org/browse/LETS-233

The single file that a TSRS actually uses/maintains is the metalog (for transaction table snapshoting). In order to allow execution of TSRS and PS on a single node, the servers needs to access separate metalog files. This should also cover the need of the recovery process on TSRS.

Implementation:
- a TSRS would create a metalog file upon boot-up, if not already found on disk
- separate name for TSRS-specific metalog file: _lgmeta_tran

Tested:
- manual tests with separate containers
- manual tests on single node
- recovery scenario with separate containers

### DIFF
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -10632,6 +10632,10 @@ log_read_metalog_from_file_with_create ()
       log_Gl.m_metainfo.set_clean_shutdown (true);
 
       err_code = log_create_metalog_file ();
+
+      // creation of an empty metalog file happens early during the log initialization process
+      // later on, there is a sequence where the clean shutdown flag is set to false such that
+      // an accidental server crash is correctly flagged
     }
   else
     {
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -1143,7 +1143,15 @@ log_initialize_internal (THREAD_ENTRY * thread_p, const char *db_fullname, const
       log_final (thread_p);
     }
 
-  error_code = log_read_metalog_from_file_with_create ();
+  if (is_tran_server_with_remote_storage ())
+    {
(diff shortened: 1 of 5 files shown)


src/transaction/log_page_buffer.c
```diff
@@ -9411,7 +9411,7 @@ logpb_copy_database (THREAD_ENTRY * thread_p, VOLID num_perm_vols, const char *t
   /*
    * Copy log meta-information file
    */
-  fileio_make_log_metainfo_name (to_volname, to_logpath, to_prefix_logname);
+  fileio_make_log_metainfo_name (to_volname, to_logpath, to_prefix_logname, false);
```
cristiarg, line 9414: I suppose, for copy and rename, both TSRS and PS (default) logmeta files would need to be copied/renamed.

src/transaction/log_manager.c
```diff
@@ -10601,12 +10618,38 @@ log_create_metalog_file ()
...
+int
+log_read_metalog_from_file_with_create ()
+{
+  int err_code = NO_ERROR;
+  struct stat sb;
+
+  if (stat (log_Name_metainfo, &sb) != 0)
+    {
+      // should not be called in a a non booting-up scenario
+      assert (log_Gl.m_metainfo.get_checkpoint_count () == 0);
```
```suggestion
      assert (!log_Gl.m_metainfo.is_loaded_from_file ());
```

ribram (approved):

src/transaction/log_page_buffer.c
ribram: No. copy/rename do not work on TSRS.

src/transaction/log_manager.c

ribram (approved):

PR #3100, opened 2021-10-18
[CBRD-24084] Problem that view-merge does not work because of the rownum of the view's subquery.

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-24084

I fix to use pt_has_inst_in_where_and_select_list() instead of mq_is_pushable_subquery() for rownum check.
pt_has_inst_in_where_and_select_list() does not check the rownum of the subquery.

The criterion of is_only_spec that does not check rownum has been changed.
If there is a where, the result may be wrong.

```
as-is : only one table
to-be : only one table + not exists predicates
```


hyunikn (approved):

src/parser/view_transform.c
```diff
@@ -1392,18 +1390,21 @@ mq_is_pushable_subquery (PARSER_CONTEXT * parser, PT_NODE * query, bool is_only_
...
-  cpi.check_xxxnum = !is_only_spec;
-
   cpi.method_found = false;
   cpi.query_found = false;
-  cpi.xxxnum_found = false;
 
   parser_walk_tree (parser, query->info.query.q.select.list, pt_check_pushable, (void *) &cpi, NULL, NULL);
 
-  if (cpi.method_found || cpi.query_found || cpi.xxxnum_found)
+  if (cpi.method_found || cpi.query_found)
```
ctshim, line 1407: Assuming that (!is_only_spec) is true AND pt_has_inst_in_where_and_select_list (parser, query) is false, it is correct not to enter the if block and return false unlike the code before modification, right?

shparkcubrid: If xxxnum is not found, true is returned. This is the same as the previous version.

ctshim: Sorry, I was confused with check_xxxnum.
thank you.

ctshim (approved):

hgryoo (approved):

beyondykk9 (approved):

youngjinj (approved):

PR #3101, opened 2021-10-18
[CBRD-24082] View merging extension : remove unnecessary restriction

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-24082

remove unnecessary restriction
 - join restriction
 - has distinct (main query)
   When view merging, if the main query has DISTINCT, the hidden column of order by in the view is not copied to main query.

add necessary restriction
 - has CONNECT BY (main query)
 - has OUTER JOIN (sub query)


src/parser/view_transform.c
```diff
@@ -1670,37 +1716,42 @@ mq_substitute_subquery_in_statement (PARSER_CONTEXT * parser, PT_NODE * statemen
...
-
-	  /* rewrite vclass_query as a derived table if spec is outer joined or if query is not pushable */
-	  rewrite_as_derived = rewrite_as_derived || !is_pushable_query || is_outer_joined;
-
-	  if (PT_IS_QUERY (tmp_result))
-	    {
-	      rewrite_as_derived = rewrite_as_derived || (tmp_result->info.query.all_distinct == PT_DISTINCT)
-		|| (PT_IS_VALUE_QUERY (query_spec));
-	    }
+	  rewrite_as_derived = false;
```
beyondykk9, line 1754: In the clause "if .... else if ... else", the last "else" and the first "if ... rewirte_as_derived = false" looks not necessary because the variable rewrite_as_derived is already set to "false" initially.

src/parser/view_transform.c
```diff
@@ -1542,26 +1542,57 @@ mq_update_order_by (PARSER_CONTEXT * parser, PT_NODE * statement, PT_NODE * quer
...
+		  statement->info.query.order_by = order->next;
+		}
+	      else
+		{
+		  prev_order->next = order->next;
+		}
+	      /* free order */
+	      save_next = order->next;
+	      order->next = NULL;
+	      parser_append_node (order, free_node);
```
beyondykk9: I think that the "free_node" never has any value other than initial NULL setting.
And parser_append_node returns PT_NODE pointer but it's not handled.

src/parser/view_transform.c
shparkcubrid: I know what you said. But I think it is good for readability.

src/parser/view_transform.c
shparkcubrid: Actually, the
return value of parser_append_node() is unnecessary. the pointer of free_node can be changed through parser_append_node().

ctshim (approved):

src/parser/view_transform.c
beyondykk9: If the "free_node" has no value pointer then it's not necessary to be free it as the last statement.
Is "free_node = parser_append_node (order, free_node)" correct?

shparkcubrid: I made a mistake. It's fixed. thank you.

hyunikn (approved):

beyondykk9 (approved):

youngjinj (approved):

hgryoo (approved):

PR #3103, opened 2021-10-18
[LETS-248] flush log page buffer after recovery undo because dangling transactions might be aborted and compensating log records added

cristiarg:
http://jira.cubrid.org/browse/LETS-248

When dangling transactions are encountered during recovery undo, compensating log records might be added - eg: `log_rv_undo_record`.
Request a flush such that all log pages are up to date.

### DIFF
--- a/src/server/tran_server.hpp
+++ b/src/server/tran_server.hpp
@@ -38,10 +38,10 @@ namespace cubpacking
 
 // Transaction server class hierarchies:
 //
-//                            tran_server
-//                               /   \
-//                              /     \
-//                             /       \
+//                              tran_server
+//                               /      |
+//                              /       |
+//                             /        |
 //             active_tran_server     passive_tran_server
 //
 //
--- a/src/transaction/log_page_buffer.c
+++ b/src/transaction/log_page_buffer.c
@@ -3400,6 +3400,10 @@ logpb_flush_all_append_pages (THREAD_ENTRY * thread_p)
     }
 }
 
+/*
+ * logpb_skip_flush_append_pages - TODO, reason
+ *
+ */
 static void
 logpb_skip_flush_append_pages ()
 {
--- a/src/transaction/log_page_buffer.c
+++ b/src/transaction/log_page_buffer.c
@@ -3394,12 +3394,6 @@ logpb_flush_all_append_pages (THREAD_ENTRY * thread_p)
       // log pages are not written to local disk; they are written by page server
       // skip flushing - aka: pretend that flushing has happened
 
-      // NOTE: this branch only updates bookkeeping and does not offer the same
-      // "functional promise" as the 'else' branch (where log pages are actually
-      // written on persistent storage); in case effective persistence is needed, one must
-      // call `logpb_flush_pages` which also makes sure that the persistence happens
-      // on the page server
-
       logpb_skip_flush_append_pages ();
       return NO_ERROR;
     }
--- a/src/transaction/log_page_buffer.c
+++ b/src/transaction/log_page_buffer.c
@@ -3391,6 +3391,15 @@ logpb_flush_all_append_pages (THREAD_ENTRY * thread_p)
 {
   if (is_tran_server_with_remote_storage ())
     {
+      // log pages are not written to local disk; they are written by page server
+      // skip flushing - aka: pretend that flushing has happened
+
+      // NOTE: this branch only updates bookkeeping and does not offer the same
+      // "functional promise" as the 'else' branch (where log pages are actually
+      // written on persistent storage); in case effective persistence is needed, one must
+      // call `logpb_flush_pages` which also makes sure that the persistence happens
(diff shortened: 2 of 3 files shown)


src/server/tran_server.hpp
```diff
@@ -38,10 +38,10 @@ namespace cubpacking
 
 // Transaction server class hierarchies:
 //
-//                            tran_server
-//                               /   \
-//                              /     \
-//                             /       \
+//                              tran_server
+//                               /      |
```
cristiarg, line 42: gcc warning about multi line comments

src/transaction/log_page_buffer.c
```diff
@@ -3400,6 +3400,10 @@ logpb_flush_all_append_pages (THREAD_ENTRY * thread_p)
     }
 }
 
+/*
+ * logpb_skip_flush_append_pages - TODO, reason
```
cristiarg: The comment for #2864 when this function was added explains what the change does, but not why the change was done. It would be better to explain here why:
```
On a transaction server with remote storage:
    Instead of writing the appended log pages to disk, set nxio_lsa and reset flush page count directly.
    Overwrite dirty old log pages in the log page buffer with new pages.
    Do not create an archive from the active log.
```

src/transaction/log_recovery.c
```diff
@@ -906,7 +908,12 @@ log_recovery_finish_transactions (THREAD_ENTRY * const thread_p)
   log_Gl.rcv_phase = LOG_RECOVERY_ANALYSIS_PHASE;
 
   // start with what metalog tells us
-  const auto[chkpt_lsa, chkpt_info] = log_Gl.m_metainfo.get_highest_lsa_checkpoint_info ();
+  // *INDENT-OFF*
+  const auto highest_lsa_chkpt_info = log_Gl.m_metainfo.get_highest_lsa_checkpoint_info ();
```
cristiarg, line 912: Older gdb does not handle well c++17 structured bindings - complains that it does not know the types, thus being unable to print values.

src/transaction/log_page_buffer.c
ribram: I agree an explanation of why would be helpful, but not here. It should be added in `logpb_flush_all_append_pages()` where the function is called. In this function an explanation of what still might help.

Something along the lines:

```
/*
 * logpb_flush_all_append_pages - Flush log append pages
 *
 * return: 1 : log flushed, 0 : do not need log flush, < 0 : error code
 *
 */
static int
logpb_flush_all_append_pages (THREAD_ENTRY * thread_p)
{
    {
      // Log pages are not written to local disk, they'll be written on the page server.
      // Skip flushing
      logpb_skip_flush_append_pages ();
      return NO_ERROR;
    }
  else
    {
      return logpb_write_append_pages_to_disk (thread_p);
    }
}

static void
logpb_skip_flush_append_pages ()
{
  // Skip flushing append pages to disk. Set nxio_lsa and reset flush page count directly.

  logpb_log ("called logpb_skip_flush_append_pages\n");

  pthread_mutex_lock (&log_Gl.flush_info.flush_mutex);
  log_Gl.flush_info.num_toflush = 0;
  pthread_mutex_unlock (&log_Gl.flush_info.flush_mutex);

  log_Gl.append.set_nxio_lsa (log_Gl.hdr.append_lsa);
}
```

cristiarg: So, actually the function `logpb_flush_all_append_pages` does not do what is says in the TSRS context.
Versus the PS/standalone context where it actually makes sure the prior list is flushed/serialized to log pages and log pages are effectively written to disk (assuming this is what the complex function `logpb_write_append_pages_to_disk` does).

In the given context (`log_initialize_internal` - `log_recovery_finish_transactions` - `logpb_flush_pages_direct` - `logpb_flush_all_append_pages` - `logpb_skip_flush_append_pages`) this might be fine since asserts following call to `log_recovery_finish_transactions` in function `log_initialize_internal` only check against these bookkeeping LSA's to ensure consistency.

But are there any other contexts where calling `logpb_flush_all_append_pages` in TSRS context actually need to have log pages effectively written to disk on the PS? I am thinking about transaction commits, for instance. In which case, this function would need a roundtrip to PS and back for acknowledgement.

ribram: > But are there any other contexts where calling logpb_flush_all_append_pages in TSRS context actually need to have log pages effectively written to disk on the PS? I am thinking about transaction commits, for instance. In which case, this function would need a roundtrip to PS and back for acknowledgement.

Normally no. Or better said, in those contexts it should call `logpb_flush_pages` instead of calling `logpb_flush_all_append_pages`.  In normal circumstances (after boot), only the log flush thread does the actual writing into pages and disk. Other threads request a flush up to a target and get blocked until that target is met. `logpb_flush_pages` waits until PS saves to disk and acknowledges writing.

PR #3104, opened 2021-10-19
[CBRD-23960] Support OID handling functionalities

hgryoo:

This PR implements Server-Side CUBRIDOID.
- `method_oid_handler` implements functionalities of CUBRIDOID
- `method_struct_oid_info` defines requests and responses of them
- Unit tests are implemented in demo/TestCUBRIDOID.java
- `include_oid` is considered in `method_query_handler` and `method_query_cursor`
  - if `include_oid` is set, the first column is set to current_oid in method_query_cursor


.gitmodules
```diff
@@ -4,5 +4,5 @@
 	branch = develop
 [submodule "cubrid-jdbc"]
 	path = cubrid-jdbc
-	url = https://github.com/cubrid/cubrid-jdbc
-	branch = feature/javasp_redesign
+	url = https://github.com/hgryoo/cubrid-jdbc
+	branch = support_oid
```
ctshim: Are you going to restore the edited contents here later?
If so, you do not need to answer.

demo/test/TestCUBRIDPreparedStatement.java
```diff
@@ -145,7 +145,7 @@ public static String test8() throws SQLException {
             stmt.execute();
             ResultSet rs = SqlUtil.executeQuery(conn, "select * from t1");
             rs.next();
-            if ((byte) 10 != (byte) rs.getObject("a")) {
+            if ((byte) 10 == (byte) rs.getObject("a")) {
```
ctshim: Shouldn't it return "t" if "a" is 10?
So != seems to be correct here.

src/jsp/com/cubrid/jsp/impl/SUConnection.java
```diff
@@ -30,174 +30,295 @@
...
+        packer.packInt(SUFunctionCode.GET_BY_OID.getCode());
+        packer.packOID(new SOID(oid.getOID()));
+        packer.packInt(attributeName.length);
+        for (int i = 0; attributeName != null && i < attributeName.length; i++) {
+            if (attributeName[i] != null) {
+                packer.packString(attributeName[i]);
+            } else {
+                packer.packString("");
+            }
+        }
```
ctshim: It doesn't seem necessary to repeatedly check the attributeName != null condition inside the for loop.
How about changing it to check once outside the for loop?
I think you should check before doing packer.packInt(attributeName.length);

src/jsp/com/cubrid/jsp/impl/SUConnection.java
```diff
@@ -30,174 +30,295 @@
...
+    }
+
+    // UFunctionCode.PUT_BY_OID
+    public void putByOID(CUBRIDOID oid, String[] attributeName, Object values[])
+            throws IOException {
+        CUBRIDPacker packer = new CUBRIDPacker(outputBuffer);
+        packer.packInt(SUFunctionCode.PUT_BY_OID.getCode());
+        packer.packOID(new SOID(oid.getOID()));
+        packer.packInt(attributeName.length);
+        for (int i = 0; attributeName != null && i < attributeName.length; i++) {
```
ctshim: It doesn't seem necessary to repeatedly check the attributeName != null condition inside the for loop.
How about changing it to check once outside the for loop?
I think you should check before doing packer.packInt(attributeName.length);

demo/test/TestCUBRIDPreparedStatement.java
beyondykk9: Maybe, the test case looks changed on this version.

src/jsp/com/cubrid/jsp/impl/SUConnection.java
```diff
@@ -30,174 +30,295 @@
...
+        packer.packInt(command);
+        packer.packOID(new SOID(oid.getOID()));
+
+        CUBRIDUnpacker unpacker = request(outputBuffer);
+
+        int result = unpacker.unpackInt();
+        if (command == CUBRIDServerSideConstants.IS_INSTANCE) {
+            if (result == 1) {
+                return oid;
+            }
```
ctshim, line 238: If result != 1, is it correct to return result?
Shouldn't it be necessary to return null?
The isInstance() function always returns true if the returned value is not null.

src/jsp/com/cubrid/jsp/jdbc/CUBRIDServerSideOID.java
```diff
@@ -0,0 +1,269 @@
...
+        try {
+            int page = Integer.parseInt(oidStringArray.nextToken().substring(1));
+            short slot = Short.parseShort(oidStringArray.nextToken());
+            short vol = Short.parseShort(oidStringArray.nextToken());
+
+            SOID sOid = new SOID(page, slot, vol);
+            return new CUBRIDServerSideOID((CUBRIDServerSideConnection) con, sOid);
+        } catch (NoSuchElementException e) {
+            throw new IllegalArgumentException();
+        }
```
ctshim, line 258: In the process of checking that the format is correct
"@123|45" will be treated as an error,
What about "@123|45|67|89" or "@ | | "?

src/jsp/com/cubrid/jsp/impl/SUConnection.java
hgryoo: Good point. I'm going to fix it to return `null`.

src/jsp/com/cubrid/jsp/jdbc/CUBRIDServerSideOID.java
hgryoo: If the OID string has an invalid format, the oidStringArray.nextToken() will throw an exception. The thrown exception will be treated as SQLException at the end.

src/method/method_invoke_java.cpp
```diff
@@ -390,16 +396,18 @@ namespace cubmethod
...
+	  std::uint64_t qid = current_result_info.query_id;
 	  const auto &iter = m_cursor_map.find (qid);
 
 	  query_cursor *cursor = nullptr;
 	  if (iter == m_cursor_map.end ())
 	    {
 	      /* not found, new cursor is created */
-	      cursor = m_cursor_map[qid] = new query_cursor (m_group->get_thread_entry(), qid);
+	      bool is_oid_included = current_result_info.include_oid;
+	      cursor = m_cursor_map[qid] = new query_cursor (m_group->get_thread_entry(), qid, is_oid_included);
```
ctshim: The new or new[] operator raises an exception if memory allocation fails.
You need to handle the exception or suppress the occurrence of the exception.

```
try
{  
  m_cursor_map[qid] = new query_cursor (m_group->get_thread_entry(), qid, is_oid_included);
}
catch( const std::bad_alloc& e )
{
   m_cursor_map[qid] = NULL;  
}
cursor = m_cursor_map[qid];
if (cursor == nullptr)
```

OR

```
cursor = m_cursor_map[qid] = new (std::nothrow) query_cursor (m_group->get_thread_entry(), qid, is_oid_included);
if (cursor == nullptr)
```

src/method/method_struct_oid_info.cpp
```diff
@@ -0,0 +1,375 @@
+/*
+ *
```
ctshim, line 2: This file should be indented as a whole.

src/method/method_oid_handler.cpp
```diff
@@ -0,0 +1,601 @@
...
+	    else
+	      {
+		set_type = DB_TYPE_NULL;
+		precision = db_domain_precision (domain);
+		scale = (short) db_domain_scale (domain);
+		charset = db_domain_codeset (domain);
+	      }
+	  }
+
+	// column_info info (db_type, set_type, scale, prec, charset, attr_name);
```
shparkcubrid: remove it.

src/jsp/com/cubrid/jsp/impl/SUConnection.java
```diff
@@ -30,174 +30,295 @@
...
 import java.nio.ByteBuffer;
-import java.sql.Connection;
 import java.util.ArrayList;
 import java.util.List;
 
-import cubrid.jdbc.jci.UError;
-
 public class SUConnection {
 
     List<SUStatement> stmts = null;
```
hyunikn: Elements are only added to this stmts list, but not retrieved from the list.
It looks OK to remove stmts.

src/jsp/com/cubrid/jsp/impl/SUConnection.java
hgryoo: I've fixed it. Thank you.

src/jsp/com/cubrid/jsp/impl/SUConnection.java
hgryoo: I've fixed it. Thank you.

src/jsp/com/cubrid/jsp/impl/SUConnection.java
hgryoo: Thank you so much!

src/method/method_oid_handler.cpp
hgryoo: fixed. thank you

demo/test/TestCUBRIDPreparedStatement.java
hgryoo: It is changed by accident. I've revert it.

src/jsp/com/cubrid/jsp/jdbc/CUBRIDServerSideStatement.java
```diff
@@ -516,6 +514,43 @@ public boolean execute(String sql, String[] columnNames) throws SQLException {
...
+    public CUBRIDOID executeInsert(String sql) throws SQLException {
+        try {
+            // 1) complete previous resultSet
+            completeResultSet();
+
+            // 2) prepare
+            prepareInternal(sql);
+
+            // 3) set autoGeneratedKeys
+            this.autoGeneratedKeys = autoGeneratedKeys;
```
hyunikn, line 538: This line looks unnecessary because it is 
this.autoGeneratedKeys = this.autoGeneratedKeys;    // ???

src/method/method_invoke_java.cpp
hyunikn: What about putting the handling at the entry point of a process or thread? instead of every usage of 'new' as described above. Any program will not work properly once memory allocation failure occurs, so we will have no other choices than shutting down gracefully. Writing error handling code for every usage of 'new' must be very burdensome.
Maybe, we need to discuss a coding standard on this issue.

hyunikn (approved):

shparkcubrid (approved):

beyondykk9 (approved):

src/jsp/com/cubrid/jsp/jdbc/CUBRIDServerSideStatement.java
hgryoo: Thank you. I'm going to fix it at another PR to support GetGenreatedKeys functionality.

src/method/method_invoke_java.cpp
ctshim: @hyunikn
That's a good comment. However, since the existing codes generally do not use the exception handling method, there seems to be a burden on the structure change to modify it.

hgryoo:
The windows build error caused by http://jira.cubrid.org/browse/CBRD-24090

youngjinj (approved):

PR #3105, opened 2021-10-19
[CBRD-24068] Add an option of the unloaddb that enables unloading data only in data volumes

wrlawodms:
http://jira.cubrid.org/browse/CBRD-24068


msg/en_US/utils.msg
```diff
@@ -522,6 +522,8 @@ valid options:\n\
   -d, --data-only             dump objects only; default: schema and objects\n\
       --output-prefix=PREFIEX PREFIX for output files; default: the database name\n\
       --hash-file=FILE        FILE for hash; default: generate by tmpnam()\n\
+      --latest-image          when dumping objects, it dumps the latest image of objects in data volumes.\n\
```
hornetmj: The latest image of objects in the page buffer can be different compared to them in the data volumes. Which version is unloaded when the user sets the --latest-image option? Or do you consider both cases to be categories of data volume? Please let me know your thought.

wrlawodms: It goes through the page buffer pool when unloading so they are the same. I thought the page buffer is another layer. If it seems confusing, we could say "it dumps the latest image of objects in data volumes, **not referring to log volumes (ignoring MVCC)**".

hornetmj: @wrlawodms Your suggestion is better than the old one. (My opinion in the previous comment said a different point. But, I think you don't need to consider it from the user's perspective.)

hornetmj (approved):

wrlawodms:
@hornetmj I have added multi-lingual msg files

hornetmj (approved):

wrlawodms:
@mhoh3963 It seems that `test_sql` is broken. Please merge this manually.

mhoh3963:
@wrlawodms 
I will force merge after 2 pending reviewers finish review.
@eido5 , @joohok please review.

joohok (approved):

wrlawodms:
@mhoh3963 I've merged develop and it works! I should've tried it first. Sorry.

PR #3106, opened 2021-10-19
[CBRD-23962] Implement Statement handler caching in CAS delegation routine

hgryoo:
http://jira.cubrid.org/browse/CBRD-23962

- It caches `method_query_handler` using SQL stmt string as key and `method_callback` class manages it.
- It caches only for SELECT query
- If `prepare()` is succeed, `method_query_handler`is registered on the cache.
- The occupied flag is set to true to indicate that one Statement of Server-side JDBC is in use, which prevents multiple Statements from being bound to one method_query_handler.


shparkcubrid (approved):

shparkcubrid:
sql error occurred
sql/_13_issues/_13_1h/cases/bug_bts_10489.sql - linux_sql_64bit_release

ctshim (approved):

hyunikn (approved):

src/method/method_callback.cpp
```diff
@@ -430,17 +450,24 @@ namespace cubmethod
...
+	else
+	  {
+	    m_query_handlers[id]->reset ();
+	  }
       }
   }
 
   void
-  callback_handler::free_query_handle_all ()
+  callback_handler::free_query_handle_all (bool is_free)
```
beyondykk9, line 400: this function looks called with "is_free = false", is "is_free" argument necessary ?

beyondykk9 (approved):

youngjinj (approved):

hgryoo:
The last commit is only for fixing code style,

PR #3107, opened 2021-10-20
[CBRD-24088] Assert_release error occurred when creating index

beyondykk9:
http://jira.cubrid.org/browse/CBRD-24088

cubrid createdb testdb ko_KR.utf8

csql>
create table tt (c1 varchar(50), c2 int);

insert into tt (
    select cast(rownum as varchar(50)), rownum from _db_class a, _db_class b, _db_class c
    union
    select cast(rownum || '  ' as varchar(50)), rownum from _db_class a, _db_class b
);

CREATE INDEX idx_tt on tt (c1);

ERROR: SYSTEM ERROR: Unable to load B+tree.

The error is caused by assert_release at **btree_get_prefix_separator** ,line 11805 or 11813.
In varchar type, not allowed ignoring trailing space as default configuration, but at splitting non-leaf node the trailing space is ignored.
So, the key comparing is not correct to find prefix of non-lead node.
I modified the code to split non-leaf node correctly not to ignore trailing space for VARCHAR type.

### DIFF
--- a/src/base/language_support.c
+++ b/src/base/language_support.c
@@ -5229,7 +5229,7 @@ lang_split_key_euckr (const LANG_COLLATION * lang_coll, const bool is_desc, cons
 	  str2_next = intl_nextchar_euc (str2, &char2_size);
 	  if (*str2 == ASCII_SPACE || *str2 == 0 || (*str2 == EUC_SPACE && char2_size == 2 && *(str2 + 1) == EUC_SPACE))
 	    {
-	      is_zero_weight = 0;
+	      is_zero_weight = 1;
 	    }
 	  str2 = str2_next;
 	  if (!is_zero_weight)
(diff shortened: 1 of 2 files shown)


hgryoo:
Can I know where assert_release() occurs and what it checks?

beyondykk9:
> Can I know where assert_release() occurs and what it checks?

at 11805 line of btree_get_prefix_separator

beyondykk9:
The below sql-test fail cases are caused by this modifying but others are related to https://github.com/CUBRID/cubrid/pull/3034

sql/_01_object/_10_system_table/_011_db_index_key/cases/1021.sql
sql/_13_issues/_14_1h/cases/bug_bts_12230_iso_bin_coll.sql

hgryoo (approved):

shparkcubrid (approved):

PR #3109, opened 2021-10-20
[LETS-250] transform global variable into a pointer and only initialize on transaction server

cristiarg:
http://jira.cubrid.org/browse/LETS-250

Avoid a crash in `tran_server::~tran_server` which does not expect to be called on a page server by not initializing `ats_Gl` global variable unless actually on an [active] transaction server.

### DIFF
--- a/src/server/active_tran_server.cpp
+++ b/src/server/active_tran_server.cpp
@@ -31,7 +31,7 @@
 #include <string>
 #include <utility>
 
-active_tran_server ats_Gl;
+std::unique_ptr<active_tran_server> ats_Gl;
 
 bool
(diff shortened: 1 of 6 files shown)


src/server/active_tran_server.cpp
```diff
@@ -31,7 +31,7 @@
 #include <string>
 #include <utility>
 
-active_tran_server ats_Gl;
+std::unique_ptr<active_tran_server> ats_Gl;
```
cristiarg, line 34: unique_ptr is not actually needed

cristiarg:
Another solution was implemented.

PR #3112, opened 2021-10-21
[CBRD-23620] It can't be found free pages, even if there are free pages on the temp temp volume

mhoh3963:
http://jira.cubrid.org/browse/CBRD-23620

fixed a problem which didn't be found free page from temp temp vol in certain cases.
And, it has been changed to expand only as much as intended in the case of temp temp volume.
The calculation method of disk_Temp_max_sects has been modified to avoid user confusion due to the difference in units between page and sector.


src/storage/disk_manager.c
```diff
@@ -1694,10 +1695,18 @@ disk_extend (THREAD_ENTRY * thread_p, DISK_EXTEND_INFO * extend_info, DISK_RESER
   assert (disk_Cache->owner_extend == thread_get_entry_index (thread_p));
 
   /* expand */
-  /* what is the desired remaining free after expand? */
-  target_free = MAX ((DKNSECTS) (total * 0.01), DISK_MIN_VOLUME_SECTS);
-  /* what is the desired expansion? do not expand less than intention. */
-  nsect_extend = MAX (target_free - free, 0) + intention;
+  if (voltype == DB_TEMPORARY_VOLTYPE)
+    {
+      nsect_extend = intention;
```
wrlawodms: Could you explain why we should extend temp volume by only `intention`?

src/storage/disk_manager.c
```diff
@@ -4939,9 +4957,15 @@ disk_manager_init (THREAD_ENTRY * thread_p, bool load_from_disk)
     {
       disk_Temp_max_sects = SECTID_MAX;	/* infinite */
     }
+  else if (disk_Temp_max_sects == 0)
+    {
+      disk_Temp_max_sects = 0;
+    }
```
wrlawodms:
```suggestion
  else if (disk_Temp_max_sects == 0)
    {
      // nothing
    }
```

src/storage/disk_manager.c
```diff
@@ -4939,9 +4957,15 @@ disk_manager_init (THREAD_ENTRY * thread_p, bool load_from_disk)
...
   else
     {
-      disk_Temp_max_sects = disk_Temp_max_sects / DISK_SECTOR_NPAGES;
+      disk_Temp_max_sects < (DISK_MIN_VOLUME_SECTS * DISK_SECTOR_NPAGES) ? (DISK_MIN_VOLUME_SECTS * DISK_SECTOR_NPAGES) : disk_Temp_max_sects;
```
wrlawodms: Did you mean to do this?
```suggestion
      disk_Temp_max_sects = disk_Temp_max_sects < (DISK_MIN_VOLUME_SECTS * DISK_SECTOR_NPAGES) ? (DISK_MIN_VOLUME_SECTS * DISK_SECTOR_NPAGES) : disk_Temp_max_sects;
```

src/storage/disk_manager.c
```diff
@@ -4939,9 +4957,15 @@ disk_manager_init (THREAD_ENTRY * thread_p, bool load_from_disk)
...
+      disk_Temp_max_sects = ceil (disk_Temp_max_sects / (double) DISK_SECTOR_NPAGES);
+      disk_Temp_max_sects = ceil (disk_Temp_max_sects / (double) DISK_MIN_VOLUME_SECTS) * DISK_MIN_VOLUME_SECTS;
```
wrlawodms: Why `disk_Temp_max_sects` has to be multiple of DISK_MIN_VOLUME_SECTS?

src/storage/disk_manager.c
wrlawodms: or you can do like this
```suggestion
      disk_Temp_max_sects = MAX (DISK_MIN_VOLUME_SECTS * DISK_SECTOR_NPAGES, disk_Temp_max_sects);
```

src/storage/disk_manager.c
```diff
@@ -4410,28 +4419,37 @@ disk_reserve_from_cache (THREAD_ENTRY * thread_p, DISK_RESERVE_CONTEXT * context
...
+	      return NO_ERROR;
+	    }
+         }
+
+      if (extend_info->nsect_total + context->n_cache_reserve_remaining >= disk_Temp_max_sects)
 	{
 	  /* too much temporary space */
 	  assert (false);
-	  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_BO_MAXTEMP_SPACE_HAS_BEEN_EXCEEDED, 1, disk_Temp_max_sects);
+	  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_BO_MAXTEMP_SPACE_HAS_BEEN_EXCEEDED, 1, disk_Temp_max_sects * DISK_SECTOR_NPAGES);
```
wrlawodms: If disk_Temp_max_sects is `SECTID_MAX` which is the default, it will be overflowed.

src/storage/disk_manager.c
wrlawodms: Why is it not set to the minimum value (`DISK_MIN_VOLUME_SECTS`) when it is 0?

src/storage/disk_manager.c
```diff
@@ -4410,28 +4419,37 @@ disk_reserve_from_cache (THREAD_ENTRY * thread_p, DISK_RESERVE_CONTEXT * context
...
+          if (context->n_cache_reserve_remaining <= 0)
+	    {
+	      /* found enough sectors */
+	      assert (context->n_cache_reserve_remaining == 0);
+	      disk_cache_unlock_reserve_for_purpose (context->purpose);
+	      return NO_ERROR;
+	    }
+         }
+
+      if (extend_info->nsect_total + context->n_cache_reserve_remaining >= disk_Temp_max_sects)
```
wrlawodms, line 4466: It seems that you should consider free sectors. There could be free sectors which of the number is less than `context->n_cache_reserve_remaining`.

src/storage/disk_manager.c
```diff
@@ -4410,28 +4419,37 @@ disk_reserve_from_cache (THREAD_ENTRY * thread_p, DISK_RESERVE_CONTEXT * context
...
+	{
+	  disk_reserve_from_cache_vols (extend_info->voltype, context);
+          if (context->n_cache_reserve_remaining <= 0)
+	    {
+	      /* found enough sectors */
+	      assert (context->n_cache_reserve_remaining == 0);
+	      disk_cache_unlock_reserve_for_purpose (context->purpose);
+	      return NO_ERROR;
+	    }
+         }
```
wrlawodms: Why is this code required even though it seems done after the big IF (`if (context->purpose == DB_TEMPORARY_DATA_PURPOSE)`) embracing this code? (from line 4450 in the original code).

wrlawodms (commented):
Regarding disk_reserve_from_cache(), you seem to be changing the spec like below:
AS-IS:
When to reserve sectors, the disk manager prioritizes a permanent type volume. For example, when 5 sectors are requested and  there are 3 sectors left in a perm type volume, and 10 sectors left in a temp volume, 3 sectors in the perm volume and 2 sectors in a temp volume are reserved.

TO-BE:
In the same case above, only 5 sectors in a temp volume is reserved.  

What I understand is right and what you intended? To review the code, please define what the problem was and elaborate on the changed design of sector reservation to handle the problem.

src/storage/disk_manager.c
wrlawodms: It has to be casted as long type and the message format has be changed together.

mhoh3963:
> Regarding disk_reserve_from_cache(), you seem to be changing the spec like below: AS-IS: When to reserve sectors, the disk manager prioritizes a permanent type volume. For example, when 5 sectors are requested and there are 3 sectors left in a perm type volume, and 10 sectors left in a temp volume, 3 sectors in the perm volume and 2 sectors in a temp volume are reserved.
> 
> TO-BE: In the same case above, only 5 sectors in a temp volume is reserved.
> 
> What I understand is right and what you intended? To review the code, please define what the problem was and elaborate on the changed design of sector reservation to handle the problem.

@wrlawodms 
Thank you for review.
It is not intended to change the specification in this section. This is a wrong fixed code. 
I will revert this code.

src/storage/disk_manager.c
```diff
@@ -4410,28 +4419,37 @@ disk_reserve_from_cache (THREAD_ENTRY * thread_p, DISK_RESERVE_CONTEXT * context
       /* if we want to allocate temporary files, we have two options: preallocated permanent volumes (but with the
        * purpose of temporary files) or temporary volumes. try first the permanent volumes */
 
-      extend_info = &disk_Cache->temp_purpose_info.extend_info;
-
-      if (disk_Cache->temp_purpose_info.nsect_perm_free > 0)
+      if (disk_Cache->temp_purpose_info.nsect_perm_free >= context->n_cache_reserve_remaining)
```
mhoh3963:
```suggestion
      if (disk_Cache->temp_purpose_info.nsect_perm_free > 0)
```

src/storage/disk_manager.c
mhoh3963: It be created two temp temp volumes for getting a free temp page, when creating database with 64 M db volume size which is the minimum unit.
So, I modified to prevent unnecessary temp volume from being additionally created.

src/storage/disk_manager.c
mhoh3963: Thank you. This code is unnecessary, so I will remove it.

hornetmj:
@mhoh3963 Can I review this PR after you fix the indent violation?

mhoh3963:
@hornetmj
completed to fix code style.

src/storage/disk_manager.c
mhoh3963: When expanding a volume, the unit of expansion is the DISK_MIN_VOLUME_SECTS (4096 pages = 64x64) value, not the DISK_SECTOR_NPAGES (64 pages) value. (The minimum size of DB volume with 16K page is 64M.)
And, the calculated values ​​was rounded up to reflect the user's intent as much as possible.
- The disk_Temp_max_sects is 64 (4096 pages), if temp_file_max_size_in_pages is a value between 1  and 4096.
- The disk_Temp_max_sects is 128 (8192 pages), if temp_file_max_size_in_pages is 4097.

src/storage/disk_manager.c
mhoh3963: thank your comment, I will modify by your comment.

src/storage/disk_manager.c
mhoh3963: The existing code occurred an error even if there is free space in temp temp vol, so it modified as follows.
- Before checking with disk_Temp_max_sects, it check if there is free space in the temp temp vol.
- if existing, it modified to reserve free pages and return.
- if not existing, it compare the values ​​of (nsect_total + n_cache_reserve_remaining) with disk_Temp_max_sects to check if can expand . It OK, proceed with the expansion. Otherwise, an error is raised.

src/storage/disk_manager.c
mhoh3963: See the comment above explained for line 4448.

hornetmj:
@mhoh3963 I executed the 'Repro' in the Jira and I faced the same problem again. Please check it again. And The changed error message still doesn't seem to be correct. The temp_file_max_size_in_pages system parameter is set to 2000. However, the error message shows 4096 pages like the following:

ERROR: The 4096 pages of total temporary space allowed have been exceeded.

mhoh3963:
@hornetmj 
When run test.sh, you can see count result after message ("run select.sql again") as following.
Do you get error, when run test.sh ?

=== <Result of SELECT Command in Line 9> ===

              count(*)
======================
               1000000 

And I modified to display the actual extended volume page, not value of temp_file_max_size_in_pages system parameter.

src/storage/disk_manager.c
```diff
@@ -4410,28 +4419,38 @@ disk_reserve_from_cache (THREAD_ENTRY * thread_p, DISK_RESERVE_CONTEXT * context
...
+	    }
+	}
+
+      if (extend_info->nsect_total + context->n_cache_reserve_remaining >= disk_Temp_max_sects)
 	{
 	  /* too much temporary space */
 	  assert (false);
-	  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_BO_MAXTEMP_SPACE_HAS_BEEN_EXCEEDED, 1, disk_Temp_max_sects);
+	  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_BO_MAXTEMP_SPACE_HAS_BEEN_EXCEEDED, 1,
+		  (long) (disk_Temp_max_sects * DISK_SECTOR_NPAGES));
```
wrlawodms: still could be overflowed.
```suggestion
		  (long) disk_Temp_max_sects * DISK_SECTOR_NPAGES);
```

msg/ko_KR.utf8/cubrid.msg
```diff
@@ -663,7 +663,7 @@ $ LOADDB
 593 갱신할 수 없는 가상 인스턴스가 회수되었으며 재생성할 수 없습니다.
 594 object_id 절에 approximate 데이터 타입(float, double)을 가지는 애트리뷰트가 올 수 없습니다. 
 595 데이터베이스 이름 "%1$s"이(가) 너무 깁니다. 데이터베이스 이름은 %2$d자 이내여야 합니다.
-596 허용된 전체 임시 공간의 %1$s 페이지들이 초과되었습니다.
+596 허용된 전체 임시 공간의 %1$d 페이지들이 초과되었습니다.
```
wrlawodms:
```suggestion
596 허용된 전체 임시 공간의 %1$lld 페이지들이 초과되었습니다.
```

src/storage/disk_manager.c
wrlawodms: To me, the original code seems reasonable because it prevents disk extension from triggering right after an extension. The disk extension is expensive operation so that It is good to extend more than enough to accept more disk reservation.

Could you let me know the case two volumes are created when getting a page?

hornetmj:
@mhoh3963 I didn't use test.sh. I just do a manual test. About the error message, the temp_file_max_size_in_pages system parameter makes the limitation of the temp file size by page count. The error message tells that our system didn't follow the system parameter setting.

src/storage/disk_manager.c
wrlawodms: It will raise an error even when the number of sectors will less than `disk_Temp_max_sects`. For example, let's say like below:
extend_info->nsect_total = 7
extend_info->nsect_free = 3
context->n_cache_reserve_remaining = 4 
disk_Temp_max_sects = 10

While sectors can be reserved by using free sectors (3) and an extended sector (1), in which the total number of sectors will be only 8 (<10), it will raise error here.

src/storage/disk_manager.c
wrlawodms: `DISK_MIN_VOLUME_SECTS` is a minimum size of a newly create volume, not the unit of extending sectors of a volume. Isn't it? I mean the number of sectors of a volume can be `DISK_MIN_VOLUME_SECTS` ~ X, not `DISK_MIN_VOLUME_SECTS`, `DISK_MIN_VOLUME_SECTS` * 2, `DISK_MIN_VOLUME_SECTS` *3, ... .

wrlawodms: And this means we don't need to set `disk_Temp_max_sects` as a multiple of `DISK_MIN_VOLUME_SECTS`. Could you tell me what I missed?

wrlawodms (changes requested):

src/storage/disk_manager.c
wrlawodms: I don't quit follow the point of doing it earlier. If this condition is true, the original disk_Temp_max_sects checking is always passed and this will done on the code below (at 4450).

extend_info->nsect_free >= context->n_cache_reserve_remaining
-> extend_info->nsect_free - context->n_cache_reserve_remaining >= 0
-> extend_info->nsect_total - extend_info->nsect_free + context->n_cache_reserve_remaining = extend_info->nsect_total - x(positive)
And 'nsect_total' is lesser than disk_Temp_max_sects at this point, so nsect_total + x is also lesser than it.

src/storage/disk_manager.c
wrlawodms: Oh, I found out that the extension unit is also 64M as you said, but it doesn't use `DISK_MIN_VOLUME_SECTS`. I think you want to use `DISK_SECTS_ROUND_UP()`, which is used when to extend.

src/storage/disk_manager.c
mhoh3963: thank your review, i will fix it.

msg/ko_KR.utf8/cubrid.msg
mhoh3963: thank you, i will apply it to all cubrid.msg files.

mhoh3963:
@hornetmj @wrlawodms 
I modified codes as the following, please start to review.
If you need to review at the off-line, let me know.

- fix to run test scenarios in the debug mode by removing assert. 
- modify how to find free spaces at temp volumes.
    AS-IS)
      * find free spaces at permanent temp volumes.
      * check that temp volume can extend, if don't find or enough its at permanent temp volumes.
    TO-BE)
      * find free spaces at permanent temp volumes.
      * find free spaces at temporary temp volumes.
      * check that temp volume can extend, if don't find or enough its at temp volumes (permanent and temporary volumes).
- separate codes to find free spaces on temp and data volume for readability.
- change that temporary temp volume doesn't extend over than disk_Temp_max_sects at disk_extend().
- didn't move a limit checking part of temporary temp volume to disk_extend(), because wait by mutex 
- when executing repro with temp_file_max_size_in_pages=2000 , you can face error because I modify to extend temporary volume to 4096 pages not 8192 pages. you need to set system parameter (max_pages_in_temp_file_cache=100) for facing success.

hornetmj:
@mhoh3963 The review started late. Sorry for this. We will proceed with the review. (@wrlawodms, @eido5)

hornetmj:
I close this PR because there has been no progress for a long time. It would be nice to make another PR after applying feedbacks or changing the way to solve the problem.

PR #3117, opened 2021-10-21
[LETS-247] Finalize thread manager after dumping error message

ribram:
http://jira.cubrid.org/browse/LETS-247

When thread manager is finalized, the error context on the main thread is lost. Calling `er_msg()` without error context crashes intentionally.

Finalize thread manager after printing `er_msg()` to avoid a crash.

### DIFF
--- a/src/executables/server.c
+++ b/src/executables/server.c
@@ -438,9 +438,7 @@ main (int argc, char **argv)
 	PRINT_AND_LOG_ERR_MSG ("%s\n", er_msg ());
 	fflush (stderr);
       }
-    // *INDENT-OFF*
-    cubthread::finalize ();
-    // *INDENT-ON*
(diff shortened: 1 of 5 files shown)


adrian-brabete (approved):

adrianbrab (approved):

src/executables/server.c
```diff
@@ -438,6 +438,9 @@ main (int argc, char **argv)
 	PRINT_AND_LOG_ERR_MSG ("%s\n", er_msg ());
 	fflush (stderr);
       }
+    // *INDENT-OFF*
+    cubthread::finalize ();
```
cristiarg: better a pair function for `thread_initialize_manager` that is called above?

cristiarg (approved):

PR #3119, opened 2021-10-21
[CBRD-24083] Even if only the column name is given in where, it evaluates to true or false.

ctshim:
http://jira.cubrid.org/browse/CBRD-24083

Purpose
Conditional clauses with only constant or column names are treated as errors. Also, a conditional clause must consist only of operations between logical expressions.

Implementation
If only a constant or column name comes alone, it was previously created as EXPR with the condition of "<> 0" added, but it is changed to prevent an error.

Remarks
N/A


src/parser/csql_grammar.y
```diff
@@ -142,11 +142,19 @@ extern int yybuffer_pos;
 #define PRINT_(a) printf(a)
 #define PRINT_1(a, b) printf(a, b)
 #define PRINT_2(a, b, c) printf(a, b, c)
+#define DBG_PRINT_MATCH(...)    do { fprintf(stdout, "  *** RULE) "__VA_ARGS__);  fflush(stdout); } while(0)
+#define DBG_PRINT_MATCH_LN(...) do {                        \
+                fprintf(stdout, "  *** RULE) "__VA_ARGS__); \
+                fprintf(stdout, "\n");                      \
+                fflush(stdout);                             \
+              } while(0)
```
ctshim, line 152: The added macro is a new addition because it is difficult to know which part is being processed with the existing DBG_PRINT.
Since then, there is a need to keep adding, so we want to keep it going.

hyunikn (approved):

beyondykk9 (approved):

shparkcubrid (approved):

hgryoo (approved):

youngjinj (approved):

PR #3123, opened 2021-10-22
[LETS-258] Fix and update test_checkpoint_info unit test after addition of tdes::commit_abort_lsa

cristiarg:
http://jira.cubrid.org/browse/LETS-258

commit_abort_lsa is a new criteria used by the checkpoint mechanism to decide whether a transactions needs to be considered for checkpointing or can actually be considered fully commited/aborted.
See https://jira.cubrid.org/browse/CBRD-24026 and corresponding patch https://github.com/CUBRID/cubrid/commit/8c0997283c0d48fce1fceaa7d147e5f4c7ea432b .

- fix unit test: the log_tdes::commit_abort_lsa is initialized in generate_tdes function similar as it is in logtb_initialize_tdes
- extend unit test: added extra transaction descriptors for which log_tdes::commit_abort_lsa is non-null; these should be skipped by the checkpoint

### DIFF
--- a/unit_tests/log/test_main_chkpt_info.cpp
+++ b/unit_tests/log/test_main_chkpt_info.cpp
@@ -739,21 +739,20 @@ logtb_get_system_tdes (THREAD_ENTRY *thread_p)
 LOG_TDES *
 logtb_rv_find_allocate_tran_index (THREAD_ENTRY *thread_p, TRANID trid, const LOG_LSA *log_lsa)
 {
-  std::unique_ptr<log_tdes> tdes {new log_tdes ()};		/* Transaction descriptor */
-
   assert (trid != NULL_TRANID);
 
   if (trid < NULL_TRANID)
     {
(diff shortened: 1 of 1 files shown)


unit_tests/log/test_main_chkpt_info.cpp
```diff
@@ -696,13 +739,13 @@ logtb_get_system_tdes (THREAD_ENTRY *thread_p)
 LOG_TDES *
 logtb_rv_find_allocate_tran_index (THREAD_ENTRY *thread_p, TRANID trid, const LOG_LSA *log_lsa)
 {
-  LOG_TDES *tdes = new log_tdes ();		/* Transaction descriptor */
+  std::unique_ptr<log_tdes> tdes {new log_tdes ()};		/* Transaction descriptor */
 
   assert (trid != NULL_TRANID);
```
ribram: Move `tdes` allocation after the `if (trid < NULL_TRANID)`.

ribram (approved):

PR #3124, opened 2021-10-24
[LETS-260] Crash involving metalog in createdb, restoredb and copydb sceanarios

cristiarg:
http://jira.cubrid.org/browse/LETS-260

`createdb` and `copydb` have a `-r` option where it is expected that there might already be an existing destination database in place which must be replaced.
In such scenarios, a metalog is first initialized with the contents of the to-be-replaced database after which the newly created, or newly copied, database is opened and the metalog should once more be initialized with new contents.

Metalog must be cleared when the to-be-removed database is closed.


ribram:
Please explain, is meta-log loaded twice during restoredb?

ribram (approved):

ribram (approved):

PR #3128, opened 2021-10-26
[CBRD-23842] modified the way of error logging in cubrid_log API

joohok:

Error logging in API level for cubrid_log api. It does only appended ERROR location, and uses only one error log file which can be overwritten when the error log file size exceeds the limit.  

This PR changed the way of error logging. 
* The error log file will be separated into 2 files, latest one and old one. never overwrite on one of those. 
* The function and Error code will be described 
* when the trace log level is greater than 0, then it leaves more messages.


joohok:
not ready for review, I'll re-request review when the test is done.

src/api/cubrid_log.c
```diff
@@ -45,20 +53,33 @@
...
+#define CUBRID_LOG_TRACELOG(msg, ...) \
+  do \
+    { \
+      if(cubrid_log_check_tracelog () == CUBRID_LOG_SUCCESS) \
+        { \
+          cubrid_log_get_time (); \
+          fprintf (g_trace_log,"[%s][FILE:%s][LINE:%d]\n" msg "\n", g_curr_time, __FILE__, __LINE__, ##__VA_ARGS__); \
+        } \
+    } \
+  while(0)
```
hornetmj: How about introducing a new function to process this mechanism together like cubrid_log_write_tracelog()?

src/api/cubrid_log.c
```diff
@@ -149,24 +218,146 @@ cubrid_log_set_extraction_timeout (int timeout)
...
+      memcpy (g_trace_log_path_old, g_trace_log_path, PATH_MAX + 1);
+    }
+  else
+    {
+      if (g_num_trace_log != 0)
+	{
+	  memcpy (g_trace_log_path_old, g_trace_log_path, PATH_MAX + 1);
+	}
+
+      g_num_trace_log++;
```
hornetmj:
```suggestion
```

src/api/cubrid_log.c
```diff
@@ -149,24 +218,146 @@ cubrid_log_set_extraction_timeout (int timeout)
...
+
+  snprintf (g_trace_log_path, PATH_MAX + 1, "%s/%s_cubridlog_%s.err", g_trace_log_base, g_dbname, curr_time);
+
+  g_trace_log = fopen (g_trace_log_path, "a+");
+  if (g_trace_log == NULL)
+    {
+      return -1;
+    }
+
+  return CUBRID_LOG_SUCCESS;
```
hornetmj, line 282:
```suggestion
  g_num_trace_log++;
  assert (g_num_trace_log <= 2);

  return CUBRID_LOG_SUCCESS;
```

src/api/cubrid_log.c
```diff
@@ -149,24 +218,146 @@ cubrid_log_set_extraction_timeout (int timeout)
...
+	{
+	  return cubrid_log_make_new_tracelog ();
+	}
     }
+
+  return CUBRID_LOG_SUCCESS;
+}
+
+static int
+cubrid_log_check_tracelog_file (char *path)
```
hornetmj:
```suggestion
cubrid_log_check_tracelog_path (char *path)
```

src/api/cubrid_log.c
```diff
@@ -179,15 +370,24 @@ cubrid_log_set_tracelog_pointer (char buf[])
...
     }
+  else
+    {
+      if ((err_code = cubrid_log_check_tracelog_file (path)) != CUBRID_LOG_SUCCESS)
+	{
+	  return err_code;
+	}
+    }
 
   if (level < 0 || level > 2)
```
hornetmj:
```suggestion
  if (level < 0 || level > 1)
```

src/api/cubrid_log.c
```diff
@@ -149,24 +218,146 @@ cubrid_log_set_extraction_timeout (int timeout)
...
- *   logsize(in):
- */
 static void
-cubrid_log_set_tracelog_pointer (char buf[])
+cubrid_log_reset_tracelog ()
+{
+  memset (g_trace_log_base, 0, PATH_MAX + 1);
+  memset (g_trace_log_path, 0, PATH_MAX + 1);
+  memset (g_trace_log_path_old, 0, PATH_MAX + 1);
+  g_num_trace_log = 0;
```
hornetmj, line 204: If g_num_trace_log is greater than zero, then does it need to close the already opened trace log file?

src/api/cubrid_log.c
```diff
@@ -206,15 +406,18 @@ cubrid_log_set_tracelog (char *path, int level, int filesize)
 	}
     }
 
-  snprintf (g_trace_log_path, PATH_MAX + 1, "%s", path);
-  g_trace_log_level = level;
-  g_trace_log_filesize = filesize * 1024 * 1024;
-  g_trace_log = fopen (path, "a+");
-  if (g_trace_log == NULL)
+  if (g_num_trace_log > 0)
```
hornetmj: Is it better to use the default setting when the scenario occurs in the comment?

src/api/cubrid_log.c
```diff
@@ -569,66 +768,69 @@ cubrid_log_connect_server (char *host, int port, char *dbname, char *id, char *p
...
+
+      strncpy (g_dbname, dbname, CUBRID_LOG_MAX_DBNAME_LEN);
+    }
+  else
+    {
+      CUBRID_LOG_ERROR_HANDLING (CUBRID_LOG_INVALID_DBNAME);
     }
 
-  if (g_stage != CUBRID_LOG_STAGE_CONFIGURATION)
+  if (g_trace_log_level > 0)
```
hornetmj:
```suggestion
  if (g_trace_log_level == 1)
```

src/api/cubrid_log.c
```diff
@@ -569,66 +768,69 @@ cubrid_log_connect_server (char *host, int port, char *dbname, char *id, char *p
...
     }
 
-  if (dbname == NULL)
+  if (port < 0 || port > USHRT_MAX)
     {
-      CUBRID_LOG_ERROR_HANDLING (CUBRID_LOG_INVALID_DBNAME, trace_errbuf);
+      CUBRID_LOG_ERROR_HANDLING (CUBRID_LOG_INVALID_PORT);
     }
 
   if (id == NULL)
```
hornetmj: Don't we replace 'id' with 'user' or 'user_name'?

src/api/cubrid_log.c
```diff
@@ -149,24 +192,184 @@ cubrid_log_set_extraction_timeout (int timeout)
...
+      if (access (path, W_OK) < 0)
+	{
+	  /* Write permission */
+	  return CUBRID_LOG_NO_FILE_PERMISSION;
+	}
     }
+  else
+    {
+      /* Directory not exist, then make directory */
+      if (mkdir (path, 0777) < 0)
```
hornetmj, line 330: How about changing the permission from 777 to 775 like CUBRID?

src/api/cubrid_log.c
```diff
@@ -149,24 +192,184 @@ cubrid_log_set_extraction_timeout (int timeout)
...
+    }
+  else
+    {
+      if (g_num_trace_log != 0)
+	{
+	  memcpy (g_trace_log_path_old, g_trace_log_path, PATH_MAX + 1);
+	}
+    }
+
+  snprintf (g_trace_log_path, PATH_MAX + 1, "%s/%s_cubridlog_%s.err", g_trace_log_base, g_dbname, curr_time);
```
hornetmj: In Windows, '/' doesn't work. Please refer to FILEIO_PATH_SEPARATOR macro.

src/api/cubrid_log.c
```diff
@@ -467,28 +676,34 @@ cubrid_log_send_configurations (void)
...
   if (reply_code == ER_CDC_NOT_AVAILABLE)
     {
-      CUBRID_LOG_ERROR_HANDLING (CUBRID_LOG_UNAVAILABLE_CDC_SERVER, trace_errbuf);
+      CUBRID_LOG_ERROR_HANDLING (CUBRID_LOG_UNAVAILABLE_CDC_SERVER,
+				 "Failed to connect to the server. Please check 'supplemental_log' parameter\n");
     }
   else if (reply_code != NO_ERROR)
     {
-      CUBRID_LOG_ERROR_HANDLING (CUBRID_LOG_FAILED_CONNECT, trace_errbuf);
+      CUBRID_LOG_ERROR_HANDLING (CUBRID_LOG_FAILED_CONNECT, NULL);
```
hornetmj: Is it possible to set it to NULL? This means the followings:

In the cubrid_log_tracelog() function.
- va_start (arg_ptr, **NULL**);
- vfprintf (g_trace_log, **NULL**, arg_ptr);

src/api/cubrid_log.c
joohok: I think it is better, no write permission to other users will be good. I will change it to 775

src/api/cubrid_log.c
joohok: OK, thanks. I will follow it.

src/api/cubrid_log.c
joohok: No, 
Second argument for CUBRID_LOG_ERROR_HANDLING indicates 'msg'. 
and if it is NULL, it just set the error code and goto error label (cubrid_log_error:) 

However, it needs to be set some error message for more clear information about why "CUBRID_LOG_FAILED_CONNECT" error is returned.

src/api/cubrid_log.c
joohok: CUBRID uses permissions of 777 or 755, and mostly used 777. 
However, I think anyone except owner is not allowed to write on an error log file. So how about change it to 755?

hornetmj: Leave it to 777. Let's change it later if it becomes a problem.

src/api/cubrid_log.c
hornetmj: Please leave additional information for us to find the problem easily when the error has occurred in the customer environment.

hornetmj:
@joohok Please let me know so I can review again.

hornetmj (approved):

PR #3129, opened 2021-10-26
[CBRD-23960] Support Server-side CallableStatement

hgryoo:

This PR supports CallableStatement functionality except returning ResultSet (Cursor).

### DIFF
--- a/src/method/method_query_handler.cpp
+++ b/src/method/method_query_handler.cpp
@@ -27,6 +27,16 @@
 
 namespace cubmethod
 {
+  prepare_call_info::prepare_call_info ()
+  {
+    //
+  }
+
+  prepare_call_info::~prepare_call_info ()
+  {
+    clear ();
+  }
+
   int
   prepare_call_info::set_is_first_out (std::string &sql_stmt)
   {
@@ -45,6 +55,15 @@ namespace cubmethod
       }
 
     return NO_ERROR;
+  }
+
+  void
+  prepare_call_info::clear ()
+  {
+    for (int i = 0; i < dbval_args.size(); i++)
+      {
+	db_value_clear (&dbval_args[i]);
+      }
   }
 
   int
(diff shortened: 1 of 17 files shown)


src/method/method_struct_query.cpp
```diff
@@ -475,6 +602,14 @@ namespace cubmethod
       }
   }
 
+  execute_info::execute_info ()
```
byungwook-kim, line 605: missed declare "void"

src/method/method_query_handler.cpp
```diff
@@ -444,7 +414,7 @@ namespace cubmethod
 
   int
   query_handler::execute_internal_call (execute_info &info, int flag, int max_col_size, int max_row,
-					std::vector<DB_VALUE> &bind_values)
+					const std::vector<DB_VALUE> &bind_values)
```
beyondykk9, line 417: max_row not used and if it will be used then it looks better to set as int64.

src/method/method_query_handler.cpp
```diff
@@ -547,7 +517,7 @@ namespace cubmethod
 
   int
   query_handler::execute_internal (execute_info &info, int flag, int max_col_size, int max_row,
-				   std::vector<DB_VALUE> &bind_values)
+				   const std::vector<DB_VALUE> &bind_values)
```
beyondykk9, line 520: max_row not used and if it will be used then it looks better to set as int64.

beyondykk9 (approved):

src/method/method_struct_query.cpp
```diff
@@ -526,6 +671,13 @@ namespace cubmethod
 	    column_infos[i].unpack (deserializator);
 	  }
       }
+
+    bool has_call_info = false;
+    deserializator.unpack_bool (has_call_info);
+    if (has_call_info)
+      {
+	call_info = new prepare_call_info (); // need to be freed;
```
shparkcubrid, line 687: I can't find free for call_info.

shparkcubrid (approved):

src/method/method_struct_query.cpp
hgryoo: I've fixed

PR #3134, opened 2021-10-29
[APIS-882] cci unintentionally changes bind type short/ushort to int/uint and sends it to CAS

airnet73:
http://jira.cubrid.org/browse/APIS-882

**Purpose**
When CCI_U_TYPE_SHORT is assigned to u_type of cci_bind_param() function, CCI sends CCI_U_TYPE_INT to CAS instead of CCI_U_TYPE_SHORT.
As a result, CAS receives the wrong type, CCI_U_TYPE_INT, instead of CCI_U_TYPE_SHORT.
This behavior can lead to errors in some situations, for example CAS Gateway to ODBC.

Since CCI_U_TYPE_INT and CCI_U_TYPE_SHORT are different types, they must be separated from each other.

**Implementation**
N/A

**Remarks**
N/A

### DIFF
--- a/src/cci/cci_query_execute.c
+++ b/src/cci/cci_query_execute.c
@@ -6563,22 +6563,16 @@ bind_value_conversion (T_CCI_A_TYPE a_type, T_CCI_U_TYPE u_type, char flag, void
     }
 
   bind_value->u_type = u_type;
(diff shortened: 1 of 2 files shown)


kisoo-han (approved):
Please note this changes and reflect it to CUBRID/cubrid-cci repo later.

PR #3136, opened 2021-10-29
[LETS-187] Server files path initialization refactoring

ribram:
http://jira.cubrid.org/browse/LETS-187

Move (and unify) the code managing the database path, the log path and the lob path respectively, from `boot_initialize_client` and `boot_initialize_server` to separate functions.

Other changes:

  - Move `boot_remove_useless_path_separator ()` to `boot.h` to be accessible on both `boot_cl` and `boot_sr`
  - Add `const` to `boot_db_path_info` fields.
  
 This change is a building block for creating a database with local permanent data and log files. The paths configuration is a common part that needs to be reused.

### DIFF
--- a/src/transaction/boot.h
+++ b/src/transaction/boot.h
@@ -101,12 +101,12 @@
 typedef struct boot_db_path_info BOOT_DB_PATH_INFO;
 struct boot_db_path_info
 {
-  char *db_path;
-  char *vol_path;
-  char *log_path;
-  char *lob_path;
-  char *db_host;
-  char *db_comments;
+  const char *db_path;
+  const char *vol_path;
+  const char *log_path;
+  const char *lob_path;
+  const char *db_host;
+  const char *db_comments;
 };
 
 /*
@@ -154,6 +154,15 @@ extern char boot_Host_name[CUB_MAXHOSTNAMELEN];
 
 #define LOB_PATH_PREFIX_MAX     ES_URI_PREFIX_MAX
 #define LOB_PATH_DEFAULT_PREFIX ES_POSIX_PATH_PREFIX
+#define ES_DEFAULT_TYPE         ES_POSIX
+
+// Inline functions
+inline void COMPOSE_FULL_NAME (char *buf, size_t buf_size, const char *path, const char *name);
+inline void boot_remove_useless_path_separator (const char *path, char *new_path);
+
+//
+// Inline implementation
+//
 
 /* Compose the full name of a database */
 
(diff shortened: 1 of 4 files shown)


src/transaction/boot_cl.c
```diff
@@ -231,6 +231,11 @@ static int boot_check_timezone_checksum (BOOT_CLIENT_CREDENTIAL * client_credent
 #endif
 static int boot_client_find_and_cache_class_oids (void);
 
+// *INDENT-OFF*
```
cristiarg: Does this need indentation rules?

src/transaction/boot_cl.c
```diff
@@ -321,6 +326,10 @@ boot_initialize_client (BOOT_CLIENT_CREDENTIAL * client_credential, BOOT_DB_PATH
   char format[BOOT_FORMAT_MAX_LENGTH];
 #endif
 
+  std::string processed_db_path;
```
cristiarg: Maybe this needs indentation rules.

src/transaction/boot_cl.c
```diff
@@ -726,6 +710,142 @@ boot_initialize_client (BOOT_CLIENT_CREDENTIAL * client_credential, BOOT_DB_PATH
   return error_code;
 }
 
+static int
```
cristiarg, line 713: Extract comment before function

src/transaction/boot_cl.c
```diff
@@ -726,6 +710,142 @@ boot_initialize_client (BOOT_CLIENT_CREDENTIAL * client_credential, BOOT_DB_PATH
...
+  boot_remove_useless_path_separator (path, remove_useless_sep_buf);
+  path = remove_useless_sep_buf;
+  
+  // done
+  strcpy (path_out, path);
+  return NO_ERROR;
+}
+
+static int
+boot_initialize_lob_path (const char *path_arg, const char *db_path, char *path_out)
```
cristiarg, line 763: Comment before function

src/transaction/boot_cl.c
```diff
@@ -726,6 +710,142 @@ boot_initialize_client (BOOT_CLIENT_CREDENTIAL * client_credential, BOOT_DB_PATH
...
+    {
+      // Convert to absolute path and check it exists
+      char absolute_path_buf[PATH_MAX];
+
+#if defined (WINDOWS)
+      struct stat stat_buf;
+      if (realpath (lob_dir_path, absolute_path_buf) != NULL
+	  && (stat (absolute_path_buf, &stat_buf) == 0 && S_ISDIR (stat_buf.st_mode)))
+#else
+      if (realpath (lob_dir_path, absolute_path_buf) != NULL)
```
cristiarg, line 819: for !`WINDOWS`, does not verify that the last path component is a directory

src/transaction/boot_cl.c
```diff
@@ -726,6 +710,142 @@ boot_initialize_client (BOOT_CLIENT_CREDENTIAL * client_credential, BOOT_DB_PATH
...
+	  if (mkdir (lob_dir_path, 0777) < 0)
+	    {
+	      cub_dirname_r (lob_dir_path, absolute_path_buf, PATH_MAX);
+	      er_set_with_oserror (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_ES_GENERAL, 2, "POSIX", absolute_path_buf);
+	      return ER_ES_GENERAL;
+	    }
+	}
+
+      // Remove useless separators and save the value to lob_dir_path (inside es_path)
+      boot_remove_useless_path_separator (absolute_path_buf, lob_dir_path);
```
cristiarg, line 837: Is this still needed? Doesn't `realpath` remove unneeded separators?

src/transaction/boot.h
```diff
@@ -175,4 +184,98 @@ COMPOSE_FULL_NAME (char *buf, size_t buf_size, const char *path, const char *nam
...
+ *       /home3/CUBRID/DB/               -->  /home3/CUBRID/DB
+ *       C:\CUBRID\\\Databases\\         -->  C:\CUBRID\Databases
+ *       \\pooh\user\                    -->  \\pooh\user
+ *
+ *       After transform..
+ *       If new path string is "/" or "\", don't remove the last slash.
+ *       It is survived.
+ */
+inline void
+boot_remove_useless_path_separator (const char *path, char *new_path)
```
cristiarg, line 207: This function should be guarded with `PATH_MAX`.

src/transaction/boot_cl.c
```diff
@@ -726,6 +710,142 @@ boot_initialize_client (BOOT_CLIENT_CREDENTIAL * client_credential, BOOT_DB_PATH
...
+	default:
+	  strncpy_bufsize (es_path, path_arg);
+	  break;
+	}
+    }
+
+  assert (es_type != ES_NONE);
+
+  // Get the path without the external storage prefix; the prefix ends with ':'
+  char *lob_dir_path = std::strchr (es_path, ':');
```
cristiarg, line 799: indent guides

src/transaction/boot.h
ribram: Please be more specific with your suggestion.

src/transaction/boot_cl.c
ribram: True. Not sure why there is this difference between OS's, but this is the code inherited from boot_sr.c. Since this is a refactoring task, I prefer keeping the code exactly the same way as before.

src/transaction/boot_cl.c
ribram: I am not sure; I assume it does not, since this is inherited from existing code.

src/transaction/boot_cl.c
ribram: done

src/transaction/boot_cl.c
ribram: I removed the strings, they're a loose end of code that has been removed.

src/transaction/boot_cl.c
ribram: done

src/transaction/boot_cl.c
ribram: done

src/transaction/boot_cl.c
ribram: done

src/transaction/boot_cl.c
cristiarg: Fair!

src/transaction/boot.h
cristiarg: I know the function is only moved, not newly implemented. :)
I was saying that it is not safe. If input `path` is larger than `PATH_MAX` there is no check to ensure that the function does not continue writing into the destination `new_path` beyond a reasonable `PATH_MAX` length.

ribram: I understand. But since this function is C-style, there is little the function can do for a real guarantee. E.g. we cannot guarantee that the caller really allocates PATH_MAX size for `new_path`.

I guess all we can do is to make the assertion that `std::strlen (path)` is less than PATH_MAX (although it is late to be caught here).

cristiarg: True that.

cristiarg (approved):

PR #3138, opened 2021-10-29
[LETS-214] Connect passive transaction server to the page server

adrianbrab:
http://jira.cubrid.org/browse/LETS-214

Created a new class for the passive transaction server, it inherits from the transaction server class.
This patch only creates the connection, and logs if successful.

### DIFF
--- a/src/connection/server_support.c
+++ b/src/connection/server_support.c
@@ -2635,6 +2635,15 @@ css_process_server_server_connect (SOCKET master_fd)
       ps_Gl.set_active_tran_server_connection (std::move (chn));
       ps_Gl.start ();
       break;
+    case cubcomm::server_server::CONNECT_PASSIVE_TRAN_TO_PAGE_SERVER:
+      // *INDENT-OFF*
+      if (!chn.send_int (static_cast <int>(cubcomm::server_server::CONNECT_PASSIVE_TRAN_TO_PAGE_SERVER)))
+        {
+          er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_NET_TRANSACTION_SERVER_CONNECTION, 0);
+          break;
+        }
+      // *INDENT-ON*
+      break;
     default:
       assert (false);
     }
--- a/src/connection/server_support.c
+++ b/src/connection/server_support.c
@@ -2626,7 +2626,7 @@ css_process_server_server_connect (SOCKET master_fd)
     {
     case cubcomm::server_server::CONNECT_ACTIVE_TRAN_TO_PAGE_SERVER:
       // *INDENT-OFF*
-      if (!chn.send_int (static_cast <int>(cubcomm::server_server::CONNECT_ACTIVE_TRAN_TO_PAGE_SERVER)))
+      if (!chn.send_int (static_cast<int> (cubcomm::server_server::CONNECT_ACTIVE_TRAN_TO_PAGE_SERVER)))
 	{
 	  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_NET_TRANSACTION_SERVER_CONNECTION, 0);
 	  break;
@@ -2637,7 +2637,7 @@ css_process_server_server_connect (SOCKET master_fd)
       break;
     case cubcomm::server_server::CONNECT_PASSIVE_TRAN_TO_PAGE_SERVER:
       // *INDENT-OFF*
-      if (!chn.send_int (static_cast <int>(cubcomm::server_server::CONNECT_PASSIVE_TRAN_TO_PAGE_SERVER)))
+      if (!chn.send_int (static_cast<int> (cubcomm::server_server::CONNECT_PASSIVE_TRAN_TO_PAGE_SERVER)))
         {
           er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_NET_TRANSACTION_SERVER_CONNECTION, 0);
           break;
--- a/src/server/passive_tran_server.cpp
+++ b/src/server/passive_tran_server.cpp
@@ -56,3 +56,41 @@ passive_tran_server::receive_saved_lsa (cubpacking::unpacker &upk)
       _er_log_debug (ARG_FILE_LINE, "[COMMIT CONFIRM] Received LSA = %lld|%d.\n", LSA_AS_ARGS (&saved_lsa));
     }
 }
+
+void
+passive_tran_server::on_boot ()
+{
+  assert (is_passive_transaction_server ());
+
+  cublog::prior_sender::sink_hook sink =
+	  std::bind (&passive_tran_server::push_request, std::ref (*this), tran_to_page_request::SEND_LOG_PRIOR_LIST,
+		     std::placeholders::_1);
+
+  log_Gl.m_prior_sender.add_sink (sink);
+}
+
+passive_tran_server::request_handlers_map_t
+passive_tran_server::get_request_handlers ()
+{
+  using map_value_t = request_handlers_map_t::value_type;
+
+  map_value_t boot_info_handler_value =
+	  std::make_pair (page_to_tran_request::SEND_BOOT_INFO,
+			  std::bind (&passive_tran_server::receive_boot_info, std::ref (*this), std::placeholders::_1));
+  map_value_t saved_lsa_handler_value =
+	  std::make_pair (page_to_tran_request::SEND_SAVED_LSA,
+			  std::bind (&passive_tran_server::receive_saved_lsa, std::ref (*this), std::placeholders::_1));
+  map_value_t log_page_handler_value =
+	  std::make_pair (page_to_tran_request::SEND_LOG_PAGE,
+			  std::bind (&passive_tran_server::receive_log_page, std::ref (*this), std::placeholders::_1));
+  map_value_t data_page_handler_value =
+	  std::make_pair (page_to_tran_request::SEND_DATA_PAGE,
+			  std::bind (&passive_tran_server::receive_data_page, std::ref (*this), std::placeholders::_1));
+
+  std::map<page_to_tran_request, std::function<void (cubpacking::unpacker &upk)>> handlers_map;
+
+  handlers_map.insert ({ boot_info_handler_value, saved_lsa_handler_value, log_page_handler_value,
+			 data_page_handler_value });
+
+  return handlers_map;
+}
--- a/src/server/passive_tran_server.cpp
+++ b/src/server/passive_tran_server.cpp
@@ -1,0 +1,58 @@
+/*
+ * Copyright 2008 Search Solution Corporation
+ * Copyright 2016 CUBRID Corporation
+ *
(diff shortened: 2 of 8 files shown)


src/server/passive_tran_server.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+#include "log_impl.h"
+#include "system_parameter.h"
+
+passive_tran_server pts_Gl;
+
+bool
+passive_tran_server::uses_remote_storage () const
+{
+  return m_uses_remote_storage;
```
cristiarg: Shouldn't this always be false for PTS?

src/server/passive_tran_server.hpp
```diff
@@ -0,0 +1,48 @@
...
+    request_handlers_map_t get_request_handlers () final override;
+
+    void receive_saved_lsa (cubpacking::unpacker &upk);
+
+  private:
+
+    bool m_uses_remote_storage = false;
+};
+
+extern passive_tran_server pts_Gl;
```
cristiarg: Not needed. [LETS-215](http://jira.cubrid.org/browse/LETS-215) will take care of instantiating either a PTS or an ATS instance via a pointer.

src/connection/server_support.c
```diff
@@ -2635,6 +2635,15 @@ css_process_server_server_connect (SOCKET master_fd)
       ps_Gl.set_active_tran_server_connection (std::move (chn));
       ps_Gl.start ();
       break;
+    case cubcomm::server_server::CONNECT_PASSIVE_TRAN_TO_PAGE_SERVER:
+      // *INDENT-OFF*
+      if (!chn.send_int (static_cast <int>(cubcomm::server_server::CONNECT_PASSIVE_TRAN_TO_PAGE_SERVER)))
```
ribram:
```suggestion
      if (!chn.send_int (static_cast<int> (cubcomm::server_server::CONNECT_PASSIVE_TRAN_TO_PAGE_SERVER)))
```

src/server/passive_tran_server.cpp
ribram: Yes, it should. Don't override `uses_remote_storage()` and remove `m_uses_remote_storage`.

src/server/passive_tran_server.cpp
```diff
@@ -0,0 +1,96 @@
...
+}
+
+bool
+passive_tran_server::get_remote_storage_config ()
+{
+  m_uses_remote_storage = prm_get_bool_value (PRM_ID_REMOTE_STORAGE);
+  return m_uses_remote_storage;
```
ribram:
```suggestion
  return false;
```

src/server/passive_tran_server.cpp
```diff
@@ -0,0 +1,96 @@
...
+  if (log_Gl.m_max_ps_flushed_lsa < saved_lsa)
+    {
+      log_Gl.update_max_ps_flushed_lsa (saved_lsa);
+    }
+
+  if (prm_get_bool_value (PRM_ID_ER_LOG_COMMIT_CONFIRM))
+    {
+      _er_log_debug (ARG_FILE_LINE, "[COMMIT CONFIRM] Received LSA = %lld|%d.\n", LSA_AS_ARGS (&saved_lsa));
+    }
+}
```
ribram: Not needed. The saved_lsa will only be sent to ATS.
```suggestion
```

src/server/passive_tran_server.cpp
```diff
@@ -0,0 +1,96 @@
...
+void
+passive_tran_server::on_boot ()
+{
+  assert (is_passive_transaction_server ());
+
+  cublog::prior_sender::sink_hook sink =
+	  std::bind (&passive_tran_server::push_request, std::ref (*this), tran_to_page_request::SEND_LOG_PRIOR_LIST,
+		     std::placeholders::_1);
+
+  log_Gl.m_prior_sender.add_sink (sink);
```
ribram: Not needed. PTS does not send the log to anyone.
```suggestion
```

src/server/passive_tran_server.cpp
```diff
@@ -0,0 +1,96 @@
...
+passive_tran_server::get_request_handlers ()
+{
+  using map_value_t = request_handlers_map_t::value_type;
+
+  map_value_t boot_info_handler_value =
+	  std::make_pair (page_to_tran_request::SEND_BOOT_INFO,
+			  std::bind (&passive_tran_server::receive_boot_info, std::ref (*this), std::placeholders::_1));
+  map_value_t saved_lsa_handler_value =
+	  std::make_pair (page_to_tran_request::SEND_SAVED_LSA,
+			  std::bind (&passive_tran_server::receive_saved_lsa, std::ref (*this), std::placeholders::_1));
```
ribram: PTS does not receive saved LSA.
```suggestion
```

src/server/passive_tran_server.hpp
ribram: I agree with @cristiarg. Maybe you two can have a call to talk about it, before the next patch.

src/server/passive_tran_server.hpp
```diff
@@ -0,0 +1,42 @@
...
+
+  private:
+    void on_boot () final override;
+    bool get_remote_storage_config () final override;
+
+    request_handlers_map_t get_request_handlers () final override;
+
+  private:
+
+    bool m_uses_remote_storage = false;
```
ribram: Not used.
```suggestion
```

ribram (approved):

src/server/passive_tran_server.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+  map_value_t boot_info_handler_value =
+	  std::make_pair (page_to_tran_request::SEND_BOOT_INFO,
+			  std::bind (&passive_tran_server::receive_boot_info, std::ref (*this), std::placeholders::_1));
+  map_value_t log_page_handler_value =
+	  std::make_pair (page_to_tran_request::SEND_LOG_PAGE,
+			  std::bind (&passive_tran_server::receive_log_page, std::ref (*this), std::placeholders::_1));
+  map_value_t data_page_handler_value =
+	  std::make_pair (page_to_tran_request::SEND_DATA_PAGE,
+			  std::bind (&passive_tran_server::receive_data_page, std::ref (*this), std::placeholders::_1));
```
cristiarg: - `SEND_BOOT_INFO`, `SEND_LOG_PAGE` and `SEND_DATA_PAGE` are common for ATS and PTS.
- They could be moved to an implementation of `tran_server::get_request_handlers` and the handling functions marked `private`.
- The derived implementations - `active/passive_tran_server::get_request_handlers` - would only add the extra handler(s).
- For `passive_tran_server::get_request_handlers` there are no extra handlers so the function would not need to be implemented.
- For `active_tran_server::get_request_handlers`, handler for `SEND_SAVED_LSA` would need to be added.

[ Up to the implementer ]

src/server/passive_tran_server.cpp
```diff
@@ -0,0 +1,56 @@
...
+  return false;
+}
+
+void
+passive_tran_server::on_boot ()
+{
+  assert (is_passive_transaction_server ());
+}
+
+passive_tran_server::request_handlers_map_t
```
cristiarg:
```suggestion
tran_server::request_handlers_map_t
```

src/server/passive_tran_server.cpp
```diff
@@ -0,0 +1,56 @@
...
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+
+#include "passive_tran_server.hpp"
+
+#include "log_impl.h"
+#include "system_parameter.h"
```
cristiarg:
```suggestion
```

src/server/passive_tran_server.cpp
```diff
@@ -0,0 +1,56 @@
...
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+
+#include "passive_tran_server.hpp"
+
+#include "log_impl.h"
```
cristiarg: A finer grained header might suffice.

src/connection/server_support.c
```diff
@@ -2635,6 +2635,15 @@ css_process_server_server_connect (SOCKET master_fd)
       ps_Gl.set_active_tran_server_connection (std::move (chn));
       ps_Gl.start ();
       break;
+    case cubcomm::server_server::CONNECT_PASSIVE_TRAN_TO_PAGE_SERVER:
+      // *INDENT-OFF*
+      if (!chn.send_int (static_cast<int> (cubcomm::server_server::CONNECT_PASSIVE_TRAN_TO_PAGE_SERVER)))
+        {
+          er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_NET_TRANSACTION_SERVER_CONNECTION, 0);
```
cristiarg, line 2642: Do we need to log different messages for ATS and PTS? The messages would be logged on different machines anyway.

src/server/passive_tran_server.cpp
ribram: This could be done. @adrianbrab will you?

src/server/passive_tran_server.cpp
ribram: Actually I think it can just be removed.

src/connection/server_support.c
ribram: I don't think this is a requirement for this patch. May be changed later.

src/server/passive_tran_server.cpp
adrianbrab: I will make the changes on this patch

src/server/tran_server.hpp
```diff
@@ -96,11 +96,6 @@ class tran_server
     virtual bool get_remote_storage_config () = 0;
     virtual void on_boot () = 0;
 
-    // Common request Handlers
-    void receive_boot_info (cubpacking::unpacker &upk);
-    void receive_log_page (cubpacking::unpacker &upk);
-    void receive_data_page (cubpacking::unpacker &upk);
-
     virtual request_handlers_map_t get_request_handlers () = 0;
```
cristiarg: Here we have an example of a function that is marked pure but also has an implementation. :)

cristiarg (approved):

ribram (approved):
Please fix the issue reported by @cristiarg

src/server/tran_server.hpp
adrianbrab: missed this

cristiarg: @adrianbrab
There was no problem with the code. It was valid. I merely pointed out a rarely occurring situation. It's perfectly valid and even needed in situation where it is needed to have an implementation but also a need to mark the class as non-instantiable. It's the trade-off for c++ not having explicit interfaces.

PR #3141, opened 2021-11-01
[LETS-221] Separate page locking from claiming BCB

cristiarg:
http://jira.cubrid.org/browse/LETS-221

Refactor to separate page access locking-or-waiting code from buffer control block code in pgbuf_claim_bcb_for_fix.

Implementation:
- move function pgbuf_lock_page one level up in pgbuf_fix and interpret the result status of the function there, to decide whether to try again to acquire page access lock ownership or continue as already having page access lock ownership
- extract functions pgbuf_buffer_lock_enlist_thread and pgbuf_buffer_lock_delist_thread from pgbuf_lock_page
- extract the perf instrumenting code in pgbuf_claim_bcb_for_fix as a separate function - pgbuf_perf_register_page_lock - at the end of pgbuf_lock_page
- move the check "er_errid () == ER_CSS_PTHREAD_MUTEX_TRYLOCK" one scope up from pgbuf_claim_bcb_for_fix in pgbuf_fix to be nearer to the scope where the error is set - in pgbuf_search_hash_chain
- extracted function pgbuf_initialize_lock_table_entry and replaced where appropriate


src/storage/page_buffer.c
```diff
@@ -7706,49 +7758,22 @@ pgbuf_claim_bcb_for_fix (THREAD_ENTRY * thread_p, const VPID * vpid, PAGE_FETCH_
...
-      pthread_mutex_unlock (&hash_anchor->hash_mutex);
-      PGBUF_BCB_CHECK_MUTEX_LEAKS ();
-      return NULL;
-    }
-
+  // TODO: more accurate comments
   /* In this case, the caller is holding only hash_anchor->hash_mutex. The hash_anchor->hash_mutex is to be
    * released in pgbuf_lock_page (). */
-  if (pgbuf_lock_page (thread_p, hash_anchor, vpid) != PGBUF_LOCK_HOLDER)
+  const PGBUF_LOCK_MODE lock_page_res = pgbuf_lock_page (thread_p, vpid, fetch_mode, hash_anchor, perf);
```
ribram: Move pgbuf_lock_page out of pgbuf_claim_bcb_for_fix. Where pgbuf_claim_bcb_for_fix is called, call pgbuf_lock_page first.

src/storage/page_buffer.c
```diff
@@ -7272,56 +7275,63 @@ pgbuf_delete_from_hash_chain (THREAD_ENTRY * thread_p, PGBUF_BCB * bufptr)
...
   TSC_TICKS start_tick, end_tick;
   UINT64 lock_wait_time = 0;
 
-  /* the caller is holding hash_anchor->hash_mutex */
-  /* check whether the page is in the Buffer Lock Chain */
-
-  if (thread_p == NULL)
+  // TODO: more accurate comments
+  /* The page is not found in the hash chain the caller is holding hash_anchor->hash_mutex */
+  if (er_errid () == ER_CSS_PTHREAD_MUTEX_TRYLOCK)
```
cristiarg: I do not understand what this checks. It seems it expects some error to have been set before , but it does not look right to me that the error being set and checked in different functions.

src/storage/page_buffer.c
```diff
@@ -7272,56 +7275,63 @@ pgbuf_delete_from_hash_chain (THREAD_ENTRY * thread_p, PGBUF_BCB * bufptr)
...
   PGBUF_BUFFER_LOCK *cur_buffer_lock;
-  THREAD_ENTRY *cur_thrd_entry;
   TSC_TICKS start_tick, end_tick;
   UINT64 lock_wait_time = 0;
 
-  /* the caller is holding hash_anchor->hash_mutex */
-  /* check whether the page is in the Buffer Lock Chain */
-
-  if (thread_p == NULL)
+  // TODO: more accurate comments
```
cristiarg: these existing comments are non-intelligible for me

src/storage/page_buffer.c
```diff
@@ -1975,15 +1976,32 @@ pgbuf_fix_release (THREAD_ENTRY * thread_p, const VPID * vpid, PAGE_FETCH_MODE f
     }
   else
     {
-      bufptr = pgbuf_claim_bcb_for_fix (thread_p, vpid, fetch_mode, hash_anchor, &perf, &retry);
+      // TODO: more accurate comments
```
ribram:
```suggestion
```

src/storage/page_buffer.c
```diff
@@ -1975,15 +1976,32 @@ pgbuf_fix_release (THREAD_ENTRY * thread_p, const VPID * vpid, PAGE_FETCH_MODE f
...
-	      retry = false;
-	      goto try_again;
-	    }
 	  ASSERT_ERROR ();
+	  // TODO: more explanatory comments
+	  /*
+	   * Now, caller is not holding any mutex.
+	   * the last argument of pgbuf_unlock_page () is true that
+	   * means hash_mutex must be held before unlocking page.
+	   */
```
ribram:
```suggestion
	  // No mutex is held. The last argument of pgbuf_unlock_page () is true  because hash_mutex must be locked before unlocking page.
```

src/storage/page_buffer.c
ribram:
```suggestion
```

src/storage/page_buffer.c
ribram: I've had a second look in the code and I have a suggestion. This error check should be moved up to pgbuf_fix.

In pgbuf_fix, the code may be something like this:

```c
  bufptr = pgbuf_search_hash_chain ();
  if (bufptr != nullptr)
    {
      if (pgbuf_bcb_is_direct_victim (bufptr))
        ...
      ...
    }
  else
    {
      if (er_errid () != NO_ERROR)
        {
          // Error searching the hash chain
          pthread_mutex_unlock (&hash_anchor->hash_mutex);
          PGBUF_BCB_CHECK_MUTEX_LEAKS ();
          return nullptr;
        }

      if (fetch_mode == OLD_PAGE_IF_IN_BUFFER)
        ...

      const PGBUF_LOCK_MODE lock_page_res = pgbuf_lock_page (thread_p, vpid, fetch_mode, hash_anchor, &perf);
      ...

      bufptr = pgbuf_claim_bcb_for_fix (...);
      ...
    }
```

As a result, `PGBUF_LOCK_UNKNOWN` may become unnecessary.

src/storage/page_buffer.c
```diff
@@ -7345,38 +7372,78 @@ pgbuf_lock_page (THREAD_ENTRY * thread_p, PGBUF_BUFFER_HASH * hash_anchor, const
...
-	  return PGBUF_LOCK_WAITER;
+	  pgbuf_lock_state = PGBUF_LOCK_WAITER;
+	  break;
 	}
       cur_buffer_lock = cur_buffer_lock->lock_next;
     }
 
-  /* buf_lock_table is implemented to have one entry for each thread. At first design, it had one entry for each
-   * thread. cur_thrd_entry->index : thread entry index cur_thrd_entry->tran_index : transaction entry index */
+  if (pgbuf_lock_state != PGBUF_LOCK_WAITER)
```
ribram: Move code to separate function to improve readability. Change the condition a bit, either by specifically setting pgbuf_lock_state to PGBUF_LOCK_HOLDER, or by checking that cur_buffer_lock does not exist.
```suggestion
  if (pgbuf_lock_state == PGBUF_LOCK_HOLDER) // or cur_buffer_lock == nullptr
    {
      cur_buffer_lock = &(pgbuf_Pool.buf_lock_table[thread_p->index]);
      pgbuf_init_buffer_lock (cur_buffer_lock, ...);
    }
```

src/storage/page_buffer.c
```diff
@@ -7345,38 +7372,78 @@ pgbuf_lock_page (THREAD_ENTRY * thread_p, PGBUF_BUFFER_HASH * hash_anchor, const
...
+    {
+      if (fetch_mode == NEW_PAGE)
+	{
+	  perf->perf_page_found = PERF_PAGE_MODE_NEW_NO_WAIT;
+	}
+      else
+	{
+	  perf->perf_page_found = PERF_PAGE_MODE_OLD_NO_WAIT;
+	}
+    }
```
ribram: Move code to separate function for readability.
```suggestion
  if (perf->is_perf_tarcking)
    {
      pgbuf_perf_register_page_lock (...);
    }
```

src/storage/page_buffer.c
```diff
@@ -7272,56 +7290,63 @@ pgbuf_delete_from_hash_chain (THREAD_ENTRY * thread_p, PGBUF_BCB * bufptr)
...
+  /* find vpid in buffer lock chain; entry being the result of indexing via a hash, there might be vpid collisions */
   while (cur_buffer_lock != NULL)
     {
       if (VPID_EQ (&(cur_buffer_lock->vpid), vpid))
 	{
-	  /* found */
-	  cur_thrd_entry->next_wait_thrd = cur_buffer_lock->next_wait_thrd;
-	  cur_buffer_lock->next_wait_thrd = cur_thrd_entry;
-	  pgbuf_sleep (cur_thrd_entry, &hash_anchor->hash_mutex);
+	  /* found page, enlist this thread at the head of the page's buffer lock chain
```
ribram, line 7390: Nitpicking: I would move this part or some of the code in this scope to another function to improve code readability.

src/storage/page_buffer.c
cristiarg: Done. Removed UNKNOWN.

src/storage/page_buffer.c
cristiarg: Extracted two functions: `pgbuf_buffer_lock_enlist_thread` and `pgbuf_buffer_lock_delist_thread`.

src/storage/page_buffer.c

src/storage/page_buffer.c
cristiarg: Moved the initialization code only to a separate function. I think that moving too much of the code to separate scopes will have the effect of affecting context rather than improve it.

src/storage/page_buffer.c
```diff
@@ -1967,37 +1972,59 @@ pgbuf_fix_release (THREAD_ENTRY * thread_p, const VPID * vpid, PAGE_FETCH_MODE f
...
-      if (fetch_mode == NEW_PAGE && pgbuf_hit == false)
+      if (fetch_mode == OLD_PAGE_IF_IN_BUFFER)
 	{
-	  pgbuf_hit = true;
+	  /* we don't need to fix page */
+	  pthread_mutex_unlock (&hash_anchor->hash_mutex);
+	  return NULL;
 	}
-      if (fetch_mode != NEW_PAGE)
+      else
```
ribram: When there is a return in an if scope, else is not necessary. It adds two additional indentations and, if the else scope is quite big, it is difficult to follow the branches.

So sometimes a condition that returns:
```c
  if (cond)
    {
      ...
      return ...;
    }
  else
    {
      // else stuff
      ...
    }
```
it is preferable to rewrite as:
```c
  if (cond)
    {
      ...
      return ...;
    }

  // else stuff
  ...
```

I think this case qualifies for the second option.

src/storage/page_buffer.c
```diff
@@ -1967,37 +1972,59 @@ pgbuf_fix_release (THREAD_ENTRY * thread_p, const VPID * vpid, PAGE_FETCH_MODE f
...
 	{
-	  CUBRID_PGBUF_MISS ();
-	}
+	  /* In this case, the caller is holding only hash_anchor->hash_mutex. The hash_anchor->hash_mutex is to be
+	   * released in pgbuf_lock_page (). */
+	  const PGBUF_LOCK_MODE lock_page_res = pgbuf_lock_page (thread_p, vpid, fetch_mode, hash_anchor, &perf);
+	  if (lock_page_res == PGBUF_LOCK_WAITER)
+	    {
+	      goto try_again;
```
ribram: I think an explanation here would help:
```suggestion
              // Somebody else loaded page from disk. Try to search the hash again.
	      goto try_again;
```

src/storage/page_buffer.c
```diff
@@ -7331,52 +7413,83 @@ pgbuf_lock_page (THREAD_ENTRY * thread_p, PGBUF_BUFFER_HASH * hash_anchor, const
...
-  pthread_mutex_unlock (&hash_anchor->hash_mutex);
-#endif /* SERVER_MODE */
+      /* vpid is not found in the Buffer Lock Chain */
+      cur_buffer_lock = &(pgbuf_Pool.buf_lock_table[thread_p->index]);
+      pgbuf_initialize_lock_table_entry (*cur_buffer_lock, *vpid);
 
+      hash_anchor->lock_next = cur_buffer_lock;
+      (void) pthread_mutex_unlock (&hash_anchor->hash_mutex);
+      perfmon_inc_stat (thread_p, PSTAT_LK_NUM_ACQUIRED_ON_PAGES);	/* monitoring */
+      pgbuf_lock_state = PGBUF_LOCK_HOLDER;
```
ribram: `pgbuf_lock_state` already is `PGBUF_LOCK_HOLDER`
```suggestion
```

src/storage/page_buffer.c
```diff
@@ -7331,52 +7413,83 @@ pgbuf_lock_page (THREAD_ENTRY * thread_p, PGBUF_BUFFER_HASH * hash_anchor, const
...
+    }
+#else
   perfmon_inc_stat (thread_p, PSTAT_LK_NUM_ACQUIRED_ON_PAGES);	/* monitoring */
-  return PGBUF_LOCK_HOLDER;
+  pgbuf_lock_state = PGBUF_LOCK_HOLDER;
```
ribram: `pgbuf_lock_state` already is `PGBUF_LOCK_HOLDER`
```suggestion
```

src/storage/page_buffer.c
```diff
@@ -7269,59 +7305,105 @@ pgbuf_delete_from_hash_chain (THREAD_ENTRY * thread_p, PGBUF_BCB * bufptr)
...
+	    }
+	  thrd_entry->next_wait_thrd = nullptr;
+	  thrd_found = true;
+	  break;
+	}
+      prev_thrd_entry = thrd_entry;
+      thrd_entry = thrd_entry->next_wait_thrd;
+    }
+
+  assert (thrd_found);
```
ribram, line 7353: Not sure this is required to be true. This kind of assertion did not exist in previous code. Please remove it if you're not 100% sure of it.

cristiarg: @ribram 
In the only context the function is currently used in, the asserted post-condition must hold because the thread is enlisted by the function `pgbuf_buffer_lock_enlist_thread` .Unless I understood it wrongly.

ribram: It matters who else delists the thread, and `pgbuf_unlock_page` also does it. But `pgbuf_unlock_page` also sets the resumed status, so at first glance, the assertion should stand. If you're confident, let's keep it.

ribram (approved):

PR #3143, opened 2021-11-02
[CBRD-24078] Fix API link error when using cubrid_log apis on windows

joohok:
http://jira.cubrid.org/browse/CBRD-24078

API link error is caused by not adding cubrid_log APIs to the cubridcs.def
So, list of cubrid log apis is added to the file.


hornetmj:
@joohok How to check the 'Repro'?

There are some warning messages in the cubrid_log.c file. Please remove them. And I think that the files changed by CDC project should be checked together.

hornetmj (approved):

PR #3147, opened 2021-11-02
[LETS-256] Delete database should delete transaction server specific metalog file

cristiarg:
http://jira.cubrid.org/browse/LETS-256

Transaction server (TS) now has its own metalog file - <dbname>_lgmeta_tran.
This file is created when needed.
When the database is deleted, this file, if found alongside others, needs to be deleted as well.

There are two scenarios that a transaction server with remote storage (TSRS) and a page server (PS) can be executed:
- on separate containers, in which case, the TS specific metalog would be located on the TS container:
- on a single node (container): in which case, all regular database files including the generic metalog file and the TS specific metalog file, would be located in the same directory.

deletedb utility must clean the TS specific metalog file in both scenarios.

### DIFF
--- a/src/transaction/log_page_buffer.c
+++ b/src/transaction/log_page_buffer.c
@@ -393,6 +393,8 @@ static bool logpb_is_log_active_from_backup_useful (THREAD_ENTRY * thread_p, con
 						    const char *db_full_name);
 static int logpb_peek_header_of_active_log_from_backup (THREAD_ENTRY * thread_p, const char *active_log_path,
 							LOG_HEADER * hdr);
+static void logpb_delete_metainfo_files_internal (THREAD_ENTRY * thread_p, const char *logpath,
+						  const char *prefix_logname);
 #if defined (SERVER_MODE)
 static void logpb_send_flushed_lsa_to_ats ();
 #endif // SERVER_MODE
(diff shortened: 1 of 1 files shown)


ribram (approved):

src/transaction/log_page_buffer.c
```diff
@@ -11403,11 +11405,34 @@ logpb_remove_all_in_log_path (THREAD_ENTRY * thread_p, const char *db_fullname,
...
+ * NOTE: a transaction server with remote storage has its own specific metalog file;
+ *      in 'single-node' install mode - where both transaction server and page server
+ *      execute on the same machine, both the regular metalog volume and the transaction server specific
+ *      metalog volume are located in the same directory
+ */
+void
+logpb_delete_metainfo_files_internal (THREAD_ENTRY * thread_p, const char *logpath, const char *prefix_logname)
+{
+  char metainfo_file_name[PATH_MAX];
+  const bool is_ts_with_remote_storage = is_tran_server_with_remote_storage ();
```
hornetmj, line 11425: Regardless of the value of the is_ts_with_remote_storage variable, this function always seems to call fileio_make_log_metainfo_name(true) and fileio_make_log_metainfo_name(false) once. Is there any reason to use this variable?

cristiarg: @hornetmj 
You are, somewhat, right.

The only difference is in relation with the order in which deletes happen.
If the database files are deleted on a transaction server:
- first delete FILEIO_SUFFIX_LOGMETA_TRAN   "_lgmeta_tran"
- then delete FILEIO_SUFFIX_LOGMETA        "_lgmeta"

If on a page server, the other way around.
The comments also specifies that the "own" metalog file is first deleted, then the other.

If there was a way to mandate to the `fileio_unformat` - `fileio_unformat_and_rename` that the files are mandatory to exist or not, it would make even more sense.

As a side comment, there are two log meta files to be able to execute the "single_node" configuration where both servers have the files in the same location and use the same config file(s).

PR #3148, opened 2021-11-03
[CBRD-24087]  Improvement of oid_compare() function and comparison logic of codes using this function.

ctshim:
http://jira.cubrid.org/browse/CBRD-24087

Purpose
The oid_compare() function is called when comparing OIDs in various places in the source code.
In particular, it is designated as a comparison function such as sort (qsort) and search (lsearch).
Since very frequent calls occur, we want to increase the efficiency of the code.

Implementation
  Remove unnecessary substitution statements

Remarks
N/A

### DIFF
--- a/src/storage/oid.h
+++ b/src/storage/oid.h
@@ -94,41 +94,29 @@
 
 #define oid_is_equal_value(oid1, oid2)  OID_EQ((oid1), (oid2))
 #define OID_EQ(oidp1, oidp2) \
-  ((oidp1) == (oidp2) || ((oidp1)->pageid == (oidp2)->pageid && \
-			  (oidp1)->slotid == (oidp2)->slotid && \
-			  (oidp1)->volid  == (oidp2)->volid))
+  ( (oidp1)->pageid == (oidp2)->pageid && \
+    (oidp1)->slotid == (oidp2)->slotid && \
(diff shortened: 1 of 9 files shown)


src/storage/oid.h
```diff
@@ -92,6 +92,7 @@
     (oidp)->volid = NULL_VOLID; \
   } while(0)
 
+#define oid_is_equal_value(oid1, oid2)  OID_EQ((oid1), (oid2))
```
hgryoo: Do we need this macro?

ctshim: I can't say it's absolutely necessary.
However, I tried to make the macro name more explicit while defining a macro that replaces the existing code of the form oid_cpmpare() == 0.

hgryoo: If it only calls OID_EQ, this macro is redundant. Also, the name of C macro function should be capital by naming convention.

ctshim: @hgryoo OK, I will fix it.

beyondykk9 (approved):

hyunikn (approved):

youngjinj (approved):

shparkcubrid (approved):

hgryoo (approved):

PR #3149, opened 2021-11-03
[CBRD-23842] Check the range of log lsa to extract when validate the log lsa

joohok:

LOG_LSA should be greater than the smallest lsa in the log volume (active or archive) and smaller than the latest lsa in the active log volume. 

If requested lsa is not in the range of LSA, then returns error (ER_CDC_INVALID_LOG_LSA) to the client.

### DIFF
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -13681,7 +13681,7 @@ cdc_check_lsa_range (THREAD_ENTRY * thread_p, LOG_LSA * lsa)
   LOG_CS_EXIT (thread_p);
 
   cdc_log ("%s : first log lsa from log volume is (%lld|%d) and last lsa is (%lld|%d). input lsa is (%lld|%d)",
-	   LSA_AS_ARGS (&first_lsa), LSA_AS_ARGS (&nxio_lsa), LSA_AS_ARGS (lsa));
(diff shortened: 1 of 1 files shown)


src/transaction/log_manager.c
```diff
@@ -13594,6 +13594,105 @@ cdc_find_lsa (THREAD_ENTRY * thread_p, time_t * time, LOG_LSA * start_lsa)
...
+	  first_lsa.offset = 0;
+	}
+    }
+
+  LOG_CS_EXIT (thread_p);
+
+  cdc_log ("%s : first log lsa from log volume is (%lld|%d) and last lsa is (%lld|%d). input lsa is (%lld|%d)",
+	   __func__, LSA_AS_ARGS (&first_lsa), LSA_AS_ARGS (&nxio_lsa), LSA_AS_ARGS (lsa));
+
+  if (LSA_GE (lsa, &first_lsa) && LSA_LE (lsa, &nxio_lsa))
```
```suggestion
  if (LSA_GE (lsa, &first_lsa) && LSA_LT (lsa, &nxio_lsa))
```

joohok: Thanks, I was confused nxio_lsa was the latest lsa that written to the log volume. I will fix it

hornetmj (approved):

PR #3151, opened 2021-11-04
[CBRD-24091] Print error messages for TDE depending on a system parameter

wrlawodms:
http://jira.cubrid.org/browse/CBRD-24091

I have removed most of `#if !defined(NDEBUG)` pairs because It cost little. However, I have left some of them if it costs more than just checking the system parameter in Release mode.


src/storage/tde.h
```diff
@@ -136,6 +136,8 @@ typedef struct tde_mk_file_item
    || (rcvindex) == RVBT_ONLINE_INDEX_UNDO_TRAN_INSERT \
    || (rcvindex) == RVBT_ONLINE_INDEX_UNDO_TRAN_DELETE)
 
+#define tde_er_log(...) if (prm_get_bool_value (PRM_ID_ER_LOG_TDE)) _er_log_debug (ARG_FILE_LINE, __VA_ARGS__)
```
joohok:
```suggestion
#define tde_er_log(...) if (prm_get_bool_value (PRM_ID_ER_LOG_TDE)) _er_log_debug (ARG_FILE_LINE, "TDE : " __VA_ARGS__)
```

How about adding "TDE : " to tde_er_log () like logpb_log () did. It seems "TDE : " prefix is used whenever er_tde_log () is called.

wrlawodms: It looks better. I'll follow it.

joohok (approved):

hornetmj (approved):

eido5 (approved):

PR #3153, opened 2021-11-04
[LETS-276] Allow ATS reconnection after a crash

ribram:
http://jira.cubrid.org/browse/LETS-276

When ATS connects to the PS, there is an assert on the PS side that no prior ATS connection exists. During a normal shutdown, the ATS will disconnect from the PS, but after a crash the connection remains. When the ATS reconnects, the PS crashes.

Fix by accepting the fact that a connection with the ATS may already exist when a new connection is created. Delete the prior connection.

### DIFF
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -52,7 +52,7 @@ void page_server::set_active_tran_server_connection (cubcomm::channel &&chn)
   er_log_debug (ARG_FILE_LINE, "Active transaction server connected to this page server. Channel id: %s.\n",
 		chn.get_channel_id ().c_str ());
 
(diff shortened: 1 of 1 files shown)


src/server/page_server.cpp
```diff
@@ -52,7 +52,13 @@ void page_server::set_active_tran_server_connection (cubcomm::channel &&chn)
   er_log_debug (ARG_FILE_LINE, "Active transaction server connected to this page server. Channel id: %s.\n",
 		chn.get_channel_id ().c_str ());
 
+  if (m_active_tran_server_conn)
```
```suggestion
  if (m_active_tran_server_conn != nullptr)
```

cristiarg (approved):

PR #3155, opened 2021-11-05
[CBRD-24107] Avoid executing multiple statements in DBLINK

ctshim:
http://jira.cubrid.org/browse/CBRD-24107

Purpose
Query statements written inside the DBLINK statement are sent to the remote DBMS without a separate parsing process, so multiple queries are prevented from being delivered at once.

Implementation
     Forces not to have multiple statements.
     When multiple statements come,  it is a policy that forces only the first statement to be executed without error handling.

Remarks
    For CUBRID, multiple syntaxes can be listed without a semicolon.
    Defending this issue is beyond the scope of this issue.

### DIFF
--- a/src/parser/csql_grammar.y
+++ b/src/parser/csql_grammar.y
@@ -28434,6 +28434,67 @@ pt_fill_conn_info_container(PARSER_CONTEXT *parser,  int buffer_pos, container_1
     ctn->c10 = FROM_NUMBER(set_bits);
 }
 
+static bool
+pt_check_one_stmt(char* p)
+{
+    char end;
+    char* t = strchr(p, ';');
+    if(!t)
+     {
+        return true;
+     }
+
+    t++;
+    while (*t == ' ' || *t == '\t' || *t == '\r' || *t == '\n')
+     {
+        t++;
+     }
+    if(*t == '\0')
+       return true;
+
+    while(*p)
+    {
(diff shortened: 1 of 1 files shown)


shparkcubrid (approved):

src/parser/csql_grammar.y
```diff
@@ -28434,6 +28434,67 @@ pt_fill_conn_info_container(PARSER_CONTEXT *parser,  int buffer_pos, container_1
...
+            p++;
+          }
+    }
+
+    while (*p == ' ' || *p == '\t' || *p == '\r' || *p == '\n')
+     {
+        p++;
+     }
+
+    return (*p == '\0');
```
hyunikn, line 28495: If there is another statement after the second semicolon, this return statement returns false, which is not in accordance with the above notice.

ctshim: The second semicolon does not meet.
Because I forcefully inserted '\0' after the first semicolon.

hyunikn: Yes, you are right. I misunderstood it.

hyunikn (approved):

beyondykk9 (approved):

youngjinj (approved):

hgryoo (approved):

PR #3156, opened 2021-11-05
[LETS-222] New page buffer fetch mode for PTS replication

cristiarg:
http://jira.cubrid.org/browse/LETS-222

A new fetch mode is required for passive transaction server (PTS) replication. A page must be fixed if one of the next two conditions are met:
- the page is in page buffer already
- the page was requested from page server (PS) and is in the process of being added to the page buffer.

After LETS-221 refactoring, add OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT fetch mode that applies to the above conditions.

Implementation
- add OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT fetch mode
- add new enum value PGBUF_LOCK_NOT_NEEDED
- in function pgbuf_lock_page, if fetch mode is OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT and no page access lock entry is found for the page VPID (ie: page is not in the process of being retrieved from the persistent storage - be that disk or PS), return PGBUF_LOCK_NOT_NEEDED
- in pgbuf_fix, if pgbuf_lock_page returns PGBUF_LOCK_NOT_NEEDED, do not fix the page and return null

### DIFF
--- a/src/storage/page_buffer.c
+++ b/src/storage/page_buffer.c
@@ -1998,6 +1998,7 @@ try_again:
       if (lock_page_res == PGBUF_LOCK_NOT_NEEDED)
 	{
 	  // No need to load the page to page buffer
+	  assert (fetch_mode == OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT);
 	  return nullptr;
 	}
 
--- a/src/storage/page_buffer.c
+++ b/src/storage/page_buffer.c
@@ -352,6 +352,7 @@ enum PGBUF_LOCK_MODE
 {
   PGBUF_LOCK_WAITER = 0,
   PGBUF_LOCK_HOLDER,
+  PGBUF_LOCK_NOT_NEEDED,
 };
 
 /* constants to indicate the content state of buffers */
@@ -1994,6 +1995,11 @@ try_again:
       /* In this case, the caller is holding only hash_anchor->hash_mutex. The hash_anchor->hash_mutex is to be
        * released in pgbuf_lock_page (). */
       const PGBUF_LOCK_MODE lock_page_res = pgbuf_lock_page (thread_p, vpid, fetch_mode, hash_anchor, &perf);
+      if (lock_page_res == PGBUF_LOCK_NOT_NEEDED)
+	{
+	  return nullptr;
+	}
+
       if (lock_page_res == PGBUF_LOCK_WAITER)
 	{
 	  // page was loaded by some other thread; search hash chain again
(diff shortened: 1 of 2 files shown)


src/storage/page_buffer.c
```diff
@@ -7425,6 +7431,12 @@ pgbuf_lock_page (THREAD_ENTRY * const thread_p, const VPID * const vpid, PAGE_FE
     }
   assert ((pgbuf_lock_state == PGBUF_LOCK_WAITER) != (cur_buffer_lock == nullptr));	// xor
 
+  if (fetch_mode == OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT)
+    {
+      return PGBUF_LOCK_NOT_NEEDED;
+    }
```
ribram: If `pgbuf_lock_state == PGBUF_LOCK_WAITER`, it means the page was loaded to page buffer and must be fixed.

If at this point `pgbuf_lock_state == PGBUF_LOCK_HOLDER` and `fetch_mode == OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT`, then page does not have to be loaded to page buffer and lock may be changed to `PGBUF_LOCK_NOT_NEEDED`.

src/storage/page_buffer.c
```diff
@@ -1994,6 +1995,11 @@ pgbuf_fix_release (THREAD_ENTRY * thread_p, const VPID * vpid, PAGE_FETCH_MODE f
       /* In this case, the caller is holding only hash_anchor->hash_mutex. The hash_anchor->hash_mutex is to be
        * released in pgbuf_lock_page (). */
       const PGBUF_LOCK_MODE lock_page_res = pgbuf_lock_page (thread_p, vpid, fetch_mode, hash_anchor, &perf);
+      if (lock_page_res == PGBUF_LOCK_NOT_NEEDED)
+	{
+	  return nullptr;
```
ribram, line 2002:
```suggestion
          // No need to load the page to page buffer.
          assert (fetch_mode == OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT);
	  return nullptr;
```

src/storage/page_buffer.c
cristiarg: Done. My bad.

src/storage/page_buffer.c
cristiarg: Done

ribram (approved):

PR #3157, opened 2021-11-08
[CBRD-24100] Problem that the view's order by is copied to the main query needlessly when view is merging.

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-24100

If an aggregate function exists, the order_by of the view is modified so that it is not copied to the main query.
```
as-is : select count(*), b.col2, b.col1 <== needless hidden columns for order_by
             from ....
to-be : select count(*)
            from ...
```


ctshim (approved):

hyunikn (approved):

src/parser/view_transform.c
```diff
@@ -1912,7 +1912,7 @@ mq_substitute_subquery_in_statement (PARSER_CONTEXT * parser, PT_NODE * statemen
 				    tmp_result->info.query.q.select.use_merge);
 
 	      assert (query_spec->info.query.orderby_for == NULL);
-	      if (!order_by && query_spec->info.query.order_by)
+	      if (!order_by && query_spec->info.query.order_by && !pt_has_aggregate (parser, tmp_result))
```
youngjinj, line 1915: Even if there is no aggregate function, the order by in the view is not necessary in the main query, right?

Is there a case where order by in the view is needed in the main query?

shparkcubrid: for backwards compatibility.
And in Oracle as well, it was tested to keep the order_by of the view.

beyondykk9 (approved):

PR #3158, opened 2021-11-08
[CBRD-24098] Improved search of memory hash table using hash value

ctshim:
http://jira.cubrid.org/browse/CBRD-24098

We want to reduce the cost of comparing key values in the use of mht_xxx() family of functions.
For this, first-order filtering is performed by comparing integers using hash values. Conventional hashkey comparison is only necessary when these values are equal.

This way the cost is reduced even when a rehash occurs.


src/cci/cci_common.c
```diff
@@ -80,6 +67,12 @@
 static const float CCI_MHT_REHASH_TRESHOLD = 0.7f;
 static const float CCI_MHT_REHASH_FACTOR = 1.3f;
 
+#if defined(USE_HASHVAL_ORDERED_LIST)
```
hgryoo, line 70: these two macros seem to disable code blocks. what is the difference?
```
#if defined(USE_HASHVAL_ORDERED_LIST)
#if defined(ENABLE_MEANINGLESS_CODE)
```

ctshim: ENABLE_MEANINGLESS_CODE
  >  Irrespective of this fix, remove meaningless code because it is not used in existing code.

USE_HASHVAL_ORDERED_LIST
   >  Allows to maintain sorted order in the newly added hash value comparison processing (not currently defined)

USE_HLS_ORDERED_LIST
   > Same purpose as USE_HASHVAL_ORDERED_LIST, except that the target is mht_hls.

In short, ENABLE_MEANINGLESS_CODE is for disabling unused parts of the existing code,
The other two are for future use through further review.

hyunikn: What about erasing code enclosed by '#if defined(ENABLE_MEANINGLESS_CODE)?

ctshim: @ hyunikn  
You may want to put the actual deletion on hold for now, and then rearrange it later.

hyunikn (approved):

beyondykk9 (approved):

src/base/memory_hash.h
```diff
@@ -76,15 +77,15 @@ struct mht_table
...
 extern unsigned int mht_3strhash (const void *key, const unsigned int ht_size);
-extern unsigned int mht_4strhash (const void *key, const unsigned int ht_size);
-extern unsigned int mht_5strhash (const void *key, const unsigned int ht_size);
+extern unsigned int mht_4strhash (const void *key, const unsigned int ht_size, unsigned int *orig_hash_value);
+extern unsigned int mht_5strhash (const void *key, const unsigned int ht_size, unsigned int *orig_hash_value);
 extern unsigned int mht_numhash (const void *key, const unsigned int ht_size);
 
 extern unsigned int mht_get_hash_number (const int unsigned ht_size, const DB_VALUE * val);
-extern unsigned int mht_ptrhash (const void *ptr, const unsigned int ht_size);
+extern unsigned int mht_ptrhash (const void *ptr, const unsigned int ht_size, unsigned int *orig_hash_value);
```
shparkcubrid, line 88: If the value type is an int, wouldn't the comparison of orig_hash_value be inefficient?

ctshim: It is inevitable to fit the common function prototype.
In comparison, one comparison is increased only when an actual match is made.

src/base/memory_hash.c
```diff
@@ -1420,26 +1493,33 @@ mht_get (MHT_TABLE * ht, const void *key)
   assert (ht != NULL);
   assert (key != NULL);
 
-  /*
-   * Hash the key and make sure that the return value is between 0 and size
-   * of hash table
+  unsigned int orig_hash_value = 0;
+  hash = (*ht->hash_func) (key, ht->size, &orig_hash_value);
```
youngjinj, line 1497: The comment position should be the same as before, before calling the function.

youngjinj (approved):

shparkcubrid:
There should be some test results stating that there is a performance improvement with this change. Has there been no performance improvement in HASH GROUP BY?

ctshim:
It has been left unattended for a long time and is dropped without additional work.

PR #3159, opened 2021-11-08
[CBRD-24109] No error produced for recursive triggering in SA mode.

beyondykk9:
http://jira.cubrid.org/browse/CBRD-24109

Configuration:
The system parameter **create_table_reuseoid** should be set "true"

```
csql -S -u dba testdb

csql>
create class oo(x int);
create trigger o_in before delete on oo execute delete from oo where x = 501;
insert into oo values (1);
insert into oo values(501);
delete from oo where x=1;

ERROR: Error evaluating action for "o_in", Maximum trigger depth 32 exceeded at trigger "o_in".

drop table oo;
create class oo(x int);
create trigger o_in before delete on oo execute delete from oo where x = 501;
insert into oo values (1);
insert into oo values(501);
delete from oo where x=1;
```

### Expected Result:
ERROR: Error evaluating action for "o_in", Maximum trigger depth 32 exceeded at trigger "o_in".

### Actual Result:
0 row affected. (0.000469 sec) Committed.


hgryoo (approved):

ctshim (approved):

shparkcubrid:
ws_abort_mops() is also called in tran_abort(). I think it is needed to check. 
And I'd like to know why it only happens with reuse_oid.

beyondykk9:
> ws_abort_mops() is also called in tran_abort(). I think it is needed to check. And I'd like to know why it only happens with reuse_oid.

ws_abort_mops() is called from tran_abort() but in SA_MODE is not. It's called only at CS_MODE, so the deleted object is not cleared from memory.
I'm not sure why it only happens in case of reuse_oid option. I could not check all of the storage manager part but at storage manager level I found that it handles specially reuse_oid on memory buffer. It might be the reason why the storage manager re-assign the reused oid more easily.

hyunikn (approved):

youngjinj (approved):

shparkcubrid (approved):

PR #3162, opened 2021-11-08
[LETS-215] Separate transaction server handling on active/passive branches

adrianbrab:
http://jira.cubrid.org/browse/LETS-215

Transformed the global variable holding the ATS  into a TS pointer that can be initialized as either a ATS or PTS.

### DIFF
--- a/src/base/server_type.cpp
+++ b/src/base/server_type.cpp
@@ -128,13 +128,13 @@ int init_server_type (const char *db_name)
     {
       assert (ts_Gl == nullptr);
 
-      if (is_active_transaction_server())
+      if (is_active_transaction_server ())
 	{
-	  ts_Gl.reset (new active_tran_server());
+	  ts_Gl.reset (new active_tran_server ());
 	}
-      else if (is_passive_transaction_server())
+      else if (is_passive_transaction_server ())
 	{
-	  ts_Gl.reset (new passive_tran_server());
+	  ts_Gl.reset (new passive_tran_server ());
 	}
       else
 	{
@@ -261,6 +261,31 @@ void finalize_server_type ()
 bool is_tran_server_with_remote_storage ()
 {
   return false;
+}
+
+bool is_active_transaction_server ()
+{
+  return true;
+}
+
+bool is_page_server ()
+{
+  return false;
+}
+
+bool is_passive_transaction_server ()
+{
+  return false;
+}
+
+bool is_passive_server ()
+{
+  return false;
+}
+
+bool is_transaction_server ()
+{
+  return true;
 }
 
 #endif // !SERVER_MODE = SA_MODE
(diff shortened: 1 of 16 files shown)


src/storage/slotted_page.c
```diff
@@ -5276,7 +5276,5 @@ spage_need_compact (THREAD_ENTRY * thread_p, PAGE_PTR page_p)
 bool
 spage_is_save_space_disabled ()
 {
-  const bool during_crash_recovery = log_is_in_crash_recovery ();
-  const bool in_page_server (get_server_type () == SERVER_TYPE_PAGE);
-  return during_crash_recovery || in_page_server;
+  return log_is_in_crash_recovery () || is_passive_server ();
```
cristiarg: The extra variables were added to ease [conditional] debugging.
They might look ugly but the generated code is the same - ie: they'll be optimized away by the compiler in optimized build configs.

cristiarg (approved):

ribram:
Fix build issues for SA_MODE:
```
../sa/libcubridsa.so.12.0: undefined reference to `is_active_transaction_server()'
../sa/libcubridsa.so.12.0: undefined reference to `is_passive_server()'
```
Define the functions in SA_MODE; stand-alone mode is considered an active transaction server.

src/base/server_type.cpp
```diff
@@ -125,7 +126,21 @@ int init_server_type (const char *db_name)
...
+	}
+      else if (is_passive_transaction_server())
+	{
+	  ts_Gl.reset (new passive_tran_server());
+	}
+      else
+	{
+	  assert (false);
+	}
+      er_code = ts_Gl->boot (db_name);
```
ribram: Space after function name.
```suggestion
      if (is_active_transaction_server ())
	{
	  ts_Gl.reset (new active_tran_server ());
	}
      else if (is_passive_transaction_server ())
	{
	  ts_Gl.reset (new passive_tran_server ());
	}
      else
	{
	  assert (false);
	}
      er_code = ts_Gl->boot (db_name);
```

src/storage/slotted_page.c
adrianbrab: i changed them back to variables

ribram (approved):

src/transaction/boot_sr.c
```diff
@@ -2659,7 +2659,7 @@ boot_restart_server (THREAD_ENTRY * thread_p, bool print_restart, const char *db
 #endif
 
 #if defined (SERVER_MODE)
-  if (get_server_type () == SERVER_TYPE_TRANSACTION)
+  if (is_active_transaction_server ())
```
hornetmj, line 2662: Compared to the develop branch, PTS also seems to have to call css_set_ha_num_of_hosts() and css_change_ha_server_state().

PR #3164, opened 2021-11-09
[CBRD-24107] Check for semicolons in comments or hint.

ctshim:
http://jira.cubrid.org/browse/CBRD-24107

Compensation for missing handling of semicolons in comments or hint syntax.


beyondykk9 (approved):

src/parser/csql_grammar.y
```diff
@@ -28466,6 +28466,28 @@ pt_check_one_stmt(char* p)
                     }
                 }
           }
+        else if((p[0] == '-' && p[1] == '-') || (p[0] == '/' && p[1] == '/'))
+        {
+                for(p += 2; *p ; p++)
+                {
+                    if (*p == '\n')
```
youngjinj, line 28473: Instead of using a for or if statement,
```
p = strchr(p, '\n');
```
How about using ?

src/parser/csql_grammar.y
```diff
@@ -28466,6 +28466,28 @@ pt_check_one_stmt(char* p)
...
+                       p++;
+                       break;
+                    }
+                }
+        }
+        else if(p[0] == '/' && p[1] == '*')
+        {
+                for(p += 2; *p ; p++)
+                {
+                    if (p[0] == '*' && p[1] == '/')
```
youngjinj, line 28484: Instead of using a for or if statement,
```
p = strstr(p, "*/");
```
How about using ?

src/parser/csql_grammar.y
ctshim: I'd like to keep the current code in case something like this happens:

```
SELECT * FROM dblink('192.168.1.8:55300:demodb:dba::',
'select 1, 'a' -- test ') as rt(col1 int);
```

src/parser/csql_grammar.y
ctshim: That's a good comment.
But in the same vein as the comment above, I'll keep it current.

src/parser/csql_grammar.y
```diff
@@ -28466,6 +28466,28 @@ pt_check_one_stmt(char* p)
...
+                    {
+                       p += 2;
+                       break;
+                    }
+                }
+        }
         else if(*p == ';')
           {
              p++;
```
youngjinj, line 28493: One of the following cases occurs when exiting the while statement.
1. If we change the first ';' to '\0'
2. When all characters are read because ';' is not encountered

If so, the next string must be ignored or is at the end of the string.

If so, wouldn't the processing below be unnecessary?
( Code below not included in this PR. )
```
while (*p == ' ' || *p == '\t' || *p == '\r' || *p == '\n')
  {
    p++;
  }
```

And is there some cases where this function returns false?

youngjinj (approved):

PR #3166, opened 2021-11-09
[CBRD-24110] Enhance performance for count(*) where no predicates

beyondykk9:
http://jira.cubrid.org/browse/CBRD-24110

I found a disabled code for enhancing the performance of count(*) query. So, I'm testing the code.


src/storage/btree.c
```diff
@@ -4581,7 +4581,7 @@ btree_dump_root_header (THREAD_ENTRY * thread_p, FILE * fp, PAGE_PTR page_ptr)
 
   fprintf (fp, "==============    R O O T    P A G E   ================\n\n");
   fprintf (fp, " Key_Type: %s\n", pr_type_name (TP_DOMAIN_TYPE (key_type)));
-  fprintf (fp, " Num OIDs: %d, Num NULLs: %d, Num keys: %d\n", root_header->num_oids, root_header->num_nulls,
+  fprintf (fp, " Num OIDs: %lld, Num NULLs: %lld, Num keys: %lld\n", root_header->num_oids, root_header->num_nulls,
```
joohok, line 4584: it seems that 'over' is required to be checked when it shows the number of oids, nulls, and keys.

src/storage/btree_load.h
```diff
@@ -210,7 +210,13 @@ struct btree_root_header
...
   OID topclass_oid;		/* topclass oid or NULL OID(non unique index) */
   int unique_pk;		/* unique or non-unique, is primary key */
-  int reverse_reserved;		/* reverse or normal *//* not used */
+  struct
+  {
+    int over:2;			/* for checking to over 32 bit */
+    int num_oids:10;		/* extend 64 bit for num_x each 10 bits */
+    int num_nulls:10;		/* extend 64 bit for num_x each 10 bits */
+    int num_keys:10;		/* extend 64 bit for num_x each 10 bits */
+  } _64;
```
wrlawodms, line 219: I think the name of this structure should have a more meaningful name. In addition. This is a temporary solution so that it would be better to leave a comment about it. With the comment, we can revise it later when we change the disk image.

src/storage/btree.c
```diff
@@ -16922,7 +16943,7 @@ int
 btree_rv_update_tran_stats (THREAD_ENTRY * thread_p, LOG_RCV * recv)
 {
   char *datap;
-  int num_nulls, num_oids, num_keys;
+  long long num_nulls, num_oids, num_keys;
```
joohok, line 16946: How about using INT64 instead of long long? 
Because it uses INT64 in OR_GET_BIGINT, and INT64 can be vary by system.

src/storage/btree.c
wrlawodms, line 4584: The type of ` root_head->num_*` are still `int32`. It seems that it needs a conversion with `_64`. Or you can change the way this function dumps.

src/storage/btree.c
```diff
@@ -16977,15 +16999,42 @@ btree_rv_roothdr_undo_update (THREAD_ENTRY * thread_p, LOG_RCV * recv)
...
+  num_oids = (unsigned int) root_header->num_oids;
+  num_keys = (unsigned int) root_header->num_keys;
+
+  if (root_header->_64.over)
+    {
+      num_nulls |= (long long) root_header->num_nulls << 32;
+      num_oids |= (long long) root_header->num_oids << 32;
+      num_keys |= (long long) root_header->num_keys << 32;
+    }
+
```
wrlawodms: `root_header` is used here, but `root_header != NULL` is checked below.

src/storage/btree.c
```diff
@@ -5639,7 +5639,11 @@ xbtree_add_index (THREAD_ENTRY * thread_p, BTID * btid, TP_DOMAIN * key_type, OI
   VFID_SET_NULL (&(root_header->ovfid));
   root_header->rev_level = BTREE_CURRENT_REV_LEVEL;
 
-  root_header->reverse_reserved = 0;	/* unused */
+  over += root_header->_64.num_oids = num_oids >> 32;
+  over += root_header->_64.num_nulls = num_nulls >> 32;
+  over += root_header->_64.num_keys = num_keys >> 32;
+
+  root_header->_64.over = (over > 0);
```
wrlawodms: I think that these should be initialized together with `root_header->num_*` above in the `if (unique_pk)` condition.

src/storage/btree.c
```diff
@@ -20882,13 +20931,24 @@ btree_scan_for_show_index_header (THREAD_ENTRY * thread_p, DB_VALUE ** out_value
     }
   else
     {
-      db_make_int (out_values[idx], root_header->num_oids);
+      num_oids = root_header->num_oids;
```
wrlawodms, line 20934: It could make a wrong result if `root_header->num_*` is bigger than 2^31-1.
See the next code:
```
  7   int a = 4294967293; // -3
  8   unsigned int b = 4294967293;
  9   long long aa = a;
 10   long long bb = b;
 11
 12   cout << "(a) " << a << " -> " << aa << endl;
 13   cout << "(b) " << b << " -> " << bb << endl;
```
Thre result is:
```
(a) -3 -> -3
(b) 4294967293 -> 4294967293
```
If it is correct that you expect it to work like (b). How about using uint32 for `root_header->num_*`.

src/storage/storage_common.h
```diff
@@ -132,6 +132,14 @@ typedef INT32 LOLENGTH;		/* Length for a large object */
 #define LEAF_RECORD_SIZE (0)
 #define SPLIT_INFO_SIZE (OR_FLOAT_SIZE + OR_INT_SIZE)
 
+/* for backward compatibility */
+typedef struct btree_version BTREE_VERSION;
```
wrlawodms: It seems that this structure is never used.

src/transaction/locator_sr.c
```diff
@@ -13879,7 +13879,35 @@ locator_multi_insert_force (THREAD_ENTRY * thread_p, HFID * hfid, OID * class_oi
...
+	  logtb_update_global_unique_stats_by_delta (thread_p, &classrepr->indexes[idx].btid,
+		recdes_array.size (), 0, recdes_array.size (), true);	
+	  stats = logtb_get_global_unique_stats_entry (thread_p, &classrepr->indexes[idx].btid, true);
+	  if (stats != NULL)
+	    {
+	      error_code = btree_reflect_global_unique_statistics (thread_p, stats, false);
+	      pthread_mutex_unlock (&stats->mutex);
+	    }
+	}
+    }
```
wrlawodms: From what I've heard and read, (1) all stats are kept on a memory cache, (2) the global cache is updated when a transaction commits and a log is left for this, (3) the global stats are reflected to root headers of btrees during checkpoint.

Let me ask some questions:
1. After the bulk update here, the transaction will commit. Why we should update global stats manually here? Won't the transaction including this operation commit?
2. If we leave a log during updating global stats, it seems that we don't need to reflect the state to btree root headers. Why do we have to reflect it here manually? From a jira issue, they remove reflecting during commit becauce of performance, but it is reflected for every BU here.
3. What happenes if a transaction aborts after reflecting by an internal error or something? btree_reflect_global_unique_statistics() does not leave log for UNDO.

src/transaction/mvcc_table.cpp
```diff
@@ -298,14 +298,18 @@ mvcctable::build_mvcc_info (log_tdes &tdes)
       trans_status.m_active_mvccs.copy_to (tdes.mvccinfo.snapshot.m_active_mvccs,
 					   mvcc_active_tran::copy_safety::THREAD_UNSAFE);
       /* load statistics temporary disabled need to be enabled when activate count optimization */
-#if 0
       /* load global statistics. This must take place here and nowhere else. */
-      if (logtb_load_global_statistics_to_tran (thread_p) != NO_ERROR)
+      tdes.mvccinfo.snapshot.valid = true;
```
wrlawodms: This flag is supposed to be set below after initializing all variables of `tdes.mvccinfo.snapshot`. I don't know in details, but I concern that there could be a race condition in which someone access uninitialized variables after checking this flag. Could you tell me the specific code building mvcc info recursively in `logtb_load_global_statistics_to_tran()`?

src/transaction/mvcc_table.cpp
```diff
@@ -298,14 +298,18 @@ mvcctable::build_mvcc_info (log_tdes &tdes)
...
+      /*
+       snapshot should be set to valid to avoid the endless mvcc_info build
+       because gtb_load_global_statistics_to_tran calls build_mvcc_info recursively
+       since the valid flag is false.
+      */
+      if (logtb_load_global_statistics_to_tran (thread_get_thread_entry_info())!= NO_ERROR)
 	{
 	  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_MVCC_CANT_GET_SNAPSHOT, 0);
-	  error_code = ER_MVCC_CANT_GET_SNAPSHOT;
+	  return;
```
wrlawodms: Is it ok to return here? Callers don't check the error code of `build_mvcc_info()` and assums the mvcc info is built. It seems that it will end up unitialized `tdes.mvccinfo`.

wrlawodms (changes requested):

src/storage/btree.c
```diff
@@ -6222,9 +6228,16 @@ btree_get_unique_statistics (THREAD_ENTRY * thread_p, BTID * btid, int *oid_cnt,
...
-  *oid_cnt = root_header->num_oids;
-  *null_cnt = root_header->num_nulls;
-  *key_cnt = root_header->num_keys;
+  *oid_cnt = (unsigned int) root_header->num_oids;
+  *null_cnt = (unsigned int) root_header->num_nulls;
+  *key_cnt = (unsigned int) root_header->num_keys;
+
+  if (root_header->_64.over)
+    {
+      *oid_cnt |= (((long long) root_header->_64.num_oids) << 32);
```
hornetmj, line 6237: Bit operation was performed after the 10-bit memory was converted to 64-bit. This code looks a bit dangerous, have any tests been performed to verify this part?

src/storage/btree.c
```diff
@@ -14089,10 +14102,18 @@ btree_reflect_global_unique_statistics (THREAD_ENTRY * thread_p, GLOBAL_UNIQUE_S
...
+
 	  /* update header information */
-	  root_header->num_nulls = unique_stat_info->unique_stats.num_nulls;
-	  root_header->num_oids = unique_stat_info->unique_stats.num_oids;
-	  root_header->num_keys = unique_stat_info->unique_stats.num_keys;
+	  root_header->num_nulls = unique_stat_info->unique_stats.num_nulls & 0xffffffff;
+	  root_header->num_oids = unique_stat_info->unique_stats.num_oids & 0xffffffff;
+	  root_header->num_keys = unique_stat_info->unique_stats.num_keys & 0xffffffff;
+
+	  over += root_header->_64.num_nulls = unique_stat_info->unique_stats.num_nulls >> 32;
```
hornetmj, line 14112: It seems safe to check if the result of the operation "unique_stat_info->unique_stats.num_nulls >> 32" is greater than 2^10.

src/storage/btree.c
```diff
@@ -16922,7 +16943,7 @@ int
 btree_rv_update_tran_stats (THREAD_ENTRY * thread_p, LOG_RCV * recv)
 {
   char *datap;
-  int num_nulls, num_oids, num_keys;
+  long long num_nulls, num_oids, num_keys;
   BTID btid;
 
   assert (recv->length >= (3 * OR_INT_SIZE) + OR_BTID_ALIGNED_SIZE);
```
hornetmj, line 16949:
```suggestion
  assert (recv->length >= (3 * OR_BIGINT_SIZE) + OR_BTID_ALIGNED_SIZE);
```

src/storage/btree.c
```diff
@@ -16977,15 +16999,42 @@ btree_rv_roothdr_undo_update (THREAD_ENTRY * thread_p, LOG_RCV * recv)
...
+
+      root_header->num_nulls = num_nulls & 0xffffffff;
+      root_header->num_oids = num_oids & 0xffffffff;
+      root_header->num_keys = num_keys & 0xffffffff;
+
+      over += root_header->_64.num_nulls = num_nulls >> 32;
+      over += root_header->_64.num_oids = num_oids >> 32;
+      over += root_header->_64.num_keys = num_keys >> 32;
+
+      root_header->_64.over = (over > 0);
```
hornetmj, line 17037: This routine is used several times. So, how about making macro or inline functions?

src/storage/btree.c
beyondykk9: This routine is not used but the it needs to avoid warning message at compiling.

src/storage/btree.c
beyondykk9: the assert clause is already prior to this code.

src/storage/btree.c
beyondykk9: good point, I will move them.

src/storage/storage_common.h
beyondykk9: it's just slip. I will remove it.

src/transaction/locator_sr.c
beyondykk9: multi_insert_force routine is called only by loaddb command. The log mechanism is not performed while the indexes are loading. So, we should reflect the statistics of the btree header inforcely.

src/transaction/mvcc_table.cpp
beyondykk9: The setting code is originally located furthur down, however, if the valid flag is not set here, this mvcc build routine can be called endless recursively by gtb_load_global_statistics_to_tran. And there are no setting conditions as you know.
The race condition will not be occurred you think because it is "thread free" (means just one transaction is concerned)

src/transaction/mvcc_table.cpp
beyondykk9: It seems not to be valid code. I will fix to reset the snapshot.valid flag to false.

src/storage/btree.c
beyondykk9: Yes, the code is verfied enough.

src/storage/btree.c
beyondykk9: unique_stat_info->unique_stats.num_nulls is 64 bit and overflown bits of 42 is not considered because the 42 bits is enough to represent the count reasonably.

src/storage/btree.c
beyondykk9: Ok, good point. I will fix it.

src/storage/btree.c
beyondykk9: I agree with you but I keep the style for easy debugging.

src/storage/btree.c
beyondykk9: the num_x's means only for unique_pk, so this routine is not important and the values is always same as initial (-1)

src/storage/btree_load.h
beyondykk9: I will change the comments.

src/storage/btree.c
beyondykk9: I checked the INT64 but the INT64 occationally means long type (means 4 bytes), and on also windows "long long"  type always means 64 bit.

src/storage/btree.c
beyondykk9: It's unused routine I remained it to avoid warning at compiling.

src/storage/btree.c
hgryoo: INT64 is defined at include/system.h, and referring to the following, it seems to always guarantee size of 8 bytes.
```
// include/system.h
#ifdef HAVE_INT64_T
typedef int64_t INT64;
#else
#if SIZEOF_LONG == 8
typedef long INT64;
#elif SIZEOF_LONG_LONG == 8
typedef long long INT64;
#else
#error "Error: INT64"
#endif
#endif
```

src/transaction/log_tran_table.c
```diff
@@ -126,10 +126,11 @@ static LOG_TRAN_BTID_UNIQUE_STATS *logtb_tran_create_btid_unique_stats (THREAD_E
 static int logtb_tran_update_delta_hash_func (THREAD_ENTRY * thread_p, void *data, void *args);
 static int logtb_tran_load_global_stats_func (THREAD_ENTRY * thread_p, void *data, void *args);
 static int logtb_tran_reset_cos_func (THREAD_ENTRY * thread_p, void *data, void *args);
-static int logtb_load_global_statistics_to_tran (THREAD_ENTRY * thread_p);
 static int logtb_create_unique_stats_from_repr (THREAD_ENTRY * thread_p, OID * class_oid);
+#if 0
 static GLOBAL_UNIQUE_STATS *logtb_get_global_unique_stats_entry (THREAD_ENTRY * thread_p, BTID * btid,
 								 bool load_at_creation);
+#endif
```
hgryoo: This function declaration moved to log_impl.h file. disabling this block is not necessary.

src/storage/btree.c
wrlawodms: Aha! this is an unused function. OK! How about commenting that "it needs computation with _64" shortly here?

src/storage/btree.c
wrlawodms: But it checks `root_header != NULL` again at the if-condition below. I suppose that the code writer wanted to skip updating root header if it fails to fix the page in **RELEASE** mode. I'm not sure whether it's good or not, but I think we should keep the intention. To do that, I reckon that we should put the code under the if-condition. What do you think?

src/storage/btree.c
wrlawodms: Oh, I got it. And it seems that you considered it already in `btree_get_unique_statistics()`. Good :)

src/transaction/locator_sr.c
wrlawodms: Could you elaborate on "The log mechanism is not performed"? Many logs are written during `multi_insert_force()`. They aren't working? What happens if a server crash during loaddb?

src/transaction/mvcc_table.cpp
wrlawodms: Yes, it is a transaction-bind variable, but some transaction-dependent code could access this. I'm not sure whether that kind of code exist or not. It needs further analysis to guarantee the safety (or you already looked through all the cases in which this variable used?). In my opinion, to keep it as it is is the simplest way if possible. So, I'd like to know the exact point calling it recursively in `gtb_load_global_statistics_to_tran()`?

src/transaction/mvcc_table.cpp
wrlawodms: Why didn't it return here and now it has to return here?

src/transaction/mvcc_table.cpp
wrlawodms: There is no memory barrier around it, so I can guess it is not referred to by others as you said. Then, we can remove `tdes.mvccinfo.snapshot.valid = true` below, but in terms of code readability, it still would be better to leave as it is.

src/transaction/log_tran_table.c
beyondykk9: Ok, I will delete the lines.

src/storage/btree.c
beyondykk9: in 64 bit linux system the "int" or "long" type has 8 bytes. But in struct it represents 4 bytes, it araise conflict at compiling.
So, I had no choice else using the "long long" type.

src/storage/btree.c
beyondykk9: Ok, I agree with you. I will do.

src/transaction/locator_sr.c
beyondykk9: In common, btree header statistics are reflected by the following mechanism, but in the case of loaddb, this mechanism does not work.

**log_commit > log_complete > logpb_checkpoint > logtb_reflect_global_unique_stats_to_btree**

So, to reflect the stats, I called the **logtb_reflect_global_unique_stats_to_btree** routine directly.
From the log manager's point of view, please comment on how it should be reflected.

src/transaction/mvcc_table.cpp
byungwook-kim: Ok, I will leave as it is.

src/storage/btree.c
```diff
@@ -5618,11 +5618,17 @@ xbtree_add_index (THREAD_ENTRY * thread_p, BTID * btid, TP_DOMAIN * key_type, OI
...
-      root_header->num_nulls = num_nulls;
-      root_header->num_keys = num_keys;
+      root_header->num_oids = num_oids & 0xffffffff;
+      root_header->num_nulls = num_nulls & 0xffffffff;
+      root_header->num_keys = num_keys & 0xffffffff;
       root_header->unique_pk = unique_pk;
 
+      over += root_header->_64.num_oids = num_oids >> 32;
+      over += root_header->_64.num_nulls = num_nulls >> 32;
+      over += root_header->_64.num_keys = num_keys >> 32;
```
shparkcubrid, line 5628: Can't 'over' overflow and be negative?

src/transaction/mvcc_table.cpp
beyondykk9: It seems "return" is not valid here. It's better to leave the error code without returning immediately.

src/transaction/mvcc_table.cpp
beyondykk9: Ok, it's better to leave as it is for readability.

src/storage/btree.c
beyondykk9: num_x's uses 42 bits of 64 bits, so the overflown is not possible.

src/transaction/locator_sr.c
wrlawodms: **log_commit -> logtb_complete_mvcc->logtb_update_global_unique_stats_by_delta** : leaves a log for updating the global stat without reflecting to a corresponding btree root header.
**logpb_checkpoint->logtb_reflect_global_unique_stats_to_btree**: reflects the global stat to the btree root header without a log.
And if the logpb_checkpoint() is not executed during running, the global stats in the tree root header are updated during recovery. That's what I understand. The way see it, if operations commits in loaddb and the local stats of each transactions are updated properly during the buik insert, it will work. 

1. Is there something I misunderstand? What do you mean by "does not work in the case of loaddb"?
2. Where are local stats updated, by the way?

src/transaction/mvcc_table.cpp
wrlawodms: I mean that we should find a way not to set true here. I'm leaving this comment to avoid miscommunication just in case.

src/transaction/locator_sr.c
beyondykk9: I checked the routine related to index header statistics log more detaily. **logtb_update_global_unique_stats_by_delta** already write undo/redo log, it's called from loaddb's  locator_multi_insert_force. Therefore, the log would be written correctly. please check the routine **logtb_update_global_unique_stats_by_delta**.

src/transaction/mvcc_table.cpp
beyondykk9: I misunderstood what you said, but I couldn't find a way to set the flag true else here.

src/transaction/locator_sr.c
wrlawodms: Yes, that is what I mean. So, the log for reflecting is written, then why should we reflect it manually?

src/transaction/mvcc_table.cpp
wrlawodms: So, I'd like to know the exact point calling it recursively in gtb_load_global_statistics_to_tran().

src/transaction/locator_sr.c
beyondykk9: loaddb doesn't use mvcc and the loaddb transaction doesn't reach to **complete_mvcc** which is used at only mvcc.
whereas in SA_MODE **logtb_tran_update_all_global_unique_stats** is called. So, I'm goning to check if I coud modify the code to call the **logtb_tran_update_all_global_unique_stats** in SERVER mode.

beyondykk9: I found a bug-like at the logtb_complete_mvcc routine as below:
```
#if defined(SA_MODE)
      if (committed && logtb_tran_update_all_global_unique_stats (thread_p) != NO_ERROR)
	{
	  assert (false);
	}
#else	/* !SA_MODE */	       /* SERVER_MODE */
```
"#if defined(SA_MODE)" looks not proper, I tested the loaddb with removing this conditional macro. The statistics was reflected successfully. However, I don't know why the code is wrtten in this way.

src/transaction/mvcc_table.cpp
beyondykk9: In partitioned table, the call stack is as below:
```
logtb_tran_load_global_stats_func
	partition_load_pruning_context
		heap_get_class_partitions
			heap_class_get_partition_info
				logtb_get_mvcc_snapshot
					logtb_tran_load_global_stats_func
```
endless recursve call stack is made because mvcc_snapshot.valid flag is not set true neverthless mvcc info lis loading yet.
If we should modify the mechanism, we should add a flag named "loading" as boolean type. The flag must be set before calling  logtb_tran_load_global_stats_func and reset it after the function call is done.

src/transaction/locator_sr.c
wrlawodms: It seems to assume that all operations that potentially update the unique stats leave some mvcc logs. Regarding HEAP operation, the bulk updating doesn't leave MVCC logs, but I don't really know the case of index.

src/transaction/mvcc_table.cpp
wrlawodms: Thank you for the information :). The way I see it, `loading` or `building` flag for a new state could be a quick solution, or we can make a state variable expending `valid`. @hornetmj What do you think about it?

src/storage/heap_file.c
```diff
@@ -10806,11 +10807,14 @@ heap_class_get_partition_info (THREAD_ENTRY * thread_p, const OID * class_oid, O
       return ER_FAILED;
     }
 
-  scan_cache.mvcc_snapshot = logtb_get_mvcc_snapshot (thread_p);
-  if (scan_cache.mvcc_snapshot == NULL)
+  if (!tdes->mvccinfo.snapshot.loading)
```
wrlawodms: 1. `is_building" looks more proper, because it is kept "on" during `build_mvcc_info()`
2. I don't think this flag is only for partitioning table. This should be checked and set in the beginning of `build_mvcc_info()` and return it immediately to avoid building mvcc info recursively. If you want to only prevent partitioning case, you should use more specific variable and mechanism.

src/storage/heap_file.c
```diff
@@ -10806,11 +10807,14 @@ heap_class_get_partition_info (THREAD_ENTRY * thread_p, const OID * class_oid, O
       return ER_FAILED;
     }
 
-  scan_cache.mvcc_snapshot = logtb_get_mvcc_snapshot (thread_p);
-  if (scan_cache.mvcc_snapshot == NULL)
+  if (tdes != NULL && !tdes->mvccinfo.snapshot.loading)
```
wrlawodms: 1. `is_building` looks more proper, because it is kept "on" during build_mvcc_info().
2. I don't think this flag is only for partitioning table. This should be checked and set in the beginning of build_mvcc_info() and return it immediately to avoid building mvcc info recursively. If you want to only prevent partitioning case, you should use more specific variable and mechanism.

wrlawodms (changes requested):

src/storage/heap_file.c
wrlawodms: And this change affects the general mvcc building, I think we should get a review from @hornetmj before applying it.

wrlawodms: It turns out that this solution has flaw and we should complete building mvcc in `heap_class_get_partition_info()`. During getting partition info it could occur core if it has to access mvcc info.

wrlawodms: We discussed it and with further analysis, it turns out that we don't need to build mvcc info in `heap_class_get_partition_info()`. We can revert the change related to mvcc build and remove mvcc building in the function to get rid of the infinite recursive calling.

src/transaction/mvcc_table.cpp
wrlawodms: We discussed it and with further analysis, it turns out that we don't need to build mvcc info in `heap_class_get_partition_info()`. We can revert the change related to mvcc build and remove `logtb_get_mvcc_snapshot()` in the function to get rid of the infinite recursive calling.

beyondykk9: logtb_get_mvcc_snapshot at heap_class_get_partition_info calling at was removed not to make endleess recursive call.

src/transaction/locator_sr.c
beyondykk9: I fixed the routine to update stats on SERVER_MODE too.

src/transaction/log_impl.h
```diff
@@ -1219,8 +1219,10 @@ extern void logtb_complete_mvcc (THREAD_ENTRY * thread_p, LOG_TDES * tdes, bool
 extern void logtb_complete_sub_mvcc (THREAD_ENTRY * thread_p, LOG_TDES * tdes);
 
 extern LOG_TRAN_CLASS_COS *logtb_tran_find_class_cos (THREAD_ENTRY * thread_p, const OID * class_oid, bool create);
-extern int logtb_tran_update_unique_stats (THREAD_ENTRY * thread_p, const BTID * btid, int n_keys, int n_oids,
-					   int n_nulls, bool write_to_log);
+extern GLOBAL_UNIQUE_STATS *logtb_get_global_unique_stats_entry (THREAD_ENTRY * thread_p, BTID * btid,
```
wrlawodms: It doesn't seem to have to be extern anymore.

src/storage/btree_load.h
wrlawodms: Let me have your opinion on the name, `_64`. It looks meaningless to me.

src/transaction/locator_sr.c
```diff
@@ -13866,6 +13866,8 @@ locator_multi_insert_force (THREAD_ENTRY * thread_p, HFID * hfid, OID * class_oi
 	}
     }
 
+  logtb_load_global_statistics_to_tran (thread_p);
```
wrlawodms: Could you tell me the reason we should load stats here?

The way I see it,
MVCC: We have to load once per transaction, and a locator function can be called several times basically.
NON-MVCC: no need to use global stats.

src/transaction/log_impl.h
beyondykk9: Ok, I fixed it.

src/storage/heap_file.c
beyondykk9: I reverted the code and remove the calling "logtb_get_mvcc_snapshot" from heap_class_get_partition_info().

src/transaction/log_tran_table.c
```diff
@@ -126,10 +126,7 @@ static LOG_TRAN_BTID_UNIQUE_STATS *logtb_tran_create_btid_unique_stats (THREAD_E
 static int logtb_tran_update_delta_hash_func (THREAD_ENTRY * thread_p, void *data, void *args);
 static int logtb_tran_load_global_stats_func (THREAD_ENTRY * thread_p, void *data, void *args);
 static int logtb_tran_reset_cos_func (THREAD_ENTRY * thread_p, void *data, void *args);
-static int logtb_load_global_statistics_to_tran (THREAD_ENTRY * thread_p);
 static int logtb_create_unique_stats_from_repr (THREAD_ENTRY * thread_p, OID * class_oid);
-static GLOBAL_UNIQUE_STATS *logtb_get_global_unique_stats_entry (THREAD_ENTRY * thread_p, BTID * btid,
```
wrlawodms, line 131: This has to be reverted too.

src/transaction/log_tran_table.c
```diff
@@ -4011,12 +4009,12 @@ logtb_complete_mvcc (THREAD_ENTRY * thread_p, LOG_TDES * tdes, bool committed)
     }
   else
     {
-#if defined(SA_MODE)
       if (committed && logtb_tran_update_all_global_unique_stats (thread_p) != NO_ERROR)
 	{
 	  assert (false);
 	}
-#else	/* !SA_MODE */	       /* SERVER_MODE */
+#if defined(SERVER_MODE)
```
wrlawodms: The comment below is explicitly saying that only the stat for db_serial has to be updated with no MVCCID being generated. If you conclude that it is not true by analysis, we have to fix the code below too. In the changed code, it seems that the stat for `db_serial` is already reflected in `logtb_tran_update_all_global_unique_stats()`.

Please share your analysis about it.

wrlawodms (changes requested):

src/transaction/locator_sr.c
wrlawodms: In the case of MVCC, it is called already in `build_mvcc_info()`.

src/transaction/log_tran_table.c
wrlawodms: In the previous code, only the stat for db_serial is reflected if no MVCCID is generated. However, in the changed code, all entries are iterated in the stat hash, and It could impact normal transaction processing performance.

src/transaction/locator_sr.c
beyondykk9: it's just a slip not to be reverted. I will fixed it.

src/transaction/log_tran_table.c
beyondykk9: Ok.

src/transaction/log_tran_table.c
beyondykk9: I don't know the reason why the reflection code for db_serial is applied in only the "SERVER_MODE".
I left the codes that reflect the serial key in SERVER_MODE since I can not figure out the intention.
I did it to avoid modifying the existing code as much as possible to avoid unintended side effects.
I will remove the code if it conclues meaningless.

wrlawodms: No, the stat for db_serial is also reflected in SA_MODE. Only the stat for db_serial is reflected in SERVER_MODE, but we are changing to reflect others too. Do I misunderstand?

beyondykk9: http://jira.cubrid.org/browse/CUBRIDSUS-17437
The db_serial-related stats reflection under SERVER_MODE is added at the above issue for patching "checkdb" bug.
However, "show index header" statement also has a bug not reflects the stats.
So, I think that the SERVER_MODE limitation and db_serial-related stats reflect code should be removed.

src/storage/btree.c
ctshim: Where is the guarantee that the value of the num_xxxs variables passed as parameters is within the 42-bit size?

beyondykk9: please see the definition of root_header, _64.num_xxx:10, it's a 64 bit number uses 42-bit only.

src/storage/btree_load.h
beyondykk9: _64 is very short name strands for 64 bit extention, and it's just temporary use until it comes next database image changes

ctshim (approved):

wrlawodms (approved):

src/storage/btree.c
```diff
@@ -16979,13 +17001,40 @@ btree_rv_roothdr_undo_update (THREAD_ENTRY * thread_p, LOG_RCV * recv)
...
+      int over = 0;
+      long long delta;
+
+      num_nulls = (unsigned int) root_header->num_nulls;
+      num_oids = (unsigned int) root_header->num_oids;
+      num_keys = (unsigned int) root_header->num_keys;
+
+      if (root_header->_64.over)
+	{
+	  num_nulls |= (long long) root_header->num_nulls << 32;
```
hyunikn, line 17013: Is it root_header->_64.num_nulls instead of root_header->num_nulls ? (and similar for the following two lines)

beyondykk9: root_header->_64 (has three 10 bit field) is used for extending the num_xxx's to 64 bit (used 42 bits only). originally num_xxx's was 32 bit and it needs to extend 64 bit (long long type).

hyunikn: Yes, and it seems that this line should set higher 10 bits of num_nulls (the lower 32 bits are set at line 17007). 
More specifically, this line should be 
"num_nulls |= (long long) root_header->_64.num_nulls << 32;"
Am I missing something here?

hgryoo (approved):

src/storage/heap_file.c
```diff
@@ -10806,13 +10806,6 @@ heap_class_get_partition_info (THREAD_ENTRY * thread_p, const OID * class_oid, O
       return ER_FAILED;
     }
 
-  scan_cache.mvcc_snapshot = logtb_get_mvcc_snapshot (thread_p);
```
shparkcubrid, line 10809: Why was this source code removed?

src/transaction/mvcc_table.cpp
```diff
@@ -297,15 +297,12 @@ mvcctable::build_mvcc_info (log_tdes &tdes)
...
-      /* load statistics temporary disabled need to be enabled when activate count optimization */
-#if 0
-      /* load global statistics. This must take place here and nowhere else. */
-      if (logtb_load_global_statistics_to_tran (thread_p) != NO_ERROR)
+
+      if (logtb_load_global_statistics_to_tran (thread_get_thread_entry_info())!= NO_ERROR)
 	{
+	  /* just error setting without returning for further processing */
 	  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_MVCC_CANT_GET_SNAPSHOT, 0);
-	  error_code = ER_MVCC_CANT_GET_SNAPSHOT;
```
shparkcubrid, line 306: Why was this error code removed?

src/storage/heap_file.c
beyondykk9: this code is not necessary because the mvcc snapshot is already built and this deletion issue has been discussed with @wrlawodms enoughly.

src/transaction/mvcc_table.cpp
beyondykk9: The error code is not used though this function.

src/storage/btree.c
beyondykk9: I fixed it as you suggested.

hyunikn (approved):

youngjinj (approved):

joohok (approved):

shparkcubrid (approved):

hornetmj (approved):

src/transaction/log_tran_table.c
```diff
@@ -4938,41 +4906,41 @@ logtb_update_global_unique_stats_by_delta (THREAD_ENTRY * thread_p, BTID * btid,
...
       datap = (char *) undo_rec.data;
       OR_PUT_BTID (datap, btid);
       datap += OR_BTID_ALIGNED_SIZE;
-      OR_PUT_INT (datap, null_delta);
-      datap += OR_INT_SIZE;
-      OR_PUT_INT (datap, oid_delta);
-      datap += OR_INT_SIZE;
-      OR_PUT_INT (datap, key_delta);
-      datap += OR_INT_SIZE;
+      OR_PUT_BIGINT (datap, &null_delta);
```
cristiarg, line 4922: @byungwook-kim 
I left a comment regarding order of packs/unpacks in issue - http://jira.cubrid.org/browse/CBRD-24110 .

PR #3168, opened 2021-11-10
[CBRD-24102] Applylogdb may skip creating TDE table with the TDE module unloaded

wrlawodms:
http://jira.cubrid.org/browse/CBRD-24102

It includes a small refactoring: tde_is_loaded() - accessing tde_Cipher indirectly to see whether it is loaded or not.


eido5 (approved):

src/communication/network_interface_cl.c
```diff
@@ -1886,6 +1886,40 @@ tde_get_data_keys ()
...
+ */
+int
+tde_is_loaded (int *is_loaded)
+{
+#if defined(CS_MODE)
+  int error = ER_NET_CLIENT_DATA_RECEIVE;
+  int req_error, area_size;
+  char *ptr;
+  OR_ALIGNED_BUF (OR_INT_SIZE) a_reply;
+  char *reply, *area;
```
```suggestion
#if defined(CS_MODE)
  int req_error;
  OR_ALIGNED_BUF (OR_INT_SIZE) a_reply;
  char *reply;
```

src/communication/network_interface_cl.c
```diff
@@ -1886,6 +1886,40 @@ tde_get_data_keys ()
...
+
+  reply = OR_ALIGNED_BUF_START (a_reply);
+
+  req_error =
+    net_client_request (NET_SERVER_TDE_IS_LOADED, NULL, 0, reply, OR_ALIGNED_BUF_SIZE (a_reply), NULL, 0, NULL, 0);
+  if (!req_error)
+    {
+      ptr = or_unpack_int (reply, is_loaded);
```
```suggestion
      (void *) or_unpack_int (reply, is_loaded);
```

src/storage/tde.c
```diff
@@ -299,6 +299,17 @@ tde_cipher_initialize (THREAD_ENTRY * thread_p, const HFID * keyinfo_hfid, const
   return err;
 }
 
+/*
+ * tde_is_loaded () - Is the TDE module initialized correctly?
+ *
+ * return                 : trun or false
```
hornetmj:
```suggestion
 * return                 : true or false
```

hornetmj (approved):

wrlawodms:
@hornetmj Thank your for your attentive review.

joohok (approved):

joohok:
it seems that code style check failure is not from your change.

wrlawodms:
@mhoh3963 the code style check failure is not from my change. Please force-merge it.

The commit message could be like below:

```
http://jira.cubrid.org/browse/CBRD-24102

In HA, if the slave node can't load the TDE module and runs into a statement related to TDE, it is supposed to just stop. However, it skips a CREATE TABLE ENCRYPT statement after about 10 minutes. 
Cause: During failure to create table due to the not loaded TDE-module, it actually creates a table (file), fails, and destroys it. These files are registered as dropped files. Once a file is registered as a dropped file, even if the file header page is deallocated, it cannot be used as a header page of another newly created file by a transaction of which mvccid is lower than the mvccid of the dropped file. The mvccid of dropped file is the 1 + mvcc id of the transaction destroying the file.
Solution: Moving the blocking point to do_create_entity() in the case of "CREATE TABLE".
```

PR #3169, opened 2021-11-10
[LETS-287] Fix system operation level after atomic start

ribram:
http://jira.cubrid.org/browse/LETS-287

Atomic sysop start is not expected to bump the system operation level during analysis. It is wrongfully bumped after checkpoint analysis.

Fix by bumping the system operation level, only if sysop start postpone is not null.

EDIT: unit test has to be updated too; also fixed some issues and reorganized some global resources/functions into test_env_chkpt.

### DIFF
--- a/src/transaction/log_checkpoint_info.cpp
+++ b/src/transaction/log_checkpoint_info.cpp
@@ -335,14 +335,7 @@ namespace cublog
 	      }
 	  }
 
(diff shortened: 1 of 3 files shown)


ribram:
@hornetmj this too should be back-ported to develop.

cristiarg (approved):

PR #3177, opened 2021-11-15
[CBRD-24099] error during recovery analysis when a sysop expected to have postpone info does not have it

cristiarg:
http://jira.cubrid.org/browse/CBRD-24099

- an uncommited transaction, that is executing either system operation postpone or an atomic system operation, is assumed to contain postpone information without actually checking.
- atomic sysop start is not expected to bump the system operation level during analysis. It is wrongfully bumped after checkpoint analysis.
- fix by bumping the system operation level, only if sysop start postpone is not null.

### DIFF
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -9891,6 +9891,8 @@ log_read_sysop_start_postpone (THREAD_ENTRY * thread_p, LOG_LSA * log_lsa, LOG_P
 {
   int error_code = NO_ERROR;
 
+  assert (!log_lsa->is_null ());
+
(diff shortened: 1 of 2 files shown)


ribram (approved):

mhoh3963:
@cristiarg 
I think @wrlawodms will do a better review than me on this part, so I add him to reviewer.

mhoh3963 (approved):

hornetmj (approved):

wrlawodms (approved):
Sorry for the late review. I was isolated because of COVID-19 out of blue.

PR #3178, opened 2021-11-15
[LETS-251] Add log records for START/END ATOMIC REPLICATION

adrianbrab:
http://jira.cubrid.org/browse/LETS-251

Added 2 new types of log records `LOG_START_ATOMIC_REPL` and `LOG_END_ATOMIC_REPL`, to be used in the replication system.

### DIFF
--- a/src/transaction/log_applier.c
+++ b/src/transaction/log_applier.c
@@ -6177,6 +6177,8 @@ la_log_record_process (LOG_RECORD_HEADER * lrec, LOG_LSA * final, LOG_PAGE * pg_
       break;
 
     case LOG_DUMMY_CRASH_RECOVERY:
+    case LOG_START_ATOMIC_REPL:
+    case LOG_END_ATOMIC_REPL:
       snprintf (buffer, sizeof (buffer), "process log record (type:%d). skip this log record. LSA: %lld|%d",
 		lrec->type, (long long int) final->pageid, (int) final->offset);
       er_set (ER_NOTIFICATION_SEVERITY, ARG_FILE_LINE, ER_HA_GENERIC_ERROR, 1, buffer);
--- a/src/transaction/log_applier.c
+++ b/src/transaction/log_applier.c
@@ -6177,8 +6177,6 @@ la_log_record_process (LOG_RECORD_HEADER * lrec, LOG_LSA * final, LOG_PAGE * pg_
       break;
 
     case LOG_DUMMY_CRASH_RECOVERY:
-    case LOG_START_ATOMIC_REPL:
-    case LOG_END_ATOMIC_REPL:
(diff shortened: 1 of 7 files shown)


src/transaction/log_applier.c
```diff
@@ -6177,6 +6177,8 @@ la_log_record_process (LOG_RECORD_HEADER * lrec, LOG_LSA * final, LOG_PAGE * pg_
       break;
 
     case LOG_DUMMY_CRASH_RECOVERY:
+    case LOG_START_ATOMIC_REPL:
+    case LOG_END_ATOMIC_REPL:
```
ribram: In this case, let the switch fall to default and ignore the log record. The notification used here for crash log record is not needed.
```suggestion
```

src/transaction/log_manager.c
```diff
@@ -7923,6 +7930,8 @@ log_rollback (THREAD_ENTRY * thread_p, LOG_TDES * tdes, const LOG_LSA * upto_lsa
 	    case LOG_END_OF_LOG:
 	    case LOG_SMALLER_LOGREC_TYPE:
 	    case LOG_LARGER_LOGREC_TYPE:
+	    case LOG_START_ATOMIC_REPL:
```
ribram: Don't add a fatal error when rollback hits a start/end atomic log record. No action is needed here; move them to the same group as `LOG_SUPPLEMENTAL_INFO`.

src/transaction/log_manager.c
```diff
@@ -8390,6 +8399,8 @@ log_do_postpone (THREAD_ENTRY * thread_p, LOG_TDES * tdes, LOG_LSA * start_postp
 		    case LOG_DUMMY_CRASH_RECOVERY:
 		    case LOG_SMALLER_LOGREC_TYPE:
 		    case LOG_LARGER_LOGREC_TYPE:
+		    case LOG_START_ATOMIC_REPL:
```
ribram: Same case here, don't log an error when finding start/end atomic replication. No action is required, so they belong to the same group as `LOG_SUPPLEMENTAL_INFO`.

src/transaction/log_recovery.c
```diff
@@ -2747,6 +2749,8 @@ log_recovery_undo (THREAD_ENTRY * thread_p)
 		case LOG_2PC_COMMIT_INFORM_PARTICPS:
 		case LOG_2PC_RECV_ACK:
 		case LOG_DUMMY_CRASH_RECOVERY:
+		case LOG_START_ATOMIC_REPL:
```
ribram: Same here, move to `LOG_SUPPLEMENTAL_INFO` group.

ribram (approved):

PR #3180, opened 2021-11-16
[LETS-301] Don't hold LOG_CS while recovery needs to fix pages

ribram:
http://jira.cubrid.org/browse/LETS-301

Recovery needs to fix pages to apply changes during redo, undo and finish postpones. In the background, page flush daemon writes bcb's to disk making room for new pages to be loaded in page buffer.

The page flush daemon needs to lock log critical section to write the log to disk, since log daemons are not yet operational (they become operational later on).

As a result, log critical section must be released during recovery phases that have to fix pages and, if necessary, victimizes pages.

### DIFF
--- a/src/transaction/log_recovery.c
+++ b/src/transaction/log_recovery.c
@@ -2350,7 +2350,7 @@ log_recovery_undo (THREAD_ENTRY * thread_p)
   //
   // Unblock log critical section while running undo.
   //
-  // During undo, pages will have to be fixed. Page buffer may be full and pages have be loaded from disk, by
+  // During undo, pages will have to be fixed. Page buffer may be full and pages have be loaded from disk only by
   // victimizing other pages. Page flush thread must be able to operate and it will require the log critical section.
   //
   // Lock log critical section again on exit.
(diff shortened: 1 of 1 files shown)


src/transaction/log_recovery.c
```diff
@@ -2347,6 +2347,20 @@ log_recovery_undo (THREAD_ENTRY * thread_p)
 
   aligned_log_pgbuf = PTR_ALIGN (log_pgbuf, MAX_ALIGNMENT);
 
+  //
+  // Unblock log critical section while running undo.
+  //
+  // During undo, pages will have to be fixed. Page buffer may be full and pages have be loaded from disk, by
```
cristiarg:
```suggestion
  // During undo, pages will have to be fixed. Page buffer may be full and pages must be loaded from disk only by
```

cristiarg (approved):
So, if the actual large parts of `log_recovery_redo` and `log_recovery_undo` unlock the log critical section, what point is there to actually need to have it owned when entering these functions?

ribram:
> So, if the actual large parts of `log_recovery_redo` and `log_recovery_undo` unlock the log critical section, what point is there to actually need to have it owned when entering these functions?

Good question. There are some actions that are done in multiple steps during recovery that require the lock on critical section. For example, whenever the log header or the log pages are flushed to disk.

`log_recovery_undo` doesn't require the log critical section lock. It is possible to unlock the LOG_CS and lock back outside the `log_recovery_undo`, in `log_recovery`.

However, in the last part of `log_recovery_redo`, log critical section is required to flush the log to disk.

The recovery system used to be single-threaded, with no parallel redo appliers and no background page buffer daemons. It did not care for how long the critical section was held; it was locked from the beginning and held for the entire recovery.

Only recently we have introduced the redo appliers and the background page buffer daemons that need concurrent access on LOG_CS. The safe approach is to release the critical section during specific parts of the recovery where we need the concurrent access on log.

PR #3181, opened 2021-11-16
[LETS-187] Add xboot_initialize_remote_storage_server()

ribram:
http://jira.cubrid.org/browse/LETS-187

Add the function that puts new database created for remote storage in databases.txt file. The function follows the code handling databases.txt entry from `xboot_initialize_server`, with the next conditions:

  - Lock dabatases.txt file while adding new entry
  - Do not overwrite existing entry; return with error if the same database already exists

Small refactoring in the xboot_initialize_server code; reorganized conditional branches to clarify the logic flow.

### DIFF
--- a/src/transaction/boot_sr.c
+++ b/src/transaction/boot_sr.c
@@ -1603,49 +1603,50 @@ xboot_initialize_server (const BOOT_CLIENT_CREDENTIAL * client_credential, BOOT_
     }
 #endif // WINDOWS && DONT_USE_MANDATORY_LOCK_IN_WINDOWS
 
-  assert (dir != NULL);
-
-  db = cfg_find_db_list (dir, client_credential->get_db_name ());
-  if (db != NULL)
+  if (dir != NULL)
     {
-      if (db_overwrite == false)
+      db = cfg_find_db_list (dir, client_credential->get_db_name ());
+      if (db != NULL)
 	{
-	  /* There is a database with the same name and we cannot overwrite it */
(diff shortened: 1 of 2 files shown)


src/transaction/boot_sr.c
```diff
@@ -1577,70 +1578,74 @@ xboot_initialize_server (const BOOT_CLIENT_CREDENTIAL * client_credential, BOOT_
...
-      fileio_dismount (thread_p, dbtxt_vdes);	/* unlock the directory file */
-      dbtxt_vdes = NULL_VOLDES;
+      if (dbtxt_vdes != NULL_VOLDES)
+	{
+	  fileio_dismount (thread_p, dbtxt_vdes);	/* unlock the directory file */
+	  dbtxt_vdes = NULL_VOLDES;
+	}
       cfg_free_directory (dir);
       dir = NULL;
+      db = NULL;
```
cristiarg, line 1648: replace with [optional] assert:
```suggestion
      assert ((db_overwrite && db == nullptr) || !db_overwrite);
```
or even better
```suggestion
      assert (db == nullptr);
```
and `db = nullptr;` should be moved to else scope ending at ~1638.
I understand that if `db` is non-null here, there's a logic problem.

cristiarg (approved):

src/transaction/boot_sr.c
ribram: db may be non-null if db_overwrite was true. db is set to null as an indication that it is not used below.

PR #3186, opened 2021-11-17
[LETS-283] Add LOG_TRANTABLE_SNAPSHOT log record

ribram:
http://jira.cubrid.org/browse/LETS-283

Add LOG_TRANTABLE_SNAPSHOT log record for each transaction table snapshot taken by transaction server.

### DIFF
--- a/src/transaction/log_page_buffer.c
+++ b/src/transaction/log_page_buffer.c
@@ -7352,6 +7352,11 @@ logpb_checkpoint_trantable (THREAD_ENTRY * const thread_p)
     LOG_LSA dummy_smallest_tran_lsa = NULL_LSA;
     trantable_checkpoint_info.load_trantable_snapshot (thread_p, dummy_smallest_tran_lsa);
 
+    // Append to log
+    log_append_trantable_snapshot (thread_p, trantable_checkpoint_info);
+
+    // Write to log meta file
+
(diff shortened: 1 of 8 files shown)


adrianbrab (approved):

src/transaction/log_page_buffer.c
```diff
@@ -7352,6 +7352,11 @@ logpb_checkpoint_trantable (THREAD_ENTRY * const thread_p)
     LOG_LSA dummy_smallest_tran_lsa = NULL_LSA;
     trantable_checkpoint_info.load_trantable_snapshot (thread_p, dummy_smallest_tran_lsa);
 
+    // Append to log
+    log_append_trantable_snapshot (thread_p, trantable_checkpoint_info);
```
cristiarg, line 7364: A comment would be useful that the trantable is saved to metafile and also added to log and for which purpose for each.
I assume that, for the long term, it can be done without writing it to metafile and, instead, recovering it from the log - maybe during a supplemental initialization step (instead of, or alongside recovery analysis).
Am I correct?

cristiarg (approved):

src/transaction/log_page_buffer.c
ribram: Correct. I updated the comment and added a todo.

PR #3187, opened 2021-11-18
[CBRD-24111] Problem with view merging when there is a variable definition ( ':=' ) in the where clause.

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-24111

if the variable definition(:=) is in the WHERE of the main query, I fix so that view merge is not performed.

### DIFF
--- a/src/parser/parser_support.c
+++ b/src/parser/parser_support.c
@@ -2975,6 +2975,7 @@ pt_is_define_vars (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int *cont
 {
   bool *is_define_vars = (bool *) arg;
   *continue_walk = PT_CONTINUE_WALK;
+  assert (has_define_vars != NULL);
 
   if (*is_define_vars)
     {
--- a/src/parser/parser_support.c
+++ b/src/parser/parser_support.c
@@ -2975,7 +2975,7 @@ pt_is_define_vars (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int *cont
 {
   bool *is_define_vars = (bool *) arg;
   *continue_walk = PT_CONTINUE_WALK;
-  assert (has_define_vars != NULL);
+  assert (is_define_vars != NULL);
 
   if (*is_define_vars)
     {
--- a/src/parser/parser_support.c
+++ b/src/parser/parser_support.c
@@ -2947,6 +2947,54 @@ pt_has_aggregate (PARSER_CONTEXT * parser, PT_NODE * node)
 }
 
 /*
+ * pt_has_define_vars () - check if a statement uses define vars ':='
+ * return	: true if the statement uses define vars ':='
+ * parser (in)	: parser context
+ * stmt (in)	: statement
+ */
+bool
+pt_has_define_vars (PARSER_CONTEXT * parser, PT_NODE * stmt)
+{
+  bool has_define_vars = false;
+
+  parser_walk_tree (parser, stmt, pt_is_define_vars, &has_define_vars, NULL, NULL);
+
+  return has_define_vars;
+}
+
+/*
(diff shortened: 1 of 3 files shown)


src/parser/view_transform.c
```diff
@@ -1735,7 +1736,7 @@ mq_substitute_subquery_in_statement (PARSER_CONTEXT * parser, PT_NODE * statemen
 	  rewrite_as_derived = true;
 	}
       /* check for CONNECT BY */
-      else if (tmp_result->info.query.q.select.connect_by)
+      else if (PT_IS_SELECT (tmp_result) && tmp_result->info.query.q.select.connect_by)
```
beyondykk9, line 1739: it looks not related to this issue.

shparkcubrid: There was a problem with node type and I fixed it together. :)

src/parser/parser_support.c
```diff
@@ -2946,6 +2946,54 @@ pt_has_aggregate (PARSER_CONTEXT * parser, PT_NODE * node)
...
+ * arg (in)    :
+ * continue_walk (in) :
+ */
+PT_NODE *
+pt_is_define_vars (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int *continue_walk)
+{
+  bool *is_define_vars = (bool *) arg;
+  *continue_walk = PT_CONTINUE_WALK;
+
+  if (*is_define_vars)
```
hgryoo, line 2980:
```suggestion
  if (is_define_vars && *is_define_vars)
```

hgryoo (approved):

src/parser/parser_support.c
ctshim: In addition to @hgryoo's comment, how about modifying it as follows?
```
 if( is_define_vars == NULL || *is_define_vars)
 {
      /* stop checking, there already is a parameter in the statement */
      return node;
    }
```

shparkcubrid: I'll put an assert.
```
assert (has_define_vars != NULL)
```

ctshim (approved):

beyondykk9 (approved):

src/parser/parser_support.c
```diff
@@ -2946,6 +2946,55 @@ pt_has_aggregate (PARSER_CONTEXT * parser, PT_NODE * node)
...
+ * node (in)   : node
+ * arg (in)    :
+ * continue_walk (in) :
+ */
+PT_NODE *
+pt_is_define_vars (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int *continue_walk)
+{
+  bool *is_define_vars = (bool *) arg;
+  *continue_walk = PT_CONTINUE_WALK;
+  assert (has_define_vars != NULL);
```
youngjinj: **has_define_vars** is a typo. should be changed to **is_define_vars**.

src/parser/parser_support.c
youngjinj: **has_define_vars** is a typo. should be changed to **is_define_vars**.

youngjinj: For pointer variables, I think it is safe to always check the address first and then check the value.
Because assert works only in debug mode, it is difficult to check in release mode.
For this reason, I agree with hgryoo's opinion.

Please consider it as one opinion.

youngjinj (approved):

youngjinj (approved):

src/parser/parser_support.c
shparkcubrid: I fixed it. thanks.

src/parser/parser_support.c
shparkcubrid: I fixed it.

PR #3191, opened 2021-11-19
[CBRD-23842] Resource is initialized regardless of success/failure of cubrid_log_finalize()

joohok:

There was a problem that, 
if cubrid_log_disconnect_server() is failed due to server down, cubrid_log_finalize() returns error without initializing the global variables and memory allocations used in API. 

So, this PR changes this behavior to clear the resources regardless of error from cubrid_log_disconnect_server().

### DIFF
--- a/src/api/cubrid_log.c
+++ b/src/api/cubrid_log.c
@@ -1823,16 +1823,6 @@ cubrid_log_disconnect_server (void)
       free_and_init (recv_data);
     }
 
-  if (g_trace_log != NULL)
-    {
-      fflush (g_trace_log);
-      fclose (g_trace_log);
-      g_trace_log = NULL;
-    }
-
-  css_free_conn (g_conn_entry);
-  g_conn_entry = NULL;
-
   return CUBRID_LOG_SUCCESS;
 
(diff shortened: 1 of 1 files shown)


src/api/cubrid_log.c
```diff
@@ -1857,6 +1847,9 @@ cubrid_log_reset_globals (void)
 {
   int i;
 
+  css_free_conn (g_conn_entry);
```
hornetmj: How about adding 'if (g_conn_entry != NULL)' check? (The g_conn_entry is needed to initialize NULL)

src/api/cubrid_log.c
```diff
@@ -1868,6 +1861,13 @@ cubrid_log_reset_globals (void)
       g_extraction_table = NULL;
     }
 
+  if (g_trace_log != NULL)
+    {
+      fflush (g_trace_log);
+      fclose (g_trace_log);
+      g_trace_log = NULL;
+    }
```
hornetmj: How about executing them in the cubrid_log_reset_tracelog() function?

hornetmj (approved):

PR #3192, opened 2021-11-19
[LETS-288] Always log assigned MVCCID before completing transaction

ribram:
http://jira.cubrid.org/browse/LETS-288

All MVCCID's assigned on ATS are also completed when the transaction is committed or aborted. The MVCCID's have to completed in the same way on the PTS.

The only way PTS may be notified of assigned/completed MVCCID's is through the log. There are MVCC log records that include the transaction MVCCID. However, it is not guaranteed that all assigned MVCCID's end up being logged in an MVCC log record.

If an assigned MVCCID is not logged before its completion through an MVCC log record, it must be logged there.

- Track the last LSA of transaction MVCC log record.
- Before completing MVCC, if the transaction is assigned an MVCCID but it its last MVCC lsa is null, append a new type of log record containing the assigned MVCCID.

The same mechanism needs to be applied when transactions are aborted during recovery undo. Therefore, last MVCC lsa was included in checkpoint/analysis.

Another patch will have to do the MVCCID assignment/completion during PTS replication.


adrianbrab (approved):

src/transaction/log_recovery.c
```diff
@@ -2809,21 +2798,7 @@ log_recovery_undo (THREAD_ENTRY * thread_p)
 		  /*
 		   * Remove the transaction from the recovery process
 		   */
-
-		  /* Clear MVCCID */
-		  tdes->mvccinfo.id = MVCCID_NULL;
```
cristiarg, line 2814: Same about `tdes->mvccinfo.id`.

src/transaction/log_recovery.c
```diff
@@ -2777,21 +2780,7 @@ log_recovery_undo (THREAD_ENTRY * thread_p)
 #endif /* CUBRID_DEBUG */
 		  /* Remove the transaction from the recovery process */
 		  assert (false);
-
-		  /* Clear MVCCID */
-		  tdes->mvccinfo.id = MVCCID_NULL;
```
cristiarg, line 2782: In the new code, tdes->mvccinfo.id is not cleared after call to log_rv_undo_end_tdes. Or so it seems. Or, does it not even matter?

src/transaction/log_recovery.c
```diff
@@ -2837,19 +2812,8 @@ log_recovery_undo (THREAD_ENTRY * thread_p)
...
-			}
-		      else
-			{
-			  (void) log_complete (thread_p, tdes, LOG_ABORT, LOG_DONT_NEED_NEWTRID,
-					       LOG_NEED_TO_WRITE_EOT_LOG);
-			  logtb_free_tran_index (thread_p, tran_index);
-			  tdes = NULL;
-			}
+		      log_rv_undo_end_tdes (thread_p, tdes);
+		      tdes = NULL;
```
cristiarg, line 2816: in the old code, `tdes` is only assigned null on `log_complete` branch

src/transaction/log_recovery_analysis.cpp
```diff
@@ -1628,18 +1657,21 @@ static void
...
     {
-    case LOG_UNDOREDO_DATA:
-    case LOG_DIFF_UNDOREDO_DATA:
-    case LOG_UNDO_DATA:
-    case LOG_REDO_DATA:
     case LOG_MVCC_UNDOREDO_DATA:
     case LOG_MVCC_DIFF_UNDOREDO_DATA:
     case LOG_MVCC_UNDO_DATA:
     case LOG_MVCC_REDO_DATA:
+      is_mvcc = true;
```
cristiarg: I think it is easier to maintain with break and explicit call to `log_rv_analysis_undo_redo` with false/true as appropriate. In the end, no line of code is saved with fall-through.

src/transaction/log_recovery_analysis.cpp
```diff
@@ -1719,6 +1751,9 @@ log_rv_analysis_record_on_tran_server (THREAD_ENTRY *thread_p, LOG_RECTYPE log_t
       (void) log_rv_analysis_atomic_sysop_start (thread_p, tran_id, log_lsa);
       break;
 
+    case LOG_ASSIGNED_MVCCID:
+      break;
```
cristiarg, line 1755: call `log_rv_analysis_assigned_mvccid` or am I completely missing the point...

src/transaction/log_recovery.c
ribram: You're right, mvccid must be cleared. Fixed.

src/transaction/log_recovery.c
ribram: Right again, fixed.

src/transaction/log_recovery.c
ribram: It does not matter, tdes may become null in both cases.

src/transaction/log_recovery_analysis.cpp
ribram: You're right, fixed.

src/transaction/log_recovery_analysis.cpp
ribram: Good catch :). Fixed.

cristiarg (approved):

PR #3193, opened 2021-11-19
[LETS-254] Atomic replication for heap_update_bigone

adrianbrab:
http://jira.cubrid.org/browse/LETS-254

Added start/end atomic replication logs to `heap_update_bigone` to mark the zones that modify more than one page.

### DIFF
--- a/src/storage/heap_file.c
+++ b/src/storage/heap_file.c
@@ -21898,6 +21898,7 @@ heap_update_bigone (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * context, b
   bool is_old_home_updated;
   RECDES new_home_recdes;
   VFID ovf_vfid;
+  bool atomic_replication_flag = false;
 
   LOG_TDES *tdes = NULL;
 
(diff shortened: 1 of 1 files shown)


src/storage/heap_file.c
```diff
@@ -22098,6 +22100,7 @@ heap_update_bigone (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * context, b
   /* Fall through to exit. */
 
 exit:
+  log_append_empty_record (thread_p, LOG_END_ATOMIC_REPL, NULL);
```
ribram: End atomic replication log record should always follow a start atomic replication log record. The code now allows one case that logs the end atomic replication, but does not log the start atomic replication, see the first `goto exit`.

Fix by:
- Move the start atomic replication log record before the first `goto exit`.
- Add a boolean to track if start atomic replication log was appended; append the end atomic replication log record only if the boolean is true.

adrianbrab: added a boolean flag to be inline with the other implementations

ribram (approved):

PR #3196, opened 2021-11-23
[LETS-219] Transfer log header and log append page to PTS

cristiarg:
http://jira.cubrid.org/browse/LETS-219

Infrastructure to transfer the log header page, log append page and prev lsa from page server to passive transaction server.

Implementation:
- new request id's for `LOG_BOOT_INFO` request and response
- add a passive transaction server 'shadow pointer' - `pts_Gl` - alongside the polymorphic one - `ts_Gl` - that is to be used only in passive transaction server scenarios - pointer is not visible globally, it needs to be accessed via a guarded function
- new task used to call a function that packs the log boot info into a string in an asynchronous manner
- logic to pack together log header page, log append page and prev lsa on the page server and transfer and unpack it on the passive transaction server side
-  move ts_Gl pointer to server_type to group initialization/reset of object in a single place (for consistency, in the future, ps_Gl also needs to be moved there and also transformed into a unique ptr)

### DIFF
--- a/src/base/server_type.cpp
+++ b/src/base/server_type.cpp
@@ -246,8 +246,15 @@ bool is_tran_server_with_remote_storage ()
 
 passive_tran_server *get_passive_tran_server_ptr ()
 {
-  assert (pts_Gl != nullptr);
-  return pts_Gl;
+  if (is_passive_transaction_server())
+    {
+      assert (pts_Gl != nullptr);
+      return pts_Gl;
+    }
+  else
+    {
+      assert (is_passive_transaction_server());
+    }
 }
 
 #else // !SERVER_MODE = SA_MODE
--- a/src/base/server_type.cpp
+++ b/src/base/server_type.cpp
@@ -19,6 +19,7 @@
 #include "server_type.hpp"
 
 #include "active_tran_server.hpp"
+#include "passive_tran_server.hpp"
 #include "communication_server_channel.hpp"
 #include "connection_defs.h"
 #include "error_manager.h"
@@ -134,7 +135,9 @@ int init_server_type (const char *db_name)
 	}
       else if (is_passive_transaction_server ())
 	{
-	  ts_Gl.reset (new passive_tran_server ());
+	  passive_tran_server *const pts_ptr = new passive_tran_server ();
+	  ts_Gl.reset (pts_ptr);
+	  init_passive_tran_server_shadow_ptr (pts_ptr);
 	}
       else
 	{
@@ -184,6 +187,10 @@ void finalize_server_type ()
 {
   if (get_server_type () == SERVER_TYPE_TRANSACTION)
     {
+      if (is_passive_transaction_server ())
+	{
+	  reset_passive_tran_server_shadow_ptr ();
+	}
       ts_Gl->disconnect_page_server ();
       ts_Gl.reset (nullptr);
     }
--- a/src/base/server_type.cpp
+++ b/src/base/server_type.cpp
@@ -25,10 +25,13 @@
 #include "error_manager.h"
 #include "log_impl.h"
 #include "page_server.hpp"
-#include "passive_tran_server.hpp"
 #include "system_parameter.h"
 
 #include <string>
+
+std::unique_ptr<tran_server> ts_Gl;
+// non-owning "shadow" pointer of globally visible ts_Gl
+passive_tran_server *pts_Gl = nullptr;
 
 SERVER_TYPE get_value_from_config (server_type_config parameter_value);
 transaction_server_type get_value_from_config (transaction_server_type_config parameter_value);
@@ -135,9 +138,9 @@ int init_server_type (const char *db_name)
 	}
       else if (is_passive_transaction_server ())
 	{
-	  passive_tran_server *const pts_ptr = new passive_tran_server ();
-	  ts_Gl.reset (pts_ptr);
-	  init_passive_tran_server_shadow_ptr (pts_ptr);
+	  assert (pts_Gl == nullptr);
+	  pts_Gl = new passive_tran_server ();
+	  ts_Gl.reset (pts_Gl);
 	}
       else
 	{
@@ -189,7 +192,8 @@ void finalize_server_type ()
     {
       if (is_passive_transaction_server ())
 	{
-	  reset_passive_tran_server_shadow_ptr ();
+	  assert (pts_Gl != nullptr);
+	  pts_Gl = nullptr;
 	}
       ts_Gl->disconnect_page_server ();
       ts_Gl.reset (nullptr);
@@ -238,6 +242,12 @@ bool is_tran_server_with_remote_storage ()
       return ts_Gl->uses_remote_storage ();
     }
   return false;
+}
+
+passive_tran_server *get_passive_tran_server_ptr ()
+{
+  assert (pts_Gl != nullptr);
+  return pts_Gl;
 }
 
 #else // !SERVER_MODE = SA_MODE
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -22,6 +22,7 @@
 #include "error_manager.h"
 #include "log_impl.h"
 #include "log_lsa_utils.hpp"
+#include "log_manager.h"
 #include "log_prior_recv.hpp"
 #include "log_replication.hpp"
 #include "packer.hpp"
@@ -36,8 +37,6 @@
 
 page_server ps_Gl;
 
-static void assert_page_server_type ();
-
 page_server::~page_server ()
 {
   assert (m_replicator == nullptr);
@@ -48,13 +47,10 @@ page_server::connection_handler::connection_handler (cubcomm::channel &chn, page
 {
   m_conn.reset (new tran_server_conn_t (std::move (chn),
   {
+    // common
     {
       tran_to_page_request::GET_BOOT_INFO,
       std::bind (&page_server::connection_handler::receive_boot_info_request, std::ref (*this), std::placeholders::_1)
-    },
-    {
-      tran_to_page_request::SEND_LOG_PRIOR_LIST,
-      std::bind (&page_server::connection_handler::receive_log_prior_list, std::ref (*this), std::placeholders::_1)
     },
     {
       tran_to_page_request::SEND_LOG_PAGE_FETCH,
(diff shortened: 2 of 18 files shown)


src/base/server_type.cpp
```diff
@@ -134,7 +135,9 @@ int init_server_type (const char *db_name)
 	}
       else if (is_passive_transaction_server ())
 	{
-	  ts_Gl.reset (new passive_tran_server ());
+	  passive_tran_server *const pts_ptr = new passive_tran_server ();
+	  ts_Gl.reset (pts_ptr);
+	  init_passive_tran_server_shadow_ptr (pts_ptr);
```
cristiarg: A "shadow pointer" is also initialized locally in `passive_tran_server.cpp` to be used there in specific contexts. This is questionable design, but best I could arrive at.

src/server/passive_tran_server.cpp
```diff
@@ -35,3 +57,84 @@ passive_tran_server::on_boot ()
...
+    });
+  }
+
+  const char *message_buf = m_log_header_log_append_prev_lsa.c_str ();
+  const int log_page_size = db_log_page_size ();
+
+  // log header, copy and initialize header
+  assert (log_Gl.loghdr_pgptr != nullptr);
+  std::memcpy (reinterpret_cast<char *> (log_Gl.loghdr_pgptr), message_buf, log_page_size);
+  LOG_HEADER *const log_hdr = reinterpret_cast<LOG_HEADER *> (log_Gl.loghdr_pgptr->area);
```
cristiarg, line 97: Is this initialization complete?
This piece of code is present in log_page_buffer.c in a few places (`logpb_fetch_header_with_buffer`, `logpb_wfetch_header_from_file`, `logpb_fetch_header_from_file`) but never isolated such that it can be reused.

src/server/passive_tran_server.cpp
```diff
@@ -35,3 +57,84 @@ passive_tran_server::on_boot ()
...
+  log_Gl.hdr = *log_hdr;
+  message_buf += log_page_size;
+
+  // log append
+  assert (log_Gl.append.log_pgptr == nullptr);
+  /* fetch the append page; append pages are always new pages */
+  log_Gl.append.log_pgptr = logpb_create_page (thread_p, log_Gl.hdr.append_lsa.pageid);
```
cristiarg, line 104: `logpb_create_page`, internally, has logic to flush a log page in case the round robin mechanism wants to re-use/overwrite an existing buffer. Given the context, that part should never be hit. Best would be to add an argument to ensure no flush ever occurs.

src/storage/async_page_fetcher.hpp
```diff
@@ -27,6 +27,25 @@
...
 namespace cublog
 {
+  /* generic task to be used with to call a engine function taking one argument - the thread context
+   * and returning a single result that is then passed back via the callback
+   *  - call func: takes as single argument the thread context and returns value
+   *  - callback func: takes as single argument the value returned by the call func
+   *
+   * TODO: error handling
+   */
+  template <typename CALL_FUNC, typename CALLBACK_FUNC>
```
cristiarg: Questionable design, as it lacks the ability to relay error, should it ever occur on the called function.
It is better to change this to be more in line with the other existing tasks - ie: template-less.

src/transaction/log_manager.c
```diff
@@ -3348,6 +3349,38 @@ log_skip_logging_set_lsa (THREAD_ENTRY * thread_p, LOG_DATA_ADDR * addr)
   return;
 }
 
+/* log_pack_log_header_log_append_prev_lsa - pack together the log header, log append pages
+ *              and prev lsa for the purpose of sending them to the passive transaction server
+ *              as part of the log initialization sequence
+ */
+// *INDENT-OFF*
+std::string log_pack_log_header_log_append_prev_lsa (THREAD_ENTRY * thread_p)
```
cristiarg: No error handling here. There should be.

src/base/server_type.cpp
ribram: I think the design is fine. What I would change is to move all the global pointers, ts_Gl, ps_Gl and ats_Gl to server_type.cpp/hpp.

src/server/tran_page_requests.hpp
```diff
@@ -31,14 +31,21 @@ enum class tran_to_page_request
   SEND_LOG_PRIOR_LIST,
 
   // Passive only
+  SEND_LOG_HEADER_LOG_APPEND_PREV_LSA_FETCH,
```
ribram: The information required to boot the PTS log may be extended in the future. It is not feasible to maintain a name that includes all the transferred information.
```suggestion
  SEND_LOG_BOOT_INFO,
```

src/transaction/log_manager.c
ribram: No error handling seems to be needed (there are no errors).

I think this function should be moved to `cublog::async_page_fetcher`; later we'll have to rename the class to something else (e.g. `page_server_response_manager`?). Let's determine this later.

The other two existing `cublog::async_page_fetcher` functionalities expose a member function that creates a task with the required actions internally. The callback handles packing and sending the response on a connection. Let's follow the same design until we decide to do things differently.

src/server/tran_page_requests.hpp
cristiarg: Done

src/transaction/log_manager.c

src/server/page_server.cpp
```diff
@@ -283,31 +321,35 @@ void page_server::disconnect_all_tran_server ()
...
+bool
+page_server::is_passive_tran_server_connected () const
 {
   assert (is_page_server ());
 
   return !m_passive_tran_server_conn.empty ();
 }
 
-cublog::async_page_fetcher &page_server::get_page_fetcher ()
+cublog::async_page_fetcher *
```
ribram: Why change from reference to pointer?

src/server/passive_tran_server.cpp
ribram: I think it is fine.

src/server/passive_tran_server.cpp
ribram: No need to do it now.

src/server/passive_tran_server.cpp
```diff
@@ -35,3 +38,84 @@ passive_tran_server::on_boot ()
...
+
+  // prev lsa
+  std::memcpy (&log_Gl.append.prev_lsa, message_buf, sizeof (struct log_lsa));
+
+  // do not leave m_log_boot_info empty as a safeguard as this function is only supposed
+  // to be called once
+  m_log_boot_info = "not empty";
+}
+
+void send_and_receive_log_boot_info (THREAD_ENTRY *thread_p)
```
ribram: Use directly `get_passive_tran_server_ptr ()->send_and_receive_log_boot_info (thread_p)`.

If you think it is better to wrap it in a function, the wrapper function should be moved to server_type and renames to clarify it is on PTS.

src/storage/async_page_fetcher.hpp
```diff
@@ -27,18 +27,20 @@
 
 namespace cublog
 {
-
   class async_page_fetcher
   {
     public:
       using log_page_callback_type = std::function<void (const LOG_PAGE *, int)>;
       using data_page_callback_type = std::function<void (const FILEIO_PAGE *, int)>;
+      using log_boot_info_callback_type = std::function<void (std::string)>;
```
ribram: Pass by reference.
```suggestion
      using log_boot_info_callback_type = std::function<void (const std::string&)>;
```

src/transaction/log_manager.c
```diff
@@ -3348,6 +3349,38 @@ log_skip_logging_set_lsa (THREAD_ENTRY * thread_p, LOG_DATA_ADDR * addr)
   return;
 }
 
+/* log_pack_log_boot_info - pack together the log header, log append pages
+ *              and prev lsa for the purpose of sending them to the passive transaction server
+ *              as part of the log initialization sequence
+ */
+// *INDENT-OFF*
+std::string log_pack_log_boot_info (THREAD_ENTRY * thread_p)
```
ribram, line 3357: I would move this function to async_page_fetcher. `log_manager` is for generic log management functions. `log_pack_log_boot_info` is specifically used only on the page server, to serve a the booting of a new passive transaction server (possibly active transaction servers too in the future).

src/transaction/log_manager.c
```diff
@@ -3348,6 +3349,38 @@ log_skip_logging_set_lsa (THREAD_ENTRY * thread_p, LOG_DATA_ADDR * addr)
...
+
+  std::string packed_message;
+
+  // log header
+  assert (log_Gl.loghdr_pgptr != nullptr);
+  packed_message.append(reinterpret_cast<const char*> (log_Gl.loghdr_pgptr), log_page_size);
+
+  // log append
+  assert (log_Gl.append.log_pgptr != nullptr);
+  packed_message.append(reinterpret_cast<const char*> (log_Gl.append.log_pgptr), log_page_size);
```
```suggestion
  packed_message.append (reinterpret_cast<const char*> (log_Gl.append.log_pgptr), log_page_size);
```

src/transaction/log_manager.c
```diff
@@ -3348,6 +3349,38 @@ log_skip_logging_set_lsa (THREAD_ENTRY * thread_p, LOG_DATA_ADDR * addr)
...
+  scope_exit log_cs_exit_ftor ( [thread_p] { LOG_CS_EXIT (thread_p); } );
+  std::lock_guard<std::mutex> { log_Gl.prior_info.prior_lsa_mutex };
+
+  const int log_page_size = db_log_page_size ();
+
+  std::string packed_message;
+
+  // log header
+  assert (log_Gl.loghdr_pgptr != nullptr);
+  packed_message.append(reinterpret_cast<const char*> (log_Gl.loghdr_pgptr), log_page_size);
```
```suggestion
  packed_message.append (reinterpret_cast<const char*> (log_Gl.loghdr_pgptr), log_page_size);
```

src/transaction/log_manager.c
```diff
@@ -3348,6 +3349,38 @@ log_skip_logging_set_lsa (THREAD_ENTRY * thread_p, LOG_DATA_ADDR * addr)
...
+
+  // log append
+  assert (log_Gl.append.log_pgptr != nullptr);
+  packed_message.append(reinterpret_cast<const char*> (log_Gl.append.log_pgptr), log_page_size);
+
+  // prev lsa
+  packed_message.append (reinterpret_cast<const char *> (&log_Gl.append.prev_lsa), sizeof (struct log_lsa));
+
+  assert (packed_message.size () == (2 * log_page_size + sizeof (struct log_lsa)));
```
ribram: Safe-guard is not really useful here.
```suggestion
```

src/server/passive_tran_server.cpp
```diff
@@ -35,3 +38,84 @@ passive_tran_server::on_boot ()
...
+  // log append
+  assert (log_Gl.append.log_pgptr == nullptr);
+  /* fetch the append page; append pages are always new pages */
+  log_Gl.append.log_pgptr = logpb_create_page (thread_p, log_Gl.hdr.append_lsa.pageid);
+  std::memcpy (reinterpret_cast<char *> (log_Gl.append.log_pgptr), message_buf, log_page_size);
+  message_buf += log_page_size;
+
+  // prev lsa
+  std::memcpy (&log_Gl.append.prev_lsa, message_buf, sizeof (struct log_lsa));
+
```
ribram: Add a safe-guard here:
```suggestion
  std::memcpy (&log_Gl.append.prev_lsa, message_buf, sizeof (struct log_lsa));
  message_buf += sizeof (struct log_lsa);
  
  // Safe-guard that the message has been consumed.
  assert (message_buf == m_log_boot_info.c_str () + m_log_boot_info.size ());
```

src/server/page_server.cpp
cristiarg: Leftover from trying to lift `ps_Gl` from page_server.hpp to server_type.hpp. Which, by the way, can only be done if `ps_Gl` is also changed to a unique_ptr.
I'll revert this change for now.

src/server/passive_tran_server.cpp
cristiarg: Done

src/server/passive_tran_server.cpp
cristiarg: Done

src/storage/async_page_fetcher.hpp
cristiarg: Changed to std::string &&. Makes more sense as the value needs to cross thread stacks.

src/transaction/log_manager.c
cristiarg: It could be done. But it leads to a lot of link errors in unit tests. Another options is to move this to a separate `log_....c` file.

src/transaction/log_manager.c

src/transaction/log_manager.c
ribram: not fixed

src/transaction/log_manager.c
ribram: not fixed

src/transaction/log_manager.c
ribram: Let's leave it as is for now.

ribram (approved):
Please fix the small spacing issues and merge.

PR #3203, opened 2021-11-25
[LETS-219] Transfer log header and log append page to PTS

cristiarg:
http://jira.cubrid.org/browse/LETS-219

Initialize log on passive transaction server by transferring log header, log append page and prev LSA from page server.

Implementation:
- function log_initialize_passive_tran_server is called instead of log_initialize when on PTS
- log_initialize_passive_tran_server is a blocking call

Other:
- logpb_fill_or_reset_header_parameters: fix extracting parameters from log header

Tests:
- executed shell_debug suite (for the modifications regarding logpb_fill_or_reset_header_parameters)

### DIFF
--- a/src/storage/async_page_fetcher.cpp
+++ b/src/storage/async_page_fetcher.cpp
@@ -114,9 +114,18 @@ namespace cublog
 
   void log_boot_info_fetch_task::execute (context_type &context)
   {
-    std::string message = log_pack_log_boot_info (&context);
+    struct log_lsa append_lsa, prev_lsa;
+
+    std::string message = log_pack_log_boot_info (&context, append_lsa, prev_lsa);
 
     m_callback (std::move (message));
+
+    if (prm_get_bool_value (PRM_ID_ER_LOG_PRIOR_TRANSFER))
+      {
+	_er_log_debug (ARG_FILE_LINE,
+		       "Sent log boot info to passive tran server with prev_lsa = (%lld|%d), append_lsa = (%lld|%d)\n",
+		       LSA_AS_ARGS (&prev_lsa), LSA_AS_ARGS (&append_lsa));
+      }
   }
 
   async_page_fetcher::async_page_fetcher ()
--- a/src/storage/async_page_fetcher.cpp
+++ b/src/storage/async_page_fetcher.cpp
@@ -114,7 +114,7 @@ namespace cublog
 
   void log_boot_info_fetch_task::execute (context_type &context)
   {
-    struct log_lsa append_lsa, prev_lsa;
+    log_lsa append_lsa, prev_lsa;
 
     std::string message = log_pack_log_boot_info (&context, append_lsa, prev_lsa);
 
--- a/src/transaction/boot_sr.c
+++ b/src/transaction/boot_sr.c
@@ -2340,6 +2340,11 @@ boot_restart_server (THREAD_ENTRY * thread_p, bool print_restart, const char *db
   if (is_passive_transaction_server ())
     {
       log_initialize_passive_tran_server (thread_p);
+      /* NOTE - Scalability:
+       * see note regarding scalability below where this function is called in the context of
+       * !passive_transaction_server
+       */
+      pgbuf_highest_evicted_lsa_init ();
     }
 #endif /* SERVER_MODE */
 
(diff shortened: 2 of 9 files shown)


src/storage/async_page_fetcher.cpp
```diff
@@ -114,9 +114,18 @@ namespace cublog
 
   void log_boot_info_fetch_task::execute (context_type &context)
   {
-    std::string message = log_pack_log_boot_info (&context);
+    struct log_lsa append_lsa, prev_lsa;
```
ribram: Having to mention struct before a type was a C thing. It is no longer required.
```suggestion
    log_lsa append_lsa, prev_lsa;
```

src/transaction/log_manager.c
```diff
@@ -3362,26 +3427,27 @@ log_skip_logging_set_lsa (THREAD_ENTRY * thread_p, LOG_DATA_ADDR * addr)
...
   std::lock_guard<std::mutex> { log_Gl.prior_info.prior_lsa_mutex };
 
-  const int log_page_size = db_log_page_size ();
-
   std::string packed_message;
 
   // log header
-  assert (log_Gl.loghdr_pgptr != nullptr);
-  packed_message.append (reinterpret_cast<const char*> (log_Gl.loghdr_pgptr), log_page_size);
+  packed_message.append( reinterpret_cast<const char*> (&log_Gl.hdr), sizeof (struct log_header));
```
```suggestion
  packed_message.append (reinterpret_cast<const char*> (&log_Gl.hdr), sizeof (log_header));
```

src/transaction/log_page_buffer.c
```diff
@@ -2549,6 +2552,32 @@ logpb_write_page_to_disk (THREAD_ENTRY * thread_p, LOG_PAGE * log_pgptr, LOG_PAG
   return NO_ERROR;
 }
 
+/*
+ * logpb_fill_or_reset_header_parameters - initialize values from header, or reset to invalid values if no header
+ *                    argument is provided
+ */
+void
+logpb_fill_or_reset_header_parameters (const struct log_header *hdr, PGLENGTH & io_page_size, PGLENGTH & log_page_size,
```
ribram: You could go with two different functions: logpb_fill_header_parameters and logpb_reset_header_parameters.

src/transaction/boot_sr.c
```diff
@@ -2334,6 +2334,15 @@ boot_restart_server (THREAD_ENTRY * thread_p, bool print_restart, const char *db
       goto error;
     }
 
+#if defined(SERVER_MODE)
+  /* for passive transaction server, log initialization occurs earlier on because it does not rely
+   * on any type of storage; information is manually/explicitly retrieved from page server */
+  if (is_passive_transaction_server ())
+    {
+      log_initialize_passive_tran_server (thread_p);
```
ribram, line 2342: On PTS, page buffer highest evicted LSA must be set before starting to read pages.
```suggestion
      log_initialize_passive_tran_server (thread_p);
      pgbuf_highest_evicted_lsa_init ();
```

src/storage/async_page_fetcher.cpp
cristiarg: Done

src/transaction/log_page_buffer.c
cristiarg: Done

src/transaction/boot_sr.c
cristiarg: It is already done sometime later? Or, for PTS, must the call be moved earlier, just after log initialize?

ribram: Actually, for both type of transaction servers it should be done before fetching data pages from page server, and after log is initialized. The current active transaction server booting sequence does not allow it, because pages are fetched before the log is initialized. If page server replication remains behind, it could be a problem; fortunately, it is most likely that the page server replication finishes up by the time ATS is restarted.

On PTS however it is mandatory to make sure that the PS replication is not behind the PTS replication while fetching data pages. Therefore, having the highest evicted LSA, used as target for PS replication, is more important.

Later on, the booting sequence of ATS could be changed to become similar to the PTS.

cristiarg: I added the call for evicted LSA initialization on possive transaction server and also guarded the existing call for !passive_tran_server.

ribram (approved):

PR #3205, opened 2021-12-01
[CBRD-24121] Avoid infinite wait on recursive Java SP calls draft

hgryoo:
http://jira.cubrid.org/browse/CBRD-24121

It is draft PR. Because of the deadlock issue described in JIRA, I've added `method_mode` flag in the connection entry (`CSS_CONN_ENTRY`). And the connection handler pushes the task by `push_external_task ()`. It avoids the nested task entering the waiting list.
The same problem can still occur when nesting calls a lot. Currently, as a draft, it uses server request worker pool (`css_Server_request_worker_pool`) by `push_external_task ()`. **This will replace it with a new worker pool only for method (SP call).**

Also, In transaction_cl.c, tran_libcas is managed by the depth of nesting. `method_callback` class manages connection info, (m_conn_info [METHOD_MAX_RECURSION_DEPTH]). With this structure, `method_callback` can know at which depth of the nesting structure of the SP it is executing and returns the result to the appropriate task.


src/transaction/transaction_cl.c
```diff
@@ -1325,6 +1327,7 @@ void
 tran_end_libcas_function (void)
 {
   tm_Is_libcas = false;
+  tm_libcas_depth--;
```
beyondykk9, line 1327: isn't it necessary to check "tm_libcas_depth > 0" ?
```
if (tran_is_in_libcas)
   {
     tm_libcas_depth--;
   }
```

hgryoo: tm_libcas_depth is 0 if CAS is not in method mode.

hgryoo: I've added assert at 879013d to check it.

beyondykk9 (approved):

shparkcubrid (approved):

PR #3207, opened 2021-12-02
[LETS-243] The transaction host is missing from logs

adrianbrab:
http://jira.cubrid.org/browse/LETS-243

Changed the connection protocol to always send the hostname when establishing a new connection. 
Also added a guard to prevent printing of invalid port connections.

### DIFF
--- a/src/communication/communication_server_channel.hpp
+++ b/src/communication/communication_server_channel.hpp
@@ -40,6 +40,7 @@ namespace cubcomm
   {
     public:
       server_channel (const char *server_name, SERVER_TYPE server_type, int max_timeout_in_ms = -1);
+      server_channel (int max_timeout_in_ms = -1);
       ~server_channel () = default;
 
       server_channel (const server_channel &) = delete;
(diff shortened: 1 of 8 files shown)


src/communication/communication_server_channel.hpp
```diff
@@ -49,10 +50,13 @@ namespace cubcomm
       server_channel &operator= (server_channel &&comm);
 
       css_error_code connect (const char *hostname, int port, css_command_type cmd_type);
+      css_error_code accept (SOCKET socket);
+      cubcomm::server_server get_conn_type ();
 
     private:
       std::string m_server_name;
       SERVER_TYPE m_server_type;
+      cubcomm::server_server m_request;
```
ribram:
```suggestion
      cubcomm::server_server m_conn_type;
```

src/server/page_server.cpp
```diff
@@ -242,8 +242,11 @@ page_server::set_active_tran_server_connection (cubcomm::channel &&chn)
   assert (is_page_server ());
 
   chn.set_channel_name ("ATS_PS_comm");
-  er_log_debug (ARG_FILE_LINE, "Active transaction server connected to this page server. Channel id: %s.\n",
-		chn.get_channel_id ().c_str ());
+  if (chn.get_port() != -1)
+    {
+      er_log_debug (ARG_FILE_LINE, "Active transaction server connected to this page server. Channel id: %s.\n",
```
ribram: The issue comment says to not include port in `chn.get_channel_id ()` if its value is -1. These message should be logged nonetheless.

ribram (approved):

PR #3219, opened 2021-12-03
[LETS-297] cubrid hb stop db_name doesn't process

adrianbrab:
http://jira.cubrid.org/browse/LETS-297

The `cubrid hb server stop command` did not actually close the server as it was looking for only Transaction and Page servers, due to the single node stop configuration.

### DIFF
--- a/src/executables/commdb.c
+++ b/src/executables/commdb.c
@@ -486,7 +486,7 @@ process_ha_server_mode (CSS_CONN_ENTRY * conn, char *server_name)
 }
 
 static int
-search_server_pid (char *server_info, char *search_pattern)
+search_server_pid (const char *server_info, const char *search_pattern)
 {
   char *p = NULL;
   int pid = 0;
@@ -539,6 +539,10 @@ process_server_info_pid (CSS_CONN_ENTRY * conn, const char *server, int server_t
 	      break;
 
 	    case SERVER_TYPE_TRANSACTION:
+	      // Transaction servers are printed in two ways:
+	      //    - as "Transaction-Server" if HA is disabled
+	      //    - as "HA-Server" if HA is enabled
+	      // Search by both patterns.
 	      sprintf (search_pattern, "Transaction-Server %s (", server);
 	      pid = search_server_pid (server_info, search_pattern);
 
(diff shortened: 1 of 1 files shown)


src/executables/commdb.c
```diff
@@ -485,6 +485,26 @@ process_ha_server_mode (CSS_CONN_ENTRY * conn, char *server_name)
     }
 }
 
+static int
+search_server_pid (char *server_info, char *search_pattern)
```
ribram:
```suggestion
search_server_pid (const char *server_info, const char *search_pattern)
```

src/executables/commdb.c
```diff
@@ -515,10 +535,18 @@ process_server_info_pid (CSS_CONN_ENTRY * conn, const char *server, int server_t
 	    {
 	    case SERVER_TYPE_PAGE:
 	      sprintf (search_pattern, "Page-Server %s (", server);
+	      pid = search_server_pid (server_info, search_pattern);
 	      break;
 
 	    case SERVER_TYPE_TRANSACTION:
 	      sprintf (search_pattern, "Transaction-Server %s (", server);
```
ribram, line 546:
```suggestion
	      // Transaction servers are printed in two ways:
	      //    - as "Transaction-Server" if HA is disabled
	      //    - as "HA-Server" if HA is enabled
	      // Search by both patterns.

	      sprintf (search_pattern, "Transaction-Server %s (", server);
```

src/executables/commdb.c
```diff
@@ -515,10 +535,18 @@ process_server_info_pid (CSS_CONN_ENTRY * conn, const char *server, int server_t
...
+	      pid = search_server_pid (server_info, search_pattern);
+
+	      if (pid == 0)
+		{
+		  sprintf (search_pattern, "HA-Server %s (", server);
+		  pid = search_server_pid (server_info, search_pattern);
+		}
 	      break;
 
 	    default:
```
ribram, line 556: `search_server_pid` should be called on default case too.

ribram (approved):

src/executables/commdb.c
```diff
@@ -485,6 +485,26 @@ process_ha_server_mode (CSS_CONN_ENTRY * conn, char *server_name)
     }
 }
 
+static int
+search_server_pid (const char *server_info, const char *search_pattern)
+{
+  char *p = NULL;
```
ribram: Build fails. Change the type of p to fix.
```suggestion
  const char *p = NULL;
```

ribram (approved):

PR #3224, opened 2021-12-06
[LETS-326] Extend requests_sync_client_server payload with the response sequence number

ribram:
http://jira.cubrid.org/browse/LETS-326

The purpose of this pull request is to adapt the request messages and request handlers to having a response sequence number in the message.

The message is decomposed into three parts:

  - The message ID, unpacked by `request_server` that determines which handler processes the request.
  - The request sequence number, unpacked by `request_sync_client_server` that determines the request-response pairing.
  - The actual user payload, that is unpacked inside the request handler.
 
The patch includes the next changes:

  - `request_sync_client_server` maintains an `internal_payload` - nested class - that pairs the user payload with a response sequence number. The response sequence number is for now always `NO_RESPONSE`.
  - `request_sync_client_server` wraps the `request_client_server` handler with its own function `unpack_and_handle`. The handlers provided to `request_sync_client_server` must be given an `internal_payload` instead of a `cubpacking::unpacker`.
  - `request_sync_client_server::push` automatically pairs the request payload with a `NO_RESPONSE` sequence number.
  - Update all handlers in transaction and page server to the new signature.
  - Updated unit tests to the new types of handlers and this patch changes.

### DIFF
--- a/src/communication/request_sync_client_server.hpp
+++ b/src/communication/request_sync_client_server.hpp
@@ -59,6 +59,27 @@ namespace cubcomm
       void push (T_OUTGOING_MSG_ID a_outgoing_message_id, T_PAYLOAD &&a_payload);
 
     private:
+
+      // Wrap the user payload with a response sequence number
+      using response_sequence_number_t = std::uint64_t;
+      static constexpr response_sequence_number_t NO_RESPONSE =
+	      std::numeric_limits<response_sequence_number_t>::max ();
+
+      class internal_payload : cubpacking::packable_object
+      {
+	public:
+	  void push_payload (T_PAYLOAD &&a_payload);
+	  T_PAYLOAD pull_payload ();
+
+	  size_t get_packed_size (cubpacking::packer &serializator, std::size_t start_offset = 0) const final override;
+	  void pack (cubpacking::packer &serializator) const final override;
+	  void unpack (cubpacking::unpacker &deserializator) final override;
+
+	private:
+	  response_sequence_number_t m_rsn = NO_RESPONSE;
+	  T_PAYLOAD m_user_payload;
+      };
+
       using request_sync_send_queue_t = cubcomm::request_sync_send_queue<request_client_server_t, T_PAYLOAD>;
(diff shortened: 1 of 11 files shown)


src/communication/request_sync_client_server.hpp
```diff
@@ -29,15 +29,55 @@ namespace cubcomm
...
+
     public:
       using outgoing_msg_id_t = T_OUTGOING_MSG_ID;
-
       using request_client_server_t = cubcomm::request_client_server<T_OUTGOING_MSG_ID, T_INCOMING_MSG_ID>;
-      using incoming_request_handler_t = typename request_client_server_t::server_request_handler;
+
+      // The user payload (of type T_PAYLOAD) is accompanied by a response sequence number set by
+      // request_sync_client_server.
+      class internal_payload : public cubpacking::packable_object
```
cristiarg, line 56: More convenient to have this as a non-nested template class - `request[_sync]_payload` - that will have to be explicitly template-instantiated "by hand".

src/communication/request_sync_client_server.hpp
```diff
@@ -29,15 +29,55 @@ namespace cubcomm
...
+	  T_PAYLOAD pull_payload ();
+
+	  size_t get_packed_size (cubpacking::packer &serializator, std::size_t start_offset = 0) const final override;
+	  void pack (cubpacking::packer &serializator) const final override;
+	  void unpack (cubpacking::unpacker &deserializator) final override;
+
+	private:
+	  response_sequence_number_t m_rsn = NO_RESPONSE;
+	  T_PAYLOAD m_user_payload;
+      };
```
cristiarg, line 78: Somehow this feels a bit anachronic. On one side it is an "internal_payload". On the other side the client still has to deal with it.
Why not unify "internal_payload" here with `request_sync_send_queue::queue_item_type`, integrate it full stack and expose to the client, with the only drawback that the client has to ignore the sequence number but still has to populate the id and the net-payload, while the infrastructure populates the sequence number.
I do not have a better understanding as of now, so it might evolve over time when we get a better hold of the design & usage.

cristiarg (approved):

PR #3229, opened 2021-12-07
[LETS-325] Add classes for request/response dispatching

ribram:
http://jira.cubrid.org/browse/LETS-325

Added two new classes into a single header file:

**cubcomm::response_sequence_number_generator**

Generates unique sequence number for responses.

**cubcomm::response_broker**

- Register responses
- Get responses

Getting a response is blocked until the response is received. When a response is registered, requesters are notified.

The response broker organizes responses into buckets. It must be initialized with a positive number of buckets. Response registering/getting is forwarded to the appropriate bucket.

**Unit test**

Add two new cases in _test_request_cs_:

- Test for `response_sequence_number_generator` that it generates unique numbers to concurrent requesters.
- Test for `response_broker` that it matches correctly the registered responses to concurrent requesters.


unit_tests/request_client_server/test_main.cpp
```diff
@@ -532,6 +533,122 @@ TEST_CASE ("Two request_sync_client_server communicate with each other", "[dbg]"
...
+
+  for (auto &ti : threads_info)
+    {
+      numbers_t &numbers = ti.second;
+      numbers.reserve (NUMBER_COUNT);
+      ti.first = std::thread ([&numbers, &rsn_gen] ()
+      {
+	numbers.push_back (rsn_gen.get_unique_number ());
+      }
+			     );
```
cristiarg: alignment, but nothing can be done about it

cristiarg (approved):

unit_tests/request_client_server/test_main.cpp
ribram: fixed

PR #3230, opened 2021-12-07
[LETS-223] Replicate changes on PTS on a single thread

cristiarg:
http://jira.cubrid.org/browse/LETS-223

Passive transaction server must replicate ATS changes on its page buffer pages to avoid fetching pages again from PS for the updated versions.
If PS only has to replicate data updates on page level, PTS has to make consistent reads concurrently with the replication.

To simplify how replication works, it will be limited to a single thread. The scope of this patch is to allow consistent queries only after the replication is completed. Consistent queries concurrent with replication will be implemented in future tasks.

Implementation:
- initialize log flush daemon on PTS to process log prior list received from PS
- start replication during passive transaction server log initialization; do this before the PTS actually starts consuming the log prior received from the PS; PTS replication is currently explicitly single threaded
- adapt to use the new page fetch mode OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT on PTS
- adapt to skip applying the log record if: page is either not returned by the attempt to fix it with the new page fetch mode; or it is returned but its LSA is greater than the current log record LSA (meaning the page retrieved from PS already contains the modifications applied by the PS replication)
- also skip applying log records for btree statistics log records
- track the highest LSA applied by replication and set it as target LSA for the data pages fetched from PS
- parameterize cublog::replicator to allow tuning it differently for PS and PTS
- gracefully shutdown replication before finalizing log on passive transaction server

### DIFF
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -390,7 +390,7 @@ page_server::start_log_replicator (const log_lsa &start_lsa)
 
   const int replication_parallel_count = prm_get_integer_value (PRM_ID_REPLICATION_PARALLEL_COUNT);
   assert (replication_parallel_count >= 0);
-  m_replicator.reset (new cublog::replicator (start_lsa, replication_parallel_count));
+  m_replicator.reset (new cublog::replicator (start_lsa, RECOVERY_PAGE, replication_parallel_count));
 }
 
 void
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -359,7 +359,7 @@ page_server::push_request_to_active_tran_server (page_to_tran_request reqid, std
 cublog::replicator &
 page_server::get_replicator ()
 {
-  assert (m_replicator);
+  assert (m_replicator != nullptr);
   return *m_replicator;
 }
 
@@ -369,7 +369,9 @@ page_server::start_log_replicator (const log_lsa &start_lsa)
   assert (is_page_server ());
   assert (m_replicator == nullptr);
 
-  m_replicator.reset (new cublog::replicator (start_lsa));
+  const int replication_parallel_count = prm_get_integer_value (PRM_ID_REPLICATION_PARALLEL_COUNT);
+  assert (replication_parallel_count >= 0);
+  m_replicator.reset (new cublog::replicator (start_lsa, replication_parallel_count));
 }
 
 void
--- a/src/server/passive_tran_server.cpp
+++ b/src/server/passive_tran_server.cpp
@@ -150,3 +150,7 @@ void passive_tran_server::start_log_replicator (const log_lsa &start_lsa)
   m_replicator.reset (new cublog::replicator (start_lsa, OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT, 0));
 }
 
+log_lsa passive_tran_server::get_replicator_lsa () const
+{
+  return m_replicator->get_redo_lsa ();
+}
--- a/src/server/passive_tran_server.cpp
+++ b/src/server/passive_tran_server.cpp
@@ -154,3 +154,11 @@ log_lsa passive_tran_server::get_replicator_lsa () const
 {
   return m_replicator->get_redo_lsa ();
 }
+
+void passive_tran_server::finish_replication_during_shutdown (cubthread::entry &thread_entry)
+{
+  assert (m_replicator != nullptr);
+
+  m_replicator->wait_replication_finish_during_shutdown ();
+  m_replicator.reset (nullptr);
+}
--- a/src/server/passive_tran_server.cpp
+++ b/src/server/passive_tran_server.cpp
@@ -147,6 +147,6 @@ void passive_tran_server::start_log_replicator (const log_lsa &start_lsa)
 
   // passive transaction server executes replication synchronously, for the time being, due to complexity of
   // executing it in parallel while also providing a consistent view of the data
-  m_replicator.reset (new cublog::replicator (start_lsa, 0));
+  m_replicator.reset (new cublog::replicator (start_lsa, OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT, 0));
 }
 
--- a/src/server/passive_tran_server.cpp
+++ b/src/server/passive_tran_server.cpp
@@ -22,6 +22,11 @@
 #include "system_parameter.h"
 #include "thread_manager.hpp"
 
+passive_tran_server::~passive_tran_server ()
+{
+  assert (m_replicator == nullptr);
+}
+
 bool
 passive_tran_server::uses_remote_storage () const
 {
@@ -125,6 +130,9 @@ void passive_tran_server::send_and_receive_log_boot_info (THREAD_ENTRY *thread_p
   // to be called once
   m_log_boot_info = "not empty";
 
+  // at this point, page server has already started log prior dispatch towards this passive transaction server;
+  // for now, log prior consumption is held back by the log prior lsa mutex held while calling this function
+
   if (prm_get_bool_value (PRM_ID_ER_LOG_PRIOR_TRANSFER))
     {
       _er_log_debug (ARG_FILE_LINE,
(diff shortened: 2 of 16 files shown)


src/server/page_server.cpp
```diff
@@ -388,7 +388,9 @@ page_server::start_log_replicator (const log_lsa &start_lsa)
   assert (is_page_server ());
   assert (m_replicator == nullptr);
 
-  m_replicator.reset (new cublog::replicator (start_lsa));
+  const int replication_parallel_count = prm_get_integer_value (PRM_ID_REPLICATION_PARALLEL_COUNT);
+  assert (replication_parallel_count >= 0);
+  m_replicator.reset (new cublog::replicator (start_lsa, RECOVERY_PAGE, replication_parallel_count));
```
cristiarg, line 389: parameterized replicator to be able to instrument it differently for PS and PTS

src/server/passive_tran_server.cpp
```diff
@@ -22,6 +22,11 @@
 #include "system_parameter.h"
 #include "thread_manager.hpp"
 
+passive_tran_server::~passive_tran_server ()
+{
+  assert (m_replicator == nullptr);
```
cristiarg, line 27: I think this assert does not currently hold

src/storage/page_buffer.c
```diff
@@ -8017,7 +8016,23 @@ pgbuf_read_page_from_file_or_page_server (THREAD_ENTRY * thread_p, const VPID *
...
+
+      /* on the passive transaction server, make sure to request that the progress on the page server - from
+       * where the data page is retrieved -  be at least the same as the local progress to avoid losing
+       * any log entries being applied on both ends
+       */
+      const LOG_LSA target_repl_lsa =
+	(is_passive_transaction_server ())? replicator_lsa_ftor () : pgbuf_Pool.get_highest_evicted_lsa ();
+
+      pgbuf_request_data_page_from_page_server (vpid, target_repl_lsa);
+      // TODO: what if, by the next line, page has already been retrieved by the page broker
```
cristiarg: what if?

src/server/page_server.cpp
ribram: Many differences may arise between PTS and PS, other than the page fetch mode. I think it's better to setup a replication type:
```c
enum replication_type
{
  PAGE_SERVER_REPLICATION,
  PASSIVE_TRAN_SERVER_REPLICATION,
};
```
Page fetch mode can be a replicator member variable determined by the replication type.

An alternative is to use `is_passive_tran_server()` / `is_page_server()` and ditch replication type enum.

src/server/passive_tran_server.cpp
ribram: Is the assert needed?

It seems to me that replicator can just be stopped; since its job is just to keep PTS up to date while running, it loses relevance during shutdown. I may just be stopped whenever.

src/storage/page_buffer.c
```diff
@@ -8017,7 +8016,23 @@ pgbuf_read_page_from_file_or_page_server (THREAD_ENTRY * thread_p, const VPID *
...
+	return pts_ptr->get_replicator_lsa ();
+      };
+      // *INDENT-ON*
+
+      /* on the passive transaction server, make sure to request that the progress on the page server - from
+       * where the data page is retrieved -  be at least the same as the local progress to avoid losing
+       * any log entries being applied on both ends
+       */
+      const LOG_LSA target_repl_lsa =
+	(is_passive_transaction_server ())? replicator_lsa_ftor () : pgbuf_Pool.get_highest_evicted_lsa ();
```
ribram: - The lambda is overkill in this case. The code is easier to read without it.
- Complete the comment with active transaction server case.
```suggestion
      /* on the passive transaction server, make sure to request that the progress on the page server - from
       * where the data page is retrieved -  be at least the same as the local progress to avoid losing
       * any log entries being applied on both ends
       *
       * on the active transaction server, use highest evicted LSA to make sure no changes are missed by the PS.
       */
      const LOG_LSA target_repl_lsa =
	is_passive_transaction_server () ? get_passive_tran_server_ptr ()->get_replicator_lsa () : pgbuf_Pool.get_highest_evicted_lsa ();
```

src/storage/page_buffer.c
ribram: Page will be found immediately. See page_broker<PageT>::wait_for_page.
```suggestion
```

src/transaction/boot_sr.c
```diff
@@ -3189,6 +3189,12 @@ xboot_shutdown_server (REFPTR (THREAD_ENTRY, thread_p), ER_FINAL_CODE is_er_fina
     }
   else if (get_server_type () == SERVER_TYPE_TRANSACTION)
     {
+      // NOTE: passive transaction server, regarding replication: since
```
ribram: The note is right. But I don't think it is justified to add comments for all the things we don't have to do. Please remove it.

src/transaction/log_manager.c
```diff
@@ -1648,6 +1648,11 @@ log_initialize_passive_tran_server (THREAD_ENTRY * thread_p)
 
     LOG_RESET_APPEND_LSA (&log_Gl.hdr.append_lsa);
     LOG_RESET_PREV_LSA (&log_Gl.append.prev_lsa);
+
+    // while still holding prior LSA lock, initialize passive transaction server replication
+    // without losing any log record
+    const log_lsa next_io_lsa = log_Gl.append.get_nxio_lsa ();
+    pts_ptr->start_log_replicator (next_io_lsa);
```
ribram: While next_io_lsa must be read while holding the mutex, replicator may be started after releasing the lock. Move replicator start outside this scope.

src/transaction/log_manager.c
```diff
@@ -10756,12 +10763,19 @@ log_flush_daemon_init ()
...
+    }
+  else
+    {
+      log_remove_log_archive_daemon_init ();
+      log_checkpoint_daemon_init ();
+      log_checkpoint_trantable_daemon_init ();
+      log_check_ha_delay_info_daemon_init ();
+      log_clock_daemon_init ();
+      log_flush_daemon_init ();
+    }
```
ribram: Nitpicking:
```suggestion
  if (!is_passive_transaction_server ())
    {
      log_remove_log_archive_daemon_init ();
      log_checkpoint_daemon_init ();
      log_checkpoint_trantable_daemon_init ();
      log_check_ha_delay_info_daemon_init ();
      log_clock_daemon_init ();
    }
  log_flush_daemon_init ();
```

src/server/page_server.cpp
cristiarg: Another way to look at it is to keep characteristics independent of whether replication is running in page server or passive transaction server context. Because some of these parameters (eg: page fetch type) will be forwarded down to the recovery functions on which replication is built on.
Of course that we could fork these params inside the ctor, or pass them from the outside.

src/server/passive_tran_server.cpp
cristiarg: Added code to stop the replication upon server shutdown. Left assert in place.

src/transaction/boot_sr.c
cristiarg: Actually replicator must be gracefully terminated before log is terminated. Otherwise, log termination will reset `nxio_lsa` which will break assert in the still running replication.

src/transaction/log_manager.c

src/transaction/log_manager.c

src/storage/page_buffer.c

src/server/passive_tran_server.hpp
```diff
@@ -27,9 +28,16 @@ class passive_tran_server : public tran_server
...
+    ~passive_tran_server () override;
 
   public:
     void send_and_receive_log_boot_info (THREAD_ENTRY *thread_p);
+    void start_log_replicator (const log_lsa &start_lsa);
+
+  public:
+    /* read replicator's current progress */
+    log_lsa get_replicator_lsa () const;
+    void finish_replication_during_shutdown (cubthread::entry &thread_entry);
```
ribram: No reason for having separated public sections here:
```suggestion
  public:
    void send_and_receive_log_boot_info (THREAD_ENTRY *thread_p);
    void start_log_replicator (const log_lsa &start_lsa);
    /* read replicator's current progress */
    log_lsa get_replicator_lsa () const;
    void finish_replication_during_shutdown (cubthread::entry &thread_entry);
```

src/transaction/log_manager.c
```diff
@@ -1634,6 +1634,12 @@ log_initialize_passive_tran_server (THREAD_ENTRY * thread_p)
       return;
     }
 
+  passive_tran_server *const pts_ptr = get_passive_tran_server_ptr ();
+  assert (pts_ptr != nullptr);
+
+  log_lsa replication_start_redo_lsa
+  {
+  NULL_LSA};
```
ribram: Alternative to fix indentation:
```suggestion
  log_lsa replication_start_redo_lsa = NULL_LSA;
```

ribram (approved):
I proposed some minor changes. Please do as you see fit and merge :)

PR #3233, opened 2021-12-09
[CBRD-23960] Support returning Cursor to client from stored function

hgryoo:

- removed legacy code related to OutResultset
- changed DB_TYPE_RESULTSET value to int64_t (DB_BIGINT)
- add routine to create SRV_HANDLE from query result in method_callback

please refer to the demo code, TestResultSet.java (Application) and SpCursorTest.java (Stored Function)


shparkcubrid (approved):

src/method/method_callback.cpp
```diff
@@ -195,6 +204,14 @@ namespace cubmethod
 
     execute_info info = handler->execute (request);
 
+    /* register query_id for out resultset */
+    cubmethod::query_result *qresult = handler->get_current_result();
+    if (qresult->stmt_type == CUBRID_STMT_SELECT)
+      {
+	uint64_t qid = (uint64_t) info.qresult_infos[0].query_id;
```
beyondykk9, line 211: In info.qresult_infos[0], 0 means that it's not supported to handle the multi-statements ?

hgryoo: yes, it assumes that a statement can have only 1 result

youngjinj (approved):

beyondykk9 (approved):

hyunikn (approved):

PR #3236, opened 2021-12-09
[CBRD-24123] Provides hidden system parameter to set the first log page id for QA or testing purposes.

hornetmj:
http://jira.cubrid.org/browse/CBRD-24123

### Purpose
- The CBRD-24123 issue fixed a problem that occurred when the log_pageid exceeds INT_MAX.
- To verify this, more than INT_MAX log pages must be created, but this is a very inefficient method.
- Therefore, a hidden system parameter that can designate the first log page id when DB is created is provided so that the test can be performed smoothly.
- If you have a better way or idea than this, please let me know.

### Implementation
- Provides hidden system parameter "first_log_pageid".
- Bypassing the assert part that occurs when createdb and server are started. 

### Remarks
- Check the "First_active_log_page" with the show log header command.

```
csql> ;line on
csql> show log header;

...
        First_active_log_page          : 2147483649
...

```

### DIFF
--- a/src/transaction/log_page_buffer.c
+++ b/src/transaction/log_page_buffer.c
@@ -1343,7 +1343,11 @@ logpb_initialize_header (THREAD_ENTRY * thread_p, LOG_HEADER * loghdr, const cha
   loghdr->avg_nlocks = LOG_ESTIMATE_NOBJ_LOCKS;
   loghdr->npages = npages - 1;	/* Hdr pg is stolen */
   loghdr->db_charset = lang_charset ();
+#if !defined(NDEBUG)
   loghdr->fpageid = (LOG_PAGEID) prm_get_bigint_value (PRM_ID_FIRST_LOG_PAGEID);	/* loghdr->fpageid should always be 0 except for QA or TEST purposes. */
+#else
+  loghdr->fpageid = 0;
+#endif
   loghdr->append_lsa.pageid = loghdr->fpageid;
   loghdr->append_lsa.offset = 0;
   LSA_COPY (&loghdr->chkpt_lsa, &loghdr->append_lsa);
@@ -2482,15 +2486,8 @@ logpb_fetch_start_append_page (THREAD_ENTRY * thread_p)
 
   logpb_log ("started logpb_fetch_start_append_page\n");
 
-  if (prm_get_bigint_value (PRM_ID_FIRST_LOG_PAGEID) != 0
-      && (LOG_PAGEID) prm_get_bigint_value (PRM_ID_FIRST_LOG_PAGEID) == log_Gl.hdr.append_lsa.pageid
-      && log_Gl.hdr.append_lsa.offset == 0)
-    {
-      flag = NEW_PAGE;
-    }
-
   /* detect empty log (page and offset of zero) */
-  if ((log_Gl.hdr.append_lsa.pageid == 0) && (log_Gl.hdr.append_lsa.offset == 0))
+  if ((log_Gl.hdr.append_lsa.pageid == log_Gl.hdr.fpageid) && (log_Gl.hdr.append_lsa.offset == 0))
     {
       flag = NEW_PAGE;
     }
--- a/src/transaction/log_page_buffer.c
+++ b/src/transaction/log_page_buffer.c
@@ -1343,7 +1343,14 @@ logpb_initialize_header (THREAD_ENTRY * thread_p, LOG_HEADER * loghdr, const cha
   loghdr->avg_nlocks = LOG_ESTIMATE_NOBJ_LOCKS;
   loghdr->npages = npages - 1;	/* Hdr pg is stolen */
   loghdr->db_charset = lang_charset ();
-  loghdr->fpageid = 0;
+  if (prm_get_bigint_value (PRM_ID_FIRST_LOG_PAGEID) != 0)
+    {
+      loghdr->fpageid = (LOG_PAGEID) prm_get_bigint_value (PRM_ID_FIRST_LOG_PAGEID);
+    }
+  else
+    {
+      loghdr->fpageid = 0;
+    }
   loghdr->append_lsa.pageid = loghdr->fpageid;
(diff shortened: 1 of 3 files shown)


mhoh3963 (approved):

src/transaction/log_page_buffer.c
```diff
@@ -1343,7 +1343,14 @@ logpb_initialize_header (THREAD_ENTRY * thread_p, LOG_HEADER * loghdr, const cha
...
   loghdr->db_charset = lang_charset ();
-  loghdr->fpageid = 0;
+  if (prm_get_bigint_value (PRM_ID_FIRST_LOG_PAGEID) != 0)
+    {
+      loghdr->fpageid = (LOG_PAGEID) prm_get_bigint_value (PRM_ID_FIRST_LOG_PAGEID);
+    }
+  else
+    {
+      loghdr->fpageid = 0;
+    }
```
wrlawodms: It looks that we can just do like below without the if-condition:
`loghdr->fpageid = (LOG_PAGEID) prm_get_bigint_value (PRM_ID_FIRST_LOG_PAGEID);` 

Or did you write this to put the case using PRM_ID_FIRST_LOG_PAGEID as an exceptional case explicitly?

hornetmj: @wrlawodms I wanted to separate this explicitly, but your suggestion is simpler and nicer. I will reflect.

src/transaction/log_page_buffer.c
```diff
@@ -2482,6 +2482,13 @@ logpb_fetch_start_append_page (THREAD_ENTRY * thread_p)
 
   logpb_log ("started logpb_fetch_start_append_page\n");
 
+  if (prm_get_bigint_value (PRM_ID_FIRST_LOG_PAGEID) != 0
+      && (LOG_PAGEID) prm_get_bigint_value (PRM_ID_FIRST_LOG_PAGEID) == log_Gl.hdr.append_lsa.pageid
+      && log_Gl.hdr.append_lsa.offset == 0)
+    {
+      flag = NEW_PAGE;
+    }
+
```
```suggestion
  if ((LOG_PAGEID) prm_get_bigint_value (PRM_ID_FIRST_LOG_PAGEID) == log_Gl.hdr.append_lsa.pageid
      && log_Gl.hdr.append_lsa.offset == 0)
    {
      flag = NEW_PAGE;
    }

```

It may remove the if conditions at line 2493 ~ 2496 
but if you intended to distinguish this exceptional case from original case, I think it is OK.

joohok (approved):

src/transaction/log_page_buffer.c
```diff
@@ -2482,6 +2482,13 @@ logpb_fetch_start_append_page (THREAD_ENTRY * thread_p)
 
   logpb_log ("started logpb_fetch_start_append_page\n");
 
+  if (prm_get_bigint_value (PRM_ID_FIRST_LOG_PAGEID) != 0
+      && (LOG_PAGEID) prm_get_bigint_value (PRM_ID_FIRST_LOG_PAGEID) == log_Gl.hdr.append_lsa.pageid
```
wrlawodms: Then, it can also be combined with the if-condition below like:

`f ((log_Gl.hdr.append_lsa.pageid == prm_get_bigint_value (PRM_ID_FIRST_LOG_PAGEID)) && (log_Gl.hdr.append_lsa.offset == 0))`

wrlawodms: Ah, joohok already pointed it out.

wrlawodms:
@hornetmj  How about making it work only in DEBUG mode? Do we have to make the regression test work also in RELEASE MODE?

eido5 (approved):

src/transaction/log_page_buffer.c
hornetmj: @joohok You are right.

hornetmj:
@wrlawodms No, some regression tests (_debug postfix) and RQG, etc are only performed in debug mode.
@mhoh3963 Can it be supported only in DEBUG mode?

mhoh3963:
@hornetmj 
Ok. Let's support only in debug mode.

mhoh3963 (approved):

joohok (approved):

src/transaction/log_page_buffer.c
```diff
@@ -2483,7 +2487,7 @@ logpb_fetch_start_append_page (THREAD_ENTRY * thread_p)
   logpb_log ("started logpb_fetch_start_append_page\n");
 
   /* detect empty log (page and offset of zero) */
-  if ((log_Gl.hdr.append_lsa.pageid == 0) && (log_Gl.hdr.append_lsa.offset == 0))
+  if ((log_Gl.hdr.append_lsa.pageid == log_Gl.hdr.fpageid) && (log_Gl.hdr.append_lsa.offset == 0))
```
wrlawodms: `log_Gl.hdr.fpageid` is the logical pageid at physical location 1, which seems to be possibly non-zero. For example, in `logpb_next_append_page()`, it is set to the first logical page id of the first one in a new log active volume. If this is comparing with the logical page id 0, it seems that we should not use `log_Gl.hdr.fpageid`.

hornetmj: @wrlawodms Thanks for your comments. First of all, I separated the release mode and debug mode behavior. So the release mode works as before without any modifications. However, there is a bit of a problem with the fix in debug mode. If the first_log_pageid setting is changed between DB creation and DB startup, unexpected results may occur. To prevent this, it is necessary to permanently store the information at the time of DB creation. However, this seems like an over-correction as it is not very valuable information. I can't come up with a good idea to solve this tonight. Do you have any good suggestions? If not, I will try to prevent problems by sharing this with the QA team first. Because they are waiting for this PR to be merged to test the 10.2.6 release.

wrlawodms: @hornetmj It seems like the simplest way. It is a hidden and debug-only parameter so that looks good to me.

wrlawodms (approved):

PR #3245, opened 2021-12-10
[LETS-284] page server relays tran table snapshot LSA to passive transaction server at boot

cristiarg:
http://jira.cubrid.org/browse/LETS-284

- page server's replication processes LOG_TRANTABLE_SNAPSHOT and saves the last such log record's LSA
- upon a passive transaction log boot sequence, it relays the current (most recent) transaction table snapshot LSA along with the rest of the log boot info
- with subsequent patches, passive transaction server will use this LSA to construct its own initial transaction table


src/transaction/log_manager.c
```diff
@@ -1655,6 +1657,24 @@ log_initialize_passive_tran_server (THREAD_ENTRY * thread_p)
...
   }
+  // NOTE: following situations can happen with regard to most recent transaction table snapshot lsa:
+  //  1. it is null here at destination on PTS:
+  //    1.1. the PTS has been started too fast after the ATS and did not get to generate and relay at least
+  //        one transaction table snapshot to the PTS via PS
+  //        - to mitigate this the PTS will have its own routine to search the log backwards for the most
+  //          recent transaction table snapshot
+  //    1.2. no transaction table snapshot has ever been generated ever
+  //        - even if the PTS searches backwards for a transaction table snapshot log record, it will not
+  //          find one; there is no solution for this state
```
```suggestion
  //          find one; the transaction table would start empty in this case.
```

ribram (approved):

PR #3247, opened 2021-12-13
[CBRD-24101] Expand in-line view : Equivalent to the view-merging criteria, in-line views can be merged.

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-24101

**Support inline view merging**
The main logic of view merging is mq_lambda() and mq_class_lambda(). In this function, table or column is replaced during view merging.
Referring to the existing routine that only works for views, functions that works the same for inline views is added.
The following changes have been made.
 - Use 'spec' instead of 'entity'.
 - Fix some routines to work in inline-view. (e.g. a function to get an attribute)

| added function                                            | existing function                                          |
| mq_substitute_inline_view_in_statement ()  | mq_substitute_subquery_list_in_statement |
| mq_substitute_select_for_inline_view ()       | mq_substitute_select_in_statement ()          |
| mq_inline_view_lambda ()                            | mq_class_lambda ()                                     |
| mq_lambda ()                                               | use same function                                       |

**Move repeating routines as common functions**
- move check routines to mq_is_pushable_subquery()
- move copying sql hint to mq_copy_sql_hint ()

**Add check for view merging**
the case that view-merging is not possible is added to check. It is also equivalently applied to existing views.
The following checks have been added.

```
check that main query has inst_num and subquery has order_by
check for correlated subquery
check selecting schema to be not pushable
subquery has CTE spec.
check for NOT FROM
check that subquery has order_by and main query has inst_num or analytic or order-sensitive aggrigation function.
check for using index keylimit.
check json function.
check for for_update
check for MERGE query
```

**Fix bug unrelated to VIEW MERGE**
```
Fix a bug that would change the correlation level absurdly.
Fix a bug where the table name was used incorrectly when index skip scan.
Fixed the problem of searching in the subquery select list up to the order_by of the main query when copying the order_by of the subquery.
Do not delete order_by of subquery when view merging is not possible.
Copy type_enum when rewriting derived table
```

**Add NO_MERGE sql hint**
View merging is not performed when the NO_MERGE hint is used in the subquery.
```
select * from (select /*+ NO_MERGE */ * from ...)
```


src/parser/xasl_generation.c
```diff
@@ -11672,7 +11672,7 @@ pt_fix_first_term_func_index_for_iss (PARSER_CONTEXT * parser, QO_INDEX_ENTRY *
   seg = QO_ENV_SEG (index_entryp->terms.env, index_entryp->seg_idxs[1]);
   head = QO_SEG_HEAD (seg);
   spec = head->entity_spec;
-  class_name = (char *) spec->info.spec.range_var->info.name.original;
+  class_name = parser_print_tree (parser, spec->info.spec.entity_name);
```
hgryoo: I found class_name is got like the following in many other codes. Is it exactly the same? 
```
classname = node->info.spec.entity_name->info.name.original;
```

src/parser/parser_support.c
```diff
@@ -709,6 +709,47 @@ pt_is_expr_wrapped_function (PARSER_CONTEXT * parser, const PT_NODE * node)
...
+	  || function_type == F_JSON_MERGE || function_type == F_JSON_MERGE_PATCH
+	  || function_type == F_JSON_OBJECT
+	  || function_type == F_JSON_PRETTY
+	  || function_type == F_JSON_QUOTE
+	  || function_type == F_JSON_REMOVE
+	  || function_type == F_JSON_REPLACE
+	  || function_type == F_JSON_SEARCH
+	  || function_type == F_JSON_SET
+	  || function_type == F_JSON_TYPE || function_type == F_JSON_UNQUOTE || function_type == F_JSON_VALID
+	  || function_type == PT_JSON_ARRAYAGG || function_type == PT_JSON_OBJECTAGG)
```
beyondykk9: how about rewrite like "if (function_type >= F_JSON_OBJECT and function_type <= F_JSON_VALID)"

ctshim: If you check many if conditions without branching, a switch statement like below might be more efficient.
```

switch(node->info.function.function_type)
{
     case F_JSON_ARRAY:
     case F_JSON_ARRAY_APPEND:
     case: F_JSON_ARRAY_INSERT:
    ...
	case PT_JSON_OBJECTAGG:
               return true;
         default:
               break;
} 
```

src/parser/parser_support.c
```diff
@@ -1049,6 +1090,105 @@ pt_find_aggregate_functions_post (PARSER_CONTEXT * parser, PT_NODE * tree, void
...
+{
+  bool *has_group_concat = (bool *) arg;
+
+  if (tree && tree->node_type == PT_FUNCTION &&
+      (tree->info.function.function_type == PT_GROUP_CONCAT
+       || tree->info.function.function_type == PT_CUME_DIST
+       || tree->info.function.function_type == PT_PERCENT_RANK
+       || tree->info.function.function_type == PT_PERCENTILE_CONT
+       || tree->info.function.function_type == PT_PERCENTILE_DISC
+       || tree->info.function.function_type == F_SEQUENCE || pt_is_json_function (parser, tree)))
```
ctshim, line 1180: The pt_is_json_function() function is only used here. 
How about removing the function and changing it to use a switch statement directly here?

src/parser/view_transform.c
```diff
@@ -444,7 +455,7 @@ mq_bump_corr_pre (PARSER_CONTEXT * parser, PT_NODE * node, void *void_arg, int *
 
   *continue_walk = PT_CONTINUE_WALK;
 
-  if (!PT_IS_QUERY_NODE_TYPE (node->node_type))
+  if (!PT_IS_SELECT (node))
```
ctshim, line 465: Is this function not entered for PT_UNION, PT_DIFFERENCE, and PT_INTERSECTION? Or has it been changed to return right here?

src/parser/view_transform.c
```diff
@@ -12029,3 +12976,79 @@ mq_copy_view_error_msgs (PARSER_CONTEXT * parser, PARSER_CONTEXT * query_cache)
...
+
+      dest_query->info.query.q.select.use_nl =
+	parser_append_node (parser_copy_tree_list (parser, src_query->info.query.q.select.use_nl),
+			    dest_query->info.query.q.select.use_nl);
+
+      dest_query->info.query.q.select.use_idx =
+	parser_append_node (parser_copy_tree_list (parser, src_query->info.query.q.select.use_idx),
+			    dest_query->info.query.q.select.use_idx);
+
+      if (!is_index_ss || dest_query->info.query.q.select.index_ss != NULL)
```
ctshim, line 13024: In the policy using hints, when INDEX_LS and INDEX_SS hints are specified at the same time, one hint is ignored.
INDEX_LS seems to be set only when (PT_HINT_NO_INDEX_SS or NOT PT_HINT_INDEX_SS).

Please refer to pt_print_select() and qo_get_ils_prefix_length().

src/parser/xasl_generation.c
shparkcubrid: Okay. Using parser_print_tree(), it seems that even unnecessary things will be printed. I'll fix it.

src/parser/parser_support.c
shparkcubrid: Okay. "if (function_type >= F_JSON_OBJECT and function_type <= F_JSON_VALID)" is good.
I'll fix it.

src/parser/parser_support.c
shparkcubrid: I think "if (function_type >= F_JSON_OBJECT and function_type <= F_JSON_VALID)" is better.

src/parser/view_transform.c
shparkcubrid: It is related to commit 093b314
The correlation level refers to the level of the external reference column.
```
select * from tbl a, (select * from tbl2 b 
                                 where b.col1 = a.col1 <== correlated column 
                               ) <== correlation level 2
```
UNION, DIFFERENCE, INTERSECTION is not related to correlation level.
It is bug fix.

ctshim: OK, I see.

src/parser/view_transform.c
shparkcubrid: Okay. I'll fix it.

src/parser/parser_support.c
shparkcubrid: PT_JSON_ARRAYAGG, PT_JSON_OBJECTAGG This is not possible because of these two. I'll change it to a switch

src/parser/view_transform.c
```diff
@@ -1334,81 +1460,232 @@ mq_substitute_spec_in_method_names (PARSER_CONTEXT * parser, PT_NODE * node, voi
...
  *  - has CONNECT BY (Hierarchical Queries)
  *  - has DISTINCT
  *  - has aggregate or orderby_for or analytic
  *  - has inst num or orderby_num
  *  - has method
  */
-static bool
-mq_is_pushable_subquery (PARSER_CONTEXT * parser, PT_NODE * query, bool is_only_spec)
+static int
+mq_is_pushable_subquery (PARSER_CONTEXT * parser, PT_NODE * subquery, PT_NODE * mainquery, PT_NODE * class_spec)
```
hgryoo: I understand the return type is changed to int type because It can return 3 types that have the following meaning: error (-1), non-pushable (0), or pushable (1). For more readability, how about defining an enum type to make it clear?

beyondykk9: I also agree with @hgryoo.

hyunikn (approved):

src/parser/xasl_generation.c
youngjinj: Aren't range_var and entity_name different?
If alias is used in the query, it seems appropriate to use range_var.

beyondykk9 (approved):

src/parser/xasl_generation.c
shparkcubrid: This is a bug where a table name is required, but an alias is used.

youngjinj (approved):

src/parser/view_transform.c
shparkcubrid: I fixed it.

ctshim (approved):

hgryoo (approved):

PR #3248, opened 2021-12-13
[CBRD-23960] Support returning Cursor between stored procedures

hgryoo:

Support functionality that a stored procedure calls another stored procedure and gets ResultSet from it. 

- implement make_out_resultset in method_invoke
  - define `make_outresult_info`
  - (CAS) find `query_handler` in `method_callback` by query_id and returns column_info and query_result_info
  - (DB Server) make `cubmethod::query_cursor` with query_id.
  - (Server-side JDBC) define `CUBRIDServerSideOutResultSet` and implement `getObject()` and `setObject()` in `CUBRIDServerSidePreparedStatement`
- returning Cursor in `method_invoke_java::receive_result`
  - If result value or out parameter is DB_TYPE_REESULTSET type, query_cursor is removed from `method_invoke_java::m_cursor_map`. It means that the query result will not be freed by `xqmgr_end_query()` to reuse the stored procedure that uses OutResultSet
  - at #3233, when Cursor is returned to the client, the query result (`QMGR_QUERY_ENTRY`) is stored in session. It means that the query result is accessible over the commit (holdable cursor). But between the stored procedure, the result doesn't have to be managed by session.

please refer to the demo examples: SpCursorTest.testOutResultWithQuery() and SpCursorTest.PyramidPrinter()


hgryoo:
The regression that is currently occurring is due to casting to client-side JDBC in the stored procedure implementation of the test case. It looks like the following example:
```
	public static ResultSet testResultSet1() { 
		try { 
			Class.forName("cubrid.jdbc.driver.CUBRIDDriver"); 
			Connection con = DriverManager.getConnection("jdbc:default:connection:"); 
	
			((CUBRIDConnection)con).setCharset("euc_kr");  // *** should be removed
			
			String query = "select * from kor order by id,name"; 
			Statement stmt = con.createStatement(); 
			ResultSet rs = stmt.executeQuery(query); 
	
			((CUBRIDResultSet)rs).setReturnable();  // *** should be removed
			
			return rs;
		} catch (Exception e) { 
			//e.printStackTrace(); 
		} 
                return null;
	} 
```

src/method/method_invoke_java.cpp
```diff
@@ -200,6 +209,15 @@ namespace cubmethod
...
 
+	if (db_value_type (&returnval) == DB_TYPE_RESULTSET)
+	  {
+	    std::uint64_t query_id = db_get_resultset (&returnval);
+	    query_cursor *cursor = m_cursor_map [query_id];
+	    cursor->close ();
+	    delete cursor;
+	    m_cursor_map.erase (query_id);
+	  }
+
```
beyondykk9, line 226: This routine has already been checked in the preceding code. It looks redundant.
I think the "temp" seems to be suitable here.

src/method/method_invoke_java.cpp
```diff
@@ -187,6 +187,15 @@ namespace cubmethod
     value_unpacker.value = &returnval;
     value_unpacker.unpack (unpacker);
 
+    if (db_value_type (&returnval) == DB_TYPE_RESULTSET)
+      {
```
beyondykk9, line 191: returnval is always DB_TYPE_NULL by precediing code "db_make_null (&returnval)".

src/jsp/com/cubrid/jsp/impl/SUStatement.java
```diff
@@ -224,9 +242,9 @@ public void execute(
 
             CallInfo callInfo = executeInfo.callInfo;
             tuples = new SUResultTuple[fetchedTupleNumber];
-            tuples[0] = callInfo.getTuple ();
-        } else {
-            cursorPosition = -1;
+            tuples[0] = callInfo.getTuple();
+        } else if (getSQLType() == true) {
+            queryId = executeInfo.getResultInfo(0).queryId;
```
beyondykk9, line 247: no "else" clause needed?

src/method/method_callback.cpp
```diff
@@ -230,6 +233,30 @@ namespace cubmethod
...
+    unpacker.unpack_all (query_id);
+
+    cubmethod::query_handler *query_handler = get_query_handler_by_qid (query_id);
+    cubmethod::query_result *qresult = query_handler->get_current_result();
+
+    make_outresult_info info;
+
+    query_handler->set_prepare_column_list_info (info.column_infos, *qresult);
+    query_handler->set_qresult_info (info.qresult_infos);
+
```
beyondykk9: necessary to check "query_handler is null"

src/method/method_scan.cpp
```diff
@@ -151,11 +151,11 @@ namespace cubscan
 	      DB_VALUE *dbval_p = (DB_VALUE *) db_private_alloc (m_thread_p, sizeof (DB_VALUE));
 	      db_make_null (dbval_p);
```
beyondykk9, line 158: necesary to assert "dbval_p is not null"

src/method/method_invoke_java.cpp
hgryoo: returnval is set by unpack at the above
```
value_unpacker.value = &returnval;
value_unpacker.unpack (unpacker);
```

src/method/method_callback.cpp
```diff
@@ -230,6 +233,30 @@ namespace cubmethod
...
+  {
+    uint64_t query_id;
+    unpacker.unpack_all (query_id);
+
+    cubmethod::query_handler *query_handler = get_query_handler_by_qid (query_id);
+    cubmethod::query_result *qresult = query_handler->get_current_result();
+
+    make_outresult_info info;
+
+    query_handler->set_prepare_column_list_info (info.column_infos, *qresult);
```
ctshim: *qresult can cause a core dump when qresult is null.

src/method/method_invoke_java.cpp
```diff
@@ -187,6 +187,15 @@ namespace cubmethod
...
+	std::uint64_t query_id = db_get_resultset (&returnval);
+	query_cursor *cursor = m_cursor_map [query_id];
+	cursor->close ();
```
ctshim: Is there any possibility that cursor is null?

src/method/method_invoke_java.cpp
```diff
@@ -635,6 +657,61 @@ namespace cubmethod
...
+	      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, sizeof (query_cursor));
+	      return ER_OUT_OF_VIRTUAL_MEMORY;
+	    }
+
+	  cursor->open ();
+	}
+
+      error = method_send_buffer_to_java (m_group->get_socket(), b);
+      return error;
+    };
```
ctshim, line 713: If there is no need to use ';', it would be better to delete it.

src/method/method_invoke_java.cpp
beyondykk9: Ok, I missed the unpack.

src/method/method_scan.cpp
hgryoo: I've fixed to check dbval_p is null and make to occur an error

src/method/method_callback.cpp
hgryoo: I've fixed

src/method/method_callback.cpp
hgryoo: I've fixed

src/method/method_invoke_java.cpp
hgryoo: I've fixed

src/jsp/com/cubrid/jsp/impl/SUStatement.java
hgryoo: there is nothing to do in "else" clause

src/method/method_invoke_java.cpp
hgryoo: I've fixed

src/method/method_invoke_java.cpp
hgryoo: ok, I'm gonna fix it

beyondykk9 (approved):

ctshim (approved):

shparkcubrid (approved):

hyunikn (approved):

youngjinj (approved):

PR #3249, opened 2021-12-13
[CBRD-24131] Fix nonexpected assignment on assert statement in debug build

lsahn-gh:
http://jira.cubrid.org/browse/CBRD-24131

The assignment expression of the same value on assert statement is not that
we expected although currently it doesn't change the behavior of the engine.

Caller function, do_execute_statement already knows that node_type would be
PT_DROP_SESSION_VARIABLES before callee function, do_drop_session_variables.


mhoh3963:
@memnoth 
Thank you for your contribution.
Let's add two reviewers and proceed with the review.

mhoh3963 (approved):

beyondykk9 (approved):

PR #3252, opened 2021-12-14
[CBRD-24127] Improved CUBRID installation method with tar.gz file

airnet73:
http://jira.cubrid.org/browse/CBRD-24127

**Purpose**
The installation program is a shell script including binary, so you can conveniently install and create demodb.
However, when installing CUBRID with a tar.gz file, the user must manually set the environment variable and create the demodb.
It is necessary to provide the install shell script to the user for more convenient installation.

**Implementation**
Create .cubrid.sh file and create .cubrid.sh.bak file
Create .cubrid.csh file and create .cubrid.csh.bak file
Register .cubrid.sh in .bash_profile.
Create databases folder and create demodb

**Remarks**
N/A

### DIFF
--- a/contrib/install/install.sh
+++ b/contrib/install/install.sh
@@ -41,8 +41,7 @@ if [ -z "$cubrid_home" ];then
   cubrid_home="`pwd`"
 fi
 
-echo "Is CUBRID installed in the directory below? [Yn]:"
-echo "\"$cubrid_home\""
+echo "Is the CUBRID installed in "$cubrid_home" ? [Yn]:"
 
 read line leftover
 is_installed_dir=TRUE
--- a/contrib/install/install.sh
+++ b/contrib/install/install.sh
@@ -1,0 +1,210 @@
+#!/bin/bash
+#
+#
+#  Copyright 2016 CUBRID Corporation
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+
+function get_cubrid_dir() {
+  cubrid_str=/CUBRID/
+  cubrid_str_len=${#cubrid_str}
+  current_dir="`pwd`"
+
+  find_dir=$(echo ${current_dir} | grep ${cubrid_str})
+
+  if [ -z "$find_dir" ];then
+    echo ""
+  else
+    temp_dir=${current_dir%%$cubrid_str*}
+    temp_dir_len=${#temp_dir}
+    cubrid_dir_len=`expr $temp_dir_len + $cubrid_str_len`
+    cubrid_dir=$(echo $current_dir | cut -c1-$cubrid_dir_len)
+    echo "$cubrid_dir"
+  fi
+}
+
+
+cubrid_home=$(get_cubrid_dir)
+
+if [ -z "$cubrid_home" ];then
+  cubrid_home="`pwd`"
+fi
+
+echo "Is CUBRID installed in the directory below? [Yn]:"
+echo "\"$cubrid_home\""
+
+read line leftover
+is_installed_dir=TRUE
+
+case ${line} in
+  n* | N*)
+    is_installed_dir=FALSE
+esac
+
+if [ "x${is_installed_dir}x" = "xFALSEx" ];then
+  echo ""
+  echo "Please enter the directory where CUBRID is installed: "
+
+  read input_dir leftover
+  cubrid_home=${input_dir}
+fi
+
+# environment variables for *csh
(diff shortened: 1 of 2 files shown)


contrib/install/install.sh
```diff
@@ -0,0 +1,210 @@
...
+
+
+cubrid_home=$(get_cubrid_dir)
+
+if [ -z "$cubrid_home" ];then
+  cubrid_home="`pwd`"
+fi
+
+echo "Is CUBRID installed in the directory below? [Yn]:"
+echo "\"$cubrid_home\""
```
kisoo-han: Suggest:
echo "Is the CUBRID installed in \"$cubrid_home\" ? [Yn]:"

contrib/install/install.sh
```diff
@@ -0,0 +1,210 @@
...
+fi
+SHLIB_PATH=$LD_LIBRARY_PATH
+LIBPATH=$LD_LIBRARY_PATH
+PATH=$CUBRID/bin:$PATH
+export CUBRID
+export CUBRID_DATABASES
+export LD_LIBRARY_PATH
+export SHLIB_PATH
+export LIBPATH
+export PATH
```
kisoo-han, line 143: What about add blank line below this line, for same effect with CUBRID install.sh
```
export PATH

LIB=$CUBRID/lib

if [ -f /etc/redhat-release ];then
```

contrib/install/install.sh
```diff
@@ -0,0 +1,210 @@
...
+LIB=$CUBRID/lib
+if [ -f /etc/redhat-release ];then
+        OS=$(cat /etc/system-release-cpe | cut -d':' -f'3-3')
+elif [ -f /etc/os-release ];then
+        OS=$(cat /etc/os-release | egrep "^ID=" | cut -d'=' -f2-2)
+fi
+case $OS in
```
kisoo-han: add intentional blank line here:

```
fi

case $OS in
```

contrib/install/install.sh
```diff
@@ -0,0 +1,210 @@
...
+esac
+
+if [ "x${is_installed_dir}x" = "xFALSEx" ];then
+  echo ""
+  echo "Please enter the directory where CUBRID is installed: "
+
+  read input_dir leftover
+  cubrid_home=${input_dir}
+fi
+
```
kisoo-han, line 61: How about checking for existence of CUBRID home before proceed ?

```
if [ ! -d $cubrid_home ];then
  echo "$cubrid_home: no such directory"
  exit
fi
```

contrib/install/install.sh
airnet73: Thanks for your advice.

contrib/install/install.sh
airnet73: Thanks for your advice.

contrib/install/install.sh
airnet73: Thanks for your advice.

contrib/install/install.sh
airnet73: If the path is long, readability is not good, so it is expressed in two lines.
let us know your opinion.

kisoo-han: It is the style of the cubrid install shell script.
Personally, I prefer single line even it is too long, but it doesn't matter.
Do as you like.

airnet73: Thanks for an opinion.
I agree with your opinion.

kisoo-han: Because the last line is not a question, but a proposed CUBRID directory, I have to read previous line to answer. It is little bit confused. Thanks for accepting my opinion.

kisoo-han (approved):

contrib/install/install.sh
```diff
@@ -0,0 +1,217 @@
...
+LIBPATH=$LD_LIBRARY_PATH
+PATH=$CUBRID/bin:$PATH
+export CUBRID
+export CUBRID_DATABASES
+export LD_LIBRARY_PATH
+export SHLIB_PATH
+export LIBPATH
+export PATH
+
+LIB=$CUBRID/lib
```
mhoh3963, line 145: Where is the CUBRID variable set ?
Is it set on line 89 ?

airnet73: The CUBRID variable is set in line 89.

mhoh3963: OK, I misunderstand the code.

contrib/install/install.sh
```diff
@@ -0,0 +1,217 @@
...
+setenv CUBRID_DATABASES $CUBRID/databases
+if (${?LD_LIBRARY_PATH}) then
+  setenv LD_LIBRARY_PATH $CUBRID/lib:${LD_LIBRARY_PATH}
+else
+  setenv LD_LIBRARY_PATH $CUBRID/lib
+endif
+setenv SHLIB_PATH $LD_LIBRARY_PATH
+setenv LIBPATH $LD_LIBRARY_PATH
+set path=($CUBRID/bin $path)
+____cubrid__here_doc____
```
mhoh3963, line 120: Even if it is csh, I think you need to check the os and lib as bash.
can you add code which make symbolic link if needed library doesn't exist ?

airnet73: fixed it

kisoo-han (approved):

mhoh3963 (approved):

mhoh3963:
@airnet73 
There are failure test case due to other commit.
so, please let me know if it is needed to force merge.

PR #3256, opened 2021-12-14
[CBRD-23842] Fix coredump due to not handling NULL exception in cubrid_log API

joohok:

Coredump is caused by strlen (cond_column_data) if cond_column_data is returned from the server as a NULL value.

### DIFF
--- a/src/api/cubrid_log.c
+++ b/src/api/cubrid_log.c
@@ -1418,7 +1418,15 @@ cubrid_log_make_dml (char **data_info, DML * dml)
 	    case 7:
 	      dml->cond_column_data[i] = ptr;
 	      ptr = or_unpack_string_nocopy (ptr, &dml->cond_column_data[i]);
-	      dml->cond_column_data_len[i] = (int) strlen (dml->cond_column_data[i]);
+	      if (dml->cond_column_data[i] == NULL)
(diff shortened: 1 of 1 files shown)


src/api/cubrid_log.c
```diff
@@ -1418,7 +1418,15 @@ cubrid_log_make_dml (char **data_info, DML * dml)
...
 	      dml->cond_column_data[i] = ptr;
 	      ptr = or_unpack_string_nocopy (ptr, &dml->cond_column_data[i]);
-	      dml->cond_column_data_len[i] = (int) strlen (dml->cond_column_data[i]);
+	      if (dml->cond_column_data[i] == NULL)
+		{
+		  dml->cond_column_data_len[i] = 0;
+		}
+	      else
+		{
+		  dml->cond_column_data_len[i] = (int) strlen (dml->cond_column_data[i]);
```
hornetmj, line 1427: How to distinguish '' and NULL value in the char and varchar domain?

joohok: if data length is 0, NULL check is required on client side. 

I provide this information about how to check NULL value and empty string in the API manual.

hornetmj (approved):

PR #3257, opened 2021-12-14
[CBRD-23960] Unallow multiple SQL statements when creating a single Statement object in server-side JDBC

hgryoo:

This PR changed structures to manage only one statement
- `method_query_handler` has only one `query_result`
- Structures for communication are changed to have only one `query_result`
  - execute_info
  - make_outresult_info
- Server-side JDBC is also changed the same way
- When preparing a statement, the `method_query_handler` checks if the parsed SQL is multiple statements, and if so, send an error.

**Example**) please refer to the demo/TestChangedSpecTest.java file
```
test1:
Statement stmt = conn.createStatement();
stmt.execute("select * from t1; insert into t1 (b) values (1); select * from t1"); // throws exception

test2:
String query = "select * from t1; insert into t1 (b) values (1); select * from t1";
PreparedStatement stmt = conn.prepareStatement(query); // throws exception
```

NOTE
- please review from 51f1b7e


src/method/method_query_handler.cpp
```diff
@@ -774,28 +601,18 @@ namespace cubmethod
...
     m_prepare_flag = flag;
 
-    m_current_result = NULL;
-    /* in cas_execute.c, srv_handle->cur_result_index = 0; is not actually index */
-    /* To make understandable, I'm not going to follow it */
-    m_current_result_index = -1;
-
-    query_result q_result;
+    query_result &q_result = m_query_result;
     q_result.stmt_type = stmt_type;
```
byungwook-kim, line 605: stmt_type is always CUBRID_STMT_NONE. it looks useless.

src/method/method_query_handler.hpp
```diff
@@ -111,13 +116,14 @@ namespace cubmethod
       int prepare_query (prepare_info &info, int &flag);
       int prepare_call (prepare_info &info, int &flag);
 
+      /* check */
+
+
```
ctshim: There is a comment related to check, but no actual execution code.
Any code to add for check?

src/method/method_query_handler.cpp
```diff
@@ -743,7 +570,7 @@ namespace cubmethod
 	db_session_set_xasl_cache_pinned (m_session, true, false);
       }
 
-    char stmt_type;
+    char stmt_type = CUBRID_STMT_NONE;
```
beyondykk9, line 580: it looks not necessary. the stmt_type is always set by the folllowing code sequences.

src/method/method_query_handler.cpp
```diff
@@ -246,218 +236,59 @@ namespace cubmethod
...
     int error = NO_ERROR;
-    int num_q_result = m_q_result.size();
     OID ins_oid = OID_INITIALIZER;
     DB_OBJECT *ins_obj_p;
 
-    qinfo.resize (num_q_result);
-    for (int i = 0; i < num_q_result; i++)
+    query_result &qresult = m_query_result;
+    DB_QUERY_RESULT *qres = (DB_QUERY_RESULT *) qresult.result;
+    if (qresult.stmt_type == CUBRID_STMT_INSERT && qres != NULL)
```
beyondykk9, line 249: It needs an explanation why stmt_type must be CUBRID_STMT_INSERT.

src/method/method_query_handler.hpp
hgryoo: I'm gonna remove it.

src/method/method_query_handler.cpp
hgryoo: It is to suppress the uninitialized variable warnings.

beyondykk9: Ok.

hyunikn (approved):

src/method/method_query_handler.cpp
hgryoo: It is for the GET_GENERATED_KEYS request. I've added a comment.

shparkcubrid (approved):

beyondykk9 (approved):

ctshim (approved):

youngjinj (approved):

PR #3262, opened 2021-12-16
[LETS-327] Use send_recv to get boot info for transaction servers

ribram:
http://jira.cubrid.org/browse/LETS-327

Replace getting boot info and log boot info for ATS and PTS respectively with send_recv/respond instead of two-way requests.

Remove the obsolete request ID's used before by page server to send back the info. Also handlers for these ID's have been removed and the mutex/condvar & other related members.


src/server/tran_server.cpp
```diff
@@ -236,13 +236,14 @@ tran_server::init_page_server_hosts (const char *db_name)
 void
 tran_server::get_boot_info_from_page_server ()
 {
-  assert (!m_is_boot_info_received);
+  std::string request_message;  // empty
```
cristiarg: remove var
```suggestion
```

src/server/passive_tran_server.cpp
```diff
@@ -87,22 +69,15 @@ passive_tran_server::receive_log_prior_list (page_server_conn_t::sequenced_paylo
 void passive_tran_server::send_and_receive_log_boot_info (THREAD_ENTRY *thread_p,
     log_lsa &most_recent_transaction_table_snapshot_lsa)
 {
-  assert (m_log_boot_info.empty ());
+  std::string request_message; // empty
```
cristiarg:
```suggestion
```

src/server/passive_tran_server.cpp
```diff
@@ -87,22 +69,15 @@ passive_tran_server::receive_log_prior_list (page_server_conn_t::sequenced_paylo
...
+  std::string log_boot_info;
 
-  // empty message request
-  push_request (tran_to_page_request::SEND_LOG_BOOT_INFO_FETCH, std::string ());
+  send_receive (tran_to_page_request::SEND_LOG_BOOT_INFO_FETCH, std::move (request_message), log_boot_info);
```
cristiarg:
```suggestion
  send_receive (tran_to_page_request::SEND_LOG_BOOT_INFO_FETCH, std::string (), log_boot_info);
```

src/server/page_server.cpp
```diff
@@ -138,12 +138,14 @@ page_server::connection_handler::receive_data_page_fetch (tran_server_conn_t::se
...
 
 void
-page_server::connection_handler::receive_log_boot_info_fetch (tran_server_conn_t::sequenced_payload &)
+page_server::connection_handler::receive_log_boot_info_fetch (tran_server_conn_t::sequenced_payload &a_sp)
 {
   // empty request message
 
-  auto callback_func =
-	  std::bind (&connection_handler::on_log_boot_info_result, this, std::placeholders::_1);
+  auto callback_func = [this, sp = a_sp] (std::string &&message) mutable
```
cristiarg, line 145: Could we implement this differently such that the copy ctor/operator on `sequenced_payload` is not needed?
Maybe by exposing some interface out of `page_server::connection_handler` and forwarding a pointer/reference to that to the page fetcher?

That copy ctor/operator on `sequenced_payload` might become a performance bottleneck if misused - right now, as I understand it, it only "copies" an empty string.

cristiarg (approved):

src/server/page_server.cpp
ribram: My intention was to use move semantics, but with all my struggle I could not find a consistent way to move the payload into the the bounded function and then move it again when the function is invoked. After half a day of trying out various suggestions, I gave up.

I think this may get another shot with LETS-329, when maybe async page fetcher could be rewritten in a way that this becomes possible (e.g. move the payload into the task).

PR #3263, opened 2021-12-16
[LETS-305] Check page desynchronization in heap_prepare_get_context

adrianbrab:
http://jira.cubrid.org/browse/LETS-305

Added steps to verify page desynchronization inside `heap_prepare_get_context`.
The checks are as follows:

1. After fixing home page and reading the object slot, if the record type is `REC_RELOCATION` or `REC_BIGONE`, check home page LSA. If it is higher than replication LSA, register the error and return.
2. After fixing forward page, in both `REC_RELOCATION` and `REC_BIGONE` cases, check its LSA. If is ahead of replication LSA, register the error and return.
3. Forward pages may also be deallocated. If fixing failed with `ER_PB_BAD_PAGEID`, the forward page must be fixed again as `OLD_PAGE_DEALLOCATED` and its LSA checked; if it higher than replication LSA, register the proper error and return. If it is not, then `ER_PB_BAD_PAGEID` remains.

### DIFF
--- a/src/storage/heap_file.c
+++ b/src/storage/heap_file.c
@@ -7468,7 +7468,6 @@ try_again:
   /* Output record type. */
   context->record_type = slot_p->record_type;
 
-#if defined (SERVER_MODE)
   if (context->record_type == REC_BIGONE || context->record_type == REC_RELOCATION)
     {
       ret = pgbuf_check_page_ahead_of_replication (thread_p, context->home_page_watcher.pgptr);
@@ -7478,7 +7477,6 @@ try_again:
 	  goto error;
 	}
     }
-#endif
 
   if (context->fwd_page_watcher.pgptr != NULL && slot_p->record_type != REC_RELOCATION
       && slot_p->record_type != REC_BIGONE)
@@ -7526,14 +7524,12 @@ try_again:
 
 	      goto error;
 	    }
-#if defined (SERVER_MODE)
 	  ret = pgbuf_check_page_ahead_of_replication (thread_p, context->fwd_page_watcher.pgptr);
 
 	  if (ret != NO_ERROR)
 	    {
 	      goto error;
 	    }
-#endif
 	  return S_SUCCESS;
 	}
       else if (ret == ER_PB_BAD_PAGEID)
@@ -7575,14 +7571,12 @@ try_again:
 	      assert (false);
 	      goto error;
 	    }
-#if defined (SERVER_MODE)
 	  ret = pgbuf_check_page_ahead_of_replication (thread_p, context->fwd_page_watcher.pgptr);
 
 	  if (ret != NO_ERROR)
 	    {
 	      goto error;
 	    }
-#endif
 	  return S_SUCCESS;
 	}
       else if (ret == ER_PB_BAD_PAGEID)
--- a/src/storage/heap_file.c
+++ b/src/storage/heap_file.c
@@ -7541,18 +7541,10 @@ try_again:
 	  VPID *vpid = pgbuf_get_vpid_ptr (context->fwd_page_watcher.pgptr);
 	  if (vpid == NULL)
 	    {
+	      ret = ER_PB_BAD_PAGEID;
 	      goto error;
 	    }
-
-	  PAGE_PTR fixed_page =
-	    pgbuf_fix (thread_p, vpid, OLD_PAGE_DEALLOCATED, context->latch_mode, PGBUF_UNCONDITIONAL_LATCH);
-	  if (fixed_page == NULL)
-	    {
-	      goto error;
-	    }
-
-	  ret = ER_PAGE_AHEAD_OF_REPLICATION;
-	  goto error;
+	  ret = pgbuf_check_for_deallocated_page_or_desyncronization (thread_p, context->latch_mode, *vpid);
 	}
 
       goto error;
@@ -7598,18 +7590,10 @@ try_again:
 	  VPID *vpid = pgbuf_get_vpid_ptr (context->fwd_page_watcher.pgptr);
 	  if (vpid == NULL)
 	    {
+	      ret = ER_PB_BAD_PAGEID;
 	      goto error;
 	    }
-
-	  PAGE_PTR fixed_page =
-	    pgbuf_fix (thread_p, vpid, OLD_PAGE_DEALLOCATED, context->latch_mode, PGBUF_UNCONDITIONAL_LATCH);
-	  if (fixed_page == NULL)
-	    {
-	      goto error;
-	    }
-
-	  ret = ER_PAGE_AHEAD_OF_REPLICATION;
-	  goto error;
+	  ret = pgbuf_check_for_deallocated_page_or_desyncronization (thread_p, context->latch_mode, *vpid);
 	}
 
       goto error;
--- a/src/storage/heap_file.c
+++ b/src/storage/heap_file.c
@@ -7468,6 +7468,18 @@ try_again:
   /* Output record type. */
   context->record_type = slot_p->record_type;
 
+#if defined (SERVER_MODE)
+  if (context->record_type == REC_BIGONE || context->record_type == REC_RELOCATION)
+    {
+      ret = pgbuf_check_page_ahead_of_replication (thread_p, context->home_page_watcher.pgptr);
+
+      if (ret != NO_ERROR)
+	{
+	  goto error;
+	}
+    }
+#endif
+
   if (context->fwd_page_watcher.pgptr != NULL && slot_p->record_type != REC_RELOCATION
       && slot_p->record_type != REC_BIGONE)
     {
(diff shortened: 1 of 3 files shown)


src/storage/page_buffer.c
```diff
@@ -2309,27 +2309,14 @@ pgbuf_fix_read_old_and_check_repl_desync (THREAD_ENTRY * thread_p, const VPID &
...
-      if (!deallocated)
-	{
-	  // Stop if other errors occurred.
-	  return nullptr;
-	}
-
-      // Force fixing the deallocated page to check its LSA
-      page = pgbuf_fix (thread_p, &vpid, OLD_PAGE_DEALLOCATED, PGBUF_LATCH_READ, PGBUF_UNCONDITIONAL_LATCH);
-      if (page == nullptr)
+      if (pgbuf_check_for_deallocated_page_or_desyncronization (thread_p, PGBUF_LATCH_READ, vpid) == ER_PB_BAD_PAGEID)
```
ribram: Should return in both cases. It is important only to overwrite ER_PB_BAD_PAGEID with 
```suggestion
      (void) pgbuf_check_for_deallocated_page_or_desyncronization (thread_p, PGBUF_LATCH_READ, vpid);
      return nullptr;
```

src/storage/page_buffer.c
```diff
@@ -2389,6 +2368,29 @@ pgbuf_check_page_ahead_of_replication (THREAD_ENTRY * thread_p, PAGE_PTR page)
 }
 #endif
 
+int
+pgbuf_check_for_deallocated_page_or_desyncronization (THREAD_ENTRY * thread_p, PGBUF_LATCH_MODE latch_mode,
+						      const VPID & vpid)
+{
+  int ret = ER_PB_BAD_PAGEID;
```
ribram, line 2376: If bad pageid error is assumed, better to safe-guard that it is true.
```suggestion
  assert (er_errid () == ER_PB_BAD_PAGEID);
  int ret = ER_PB_BAD_PAGEID;
```

src/storage/page_buffer.c
```diff
@@ -2389,6 +2368,29 @@ pgbuf_check_page_ahead_of_replication (THREAD_ENTRY * thread_p, PAGE_PTR page)
...
+
+  PAGE_PTR fixed_page = pgbuf_fix (thread_p, &vpid, OLD_PAGE_DEALLOCATED, latch_mode, PGBUF_UNCONDITIONAL_LATCH);
+  if (fixed_page == NULL)
+    {
+      return ER_PB_BAD_PAGEID;
```
ribram: Maybe this time there's another error.
```suggestion
      ASSERT_ERROR_AND_SET (ret);
      return ret;
```

src/storage/page_buffer.c
```diff
@@ -2389,6 +2368,29 @@ pgbuf_check_page_ahead_of_replication (THREAD_ENTRY * thread_p, PAGE_PTR page)
...
+
+#if defined (SERVER_MODE)
+  ret = pgbuf_check_page_ahead_of_replication (thread_p, fixed_page);
+  if (ret == NO_ERROR)
+    {
+      return ER_PB_BAD_PAGEID;
+      pgbuf_unfix (thread_p, fixed_page);
+    }
+#endif
+  return ret;
```
ribram:
```suggestion
  return pgbuf_check_page_ahead_of_replication (thread_p, fixed_page);
```

src/storage/page_buffer.h
```diff
@@ -468,6 +468,8 @@ extern void pgbuf_daemons_destroy ();
 // Check if page is ahead of replication; only relevant on passive transaction server, don't call elsewhere.
 extern int pgbuf_check_page_ahead_of_replication (THREAD_ENTRY * thread_p, PAGE_PTR page);
```
ribram, line 468: Activate pgbuf_check_page_ahead_of_replication declaration/definition also in not SERVER_MODE, to avoid putting #ifdef SERVER_MODE/#endif wherever it is being used.

Change its definition to:
```c
int
pgbuf_check_page_ahead_of_replication  (...)
{
#if defined (SERVER_MODE)
  // current body
#else
  return NO_ERROR;
#endif
}
```

Do something similar for pgbuf_check_for_deallocated_page_or_desyncronization:
``` c
int
pgbuf_check_for_deallocated_page_or_desyncronization (...)
{
  assert (err_errid () == ER_PB_BAD_PAGEID);
  ret = ER_PB_BAD_PAGEID;
#if defined (SERVER_MODE)
  // rest of current body
#else
  return ret;
#endif
}
```

src/storage/heap_file.c
```diff
@@ -7468,6 +7468,16 @@ heap_prepare_get_context (THREAD_ENTRY * thread_p, HEAP_GET_CONTEXT * context, b
   /* Output record type. */
   context->record_type = slot_p->record_type;
 
+  if (context->record_type == REC_BIGONE || context->record_type == REC_RELOCATION)
+    {
+      ret = pgbuf_check_page_ahead_of_replication (thread_p, context->home_page_watcher.pgptr);
+
+      if (ret != NO_ERROR)
```
ribram:
```suggestion
      ret = pgbuf_check_page_ahead_of_replication (thread_p, context->home_page_watcher.pgptr);
      if (ret != NO_ERROR)
```

src/storage/heap_file.c
```diff
@@ -7512,10 +7522,26 @@ heap_prepare_get_context (THREAD_ENTRY * thread_p, HEAP_GET_CONTEXT * context, b
 			  context->forward_oid.pageid);
 		}
 
+	      goto error;
+	    }
+	  ret = pgbuf_check_page_ahead_of_replication (thread_p, context->fwd_page_watcher.pgptr);
+
+	  if (ret != NO_ERROR)
```
ribram:
```suggestion
	  ret = pgbuf_check_page_ahead_of_replication (thread_p, context->fwd_page_watcher.pgptr);
	  if (ret != NO_ERROR)
```

src/storage/heap_file.c
```diff
@@ -7545,8 +7571,24 @@ heap_prepare_get_context (THREAD_ENTRY * thread_p, HEAP_GET_CONTEXT * context, b
 	      assert (false);
 	      goto error;
 	    }
+	  ret = pgbuf_check_page_ahead_of_replication (thread_p, context->fwd_page_watcher.pgptr);
+
+	  if (ret != NO_ERROR)
```
ribram:
```suggestion
	  ret = pgbuf_check_page_ahead_of_replication (thread_p, context->fwd_page_watcher.pgptr);
	  if (ret != NO_ERROR)
```

ribram (approved):
Usually, we don't leave an empty line between a function call and the following returned error code check. Please remove the empty lines, and merge.

src/storage/page_buffer.c
```diff
@@ -2309,29 +2309,14 @@ pgbuf_fix_read_old_and_check_repl_desync (THREAD_ENTRY * thread_p, const VPID &
...
-
-      // Force fixing the deallocated page to check its LSA
-      page = pgbuf_fix (thread_p, &vpid, OLD_PAGE_DEALLOCATED, PGBUF_LATCH_READ, PGBUF_UNCONDITIONAL_LATCH);
-      if (page == nullptr)
-	{
-	  // Another error occurred. Stop.
-	  ASSERT_ERROR ();
-	  return nullptr;
-	}
+      (void) pgbuf_check_for_deallocated_page_or_desyncronization (thread_p, PGBUF_LATCH_READ, vpid);
```
wrlawodms, line 2318: What if not `ER_PB_BAD_PAGEID`? It seems that the `assert (er_errid () == ER_PB_BAD_PAGEID);` in `pgbuf_check_for_deallocated_page_or_desyncronization()`  fails.

src/storage/heap_file.c
```diff
@@ -7545,8 +7570,24 @@ heap_prepare_get_context (THREAD_ENTRY * thread_p, HEAP_GET_CONTEXT * context, b
...
 	}
+      else if (ret == ER_PB_BAD_PAGEID)
+	{
+	  VPID *vpid = pgbuf_get_vpid_ptr (context->fwd_page_watcher.pgptr);
+	  if (vpid == NULL)
+	    {
+	      ret = ER_PB_BAD_PAGEID;
+	      goto error;
+	    }
+	  ret = pgbuf_check_for_deallocated_page_or_desyncronization (thread_p, context->latch_mode, *vpid);
```
wrlawodms, line 7589: It seems that if `ER_PB_BAD_PAGEID` occurs because of other cause than the desynchronization, `pgbuf_check_for_deallocated_page_or_desyncronization` returns NO_ERROR, which ends up skipping the error. What do you think?

wrlawodms: Or is it guaranteed that the error from pgbuf_fix() with OLD_PAGE_DEALLOCATED is always the same as that with OLD_PAGE when it's not from the desynchronization?

wrlawodms (commented):
@ribram @adrianbrab Sorry for the late comment after the merge, but there seem to be bugs.

src/storage/page_buffer.c
ribram: thanks for pointing this, we will correct it.

PR #3265, opened 2021-12-17
[CBRD-24105] Addition '-z' and '--no-check' options with 'backupdb' u…

eido5:
http://jira.cubrid.org/browse/CBRD-24105

**Purpose**
Addition '-z' and '--no-check' options with 'backupdb' utility in the 'ha_make_slavedb.sh'. 

**Implementation**
in `$ <user>/cubrid/contrib/scripts/ha`
"ha_make_slavedb.sh" shell script is modified;
at `# optional`  category, `backup_option="-z --no-check" is written`.

So this line is affected in the "ha_make_slavedb.sh" 
 `ssh_cubrid $target_host "cubrid backupdb $backup_option -C -D $backup_dest_path -o $backupdb_output $db_name@localhost"`

**Remarks**
N/A

### DIFF
--- a/contrib/scripts/ha/ha_make_slavedb.sh
+++ b/contrib/scripts/ha/ha_make_slavedb.sh
@@ -1,4 +1,21 @@
 #!/bin/bash
+#
+#  Copyright 2008 Search Solution Corporation
+#  Copyright 2016 CUBRID Corporation
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+#
 
 CURR_DIR=$(dirname $0)
 WORK_DIR=$(pwd)
--- a/contrib/scripts/ha/ha_make_slavedb.sh
+++ b/contrib/scripts/ha/ha_make_slavedb.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 #
-#  Copyright 2008 Search Solution Corporation
+#  
 #  Copyright 2016 CUBRID Corporation
 #
 #   Licensed under the Apache License, Version 2.0 (the "License");
--- a/contrib/scripts/ha/ha_make_slavedb.sh
+++ b/contrib/scripts/ha/ha_make_slavedb.sh
@@ -1,20 +1,20 @@
 #!/bin/bash
 #
-#  
+#  Copyright 2008 Search Solution Corporation
 #  Copyright 2016 CUBRID Corporation
-#
+# 
 #   Licensed under the Apache License, Version 2.0 (the "License");
-#   you may not use this file except in compliance with the License
+#   you may not use this file except in compliance with the License.
 #   You may obtain a copy of the License at
-#
+# 
 #       http://www.apache.org/licenses/LICENSE-2.0
-#
+# 
 #   Unless required by applicable law or agreed to in writing, software
 #   distributed under the License is distributed on an "AS IS" BASIS,
 #   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
-#
+# 
 
 CURR_DIR=$(dirname $0)
 WORK_DIR=$(pwd)
--- a/contrib/scripts/ha/ha_make_slavedb.sh
+++ b/contrib/scripts/ha/ha_make_slavedb.sh
@@ -16,7 +16,7 @@ repl_log_home=$CUBRID_DATABASES
 db_name=
 
 backup_dest_path=
-backup_option=
+backup_option="-z --no-check"
 restore_option=
 scp_option="-l 131072"		# default - limit : 16M*1024*8=131072
 ssh_port=22
--- a/contrib/scripts/ha/ha_make_slavedb.sh
+++ b/contrib/scripts/ha/ha_make_slavedb.sh
@@ -15,7 +15,6 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 #
-#
 
 CURR_DIR=$(dirname $0)
 WORK_DIR=$(pwd)


eido5:
I'm trying to solve why the license is failed.

hornetmj:
@mhoh3963 @ssihil git_checks complains about the license. It seems that there is no copyright in the modified file, how should I handle it?

wrlawodms:
@hornetmj 
Possibilities to deal with the license issue:
1. Just force-merge for this case.
2. Add a license header into the script too.
3. Add a file list to skip license check. A skip mechanism is required for this.

contrib/scripts/ha/ha_make_slavedb.sh
```diff
@@ -16,7 +16,7 @@ repl_log_home=$CUBRID_DATABASES
 db_name=
 
 backup_dest_path=
-backup_option=
+backup_option="-z --no-check"
```
wrlawodms, line 35: 1. These two options seem to be always set if no options are given.
2.  Given any option, these seem to be overwritten.

Am I correct and is it correct?

joohok:
The license header was written only to build.sh among scripts, not to scripts in "contrib/". 
If license header is required to be added, I think it needs consistency.

contrib/scripts/ha/ha_make_slavedb.sh
joohok: The parameters in this "User Configuration" part are set manually by modifying the script if necessary. 
So, I wonder why the backup option needs this initial value ?

mhoh3963:
@hornetmj @eido5 
When I changed the license, I decided to add the missing license information when editing the file later.
How about adding license header to ha_make_slavedb.sh?

hornetmj:
@hornetmj I got it. @wrlawodms, @joohok Thanks for your suggestions.

contrib/scripts/ha/ha_make_slavedb.sh
eido5: Before each replies is tagged, there is the notice.
Tech team requested that two options,-z and --no-check, are written explicitly.

eido5: @wrlawodms 

1. --no-check option is set to default and -z option is set to default.

> --no-check               don't check a consistency of the database; default: check

2. Can I ask the meaning "overwritten"? does it stands for options written more than once?
Are there any problems if they are written more than once?

eido5: @joohok 
Did my first reply help your reply? Or are there any something what I missed?

eido5:
@joohok 
According to @mhoh3963, addition license is decided when the file or files is(are) needed to be edit.
I wonder that is the consistency critical problem? Because If not, I think that insertion the license by a case seems so efficient.

eido5:
> @hornetmj Possibilities to deal with the license issue:
> 
> 1. Just force-merge for this case.
> 2. Add a license header into the script too.
> 3. Add a file list to skip license check. A skip mechanism is required for this.

Can I ask the reason of 3rd suggestion? Have there been any similar problems?

wrlawodms:
@eido5 There are other files including no license header, but (2) has been chosen. Let's deal with this kind of issue in a greedy manner as @mhoh3963 said.

contrib/scripts/ha/ha_make_slavedb.sh
wrlawodms: @eido5 What do you mean by "explicitly"? `z` and `--no-check` is special ones than other options?
Other options have to be written into `backup_option` to be used with `cubrid backupdb` below, but why are the two options set by default?

wrlawodms: Furthermore, `-z` option is even not printed in the help message of `cubrid backupdb` and compression is applied by default now.

eido5: > Furthermore, `-z` option is even not printed in the help message of `cubrid backupdb` and compression is applied by default now.

Right. I should write down about the -z option with no.2 . 
[CBRD-24032](http://jira.cubrid.org/browse/CBRD-24032)

I can guess from your reply that "overwritten" seems no problem.
Tech team requested the addition options explicitly for the compatibility with the previous CUBRID version.

eido5:
> @eido5 There are other files including no license header, but (2) has been chosen. Let's deal with this kind of issue in a greedy manner as @mhoh3963 said.

Ok. Let's handle this problem in the greedy manner.
And I still wonder why you suggest 3rd method. Can I ask about the detailed reasons?

wrlawodms:
@eido5 There are other files including no license header. This is the reason. The third one could be a general way to deal with this kind of issue. We can define that some files must have license headers, and some don't.

contrib/scripts/ha/ha_make_slavedb.sh
wrlawodms: Regrading "overwritten", I supposed that the variables can be set out of the script, but I'm not sure it's possible. If it is impossible, never mind. My opinion is basically the same as @joohok's.

wrlawodms: To clarify, the request is that '-z' and '--no-check' have to be always set when no option is given, and if the option is given(means `backup_option` is changed) it doesn't need to be set?

eido5:
> @eido5 There are other files including no license header. This is the reason. The third one could be a general way to deal with this kind of issue. We can define that some files must have license headers, and some don't.

Oh I see. I understood definitely.

contrib/scripts/ha/ha_make_slavedb.sh
eido5: > To clarify, the request is that '-z' and '--no-check' have to be always set when no option is given, and if the option is given(means `backup_option` is changed) it doesn't need to be set?

I understood your reply that '-z' and '--no-check' is always set whether `backup_option` is changed.
If it, yes it is. 

the request includes that it is necessary for the compatibility with any version CUBRID. 
So `backup_option` is set to  '-z' and '--no-check' options.

Can i ask the case of `backup_option` is given?(I may guess it is transfer by any argument.)

wrlawodms: > `Can i ask the case of backup_option is given?(I may guess it is transfer by any argument.)`

I thought something like `backup_option='[other option]' ./ha_make_slavedb.sh `, but it will be overitten by `backup_option="-z --no-check"` here. Never mind.

> If it, yes it is.

1. When a user changes the `backup-option` , the -z and --no-check can be removed. Is it part of the request?
2. In other cubrid versions, `--no-check` and '-z' is the default behavior of backupdb? There are -z and --no-check in 10.1 too. What do you mean by "the compatibility"?

eido5: @wrlawodms 
> 1. When a user changes the `backup-option` , the -z and --no-check can be removed. Is it part of the request?

 Yes. If user wants to remove or change options, it allows. 
 The point is that Tech team usually sets the `-z` and `--no-check` option, so they want to write on script and always set these options automatically.
 
> 2. In other cubrid versions, `--no-check` and '-z' is the default behavior of backupdb? There are -z and --no-check in 10.1 too. What do you mean by "the compatibility"?

I used "the compatibility" for the previous and next version cubrid. 
I know that after this issue, [CBRD-24032](http://jira.cubrid.org/browse/CBRD-24032), compress '-z' option is set to default.
So, I think these options can be set to default or not in any versions.

joohok: I understand what you mean. 
Before requesting review, please explain why this work is required in the jira. 
I thought this default option was necessary to set the minimum options the script would run on, and I thought there is no problem to execute this script without that backup options you describe. 
So if the user wants the '-z' and '--no-check' options, the user can change the options themselves. But with the "compatibility" view, I think it is OK. 
 is '--no-check' option also added for the compatibility issue?

eido5: @joohok 
> Before requesting review, please explain why this work is required in the jira. 

Oh I see. My one is lack of explanation. 

> Is '--no-check' option also added for the compatibility issue?

In my opinion, the compatibility includes '--no-check' option. It is related to the back-up performance. 
So Tech team needs to write on scripts even '--no-check' option.

Is '--no-check' option not enough relation with 'compatibility'?

hornetmj (approved):

joohok (approved):

wrlawodms (approved):

PR #3267, opened 2021-12-17
[LETS-306] Handle pages desynchronization errors in heap_get_visible_version_internal

adrianbrab:
http://jira.cubrid.org/browse/LETS-306

Added a new `heap_get_visible_version_with_repl_desync` function to handle page desyncronization situations that occur inside `heap_get_visible_version_internal`.

The function is only active on passive transaction servers and will replace all calls of `heap_get_visible_version_internal`.

The function checks for the `ER_PAGE_AHEAD_OF_REPLICATION` error and unfixes the pages waits for replication and calls `heap_get_visible_version_internal` again until the error does not occur.

### DIFF
--- a/src/storage/heap_file.c
+++ b/src/storage/heap_file.c
@@ -24983,7 +24983,6 @@ heap_get_visible_version_with_repl_desync (THREAD_ENTRY * thread_p, HEAP_GET_CON
   do
     {
       ret = heap_get_visible_version_with_repl_desync (thread_p, context, is_heap_scan);
-
       if (ret != S_ERROR || er_errid () != ER_PAGE_AHEAD_OF_REPLICATION)
 	{
 	  // We'll continue here only if we have a page desynchronization error. Break the loop in any other cases.
--- a/src/storage/heap_file.c
+++ b/src/storage/heap_file.c
@@ -24981,7 +24981,7 @@ heap_get_visible_version_with_repl_desync (THREAD_ENTRY * thread_p, HEAP_GET_CON
 
       if (ret != S_SUCCESS)
 	{
-	  if (er_errid () == ER_PB_BAD_PAGEID)
+	  if (er_errid () == ER_PAGE_AHEAD_OF_REPLICATION)
 	    {
 	      if (context->home_page_watcher.pgptr)
 		{
--- a/src/storage/heap_file.c
+++ b/src/storage/heap_file.c
@@ -24991,7 +24991,6 @@ heap_get_visible_version_with_repl_desync (THREAD_ENTRY * thread_p, HEAP_GET_CON
 	  pgbuf_ordered_unfix (thread_p, &context->fwd_page_watcher);
 	}
 
-      assert (is_passive_transaction_server ());
       passive_tran_server *const pts_ptr = get_passive_tran_server_ptr ();
       LOG_TDES *tdes = LOG_FIND_CURRENT_TDES (thread_p);
       assert (tdes->page_desync_lsa.is_null ());
--- a/src/storage/heap_file.c
+++ b/src/storage/heap_file.c
@@ -24969,26 +24969,15 @@ heap_get_visible_version_with_repl_desync (THREAD_ENTRY * thread_p, HEAP_GET_CON
 {
   SCAN_CODE ret;
 #if defined (SERVER_MODE)
-  if (is_active_transaction_server ())
-    {
-      // Active transaction server doesn't have the page desynchronization issue.
-      ret = heap_get_visible_version_internal (thread_p, context, is_heap_scan);
-      if (ret == S_ERROR)
-	{
-	  assert (er_errid () != ER_PAGE_AHEAD_OF_REPLICATION);
-	}
-      return ret;
-    }
-
   do
     {
-      ret = heap_get_visible_version_with_repl_desync (thread_p, context, is_heap_scan);
+      ret = heap_get_visible_version_internal (thread_p, context, is_heap_scan);
       if (ret != S_ERROR || er_errid () != ER_PAGE_AHEAD_OF_REPLICATION)
 	{
 	  // We'll continue here only if we have a page desynchronization error. Break the loop in any other cases.
 	  break;
 	}
-
+      assert (is_passive_transaction_server ());
       // Handle the page desynchronization error
       if (context->home_page_watcher.pgptr)
 	{
@@ -25007,6 +24996,7 @@ heap_get_visible_version_with_repl_desync (THREAD_ENTRY * thread_p, HEAP_GET_CON
       LOG_TDES *tdes = LOG_FIND_CURRENT_TDES (thread_p);
       assert (tdes->page_desync_lsa.is_null ());
       pts_ptr->wait_replication_pasts_target_lsa (tdes->page_desync_lsa);
+      tdes->page_desync_lsa.set_null ();
     }
   while (true);
 #else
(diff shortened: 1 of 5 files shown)


src/storage/heap_file.c
```diff
@@ -24946,13 +24948,72 @@ heap_scan_get_visible_version (THREAD_ENTRY * thread_p, const OID * oid, OID * c
...
+ *  is_heap_scan (in): required for heap_prepare_get_context
+ */
+SCAN_CODE
+heap_get_visible_version_with_repl_desync (THREAD_ENTRY * thread_p, HEAP_GET_CONTEXT * context, bool is_heap_scan)
+{
+#if defined (SERVER_MODE)
+
+  if (is_active_transaction_server ())
+    {
+      return heap_get_visible_version_with_repl_desync (thread_p, context, is_heap_scan);
```
ribram: The function is not meant to be called recursively. Call heap_get_visible_version_internal inside.
Also add an explanation comment.
```suggestion
      // Active transaction server doesn't have the page desynchronization issue.
      return heap_get_visible_version_internal (thread_p, context, is_heap_scan);
```

src/storage/heap_file.c
```diff
@@ -24946,13 +24948,72 @@ heap_scan_get_visible_version (THREAD_ENTRY * thread_p, const OID * oid, OID * c
...
+	  else
+	    {
+	      return ret;
+	    }
+	}
+    }
+  while (ret != S_SUCCESS);
+  return ret;
+#else
+  return heap_get_visible_version_with_repl_desync (thread_p, context, is_heap_scan);
```
ribram:
```suggestion
  // SA_MODE doesn't have the page desynchronization issue.
  return heap_get_visible_version_internal (thread_p, context, is_heap_scan);
```

src/storage/heap_file.c
```diff
@@ -24946,13 +24948,72 @@ heap_scan_get_visible_version (THREAD_ENTRY * thread_p, const OID * oid, OID * c
...
+	      assert (tdes->page_desync_lsa.is_null ());
+	      pts_ptr->wait_replication_pasts_target_lsa (tdes->page_desync_lsa);
+	    }
+	  else
+	    {
+	      return ret;
+	    }
+	}
+    }
+  while (ret != S_SUCCESS);
```
ribram: Condition is already checked inside the loop. Break the loop manually.
```suggestion
  while (true);
```

src/storage/heap_file.c
```diff
@@ -24946,13 +24948,72 @@ heap_scan_get_visible_version (THREAD_ENTRY * thread_p, const OID * oid, OID * c
...
+	      passive_tran_server *const pts_ptr = get_passive_tran_server_ptr ();
+	      LOG_TDES *tdes = LOG_FIND_CURRENT_TDES (thread_p);
+	      assert (tdes->page_desync_lsa.is_null ());
+	      pts_ptr->wait_replication_pasts_target_lsa (tdes->page_desync_lsa);
+	    }
+	  else
+	    {
+	      return ret;
+	    }
+	}
```
ribram: I suggest a rewrite like:

```c
ret = heap_get_visible_version_internal (thread_p, context, is_heap_scan);
if (ret != S_ERROR || er_errid () != ER_PAGE_AHEAD_OF_REPLICATION)
  {
    // We'll continue here only if we have a page desynchronization error. Break the loop in any other cases.
    break;
  }

// Handle the page desynchronization error
...
```

Also make sure `heap_get_visible_version_internal()` is called.

src/storage/heap_file.c
```diff
@@ -24946,13 +24948,78 @@ heap_scan_get_visible_version (THREAD_ENTRY * thread_p, const OID * oid, OID * c
...
+      if (ret == S_ERROR)
+	{
+	  assert (er_errid () != ER_PAGE_AHEAD_OF_REPLICATION);
+	}
+      return ret;
+    }
+
+  do
+    {
+      ret = heap_get_visible_version_with_repl_desync (thread_p, context, is_heap_scan);
```
ribram: `heap_get_visible_version_with_repl_desync()` always calls `heap_get_visible_version_internal()` and never itself.
```suggestion
      ret = heap_get_visible_version_internal (thread_p, context, is_heap_scan);
```

src/storage/heap_file.c
```diff
@@ -24946,13 +24948,78 @@ heap_scan_get_visible_version (THREAD_ENTRY * thread_p, const OID * oid, OID * c
...
+    {
+      // Active transaction server doesn't have the page desynchronization issue.
+      ret = heap_get_visible_version_internal (thread_p, context, is_heap_scan);
+      if (ret == S_ERROR)
+	{
+	  assert (er_errid () != ER_PAGE_AHEAD_OF_REPLICATION);
+	}
+      return ret;
+    }
+
```
ribram: On second thought, I don't think we need this; the ATS can also go into the do/while loop, but it should break the loop on the first iteration.
```suggestion
```

src/storage/heap_file.c
```diff
@@ -24946,13 +24948,78 @@ heap_scan_get_visible_version (THREAD_ENTRY * thread_p, const OID * oid, OID * c
...
+    {
+      ret = heap_get_visible_version_with_repl_desync (thread_p, context, is_heap_scan);
+      if (ret != S_ERROR || er_errid () != ER_PAGE_AHEAD_OF_REPLICATION)
+	{
+	  // We'll continue here only if we have a page desynchronization error. Break the loop in any other cases.
+	  break;
+	}
+
+      // Handle the page desynchronization error
+      if (context->home_page_watcher.pgptr)
```
ribram, line 24982: Only PTS should reach this part of the loop.
```suggestion
      assert (is_passive_transaction_server ());
      if (context->home_page_watcher.pgptr)
```

src/storage/heap_file.c
```diff
@@ -24946,13 +24948,78 @@ heap_scan_get_visible_version (THREAD_ENTRY * thread_p, const OID * oid, OID * c
...
+	{
+	  /* Unfix forward page. */
+	  pgbuf_ordered_unfix (thread_p, &context->fwd_page_watcher);
+	}
+
+      assert (is_passive_transaction_server ());
+      passive_tran_server *const pts_ptr = get_passive_tran_server_ptr ();
+      LOG_TDES *tdes = LOG_FIND_CURRENT_TDES (thread_p);
+      assert (tdes->page_desync_lsa.is_null ());
+      pts_ptr->wait_replication_pasts_target_lsa (tdes->page_desync_lsa);
```
ribram, line 24997: also reset the page_desync_lsa
```suggestion
      pts_ptr->wait_replication_pasts_target_lsa (tdes->page_desync_lsa);
      tdes->page_desync_lsa.set_null ();
```

src/storage/heap_file.c
adrianbrab: I will also test this to make sure

src/storage/heap_file.c
```diff
@@ -24946,13 +24948,68 @@ heap_scan_get_visible_version (THREAD_ENTRY * thread_p, const OID * oid, OID * c
...
+	  pgbuf_ordered_unfix (thread_p, &context->home_page_watcher);
+	}
+
+      if (context->fwd_page_watcher.pgptr != NULL)
+	{
+	  /* Unfix forward page. */
+	  pgbuf_ordered_unfix (thread_p, &context->fwd_page_watcher);
+	}
+
+      assert (is_passive_transaction_server ());
```
ribram: Already asserted.
```suggestion
```

ribram (approved):
Remove duplicate assert and merge.

PR #3269, opened 2021-12-21
[LETS-328] Use send_recv() to get log and data page from page server

ribram:
http://jira.cubrid.org/browse/LETS-328

Replace the page broker system with send_recv() calls.

List of changes:

  - Instead of sending a request to get the data/log page, and then wait for the page broker to return the requested page, use direcly send_recv. `pgbuf_request_data_page_from_page_server()` and `request_log_page_from_page_server()` were changed to handle the response unpacking too.
  - Remove the SEND_LOG_PAGE and SEND_DATA_PAGE requests from page server to transaction server. Also remove the handlers on the transaction server side (as a result, the common passive/active handler map is now empty).
  - Adapt page server code to use `respond()` (some copy payload issue from LETS-327 is present here; to be solved in LETS-329).
  - Remove page_broker.hpp, all related functions in `tran_server` and its unit test.


src/server/page_server.cpp
```diff
@@ -121,14 +121,18 @@ page_server::connection_handler::receive_log_page_fetch (tran_server_conn_t::seq
       _er_log_debug (ARG_FILE_LINE, "Received request for log from Transaction Server. Page ID: %lld \n", pageid);
     }
 
-  m_ps.get_page_fetcher ().fetch_log_page (pageid, std::bind (&page_server::connection_handler::on_log_page_read_result,
-      this, std::placeholders::_1, std::placeholders::_2));
+  auto callback_func = [this, sp = a_sp] (const LOG_PAGE *log_page, int error_code) mutable
```
cristiarg, line 124: I checked and the compiler says that:
```
      sequenced_payload &operator= (const sequenced_payload &) = delete;
```
is ok. It only needs the copy ctor.

src/storage/page_buffer.c
```diff
@@ -8215,11 +8211,56 @@ pgbuf_request_data_page_from_page_server (const VPID * vpid, log_lsa target_repl
...
+    }
+
+  std::string response_message;
+  ts_Gl->send_receive (tran_to_page_request::SEND_DATA_PAGE_FETCH, std::move (request_message), response_message);
+
+  const char *message_buf = response_message.c_str ();
+  int error_code = NO_ERROR;
+  std::memcpy (&error_code, message_buf, sizeof (error_code));
+
+
```
cristiarg, line 8228:
```suggestion
```

src/storage/page_buffer.c
```diff
@@ -8215,11 +8211,56 @@ pgbuf_request_data_page_from_page_server (const VPID * vpid, log_lsa target_repl
...
+      if (error_code == ER_PB_BAD_PAGEID)
+	{
+	  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_PB_BAD_PAGEID, 2, VPID_AS_ARGS (vpid));
+	}
+      else
+	{
+	  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_GENERIC_ERROR, 0);
+	}
+
+      return error_code;
```
cristiarg: wipe `io_page` before return?

src/storage/page_buffer.c
```diff
@@ -8215,11 +8211,56 @@ pgbuf_request_data_page_from_page_server (const VPID * vpid, log_lsa target_repl
...
-      _er_log_debug (ARG_FILE_LINE, "[READ DATA] Sent request for Page to Page Server. VPID: %d|%d\n", VPID_AS_ARGS (vpid));
+      _er_log_debug (ARG_FILE_LINE, "[READ DATA] Send request for Page to Page Server. VPID: %d|%d\n", VPID_AS_ARGS (vpid));
+    }
+
+  std::string response_message;
+  ts_Gl->send_receive (tran_to_page_request::SEND_DATA_PAGE_FETCH, std::move (request_message), response_message);
+
+  const char *message_buf = response_message.c_str ();
+  int error_code = NO_ERROR;
+  std::memcpy (&error_code, message_buf, sizeof (error_code));
```
cristiarg, line 8226:
```suggestion
```
done in the if scope

src/server/page_server.cpp
```diff
@@ -245,36 +252,30 @@ page_server::connection_handler::on_log_page_read_result (const LOG_PAGE *log_pa
...
-    {
-      char buffer[IO_MAX_PAGE_SIZE];
-      std::memcpy (buffer, io_page, db_io_page_size ());
-      message = std::string (buffer, db_io_page_size ());
-    }
-
-  if (prm_get_bool_value (PRM_ID_ER_LOG_READ_DATA_PAGE))
-    {
-      if (error_code == NO_ERROR)
+      message.append (reinterpret_cast<const char *> (&error_code), sizeof (error_code));
```
cristiarg: We can unify the approaches for data and log pages. Either always bundle the error code, or add alternatively error code or the actual page. Currently, in both cases, any errors are fatal. But, can it be that some retry/failover mechanism is implemented when more than one page servers would be present?

src/transaction/log_page_buffer.c
```diff
@@ -2176,25 +2193,27 @@ logpb_read_page_from_file_or_page_server (THREAD_ENTRY * thread_p, LOG_PAGEID pa
...
+  const SERVER_TYPE server_type = get_server_type ();
   if (server_type == SERVER_TYPE_TRANSACTION && ts_Gl->is_page_server_connected ())
     {
-      // *INDENT-OFF*
-      std::shared_ptr<log_page_owner> brokered_log_page
-          = ts_Gl->get_log_page_broker ().wait_for_page (pageid);
-      // *INDENT-ON*
-      const LOG_PAGE *const log_page_from_page_server = brokered_log_page->get_log_page ();
-      if (read_from_disk)
+      // context 1) or 2)
```
cristiarg:
```suggestion
```

src/transaction/log_page_buffer.c
```diff
@@ -2176,25 +2193,27 @@ logpb_read_page_from_file_or_page_server (THREAD_ENTRY * thread_p, LOG_PAGEID pa
...
+      if (!read_from_disk)
 	{
-	  // context 2)
-	  // log_pgptr already contains value read from local storage
-	  logpb_verify_page_read (pageid, log_page_from_page_server, log_pgptr);
+	  return request_log_page_from_page_server (pageid, log_pgptr);
```
cristiarg, line 2202:
```suggestion
          // context 1)
	  return request_log_page_from_page_server (pageid, log_pgptr);
```

src/transaction/log_page_buffer.c
```diff
@@ -2176,25 +2193,27 @@ logpb_read_page_from_file_or_page_server (THREAD_ENTRY * thread_p, LOG_PAGEID pa
...
 	}
       else
 	{
-	  // context 1)
-          // *INDENT-OFF*
-	  std::memcpy (log_pgptr, log_page_from_page_server, LOG_PAGESIZE);
-          // *INDENT-ON*
+	  // *INDENT-OFF*
```
cristiarg, line 2207:
```suggestion
          // context 2)
	  // *INDENT-OFF*
```

src/server/page_server.cpp
ribram: Great, I'll remove both operator= and copy constructor in LETS-329. Neither should be necessary anymore.

src/server/page_server.cpp
ribram: > We can unify the approaches for data and log pages. Either always bundle the error code, or add alternatively error code or the actual page.

Yes, you're right. I'll make the change to always bundle them together.

src/storage/page_buffer.c
ribram: I don't think it is necessary; caller should ignore io_page in case of errors.

src/storage/page_buffer.c
ribram: Code was changed.

src/storage/page_buffer.c
ribram: done

src/transaction/log_page_buffer.c
ribram: done

src/transaction/log_page_buffer.c
ribram: done

cristiarg (approved):

PR #3272, opened 2021-12-22
[LETS-329] Add a class to handle server request responses asynchronously 

ribram:
http://jira.cubrid.org/browse/LETS-329

Replace `cublog::async_page_fetcher` with `server_request_responder`; they serve a similar purpose and execute the same functions, but the implementation was rewritten to handle all future requests that require asynchronous response handling.

`server_request_responder` uses a worker pool and `server_request_responder::task` - a cubthread::entry_task derivation. `server_request_responder::task` overrides the execute function to call a request handler and to automatically send the response on `request_sync_client_server`.

The server_request_responder is a template class; the template type must be a request_sync_client_server specialization. In case of page server, the specialization used is `server_request_responder<page_server::connection_handler::tran_server_conn_t>`.

The server_request_handler was added into the server module instead of communication, because of its usage of thread entries. Thread entries are limited to CUBRID servers only.

Complete list of changes:

  - Add `server_request_responder` class and a unit test to verify its functionality.
  - Remove `cublog::async_page_fetcher`, all its references and unit tests.
  - Add functions in `page_server::connection_handler` that unpack input payload, execute the request and pack the response payload. These functions are put into `page_server::connection_handler` out of convenience, but they may be moved anywhere.
  - Push these functions to `page_server.m_responder`
  - Remove request_sync_client_server::sequenced_payload copy constructor and operator=. Add request_sync_client_server::payload_t.


src/communication/request_sync_client_server.hpp
```diff
@@ -46,6 +46,7 @@ namespace cubcomm
     public:
       using outgoing_msg_id_t = T_OUTGOING_MSG_ID;
       using request_client_server_t = cubcomm::request_client_server<T_OUTGOING_MSG_ID, T_INCOMING_MSG_ID>;
+      using payload_t = T_PAYLOAD;
```
cristiarg, line 49: typedef is not used

src/server/page_server.cpp
```diff
@@ -107,64 +109,151 @@ page_server::connection_handler::receive_log_prior_list (tran_server_conn_t::seq
...
-page_server::connection_handler::receive_log_page_fetch (tran_server_conn_t::sequenced_payload &a_sp)
+page_server::connection_handler::push_async_response (F &&a_func, tran_server_conn_t::sequenced_payload &&a_sp)
+{
+  m_ps.get_responder ().async_execute (std::ref (*m_conn), std::move (a_sp),
+				       std::bind (std::forward<F> (a_func), this,
+					   std::placeholders::_1, std::placeholders::_2));
+}
+
+void
+page_server::connection_handler::fetch_log_page (cubthread::entry &context, std::string &a_payload)
```
cristiarg: I see nothing in this function that mandates it being a member of the connection handler. It can be moved to `log_page_buffer`.
If there is ever a need to have member functions as handler functions, `push_async_response` can have two flavors, one that binds a member function; one that maps a global function.

src/server/page_server.cpp
```diff
@@ -107,64 +109,151 @@ page_server::connection_handler::receive_log_prior_list (tran_server_conn_t::seq
...
 
 void
-page_server::connection_handler::receive_data_page_fetch (tran_server_conn_t::sequenced_payload &a_sp)
+page_server::connection_handler::receive_log_page_fetch (tran_server_conn_t::sequenced_payload &a_sp)
+{
+  push_async_response (&connection_handler::fetch_log_page, std::move (a_sp));
+}
+
+void
+page_server::connection_handler::fetch_data_page (cubthread::entry &context, std::string &a_payload)
```
cristiarg: same here; move to `page_buffer`; `m_ps.get_replicator` can be replaced by the call to he global variable

src/server/page_server.cpp
```diff
@@ -107,64 +109,151 @@ page_server::connection_handler::receive_log_prior_list (tran_server_conn_t::seq
...
+	}
+    }
 }
 
 void
-page_server::connection_handler::receive_log_boot_info_fetch (tran_server_conn_t::sequenced_payload &a_sp)
+page_server::connection_handler::receive_data_page_fetch (tran_server_conn_t::sequenced_payload &a_sp)
 {
-  // empty request message
+  push_async_response (&connection_handler::fetch_data_page, std::move (a_sp));
```
cristiarg: this is where `push_async_response` must be added a new flavor for non-member handlers

src/server/server_request_responder.hpp
```diff
@@ -0,0 +1,141 @@
...
+//
+// server_request_responder implementation
+//
+
+template<typename T_CONN>
+server_request_responder<T_CONN>::server_request_responder ()
+  : m_threads_context_manager (std::make_unique<cubthread::system_worker_entry_manager> (TT_SYSTEM_WORKER))
+{
+  const auto THREAD_COUNT = std::thread::hardware_concurrency ();
+  const auto TASK_MAX_COUNT = THREAD_COUNT * 4;
```
cristiarg, line 90: why 4?

unit_tests/server/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+# 
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+# 
+#
+
+# Project to test memory allocation.
```
cristiarg: c&p

src/communication/request_sync_client_server.hpp
ribram: It is used in server_request_responder.

src/server/page_server.cpp
ribram: I hesitate to move the function to log_page_buffer (and others too), because it is tied to the request payload; beside doing the log page fetch, we still need to unpack the request payload and pack the response payload.

The function does not really belong to connection_handler, nor to log_page_buffer, it is somewhere in between. I considered first making a new file, but it seemed too much for three tiny functions :))

src/server/server_request_responder.hpp
ribram: It is arbitrarily chosen; actually, the value has no real effect unless `worker_pool::try_execute()` is used instead of `execute()`.

unit_tests/server/CMakeLists.txt
ribram: fixed

cristiarg (approved):

PR #3273, opened 2021-12-23
[CBRD-24140] Change the javasp utility behavior

hgryoo:
http://jira.cubrid.org/browse/CBRD-24140

- implements get_server_names() which requests server names from master process
- get_server_names() is used for `cubrid service stop`, `cubrid service status`, and `cubrid javasp status` where db name is not specified.
- In `cubrid server start <db_name>` and `cubrid server stop <db_name>`, respectively, `cubrid javasp start <db_name>` and `cubrid javasp stop <db_name>` are executed together. In this case, the java_stored_procedure parameter has to be set to yes at startup.

### DIFF
--- a/conf/cubrid.conf.small
+++ b/conf/cubrid.conf.small
@@ -8,7 +8,7 @@ COMMENT_THIS_LINE = TO_USE_THIS_CONFIGURATION
 [service]
 
 # The list of processes to be started automatically by 'cubrid service start' command
-# Any combinations are available with server, broker, manager, heartbeat and javasp.
+# Any combinations are available with server, broker, manager, heartbeat.
 service=server,broker,manager
 
 # The list of database servers in all by 'cubrid service start' command.
--- a/conf/cubrid.conf.small
+++ b/conf/cubrid.conf.small
@@ -8,7 +8,7 @@ COMMENT_THIS_LINE = TO_USE_THIS_CONFIGURATION
 [service]
 
 # The list of processes to be started automatically by 'cubrid service start' command
-# Any combinations are available with server, broker, manager, heartbeat.
+# Any combinations are available with server, broker, manager and heartbeat.
 service=server,broker,manager
 
 # The list of database servers in all by 'cubrid service start' command.
--- a/src/executables/util_service.c
+++ b/src/executables/util_service.c
@@ -1592,6 +1592,12 @@ process_server (int command_type, int argc, char **argv, bool show_usage, bool c
 		      status = ER_GENERIC_ERROR;
 		    }
 		  print_result (PRINT_SERVER_NAME, status, command_type);
+
+		  /* run javasp utility if server is started successfully */
+		  if (status == NO_ERROR && (is_javasp_running (token) != JAVASP_SERVER_RUNNING))
+		    {
+		      (void) process_javasp (command_type, 1, (const char **) &token, true);
+		    }
 		}
 	    }
 #endif
@@ -1613,11 +1619,6 @@ process_server (int command_type, int argc, char **argv, bool show_usage, bool c
 	      if (token == NULL)
 		{
 		  break;
-		}
-
-	      if (is_javasp_running (token) != JAVASP_SERVER_RUNNING)
-		{
-		  (void) process_javasp (command_type, 1, (const char **) &token, false);
 		}
 
 	      print_message (stdout, MSGCAT_UTIL_GENERIC_START_STOP_3S, PRINT_SERVER_NAME, PRINT_CMD_START, token);
@@ -1662,6 +1663,12 @@ process_server (int command_type, int argc, char **argv, bool show_usage, bool c
 		      status = ER_GENERIC_ERROR;
 		    }
 		  print_result (PRINT_SERVER_NAME, status, command_type);
+
+		  /* run javasp server if DB server is started successfully */
+		  if (status == NO_ERROR && (is_javasp_running (token) != JAVASP_SERVER_RUNNING))
+		    {
+		      (void) process_javasp (command_type, 1, (const char **) &token, false);
+		    }
 		}
 	    }
 	}
@@ -1675,9 +1682,10 @@ process_server (int command_type, int argc, char **argv, bool show_usage, bool c
 	      break;
 	    }
 
+	  /* try to stop javasp server first */
 	  if (is_javasp_running (token) == JAVASP_SERVER_RUNNING)
 	    {
-	      (void) process_javasp (command_type, 1, (const char **) &token, false);
+	      (void) process_javasp (command_type, 1, (const char **) &token, process_window_service);
 	    }
 
 	  print_message (stdout, MSGCAT_UTIL_GENERIC_START_STOP_3S, PRINT_SERVER_NAME, PRINT_CMD_STOP, token);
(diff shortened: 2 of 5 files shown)


src/executables/util_service.c
```diff
@@ -1362,20 +1327,14 @@ process_service (int command_type, bool process_window_service)
 	{
 	  if (!are_all_services_stopped (0, process_window_service))
 	    {
+	      (void) process_javasp (command_type, 0, NULL, false, process_window_service);
```
mhoh3963, line 1330: I think that javasp will stop a process server() or process heartbeat().
Is this code for stopping the javasp if db server doesn't execute and javasp only execute ?

src/executables/util_service.c
```diff
@@ -1665,6 +1674,7 @@ process_server (int command_type, int argc, char **argv, bool show_usage, bool c
 		{
 		  break;
 		}
+
```
mhoh3963: Did you include line feed for readability?

src/executables/util_service.c
```diff
@@ -1442,6 +1396,61 @@ process_service (int command_type, bool process_window_service)
...
 
+/*
+ * get_server_names -
+ *
+ * return:
+ *
+ *      out_buffer (out):
+ */
+static void
+get_server_names (char *out_buffer)
```
mhoh3963: I think that this function collect server name which execute by heartbeat and return collecting value via out_buffer.
Is it right ?
If it right, how about change function name from get_server_names() to get_ha_servers() ?

src/executables/util_service.c
```diff
@@ -1442,6 +1396,61 @@ process_service (int command_type, bool process_window_service)
...
+  if (out_buffer == NULL)
+    {
+      return;
+    }
+
+  make_exec_abspath (cmd, PATH_MAX, (char *) UTIL_COMMDB_NAME " " COMMDB_ALL_STATUS);
+  input = popen (cmd, "r");
+  if (input == NULL)
+    {
+      return;
```
mhoh3963, line 1421: I think that a clear code should be added in case the out_buffer is not initialized.

src/executables/util_service.c
```diff
@@ -1442,6 +1396,61 @@ process_service (int command_type, bool process_window_service)
...
+	    continue;
+      }
+
+      /* find Server */
+      size_t start = s.find (delimiter, 1) + 1;
+      size_t end = s.find (delimiter, start);
+
+      std::string server_name = s.substr (start, end - start);
+
+      snprintf (out_buffer + offset, 4096, "%s,", server_name.c_str ());
```
mhoh3963: Since you don't know the out buffer size, out_buffer overflow can occur if you have many ha servers.
You need to add code to prevent overflow.

src/executables/util_service.c
```diff
@@ -2540,34 +2560,44 @@ process_javasp_status (const char *db_name)
...
+  int master_port = prm_get_master_port_id ();
 
   if (argc == 0)		/* cubrid service command */
     {
-      if (us_Property_map[SERVER_START_LIST].property_value != NULL)
+      /* get all server names from master request */
+      if (css_does_master_exist (master_port))
 	{
-	  strncpy (buf, us_Property_map[SERVER_START_LIST].property_value, sizeof (buf) - 1);
+	  get_server_names (buf);
```
mhoh3963: I think that get_server_names() function only can get list of servers except heartbeat.
But, it seem that here line need all executed server, isn't it ?

src/executables/util_service.c
hgryoo: If a DB server name is not registered in the service section, DB server is not actually stopped by `cubrid service stop` command. Because the master server shuts down by the command, the server loses connection with the master and halts.
Therefore, this line is needed to explicitly pass the stop command.

src/executables/util_service.c
hgryoo: It collects the name of the non-HA servers.

src/executables/util_service.c
mhoh3963: OK, i got.

src/executables/util_service.c
hgryoo: I'm gonna remove this line.

src/executables/util_service.c
mhoh3963: sorry, since I am not good at C++ function, I mis-understand.

hgryoo:
@mhoh3963 I'm gonna fix to handle out_buffer in a safe way soon.

conf/cubrid.conf.small
```diff
@@ -8,7 +8,7 @@ COMMENT_THIS_LINE = TO_USE_THIS_CONFIGURATION
 [service]
 
 # The list of processes to be started automatically by 'cubrid service start' command
-# Any combinations are available with server, broker, manager, heartbeat and javasp.
+# Any combinations are available with server, broker, manager, heartbeat.
```
beyondykk9:
```suggestion
# Any combinations are available with server, broker, manager, and heartbeat.
```

src/executables/util_service.c
hgryoo: I've fixed to get all executed servers. By `cub_commdb -O`, HA-applylogdb and HA-copylogdb are also returned. So I've changed only to handle server names including HA-Server.

conf/cubrid.conf.small
hgryoo: thank you

src/executables/util_service.c
hgryoo: I've changed to allocate buffer (`name_buffer`) for server names in the get_server_names(). a function that calls the get_server_names() must free the buffer if it is not NULL.

src/executables/util_service.c
```diff
@@ -1442,6 +1396,75 @@ process_service (int command_type, bool process_window_service)
...
+  while (fgets (buf, 4096, input) != NULL)
+    {
+      /* *INDENT-OFF* */
+      std::string s (buf);
+
+      /*
+      * ignore HA-applylogdb and HA-copylogdb
+      * check Server and HA-Server
+      */
+      std::string::size_type server_pos = s.find (CHECK_SERVER);
```
mhoh3963, line 1438: It seems to be possible to check with HA-server with CHECK_SERVER.
Do you need to explicitly check the HA-server?

src/executables/util_service.c
mhoh3963: OK

mhoh3963 (approved):

src/executables/util_service.c
hgryoo: Both "HA-Server" and "Server" have the string, "SERVER".  The purpose of this code is to get the name of the server after that space if there exists a string, "Server".
In HA mode, for example, the following string is returned from cub_comm
```
bash> cubrid heartbeat status

HA-applylogdb demodb@localhost:/home/cubrid/javasp_cubrid/build_x86_64_debug/_install/CUBRID/databases/demodb_hgryoo202 (rel 11.2, pid 197770)\n"
HA-copylogdb demodb@hgryoo202:/home/cubrid/javasp_cubrid/build_x86_64_debug/_install/CUBRID/databases/demodb_hgryoo202 (rel 11.2, pid 197768)\n"
HA-Server demodb (rel 11.2, pid 197331)\n
...
```

Else,
```
bash> cubrid service status

Server demodb (rel 11.2, pid 197331)\n
...
```

beyondykk9 (approved):

PR #3275, opened 2021-12-27
[CBRD-24143] Allow multiple CREATE INDEX on filtered column.

beyondykk9:
http://jira.cubrid.org/browse/CBRD-24143

Currently, an error occurs when creating an index with a different filter predicate in the same column as follows:

```
create table foo (a int not null, b int not null, c int);
create index idx1 on foo(a, b) where c=1;
create index idx2 on foo(a, b) where c=2; <-- error occurred 
```
If the filter predicate is different, it should be done so that it can be created without error.

### DIFF
--- a/src/object/class_object.c
+++ b/src/object/class_object.c
@@ -4110,6 +4110,11 @@ classobj_find_constraint_by_attrs (SM_CLASS_CONSTRAINT * cons_list, DB_CONSTRAIN
 			  continue;
 			}
 
+		      if (!filter_predicate->pred_string || !cons->filter_predicate->pred_string)
+			{
+			  continue;
+			}
+
 		      if (strcmp (filter_predicate->pred_string, cons->filter_predicate->pred_string))
 			{
 			  continue;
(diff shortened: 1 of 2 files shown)


mhoh3963 (approved):

src/object/class_object.c
```diff
@@ -4103,9 +4103,23 @@ classobj_find_constraint_by_attrs (SM_CLASS_CONSTRAINT * cons_list, DB_CONSTRAIN
...
+		    {
+		      if (!cons->filter_predicate)
+			{
+			  continue;
+			}
+
+		      if (strcmp (filter_predicate->pred_string, cons->filter_predicate->pred_string))
+			{
+			  continue;
+			}
```
hgryoo, line 4121: NULL check for pred_string is required.

hgryoo (approved):

ctshim (approved):

hyunikn (approved):

src/object/class_object.c
beyondykk9: the "pred_string" is set only when the filter_predicate is set, therefore pred_string can not be null in this case.
but I put the code to avoid compare with null string.

youngjinj (approved):

shparkcubrid (approved):

PR #3277, opened 2021-12-27
[CBRD-24141] Changed SQL keyword to be usable as table name and column name

ctshim:
http://jira.cubrid.org/browse/CBRD-24141

Changes are made so that they can be used as table names, column names, etc. using keywords newly added in version 10.0 or later. (Make sure it can be used as an identifier.)

However, the following cases are excluded.
1) Keyword indicating data type
2) DBTIMEZONE
3) SESSIONTIMEZONE
4) TIMEZONES
5) NONE

### DIFF
--- a/src/parser/csql_grammar.y
+++ b/src/parser/csql_grammar.y
@@ -1251,6 +1251,7 @@ int g_original_buffer_len;
 %token SEQUENCE_OF
 %token SERIALIZABLE
 %token SESSION
+%token SESSION_TIMEZONE
 %token SESSION_USER
 %token SET
 %token SET_OF
@@ -1563,7 +1564,6 @@ int g_original_buffer_len;
 %token <cptr> REGEXP_LIKE
 %token <cptr> REGEXP_REPLACE
 %token <cptr> REGEXP_SUBSTR
-%token <cptr> SESSION_TIMEZONE
 %token <cptr> TIMEZONE
 
 %token <cptr> IdName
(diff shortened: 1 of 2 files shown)


hyunikn (approved):

src/parser/csql_grammar.y
```diff
@@ -21703,1715 +21695,196 @@ identifier
...
-
-		DBG_PRINT}}
-	| TEXT
-		{{
-
-			PT_NODE *p = parser_new_node (this_parser, PT_NAME);
-			if (p)
-			  p->info.name.original = $1;
-			$$ = p;
-
```
hyunikn, line 23004: TEXT does not have PARSE_SAVE_ERR_CONTEXT(...). Is applying SET_CPTR_2_PTNAME(...) to TEXT OK?

beyondykk9 (approved):

hgryoo (approved):

src/parser/csql_grammar.y
ctshim: No problem.
Rather, it is reasonable to assume that "PARSER_SAVE_ERR_CONTEXT ($$, @$.buffer_pos)" was missing.

hyunikn: OK

shparkcubrid (approved):

youngjinj (approved):

PR #3278, opened 2021-12-28
[CBRD-24148] Core dump occur when using hash list scan

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-24148

It is a core dump caused by the problem that the result is located in the middle of duplicate data during binary search.
In order to be located on the right side of the duplicate value, flag is added to middle.

### DIFF
--- a/src/query/query_hash_scan.c
+++ b/src/query/query_hash_scan.c
@@ -73,7 +73,7 @@ static DB_VALUE_COMPARE_RESULT qdata_hscan_key_compare (HASH_SCAN_KEY * ckey1, H
 typedef unsigned int FHS_HASH_KEY;
 #define FHS_KEY_SIZE (sizeof (FHS_HASH_KEY))
 #define FHS_ALIGNMENT ((char) FHS_KEY_SIZE)
-#define FHS_MAX_DUP_KEY 3	/* 10% of MAXNUM (PAGE 16K / RECORD 14 bytes) */
+#define FHS_MAX_DUP_KEY 100	/* 10% of MAXNUM (PAGE 16K / RECORD 14 bytes) */
 				/* TO_DO : adjust it properly using statistical information in optimizer */
 
 #define SET_VPID(dest_vpid, vol_id, page_id)  \
--- a/src/query/query_hash_scan.c
+++ b/src/query/query_hash_scan.c
@@ -1924,7 +1924,7 @@ fhs_binary_search_bucket (THREAD_ENTRY * thread_p, PAGE_PTR bucket_page_p, PGSLO
 	}
       else
 	{
-	  middle += (flag > 0) ? flag - 1 : 0; /* go to last slot using flag */
+	  middle += (flag > 0) ? flag - 1 : 0;	/* go to last slot using flag */
 	  low = middle + 1;
 	}
     }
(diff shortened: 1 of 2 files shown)


src/query/query_hash_scan.c
```diff
@@ -1924,7 +1924,12 @@ fhs_binary_search_bucket (THREAD_ENTRY * thread_p, PAGE_PTR bucket_page_p, PGSLO
 	}
       else
 	{
-	  low = (flag > 0) ? middle + flag : middle + 1;
+	  if (flag > 0)
+	    {
+	      /* go to last slot using flag */
+	      middle = middle + flag - 1;
+	    }
+	  low = middle + 1;
```
hgryoo:
```suggestion
	  middle += (flag > 0) ? flag - 1 : 0; /* go to last slot using flag */
	  low = middle + 1;
```
the same code format as before

hgryoo (approved):

src/query/query_hash_scan.c
```diff
@@ -1924,7 +1924,8 @@ fhs_binary_search_bucket (THREAD_ENTRY * thread_p, PAGE_PTR bucket_page_p, PGSLO
 	}
       else
 	{
-	  low = (flag > 0) ? middle + flag : middle + 1;
+	  middle += (flag > 0) ? flag - 1 : 0; /* go to last slot using flag */
+	  low = middle + 1;
```
hyunikn, line 1928: The variable low will have the same value after this change as before regardless of flag > 0 or not because the earlier code is run as "low = (flag > 0) ? (middle + flag) : (middle + 1);"
Then, is it the real point of this change to increase middle by (flag - 1) when (flag > 0) ?

youngjinj (approved):

src/query/query_hash_scan.c
beyondykk9: It looks like the logic to set the **middle** value, keeping the **low** value as before.
If that's the intention, it's the right code.

shparkcubrid: The value of middle becomes the final return value. The problem occurred because middle is not changed.

src/query/query_hash_scan.c
```diff
@@ -1924,7 +1924,8 @@ fhs_binary_search_bucket (THREAD_ENTRY * thread_p, PAGE_PTR bucket_page_p, PGSLO
 	}
       else
 	{
-	  low = (flag > 0) ? middle + flag : middle + 1;
+	  middle += (flag > 0) ? flag - 1 : 0; /* go to last slot using flag */
```
shparkcubrid:
```suggestion
	  middle += (flag > 0) ? flag - 1 : 0;	/* go to last slot using flag */
```

src/query/query_hash_scan.c
hyunikn: OK. I got it.

hyunikn (approved):

src/query/query_hash_scan.c
ctshim: If (compare_result < 0) does not need to adjust the middle value?

ctshim (approved):

src/query/query_hash_scan.c
shparkcubrid: The flag only can be used when moving to the end.

beyondykk9 (approved):

PR #3282, opened 2021-12-29
[LETS-308] Handle page ahead replication after btree_range_scan navigation

ribram:
http://jira.cubrid.org/browse/LETS-308

Page ahead of replication errors may show up during b-tree range scan operations. This patch handles the errors after "navigation" functions (while advancing from root to current key, while reading overflow keys or while advancing from a leaf page to the next (or prev)).

The page ahead of replication problem is handled by waiting for the replication to catch-up with the page changes. If OID's that may be processed have been selected, the btree_range_scan last iteration is ended to process the OID's while replication catches up.

EDIT: the page ahead of replication error that is discovered while selecting OID's from b-tree keys is handled separately in LETS_309.

### DIFF
--- a/src/storage/btree.c
+++ b/src/storage/btree.c
@@ -41,6 +41,7 @@
 #include "utility.h"
 #include "transform.h"
 #include "partition_sr.h"
+#include "passive_tran_server.hpp"
 #include "porting_inline.hpp"
 #include "query_executor.h"
 #include "query_opfunc.h"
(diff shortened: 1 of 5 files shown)


src/storage/btree.c
```diff
@@ -24894,6 +24899,63 @@ btree_range_scan_descending_fix_prev_leaf (THREAD_ENTRY * thread_p, BTREE_SCAN *
...
+  er_clear ();
+
+  bts.force_restart_from_root = true;
+
+  // If there are objects selected, end this iteration and maybe replication will catch-up in the meantime.
+  // If MRO is used or if only counting is required, ending an iteration will not help.
+  //
+  if (bts.n_oids_read_last_iteration > 0 && !BTS_IS_INDEX_MRO (&bts) && !BTS_NEED_COUNT_ONLY (&bts))
+    {
+      bts.end_one_iteration;
```
cristiarg, line 24923: `bts.end_one_iteration;` what?

src/storage/btree.c
```diff
@@ -13569,7 +13572,7 @@ btree_split_root (THREAD_ENTRY * thread_p, BTID_INT * btid, PAGE_PTR P, PAGE_PTR
     }
 
 #if !defined(NDEBUG)
-  node_level = btree_get_node_level (thread_p, P);
+  int node_level = btree_get_node_level (thread_p, P);
```
cristiarg, line 13575: hide in block context if `node_level` is not used further

razvanradulescu (approved):

PR #3283, opened 2021-12-30
[CBRD-24149] Change the comparison spec for char-type and varchar-type.

beyondykk9:
http://jira.cubrid.org/browse/CBRD-24149

Since version 11.0 in char string comparison spec is as below:
    char-type vs. char-type => char-type comparison
    char-type vs. varchar-type => varchar-type comparison
    * commonly char-type string's trailing space is trimmed first before comparing.

In the index search, however, the above comparison spec is not  applied. The char-type and varchar-type comparison is done as like char-type and char-type (ignores trailing space).


ctshim:
Is it unnecessary to trim separately in the mr_cmpval_string() and mr_cmpval_char() functions, and is it solved by appropriately adjusting the ti parameter when calling QSTR_COMPARE()?

Even so, isn't the code below necessary in the mr_cmpval_char() function?
```
if (type1 != DB_TYPE_CHAR && type1 != DB_TYPE_NCHAR && type2 != DB_TYPE_CHAR && type2 != DB_TYPE_NCHAR)
   {
        ti = false;
    }
```

shparkcubrid:
I commented a test case that caused the problem in jira.

src/object/object_primitive.c
```diff
@@ -11134,33 +11134,10 @@ mr_cmpval_string (DB_VALUE * value1, DB_VALUE * value2, int do_coercion, int tot
 
   if (!ignore_trailing_space)
     {
-      int i;
-
-      if (type1 == DB_TYPE_CHAR || type1 == DB_TYPE_NCHAR)
+      if (type1 != DB_TYPE_CHAR && type1 != DB_TYPE_NCHAR && type2 != DB_TYPE_CHAR && type2 != DB_TYPE_NCHAR)
```
hgryoo:
```suggestion
      if (!TP_IS_FIXED_LEN_CHAR_TYPE (type1) && !TP_IS_FIXED_LEN_CHAR_TYPE (type2))
```

hyunikn (approved):

src/parser/csql_grammar.y
```diff
@@ -27903,6 +27903,7 @@ pt_create_char_string_literal (PARSER_CONTEXT *parser, const PT_TYPE_ENUM char_t
         length = node->info.value.data_value.str->length;
 
         node->type_enum = char_type;
+/*
```
ctshim: In the query below, the tailing space of a varchar column is treated like a char column. Is your intention correct?
`select * from tbl where varchar_col = 'abc  '`

beyondykk9: yes, the policy of varchar-type constant is changed.

src/object/object_primitive.c
beyondykk9: I agree with you.

src/parser/csql_grammar.y
ctshim: It would be nice if you could write in a comment that you would like to comment on the policy change.

ctshim (approved):

hyunikn (commented):
I heard that Cubrid regards a text literal as a varchar if it ends with trailing spaces and as a char if it does not, while Oracle always regards it as a char. Do we keep this policy unchanged?

src/parser/csql_grammar.y
beyondykk9: Ok, I commented at JIRA as you said.

src/object/object_primitive.c
```diff
@@ -11132,35 +11132,9 @@ mr_cmpval_string (DB_VALUE * value1, DB_VALUE * value2, int do_coercion, int tot
...
-	      if (string2[i - 1] != 0x20)
-		{
-		  break;
-		}
-	    }
-	  size2 = i;
-	}
-
-      ti = false;
+      ti = true;
```
hyunikn: We can now remove ti and use ignore_trailing_space instead.

beyondykk9: **ignore_trailing_space** is a static global and it is set only once, however, the variable **ti** is set whenever this routine is called. Therefore **ti** has a role to apply comparison method.

src/object/object_primitive.c
```diff
@@ -12015,55 +11989,20 @@ mr_cmpval_char (DB_VALUE * value1, DB_VALUE * value2, int do_coercion, int total
...
-    }
-  else if (!ignore_trailing_space &&
-	   (type1 == DB_TYPE_STRING || type1 == DB_TYPE_VARNCHAR || type2 == DB_TYPE_STRING
-	    || type2 == DB_TYPE_VARNCHAR))
+  if (!ignore_trailing_space)
     {
-      int i;
-
-      if (type1 == DB_TYPE_CHAR || type1 == DB_TYPE_NCHAR)
+      if (do_coercion == 2)
```
hyunikn:
```suggestion
      if (do_coercion != 2 && 
          (!TP_IS_FIXED_LEN_CHAR_TYPE (type1) || !TP_IS_FIXED_LEN_CHAR_TYPE (type2)))
        {
          ti = false;
        }
```

src/object/object_primitive.c
```diff
@@ -12015,55 +11989,20 @@ mr_cmpval_char (DB_VALUE * value1, DB_VALUE * value2, int do_coercion, int total
...
 	{
-	  for (i = size1; i > 1; i--)
+	  /* only for btree_get_prefix_sparator */
+	  if (TP_IS_FIXED_LEN_CHAR_TYPE (type1) || TP_IS_FIXED_LEN_CHAR_TYPE (type2))
 	    {
-	      if (string1[i - 1] != 0x20)
-		{
-		  break;
-		}
+	      ti = true;
```
hyunikn: ti is already true, and hence the above suggestion.

src/object/object_primitive.c
```diff
@@ -12963,55 +12902,20 @@ mr_cmpval_nchar (DB_VALUE * value1, DB_VALUE * value2, int do_coercion, int tota
...
-   * we need to process enforcing no-ignore-trailing space.
-   */
-  else if (do_coercion == 3)
-    {
-      ti = false;
-    }
-  else if (!ignore_trailing_space &&
-	   (type1 == DB_TYPE_STRING || type1 == DB_TYPE_VARNCHAR || type2 == DB_TYPE_STRING
-	    || type2 == DB_TYPE_VARNCHAR))
+  if (!ignore_trailing_space)
```
hyunikn, line 12898: This if block can be erased because ti is already true.

src/object/object_primitive.c
hyunikn: ti always has the same value as ignore_trailing_space at line 11140, and there is no other use of ti.

src/object/object_primitive.c
beyondykk9: I understand your comment, but I keep the code and comment for readability.

src/object/object_primitive.c
beyondykk9: Ok, I refactored the code as your comment.

PR #3290, opened 2022-01-04
[CBRD-24152] Improved readability of qexec_intprt_fnc() function

ctshim:
http://jira.cubrid.org/browse/CBRD-24152

qexec_intprt_fnc(), qexec_merge_fnc() The common parts of the two functions are rewritten as separate functions to improve code readability.
Removed some unnecessary variable usage.

Also, in the qexec_execute_scan() function, unnecessary if statement checks are removed and cleaned up.


src/query/query_executor.c
```diff
@@ -7919,11 +8016,10 @@ qexec_intprt_fnc (THREAD_ENTRY * thread_p, XASL_NODE * xasl, XASL_STATE * xasl_s
...
 	      if (cte_start_new_iteration)
 		{
-		  recursive_iterations++;
-		  if (recursive_iterations >= xasl->max_iterations)
+		  if (++recursive_iterations >= xasl->max_iterations)
 		    {
-		      max_recursive_iterations_reached = true;
-		      break;
+		      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_CTE_MAX_RECURSION_REACHED, 1, xasl->max_iterations);
+		      return S_ERROR;
```
shparkcubrid, line 7990: is it okay that return S_ERROR instead of break?

ctshim: When a break is encountered in the current code, er_set() is executed at line 8190, and eventually “return S_ERROR;” at line 8205 is encountered.

src/query/query_executor.c
```diff
@@ -692,6 +692,11 @@ static int qexec_locate_agg_hentry_in_list (THREAD_ENTRY * thread_p, AGGREGATE_H
 					    AGGREGATE_HASH_KEY * key, bool * found);
 static int qexec_get_attr_default (THREAD_ENTRY * thread_p, OR_ATTRIBUTE * attr, DB_VALUE * default_val);
 
+static inline SCAN_CODE evaluate_xptr_list (THREAD_ENTRY * thread_p, XASL_STATE * xasl_state, XASL_NODE * xptr_list,
+					    SCAN_OPERATION_TYPE scan_operation_type, bool * keep_going);
+static inline SCAN_CODE evaluate_one_scanned_tuple (THREAD_ENTRY * thread_p, XASL_NODE * xasl, XASL_STATE * xasl_state,
+						    QFILE_TUPLE_RECORD * tplrec, bool * keep_going);
+
```
hgryoo: Can I know why these functions are inlined? I think these functions are too long, so the inline keyword is going to be ignored.

ctshim: Even if inline is specified, it may be excluded by the compiler.

src/query/query_executor.c
```diff
@@ -7792,6 +7797,99 @@ qexec_init_next_partition (THREAD_ENTRY * thread_p, ACCESS_SPEC_TYPE * spec)
...
+	      return S_SUCCESS;
+	    }
+
+	  if ((xasl->instnum_flag & XASL_INSTNUM_FLAG_SCAN_STOP))
+	    {
+	      return S_SUCCESS;
+	    }
+	}
+
+      if (xasl->instnum_pred == NULL || ev_res == V_TRUE)
```
hyunikn, line 7843: ev_res has not been initialized at this line if xasl->instnum_val is falsy.
Seeing the current code, it should be initialized with V_UNKNOWN at the declaration.

hyunikn: Similar problem seems to exist also in the current code.
For example, reading ev_res at line 8062 (develop branch) can have a value set at line 7969 or 7983
which seem very far from line 8062 and suspicious.

shparkcubrid:
I think performance is more important than readability in executor. In that regard, it seems good to keep the old source code style. Personally, I think the 'qualified' variable is more suitable than 'keep_going'.

youngjinj (approved):

youngjinj:
When looking only at the **evaluate_xptr_list** and **evaluate_one_scanned_tuple** functions, it is difficult to understand the necessity of **keep_going**.

I think it would be nice to use something other than **keep_going** variables.

src/query/query_executor.c
ctshim: I will re check it.

hyunikn (approved):

beyondykk9 (approved):

youngjinj (approved):

hyunikn (approved):

beyondykk9 (approved):

src/query/query_executor.c
```diff
@@ -7261,156 +7265,130 @@ qexec_execute_scan (THREAD_ENTRY * thread_p, XASL_NODE * xasl, XASL_STATE * xasl
...
+	  for (xptr = xasl->bptr_list; xptr != NULL; xptr = xptr->next)
 	    {
 	      if (qexec_execute_obj_fetch (thread_p, xptr, xasl_state, scan_operation_type) != NO_ERROR)
 		{
 		  return S_ERROR;
 		}
 	      else if (xptr->proc.fetch.fetch_res == false)
 		{
-		  qualified = false;
+		  goto try_scan_next_scan;
```
shparkcubrid, line 7291: I disagree that using the goto statement is better for readability.

ctshim: In general, goto statements reduce readability.
But wouldn't this be an efficient method for sources that are long due to many repeated and nested if() blocks?

ctshim:
It has been left unattended for a long time and is dropped without additional work.

PR #3293, opened 2022-01-05
[LETS-380] correct make_unique char array

cristiarg:
http://jira.cubrid.org/browse/LETS-380

Regression.
Went under the radar during a recent commit.

### DIFF
--- a/src/storage/page_buffer.c
+++ b/src/storage/page_buffer.c
@@ -8140,7 +8140,7 @@ pgbuf_read_page_from_file_or_page_server (THREAD_ENTRY * thread_p, const VPID * 
       if (read_from_local)
 	{
 	  // *INDENT-OFF*
-	  auto buffer_uptr = std::make_unique<char> (db_io_page_size ());
+	  auto buffer_uptr = std::make_unique<char []> (db_io_page_size ());
 	  FILEIO_PAGE *second_io_page = reinterpret_cast<FILEIO_PAGE *> (buffer_uptr.get ());
 	  error_code = pgbuf_request_data_page_from_page_server (vpid, target_repl_lsa, second_io_page);
 	  if (error_code != NO_ERROR)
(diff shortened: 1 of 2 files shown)


src/transaction/log_page_buffer.c
```diff
@@ -2250,7 +2250,7 @@ logpb_read_page_from_file_or_page_server (THREAD_ENTRY * thread_p, LOG_PAGEID pa
 	{
 	  // context 2)
 	  // *INDENT-OFF*
-	  auto log_page_buffer_uptr = std::make_unique<char> (IO_MAX_PAGE_SIZE);
+	  auto log_page_buffer_uptr = std::make_unique<char []> (IO_MAX_PAGE_SIZE);
```
razvanradulescu: Nit :
Use db_log_page_size () or LOG_PAGESIZE

src/storage/page_buffer.c
```diff
@@ -8140,7 +8140,7 @@ pgbuf_read_page_from_file_or_page_server (THREAD_ENTRY * thread_p, const VPID *
       if (read_from_local)
 	{
 	  // *INDENT-OFF*
-	  auto buffer_uptr = std::make_unique<char> (db_io_page_size ());
+	  auto buffer_uptr = std::make_unique<char []> (db_io_page_size ());
```
razvanradulescu: This style with auto looks dangerous:
According to form (2), if return type of db_io_page_size is size_t, then you would get the behavior expected from previous version.
What if we change the signature of db_io_page_size () in the future ?

https://docs.w3cub.com/cpp/memory/unique_ptr/make_unique

razvanradulescu (approved):

src/storage/page_buffer.c
cristiarg: You meant:
```
According to form (2), if return type of db_io_page_size is _NOT_ size_t, then you would get the behavior expected from previous version.
```
You are correct.

cristiarg: Changed the code to be more explicit.

src/transaction/log_page_buffer.c
cristiarg: Done. Changed the code to be more explicit.

razvanradulescu (approved):

PR #3294, opened 2022-01-05
[LETS-379] more accurate comments regarding transaction table checkpoint

cristiarg:
http://jira.cubrid.org/browse/LETS-379

Extra:
When adding a checkpoint to the metalog, first check that a same-LSA checkpoint already exists.
Technically, this should never happen:
- on active transaction server (ATS), every checkpoint created is also added to the log before being added to the metalog, thus increasing the LSA
- on page server (PS), since page server consumes (replicates) the log generated by the active transaction server, the same pattern could happen there as well; the edge case, in which the page server could add a same-LSA again in the metalog, is if the checkpoint daemon has such a short execution interval that it executes two times consecutively between two log records created by the ATS

### DIFF
--- a/src/transaction/log_meta.cpp
+++ b/src/transaction/log_meta.cpp
@@ -178,7 +178,14 @@ namespace cublog
   void
   meta::add_checkpoint_info (const log_lsa &chkpt_lsa, const checkpoint_info &chkpt_info)
   {
-    assert (m_checkpoints.find (chkpt_lsa) == m_checkpoints.cend ());
+    const checkpoint_container_t::const_iterator found_it = m_checkpoints.find (chkpt_lsa);
+    if (found_it != m_checkpoints.cend ())
+      {
+	// if same LSA checkpoint already exists, replace with new one
+	// to be in line with what happens in calling code where the checkpoint is also
+	// added to the log for the purpose of being transferred to passive transaction servers
+	m_checkpoints.erase (found_it);
+      }
     m_checkpoints.insert ({ chkpt_lsa, chkpt_info });
   }
 
--- a/src/transaction/log_meta.cpp
+++ b/src/transaction/log_meta.cpp
@@ -178,6 +178,7 @@ namespace cublog
   void
   meta::add_checkpoint_info (const log_lsa &chkpt_lsa, const checkpoint_info &chkpt_info)
   {
+    assert (m_checkpoints.find (chkpt_lsa) == m_checkpoints.cend ());
     m_checkpoints.insert ({ chkpt_lsa, chkpt_info });
   }
 
--- a/src/transaction/log_meta.hpp
+++ b/src/transaction/log_meta.hpp
@@ -72,12 +72,18 @@ namespace cublog
        */
       bool m_clean_shutdown = false;
 
-      /* as the system is designed, it is not needed to hold a map of checkpoints since there should
-       * be, at most, 2 checkpoints:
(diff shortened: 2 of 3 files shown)


razvanradulescu (approved):

src/transaction/log_meta.cpp
```diff
@@ -178,6 +178,7 @@ namespace cublog
   void
   meta::add_checkpoint_info (const log_lsa &chkpt_lsa, const checkpoint_info &chkpt_info)
   {
+    assert (m_checkpoints.find (chkpt_lsa) == m_checkpoints.cend ());
```
wrlawodms: If no log records are added during two checkpoints, it seems that `chkpt_lsa` could be in `m_checkpoints`. This can't happen?

cristiarg: `log_Checkpoint_trantable_daemon` is set to execute every 60 seconds (`log_checkpoint_trantable_daemon_init`).
I know that there are status/dummy log records which are generated more often (eg: `LOG_DUMMY_HA_SERVER_STATE`) - which replication also uses to piggyback replication delay statistics calculation on page server and, in the future, passive transaction server.

Later edit: Might it be safer to just, if a same LSA checkpoint is found already in the map, replace with the new one?

wrlawodms: > log_Checkpoint_trantable_daemon is set to execute every 60 seconds (log_checkpoint_trantable_daemon_init).

In ATS, it doesn't seem to be a problem because it generates the tran-table snapshot log record together with adding a checkping info in the meta log, which makes LSA advanced. Where I think it could matter is in PS. 

> know that there are status/dummy log records which are generated more often (eg: LOG_DUMMY_HA_SERVER_STATE) - which replication also uses to piggyback replication delay statistics calculation on page server and, in the future, passive transaction server.
Later edit: Might it be safer to just, if a same LSA checkpoint is found already in the map, replace with the new one?

I suppose it is better to decouple what generates logs and the checkpoint/log mechanism. Your suggestion later edited looks better to me if the situation I mentioned can occur :).

wrlawodms: Or can we just skip adding checkpoint? No record after the previous checkpoint means no need to do a checkpoint?

cristiarg: PS's do not generate log and, I think, do not perform checkpoints.
On ATS, indeed, a log record is added prior to saving the checkpoint in meta log.
To be congruent with this addition of checkpoint as a log record, I replaced the checkpoint in metalog if a same-LSA checkpoint is already found there.

wrlawodms: I'd like to correct If I have misunderstood. Could you let me know the followings are correct?
1. On PS, no log records are generated, but checkpoint is performed in logpb_checkpoint(), in which checkpoint info is added in the meta log.
2. On ATS, there is no chance to get the same LSA as the argument in add_checkpoint_info() because a log record is added prior to saving the checkpoint in meta log as you said, and it will advance the snapshot lsa of the checkpoint info in the next time.

wrlawodms (approved):

hornetmj (approved):

src/transaction/log_meta.cpp
cristiarg: 1. You are correct. I was mistaken. Checkpoint daemon does execute on PS.
2. correct again

wrlawodms: Thank you for the answer.

PR #3295, opened 2022-01-06
[LETS-286] Build PTS starting MVCC table based on transaction table

cristiarg:
http://jira.cubrid.org/browse/LETS-286

On passive transaction server, having the starting transaction table, build the starting MVCC table:
- set log_Gl.hdr.mvcc_next_id to the smallest MVCCID in transaction table and call log_Gl.mvcc_table.reset_start_mvccid ()
- For each MVCCID between smallest MVCCID in transaction table and highest MVCCID in transaction table that is missing from the table, call log_Gl.mvcc_table.complete_mvccid ()
- starting_point for MVCC table is the minimum MVCCID from transaction table
- log_Gl.hdr.next_mvccid is the maximum MVCCID from transaction table + 1
- any MVCCID between minimum and maximum MVCCID in transaction table, that is not in transaction table becomes inactive (completed)
- other MVCCID in transaction table remain active and will be completed by subsequent commits/aborts processed during replication


src/transaction/log_recovery_analysis.cpp
```diff
@@ -2204,35 +2210,19 @@ corruption_checker::check_log_record (const log_lsa &record_lsa, const log_rec_h
...
- *                  the active transaction server
- *
- * most_recent_trantable_snapshot_lsa (in): the lsa where a record containing, as payload, the packed contents
- *                                of a recent transaction table snapshot; starting from that snapshot, analyze the log
- *                                to construct an actual starting transaction table
- */
-void
-log_recovery_analysis_from_transaction_table_snapshot (THREAD_ENTRY *thread_p,
-    log_lsa most_recent_trantable_snapshot_lsa)
+static int
```
razvanradulescu, line 2216: A comment header to describe the usage would be useful

src/transaction/log_recovery_analysis.cpp
```diff
@@ -2241,20 +2231,105 @@ log_recovery_analysis_from_transaction_table_snapshot (THREAD_ENTRY *thread_p,
...
+		}
+	      if (tdes->mvccinfo.id > largest_mvccid)
+		{
+		  largest_mvccid = tdes->mvccinfo.id;
+		}
+	      present_mvccids.insert (tdes->mvccinfo.id);
+	    }
+	}
+    }
+  log_Gl.hdr.mvcc_next_id = smallest_mvccid;
```
razvanradulescu, line 2280: Do we need to keep log_Gl.hdr.mvcc_next_id == smallest_mvccid during the loop which "commits" missing MVCCIDs or is needed only for reset_start_mvccid () ?

razvanradulescu (approved):

src/transaction/log_recovery_analysis.cpp
cristiarg: Done

src/transaction/log_recovery_analysis.cpp
cristiarg: It is needed for `mvcctable::reset_start_mvccid`. But I am not sure about the order in relation with "committing" missing MVCCID's.

PR #3297, opened 2022-01-07
[CBRD-24155] Add a system parameter to set core count for thread worker pool

hgryoo:
http://jira.cubrid.org/browse/CBRD-24155

- a new parameter, "thread_core_count" is added
  - Change (Tune) the value entered by the user considering max_client and system core count
  - The behavior of partitioning workers according to the # of cores in the thread worker pool is changed to use "thread_core_count" as partitioning factor

please refer to the JIRA description for detailed implementation.

### DIFF
--- a/src/base/system_parameter.c
+++ b/src/base/system_parameter.c
@@ -2380,13 +2380,14 @@ static UINT64 prm_first_log_pageid_lower = 0LL;
 static UINT64 prm_first_log_pageid_upper = LOGPAGEID_MAX;
 static unsigned int prm_first_log_pageid_flag = 0;
 
-static const int lower_bound_core_count = (prm_css_max_clients_default / 3);
-static const int upper_bound_core_count = (prm_css_max_clients_upper / 3);
-
-int PRM_THREAD_CORE_COUNT = lower_bound_core_count;	// this value will be tuned
-static int prm_thread_core_count_default = lower_bound_core_count;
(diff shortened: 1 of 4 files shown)


src/base/system_parameter.c
```diff
@@ -2377,6 +2380,15 @@ static UINT64 prm_first_log_pageid_lower = 0LL;
 static UINT64 prm_first_log_pageid_upper = LOGPAGEID_MAX;
 static unsigned int prm_first_log_pageid_flag = 0;
 
+static const int lower_bound_core_count = (prm_css_max_clients_default / 3);
```
mhoh3963, line 2383: How about use prm_css_max_clients_lower in stead of prm_css_max_clients_default ?

beyondykk9 (approved):

beyondykk9 (approved):

src/base/system_parameter.c
```diff
@@ -10711,6 +10749,7 @@ prm_tune_parameters (void)
   SYSPRM_PARAM *ha_check_disk_failure_interval_prm;
   SYSPRM_PARAM *test_mode_prm;
   SYSPRM_PARAM *tz_leap_second_support_prm;
+  SYSPRM_PARAM *thread_core_count_prm;
```
hgryoo:
```suggestion
```

PR #3298, opened 2022-01-07
[CBRD-23844] Support User Schema

youngjinj:
http://jira.cubrid.org/browse/CBRD-23844

User schema logically categorizes tables, views, indexes, serials, triggers, functions and procedures created by user accounts.
There is a strict distinction between user accounts and user schema, but if only one user schema can be used by one user account, then the user account and user schema are the same.


youngjinj:
Close by changing the source branch.

PR #3271, opened 2021-12-22
[LETS-307] Check page desynchronization of b-tree overflow OID

adrianbrab:
http://jira.cubrid.org/browse/LETS-307

Changed calls to `pgbuf_fix` for overflow OID pages to `pgbuf_fix_read_old_and_check_repl_desync` to also check for desync.
The functions changed are:

-     btree_get_num_visible_oids_from_all_ovf
-     btree_find_oid_and_its_page
-     btree_key_find_first_visible_row_from_all_ovf
-     btree_key_process_objects
-     btree_range_scan_select_visible_oids
-     btree_range_scan_find_fk_any_object

PR #3299, opened 2022-01-07
[LETS-381] regression, correct pgbuf_fix latch mode

cristiarg:
http://jira.cubrid.org/browse/LETS-381

Fallout from #3271 - http://jira.cubrid.org/browse/LETS-307

In function btree_find_oid_and_its_page page, `overflow_vpid` page was previously fixed with PGBUF_LATCH_WRITE which previous patch changed to PGBUF_LATCH_READ by means of `pgbuf_fix_read_old_and_check_repl_desync` implementation.
Refactor to make latch mode explicit as argument for function pgbuf_fix_old_and_check_repl_desync and adapted callers.
Corrected minor typo.

### DIFF
--- a/src/storage/btree.c
+++ b/src/storage/btree.c
@@ -2497,7 +2497,8 @@ btree_get_num_visible_oids_from_all_ovf (THREAD_ENTRY * thread_p, BTID_INT * bti
   /* search for OID into overflow page */
   while (!VPID_ISNULL (&next_ovfl_vpid))
     {
-      ovfl_page = pgbuf_fix_read_old_and_check_repl_desync (thread_p, next_ovfl_vpid, PGBUF_UNCONDITIONAL_LATCH);
+      ovfl_page = pgbuf_fix_old_and_check_repl_desync (thread_p, next_ovfl_vpid, PGBUF_LATCH_READ,
+						       PGBUF_UNCONDITIONAL_LATCH);
       if (ovfl_page == NULL)
 	{
 	  ASSERT_ERROR_AND_SET (ret);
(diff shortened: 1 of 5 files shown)


src/storage/btree.c
```diff
@@ -11137,7 +11138,8 @@ btree_find_oid_and_its_page (THREAD_ENTRY * thread_p, BTID_INT * btid_int, OID *
   do
     {
       PERF_UTIME_TRACKER_START (thread_p, &ovf_fix_time_track);
-      overflow_page = pgbuf_fix_read_old_and_check_repl_desync (thread_p, overflow_vpid, PGBUF_UNCONDITIONAL_LATCH);
+      overflow_page = pgbuf_fix_old_and_check_repl_desync (thread_p, overflow_vpid, PGBUF_LATCH_WRITE,
```
razvanradulescu, line 11141: As far as I understand, the purpose of 'pgbuf_fix_read_old_and_check_repl_desync' function was to be run only on PTS for online transactions (which should all be read-only).
Do we want to adapt  'pgbuf_fix_read_old_and_check_repl_desync' to be used for replication (or vacuum after replication) purposes which is the only context where PTS performs writes ?

What if in context of vacuum, the error   ER_PAGE_AHEAD_OF_REPLICATION is ignored, and the page is not vacuumed at all ?

cristiarg: @razvanradulescu 
I do not think that `pgbuf_fix_read_old_and_check_repl_desync` was meant to be only called on PTS - since in https://github.com/CUBRID/cubrid/pull/3271/files it is in-plare replacement of `pgbuf_fix`.
I think that whether read-write or read-only transactions are allowed on PTS is controlled from a higher layer in the decision stack (somewhere at the level of parsing a query maybe?).
Therefore, these `pgbuf_fix_old_and_check_repl_desync` will be called on both PTS and ATS - for whatever purpose - the forking between ATS and PTS is done inside the function. I am not able to answer myself why they are not called on the PS as well - LE: maybe because there are no queries being run on PS, therefore no index searches?.
As for the vacuum part, I do not have an answer.

razvanradulescu: After checking code, the WRITE latch in btree_find_oid_and_its_page  is correct.
As a general rule, the presence of BTREE_OP_PURPOSE  arguments suggests that function is called in a context where at least one page needs to be WRITE latched.
This PR is not concerned with refactoring/optimization of page latching, so it is enough.

razvanradulescu (approved):

PR #3300, opened 2022-01-10
[CBRD-24160] Changed or_class_name function to unpack names longer than 255.

youngjinj:
http://jira.cubrid.org/browse/CBRD-24160

Purpose
- When the or_class_name function gets class_name from a record, it decompresses it if it is compressed.

Implementation
- When the or_class_name function gets class_name from a record, it decompresses it if it is compressed.
- If class_name exceeds 255 characters, dynamic memory release is required after calling the or_class_name function.

Remarks
- N/A

### DIFF
--- a/src/object/object_representation.c
+++ b/src/object/object_representation.c
@@ -316,7 +316,7 @@ or_class_name (RECDES * record, char **string, int *alloced_string)
    * Knowledge of the format of packed varchars should be in a different
    * or_ function.
    */
-  len = (int) * ((unsigned char *) start);
+  len = (int) *((unsigned char *) start);
   if (len < OR_MINIMUM_STRING_LENGTH_FOR_COMPRESSION)
     {
       *string = start + 1;
--- a/src/object/object_representation.c
+++ b/src/object/object_representation.c
@@ -293,6 +293,9 @@ or_class_name (RECDES * record, char **string, int *alloced_string)
   int decompressed_length = 0;
   int rc = NO_ERROR;
 
+  assert (string != NULL);
+  assert (alloced_string != NULL);
+
   /* Initialization */
   *string = NULL;
   *alloced_string == 0;
--- a/src/object/object_representation.c
+++ b/src/object/object_representation.c
@@ -293,7 +293,9 @@ or_class_name (RECDES * record, char **string, int *alloced_string)
   int decompressed_length = 0;
   int rc = NO_ERROR;
 
-  assert (*alloced_string == 0);
+  /* Initialization */
+  *string = NULL;
+  *alloced_string == 0;
 
   /*
    * the first variable attribute for both classes and the rootclass
@@ -314,8 +316,8 @@ or_class_name (RECDES * record, char **string, int *alloced_string)
    * Knowledge of the format of packed varchars should be in a different
    * or_ function.
    */
-  len = (int) *((unsigned char *) start);
-  if (len != 0xFF)
+  len = (int) * ((unsigned char *) start);
+  if (len < OR_MINIMUM_STRING_LENGTH_FOR_COMPRESSION)
     {
       *string = start + 1;
     }
@@ -327,28 +329,25 @@ or_class_name (RECDES * record, char **string, int *alloced_string)
       if (rc != NO_ERROR)
 	{
 	  ASSERT_ERROR ();
-	  *string = NULL;
 	  return rc;
 	}
 
-      assert (*string == NULL);
       *string = (char *) db_private_alloc (NULL, decompressed_length + 1);
       if (*string == NULL)
 	{
 	  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, decompressed_length + 1);
 	  return ER_OUT_OF_VIRTUAL_MEMORY;
 	}
-      *alloced_string = 1;
 
       rc = pr_get_compressed_data_from_buffer (&buffer, *string, compressed_length, decompressed_length);
       if (rc != NO_ERROR)
 	{
 	  ASSERT_ERROR ();
-	  db_private_free (NULL, *string);
-	  *alloced_string = 0;
-	  *string = NULL;
+	  db_private_free_and_init (NULL, *string);
 	  return rc;
 	}
+
+      *alloced_string = 1;
     }
 
   return rc;
--- a/src/object/object_representation.c
+++ b/src/object/object_representation.c
@@ -281,11 +281,19 @@ int or_Type_sizes[] = {
  *    need to copy the string out of the record for some architectures
  *    if there are weird alignment or character set problems.
  */
-char *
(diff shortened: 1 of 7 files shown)


src/object/object_representation.c
```diff
@@ -309,14 +317,41 @@ or_class_name (RECDES * record)
   len = (int) *((unsigned char *) start);
   if (len != 0xFF)
     {
-      name = start + 1;
+      *string = start + 1;
```
ctshim, line 325: Looks like the following code is missing
`*alloced_string = 0;`

src/query/query_executor.c
```diff
@@ -21843,11 +21843,24 @@ qexec_execute_build_indexes (THREAD_ENTRY * thread_p, XASL_NODE * xasl, XASL_STA
...
+      ASSERT_ERROR ();
+      GOTO_EXIT_ON_ERROR;
+    }
+  class_name = string;
+
   for (i = 0; i < rep->n_indexes; i++)
     {
       /* class name */
       db_make_string (out_values[0], class_name);
+      if (string != NULL && alloced_string == 1)
```
beyondykk9: why is the index of out_values is always 0 ?

src/storage/statistics_sr.c
```diff
@@ -392,9 +394,22 @@ xstats_update_all_statistics (THREAD_ENTRY * thread_p, bool with_fullscan)
...
+      error = or_class_name (&recdes, &string, &alloced_string);
+      if (error != NO_ERROR)
+	{
+	  ASSERT_ERROR ();
+	  goto exit_on_error;
+	}
+      classname = string;
       assert (classname != NULL);
-      assert (strlen (classname) < 255);
+      assert (strlen (classname) < 255);	// to be: DB_MAX_FULL_CLASS_LENGTH
```
beyondykk9: A macro definition looks better to hard-coded 255.

src/object/object_representation.c
```diff
@@ -309,14 +317,41 @@ or_class_name (RECDES * record)
...
     }
   else
     {
-      name = start + 1 + OR_INT_SIZE;
+      OR_BUF_INIT (buffer, start, -1);
+
+      rc = or_get_varchar_compression_lengths (&buffer, &compressed_length, &decompressed_length);
```
hgryoo, line 331: If the string is not compressed, compresed_length and decompressed_length are set as 0 and 0xff respectively. It may depend on the "enable_string_compression" parameter.

src/query/query_executor.c
```diff
@@ -21843,11 +21843,24 @@ qexec_execute_build_indexes (THREAD_ENTRY * thread_p, XASL_NODE * xasl, XASL_STA
...
+	{
+	  out_values[0]->need_clear = true;
+	}
```
ctshim: You need to be prepared in case db_make_string() fails.
See db_make_string_copy().

src/storage/heap_file.c
```diff
@@ -2719,6 +2728,11 @@ heap_classrepr_dump (THREAD_ENTRY * thread_p, FILE * fp, const OID * class_oid,
 	   (repr->n_attributes - repr->n_variable - repr->n_shared_attrs - repr->n_class_attrs), repr->n_variable,
 	   repr->n_shared_attrs, repr->n_class_attrs, repr->fixed_length);
 
+  if (string != NULL && alloced_string == 1)
```
ctshim: It seems like you only need to check for alloced_string.

src/object/object_representation.c
youngjinj: Yes. I forgot that asserts can only be checked in debug mode.

src/query/query_executor.c
youngjinj: It follows the schema defined in the **pt_make_query_show_index** <sup>src/parser/parser_support.c</sup> function.
The table name must come with index 0.
```
const char *aliases[] = {
  "Table", "Non_unique", "Key_name", "Seq_in_index", "Column_name",
  "Collation", "Cardinality", "Sub_part", "Packed", "Null", "Index_type",
  "Func", "Comment", "Visible"
};
```

src/storage/heap_file.c
youngjinj: Okay. It seems like it would be better to exclude unnecessary comparisons.

src/storage/statistics_sr.c
youngjinj: Does that mean it's better to change it to a macro variable?
Later this will be changed to a macro variable by adding DB_MAX_FULL_CLASS_LENGTH.
For now, I'll set it to the DB_MAX_IDENTIFIER_LENGTH macro variable.

src/query/query_executor.c
youngjinj: We talked separately.
We decided not to handle exceptions for db_make_string () calls like other codes.
If exception handling is required, it seems to be resolved consistently through other issues.

src/object/object_representation.c
```diff
@@ -281,11 +281,19 @@ int or_Type_sizes[] = {
...
+  char *start = NULL;
+  int offset = 0;
+  int len = 0;
+
+  OR_BUF buffer;
+  int compressed_length = 0;
+  int decompressed_length = 0;
+  int rc = NO_ERROR;
+
+  assert (*alloced_string == 0);
```
hgryoo: NULL check assertion
```suggestion
  assert (string != NULL);
  assert (alloced_string != NULL);
```
for the `assert (*alloced_string == 0);`, There is no need to check that this variable is initialized to zero.

youngjinj: I've been talking about the same topic with @ctshim.
In conclusion, I decided to initialize it to NULL in or_class_name.

Since it is a variable passed as an argument to a function,
it seems to be recommended to initialize NULL before calling the function.
However, this increases unnecessary code, so I will initialize it to NULL and use it internally.

src/object/object_representation.c
hgryoo: I've understood that `or_get_varchar_compression_lengths` works well regardless of "enable_string_compression". I'm going to resolve it.

src/object/object_representation.c
hgryoo: When ` or_class_name` is called as the following, it should be checked on debug mode.
```
or_class_name (record, NULL, NULL);
```

youngjinj: Yes, I added the two check codes below.
```
assert (string != NULL);
assert (alloced_string != NULL);
```

src/object/object_representation.c
youngjinj: Thank you.

hgryoo (approved):

src/query/query_executor.c
beyondykk9: OK, I see.

beyondykk9 (approved):

ctshim (approved):

shparkcubrid (approved):

hyunikn (approved):

PR #3311, opened 2022-01-12
[CBRD-24077] Fixed coredump occurring in vacuum when restoredb is executed with -d option.

hornetmj:
http://jira.cubrid.org/browse/CBRD-24077

### Purpose
- The log_Gl.hdr.vacuum_last_blockid indicates the last block id that has been vacuumed in SA_MODE.
- When restoredb is executed, the vacuum data file also returns to the restore point, so the vacuum information performed in the last SA_MODE is not valid. (The important thing is that the record left by SA_MODE cannot be trusted because restoredb execution can change the database time point.)
- Looking at the reproduced cases registered in the issue, log_Gl.hdr.vacuum_last_blockid was updated with the latest information through vacuum_sa_reflect_last_blockid while diagdb was executed.
- Immediately afterward, it was restored to the past point in time through restoredb, and the vacuum data file was also restored to the past point in time.
- At this time, if the log_Gl.hdr.vacuum_last_blockid information is used, the information at the wrong time is used and a coredump occurs.

### Implementation
- N/A

### Remarks
- When vacuuming with SA_MODE, vacuum_sa_reflect_last_blockid() function always updates vacuum_sa_reflect_last_blockid and vacuum_Data.last_page->data->blockid to the same value.
- In case of abnormal termination in SA_MODE, vacuum_Data.last_page->data->blockid may be outdated than log_Gl.hdr.vacuum_last_blockid.
- A problem was found that the vacuum_sa_reflect_last_blockid() function was called even when the vacuum was set to skip the vacuum when csql and restoredb were executed. I am reviewing any issues that may arise from this, and I would like to address this as a different PR or issue.

### DIFF
--- a/src/query/vacuum.c
+++ b/src/query/vacuum.c
@@ -4178,19 +4178,7 @@ vacuum_data_load_and_recover (THREAD_ENTRY * thread_p)
 	   * in the log_Gl header. After a long session in SA_MODE, the vacuum_Data.last_page->data->blockid will
 	   * be outdated. Instead, SA_MODE updates log_Gl.hdr.vacuum_last_blockid before removing old archives.
 	   */
-	  VACUUM_LOG_BLOCKID last_blockid;
-
-	  if (vacuum_Data.is_restoredb_session)
-	    {
-	      last_blockid = vacuum_Data.last_page->data->blockid;
-	      log_Gl.hdr.vacuum_last_blockid = 0;
-	    }
-	  else
-	    {
-	      last_blockid = MAX (log_Gl.hdr.vacuum_last_blockid, vacuum_Data.last_page->data->blockid);
-	    }
-
-	  vacuum_Data.set_last_blockid (last_blockid);
+	  vacuum_Data.set_last_blockid (MAX (log_Gl.hdr.vacuum_last_blockid, vacuum_Data.last_page->data->blockid));
 
 	  vacuum_er_log (VACUUM_ER_LOG_VACUUM_DATA | VACUUM_ER_LOG_RECOVERY,
 			 "vacuum_data_load_and_recover: set last_blockid = %lld to MAX("
--- a/src/query/vacuum.c
+++ b/src/query/vacuum.c
@@ -4178,7 +4178,19 @@ vacuum_data_load_and_recover (THREAD_ENTRY * thread_p)
 	   * in the log_Gl header. After a long session in SA_MODE, the vacuum_Data.last_page->data->blockid will
 	   * be outdated. Instead, SA_MODE updates log_Gl.hdr.vacuum_last_blockid before removing old archives.
 	   */
-	  vacuum_Data.set_last_blockid (MAX (log_Gl.hdr.vacuum_last_blockid, vacuum_Data.last_page->data->blockid));
+	  VACUUM_LOG_BLOCKID last_blockid;
+
+	  if (vacuum_Data.is_restoredb_session)
+	    {
+	      last_blockid = vacuum_Data.last_page->data->blockid;
+	      log_Gl.hdr.vacuum_last_blockid = 0;
+	    }
+	  else
+	    {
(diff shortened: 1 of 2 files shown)


eido5 (approved):

src/query/vacuum.c
```diff
@@ -4178,7 +4178,19 @@ vacuum_data_load_and_recover (THREAD_ENTRY * thread_p)
 	   * in the log_Gl header. After a long session in SA_MODE, the vacuum_Data.last_page->data->blockid will
 	   * be outdated. Instead, SA_MODE updates log_Gl.hdr.vacuum_last_blockid before removing old archives.
 	   */
-	  vacuum_Data.set_last_blockid (MAX (log_Gl.hdr.vacuum_last_blockid, vacuum_Data.last_page->data->blockid));
+	  VACUUM_LOG_BLOCKID last_blockid;
+
+	  if (vacuum_Data.is_restoredb_session)
+	    {
+	      last_blockid = vacuum_Data.last_page->data->blockid;
+	      log_Gl.hdr.vacuum_last_blockid = 0;
```
joohok: How about put this initialization code into vacuum_initialize()

when the argument 'is_restore' is true, then set the log_Gl.hdr.vacuum_last_blockid to 0.

`int vacuum_initialize (...., bool is_restore)`
`{`
`   if (is_restore)` 
`    {`
`        log_Gl.hdr.vacuum_last_blockid = 0;`
`    }`
`}`

I think it is better to set vacuum_last_blockid when restoredb is executed.
If log_Gl.hdr.vacuum_last_blockid is set 0 when restoredb is executed, then original code is not required to be modified 
`vacuum_Data.set_last_blockid (MAX (log_Gl.hdr.vacuum_last_blockid, vacuum_Data.last_page->data->blockid));` 
what do you think?

hornetmj: - If your feedback is to initialize log_Gl.hdr.vacuum_last_blockid at the time of restoredb, the code I modified is also initialized at the time of restoredb.
- It would be better to change log_Gl.hdr.vacuum_last_blockid after restoredb completes recovery in log_initialize().
- log_Gl.hdr.vacuum_last_blockid is only meaningful when vacuum_empty().

joohok: OK, I understand. 
I didn't consider the `It would be better to change log_Gl.hdr.vacuum_last_blockid after restoredb completes recovery in log_initialize().`.

joohok (approved):

src/query/vacuum.c
wrlawodms: I have two suggestions.

1. How about using `logpb_last_complete_blockid()` to initiaizlize `log_Gl.hdr.vacuum_last_blockid`. It is normally  used to set a value to the variable with `log_Gl.append.prev_lsa.pageid`.
2. How about initializing it in `log_recovery_resetlog()`, which resets some variables in the log header when resetting log to a certain past time. It is called when time-specific(-d) restoredb or incomplete recovery occurred by a page corruption. It seems more natural and safer because some post/pre-processing seems to be done here (e.g. `logpb_flush_header (thread_p);`)

My suggestion is basically to handle the problem where it causes, not it is used or detected.

hornetmj: vacuum_Data.m_last_blockid is a low bound used to find log blocks that could not be added to the vacuum data file. If vacuum_Data.last_page->data->blockid < block id of log_Gl.append.prev_lsa.pageid, a log block that cannot be found may occur.

hornetmj: Modified to initialize log_Gl.hdr.vacuum_last_blockid using log_recovery_resetlog() function. Thanks for the good comments.

wrlawodms (approved):

hornetmj:
@mhoh3963 Although circleci complains, this appears to be a problem with the previous fix. So, I'm requesting a merge.

mhoh3963:
@hornetmj 
OK, I merged it.

PR #3284, opened 2021-12-30
[CBRD-23839] unloaddb.sh supports multi-process

kisoo-han:
http://jira.cubrid.org/browse/CBRD-23839

**Purpose**
The **unloaddb** utility that provides multi-process. 

**Implementation**
example:
sh unloaddb.sh -v -i input_file -u user -D /tmp -t num_process _database_name_

option summary:
  -   -v: if given, print verbose information
  -  -i: if given, table names to unload data of specified tables in this file, otherwise all tables in database will be unloaded
  - -u: if given, use user name **user** for unloaddb, otherwise **dba** will be used
  - -t: if given, maximum num_process unloaddb process will be invoked, otherwise -t 8 will be used (default)
  - -D: if given, target directory to create unloaddb output, otherwise current working directory will be used

If executed, for example dbname is demodb and num process is 4, following 4 directories will be created on current working directory:
  demodb.0, demodb.1, demodb.2, demodb.3

For all tables to be unloaded, calculate number of pages for each table and pick a process which has minimum pages to unload, and assign the table to the process. (using csql -c "show heap capacity of table _table_name_")

And each directory contains file named, **Table_list_demodb**.1, it contains table names unloaded. And each directory will contain unloaddb output for tables listed in _Table_list_database_name.x_. 
 (demodb.0/Table_list_demodb.0, demodb.1/Table_list_demodb.1, ...)

**Remarks**
NA

PR #3317, opened 2022-01-13
[CBRD-23839] shell script for unloaddb concurrently by sub-processes

kisoo-han:
http://jira.cubrid.org/browse/CBRD-23839

**Purpose**
The unloaddb utility that provides multi-process (**only DBA or DBA group user able to run this script**)

**Implementation**
example:
sh unloaddb.sh -v -i input_file -u user -D /tmp -t num_process database_name

option summary:

-v: if given, print verbose information
-i: if given, table names to unload data of specified tables in this file, otherwise all tables in database will be unloaded
-u: if given, use user name user for unloaddb, otherwise dba will be used
-t: if given, maximum num_process unloaddb process will be invoked, otherwise -t 8 will be used (default)
-D: if given, target directory to create unloaddb output, otherwise current working directory will be used
-d: if given, dump objects only; default: schema and objects (same context with cubrid unloaddb utility)
-s: if given, dump schema only; default: schema and objects (same context with cubrid unloaddb utility)

If executed, for example dbname is demodb and num process is 4, following 4 files will be created on _target_ directory:
- demodb_0_objects
- demodb_1_objects
- demodb_2_objects
- demodb_3_objects

Compared to data files, schema, index, trigger files will not be splitted and containing schama/index/trigger information for subject tables, for example:
demodb_indexes
demodb_schema
demodb_trigger

Additional information in database_unloaddb.log (by example)
- demodb_unloaddb.log/demodb_0.files: unloaded table names by process 0
- demodb_unloaddb.log/demodb_0.pid: child PID of unloaddb (for kill the process by parent)
- demodb_unloaddb.log/demodb_0.status: 0th Child process result (success or fail)

**Algorithm for allocating tables to processes:**

For all tables to be unloaded, do following to assign each table to a process with the result that all process have same load to unload:

1. calculate size each table (do '**csql -u user -l -c "show heap capacity of table" database**), and pick **Num_recs**, **Avg_rec_len** as number of rows and estimated schema size respectively (Num_recs * Avg_rec_len)
2. Pick a process which has minimum size to be processed.
3. Add current table size to the process to be processed.

Remarks
This PR is created due to some inconsistency between base branch and forked one.
Refer previous PR which is closed, https://github.com/CUBRID/cubrid/pull/3284
NA

### DIFF
--- a/contrib/CMakeLists.txt
+++ b/contrib/CMakeLists.txt
@@ -105,6 +105,6 @@ endif(UNIX)
 # multi-process unloaddb scripts
 if(UNIX)
   install(PROGRAMS
-    ${CMAKE_SOURCE_DIR}/contrib/script/unloaddb.sh.sh
+    ${CMAKE_SOURCE_DIR}/contrib/scripts/unloaddb.sh
     DESTINATION ${CUBRID_DATADIR}/scripts/)
 endif(UNIX)
--- a/contrib/CMakeLists.txt
+++ b/contrib/CMakeLists.txt
@@ -102,3 +102,9 @@ if(UNIX)
     DESTINATION ${CUBRID_DATADIR}/install/)
 endif(UNIX)
 
+# multi-process unloaddb scripts
+if(UNIX)
+  install(PROGRAMS
+    ${CMAKE_SOURCE_DIR}/contrib/script/unloaddb.sh.sh
+    DESTINATION ${CUBRID_DATADIR}/scripts/)
+endif(UNIX)
--- a/contrib/scripts/unloaddb.sh
+++ b/contrib/scripts/unloaddb.sh
@@ -24,12 +24,14 @@ num_tables=0
 slot_selected=0
 slot_size=(0 0 0 0 0 0 0 0)
 num_tables_slot=(0 0 0 0 0 0 0 0)
-database=
+database=""
 user="-u dba"
 pass=""
 total_pages=0
-from_file=""
+filename=""
 target_dir=$(pwd)
+from_file=0
+num_args_remain=0
 
 slot=()
 
@@ -125,7 +127,7 @@ function get_options ()
          while getopts ":D:u:i:t:v" opt; do
                 case $opt in
                         u ) user="-u $OPTARG" ;;
-                        i ) from_file="$OPTARG" ;;
+                        i ) filename="$OPTARG" ;from_file=1 ;;
                         t ) num_proc="$OPTARG" ;;
                         D ) target_dir="$OPTARG" ;;
                         v ) verbose="yes" ;;
@@ -133,8 +135,9 @@ function get_options ()
         done
 
         shift $(($OPTIND - 1))
-        database="$*"
-        echo $#
+
+        num_args_remain=$#
+        database=$*
 }
 
 function silent_cd ()
@@ -259,16 +262,15 @@ function analyze_table_info ()
         local db=$database
 
         # Get all table names from CATALOG
-
-        if [ X$from_file != X"" ];then  # Read table name from file
-                result=$(cat $from_file)
+        if [ $from_file -eq 1 ];then  # Read table name from file
+                result=$(cat $filename)
         else
                 result=$(csql $user $pass -c "select class_name from db_class where is_system_class = 'NO' AND class_type = 'CLASS' order by class_name" $db)
         fi
 
         for token in $result
         do
-                if  [ X$from_file = X"" ] && [ $found -eq 0 ];then
+                if  [ $from_file -eq 0 ] && [ $found -eq 0 ];then
                         str=$(echo $token | grep "====") # skip until we found table names in csql
                         if [ $? -eq 0 ];then
                                 found=1
@@ -276,13 +278,13 @@ function analyze_table_info ()
                         continue
                 fi
 
-                if [ X$from_file = X"" ] && [ ${token:0:1} != "'" ];then
+                if [ $from_file -eq 0 ] && [ ${token:0:1} != "'" ];then
                         break
                 fi
 
                 table_selected[idx]="$token"
 
-                if [ X$from_file = X"" ];then        # remove single quota from csql output
+                if [ $from_file -eq 0 ];then        # remove single quota from csql output
                         table_selected[idx]=$(echo ${table_selected[idx]} | sed "s/[\']//g")
                 fi
 
@@ -312,8 +314,9 @@ function analyze_table_info ()
 trap "cleanup" SIGHUP SIGINT SIGTERM
 set -o monitor
 
-extract_db_name $*
-if [ $(get_options "$@") -ne 1 ];then
+get_options "$@"
+
+if [ $num_args_remain -ne 1 ] || [ -z $database ];then
         show_usage
         exit 1
 fi
@@ -336,7 +339,7 @@ verify_user_pass
 
 echo -n "Analyzing table spacace ..."
 
-analyze_table_info $*
+analyze_table_info
 
 if [ $num_tables -eq 0 ];then
         echo "No Table SELECTED."
--- a/contrib/scripts/unloaddb.sh
+++ b/contrib/scripts/unloaddb.sh
@@ -277,12 +277,13 @@ function analyze_table_info ()
         local idx=0
         local table_name
         local db=$database
+        local query="show tables"
 
-        # Get all table names from CATALOG
+        # Get all table names if $from_file equals 0, otherwise from input file
         if [ $from_file -eq 1 ];then  # Read table name from file
                 result=$(cat $filename)
         else
-                result=$(csql $user $pass -c "select class_name from db_class where is_system_class = 'NO' AND class_type = 'CLASS' order by class_name" $db)
+                result=$(csql $user $pass -c "$query" $db)
         fi
 
         for token in $result
--- a/contrib/scripts/unloaddb.sh
+++ b/contrib/scripts/unloaddb.sh
@@ -94,18 +94,9 @@ function verify_user_pass ()
 	local dbuser
 	local passwd
 	local qry_dba_grp="SELECT u.name FROM db_user AS u, TABLE(u.groups) AS g(x) where x.name = 'DBA'"
-	local retcode
 
 	dbuser=$(echo ${user} | cut -d' ' -f2-2)
 	USERNAME=$(echo ${dbuser} | tr [:lower:] [:upper:])
-
-	# check the database server is running
-	retcode=$(ps -ef | grep cub_server | grep $database | wc -l)
-
-	if [ $retcode -eq 0 ];then
-		echo "Database server '$database' is not running"
-		exit 1
-	fi
 
 	dba_groups=$(csql -u public -l -c "$qry_dba_grp" $database | grep -w $USERNAME | wc -l)
 
@@ -339,7 +330,14 @@ if [ ! -d $target_dir ];then
         echo "$target_dir: directory not exists or permission denied"
         exit
 else
-   silent_cd $target_dir
+        silent_cd $target_dir
+fi
+
+# check the database server is running
+retcode=$(ps -ef | grep cub_server | grep $database | wc -l)
+if [ $retcode -eq 0 ];then
+        echo "Database server '$database' is not running"
+        exit 1
 fi
 
 verify_user_pass
--- a/contrib/scripts/unloaddb.sh
+++ b/contrib/scripts/unloaddb.sh
@@ -168,7 +168,6 @@ function get_table_size ()
         local table_name=$1
         local Avg_rec_len=0
         local table_size=0
-        local csql_output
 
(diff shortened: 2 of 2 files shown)


contrib/scripts/unloaddb.sh
```diff
@@ -0,0 +1,402 @@
...
+
+	# check the database server is running
+	retcode=$(ps -ef | grep cub_server | grep $database | wc -l)
+
+	if [ $retcode -eq 0 ];then
+		echo "Database server '$database' is not running"
+		exit 1
+	fi
+
+	dba_groups=$(csql -u public -l -c "$qry_dba_grp" $database | grep -w $USERNAME | wc -l)
```
mhoh3963: If public user have password, you can't get the result what you want.
How about change that shell script have to be executed by only dba ?

contrib/scripts/unloaddb.sh
```diff
@@ -0,0 +1,402 @@
...
+
+function get_table_size ()
+{
+        local num_rows=0
+        local table_name=$1
+        local row_size=0
+        local table_size=0
+
+        num_rows=$(csql $user $pass -l -c "show heap capacity of $table_name" $db | grep Num_recs | awk '{print $3}')
+        row_size=$(get_schema_size $table_name)
```
mhoh3963: Since we can get  avg_rec_len of the table from "show heap capacity of", 
how about replacing get_schema_size function to avg_rec_len value ?

contrib/scripts/unloaddb.sh
```diff
@@ -0,0 +1,402 @@
...
+	fi
+}
+
+function get_options ()
+{
+         while getopts ":D:u:i:t:v" opt; do
+                case $opt in
+                        u ) user="-u $OPTARG" ;;
+                        i ) from_file="$OPTARG" ;;
+                        t ) num_proc="$OPTARG" ;;
```
mhoh3963: Is it possible to run more than 8 process ?
if it can't be executed and num_proc greater than 8, I think that you have to tune num_proc to 8.

contrib/scripts/unloaddb.sh
```diff
@@ -0,0 +1,402 @@
...
+        local found=0
+        local line
+        local idx=0
+        local table_name
+        local db=$database
+
+        # Get all table names from CATALOG
+
+        if [ X$from_file != X"" ];then  # Read table name from file
+                result=$(cat $from_file)
```
mhoh3963: When tables is gotten from file, I seem like that you are better to set a flag and check the flag at below routines for readability.

contrib/scripts/unloaddb.sh
```diff
@@ -0,0 +1,402 @@
...
+        else
+                result=$(csql $user $pass -c "select class_name from db_class where is_system_class = 'NO' AND class_type = 'CLASS' order by class_name" $db)
```
mhoh3963: You can use "show tables" instead of this query.
And you can also remove column headers with --skip-column-names at csql.

contrib/scripts/unloaddb.sh
```diff
@@ -0,0 +1,402 @@
...
+extract_db_name $*
+get_options "$@"
+
+if [ ! -d $target_dir ];then
+        echo "$target_dir: directory not exists or permission denied"
+        exit
+else
+   silent_cd $target_dir
+fi
+
```
mhoh3963, line 413: Instead of checking whether the cub_server of database is running in verify_user_pass,
How about check it here explicitly ?

kisoo-han: I agree, it is moved from verify_user_pass () to MAIN.
Thank you for your comment!

contrib/scripts/unloaddb.sh
```diff
@@ -0,0 +1,402 @@
...
+	local dbuser
+	local passwd
+	local qry_dba_grp="SELECT u.name FROM db_user AS u, TABLE(u.groups) AS g(x) where x.name = 'DBA'"
+	local retcode
+
+	dbuser=$(echo ${user} | cut -d' ' -f2-2)
+	USERNAME=$(echo ${dbuser} | tr [:lower:] [:upper:])
+
+	# check the database server is running
+	retcode=$(ps -ef | grep cub_server | grep $database | wc -l)
```
mhoh3963: If database has hostname with @ string such as demodb@localhost on HA mode, it can't be found the cub_server's pid from process list.
I think that you need to consider a hostname at database.

contrib/scripts/unloaddb.sh
kisoo-han: Thank you for your comment, I didn't think about the case that user public has password.
I fixed it.

contrib/scripts/unloaddb.sh
```diff
@@ -0,0 +1,402 @@
...
+                fi
+        done
+
+        if [ $verbose = "yes" ];then
+                echo "Proc $slot_num: num tables: $num_tables_in_slot, ${slot_size[$slot_num]} bytes"
+        fi
+
+        echo "process $slot_num: starting" > unloaddb_$slot_num.status
+
+	cubrid unloaddb $user $pass --input-class-only --input-class-file $file $database &
```
mhoh3963: How about using the -d or --data-only option here?
And, I think that it would be better not to unload schema files at this script.
If it is necessary to unload schema file, it is better to unload it all at once rather than dividing it.

contrib/scripts/unloaddb.sh
kisoo-han: Comment accepted, I add a flag '**from_file**' for that and changed variable for file to 'filename'

contrib/scripts/unloaddb.sh
kisoo-han: I set default num proc to 8, and max num proc to 16. 
And if a user set more than 16, for example '-t 17', it will be set to 16.
Thank you for your comment!

contrib/scripts/unloaddb.sh
kisoo-han: In that case, neither **index** nor **trigger** could be unloaded as well, what do you think? 
@mhoh3963 , please comment this.

contrib/scripts/unloaddb.sh
kisoo-han: Thank you for your comment, I didn't consider that.
Changed for check running db server excluding @hostname if it user gives it.
(demodb, demodb@localhost, demodb@ipaddr => only check for 'demodb' server is running)

contrib/scripts/unloaddb.sh
kisoo-han: Comment accepted, I changed the query to get all table names to "show tables" (it is quite simple, thank you).
In the for_loop to get table name in csql, I did
1. skip line until ===
2. remove single quota '**code**' to **code**
3. loop until no single quota is found

And if it is from file, I assumed there is one table name per line. (no single quota)

contrib/scripts/unloaddb.sh
kisoo-han: I saw that parameter of csql output but not sure it is correct answer for record length.
I accept your recommendation and change table size as (two args are output of "**csql** show heap capacity of _table_")
size of **table** = **Num_recs** * **Avg_rec_len**

contrib/scripts/unloaddb.sh
```diff
@@ -0,0 +1,387 @@
...
+from_file=0
+num_args_remain=0
+
+slot=()
+
+function show_usage ()
+{
+         echo "Usage: $0 [OPTIONS] [database]"
+         echo " OPTIONS"
+         echo "  -t arg  Set number of parallel process; default 8"
```
hwany7seo: I like to add maximum value description. (e.g. max 16)

contrib/scripts/unloaddb.sh
```diff
@@ -0,0 +1,387 @@
...
+         echo "  -i arg  input FILE of table names; default: dump all classes"
+         echo "  -u arg  Set database user name; default dba"
+         echo "  -p arg  Set dbuser password"
+         echo "  -D arg  Set directory for unloaddb output dir/files"
```
hwany7seo: 'password option' are unused. please delete it.

contrib/scripts/unloaddb.sh
```diff
@@ -0,0 +1,387 @@
...
+         echo "  -v      Set verbose mode on"
+
+         echo ""
+         echo " EXAMPLES"
+         echo "  $0 -t 4 -v demodb          # unload all tables in demodb"
+         echo "  $0 -i file demodb          # unload tables listed in file in demodb"
```
hwany7seo: The '-i' option are based on 'target option' folder.
it would be better based on 'shell file for execute'. 
an error occurs when input as in the example below.  (target file is in the same path as execute file).
(e.g. ./unloaddb.sh -u dba -t 17 -i ./search -v -D aaa demodb -> ERROR)
I think users will think so.

if not, explain to enter the absolute path in 'show usage'.

hwany7seo: Don't you need an installation for engine release?
If necessary, add it to CMAKE file.

hwany7seo:
Don't you need an installation for engine release?
If necessary, add it to CMAKE file.

contrib/scripts/unloaddb.sh
mhoh3963: I think that it should be handled with trigger and index files in the same way as schema file.
If we decide that it don't be unloaded the schema file at this shell script, trigger and index won't need to unloaded too.

contrib/scripts/unloaddb.sh
kisoo-han: Thank you for your comments, I missed to delete it.
Fixed.

contrib/scripts/unloaddb.sh
kisoo-han: You right, target should be applied to unloaddb output only, -i option should be applied the current.
Thank for your comment!!
Fixed it.

kisoo-han:
Added lines to contrib/CMakeLists.txt to include unloaddb.sh to dist package. Thank you @hwany7seo

contrib/CMakeLists.txt
```diff
@@ -102,3 +102,9 @@ if(UNIX)
     DESTINATION ${CUBRID_DATADIR}/install/)
 endif(UNIX)
 
+# multi-process unloaddb scripts
+if(UNIX)
+  install(PROGRAMS
+    ${CMAKE_SOURCE_DIR}/contrib/script/unloaddb.sh.sh
```
hwany7seo: Invalid path. (script/unloaddb.sh.sh -> scripts/unloaddb.sh)

hwany7seo (approved):

contrib/scripts/unloaddb.sh
mhoh3963: Since it can be executed by only dba, I think that  '-u option' is unnecessary. Isn't it ?

mhoh3963 (commented):
As you said in the offline meeting, Do you modify codes to unload into one file rather than separate schema, trigger and index files per process, after changing flatten the directory structure ?
If you work it, after changing it, I will continue to review.
Please, let me know.

contrib/scripts/unloaddb.sh
kisoo-han: Fixed it. Thank you!

contrib/CMakeLists.txt
kisoo-han: Fixed. Thank you!

kisoo-han:
@mhoh3963 , @hwany7seo, @airnet73 
I think I finished the functions I can imagine. And you are re-invited to this pr's review.
Major changes after your review:
1. introduce -d, -s option for data-only and schema-only (default both data and schema)
2. data files will be created in the target directory as a form of 
   _database_\_0_objects, ..., _database_\_n_objects
3. On the other hand, a log directory as a form of _database_\_unloaddb.log will contain all temporary files to process
    for example, PID, tables names contained in a process, processing result
3. if schema/trigger/index file need to be created, it will be created as one file for all schema not splitted fasion
  _database_\_schema, _database_\_trigger, _database_\_indexes
4. Exit on unknown options

contrib/scripts/unloaddb.sh
```diff
@@ -0,0 +1,468 @@
...
+fi
+
+echo ""
+
+# Assign all tables to unloaddb slot
+# Assign and calculate total table size & num tables assigned in each slot
+for ((i = 0; i < $num_tables; i++))
+do
+	this_slot=$(find_least_loaded_proc)
+	tbl=${table_selected[i]}
```
mhoh3963: I think that tbl variable doesn't use at codes. Can you remove that ?

mhoh3963 (approved):
LGMT :)

contrib/scripts/unloaddb.sh
kisoo-han: That line was removed, thank you for your comment!

hwany7seo (approved):

airnet73 (approved):

PR #3318, opened 2022-01-14
[CBRD-24171] Fix cubmem::extensible_block's move constructor

hgryoo:
http://jira.cubrid.org/browse/CBRD-24171

Fixed the move constructor of `cubmem::extensible_block`. m_block's content is not moved.

### DIFF
--- a/src/base/mem_block.hpp
+++ b/src/base/mem_block.hpp
@@ -306,7 +306,7 @@ namespace cubmem
   extensible_block::extensible_block (extensible_block &&b)
     : extensible_block { *b.m_allocator }
   {
-    m_block = std:move (b.m_block);
+    m_block = std::move (b.m_block);
   }
 
   extensible_block::extensible_block (const block_allocator &alloc)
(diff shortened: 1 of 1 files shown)


src/base/mem_block.hpp
```diff
@@ -306,6 +306,8 @@ namespace cubmem
   extensible_block::extensible_block (extensible_block &&b)
     : extensible_block { *b.m_allocator }
   {
+    m_block.dim = b.m_block.dim;
+    m_block.ptr = b.m_block.ptr;
     b.m_block.dim = 0;
     b.m_block.ptr = NULL;
```
wrlawodms:
```suggestion
    m_block = std:move (b.m_block);
```
How about it to make use of the defend move op of `struct block`?

wrlawodms: However, `extensible_block::operator= (extensible_block &&b)` also define its own move. We can go with your opinion.

wrlawodms (approved):

beyondykk9 (approved):

src/base/mem_block.hpp
hgryoo: Thank you. I'd like to accept this suggestion.

PR #3320, opened 2022-01-17
[CRBD-24165] Remove unused function disk_auto_volume_expansion_daemon…

minjune8506:
http://jira.cubrid.org/projects/CBRD/issues/CBRD-24165?filter=allopenissues

**Purpose**
Remove unused function disk_auto_volume_expansion_daemon_init() and the related code.

**Implementation**
disk_auto_volume_expansion_daemon_init() and disk_auto_volume_expansion_daemon_destroy() are not implemented.
So, I commented out related codes

**Remarks**
N/A

### DIFF
--- a/src/storage/disk_manager.c
+++ b/src/storage/disk_manager.c
@@ -478,10 +478,10 @@ STATIC_INLINE void disk_check_own_reserve_for_purpose (DB_VOLPURPOSE purpose) __
 static DISK_ISVALID disk_check_volume (THREAD_ENTRY * thread_p, INT16 volid, bool repair);
 
 // *INDENT-OFF*
-static cubthread::daemon *disk_Auto_volume_expansion_daemon = NULL;
+// static cubthread::daemon *disk_Auto_volume_expansion_daemon = NULL;
 
-static void disk_auto_volume_expansion_daemon_init ();
-static void disk_auto_volume_expansion_daemon_destroy ();
+// static void disk_auto_volume_expansion_daemon_init ();
+// static void disk_auto_volume_expansion_daemon_destroy ();
 // *INDENT-ON*
 
 /************************************************************************/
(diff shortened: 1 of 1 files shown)


wrlawodms:
@minjune8506 Thank you for your first contribution. We start the PR description with a Jira URL, which is also quite a different form than what you have used. Please see other PRs and fix it.

hornetmj:
@minjune8506 @wrlawodms @joohok @eido5 What do you think about removing unnecessary code instead of disabling it?

joohok:
If there is something implemented in the function, I thought it is better to wrap the code with 
`#if defined ENABLE_UNUSED_FUNCTION ... #endif ` .
However, it seems to be just an empty function, so I think it can be removed rather than disabled.

wrlawodms:
@hornetmj I have taken a look at the past code using "auto extension". It seems to be a sort of background volume extension concept, in which some volume extension was done by a background thread, not by a transaction itself. 

Now we know what the code was for, I think we can remove that code. 

FYI:
commit: 147ff59f084b9709ab6ae35c992558afaa1d9b97
The related system parameter: PRM_ID_GENERIC_VOL_PREALLOC_SIZE (which is obsolete now)
The releated functions: boot_add_auto_volume_extension(), thread_auto_volume_expansion_thread()

wrlawodms:
@minjune8506 I reckon that you should consider removing `disk_auto_expand()` and `disk_auto_expansion_execute()` too.

hornetmj:
@minjune8506 Remove unnecessary codes instead of disabling them.

minjune8506:
@hornetmj @wrlawodms @joohok
Thank you for your comments.
I accepted your advice and changed my commit.

wrlawodms:
@minjune8506 You seem to commit the last change to the develop branch on your forked repo. It has to be pushed to `minjune8506:CRBD-24165`. Please check it. Plus, fix the JIRA url at the head of the description.

PR #3322, opened 2022-01-18
[CBRD-24133] Implement flashback utility options	

joohok:
http://jira.cubrid.org/browse/CBRD-24133

**Purpose**
Skeleton code for flashback utility. 
Register a flashback utility, and check arguments from an user whether they are valid. 

**Implementation**
- flashback () 
   - implementation for flashback utility
   - checks number of arguments ( dbname, table list)
   - sets and verifies options
- parse_date_string_to_time ()
   - parse date string (DD-MM-YYYY:hh:mm:ss) and convert it to 'time_t' value

**Remarks**
N/A


src/executables/util_cs.c
```diff
@@ -3993,3 +3993,311 @@ tde (UTIL_FUNCTION_ARG * arg)
 error_exit:
   return EXIT_FAILURE;
 }
+
+static time_t
+parse_date_string_to_time (char *date_string)
```
wrlawodms, line 3998: Why don't you use the existing standard solutions? https://stackoverflow.com/questions/11213326/how-to-convert-a-string-variable-containing-time-to-time-t-type-in-c

src/executables/util_cs.c
```diff
@@ -3993,3 +3993,311 @@ tde (UTIL_FUNCTION_ARG * arg)
...
+  /* class existence and classoid will be found at server side. if is checked at utility side, it needs addtional access to the server through locator */
+  darray = da_create (num_tables, SM_MAX_IDENTIFIER_LENGTH);
+  if (darray == NULL)
+    {
+      perror ("calloc");
+      util_log_write_errid (MSGCAT_UTIL_GENERIC_NO_MEM);
+      goto error_exit;
+    }
+
+  if (num_tables > 0)
```
wrlawodms: It seems always true.

src/executables/util_cs.c
```diff
@@ -3993,3 +3993,311 @@ tde (UTIL_FUNCTION_ARG * arg)
...
+    {
+      da_destroy (darray);
+    }
+
+  return EXIT_SUCCESS;
+
+print_flashback_usage:
+  util_log_write_errid (MSGCAT_UTIL_GENERIC_INVALID_ARGUMENT);
+error_exit:
+  fprintf (stderr, "Flashback Error Exit\n");
```
wrlawodms: It seems to print the Error when printing the Usage message.

src/executables/util_cs.c
joohok: Yes, I will remove it

src/executables/util_cs.c
joohok: Yes, I will remove it. It was used just for testing.

src/executables/util_cs.c
joohok: Oh.. Thank you. I will take a look at what you sent.

joohok: I replace it with ssanf function.

There are two  candidate functions for parsing date format string and converting it to time_t. 
1. std::get_time
   - Can not handle a exception if some parts(day, month, year, hour, minutes or seconds) are not input 
2. sscanf 
   - Have to check that the parts are set correctly, but can handle the exception mentioned above.

src/executables/util_cs.c
```diff
@@ -3993,3 +3993,255 @@ tde (UTIL_FUNCTION_ARG * arg)
...
+    }
+
+  if (time_data.tm_mday < 1 || time_data.tm_mday > 31
+      || time_data.tm_mon < 1 || time_data.tm_mon > 12
+      || time_data.tm_year < 1900
+      || time_data.tm_hour < 0 || time_data.tm_hour > 23
+      || time_data.tm_min < 0 || time_data.tm_min > 59 || time_data.tm_sec < 0 || time_data.tm_sec > 59)
+    {
+      return 0;
+    }
```
wrlawodms, line 4019: Do we have to validate the time string? it even can't filter all wrong date like 2/30

src/executables/util_cs.c
```diff
@@ -3993,3 +3993,255 @@ tde (UTIL_FUNCTION_ARG * arg)
...
+
+  time_data.tm_mon -= 1;
+  time_data.tm_year -= 1900;
+  time_data.tm_isdst = -1;
```
wrlawodms, line 4023: Why does it see the time from 1900 and 0-th month? Timestamps on the server do?

joohok: yes, 
it is spec of struct tm.
- tm_year : years since 1900
- tm_mon : months since January [0 - 11]

src/executables/util_cs.c
joohok: Yes, you are right, it can not filter like 2/30, 4/31, ..etc. 
we don't have to validate the time string, but I just follow the spec of the members in `struct tm` which is used to get `time_t` value.

src/executables/util_cs.c
```diff
@@ -3993,3 +3993,255 @@ tde (UTIL_FUNCTION_ARG * arg)
...
+    {
+      perror ("calloc");
+      util_log_write_errid (MSGCAT_UTIL_GENERIC_NO_MEM);
+      goto error_exit;
+    }
+
+  for (i = 0; i < num_tables; i++)
+    {
+      table_name = utility_get_option_string_value (arg_map, OPTION_STRING_TABLE, i + 1);
+      if (table_name == NULL)
```
hornetmj: I think a table name is never NULL. Are there any cases where a table name can be NULL?

src/executables/util_cs.c
```diff
@@ -3993,3 +3993,255 @@ tde (UTIL_FUNCTION_ARG * arg)
...
+
+  output_file = utility_get_option_string_value (arg_map, FLASHBACK_OUTPUT_S, 0);
+  user = utility_get_option_string_value (arg_map, FLASHBACK_USER_S, 0);
+  dba_password = utility_get_option_string_value (arg_map, FLASHBACK_DBA_PASSWORD_S, 0);
+  start_date = utility_get_option_string_value (arg_map, FLASHBACK_START_DATE_S, 0);
+  end_date = utility_get_option_string_value (arg_map, FLASHBACK_END_DATE_S, 0);
+  is_detail = utility_get_option_bool_value (arg_map, FLASHBACK_DETAIL_S);
+  is_oldest = utility_get_option_bool_value (arg_map, FLASHBACK_OLDEST_S);
+
+  if (check_database_name (database_name))
```
hornetmj: It would be nice to execute just after getting the database_name.

src/executables/util_cs.c
```diff
@@ -3993,3 +3993,255 @@ tde (UTIL_FUNCTION_ARG * arg)
...
+  else if (start_time == 0 && end_time == 0)
+    {
+      end_time = time (NULL);
+      start_time = end_time - 600;
+    }
+
+  /* 1. start time < end time
+   * 2. start time is required to be set after db_creation time (server side check)
+   * 3. start time, and end time are required to be set within the log volume range (server side check) */
+  if (start_time > end_time)
```
```suggestion
  if (start_time >= end_time)
```

src/executables/util_cs.c
```diff
@@ -3993,3 +3993,255 @@ tde (UTIL_FUNCTION_ARG * arg)
...
+	  goto error_exit;
+	}
+    }
+
+  /* error message log file */
+  snprintf (er_msg_file, sizeof (er_msg_file) - 1, "%s_%s.err", database_name, arg->command_name);
+  er_init (er_msg_file, ER_NEVER_EXIT);
+
+  db_set_client_type (DB_CLIENT_TYPE_ADMIN_UTILITY);
+  if (db_login ("DBA", dba_password) != NO_ERROR)
```
hornetmj, line 4181:
```suggestion
  if (db_login (user, dba_password) != NO_ERROR)
```

src/executables/util_cs.c
```diff
@@ -3993,3 +3993,255 @@ tde (UTIL_FUNCTION_ARG * arg)
...
+	  dba_password = passbuf;
+	  if (db_login ("DBA", dba_password) != NO_ERROR)
+	    {
+	      PRINT_AND_LOG_ERR_MSG ("%s\n", db_error_string (3));
+	      goto error_exit;
+	    }
+	  else
+	    {
+	      error = db_restart (arg->command_name, TRUE, database_name);
+	    }
```
hornetmj, line 4215: Do we need this mechanism? It is enough to return an error that informs invalid passwords.

src/executables/util_cs.c
```diff
@@ -3993,3 +3993,255 @@ tde (UTIL_FUNCTION_ARG * arg)
...
+
+  if (darray != NULL)
+    {
+      da_destroy (darray);
+    }
+
+  return EXIT_SUCCESS;
+
+print_flashback_usage:
+  util_log_write_errid (MSGCAT_UTIL_GENERIC_INVALID_ARGUMENT);
```
hornetmj, line 4235: Do you change this to a usages message in the next PR?

src/executables/util_cs.c
hornetmj: And how do you check whether the user has DBA privileges? It looks like you need to use the au_is_dba_group_member() function, but you don't use it, so I leave a question.

src/executables/util_cs.c
hornetmj: It would be better to change the label(print_flashback_usage) If you just leave an error.

src/executables/util_cs.c
joohok: No, it can not be a NULL. I will remove it.

src/executables/util_cs.c
joohok: Yes, I thought it would be more readable but I did it for a consistency with legacy.
But I will move it, because this function just check the name of the database.

src/executables/util_cs.c
joohok: ok, I will change it

src/executables/util_cs.c
joohok: 'user' option is not a user that execute a flashback utility. It is a filter option that extract SQL logs executed by a transaction 'user'. 
All the other utilities which need a dba privilege doesn't specify the user who executed.

src/executables/util_cs.c
joohok: This is a additional mechanism, not mandatory, for user who doesn't input a password. 
I think it is nice to provide a chance to input a password, and this is what other utilities, which require password, did.

src/executables/util_cs.c
joohok: I will change the usage messages in other issue which is related to working on error and help message.

src/executables/util_cs.c
hornetmj: @joohok Oh. I'm confused. I hope you can answer the second question as well.

src/executables/util_cs.c
hornetmj: OK.

src/executables/util_cs.c
hornetmj: OK. You need to think about whether to print usage or output explicit errors.

src/executables/util_cs.c
joohok: Utility will be executed as a "DBA" user, so I think it doesn't need to check the dba privilege. 
How do you think?

hornetmj: @joohok I got it.

wrlawodms (approved):

eido5 (approved):

hornetmj (approved):

PR #3327, opened 2022-01-18
[CBRD-24166] Remove unnecessary validation code in the dwb_flush_next_block function.

hyahong:
http://jira.cubrid.org/browse/CBRD-24166

Purpose
remove unnecessary code in dwb_flush_next_block function.

Implementation
N/A

Remarks
N/A

### DIFF
--- a/src/storage/double_write_buffer.c
+++ b/src/storage/double_write_buffer.c
@@ -3376,7 +3376,7 @@ start:
 
       /* If version of current block is UINT64_MAX, previous block version < current block version is true */
       assert (DWB_GET_PREV_BLOCK (flush_block->block_no)->version > flush_block->version
-	      || flush_block->version == UINT64_MAX);
+	      || flush_block->block_no == 0 || flush_block->version == UINT64_MAX);
 
       error_code = dwb_flush_block (thread_p, flush_block, true, NULL);
       if (error_code != NO_ERROR)
--- a/src/storage/double_write_buffer.c
+++ b/src/storage/double_write_buffer.c
@@ -3378,7 +3378,7 @@ start:
 	      (flush_block->block_no == 0
 	       && (DWB_GET_PREV_BLOCK (flush_block->block_no)->version == flush_block->version))
 	      || (flush_block->version == UINT64_MAX
-		  && (DWB_GET_PREV_BLOCK (flush_block->block_no)->version < flush_block->version));
+		  && (DWB_GET_PREV_BLOCK (flush_block->block_no)->version < flush_block->version)));
 
       error_code = dwb_flush_block (thread_p, flush_block, true, NULL);
       if (error_code != NO_ERROR)
--- a/src/storage/double_write_buffer.c
+++ b/src/storage/double_write_buffer.c
@@ -3374,6 +3374,10 @@ start:
       /* Flush all pages from current block */
       assert (flush_block != NULL && flush_block->count_wb_pages == DWB_BLOCK_NUM_PAGES);
 
+      /* If version of current block is UINT64_MAX, previous block version < current block version is true */
+      assert (DWB_GET_PREV_BLOCK (flush_block->block_no)->version > flush_block->version
+        || flush_block->version == UINT64_MAX);
+
       error_code = dwb_flush_block (thread_p, flush_block, true, NULL);
       if (error_code != NO_ERROR)
 	{
--- a/src/storage/double_write_buffer.c
+++ b/src/storage/double_write_buffer.c
@@ -3374,16 +3374,6 @@ start:
       /* Flush all pages from current block */
       assert (flush_block != NULL && flush_block->count_wb_pages == DWB_BLOCK_NUM_PAGES);
 
-      if (DWB_GET_PREV_BLOCK (flush_block->block_no)->version < flush_block->version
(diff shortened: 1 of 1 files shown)


joohok (approved):
LGTM.

wrlawodms:
@Yechan0815 Please use the following URL format: http://jira.cubrid.org/browse/CBRD-24166

hornetmj:
@Yechan0815 How did you verify it?

wrlawodms:
In my opinion, I doubt that this assertion is useless. As you mentioned in the JIRA description, it may not happen and must not happen. I suppose this is the reason this assertion is here. When the wrong behavior happens, we have to notice it happens with the assertion. 

I don't really know the details of DWB, but the assertion seems to define the pre-condition of `dwb_flush_next_block()`,  which may be determined by other functions. I'd like to have others' opinions on it. @hornetmj @joohok @eido5

hyahong:
@wrlawodms It's corrected. I will be more careful next time.

hyahong:
@hornetmj 
this assertion can occur when pages are not stacked or flushed in order.

since the fact that pages are stacked in blocks in order is described in the issue, just need to prove that flushes are done in block order.

function dwb_flush_block where flush occurs is called in a total of three cases. 
and in the following three cases, flush must be done in block order.

```
1. dwb_starts_structure_modification. when dwb is created
2. dwb_add_page. when page is added to dwb
3. dwb_flush_next_block. when daemon flushes the dwb block.
```

1.
while dwb is being created, flag DWB_MODIFY_STRUCTURE is set, so page addition cannot occur in other areas.
so, while dwb is being created, dwb is not affected other than dwb_starts_structure_modification logic and so flush is in order.


2.
dwb_add_page adds a page to dwb. at this time, if block with pages added is full, the processing is divided according to whether the flush daemon is used.
if the flush daemon is available, dwb_add_page exits without flush request.
if the flush daemon is unavailable, dwb_add_page will request flush directly. In this case, since dwb_add_page is the only flush requestor, and the pages are stacked in order unconditionally, flush is in order.


3.
dwb_flush_next_block is only called from the daemon.
dwb_flush_next_block flushes "dwb_Global.next_block_to_flush"th block.
the variable "dwb_Global.next_block_to_flush" is incremented after flushing the block, and it continues to traverse all block numbers.
so even if use the daemon, the flush is in order.

Therefore, pages are stacked in blocks in order, and flush is done in order, so the above verification code is unnecessary.

hornetmj:
@Yechan0815 You mean the following condition is always true. right?

DWB_GET_PREV_BLOCK (flush_block->block_no)->version > flush_block->version

hyahong:
@hornetmj 
because version is incremented with each flush

`DWB_GET_PREV_BLOCK (flush_block->block_no)->version > flush_block->version` 
is true before flushing the current block,

and
`DWB_GET_PREV_BLOCK (flush_block->block_no)->version == flush_block->version`
is true after flushing the current block.

hornetmj:
@Yechan0815 Yes. It would be nice if you put assert (DWB_GET_PREV_BLOCK (flush_block->block_no)->version > flush_block->version) instead of the code you removed.

hornetmj:
@Yechan0815 I have something to check. Please wait until I give you more comments.

hornetmj:
@Yechan0815 The condition "DWB_GET_PREV_BLOCK (flush_block->block_no)->version < flush_block->version" can occur very rarely when the version of the current block is UINT64_MAX. You need to take this into account when adding assert. And in my opinion, if the current flush block id > the previous flushed block id, then the previous version > the current version, and if the current flush block id < the previous flushed block id, then the previous version == the current version at the code you removed. Considering these two, it would be good to insert the assert code.

The second condition of OR is judged as a condition that can never occur. If you have any further comments on this, please leave a comment. (@eido5, @wrlawodms, @joohok, @Yechan0815)

hyahong:
@hornetmj I also agree. I don't think the second condition can happen.

hyahong (commented):
@hornetmj 
```c
/* If version of current block is UINT64_MAX, previous block version < current block version is true */
assert (DWB_GET_PREV_BLOCK (flush_block->block_no)->version > flush_block->version
  || flush_block->version == UINT64_MAX);
```

and I added this assertion, would this be enough?

hornetmj (approved):
LGTM

hornetmj:
@Yechan0815 We lost the following condition:
- if flush_block->block_no == 0, then DWB_GET_PREV_BLOCK (flush_block->block_no)->version == flush_block->version.

wrlawodms (approved):

hornetmj:
@Yechan0815 How about this? If then you can remove the comment you leave.  
```
assert ((DWB_GET_PREV_BLOCK (flush_block->block_no)->version > flush_block->version) || 
(flush_block->block_no == 0 && (DWB_GET_PREV_BLOCK (flush_block->block_no)->version == flush_block->version)) || 
(flush_block->version == UINT64_MAX && (DWB_GET_PREV_BLOCK (flush_block->block_no)->version < flush_block->version));
```

hyahong:
@hornetmj 
This is more intuitive! I changed the code to this.

hornetmj (approved):

eido5 (approved):

PR #3330, opened 2022-01-19
[LETS-383] core read heap page from file or page server

cristiarg:
http://jira.cubrid.org/browse/LETS-383

In catalog_drop_representation_helper function an undo RVCT_UPDATE log record was being added (log_append_undo_recdes2) without the corresponding redo RVCT_UPDATE log record (log_append_redo_recdes2) when the representation did not overflow (which is, probably, most of the times).
This had the effect of breaking validation in scalability validation scenarios (which involve executing active transaction server with local storage and comparing pages from local storage and from page server storage):
- the heap page received from the page server had a lower prv.lsa than the one read from the active transaction server's page buffer
- because the replication on page server did not update the page's prv.lsa in the absence of a redo log record

Other:
- harmonized some log prior transfer logging
- in pgbuf_respond_data_fetch_page_request only unfix page after having used the casted pointer for logging


src/storage/system_catalog.c
```diff
@@ -1614,13 +1614,13 @@ catalog_drop_representation_helper (THREAD_ENTRY * thread_p, PAGE_PTR page_p, VP
 
   spage_get_record (thread_p, page_p, CATALOG_HEADER_SLOT, &record, COPY);
 
-  log_append_undo_recdes2 (thread_p, RVCT_UPDATE, &catalog_Id.vfid, page_p, CATALOG_HEADER_SLOT, &record);
-
   overflow_vpid.pageid = CATALOG_GET_PGHEADER_OVFL_PGID_PAGEID (record.data);
   overflow_vpid.volid = CATALOG_GET_PGHEADER_OVFL_PGID_VOLID (record.data);
 
   if (overflow_vpid.pageid != NULL_PAGEID)
```
alexstanro, line 1620: The code from "while (overflow_vpid.pageid != NULL_PAGEID)" may be included here, to avoid checking overflow_vpid.pageid twice.

src/storage/page_buffer.c
```diff
@@ -8345,14 +8345,16 @@ pgbuf_respond_data_fetch_page_request (THREAD_ENTRY &thread_r, std::string &payl
       // add io_page
       payload_in_out.append (reinterpret_cast<const char *> (io_pgptr), (size_t) db_io_page_size ());
 
-      pgbuf_unfix (&thread_r, page_ptr);
-
       if (prm_get_bool_value (PRM_ID_ER_LOG_READ_DATA_PAGE))
 	{
 	  _er_log_debug (ARG_FILE_LINE,
-			 "[READ DATA] Successful while fixing page %d|%d, replication LSA = %lld|%d\n",
```
alexstanro, line 8353: You can use a new macro here that check PRM_ID_ER_LOG_READ_DATA_PAGE parameter and call er_log_debug.

alexstanro (approved):
The patch looks correct. 

I only have small remarks without affecting the correctness.

src/storage/system_catalog.c
cristiarg: Done

alexstanro (approved):

PR #3332, opened 2022-01-20
[LETS-278] Protect index lookups on PTS from page desynchronization

adrianbrab:
http://jira.cubrid.org/browse/LETS-278

There are two index traversion functions for reading purpose: `btree_locate_key` and `xbtree_find_unique`; there are other traversing functions that manipulate the index, but they're forbidden on PTS.

Both traversing functions rely on two other functions: `btree_get_root_with_key` and `btree_advance_and_find_key`. Modify them to check against fixing pages more advanced than the replication, on PTS only.

`btree_get_root_with_key:`
 - fixed with check for desync
 - if page is ahead set the return flag and wait for replication
 
The same goes for `btree_advance_and_find_key`.

### DIFF
--- a/src/storage/btree.c
+++ b/src/storage/btree.c
@@ -1256,8 +1256,7 @@ struct btree_helper
 
 STATIC_INLINE PAGE_PTR btree_fix_root_with_info (THREAD_ENTRY * thread_p, BTID * btid, PGBUF_LATCH_MODE latch_mode,
 						 VPID * root_vpid_p, BTREE_ROOT_HEADER ** root_header_p,
-						 BTID_INT * btid_int_p, bool check_for_desync)
-  __attribute__ ((ALWAYS_INLINE));
+						 BTID_INT * btid_int_p) __attribute__ ((ALWAYS_INLINE));
 
 STATIC_INLINE bool btree_is_fence_key (PAGE_PTR leaf_page, PGSLOTID slotid) __attribute__ ((ALWAYS_INLINE));
 
@@ -1805,7 +1804,7 @@ static bool btree_check_locking_for_delete_unique (THREAD_ENTRY * thread_p, cons
  */
 STATIC_INLINE PAGE_PTR
 btree_fix_root_with_info (THREAD_ENTRY * thread_p, BTID * btid, PGBUF_LATCH_MODE latch_mode, VPID * root_vpid_p,
-			  BTREE_ROOT_HEADER ** root_header_p, BTID_INT * btid_int_p, bool check_for_desync)
+			  BTREE_ROOT_HEADER ** root_header_p, BTID_INT * btid_int_p)
 {
   PAGE_PTR root_page = NULL;
   VPID vpid;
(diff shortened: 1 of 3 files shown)


src/storage/btree.c
```diff
@@ -22812,9 +22820,18 @@ btree_get_root_with_key (THREAD_ENTRY * thread_p, BTID * btid, BTID_INT * btid_i
...
-    btree_fix_root_with_info (thread_p, btid, PGBUF_LATCH_READ, NULL, &root_header, (reuse_btid_int ? NULL : btid_int));
+    btree_fix_root_with_info (thread_p, btid, PGBUF_LATCH_READ, NULL, &root_header, (reuse_btid_int ? NULL : btid_int),
+			      true);
   if (*root_page == NULL)
     {
+      if (er_errid () == ER_PAGE_AHEAD_OF_REPLICATION)
+	{
+	  *restart = true;
+	  constexpr VPID null_vpid = VPID_INITIALIZER;
+	  pgbuf_wait_for_replication (thread_p, &null_vpid);
```
cristiarg: Please change the name of `pgbuf_wait_for_replication` vpid argument as `optional_vpid_for_logging`. With the param named just `vpid` the function's API is confusing. My bad for suggesting this earlier. And add comment in the function comment header about this aspect.

src/storage/btree.c
```diff
@@ -1804,7 +1805,7 @@ static bool btree_check_locking_for_delete_unique (THREAD_ENTRY * thread_p, cons
  */
 STATIC_INLINE PAGE_PTR
 btree_fix_root_with_info (THREAD_ENTRY * thread_p, BTID * btid, PGBUF_LATCH_MODE latch_mode, VPID * root_vpid_p,
-			  BTREE_ROOT_HEADER ** root_header_p, BTID_INT * btid_int_p)
+			  BTREE_ROOT_HEADER ** root_header_p, BTID_INT * btid_int_p, bool check_for_desync)
```
cristiarg: I do not think the extra param is needed. If:
- on ATS, `pgbuf_fix_old_and_check_repl_desync` will behave just like `pgbuf_fix`
- on PTS, and desync is met, will do it's job of checking desync

What do you think? :)

src/storage/btree.c
```diff
@@ -1822,7 +1823,14 @@ btree_fix_root_with_info (THREAD_ENTRY * thread_p, BTID * btid, PGBUF_LATCH_MODE
...
   /* Fix root page. */
-  root_page = pgbuf_fix (thread_p, root_vpid_p, OLD_PAGE, latch_mode, PGBUF_UNCONDITIONAL_LATCH);
+  if (check_for_desync)
+    {
+      root_page = pgbuf_fix_old_and_check_repl_desync (thread_p, *root_vpid_p, latch_mode, PGBUF_UNCONDITIONAL_LATCH);
+    }
+  else
+    {
+      root_page = pgbuf_fix (thread_p, root_vpid_p, OLD_PAGE, latch_mode, PGBUF_UNCONDITIONAL_LATCH);
+    }
```
cristiarg: As suggested above.
```suggestion
      root_page = pgbuf_fix_old_and_check_repl_desync (thread_p, *root_vpid_p, latch_mode, PGBUF_UNCONDITIONAL_LATCH);
```

src/storage/btree.c
adrianbrab: you are right, my bad, `pgbuf_fix_old_and_check_repl_desync` already does this part inside

src/storage/btree.c
adrianbrab: Added a multi-line comment before the function and renamed the arg

src/storage/page_buffer.c
```diff
@@ -2347,8 +2347,14 @@ pgbuf_fix_old_and_check_repl_desync (THREAD_ENTRY * thread_p, const VPID & vpid,
   return page;
 }
 
+/*
+ * pgbuf_wait_for_replication () - Wait for replication to catch up
+ *   return: void
+ *   thread_p:                  thread pointer
+ *   optional_vpid_for_logging: Optional vpid argument used for logging, a null vpid can be provided if no valid value is available
```
cristiarg: if this line exceeds the maximum 120 length, wrap it

cristiarg (approved):

PR #3338, opened 2022-01-22
[CBRD-24134] Data transmission between flashback utility and server

joohok:
http://jira.cubrid.org/browse/CBRD-24134

**Purpose**
Data transmission between server and client for flashback. 

**Implementation**
sflashback_get_summary () 
- get classname list from client
  - xlocator_find_class_oid () : check whether class exists, and return OID of the class to the client. 
- get start, end time 
  - flashback_verify_time () : verifying the start, end time and get LOG_LSA from that time

sflashback_get_loginfo ()
- get required information to generate log infos for flashback
- returns loginfos

**Remarks**
N/A


src/communication/network_interface_cl.h
```diff
@@ -442,4 +443,9 @@ extern int loaddb_fetch_status (load_status & status);
 extern int loaddb_destroy ();
 extern int loaddb_interrupt ();
 extern int loaddb_update_stats ();
+
+extern int flashback_get_summary (dynamic_array * class_list, const char *user, time_t start_time, time_t end_time,
+				  void *summary_list, OID ** oid_list);
+extern int flashback_get_loginfo (int trid, char *user, OID * classlist, int num_table, LOG_LSA * start_lsa,
```
hornetmj:
```suggestion
extern int flashback_get_loginfo (int trid, char *user, OID * classlist, int num_class, LOG_LSA * start_lsa,
```

src/communication/network_interface_cl.c
```diff
@@ -71,6 +71,7 @@
 #include "xasl.h"
 #include "lob_locator.hpp"
 #include "crypt_opfunc.h"
+#include "dynamic_array.h"
```
```suggestion
```

hornetmj: already included in the network_interface_cl.h

src/communication/network_interface_cl.c
```diff
@@ -10560,3 +10561,160 @@ loaddb_update_stats ()
...
+  for (int i = 0; i < num_tables; i++)
+    {
+      if (da_get (class_list, i, table) != NO_ERROR)
+	{
+	  return ER_FAILED;
+	}
+      request_size += or_packed_string_length (table, NULL);
+    }
+
+  request = (char *) malloc (request_size + MAX_ALIGNMENT);
```
hornetmj, line 10597: plus sizeof ('num tables')

src/communication/network_interface_cl.c
```diff
@@ -10560,3 +10561,160 @@ loaddb_update_stats ()
...
+  ptr = or_pack_int64 (ptr, start_time);
+  ptr = or_pack_int64 (ptr, end_time);
+
+  request_size = ptr - start_ptr;
+
+  req_error =
+    net_client_request2 (NET_SERVER_FLASHBACK_GET_SUMMARY, start_ptr, request_size, reply,
+			 OR_ALIGNED_BUF_SIZE (a_reply), NULL, 0, &area, &area_size);
+
+  if (!req_error)
```
```suggestion
  if (req_error == NO_ERROR)
```

src/communication/network_interface_cl.c
```diff
@@ -10560,3 +10561,160 @@ loaddb_update_stats ()
...
+    {
+      ptr = or_unpack_int (reply, &area_size);
+      ptr = or_unpack_int (ptr, &rep_error);
```
hornetmj, line 10630:
```suggestion
      ptr = or_unpack_int (ptr, &rep_error);
      ptr = or_unpack_int (reply, &area_size);
```

hornetmj: I think it needs to change like that. (refers to the specification of the issue.)

src/communication/network_interface_sr.c
```diff
@@ -10698,3 +10698,231 @@ scdc_end_session (THREAD_ENTRY * thread_p, unsigned int rid, char *request, int
...
+  area = (char *) db_private_alloc (thread_p, area_size);	/* summary info size will be added */
+  if (area == NULL)
+    {
+      error_code = ER_OUT_OF_VIRTUAL_MEMORY;
+      goto error;
+    }
+
+  /* reply packing : error_code | area size */
+  ptr = or_pack_int (reply, area_size);
+  ptr = or_pack_int (ptr, error_code);
```
hornetmj, line 10823:
```suggestion
  ptr = or_pack_int (reply, error_code);
  ptr = or_pack_int (ptr, area_size);
```

src/communication/network_interface_sr.c
```diff
@@ -10698,3 +10698,231 @@ scdc_end_session (THREAD_ENTRY * thread_p, unsigned int rid, char *request, int
...
+    }
+
+  css_send_reply_and_data_to_client (thread_p->conn_entry, rid, reply, OR_ALIGNED_BUF_SIZE (a_reply), area, area_size);
+
+  db_private_free_and_init (thread_p, area);
+  db_private_free_and_init (thread_p, oid);
+
+  return;
+error:
+  ptr = or_pack_int (reply, 0);
```
hornetmj, line 10839: In my opinion, It seems that there is no need to send area_size information when an error occurs in the client. (If error_code is written before area_size)

src/executables/util_cs.c
```diff
@@ -4040,9 +4040,11 @@ flashback (UTIL_FUNCTION_ARG * arg)
   char table_name_buf[SM_MAX_IDENTIFIER_LENGTH];
   char *table_name;
 
+  OID *oid_list = NULL;
```
hornetmj, line 4043: I would like to give you one opinion. In the next PR, how about combining the variables used in the flashback operation into a single context?

src/communication/network_interface_cl.c
joohok: Oh, Sorry. I have change the specification in the issue. 
I've found this while testing, so I missed to change the specification. 

Network functions (net_client_request2 ()) requires the area_size to be packed first before the reply error.

src/communication/network_interface_cl.c
joohok: Thank you

src/communication/network_interface_cl.c
joohok: Oh.. Thank you, I missed it.

src/communication/network_interface_cl.c
joohok: Yes, I will remove it

src/communication/network_interface_cl.h
joohok: I will change it

src/communication/network_interface_sr.c
joohok: This is due to network interface in client (net_client_request2), which requires the order of packing area size and error code like above.

src/communication/network_interface_sr.c
joohok: Yes, I understand what you said. but I have to follow the 'reply' packing order that net_client_request2() require.

src/executables/util_cs.c
joohok: Ok, that will be more readable when pass these variables as arguments. 
I will follow it. 
In this PR, these variables are used for testing.

src/communication/network_interface_cl.c
```diff
@@ -10560,3 +10561,160 @@ loaddb_update_stats ()
...
+  request_size =
+    OR_INT_SIZE + or_packed_string_length (user,
+					   NULL) + OR_OID_SIZE * num_table + OR_LOG_LSA_ALIGNED_SIZE * 2 + OR_INT_SIZE +
+    OR_INT_SIZE;
+
+  request = (char *) malloc (request_size + MAX_ALIGNMENT);
+  if (request == NULL)
+    {
+      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, (size_t) request_size);
+      return ER_FAILED;
```
```suggestion
      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, (size_t) request_size);
      return ER_OUT_OF_VIRTUAL_MEMORY;
```
It doesn't need to match?

src/communication/network_interface_cl.c
```diff
@@ -10560,3 +10561,160 @@ loaddb_update_stats ()
...
+	  return ER_FAILED;
+	}
+      request_size += or_packed_string_length (table, NULL);
+    }
+
+  request = (char *) malloc (request_size + MAX_ALIGNMENT);
+  if (request == NULL)
+    {
+      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, (size_t) request_size);
+      return ER_FAILED;
```
```suggestion
      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, (size_t) request_size);
      return ER_OUT_OF_VIRTUAL_MEMORY;
```
It doesn't need to match?

src/communication/network_interface_cl.c
```diff
@@ -10560,3 +10561,160 @@ loaddb_update_stats ()
...
+  num_tables = da_size (class_list);
+
+  /* request : num tables | table name list | user | start_time | end_time */
+  request_size = or_packed_string_length (user, NULL) + OR_INT64_SIZE + OR_INT64_SIZE;
+
+  for (int i = 0; i < num_tables; i++)
+    {
+      if (da_get (class_list, i, table) != NO_ERROR)
+	{
+	  return ER_FAILED;
```
wrlawodms, line 10592: What makes you decide to use `er_set` or not? You use it for some and don't for others.

src/communication/network_interface_cl.c
```diff
@@ -10560,3 +10561,160 @@ loaddb_update_stats ()
...
+  char table[SM_MAX_IDENTIFIER_LENGTH];
+
+  OR_ALIGNED_BUF (OR_INT_SIZE + OR_INT_SIZE) a_reply;
+  char *reply = OR_ALIGNED_BUF_START (a_reply);
+  char *area;
+  int area_size;
+
+  num_tables = da_size (class_list);
+
+  /* request : num tables | table name list | user | start_time | end_time */
```
wrlawodms: I think variable names should have consistency. Some use term `class`, others use `table`.

src/communication/network_interface_cl.c
```diff
@@ -10560,3 +10561,160 @@ loaddb_update_stats ()
...
+	    }
+	  /* get summary info */
+	}
+
+      free_and_init (area);
+    }
+
+  free_and_init (request);
+
+  return rep_error;
```
wrlawodms: It seems to have to return `req_error` if `req_error != NO_ERROR`.

src/communication/network_interface_sr.c
```diff
@@ -10698,3 +10698,231 @@ scdc_end_session (THREAD_ENTRY * thread_p, unsigned int rid, char *request, int
...
+  for (int i = 0; i < num_table; i++)
+    {
+      char *classname = NULL;
+
+      ptr = or_unpack_string_nocopy (ptr, &classname);
+
+      status = xlocator_find_class_oid (thread_p, classname, &oid[i], NULL_LOCK);
+      if (status != LC_CLASSNAME_EXIST)
+	{
+	  error_code = ER_FLASHBACK_INVALID_CLASS;
```
wrlawodms, line 10798: No need to leave an error message with `er_set`? IMO, If an error comes from another function, we can expect it to leave the error inside, but if we set the error for the first time, we should leave an error message for it.

src/communication/network_interface_sr.c
```diff
@@ -10698,3 +10698,231 @@ scdc_end_session (THREAD_ENTRY * thread_p, unsigned int rid, char *request, int
...
+		       LOG_LSA * end_lsa)
+{
+  int error_code = NO_ERROR;
+  time_t ret_time = 0;
+
+
+  /* 1. Check start_time */
+  if (start_time < log_Gl.hdr.db_creation)
+    {
+      return ER_FLASHBACK_INVALID_TIME;
```
wrlawodms, line 10714: er_set()?

wrlawodms: and for other errors in this function. Or, do you mean to deal with errors outside?

src/communication/network_interface_sr.c
```diff
@@ -10698,3 +10698,231 @@ scdc_end_session (THREAD_ENTRY * thread_p, unsigned int rid, char *request, int
...
+    {
+      ret_time = start_time;
+
+      error_code = cdc_find_lsa (thread_p, &ret_time, start_lsa);
+      if (error_code == NO_ERROR || error_code == ER_CDC_ADJUSTED_LSA)
+	{
+	  /* find log record at the time
+	   * ret_time : time of the commit record or dummy record that is found to be greater than or equal to the start_time at first */
+
+	  if (ret_time > end_time)
```
wrlawodms: 1. Q. What does `start_lsa` mean exactly from `cdc_find_lsa()`? the LSA of the first downtime record after `start_time` or before?
2. What if start_time == end_time. It's valid? and this case could be validated on the client-side.

src/communication/network_interface_sr.c
```diff
@@ -10698,3 +10698,231 @@ scdc_end_session (THREAD_ENTRY * thread_p, unsigned int rid, char *request, int
...
+	    {
+	      /* out of range : start_time (ret_time) can not be greater than end_time */
+	      return ER_FLASHBACK_INVALID_TIME;
+	    }
+	}
+      else
+	{
+	  /* failed to find a log at the time due to failure while reading log page or volume (ER_FAILED, ER_LOG_READ) */
+	  return ER_FAILED;
```
wrlawodms, line 10737: What if it can't find the log record because the time is too large or small? It isn't `ER_FLASHBACK_INVALID_TIME`?

src/communication/network_interface_sr.c
joohok: No, I will use er_set() here. 
I just do append er_set() functions after setting error messages for each errors, which will be done in other issue. 
because error message format is not decided yet.

src/communication/network_interface_cl.c
joohok: yes, I agree. I will change it

src/communication/network_interface_cl.c
joohok: rep_error is initialized as ER_FAILED, so I think it is ok to return rep_error when request failed, like other code. 
but it seems more appropriate to return req_error when request failed.

src/communication/network_interface_cl.c
joohok: OK, I will change it. Thanks

src/communication/network_interface_cl.c
joohok: Yes, that was what @hornetmj  said. I will change the variables for consistency.

src/communication/network_interface_sr.c
joohok: 1. first downtime record **_after_** `start_time` 
2. comparing start_time and end_time is validated in client-side. (start_time >= end_time)
    I think start_time and end_time are able to have same value, and if they have same value, flashback just return at most one record.

src/communication/network_interface_sr.c
joohok: validate if the time is too small with comparing `db_creation` .
and compare `end_time` to see if time is too large. 
end_time should be less than current time.

src/communication/network_interface_sr.c
joohok: I agree with you, I will leave an error message when it is set. 
I thought er_set() should be done after setting error messages for each error code. 

However, it seems better to add er_set() first and modify it according to the message format when the error message is determined. Thanks.

src/communication/network_interface_sr.c
wrlawodms: If they are the same, the only log record we can get is a donetime record. and it is meaningless. Isn't it?

src/communication/network_interface_sr.c
wrlawodms: OK. I recommend you leave a TODO comment or some noticeable code in where you want to handle later. There are so many, and it seems to be easy to lose track of. Just a recommendation, it's up to you.

src/communication/network_interface_cl.c
wrlawodms: OK, let's go with your opinion. Other codes deal with it in their own way..

src/communication/network_interface_sr.c
wrlawodms: OK!

src/communication/network_interface_sr.c
wrlawodms: + I thought `start_time`- 1 sec is used here.

src/communication/network_interface_sr.c
joohok: Thank you, I will leave some mark.

src/communication/network_interface_sr.c
joohok: yes, it is meaningless. 

`start_time (returned from cdc_find_lsa) -1 ` was decided to use when making summary info per transaction, when fill the start_time of the transaction. 
 
but it is better to do here too, and maybe client-side validation (start_time >= end_time) will be modified to (start_time > end_time).

hornetmj (approved):

src/communication/network_interface_sr.c
joohok: sorry for above comment. please ignore it. 
1. (start_time <  end_time) is right  : start_time never be same with end_time. 
2. returned time from cdc_find_lsa() can not be same with end_time. 
    -  `if (ret_time >= end_time) return error; `

joohok: 3. cdc_find_lsa() returns `start_lsa`
  - compared with log record containing time info : LOG_COMMIT, LOG_ABORT, or LOG_DUMMY_HA_SERVER_STATE  
  - traversing logs in time order and if input time >= time of log record, then return time of log record, and LSA. 

So, I think `start_time -1` is not required here, because it can return LSA of former log record.

eido5 (approved):

wrlawodms (approved):

PR #3340, opened 2022-01-24
[CBRD-24180] Change unloaddb schema file output regarding newline

kisoo-han:
http://jira.cubrid.org/browse/CBRD-24180

**Purpose**
Improved loaddb schema file output readability from a schema perspective.

1. Add one or more newline if the operation is for new object compare to previous line.
2. Remove new line, if subject object is for same object with the previous line

**Implementation**

**Remarks**
Refer attached files of http://jira.cubrid.org/browse/CBRD-24180, for ASIS and TOBE demodb_schema for changes.


mhoh3963 (approved):
LGTM

PR #3342, opened 2022-01-25
[CBRD-24189] Support Unix Domain Socket for Java SP server

hgryoo:
http://jira.cubrid.org/browse/CBRD-24189


src/base/system_parameter.c
```diff
@@ -6017,13 +6017,13 @@ static SYSPRM_PARAM prm_Def[] = {
    (char *) NULL,
    (DUP_PRM_FUNC) NULL,
    (DUP_PRM_FUNC) NULL},
-  {PRM_ID_JAVA_STORED_PROCEDURE_RESERVE_01,
-   PRM_NAME_JAVA_STORED_PROCEDURE_RESERVE_01,
-   (PRM_FOR_SERVER | PRM_HIDDEN),
+  {PRM_ID_JAVA_STORED_PROCEDURE_UDS,
+   PRM_NAME_JAVA_STORED_PROCEDURE_UDS,
+   (PRM_FOR_SERVER),
```
ctshim, line 6022: Since Windows can use only TCP and UNIX (LINUX) uses UDS by default, it would be better to set it as a hidden property and specify TCP only for specific test purposes.

src/jsp/jsp_comm.c
```diff
@@ -263,6 +221,134 @@ jsp_readn (SOCKET fd, void *vptr, int n)
...
+  sockfd = socket (saddr->sa_family, SOCK_STREAM, 0);
+  if (IS_INVALID_SOCKET (sockfd))
+    {
+      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_SP_CANNOT_CONNECT_JVM, 1, "socket()");
+      return INVALID_SOCKET;
+    }
+  else
+    {
+      b = 1;
+      setsockopt (sockfd, IPPROTO_TCP, TCP_NODELAY, (char *) &b, sizeof (b));
```
beyondykk9, line 339: need to check the return value, and the variable b looks not familiar

src/jsp/com/cubrid/jsp/Server.java
```diff
@@ -53,68 +54,69 @@
...
         serverName = name;
         spPath = path;
         rootPath = rPath;
         shutdown = new AtomicBoolean(false);
 
-        try {
-            int port_number = Integer.parseInt(port);
-            serverSocket = new ServerSocket(port_number);
+        if (OSValidator.IS_UNIX) {
+            String socketName = rPath + "/tmp/junixsocket-" + name + ".sock";
```
ctshim: Please refer to the jsp_get_socket_file_path() function in jsp_comm.c. 
"/tmp" is used only when there is no result of envvar_get ("TMP"), but here "/tmp" is used unconditionally. 
The process of checking the value of envvar_get ("TMP") should be added.

src/jsp/com/cubrid/jsp/OSValidator.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+package com.cubrid.jsp;
+
+public class OSValidator {
+    private static String OS = System.getProperty("os.name").toLowerCase();
+    public static boolean IS_WINDOWS = (OS.indexOf("win") >= 0);
+    public static boolean IS_MAC = (OS.indexOf("mac") >= 0);
+    public static boolean IS_UNIX =
+            (OS.indexOf("nix") >= 0 || OS.indexOf("nux") >= 0 || OS.indexOf("aix") > 0);
```
hyunikn, line 39: Is > (rather than >=) for aix intended?

hyunikn (approved):

shparkcubrid (approved):

beyondykk9 (approved):

src/jsp/com/cubrid/jsp/Server.java
hgryoo: I've fixed by passing the envvar_get("TMP") value to the Java SP Server

src/jsp/com/cubrid/jsp/OSValidator.java
hgryoo: I've fixed OSValidator

ctshim (approved):

youngjinj (approved):

PR #3344, opened 2022-01-25
[CBRD-24170] Remove redundant logical code in the log_recovery_analysis function.

42-saoh:
http://jira.cubrid.org/browse/CBRD-24170

### Purpose
Remove redundant logical code.
log_recovery_analysis function in log_recovery.c
After passing the if statement, it is checked with assert_release() under the same condition.

### Implementation
removed the redundant assert_release()

### Remarks
N/A

### DIFF
--- a/src/transaction/log_recovery.c
+++ b/src/transaction/log_recovery.c
@@ -2864,7 +2864,6 @@ log_recovery_analysis (THREAD_ENTRY * thread_p, LOG_LSA * start_lsa, LOG_LSA * s
 	    }
 	  if (LSA_EQ (end_redo_lsa, &lsa))
 	    {
-	      assert_release (!LSA_EQ (end_redo_lsa, &lsa));
 	      LSA_SET_NULL (&lsa);
 	      break;
 	    }


hornetmj (approved):

joohok (approved):

wrlawodms (commented):
The if-condition and the assertion are not redundant but contradict (opposite). We have to make sure which is true to fix it. I think this change may be right, but we need reasoning and description about the context of this code. The existing code says `assert_release (false)`, which could mean an error case.

hornetmj:
@wrlawodms When I checked the code after reading your comments, we need to check that the end_redo_lsa can be equal to the lsa. The followings are the code which are set the related variables. Please check it for me.
```
LSA_COPY (end_redo_lsa, &lsa);
LSA_COPY (&lsa, &log_rec->forw_lsa);
```

wrlawodms:
@hornetmj Looking through `log_recovery_analysis()`, there seems to be no chance that they are the same. However, to make sure, we have to consider other log mechanisms. I'm not sure that `log_rec->forw_lsa` can't be the same as its own lsa. There seem to be some tricky ways regarding archiving and flushing. It needs more analysis or I hope that @joohok can answer it.

joohok:
In general, `log_rec->forw_lsa` cannot be equal to `lsa of log_rec`, because forw_lsa is written at end of the appending log record using `log_Gl.prior_info.prior_lsa`. 
Some corruption may cause LSA to be NULL while it is not a LOG_END_OF_LOG. (I guess it occurs when the failure of flushing log page buffer to the volume)

And I think `end_redo_lsa` and `lsa` has same values  just when they have NULL_LSA. So, I think `end_redo_lsa` and `lsa` are required to be different. 

we need more time to investigate about this.

eido5 (approved):

hornetmj:
@42-saoh I'm going to postpone the merging of this issue a bit. It's a simple fix, but I'm not going to make it urgent because we might lose information written to the error log. This part will be confirmed by @joohok, and will be merged when completed.

@joohok Please check it on fig.

PR #3349, opened 2022-01-26
[CBRD-24195] Implement User Schema feature

youngjinj:
http://jira.cubrid.org/browse/CBRD-24195

Independently manage user objects with user_specified_name of "owner_name + dot(.) + object_name".
Added unique_name column to manage user_specified_name by object.

Objects that can be managed independently are:
- table, view, serial, trigger

There are a number of changes to finding and using objects by user_specified_name.


msg/de_DE.utf8/utils.msg
```diff
@@ -490,7 +490,9 @@ gültige Optionen:\n\
   -d, --data-file=DATEI            DATEI laden\n\
   -t, --table=TABLE                Name der Tabelle, die für das fehlende Klassenheader in der Datei ersetzt wird\n\
       --error-control-file=DATEI   DATEI für Fehlerkontrolle während Ladung\n\
-      --ignore-class-file=DATEI    Eingangsdatei für Klassenamen, die nicht geladen werden\n
+      --ignore-class-file=DATEI    Eingangsdatei für Klassenamen, die nicht geladen werden\n\
+      --no-user-specified-name     Find class, serial, and trigger by object-name and don't use user-specified-name.\n
```
shparkcubrid: How about below?
no-use-user-schema For backward compatibility with the load file of the previous version of V11.2.

src/base/object_representation.h
```diff
@@ -757,22 +757,23 @@ enum
...
-  ORC_METHODS_INDEX = 8,
-  ORC_CLASS_METHODS_INDEX = 9,
-  ORC_METHOD_FILES_INDEX = 10,
-  ORC_RESOLUTIONS_INDEX = 11,
-  ORC_QUERY_SPEC_INDEX = 12,
-  ORC_TRIGGERS_INDEX = 13,
-  ORC_PROPERTIES_INDEX = 14,
-  ORC_COMMENT_INDEX = 15,
-  ORC_PARTITION_INDEX = 16,
+  ORC_SIMPLE_NAME_INDEX = 1,
```
shparkcubrid, line 760: Is there any reason to put a new value in the middle?

src/broker/cas_schema_info.h
```diff
@@ -32,11 +32,11 @@
 #include "cas_net_buf.h"
 
 #if defined(CAS_FOR_ORACLE) || defined(CAS_FOR_MYSQL)
-#define SCH_STR_LEN	255
+#define SCH_STR_LEN	287
 #define SCH_CLASS_REMARKS_STR_LEN       2048
 #define SCH_REMARKS_STR_LEN     1024
 #else
-#define SCH_STR_LEN	DB_MAX_IDENTIFIER_LENGTH
+#define SCH_STR_LEN	DB_MAX_IDENTIFIER_LENGTH_287
```
shparkcubrid: 'DB_MAX_IDENTIFIER_LENGTH_287' is too meaningless.
Can't 'DB_MAX_IDENTIFIER_LENGTH' changed to 287?

msg/en_US.utf8/cubrid.msg
```diff
@@ -1874,6 +1874,7 @@ $set 8 MSGCAT_SET_PARSER_SEMANTIC
 304 Could not find compatible signature for function %1$s()
 305 Signature %1$s is not compatible with function arguments.
 306 dblink - Required information is missing.
+400 A member of owner group or a member of DBA group can create table.
```
ctshim: It should be 307 in order, so why skip to 400?

src/base/system_parameter.c
```diff
@@ -2390,6 +2392,11 @@ static int prm_thread_core_count_upper = 1024;
 int PRM_THREAD_CORE_COUNT = prm_thread_core_count_default;	// this value will be tuned
 static unsigned int prm_thread_core_count_flag = 0;
 
+bool PRM_NO_USER_SPECIFIED_NAME = false;
+/* set to true for testing. It should be false when the test is complete. (by youngjinj) */
+static bool prm_no_user_specified_name_default = true;
```
ctshim: If the value is not permanently changed, it is better to declare it as "static const".

src/parser/csql_grammar.y
```diff
@@ -21791,6 +21796,11 @@ identifier
 
 			    size_in = strlen(str_name);
 
+			    if (!strchr (str_name, '.') && size_in >= (DB_MAX_IDENTIFIER_LENGTH - 1))
+			      {
+				str_name[DB_MAX_IDENTIFIER_LENGTH - 1] = '\0';	// truncate
+			      }
+
```
ctshim: I would like to comment on the same structure here and below.

It comes after performing a 286 length excess check in csql_lexer.l .
Is there any reason to check DB_MAX_IDENTIFIER_LENGTH again here?

src/object/trigger_manager.c
```diff
@@ -4225,6 +4232,24 @@ tr_find_trigger (const char *name)
...
     {
+      /* This is the case when the loaddb utility is executed with the --no-user-specified-name option as the dba user. */
+      if (db_get_client_type () == DB_CLIENT_TYPE_ADMIN_UTILITY && prm_get_bool_value (PRM_ID_NO_USER_SPECIFIED_NAME))
+	{
+	  if (object == NULL)
+	    {
+	      char other_trigger_name[DB_MAX_IDENTIFIER_LENGTH_287] = { '\0' };
+
+	      do_find_trigger_by_query (name, other_trigger_name, DB_MAX_IDENTIFIER_LENGTH_287);
+	      if (other_trigger_name)
```
hgryoo: Do you intend to the following?
```suggestion
	      if (other_trigger_name[0] != '\0')
```

src/transaction/locator_cl.c
```diff
@@ -3141,15 +3144,33 @@ locator_find_class_by_name (const char *classname, LOCK lock, MOP * class_mop)
...
     }
 
-  return class_mop;
+  /* This is the case when the loaddb utility is executed with the --no-user-specified-name option as the dba user. */
+  if (db_get_client_type () == DB_CLIENT_TYPE_ADMIN_UTILITY && prm_get_bool_value (PRM_ID_NO_USER_SPECIFIED_NAME))
+    {
+      char other_class_name[DB_MAX_IDENTIFIER_LENGTH_287] = { '\0' };
+
+      do_find_class_by_query (classname, other_class_name, DB_MAX_IDENTIFIER_LENGTH_287);
+      if (other_class_name)
```
hgryoo:
```suggestion
      if (other_class_name[0] != '\0')
```

src/parser/scanner_support.c
```diff
@@ -617,12 +617,28 @@ pt_check_hint (const char *text, PT_HINT hint_table[], PT_HINT_ENUM * result_hin
...
 					{
+
 					  *temp = '\0';
 					  arg->info.name.resolved = pt_trim_as_identifier (arg_start);
 					  arg->info.name.resolved = pt_makename (arg->info.name.resolved);
-					  *temp++ = '.';
+					  *temp = '.';
+
+					  *temp++;
+
```
hgryoo:
```suggestion
					  *temp++ = '.';
```

src/parser/scanner_support.c
```diff
@@ -676,6 +692,17 @@ pt_check_hint (const char *text, PT_HINT hint_table[], PT_HINT_ENUM * result_hin
...
+				  /*
+				   * "user_name + dot(.)" is added in front of the table name to distinguish
+				   * the user schema. So we cannot separate index names with dot(.).
+				   * dot(.) does not appear more than 2 times, and the last name is the index name.
+				   */
+				  if (temp && strstr (temp + 1, ".") != NULL)
+				    {
+				      temp = strstr (temp + 1, ".");
+				    }
```
hgryoo: `strstr()` is performed twice unnecessary.

src/broker/cas_schema_info.h
youngjinj: Remove DB_MAX_IDENTIFIER_LENGTH_287 and only use DB_MAX_IDENTIFIER_LENGTH.
The maximum length of class_name is changed from 255 bytes to 223 bytes.

msg/en_US.utf8/cubrid.msg
youngjinj: This is a margin to avoid error number collisions when merging.
However, it seems unnecessary to do this, so I will change the numbers consecutively.

src/base/system_parameter.c
youngjinj: Since it is a default value, it seems appropriate to set const. Why didn't the existing codes use const?
Let's set it to const.

src/object/trigger_manager.c
youngjinj: I will change it.

src/transaction/locator_cl.c
youngjinj: I will change it.

src/parser/scanner_support.c
youngjinj: Are there any other benefits besides shorter code?
Wouldn't it be more readable to write it separately?

hgryoo: If you want to separate assigning value and increment pointer, `*temp++;` seems not the intended code. No need to use dereference (`*`). `temp++;`.

youngjinj: I used dereference unnecessarily while separating existing code.
Is it ok if I do it like below without using dereference?

```
*temp = '.';
temp++;
```

src/parser/csql_grammar.y
youngjinj: It's a change from a recent commit, but I'd like to explain it.

Except for user_specified_name, other identifiers are limited to 255 characters.
In order to store user_specified_name as it is, a maximum of 287 bytes is allowed in csql_lexer.l.
However, identifiers that should be limited to 255 bytes can be allowed up to 287 bytes, so check once more in csql_grammar.y.
If not user_specified_name, truncate it to 255 bytes.

During team discussion, we decided to accept the suggestion that user_specified_name should also be limited to 255 bytes.
So there is no need to allow up to 287 bytes in csql_lexer.l, and there is no need to check again in csql_grammar.y.
Because we limit user_specified_name to 255 bytes, class_name stores 223 bytes excluding 32 bytes storing user_name.

msg/de_DE.utf8/utils.msg
youngjinj: Yes, I think it should be included. I will add.

src/parser/semantic_check.c
```diff
@@ -8359,6 +8368,24 @@ pt_check_create_entity (PARSER_CONTEXT * parser, PT_NODE * node)
...
+      MOP owner = NULL;
+      DB_SET *groups = NULL;
+
+      owner = db_find_user (owner_name);
+
+      if (ws_is_same_object (owner, Au_user) == false)
+	{
+	  if (au_is_user_group_member (owner, Au_user) == false)
+	    {
+	      PT_ERRORm (parser, name, MSGCAT_SET_PARSER_SEMANTIC, MSGCAT_SEMANTIC_NOT_MEMBER_CREATE_NOT_ALLOWED);
```
hgryoo: return is not needed?
```suggestion
	      PT_ERRORm (parser, name, MSGCAT_SET_PARSER_SEMANTIC, MSGCAT_SEMANTIC_NOT_MEMBER_CREATE_NOT_ALLOWED);
	      return;
```

src/query/execute_statement.c
```diff
@@ -6290,7 +6360,7 @@ int
 do_create_trigger (PARSER_CONTEXT * parser, PT_NODE * statement)
 {
   PT_NODE *cond, *action, *target, *attr, *pri, *comment_node;
-  const char *name, *comment;
+  const char *name, *target_class_name, *comment;
```
hgryoo: redundant change
```suggestion
  const char *name, *comment;
```

src/parser/semantic_check.c
youngjinj: Yes. I'll add it.

src/query/execute_statement.c
youngjinj: No need. I will remove it.

src/parser/scanner_support.c
ctshim: {*temp = '1'; *temp++;} and {*temp = '2';} are equivalent.
{*temp++ = '1';} and {*temp = '1'; temp++;} is the same.

src/object/schema_template.c
```diff
@@ -699,13 +699,14 @@ def_class_internal (const char *name, int class_type)
   char realname[SM_MAX_IDENTIFIER_LENGTH];
   SM_TEMPLATE *template_ = NULL;
   PR_TYPE *type;
+  const char *simple_name = sm_simple_name (name);
 
   if (sm_check_name (name))
     {
-      type = pr_find_type (name);
+      type = pr_find_type (simple_name);
```
ctshim: I think it's better to modify it as follows.
```
  if (sm_check_name (name))
    {
      const char *simple_name = sm_simple_name (name);
      type = pr_find_type (simple_name);
```

src/executables/unload_schema.c
```diff
@@ -3157,7 +3286,14 @@ emit_partition_info (print_output & output_ctx, MOP clsobj)
       return;
     }
 
-  output_ctx ("\nALTER CLASS %s%s%s ", PRINT_IDENTIFIER (name));
+  class_name_copy = strdup (name);
+  owner_name_p = strtok_r (class_name_copy, ".", &save_token);
+  class_name_p = strtok_r (NULL, ".", &save_token);
+
```
beyondykk9:
```suggestion
  class_name_copy = strdup (name);
  owner_name_p = strtok_r (class_name_copy, ".", &save_token);
  class_name_p = strtok_r (NULL, ".", &save_token);
```
It looks better to replace with a macro for repeated routine.

src/loaddb/load_server_loader.cpp
```diff
@@ -703,12 +703,12 @@ namespace cubload
 	m_error_handler.log_date_time_conversion_error (token, pr_type_name (attr.get_domain ().type->get_id ()));
       }
     else if (error_code == ER_OBJ_ATTRIBUTE_CANT_BE_NULL)
-           {
-              char class_attr[512];
+      {
+	char class_attr[512];
```
beyondykk9: It looks better to replace with a macro for the hard coded value 512.

src/object/schema_manager.c
```diff
@@ -3032,6 +3117,144 @@ sm_is_system_class (MOP op)
...
+sm_check_system_class_by_name (const char *name)
+{
+  // *INDENT-OFF*
+  static
+  struct system_class_list
+  {
+    const char *name;
+    int len;
+  }
+  system_classes[] =
```
byungwook-kim, line 3136: This is a local variable and it is always initialized whenever it's called. I think it looks better to move the local variable to a static global variable.

src/object/schema_manager.c
```diff
@@ -3032,6 +3117,144 @@ sm_is_system_class (MOP op)
   return sm_get_class_flag (op, SM_CLASSFLAG_SYSTEM);
 }
 
+/*
+ * sm_check_system_class_by_name () - Checks whether the class name is
+ *    the same as the system class name.
+ * return: int
+ * name(in): class simple name
+ */
+int
```
beyondykk9:
```suggestion
bool
```
this function looks return true or false

src/object/trigger_description.cpp
```diff
@@ -255,11 +259,17 @@ tr_dump_trigger (print_output &output_ctx, DB_OBJECT *trigger_object)
     {
       /* automatically filter out invalid triggers */
 
+      trigger_name_copy = strdup (trigger->name);
+      owner_name_p = strtok_r (trigger_name_copy, ".", &save_token);
+      trigger_name_p = strtok_r (NULL, ".", &save_token);
```
ctshim, line 264: How about modifying it like below?
     ```
 owner_name_p = trigger_name_copy:
      trigger_name_p = strchr (trigger_name_copy, '.');
      *trigger_name_p = '\0';
      trigger_name_p++;
```

Comment (ctshim) [inline src/object/trigger_description.cpp]: How about modifying it like below?

```
owner_name_p = trigger_name_copy:
trigger_name_p = strchr (trigger_name_copy, '.');
*trigger_name_p = '\0';
trigger_name_p++;
```

File: src/loaddb/load_db.c [hunk 19]
```diff
@@ -549,6 +549,10 @@ loaddb_internal (UTIL_FUNCTION_ARG * arg, int dba_mode)
   /* login */
   if (!args.user_name.empty () || !dba_mode)
     {
+      if (strcasecmp (args.user_name.c_str (), "DBA") == 0)
```
shparkcubrid, line 552: Can't it check with dba_mode?

youngjinj: dba_mode is only set when calling the loaddb_dba() function.
loaddb_dba () is set as an unused function.
So we can only check by username.

src/parser/scanner_support.c
youngjinj: "*temp++;" was the wrong change.
To improve readability, I wanted to separate value assignment and repositioning.

src/parser/scanner_support.c
youngjinj: Looking back now, I don't need to add this code.
I will comment again after more checking.

src/object/schema_template.c
youngjinj: Okay, I'll change it.

src/object/schema_manager.c
youngjinj: Okay, I'll change it.

src/object/trigger_manager.c
```diff
@@ -6847,12 +6879,25 @@ tr_rename_trigger (DB_OBJECT * trigger_object, const char *name, bool call_from_
 	      trigger->name = newname;
 	      db_make_string (&value, newname);
 	      newname = NULL;
-	      error = db_put_internal (trigger_object, TR_ATT_NAME, &value);
-	      if (error == NO_ERROR)
+
+	      db_make_string (&value, trigger->name);
+	      error = db_put_internal (trigger_object, TR_ATT_FULL_NAME, &value);
+	      if (error != NO_ERROR)
```
ctshim: It should be modified in the following way.
```
error = db_put_internal (trigger_object, TR_ATT_FULL_NAME, &value);
if (error == NO_ERROR)
{
    db_make_string (&value, sm_simple_name (trigger->name));
    error = db_put_internal (trigger_object, TR_ATT_NAME, &value);
}
```

src/object/schema_manager.c
```diff
@@ -2168,6 +2169,89 @@ sm_downcase_name (const char *name, char *buf, int maxlen)
   intl_identifier_lower (name, buf);
 }
 
+/* AUTO_INCREMENT_SERIAL_NAME_MAX_LENGTH is not a target. 
+ * Trigger and serial names can be longer than DB_MAX_IDENTIFIER_LENGTH.
+ */
+int
+sm_user_specified_name (const char *name, const char *owner_name, char *buf, int buf_size)
```
ctshim: In all cases where this function is called, owner_name is specified as NULL. 
If it is unnecessary, it is better to organize it.

src/executables/unload_schema.c
```diff
@@ -773,7 +780,15 @@ export_serial (print_output & output_ctx)
 
       output_ctx ("call [find_user]('%s') on class [db_user] to [auser];\n",
 		  db_get_string (&values[SERIAL_OWNER_NAME]));
-      output_ctx ("create serial %s%s%s\n", PRINT_IDENTIFIER (db_get_string (&values[SERIAL_NAME])));
+
+      serial_name_copy = strdup (db_get_string (&values[SERIAL_FULL_NAME]));
```
ctshim: Rather than using strdup(), it is more efficient to prepare a static array in advance and use memcpy().

src/executables/unload_schema.c
```diff
@@ -773,7 +780,15 @@ export_serial (print_output & output_ctx)
...
+      owner_name_p = strtok_r (serial_name_copy, ".", &save_token);
+      serial_name_p = strtok_r (NULL, ".", &save_token);
```
ctshim: Consider using strchr()

src/executables/unload_schema.c
```diff
@@ -1122,7 +1141,14 @@ emit_schema (print_output & output_ctx, DB_OBJLIST * classes, int do_auth, DB_OB
...
 
-      output_ctx ("CREATE %s %s%s%s", is_vclass ? "VCLASS" : "CLASS", PRINT_IDENTIFIER (name));
+      class_name_copy = strdup (name);
+      owner_name_p = strtok_r (class_name_copy, ".", &save_token);
+      class_name_p = strtok_r (NULL, ".", &save_token);
+
+      output_ctx ("CREATE %s %s%s%s.%s%s%s", is_vclass ? "VCLASS" : "CLASS", PRINT_IDENTIFIER (owner_name_p),
+		  PRINT_IDENTIFIER (class_name_p));
+
+      free_and_init (class_name_copy);
```
ctshim: Please refer to the code below
```
      char owner_name_p[512]; 
      strcpy(owner_name_p, name);
      class_name_p = strtchr (owner_name_p, '.');
      *class_name_p++ = '\0';

      output_ctx ("CREATE %s %s%s%s.%s%s%s", is_vclass ? "VCLASS" : "CLASS", PRINT_IDENTIFIER (owner_name_p),
		  PRINT_IDENTIFIER (class_name_p));
```

src/object/schema_manager.c
```diff
@@ -2168,6 +2169,89 @@ sm_downcase_name (const char *name, char *buf, int maxlen)
...
+
+  dot = (char *) strchr (name, '.');
+  if (dot)
+    {
+      name_size = intl_identifier_lower_string_size (name);
+      assert (name_size < buf_size);
+      assert (name_size < DB_MAX_IDENTIFIER_LENGTH);
+
+      intl_identifier_lower (name, buf);
+
```
ctshim: How about replacing it like below?
`sm_downcase_name (name, buf, buf_size);`

src/object/schema_manager.c
```diff
@@ -2168,6 +2169,89 @@ sm_downcase_name (const char *name, char *buf, int maxlen)
...
+  if (sm_check_system_class_by_name (name)
+      && (owner_name == NULL || owner_name[0] == '\0' || intl_identifier_casecmp (owner_name, "DBA") == 0))
+    {
+      name_size = intl_identifier_lower_string_size (name);
+      assert (name_size < buf_size);
+      assert (name_size < DB_MAX_IDENTIFIER_LENGTH);
+
+      intl_identifier_lower (name, buf);
+
```
ctshim: How about replacing it like below?
`sm_downcase_name (name, buf, buf_size);`

src/object/authenticate.c
```diff
@@ -5129,31 +5175,63 @@ au_change_owner (MOP classmop, MOP owner)
...
+
+      /*
+       * Since class_full_name includes owner_name, if owner of class is changed,
+       * class_full_name must be changed as well.
+       */
+      class_name = sm_ch_simple_name ((MOBJ) class_);
 
-	  /* Change class owner */
-	  class_->owner = owner;
+      if (sm_check_system_class_by_name (class_name))
```
ctshim: If (sm_check_system_class_by_name (class_name)) is not satisfied, do I not have to perform locator_flush_class (classmop)?

src/object/authenticate.c
```diff
@@ -5129,31 +5175,63 @@ au_change_owner (MOP classmop, MOP owner)
...
+	{
 	  error = locator_flush_class (classmop);
+	  goto exit_on_error;
 	}
+
+      owner_name = au_get_user_name (owner);
+
+      snprintf (class_full_name, DB_MAX_IDENTIFIER_LENGTH, "%s.%s", owner_name, class_name);
+
+      error = sm_rename_class (classmop, class_full_name);
```
ctshim, line 5223: Shouldn't locator_flush_class (classmop); be called when there is no error after executing this statement?

src/object/authenticate.c
ctshim: I understand your explanation.
thank you.

src/object/authenticate.c
ctshim: I understand your explanation.
thank you.

src/object/class_object.c
```diff
@@ -6765,6 +6774,7 @@ classobj_class_size (SM_CLASS * class_)
 
   size = sizeof (SM_CLASS);
   size += strlen (sm_ch_name ((MOBJ) class_)) + 1;
+  size += strlen (sm_ch_simple_name ((MOBJ) class_)) + 1;
```
ctshim: Since ch_simple_name does not have any real data and only refers to a pointer, isn't it only necessary to increase the size of sizeof(char*)?

`size += sizeof(char*);`

src/parser/scanner_support.c
youngjinj: I changed it to enable the query below.
```
select /*+ select_key_info (dba.t1.pk_t1_c1) */ * from t1;
```
However, I think it would be better to just write the index name, so I delete the related changes.

src/loaddb/load_server_loader.cpp
youngjinj: I will change it to "DB_MAX_IDENTIFIER_LENGTH * 2".

src/object/trigger_manager.c
youngjinj: Okay, I'll change it.

src/object/class_object.c
youngjinj: Since the memory size is calculated, the size for ch_simple_name can be excluded.
The size of the ch_simple_name pointer is contained in "size = sizeof (SM_CLASS);".

Is it okay to exclude the size of ch_simple_name?

src/object/schema_manager.c
youngjinj: Okay, I'll change it.

src/object/schema_manager.c
youngjinj: Okay, I'll change it.

src/executables/unload_schema.c
youngjinj: I will change it to a macro.

src/object/trigger_description.cpp
youngjinj: I will use this method while changing to a macro.

src/executables/unload_schema.c
youngjinj: I will use this method while changing to a macro.

src/executables/unload_schema.c
youngjinj: I will use this method while changing to a macro.

src/executables/unload_schema.c
youngjinj: I will use this method while changing to a macro.

src/object/schema_manager.c
youngjinj: Thanks to you, I realized I was calling owner_name all NULL.

So I changed the sm_user_specified_name function.
1. It does not receive owner_name.
2. When name is not user_specified_name, current_user_name is used to make it user_specified_name.
3. The sm_user_specified_name function is called in sm_find_class.
4. Calls from other unnecessary locations have been removed.

src/object/class_object.c
ctshim: I think so.
What is the purpose of the size obtained?

src/executables/unloaddb.h
```diff
@@ -59,6 +59,21 @@ extern int lo_count;
...
+/* 
+ * name is user_specified_name.
+ * owner_name must be a char array of size DB_MAX_IDENTIFIER_LENGTH * 2 to copy user_specified_name.
+ * class_name refers to class_name after dot(.).
+ */
+#define SPLIT_USER_SPECIFIED_NAME(name, owner_name, class_name) \
+	do \
+	  { \
+	    memset (owner_name, 0, sizeof (owner_name)); \
+	    memcpy (owner_name, name, strlen (name) + 1); \
```
ctshim: If you must use strlen() anyway, use strcpy()
`strcpy(owner_name, name);`

youngjinj: Okay, I'll change it.

src/object/class_object.c
youngjinj: It is used in the calculation of "Total bytes for class storage" when printing workspace information in csql.

```
csql> ;info workspace
~
Total bytes for class storage     28668
~
```

src/object/trigger_manager.c
```diff
@@ -4225,6 +4232,24 @@ tr_find_trigger (const char *name)
 
   if (trigger_table_find (name, &object) == NO_ERROR)
     {
+      /* This is the case when the loaddb utility is executed with the --no-user-specified-name option as the dba user. */
+      if (db_get_client_type () == DB_CLIENT_TYPE_ADMIN_UTILITY && prm_get_bool_value (PRM_ID_NO_USER_SPECIFIED_NAME))
+	{
+	  if (object == NULL)
```
ctshim: How likely is object to be null in trigger_table_find()?
It would be better to change the order of checking if() statements.
```
      if (object == NULL)
	{
	  if (db_get_client_type () == DB_CLIENT_TYPE_ADMIN_UTILITY && prm_get_bool_value (PRM_ID_NO_USER_SPECIFIED_NAME))
```

src/object/trigger_manager.c
```diff
@@ -6866,6 +6900,12 @@ tr_rename_trigger (DB_OBJECT * trigger_object, const char *name, bool call_from_
 		  (void) trigger_table_rename (trigger_object, oldname);
 		  oldname = NULL;
 		}
+
+	      if (error == NO_ERROR)
+		{
+		  error = locator_flush_instance (trigger_object);
+		}
+
```
ctshim: It seems to have to be placed before line 6890 if (error != NO_ERROR).

src/parser/parser_support.c
```diff
@@ -9887,3 +9898,255 @@ pt_has_name_oid (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int *contin
...
+       */
+      return name;
+    }
+
+  current_user_name = db_get_user_name ();
+  assert (current_user_name != NULL);
+
+  dot[0] = '\0';
+
+  if (name && current_user_name && intl_identifier_casecmp (name, current_user_name) == 0)
```
ctshim: It does not come this far when name is null. The name check seems unnecessary.

src/object/trigger_manager.c
youngjinj: I'm changing the tr_rename_trigger function a bit.
I will consider it together.

src/parser/parser_support.c
youngjinj: Okay. I'll remove the unnecessary null check for name, current_user_name.

src/executables/unloaddb.h
```diff
@@ -59,6 +59,21 @@ extern int lo_count;
...
+ * owner_name must be a char array of size DB_MAX_IDENTIFIER_LENGTH * 2 to copy user_specified_name.
+ * class_name refers to class_name after dot(.).
+ */
+#define SPLIT_USER_SPECIFIED_NAME(name, owner_name, class_name) \
+	do \
+	  { \
+	    memset (owner_name, 0, sizeof (owner_name)); \
+	    strcpy (owner_name, name); \
+	    class_name = strchr (owner_name, '.'); \
+	    *class_name++ = '\0'; \
```
ctshim: memset() seems unnecessary.
*class_name++ = '\0'; is *(class_name)++ = '\0'; It would be better to enclose it in parentheses.

msg/km_KH.utf8/utils.msg
```diff
@@ -491,7 +491,9 @@ valid options:\n\
   -d, --data-file=FILE           load data with FILE\n\
   -t, --table=TABLE              table name that is substituted for missing class header in data file\n\
       --error-control-file=FILE  FILE to control error(s) during loading\n\
-      --ignore-class-file=FILE   input file of class names that skip load\n
+      --ignore-class-file=FILE   input file of class names that skip load\n\
+      --no-user-specified-name   Find class, serial, and trigger by object-name and don't use user-specified-name.\n\
+                                 For backward compatibility with the load file of the previous version of 11.2.\n
```
hyunikn: What about this?
"Find classes, serials, and triggers only by their object names without their user names.
 This is to load files generated by Cubrid before version 11.2."

src/parser/name_resolution.c
```diff
@@ -7152,6 +7167,82 @@ pt_str_compare (const char *p, const char *q, CASE_SENSITIVENESS case_flag)
...
+      /*
+       * e.g. original_p : object_name.          -> NULL
+       *      original_q : user_name.object_name -> object_name
+       *
+       *      or
+       * 
+       *      original_p : user_name.object_name -> object_name
+       *      original_q : object_name.          -> NULL
+       */
+      if (!original_p || !original_q)
```
ctshim: `if (original_p[0] == '\0' ||  original_q[0] == '\0')`

src/parser/parser_support.c
```diff
@@ -9887,3 +9898,255 @@ pt_has_name_oid (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int *contin
...
+
+  return node;
+}
+
+const char *
+pt_get_qualifier_name (PARSER_CONTEXT * parser, PT_NODE * name)
+{
+  char *dot = NULL;
+  const char *original_name = NULL;
+  const char *resolved_name = NULL;
```
ctshim: Even if the two variables original_name and resolved_name do not exist, it seems to be sufficient to handle them.

src/parser/query_result.c
```diff
@@ -659,9 +660,11 @@ pt_get_titles (PARSER_CONTEXT * parser, PT_NODE * query)
 	}
       else
 	{
+	  save_custom = parser->custom_print;
 	  parser->custom_print |= PT_SUPPRESS_CHARSET_PRINT;
+	  parser->custom_print |= PT_PRINT_NO_CURRENT_USER_NAME;
 	  t = pt_get_node_title (parser, s, f);
-	  parser->custom_print &= ~PT_SUPPRESS_CHARSET_PRINT;
+	  parser->custom_print = save_custom;
```
ctshim, line 667: The semantics are slightly different from the existing code.
If the purpose of the existing code is preserved, it should be modified as follows.

```
	  parser->custom_print |= (PT_SUPPRESS_CHARSET_PRINT | PT_PRINT_NO_CURRENT_USER_NAME);
	  t = pt_get_node_title (parser, s, f);
	  parser->custom_print &= ~(PT_SUPPRESS_CHARSET_PRINT | PT_PRINT_NO_CURRENT_USER_NAME);
```
In current code, "parser->custom_print &= ~PT_SUPPRESS_CHARSET_PRINT;" Strictly speaking, it is not to undo, but to remove PT_SUPPRESS_CHARSET_PRINT.

src/parser/query_result.c
```diff
@@ -732,17 +736,46 @@ pt_get_node_title (PARSER_CONTEXT * parser, const PT_NODE * col, const PT_NODE *
 	    {
 	      if (col->info.name.meta_class == PT_META_ATTR)
 		{
-		  name =
-		    pt_append_string (parser, pt_append_string (parser, (char *) col->info.name.resolved, "."), name);
-		  name = pt_append_string (parser, pt_append_string (parser, NULL, "class "), name);
+		  name = pt_append_string (parser, NULL, "class ");
+		  if (parser->custom_print & PT_PRINT_NO_CURRENT_USER_NAME)
```
ctshim: Is there any possibility that PT_PRINT_NO_CURRENT_USER_NAME and PT_PRINT_NO_SPECIFIED_USER_NAME are set at the same time?

src/query/execute_schema.c
```diff
@@ -2485,7 +2485,13 @@ acquire_locks_for_multiple_rename (const PT_NODE * statement)
...
       for (i = 0; i < num_names; i++)
 	{
-	  if (strcmp (name_set[i], old_name) == 0)
+	  /*
+	   * Check if old_name to be changed next is in the list of new_name that has been changed before.
+	   *   - e.g. rename table a as b, b as c, c as a;
+	   *     1. b vs b
+	   *     2. b vs c, c vs c
+	   */
+	  if (pt_str_compare (name_set[i], old_name, CASE_INSENSITIVE) == 0)
```
ctshim, line 2498: In the mq_check_delete() function, pt_str_compare() is replaced with pt_user_specified_name_compare().
Here we are replacing strcmp() with pt_str_compare(CASE_INSENSITIVE). Maybe I should use pt_user_specified_name_compare()?
Was strcmp() modified considering that it was case-sensitive?

src/object/class_object.c
ctshim: If it's a simple output of information, it doesn't seem that the calculated value is sensitive.

msg/en_US.utf8/cubrid.msg
```diff
@@ -1407,8 +1407,10 @@ Check the path of the key file (_keys) and make sure that it includes proper key
 1331 Error reserved for dblink development.
 
 1332 The active log volume (%1$s) is too sane to recreate. If you remove the active log, it could cause an unexpected and irreversible problem.
+1333 'auto_increment' serial does not allow change of owner.
+1334 RENAME does not allow change of owner.
 
-1333 Last Error
+1335 Last Error
```
hgryoo: Other cubrid.msg language files don't have these change

src/query/execute_schema.c
```diff
@@ -2612,6 +2618,15 @@ do_rename (const PARSER_CONTEXT * parser, const PT_NODE * statement)
       const char *old_name = current_rename->info.rename.old_name->info.name.original;
       const char *new_name = current_rename->info.rename.new_name->info.name.original;
 
+      const char *old_qualifier_name = pt_get_qualifier_name (parser, current_rename->info.rename.old_name);
+      const char *new_qualifier_name = pt_get_qualifier_name (parser, current_rename->info.rename.new_name);
+
+      if (old_qualifier_name && new_qualifier_name && intl_identifier_casecmp (old_qualifier_name, new_qualifier_name))
```
hgryoo: To make clear
```suggestion
      if (old_qualifier_name && new_qualifier_name && intl_identifier_casecmp (old_qualifier_name, new_qualifier_name) != 0)
```

src/object/trigger_manager.c
youngjinj: Okay, I'll change it. It seems good to check for NULL first.

src/executables/unloaddb.h
youngjinj: Okay. I enclose the argument in parentheses.

src/parser/name_resolution.c
youngjinj: Okay, I'll change it. I should have checked the value.

src/object/trigger_manager.c
youngjinj: Please review the changed tr_rename_trigger() function again.

src/parser/query_result.c
youngjinj: The code to remove PT_SUPPRESS_CHARSET_PRINT may have been intentional, but I don't think it matters.
The previous code is misleading, so I'll keep the changes.

src/parser/query_result.c
youngjinj: PT_PRINT_NO_SPECIFIED_USER_NAME implicitly includes PT_PRINT_NO_CURRENT_USER_NAME.

I will not deliberately set both, but since it is not impossible, I will set the priority by checking PT_PRINT_NO_SPECIFIED_USER_NAME before PT_PRINT_NO_CURRENT_USER_NAME.

msg/en_US.utf8/cubrid.msg
youngjinj: Okay. Changed.

src/query/execute_schema.c
youngjinj: Okay. Changed.

src/query/execute_schema.c
youngjinj: Changed comments to clarify the reason for the change.

```
/*
 * Check if old_name to be changed next is in the list of new_name that has been changed before.
 *   - e.g. rename table a as b, B as c, C as a;
 *     1. dba.b vs dba.B
 *     2. dba.b vs dba.C, dba.c vs dba.C
 *
 *     ERROR: Unknown class "b".
 *
 * Changed comparison function to pt_str_compare() function to be case insensitive.
 */
if (pt_str_compare (name_set[i], old_name, CASE_INSENSITIVE) == 0)
  {
    found = true;
    break;
  }
```

src/base/error_manager.h
```diff
@@ -85,6 +85,86 @@
     } \
   while (0)
 
+#define ERROR_SET_WARNING(error, code) \
+  do \
+    { \
+      error = code; \
+      er_set (ER_WARNING_SEVERITY, ARG_FILE_LINE, code, 0); \
+    } \
+  while (0)
```
ctshim: How about using a variable argument macro?

```
#define ERROR_SET_WARNING(error, code, arg_cnt, ...) \
  do \
    { \
      error = code; \
      er_set (ER_WARNING_SEVERITY, ARG_FILE_LINE, code, arg_cnt, ##__VA_ARGS__); \
    } \
  while (0)
```
  
or

```
#define ERROR_SET_WARNING(error, code, arg_cnt, args...) \
  do \
    { \
      error = code; \
      er_set (ER_WARNING_SEVERITY, ARG_FILE_LINE, code, arg_cnt, ##args); \
    } \
  while (0)  

```

src/query/serial.c
```diff
@@ -1486,8 +1491,8 @@ serial_cache_index_btid (THREAD_ENTRY * thread_p)
   oid_get_serial_oid (&serial_oid);
   assert (!OID_ISNULL (&serial_oid));
 
-  /* Now try to get index BTID. Serial index name is "pk_db_serial_name". */
-  error_code = heap_get_btid_from_index_name (thread_p, &serial_oid, "pk_db_serial_name", &serial_Cached_btid);
+  /* Now try to get index BTID. Serial index name is "pk_db_serial_name_owner". */
+  error_code = heap_get_btid_from_index_name (thread_p, &serial_oid, "u_db_serial_full_name", &serial_Cached_btid);
```
ctshim: According to the comment, shouldn't we use "pk_db_serial_name_owner"?

src/loaddb/load_sa_loader.cpp
```diff
@@ -1417,33 +1419,159 @@ ldr_assign_class_id (DB_OBJECT *class_, int id)
...
+  snprintf (query_buf, sizeof (query_buf), query, CT_CLASS_NAME, name_p, current_user_name);
+
+  error = db_compile_and_execute_local (query_buf, &query_result, &query_error);
+  if (error < NO_ERROR)
+    {
+      ASSERT_ERROR ();
+      goto end;
+    }
+
+  db_private_free_and_init (NULL, current_user_name);
```
ctshim, line 1523: Even if an error occurs, db_private_free_and_init() must be called.
```
  db_private_free_and_init (NULL, current_user_name);
  if (error < NO_ERROR)
    {
      ASSERT_ERROR ();
      goto end;
    }
```

ctshim: Even if an error occurs, db_private_free_and_init() must be called.
```
 db_private_free_and_init (NULL, current_user_name);

 error = db_compile_and_execute_local (query_buf, &query_result, &query_error);
  if (error < NO_ERROR)
    {
      ASSERT_ERROR ();
      goto end;
    }
```

src/loaddb/load_sa_loader.cpp
```diff
@@ -1417,33 +1419,159 @@ ldr_assign_class_id (DB_OBJECT *class_, int id)
...
 
-  return (class_);
+  return class_;
+}
+
+static int
+ldr_find_class_by_query (const char *name, char *buf, size_t buf_size)
+{
+#define QUERY_BUF_SIZE 2048
+#define CT_CLASS_NAME "_db_class"
```
ctshim: Is there any need to define it if it is to be used only within this function?
How about using const variables?

src/parser/parser_support.c
```diff
@@ -9887,3 +9898,258 @@ pt_has_name_oid (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int *contin
...
+	  ASSERT_ERROR_AND_SET (error);
+	  PT_ERROR (parser, node, er_msg ());
+	  *continue_walk = PT_STOP_WALK;
+	  return node;
+	}
+
+      resolved_name = current_user_name;
+    }
+
+  if (strlen (resolved_name) >= DB_MAX_USER_LENGTH)
```
ctshim: I don't think the user name obtained through db_get_user_name() can exceed the length limit.
So how about doing something like below?
`else if (strlen (resolved_name) >= DB_MAX_USER_LENGTH)`

src/parser/parser_support.c
```diff
@@ -9887,3 +9898,258 @@ pt_has_name_oid (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int *contin
...
+      *continue_walk = PT_STOP_WALK;
+      return node;
+    }
+
+  intl_identifier_lower (resolved_name, downcase_resolved_name);
+
+  if (strlen (original_name) >= DB_MAX_IDENTIFIER_LENGTH - DB_MAX_SCHEMA_LENGTH)
+    {
+      char *name = (char *) original_name;
+      strcpy (name + 60, "...");
```
ctshim: This is destroying the original data.
For simple error message output, it should be copied to a separate variable and used.

src/base/error_manager.h
youngjinj: Not bad, but isn't it the same as calling the er_set function directly?

src/query/serial.c
youngjinj: I'm missing a comment change. I'll remove it from the comments.

src/loaddb/load_sa_loader.cpp
youngjinj: Okay, I'll change it.

src/object/trigger_manager.c
ctshim: (void) trigger_table_rename (trigger_object, oldname); should be called even if locator_flush_instance() failed.

src/loaddb/load_sa_loader.cpp
youngjinj: Remove #define CT_CLASS_NAME "_db_class" and add #include "transform.h" .

src/parser/parser_support.c
youngjinj: Like the code below?
```
if (resolved_name == NULL || resolved_name[0] == '\0')
  {
    current_user_name = db_get_user_name ();
    if (!current_user_name)
      {
        ASSERT_ERROR_AND_SET (error);
        PT_ERROR (parser, node, er_msg ());
        return node;
      }

    resolved_name = current_user_name;
  }
else if (strlen (resolved_name) >= DB_MAX_USER_LENGTH)
  {
    PT_ERRORf2 (parser, node,
                "User name [%s] not allowed. It cannot exceed %d bytes.", resolved_name, DB_MAX_USER_LENGTH);
    return node;
  }
```

src/parser/parser_support.c
youngjinj: Okay, I'll change it like below.
```
PT_ERRORf2 (parser, node,
            "Identifier name [%s] not allowed. It cannot exceed %d bytes.",
            pt_short_print (parser, node), DB_MAX_IDENTIFIER_LENGTH - DB_MAX_USER_LENGTH);
```

beyondykk9 (approved):

shparkcubrid:
Please add the test case for this issue in JIRA.

src/parser/parser_support.c
ctshim: yes!

hyunikn (approved):

ctshim (approved):

src/executables/csql.c
```diff
@@ -1076,7 +1086,17 @@ csql_do_session_cmd (char *line_read, CSQL_ARGUMENT * csql_arg)
...
     case S_CMD_TRIGGER:
-      csql_help_trigger ((argument[0] == '\0') ? NULL : argument);
+      if (argument[0] == '\0')
+	{
+	  csql_help_trigger (NULL);
+	}
+      else
+	{
+	  char realname[SM_MAX_IDENTIFIER_LENGTH] = { '\0' };
+	  sm_user_specified_name (argument, realname, SM_MAX_IDENTIFIER_LENGTH);
```
shparkcubrid, line 1096: Isn't it necessary for the S_CMD_INFO case as well?
```
info stat table
```

src/loaddb/load_sa_loader.cpp
```diff
@@ -1417,33 +1420,157 @@ ldr_assign_class_id (DB_OBJECT *class_, int id)
...
+
+  if (!DB_IS_NULL (&value))
+    {
+      memset (buf, 0, buf_size);
+      strcpy(buf, db_get_string (&value));
+    }
+  else
+    {
+      /* class_full_name must not be null. */
+      assert (false);
```
shparkcubrid:
```suggestion
ASSERT_ERROR_AND_SET (error);
goto end;
```

src/base/error_manager.h
youngjinj: That's a way to reduce multiple macro functions into one, but I'll keep it this way for now.

src/executables/csql.c
youngjinj: Below, the stats_dump() function is called and the class-name is found in the sm_find_class() function.
The sm_find_class() function calls the sm_user_specified_name() function to find the class-name.
```
;info stat class-name
```

src/loaddb/load_sa_loader.cpp
youngjinj: Okay. Changed.

msg/km_KH.utf8/utils.msg
youngjinj: Okay, I changed it.
I used it by changing "user name" to "owner name".

shparkcubrid (approved):

hgryoo (approved):

src/parser/query_result.c
```diff
@@ -836,14 +836,31 @@ pt_get_node_title (PARSER_CONTEXT * parser, const PT_NODE * col, const PT_NODE *
 			  if (original_name)
 			    {
 			      /* strip off classname.* */
-			      name += strlen (node->info.name.original) + 1;
+			      if (strchr (node->info.name.original, '.'))
+				{
+				  name = strchr (original_name, '.');
```
shparkcubrid, line 841: Are node->info.name.original and original_name the same value?

youngjinj: node->info.name.original and original name do not have to be the same.

This is to remove [owner_name.class_name] when classname.* is [owner_name.class_name].*.

[owner_name.class_name] is stored in node->info.dot.arg1 in case of PT_DOT_ node.

If col->info.dot.arg1->info.name.original (node->info.name.original) contains dot(.), point to after the second dot(.).

src/object/authenticate.h
```diff
@@ -138,6 +138,21 @@ MOP au_get_dba_user (void);
...
 
+/* 
+ * name is user_specified_name.
+ * owner_name must be a char array of size DB_MAX_IDENTIFIER_LENGTH to copy user_specified_name.
+ * class_name refers to class_name after dot(.).
+ */
+#define SPLIT_USER_SPECIFIED_NAME(name, owner_name, class_name) \
+	do \
+	  { \
+	    memset ((owner_name), '\0', DB_MAX_IDENTIFIER_LENGTH); \
```
ctshim: memset() is unnecessary code.

src/query/execute_schema.c
```diff
@@ -2625,7 +2625,7 @@ do_rename (PARSER_CONTEXT * parser, PT_NODE * statement)
       const char *old_qualifier_name = pt_get_qualifier_name (parser, current_rename->info.rename.old_name);
       const char *new_qualifier_name = pt_get_qualifier_name (parser, current_rename->info.rename.new_name);
 
-      if (pt_str_compare (old_qualifier_name, new_qualifier_name, CASE_INSENSITIVE) != 0)
+      if (old_qualifier_name && new_qualifier_name && pt_str_compare (old_qualifier_name, new_qualifier_name, CASE_INSENSITIVE) != 0)
```
ctshim: How about modifying it like below?
`if (old_qualifier_name && new_qualifier_name && intl_identifier_casecmp(old_qualifier_name, new_qualifier_name) != 0)`

But if any of old_qualifier_name or new_qualifier_name are null, is it correct to judge that it is normal without having to compare? Isn't that an error?

src/object/authenticate.h
youngjinj: Sometimes it is called in a loop. In this case, memset () is required.

src/query/execute_schema.c
youngjinj: If it is a system class, new_qualifier_name is null because unique_name does not contain an owner.
So the error occurs because the owner is different.

The error is not appropriate, and it has been changed to match the existing test case answer.

I would change to intl_identifier_casecmp () instead of pt_str_compare () function.

src/executables/unloaddb.h
```diff
@@ -59,6 +59,21 @@ extern int lo_count;
...
 
+/* 
+ * name is user_specified_name.
+ * owner_name must be a char array of size DB_MAX_IDENTIFIER_LENGTH to copy user_specified_name.
+ * class_name refers to class_name after dot(.).
+ */
+#define SPLIT_USER_SPECIFIED_NAME(name, owner_name, class_name) \
+	do \
+	  { \
+	    memset ((owner_name), '\0', DB_MAX_IDENTIFIER_LENGTH); \
```
ctshim: memset() is unnecessary code.

src/object/authenticate.h
youngjinj: It is checked with unnecessary memset() and will be removed.

src/executables/unloaddb.h
youngjinj: It is checked with unnecessary memset() and will be removed.

src/loaddb/load_sa_loader.cpp
```diff
@@ -1417,33 +1420,152 @@ ldr_assign_class_id (DB_OBJECT *class_, int id)
...
+  char query_buf[QUERY_BUF_SIZE] = { '\0' };
+  char current_user_name[DB_MAX_USER_LENGTH] = { '\0' };
+  const char *class_name = NULL;
+  int error = NO_ERROR;
+
+  db_make_null (&value);
+  query_error.err_lineno = 0;
+  query_error.err_posno = 0;
+
+  if (name == NULL || name[0] == '\0' || buf == NULL || buf_size < 0)
```
ctshim: If buf_size is 0 or 5, it seems to be a problem. Does it make sense to check here?
It seems to be larger than the minimum size to contain unique_name.

youngjinj: buf_size is only used by assert(), so checking "buf_size < 0" is unnecessary.
Getting the buf_size in itself seems unnecessary.
I will improve.

ctshim (approved):

src/executables/unload_schema.c
```diff
@@ -1178,6 +1191,35 @@ emit_schema (print_output & output_ctx, DB_OBJLIST * classes, int do_auth, DB_OB
...
+       *      create table t1;
+       *      call change_owner ('t1', 'u1') on class db_root;
+       *      alter table t1 add attribute c1 int auto_increment;
+       *      select c.clasS_name, c.owner.name, s.name, s.owner.name
+       *      from _db_class c, db_serial s
+       *      where c.class_name = s.class_name;
+       *
+       *        clasS_name            owner.name            name                  owner.name
+       *      ========================================================================================
+       *        't2'                  'U1'                  't2_ai_c1'            'DBA'
```
hyunikn: Prabably, it is not t2 but t1.

src/executables/unload_schema.c
```diff
@@ -1378,8 +1419,9 @@ emit_query_specs (print_output & output_ctx, DB_OBJLIST * classes)
 	      if (query_ptr != NULL)
 		{
 		  null_spec = pt_print_query_spec_no_list (parser, *query_ptr);
-
-		  output_ctx ("ALTER VCLASS %s%s%s ADD QUERY %s ; \n", PRINT_IDENTIFIER (name), null_spec);
+		  SPLIT_USER_SPECIFIED_NAME (name, owner_name, class_name);
```
hyunikn, line 1419: Is the 'name' a unique_name?

hyunikn (approved):

src/executables/unload_schema.c
youngjinj: Okay, it's a typo.

src/executables/unload_schema.c
youngjinj: The name of SM_CLASS_HEADER is unique_name .

Get the name with the code below.
```
name = db_get_class_name (cl->op);
```

The db_get_class_name() function gets the name of SM_CLASS_HEADER in the calling order below.
```
db_get_class_name ()
-> sm_get_ch_name ()
-> sm_ch_name ()
```

hgryoo (approved):

beyondykk9 (approved):

PR #3350, opened 2022-01-27
[LETS-279] Set page desynchronization error when fixing leaf page for btree range scan

adrianbrab:
http://jira.cubrid.org/browse/LETS-279

On PTS, index range scans may move from a leaf page to the next/prev. The next page may be fetched from PS and may be desynchronized if that is the case the page ahead of replication error must be set.

In  the following cases the `pgbuf_fix` was changed to also check for desync: 
 -  `next_node_page` in `btree_range_scan_advance_over_filtered_keys`
 -  `prev_leaf` in `btree_range_scan_descending_fix_prev_leaf`
 -  `bts->C_page` when `bts->force_restart_from_root` is false

### DIFF
--- a/src/storage/btree.c
+++ b/src/storage/btree.c
@@ -24860,7 +24860,7 @@ btree_range_scan_descending_fix_prev_leaf (THREAD_ENTRY * thread_p, BTREE_SCAN *
     }
   if (prev_leaf == NULL)
     {
-      /* deallocated */
+      /* deallocated or found to be ahead of replication, in either case the scan needs to restart */
       bts->force_restart_from_root = true;
       return NO_ERROR;
     }
--- a/src/storage/btree.c
+++ b/src/storage/btree.c
@@ -24851,7 +24851,8 @@ btree_range_scan_descending_fix_prev_leaf (THREAD_ENTRY * thread_p, BTREE_SCAN *
   /* Unfix current page and retry. */
   pgbuf_unfix_and_init (thread_p, bts->C_page);
   error_code =
-    pgbuf_fix_if_not_deallocated (thread_p, &prev_leaf_vpid, PGBUF_LATCH_READ, PGBUF_UNCONDITIONAL_LATCH, &prev_leaf);
+    pgbuf_fix_if_not_deallocated_with_repl_desync_check (thread_p, &prev_leaf_vpid, PGBUF_LATCH_READ,
+							 PGBUF_UNCONDITIONAL_LATCH, &prev_leaf);
   if (error_code != NO_ERROR)
     {
       ASSERT_ERROR ();
@@ -24891,7 +24892,7 @@ btree_range_scan_descending_fix_prev_leaf (THREAD_ENTRY * thread_p, BTREE_SCAN *
       if (er_errid () == ER_PAGE_AHEAD_OF_REPLICATION)
 	{
 	  bts->force_restart_from_root = true;
-	  pgbuf_unfix_and_init (thread_p, bts->C_page);
+	  pgbuf_unfix_and_init (thread_p, prev_leaf);
 	  return NO_ERROR;
 	}
 
--- a/src/storage/btree.c
+++ b/src/storage/btree.c
@@ -24711,7 +24711,8 @@ btree_range_scan_advance_over_filtered_keys (THREAD_ENTRY * thread_p, BTREE_SCAN
 	  else
 	    {
 	      /* Fix next leaf page. */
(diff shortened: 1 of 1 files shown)


src/storage/btree.c
```diff
@@ -24877,7 +24878,16 @@ btree_range_scan_descending_fix_prev_leaf (THREAD_ENTRY * thread_p, BTREE_SCAN *
   /* Pages are still linked. */
 
   /* Fix current page too. */
-  bts->C_page = pgbuf_fix (thread_p, &bts->C_vpid, OLD_PAGE, PGBUF_LATCH_READ, PGBUF_UNCONDITIONAL_LATCH);
+  if (bts->force_restart_from_root)
+    {
+      bts->C_page = pgbuf_fix (thread_p, &bts->C_vpid, OLD_PAGE, PGBUF_LATCH_READ, PGBUF_UNCONDITIONAL_LATCH);
```
razvanradulescu: Why we don't always use pgbuf_fix_old_and_check_repl_desync ?
The flag force_restart_from_root can be set for various reasons (btree changed by other transaction, page ahead  due to replication).
By always using pgbuf_fix_old_and_check_repl_desync, we can catch the "page ahead error", clear it  and make sure to retry, otherwise the scan is stopped with error.

cristiarg: My understanding in this situation is that the page is "most probably" still in the page buffer. Thus, it will be in-sync with replication when re-fixed again after having been unfixed at the beginning of the function.
To be perfectly safe, I would still `pgbuf_fix_old_and_check_repl_desync` and, if `C_page` remains null, assert `assert (er_errid () != ER_PAGE_AHEAD_OF_REPLICATION);`

src/storage/btree.c
```diff
@@ -24711,7 +24711,8 @@ btree_range_scan_advance_over_filtered_keys (THREAD_ENTRY * thread_p, BTREE_SCAN
 	  else
 	    {
 	      /* Fix next leaf page. */
-	      next_node_page = pgbuf_fix (thread_p, &next_vpid, OLD_PAGE, PGBUF_LATCH_READ, PGBUF_UNCONDITIONAL_LATCH);
+	      next_node_page =
+		pgbuf_fix_old_and_check_repl_desync (thread_p, next_vpid, PGBUF_LATCH_READ, PGBUF_UNCONDITIONAL_LATCH);
```
cristiarg, line 24715: How is `ER_PAGE_AHEAD_OF_REPLICATION` error handled here?

src/storage/btree.c
```diff
@@ -24826,7 +24827,7 @@ btree_range_scan_descending_fix_prev_leaf (THREAD_ENTRY * thread_p, BTREE_SCAN *
   VPID_COPY (&prev_leaf_vpid, next_vpid);
 
   /* Conditional latch for previous page. */
-  prev_leaf = pgbuf_fix (thread_p, &prev_leaf_vpid, OLD_PAGE, PGBUF_LATCH_READ, PGBUF_CONDITIONAL_LATCH);
+  prev_leaf = pgbuf_fix_old_and_check_repl_desync (thread_p, prev_leaf_vpid, PGBUF_LATCH_READ, PGBUF_CONDITIONAL_LATCH);
```
cristiarg, line 24830: below, after
```
  /* Conditional latch failed. */
```
add a comment explaining that the page can be null also because of page desync on PTS and that the behavior is the same as in other previouslty existing situations.

src/storage/btree.c
```diff
@@ -24840,11 +24841,18 @@ btree_range_scan_descending_fix_prev_leaf (THREAD_ENTRY * thread_p, BTREE_SCAN *
...
+      bts->force_restart_from_root = true;
+      pgbuf_unfix_and_init (thread_p, bts->C_page);
+      return NO_ERROR;
+    }
   /* Unfix current page and retry. */
   pgbuf_unfix_and_init (thread_p, bts->C_page);
   error_code =
-    pgbuf_fix_if_not_deallocated (thread_p, &prev_leaf_vpid, PGBUF_LATCH_READ, PGBUF_UNCONDITIONAL_LATCH, &prev_leaf);
+    pgbuf_fix_if_not_deallocated_with_repl_desync_check (thread_p, &prev_leaf_vpid, PGBUF_LATCH_READ,
+							 PGBUF_UNCONDITIONAL_LATCH, &prev_leaf);
```
cristiarg, line 24855: Add a comment in the scope
```
  if (prev_leaf == NULL)
    {
      /* deallocated */
      bts->force_restart_from_root = true;
      return NO_ERROR;
    }
```
below that the restart can also happen due to desynchronization

cristiarg (approved):

PR #3348, opened 2022-01-26
[CBRD-24173] Generate summary information for flashback

joohok:
http://jira.cubrid.org/browse/CBRD-24173

Purpose
1. flashback utility request summary information (flashback_get_summary())
2. server generate the summary information and return (sflashback_get_summary())
3. flashback utility will show generated summary information to the client (in other PR)

Implementation

1. flashback_get_summary () request to sflashback_get_summary ()
2. sflashback_get_summary () calls flashback_make_summary_list () to generate summary information
3. Traversing log records backward and make summary entry when meets LOG_COMMIT 

Remark
N/A

PR #3353, opened 2022-01-28
[CBRD-24175] Generate log items for flashback

joohok:
http://jira.cubrid.org/browse/CBRD-24175

Purpose

Flashback utility request flashback using specific transaction ID, start lsa, end lsa and direction.
Server generates log items using above information.


src/communication/network_interface_sr.c
```diff
@@ -10849,6 +10849,32 @@ sflashback_get_summary (THREAD_ENTRY * thread_p, unsigned int rid, char *request
...
+static char *
+packing_loginfos (THREAD_ENTRY * thread_p, char *ptr, FLASHBACK_LOGINFO_CONTEXT context)
+{
+  CDC_LOGINFO_ENTRY *entry;
+
+  for (int i = 0; i < context.num_item; i++)
+    {
+      ptr = PTR_ALIGN (ptr, MAX_ALIGNMENT);
+      // *INDENT-OFF*
+      entry = context.loginfo_queue.front ();
```
hornetmj, line 10926: FYI.

If you use deque instead of queue container, then you can simplify the log read mechanism (with direction). I mean (for example ... :D)

if (direction == forward)
{
    deque.pop_front();
}
else
{
    deque.pop_back();
}

hornetmj: And I think you just use pop() instead of front() if there is no reason why you need to use front().

src/transaction/log_manager.h
```diff
@@ -257,6 +257,9 @@ extern void cdc_kill_producer ();
 
 extern void cdc_pause_consumer ();
 extern void cdc_wakeup_consumer ();
+
+/* flashback */
+extern int flashback_make_loginfo (THREAD_ENTRY * thread_p, FLASHBACK_LOGINFO_CONTEXT * context);
```
hornetmj, line 262: It would be nice to declare and define in the flashback.c/h.

src/transaction/log_manager.c
```diff
@@ -14684,6 +14733,457 @@ cdc_is_filtered_user (char *user)
...
+  LSA_COPY (&process_lsa, &context->end_lsa);
+
+  if (logpb_fetch_page (thread_p, &process_lsa, LOG_CS_SAFE_READER, log_page_p) != NO_ERROR)
+    {
+      /* can not find the log page */
+      error = ER_FAILED;
+      goto exit;
+    }
+
+begin:
```
hornetmj: I think a loop is more readable than goto.

src/transaction/log_manager.c
```diff
@@ -14684,6 +14733,457 @@ cdc_is_filtered_user (char *user)
...
+  else
+    {
+      LSA_COPY (&process_lsa, &log_rec_header->prev_tranlsa);
+    }
+
+  if (log_page_p->hdr.logical_pageid != process_lsa.pageid)
+    {
+      if (logpb_is_page_in_archive (process_lsa.pageid))
+	{
+	  int arv_num = 0;
```
hornetmj: If you don't have a plan to use this local variable, then you can remove it. (you can call logpb_fetch_from_archive without arv_num.)

src/transaction/log_manager.c
```diff
@@ -14684,6 +14733,457 @@ cdc_is_filtered_user (char *user)
...
+  RECDES supp_recdes = RECDES_INITIALIZER;
+  RECDES undo_recdes = RECDES_INITIALIZER;
+  RECDES redo_recdes = RECDES_INITIALIZER;
+
+  int num_loginfo = 0;
+  CDC_LOGINFO_ENTRY *log_info_entry = NULL;
+
+  if (LSA_ISNULL (&context->start_lsa))
+    {
+      if ((error = flashback_find_start_lsa (thread_p, context)) != NO_ERROR)
```
hornetmj: This makes additional log page reads. So, I wonder do you think about finding start lsa and making log info together?

src/transaction/log_manager.c
```diff
@@ -14684,6 +14733,457 @@ cdc_is_filtered_user (char *user)
...
+  LSA_COPY (&cur_log_rec_lsa, &process_lsa);
+
+  log_rec_header = LOG_GET_LOG_RECORD_HEADER (log_page_p, &process_lsa);
+
+  log_type = log_rec_header->type;
+  trid = log_rec_header->trid;
+
+  if (context->forward)
+    {
+      if (LSA_ISNULL (&log_rec_header->forw_lsa) && log_type != LOG_END_OF_LOG)
```
hornetmj: When the direction is forward, the end_lsa is bounded. So I think that this line doesn't need. Is there any reason for this?

src/transaction/log_manager.c
```diff
@@ -14684,6 +14733,457 @@ cdc_is_filtered_user (char *user)
...
+
+      LOG_READ_ADD_ALIGN (thread_p, sizeof (*supplement), &process_lsa, log_page_p);
+
+      if (cdc_get_undo_record (thread_p, log_page_p, cur_log_rec_lsa, &supp_recdes) != S_SUCCESS)
+	{
+	  error = ER_FAILED;
+	  goto exit;
+	}
+
+      supplement_length = sizeof (supp_recdes.type) + supp_recdes.length;
```
hornetmj: These codes around this line are very similar to the codes in the flashback_make_summary_list function (case LOG_SUPPLEMENTAL_INFO:). So, make a function for them.

src/communication/network_interface_sr.c
joohok: 1. deque will be nice to use for log read mechanism. 
but we don't know exact range of LSA to read log records.

I mean, 
We create log items by searching from `start_lsa` to `end_lsa`, but it is difficult to do in the way as above because we only create as many as the number input by the user. 

if user wants 5 log items per one connection, it traverse start_lsa to next start_lsa to make 5 log items.  
start_lsa ----------|next start lsa -------------------end_lsa

if only 5 log items exists within the range, then deque will be nice to use. 

2. pop() function does not returns the element, just remove the element. So, I used front() first then pop().

src/transaction/log_manager.h
joohok: yes, I will move it to flashback.c/h when previous PR is merged (generating flashback summary info (#3348))

src/transaction/log_manager.c
joohok: yes, I will change it

src/transaction/log_manager.c
joohok: oh, Thank you. I will remove it

src/transaction/log_manager.c
joohok: yes, that can be done only for when direction is forward, and just for the first time user request. 

this is processed only for rare case that transaction is started out of the time range (transaction start time > flashback user input start time).

src/transaction/log_manager.c
joohok: Yes, I think it doesn't need. 
It is for the log record that is not flushed to the page buffer  yet. thank you.

src/transaction/log_manager.c
joohok: Yes, it will be better

eido5 (approved):
there are no typos.

src/communication/network_interface_cl.c
```diff
@@ -10649,6 +10649,118 @@ flashback_get_summary (dynamic_array * class_list, const char *user, time_t star
...
+      ptr = or_unpack_int (ptr, &num_change_col);
+
+      for (int j = 0; j < num_change_col; j++)
+	{
+	  ptr = or_unpack_int (ptr, &change_index);
+	}
+
+      for (int j = 0; j < num_change_col; j++)
+	{
+	  ptr = or_unpack_int (ptr, &func_code);
```
wrlawodms, line 10770: What does `func_code` mean and why don't you define an ENUM?

src/transaction/log_manager.c
```diff
@@ -14684,6 +14733,457 @@ cdc_is_filtered_user (char *user)
...
+      memcpy (supplement_data + sizeof (supp_recdes.type), supp_recdes.data, supp_recdes.length);
+
+      free_and_init (supp_recdes.data);
+
+      switch (rec_type)
+	{
+	case LOG_SUPPLEMENT_TRAN_USER:
+	  strncpy (tran_user, supplement_data, supplement_length);
+
+	  break;
```
wrlawodms: `context` already knows the user name. Can't we use it?

src/transaction/log_manager.c
```diff
@@ -14684,6 +14733,457 @@ cdc_is_filtered_user (char *user)
...
+      LOG_READ_ADD_ALIGN (thread_p, sizeof (*supplement), &process_lsa, log_page_p);
+
+      if (cdc_get_undo_record (thread_p, log_page_p, cur_log_rec_lsa, &supp_recdes) != S_SUCCESS)
+	{
+	  error = ER_FAILED;
+	  goto exit;
+	}
+
+      supplement_length = sizeof (supp_recdes.type) + supp_recdes.length;
+      supplement_data = (char *) malloc (supplement_length);
```
wrlawodms: Some use `db_private_alloc()`, and some use `malloc()`. What's different?

src/transaction/log_manager.c
```diff
@@ -14684,6 +14733,457 @@ cdc_is_filtered_user (char *user)
...
+
+      supplement = (LOG_REC_SUPPLEMENT *) (log_page_p->area + process_lsa.offset);
+      supplement_length = supplement->length;
+      rec_type = supplement->rec_type;
+
+      LOG_READ_ADD_ALIGN (thread_p, sizeof (*supplement), &process_lsa, log_page_p);
+
+      if (cdc_get_undo_record (thread_p, log_page_p, cur_log_rec_lsa, &supp_recdes) != S_SUCCESS)
+	{
+	  error = ER_FAILED;
```
wrlawodms: No need to `er_set()?`

src/transaction/log_manager.c
```diff
@@ -14684,6 +14733,457 @@ cdc_is_filtered_user (char *user)
...
+	  if (cdc_get_recdes (thread_p, NULL, NULL, &redo_lsa, &redo_recdes, true) != NO_ERROR)
+	    {
+	      goto exit;
+	    }
+
+	  log_info_entry = (CDC_LOGINFO_ENTRY *) db_private_alloc (thread_p, sizeof (CDC_LOGINFO_ENTRY));
+	  if (log_info_entry == NULL)
+	    {
+	      error = ER_OUT_OF_VIRTUAL_MEMORY;
+	      goto exit;
```
wrlawodms: No need to er_set()?

src/transaction/log_manager.c
```diff
@@ -14684,6 +14733,457 @@ cdc_is_filtered_user (char *user)
...
+	  if (!flashback_class_filter (context, classoid))
+	    {
+	      goto end;
+	    }
+
+	  memcpy (&redo_lsa, supplement_data + sizeof (OID), sizeof (LOG_LSA));
+
+	  if (cdc_get_recdes (thread_p, NULL, NULL, &redo_lsa, &redo_recdes, true) != NO_ERROR)
+	    {
+	      goto exit;
```
wrlawodms: error not set

src/transaction/log_manager.c
```diff
@@ -14684,6 +14733,457 @@ cdc_is_filtered_user (char *user)
...
+              context->loginfo_queue.push (log_info_entry);
+              // *INDENT-ON*
+	      context->queue_size += log_info_entry->length;
+	      num_loginfo++;
+	      log_info_entry = NULL;
+	    }
+	  else
+	    {
+	      goto exit;
+	    }
```
wrlawodms: Like others do
```suggestion
	  if (error != NO_ERROR)
	    {
	      goto exit;
	    }
	    // *INDENT-OFF*
              context->loginfo_queue.push (log_info_entry);
              // *INDENT-ON*
	      context->queue_size += log_info_entry->length;
	      num_loginfo++;
	      log_info_entry = NULL;
```

src/transaction/log_manager.c
```diff
@@ -14684,6 +14733,457 @@ cdc_is_filtered_user (char *user)
...
+	  if ((error = cdc_get_recdes (thread_p, &undo_lsa, &undo_recdes, &redo_lsa, &redo_recdes, true)) != NO_ERROR)
+	    {
+	      goto exit;
+	    }
+
+	  log_info_entry = (CDC_LOGINFO_ENTRY *) db_private_alloc (thread_p, sizeof (CDC_LOGINFO_ENTRY));
+	  if (log_info_entry == NULL)
+	    {
+	      error = ER_OUT_OF_VIRTUAL_MEMORY;
+	      goto exit;
```
wrlawodms: er_set?

src/transaction/log_manager.c
```diff
@@ -14684,6 +14733,457 @@ cdc_is_filtered_user (char *user)
...
+              context->loginfo_queue.push (log_info_entry);
+              // *INDENT-ON*
+	      context->queue_size += log_info_entry->length;
+	      num_loginfo++;
+	      log_info_entry = NULL;
+	    }
+	  else
+	    {
+	      goto exit;
+	    }
```
wrlawodms: like others do. Just suggestion.
```suggestion

	  if (error != NO_ERROR)
	    {
	      goto exit;
	    }
	   // *INDENT-OFF*
              context->loginfo_queue.push (log_info_entry);
              // *INDENT-ON*
	      context->queue_size += log_info_entry->length;
	      num_loginfo++;
	      log_info_entry = NULL;
```

src/transaction/log_manager.c
```diff
@@ -14684,6 +14733,457 @@ cdc_is_filtered_user (char *user)
...
+
+end:
+  if (context->forward)
+    {
+      if (LSA_EQ (&next_log_rec_lsa, &context->end_lsa) || num_loginfo == context->num_item)
+	{
+	  context->num_item = num_loginfo;
+	  LSA_COPY (&context->start_lsa, &next_log_rec_lsa);
+
+	  error = NO_ERROR;
```
wrlawodms: Why does it have to be overwritten? and `error` can be not NO_ERROR here?

src/transaction/log_manager.c
```diff
@@ -14684,6 +14733,457 @@ cdc_is_filtered_user (char *user)
...
+      if (logpb_fetch_page (thread_p, &process_lsa, LOG_CS_SAFE_READER, log_page_p) != NO_ERROR)
+	{
+	  error = ER_FAILED;
+	  goto exit;
+	}
+
+      if (process_lsa.offset == NULL_OFFSET && (process_lsa.offset = log_page_p->hdr.offset) == NULL_OFFSET)
+	{
+	  /* can not find the first log record in the page */
+	  error = ER_FAILED;
```
wrlawodms: er_set?

src/transaction/log_manager.c
```diff
@@ -14684,6 +14733,457 @@ cdc_is_filtered_user (char *user)
...
+  log_page_p = (LOG_PAGE *) PTR_ALIGN (log_pgbuf, MAX_ALIGNMENT);
+
+  /* fetch log page */
+  if (logpb_fetch_page (thread_p, &process_lsa, LOG_CS_SAFE_READER, log_page_p) != NO_ERROR)
+    {
+      error = ER_FAILED;
+      goto exit;
+    }
+
+begin:
```
wrlawodms: Are you going to use `while` instead of `goto` like https://github.com/CUBRID/cubrid/pull/3348?

src/communication/network_interface_cl.c
```diff
@@ -10649,6 +10649,118 @@ flashback_get_summary (dynamic_array * class_list, const char *user, time_t star
   return ER_NOT_IN_STANDALONE;
 }
 
+static char *
+unpacking_loginfo (char *ptr, int num_item)
```
wrlawodms, line 10728: For consistency with `packing_loginfos`
```suggestion
unpacking_loginfos (char *ptr, int num_item)
```

src/communication/network_interface_sr.c
```diff
@@ -10849,6 +10849,32 @@ sflashback_get_summary (THREAD_ENTRY * thread_p, unsigned int rid, char *request
   return;
 }
 
+static char *
+packing_loginfos (THREAD_ENTRY * thread_p, char *ptr, FLASHBACK_LOGINFO_CONTEXT context)
```
wrlawodms: It seems to need a prefix.

src/transaction/log_manager.c
joohok: Yes, I will avoid using label.

src/communication/network_interface_cl.c
joohok: I will fix it to clarify what func_code means and set ENUM. 
func_code is just what or_pack_* function is used when packing data.

src/transaction/log_manager.c
joohok: Oh, yes you are right.

src/transaction/log_manager.c
joohok: Oh... No, I will fix it like previous PR (generating summary info #3348)

src/transaction/log_manager.c
joohok: T.T .. I will find all the error, and do er_set. Thanks

src/transaction/log_manager.c
joohok: OK, It seems better

src/transaction/log_manager.c
joohok: OK, thank you!

src/transaction/log_manager.c
joohok: I think there is no need to overwrite. 
Thanks, there is no case that `error` can be not NO_ERROR.

src/communication/network_interface_sr.c
hornetmj: 1. I don't understand the reason you explain for me. I think it is not related to the log info count per request. Please leave more detail for me or we can discuss this online first.
2. OK.

src/transaction/log_manager.h
hornetmj: OK.

src/transaction/log_manager.c
hornetmj: OK.

src/transaction/log_manager.c
hornetmj: OK.

src/transaction/log_manager.c
hornetmj: No. My comment means that how about deferring to find the start_lsa. Finding the start_lsa is always executed if the transaction didn't find the start_lsa in the summary step. You said this is a rare case. But I don't agree with it because the time range inputted to flashback utility by a user is an approximate time.

I think there is one case too hard to adapt my opinion(have to find the start_lsa in advance as you did). If the start_lsa is null and the direction is forward, then we need to search backward to find the start_lsa. On the contrary, we need to search forward to make log infos. But, I think that a user normally wants to use the backward direction to flashback. So, the previous situation is rare I think.

Getting back to the point, let's first focus on getting the flashback utility to work properly.

src/transaction/log_manager.c
hornetmj: We already find the end_lsa in the summary step. And I think that the forw_lsa is never null until the end_lsa. Are there any cases I didn't know(catch)?

src/transaction/log_manager.c
hornetmj: OK.

src/communication/network_interface_cl.c
```diff
@@ -10649,6 +10649,134 @@ flashback_get_summary (dynamic_array * class_list, const char *user, time_t star
   return ER_NOT_IN_STANDALONE;
 }
 
+/* unpacking_loginfo : test purpose, will be replaced */
```
joohok, line 10726: This is test code.
Here will be dealt in other issue (#3361)

src/communication/network_interface_sr.c
joohok: 1. I think it is better to discussing it online. 

- if loginfos are generated backward, and 10 loginfo are in one transaction. 
- user wants 5 loginfos per connection
- Loginfos are : | Log info 1 ~ Log info 5 | Log info 6 ~ Log info 10 |
  - Loginfo # is generated in reverse time order
  - if direction is backward, then it return at first | Loginfo 1 ~ Loginfo 5 | then  | Loginfo6 ~ Loginfo 10 |, using deque.pop_front
  - if direction is forward, then it returns at first | Loginfo 5 ~ Loginfo 1 | then | Loginfo10 ~ Loginfo6 |, using deque.pop_back
     - this case is problem, because user gets Loginfo 5, 4, 3, 2, 1 and then Loginfo 10,9,8,7,6

src/transaction/log_manager.c
joohok: Yes, I agree with you.
I will leave this one to refactor. 
it may be modified to `flashback_find_start_lsa (thread_p, log_page_p, context)` to use `log_page_p` which is fetched in the function.

src/transaction/log_manager.c
```diff
@@ -14684,6 +14733,441 @@ cdc_is_filtered_user (char *user)
...
+		cdc_make_dml_loginfo (thread_p, trid, context->user,
+				      rec_type == LOG_SUPPLEMENT_INSERT ? CDC_INSERT : CDC_TRIGGER_INSERT, classoid,
+				      NULL, &redo_recdes, log_info_entry, true);
+
+	      if (error != NO_ERROR)
+		{
+		  goto error;
+		}
+
+              // *INDENT-OFF*
```
wrlawodms: It needs `// *INDENT-OFF*`?

wrlawodms (approved):

src/transaction/log_manager.c
joohok: Oh.. no, I will remove it. At first, I thought I need to `INDENT-OFF` to c++ functions.

src/communication/network_interface_sr.c
hornetmj: OK. Let me know when you are ready.

hornetmj: I understand your feedback. Thanks.

src/transaction/log_manager.c
hornetmj: OK.

hornetmj (approved):

PR #3354, opened 2022-01-28
[CBRD-24191] Refactoring function 'dwb_adjust_write_buffer_values'.

hyahong:
http://jira.cubrid.org/browse/CBRD-24191

Purpose
Refactor the code to be a little more intuitive and consistent.

Implementation
N/A

Remarks
N/A

### DIFF
--- a/src/storage/double_write_buffer.c
+++ b/src/storage/double_write_buffer.c
@@ -737,6 +737,8 @@ dwb_adjust_write_buffer_values (unsigned int *p_double_write_buffer_size, unsign
   min_size = DWB_MIN_SIZE;
   max_size = DWB_MAX_SIZE;
 
+  assert (IS_POWER_OF_2 (min_size) && IS_POWER_OF_2 (max_size));
+
   if (*p_double_write_buffer_size < min_size)
     {
       *p_double_write_buffer_size = min_size;
@@ -762,6 +764,7 @@ dwb_adjust_write_buffer_values (unsigned int *p_double_write_buffer_size, unsign
   min_size = DWB_MIN_BLOCKS;
   max_size = DWB_MAX_BLOCKS;
 
+  assert (IS_POWER_OF_2 (min_size) && IS_POWER_OF_2 (max_size));
   assert (*p_num_blocks >= min_size);
 
   if (*p_num_blocks > max_size)
@@ -770,7 +773,7 @@ dwb_adjust_write_buffer_values (unsigned int *p_double_write_buffer_size, unsign
     }
   else if (!IS_POWER_OF_2 (*p_num_blocks))
     {
-      unsigned int block_limit = min_size << 2;
+      unsigned int block_limit = min_size << 1;
 
       while (block_limit < DWB_MAX_BLOCKS && block_limit < *p_num_blocks)
 	{
--- a/src/storage/double_write_buffer.c
+++ b/src/storage/double_write_buffer.c
@@ -337,8 +337,10 @@ STATIC_INLINE void dwb_signal_waiting_threads (DWB_WAIT_QUEUE * wait_queue, pthr
 STATIC_INLINE void dwb_destroy_wait_queue (DWB_WAIT_QUEUE * wait_queue, pthread_mutex_t * mutex);
 
 /* DWB functions */
-STATIC_INLINE void dwb_adjust_write_buffer_values (unsigned int *p_double_write_buffer_size,
-						   unsigned int *p_num_blocks) __attribute__ ((ALWAYS_INLINE));
+STATIC_INLINE void dwb_adjust_value (unsigned int min, unsigned int max, unsigned int *p_value)
+  __attribute__ ((ALWAYS_INLINE));
+STATIC_INLINE bool dwb_load_buffer_size (unsigned int *p_double_write_buffer_size) __attribute__ ((ALWAYS_INLINE));
+STATIC_INLINE bool dwb_load_block_count (unsigned int *p_num_blocks) __attribute__ ((ALWAYS_INLINE));
 STATIC_INLINE int dwb_wait_for_block_completion (THREAD_ENTRY * thread_p, unsigned int block_no)
   __attribute__ ((ALWAYS_INLINE));
 STATIC_INLINE int dwb_signal_waiting_thread (void *data) __attribute__ ((ALWAYS_INLINE));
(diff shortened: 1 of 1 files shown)


src/storage/double_write_buffer.c
```diff
@@ -741,57 +741,46 @@ dwb_adjust_write_buffer_values (unsigned int *p_double_write_buffer_size, unsign
...
 
-  if (*p_num_blocks > min_size)
+  if (*p_num_blocks > max_size)
+    {
+      *p_num_blocks = max_size;
+    }
+  else if (!IS_POWER_OF_2 (*p_num_blocks))
     {
-      if (*p_num_blocks > max_size)
+      unsigned int block_limit = min_size << 2;
```
wrlawodms: It looks correct now, but this seems to assume that `DWB_MIN_BLOCKS` is 1. If then, I have a comment. The value is defined separately, which means it should be decoupled with this code. We should be able to change the value without considering this code. And to use `min_size` to set this, it'd be better to `assert (IS_POWER_OF_2(min_size))`.
```suggestion
      assert (IS_POWER_OF_2(min_size));
      unsigned int block_limit = min_size << 1;
```

wrlawodms (commented):
Looks better and correct. Good job. 👍
I have an opinion, though.

hyahong:
@wrlawodms 
I couldn't think that the value should be decoupled.
Also, including the parts you suggested, it seems that assertion about min_size and max_size being a power of 2 is necessary.
Thank you!

wrlawodms (approved):
LGTM. Thank you for your continued contribution.
It seems to require a simple test case. @hornetmj will decide it.

src/storage/double_write_buffer.c
```diff
@@ -737,61 +737,53 @@ dwb_adjust_write_buffer_values (unsigned int *p_double_write_buffer_size, unsign
...
+  else if (*p_double_write_buffer_size > max_size)
+    {
+      *p_double_write_buffer_size = max_size;
+    }
+  else if (!IS_POWER_OF_2 (*p_double_write_buffer_size))
     {
-      if (*p_double_write_buffer_size > max_size)
+      unsigned int dwb_limit = min_size << 1;
+
+      while (dwb_limit < DWB_MAX_SIZE && dwb_limit < *p_double_write_buffer_size)
```
hornetmj:
```suggestion
      while (dwb_limit < *p_double_write_buffer_size)
```

hornetmj: "dwb_limit >= DWB_MAX_SIZE" is never happen.

src/storage/double_write_buffer.c
```diff
@@ -737,61 +737,53 @@ dwb_adjust_write_buffer_values (unsigned int *p_double_write_buffer_size, unsign
...
-		}
-	      limit1 = limit1 << 1;
-	    }
 
-	  *p_double_write_buffer_size = limit1;
-	}
+      *p_double_write_buffer_size = dwb_limit;
     }
 
+  assert (IS_POWER_OF_2 (*p_double_write_buffer_size));
```
hornetmj:
```suggestion
  assert (IS_POWER_OF_2 (*p_double_write_buffer_size) && *p_double_write_buffer_size >= min_size && *p_double_write_buffer_size <= max_size);
```

src/storage/double_write_buffer.c
```diff
@@ -737,61 +737,53 @@ dwb_adjust_write_buffer_values (unsigned int *p_double_write_buffer_size, unsign
...
+  if (*p_num_blocks > max_size)
+    {
+      *p_num_blocks = max_size;
+    }
+  else if (!IS_POWER_OF_2 (*p_num_blocks))
     {
-      if (*p_num_blocks > max_size)
+      unsigned int block_limit = min_size << 1;
+
+      while (block_limit < DWB_MAX_BLOCKS && block_limit < *p_num_blocks)
```
hornetmj: *p_num_blocks is always in the range min_size to max_size. (min_size < *p_num_blocks < max_size) So, "block_limit < DWB_MAX_BLOCKS &&" can be removed.

src/storage/double_write_buffer.c
```diff
@@ -737,61 +737,53 @@ dwb_adjust_write_buffer_values (unsigned int *p_double_write_buffer_size, unsign
...
+  assert (IS_POWER_OF_2 (*p_double_write_buffer_size));
+
   min_size = DWB_MIN_BLOCKS;
   max_size = DWB_MAX_BLOCKS;
 
+  assert (IS_POWER_OF_2 (min_size) && IS_POWER_OF_2 (max_size));
   assert (*p_num_blocks >= min_size);
 
-  if (*p_num_blocks > min_size)
+  if (*p_num_blocks > max_size)
```
hornetmj: In my opinion, it is better to check '*p_num_blocks < min_size' too. It is more readable even if it is already checked in the debug mode by assert.

hornetmj (approved):

src/storage/double_write_buffer.c
```diff
@@ -737,61 +737,53 @@ dwb_adjust_write_buffer_values (unsigned int *p_double_write_buffer_size, unsign
   min_size = DWB_MIN_SIZE;
   max_size = DWB_MAX_SIZE;
 
+  assert (IS_POWER_OF_2 (min_size) && IS_POWER_OF_2 (max_size));
```
joohok: is this assertion required? 
min_size and max_size are initialized using constant values ​​that satisfy the 'assert' condition.

And I think this min/max_size variable refers to the size of the buffer, not the number of blocks. So the assert should check if it is a multiple of 512 * 1024.

src/storage/double_write_buffer.c
```diff
@@ -737,61 +737,53 @@ dwb_adjust_write_buffer_values (unsigned int *p_double_write_buffer_size, unsign
...
-	}
+      *p_double_write_buffer_size = dwb_limit;
     }
 
+  assert (IS_POWER_OF_2 (*p_double_write_buffer_size));
+
   min_size = DWB_MIN_BLOCKS;
   max_size = DWB_MAX_BLOCKS;
 
+  assert (IS_POWER_OF_2 (min_size) && IS_POWER_OF_2 (max_size));
```
joohok: How about replacing `min/max_size` variables with `min/max_block` ?
I think it is more readable, and also is this assertion required? 

if the check is required, I think it is required to be mentioned when defining `DWB_MAX/MIN_BLOCKS` , `DBW_MAX/MIN_SIZE` at the top of `double_writer_buffer.c`.

hornetmj:
@Yechan0815 The behavior for determining *p_double_write_buffer_size and *p_num_blocks is exactly the same. Since it is the same pattern, How about creating a new function for it?

I recommend the following method. How about introducing the dwb_load_buffer_size() and dwb_load_block_count() functions instead of the dwb_adjust_write_buffer_values()? You can also remove the dwb_adjust_write_buffer_values(). Also, create a new function that takes the MIN/MAX value and returns it into a power of two.

# dwb_load_buffer_size(&double_write_buffer_size)
- double_write_buffer_size = prm_get_integer_value (PRM_ID_DWB_SIZE);
- check the double_write_buffer_size whether or not 0
- process other part as a function
# dwb_load_block_count(&num_blocks)
- num_blocks = prm_get_integer_value (PRM_ID_DWB_BLOCKS);
- check the num_blocks whether or not 0
- process other part as the function

hyahong:
@hornetmj 
It is definitely a effective structure. I will try it!

hyahong:
@joohok 
Because all calculations are done under the assumption that min/max_size is power of 2, I added debug assertion just in case. I will reflect your suggestion in structure to be changed!

eido5:
> @Yechan0815 The behavior for determining *p_double_write_buffer_size and *p_num_blocks is exactly the same. Since it is the same pattern, How about creating a new function for it?
> 
> I recommend the following method. How about introducing the dwb_load_buffer_size() and dwb_load_block_count() functions instead of the dwb_adjust_write_buffer_values()? You can also remove the dwb_adjust_write_buffer_values(). Also, create a new function that takes the MIN/MAX value and returns it into a power of two.
> 
> # dwb_load_buffer_size(&double_write_buffer_size)
> * double_write_buffer_size = prm_get_integer_value (PRM_ID_DWB_SIZE);
> * check the double_write_buffer_size whether or not 0
> * process other part as a function
> 
> # dwb_load_block_count(&num_blocks)
> * num_blocks = prm_get_integer_value (PRM_ID_DWB_BLOCKS);
> * check the num_blocks whether or not 0
> * process other part as the function

I may guess we can make "DWB_adjust()" using by overload in C++.
And then we can use just 1 function. because it is mentioned as "it works exaclty same".
And we can shorten the code length.
I cannot implement it right now as pseudo code but try to make pseudo code.

hyahong:
@hornetmj I changed it first. I don't know if it's okay.

hyahong:
@eido5 
Is it this kind of structure? This is a code that has been slightly changed in the current commit, and it has been used just one function.

(I'm attaching an image because I don't know how to upload the code changed without changing the commit.)

eido5:
> 

Totally nice.
In my opinion, it is more readable code and simple.
How about this implementation? @hornetmj , @wrlawodms , @joohok

src/storage/double_write_buffer.c
```diff
@@ -717,81 +719,86 @@ dwb_destroy_wait_queue (DWB_WAIT_QUEUE * wait_queue, pthread_mutex_t * mutex)
...
+ * dwb_load_buffer_size () - Load the buffer size value of double write buffer.
+ *
+ * return   : true of false.
+ * p_double_write_buffer_size (in/out) : the buffer size of double write buffer.
+ *
+ */
+STATIC_INLINE bool
+dwb_load_buffer_size (unsigned int *p_double_write_buffer_size)
+{
+  assert (IS_POWER_OF_2 (DWB_MIN_SIZE) && IS_POWER_OF_2 (DWB_MAX_SIZE));
```
joohok, line 772: If the reason you need assert() here is that `DWB_MIN_SIZE` and `DWB_MAX_SIZE`are defined outside of this code, then I think you also need to compare the sizes of `DWB_MIN_SIZE` and `DWB_MAX_SIZE` and check if `DWB_MIN_SIZE` is greater than 512K. what do you think?
- DWB_MAX_SIZE >= DWB_MIN_SIZE
- DWB_MIN_SIZE >= 512 * 1024

And how about mentioning these conditions where they are defined?

hyahong:
@joohok 
Values are defined by people, so I added a debug assertion to make sure that the values are valid.
adding the above condition such as "DWB_MAX_SIZE >= DWB_MIN_SIZE" makes it seem clearer.
If the structure doesn't change anymore, I will add conditions to dwb_load_buffer_size and dwb_load_block_count!

and sorry, I couldn't understand exactly what "mentioning these conditions where they are defined" means.
Could you please explain a little more in detail?

src/storage/double_write_buffer.c
joohok: @Yechan0815  
Please, reply to this comment. 

> @joohok
Values are defined by people, so I added a debug assertion to make sure that the values are valid.
adding the above condition such as "DWB_MAX_SIZE >= DWB_MIN_SIZE" makes it seem clearer.
If the structure doesn't change anymore, I will add conditions to dwb_load_buffer_size and dwb_load_block_count!

I understand what you mean, it will be nice! 

> and sorry, I couldn't understand exactly what "mentioning these conditions where they are defined" means.
Could you please explain a little more in detail?

I mean, it will be better if there is some comment where these values are defined. it is just an opinion.
```
#define DWB_MAX_SIZE          .... /* add some comment that this value is required to be power of 2 */
#define DWB_MIN_SIZE           .... /* add some comment that this value is required to be power of 2  and greater than 512K */
```

joohok (approved):
Nice work, Thanks for contribution

src/storage/double_write_buffer.c
```diff
@@ -717,81 +719,86 @@ dwb_destroy_wait_queue (DWB_WAIT_QUEUE * wait_queue, pthread_mutex_t * mutex)
...
-  max_size = DWB_MAX_BLOCKS;
+  assert (IS_POWER_OF_2 (*p_value));
+  assert (*p_value >= min && *p_value <= max);
+}
 
-  assert (*p_num_blocks >= min_size);
+/*
+ * dwb_load_buffer_size () - Load the buffer size value of double write buffer.
+ *
+ * return   : true of false.
```
wrlawodms:
```suggestion
 * return   : true or false.
```

src/storage/double_write_buffer.c
```diff
@@ -717,81 +719,86 @@ dwb_destroy_wait_queue (DWB_WAIT_QUEUE * wait_queue, pthread_mutex_t * mutex)
...
+    }
+  dwb_adjust_value (DWB_MIN_SIZE, DWB_MAX_SIZE, p_double_write_buffer_size);
+  return true;
+}
 
-	  *p_num_blocks = num_blocks << 1;
+/*
+ * dwb_load_block_count () - Load the block count value of double write buffer.
+ *
+ * return   : true of false.
```
wrlawodms:
```suggestion
 * return   : true or false.
```

src/storage/double_write_buffer.c
```diff
@@ -717,81 +719,86 @@ dwb_destroy_wait_queue (DWB_WAIT_QUEUE * wait_queue, pthread_mutex_t * mutex)
...
+ * dwb_adjust_value () - Adjust value to power of 2.
  *
  * return   : Error code.
- * p_double_write_buffer_size (in/out) : Double write buffer size.
- * p_num_blocks (in/out): The number of blocks.
+ * min (in) : minimum of value to be adjusted.
+ * max (in) : maximum of value to be adjusted.
+ * p_value (in/out) : value to be adjusted.
  *
- *  Note: The buffer size must be a multiple of 512 K. The number of blocks must be a power of 2.
```
wrlawodms, line 726: I think that this comment should be alive on `dwb_load_buffer_size()`. The buffer size must be a multiple of 512k, not only a power of 2.

src/storage/double_write_buffer.c
```diff
@@ -717,81 +719,86 @@ dwb_destroy_wait_queue (DWB_WAIT_QUEUE * wait_queue, pthread_mutex_t * mutex)
 }
 
 /*
- * dwb_adjust_write_buffer_values () - Adjust double write buffer values.
+ * dwb_adjust_value () - Adjust value to power of 2.
```
wrlawodms: I want it to be more meaningful so that it says the purpose or role of this function. For example, It can be `dwb_adjust_[something]_value()`, `dwb_power2_ceil()` or something. I suppose you can make a better choice than me.

wrlawodms (commented):
review on the changed code.

src/storage/double_write_buffer.c
```diff
@@ -717,81 +719,86 @@ dwb_destroy_wait_queue (DWB_WAIT_QUEUE * wait_queue, pthread_mutex_t * mutex)
...
-	  do
-	    {
-	      num_blocks = num_blocks & (num_blocks - 1);
-	    }
-	  while (!IS_POWER_OF_2 (num_blocks));
+      /* Do not use double write buffer. */
+      return false;
+    }
+  dwb_adjust_value (DWB_MIN_SIZE, DWB_MAX_SIZE, p_double_write_buffer_size);
+  return true;
```
hornetmj:
```suggestion
    }

  dwb_adjust_value (DWB_MIN_SIZE, DWB_MAX_SIZE, p_double_write_buffer_size);

  return true;
```

src/storage/double_write_buffer.c
```diff
@@ -717,81 +719,86 @@ dwb_destroy_wait_queue (DWB_WAIT_QUEUE * wait_queue, pthread_mutex_t * mutex)
...
-	  assert (*p_num_blocks <= max_size);
-	}
+  *p_num_blocks = prm_get_integer_value (PRM_ID_DWB_BLOCKS);
+  if (*p_num_blocks == 0)
+    {
+      /* Do not use double write buffer. */
+      return false;
     }
+  dwb_adjust_value (DWB_MIN_BLOCKS, DWB_MAX_BLOCKS, p_num_blocks);
+  return true;
```
hornetmj:
```suggestion

  dwb_adjust_value (DWB_MIN_BLOCKS, DWB_MAX_BLOCKS, p_num_blocks);

  return true;
```

hornetmj (approved):

hyahong:
@wrlawodms @joohok It's corrected. Thank you!

eido5 (approved):

hornetmj:
@Yechan0815 First, sorry for the late reply.

This is a refactoring and we expect that there is no execution change after merging this PR. So, I wonder how do you process the developer test to verify this. Please let me know when you are ready. I will make test scenarios but I want to know it.

hyahong:
@hornetmj 
First of all, I confirmed the same with the code flow.
And I verified it by debugger, setting values smaller than min, min value, values between min and max, max value, and values larger than max.
(Not all values between min and max are checked.)

+ I modified dwb, And I don't know why sql fails. Can I get some advice?

hornetmj:
@Yechan0815 The fails of the test are not related to your changes. So, don't worry about that. And I will merge this PR when we prepared the test scenarios to verify this. Thanks to share how you do the test.

hornetmj:
@mhoh3963 Please merge this PR. The failures of the CircleCI are not related to this PR.

hornetmj:
@mhoh3963 Please merge this PR. The failures of the CircleCI are not related to this PR. I am re-requesting because it seems that the previous request has not been confirmed.

PR #3355, opened 2022-01-31
[LETS-384] Core dumped in vacuum_check_not_vacuumed_rec_header

cristiarg:
http://jira.cubrid.org/browse/LETS-384

When an ATS is booted up, the mvcc/vacuum information in log header structure is relayed from PS to ATS via retrieval of entire log header page and copying the structure.
[Some of] the mvcc/vacuum fields must be kept up to date on page server to allow a restarted ATS to correctly initialize vacuum module at boot.
Implementation:
- field `log_Gl.hdr.mvcc_next_id` was already update in `replicator::read_and_redo_record`
- implemented function `log_replication_update_header_mvcc_info` to update mvcc/vacuum fields in log header based on investigating log records which are about to be replicated in heap pages
- some of the fields in log header are supposed to be kept on a per-vacuum-block basis; others are independent of vacuum block [id]
- in log replication, added handling for `LOG_MVCC_UNDO_DATA` and `LOG_SYSOP_END` to also process mvccid's present in these log records
- only bookkeep mvcc/vacuum log header fields on page server; on passive transction server, which also performs replication, it is not needed

Other:
- completed `log_dump_header` with missing fields; the logging remains inactivated to not break possible tests relying on comparing log
- Extracted function `prior_extract_vacuum_info_from_prior_node` for reuse in newly added logic executed on PS side.


src/transaction/log_append.cpp
```diff
@@ -151,6 +156,39 @@ log_prior_lsa_info::push_list (log_prior_node *&list_head, log_prior_node *&list
...
+	  prior_extract_vacuum_info_from_prior_node (node, dummy_vacuum_info, mvccid);
+	  assert (dummy_vacuum_info != nullptr);
+	  assert (mvccid != MVCCID_NULL);
+
+	  if (log_Gl.hdr.newest_block_mvccid < mvccid)
+	    {
+	      log_Gl.hdr.newest_block_mvccid = mvccid;
+	    }
+	  log_Gl.hdr.mvcc_op_log_lsa = node->start_lsa;
+	  log_Gl.hdr.does_block_need_vacuum = true;
```
cristiarg: The flag is not reset yet. I checked but don't know where is the appropriate place to reset it.

alexstanro: does_block_need_vacuum  keeps the "need vacuum" state regarding latest block. It must be reset on PS, when  the following condition is true:
"(vacuum_get_log_blockid (log_Gl.hdr.mvcc_op_log_lsa.pageid) != vacuum_get_log_blockid (start_lsa.pageid))".
This must be done when iterate through nodes.
This means that new block start, so, no need to vacuum the latest block.

cristiarg:
@alexstanro this PR is ready for final review

alexstanro (approved):

PR #3358, opened 2022-02-04
[CBRD-24181] Remove unnecessary select-list for unpushable subquery

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-24181

**Implementation**
The following functions have been added or modified

- mq_is_removable_select_list () : Check is it removal.
- mq_get_references() : Returns the columns referenced from mainquery.
- mq_rewrite_vclass_spec_as_derived () : Create a new subquery without unnecessary select list. (case of view)
- mq_remove_select_list_for_inline_view () : Create a new subquery without unnecessary select list (case of inline view)
Do not copy order_by when using aggregate functions, except for order-sensitive aggregate functions. In the case below, if order by is not deleted, unnecessary a, b, c cannot be deleted.
```
select count(*) from (select a,b,c from .. order by 1,2,3)
```
When using NO_MERGE hint, the select list is not deleted.

**Restrictions**
only for select query, Not for update, delete query
Not for merge query, CTE, HQ , value query
Not for method, distinct, define_vars

**bug fix**
Problem with doing not free intermediate tree when merging view
Problem with doing not copy data_type when copying order_by
Problem with using freed tree when semantic checking
Core dump occur when printing plan with constant value


shparkcubrid:
It is related to this issue from 036cef4 commit. sorry.

src/optimizer/query_graph.c
```diff
@@ -1494,13 +1494,11 @@ qo_insert_segment (QO_NODE * head, QO_NODE * tail, PT_NODE * node, QO_ENV * env,
...
   QO_SEG_IDX (seg) = env->nsegs;
-  /* add dummy name to segment example: dummy attr from view transfrom select count(*) from v select count(*) from
-   * (select {v}, 1 from t) v (v, 1) here, '1' is dummy attr set empty string to avoid core crash
-   */
   if (node)
     {
+      /* If it is not PT_NAME, an empty string is used for name. */
       QO_SEG_NAME (seg) =
-	node->info.name.original ? node->info.name.original : pt_append_string (QO_ENV_PARSER (env), NULL, "");
+	node->node_type == PT_NAME ? node->info.name.original : pt_append_string (QO_ENV_PARSER (env), NULL, "");
```
ctshim: Unlike before, is there no need to check the possibility that node->info.name.original is null?
Shouldn't it be something like the following?
`(node->node_type == PT_NAME && node->info.name.original) ? node->info.name.original : pt_append_string (QO_ENV_PARSER (env), NULL, "");`

src/parser/view_transform.c
```diff
@@ -1715,6 +1875,126 @@ mq_is_pushable_subquery (PARSER_CONTEXT * parser, PT_NODE * subquery, PT_NODE *
...
+    {
+      return NON_PUSHABLE;
+    }
+
+  /* check for schema query */
+  if (PT_SELECT_INFO_IS_FLAGED (mainquery, PT_SELECT_INFO_COLS_SCHEMA)
+      || PT_SELECT_INFO_IS_FLAGED (mainquery, PT_SELECT_FULL_INFO_COLS_SCHEMA))
+    {
+      return NON_PUSHABLE;
+    }
```
ctshim: How about modifying it like below?  
```
/* check for merge/ update, delete/schema   query */
static int non_pushable_flag = PT_SELECT_INFO_IS_MERGE_QUERY | PT_SELECT_INFO_IS_UPD_DEL_QUERY | PT_SELECT_INFO_COLS_SCHEMA | PT_SELECT_FULL_INFO_COLS_SCHEMA;

  if (PT_SELECT_INFO_IS_FLAGED (mainquery, non_pushable_flag ))
    {
      return NON_PUSHABLE;
    }
```

src/optimizer/query_graph.c
shparkcubrid: I think the previous check was wrong. But adding a null check is fine. I'll fix it.

src/parser/view_transform.c
shparkcubrid: okay. I'll fix it.

ctshim (approved):

youngjinj (approved):

beyondykk9 (approved):

hgryoo (approved):

hyunikn (approved):

PR #3360, opened 2022-02-09
[CBRD-24205] Invalid xasl error occurs when plan cache removed while executing executeBatch

airnet73:
http://jira.cubrid.org/browse/CBRD-24205

Purpose
If the plan cache is deleted while executing multiple queries using executeBatch, all subsequent queries are not executed.
In this case, if the plan cache is regenerated, the query can be executed normally.
Since an error does not always occur, it can be reproduced only by forcibly deleting the plan cache during the test.

Implementation
N/A

Remarks
N/A

### DIFF
--- a/src/broker/cas_execute.c
+++ b/src/broker/cas_execute.c
@@ -2593,6 +2593,7 @@ ux_execute_array (T_SRV_HANDLE * srv_handle, int argc, void **argv, T_NET_BUF * 
   DB_VALUE val;
   DB_OBJECT *ins_obj_p;
   T_BROKER_VERSION client_version = req_info->client_version;
+  int retried_query_num = 0;
 
   if (srv_handle == NULL || srv_handle->schema_type >= CCI_SCH_FIRST)
     {
@@ -2708,6 +2709,10 @@ ux_execute_array (T_SRV_HANDLE * srv_handle, int argc, void **argv, T_NET_BUF * 
 
       if (res_count < 0)
 	{
+	  if (res_count == ER_QPROC_INVALID_XASLNODE && retried_query_num != num_query)
+	    {
+	      goto exec_retry_xasl_error;
+	    }
 	  goto exec_db_error;
 	}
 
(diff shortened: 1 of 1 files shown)


src/broker/cas_execute.c
```diff
@@ -2804,6 +2809,38 @@ ux_execute_array (T_SRV_HANDLE * srv_handle, int argc, void **argv, T_NET_BUF *
...
+	{
+	  goto exec_db_error;
+	}
+
+      retried_query_num = num_query;
+
+      srv_handle->session = session;
+      srv_handle->q_result->stmt_id = stmt_id;
+
+      num_query--;
```
kisoo-han: What about rewrite codes from `exec_retry_xasl_error:` to the end as a function for readability.

airnet73: I made part of exec_retry_xasl_error: into a function.

src/broker/cas_execute.c
```diff
@@ -2708,6 +2710,10 @@ ux_execute_array (T_SRV_HANDLE * srv_handle, int argc, void **argv, T_NET_BUF *
 
       if (res_count < 0)
 	{
+	  if (res_count == ER_QPROC_INVALID_XASLNODE && retried_query_num != num_query)
+	    {
+	      goto exec_retry_xasl_error;
+	    }
```
kisoo-han, line 2724: My suggestion (it is trivial and do as you like):
1. remove goto statement & copy goto block here (2816 ~ 2824) like this:
```
 if (res_count == ER_QPROC_INVALID_XASLNODE && retried_query_num != num_query)
  {
  err_code = recompile_statement (srv_handle);
  if (err_code < 0)
    {
      goto exec_db_error;
    }
  session = (DB_SESSION *) srv_handle->session;
  retried_query_num = num_query;
  num_query--;
  continue;
}
```

2. Remove `continue `of line# 2813, and 'exec_retry_xasl_error:' of line# 2815

kisoo-han (approved):
It looks better than before!

src/broker/cas_execute.c
airnet73: Thank you for your advice.

PR #3361, opened 2022-02-09
[CBRD-24176] Construct a SQL statement with the received log items

joohok:
http://jira.cubrid.org/browse/CBRD-24176

Purpose

Flashback utility gets log items from server, and it needs to interpret those log items and make SQL statement. 
This PR shows how to construct SQL statements using log item (loginfo list).


src/executables/util_cs.c
```diff
@@ -4250,23 +5042,35 @@ flashback (UTIL_FUNCTION_ARG * arg)
...
+  do
     {
-      db_shutdown ();
-      goto error_exit;
+      error =
+	flashback_get_loginfo (trid, user, oid_list, num_tables, &start_lsa, &end_lsa, &num_item, is_oldest,
+			       &loginfo_list);
+      if (error != NO_ERROR)
+	{
+	  db_shutdown ();
```
hornetmj: How about using a flag for it and processing it at the end of the function?

src/executables/util_cs.c
```diff
@@ -3994,6 +3994,798 @@ tde (UTIL_FUNCTION_ARG * arg)
...
+  ptr = *loginfo;
+
+  classop = sm_find_class (classname);
+  error = au_fetch_class (classop, &class_, AU_FETCH_READ, AU_SELECT);
+  if (error != NO_ERROR)
+    {
+      goto error;
+    }
+
+  sprintf (sql, "delete from %s where ", classname);
```
hornetmj: The table name should be enclosed by [ and ]. And maybe other identifiers. The following test is successfully executed, unfortunately.
- create table [test 2] (c1 int);
- insert into [test 2] valus (1);
- select * from [test 2];

But, an sql fails to execute if there is no [ and ].

hornetmj: Should we support an additional utility option for that? What do guys(@joohok @wrlawodms @eido5) think about it?

wrlawodms: All object name including column's and table's has to be enclosed with []. It lets the names be distinguished with reserved keywords such as 'table' and 'create'. For example, we can do `create table [table] (a int);` with this. @hornetmj You mean an option that enables printing []? It looks mandatory and not optional to me.

joohok: I agree with you. 
I didn't consider about it. I will fix it

eido5 (approved):
It looks there are no typos.

wrlawodms (commented):
I think all other functions except `flashback()` should move to another file like `flashback.c` and they need a prefix for their names. Let me have your opinion on it. They also need function comments.

src/executables/util_cs.c
```diff
@@ -3994,6 +3994,798 @@ tde (UTIL_FUNCTION_ARG * arg)
...
+      ptr = or_unpack_int64 (ptr, &b_classoid);
+      memcpy (&classoid, &b_classoid, sizeof (OID));
+
+      class_index = get_class_index (da_size (classlist), oidlist, classoid);
+
+      da_get (classlist, class_index, classname);
+
+      switch (dml_type)
+	{
+	case 0:
```
wrlawodms: Why don't you define an enumeration for it?

src/executables/util_cs.c
```diff
@@ -3994,6 +3994,798 @@ tde (UTIL_FUNCTION_ARG * arg)
...
+	case 1:
+	  /* UPDATE */
+	  error = print_flashback_update (&ptr, trid, user, classname, is_detail, outfp);
+	  break;
+	case 2:
+	  /* DELETE */
+	  error = print_flashback_delete (&ptr, trid, user, classname, is_detail, outfp);
+	  break;
+	default:
+	  assert (dml_type < 3 || dml_type >= 0);
```
wrlawodms: Maybe `dml_type < 3 && dml_type >= 0`? and it never happens.
```suggestion
	  assert (false);
```

src/executables/util_cs.c
```diff
@@ -3994,6 +3994,798 @@ tde (UTIL_FUNCTION_ARG * arg)
...
+	{
+	  return idx;
+	}
+    }
+
+  return -1;
+}
+
+static int
+check_sql_memory (char **sql, int req_size, int *max_sql_size)
```
wrlawodms: It does not only check but also resize. The name should include this work.

joohok:
> I think all other functions except `flashback()` should move to another file like `flashback.c` and they need a prefix for their names. Let me have your opinion on it. They also need function comments.

Yes, I will move it to flashback files.
I'm now planning a flashback_sr.c/h for server side flashback processing, and flashback_cl.c/h for utility side flashback processing. 
But, now the PRs are processed in parallel , so I can not move the functions.. 
when PRs are merged, I will refactor these. 

In the integrate version that I used for testing, I moved the functions to flashback files.

and I will add function comments

src/executables/util_cs.c
```diff
@@ -3994,6 +3994,798 @@ tde (UTIL_FUNCTION_ARG * arg)
...
+  ptr = or_unpack_int (ptr, &num_change_col);
+  for (i = 0; i < num_change_col; i++)
+    {
+      ptr = or_unpack_int (ptr, &change_index);
+    }
+
+  attr = class_->ordered_attributes;
+  for (i = 0; i < num_change_col; i++)
+    {
+      error = check_sql_memory (&sql, strlen (sql) + strlen (attr->header.name) + 12, &max_sql_size);
```
wrlawodms: \+ 12 for what? It needs a comment.

src/executables/util_cs.c
```diff
@@ -3994,6 +3994,798 @@ tde (UTIL_FUNCTION_ARG * arg)
...
+
+  int num_cond_col;
+  int cond_index;
+
+  int i = 0;
+
+  sql = (char *) malloc (max_sql_size);
+  if (sql == NULL)
+    {
+      error = ER_OUT_OF_VIRTUAL_MEMORY;
```
wrlawodms, line 4214: er_set(). and check others too.

wrlawodms: or you think it doesn't need?

src/executables/util_cs.c
joohok: yes, I agree with you

src/executables/util_cs.c
```diff
@@ -3994,6 +3994,798 @@ tde (UTIL_FUNCTION_ARG * arg)
...
+  float f_data;
+  short sh_data;
+  char *s_data;
+  INT64 b_data;
+
+  int max_decimal_digits_len = 20;
+
+  ptr = or_unpack_int (ptr, &func_code);
+
+  switch (func_code)
```
joohok: I will add enumeration for func_code

src/executables/util_cs.c
joohok: Oh.. I will add 
`util_log_write_errid (MSGCAT_UTIL_GENERIC_NO_MEM)`

I will do other error setting and message setting, too

src/executables/util_cs.c
joohok: Yes, I will add comment.

src/executables/util_cs.c
joohok: OK, I will fix it.

src/executables/util_cs.c
joohok: Yes, I will use a flag for db_shutdown.

hornetmj: OK.

src/executables/util_cs.c
hornetmj: OK. I agree with the opinion of @wrlawodms.

src/executables/util_cs.c
joohok: I set `need_shutdown` flag, 
and if this flag is set to true, then `db_shutdown()` is called at the `error_exit` label. 

this flag is already set in previous issue (#3348), so I will apply it when it merged.

hornetmj (approved):

src/executables/util_cs.c
```diff
@@ -3994,6 +3994,827 @@ tde (UTIL_FUNCTION_ARG * arg)
...
+  return error;
+}
+
+typedef enum
+{
+  FLASHBACK_INSERT = 0,
+  FLASHBACK_UPDATE,
+  FLASHBACK_DELETE
+} FLASHBACK_DML_TYPE;
+static int
```
```suggestion
} FLASHBACK_DML_TYPE;

static int
```

src/executables/util_cs.c
```diff
@@ -3994,6 +3994,827 @@ tde (UTIL_FUNCTION_ARG * arg)
...
+
+      ptr = or_unpack_int (ptr, (int *) &dml_type);
+      ptr = or_unpack_int64 (ptr, &b_classoid);
+      memcpy (&classoid, &b_classoid, sizeof (OID));
+
+      class_index = get_class_index (da_size (classlist), oidlist, classoid);
+
+      da_get (classlist, class_index, classname);
+
+      /* TODO: make enum for dml type */
```
wrlawodms: Seems that you already have made enum.

src/executables/util_cs.c
```diff
@@ -3994,6 +3994,827 @@ tde (UTIL_FUNCTION_ARG * arg)
...
+  float f_data;
+  short sh_data;
+  char *s_data;
+  INT64 b_data;
+
+  int max_decimal_digits_len = 20;
+
+  ptr = or_unpack_int (ptr, (int *) &func_type);
+
+  /* TODO: make enum for func code */
```
wrlawodms: Seems that you already have made enum for it.

src/executables/util_cs.c
```diff
@@ -3994,6 +3994,827 @@ tde (UTIL_FUNCTION_ARG * arg)
...
+  PACK_FUNC_TYPE func_type;
+
+  int i_data;
+  double d_data;
+  float f_data;
+  short sh_data;
+  char *s_data;
+  INT64 b_data;
+
+  int max_decimal_digits_len = 20;
```
wrlawodms: Could you tell me what does it mean?

src/executables/util_cs.c
joohok: Thanks! I will remove it

src/executables/util_cs.c
joohok: OK, I will fix it

src/executables/util_cs.c
joohok: Thanks, it was progress of applying feedback. I will remove it

src/executables/util_cs.c
joohok: Simply, it is maximum length of the converted string when converting a decimal number to a string.
INT64 presents 9223372036854775807, and this consist of 20 char.

wrlawodms: Thank you for the answer. It would be better to leave a comment for it and make it a constexpr.

wrlawodms (approved):

src/executables/util_cs.c
joohok: Thanks, I will change it to constexpr, and leave a comment

PR #3362, opened 2022-02-09
[LETS-386] on server boot, during recovery and until log recovery undo, inhibit comparison between local and remote storage heap pages

cristiarg:
http://jira.cubrid.org/browse/LETS-386

During recovery, do not strictly compare local and page server retrieved heap pages because local retrieved pages might be behind actual state due to recovery not having been performed yet.
This occurs in early stages of server boot. Once log recovery passes the "redo" sequence, pages on local (ATS) and remote (PS) storage must already be in sync and strict comparison can resume.

Other:
- made some functions inline

### DIFF
--- a/src/storage/page_buffer.c
+++ b/src/storage/page_buffer.c
@@ -8202,7 +8202,7 @@ pgbuf_read_page_from_file_or_page_server (THREAD_ENTRY * thread_p, const VPID * 
 	   *    active transaction server (ie: the page still exists), but the log has been applied on
 	   *    page server and the page cannot be retrieved
 	   */
-	  if (log_is_restarted_or_in_crash_recovery_but_past_redo ())
+	  if (log_is_in_crash_recovery_but_past_redo_or_restarted ())
 	    {
 #if !defined(NDEBUG)
 	      if (!(io_page->prv == second_io_page->prv))
@@ -8226,7 +8226,7 @@ pgbuf_read_page_from_file_or_page_server (THREAD_ENTRY * thread_p, const VPID * 
 	      const bool local_lsa_is_less_than_or_equal_to_page_server_lsa
 		= (io_page->prv.lsa <= second_io_page->prv.lsa);
 	      const bool equal_vpid = io_page->prv.volid == second_io_page->prv.volid
-		&& io_page->prv.pageid == second_io_page->prv.pageid && io_page->prv.ptype == second_io_page->prv.ptype;
+		&& io_page->prv.pageid == second_io_page->prv.pageid;
 #if !defined(NDEBUG)
 	      if (!local_lsa_is_less_than_or_equal_to_page_server_lsa && !equal_vpid)
 		{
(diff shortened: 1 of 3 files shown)


src/storage/page_buffer.c
```diff
@@ -8185,15 +8185,66 @@ pgbuf_read_page_from_file_or_page_server (THREAD_ENTRY * thread_p, const VPID *
...
+		  er_log_debug (ARG_FILE_LINE, "pgbuf_read_page_from_file_or_page_server"
+				"past recovery redo pages not equal\n"
+				"    target repl LSA: %lld|%d\n"
+				"    local page  VPID: %d|%d  LSA: %lld|%d  ptype: %d\n"
+				"    remote page VPID: %d|%d  LSA: %lld|%d  ptype: %d\n",
+				LSA_AS_ARGS (&target_repl_lsa),
+				io_page->prv.volid, io_page->prv.pageid,
+				LSA_AS_ARGS (&io_page->prv.lsa), io_page->prv.ptype,
+				second_io_page->prv.volid, second_io_page->prv.pageid,
+				LSA_AS_ARGS (&second_io_page->prv.lsa), second_io_page->prv.ptype);
```
alexstanro, line 8219: How about adding assert (false) here and remove "assert (io_page->prv == second_io_page->prv);"?

src/transaction/log_manager.h
```diff
@@ -288,4 +289,32 @@ void LOG_CS_PROMOTE (THREAD_ENTRY * thread_p);
...
+  else
+    {
+      return false;
+    }
+}
+
+inline bool
+log_is_in_crash_recovery_but_past_redo_or_restarted (void)
+{
+  if (LOG_RESTARTED == log_Gl.rcv_phase || LOG_RECOVERY_UNDO_PHASE == log_Gl.rcv_phase
```
alexstanro, line 312: How about  changing condition as : "if (log_Gl.rcv_phase  == LOG_RESTARTED || log_Gl.rcv_phase == LOG_RECOVERY_UNDO_PHASE)"?

alexstanro (approved):
The patch looks correct. I have only small observations.

I think would be helpful to check BOOT_DB_PARM after redo phase. Must be the same LSA on PS/ATS after redo. It can be done in a separate patch.

src/transaction/log_manager.h
cristiarg: This form, although ugly, is safer for c-style code because of the inherent unsafe nature of the language - avoids accidental assignment (`log_Gl.rcv_phase = LOG_RESTARTED`) which would also evaluate to true.

src/storage/page_buffer.c
cristiarg: I was thinking that, even with `NDEBUG` defined, there are still compilations flags that would still allow the `assert` to be raised. That is the reason why the logging block is separated from the assert.

PR #3364, opened 2022-02-10
[LETS-386] on ATS boot, during recovery analysis, inhibit full comparison between local and page server log pages

cristiarg:
http://jira.cubrid.org/browse/LETS-386

On active transaction server with local storage log pages are retrieved from both local storage and also from page server. They are then compared to ensure they have the same layout.
On a booting up active transaction server performing recovery, the value of `log_Gl.hdr.append_lsa.offset` is only updated at the end of the log recovery analysis phase (callstack: log_recovery_analysis -> log_rv_analysis_record -> log_rv_analysis_record_on_tran_server -> log_rv_analysis_log_end).
However, during this phase, log pages are also retrieved (and compared) - for the purpose of performing log recovery analysis on them.
The pages have to be the same in entirety, except the last page that contains the end of log record for which the `tranid` is different (Active Transaction Server and Page Server have different mechanism of appending this end of log record).
This patch ensures that:
- all but the last log pages are identically equal (as before)
- in case a page is not identically equal, it has to be the last page containing the end of log record
- this difference regarding end of log record can only happen during the log recovery analysis phase.

### DIFF
--- a/src/transaction/log_page_buffer.c
+++ b/src/transaction/log_page_buffer.c
@@ -2259,6 +2259,7 @@ logpb_verify_page_read (LOG_PAGEID pageid, const LOG_PAGE * left_log_pgptr, cons
 	      pages_equal = (!left_has_valid_checksum || !rite_has_valid_checksum);
 	      break;
 	    }
+
 	  // next log record still in the same page, advance to it
 	  curr_left_lsa = left_log_rec_header->forw_lsa;
 	  curr_rite_lsa = rite_log_rec_header->forw_lsa;
--- a/src/transaction/log_page_buffer.c
+++ b/src/transaction/log_page_buffer.c
@@ -2242,9 +2242,10 @@ logpb_verify_page_read (LOG_PAGEID pageid, const LOG_PAGE * left_log_pgptr, cons
 	      break;
 	    }
 
+	  // if not the last page, headers must be equal
 	  assert (*left_log_rec_header == *rite_log_rec_header);
 
-	  // might not be the last page, must stop, the pages are not equal
+	  // if not the last log page, must stop, the pages are not equal
 	  if (left_log_rec_header->forw_lsa.pageid != curr_left_lsa.pageid
 	      || rite_log_rec_header->forw_lsa.pageid != curr_rite_lsa.pageid)
 	    {
@@ -2252,7 +2253,7 @@ logpb_verify_page_read (LOG_PAGEID pageid, const LOG_PAGE * left_log_pgptr, cons
 		      && rite_log_rec_header->forw_lsa.pageid != curr_rite_lsa.pageid);
 
 	      // it is not this function's job to validate page corruption
-	      // if any of the pages is corrupted, decide withhold decision for log recovery analysis logic
+	      // if any of the pages is corrupted, withhold decision for log recovery analysis logic
 	      const bool left_has_valid_checksum = logpb_page_has_valid_checksum (left_log_pgptr);
 	      const bool rite_has_valid_checksum = logpb_page_has_valid_checksum (rite_log_pgptr);
 	      pages_equal = (!left_has_valid_checksum || !rite_has_valid_checksum);
--- a/src/transaction/log_page_buffer.c
+++ b/src/transaction/log_page_buffer.c
@@ -2195,13 +2195,15 @@ logpb_verify_page_read (LOG_PAGEID pageid, const LOG_PAGE * left_log_pgptr, cons
   if (!pages_equal && log_Gl.rcv_phase == LOG_RECOVERY_ANALYSIS_PHASE)
     {
       /* NOTE:
-       *  - during early stages of the recovery phase, we cannot rely on log header append lsa to
-       *    know how much of the page area to compare because it is not updated until later in the
-       *    recovery process
+       *  - during analysis phase of the recovery on ATS, the value of log header append lsa cannot be relied
+       *    upon to know how much of the page area to compare because it is not updated until the end of the
+       *    analysis (see call stack: log_recovery_analysis -> log_rv_analysis_record ->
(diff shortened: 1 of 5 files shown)


src/transaction/log_page_buffer.c
```diff
@@ -2192,7 +2192,69 @@ logpb_verify_page_read (LOG_PAGEID pageid, const LOG_PAGE * left_log_pgptr, cons
...
+	  assert (left_log_rec_header->prev_tranlsa == rite_log_rec_header->prev_tranlsa
+		  && left_log_rec_header->back_lsa == rite_log_rec_header->back_lsa
+		  && left_log_rec_header->forw_lsa == rite_log_rec_header->forw_lsa
+		  && left_log_rec_header->trid == rite_log_rec_header->trid
+		  && left_log_rec_header->type == rite_log_rec_header->type);
+
+	  // might not be the last page, must stop, the pages are not equal
+	  if (left_log_rec_header->forw_lsa.pageid != curr_left_lsa.pageid
+	      || rite_log_rec_header->forw_lsa.pageid != curr_rite_lsa.pageid)
+	    {
```
alexstanro, line 2251: Please add a comment: it is not the last page, and the content differs => should not happen, force crash!

alexstanro (approved):
You can consider also checking log page corruption.

src/transaction/log_page_buffer.c
cristiarg: Made the comment before `if` more clear.

src/transaction/log_page_buffer.c
```diff
@@ -2192,8 +2192,85 @@ logpb_verify_page_read (LOG_PAGEID pageid, const LOG_PAGE * left_log_pgptr, cons
...
+		      //&& left_log_rec_header->trid == rite_log_rec_header->trid
+		      && left_log_rec_header->type == rite_log_rec_header->type);
+
+	      assert (left_log_rec_header->forw_lsa.is_null () && rite_log_rec_header->forw_lsa.is_null ());
+	      assert (left_log_rec_header->type == LOG_END_OF_LOG);
+
+	      // compare everything until end of log
+	      const char *const left_log_page_area = left_log_pgptr->area;
+	      const char *const rite_log_page_area = rite_log_pgptr->area;
+	      const int cmp_res = strncmp (left_log_page_area, rite_log_page_area, curr_left_lsa.offset
```
alexstanro, line 2239: Shouldn't be memcmp here?

src/transaction/log_page_buffer.c
```diff
@@ -2192,8 +2192,85 @@ logpb_verify_page_read (LOG_PAGEID pageid, const LOG_PAGE * left_log_pgptr, cons
...
+	  // if not the last log page, must stop, the pages are not equal
+	  if (left_log_rec_header->forw_lsa.pageid != curr_left_lsa.pageid
+	      || rite_log_rec_header->forw_lsa.pageid != curr_rite_lsa.pageid)
+	    {
+	      assert (left_log_rec_header->forw_lsa.pageid != curr_left_lsa.pageid
+		      && rite_log_rec_header->forw_lsa.pageid != curr_rite_lsa.pageid);
+
+	      // it is not this function's job to validate page corruption
+	      // if any of the pages is corrupted, withhold decision for log recovery analysis logic
+	      const bool left_has_valid_checksum = logpb_page_has_valid_checksum (left_log_pgptr);
```
alexstanro, line 2257: I think that this check must be done in all cases.

src/transaction/log_page_buffer.c
```diff
@@ -2192,8 +2192,85 @@ logpb_verify_page_read (LOG_PAGEID pageid, const LOG_PAGE * left_log_pgptr, cons
...
+       *  - check if page is the last one in the log record by iterating log records
+       * the last log record (the one that contains the end of the log) is allowed to be different
+       *
+       * NOTE:
+       *  this iteration is too simplistic and brittle (as opposed to the one in log recovery analysis); but,
+       *  as this code has more of a temporary nature, it is considered good enough
+       */
+      const LOG_RECORD_HEADER *left_log_rec_header = nullptr;
+      const LOG_RECORD_HEADER *rite_log_rec_header = nullptr;
+      log_lsa curr_left_lsa = { pageid, left_log_pgptr->hdr.offset };
```
alexstanro, line 2216: left_log_pgptr->hdr.offset must be compared with rite_log_pgptr->hdr.offset for safety.

Also, pageid must be compared.

src/transaction/log_page_buffer.c
```diff
@@ -2192,8 +2192,85 @@ logpb_verify_page_read (LOG_PAGEID pageid, const LOG_PAGE * left_log_pgptr, cons
...
+	  left_log_rec_header = (const LOG_RECORD_HEADER *) (left_log_pgptr->area + curr_left_lsa.offset);
+	  rite_log_rec_header = (const LOG_RECORD_HEADER *) (rite_log_pgptr->area + curr_rite_lsa.offset);
+	}
+    }
+  else if (pageid == log_Gl.hdr.append_lsa.pageid)
+    {
+      // last page, not in recovery analysis
       const char *const left_log_page_area = left_log_pgptr->area;
       const char *const rite_log_page_area = rite_log_pgptr->area;
       const int cmp_res = strncmp (left_log_page_area, rite_log_page_area, log_Gl.hdr.append_lsa.offset);
```
alexstanro, line 2276: log_Gl.hdr.append_lsa may be higher on PTS than on ATS. Thus, latest log records may not arrived yet on PTS.

src/transaction/log_page_buffer.c
cristiarg: @alexstanro 
Right. Since this issue manifests in more places, I have added the new issue http://jira.cubrid.org/browse/LETS-396 to address it.

PR #3368, opened 2022-02-11
[CBRD-24184] Fix compile warning: PACKAGE_STRING redefined

airnet73:
http://jira.cubrid.org/browse/CBRD-24184

Purpose
PACKAGE_STRING is defined in both version.h and unixodbc_conf.h of unixODBC downloaded.
If waring occurs during compilation, CUBRID may have incorrect information.
The waring has been removed so that CUBRID does not get incorrect information.

Implementation
N/A

Remarks
N/A


kisoo-han (approved):
I suggest some different point of view:

**SIZEOF_LONG_INT** is used to determine the **SQLLEN/SQLULEN** size. It is defined as 32 in 32-bit OS and 64 in 64-bit OS.

There is no problem to include **unixodbc_conf.h** when **_building unixodbc manager itself_**. However, in the unixodbc application, is it necessary to include unixodbc_conf.h? There is no definition referenced in unixodbc_defs.h other than **SIZEOF_LONG_INT** in **_unixodbc_** application.
So before including **sqltypes.h**, what about insert following codes (**cas_cgw.h, cas_handle.h**):

```
#if !defined (SIZEOF_LONG_INT)
#define SIZEOF_LONG_INT 8
#endif

```
Then, unixodbc_conf.h will not be included. The **cas_cgw** is a part of CUBRID Engine, and we do not support 32-bit OS. So, we do not need to consider whether this machine is 32/64. That is why we simply define SIZEOF_LONG_INT as 8, not considering 4.

However, if you choose this, you have to test functionality again for cas_cgw. **The decision is up to you.**

kisoo-han (approved):
Thank you for accepting my comments with an open mind!!

PR #3369, opened 2022-02-11
[LETS-382] do not compare page LSA for btree root pages due to btree statistics

cristiarg:
http://jira.cubrid.org/browse/LETS-382

- btree root stats are updated differently on active transaction server and page server:
  - on active transaction server, there exists a mechanism to cache the statistics in memory and only write to the page at checkpoint
  - on page server, the statistics are applied as the log is replicated
- therefore, for btree root page's, do not check equality of page LSA


src/transaction/log_replication.cpp
```diff
@@ -518,7 +518,12 @@ namespace cublog
       }
 
     btree_root_update_stats (&thread_entry, root_page, stats);
-    pgbuf_set_lsa (&thread_entry, root_page, &record_lsa);
+
+    // it is not needed to update page LSA for btree root statistics
+    // if this is done, in scalability scenarios with local storage, the LSA's of active transaction server's
+    // and page server's pages will differ because of this update
```
alexstanro: I don't think is correct. If the system crashes and restarts, I guess that unique stats are wrong. You may check that the difference is unique stats and ignore it.

cristiarg: Refactored to not compare btree root pages' LSAs

src/storage/btree.c
```diff
@@ -33431,6 +33431,43 @@ btree_get_perf_btree_page_type (THREAD_ENTRY * thread_p, PAGE_PTR page_ptr)
...
 
+/* btree_is_btree_root_page - investigative check whether a page is a btree root page
+ *
+ * NOTE: function just assumes that it investigates a PAGE_BTREE; however, due to the low
+ *    layer the function is used within, it cannot assert that the page is of the correct
+ *    type; it is the collers responsability to ensure this
+ */
+bool
+btree_is_btree_root_page (THREAD_ENTRY * thread_p, const PAGE_PTR page_ptr)
+{
```
alexstanro, line 33442: assert (page_ptr != NULL)

src/storage/page_buffer.c
```diff
@@ -8196,32 +8196,14 @@ pgbuf_read_page_from_file_or_page_server (THREAD_ENTRY * thread_p, const VPID *
...
-				LSA_AS_ARGS (&second_io_page->prv.lsa), second_io_page->prv.ptype);
-		}
-#endif
-	      assert (io_page->prv == second_io_page->prv);
-	    }
-	  else
+	  if (log_is_in_crash_recovery_and_not_yet_completes_redo ())
 	    {
 	      const bool local_lsa_is_less_than_or_equal_to_page_server_lsa
 		= (io_page->prv.lsa <= second_io_page->prv.lsa);
```
alexstanro: You may use LSA_LE.

src/storage/page_buffer.c
```diff
@@ -8244,6 +8226,53 @@ pgbuf_read_page_from_file_or_page_server (THREAD_ENTRY * thread_p, const VPID *
...
+		  if (io_page->prv.ptype == PAGE_BTREE && second_io_page->prv.ptype == PAGE_BTREE
+		      && btree_is_btree_root_page (thread_p, pgptr)
+		      && btree_is_btree_root_page (thread_p, second_pgptr))
+		    {
+		      // page server might already contain btree stats which are still kept in memory on the
+		      // active transaction server
+		      assert (io_page->prv.lsa <= second_io_page->prv.lsa
+			      && io_page->prv.pageid == second_io_page->prv.pageid
+			      && io_page->prv.volid == second_io_page->prv.volid
+			      && io_page->prv.pflag == second_io_page->prv.pflag
```
alexstanro: Duplicate condition

src/storage/page_buffer.c
```diff
@@ -8244,6 +8226,53 @@ pgbuf_read_page_from_file_or_page_server (THREAD_ENTRY * thread_p, const VPID *
...
+		   */
+		  const PAGE_PTR pgptr = (PAGE_PTR) ((char *) io_page->page);
+		  PAGE_PTR second_pgptr = (PAGE_PTR) ((char *) second_io_page->page);
+		  if (io_page->prv.ptype == PAGE_BTREE && second_io_page->prv.ptype == PAGE_BTREE
+		      && btree_is_btree_root_page (thread_p, pgptr)
+		      && btree_is_btree_root_page (thread_p, second_pgptr))
+		    {
+		      // page server might already contain btree stats which are still kept in memory on the
+		      // active transaction server
+		      assert (io_page->prv.lsa <= second_io_page->prv.lsa
```
alexstanro: Please, use LSA_LE

src/storage/page_buffer.c
```diff
@@ -8244,6 +8226,53 @@ pgbuf_read_page_from_file_or_page_server (THREAD_ENTRY * thread_p, const VPID *
...
+			      && io_page->prv.pageid == second_io_page->prv.pageid
+			      && io_page->prv.volid == second_io_page->prv.volid
+			      && io_page->prv.pflag == second_io_page->prv.pflag
+			      && io_page->prv.pflag == second_io_page->prv.pflag
+			      && io_page->prv.tde_nonce == second_io_page->prv.tde_nonce);
+		    }
+		  else
+		    {
+#if !defined(NDEBUG)
+		      if (!fileio_pages_equal)
```
alexstanro: previously checked

src/storage/page_buffer.c
```diff
@@ -8244,6 +8226,53 @@ pgbuf_read_page_from_file_or_page_server (THREAD_ENTRY * thread_p, const VPID *
...
+					"    local page  VPID: %d|%d  LSA: %lld|%d  ptype: %d\n"
+					"    remote page VPID: %d|%d  LSA: %lld|%d  ptype: %d\n",
+					LSA_AS_ARGS (&target_repl_lsa),
+					io_page->prv.volid, io_page->prv.pageid,
+					LSA_AS_ARGS (&io_page->prv.lsa), io_page->prv.ptype,
+					second_io_page->prv.volid, second_io_page->prv.pageid,
+					LSA_AS_ARGS (&second_io_page->prv.lsa), second_io_page->prv.ptype);
+			}
+#endif
+		      assert (fileio_pages_equal);
```
alexstanro: I guess that assert must be removed.

src/storage/page_buffer.c
cristiarg: Condition is not duplicated because `io_page->prv.lsa <= second_io_page->prv.lsa` (less than).

cristiarg: Should have been `ptype`.

src/storage/page_buffer.c
cristiarg: Removed

src/storage/page_buffer.c
cristiarg: The assert must stay. If the page is not a btree root page (checked in the previous if), then check must fail.

src/storage/btree.c
cristiarg: Done

src/storage/page_buffer.c
cristiarg: Done.

src/storage/page_buffer.c

src/storage/page_buffer.c
alexstanro: My expectation was assert (false), but is ok.

alexstanro (approved):

PR #3376, opened 2022-02-22
[CBRD-23986] Add an index on data_type of _db_domain

wrlawodms:
http://jira.cubrid.org/browse/CBRD-23986

### DIFF
--- a/src/transaction/boot_cl.c
+++ b/src/transaction/boot_cl.c
@@ -2121,7 +2121,8 @@ boot_define_domain (MOP class_mop)
   SM_TEMPLATE *def;
   char domain_string[32];
   int error_code = NO_ERROR;
-  const char *index_col_names[2] = { "object_of", NULL };
+  const char *index_col_names1[2] = { "object_of", NULL };
+  const char *index_col_names2[2] = { "data_type", NULL };
 
   def = smt_edit_class_mop (class_mop, AU_ALTER);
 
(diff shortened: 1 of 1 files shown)


src/transaction/boot_cl.c
```diff
@@ -2121,7 +2121,8 @@ boot_define_domain (MOP class_mop)
   SM_TEMPLATE *def;
   char domain_string[32];
   int error_code = NO_ERROR;
-  const char *index_col_names[2] = { "object_of", NULL };
+  const char *index_col_names1[2] = { "object_of", NULL };
+  const char *index_col_names2[2] = { "data_type", NULL };
```
joohok, line 2125: In my opinion, I think it would be better to add a comment for this index.
 Because the other system classes have only one index each. Perhaps this is to reduce the cost of 'select' that system class. 
 However, it will be helpful for maintenance if you give a brief comment of why this system class has two indexes. 
 it is just a opinion.

joohok (approved):

eido5 (approved):
It look so good.

src/transaction/boot_cl.c
wrlawodms: Thank you for the feedback. As you said, other system classes have only one index. They seem for some internal queries, and the newly added one has the same goal. With this PR, only _db_domain will have two indexes, but they have the same context, so I don't think we should explain why.

joohok: Thanks for explain. I understand what you mean.

hornetmj (approved):

hornetmj:
@wrlawodms Please follow the description format when PR next time.

wrlawodms:
@hornetmj OK, I'll follow if an additional description is needed.

PR #3377, opened 2022-02-22
[CBRD-24101] Expand in-line view : patch legacy bug

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-24101

reproduction
```
DROP TABLE IF EXISTS t1;
CREATE TABLE t1(id INT, title VARCHAR(10), read_count INT);
INSERT INTO t1 VALUES(1,'book1',3),(2,'book2',5),(3,'book3',1),(4,'book4',0),(5,'book5',3),(6,'book6',2),(7,'book7',0);
COMMIT WORK;
SELECT DECR(read_count) FROM (select sleep(1)) x, t1 WHERE id >6;
```
result
```
ERROR: Attribute of derived class is not permitted in DECR()
```
expected
```
no error
```

When view merging, a new node is used and the first node is freed. ( mq_translate_tree())
A problem with using a freed node has been detected. I change to use the latest node.


shparkcubrid:
2d04055 is only for this issue. sorry.

hgryoo (approved):

ctshim (approved):

youngjinj (approved):

hyunikn (approved):

beyondykk9 (approved):

PR #3378, opened 2022-02-23
[CBRD-24174] Prints out flashback summary information

joohok:
http://jira.cubrid.org/browse/CBRD-24174

purpose
prints summary information. 

Implementation
flashback_unpack_and_print_summary ()
- unpacking the area retrieved from server
- make brief summary info map
- print the summary info
- If the summary is printed beyond the window size, check whether to print more


joohok:
Please, review the code in flashback_cl.c 
main function for printing summary information is `flashback_unpack_and_print_summary` 

flashback_get_summary () calls it, and get a trid input from user.
and flashback() uses the trid to get log info. 

Modification on existing code which calls `flashback_unpack_and_print_summary` is just for showing how this function works. 

I will notify you again when previous PRs are merged and "Modification on existing code" are ready to review.

hornetmj:
@joohok Can you share the execution result on a terminal as a screenshot or something like that on the jira issue? I'd like to know what it looks like bofore I complete the code review.

src/communication/network_interface_cl.c
```diff
@@ -10637,14 +10639,18 @@ flashback_get_summary (dynamic_array * class_list, const char *user, time_t star
 	      ptr = or_unpack_oid (ptr, &(*oid_list)[i]);
 	    }
 	  /* get summary info */
+
+	  ptr = or_unpack_int (ptr, &num_summary);
+
+	  error_code = flashback_unpack_and_print_summary (&ptr, summary, num_summary, class_list, *oid_list);
```
wrlawodms, line 10657: `flashback_get_summary()` has two jobs, retrieving and printing. In my opinion, it would be better to separate them if possible for maintenance and flexibility. I recommend just retrieving here and printing out of this function.

joohok:
> @joohok Can you share the execution result on a terminal as a screenshot or something like that on the jira issue? I'd like to know what it looks like bofore I complete the code review.

OK, I will share it

src/transaction/flashback_cl.c
```diff
@@ -0,0 +1,197 @@
...
+ * flashback_find_class_index () - find the index of oidlist which contains classoid
+ *
+ * return              : class index
+ * classname_list (in) : classnames
+ * oidlist (in)        : OID of classnames
+ * classoid (in)       : classoid to find
+ */
+
+static int
+flashback_find_class_index (dynamic_array * classname_list, OID * oidlist, OID classoid)
```
wrlawodms: How about just the size of the oid list? It seems more flexible and that classname_list is not needed.
```suggestion
flashback_find_class_index (OID * oidlist, int list_size, OID classoid)
```

src/transaction/flashback_cl.c
```diff
@@ -0,0 +1,197 @@
...
+      tmp_ptr = or_unpack_log_lsa (tmp_ptr, &end_lsa);
+      tmp_ptr = or_unpack_int (tmp_ptr, &num_table);
+
+      info.trid = trid;
+      info.start_lsa = start_lsa;
+      info.end_lsa = end_lsa;
+
+      /* *INDENT-OFF* */
+      summary->emplace (trid, info);
+      /* *INDENT-ON* */
```
wrlawodms:
```suggestion
      summary->emplace (trid, info);
```

src/transaction/flashback_cl.c
```diff
@@ -0,0 +1,197 @@
...
+	  char c;
+	  printf ("press 'q' to quit or press anything to continue");
+	  c = getchar ();
+	  if (c == 'q')
+	    {
+	      stop_print = true;
+	    }
+
+	  len = 0;
+	  printf ("\033[A\33[2K\r");
```
wrlawodms, line 190: What does it mean?

src/transaction/flashback_cl.c
```diff
@@ -0,0 +1,197 @@
...
+  char *user = NULL;
+  time_t start_time, end_time;
+  int num_insert, num_update, num_delete;
+  LOG_LSA start_lsa, end_lsa;
+  int num_table;
+  OID classoid;
+  char *tmp_ptr = *ptr;
+
+  bool stop_print = false;
+  int max_window_size = flashback_util_get_winsize ();
```
wrlawodms:
```suggestion
  const int max_window_size = flashback_util_get_winsize ();
```

src/transaction/flashback_cl.c
```diff
@@ -0,0 +1,197 @@
...
+  int len = 0;
```
wrlawodms: better to use a more meaningful one like line_len or line_cnt.

src/communication/network_interface_cl.c
joohok: Yes, you are right. 
I really wanted to somehow separate the two functions of retrieving and printing.
That was the detailed design point, but when I tried to separate the two functions and perform the printing function outside, 
the total amount of memory to be maintained in the utility increased, and number of memory copy increased.

To minimize the memory to be maintained, only (trid, start_lsa, end_lsa) is saved, but all summary information must be maintained for printing outside. what do you think?

src/transaction/flashback_cl.c
joohok: Oh, you are right. Thank you

src/transaction/flashback_cl.c
joohok: OK, I will change it

src/transaction/flashback_cl.c
joohok: OK, I will modify it to line_cnt. Thanks

src/transaction/flashback_cl.c
joohok: Thanks!

src/transaction/flashback_cl.c
joohok: It means, erase the line `"press 'q' to quit or press anything to continue"`.
because above message is no more required when user press any key.

wrlawodms: Aha! I've got it. How about leaving a comment?

joohok: Oh, It seems better! I will leave a comment

src/transaction/flashback_cl.c
```diff
@@ -0,0 +1,197 @@
...
+
+  printf
+    ("  transaction id  |    user   |   start time    |   end time   |   num_insert   |   num_update   |   num_delete   |   tables   ");
+  len++;
+
+  for (int i = 0; i < num_summary; i++)
+    {
+      FLASHBACK_SUMMARY_INFO info = { 0, LSA_INITIALIZER, LSA_INITIALIZER };
+
+      /* testing code, it will be replaced with print function */
```
wrlawodms: It is a still test code?

src/transaction/flashback_cl.c
```diff
@@ -0,0 +1,197 @@
...
+ * return              : error code
+ * ptr (in/out)        : memory pointer where to unpack
+ * summary (out)       : brief summary information
+ * num_summary (in)    : number of summary to unpack
+ * classname_list (in) : classnames
+ * oidlist (in)        : OID of classnames
+ */
+
+int
+flashback_unpack_and_print_summary (char **ptr, FLASHBACK_SUMMARY_INFO_MAP * summary, int num_summary,
```
wrlawodms: `ptr` means nothing. Can we use a more meaningful one?

src/communication/network_interface_cl.c
wrlawodms: Ok, now I understand and it sounds reasonable. I think a quick solution is renaming the function name. `flashback_get_summary` to `flashback_get_and_show_summary` or something else.

joohok: OK, it seems better to change the function name.

src/transaction/flashback_cl.c
joohok: Yes, I will change it to more meaningful one.
I used it following `or_unpack_* (char *ptr, ..)`, but it needs to be changed.

src/transaction/flashback_cl.c
joohok: Oh.. I will remove it

src/communication/network_interface_cl.c
joohok: I will change the function name when previous PRs are merged.

src/transaction/flashback_cl.c
```diff
@@ -0,0 +1,196 @@
...
+  char *tmp_ptr = *summary_buffer;
+
+  bool stop_print = false;
+  const int max_window_size = flashback_util_get_winsize ();
+  int line_cnt = 0;
+
+  printf
+    ("  transaction id  |    user   |   start time    |   end time   |   num_insert   |   num_update   |   num_delete   |   tables   ");
+
+  line_cnt++;
```
eido5, line 116: Why the `line_cnt` is not 1?
Are there any reasons that `line_cnt` should be  initialized 0?

joohok: Since no lines were initially printed on the screen, line_cnt is initialized to 0. 
It is more intuitive to increment line_cnt after every line is printed, even if it initially outputs one line unconditionally.

src/communication/network_interface_cl.c
```diff
@@ -10561,58 +10561,8 @@ loaddb_update_stats ()
...
-    }
-
-  *ptr = tmp_ptr;
-
-  return NO_ERROR;
-}
-
 /*
- * flashabck_get_summary () - client-side function to get flashback summary
+ * flashabck_get_and_show_summary () - client-side function to get flashback summary
```
hornetmj: The description in the comment should be changed together to explain the show part.

src/transaction/flashback_cl.h
```diff
@@ -0,0 +1,63 @@
...
+
+#ifndef _FLASHBACK_CL_H_
+#define _FLASHBACK_CL_H_
+
+#include <unordered_map>
+#include "dynamic_array.h"
+#include "storage_common.h"
+#include "log_lsa.hpp"
+
+#define FLASHBACK_FIND_SUMMARY_ENTRY(trid, summary_entry, summary_info) \
```
hornetmj: IMHO, it would be nice to make the order of parameters 'in in out'.

But, this is not a strong request. So, you can decide whether or not to change it.

src/transaction/flashback_cl.c
```diff
@@ -0,0 +1,197 @@
...
+#include <sys/ioctl.h>
+#include <unistd.h>
+#endif /* LINUX */
+
+#include "flashback_cl.h"
+#include "object_representation.h"
+
+#if defined(WINDOWS)
+static int
+flashback_util_get_winsize ()
```
hornetmj, line 40: We need to test on Windows for this. Please add test scenarios for it to the test list.
- If the size of the windows is changed during the execution.
- If the size of the windows is too small to output even one line (need to consider the heading, information messages, etc).

src/transaction/flashback_cl.c
hornetmj: Is it work on Windows too?

And when the user chooses the 'q', I think that it is enough just to finish the output of the summary info and return success in the function. Is there any reason to use the stop_print flag and to keep going the execution? I think that you don't gather rest summary info because it is determined that the user already found the necessary information(transaction id).

hornetmj (approved):

src/communication/network_interface_cl.c
joohok: OK, I will change the comment

src/transaction/flashback_cl.c
joohok: OK, I will add it to the test scenario.

src/transaction/flashback_cl.h
joohok: OK, it seems better. 
I will change it

src/transaction/flashback_cl.c
joohok: @hornetmj 
1. I have to test on Windows.
2. The case you mentioned is when the user uses it normally. 
    This is in case the user executes it more than once. 
    So user knows that there is a transaction after the printed summary information so far.

src/transaction/flashback_cl.h
```diff
@@ -0,0 +1,63 @@
...
+  TRANID trid;
+  LOG_LSA start_lsa;
+  LOG_LSA end_lsa;
+} FLASHBACK_SUMMARY_INFO;
+
+// *INDENT-OFF*
+typedef std::unordered_map<TRANID, FLASHBACK_SUMMARY_INFO> FLASHBACK_SUMMARY_INFO_MAP;
+// *INDENT-ON*
+
+extern int flashback_unpack_and_print_summary (char **ptr, FLASHBACK_SUMMARY_INFO_MAP * summary, int num_summary,
```
wrlawodms:
```suggestion
extern int flashback_unpack_and_print_summary (char **summary_buffer, FLASHBACK_SUMMARY_INFO_MAP * summary, int num_summary,
```

wrlawodms (approved):

src/transaction/flashback_cl.c
eido5: Oh, I see. Thank you for the explaination.

eido5 (approved):

PR #3380, opened 2022-02-23
[CBRD-24210] Fix an error when executing SHOW EXEC STATISTICS;

ctshim:
http://jira.cubrid.org/browse/CBRD-24210

After applying CBRD-24117, the error occurs.
The error occurs in the following two statements:

```
SHOW EXEC STATISTICS;
SHOW EXEC STATISTICS ALL;
```

### DIFF
--- a/src/parser/csql_grammar.y
+++ b/src/parser/csql_grammar.y
@@ -1652,6 +1652,8 @@ stmt_list
 			  }
 
 		DBG_PRINT}}
+        | ';'
+                {{ /* empty line*/ }}
 	;
 
 


hyunikn (approved):

youngjinj (approved):

src/parser/csql_grammar.y
```diff
@@ -1652,6 +1652,8 @@ stmt_list
 			  }
 
 		DBG_PRINT}}
+        | ';'
```
youngjinj, line 1655: Is it for a quick fix?

Is it difficult to change the following?
It seems that the following are not simple changes.
(Errors related to CUBRID_STMT_TYPE, etc.)
```
stmt_list	
        : stmt_list stmt
        | stmt
        ;

stmt
        : stmt_
        | ';'
```

ctshim: The example provided is the format when stmts do not need to be separated by semicolons.

I didn't quite understand what you said about CUBRID_STMT_TYPE.

shparkcubrid (approved):

hgryoo (approved):

beyondykk9 (approved):

PR #3383, opened 2022-02-24
[CBRD-24177] Block archive log volumes from being removed during flashback

joohok:
http://jira.cubrid.org/browse/CBRD-24177

Purpose

This is a PR to prevent the archive volume referenced by flashback from being deleted. 
if it continue to prevent deletion of archive volumes, archive volumes may accumulate and the disk may become full.

Implementation

To prevent archive log volumes from being removed 
1. set flashback_Min_log_pageid_to_keep which is used in logpb_remove_archive_logs, logpb_remove_archive_logs_exceeds_limit.
2. set flashback_Is_active that represents state of flashback processing the request. 
    when flashback is active, then flashback_Min_log_pageid_to_keep will not be unset.

To prevent disk full, minimum log pageid to keep is unset under the following 2 conditions.
1. Interval between requests exceeds threshold
3. when flashback is finished.


src/communication/network_interface_sr.c
```diff
@@ -10849,13 +10849,17 @@ sflashback_get_summary (THREAD_ENTRY * thread_p, unsigned int rid, char *request
...
+  FLASHBACK_SET_REQUEST_TIME ();
+
   /* get summary list */
   error_code = flashback_make_summary_list (thread_p, &context);
   if (error_code != NO_ERROR)
     {
       goto error;
     }
 
+  FLASHBACK_SET_MIN_LOG_PAGEID_TO_KEEP (&context.start_lsa);
```
wrlawodms: What if the log volume is truncated between `flashback_make_summary_list()` and here? It seems that logs after `context.start_lsa` are possibly deleted.

src/communication/network_interface_sr.c
```diff
@@ -10972,12 +10978,23 @@ sflashback_get_loginfo (THREAD_ENTRY * thread_p, unsigned int rid, char *request
...
       goto error;
     }
 
+  if (flashback_is_loginfo_generation_finished (&context.start_lsa, &context.end_lsa))
+    {
+      FLASHBACK_UNSET_MIN_LOG_PAGEID_TO_KEEP ();
+    }
+  else
+    {
+      FLASHBACK_SET_MIN_LOG_PAGEID_TO_KEEP (&context.start_lsa);
```
wrlawodms: I don't remember exactly, so I have a question: Is `context.start_lsa` <= `context.end_lsa` always true regardless of forward/backward?

src/communication/network_interface_sr.c
hornetmj: I think this mechanism need to consider the following things:
- context.start_lsa == NULL_LSA
- between getting summaries and getting log infos

src/communication/network_interface_sr.c
```diff
@@ -10972,12 +10978,23 @@ sflashback_get_loginfo (THREAD_ENTRY * thread_p, unsigned int rid, char *request
   ptr = or_unpack_int (ptr, &context.num_loginfo);
   ptr = or_unpack_int (ptr, &context.forward);
 
+  FLASHBACK_SET_REQUEST_TIME ();
```
hornetmj: IMHO, it would be nice to place at the position immediately after the function call.

src/transaction/flashback.h
```diff
@@ -55,6 +55,49 @@
...
+#define FLASHBACK_SET_REQUEST_TIME() \
+  do \
+    { \
+      flashback_Last_request_time = time (NULL); \
+    } \
+  while (0)
+
+extern LOG_PAGEID flashback_Min_log_pageid;	// Minumun log pageid to keep archive log volume from being removed
+extern time_t flashback_Last_request_time;	// The time most recently requested by flashback
+extern int flashback_Threshold_to_remove_archive;	/* If the difference between the time at which the archive log is deleted
```
wrlawodms: In my opinion, it would be better to make these static global variables and provide interfaces to set them instead of macros. They don't have to be exposed to other modules. You already define getters. Could you tell me the reason for making setters as macros and variables accessible from outside?

src/communication/network_interface_sr.c
hornetmj: How about checking the value of the 'supplemental_log' and returning an error with an error_code if its value is 0 in the first place of this function?

src/transaction/log_page_buffer.c
```diff
@@ -6240,6 +6277,23 @@ logpb_remove_archive_logs (THREAD_ENTRY * thread_p, const char *info_reason)
 	  min_arv_required_for_cdc--;
 	  last_deleted_arv_num = MIN (last_deleted_arv_num, min_arv_required_for_cdc);
 	}
+
```
wrlawodms, line 6277: How about adding the flashback and the cdc case to the function comment? The comment only mentions recovery and vacuum.

wrlawodms (commented):
Do we have a mechanism to prevent users from running the flashback utility concurrently?

src/transaction/log_page_buffer.c
```diff
@@ -6092,6 +6096,36 @@ logpb_remove_archive_logs_exceed_limit (THREAD_ENTRY * thread_p, int max_count)
 		  assert (false);
 		}
 	    }
+
+	  /* check flashback */
+	  if (flashback_check_time_to_remove_archive ())
```
hornetmj: This mechanism makes a timing problem. I mean that every request should be completed within the threshold. But, the archive volumes should never be deleted during the request is processed. So, you just consider the time between a request and the next request and abnormal termination of connection.

wrlawodms:
What if a user request arrives after timeout? In that case, it seems that archives are removed. Have you already handled that case?

src/communication/network_interface_sr.c
joohok: Yes, you are right. 
I will set the flag that represents flashback state (run or stop), and prevent the archive log volumes from being deleted.

joohok: 1. yes
   * I will add `assert(!LSA_ISNULL(&context.start_lsa))`
   * I will check intervals between requests, and if intervals exceeds threshold then return error
2. Yes, I will check supplemental_log is on at utility side, and add `assert (supplemental_log >=1)` here.

src/communication/network_interface_sr.c
joohok: Yes, this will be used for check interval, so I will move it to the end of the function.

src/communication/network_interface_sr.c
joohok: Yes, `context.start_lsa` <= `context.end_lsa` is always true.

src/transaction/flashback.h
joohok: Oh.. yes you are right. Thank you. I will change it to static global variables, and provide setter function.
It was my mistake to set these variable as global variables that can be accessed from outside.

src/transaction/log_page_buffer.c
joohok: Yes, I will consider the intervals between requests and flashback state.

src/transaction/log_page_buffer.c
joohok: Ah, OK. I didn't consider it. Thanks.

joohok:
> Do we have a mechanism to prevent users to run the flashback utility concurrently?

Oh no.. I didn't consider it. I have to think about it and it may be dealt in next PR. Thank you

> What if a user request arrives after timeout? In that case, it seems that archives are removed. Have you already handled that case?

When we discuss the specification, we decided to return error. And it is not handled yet. 
@hornetmj, @wrlawodms both of you mentioned this issue at the same time. Thank you. I'll deal with it right away.

src/transaction/flashback.c
```diff
@@ -55,6 +55,130 @@
...
+/*
+ * flashback_set_threshold - set flashback_Threshold_to_remove_archive
+ *                           it follows PRM_ID_REMOVE_LOG_ARCHIVES_INTERVAL
+ */
+
+void
+flashback_set_threshold ()
+{
+  flashback_Threshold_to_remove_archive = prm_get_integer_value (PRM_ID_REMOVE_LOG_ARCHIVES_INTERVAL);
+  if (flashback_Threshold_to_remove_archive == 0)
```
hornetmj: How about < 60 instead of == 0?

src/transaction/flashback.c
```diff
@@ -55,6 +55,130 @@
...
+}
+
+/*
+ * flashback_is_loginfo_generation_finished - check if it is finished to generate log infos
+ *
+ * return   : true or false
+ */
+
+bool
+flashback_is_loginfo_generation_finished (LOG_LSA * start_lsa, LOG_LSA * end_lsa)
```
hornetmj, line 194: Can you remind me about the start_lsa and the end_lsa? And it would be nice to describe them in the comment.

src/executables/util_cs.c
```diff
@@ -4940,6 +4940,12 @@ flashback (UTIL_FUNCTION_ARG * arg)
   is_detail = utility_get_option_bool_value (arg_map, FLASHBACK_DETAIL_S);
   is_oldest = utility_get_option_bool_value (arg_map, FLASHBACK_OLDEST_S);
 
+  if (!prm_get_integer_value (PRM_ID_SUPPLEMENTAL_LOG))
+    {
+      fprintf (stderr, "please set \"supplemental_log=1\" in conf/cubrid.conf\n");
```
hornetmj: 'supplemental_log=2' is also supported to the flashback.

hornetmj (commented):
We discussed it online meeting. Please check my opinions and arrange an online meeting to review this code together.
- The global variables are needed to protect from accessing multi-threads (worker, daemon).
- The state changes should be done in the very first place in callback functions.
- The remove archive daemon thread only considers the flashback info when the flashback utility is running.
- etc.

src/executables/util_cs.c
joohok: OK, I changed it

src/transaction/flashback.c
joohok: OK, I leave a comment for it

src/transaction/flashback.c
joohok: Yes, it will be nice. there will be frequent error when interval is too short

src/base/error_code.h
```diff
@@ -1703,8 +1703,9 @@
 #define ER_FLASHBACK_EXCEED_MAX_NUM_TRAN_TO_SUMMARY -1335
 #define ER_FLASHBACK_SCHEMA_CHANGED                 -1336
 #define ER_FLASHBACK_LOG_NOT_EXIST                  -1337
+#define ER_FLASHBACK_EXCEED_THRESHOLD               -1338
```
wrlawodms: How about `ER_FLASHBACK_TIMEOUT` or something? This name doesn't contain the meaning of what kind of THRESHOLD it is.

joohok: Yes, I will change it to something. 
this error occurs when timeout between requests, so it is better to be replaced with `ER_FLASHBACK_TIMEOUT`

src/transaction/flashback.c
```diff
@@ -55,6 +55,177 @@
...
+/*
+ * flashback_set_threshold - set flashback_Threshold_to_remove_archive
+ *                           it follows PRM_ID_REMOVE_LOG_ARCHIVES_INTERVAL
+ */
+
+static void
+flashback_set_threshold ()
+{
+  flashback_Threshold_to_remove_archive = prm_get_integer_value (PRM_ID_REMOVE_LOG_ARCHIVES_INTERVAL);
+  if (flashback_Threshold_to_remove_archive < 60)
```
eido5: 1. Is this value, 60, fixed(maybe seconds? or ms?)? If not, what about expression macro? I suggest `THRESHOLD_LIMIT_SECOND` rough.
3. Where can I find the reason of this policy(60)?

src/transaction/flashback.c
```diff
@@ -171,6 +171,21 @@ flashback_is_needed_to_keep_archive ()
...
 }
 
+
+/*
+ * flashback_reset_variables - reset flashback global variables
+ *
+ */
+
+void
+flashback_reset_variables ()
```
wrlawodms: The way I see it, `flashback_reset` is enough and better. Just an opinion, it's up to you.

src/transaction/flashback.c
joohok: Thank you for feedback, 
1. I will define constant value for it. it may seems more readable
2. 60 is the recommended value for system parameter "remove_log_archive_interval_in_secs" 
    in manual. (https://www.cubrid.org/manual/en/11.0/admin/config.html#logging-related-parameters)

eido5: Oh, I see. 
Thank you for the link. I checked it.

src/transaction/flashback.c
```diff
@@ -55,6 +55,177 @@
...
+flashback_is_needed_to_keep_archive ()
+{
+  if (flashback_Min_log_pageid == NULL_LOG_PAGEID)
+    {
+      /* if nothing is set on flashback_Min_log_pageid, then there is no need to keep archive */
+
+      return false;
+    }
+
+  if (!flashback_Is_active && flashback_check_time_exceed_threshold ())
```
wrlawodms, line 139: Is it OK not using `flashback_Is_active.load()`?

src/transaction/flashback.c
```diff
@@ -55,6 +55,177 @@
...
+
+  flashback_Min_log_pageid = lsa->pageid;
+}
+
+/*
+ * flashback_unset_min_log_pageid_to_keep - unset flashback_Min_log_pageid to NULL_LOG_PAGEID
+ */
+
+void
+flashback_unset_min_log_pageid_to_keep ()
```
wrlawodms: It seems that this is not used any more.

src/transaction/flashback.c
```diff
@@ -55,6 +55,177 @@
...
+/*
+ * flashback_check_time_to_remove_archive - check the time if the archive can be removed
+ *
+ * return   : true or false
+ */
+
+bool
+flashback_check_time_exceed_threshold ()
+{
+  if (flashback_Threshold_to_remove_archive == 0)
```
wrlawodms: `flashback_Threshold_to_remove_archive` is used only in this function. Do we really need to declare this as a global variable? It seems that one local variable can be used for it like:
```
int threshold = prm_get_integer_value (PRM_ID_REMOVE_LOG_ARCHIVES_INTERVAL);
threshold = threshold < 60 ? 60 : threshold.
```

wrlawodms: `flashback_set_threshold` can be also removed.

src/transaction/log_page_buffer.c
```diff
@@ -6092,6 +6096,33 @@ logpb_remove_archive_logs_exceed_limit (THREAD_ENTRY * thread_p, int max_count)
 		  assert (false);
 		}
 	    }
+
+	  /* check flashback */
+	  if (flashback_is_needed_to_keep_archive ())
+	    {
+	      flashback_first_pageid = flashback_min_log_pageid_to_keep ();
```
wrlawodms, line 6103: There seems to be a race condition. After `flashback_is_needed_to_keep_archive()` returns true, `flashback_Min_log_pageid` can be set `NULL_LOG_PAGEID` because a flashback request faces an error or finishes. What do you think?

src/transaction/flashback.c
joohok: Yes, you are right. I'll remove it.

src/transaction/flashback.c
joohok: Yes, you are right. 
`flashback_Threshold_to_remove_archive` is used only in this function. I tried to process it locally, but I wondered if it would be right to set the variable every time this function is called. 
However I don't think it need to be set globally. I will fix it.

src/transaction/flashback.c
joohok: Load() is executed implicitly. But I'll change it to load() for readability.

src/transaction/flashback.c
joohok: Oh.. I think it is better, I will change it.

src/transaction/log_page_buffer.c
joohok: yes, you are right. 
I need to consider that case, and add condition in line 6108 
`if (flashback_first_pageid != NULL_LOG_PAGEID && logpb_is_page_in_archive (flashback_first_pageid)) `

src/transaction/flashback.c
wrlawodms: No, It is Ok if it is done implicitly. We do neither store() explicilty.

src/transaction/flashback.c
wrlawodms: If that is your concern, you can use a static local variable, but I think it doesn't matter.

src/transaction/log_page_buffer.c
wrlawodms: How about just using mutex and try_lock instead of atomic variable and the active variable. It seems that this is not performance-intensive code, which means we don't need to make it lock-free and worry about tricky race conditions and memory barriers.

src/communication/network_interface_sr.c
```diff
@@ -10888,17 +10905,35 @@ sflashback_get_summary (THREAD_ENTRY * thread_p, unsigned int rid, char *request
...
-  css_send_reply_and_data_to_client (thread_p->conn_entry, rid, reply, OR_ALIGNED_BUF_SIZE (a_reply), area, area_size);
+  error_code =
+    css_send_reply_and_data_to_client (thread_p->conn_entry, rid, reply, OR_ALIGNED_BUF_SIZE (a_reply), area,
+				       area_size);
 
   db_private_free_and_init (thread_p, area);
 
+  if (error_code != NO_ERROR)
+    {
+      flashback_reset_variables ();
```
hornetmj: How about using the same error handling mechanism?

e.g. goto css_error

wrlawodms (approved):

src/transaction/log_page_buffer.c
wrlawodms: > How about just using mutex and try_lock instead of atomic variable and the active variable. It seems that this is not performance-intensive code, which means we don't need to make it lock-free and worry about tricky race conditions and memory barriers.

I don't think we have enough time to consider another detailed design. Just let it go.

src/communication/network_interface_sr.c
```diff
@@ -11005,16 +11049,37 @@ sflashback_get_loginfo (THREAD_ENTRY * thread_p, unsigned int rid, char *request
...
   ptr = flashback_pack_loginfo (thread_p, ptr, context);
 
-  css_send_reply_and_data_to_client (thread_p->conn_entry, rid, reply, OR_ALIGNED_BUF_SIZE (a_reply), area, area_size);
+  error_code =
+    css_send_reply_and_data_to_client (thread_p->conn_entry, rid, reply, OR_ALIGNED_BUF_SIZE (a_reply), area,
+				       area_size);
 
   db_private_free_and_init (thread_p, area);
 
+  if (flashback_is_loginfo_generation_finished (&context.start_lsa, &context.end_lsa) || error_code != NO_ERROR)
```
hornetmj: How about using the same error handling mechanism when the error_code is checked?

e.g. goto css_error

eido5 (approved):

src/transaction/log_page_buffer.c
```diff
@@ -6240,6 +6274,21 @@ logpb_remove_archive_logs (THREAD_ENTRY * thread_p, const char *info_reason)
...
+	{
+
+	  flashback_first_pageid = flashback_min_log_pageid_to_keep ();
+
+	  /* NULL check for flashback_first_pageid is done in flashback_is_needed_to_keep_archive () */
+	  if (logpb_is_page_in_archive (flashback_first_pageid))
+	    {
+	      min_arv_required_for_flashback = logpb_get_archive_number (thread_p, flashback_first_pageid);
+	      min_arv_required_for_flashback--;
+	      last_deleted_arv_num = MIN (last_deleted_arv_num, min_arv_required_for_flashback);
```
hornetmj, line 6288: Please add a debugging log here too. And It would be nice to add a prefix that can be distinguished from the log left by the daemon.

hornetmj (approved):

src/communication/network_interface_sr.c
joohok: OK, I will handle the error as you mentiond

src/transaction/flashback.c
joohok: OK !

src/transaction/log_page_buffer.c
joohok: I'll put it on my to-do list and put it together when I refactor. thank you.

src/transaction/log_page_buffer.c
joohok: OK, I will add debug message here too.

src/base/error_code.h
```diff
@@ -1703,8 +1703,9 @@
 #define ER_FLASHBACK_EXCEED_MAX_NUM_TRAN_TO_SUMMARY -1335
 #define ER_FLASHBACK_SCHEMA_CHANGED                 -1336
 #define ER_FLASHBACK_LOG_NOT_EXIST                  -1337
+#define ER_FLASHBACK_TIMEOUT                        -1338
```
hornetmj, line 1709: How about using ER_FLASHBACK_EXCEED_THRESHOLD?

hornetmj (approved):

src/base/error_code.h
joohok: Initially, it was specified with that name, but it was changed to the current name according to the review.
Since people don't know which THRESHOLD was exceeded, the meaning of the error can become ambiguous.

so, I tried to change it to `EXCEED_TIME_THRESHOLD` but it means TIMEOUT..

PR #3386, opened 2022-02-26
[CBRD-24215] make temp filenames unique in CMS

kisoo-han:
http://jira.cubrid.org/browse/CBRD-24215

**Purpose**
- This is a defensive code in some sense of meaning.
- In CUBRID Manager Server, most temp file names looks like:
  cmd_res_{**pid**}, for example **_cmd_res_29470_**
- In the environment of multiple clients request massive CMS api at the same time, temp file name of this style may cause some problem because every clients use same file for intermediate result create/writing/reading/delete
- We must guarantee that every temp files for any client is **unique** to avoid such a contention, new temp file name style will be:
cmd_res_{**seconds**}\_{**usecs**}_{random number less than 997} , for example **_cmd_res_1645860068_17325_624_**
(it is number of seconds and microseconds since the Epoch)

**Implementation**
1. We will introduce two file name generation functions in src/cm_common/cm_utils.c
  `make_temp_filename ( )`, `make_temp_filepath ( )`
2. And change temp file name generation code in src/cm_common
3. After merging of this PR, we will replace temp file name generations code in **CUBRID/cubrid-manager-server** repo using above two functions.

**Remarks**
`gettimeofday` () is not defined in windows, so we got the code snippet from **base/porting.c**.

### DIFF
--- a/src/cm_common/cm_dep_tasks.c
+++ b/src/cm_common/cm_dep_tasks.c
@@ -1379,6 +1379,7 @@ user_login_sa (nvplist * out, char *_dbmt_error, char *dbname, char *dbuser, cha
 
   if (make_temp_filename (tmpfile, "DBMT_ems_sa.", PATH_MAX) < 0)
     {
+      strcpy (_dbmt_error, "make_temp_filename: filename creation error");
       goto login_err;
     }
 
--- a/src/cm_common/cm_dep_tasks.c
+++ b/src/cm_common/cm_dep_tasks.c
@@ -1377,9 +1377,11 @@ user_login_sa (nvplist * out, char *_dbmt_error, char *dbname, char *dbuser, cha
   char cmd_name[PATH_MAX];
   char *outmsg = NULL, *errmsg = NULL;
 
-  if (make_temp_filename (tmpfile, "DBMT_ems_sa.", PATH_MAX) < 0) {
-    goto login_err;
-  }
+  if (make_temp_filename (tmpfile, "DBMT_ems_sa.", PATH_MAX) < 0)
+    {
+      goto login_err;
+    }
+
 
   (void) envvar_tmpdir_file (outfile, PATH_MAX, tmpfile);
   if (snprintf (errfile, PATH_MAX - 1, "%s.err", outfile) < 0)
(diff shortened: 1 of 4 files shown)


src/cm_common/cm_dep_tasks.c
```diff
@@ -1372,12 +1372,17 @@ user_login_sa (nvplist * out, char *_dbmt_error, char *dbname, char *dbuser, cha
...
-  char tmpfile[100];
+  char tmpfile[PATH_MAX];
   const char *argv[10];
   char cmd_name[PATH_MAX];
   char *outmsg = NULL, *errmsg = NULL;
 
-  snprintf (tmpfile, sizeof (tmpfile) - 1, "%s%d", "DBMT_ems_sa.", getpid ());
+  if (make_temp_filename (tmpfile, "DBMT_ems_sa.", PATH_MAX) < 0)
+    {
+      goto login_err;
```
mhoh3963, line 1383: I think that you need to add error setting code before goto statement as following.
strcpy (_dbmt_error, "make temp file error");

src/cm_common/cm_utils.c
```diff
@@ -410,7 +413,10 @@ cmd_server_status (void)
     }
 #endif
 
-  snprintf (tmpfile, sizeof (tmpfile) - 1, "%s%d", "DBMT_util_001.", getpid ());
+  if (make_temp_filename (tmpfile, "DBMT_util_001.", sizeof (tmpfile)) < 0)
+    {
+      return res;
```
mhoh3963: You would better to return NULL rather than res, because res is not initialize.

mhoh3963 (approved):

src/cm_common/cm_utils.c
```diff
@@ -395,7 +398,7 @@ cmd_server_status (void)
   char out_file[PATH_MAX];
   char cmd_name[PATH_MAX];
   const char *argv[5];
-  char tmpfile[100];
+  char tmpfile[512];
```
airnet73: How about using tmpfile[PATH_MAX] instead of tmpfile[512]?

airnet73 (approved):

src/cm_common/cm_dep_tasks.c
kisoo-han: Thank you for your comment!
Comment accepted.

src/cm_common/cm_utils.c
kisoo-han: Thank you for your comment, comment accepted!

src/cm_common/cm_utils.c
kisoo-han: Thank you for your comment, it is changed to **PATH_MAX**.

PR #3387, opened 2022-02-28
[LETS-367] Active transaction server transaction table checkpointing daemon must execute upfront after instantiation

adrianbrab:
http://jira.cubrid.org/browse/LETS-367

Currently, the daemon is set to execute at fixed 60 seconds intervals.

Because of this, after instantiation, it needs 60 seconds to create a transaction table checkpoint and add it to the log.

Transaction table checkpoints are created to be consumed when one passive transaction server comes online in order to have a baseline transaction table from which it can then start to construct a consistent transaction table in order to provide consistent responses to queries.

As such, if a passive transaction server is started within the 60 seconds interval that it takes the active transaction server to create a first transaction table checkpoint, it will not obtain the information about such a table recorded in the from the active transaction server (via the page server).

To fix this the snapshot has to be created earlier as a baseline for PTS before the daemon will enter his normal cycle.

### DIFF
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -10729,12 +10729,12 @@ log_checkpoint_daemon_init ()
 void
 log_get_checkpoint_trantable_interval (bool & is_timed_wait, cubthread::delta_time & period)
 {
+  // will only be accessed
   static bool first_call = true;
   is_timed_wait = true;
   if (!BO_IS_SERVER_RESTARTED)
     {
       period = std::chrono::milliseconds (100);
-      return;
     }
   else
     {
@@ -10742,7 +10742,6 @@ log_get_checkpoint_trantable_interval (bool & is_timed_wait, cubthread::delta_ti
         {
           first_call = false;
           period = std::chrono::milliseconds (100);
-          return;
         }
       else
         {
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -1548,10 +1548,10 @@ log_initialize_internal (THREAD_ENTRY * thread_p, const char *db_fullname, const
 	}
     }
 
-  if (is_tran_server_with_remote_storage () && is_active_transaction_server ())
-    {
-      logpb_checkpoint_trantable (thread_p);
-    }
+//  if (is_tran_server_with_remote_storage () && is_active_transaction_server ())
+//    {
+//      logpb_checkpoint_trantable (thread_p);
+//    }
 
   LOG_CS_EXIT (thread_p);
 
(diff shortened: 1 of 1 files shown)


src/transaction/log_manager.c
```diff
@@ -1548,6 +1548,11 @@ log_initialize_internal (THREAD_ENTRY * thread_p, const char *db_fullname, const
 	}
     }
 
+  if (is_tran_server_with_remote_storage () && is_active_transaction_server ())
```
cristiarg: If `log_recovery_finish_transactions` is also executed, it also executes a `logpb_checkpoint_trantable`.

I would like us to investigate also a method that involves the daemon.
Implement a custom looper `cubthread::period_function` (see example: `log_get_log_group_commit_interval`) with the following logic:
- when called, as long as `!BO_IS_SERVER_RESTARTED` (this is also the condition used in `log_checkpoint_trantable_execute`), return a wait interval of 1 second - this will cause the daemon thread to call the daemon execute function every second - but it will be useless as the `!BO_IS_SERVER_RESTARTED` condition will also impede the actual creation of the checkpoint
- the first time the condition `BO_IS_SERVER_RESTARTED` is met, also return a wait interval of 1 second to cause the daemon to execute immediately - in this case, the interval can be even smaller
- the rest of the times, return unconditionally an interval of 60 seconds
In order to provide state, use a local static variable in the "period_function" implementation - it is safe to use the static variable as long as only one thread (even if not necessarily the same thread every time) will call it.

src/transaction/log_manager.c
```diff
@@ -10725,6 +10725,33 @@ log_checkpoint_daemon_init ()
...
 
+#if defined (SERVER_MODE)
+void
+log_get_checkpoint_trantable_interval (bool & is_timed_wait, cubthread::delta_time & period)
+{
+  static bool first_call = true;
+  is_timed_wait = true;
+  if (!BO_IS_SERVER_RESTARTED)
+    {
+      period = std::chrono::milliseconds (100);
```
cristiarg, line 10737: make these `constexpr`

src/transaction/log_manager.c
```diff
@@ -10725,6 +10725,33 @@ log_checkpoint_daemon_init ()
 }
 #endif /* SERVER_MODE */
 
+#if defined (SERVER_MODE)
+void
+log_get_checkpoint_trantable_interval (bool & is_timed_wait, cubthread::delta_time & period)
+{
+  static bool first_call = true;
```
cristiarg, line 10733: Add a comment this static is safe because only one thread accesses it at a certain moment.

src/transaction/log_manager.c
```diff
@@ -10725,6 +10725,33 @@ log_checkpoint_daemon_init ()
...
+      period = std::chrono::milliseconds (100);
+      return;
+    }
+  else
+    {
+      if (first_call)
+        {
+          first_call = false;
+          period = std::chrono::milliseconds (100);
+          return;
```
cristiarg: no need to `return` anywhere

cristiarg (approved):

PR #3388, opened 2022-02-28
[LETS-310] Handle page desynchronization during btree_range_scan OID selection

adrianbrab:
http://jira.cubrid.org/browse/LETS-310

`btree_range_scan_select_visible_oids` function selects all current key OID's. If the key has overflow OID's, the selection may be interrupted if the OID buffer is full.
Additional to this, on PTS, the fixing of the overflow OID page may return page desynchronization error. The case must be treated similarly with the full buffer issue.
The existing mechanism that interrupts the key processing and resumes on next `btree_range_scan` iteration relies on the fact that it won't interrupt immediately after processing the leaf record; it always interrupts after an overflow page.
On PTS, the page desync error may occur on the first overflow page too. Therefore, on PTS things need to be done a little differently to apply the same kind of mechanism:
- When a key is processed the first time, and it has an overflow page, fix the overflow page before processing the leaf record objects. If page desynchronization error occurs, interrupt `btree_range_scan_select_visible_oids`. If overflow page is fixed, keep it fixed.
- In the while that fixes overflow pages, fix overflow page only if it is not fixed already. On first iteration, the overflow page would have already been fixed, on subsequent iterations, not.
- In the while that fixed overflow pages, if page desynchronization occurs, set `bts->is_key_partially_processed` to `true` and save `bts->O_vpid` on the `last_visible_overflow`.
- `btree_range_scan` calls `btree_range_scan_handle_error` -> `btree_range_scan_handle_page_ahead_repl_error` after `btree_range_scan_select_visible_oids` and if `bts.repl_sync_lsa` is set, the scan iteration is interrupted.


src/storage/btree.c
```diff
@@ -25496,24 +25518,44 @@ btree_range_scan_select_visible_oids (THREAD_ENTRY * thread_p, BTREE_SCAN * bts)
...
+	}
+
       /* Start processing key with leaf record objects. */
       error_code =
 	btree_record_process_objects (thread_p, &bts->btid_int, BTREE_LEAF_NODE, &bts->key_record, bts->offset, &stop,
 				      btree_select_visible_object_for_range_scan, bts);
       if (error_code != NO_ERROR)
 	{
 	  ASSERT_ERROR ();
+	  // TODO: what about the possibly-fixed first overflow page?
```
cristiarg, line 25549: This TODO

src/storage/btree.c
```diff
@@ -25496,24 +25518,44 @@ btree_range_scan_select_visible_oids (THREAD_ENTRY * thread_p, BTREE_SCAN * bts)
...
 	  return error_code;
 	}
       if (stop || bts->end_one_iteration || bts->end_scan || bts->is_interrupted)
 	{
 	  /* Early out. */
+	  // TODO: what about the possibly-fixed first overflow page?
```
cristiarg: This TODO

src/storage/btree.c
```diff
@@ -25535,13 +25577,24 @@ btree_range_scan_select_visible_oids (THREAD_ENTRY * thread_p, BTREE_SCAN * bts)
...
+	}
       if (overflow_page == NULL)
 	{
 	  ASSERT_ERROR_AND_SET (error_code);
+	  if (er_errid () == ER_PAGE_AHEAD_OF_REPLICATION)
+	    {
+	      // save state; upon next entry/retry will directly fix the overflow page and reach here
+	      bts->O_vpid = overflow_vpid;
+	      bts->is_key_partially_processed = true;
+	      // TODO: these are useless as - due to handling in btree_range_scan_handle_error and
```
cristiarg, line 25595: This TODO

src/query/scan_manager.c
```diff
@@ -2379,11 +2379,26 @@ scan_get_index_oidset (THREAD_ENTRY * thread_p, SCAN_ID * s_id, DB_BIGINT * key_
 	  assert (er_errid () != NO_ERROR);
 	  goto exit_on_error;
 	}
-      ret = btree_range_scan (thread_p, bts, btree_range_scan_select_visible_oids);
-      if (ret != NO_ERROR)
+      bool retry = true;
```
cristiarg: @adrianbrab 
I do not think this handling is needed.
I have added generic handling in function `btree_range_scan` for all cases where `btree_range_scan_select_visible_oids` is used as key handling/processing function (`btree_range_scan_select_visible_oids` is only used in this context as argument for `btree_range_scan`).
Does not even compile.

razvanradulescu (approved):

PR #3390, opened 2022-03-02
[LETS-340] server-server connection errors

cristiarg:

There are two levels of error handling:
  1. a generic instance-wide error handling per innstance of `request_sync_send_queue`
    - used mostly for handling communication errors on the page server side
  3. a message-specific error handling for each dispatched message by means of `request_sync_send_queue::push` function
    - used for handling communication errors on the active/passive transaction server side
   
Implementation:
- changed to return error codes in `cubcomm::channel`
- changed to return error code (instead of ints) in `cubcomm::request_client_server`
- added error handling new param to `request_sync_client_server` ctor
- this is used in `page_server::connection_handler::connection_handler` to provide a handler which, when triggered will automatically disconnect the connection handler - effectively dropping the connection with the respective transaction server - in `page_server::connection_handler::abnormal_tran_server_disconnect` - this will happen when a transaction server goes suddenly offline
- error handling happens in function `request_sync_send_queue::send_queue` where the message-specific error handling is first checked and invoked, then the generic instance-wide one
- `request_sync_client_server::send_recv` is the function that is used on the transaction server side; it will invoke the communication layer with its own custom error handler - `error_handler_ftor` - and report to the callers via `css_error_code`
- via the function `tran_server::send_receive` this ends up being used to retrieve heap pages (`pgbuf_request_data_page_from_page_server`), log pages (`logpb_request_log_page_from_page_server`), passive transaction server log boot info (`send_and_receive_log_boot_info`), command page server to stop dispatching log prior info to the passive transaction server (`passive_tran_server::send_and_receive_stop_log_prior_dispatch`) which are all handled according to needs
- NOTE: some calls to communicate message to the other side have not been enhanced with error handling - `request_sync_client_server::push`, `request_sync_client_server::respond` - and will be done if&when needed
- adapted unit tests

Other:
- deleted copy/move ctors/operator here&there
- removed some unneeded includes in `page_buffer.c`

Caveat:
- a mistake has been identified in the implementation of error handing in `request_sync_client_server::send_recv`: because the function `m_queue->push` only dispatches the request for async execution, the error handling must be implemented asynchronously as well; because of the complexity of the implementation, this will be done in an separate patch


cristiarg:
Reviewers: please note the caveat in the description

cristiarg:
This PR is super-seeded by a few smaller PR's to the same end effect.

PR #3392, opened 2022-03-03
[CBRD-24212] When executing a phrase that performs multiple sentences at once, the hint is grouped into one.

ctshim:
http://jira.cubrid.org/browse/CBRD-24212

Fix to output hint information per SQL statement.

Schema:
```
create table tbl (id varchar);
create index i_tbl on tbl(id);
insert into tbl values ('T'), ('X');

create table tbl_a (a varchar, aaa varchar);
create index i_tbl_a on tbl_a(a);
insert into tbl_a values ('TA', 'Y');
```

Query:
```
SELECT  a as TEST from tbl_a
UNION all
select /*++ INDEX_LS(tbl_a) */ id as TEST from tbl
UNION all
select /*++ INDEX_LS() */ id as TEST from tbl;
```

Hint information is displayed after all plan information for a single statement is output.
```
seg[0]: [0]
seg[1]: a[0] (f)
Join graph nodes:
node[0]: tbl_a tbl_a(0/1) (loc 0)

Query plan:

sscan
    class: tbl_a node[0]
    cost:  1 card 0

Query stmt:

select tbl_a.a from tbl_a tbl_a

seg[0]: [0]
seg[1]: id[0] (f)
Join graph nodes:
node[0]: tbl tbl(0/1) (loc 0)

Query plan:

sscan
    class: tbl node[0]
    cost:  1 card 0

Query stmt:

select tbl.id from tbl tbl

seg[0]: [0]
seg[1]: id[0] (f)
Join graph nodes:
node[0]: tbl tbl(0/1) (loc 0)

Query plan:

sscan
    class: tbl node[0]
    cost:  1 card 0

Query stmt:

select tbl.id from tbl tbl


Query hints:
    Input)  INDEX_LS(tbl_a) 
    Hit)  INDEX_LS([tbl_a])
    Input)  INDEX_LS() 
    Hit)  INDEX_LS( )

```


shparkcubrid (approved):

src/parser/scanner_support.c
```diff
@@ -145,23 +215,111 @@ set_plan_include_hint (bool is_include)
...
+	free (ptr);
+      }
+
+    ptr = buf;
+    size = sizeof (buf);
+    buf[0] = 0x00;
+  }
+};
+
+static struct s_string tempory_hint_string;
```
hgryoo: How about using `std::string` instead of defining this new struct? `std::string` seems to have the same functionality as the `s_string`

hgryoo, line 321: How about using `std::string` instead of defining this new struct? `std::string` seems to have the same functionality as the `s_string`

src/parser/scanner_support.c
```diff
@@ -88,6 +107,52 @@ struct st_hint_msg
...
+
+	prev = m_hint_ptr;
+      }
+
+    m_hint_ptr = (STMT_HINT *) malloc (sizeof (STMT_HINT));
+    assert (m_hint_ptr != NULL);
+
+    m_hint_ptr->stmt_no = no;
+    m_hint_ptr->offset = m_used;
+    m_hint_ptr->next = prev->next;
```
hyunikn, line 155: Is prev->next is always NULL at this line? If so, NULL seems to be more clear.

src/parser/scanner_support.c
ctshim: OK, I will change the code to use std::string.

src/parser/scanner_support.c
ctshim: You are right.
It is a sign of considering the sorted order using stmt_no.
There is no need to modify it.

src/parser/xasl_generation.c
```diff
@@ -21798,6 +21799,11 @@ parser_generate_xasl (PARSER_CONTEXT * parser, PT_NODE * node)
   node->next = NULL;
   parser->dbval_cnt = 0;
 
+  {
```
beyondykk9, line 21802: Is there any reason this code block is enclosed in braces?

src/parser/scanner_support.c
```diff
@@ -145,23 +219,85 @@ set_plan_include_hint (bool is_include)
...
 
-      s_hint_msg.msg_ptr[s_hint_msg.m_used] = '\0';
-      fprintf (output, "%s", s_hint_msg.msg_ptr);
+      if (hint_ptr)
+	{
+	  fputs ("\nQuery hints:\n", output);
+
+	  s_hint_msg.msg_ptr[s_hint_msg.m_used] = '\0';
+	  fprintf (output, "%s", s_hint_msg.msg_ptr + hint_ptr->offset);
+	  hint_ptr->stmt_no = -2;	/* change to no more use */
```
beyondykk9, line 242: what does mean "-2" ?
if it needs some meaning, it would be better to use a macro defines a special meaning.

src/parser/scanner_support.c
```diff
@@ -133,8 +198,17 @@ struct st_hint_msg
     check_buffer (snprintf (NULL, 0, fmt, str1, str2));
     m_used += sprintf (msg_ptr + m_used, fmt, str1, str2);
   }
+  void addendstring ()
+  {
+    check_buffer (2);
```
beyondykk9: what does mean 2?

src/parser/scanner_support.c
```diff
@@ -133,8 +198,17 @@ struct st_hint_msg
     check_buffer (snprintf (NULL, 0, fmt, str1, str2));
     m_used += sprintf (msg_ptr + m_used, fmt, str1, str2);
   }
+  void addendstring ()
```
beyondykk9:
```suggestion
  void add_end_string ()
```

src/parser/scanner_support.c
```diff
@@ -88,6 +107,52 @@ struct st_hint_msg
...
+      {
+	STMT_HINT *t = m_hint_ptr;
+	m_hint_ptr = t->next;
+	free (t);
+      }
+    memset (&m_hint_head, 0x00, sizeof (STMT_HINT));
+    m_hint_head.stmt_no = -1;
+
+    m_stmt_no = -1;
+    m_select_cnt = -1;
```
beyondykk9, line 123: why m_select_cnt initial value is not 0 ?

src/parser/scanner_support.c
hyunikn: Aha! I got it.

src/parser/scanner_support.c
```diff
@@ -75,6 +87,13 @@ struct st_hint_msg
     m_used = 0;
     msg_ptr = msg_buf;
     msg_buf[0] = 0x00;
+
+    m_hint_ptr = NULL;
+    memset (&m_hint_head, 0x00, sizeof (STMT_HINT));
+    m_hint_head.stmt_no = -1;
```
hgryoo, line 94: How about implementing the default constructor for STMT_HINT?

src/parser/scanner_support.c
```diff
@@ -59,6 +59,13 @@ extern int yycolumn_end;
 
 
 /* matching hint information to output*/
+typedef struct _stmt_hint STMT_HINT;
+struct _stmt_hint
```
hgryoo: The two structs `_stmt_hint` and `st_hint_msg` seem to be related, but why is there no `st_` prefix like `st_hint_msg`?

hyunikn (approved):

src/parser/xasl_generation.c
ctshim: Declare and call extern functions that are only used here. I was trying to wrap it up with blocks to make it more visible.

src/parser/scanner_support.c
ctshim: I will edit according to your comments. Thank you

src/parser/scanner_support.c
ctshim: Check the space required to append a string to msg_ptr. Make room for '\0'.

src/parser/scanner_support.c
ctshim: Setting it to -2 (negative number) means the information is no longer used.
I used -2 to distinguish it from the initial value of -1.

src/parser/scanner_support.c
ctshim: When used, set it to 0 through the calc_hint_statement_count() function.
It is set to -1 in the initialization state because it can also be identified whether the calc_hint_statement_count() function is called through the value of this variable.

src/parser/scanner_support.c
ctshim: I don't think there's any need to introduce a constructor as it's a simple code.

src/parser/scanner_support.c
ctshim: Like struct _xxx, starting with a _ is a common practice when the name is not going to be used in the future. Instead, use the typedef name. However, according to your comments, we will make corrections for consistency.

beyondykk9 (approved):

hgryoo:
`add_string` functions could be implemented by using _variable arguments_
```
#include <stdarg.h>

void add_string (const char *fmt, ...)
{
  va_list ap;
  va_start (ap, fmt ? fmt : "%s");
  int count = vsnprintf (NULL, 0, fmt, ap);
  if (count >= 0)
  {
    check_buffer (count + 1);
    m_used += vsnprintf (msg_ptr + m_used, count + 1, fmt, ap);
  }
  va_end (ap);
}

void end_string ()
{
  add_string (NULL, '\0');
}
```

beyondykk9 (approved):

ctshim:
@hgryoo
According to hgryoo's comment, I changed the code using variable arg.
However, some of the provided codes have been modified and applied.

void add_string (const char *fmt, ...)
  {
    va_list ap;

    va_start (ap, fmt);
    int count = vsnprintf (NULL, 0, fmt, ap);
    **va_end (ap);**

    if (count > 0)
      {
	check_buffer (count);

        va_start (ap, fmt);
	m_used += vsprintf (msg_ptr + m_used, fmt, ap);
	va_end (ap);
      }
  }

  void add_end_string ()
  {
    add_string ("%c", '\0');
  }

src/parser/scanner_support.c
```diff
@@ -59,6 +60,13 @@ extern int yycolumn_end;
 
 
 /* matching hint information to output*/
+typedef struct st_stmt_hint STMT_HINT;
```
youngjinj, line 63: Isn't there a convention that struct type names are capitalized for struct names?

ctshim: The company code convention rules are not clear.
However, in the previous example, it was used as follows.

```
typedef struct parser_node PT_NODE;
typedef struct pt_alter_info PT_ALTER_INFO;
```

youngjinj: Yes, I understand.

youngjinj (approved):

shparkcubrid (approved):

hgryoo (approved):

hyunikn (approved):

PR #3396, opened 2022-03-03
[CBRD-24217] Error in data communication protocol between Driver and CAS

airnet73:
http://jira.cubrid.org/browse/CBRD-24217

Purpose
In the FETCH phase of the data communication protocol between the driver and the CAS, the data size and data are put into the network buffer. At this time, the data structure according to the data size and data type must match.
In case of SQL_TINYINT type, data size should be 2 bytes, and data type should be short.

Implementation
N/A

Remarks
N/A

### DIFF
--- a/src/broker/cas_execute.c
+++ b/src/broker/cas_execute.c
@@ -1119,8 +1119,9 @@ ux_cgw_prepare (char *sql_stmt, int flag, char auto_commit_mode, T_NET_BUF * net
 	  goto prepare_error;
 	}
 
-      cgw_set_stmt_attr (srv_handle->cgw_handle->hstmt, SQL_ATTR_CURSOR_TYPE, (SQLPOINTER) SQL_CURSOR_STATIC,
-			 SQL_IS_INTEGER);
+      err_code =
+	cgw_set_stmt_attr (srv_handle->cgw_handle->hstmt, SQL_ATTR_CURSOR_TYPE, (SQLPOINTER) SQL_CURSOR_STATIC,
+			   SQL_IS_INTEGER);
       if (err_code < 0)
 	{
 	  err_code = ERROR_INFO_SET (db_error_code (), DBMS_ERROR_INDICATOR);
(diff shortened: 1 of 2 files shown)


src/broker/cas_execute.c
```diff
@@ -1118,6 +1118,14 @@ ux_cgw_prepare (char *sql_stmt, int flag, char auto_commit_mode, T_NET_BUF * net
 	  err_code = ERROR_INFO_SET (db_error_code (), DBMS_ERROR_INDICATOR);
 	  goto prepare_error;
 	}
+
+      cgw_set_stmt_attr (srv_handle->cgw_handle->hstmt, SQL_ATTR_CURSOR_TYPE, (SQLPOINTER) SQL_CURSOR_STATIC,
```
beyondykk9: hstmt can be NULL. it needs to check if it's NULL ?

airnet73: It's checking for null in cgw_set_stmt_attr().

beyondykk9 (approved):

src/broker/cas_execute.c
```diff
@@ -1118,6 +1118,14 @@ ux_cgw_prepare (char *sql_stmt, int flag, char auto_commit_mode, T_NET_BUF * net
...
+			 SQL_IS_INTEGER);
+      if (err_code < 0)
```
beyondykk9, line 1125: this code block looks not necessary because the error handling is done by the prior code block.

airnet73: fixed it

kisoo-han (approved):

src/broker/cas_execute.c
```diff
@@ -3228,6 +3237,7 @@ ux_cursor_close (T_SRV_HANDLE * srv_handle)
     }
 #if defined(CAS_FOR_CGW)
   cgw_cursor_close (srv_handle->cgw_handle->hstmt);
```
beyondykk9: no need to check error?

beyondykk9 (approved):

src/broker/cas_execute.c
airnet73: Thanks for your comment.
Added error check code.
if (cgw_cursor_close (srv_handle->cgw_handle->hstmt) > -1)
    {
      srv_handle->cgw_handle->hstmt = NULL;
    }

PR #3400, opened 2022-03-04
[LETS-340] server-server communication error handling for page server

cristiarg:

An error handler is added for the entire duration of the `request_sync_send_queue` lifetime.
   
Implementation:
- added error handling new param to `request_sync_client_server` ctor which forwards the error handler functor to `request_sync_send_queue` ctor
- this is used in `page_server::connection_handler::connection_handler` ctor to provide a handler
- the handler `page_server::connection_handler::abnormal_tran_server_disconnect` will be called when a message cannot be sent in `request_sync_send_queue::send_queue` and will "ask" the parent instance - `page_server` - to delete "this" connection handler - effectively dropping the connection with the respective transaction server
- the "abort_further_processing" is a flag which the error handler can use to ask that no further communication needs to be handled by the queue
- adapted unit tests

Other:
- deleted copy/move ctors/operator here&there
- removed some unneeded includes in `page_buffer.c`
- const-ed some functions where possible


src/server/page_server.cpp
```diff
@@ -170,6 +178,51 @@ page_server::connection_handler::receive_disconnect_request (tran_server_conn_t:
...
+			(int)error_code);
+
+	  {
+	    std::lock_guard<std::mutex> lockg { m_prior_sender_sink_removal_mtx };
+
+	    log_Gl.m_prior_sender.remove_sink (m_prior_sender_sink_hook_func);
+	    m_prior_sender_sink_hook_func = nullptr;
+	  }
+
+	  std::thread disconnect_thread { &page_server::disconnect_tran_server, std::ref (m_ps), this };
```
razvanradulescu: This codes seems identical to the one in else {

razvanradulescu: Nit : do we really need another thread just to close the connection ?

razvanradulescu (approved):

src/server/page_server.cpp
cristiarg: Fixed

cristiarg: The mechanism existed before.
The parent class will destroy the child (invoking class).
I think it should work provided that no `this` members are accessed after call to destroy `this`.

cristiarg: @razvanradulescu 
As per @alexstanro review, the mechanism will be changed to be more robust by introducing a permanent worker thread with a lifetime parallel to that of `page_server`'s instance.

alexstanro (approved):

PR #3402, opened 2022-03-05
[CBRD-24226] 10.2 cm_common, pull defensive code in develop

kisoo-han:
http://jira.cubrid.org/browse/CBRD-24226

**Purpose**
To pull some defensive code in **_develop_** into 10.2 branch, for example do **fclose ()** before return if subject **fd** is active.

Implementation
N/A

**Remarks**
N/A

### DIFF
--- a/src/cm_common/cm_utils.c
+++ b/src/cm_common/cm_utils.c
@@ -520,11 +520,16 @@ read_server_status_output (T_SERVER_STATUS_RESULT * res, char *out_file)
       if (num_info > num_alloc)
 	{
 	  num_alloc += 5;
-	  info = (T_SERVER_STATUS_INFO *) realloc (info, sizeof (T_SERVER_STATUS_INFO) * num_alloc);
-	  if (info == NULL)
+	  T_SERVER_STATUS_INFO *const new_info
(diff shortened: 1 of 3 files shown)


src/cm_common/cm_utils.c
```diff
@@ -520,12 +520,17 @@ read_server_status_output (T_SERVER_STATUS_RESULT * res, char *out_file)
       if (num_info > num_alloc)
 	{
 	  num_alloc += 5;
-	  info = (T_SERVER_STATUS_INFO *) realloc (info, sizeof (T_SERVER_STATUS_INFO) * num_alloc);
-	  if (info == NULL)
+	  T_SERVER_STATUS_INFO *const new_info
+	    = (T_SERVER_STATUS_INFO *) realloc (info, sizeof (T_SERVER_STATUS_INFO) * num_alloc);
+	  if (new_info == NULL)
 	    {
 	      fclose (fp);
```
airnet73, line 528: If new_info is NULL, info also needs to free memory.

airnet73 (approved):

src/cm_common/cm_utils.c
kisoo-han: Your comment is accepted. Thank you.

PR #3407, opened 2022-03-06
[LETS-340] async disconnect handler in page server

cristiarg:

Implement a `disconnect_handler` in `page_server` that reliably makes sure to disconnect (aka: destroy) a `connection_handler`. Previous method using detached threads was not deterministic and could cause indefinite hangs.
- in `page_server::connection_handler::connection_handler` dispatch `this` to be deleted to the parent's disconnect handler
- same in `page_server::connection_handler::abnormal_tran_server_disconnect`
- `page_server::disconnect_tran_server_async` just moves the connect handler's pointer to the disconnect handler

Related changes:
- in `xboot_shutdown_server` the page server is explicitly asked to terminate all connections to transaction servers
- in `cublog::replicator` introduced an internal flag to enforce a invariant relation: after the replicator is asked to `wait_replication_finish_during_shutdown`, no further log records can be processed

Other:
- deleted some unneeded ctors

### DIFF
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -266,11 +266,14 @@ page_server::disconnect_handler::disconnect_handler ()
 
 page_server::disconnect_handler::~disconnect_handler ()
 {
-  assert (m_terminate);
-  if (m_terminate && m_thread.joinable ())
+  if (m_terminate.load () && m_thread.joinable ())
     {
       m_queue_cv.notify_one ();
       m_thread.join ();
+    }
+  else
+    {
+      assert (m_terminate.load ());
     }
 
   assert (m_disconnect_queue.empty ());
@@ -279,7 +282,7 @@ page_server::disconnect_handler::~disconnect_handler ()
 void
 page_server::disconnect_handler::disconnect (connection_handler_uptr_t &&handler)
 {
-  if (!m_terminate)
+  if (!m_terminate.load ())
     {
       std::lock_guard<std::mutex> lockg { m_queue_mtx };
       m_disconnect_queue.emplace (std::move (handler));
@@ -295,7 +298,7 @@ page_server::disconnect_handler::disconnect (connection_handler_uptr_t &&handler
 void
 page_server::disconnect_handler::terminate ()
 {
-  m_terminate = true;
+  m_terminate.store (true);
   m_queue_cv.notify_one ();
 }
 
@@ -303,13 +306,13 @@ void
 page_server::disconnect_handler::disconnect_loop ()
 {
   std::queue<connection_handler_uptr_t> disconnect_work_buffer;
-  while (!m_terminate)
+  while (!m_terminate.load ())
     {
       {
 	std::unique_lock<std::mutex> ulock { m_queue_mtx };
 	m_queue_cv.wait (ulock, [this]
 	{
-	  return !m_disconnect_queue.empty () || m_terminate;
+	  return !m_disconnect_queue.empty () || m_terminate.load ();
 	});
 
 	m_disconnect_queue.swap (disconnect_work_buffer);
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -312,10 +312,11 @@ page_server::async_disconnect_handler::disconnect_loop ()
     {
       {
 	std::unique_lock<std::mutex> ulock { m_queue_mtx };
-	while (!m_queue_cv.wait_for (ulock, one_second, [this]
-	{
-	  return !m_disconnect_queue.empty () || m_terminate.load ();
-	  }));
+	if (!m_queue_cv.wait_for (ulock, one_second,
+				  [this] { return !m_disconnect_queue.empty () || m_terminate.load (); }))
+	  {
+	    continue;
+	  }
 
 	m_disconnect_queue.swap (disconnect_work_buffer);
       }
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -305,15 +305,17 @@ page_server::async_disconnect_handler::terminate ()
 void
 page_server::async_disconnect_handler::disconnect_loop ()
 {
+  constexpr std::chrono::seconds one_second { 1 };
+
   std::queue<connection_handler_uptr_t> disconnect_work_buffer;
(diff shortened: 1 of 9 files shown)


src/server/page_server.hpp
```diff
@@ -110,18 +111,51 @@ class page_server
...
+	disconnect_handler &operator = (disconnect_handler &&) = delete;
+
+	void disconnect (connection_handler_uptr_t &&handler);
+	void terminate ();
+
+      private:
+	void disconnect_loop ();
+
+      private:
+	bool m_terminate = false;
```
alexstanro: Shouldn't be an std::atomic<bool> m_terminate here?

src/server/page_server.cpp
```diff
@@ -244,6 +258,72 @@ page_server::connection_handler::prior_sender_sink_hook (std::string &&message)
...
+
+void
+page_server::disconnect_handler::disconnect_loop ()
+{
+  std::queue<connection_handler_uptr_t> disconnect_work_buffer;
+  while (!m_terminate)
+    {
+      {
+	std::unique_lock<std::mutex> ulock { m_queue_mtx };
+	m_queue_cv.wait (ulock, [this]
```
alexstanro: How about waiting with timeout?

src/server/page_server.hpp
cristiarg: True. Done.

src/server/page_server.cpp
cristiarg: What is the point of waiting with a timeout?
By the nature of the application logic, the thread will have very little work to do, so spinning it uselessly makes no sense.
Whenever
1. work is pushed
2. terminate is set

the condition variable is notified.

alexstanro: Is not about spinning. It is about calling wait_for instead wait. It is used for safety, to avoid infinite blocking

src/server/page_server.cpp
```diff
@@ -244,6 +258,75 @@ page_server::connection_handler::prior_sender_sink_hook (std::string &&message)
...
+  assert (m_disconnect_queue.empty ());
+}
+
+void
+page_server::disconnect_handler::disconnect (connection_handler_uptr_t &&handler)
+{
+  if (!m_terminate.load ())
+    {
+      std::lock_guard<std::mutex> lockg { m_queue_mtx };
+      m_disconnect_queue.emplace (std::move (handler));
```
alexstanro, line 287: You can release the mutex before notify

src/server/page_server.cpp
```diff
@@ -244,6 +258,75 @@ page_server::connection_handler::prior_sender_sink_hook (std::string &&message)
...
+    {
+      // cannot ask for disconnect after termination
+      assert (false);
+    }
+}
+
+void
+page_server::disconnect_handler::terminate ()
+{
+  m_terminate.store (true);
```
alexstanro, line 301: Do we really need this function? How about moving "m_terminate.store (true)" into destructor and get rid og terminate()?

src/server/page_server.cpp
```diff
@@ -244,6 +258,75 @@ page_server::connection_handler::prior_sender_sink_hook (std::string &&message)
...
+  else
+    {
+      assert (m_terminate.load ());
+    }
+
+  assert (m_disconnect_queue.empty ());
+}
+
+void
+page_server::disconnect_handler::disconnect (connection_handler_uptr_t &&handler)
```
alexstanro: How about renaming it as async_disconnect?

src/server/page_server.cpp
```diff
@@ -244,6 +258,75 @@ page_server::connection_handler::prior_sender_sink_hook (std::string &&message)
...
+  : m_terminate { false }
+{
+  m_thread = std::thread (&page_server::disconnect_handler::disconnect_loop, std::ref (*this));
+}
+
+page_server::disconnect_handler::~disconnect_handler ()
+{
+  if (m_terminate.load () && m_thread.joinable ())
+    {
+      m_queue_cv.notify_one ();
```
alexstanro: Do we really need this notify? It was already called when terminate was set.

src/server/page_server.cpp
```diff
@@ -277,18 +364,27 @@ page_server::set_passive_tran_server_connection (cubcomm::channel &&chn)
...
 
 void
-page_server::disconnect_tran_server (connection_handler *conn)
+page_server::disconnect_tran_server_async (connection_handler *conn)
 {
   assert (conn != nullptr);
   if (conn == m_active_tran_server_conn.get ())
     {
-      disconnect_active_tran_server ();
+      m_disconnect_handler.disconnect (std::move (m_active_tran_server_conn));
```
alexstanro: Adding/removing m_active_tran_server_conn, m_passive_tran_server_conn should be protected.

src/server/page_server.cpp
cristiarg: Maybe not. It is covered indeed by the terminate notify an then by the assert in `disconnect`.

src/server/page_server.cpp
cristiarg: Correct. It was a mistake.

src/server/page_server.cpp
cristiarg: Renamed

src/server/page_server.cpp
cristiarg: Indeed, this discussion came in the past, but we thought it is not yet "time" to consider it because the probability for collision was considered to be very low. However it should be taken up.

For `m_active_tran_server_conn`:
- adding: is only done once, so it does not need to be
- deleting it sync or async: indeed it needs to be
- using it (eg: in `push_request_to_active_tran_server`) is only done in one place, but usage is not synchronized with deletion
Same goes for each instance of `m_passive_tran_server_conn`.

I think it is better to take this as a separate issue - I've added http://jira.cubrid.org/browse/LETS-408 .

alexstanro: I agree to handle it in a separate issue.

src/server/page_server.cpp
cristiarg: Done

alexstanro (approved):

PR #3419, opened 2022-03-10
[CBRD-24229] Reserve a flag bit in class record to know whether supplemental log is appended

joohok:
http://jira.cubrid.org/browse/CBRD-24229

Purpose

The flag bit is reserved to know if the supplemental log is appended for a specific class.
One of the remaining bits of SM_CLASS_FLAG is used for the future use.


hornetmj (approved):

wrlawodms:
1. When will this column be set and unset?
2. Why don't you add it to the db_class view?
3. When the schema of the class record is changed, it is not compatible with the previous CUBRID version. Is it OK?
4. the column is just a flag. So, I think it should use a bit of the flag column in the root class.

eido5 (approved):
There looks no typos.

src/storage/storage_common.h
```diff
@@ -1089,6 +1089,19 @@ extern const int SM_MAX_STRING_LENGTH;
...
+ *    as powers of two because they are stored packed in a single integer.
+ */
+typedef enum
+{
+  SM_CLASSFLAG_SYSTEM = 1,	/* a system defined class */
+  SM_CLASSFLAG_WITHCHECKOPTION = 2,	/* a view with check option */
+  SM_CLASSFLAG_LOCALCHECKOPTION = 4,	/* view w/local check option */
+  SM_CLASSFLAG_REUSE_OID = 8,	/* the class can reuse OIDs */
+  SM_CLASSFLAG_SUPPLEMENTAL_LOG = 16	/* the supplemental log is appended for this class */
+} SM_CLASS_FLAG;
```
wrlawodms: Could you tell me the reason to move it?

wrlawodms (commented):
Have you considered the way other flags in SM_CLASS_FLAG do? They aren't added as columns in _db_class, but just being integrated into the `is_system_class` column (`flags` in code). And when they are inquired from the db_class view, they are shown as each flag. See `boot_define_view_class()`. I think we should follow that rule. What do you think?

joohok:
> Have you considered the way other flags in SM_CLASS_FLAG do? They aren't added as columns in _db_class, but just being integrated into the `is_system_class` column (`flags` in code). And when they are inquired from the db_class view, they are shown as each flag. See `boot_define_view_class()`. I think we should follow that rule. What do you think?

Yes, I saw that usage. 
all the SM_CLASS_FLAGs are integrated into the 'is_system_class', 
and only if REUSE_OID flag is required to be shown to user, then it is extracted from is_system_class. 

But I wonder if legacy is correct.. 
because, `is_system_class` is literally just a boolean value, not a set of flags. 
and it may provides poor readability when we need to write a query for db_class for getting specific flag (`boot_define_view_class()`).

The case of extracting the flag bit from `is_system_class` and showing it through db_class is only in the case of `reuse_oid`. 

In setting other system class(_db_attribute) (`catcl_get_or_value_from_attribute()`), it extract `is_nullable` from the flags. 

I think adding a new column to _db_class is better. 
The system class containing class information is _db_class, 
and among the information of _db_class, the aggregate of information that can be seen by all users is considered to be db_class. 
Therefore, class information should be added to _db_class, and I think that it is possible to selectively decide whether to add information to db_class. 
In the case of `reuse_oid`.., it seems that it was not added to _db_class due to migration.

wrlawodms:
> But I wonder if legacy is correct..
> because, is_system_class is literally just a boolean value, not a set of flags.
> and it may provides poor readability when we need to write a query for db_class for getting specific flag (reuse_oid).

I totally agree with you that it has poor readability, and it should be refactored.

> The case of extracting the flag bit from is_system_class and showing it through db_class is only in the case of reuse_oid.

Yes, but other flags are also stored in the `is_system_class` column and used internally. 

> I think adding a new column to _db_class is better.
> The system class containing class information is _db_class,
> and among the information of _db_class, the aggregate of information that can be seen by all users is considered to be db_class.
> Therefore, class information should be added to _db_class, and I think that it is possible to selectively decide whether to add information to db_class.
> In the case of reuse_oid.., it seems that it was not added to _db_class due to migration.

In my opinion, what has to be fixed is the name of the column regarding the followings:
1. `flags` are enough to represent all the flags and a new column is redundant.
2. _db_class is just a class representing classes shown to DBA and stored in disk. 

Let's have a discussion at the next meeting with @hornetmj.

src/storage/storage_common.h
joohok: it is reverted

wrlawodms (approved):
You should change the PR title regarding the change.

joohok:
> You should change the PR title regarding the change.

OK, I will change the name of PR and issue.

joohok:
test_sql is failed due to test case modification in (https://github.com/CUBRID/cubrid-testcases/pull/1105).

hornetmj (approved):

eido5 (approved):

joohok:
@mhoh3963  
Merge this PR please, the test is failed not from this PR.
tests are failed due to changes of test cases for user schema.

PR #3422, opened 2022-03-12
[LETS-340] server-server connection error handling infrastructure for transaction server  

cristiarg:

Infrastructure to allow message-specific error handling.
Extend error handling with a message-specific error handling for each dispatched message by means of `request_sync_send_queue::push` function.
Used for handling communication errors on the active/passive transaction server side.
   
Implementation:
- extended `request_sync_send_queue::queue_item_type` with an error handler, which is first checked and handled in the `push` function in case sending fails (as a remark, network communication could be detected immediately only if the application protocol is a request-ACK type of; in this case the application protocol is more has a loose request-response structure in the sense that the response is waited for but not immediately, rather in an async optimistic-like manner; this has been chosen for performance reasons; as such, should there be a failure of the network connection, it will be detected reactively rather than pro-actively - eg. when the internal/socket's send bufffer would have filled)
- class `response_broker` is a class that facilitates simple send-and-wait-for-response roundtrips; extended with the ability to register either response or error response; the error type is a new template argument `T_ERROR` which can be registered via `register_error` and will be returned together with the response as a tuple in `response_broker::bucket::get_response`
- `response_broker` will take additional `T_ERROR` ctor arguments that tells it which is the no-error (idle value) and also a generic error for the error response value - this is needed because the used typename used for `T_ERROR` is an enum - `enum css_error_code`
- in case the response broker needs to terminate without having receive either an actual response or an error, the generic error result will be returned
- add an error handler to `request_sync_client_server::send_recv` which uses the infrastructure in `response_broker` to check whether sending succeeded or not
- `tran_server::send_receive` sets an error and returns error code which is then further handled upstream
- adapted unit tests

Other:
- more granular locking when sending messages over the network in `request_sync_send_queue::push` with the purpose of avoiding double mutex locking as much as possible added a stress unit test to ensure correctness of this change - "Test out-of-order request_queue_autosend"

### DIFF
--- a/src/communication/request_sync_client_server.hpp
+++ b/src/communication/request_sync_client_server.hpp
@@ -76,7 +76,8 @@ namespace cubcomm
       std::string get_underlying_channel_id () const;
 
       void push (T_OUTGOING_MSG_ID a_outgoing_message_id, T_PAYLOAD &&a_payload);
-      void send_recv (T_OUTGOING_MSG_ID a_outgoing_message_id, T_PAYLOAD &&a_request_payload, T_PAYLOAD &a_response_payload);
+      css_error_code send_recv (T_OUTGOING_MSG_ID a_outgoing_message_id, T_PAYLOAD &&a_request_payload,
+				T_PAYLOAD &a_response_payload);
       void respond (sequenced_payload &&seq_payload);
 
     private:
@@ -92,10 +93,10 @@ namespace cubcomm
       std::unique_ptr<request_sync_send_queue_t> m_queue;
       std::unique_ptr<request_queue_autosend_t> m_queue_autosend;
 
-      T_OUTGOING_MSG_ID m_outgoing_response_msgid;
-      T_INCOMING_MSG_ID m_incoming_response_msgid;
+      const T_OUTGOING_MSG_ID m_outgoing_response_msgid;
+      const T_INCOMING_MSG_ID m_incoming_response_msgid;
       response_sequence_number_generator m_rsn_generator;
-      response_broker<T_PAYLOAD> m_response_broker;
+      response_broker<T_PAYLOAD, css_error_code> m_response_broker;
   };
 
   template <typename T_OUTGOING_MSG_ID, typename T_INCOMING_MSG_ID, typename T_PAYLOAD>
@@ -147,7 +148,7 @@ namespace cubcomm
   , m_queue_autosend { new request_queue_autosend_t (*m_queue) }
   , m_outgoing_response_msgid { a_outgoing_response_msgid }
   , m_incoming_response_msgid { a_incoming_response_msgid }
-  , m_response_broker { response_partition_count }
+  , m_response_broker { response_partition_count, NO_ERRORS }
   {
     assert (a_incoming_request_handlers.size () > 0);
     for (const auto &pair: a_incoming_request_handlers)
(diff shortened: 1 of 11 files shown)


src/communication/response_broker.hpp
```diff
@@ -82,76 +105,112 @@ namespace cubcomm
...
-    auto condvar_pred = [this, &payload, a_rsn] ()
+    auto condvar_pred = [this, &payload_or_error, a_rsn] ()
     {
       auto it = m_response_payloads.find (a_rsn);
       if (it == m_response_payloads.end ())
 	{
 	  return false;
 	}
-      payload = std::move ((*it).second);
+      payload_or_error = { std::move ((*it).second.m_payload), std::move ((*it).second.m_error) };
```
alexstanro: I guess that this code may be moved after m_condvar.wait

src/communication/response_broker.hpp
```diff
@@ -82,76 +105,112 @@ namespace cubcomm
...
+  template <typename T_PAYLOAD, typename T_ERROR>
+  std::tuple<T_PAYLOAD, T_ERROR>
+  response_broker<T_PAYLOAD, T_ERROR>::bucket::get_response (response_sequence_number a_rsn)
   {
     std::unique_lock<std::mutex> ulock (m_mutex);
-    T_PAYLOAD payload;
+    std::tuple<T_PAYLOAD, T_ERROR> payload_or_error;
 
-    auto condvar_pred = [this, &payload, a_rsn] ()
+    auto condvar_pred = [this, &payload_or_error, a_rsn] ()
```
alexstanro: Please, be sure that the thread does not hang - use a flag and cond_wait with timeout for instance.

src/communication/request_sync_client_server.hpp
```diff
@@ -193,31 +194,50 @@ namespace cubcomm
...
+    send_queue_error_handler error_handler_ftor = [&rsn, this] (
+		css_error_code error_code, bool &abort_further_processing)
+    {
+      abort_further_processing = false;
+      this->m_response_broker.register_error (rsn, std::move (error_code));
+    };
+
     // Send the request
-    m_queue->push (a_outgoing_message_id, std::move (seq_payload));
+    m_queue->push (a_outgoing_message_id, std::move (seq_payload), nullptr /*std::move (error_handler_ftor)*/);
```
alexstanro: Why  error_handler ftor is disabled here?

src/communication/response_broker.hpp
```diff
@@ -82,76 +105,112 @@ namespace cubcomm
...
+    }
+    m_condvar.notify_all (); // all because there is more than one thread waiting for data
+  }
+
+  template <typename T_PAYLOAD, typename T_ERROR>
+  void
+  response_broker<T_PAYLOAD, T_ERROR>::bucket::register_error (response_sequence_number a_rsn, T_ERROR &&a_error)
+  {
+    {
+      std::lock_guard<std::mutex> lockg (m_mutex);
```
alexstanro, line 195: m_send_mutex is already acquired. This may be source of deadlocks. At least add a comment, please.

src/communication/request_sync_client_server.hpp
cristiarg: Corrected mistake.

src/communication/response_broker.hpp
cristiarg: I do not understand the comment.
Either `register_error` or `register_response` is indeed called from the underlying sending thread - the one in `request_queue_autosend::start_thread` via `request_queue_autosend::loop_send_requests` and then `wait_not_empty_and_send_all` and then `send_queue`.
During this time, the client thread - maybe a transaction thread or a system thread - waits in `request_sync_client_server::send_recv` via `response_broker::bucket::get_response` but the wait is guarded by `m_mutex` and `m_condvar` so there should be no problem.

src/communication/response_broker.hpp
cristiarg: Indeed `get_response` might hang when the connection is dropped for whatever reason and neither response nor error will be registered.
Added a new flag - `m_terminate` - to the `bucket` and set it for all buckets in the `request_sync_client_server` dtor after the sending/receiving of network messages is terminated.

src/communication/response_broker.hpp
cristiarg: I see no way of moving it out since its access to the protected container - `m_response_payloads` and it needs to be guarded.

src/storage/page_buffer.c
```diff
@@ -8336,7 +8336,7 @@ pgbuf_request_data_page_from_page_server (const VPID * vpid, log_lsa target_repl
     }
 
   std::string response_message;
-  ts_Gl->send_receive (tran_to_page_request::SEND_DATA_PAGE_FETCH, std::move (request_message), response_message);
+  (void) ts_Gl->send_receive (tran_to_page_request::SEND_DATA_PAGE_FETCH, std::move (request_message), response_message);
```
cristiarg, line 8339: temporary, next patch will handle errors

src/transaction/log_page_buffer.c
```diff
@@ -2104,7 +2104,7 @@ logpb_request_log_page_from_page_server (LOG_PAGEID log_pageid, LOG_PAGE * log_p
       _er_log_debug (ARG_FILE_LINE, "[READ LOG] Sent request for log to Page Server. Page ID: %lld \n", log_pageid);
     }
   std::string response_message;
-  ts_Gl->send_receive (tran_to_page_request::SEND_LOG_PAGE_FETCH, std::move (request_message), response_message);
+  (void) ts_Gl->send_receive (tran_to_page_request::SEND_LOG_PAGE_FETCH, std::move (request_message), response_message);
```
cristiarg, line 2107: temporary, next patch will handle errors

src/server/tran_server.cpp
```diff
@@ -237,7 +237,7 @@ void
 tran_server::get_boot_info_from_page_server ()
 {
   std::string response_message;
-  send_receive (tran_to_page_request::GET_BOOT_INFO, std::string (), response_message);
+  (void) send_receive (tran_to_page_request::GET_BOOT_INFO, std::string (), response_message);
```
cristiarg, line 240: temporary, next patch will handle errors

src/server/passive_tran_server.cpp
```diff
@@ -129,7 +129,7 @@ void passive_tran_server::send_and_receive_stop_log_prior_dispatch ()
 
   std::string expected_empty_answer;
   // blocking call
-  send_receive (tran_to_page_request::SEND_STOP_LOG_PRIOR_DISPATCH, std::string (), expected_empty_answer);
+  (void) send_receive (tran_to_page_request::SEND_STOP_LOG_PRIOR_DISPATCH, std::string (), expected_empty_answer);
```
cristiarg, line 132: temporary, next patch will handle errors

src/server/passive_tran_server.cpp
```diff
@@ -71,7 +71,7 @@ void passive_tran_server::send_and_receive_log_boot_info (THREAD_ENTRY *thread_p
 {
   std::string log_boot_info;
 
-  send_receive (tran_to_page_request::SEND_LOG_BOOT_INFO_FETCH, std::string (), log_boot_info);
+  (void) send_receive (tran_to_page_request::SEND_LOG_BOOT_INFO_FETCH, std::string (), log_boot_info);
```
cristiarg, line 74: temporary, next patch will handle errors

alexstanro (approved):

PR #3425, opened 2022-03-14
[CBRD-24208] Cannot REORGANIZE PARTITION after partition rename.

youngjinj:
http://jira.cubrid.org/browse/CBRD-24208

Changing pexpr when renaming partition table.


src/query/execute_schema.c
```diff
@@ -4573,12 +4577,24 @@ do_rename_partition (MOP old_class, const char *newname)
...
-  error = au_fetch_class (old_class, &smclass, AU_FETCH_READ, AU_SELECT);
+  error = au_fetch_class (old_class, &smclass, AU_FETCH_UPDATE, AU_ALTER);
   if (error != NO_ERROR)
     {
       goto end_rename;
     }
 
+  strncpy (expr_copy, smclass->partition->expr, DB_MAX_PARTITION_EXPR_LENGTH);
+  keycol = expr_copy + 7;	/* strlen ("SELECT") + empty_string = 7 */
+  keycol_end = strstr (keycol, " ");
```
ctshim: Column names may contain spaces.
I don't think a simple whitespace check will solve it.

`create table t1 ([cond name] int primary key, c2 varchar);`
`SELECT [cond  name] FROM ~`

src/query/execute_schema.c
```diff
@@ -4573,12 +4577,24 @@ do_rename_partition (MOP old_class, const char *newname)
...
+  keycol_end[0] = '\0';
+
+  assert (snprintf (NULL, 0, "SELECT %s FROM [%s]", keycol, newname) < DB_MAX_PARTITION_EXPR_LENGTH);
```
hgryoo: In release mode, unexpected behavior could occur, How about `return ER_FAILED` instead of assert?

src/query/execute_schema.c
youngjinj: Thank you. I didn't think of a blank.
I'll think of another way.

youngjinj:
There was a problem when the identifier name contained spaces. (Refer to ctshim's comment)

So I changed the method to change after FROM.

src/query/execute_schema.c
youngjinj: Removed assert() check from commit in "Change only after FROM in pexpr. (2bcc3b8)".

Since DB_MAX_PARTITION_EXPR_LENGTH has a length of 2048 and newname has a maximum length of 255,
I thought the length didn't need to be checked.

src/query/execute_schema.c
```diff
@@ -4573,12 +4575,22 @@ do_rename_partition (MOP old_class, const char *newname)
...
+  error = au_fetch_class (old_class, &smclass, AU_FETCH_UPDATE, AU_ALTER);
   if (error != NO_ERROR)
     {
       goto end_rename;
     }
 
+  /* The pexpr format is defined in the function pt_node_to_partition_info. */
+  strncpy (expr, smclass->partition->expr, DB_MAX_PARTITION_EXPR_LENGTH);
+  expr_ptr = strstr (expr, "FROM");
+  expr_ptr = strstr (expr_ptr, "[");
```
beyondykk9: expr_ptr can be null. it needs to check if it's null.

youngjinj: Execute the do_rename_partition() function through the condition below.

``
if (is_partition == DB_PARTITIONED_CLASS)
``

pexpr of DB_PARTITION_CLASS can be NULL, but pexpr of DB_PARTITIONED_CLASS) cannot be NULL.

ctshim: in the case below
create table "from wild" ( "from river" int ):
==> SELECT [from wild] FROM [from river]

Please refer to the code below
```
// SELECT [ ~ ] FROM [ ~ ]
expr_ptr = strstr (expr, "]"); // skip column name
expr_ptr = strstr (expr_ptr + 1, "[");  // find table name position
```

hgryoo (approved):

src/query/execute_schema.c
youngjinj: I'll add a comment.

Added above case to test case attached to jira.

youngjinj: As you said, the method of finding table names in pexpr has been improved.

ctshim (approved):

src/query/execute_schema.c
beyondykk9: at the clause expr_ptr = strstr (expr, "FROM"), expr_ptr can be null if the "FROM" is not found.

youngjinj: The pexpr is created in the pt_node_to_partition_info() function.
It is not created by the user's query.

In the pt_node_to_partition_info () function, the pexpr query is fixed as "SELECT %s FROM [%s]".

The way pexpr looks up table names has changed, but pexpr can't be without "FROM".

hyunikn (approved):

beyondykk9 (approved):

shparkcubrid (approved):

PR #3426, opened 2022-03-15
[LETS-420] Mechanism to execute a trantable checkpoint right after boot is not reliable

adrianbrab:
http://jira.cubrid.org/browse/LETS-420

Changed the trantable checkpoint launch mechanism to wait in the daemon function until the first viable run.

### DIFF
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -10727,7 +10727,8 @@ log_checkpoint_trantable_daemon_init ()
 
   if (is_tran_server_with_remote_storage ())
     {
-      cubthread::looper looper (std::chrono::seconds (60));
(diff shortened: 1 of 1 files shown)


src/transaction/log_manager.c
```diff
@@ -10750,7 +10727,7 @@ log_checkpoint_trantable_daemon_init ()
 
   if (is_tran_server_with_remote_storage ())
     {
-      cubthread::looper looper (log_get_checkpoint_trantable_interval);
+      cubthread::looper looper (std::chrono::seconds (60));
```
cristiarg: constexpr

cristiarg (approved):

PR #3427, opened 2022-03-16
[CBRD-24225] Added information for debugging to csql_grammar.y and csql_lexer.l

ctshim:
http://jira.cubrid.org/browse/CBRD-24225

Added information for debugging to csql_grammar.y and csql_lexer.l

This is used when debugging information about grammar rules or information about the progress of the lexer is needed.
Change the `DBG_TRACE_LEVEL` value in the csql_grammar.y file and use it after building.
This feature is only available in SA_MODE.
The selectable option values are as follows.
```
   0: No displayed on the screen
   1: Words extracted from csql_lexer are displayed on the screen
   2: Display the contents of application of grammar rules in csql_grammar on the screen
  3: Expression of both 1 and 2 above
```

### DIFF
--- a/src/parser/csql_grammar.y
+++ b/src/parser/csql_grammar.y
@@ -84,6 +84,36 @@ extern int g_msg[1024];
 extern int msg_ptr;
 extern int yybuffer_pos;
 
+#if defined(SA_MODE)
+     /*
+     ** DBG_TRACE_LEVEL is specified by referring to the following.
+     ** 0: No displayed on the screen
+     ** 1: Words extracted from csql_lexer are displayed on the screen
+     ** 2: Display the contents of application of grammar rules in csql_grammar on the screen
+     ** 3: Expression of both 1 and 2 above
+     */
+#    define DBG_TRACE_LEVEL      0
+#endif
+
+#if (DBG_TRACE_LEVEL == 1 || DBG_TRACE_LEVEL == 3) 
+#    define DBG_PRINT_TOKEN(token)            fprintf(stdout, " *** Token) [%s]\n", token);
+#    define DBG_PRINT_STRING(pre, str, post)  fprintf(stdout, " *** Token) [%s%s%s]\n", pre, str, post);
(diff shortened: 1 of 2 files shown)


src/parser/csql_grammar.y
```diff
@@ -472,6 +494,19 @@ int g_original_buffer_len;
    DBG_PRINT                                            \
 } while (0)
 
+#define MAKE_MONETARY_LITERAL(rv, iv, b_p, sign_text, currency_type) do {    \
```
hyunikn, line 497: This macro has no use now. Is it used sometime?

hyunikn (approved):

src/parser/csql_grammar.y
ctshim: Ah, I found some repetitive code and prepared to replace it with this macro.
Since it is not directly related to this issue, it has not been modified, but the defined macro is not deleted and remains as it is.



The code below is repeatedly written in 24 blocks. 
The difference is only the two strings are different.
```
		{{

			char *str, *txt;
			PT_NODE *val = parser_new_node (this_parser, PT_VALUE);

			str = pt_append_string (this_parser, NULL, SERBIAN_DINAR_TEXT);
			txt = $2;

			if (val)
			  {
			    pt_value_set_monetary (this_parser, val, str, txt, PT_CURRENCY_SERBIAN_DINAR);
			  }

			$$ = val;

		DBG_PRINT}}
```
It can be replaced in the following way.
  { MAKE_MONETARY_LITERAL($$, $2, @$.buffer_pos,  **_SERBIAN_DINAR_TEXT_**, **_PT_CURRENCY_SERBIAN_DINAR_**); } 
  
  I don't know if I can fix it together in this PR.

hyunikn: Yep

beyondykk9 (approved):

shparkcubrid (approved):

youngjinj (approved):

hgryoo (approved):

PR #3429, opened 2022-03-16
[CBRD-24245] Square brackets([ ]) should not be used in identifier names.

youngjinj:
http://jira.cubrid.org/browse/CBRD-24245

- Square brackets should not be used in identifier names. However, it can be used in exception rules.
- Square brackets cannot be used again inside square brackets, but can be used inside double quotes and backtick symbols.
- Internally, when using reserved words, etc., an exception rule is used by using square brackets.
- Allowing the use of Square brackets in identifier names is problematic when wrapping them in Square brackets again.

```
create table t1 ("[c1]" int);
create view v1 as select * from t1;
select * from v1;

ERROR: In line 2, column 9 before ' from [t1] [t1]'
Syntax error: unexpected ']', expecting $end

csql> ;sc v1

 <VClass Name>

     v1

 <Attributes>

     [c1]                 INTEGER

 <Query_specifications>

     select [t1].[[c1]] from [t1] [t1]
```


shparkcubrid (approved):

src/parser/csql_grammar.y
```diff
@@ -26290,6 +26290,13 @@ pt_check_identifier (PARSER_CONTEXT *parser, PT_NODE *p, const char *str,
   char *invalid_pos = NULL;
   int composed_size;
 
+  if (strchr (str, '[') || strchr (str, ']'))
+    {
+      PT_ERRORf (this_parser, p,
+		 "Identifier name [%s] not allowed. It cannot contain '[' or ']'.",
+		 str);
```
hgryoo:
```suggestion
      PT_ERRORf (this_parser, p,
		 "Identifier name [%s] not allowed. It cannot contain '[' or ']'.",
		 str);
      return NULL;
```

youngjinj: Returning NULL causes ane->info.name.original to be NULL, and the core is raised by executing intl_identifier_casecmp(NULL, "...") on path_expression.

I know this in itself is a problem.
However, the purpose here is to pass the existing test cases as they are.

Problems caused by returning NULL will be resolved through other issues.

ctshim (approved):

src/parser/csql_grammar.y
```diff
@@ -26290,6 +26290,13 @@ pt_check_identifier (PARSER_CONTEXT *parser, PT_NODE *p, const char *str,
   char *invalid_pos = NULL;
   int composed_size;
 
+  if (strchr (str, '[') || strchr (str, ']'))
+    {
+      PT_ERRORf (this_parser, p,
+		 "Identifier name [%s] not allowed. It cannot contain '[' or ']'.",
```
beyondykk9:
```suggestion
		 "Identifier name \"%s\" not allowed. It cannot contain '[' or ']'.",
```

beyondykk9 (approved):

hgryoo (approved):

hyunikn (approved):

PR #3385, opened 2022-02-25
[CBRD-24178] Add help message and error message for flashback utility

joohok:
http://jira.cubrid.org/browse/CBRD-24178

Purpose

1. Set messages for utility
2. Print usage and error message at utility
3. Set error message for each error code in engine

PR #3433, opened 2022-03-18
[CBRD-24227] Handling multiple executions of flashback

joohok:
http://jira.cubrid.org/browse/CBRD-24227

Purpose

For Blocking duplicated flashback execution,
it need to check if previous flashback is in progress or shutdown abnormally or exited properly. 

Implementation
To know status of previous flashback request, 
1. added flashback status in css_conn_entry for request handler (conn_entry->in_flashback)
2. maintained flashback connection entry (flashback_Current_conn)

### DIFF
--- a/src/base/error_code.h
+++ b/src/base/error_code.h
@@ -1707,8 +1707,9 @@
 #define ER_FLASHBACK_SCHEMA_CHANGED                 -1336
 #define ER_FLASHBACK_LOG_NOT_EXIST                  -1337
 #define ER_FLASHBACK_TIMEOUT                        -1338
+#define ER_FLASHBACK_DUPLICATED_REQUEST             -1339
 
-#define ER_LAST_ERROR                               -1339
+#define ER_LAST_ERROR                               -1340
 /*
  * CAUTION!
  *
--- a/src/communication/network_interface_sr.c
+++ b/src/communication/network_interface_sr.c
@@ -10820,14 +10820,27 @@ sflashback_get_summary (THREAD_ENTRY * thread_p, unsigned int rid, char *request
   time_t start_time = 0;
   time_t end_time = 0;
 
-  if (flashback_is_duplicated_request (thread_p))
+  bool is_duplicated_request = false;
+
+  /* If multiple requests come at the same time,
+   * they can all be treated as non-duplicate requests.
+   * So, latch for check and set flashback connection is required */
+
+  flashback_lock_request ();
+
+  is_duplicated_request = flashback_is_duplicated_request (thread_p);
+  if (is_duplicated_request)
     {
       error_code = ER_FLASHBACK_DUPLICATED_REQUEST;
+
+      flashback_unlock_request ();
       goto error;
     }
   else
     {
       flashback_initialize (thread_p);
+
+      flashback_unlock_request ();
     }
 
   flashback_set_status_active ();
@@ -10929,7 +10942,12 @@ error:
 
   (void) css_send_data_to_client (thread_p->conn_entry, rid, reply, OR_ALIGNED_BUF_SIZE (a_reply));
 
-  flashback_reset ();
+  if (!is_duplicated_request)
+    {
+      /* if flashback variables are reset by duplicated request error,
+       * variables for existing connection (valid connection) can be reset */
+      flashback_reset ();
+    }
 
   return;
 
--- a/src/communication/network_interface_sr.c
+++ b/src/communication/network_interface_sr.c
@@ -10820,27 +10820,10 @@ sflashback_get_summary (THREAD_ENTRY * thread_p, unsigned int rid, char *request
   time_t start_time = 0;
   time_t end_time = 0;
 
-  bool is_duplicated_request = false;
-
-  /* If multiple requests come at the same time,
-   * they can all be treated as non-duplicate requests.
-   * So, latch for check and set flashback connection is required */
-
-  flashback_lock_request ();
-
-  is_duplicated_request = flashback_is_duplicated_request (thread_p);
-  if (is_duplicated_request)
+  error_code = flashback_initialize (thread_p);
+  if (error_code != NO_ERROR)
     {
-      error_code = ER_FLASHBACK_DUPLICATED_REQUEST;
-
-      flashback_unlock_request ();
       goto error;
-    }
-  else
-    {
-      flashback_initialize (thread_p);
-
-      flashback_unlock_request ();
     }
 
   flashback_set_status_active ();
@@ -10942,7 +10925,7 @@ error:
 
   (void) css_send_data_to_client (thread_p->conn_entry, rid, reply, OR_ALIGNED_BUF_SIZE (a_reply));
 
-  if (!is_duplicated_request)
+  if (error_code != ER_FLASHBACK_DUPLICATED_REQUEST)
     {
       /* if flashback variables are reset by duplicated request error,
        * variables for existing connection (valid connection) can be reset */
--- a/src/communication/network_interface_sr.c
+++ b/src/communication/network_interface_sr.c
@@ -10820,15 +10820,14 @@ sflashback_get_summary (THREAD_ENTRY * thread_p, unsigned int rid, char *request
   time_t start_time = 0;
   time_t end_time = 0;
 
-  if (flashback_min_log_pageid_to_keep () != NULL_LOG_PAGEID)
+  if (flashback_is_duplicated_request (thread_p))
     {
-      /* if flashback was shutdown abnormally, flashback_min_log_pageid can not be cleared
-       * If the previous flashback was abnormally terminated,
-       * this flashback_min_log_pageid  would not have been initialized.
-       * Therefore, after all variables related to flashback are initialized, subsequent operations should be performed
-       */
-
-      flashback_reset ();
+      error_code = ER_FLASHBACK_DUPLICATED_REQUEST;
+      goto error;
+    }
+  else
+    {
+      flashback_initialize (thread_p);
     }
 
   flashback_set_status_active ();
--- a/src/executables/util_cs.c
+++ b/src/executables/util_cs.c
@@ -4922,6 +4922,12 @@ flashback (UTIL_FUNCTION_ARG * arg)
   is_detail = utility_get_option_bool_value (arg_map, FLASHBACK_DETAIL_S);
   is_oldest = utility_get_option_bool_value (arg_map, FLASHBACK_OLDEST_S);
 
+  if (!prm_get_integer_value (PRM_ID_SUPPLEMENTAL_LOG))
+    {
+      fprintf (stderr, "please set \"supplemental_log\" in conf/cubrid.conf\n");
+      goto error_exit;
+    }
+
   /* create table list */
   /* class existence and classoid will be found at server side. if is checked at utility side, it needs addtional access to the server through locator */
   darray = da_create (num_tables, SM_MAX_IDENTIFIER_LENGTH);
@@ -5059,12 +5065,6 @@ flashback (UTIL_FUNCTION_ARG * arg)
     }
 
   need_shutdown = true;
-
-  if (!prm_get_integer_value (PRM_ID_SUPPLEMENTAL_LOG))
-    {
-      fprintf (stderr, "please set \"supplemental_log\" in conf/cubrid.conf\n");
-      goto error_exit;
-    }
 
   oid_list = (OID *) malloc (sizeof (OID) * num_tables);
   if (oid_list == NULL)
--- a/src/executables/util_cs.c
+++ b/src/executables/util_cs.c
@@ -4922,12 +4922,6 @@ flashback (UTIL_FUNCTION_ARG * arg)
   is_detail = utility_get_option_bool_value (arg_map, FLASHBACK_DETAIL_S);
   is_oldest = utility_get_option_bool_value (arg_map, FLASHBACK_OLDEST_S);
 
-  if (!prm_get_integer_value (PRM_ID_SUPPLEMENTAL_LOG))
-    {
-      fprintf (stderr, "please set \"supplemental_log\" in conf/cubrid.conf\n");
-      goto error_exit;
-    }
-
   /* create table list */
   /* class existence and classoid will be found at server side. if is checked at utility side, it needs addtional access to the server through locator */
   darray = da_create (num_tables, SM_MAX_IDENTIFIER_LENGTH);
@@ -5065,6 +5059,12 @@ flashback (UTIL_FUNCTION_ARG * arg)
     }
 
   need_shutdown = true;
+
+  if (!prm_get_integer_value (PRM_ID_SUPPLEMENTAL_LOG))
+    {
+      fprintf (stderr, "please set \"supplemental_log\" in conf/cubrid.conf\n");
+      goto error_exit;
+    }
 
   oid_list = (OID *) malloc (sizeof (OID) * num_tables);
   if (oid_list == NULL)
--- a/src/transaction/flashback.c
+++ b/src/transaction/flashback.c
@@ -70,11 +70,25 @@ static std::atomic_bool flashback_In_progress = false;  // the status value that
 
 static CSS_CONN_ENTRY *flashback_Current_conn = NULL;	// the connection entry for a flashback requests
 
+static pthread_mutex_t flashback_Request_lock = PTHREAD_MUTEX_INITIALIZER;
+
 /*
  * flashback_is_duplicated_request - check if the caller is duplicated request for flashback
  *
  * return   : duplicated or not
  */
+
+void
+flashback_lock_request ()
+{
+  pthread_mutex_lock (&flashback_Request_lock);
+}
+
+void
+flashback_unlock_request ()
+{
+  pthread_mutex_unlock (&flashback_Request_lock);
+}
 
 bool
 flashback_is_duplicated_request (THREAD_ENTRY * thread_p)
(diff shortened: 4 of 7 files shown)


src/executables/util_cs.c
```diff
@@ -5066,6 +5060,12 @@ flashback (UTIL_FUNCTION_ARG * arg)
 
   need_shutdown = true;
 
+  if (!prm_get_integer_value (PRM_ID_SUPPLEMENTAL_LOG))
```
wrlawodms: Why has it been moved?

src/communication/network_interface_sr.c
```diff
@@ -10820,15 +10820,14 @@ sflashback_get_summary (THREAD_ENTRY * thread_p, unsigned int rid, char *request
   time_t start_time = 0;
   time_t end_time = 0;
 
-  if (flashback_min_log_pageid_to_keep () != NULL_LOG_PAGEID)
+  if (flashback_is_duplicated_request (thread_p))
```
wrlawodms: If more than two requests come simultaneously, all they pass this condition because no one has initialized 'flashback_Current_conn` yet. It seems to need a latch in the function.

wrlawodms (commented):
1. Could you explain shortly the connection status such as CONN_OEPN and when the transition among them happens?
2. How long does it take usually from OPEN to CLOSING or CLOSED? I think users feel it is a delay of the exit of flashback when they do execute-exit-execute.

src/communication/network_interface_sr.c
joohok: Yes, you are right. I will change it

src/executables/util_cs.c
joohok: Oh.. this is unrelated change. it is done in other PR (#3385) but not merged.
I moved this because parameters are loaded after db_restart ().

src/base/error_code.h
```diff
@@ -1707,8 +1707,9 @@
 #define ER_FLASHBACK_SCHEMA_CHANGED                 -1336
 #define ER_FLASHBACK_LOG_NOT_EXIST                  -1337
 #define ER_FLASHBACK_TIMEOUT                        -1338
+#define ER_FLASHBACK_DUPLICATED_REQUEST             -1339
```
hornetmj, line 1710: How about ER_FLASHBACK_ALREADY_IN_PROGRESS_BY_DBA?

src/transaction/flashback.c
```diff
@@ -64,9 +64,88 @@ static volatile int flashback_Threshold_to_remove_archive = 0;	/* If the differe
...
+static CSS_CONN_ENTRY *flashback_Current_conn = NULL;	// the connection entry for a flashback requests
+
+/*
+ * flashback_is_duplicated_request - check if the caller is duplicated request for flashback
+ *
+ * return   : duplicated or not
+ */
+
+bool
+flashback_is_duplicated_request (THREAD_ENTRY * thread_p)
```
hornetmj, line 81: How about the following code flow? IMO, it is more readable.

```
if (flashback_In_progress == false)
{
    // 이전 flashback 연결이 정상적으로 종료된 경우. 현재 수행 중인 flashback 없음.
    return false;
}
else
{
    if (flashback_Current_conn->in_flashback == true)
    {
        // 이전 flashback 요청이 아직 처리 중인 경우
        return true;
    }
    else
    {
        // 이전 flashback 요청이 비정상 종료된 경우
        return false;
    }
}
```

src/base/error_code.h
joohok: I followed ER_LOG_BKUP_DUPLICATE_REQUESTS, which is set when a new backup request comes in while the backup is in progress.

What about ER_FLASHBACK_ALREADY_IN_PROGRESS if the error code is not indicating a duplicate request, but rather an existing flashback in progress?

src/transaction/flashback.c
joohok: Does this code mean to remove `if(previous_conn->status == CONN_OPEN)`? 
or 
Does it mean to first classify whether or not flashback is normally terminated using `flashback_In_progress`?

joohok:
@wrlawodms Sorry, I missed the comments 
> * Could you explain shortly the connection status such as CONN_OPEN and when the transition among them happens?

CONN_OPEN : initial status, when conn_entry is allocated.

CONN_CLOSED : connection closed due to shutdown or when connection error handler (net_server_conn_down) detects disconnection with clients, it changes the status from CONN_OPEN, CONN_CLOSING to CONN_CLOSED

CONN_CLOSING : CONN_CLOSING is soon to be CONN_CLOSED
if request is done or client is unregistered or some error is detected 
then server sets status to CON_CLOSING and, at the end of the connection handler (net_server_conn_down), it changes this status to CONN_CLOSED. 

> * How long does it take usually from OPEN to CLOSING or CLOSED? I think users feel it is a delay of the exit of flashback when they do execute-exit-execute.

I don't know exactly, but it seems connection handler uses poll() with 100 miliseconds timeout to check client is alive. (css_connection_handler_thread)
and if there is error while polling, it would changes the status to CONN_CLOSED.

src/transaction/flashback.c
hornetmj: @joohok Both of them. I think it is good to discuss off-line first.

src/base/error_code.h
hornetmj: @joohok Thanks for your reply. Let's use ER_FLASHBACK_DUPLICATED_REQUEST as is like ER_LOG_BKUP_DUPLICATE_REQUESTS.

src/transaction/flashback.h
```diff
@@ -108,6 +109,8 @@ typedef struct flashback_summary_context
 extern int flashback_make_summary_list (THREAD_ENTRY * thread_p, FLASHBACK_SUMMARY_CONTEXT * context);
 extern void flashback_cleanup (THREAD_ENTRY * thread_p, FLASHBACK_SUMMARY_CONTEXT * context);
 
+extern bool flashback_is_duplicated_request (THREAD_ENTRY * thread_p);
+extern void flashback_initialize (THREAD_ENTRY * thread_p);
```
eido5: Is `flashback_init()` better to keep name consistency?

And is `flashback_initialize()` function which initializes the module?

If it is initializing module, I usually watched the name as `[module_name]_init()` that the function of name executes to initialize the module.

I executed '**grep -r [keyword] ***' and then rarely its name is 'initialize()'. 

Actually, it is **not** related with **functional view**. It is only about the name **consistency**.

what do you think?

Briefly,
**AS-IS**
flashback_initialize()

**TO-BE**
flashback_init()

src/base/error_code.h
joohok: OK, Thanks. I will leave it.

src/transaction/flashback.c
joohok: OK, lets discuss it off-line.

src/transaction/flashback.h
joohok: Thanks for feedback. 
'init' is just an abbreviation for 'initialize'. 
And there seems to be no consistency in the function names related to 'init' or 'initialize' in legacy code.
If it is required to be consistent, perfmon_initialize, locator_initialize, etc. should be modified too.

eido5: Oh I see thank you for replying.
I knew the name consistency should be kept. 
Thank you.

eido5 (approved):

src/executables/util_cs.c
joohok: I will revert it when all the feedbacks are applied. 
Now, I just leave it for a smooth test.

src/communication/network_interface_sr.c
```diff
@@ -10820,15 +10820,27 @@ sflashback_get_summary (THREAD_ENTRY * thread_p, unsigned int rid, char *request
   time_t start_time = 0;
   time_t end_time = 0;
 
-  if (flashback_min_log_pageid_to_keep () != NULL_LOG_PAGEID)
+  bool is_duplicated_request = false;
```
wrlawodms: This can be removed if we check it with `error_code == ER_FLASHBACK_DUPLICATED_REQUEST`.

src/communication/network_interface_sr.c
```diff
@@ -10820,15 +10820,27 @@ sflashback_get_summary (THREAD_ENTRY * thread_p, unsigned int rid, char *request
...
+
+  /* If multiple requests come at the same time,
+   * they can all be treated as non-duplicate requests.
+   * So, latch for check and set flashback connection is required */
+
+  flashback_lock_request ();
+
+  is_duplicated_request = flashback_is_duplicated_request (thread_p);
```
wrlawodms: How about putting the concurrent request check into the `flashback_initialize()` because the latch for it seems only used during initialization. We can remove global variables of mutex and functions (`flashback_lock_request()` and `flashback_unlock_request()`) only used here. The code I'm suggesting will be like this:
```suggestion
  error = flashback_initialize (thread_p);
  if (error != NO_ERROR)
  {
    /* maybe only ER_FLASHBACK_DUPLICATED_REQUEST */
    goto error;
  }
```

We can also make the mutex variable a static local variable, but it's up to you (I'm not sure which is better between using a static local variable and a static global variable.). Let me have your opinion on this.

wrlawodms: It even helps us shorten the critical section. However, if this critical section is shared with others, you should go as you did.

src/transaction/flashback.c
```diff
@@ -64,9 +64,92 @@ static volatile int flashback_Threshold_to_remove_archive = 0;	/* If the differe
...
+  if (flashback_In_progress == false)
+    {
+      /* previous flashback set flashback_In_progress to false prolperly. (exited well) */
+      return false;
+    }
+  else
+    {
+      /* flashback_In_progress == true */
+
+      if (previous_conn->in_flashback == true)
```
wrlawodms: For safety, we have to make sure that `CSS_CONN_ENTRY` is not deallocated during running. It does?

src/communication/network_interface_sr.c
joohok: Oh, Thanks. I will change it

src/communication/network_interface_sr.c
joohok: Thanks for the detailed feedback. 
Lock is also required in checking if it is duplicated request.
If two connections are simultaneously enter the 'flashback_is_duplicated_request()' and 'flashback_Current_conn' is null, it is determined that both connections are not duplicated and both can proceed with initialization.

src/transaction/flashback.c
joohok: Yes, CSS_CONN_ENTRY is the one of the entries in css_Conn_array, which is allocated at server startup. 
And css_Conn_array is deallocated on server shutdown.

Or does it mean that the connection is lost in the process rather than deallocating the memory?

wrlawodms: No, I meant deallocating. Thank you!

src/communication/network_interface_sr.c
wrlawodms: > Lock is also required in checking if it is duplicated request

Yes, but it seems that the check is done only during initializing the flashback. Isn't it?

joohok: Oh, now I understand what you mean. 
check and settings (initialization) are only done here, and are not shared anywhere else. 

Other mutex locks (css_Client_id_lock, css_Conn_rule_lock,.. ) are declared as static global variables, even if they are only used in specific function.

But I think static local variable is sufficient because it is only used in initialization functions. If it is used in other functions in the future, it would be better to modify it then.

wrlawodms: Yes, it looks better to me. My point was, by the way, that we don't have to open latch interfaces, but can just acquire and release the latch locally in the initialization function.

wrlawodms: Either the duplication check does.

joohok: Yes, I agree with you

src/transaction/flashback.c
```diff
@@ -64,9 +64,97 @@ static volatile int flashback_Threshold_to_remove_archive = 0;	/* If the differe
...
+  /* If multiple requests come at the same time,
+   * they all can be treated as non-duplicate requests.
+   * So, latch for check and set flashback connection is required */
+
+  pthread_mutex_lock (&conn_mutex);
+
+  if (flashback_is_duplicated_request (thread_p))
+    {
+      /* er_set() */
+      return ER_FLASHBACK_DUPLICATED_REQUEST;
```
wrlawodms, line 129:
```suggestion
      pthread_mutex_unlock (&conn_mutex); 
      /* er_set() */
      return ER_FLASHBACK_DUPLICATED_REQUEST;
```

src/transaction/flashback.c
```diff
@@ -64,9 +64,97 @@ static volatile int flashback_Threshold_to_remove_archive = 0;	/* If the differe
...
+
+/*
+ * flashback_initialize - check request if is duplicated, 
+ *                        then initialize variables and connection when flashback request is started
+ */
+
+int
+flashback_initialize (THREAD_ENTRY * thread_p)
+{
+  static pthread_mutex_t conn_mutex = PTHREAD_MUTEX_INITIALIZER;
```
wrlawodms: I'm still not sure whether global is better or local is. The target of this mutex is global, so it seems to have to be on the same level (as a global var), but it is used only here. Now I think it as a global var is better, but I'll follow your choice.

joohok: Yes, I agree with what you said. 
I think it's better to set this as a global variable according to the target of this lock.

src/transaction/flashback.c
joohok: Thank you

src/transaction/flashback.c
```diff
@@ -64,9 +64,99 @@ static volatile int flashback_Threshold_to_remove_archive = 0;	/* If the differe
...
+
+  if (flashback_is_duplicated_request (thread_p))
+    {
+      pthread_mutex_unlock (&flashback_Conn_lock);
+
+      /* er_set() */
+      return ER_FLASHBACK_DUPLICATED_REQUEST;
+    }
+
+  flashback_Current_conn = thread_p->conn_entry;
```
hornetmj, line 134:
```suggestion
  assert (flashback_In_progress == false && flashback_Current_conn == NULL);

  flashback_Current_conn = thread_p->conn_entry;
```

src/transaction/flashback.c
```diff
@@ -64,9 +64,99 @@ static volatile int flashback_Threshold_to_remove_archive = 0;	/* If the differe
...
+
+  /* When flashback is first called or exited properly,
+   * flashback_Current_conn is set to NULL */
+  if (flashback_Current_conn == NULL)
+    {
+      return false;
+    }
+
+  /* flashback_Current_conn indicates conn_entry in thread_p, and conn_entry can be reused by request handler.
+   * So, status in flashback_Current_conn can be overwritten. */
```
hornetmj:
```suggestion
```

src/transaction/flashback.c
```diff
@@ -64,9 +64,99 @@ static volatile int flashback_Threshold_to_remove_archive = 0;	/* If the differe
...
+  previous_conn = flashback_Current_conn;
+
+  if (flashback_In_progress == false)
+    {
+      /* previous flashback set flashback_In_progress to false prolperly. (exited well) */
+      return false;
+    }
+  else
+    {
+      /* flashback_In_progress == true */
```
hornetmj:
```suggestion
      /* flashback_In_progress == true */
      
      assert (previous_conn != NULL);
```

hornetmj: It would be not bad to use just the flashback_Current_conn global variable.

hornetmj: And I think that the flashback_Current_conn global variable can replace the purpose of the flashback_In_progress global variable. Then you can reduce/remove the usage of the global variable flashback_In_progress.

src/transaction/flashback.c
joohok: Yes, this is a duplicate condition. I'll remove it

src/transaction/flashback.c
joohok: there is a case that previous flashback request has been shut down abnormally.
if it is shutdown abnormally, then the condition is as below
```
flashback_In_progress == true && 
flashback_Current_conn != NULL && 
flashback_Current_conn->in_flashback == false
```

So the above assert cannot be appended.

src/transaction/flashback.c
joohok: I used `previous_conn` to make the code more understandable. 

do you mean that the condition in 'flashback_In_progress' can be changed to the condition in 'flashback_Current_conn' like below?
```
flashback_In_progress == false    ->  flashback_Current_conn == NULL
flashback_In_progress == true     ->  flashback_Current_conn != NULL 
```

joohok: I understand what you mean.
I changed it as you said and tested it, and it worked well for the purpose.

src/transaction/flashback.c
hornetmj: If you found the situation that the previous flashback request was shut down abnormally, then you need to reset the global variables correctly. isn't it?

joohok: Yes, you are right. 
I will reset the variables.

src/transaction/flashback.c
```diff
@@ -67,6 +67,87 @@ static volatile int flashback_Threshold_to_remove_archive = 0;	/* If the differe
...
+/*
+ * flashback_is_duplicated_request - check if the caller is duplicated request for flashback
+ *
+ * return   : duplicated or not
+ */
+
+static bool
+flashback_is_duplicated_request (THREAD_ENTRY * thread_p)
+{
+  CSS_CONN_ENTRY *previous_conn = NULL;
```
hornetmj:
```suggestion
```

hornetmj (approved):

eido5 (approved):

wrlawodms (approved):

joohok:
@mhoh3963 
Please merge this PR.
This PR did not cause the test to fail.

PR #3441, opened 2022-03-22
[LETS-429] proper network communication stopping order for page [and transaction] server

cristiarg:
http://jira.cubrid.org/browse/LETS-429

Race condition stopping page server communication infrastructure when stopping page server.
When page server is stopped, first all currently async executing responses - which are handled via `server_request_responder` - must be waited for termination.
Another scenario when it is required is when a connection must be stopped - eg, as when a passive transaction server stops. Only the async processing for that connection need to be waited to terminate.

Implementation:
- change the implementation of `page_server`'s each `connection_handler`'s dtor to first stop incoming communication thread, then wait for all async processing to finish (which, will also dispatch responses over the network via the outgoing communication thread) and then, as a final step, stop the outgoing communication thread
- exposed function out of `request_sync_client_server` for stopping the incoming/outgoing communication threads
- to allow `page_server` to wait for async processing to terminate on a per-connection basis, `server_request_responder` is augmented with a bookkeeping mechanism for tasks of a certain connection - `connection_executing_task_type` is a container item of which one exists per each (page server) connection - when a new async task is dispatched for processing for a connection, a number is incremented - when the task finishes execution, the number is decremented - thus allowing an external agent to wait for the processing to terminate


src/server/server_request_responder.hpp
```diff
@@ -120,17 +179,70 @@ void
...
+  auto found_it = m_executing_tasks.find (connection_ptr);
+  if (found_it != m_executing_tasks.end ())
+    {
+      connection_executing_task_type &executing_task_for_connection = (*found_it).second;
+      while (true)
+	{
+	  if (executing_task_for_connection.m_cv.wait_for (ulock, millis_20,
+	      [&executing_task_for_connection] { return (executing_task_for_connection.n_count == 0); }))
+	    {
+	      m_executing_tasks.erase (found_it);
```
cristiarg, line 265: Might be an overkill to create and erase the container element each time its bookkeeping reaches `0`.
For a connection which only has consecutive requests, this will end repeatedly adding & removing instances of `connection_executing_task_type` from the bookkeeping container (`m_executing_tasks`).
Might be better to just leave the entry there at the cost of actually cluttering the container with entries. The way entries are added, in the scalability architecture, is by passive transaction servers coming online/going offline as part of normal functioning of a cluster.
Implementing the latter would also supersede the need to notify the enclosing wait loop from being notified from under the mutex in `retire_task`.

alexstanro (approved):

src/server/server_request_responder.hpp
cristiarg: Incorrect comment because the responder is only waited for once to become idle for every connection. Refactored the code to formally enforce this invariant.

PR #3443, opened 2022-03-23
[CBRD-24182] Remove unnecessary table such as outer joined table with unique predicates

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-24182

The case of outer joined table with unique predicates.

```
SELECT count(*)  <== referenced column from tbl2 is not exists.
   FROM  tbl1 left outer join tbl2  <== tbl2 is outer joined.
 on tbl1.col1 = tbl2.PK <== tbl2.PK is unique predicate.
```
In this case, tbl2 can be removed.

How to check unique predicates.
 - All columns of a unique constraint must exist in the on_cond clause.
 - Only predcates with the following conditions are included.
 |              arg1                       |       op      |          arg2                      |
 | only have one of my spec  |       '='      |  not have my spec node|
 | e.g.)      col(my spec)          |        =       |  2 (constant)                  |
 |            a.col(my spec)         |        =       |  b.col                             |
 |            a.col(my spec)         |        =       |  b.col + c.col                 |

Restrictions
 - CTE spec
 - Right outer join


src/optimizer/query_rewrite.c
```diff
@@ -3074,6 +3088,410 @@ qo_reduce_comp_pair_terms (PARSER_CONTEXT * parser, PT_NODE ** wherep)
...
+ *
+ * Note:
+ */
+static void
+qo_reset_spec_location (PARSER_CONTEXT * parser, PT_NODE * spec, PT_NODE * query)
+{
+  short curr_loc, after_loc;
+  PT_NODE *where;
+  RESET_LOCATION_INFO locate_info;
+
```
byungwook-kim, line 3291: PT_NODE *p;

beyondykk9, line 3291:
```suggestion
PT_NODE *sp

sp = spec;
while (sp != NULL)
    {
      ...
      sp = sp->next;
    }
```

shparkcubrid: Since the spec is only used here, I think it would be ok to keep it.

beyondykk9 (approved):

hyunikn (approved):

ctshim (approved):

src/parser/parse_tree_cl.c
```diff
@@ -3103,6 +3103,27 @@ parser_append_node (PT_NODE * node, PT_NODE * list)
   return list;
 }
 
+/*
+ * parser_append_previous_node() -
+ *   return:
+ *   node(in/out):
```
hgryoo:
```suggestion
 *   node(in):
```

src/parser/parse_tree_cl.c
```diff
@@ -3103,6 +3103,27 @@ parser_append_node (PT_NODE * node, PT_NODE * list)
...
+  if (list)
+    {
+      node->next = list;
+      list = node;
+    }
+  else
+    {
+      list = node;
+    }
+  return list;
```
hgryoo, line 3124: It is shortened version.
```suggestion
  if (list)
    {
      node->next = list;
      node = list;
    }
  return node;
```

src/optimizer/query_rewrite.c
```diff
@@ -3074,6 +3088,410 @@ qo_reduce_comp_pair_terms (PARSER_CONTEXT * parser, PT_NODE ** wherep)
...
+ *                select count(*)
+ *                  from tbl1 a
+ */
+static PT_NODE *
+qo_reduce_outer_joined_tables (PARSER_CONTEXT * parser, PT_NODE * spec, PT_NODE * query)
+{
+  MOP cls;
+  SM_CLASS_CONSTRAINT *consp;
+  SM_ATTRIBUTE *attrp;
+  PT_NODE *point_list, *point, *where, *col, *tmp_spec, *prev_spec, *pred, *prev_pred, *next_pred;
```
hgryoo:
```suggestion
  PT_NODE *point_list = NULL, *point, *where, *col, *tmp_spec, *prev_spec, *pred, *prev_pred, *next_pred;
```

at the `end` flag, point_list is freed according to the case this variable is NULL.

hgryoo (approved):

src/optimizer/query_rewrite.c
shparkcubrid: I'll fix it.

src/parser/parse_tree_cl.c
shparkcubrid: I'll fix it.

src/parser/parse_tree_cl.c
shparkcubrid: I think it is not same.

youngjinj (approved):

src/parser/parse_tree_cl.c
hgryoo: Assigning to the `list` variable changes the local variable, so there is no change outside this function.

hgryoo: I wonder if you are intending to change the reference of the list variable passed when calling this function.

shparkcubrid: `list` is return var. so I think it is okay.

PR #3445, opened 2022-03-23
[LETS-421] Separate replication LSA in two separate concepts

adrianbrab:
http://jira.cubrid.org/browse/LETS-421

Separated `get_replicator_lsa` into 2 new functions:

- `passive_tran_server::get_highest_processed_lsa`
  - to be used for retrieving data pages from page server
  - in current replicator, it is m_redo_lsa
- `passive_tran_server::get_min_unapplied_lsa`
  - to be used to wait passive transaction server to catch up with a desynchronized page (a too advanced data page that has been retrieved from page server)
  - in current log replicator:
    - if parallel replication is not used it is also `m_redo_lsa`
    - if parallel replication is used, the value is given by the parallel replication infrastructure
  - also, in the current replicator, `get_min_unapplied_lsa` should not be called, therefore function should assert, function will be used later when replicator will be executed with two separate flavors - for passive transaction server and page server

### DIFF
--- a/src/server/passive_tran_server.cpp
+++ b/src/server/passive_tran_server.cpp
@@ -153,10 +153,9 @@ log_lsa passive_tran_server::get_highest_processed_lsa () const
   return m_replicator->get_highest_processed_lsa ();
 }
 
-log_lsa passive_tran_server::get_min_unapplied_lsa () const
+log_lsa passive_tran_server::get_lowest_unapplied_lsa () const
 {
-  assert (false);
-  return m_replicator->get_min_unapplied_lsa ();
+  return m_replicator->get_lowest_unapplied_lsa ();
 }
 
 void passive_tran_server::finish_replication_during_shutdown (cubthread::entry &thread_entry)
--- a/src/server/passive_tran_server.cpp
+++ b/src/server/passive_tran_server.cpp
@@ -148,9 +148,15 @@ void passive_tran_server::send_and_receive_stop_log_prior_dispatch ()
   // needs to be consumed (aka: waited to be consumed/serialized to log)
 }
 
-log_lsa passive_tran_server::get_replicator_lsa () const
+log_lsa passive_tran_server::get_highest_processed_lsa () const
 {
-  return m_replicator->get_redo_lsa ();
+  return m_replicator->get_highest_processed_lsa ();
+}
+
+log_lsa passive_tran_server::get_min_unapplied_lsa () const
+{
+  assert (false);
+  return m_replicator->get_min_unapplied_lsa ();
 }
 
 void passive_tran_server::finish_replication_during_shutdown (cubthread::entry &thread_entry)
--- a/src/server/passive_tran_server.hpp
+++ b/src/server/passive_tran_server.hpp
@@ -37,7 +37,8 @@ class passive_tran_server : public tran_server
     void send_and_receive_stop_log_prior_dispatch ();
 
     /* read replicator's current progress */
-    log_lsa get_replicator_lsa () const;
+    log_lsa get_highest_processed_lsa () const;
+    log_lsa get_min_unapplied_lsa () const;
     void finish_replication_during_shutdown (cubthread::entry &thread_entry);
     void wait_replication_past_target_lsa (LOG_LSA lsa);
 
--- a/src/server/passive_tran_server.hpp
+++ b/src/server/passive_tran_server.hpp
@@ -36,9 +36,10 @@ class passive_tran_server : public tran_server
     void start_log_replicator (const log_lsa &start_lsa);
     void send_and_receive_stop_log_prior_dispatch ();
 
-    /* read replicator's current progress */
+    /* highest processed lsa, to be used for retrieve pages from PS */
     log_lsa get_highest_processed_lsa () const;
-    log_lsa get_min_unapplied_lsa () const;
+    /* lowest unapplied lsa, to be used to wait for page sync */
+    log_lsa get_lowest_unapplied_lsa () const;
     void finish_replication_during_shutdown (cubthread::entry &thread_entry);
     void wait_replication_past_target_lsa (LOG_LSA lsa);
 
--- a/src/storage/page_buffer.c
+++ b/src/storage/page_buffer.c
@@ -2373,14 +2373,24 @@ pgbuf_check_page_ahead_of_replication (THREAD_ENTRY * thread_p, PAGE_PTR page)
   // Compare page LSA to replication LSA. If page is ahead, there may be a desynchronization issue; its LSA must be
   // saved in transaction descriptor and the appropriate error must be set and returned.
   LOG_LSA page_lsa = *pgbuf_get_lsa (page);
-  LOG_LSA repl_lsa = get_passive_tran_server_ptr ()->get_lowest_unapplied_lsa ();
-  if (page_lsa > repl_lsa)
+  LOG_LSA lowest_unapplied_lsa = get_passive_tran_server_ptr ()->get_lowest_unapplied_lsa ();
+
+  // Print the current replication progress to aid in debug scenarios
+  if (prm_get_bool_value (PRM_ID_ER_LOG_DEBUG))
+    {
+      LOG_LSA highest_applied_lsa = get_passive_tran_server_ptr ()->get_highest_processed_lsa ();
+      _er_log_debug (ARG_FILE_LINE,
+		     "Current replication progress is situated at: lowest unapplied lsa: %lld|%d, highest processed lsa: %lld|%d.",
+		     LSA_AS_ARGS (lowest_unapplied_lsa), LSA_AS_ARGS (highest_applied_lsa));
+    }
+
+  if (page_lsa > lowest_unapplied_lsa)
     {
       LOG_TDES *const tdes = LOG_FIND_CURRENT_TDES (thread_p);
       assert (tdes != nullptr && tdes->page_desync_lsa.is_null ());
       tdes->page_desync_lsa = page_lsa;
       er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_PAGE_AHEAD_OF_REPLICATION, 6, PGBUF_PAGE_VPID_AS_ARGS (page),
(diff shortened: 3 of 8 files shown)


src/server/passive_tran_server.hpp
```diff
@@ -37,7 +37,8 @@ class passive_tran_server : public tran_server
     void send_and_receive_stop_log_prior_dispatch ();
 
     /* read replicator's current progress */
-    log_lsa get_replicator_lsa () const;
+    log_lsa get_highest_processed_lsa () const;
+    log_lsa get_min_unapplied_lsa () const;
```
cristiarg: comments

src/server/passive_tran_server.cpp
```diff
@@ -148,9 +148,15 @@ void passive_tran_server::send_and_receive_stop_log_prior_dispatch ()
...
 }
 
-log_lsa passive_tran_server::get_replicator_lsa () const
+log_lsa passive_tran_server::get_highest_processed_lsa () const
 {
-  return m_replicator->get_redo_lsa ();
+  return m_replicator->get_highest_processed_lsa ();
+}
+
+log_lsa passive_tran_server::get_min_unapplied_lsa () const
```
cristiarg: I think it is better to rename to `get_lowest_unapplied_lsa` to make it the obvious counterpart of `highest_processed`

src/transaction/log_replication.cpp
```diff
@@ -438,12 +438,23 @@ namespace cublog
...
   }
 
-  log_lsa replicator::get_redo_lsa () const
+  log_lsa replicator::get_highest_processed_lsa () const
   {
     std::lock_guard<std::mutex> lockg (m_redo_lsa_mutex);
     return m_redo_lsa;
   }
 
+  log_lsa replicator::get_min_unapplied_lsa () const
```
cristiarg: same rename here

src/server/passive_tran_server.cpp
```diff
@@ -148,9 +148,15 @@ void passive_tran_server::send_and_receive_stop_log_prior_dispatch ()
...
+{
+  assert (false);
```
cristiarg: maybe not such a good idea to assert here; let's remove the assert and modify the usage of this function according to spec

src/transaction/log_replication.cpp
```diff
@@ -438,12 +438,23 @@ namespace cublog
...
+  {
+    if (m_parallel_replication_redo == nullptr)
+      {
+	//sync
+	return m_redo_lsa;
+      }
+    //async
+    return m_parallel_replication_redo->get_min_unapplied_log_lsa ();
```
cristiarg: instead of the assert in `log_lsa passive_tran_server::get_min_unapplied_lsa ()`, assert here
when we refactor to have two flavors of the replicator, we'll also implement here
add a TODO comment about that

src/transaction/log_replication.cpp
```diff
@@ -438,12 +438,23 @@ namespace cublog
...
     std::lock_guard<std::mutex> lockg (m_redo_lsa_mutex);
     return m_redo_lsa;
   }
 
+  log_lsa replicator::get_min_unapplied_lsa () const
+  {
+    if (m_parallel_replication_redo == nullptr)
+      {
+	//sync
+	return m_redo_lsa;
```
cristiarg: access to `m_redo_lsa` needs to be synched

src/transaction/log_replication.cpp
```diff
@@ -438,12 +438,23 @@ namespace cublog
     return m_most_recent_trantable_snapshot_lsa.load ();
   }
 
-  log_lsa replicator::get_redo_lsa () const
+  log_lsa replicator::get_highest_processed_lsa () const
```
cristiarg, line 442: in `replicator::wait_past_target_lsa` add a TODO comment that it needs to be refactored to take into account the two flavors once atomic replication is added

src/transaction/log_replication.hpp
```diff
@@ -71,7 +71,8 @@ namespace cublog
       /* wait until replication advances past the target lsa; blocking call */
       void wait_past_target_lsa (const log_lsa &a_target_lsa);
       /* return current progress of the replicator; non-blocking call */
-      log_lsa get_redo_lsa () const;
+      log_lsa get_highest_processed_lsa () const;
+      log_lsa get_lowest_unapplied_lsa () const;
```
cristiarg, line 76: comment

src/storage/page_buffer.c
```diff
@@ -2373,7 +2373,7 @@ pgbuf_check_page_ahead_of_replication (THREAD_ENTRY * thread_p, PAGE_PTR page)
   // Compare page LSA to replication LSA. If page is ahead, there may be a desynchronization issue; its LSA must be
   // saved in transaction descriptor and the appropriate error must be set and returned.
   LOG_LSA page_lsa = *pgbuf_get_lsa (page);
-  LOG_LSA repl_lsa = get_passive_tran_server_ptr ()->get_replicator_lsa ();
+  LOG_LSA repl_lsa = get_passive_tran_server_ptr ()->get_highest_processed_lsa ();
```
cristiarg: I think, `get_lowest_unapplied_lsa`

src/transaction/log_replication.cpp
```diff
@@ -438,12 +439,28 @@ namespace cublog
...
+  log_lsa replicator::get_lowest_unapplied_lsa () const
+  {
+    // TODO: needs to be refactored to work with the new replicators flavors
+    // for now it should not be called
+    assert (false);
+
+    if (m_parallel_replication_redo == nullptr)
+      {
+	//sync
+	std::lock_guard<std::mutex> lockg (m_redo_lsa_mutex);
```
cristiarg: redirect to `get_highest_processed_lsa`

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -719,6 +725,12 @@ namespace cublog
     m_min_unapplied_log_lsa_calculation.wait_past_target_log_lsa (a_target_lsa);
   }
 
+  log_lsa
+  redo_parallel::get_min_unapplied_log_lsa ()
```
cristiarg: might not be needed; please remove

src/transaction/log_replication.cpp
```diff
@@ -438,12 +439,28 @@ namespace cublog
...
+	return m_redo_lsa;
+      }
+    //async
+    return m_parallel_replication_redo->get_min_unapplied_log_lsa ();
```
cristiarg: when atomic replication will be implemted; this function will return the value of atomic replication "minimum unapplied lsa"; therefore `m_parallel_replication_redo->get_min_unapplied_log_lsa` is not needed

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -524,6 +524,12 @@ namespace cublog
     }));
   }
 
+  log_lsa
```
cristiarg: This function is also not needed.
Side note: the function should have been `const`

src/storage/page_buffer.c
```diff
@@ -2352,7 +2352,7 @@ pgbuf_wait_for_replication (THREAD_ENTRY * thread_p, const VPID * optional_vpid_
   assert (tdes != nullptr && !tdes->page_desync_lsa.is_null ());
   pts_ptr->wait_replication_past_target_lsa (tdes->page_desync_lsa);
 
-  const LOG_LSA replication_lsa = pts_ptr->get_replicator_lsa ();
+  const LOG_LSA replication_lsa = pts_ptr->get_highest_processed_lsa ();
```
cristiarg: I actually thing we need to log both replication LSA's here. Just to be on the safe side.
Also, for performance - because both functions [will] lock some mutexes - and, we do not want to gratuitously do than when the code is compiled in release - refactor to:
```
if (prm_get_bool_value (PRM_ID_ER_LOG_DEBUG))
  {
    // get values
    // _er_log_debug(..
  }
```
Note: when the variable `replication_lsa` would not be needed - in release mode - the compiler would usually factor away the code, unless there are side effects; and a mutex lock is a side effect

src/storage/page_buffer.c
```diff
@@ -2373,7 +2373,7 @@ pgbuf_check_page_ahead_of_replication (THREAD_ENTRY * thread_p, PAGE_PTR page)
   // Compare page LSA to replication LSA. If page is ahead, there may be a desynchronization issue; its LSA must be
   // saved in transaction descriptor and the appropriate error must be set and returned.
   LOG_LSA page_lsa = *pgbuf_get_lsa (page);
-  LOG_LSA repl_lsa = get_passive_tran_server_ptr ()->get_replicator_lsa ();
+  LOG_LSA repl_lsa = get_passive_tran_server_ptr ()->get_lowest_unapplied_lsa ();
```
cristiarg: proper variable name here and in the logging below

cristiarg (commented):
Also, remove empty `atomic_replication_helper` files.

src/storage/page_buffer.c
```diff
@@ -2373,14 +2373,24 @@ pgbuf_check_page_ahead_of_replication (THREAD_ENTRY * thread_p, PAGE_PTR page)
   // Compare page LSA to replication LSA. If page is ahead, there may be a desynchronization issue; its LSA must be
   // saved in transaction descriptor and the appropriate error must be set and returned.
   LOG_LSA page_lsa = *pgbuf_get_lsa (page);
-  LOG_LSA repl_lsa = get_passive_tran_server_ptr ()->get_replicator_lsa ();
-  if (page_lsa > repl_lsa)
+  LOG_LSA lowest_unapplied_lsa = get_passive_tran_server_ptr ()->get_lowest_unapplied_lsa ();
+
+  // Print the current replication progress to aid in debug scenarios
+  if (prm_get_bool_value (PRM_ID_ER_LOG_DEBUG))
```
cristiarg: The comment above was not for this function. It was for  `pgbuf_wait_for_replication`.
Please revert this and change the other `pgbuf_wait_for_replication`.

cristiarg (approved):

PR #3449, opened 2022-03-24
[CBRD-24177] Modify to check whether the client is alive in removing archive

joohok:
http://jira.cubrid.org/browse/CBRD-24177

Purpose

As client checking mechanism is added in previous PR (#3433)
mechanism for removing archive logs also need to be modified.

Existing timeout value is handled in utility side, not in removing archive side

Implementation
1. if client is alive, then block archive logs from removing
2. if client is not alive, then archive logs are able to be removed
3. Existing timeout error is dealt in utility side
   -  Use client system parameter (PRM_ID_FLASHBACK_TIMEOUT)
   -  Use poll() in utility to get timeout error during scanf ()
   -  if user doesn't input a transaction id for a TIMEOUT, then it quits and it makes archive logs to be able to be removed

### DIFF
--- a/src/executables/util_cs.c
+++ b/src/executables/util_cs.c
@@ -75,6 +75,10 @@
 	    : (VOL_PURPOSE == DB_TEMPORARY_DATA_PURPOSE) ? "TEMPORARY DATA"     \
 	    : "UNKNOWN")
 
+#if defined(WINDOWS)
+#define STDIN_FILENO  _fileno (stdin)
+#endif
+
 typedef enum
 {
   SPACEDB_SIZE_UNIT_PAGE = 0,
--- a/src/executables/util_cs.c
+++ b/src/executables/util_cs.c
@@ -60,6 +60,7 @@
 #include "log_volids.hpp"
 #include "tde.h"
 #include "flashback_cl.h"
+#include "connection_support.h"
 #if !defined(WINDOWS)
 #include "heartbeat.h"
 #endif
@@ -4922,12 +4923,6 @@ flashback (UTIL_FUNCTION_ARG * arg)
   is_detail = utility_get_option_bool_value (arg_map, FLASHBACK_DETAIL_S);
   is_oldest = utility_get_option_bool_value (arg_map, FLASHBACK_OLDEST_S);
 
-  if (!prm_get_integer_value (PRM_ID_SUPPLEMENTAL_LOG))
-    {
-      fprintf (stderr, "please set \"supplemental_log\" in conf/cubrid.conf\n");
-      goto error_exit;
-    }
-
   /* create table list */
   /* class existence and classoid will be found at server side. if is checked at utility side, it needs addtional access to the server through locator */
   darray = da_create (num_tables, SM_MAX_IDENTIFIER_LENGTH);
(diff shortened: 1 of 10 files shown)


src/transaction/flashback.c
```diff
@@ -121,7 +117,7 @@ flashback_initialize (THREAD_ENTRY * thread_p)
 
   pthread_mutex_lock (&flashback_Conn_lock);
 
-  if (flashback_is_duplicated_request (thread_p))
+  if (flashback_is_connection_alive (true))
```
hornetmj: How about using flashback_is_active() or flashback_is_in_progress()?

src/transaction/flashback.c
```diff
@@ -99,8 +93,10 @@ flashback_is_duplicated_request (THREAD_ENTRY * thread_p)
 	{
 	  /* - flashback_Current_conn is overwritten with new connection, so in_flashback value is initialized to false.
 	   * - previous flashback connection has been exited abnormally. */
-
-	  flashback_reset ();
+	  if (need_reset)
+	    {
+	      flashback_reset ();
```
wrlawodms: How about resetting in `flashback_initialize()` and leaving this function as just a checking function.

src/transaction/flashback.c
```diff
@@ -208,24 +170,23 @@ flashback_check_time_exceed_threshold ()
...
 flashback_is_needed_to_keep_archive ()
 {
-  if (flashback_Min_log_pageid == NULL_LOG_PAGEID)
-    {
-      /* if nothing is set on flashback_Min_log_pageid, then there is no need to keep archive */
+  bool is_needed = false;
 
-      return false;
-    }
+  pthread_mutex_lock (&flashback_Conn_lock);
```
wrlawodms: Coul you tell me the case this lock is needed?

src/executables/util_cs.c
```diff
@@ -5092,9 +5097,23 @@ flashback (UTIL_FUNCTION_ARG * arg)
       goto error_exit;
     }
 
-  printf ("Enter transaction id : ");
-  scanf ("%d", &trid);
+  {
+    POLL_FD input_fd = { STDIN_FILENO, POLLIN | POLLPRI };
+    int timeout = prm_get_integer_value (PRM_ID_FLASHBACK_TIMEOUT);
```
wrlawodms: 1. PRM_ID_FLASHBACK_TIMEOUT is just for flashback utility, not for other clients. So I think it should be a utility option, not a system parameter.
2. One concern is that this is a client-side parameter (or option), so one client can monopolize the flashback with a very long timeout. But this utility is only executed by DBA, so the problem is alleviated.

wrlawodms (commented):
Timeout on the client-side! Clever!

src/transaction/flashback.c
joohok: OK, it will be better.

src/transaction/flashback.c
joohok: OK, I will change it

src/transaction/flashback.c
joohok: At the beginning of the modification, 
a lock was needed because the flashback-related variables were reset at `flashback_is_connection_alive()`. 
However, as that part was changed, the lock was no longer needed.

I'll remove it.

src/executables/util_cs.c
joohok: Thanks for the feedback. 

1. I thought a lot about which of the two options to choose. Whether to give the timeout as an option or to set it as a system parameter. 

But it is weird for me that 
user decide the timeout as an option, 
and the transaction ID has to be entered by user only within that timeout.

It seems right for the system to determine if the situation is a timeout and outputs an error.
So I think it is better to set the timeout as a system parameter.

wrlawodms (approved):

hornetmj (approved):

PR #3450, opened 2022-03-24
[CBRD-24174] Refactor the printing summary information

joohok:
http://jira.cubrid.org/browse/CBRD-24174

Purpose

the output format is refined more neatly, and the following items are added

- add summary start-time ~ end-time
- add total transaction number
- fix the order of classoid
- add the user info when printing summary
- add retry option when user enter the wrong trained
- Only print the summary of log items that have DML
- fix 'press 'q' to quit or press anything to quit' to get character without pressing enter

### DIFF
--- a/src/communication/network_interface_sr.c
+++ b/src/communication/network_interface_sr.c
@@ -10711,7 +10711,7 @@ scdc_end_session (THREAD_ENTRY * thread_p, unsigned int rid, char *request, int 
  * TODO : move to flashback_sr.c
  */
 static int
-flashback_verify_time (THREAD_ENTRY * thread_p, time_t start_time, time_t end_time, LOG_LSA * start_lsa,
+flashback_verify_time (THREAD_ENTRY * thread_p, time_t * start_time, time_t * end_time, LOG_LSA * start_lsa,
 		       LOG_LSA * end_lsa)
 {
   int error_code = NO_ERROR;
@@ -10719,14 +10719,14 @@ flashback_verify_time (THREAD_ENTRY * thread_p, time_t start_time, time_t end_ti
 
 
   /* 1. Check start_time */
-  if (start_time < log_Gl.hdr.db_creation)
+  if (*start_time < log_Gl.hdr.db_creation)
     {
       /* TODO : er_set() */
       return ER_FLASHBACK_INVALID_TIME;
     }
   else
     {
-      ret_time = start_time;
+      ret_time = *start_time;
 
       error_code = cdc_find_lsa (thread_p, &ret_time, start_lsa);
       if (error_code == NO_ERROR || error_code == ER_CDC_ADJUSTED_LSA)
@@ -10734,7 +10734,7 @@ flashback_verify_time (THREAD_ENTRY * thread_p, time_t start_time, time_t end_ti
 	  /* find log record at the time
 	   * ret_time : time of the commit record or dummy record that is found to be greater than or equal to the start_time at first */
 
-	  if (ret_time >= end_time)
+	  if (ret_time >= *end_time)
 	    {
 	      /* out of range : start_time (ret_time) can not be greater than end_time */
 	      /* TODO : er_set() */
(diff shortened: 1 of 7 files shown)


src/communication/network_interface_sr.c
```diff
@@ -10813,13 +10819,22 @@ sflashback_get_summary (THREAD_ENTRY * thread_p, unsigned int rid, char *request
...
 
   FLASHBACK_SUMMARY_CONTEXT context = { LSA_INITIALIZER, LSA_INITIALIZER, NULL, 0, 0, };
 
   time_t start_time = 0;
   time_t end_time = 0;
 
+  char *classname = NULL;
+
+  /* *INDENT-OFF* */
+  std::queue <OID> classoid_queue; //queue to return class OID in the order of the class names received from the user
```
wrlawodms: You don't push to and pop from the queue while running, but just collect and iterate in order. I recommend you to use just `std::vector`. push_back() where `classoid_queue.push (classoid);`, just iterating as you did before when `classoid_queue.front ();`

wrlawodms: I don't remember the context, so I'm asking. Do we need to maintain two data structures containing the same elements, `classoid_queue` and `context.classoid_set`. Can't we just keep one and make the other from it?

src/communication/network_interface_sr.c
```diff
@@ -10890,15 +10905,25 @@ sflashback_get_summary (THREAD_ENTRY * thread_p, unsigned int rid, char *request
...
+
+  num_ptr = ptr;
 
   ptr = or_pack_int (ptr, context.num_summary);
-  ptr = flashback_pack_summary_entry (ptr, context);
+  ptr = flashback_pack_summary_entry (ptr, context, &tmp_num);
+
+  if (context.num_summary != tmp_num)
+    {
+      or_pack_int (num_ptr, tmp_num);
```
wrlawodms, line 10921: 1. I can't find the client-side considering this new packed one.
2. Do `context.num_summary` has to be packed? `tmp_num` seems to be the one reliable.

src/executables/util_cs.c
```diff
@@ -5092,17 +5092,34 @@ flashback (UTIL_FUNCTION_ARG * arg)
...
-  if (summary_entry == NULL)
+  if (summary_info.empty ())
     {
-      /* add message that can not find transaction id */
       goto error_exit;
     }
 
+  while (summary_entry == NULL)
+    {
+      printf ("Enter transaction id ( press -1 to quit) : ");
```
```suggestion
      printf ("Enter transaction id (press -1 to quit) : ");
```
or
```suggestion
      printf ("Enter transaction id ( press -1 to quit ) : ");
```

src/transaction/flashback_cl.c
```diff
@@ -164,21 +233,25 @@ flashback_unpack_and_print_summary (char **summary_buffer, FLASHBACK_SUMMARY_INF
...
 	      da_get (classname_list, idx, classname);
-	      printf ("%s ", classname);
+	      if (j == 0)
+		{
+		  printf ("%s\n", classname);
+		  line_cnt++;
+		}
+	      else
+		{
+		  printf ("%134s\n", classname);
```
wrlawodms: What's this format for?

src/executables/util_cs.c
```suggestion
      printf ("Enter transaction id (press -1 to quit) : ");
```

src/communication/network_interface_sr.c
joohok: I'll use vector to replace queue and set.

src/transaction/flashback_cl.c
joohok: I'll modify the format so that the table names are left-aligned.

src/communication/network_interface_sr.c
joohok: As we discussed offline, num_summary is overwritten by tmp_num

wrlawodms (approved):

hornetmj (approved):

PR #3452, opened 2022-03-25
[CBRD-24254] Implement runtime context class for method (Java SP)

hgryoo:
http://jira.cubrid.org/browse/CBRD-24254


src/method/method_query_cursor.cpp
```diff
@@ -22,62 +22,52 @@
...
     return NO_ERROR;
   }
 
   int
   query_cursor::open ()
   {
-    return qfile_open_list_scan (m_list_id, &m_scan_id);
+    if (m_is_opened == false)
+      {
+	qfile_open_list_scan (m_list_id, &m_scan_id);
```
shparkcubrid, line 56: Error handling seems to be needed. And m_is_opened is always true.

src/method/method_invoke_group.cpp
```diff
@@ -210,7 +228,7 @@ namespace cubmethod
 	  }
```
beyondykk9, line 228: not necessary to check m_connection is null?

src/method/method_invoke_group.cpp
```diff
@@ -219,28 +237,70 @@ namespace cubmethod
...
+      {
+	return;
+      }
+
     reset (true);
 
     get_connection_pool ()->retire (m_connection);
     m_connection = nullptr;
 
-    return error;
```
beyondykk9, line 244: not necessary to check m_connection is null at "retire"?

src/method/method_invoke_java.cpp
```diff
@@ -433,48 +410,48 @@ namespace cubmethod
...
 
-	int i = 0;
-	SCAN_CODE s_code = S_SUCCESS;
-	while (s_code == S_SUCCESS)
+    int i = 0;
+    SCAN_CODE s_code = S_SUCCESS;
+    while (s_code == S_SUCCESS)
+      {
+	s_code = cursor->next_row ();
+	if (s_code == S_END || i > 1000)
```
beyondykk9: 1000 can be defined as a macro and commented the meaning.

src/method/method_query_cursor.cpp
```diff
@@ -22,62 +22,52 @@
...
 
   void
   query_cursor::close ()
   {
-    clear ();
-    qfile_close_scan (m_thread, &m_scan_id);
+    if (m_is_opened)
+      {
+	clear ();
+	qfile_close_scan (m_thread, &m_scan_id);
```
beyondykk9, line 69: not necessary to check error?

src/method/query_method.cpp
```diff
@@ -598,25 +598,22 @@ int xmethod_invoke_fold_constants (THREAD_ENTRY *thread_p, const method_sig_list
...
-  DB_VALUE &res = method_group.get_return_value (0);
+  DB_VALUE &res = method_group->get_return_value (0);
   db_value_clone (&res, &result);
-
-  /* end() will be called at ~method_invoke_group */
-  // method_group.end ();
   return error_code;
 }
-#endif
+#endif
```
ctshim: Please add a newline

src/method/method_invoke_java.cpp
```diff
@@ -582,22 +559,13 @@ namespace cubmethod
...
-	      is_oid_included);
-	  if (cursor == nullptr)
-	    {
-	      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, sizeof (query_cursor));
-	      return ER_OUT_OF_VIRTUAL_MEMORY;
-	    }
-
-	  cursor->open ();
+	  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, sizeof (query_cursor));
+	  return ER_OUT_OF_VIRTUAL_MEMORY;
```
ctshim: Inside the m_group->get_cursor() function, there doesn't seem to be an operation to allocate memory.
How about replacing it with another appropriate error code?

src/method/method_runtime_context.cpp
```diff
@@ -0,0 +1,229 @@
...
+	  {
+	    // m_list_id is going to be destoryed on server-side, so that qlist_count has to be updated
+	    qfile_update_qlist_count (thread_p, query_entry_p->list_id, 1);
+
+	    // store a new cursor in map
+	    cursor = new (std::nothrow) query_cursor (thread_p, query_entry_p, is_oid_included);
+	    m_cursor_map [query_id] = cursor;
+	  }
+      }
+
```
ctshim, line 158: Wouldn't the assert(cursor != nullptr) check be necessary?

src/method/method_runtime_context.cpp
```diff
@@ -0,0 +1,229 @@
...
+    query_cursor *cursor = nullptr;
+
+    // find in map
+    auto search = m_cursor_map.find (query_id);
+    if (search != m_cursor_map.end ())
+      {
+	// found
+	cursor = search->second;
+	if (cursor)
```
hyunikn: It seems that cursor cannot be null at this line. If so, what about assert(cursor) instead of this null check?

src/method/method_runtime_context.cpp
```diff
@@ -0,0 +1,229 @@
...
+	  {
+	    cursor->change_owner (thread_p);
```
hyunikn: [Q] In which case this owner changing happens? (that is, for the same query_id and a thread_p that is different from the thread_p given at the cursor creation time)

src/method/method_runtime_context.cpp
```diff
@@ -0,0 +1,229 @@
...
+  void
+  runtime_context::register_returning_cursor (cubthread::entry *thread_p, QUERY_ID query_id)
+  {
+    if (query_id == NULL_QUERY_ID)
+      {
+	/* do nothing */
+	return;
+      }
+
+    m_returning_cursors.insert (query_id);
```
hyunikn, line 196: [Q] Returning cursors are just inserted at this line and are not processed further. Why are they kept in the set m_returning_cursors?

hyunikn: First of all, there is no call of this method in the whole code.

src/method/method_invoke_java.cpp
```diff
@@ -388,35 +385,15 @@ namespace cubmethod
...
 	      bool is_oid_included = current_result_info.include_oid;
-	      cursor = m_cursor_map[qid] = new (std::nothrow) query_cursor (m_group->get_thread_entry(), qid, is_oid_included);
-	      if (cursor == nullptr)
-		{
-		  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, sizeof (query_cursor));
-		  return ER_OUT_OF_VIRTUAL_MEMORY;
-		}
-
-	      cursor->open ();
+	      query_cursor *cursor = m_group->create_cursor (qid, is_oid_included);
```
hyunikn, line 384: The variable cursor is defined but not referenced.

src/method/method_query_cursor.cpp
shparkcubrid: I don't think this is the problem. Just check it once.

shparkcubrid (approved):

src/method/method_invoke_group.cpp
hgryoo: retire returns immediately if m_connection is null and does nothing

src/method/method_query_cursor.cpp
hgryoo: qfile_close_scan doesn't return any value and there is no error handling for this function as fas as I know

src/method/method_query_cursor.cpp
hgryoo: Thank you for checking it. There is no way to handle errors, I think the list file module will take the error handling.

src/method/method_invoke_group.cpp
hgryoo: If m_connection is nullptr, get_socket() will return INVALID_SOCKET. and the method_invoke's will handle this case as connection failed error.

src/method/method_invoke_java.cpp
hgryoo: I've changed to get `1000` by calling `cursor->get_fetch_count ()`

src/method/method_invoke_java.cpp
hgryoo: the assertion is ommitted. I've added assertion at the following line
```
assert (cursor != nullptr);
```

src/method/query_method.cpp
hgryoo: thank you

src/method/method_runtime_context.cpp
hgryoo: I've added it. thank you.

src/method/method_runtime_context.cpp
hgryoo: When returning a cursor from a procedure (child) to another procedure (parent), the owner of a cursor is changed.

beyondykk9 (approved):

ctshim (approved):

src/method/method_invoke_java.cpp
hgryoo: I've just changed it to return ER_FAILED. get_cursor() doesn't allocate memory.

src/method/method_runtime_context.cpp
hgryoo: Ok, I'm going to add assert here

hyunikn (approved):

youngjinj (approved):

PR #3456, opened 2022-03-28
[CBRD-24257] It prevents the CAS memory size from increasing when a large amount of data is fetched through a query.

airnet73:
http://jira.cubrid.org/browse/CBRD-24257

Purpose
When a large amount of data was imported with a cursor type other than SQL_CURSOR_FORWARD_ONLY among the ODBC cursor types, the CAS (CGW) memory size increased.
It was confirmed that the memory size increases in SQLExecute for MySQL and SQLFetchScroll for Oracle.
When importing large amounts of data, if the memory size of CAS (CGW) exceeds the set memory size, CAS (CGW) restarts. At this time, the driver cannot get the data and an error occurs.

Implementation
N/A

Remarks
N/A

### DIFF
--- a/src/broker/cas_execute.c
+++ b/src/broker/cas_execute.c
@@ -5747,7 +5747,7 @@ cgw_fetch_result (T_SRV_HANDLE * srv_handle, int cursor_pos, int fetch_count, ch
     }
   else if (srv_handle->is_cursor_open && cursor_pos == 1 && srv_handle->cursor_pos > 1)
     {
-      cgw_cursor_close (srv_handle);
+      err_code = cgw_cursor_close (srv_handle);
       if (err_code < 0)
 	{
 	  err_code = ERROR_INFO_SET (db_error_code (), DBMS_ERROR_INDICATOR);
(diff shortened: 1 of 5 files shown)


src/broker/cas_execute.c
```diff
@@ -5738,6 +5735,33 @@ cgw_fetch_result (T_SRV_HANDLE * srv_handle, int cursor_pos, int fetch_count, ch
...
+      err_code = cgw_execute (srv_handle);
+      if (err_code < 0)
+	{
+	  err_code = ERROR_INFO_SET (db_error_code (), DBMS_ERROR_INDICATOR);
+	  goto fetch_error;
+	}
+    }
+  else if (srv_handle->is_cursor_open && cursor_pos == 1 && srv_handle->cursor_pos > 1)
+    {
+      cgw_cursor_close (srv_handle);
```
beyondykk9:
```suggestion
      err_code = cgw_cursor_close (srv_handle);
```

kisoo-han (approved):

beyondykk9 (approved):

PR #3457, opened 2022-03-29
[CBRD-24256] If there is Unicode, the user specified name must be in fully composed form.

youngjinj:
http://jira.cubrid.org/browse/CBRD-24256

- When changing a name containing Unicode to a user specified name, it must be made in a fully composed form again.
  - Otherwise, it is different from the expected value when changing the user specified name to lowercase.
- The following system parameters must be set.
```
unicode_input_normalization=yes
unicode_output_normalization=yes
```
- When parsing a query, the identifier name is changed to a fully composed form in the pt_check_identifier () function.
- In the csql_help_schema () and csql_help_trigger () functions, after normalizing the identifier name in Unicode, it must be changed to the user specified name.
- Fixed an issue when using "*" in the csql session command ";TRIgger ['\*'|trigger-name]".
- Fixed an issue that caused core to occur when exceeding the maximum identifier length in the loaddb utility.
  - Core occurs because the maximum identifier length is not checked when the sm_user_specified_name () function is called in the ldr_find_class () function.
- Data file can be loaded in --CS-mode with --no-user-specified-name option.
- When unloading, if there are "[", "]" in the column name in the ALTER VCLASS statement, use QUOTE to handle the exception.
- Changed SHOW statement to use user specified name as table name.
- Allows the use of user specified names in checkdb, compactdb, loaddb, and unloaddb.


src/object/schema_manager.c
```diff
@@ -2249,23 +2277,88 @@ sm_qualifier_name (const char *name, char *buf, int buf_size)
...
+
+      if (name_size < DB_MAX_USER_LENGTH)
+	{
+	  memcpy (name_copy, name, name_size);
+	  name_copy[name_size] = '\0';
+	}
+      else
+	{
+	  memcpy (name_copy, name, (DB_MAX_USER_LENGTH - 4));
+	  memcpy (name_copy, "...", 4);
```
beyondykk9: it's a proper copy operation ?  maybe, name_copy has only string "..."

src/object/schema_manager.c
```diff
@@ -2249,23 +2277,88 @@ sm_qualifier_name (const char *name, char *buf, int buf_size)
...
+  name_size = STATIC_CAST (int, dot - name);
+  if (name_size >= DB_MAX_USER_LENGTH)
+    {
+      /* exit on error */
+
+      char name_copy[DB_MAX_USER_LENGTH] = { '\0' };
+      int n_bytes = DB_MAX_USER_LENGTH;
+
+      memcpy (name_copy, name, (DB_MAX_USER_LENGTH - 4));
+      memcpy (name_copy, "...", 4);
```
beyondykk9: it's a proper copy operation ? maybe, name_copy has only string "..."

src/parser/parser_support.c
```diff
@@ -10290,6 +10293,31 @@ pt_set_user_specified_name (PARSER_CONTEXT * parser, PT_NODE * node, void *arg,
   user_specified_name = pt_append_string (parser, downcase_resolved_name, ".");
   user_specified_name = pt_append_string (parser, user_specified_name, original_name);
 
+  name_size = (int) strlen (user_specified_name);
```
byungwook-kim, line 10296: name_size should be moved into the next "if" clause.

src/parser/parser_support.c
```diff
@@ -10290,6 +10293,31 @@ pt_set_user_specified_name (PARSER_CONTEXT * parser, PT_NODE * node, void *arg,
...
+
+      composed = STATIC_CAST (char *, (parser_allocate_string_buffer (parser, composed_size + 1, sizeof (char))));
+      if (composed == NULL)
+	{
+	  PT_ERRORf (parser, node, "cannot alloc %d bytes", composed_size + 1);
+	  return NULL;
+	}
+
+      unicode_compose_string (user_specified_name, name_size, composed, &composed_size, &is_composed,
+			      lang_get_generic_unicode_norm ());
```
beyondykk9: no necessary to check "is_composed" is true or false ?

src/object/schema_manager.c
youngjinj: Changed. Thank you.

src/object/schema_manager.c
youngjinj: Changed. Thank you.

shparkcubrid (approved):

hyunikn (approved):

src/parser/parser_support.c
youngjinj: The same code is being used elsewhere.
Maybe check is_composed for stability?

For consistency with other code, I'll keep the check for is_composed.

beyondykk9 (approved):

src/executables/unload_schema.c
```diff
@@ -2399,12 +2399,26 @@ emit_attribute_def (print_output & output_ctx, DB_ATTRIBUTE * attribute, ATTRIBU
...
       {
-	output_ctx ("       %s%s%s ", PRINT_IDENTIFIER (name));
+	if (strchr (name, ']') != NULL)
+	  {
+	    output_ctx ("       %s%s%s ", PRINT_IDENTIFIER_WITH_QUOTE (name));
+	  }
+	else
+	  {
+	    output_ctx ("       %s%s%s ", PRINT_IDENTIFIER (name));
+	  }
```
ctshim: All three cases seem to do the same thing.
How about changing it like below?

```

case CLASS_ATTRIBUTE:
         /*
          * NOTE: The parser no longer recognizes a CLASS prefix for class
          * attributes, this will have been encoded in the surrounding
          * "ADD CLASS ATTRIBUTE" clause
          */
case INSTANCE_ATTRIBUTE:
case SHARED_ATTRIBUTE:
         if (strchr (name, ']') != NULL)
         {
               output_ctx(" %s%s%s ", PRINT_IDENTIFIER_WITH_QUOTE (name)); 
          }
        else
          { 
                output_ctx(" %s%s%s ", PRINT_IDENTIFIER (name));
           }
      break;
```

src/loaddb/load_db.c
```diff
@@ -1157,7 +1157,9 @@ get_loaddb_args (UTIL_ARG_MAP * arg_map, load_args * args)
   args->compare_storage_order = utility_get_option_bool_value (arg_map, LOAD_COMPARE_STORAGE_ORDER_S);
   args->table_name = table_name ? table_name : empty;
   args->ignore_class_file = ignore_class_file ? ignore_class_file : empty;
-  args->no_user_specified_name = utility_get_option_bool_value (arg_map, LOAD_NO_USER_SPECIFIED_NAME_S);
+  /* set to true for testing. When the test is complete, the option should be checked. (by youngjinj) */
+  // args->no_user_specified_name = utility_get_option_bool_value (arg_map, LOAD_NO_USER_SPECIFIED_NAME_S);
+  args->no_user_specified_name = true;
```
ctshim, line 1159: Is the test not complete?
Is it correct to reflect the code like this?

src/executables/unload_schema.c
youngjinj: I will change it like this. Thank you.

src/loaddb/load_db.c
youngjinj: Shell testing is still in progress in QA.

Previously, the system parameter PRM_ID_NO_USER_SPECIFIED_NAME was set to true.
However, changing system parameters affects utilities other than loaddb.
So change it like this.

When the shell test is finished in QA, it will be changed to receive the option value again.

src/object/schema_manager.c
```diff
@@ -2167,10 +2169,35 @@ sm_downcase_name (const char *name, char *buf, int buf_size)
...
+	  memcpy (name_copy, name, name_size);
+	  name_copy[name_size] = '\0';
+	}
+      else
+	{
+	  memcpy (name_copy, name, (MAX_PRINT_SIZE - 4));
+	  memcpy (name_copy + (MAX_PRINT_SIZE - 4), "...", 4);
+	}
+
+      ERROR_SET_ERROR_2ARGS (error, ER_SM_MAX_LENGTH_EXCEEDED_IDENTIFIER, name_copy, n_bytes);
```
ctshim: Is there a length limit for error messages?
I think it can be solved with just two lines like below.

```
int n_bytes = (buf_size <= DB_MAX_IDENTIFIER_LENGTH) ? buf_size : DB_MAX_IDENTIFIER_LENGTH;
ERROR_SET_ERROR_2ARGS (error, ER_SM_MAX_LENGTH_EXCEEDED_IDENTIFIER, name, n_bytes);
```

src/object/schema_manager.c
```diff
@@ -2249,23 +2277,88 @@ sm_qualifier_name (const char *name, char *buf, int buf_size)
...
+    {
+      /* exit on error */
+
+      char name_copy[DB_MAX_USER_LENGTH] = { '\0' };
+      int n_bytes = DB_MAX_USER_LENGTH;
+
+      memcpy (name_copy, name, (DB_MAX_USER_LENGTH - 4));
+      memcpy (name_copy + (DB_MAX_USER_LENGTH - 4), "...", 4);
+
+      ERROR_SET_ERROR_2ARGS (error, ER_SM_MAX_LENGTH_EXCEEDED_USER, name_copy, n_bytes);
```
ctshim: Is there a length limit for error messages?
I think it can be solved with just one lines like below.

ERROR_SET_ERROR_2ARGS (error, ER_SM_MAX_LENGTH_EXCEEDED_USER, name, B_MAX_USER_LENGTH);

src/loaddb/load_server_loader.cpp
```diff
@@ -114,8 +114,171 @@ namespace cubload
...
+	      if (error != NO_ERROR)
+		{
+		  ASSERT_ERROR ();
+		  heap_attrinfo_end (&thread_ref, &attr_info);
+		  break;
+		}
+
+	      if (string != NULL && strcmp ("name", string) == 0)
+		{
+		  attr_idx = i;
```
ctshim: The code is written on the assumption that "name" must be present among att_names.
Is that right? Otherwise, there will be a problem on line 242.

youngjinj: The db_user system class has a "name" column.
So there should never be a case where attr_idx is not set.

But even if it is not set, the code below can be executed, so does the defense code need to be added?

ctshim: If it's guaranteed, there's no need to fix it.
But, If it is a fixed system table, wouldn't it be possible to define idx as a constant value without looking for "name"?

src/parser/parser_support.c
youngjinj: After analyzing with @ctshim, I found that this part of the code was unnecessary.
So, the code that normalizes Unicode has been removed here.

src/object/schema_manager.c
```diff
@@ -2207,11 +2232,15 @@ sm_user_specified_name (const char *name, char *buf, int buf_size)
        *        name: system_class_name
        *      return: system_class_name
        */
+      if (sm_check_qualifier_name_size (name) == false)
+	{
+	  ASSERT_ERROR ();
+	  return NULL;
+	}
+
```
ctshim: The sm_check_qualifier_name_size() function is only called here in one place. If you remove all unnecessary tasks, you will be able to replace them like this:
  If you modify it as below, the sm_check_qualifier_name_size() function is no longer needed.

```
/* If the name is already a user-specified name or a system class name, do not recreate it. */
char* dot = strchr (name, '.');  
if ( dot)
{  
     /*
       * e.g.   name: user_name.object_name
       *      return: user_name.object_name
       * 
       *      or
       * 
       *        name: system_class_name
       *      return: system_class_name
       */
   if ( STATIC_CAST (int, dot - name) >= DB_MAX_USER_LENGTH)
	{
	  ASSERT_ERROR ();
	  return NULL;
	}
	return sm_downcase_name (name, buf, buf_size);
}
else if ( sm_check_system_class_by_name (name))
{  
	return sm_downcase_name (name, buf, buf_size);
}

/* Appends the current username, making it a user-specified name. */
...
```

src/object/schema_manager.c
```diff
@@ -2158,6 +2158,8 @@ sm_check_name (const char *name)
 char *
 sm_downcase_name (const char *name, char *buf, int buf_size)
 {
+#define MAX_PRINT_SIZE 64
+  int name_size = 0;
```
ctshim: Is there any need to check the length in sm_downcase_name() function? In my opinion, an existing function without a length check would suffice.

src/object/schema_manager.c
youngjinj: I changed it as you said. Thank you.

src/object/schema_manager.c
youngjinj: As you said, I removed this code because the function to check the length was run before.
Thank you.

src/object/schema_manager.c
youngjinj: As you said, I removed this code because the function to check the length was run before.
Thank you.

src/object/schema_manager.c
youngjinj: As you said, I removed this code because the function to check the length was run before.
Thank you.

src/loaddb/load_server_loader.cpp
youngjinj: ATTR_IDX is fixed to 7 unless the column order in the db_user table is changed or a new column is added after the "name" column.

I will improve it by adding the macro variable below.
```
#define CATCLS_USER_ATTR_IDX_NAME 7
```

And if a change occurs in the code for adding a column to the db_user system table, I will leave a comment requesting to change CATCLS_USER_ATTR_IDX_NAME together.

src/parser/parser_support.c
```diff
@@ -10187,6 +10187,8 @@ pt_set_user_specified_name (PARSER_CONTEXT * parser, PT_NODE * node, void *arg,
   char downcase_resolved_name[DB_MAX_USER_LENGTH] = { '\0' };
   char current_user_name[DB_MAX_USER_LENGTH] = { '\0' };
   const char *user_specified_name = NULL;
+  int name_size = 0;
+  int composed_size = 0;
```
hgryoo: unused variables

src/loaddb/load_server_loader.cpp
```diff
@@ -114,8 +114,171 @@ namespace cubload
...
+		ASSERT_ERROR ();
+		heap_scancache_end (&thread_ref, &scan_cache);
+		heap_attrinfo_end (&thread_ref, &attr_info);
+		return LC_CLASSNAME_ERROR;
+	      }
+	  }
+
+	heap_scancache_end (&thread_ref, &scan_cache);
+	heap_attrinfo_end (&thread_ref, &attr_info);
+      }
```
hgryoo: please extract this code to a private function. If the public function is too long, It can be difficult to understand how this class works.

hgryoo (approved):

src/parser/parser_support.c
youngjinj: I will remove it. Thank you.

src/loaddb/load_server_loader.cpp
youngjinj: Added the locate_class_for_all_users () function to separate it.
Thank you.

ctshim (approved):

PR #3458, opened 2022-03-29
[CBRD-24183] Rewrite optimization for limit clause of union

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-24183

The limit clause of mainquery of union is pushed to subquery of union.
If the subquery of union has order_by or group_by, limit is changed to order_by_num or group_by_num. it is existing routines.

Restriction
```
main query has order_by
union
```
NO_PUSH_PRED sql hint : don't push limit to subquery of union

### DIFF
--- a/src/optimizer/query_rewrite.c
+++ b/src/optimizer/query_rewrite.c
@@ -7851,7 +7851,10 @@ qo_optimize_queries (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int *co
 		{
 		  (void) mq_copypush_sargable_terms (parser, node, spec);
 		}
-	      point_list = parser_append_previous_node (pt_point (parser, spec), point_list);
+	      else
+		{
+		  point_list = parser_append_previous_node (pt_point (parser, spec), point_list);
+		}
 	      spec = spec->next;
 	    }
 	  /* reduce unnecessary tables */
--- a/src/optimizer/query_rewrite.c
+++ b/src/optimizer/query_rewrite.c
@@ -7222,9 +7222,6 @@ qo_check_distinct_union (PARSER_CONTEXT * parser, PT_NODE * node)
 
   switch (node->node_type)
     {
-    case PT_SELECT:
-      break;
-
     case PT_UNION:
       if (node->info.query.all_distinct == PT_DISTINCT)
 	{
@@ -7393,7 +7390,7 @@ qo_optimize_queries (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int *co
 
 	      /* order by, union check */
 	      if (node->info.query.order_by == NULL && !qo_check_distinct_union (parser, node)
-		  && !node->info.query.q.select.hint & PT_HINT_NO_PUSH_PRED)
+		  && !(node->info.query.q.select.hint & PT_HINT_NO_PUSH_PRED))
 		{
 		  node = qo_push_limit_to_union (parser, node, limit_node);
 		}
--- a/src/optimizer/query_rewrite.c
+++ b/src/optimizer/query_rewrite.c
@@ -7394,9 +7394,9 @@ qo_optimize_queries (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int *co
 	      /* order by, union check */
 	      if (node->info.query.order_by == NULL && !qo_check_distinct_union (parser, node)
 		  && !node->info.query.q.select.hint & PT_HINT_NO_PUSH_PRED)
-	        {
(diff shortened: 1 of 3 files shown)


src/optimizer/query_rewrite.c
```diff
@@ -6906,6 +7391,12 @@ qo_optimize_queries (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int *co
 	      single_tuple_bak = node->info.query.flag.single_tuple;
 	      node->info.query.flag.single_tuple = false;
 
+	      /* order by, union check */
+	      if (node->info.query.order_by == NULL && !qo_check_distinct_union (parser, node)
+		  && !node->info.query.q.select.hint & PT_HINT_NO_PUSH_PRED)
```
ctshim: Shouldn't it be modified as follows in consideration of operator priotity?

 if (node->info.query.order_by == NULL && !qo_check_distinct_union (parser, node)
		  && **!(node->info.query.q.select.hint & PT_HINT_NO_PUSH_PRED)**)

src/optimizer/query_rewrite.c
```diff
@@ -7174,6 +7174,73 @@ qo_move_on_clause_of_explicit_join_to_where_clause (PARSER_CONTEXT * parser, PT_
...
+      break;
+
+    case PT_UNION:
+      if (node->info.query.all_distinct == PT_DISTINCT)
+	{
+	  return true;
+	}
+
+      result |= qo_check_distinct_union (parser, node->info.query.q.union_.arg1);
+      result |= qo_check_distinct_union (parser, node->info.query.q.union_.arg2);
```
ctshim, line 7232: If any part is true, all will be true.
How about doing something like below?

```
if(qo_check_distinct_union (parser, node->info.query.q.union_.arg1))
  {
    return true;
  }
return qo_check_distinct_union (parser, node->info.query.q.union_.arg2);
```

src/optimizer/query_rewrite.c
```diff
@@ -7174,6 +7174,73 @@ qo_move_on_clause_of_explicit_join_to_where_clause (PARSER_CONTEXT * parser, PT_
...
+static PT_NODE *
+qo_push_limit_to_union (PARSER_CONTEXT * parser, PT_NODE * node, PT_NODE * limit)
+{
+  switch (node->node_type)
+    {
+    case PT_SELECT:
+      if (!pt_has_inst_in_where_and_select_list (parser, node))
+	{
+	  node->info.query.flag.rewrite_limit = 1;
+	  node->info.query.limit = limit;
```
ctshim, line 7194: Isn't it necessary to guarantee that node->info.query.limit is null when incorporated here?
`assert(node->info.query.limit == NULL);`

src/optimizer/query_rewrite.c
beyondykk9: It is better to modify it considering operator precedence.

src/optimizer/query_rewrite.c
```diff
@@ -6756,6 +7174,73 @@ qo_move_on_clause_of_explicit_join_to_where_clause (PARSER_CONTEXT * parser, PT_
...
+ *   continue_walk(in):
+ */
+static bool
+qo_check_distinct_union (PARSER_CONTEXT * parser, PT_NODE * node)
+{
+  bool result = false;
+
+  switch (node->node_type)
+    {
+    case PT_SELECT:
```
hyunikn: It seems that PT_SELECT case is covered by the default case and can be erased.

hyunikn (approved):

beyondykk9 (approved):

src/optimizer/query_rewrite.c
shparkcubrid: I'll fix it

src/optimizer/query_rewrite.c
shparkcubrid: It is checked in pt_has_inst_in_where_and_select_list (). limit is changed to instnum on semantic check.

src/optimizer/query_rewrite.c
shparkcubrid: Okay. I'll fix it.

src/optimizer/query_rewrite.c
shparkcubrid: You are right, but I want to keep this code style.

ctshim (approved):

youngjinj (approved):

hgryoo (approved):

PR #3466, opened 2022-03-31
[CBRD-24176] Support escape character for single quotation in string value

joohok:
http://jira.cubrid.org/browse/CBRD-24176

Purpose

This is to insert an escape character to process a single quotation character. 
In order to use single quotation in the string value, we need to add escape (\') character.

if user wants to insert a value `aa'aa`, then SQL for it is as below
`INSERT INTO tbl values (' aa''aa ');`

And flashback should provide SQL statement like above. 

Implementation

Used existing csql function (string_to_string ()) which is used for making raw string value to executable string.
- string_value : original string value (aa'aa)
- delimiter : string enclosure ('\'')
- string_introducer : character that must come at the beginning of the string (e.g `X'1001'`,` B'1010'`)
- plain string : string that can contain '\n', '\t', ...
- change_single_quote : if there is a single quote in the string, then change it from `'\''` to `'\'\''`

### DIFF
--- a/src/transaction/flashback_cl.c
+++ b/src/transaction/flashback_cl.c
@@ -430,6 +430,8 @@ flashback_process_column_data (char **data, char **sql, int *max_sql_size, DB_TY
 	  error = flashback_check_and_resize_sql_memory (sql, sql_length + result_length + 1, max_sql_size);
 	  if (error != NO_ERROR)
 	    {
+	      free_and_init (result_string);
+
 	      return error;
 	    }
 
@@ -792,7 +794,7 @@ flashback_print_delete (char **loginfo, int trid, char *user, const char *classn
     {
       /* check SQL length
        * sql + length of ", " */
-      error = flashback_check_and_resize_sql_memory (&sql, strlen (sql) + 3, &max_sql_size);
+      error = flashback_check_and_resize_sql_memory (&sql, strlen (sql) + 4, &max_sql_size);
       if (error != NO_ERROR)
 	{
 	  goto error;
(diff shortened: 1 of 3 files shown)


src/transaction/flashback_cl.c
```diff
@@ -774,7 +805,7 @@ flashback_print_delete (char **loginfo, int trid, char *user, const char *classn
 	}
       else
 	{
-	  strcat (sql, " )");
+	  strcat (sql, " );");
```
wrlawodms: Doesn't this affect the length of this sql?

src/transaction/flashback_cl.c
```diff
@@ -978,7 +1009,7 @@ flashback_print_insert (char **loginfo, int trid, char *user, const char *classn
 	    }
 	  else
 	    {
-	      strcat (original_sql, " )");
+	      strcat (original_sql, " );");
```
wrlawodms: The same above goes for this.

src/transaction/flashback_cl.c
```diff
@@ -412,18 +414,47 @@ flashback_process_column_data (char **data, char **sql, int *max_sql_size, DB_TY
...
+	    {
+	      /* internally stirng_to_string() allocates memory for string */
+	      util_log_write_errid (MSGCAT_UTIL_GENERIC_NO_MEM);
+	      return ER_OUT_OF_VIRTUAL_MEMORY;
+	    }
+
+	  error = flashback_check_and_resize_sql_memory (sql, sql_length + result_length + 1, max_sql_size);
+	  if (error != NO_ERROR)
+	    {
+	      return error;
```
wrlawodms, line 435:
```suggestion
              free_and_init (result_string);
	      return error;
```

joohok: Oh.. Thank you

src/transaction/flashback_cl.c
joohok: yes, 
`error = flashback_check_and_resize_sql_memory (&sql, strlen (sql) + 3, &max_sql_size);`

it means the maximum value among the lengths of the two strings (", " or " );" ), and 3 is sufficient.

Oh.. no, it is not sufficient, I will change the `strlen (sql) + 3` to `strlen (sql) + 4` . 

Thank you

hornetmj: Or remove space in front of ).

hornetmj (approved):

eido5 (approved):

wrlawodms (approved):

PR #3468, opened 2022-03-31
[CBRD-24173] Fix bugs occurred when getting summary information with default option

joohok:
http://jira.cubrid.org/browse/CBRD-24173

Purpose

Fix bugs when
1. Get flashback summary with a default option
  - cubrid createdb testdb en_US 
  - cubrid flashback testdb
  -> start_date is set to 10 minutes before current
  -> then start_date can be set earlier than db_creation
  -> So, if start_date is earlier than db_creation, then set start_date to db_creation. 

2. Get transaction ID as value ​​other than integer
  - Since the input buffer is not flushed by scanf, scanf can continue to run even if there is no input.
  -> if an error occurs in scanf, then flushed an input buffer

3. Missed argument for "output file" error message

### DIFF
--- a/src/executables/util_cs.c
+++ b/src/executables/util_cs.c
@@ -4357,7 +4357,7 @@ flashback (UTIL_FUNCTION_ARG * arg)
 
       if (poll (&input_fd, 1, timeout * 1000))
 	{
-	  if (!scanf ("%d", &trid))
+	  if (scanf ("%d", &trid) != 1)
 	    {
 	      /* When non integer value is input, the input buffer must be flushed. */
 	      clean_stdin ();
--- a/src/executables/util_cs.c
+++ b/src/executables/util_cs.c
@@ -4361,6 +4361,8 @@ flashback (UTIL_FUNCTION_ARG * arg)
 	    {
 	      /* When non integer value is input, the input buffer must be flushed. */
 	      clean_stdin ();
+
+	      continue;
 	    }
 	}
       else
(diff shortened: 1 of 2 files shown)


src/executables/util_cs.c
```diff
@@ -4336,12 +4350,20 @@ flashback (UTIL_FUNCTION_ARG * arg)
...
     {
+      POLL_FD input_fd = { STDIN_FILENO, POLLIN | POLLPRI, 0 };
+
       printf ("Enter transaction id (press -1 to quit): ");
       fflush (stdout);
 
       if (poll (&input_fd, 1, timeout * 1000))
 	{
-	  scanf ("%d", &trid);
+	  if (!scanf ("%d", &trid))
```
eido5: When `if` condition has  -1(EOF), does it work normally?

joohok: Thanks for feedback,

I didn't consider that users can input EOF with ctrl+D. 

I will apply your feedback

eido5 (approved):

hornetmj (approved):

wrlawodms (approved):
When a user gives an input time earlier than the database creation time, it's also set to the creation time. In this case, It would be good to notify the user that the start time is replaced. Or, it should be described in the usage or manual.

joohok:
> When a user gives an input time earlier than the database creation time, it's also set to the creation time. In this case, It would be good to notify the user that the start time is replaced. Or, it should be described in the usage or manual.

Yes, you are right.
For now, adjusted time is shown to user when summary information is printed.
(e.g. `Start date - End date: 01-04-2022:09:43:00 - 01-04-2022:09:53:00` )

PR #3469, opened 2022-04-01
[CBRD-24258] Rewrite optimization for oracle style partial range scan

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-24258

Implementation
case1 : oracle style limit
Modify so that view merge is possible for cases that satisfy all of the conditions below.
1. No join (only spec)
2. Only rownum related predicate exists in where clause

If subquery has order by, change rownum to orderby_num.

TO_DO
UNION query: Since union query can not be view merged, need to add a new routine.

case2 : push limit
The limit clause of mainquery of union is pushed to subquery of union.
If the subquery of union has order_by or group_by, limit is changed to order_by_num or group_by_num. it is existing routines.

'limit 10,10' is changed to 'limit 10+10' and it is pushed to subquery.

Restriction
- main query has order_by
- union

NO_PUSH_PRED sql hint : don't push limit to subquery of union


src/optimizer/query_graph.c
```diff
@@ -8699,6 +8704,14 @@ qo_discover_sort_limit_nodes (QO_ENV * env)
   return;
 
 abandon_stop_limit:
+  /* remove error which is occured in this function */
+  if (pt_has_error (QO_ENV_PARSER (env)))
+    {
+      pt_reset_error (QO_ENV_PARSER (env));
+      (QO_ENV_PARSER (env))->flag.has_internal_error = 0;
+    }
+
```
ctshim, line 8758: On lines 8617 and 8626, is not bitset_delset() necessary for expr_segs and tmp_bitset variables?

src/optimizer/query_graph.c
```diff
@@ -8699,6 +8704,14 @@ qo_discover_sort_limit_nodes (QO_ENV * env)
   return;
 
 abandon_stop_limit:
+  /* remove error which is occured in this function */
+  if (pt_has_error (QO_ENV_PARSER (env)))
+    {
+      pt_reset_error (QO_ENV_PARSER (env));
```
ctshim, line 8755: In line 8617, assert_release() is executed and the error setting is removed even though it is coming?
Or is it ignoring the case because it is unlikely to happen?

shparkcubrid: This routine checks if sort limit optimization is possible. If an error occurs in this function, only sort limit optimization should be disabled, and it should not have any other effect.

src/optimizer/query_graph.c
shparkcubrid: No. Since it is not allocated, bitset_delset() is not needed.

src/parser/parse_tree.h
```diff
@@ -467,6 +467,17 @@ struct json_t;
...
+#define PT_IS_EXPR_NODE_WITH_COMP_OP(n) \
+        ( (PT_IS_EXPR_NODE (n)) && \
+          ((n)->info.expr.op == PT_EQ || \
+           (n)->info.expr.op == PT_GE || \
+           (n)->info.expr.op == PT_GT || \
+           (n)->info.expr.op == PT_LT || \
+           (n)->info.expr.op == PT_LE || \
+           (n)->info.expr.op == PT_GT_INF || \
+           (n)->info.expr.op == PT_LT_INF || \
+           (n)->info.expr.op ==PT_RANGE ))
```
hgryoo: trivial :)
```suggestion
#define PT_IS_EXPR_NODE_WITH_COMP_OP(n) \
        ( (PT_IS_EXPR_NODE (n)) && \
          ((n)->info.expr.op == PT_EQ || \
           (n)->info.expr.op == PT_GE || \
           (n)->info.expr.op == PT_GT || \
           (n)->info.expr.op == PT_LT || \
           (n)->info.expr.op == PT_LE || \
           (n)->info.expr.op == PT_GT_INF || \
           (n)->info.expr.op == PT_LT_INF || \
           (n)->info.expr.op == PT_RANGE ))
```

shparkcubrid: thanks.

beyondykk9 (approved):

ctshim (approved):

hgryoo (approved):

youngjinj (approved):

hyunikn (approved):

src/parser/view_transform.c
```diff
@@ -358,6 +359,7 @@ static const char *get_authorization_name (DB_AUTH auth);
 static PT_NODE *mq_add_dummy_from_pre (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int *continue_walk);
 static PT_NODE *mq_update_order_by (PARSER_CONTEXT * parser, PT_NODE * statement, PT_NODE * query_spec,
 				    PT_NODE * class_, PT_NODE * derived_spec);
+static PT_NODE *mq_update_orderby_for (PARSER_CONTEXT * parser, PT_NODE * statement);
```
youngjinj: It is trivial.

mq_update_orderby_for() is a newly added function, but it is not used.
The comment says it needs to be removed, so why was it added to the PR?

shparkcubrid: I'll check it. thanks.

shparkcubrid: I removed mq_update_orderby_for ()

PR #3473, opened 2022-04-01
[CBRD-23981] fix the replication failures when the execution of bulk-insert by the loaddb on the CS-mode

hornetmj:
http://jira.cubrid.org/browse/CBRD-23981

### Purpose
* fix a bug that a bulk-insert by loaddb utility with the CS-mode makes a coredump on the slave node.
* this bug has occurred because there is no room for the mvcc insert id on the slave node.

### Implementation
* change the recdes from the non-mvcc format to the mvcc format on the applylogdb
* if the log record type is the RVHF_INSERT and it relates to the class supporting the mvcc, then the recdes is converted to the mvcc format

### Remarks
N/A


eido5 (approved):
Looks so good. Arguments are inserted correct as the pointer.

wrlawodms (commented):
Have you considered to handle this issue when generating the log records? This seems to  occur only when bulk-inserting with HA-mode, so that it seems better to handle it where it causes if it's not complicated. Let me have your opinion. 

For example (but, I'm not sure this is the proper location to handle it):
in `heap_insert_adjust_recdes_header()`:

19789       if (is_mvcc_class && **!(insert_context->is_bulk_op && !log_does_allow_replication())** )
19790         {
19791           /* get MVCC id */
19792           mvcc_id = logtb_get_current_mvccid (thread_p);
19793
19794           /* set MVCC INSID if necessary */
19795           if (!MVCC_IS_FLAG_SET (&mvcc_rec_header, OR_MVCC_FLAG_VALID_INSID))
19796             {
19797               MVCC_SET_FLAG (&mvcc_rec_header, OR_MVCC_FLAG_VALID_INSID);
19798               record_size += OR_MVCCID_SIZE;
19799             }
19800           MVCC_SET_INSID (&mvcc_rec_header, mvcc_id);
19801         }

hornetmj:
@wrlawodms
Your suggestion act like the following and I prospect it works well(But, needs more cosideration). But I don't want to make any vacuum job on the master node when executing the bulk operations. So, I choose my approach.

Non-HA configuration:
* insert_context->is_bulk_op == true -> non-mvcc recdes
* insert_context->is_bulk_op == false -> mvcc recdes

HA configuration: on the master (HA_SERVER_STATE_ACTIVE)
* insert_context->is_bulk_op == true -> mvcc recdes (due to !log_does_allow_replication() always false)
* insert_context->is_bulk_op == false -> mvcc recdes

HA configuration: on the slave (insert_context->is_bulk_op always false)
* mvcc recdes

wrlawodms (approved):
I didn't consider the vacuum. Thank you.

PR #3484, opened 2022-04-06
[LETS-422] Atomic replication helper class

adrianbrab:
http://jira.cubrid.org/browse/LETS-422

A new helper class that is able to hold log records information for all log records that pertain to an atomic redo sequence - which is, in itself, part of a single transaction - and apply them.

* Implementation:

Created 2 new classes `atomic_replication_helper` and `atomic_replication_unit`. The `atomic_replication_unit` holds the information about one log record and exposes methods to fix and unfix the page and apply the redo. The unit class is a private internal class of the `atomic_replication_helper`, the helper class holds the map of `TRANID` to  `atomic_replication_unit`.

The classes will be used in the new replicator flavor on the passive transaction server in the upcoming issues.


src/transaction/atomic_replication_helper.cpp
```diff
@@ -0,0 +1,184 @@
...
+   *********************************************************************/
+
+  void atomic_replication_helper::add_atomic_replication_unit (THREAD_ENTRY *thread_p, PGBUF_WATCHER *watcher,
+      TRANID tranid, log_lsa record_lsa, log_rectype record_type, VPID vpid)
+  {
+    auto sequence = m_atomic_sequences_map.find (tranid);
+    if (sequence == m_atomic_sequences_map.end ())
+      {
+	std::deque<atomic_replication_unit> new_sequence;
+	m_atomic_sequences_map.insert ({tranid, new_sequence});
```
cristiarg: insert will return an interator to the added element; no need for extra find

src/transaction/atomic_replication_helper.cpp
```diff
@@ -0,0 +1,184 @@
...
+  {
+    // call log apply function
+  }
+
+  void atomic_replication_helper::atomic_replication_unit::fix_page (THREAD_ENTRY *thread_p,
+      PGBUF_WATCHER *pg_watcher)
+  {
+    switch (m_record_type)
+      {
+      case LOG_REDO_DATA:
```
cristiarg: I do not think this is the criteria to use for differentiating between ordered fix and regular fix; rather `rcvindex` is:
```
  RVHF_INSERT = 37,
  RVHF_MVCC_INSERT = 43,
  RVHF_DELETE = 38,
  RVHF_MVCC_DELETE_REC_HOME = 44,
  RVHF_MVCC_DELETE_OVERFLOW = 45,
  RVHF_MVCC_DELETE_REC_NEWHOME = 46,
  RVHF_MVCC_DELETE_MODIFY_HOME = 47,
  RVHF_UPDATE = 39,
  RVHF_MVCC_UPDATE_OVERFLOW = 52,
```

src/transaction/atomic_replication_helper.hpp
```diff
@@ -0,0 +1,68 @@
...
+	private:
+	  log_lsa m_record_lsa;
+	  log_rectype m_record_type;
+	  VPID m_vpid;
+	  PAGE_PTR m_page_ptr;
+	  TRANID m_record_tranid;
+      };
+
+      //Hashmap
+      std::map<TRANID, std::deque<atomic_replication_unit>> m_atomic_sequences_map;
```
cristiarg: use some typedefs here
maybe a `std::vector` instead of deque? because the access is faster and we don't need to continuosly add and remove elements; rather, we add elements one by one; and remove all of them at the end - eg: iteration in `unfix_atomic_replication_sequence` does not need to pop one by one

src/transaction/atomic_replication_helper.hpp
```diff
@@ -0,0 +1,68 @@
...
+      bool is_part_of_atomic_replication (TRANID tranid) const;
+#if !defined (NDEBUG)
+      bool is_page_part_of_atomic_replication_sequence (TRANID tranid, VPID vpid) const;
+#endif
+
+    private:
+      class atomic_replication_unit
+      {
+	public:
+	  atomic_replication_unit (log_lsa lsa, log_rectype rectype, VPID vpid, TRANID record_tranid);
```
cristiarg: delete/default/implement copy/move ctors/ops

src/transaction/atomic_replication_helper.hpp
```diff
@@ -0,0 +1,68 @@
...
+#include "log_record.hpp"
+#include "page_buffer.h"
+#include "thread_entry.hpp"
+
+namespace cublog
+{
+
+  class atomic_replication_helper
+  {
+    public:
```
cristiarg, line 39: delete/default/implement copy/move ctors/ops

src/transaction/atomic_replication_helper.cpp
```diff
@@ -0,0 +1,184 @@
...
+  void atomic_replication_helper::unfix_atomic_replication_sequence (THREAD_ENTRY *thread_p, PGBUF_WATCHER *pg_watcher,
+      TRANID tranid)
+  {
+    auto sequence = m_atomic_sequences_map.find (tranid);
+    if (sequence == m_atomic_sequences_map.end ())
+      {
+	return;
+      }
+
+    std::deque<atomic_replication_unit> sequence_q = sequence->second;
```
cristiarg: adress of

src/transaction/atomic_replication_helper.cpp
```diff
@@ -0,0 +1,184 @@
...
+    if (vpid.pageid != vpid_Null_vpid.pageid && vpid.volid != vpid_Null_vpid.volid
+	&& is_page_part_of_atomic_replication_sequence (tranid, vpid))
+      {
+	// page is already part of atomic replication
+	return;
+      }
+#endif
+
+    atomic_replication_unit atomic_unit (record_lsa, record_type, vpid, tranid);
+    atomic_unit.fix_page (thread_p, watcher);
```
cristiarg: `watcher` should probably be a member of the unit

src/transaction/atomic_replication_helper.cpp
```diff
@@ -0,0 +1,193 @@
...
+   *********************************************************************/
+
+  template <typename T>
+  void atomic_replication_helper::add_atomic_replication_unit (THREAD_ENTRY *thread_p, TRANID tranid, log_lsa record_lsa,
+      LOG_RCVINDEX rcvindex, VPID vpid, log_rv_redo_context &redo_context, const log_rv_redo_rec_info<T> &record_info)
+  {
+    auto sequence = m_atomic_sequences_map.find (tranid);
+    if (sequence == m_atomic_sequences_map.end ())
+      {
+	std::vector<atomic_replication_unit> new_sequence;
```
cristiarg: `new_sequence` can be constructed in place so only the move ctor is used

src/transaction/atomic_replication_helper.cpp
```diff
@@ -0,0 +1,193 @@
...
+	sequence = m_atomic_sequences_map.insert ({tranid, new_sequence});
+      }
+
+#if !defined (NDEBUG)
+    if (vpid.pageid != NULL_PAGEID && vpid.volid != NULL_VOLID
```
cristiarg: use `VPID_ISNULL`

src/transaction/atomic_replication_helper.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+    atomic_replication_unit atomic_unit (record_lsa, vpid, rcvindex);
+    atomic_unit.fix_page (thread_p);
+    sequence->second.emplace_back (atomic_unit);
+
+    atomic_unit.apply_log_redo (thread_p, redo_context, record_info);
+  }
+
+#if !defined (NDEBUG)
+  bool atomic_replication_helper::is_page_part_of_atomic_replication_sequence (TRANID tranid, VPID vpid) const
```
cristiarg: This function should perform checks from the `vpid` point of view.
It should check for the scenario:
- one page is only part of one single atomic replication

As it is implemented it only it only answers to the question: is the vpid part of at least one atomic replication?

src/transaction/atomic_replication_helper.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+    return true;
+  }
+
+  void atomic_replication_helper::unfix_atomic_replication_sequence (THREAD_ENTRY *thread_p, TRANID tranid)
+  {
+    auto sequence = m_atomic_sequences_map.find (tranid);
+    if (sequence == m_atomic_sequences_map.end ())
+      {
+	return;
```
cristiarg, line 94: assert false

src/transaction/atomic_replication_helper.cpp
```diff
@@ -0,0 +1,193 @@
...
+   ****************************************************************************/
+
+  atomic_replication_helper::atomic_replication_unit::atomic_replication_unit (log_lsa lsa, VPID vpid,
+      LOG_RCVINDEX rcvindex)
+    : m_record_lsa { lsa }
+    , m_vpid { vpid }
+    , m_record_index { rcvindex }
+  {
+    assert (lsa != NULL_LSA);
+    PGBUF_INIT_WATCHER (&m_watcher, PGBUF_ORDERED_HEAP_NORMAL, PGBUF_ORDERED_NULL_HFID);
```
cristiarg, line 123: `PGBUF_CLEAR_WATCHER` in dtor

src/transaction/atomic_replication_helper.cpp
```diff
@@ -0,0 +1,193 @@
...
+    m_page_ptr == nullptr;
```
cristiarg: init in initialization list

src/transaction/atomic_replication_helper.hpp
```diff
@@ -0,0 +1,89 @@
...
+#include "log_lsa.hpp"
+#include "log_record.hpp"
+#include "log_recovery_redo.hpp"
+#include "page_buffer.h"
+#include "thread_entry.hpp"
+
+namespace cublog
+{
+
+  class atomic_replication_helper
```
cristiarg, line 37: comment

src/transaction/atomic_replication_helper.hpp
```diff
@@ -0,0 +1,89 @@
...
+      void add_atomic_replication_unit (THREAD_ENTRY *thread_p, TRANID tranid, log_lsa record_lsa, LOG_RCVINDEX rcvindex,
+					VPID vpid, log_rv_redo_context &redo_context, const log_rv_redo_rec_info<T> &record_info);
+      void unfix_atomic_replication_sequence (THREAD_ENTRY *thread_p, TRANID tranid);
+      bool is_part_of_atomic_replication (TRANID tranid) const;
+#if !defined (NDEBUG)
+      bool is_page_part_of_atomic_replication_sequence (TRANID tranid, VPID vpid) const;
+#endif
+
+    private:
+      class atomic_replication_unit
```
cristiarg, line 63: comment

src/transaction/atomic_replication_helper.cpp
cristiarg, line 123: also, add a comment here:
```
using null hfid here as the watcher->group_id is initialized internally by pgbuf_ordered_fix at a cost
```

src/transaction/atomic_replication_helper.cpp
```diff
@@ -0,0 +1,203 @@
...
+    rcv.reference_lsa = m_record_lsa;
+    log_rv_redo_record_sync_apply (thread_p,redo_context, record_info, m_vpid, rcv);
+  }
+
+  void atomic_replication_helper::atomic_replication_unit::fix_page (THREAD_ENTRY *thread_p)
+  {
+    switch (m_record_index)
+      {
+      case RVHF_INSERT:
+      case RVHF_MVCC_INSERT:
```
cristiarg, line 153: `RVHF_INSERT_NEWHOME` should also be here:
- as visible from `heap_log_insert_physical` where `RVHF_INSERT_NEWHOME` log record is added
- called from `heap_insert_newhome` where pgptr is supplied from a `ins_context.home_page_watcher_p->pgptr` which means it used ordered fix to fix that pgptr

src/transaction/atomic_replication_helper.cpp
```diff
@@ -0,0 +1,203 @@
...
+      case RVHF_UPDATE:
+      case RVHF_MVCC_UPDATE_OVERFLOW:
+	if (pgbuf_ordered_fix (thread_p, &m_vpid, OLD_PAGE, PGBUF_LATCH_WRITE, &m_watcher) != NO_ERROR)
+	  {
+	    er_log_debug (ARG_FILE_LINE, "[ATOMIC REPLICATION] Unnable to apply ordered fix on page %d|%d.",
+			  VPID_AS_ARGS (&m_vpid));
+	    assert (false);
+	    return;
+	  }
+	break;
```
cristiarg, line 170: also add `RVHF_UPDATE_NOTIFY_VACUUM` under a new `case` group that `assert`s; we'll see later when testing if it needs to be added under the `pgbuf_ordered_fix`;

src/transaction/atomic_replication_helper.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+  template <typename T>
+  void atomic_replication_helper::add_atomic_replication_unit (THREAD_ENTRY *thread_p, TRANID tranid, log_lsa record_lsa,
+      LOG_RCVINDEX rcvindex, VPID vpid, log_rv_redo_context &redo_context, const log_rv_redo_rec_info<T> &record_info)
+  {
+#if !defined (NDEBUG)
+    if (!VPID_ISNULL (&vpid) && is_page_part_of_atomic_replication_sequence (vpid))
+      {
+	// page is already part of atomic replication
+	return;
```
cristiarg: does this mean that; if this `vpid` is already part of an atomic replication unit, the current log record will not be processed?
IMHO, it should do the following:
- if the vpid is already part of an atomic replication
- the tranid of the existing atomic replication should be the same as the tranid of the new log record
If the new log record which is supposed to be part of an atomic replication has a different tranid; it means there are two transactions changing the same page which is a situation we should not see, but we need to make sure it does not occur

cristiarg (commented):
Please write a more complete PR description.

src/transaction/atomic_replication_helper.cpp
```diff
@@ -57,20 +55,25 @@ namespace cublog
   }
 
 #if !defined (NDEBUG)
-  bool atomic_replication_helper::is_page_part_of_atomic_replication_sequence (TRANID tranid, VPID vpid) const
+  bool atomic_replication_helper::is_page_part_of_atomic_replication_sequence (VPID vpid) const
```
cristiarg: - rename to `void check_log_record_vs_transaction`
- if vpid is not part of an atomic replication sequence, ok, don't care
- if it is part of an atomic replication sequence, its `tranid` must be the same.
- also, try to avoid nested iteration - with a `std::set` of vpids

src/transaction/atomic_replication_helper.cpp
```diff
@@ -0,0 +1,209 @@
...
+      LOG_RCVINDEX rcvindex, VPID vpid, log_rv_redo_context &redo_context, const log_rv_redo_rec_info<T> &record_info)
+  {
+#if !defined (NDEBUG)
+    if (!VPID_ISNULL (&vpid) && is_page_part_of_atomic_replication_sequence (vpid))
+      {
+	// page is already part of atomic replication
+	return;
+      }
+#endif
+    auto sequence = m_atomic_sequences_map.find (tranid);
```
cristiarg: `sequence` is in fact an iterator
either emplace or retrieve:
```
std::vector<atomic_replication_unit>& sequence = m_atomic_sequences_map[tranid];
```

src/transaction/atomic_replication_helper.hpp
```diff
@@ -0,0 +1,94 @@
...
+
+	  VPID m_vpid;
+	private:
+	  log_lsa m_record_lsa;
+	  PAGE_PTR m_page_ptr;
+	  PGBUF_WATCHER m_watcher;
+	  LOG_RCVINDEX m_record_index;
+      };
+
+      std::map<TRANID, std::vector<atomic_replication_unit>> m_atomic_sequences_map;
```
cristiarg: using/typedef:
```
using atomic_replication_sequence_type = std::vector<atomic_replication_unit>;
```
and use the typedef all over

src/transaction/atomic_replication_helper.cpp
```diff
@@ -0,0 +1,209 @@
...
+      }
+    else
+      {
+	rcv.pgptr = m_watcher.pgptr;
+      }
+    rcv.reference_lsa = m_record_lsa;
+    log_rv_redo_record_sync_apply (thread_p,redo_context, record_info, m_vpid, rcv);
+  }
+
+  void atomic_replication_helper::atomic_replication_unit::fix_page (THREAD_ENTRY *thread_p)
```
cristiarg: propagate error upwards

src/transaction/atomic_replication_helper.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+namespace cublog
+{
+
+  /*********************************************************************
+   * atomic_replication_helper function definitions                    *
+   *********************************************************************/
+
+  template <typename T>
+  void atomic_replication_helper::add_atomic_replication_unit (THREAD_ENTRY *thread_p, TRANID tranid, log_lsa record_lsa,
```
cristiarg: propagate error upwards

src/transaction/atomic_replication_helper.cpp
```diff
@@ -157,8 +163,7 @@ namespace cublog
 	  {
 	    er_log_debug (ARG_FILE_LINE, "[ATOMIC REPLICATION] Unnable to apply ordered fix on page %d|%d.",
 			  VPID_AS_ARGS (&m_vpid));
-	    assert (false);
-	    return;
+	    return ER_FAILED;
```
cristiarg: return the specific error returned by `pgbuf_ordered_fix`

src/transaction/atomic_replication_helper.cpp
```diff
@@ -170,11 +175,12 @@ namespace cublog
 	if (m_page_ptr == nullptr)
 	  {
 	    er_log_debug (ARG_FILE_LINE, "[ATOMIC REPLICATION] Unnable to apply fix on page %d|%d.", VPID_AS_ARGS (&m_vpid));
-	    assert (false);
-	    return;
+	    return ER_FAILED;
```
cristiarg, line 181: same here

src/transaction/atomic_replication_helper.cpp
```diff
@@ -0,0 +1,207 @@
...
+    for (auto const &sequence : m_atomic_sequences_map)
+      {
+	for (size_t i = 0; i < sequence.second.size (); i++)
+	  {
+	    const VPID element_vpid = sequence.second[i].m_vpid;
+	    if (element_vpid.pageid == vpid.pageid && element_vpid.volid == vpid.volid)
+	      {
+		if (sequence.first != tranid)
+		  {
+		    return false;
```
cristiarg: Add logging here.

src/transaction/atomic_replication_helper.cpp
```diff
@@ -0,0 +1,207 @@
...
+   *********************************************************************/
+
+  template <typename T>
+  int atomic_replication_helper::add_atomic_replication_unit (THREAD_ENTRY *thread_p, TRANID tranid, log_lsa record_lsa,
+      LOG_RCVINDEX rcvindex, VPID vpid, log_rv_redo_context &redo_context, const log_rv_redo_rec_info<T> &record_info)
+  {
+#if !defined (NDEBUG)
+    if (!VPID_ISNULL (&vpid) && !check_for_page_validity (vpid, tranid))
+      {
+	// page is already part of atomic replication
```
cristiarg: this document is not relevant/up to date

src/transaction/atomic_replication_helper.cpp
```diff
@@ -0,0 +1,209 @@
...
+  }
+
+#if !defined (NDEBUG)
+  bool atomic_replication_helper::check_for_page_validity (VPID vpid, TRANID tranid) const
+  {
+    for (auto const &vpid_sets_iterator : m_atomic_sequences_vpids_map)
+      {
+	if (vpid_sets_iterator->first != tranid)
+	  {
+	    if (vpid_sets_iterator->second.find (vpid) != vpid_sets_iterator->second.end ())
```
cristiarg: `cend`

src/transaction/atomic_replication_helper.cpp
```diff
@@ -0,0 +1,209 @@
...
+	      {
+		er_log_debug (ARG_FILE_LINE, "[ATOMIC REPLICATION] Page %d|%d is part of multiple atomic replication sequences.",
+			      VPID_AS_ARGS (&vpid));
```
cristiarg: Log existing and new tranid.

cristiarg (approved):
still small remarks

PR #3499, opened 2022-04-12
[CBRD-24280] [Regression] Core dumped in mspace_free at src/heaplayers/malloc_2_8_3.c:5042

youngjinj:
http://jira.cubrid.org/browse/CBRD-24280

In the sm_rename_class() function, class_new_name is freed if the following code execution fails after the sm_flush_objects() function is executed. However, a core occurs when the ws_abort_mops() function tries to free class_new_name (class_->header.ch_name) once more while freeing the flushed object in the abort process.

Even if rename fails in sm_rename_class() function, class_new_name is not freed when sm_flush_objects() function execution is completed.


shparkcubrid (approved):

src/object/schema_manager.c
```diff
@@ -2867,23 +2873,10 @@ sm_rename_class (MOP class_mop, const char *new_name)
 	      tran_abort_upto_system_savepoint (UNIQUE_PARTITION_SAVEPOINT_RENAME);
 	    }
```
byungwook-kim, line 2875:
```suggestion
db_private_free_and_init (NULL, class_new_name);
```

youngjinj: ~After sm_flush_objects(), class_new_name must not be freed.~

After changing to "class_->header.ch_name = class_new_name;", class_new_name must not be freed separately.

beyondykk9 (approved):

hyunikn (approved):

hgryoo (approved):

src/object/schema_manager.c
```diff
@@ -2815,10 +2819,13 @@ sm_rename_class (MOP class_mop, const char *new_name)
 
   class_->header.ch_name = class_new_name;
 
+  need_free_old_name = true;
+  need_free_new_name = false;
+
```
ctshim, line 2824: If sm_flush_objects() fails
No need to keep "need_free_new_name = true" state?

ctshim (approved):

src/object/schema_manager.c
youngjinj: After changing to "class_->header.ch_name = class_new_name;", class_new_name must not be freed separately.

What I said earlier was incorrect.

PR #3500, opened 2022-04-12
[CBRD-24274] Fix the bug occurred when read from the dismounted log volume

joohok:
http://jira.cubrid.org/browse/CBRD-24274

Purpose

* The most recently used archive log volume can be mounted and information can be saved in log_Gl.archive (log_Gl.archive.vdes)
* In cdc_get_start_point_from_file(), if the most recent archive log volume is used, the operation is finished and dismounted.
* If another thread tries to access the archive log volume using log_Gl.archive.vdes, an error occurs because it was dismounted in the above function.

Implementation
* When using the archive volume stored in log_Gl.archive, the stored information is used
* if do not use the archive volume stored in log_Gl.archive, mount the required log volume in the same way as before to obtain information and perform dismount.

### DIFF
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -14137,6 +14137,8 @@ cdc_get_start_point_from_file (THREAD_ENTRY * thread_p, int arv_num, LOG_LSA * r
     }
   else
     {
+      LOG_ARCHIVE_CS_ENTER (thread_p);
+
       if (log_Gl.archive.vdes != NULL_VOLDES && log_Gl.archive.hdr.arv_num == arv_num)
 	{
 	  /* if target archive log volume is currenty mounted, then use that */
@@ -14145,8 +14147,6 @@ cdc_get_start_point_from_file (THREAD_ENTRY * thread_p, int arv_num, LOG_LSA * r
 	}
       else
 	{
-	  LOG_ARCHIVE_CS_ENTER (thread_p);
-
 	  aligned_hdr_pgbuf = PTR_ALIGN (hdr_pgbuf, MAX_ALIGNMENT);
 
 	  hdr_pgptr = (LOG_PAGE *) aligned_hdr_pgbuf;
@@ -14188,10 +14188,11 @@ cdc_get_start_point_from_file (THREAD_ENTRY * thread_p, int arv_num, LOG_LSA * r
 
 		  fileio_dismount (thread_p, vdes);
 
-		  LOG_ARCHIVE_CS_EXIT (thread_p);
 		}
 	    }
 	}
+
+      LOG_ARCHIVE_CS_EXIT (thread_p);
     }
 
   LOG_CS_EXIT (thread_p);
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -14119,8 +14119,8 @@ cdc_get_start_point_from_file (THREAD_ENTRY * thread_p, int arv_num, LOG_LSA * r
   char ctime_buf[CTIME_MAX];
   int error_code;
 
-  LOG_LSA process_lsa;
-  LOG_LSA forw_lsa;
+  LOG_LSA process_lsa = LSA_INITIALIZER;
+  LOG_LSA forw_lsa = LSA_INITIALIZER;
 
   LOG_RECORD_HEADER *log_rec_header;
   LOG_REC_DONETIME *donetime;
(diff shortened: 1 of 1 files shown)


src/transaction/log_manager.c
```diff
@@ -14137,47 +14137,59 @@ cdc_get_start_point_from_file (THREAD_ENTRY * thread_p, int arv_num, LOG_LSA * r
...
-      aligned_hdr_pgbuf = PTR_ALIGN (hdr_pgbuf, MAX_ALIGNMENT);
+      if (log_Gl.archive.vdes != NULL_VOLDES && log_Gl.archive.hdr.arv_num == arv_num)
+	{
+	  /* if target archive log volume is currenty mounted, then use that */
+	  process_lsa.pageid = log_Gl.archive.hdr.fpageid;
+	  process_lsa.offset = 0;
+	}
+      else
+	{
+	  LOG_ARCHIVE_CS_ENTER (thread_p);
```
wrlawodms: Not related with this issue, but there is no `LOG_ARCHIVE_CS_EXIT (thread_p);` when `fileio_is_volume_exist (arv_name)` is not true.

src/transaction/log_manager.c
```diff
@@ -14137,47 +14137,59 @@ cdc_get_start_point_from_file (THREAD_ENTRY * thread_p, int arv_num, LOG_LSA * r
...
   else
     {
-      LOG_ARCHIVE_CS_ENTER (thread_p);
-      aligned_hdr_pgbuf = PTR_ALIGN (hdr_pgbuf, MAX_ALIGNMENT);
+      if (log_Gl.archive.vdes != NULL_VOLDES && log_Gl.archive.hdr.arv_num == arv_num)
+	{
+	  /* if target archive log volume is currenty mounted, then use that */
+	  process_lsa.pageid = log_Gl.archive.hdr.fpageid;
+	  process_lsa.offset = 0;
+	}
```
wrlawodms, line 14147: 1. I don't know well, but please check whether it needs `LOG_ARCHIVE_CS_ENTER()` because it seems that `vdes` and `hdr` of `log_Gl.archive` are changed not atomically in other threads.
2. What happens if `log_Gl.archive.vdes` is dismounted after this in another thread? Is it OK?

joohok: Oh.. yes, I will check it and move the LOG_ARCHIVE_CS_ENTER() and EXIT(). thank you.

src/transaction/log_manager.c
joohok: Thank you, I will look into it

eido5 (approved):

src/transaction/log_manager.c
joohok: 1. information in log_Gl.archive is protected by LOG_ARCHIVE_CS_ENTER()/EXIT() 
2. After getting fpageid of archive log volume, it doesn't matter if it is dismounted. 
    This is because it is mounted again with the corresponding fpageid in the logpb_fetch_page() below.

src/transaction/log_manager.c
joohok: I moved the LOG_ARCHIVE_CS_EXIT()

src/transaction/log_manager.c
```diff
@@ -14138,48 +14138,61 @@ cdc_get_start_point_from_file (THREAD_ENTRY * thread_p, int arv_num, LOG_LSA * r
...
-      fileio_make_log_archive_name (arv_name, log_Archive_path, log_Prefix, arv_num);
+	  hdr_pgptr = (LOG_PAGE *) aligned_hdr_pgbuf;
 
-      if (fileio_is_volume_exist (arv_name) == true)
-	{
-	  vdes = fileio_mount (thread_p, log_Db_fullname, arv_name, LOG_DBLOG_ARCHIVE_VOLID, false, false);
-	  if (vdes != NULL_VOLDES)
+	  fileio_make_log_archive_name (arv_name, log_Archive_path, log_Prefix, arv_num);
+
+	  if (fileio_is_volume_exist (arv_name) == true)
```
hornetmj, line 14156: It would be nice if handling 'else' case too. In the perspective of code, the process_lsa was not initialized and it is passed to the logpb_fetch_page().

hornetmj (approved):

src/transaction/log_manager.c
hornetmj: 11.1 patch please.

wrlawodms (approved):

src/transaction/log_manager.c
joohok: Yes, you are right. 
Attempts were made only to check against archive log volumes that already exist through the information stored in log_Gl. 

However, I'll add an error along with assert() as it can cause an error when that archive volume is deleted.

joohok:
@mhoh3963
Please, merge this PR

PR #3501, opened 2022-04-12
[LETS-446] Refactor PARALLEL_RECOVERY_REDO/RECOVERY_PARALLEL_COUNT_TSLS tests to not use pstack

adrianbrab:
http://jira.cubrid.org/browse/LETS-446

Added a new hidden parameter `recovery_parallel_task_debug`, in order to aid the parallel recovery automated test to more easily track the parallel task.


src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -222,6 +222,13 @@ namespace cublog
     // which should help to only allocate/reserve once
     jobs_vec.reserve (PARALLEL_REDO_JOB_VECTOR_RESERVE_SIZE);
 
+#if defined(NDEBUG)
```
razvanradulescu:
```suggestion
#if !defined(NDEBUG)
```

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -222,6 +222,13 @@ namespace cublog
...
+    if (prm_get_bool_value (PRM_ID_RECOVERY_PARALLEL_TASK_DEBUG))
+      {
+	er_log_debug (ARG_FILE_LINE, "[PARALLEL RECOVERY] Call to parallel execute with task index: %d. ", m_task_idx);
```
cristiarg:
```
- keep an on-stack counter about the number of jobs the task has executed, debug mode only
- use the parameter in redo_parallel::redo_task::execute to log the number of jobs executed along with m_task_idx, debug mode only - comment the motives behind this logging
```
This is needed to ensure that the task really does process log records. Entering the `execute` function does not guarantee that the task also processes log records.

src/transaction/log_recovery_redo_parallel.cpp
```diff
@@ -243,6 +247,9 @@ namespace cublog
 		  {
 		    set_unapplied_log_lsa_from_execute_func (job->get_log_lsa ());
 		  }
+#if !defined(NDEBUG)
+		task_execute_counter++;
```
cristiarg: Nit: `++value`

cristiarg (approved):

PR #3504, opened 2022-04-13
[LETS-210] Logging b-tree operations for atomic replication: btree_delete_key_from_leaf

cristiarg:

- extracted function `btree_get_first_overflow_page_vpid_from_leaf` that gets the vpid of the first overflow page out of the leaf node where the key is stored
- used the function in `btree_delete_overflow_key` to first get the vpid and then, after the key entry in the leaf page is removed, also remove the overflow page [chain]

Other:
- added `const` in a lot of places where it makes sense (during investigation to understand the code)
- added a const version of a macro - `CAST_CONST_PGPTR_TO_CONST_BFPTR`


razvanradulescu (approved):

cristiarg:
@razvanradulescu see the last commit

adrianbrab (approved):

PR #3509, opened 2022-04-14
[CBRD-24285] Problem of sort limit optimization without checking not null of foreign key

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-24285

check not null of fk when sort limit optimization


src/optimizer/query_graph.c
```diff
@@ -1518,6 +1522,42 @@ qo_insert_segment (QO_NODE * head, QO_NODE * tail, PT_NODE * node, QO_ENV * env,
...
+	    }
+	  if (found)
+	    {
+	      break;
+	    }
+	  constraints = constraints->next;
+	}
+    }
+
   env->nsegs++;
```
ctshim, line 1565: Same content, but I think the code could be shorter

```suggestion
  env->nsegs++;
    
    /* check is_not_null */
  entity = QO_NODE_ENTITY_SPEC (head);
  if (pt_is_name_node (node) && PT_SPEC_IS_ENTITY (entity))
    {
      int i;
      cls = sm_find_class (entity->info.spec.entity_name->info.name.original);
      constraints = sm_class_constraints (cls);

      while (constraints != NULL)
	{
	  if (SM_IS_CONSTRAINT_NOT_NULL_FAMILY (constraints->type))
	    {	      
	       /* check columns on this constraint */
	      for (i = 0; constraints->attributes[i]; i++)
	        {
	          attrp = constraints->attributes[i];
	          if (intl_identifier_casecmp (node->info.name.original, attrp->header.name) == 0)
		    {
		      QO_SEG_IS_NOT_NULL (seg) = true;
		     return seg;
		    }
	        }
	    }
	  constraints = constraints->next;
	}
    }
```

ctshim (approved):

src/optimizer/query_graph.c
shparkcubrid: Thanks for your comments. But I don't want to put 'return' in the middle. I'll keep my style.

src/optimizer/query_graph.c
```diff
@@ -1518,6 +1522,42 @@ qo_insert_segment (QO_NODE * head, QO_NODE * tail, PT_NODE * node, QO_ENV * env,
...
+  /* check is_not_null */
+  entity = QO_NODE_ENTITY_SPEC (head);
+  if (pt_is_name_node (node) && PT_SPEC_IS_ENTITY (entity))
+    {
+      int i;
+      bool found = false;
+      cls = sm_find_class (entity->info.spec.entity_name->info.name.original);
+      constraints = sm_class_constraints (cls);
+
+      while (constraints != NULL)
```
beyondykk9, line 1538:
```suggestion
      while (constraints != NULL && !found)
```

src/optimizer/query_graph.c
```diff
@@ -1518,6 +1522,42 @@ qo_insert_segment (QO_NODE * head, QO_NODE * tail, PT_NODE * node, QO_ENV * env,
...
+	    {
+	      attrp = constraints->attributes[i];
+	      if (intl_identifier_casecmp (node->info.name.original, attrp->header.name) == 0)
+		{
+		  QO_SEG_IS_NOT_NULL (seg) = true;
+		  found = true;
+		  break;
+		}
+	    }
+	  if (found)
```
beyondykk9, line 1557: no needed for moving while condition (!found)

shparkcubrid: Thanks for the comment. but I think these are similar. I'll keep my source code.

src/optimizer/query_graph.c
```diff
@@ -9596,6 +9637,12 @@ qo_is_pk_fk_full_join (QO_ENV * env, QO_NODE * fk_node, QO_NODE * pk_node)
 	  fk_seg = QO_TERM_INDEX_SEG (term, 1);
 	}
 
+      /* fk must have the NOT_NULL constraint */
+      if (!QO_SEG_IS_NOT_NULL (fk_seg))
```
hgryoo, line 9645: Is it typo?
```suggestion
      /* fk must not have the NOT_NULL constraint */
      if (!QO_SEG_IS_NOT_NULL (fk_seg))
```

hgryoo (approved):

src/optimizer/query_graph.c
shparkcubrid: No. if not have NOT_NULL, return false.

beyondykk9 (approved):

src/optimizer/query_graph.c
hgryoo: I see

youngjinj (approved):

hyunikn (approved):

PR #3510, opened 2022-04-14
[CBRD-24286] [Regression] [isolation] It is not possible to create a db with a name that starts with the system class.

youngjinj:
http://jira.cubrid.org/browse/CBRD-24286

Resolves the problem of checking as a system class if the prefix is the same as the system class name.

The reason was that only the length of the system class name was compared with strncmp ().
This problem has been resolved.
sm_check_system_class_by_name() was also improved to perform binary search.


src/object/schema_manager.c
```diff
@@ -3312,10 +3321,22 @@ sm_check_system_class_by_name (const char *name)
       return false;
     }
 
-  count = sizeof (system_classes) / sizeof (system_classes[0]);
-  for (i = 0; i < count; i++)
+  /* binary search */
```
beyondykk9: How about using bsearch()
 ```
      #include <stdlib.h>

       void *bsearch(const void *key, const void *base,
                     size_t nmemb, size_t size,
                     int (*compar)(const void *, const void *));
```

**Review by shparkcubrid — APPROVED** [review]

File: src/object/schema_manager.c [hunk 2]
```diff
@@ -3312,10 +3321,22 @@ sm_check_system_class_by_name (const char *name)
...
+  left = 0;
+  right = count;
+  while (left <= right)
     {
-      if (strncmp (downcase_name, system_classes[i].name, system_classes[i].len) == 0)
```
ctshim, line 3326: how about below?
`if (memcmp (downcase_name, system_classes[i].name, system_classes[i].len+1) == 0)`

src/object/schema_manager.c
```diff
@@ -131,9 +131,86 @@ typedef struct schema_def
...
 static SCHEMA_DEF Current_Schema = { {'\0'}, NULL, NULL, NULL, NULL };
 
+typedef struct system_class_def
+{
+  const char *name;
+  int len;
+} SYSTEM_CLASS_DEF;
 
+// *INDENT-OFF*
+static SYSTEM_CLASS_DEF system_classes[] = {
```
ctshim, line 141: If you collect names starting with "_db_" and "db_" and divide them into two arrays, the size is small enough that you do not need to use binary search.
In that case, sequential search seems to be more efficient.
Binary search becomes rather inefficient when the number of data is less than a certain number. This is my personal opinion. Note please

hyunikn (approved):

hgryoo (approved):

src/object/schema_manager.c
youngjinj: Thank you.
However, it becomes difficult to manage the list of system classes.
Instead, I will improve it in other ways.

src/object/schema_manager.c
youngjinj: If it was the old code, it would have been a good way.

I improved it by using the length comparison method a while ago.
So it seems there is no problem to use strcmp ().

src/object/schema_manager.c
youngjinj: Because of the small amount of data, I decided not to use binary search.
Instead, it was improved as a way to compare lengths.

src/object/schema_manager.c
```diff
@@ -3189,6 +3269,24 @@ sm_is_system_class (MOP op)
...
+    {
+      return -1;
+    }
+  else if (a_len > b_len)
+    {
+      return 1;
+    }
+
+  return 0;
+}
```
ctshim, line 3288: Please make the function as below.
```
static int
system_class_def_compare (const void *a, const void *b)
{
    const SYSTEM_CLASS_DEF * sa = (const SYSTEM_CLASS_DEF *) a;
    const SYSTEM_CLASS_DEF * sb = (const SYSTEM_CLASS_DEF *) b;

    if( sa->len != sb->len)
    {
        return sa->len - sb->len;
    }
    return strcmp(sa->name, sb->name);
}
```

That way you can make the bottom for loop more efficient like this:
```
int cmp;
for (i = 0; i < count; i++)
    {
        cmp = system_class_def_compare(downcase_name, system_classes[i].name) ;
        if(cmp <= 0)
          {
              return  (cmp == 0);
           }        
    }
```

src/object/schema_manager.c
```diff
@@ -3189,6 +3269,24 @@ sm_is_system_class (MOP op)
...
+
```
ctshim: Please make the function as below.
```
static int
system_class_def_compare (const void *a, const void *b)
{
    const SYSTEM_CLASS_DEF * sa = (const SYSTEM_CLASS_DEF *) a;
    const SYSTEM_CLASS_DEF * sb = (const SYSTEM_CLASS_DEF *) b;

    if( sa->len != sb->len)
    {
        return sa->len - sb->len;
    }
    return strcmp(sa->name, sb->name);
}
```
That way you can make the bottom for loop more efficient like this:
```
SYSTEM_CLASS_DEF  sa; 
int cmp;
sa.name = downcase_name;
sa.len = len;
for (i = 0; i < count; i++)
    {
        cmp = system_class_def_compare(&sa, system_classes+i) ;
        if(cmp <= 0)
          {
              return  (cmp == 0);
           }        
    }
```

beyondykk9 (approved):

src/object/schema_manager.c
youngjinj: Thank you!
I will change it like the code above.

ctshim (approved):

PR #3511, opened 2022-04-14
[CBRD-24268] Fix to hide information about system class in flashback

joohok:
http://jira.cubrid.org/browse/CBRD-24268

Purpose

Fix the bug where transactions for system classes could appear in the flashback summary.

Implementation

turn off supplemental_log parameter when creating database (createdb)

### DIFF
--- a/src/transaction/flashback.c
+++ b/src/transaction/flashback.c
@@ -459,6 +459,13 @@ flashback_make_summary_list (THREAD_ENTRY * thread_p, FLASHBACK_SUMMARY_CONTEXT 
 		{
 		  memcpy (&classoid, supplement_data, sizeof (OID));
 
+		  if (oid_is_system_class (&classoid))
+		    {
+		      /* while creating database, there is no information about system class in server
+		       * so that supplemental log can be logged */
+		      break;
+		    }
+
                   // *INDENT-OFF*
 		  if (std::find (context->classoids.begin(), context->classoids.end(), classoid) == context->classoids.end())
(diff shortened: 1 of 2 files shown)


wrlawodms (approved):

src/transaction/flashback.c
```diff
@@ -459,6 +459,13 @@ flashback_make_summary_list (THREAD_ENTRY * thread_p, FLASHBACK_SUMMARY_CONTEXT
 		{
 		  memcpy (&classoid, supplement_data, sizeof (OID));
 
+		  if (oid_is_system_class (&classoid))
```
hornetmj: How about turning off the 'supplemental_log' in the createdb() using sysprm_set_force()?

joohok: Oh.. it is really good idea. 
I will apply it then test if it works.

joohok: it works well, I applied it

eido5 (approved):

src/transaction/flashback.c
hornetmj: This fix needs to apply to the 11.1 branch.

And how about leave hear 'assert (oid_is_system_class (&classoid) == false)'?

joohok: OK, I know what you mean. I will add assert.

hornetmj (approved):

joohok:
@mhoh3963
Please, merge this PR

PR #3514, opened 2022-04-14
[CBRD-24273] Fix the bug occurred when find start lsa from archive log volume

joohok:
http://jira.cubrid.org/browse/CBRD-24273

Purpose

* If the start_time of the selected transaction does not exist, flashback try to find start_lsa 
   * At this time, if the log corresponding to start_lsa exists in the archive log volume, an error occurs.
   * LOG_CS_ENTER() is required before calling logpb_fetch_from_archive()
* When exiting the while loop, the LSA of the current log record should be fetched
   * But the prev_tranlsa value having a NULL value is fetched 
   * Therefore, start_lsa is a NULL value, and it behaves incorrectly

Implementation

* add LOG_CS_ENTER_READ_MODE () and LOG_CS_EXIT() around logpb_fetch_from_archive()
* add variable (cur_log_rec_lsa) to store LSA of current log record

### DIFF
--- a/src/transaction/flashback.c
+++ b/src/transaction/flashback.c
@@ -693,14 +693,6 @@ flashback_find_start_lsa (THREAD_ENTRY * thread_p, FLASHBACK_LOGINFO_CONTEXT * c
   assert (!LSA_ISNULL (&context->end_lsa));
   LSA_COPY (&process_lsa, &context->end_lsa);
 
-  /* fetch log page */
-  error = logpb_fetch_page (thread_p, &process_lsa, LOG_CS_SAFE_READER, log_page_p);
-  if (error != NO_ERROR)
-    {
-      logpb_fatal_error (thread_p, false, ARG_FILE_LINE, "flashback_make_loginfo");
-      goto error;
-    }
-
   while (!LSA_ISNULL (&process_lsa))
     {
       /* fetch page from archive or active, if previous page is needed */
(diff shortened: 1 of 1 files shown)


src/transaction/flashback.c
```diff
@@ -692,28 +693,26 @@ flashback_find_start_lsa (THREAD_ENTRY * thread_p, FLASHBACK_LOGINFO_CONTEXT * c
...
-  if (error != NO_ERROR)
-    {
-      logpb_fatal_error (thread_p, false, ARG_FILE_LINE, "flashback_make_loginfo");
-      goto error;
-    }
-
   while (!LSA_ISNULL (&process_lsa))
     {
       /* fetch page from archive or active, if previous page is needed */
       if (log_page_p->hdr.logical_pageid != process_lsa.pageid)
```
wrlawodms, line 701: `log_page_p->hdr.logical_pageid` seems uninitialized.

joohok: yes, there can be a trash value in `log_page_p->hdr.logical_pageid`.
I will initialize it to -1. Thanks

eido5 (approved):

hornetmj (approved):

wrlawodms (approved):

joohok:
@mhoh3963 
Please, merge this PR

PR #3516, opened 2022-04-15
[CBRD-24287] Query generated inside the SP incorrectly affects the query plan result

hgryoo:
http://jira.cubrid.org/browse/CBRD-24287

**Purpose**
When displaying the query plan, method or stored procedure should not be executed.

**Implementation**
- Fixed the stored procedure not to be executed by constant folding by checking the optimization level parameter.
- Controlled to open and close before and after the main code that generates XASL so that the file that query plan dump is not kept open.


src/compat/db_query.c
```diff
@@ -3572,6 +3572,91 @@ db_query_plan_dump_file (char *filename)
...
+}
+
+/*
+ * db_query_plan_dump_fp_open() -
+ * return : FILE *
+ */
+FILE *
+db_query_plan_dump_fp_open ()
+{
+  if (query_Plan_dump_fp_open)
```
beyondykk9, line 3602: the status of plan dump fp can be checked by just checking query_Plan_dump_fp is NULL or not.
Therefore, query_Plan_dump_fp_open is not necessary.

ctshim (approved):

src/compat/db_query.c
hgryoo: At first, `query_Plan_dump_fp` is initialized to `NULL`. 
But in many places if `query_Plan_dump_fp` is `NULL`, set `query_Plan_dump_fp` to `stdout`. 
For readability, checking `query_Plan_dump_fp` is `NULL` or `stdout` seems not good for me.

beyondykk9 (approved):

hyunikn (approved):

src/parser/xasl_generation.c
```diff
@@ -21924,9 +21881,14 @@ parser_generate_xasl (PARSER_CONTEXT * parser, PT_NODE * node)
 	  /* XASL cache related information */
 	  pt_init_xasl_supp_info ();
 
+	  db_query_plan_dump_fp_open ();
```
shparkcubrid, line 21888: pt_to_insert_xasl() also calls parser_generate_xasl_post().
It is necessary to check whether the plan of the query below is normal.
```
create table t (col1 int, col2 int);
set optimization level 513;
insert /*+ recompile */ into t values(1,(select 1 from dual limit 1));
```

hgryoo: Query plan is shown as following
```
csql> insert /*+ recompile */ into t values(1,(select 1 from dual limit 1));
seg[0]: [0]
Join graph nodes:
node[0]: dual dual(0/1) (loc 0)
Join graph terms:
term[0]: inst_num() range (min inf_le 1) (sel 0.1) (rank 3) (instnum term) (not-join eligible) (loc 0)

Query plan:

sscan
    class: dual node[0]
    sargs: term[0]
    cost:  1 card 0

Query stmt:

(select 1 from dual dual where (inst_num()<= ?:0 ))
```

Additionally, I've tested too for the case that the stored procedure is called.
```
set optimization level 513;
create table t2 (col1 int, col2 string);
insert /*+ recompile */ into t2 values(1,(select hello() from dual limit 1)); -- hello() returns string type
```

Result
```
seg[0]: [0]
Join graph nodes:
node[0]: dual dual(0/1) (loc 0)
Join graph terms:
term[0]: inst_num() range (min inf_le 1) (sel 0.1) (rank 3) (instnum term) (not-join eligible) (loc 0)

Query plan:

sscan
    class: dual node[0]
    sargs: term[0]
    cost:  1 card 0

Query stmt:

(select hello() from dual dual where (inst_num()<=1))

```

shparkcubrid: In case of insert query, db_query_plan_dump_fp_open() does not seem to be called differently than before. is this ok?

hgryoo: I think yes. Insert query doesn't generate XASL, so opening a file pointer for the query plan is not necessary.

ctshim (approved):

shparkcubrid (approved):

youngjinj (approved):

PR #3521, opened 2022-04-15
[LETS-210] Logging b-tree operations for atomic replication: btree_delete_key_from_leaf

cristiarg:

- extracted function `btree_get_first_overflow_page_vpid_from_leaf` that gets the vpid of the first overflow page out of the leaf node where the key is stored
- used the function in `btree_delete_overflow_key` to first get the vpid and then, after the key entry in the leaf page is removed, also remove the overflow page [chain]


src/storage/btree.c
```diff
@@ -2179,24 +2183,18 @@ btree_load_overflow_key (THREAD_ENTRY * thread_p, BTID_INT * btid, VPID * first_
...
- *   node_type(in): Leaf or NonLeaf page
- *
- * Note: The overflow key is deleted. This routine will not delete the btree slot containing the key.
+ * btree_get_first_overflow_page_vpid_from_leaf - extract the vpid of the first overflow key page;
+ *                         the leaf page ptr supplied is assumed to be a leaf page with overflow
  */
 static int
-btree_delete_overflow_key (THREAD_ENTRY * thread_p, BTID_INT * btid, PAGE_PTR page_ptr, INT16 slot_id,
-			   BTREE_NODE_TYPE node_type)
+btree_get_first_overflow_page_vpid_from_leaf (THREAD_ENTRY * thread_p, const BTID_INT * btid,
```
razvanradulescu, line 2190: Function name is not accurate: "_from_leaf". It actually gets the VPID of the overflow page for both leaf and non-leaf nodes.

razvanradulescu (approved):

PR #3523, opened 2022-04-17
[CBRD-24281] [Regression] [isolation] In isolation level repeatable read, the select  result of client2 is different from the previous version.

youngjinj:
http://jira.cubrid.org/browse/CBRD-24281

When creating a user-specified name, the current schema name is checked instead of the current user name.

### DIFF
--- a/src/object/authenticate.c
+++ b/src/object/authenticate.c
@@ -7146,8 +7146,6 @@ au_perform_login (const char *name, const char *password, bool ignore_dba_privil
 		{
 		  error = AU_SET_USER (user);
 
-		  error = sc_set_current_schema (user);
-
 		  /* necessary to invalidate vclass cache */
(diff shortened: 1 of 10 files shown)


src/object/authenticate.c
```diff
@@ -7205,6 +7146,8 @@ au_perform_login (const char *name, const char *password, bool ignore_dba_privil
 		{
 		  error = AU_SET_USER (user);
```
beyondykk9: no need to check error?
finally AU_SET_USER calls "sc_set_current_schema".
it's not redundant to call "sc_set_current_schema" again?

youngjinj: Removed call sc_set_current_schema in next commit.

beyondykk9 (approved):

PR #3524, opened 2022-04-17
[CBRD-24121] Support for handling interrupts and recursion limits in stored procedure calls

hgryoo:
http://jira.cubrid.org/browse/CBRD-24121

**Purpose**
To support error handling for interrupts and recursion limits, 

**Implementation**
- (server_support.c) 
   Revert to previous implementation. pass `is_method` value into `push_task_on_core ()`
- (thread_worker_pool.hpp) 
   Add new worker routine only for method mode (calling recursive C method/stored procedure)
  - if there is no active worker, a new thread is created and deleted immediately after executing the task
- Supports an interrupt routine that resets the running (method) runtime context and raises an proper error
  - (log_tran_table.c) 
     here, interrupt is set in runtime context when `^C` or `query_cancel()` is performed at the client.
  - (method_invoke_group)
     set an error if interrupt is happened
- (method_connection_sr.cpp) 
  Add checking routine that the runtime context is in interrupting
  - (jsp_comm.c)
     jsp_readn() wraps css_readn() with timeout (5000). whenever timeout occurs, the function checks interrupt is happened
- misc
  - (method_connection_pool.cpp) 
     If java's thread is in unexpected state (is_valid() is false), disconnect

**Remarks**
N/A


beyondykk9 (approved):

src/method/method_connection_sr.cpp
```diff
@@ -75,10 +79,14 @@ namespace cubmethod
...
 
     int error = xs_receive_data_from_client (thread_p, &buffer.ptr, (int *) &buffer.dim);
-    if (error == NO_ERROR)
+    if (error == NO_ERROR && er_errid () == NO_ERROR)
       {
 	error = func (socket, buffer);
       }
+    else
+      {
+	error = (error == NO_ERROR) ? er_errid () : NO_ERROR; // ER_SP_TOO_MANY_NESTED_CALL
```
ctshim: If (error != NO_ERROR) are you trying to make it error = NO_ERROR?
Please check once more whether you are trying to process in the following meanings.
```
if(error == NO_ERROR) 
{
   error = er_errid ();
}
```

src/method/method_invoke_group.cpp
```diff
@@ -198,8 +206,15 @@ namespace cubmethod
 	  }
 
 	error = m_method_vector[i]->get_return (m_thread_p, arg_base, m_result_vector[i]);
+	if (m_rctx->is_interrupted ())
+	  {
+	    error = m_rctx->get_interrupt_reason ();
+	  }
+
```
ctshim, line 213: If there is an interrupt, is it correct to ignore the return value of m_method_vector[i]->get_return() and substitute the error variable for the reason of the interrupt?
If the return value of m_method_vector[i]->get_return() is an error, how about handling it first and then checking whether there is an interrupt?

src/method/method_invoke_group.cpp
```diff
@@ -198,8 +206,15 @@ namespace cubmethod
 	  }
 
 	error = m_method_vector[i]->get_return (m_thread_p, arg_base, m_result_vector[i]);
+	if (m_rctx->is_interrupted ())
+	  {
+	    error = m_rctx->get_interrupt_reason ();
```
ctshim, line 211: Here, the return value of m_rctx->get_interrupt_reason() cannot be NO_ERROR. Also, you don't have to call m_rctx->set_interrupt_by_reason(error) again.
So, it looks like you could do something like this:
`return m_rctx->get_interrupt_reason();`

hgryoo: At m_rctx->set_interrput (error), the `error` may not interrupt reason such as ER_INTERRUPTED. If an error is not interrupt-related code, It does not enter interrupt mode.

hyunikn (approved):

src/method/method_invoke_group.cpp
hgryoo: Yes, it is correct. In the interrupt situation, the return value of a method_invoke_group does not have to be handled, and only **the interrupt reason of the runtime context should be set as an error and returns the reason code to the client directly**.
For this reason, the error code corresponding to the reason for the interrupt has a higher priority than other execution error codes, so if it is in the interrupt, there is no need to handle it.

shparkcubrid (approved):

ctshim (approved):

src/method/method_connection_sr.cpp
hgryoo: Thank you. I've fixed it as your suggestion.

youngjinj (approved):

PR #3526, opened 2022-04-18
[CBRD-24279] Support user schema feature in flashback utility

joohok:
http://jira.cubrid.org/browse/CBRD-24279

Purpose

As user schema feature is supported in 11.2, flashback utility also need to support it.
When users input table name without specifying owner name, it needs to attach the owner name to the table name.
So if the user does not specify an owner name, it will cause the error

Implement
* Output an error when user doesn't specify the owner name
* Modify usage to let users to specify owner name


msg/en_US/utils.msg
```diff
@@ -1276,7 +1276,7 @@ $set 58 MSGCAT_UTIL_SET_FLASHBACK
 14 Flashback cannot run as standalone mode.\n
 60 \
 flashback: Provides statements to flashback a specific transaction.\n\
-usage: %1$s flashback [OPTION] database-name class_name1 [class_name2 ...]\n\
+usage: %1$s flashback [OPTION] database-name owner_name.class_name1 [owner_name.class_name2 ...]\n\
```
wrlawodms, line 1281: The owner name is not necessary. Don't we need to tell users that owner_name will be set to dba if not given?
```suggestion
usage: %1$s flashback [OPTION] database-name [owner_name.]class_name1 [[owner_name.]class_name2 ...]\n\
```

joohok:
@mhoh3963  @youngjinj 
I have registered you as reviewers because the usage modification will affect the modifications of other utilities (checkdb, compactdb) in the future. If you have any comments regarding usage, please review the usage.

src/executables/util_cs.c
```diff
@@ -4291,6 +4279,20 @@ flashback (UTIL_FUNCTION_ARG * arg)
       goto error_exit;
     }
 
+  for (i = 0; i < num_tables; i++)
+    {
+      table_name = utility_get_option_string_value (arg_map, OPTION_STRING_TABLE, i + 1);
+
+      /* when table_name doesn't contain the user name (owner), then attach the user name to the table name */
+      sm_user_specified_name (table_name, table_name_buf, SM_MAX_IDENTIFIER_LENGTH);
```
youngjinj: It should be checked that the length of table_name does not exceed the length of table_name_buf (SM_MAX_IDENTIFIER_LENGTH).

Please refer to the code that checks the length of user_specified_name in get_class_mops_from_file().

```
/* src/executables/compactdb_common.c */

int
get_class_mops_from_file (const char *input_filename, MOP ** class_list, int *num_class_mops)
{
  ~
  len = STATIC_CAST (int, strlen (buffer));
  sub_len = len;
  ~
  dot = strchr (buffer, '.');
  if (dot)
    {
      /* user specified name */
  
      /* user name of user specified name */
      sub_len = STATIC_CAST (int, dot - buffer);
      if (sub_len >= DB_MAX_USER_LENGTH)
        {
          PRINT_AND_LOG_ERR_MSG (msgcat_message (MSGCAT_CATALOG_UTILS, MSGCAT_UTIL_SET_COMPACTDB,
                                                 COMPACTDB_MSG_EXCEED_MAX_USER_LEN), DB_MAX_USER_LENGTH - 1);
          status = ER_FAILED;
          goto end;
        }
  
      /* class name of user specified name */
      sub_len = STATIC_CAST (int, strlen (dot + 1));
    }
  
  if (sub_len >= DB_MAX_IDENTIFIER_LENGTH - DB_MAX_USER_LENGTH)
    {
      PRINT_AND_LOG_ERR_MSG (msgcat_message (MSGCAT_CATALOG_UTILS, MSGCAT_UTIL_SET_COMPACTDB,
                                             COMPACTDB_MSG_EXCEED_MAX_LEN),
                             DB_MAX_IDENTIFIER_LENGTH - DB_MAX_USER_LENGTH - 1);
      status = ER_FAILED;
      goto end;
    }
~
```

msg/en_US/utils.msg
joohok: You are right. This is the main concern for this modification. 
users can run it without specifying the owner name. 

But I think it's better to entice them to run by specifying the owner name to get the desired result.

src/executables/util_cs.c
joohok: Okay, I understand what you mean.
I will check a length of table_name.

eido5 (approved):

msg/en_US/utils.msg
wrlawodms: Ok, I understand. However, if omitting owner_name possibly makes a wrong result, I can't find the point to allow this. How about just preventing it?

joohok: There seem to be two ways.
1. Proceed as it is now, but if the owner is not entered, **the manual** specifies that a table owned by dba will be searched.
2. If the owner is not specified, an error occurred from the flashback.

I think 2 is better, and other utilities that take the class names as arguments should also be fixed like this.

but it is related to flashback spec, so I think we need to discuss it a bit more and decide.

wrlawodms: Yep, leave a comment here when it is decided, or have a discussion here by mentioning others.

src/executables/util_cs.c
hornetmj: trivial;

Is there any reason that this for loop moves here?

hornetmj (approved):

msg/en_US/utils.msg
joohok: The decision was made according to your opinion.
We will force the user to specify the owner.

joohok: The decision was made according to your opinion.
We will force the user to specify the owner.

wrlawodms: Thank you for letting me know!

src/executables/util_cs.c
```diff
@@ -4000,6 +4000,58 @@ tde (UTIL_FUNCTION_ARG * arg)
...
+      PRINT_AND_LOG_ERR_MSG (msgcat_message
+			     (MSGCAT_CATALOG_UTILS, MSGCAT_UTIL_SET_FLASHBACK, FLASHBACK_MSG_OWNER_NOT_SPECIFIED),
+			     table_name);
+
+      return ER_FAILED;
+    }
+
+  /* check length of owner name */
+  sub_len = STATIC_CAST (int, dot - table_name);
+  if (sub_len < 1 || sub_len >= SM_MAX_USER_LENGTH)
```
hornetmj: Does it need to check the 'sub_len < 1'  condition?

msg/en_US/utils.msg
```diff
@@ -1274,9 +1274,13 @@ $set 58 MSGCAT_UTIL_SET_FLASHBACK
 12 Transaction ID should be entered in %1$d seconds. The timeout can be adjusted through the 'flashback_timeout' setting in 'cubrid.conf'\n
 13 Flashback is already running. Flashback cannot be run by two processes.\n
 14 Flashback cannot run as standalone mode.\n
+15 Length of owner_name.class_name is required to be greater than 0 or less than %1$d bytes.\n
+16 Length of class name is required to be greater than 0 or less than %1$d bytes.\n
+17 Length of owner name is required to be greater than 0 or less than %1$d bytes.\n
```
wrlawodms:
```suggestion
15 Length of owner_name.class_name is required to be greater than 0 and less than %1$d bytes.\n
16 Length of class_name is required to be greater than 0 and less than %1$d bytes.\n
17 Length of owner_name is required to be greater than 0 and less than %1$d bytes.\n
```

src/executables/util_cs.c
```diff
@@ -4000,6 +4000,58 @@ tde (UTIL_FUNCTION_ARG * arg)
...
+      PRINT_AND_LOG_ERR_MSG (msgcat_message
+			     (MSGCAT_CATALOG_UTILS, MSGCAT_UTIL_SET_FLASHBACK, FLASHBACK_MSG_INVALID_OWNER_LENGTH),
+			     SM_MAX_USER_LENGTH);
+
+      return ER_FAILED;
+    }
+
+  /* check length of class name */
+  sub_len = STATIC_CAST (int, strlen (dot + 1));
+  if (sub_len < 1 || sub_len >= SM_MAX_IDENTIFIER_LENGTH - SM_MAX_USER_LENGTH)
```
wrlawodms: What is the maximum length of the class name in CUBRID? If it is SM_MAX_IDENTIFIER_LENGTH, this requirement prevents the flashback if the length of a user name + a length of a class name is bigger than SM_MAX_IDENTIFIER_LENGTH.

If so, the max length of owner_name.class_name, `table_name_len`,  should be SM_MAX_IDENTIFIER_LENGTH + SM_MAX_USER_LENGTH.

joohok: What has changed when user schema is applied is this length spec (`SM_MAX_IDENTIFIER_LENGTH`).
SM_MAX_IDENTIFIER_LEGNTH represents maximum value of (length of owner name + length of class name).

So, I compared class name length with (SM_MAX_IDENTIFIER_LENGTH - SM_MAX_USER_LENGTH)

msg/en_US/utils.msg
joohok: Oh, Thank you

src/executables/util_cs.c
wrlawodms: OK, I got it. Then, what about the backward compatibility? Do the existing table and user name have to be changed to fit that restriction? I don't think it's a flashback's issue, but I just wonder.

msg/en_US/utils.msg
```diff
@@ -1274,9 +1274,13 @@ $set 58 MSGCAT_UTIL_SET_FLASHBACK
...
+18 Owner name for '%1$s' is required to be specified.\n
```
hornetmj: How about reducing the number of error messages. My suggestion is the following.
- Invalid format of owner_name.class_name ('%1$s').
- The length of the owner_name must be less than %2$d and the length of the table_name must be less than %2$d.

hornetmj (approved):

src/executables/util_cs.c
joohok: Yes. As I know, the functions that check the length of the class name and owner name have been changed according to the specification. (`get_class_mops_from_file()`, `get_class_mops()`, `get_requested_classes()`,.. )

`SM_MAX_IDENTIFIER_LENGTH` is used for unique_name (owner_name.class_name), and unique name is used in server side instead of class name through parser where makes classname to unique_name.

src/executables/util_cs.c
joohok: because, users can input class_name without owner name, and just add 'dot' to the class_name like below
`cubrid flashback demodb .test_table` 

So, I added a condition `sub_len<1`

src/executables/util_cs.c
joohok: But, I don't know about the existing tables. I agree with you, it can cause a problem.

src/executables/util_cs.c
hornetmj: If then, you need to use 'sub_len <= 1 or sub_len < 2'. isn't it?

src/executables/util_cs.c
wrlawodms: Thank you for the answer 👍 .

eido5 (approved):

src/executables/util_cs.c
joohok: if owner name is 'a' , `sub_len` will be 1, and user 'a' could exist. 
`sub_len` means the length of owner_name.
length for a dot is not included in `sub_len`.

msg/en_US/utils.msg
joohok: > Invalid format of owner_name.class_name ('%1$s').
It seems better

In the beginning, as your suggestion, I tried to reduce the number of error messages. 
I thought, however, it would be nice for the user to output an error according to each situation, so I modified it like this.

Also, the length of owner_name and class_name must both be greater than 0. 
If I put them all in one message, the message will be too long and the meaning will be unclear.

src/executables/util_cs.c
hornetmj: Oh. I understand.

msg/en_US/utils.msg
joohok: I will change it following your suggestion.

youngjinj (approved):

src/executables/util_cs.c
```diff
@@ -4000,6 +4000,81 @@ tde (UTIL_FUNCTION_ARG * arg)
...
+{
+  int table_name_len = STATIC_CAST (int, strlen (table_name));
+  int sub_len = 0;
+  const char *dot = NULL;
+
+  if (table_name_len >= SM_MAX_IDENTIFIER_LENGTH)
+    {
+      PRINT_AND_LOG_ERR_MSG (msgcat_message
+			     (MSGCAT_CATALOG_UTILS, MSGCAT_UTIL_SET_FLASHBACK,
+			      FLASHBACK_MSG_EXCEED_MAX_CLASSNAME_LENGTH), SM_MAX_USER_LENGTH,
```
hornetmj, line 4014: When I tested, I was able to create a user with a length of 31 instead of 32.

src/executables/util_cs.c
```diff
@@ -4000,6 +4000,81 @@ tde (UTIL_FUNCTION_ARG * arg)
...
+			     table_name);
+
+      return ER_FAILED;
+    }
+
+  /* check length of owner name */
+  sub_len = STATIC_CAST (int, dot - table_name);
+  if (sub_len < 1)
+    {
+      /* owner name is not specified (e.g. '.table' */
```
```suggestion
      /* owner name is not specified (e.g. '.table') */
```

src/executables/util_cs.c
```diff
@@ -4000,6 +4000,81 @@ tde (UTIL_FUNCTION_ARG * arg)
...
+    {
+      /* class name is not specified (e.g. 'dba.') */
+      PRINT_AND_LOG_ERR_MSG (msgcat_message
+			     (MSGCAT_CATALOG_UTILS, MSGCAT_UTIL_SET_FLASHBACK, FLASHBACK_MSG_INVALID_CLASSNAME_FORMAT),
+			     table_name);
+
+      return ER_FAILED;
+    }
+
+  if (sub_len >= SM_MAX_IDENTIFIER_LENGTH - SM_MAX_USER_LENGTH)
```
hornetmj, line 4065:
```suggestion
  if (sub_len > SM_MAX_IDENTIFIER_LENGTH - SM_MAX_USER_LENGTH)
```

src/executables/util_cs.c
joohok: Thanks for testing.

Yes, that is intended result.

The length of the owner_name.class_name in 'user schema' is described as follows.
> Length = user_name (31 bytes) + dot (1 byte) + object_name (222 bytes) + null (\0) (1 byte)

So, I asked the user to set the owner name to less than 32 bytes through an error message.

src/executables/util_cs.c
joohok: If this change allowed the user to use class names of 223 bytes, it would be against the specification of the user schema.

src/executables/util_cs.c
joohok: Thanks

wrlawodms (approved):

hornetmj:
@mhoh3963 Please review and merge this PR.

joohok:
@mhoh3963 
It is said that QA build is performed at 3 o'clock. Therefore, I would appreciate it if you could do a review before that.

mhoh3963 (approved):

PR #3537, opened 2022-04-20
[CBRD-24301] Clearly display the owner in "csql> ;info schema" and "show tables".

youngjinj:
http://jira.cubrid.org/browse/CBRD-24301

### Clearly display the owner in "csql> ;info schema" and "show tables".
- In "csql> ;info schema", the owner name is repeatedly output.
  - Output owner_name only once.
  - The owner name is not displayed in the system table.
- In "show tables", the owner name is not displayed, so tables with the same name cannot be distinguished.
  - Add "Owner" column in "show full tables".


src/object/object_print.c
```diff
@@ -385,24 +384,15 @@ help_class_names (const char *qualifier)
...
-			{
-			  snprintf (buffer, sizeof (buffer) - 1, "%s.%s", "unknown_user", cname);
-			}
-		      db_value_clear (&owner_name);
+		      snprintf (buffer, sizeof (buffer) - 1, "%s", unique_name);
 		    }
 		  else
 		    {
-		      snprintf (buffer, sizeof (buffer) - 1, "%s", cname);
+		      snprintf (buffer, sizeof (buffer) - 1, "%s", sm_remove_qualifier_name (unique_name));
```
hgryoo: What If `sm_remove_qualifier_name` returns NULL?

ctshim (approved):

src/object/object_print.c
youngjinj: Since the sm_remove_qualifier_name() function returns a const char * type,
how about changing it like the one below?

```
if (name == NULL || name[0] == '\0')
  {
    ERROR_SET_WARNING_1ARG (error, ER_SM_INVALID_NAME, name);
    return "";
  }
```

shparkcubrid (approved):

src/object/object_print.c
youngjinj: Thank you.
Added exception handling code for returning NULL.

hgryoo:
please sync with develop. It will pass circleci tests then you can merge this PR.

hgryoo (approved):

beyondykk9 (approved):

hyunikn (approved):

PR #3538, opened 2022-04-21
[LETS-454] add page desync asserts

cristiarg:
http://jira.cubrid.org/browse/LETS-454

Added `ASSERT_NO_ERROR` in `btree_find_boundary_leaf_with_repl_desync_check` failed with sometimes with `ER_INTERRUPTED` (this occured not in scalability context, but in monolithic context).
Replace with a more narrow assertion that, in the functions where page desynch might occur, the error, if existing, is something else than `ER_PAGE_AHEAD_OF_REPLICATION`.
This will help in pinpointing two things, once the passive transaction server tests are in place:
- if page desync is occuring and it is not cleared
- where on the calling stack it should be cleared

### DIFF
--- a/src/base/error_manager.h
+++ b/src/base/error_manager.h
@@ -311,10 +311,11 @@ extern "C"
 #endif
 
 STATIC_INLINE void
-ASSERT_NOT_ERROR (int not_error_code)
+ASSERT_NOT_ERROR (const int not_error_code)
 {
+  assert (not_error_code != NO_ERROR);
   const int error_code = er_errid ();
-  assert (error_code == NO_ERROR || error_code != not_error_code);
+  assert (error_code != not_error_code);
 }
 
 #ifdef __cplusplus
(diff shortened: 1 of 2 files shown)


src/base/error_manager.h
```diff
@@ -309,6 +310,13 @@ extern "C"
 }
 #endif
 
+STATIC_INLINE void
+ASSERT_NOT_ERROR (int not_error_code)
+{
+  const int error_code = er_errid ();
+  assert (error_code == NO_ERROR || error_code != not_error_code);
```
razvanradulescu: Is there any case other then (not_error_code == NO_ERROR), for which 
the part  ("error_code == NO_ERROR") is not redundant ?

razvanradulescu (approved):

src/base/error_manager.h
cristiarg: I wanted to make sure that it is not called with `NO_ERROR`. Maybe it should just `assert (not_error_code != NO_ERROR)`. Because NO_ERROR already has a dedicated macro.

PR #3544, opened 2022-04-25
[CBRD-23846] [regression] Fix unexpected "SELECT authorization failure"

hgryoo:
http://jira.cubrid.org/browse/CBRD-23846

**Purpose**
Some regression tests failed caused by "SELECT authorization failure". I found some code that authentication check for java stored procedure is not disabled.

**Implementation**
Owner of stored procedure should not be checked. When getting MOP of a stored procedure, the authentication check was disabled. But when getting any attribute from the MOP, `SELECT authorization failure` occurred.

**Remarks**
For the `CALL` statement, it occurs the error. But for the `SELECT` statement, the error does not occur.


youngjinj (approved):

shparkcubrid (approved):

hyunikn (approved):

src/jsp/jsp_cl.c
```diff
@@ -1328,6 +1329,8 @@ jsp_make_method_sig_list (PARSER_CONTEXT * parser, PT_NODE * node, method_sig_li
     DB_OBJECT *mop_p = jsp_find_stored_procedure (parsed_method_name);
     if (mop_p)
       {
+	AU_DISABLE (save);
```
ctshim: When mop is null, AU_ENABLE (save) is performed with the 'save' variable never initialized.

beyondykk9 (approved):

src/jsp/jsp_cl.c
hgryoo: Thank you!. I'm gonna move this line out of the if-statement.

ctshim (approved):

PR #3545, opened 2022-04-25
[CBRD-24311] When using an internal select query, an mvcc snapshot occurs and the isolation test result is different.

youngjinj:
http://jira.cubrid.org/browse/CBRD-24311

### When using an internal select query, an mvcc snapshot occurs and the isolation test result is different.	
- In Repeatable Read, the MVCC snapshot is created when the SELECT query is executed. Even if the SELCET query is not explicitly executed, the MVCC snapshot is created when the index is scanned.
```
/* src/storage/btree.c */

BTREE_SEARCH
xbtree_find_unique (THREAD_ENTRY * thread_p, BTID * btid, SCAN_OPERATION_TYPE scan_op_type, DB_VALUE * key,
            OID * class_oid, OID * oid, bool is_all_class_srch)
{
  ~
  if (logtb_find_current_isolation (thread_p) >= TRAN_REP_READ || (find_unique_helper.lock_mode >= S_LOCK))
    {
      /*
       * Acquire snapshot in RR if not already acquired. This is needed since
       * the transaction need to know the actual visible objects - before
       * instance locking. In this way future commands of current transaction
       * may correctly detect visible objects.
       */
      (void) logtb_get_mvcc_snapshot (thread_p);
    }
~
```
- For this reason, MVCC snapshots are created when looking up a user, getting the current username, and looking up a Synonym. Creating an MVCC snapshot when executing an internal query on the system may not be the intended result of the user. Therefore, in case of index scan of the db_user and _db_synonym system tables, do not create an MVCC snapshot so that it does not differ from the previous answer in the isolation test case.

### DIFF
--- a/src/object/authenticate.c
+++ b/src/object/authenticate.c
@@ -1192,7 +1192,8 @@ au_find_user (const char *user_name)
      * When au_login_method () is called, Au_user_name is not changed.
      */
     const char *current_schema_name = sc_current_schema_name ();
-    if (current_schema_name && intl_identifier_casecmp (current_schema_name, user_name) == 0)
+    if (current_schema_name && current_schema_name[0] != '\0'
+	&& intl_identifier_casecmp (current_schema_name, user_name) == 0)
       {
 	return Au_user;
       }
@@ -6308,7 +6309,7 @@ au_user_name (void)
        * When au_login_method () is called, Au_user_name is not changed.
        */
       const char *current_schema_name = sc_current_schema_name ();
-      if (current_schema_name)
+      if (current_schema_name && current_schema_name[0] != '\0')
 	{
 	  return ws_copy_string (current_schema_name);
 	}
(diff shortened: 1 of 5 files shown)


beyondykk9 (approved):

hgryoo (approved):

shparkcubrid (approved):

hyunikn (approved):

src/object/authenticate.c
```diff
@@ -6283,6 +6300,21 @@ au_user_name (void)
...
 
+      /*
+       * To reduce unnecessary code execution,
+       * the current schema name can be used instead of the current user name.
+       * 
+       * Au_user_name cannot be used because it does not always store the current user name.
+       * When au_login_method () is called, Au_user_name is not changed.
+       */
+      const char *sc_name = sc_current_schema_name ();
+      char upper_sc_name[DB_MAX_USER_LENGTH];
```
ctshim, line 6311: Even if the values of DB_MAX_USER_LENGTH and DB_MAX_SCHEMA_LENGTH are currently the same, it would be better to use DB_MAX_SCHEMA_LENGTH semantically.

`char upper_sc_name[DB_MAX_SCHEMA_LENGTH ];`

For reference, the size of schema_def is as follows
  ` char name[DB_MAX_SCHEMA_LENGTH * INTL_UTF8_MAX_CHAR_SIZE + 4];`

PR #3548, opened 2022-04-26
[CBRD-24060] Requires synonym support in unloaddb, loaddb utility.

youngjinj:
http://jira.cubrid.org/browse/CBRD-24060

Requires synonym support in unloaddb, loaddb utility.


beyondykk9 (approved):

ctshim (approved):

shparkcubrid (approved):

src/executables/unload_schema.c
```diff
@@ -815,6 +828,163 @@ export_serial (print_output & output_ctx)
...
+	}
+
+      synonym_name = db_get_string (&values[SYNONYM_NAME]);
+      synonym_owner = db_get_object (&values[SYNONYM_OWNER]);
+      synonym_owner_name = db_get_string (&values[SYNONYM_OWNER_NAME]);
+      is_public = db_get_int (&values[SYNONYM_IS_PUBLIC]);
+      target_name = db_get_string (&values[SYNONYM_TARGET_NAME]);
+      target_owner_name = db_get_string (&values[SYNONYM_TARGET_OWNER_NAME]);
+
+      if (!is_dba_group_member && !ws_is_same_object (Au_user, synonym_owner))
```
hyunikn, line 951: This check and continuing can be done earlier in the while loop, that is, before the for-loop.
It will reduce the execution time if db_query_get_tuple_value() is an expensive operation.

hyunikn (approved):

hgryoo (approved):

src/executables/unload_schema.c
youngjinj: Thank you. I agree with that.

However, unloading a schema rather than an object is not a big load,
so I will invest more in code brevity.

PR #3550, opened 2022-04-27
[CBRD-24313] Don't push the correlated predicate which includes dblink subquery

beyondykk9:
http://jira.cubrid.org/browse/CBRD-24313

The predicate should be pushed to the subquery for performance enhance, however, the correlated predicate should be handled carefully. The subquery includes dblink is executed at remote server not current local server, so the correlated predicate can not be handled at the server.

We should prevent from pushing the correlated predicate when the query has a dblink subquery.

The below query should be executed successfully.

```
select DISTINCT ( 
                  select col8 
                  from dblink(
                         srv1,
                         'select col1, col8 from remote_t') as t (col1 int, col8 varchar)
                  where t.col1 = local_t.col1
               ) col8
from local_t local_t;
```

to be rewritten:
```
select distinct (
	select t.col8 
	from (
		select [_dbl].col1, [_dbl].col8 
		from DBLINK(srv1, 'select col1, col8 from remote_t') 
                         as [_dbl](col1 integer, col8 varchar)) t (col1, col8) 
		where t.col1=local_t.col1
	) 
from local_t local_t
```

The below rewritten query is not correct
```
select distinct (
        select t.col8 
        from (
                select [_dbl].col1, [_dbl].col8 
                from DBLINK(srv1, 'SELECT * FROM (select col1, col8 from remote_t) AS _r(col1, col8) WHERE col1=col1'
        ) as [_dbl](col1 integer, col8 varchar)) t (col1, col8)) 
from local_t local_t
```

### DIFF
--- a/src/parser/parser.h
+++ b/src/parser/parser.h
@@ -530,6 +530,7 @@ extern "C"
   extern PT_NODE *pt_get_cursor (const PT_HOST_VARS * hv);
   extern PT_NODE *pt_get_parameters (PARSER_CONTEXT * parser, PT_NODE * statement);
 
+  extern bool pt_has_dblink (PARSER_CONTEXT * parser, PT_NODE * node);
   extern bool pt_has_aggregate (PARSER_CONTEXT * parser, PT_NODE * node);
   extern bool pt_has_analytic (PARSER_CONTEXT * parser, PT_NODE * node);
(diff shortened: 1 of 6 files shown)


beyondykk9:
The issue is closed by another PR (#3559)

PR #3554, opened 2022-04-27
[CBRD-24309] Change the logic for checking if schema has been changed

joohok:
http://jira.cubrid.org/browse/CBRD-24309

Purpose

Changed the logic to check whether the schema has been changed. 
Previously, it was checked by checking the number of columns (attributes).
but with this PR, it was modified to check through the **representation ID**

Implementation

1. Returns an error if the latest representation ID is different from the representation ID in RECDES.
2. Remove the existing logic for detecting schema change

### DIFF
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -12449,6 +12449,26 @@ error:
     } \
   while (0)
 
+static bool
+cdc_check_if_schema_changed (RECDES * recdes, HEAP_CACHE_ATTRINFO * attr_info)
+{
+  REPR_ID reprid = NULL_REPRID;
+  REPR_ID latest_reprid = NULL_REPRID;
+
+  reprid = or_rep_id (recdes);
+  latest_reprid = attr_info->last_classrepr->id;
+
+  if (reprid != last_reprid)
+    {
+      /* schema has been changed */
+      return true;
+    }
+  else
+    {
+      return false;
+    }
+}
+
 int
 cdc_make_dml_loginfo (THREAD_ENTRY * thread_p, int trid, char *user, CDC_DML_TYPE dml_type,
 		      OID classoid, RECDES * undo_recdes, RECDES * redo_recdes, CDC_LOGINFO_ENTRY * dml_entry,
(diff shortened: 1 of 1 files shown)


src/transaction/log_manager.c
```diff
@@ -12449,6 +12449,26 @@ cdc_make_error_loginfo (int trid, char *user, CDC_DML_TYPE dml_type, OID classoi
     } \
   while (0)
 
+static bool
```
hornetmj, line 12467: Please leave a comment about the followings. So let other developers know why you don't service if there is a change of any schema.
1. cdc - design issue (Issues due to the difference between the time the client obtains schema information through jdbc and the time it takes to interpret log records)
2. flashback - cannot flashback even if you generate any SQL formatted by old schema

joohok: OK, I will add the comment for it.

eido5 (approved):

src/transaction/log_manager.c
```diff
@@ -12449,6 +12449,41 @@ cdc_make_error_loginfo (int trid, char *user, CDC_DML_TYPE dml_type, OID classoi
...
+
+  if (reprid != latest_reprid)
+    {
+      /* schema has been changed */
+      return true;
+    }
+  else
+    {
+      return false;
+    }
```
```suggestion
      /* schema has been changed */
      return reprid != latest_reprid;
```

src/transaction/log_manager.c
```diff
@@ -12449,6 +12449,41 @@ cdc_make_error_loginfo (int trid, char *user, CDC_DML_TYPE dml_type, OID classoi
...
+}
```
wrlawodms: The above or
```suggestion
{
  /* schema has been changed */
  return or_rep_id (recdes) != attr_info->last_classrepr->id;
}
```

wrlawodms (approved):

src/transaction/log_manager.c
joohok: Wow.. this is super simple, thanks

hornetmj (approved):

PR #3555, opened 2022-04-27
[CBRD-24279] Check if user inputs the system class name to flashback

joohok:
http://jira.cubrid.org/browse/CBRD-24279

Purpose

As the owner name is omitted for the system table (#3537), 
the error message is unclear when searching the system class by flashback.

For example, 
1. cubrid flashback demodb _db_class
    invalid format for owner_name.class_name (_db_class)
2. cubrid flashback demodb dba._db_class
    Table (dba._db_class) does not exist in database

In order to clarify the error situation, 
logic to check whether the system table exists is added, and an error message is also added.

Implementation
1. used sm_check_system_class_by_name() to check system class
2. Since sm_check_system_class_by_name() can be used only when the language module is initialized, 
    it has been moved after db_restart().


wrlawodms (approved):

eido5 (approved):

msg/en_US/utils.msg
```diff
@@ -1278,6 +1278,7 @@ $set 58 MSGCAT_UTIL_SET_FLASHBACK
 14 Flashback cannot run as standalone mode.\n
 15 The length of owner_name is required to be less than %1$d and the length of class_name is required to be less than %2$d bytes.\n
 16 Invalid format for owner_name.class_name ('%1$s').\n
+17 System catalog classes (%1$s) are not supported in flashback.\n
```
hornetmj:
```suggestion
17 The system catalog class (%1$s) is not supported in flashback.\n
```

hornetmj (approved):

PR #3560, opened 2022-04-28
[CBRD-24292] Checksumdb results performed on the slave are different from before.

eido5:
http://jira.cubrid.org/browse/CBRD-24292

**Purpose**
[checksumdb provides a simple way to check replication integrity](https://www.cubrid.org/manual/en/11.0/ha.html?highlight=checksumdb#checksumdb)
And checksum integrity is invalid in the `db_ha_checksum_schema` even when there is no other different schemas between master node and slave node.
(Checksum result of `db_ha_checksum_schema` is stored in the catalog class `db_ha_checksum`)

So, we found that there are no catalog classes registration about the `db_ha_checksum_schema` and `db_ha_checksum_schema`. And then we registered them not to occur the error.

**Implementation**
we registered `db_ha_checksum_schema` and `db_ha_checksum_schema` in the `SYSTEM_CLASS_DEF`. And we registered `db_ha_checksum_schema` and `db_ha_checksum_schema` in the  'Meta OID information' of `transform.h`

**Remarks**
N/A

### DIFF
--- a/msg/de_DE.utf8/utils.msg
+++ b/msg/de_DE.utf8/utils.msg
@@ -1232,7 +1232,7 @@ gültigen Optionen:\n\
   -e, --exclude-class-file=FILE Eingabedatei der Tabellennamen, um auszuschließen; entweder -i oder -e,\n\
                                 verwendet werden, jedoch nicht beides\n\
   -t, --timeout=NUMBER          Abfrage-Timeout in Millisekunden; Standard: 1000 ms\n\
-  -n, --table-name=NAME         Tabellennamen, um Prüfsumme Ergebnisse zu speichern; Standard: db_ha_checksum\n\
+  -n, --table-name=NAME         Tabellennamen, um Prüfsumme Ergebnisse zu speichern; Standard: dba.db_ha_checksum\n\
   -r, --report-only             Bericht Replikationsintegrität auf der Grundlage Prüfsummenberechnung\n\
                                 alle Optionen außer --table-name werden ignoriert\n\
       --resume                  Wiederaufnahme der Berechnung Prüfsumme\n\
--- a/msg/de_DE.utf8/utils.msg
+++ b/msg/de_DE.utf8/utils.msg
@@ -1232,7 +1232,7 @@ gültigen Optionen:\n\
   -e, --exclude-class-file=FILE Eingabedatei der Tabellennamen, um auszuschließen; entweder -i oder -e,\n\
                                 verwendet werden, jedoch nicht beides\n\
   -t, --timeout=NUMBER          Abfrage-Timeout in Millisekunden; Standard: 1000 ms\n\
-  -n, --table-name=NAME         Tabellennamen, um Prüfsumme Ergebnisse zu speichern; Standard: db_ha_checksums\n\
+  -n, --table-name=NAME         Tabellennamen, um Prüfsumme Ergebnisse zu speichern; Standard: db_ha_checksum\n\
   -r, --report-only             Bericht Replikationsintegrität auf der Grundlage Prüfsummenberechnung\n\
                                 alle Optionen außer --table-name werden ignoriert\n\
       --resume                  Wiederaufnahme der Berechnung Prüfsumme\n\
--- a/msg/en_US.utf8/utils.msg
+++ b/msg/en_US.utf8/utils.msg
@@ -1233,7 +1233,7 @@ valid options:\n\
   -i, --include-class-file=FILE input FILE of table names to include\n\
   -e, --exclude-class-file=FILE input FILE of table names to exclude; either -i or -e can be used, not both\n\
   -t, --timeout=NUMBER          query timeout in milliseconds; default: 1000 ms\n\
-  -n, --table-name=NAME         table name to save checksum results; default: db_ha_checksum\n\
+  -n, --table-name=NAME         table name to save checksum results; default: dba.db_ha_checksum\n\
   -r, --report-only             report replication integrity based on checksum calculation\n\
                                 any options except --table-name will be ignored\n\
       --resume                  resume calculating checksum\n\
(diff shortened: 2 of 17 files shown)


hornetmj:
@eido5 When you describe the PR, follow the format guided on the cubrid development process.

joohok:
Please check the PR title, and

It would be nice if you could explain what the purpose of this PR is and what kind of situation you are trying to fix. 
From the description in the issue, I don't know what PR is for. I hope you can explain why the system class was added.

wrlawodms:
@eido5 The Jira issue is a bug report, so we don't know how you deal with the issue. Please elaborate a little bit.

youngjinj:
Please refer to the PR I was going to proceed with.
- https://github.com/CUBRID/cubrid/pull/3556

The code below should be fixed.
- AS-IS
```
/* src/executables/checksumdb.c */

static bool
chksum_need_skip_table (const char *table_name, CHKSUM_ARG * chksum_arg)
{
  ~
  if (table_name == NULL || (strcasecmp (table_name, chksum_result_Table_name) == 0)
      || (strcasecmp (table_name, chksum_schema_Table_name) == 0))
    {
      return true;
    }
~
```
- TO-BE
  - strcasecmp () → pt_user_specified_name_compare ()
  - If the user uses the "-n, --table-name=STRING" option, "owner_name.chksum_schema_Table_name" and "chksum_schema_Table_name" should be compared.
```
/* src/executables/checksumdb.c */

static bool
chksum_need_skip_table (const char *table_name, CHKSUM_ARG * chksum_arg)
{
  ~
  if (table_name == NULL || (pt_user_specified_name_compare (table_name, chksum_result_Table_name) == 0)
      || (pt_user_specified_name_compare (table_name, chksum_schema_Table_name) == 0))
    {
      return true;
    }
~
```

eido5:
I modified the PR tile and description. And there is one more commit. Please wait for one more commit. It will be committed soon.

youngjinj (approved):

eido5:
The review is consisted three parts.

1. Usage message typo
2. COUNT() int cast
3. `dba.` is attached before CHKSUM_DEFAULT_TABLE_NAME.
ex) dba.db_ha_checksum_schema

And,
The testcase `cubrid checksumdb -n pp demodb` will be blocked. 
The testcase only `cubrid checksumdb -n [user_name].pp demodb` will be permitted.

wrlawodms (approved):

msg/de_DE.utf8/utils.msg
```diff
@@ -1232,7 +1232,7 @@ gültigen Optionen:\n\
   -e, --exclude-class-file=FILE Eingabedatei der Tabellennamen, um auszuschließen; entweder -i oder -e,\n\
                                 verwendet werden, jedoch nicht beides\n\
   -t, --timeout=NUMBER          Abfrage-Timeout in Millisekunden; Standard: 1000 ms\n\
-  -n, --table-name=NAME         Tabellennamen, um Prüfsumme Ergebnisse zu speichern; Standard: db_ha_checksums\n\
+  -n, --table-name=NAME         Tabellennamen, um Prüfsumme Ergebnisse zu speichern; Standard: db_ha_checksum\n\
```
joohok: Shouldn't this also be changed to dba.db_ha_checksum?

eido5: I am sorry to reply so late.

I think default value of class is better that `db_ha_checksum`. 

Because I think it is similar normal table name. 
(There is no description, default table owner  is 'public' . )The verification of owner is up to the user. 

Many tables have each owner but in CUBRID, user should find the owner himself.

hornetmj (approved):

msg/de_DE.utf8/utils.msg
eido5: But I think your suggestion is also nice. What about fix the manual? 

https://www.cubrid.org/manual/en/11.0/ha.html?highlight=checksumdb#cmdoption-checksumdb-n
From
`You can specify a table name to save checksum results. (default: db_ha_checksum)`

to

`You can specify a table name to save checksum results. (default: db_ha_checksum, default owner: dba)`

joohok: I think it is essential to give accurate information through help messages. Depending on who owns the table, there can be more than one table named db_ha_checksum, and I think it is necessary to tell which table the information is stored in. Even though the help message can provide enough information, I don't understand why the user has to find out which table the information is stored in with only the name db_ha_checksum.

> Because I think it is similar normal table name.
> (There is no description, default table owner is 'public' . )

I don't understand what this means

eido5: `Because I think it is similar normal table name.
(There is no description, default table owner is 'public' . )`

This is what I mean, 
For example, I can create t0 table by csql demodb -c "create table t0(col1 int primary key,col2 varchar(10),col3 varchar(3));". And its owner is "public". But There is no usage that "public" is default table owner.

So, I suggested that manual is better to explain default owner.

joohok: I don't know why you are looking for the default owner. Owner means the subject who created the table. Since csql was executed without the -u option, csql was executed as a public user, and since the public user created a table, the owner of the table only became public.

joohok: And users don't know that utility is executed by the dba user.

hornetmj: It would be nice to follow @joohok's suggestion. In this approach, tables with the same name but different owners can exist(e.g. user1.db_ha_checksum, user2.db_ha_checksum, dba.db_ha_checksum ...). If the usages doesn't describe the owner of the db_ha_checksum, the user doesn't know which db_ha_checksum table they're talking about.

eido5: `I don't know why you are looking for the default owner.`

Because db_ha_checksum of default owner is dba.
(cubrid checksumdb db_name)

`users don't know that utility is executed by the dba user.`
Right, I agree. yes, it is definitely essential to explain. 

So, Could you tell me that fix the manual? I think usage should be included all cases. ("show tables" shows db_ha_checksum.)

+) Actually I am trying to negotiate and trying to persuade myself:)

joohok: if you change this to dba.db_ha_checksum, you also need to edit the manual (https://www.cubrid.org/manual/en/11.0/ha.html#checksumdb). 

* description of the option `--table-name`
* examples

joohok (approved):

msg/de_DE.utf8/utils.msg
eido5: Thank you I will apply it. :)

PR #3563, opened 2022-04-29
[CBRD-24304] The problem that the optimal plan cannot be found because the first node is not properly selected

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-24304

Deleting a specific node from the first node is to reduce the number of plan comparisons.
The current logic is to remove nodes with both high cost and cardinality. I add routines to whether index scan is possible in inner. This is the routines that was previously written only on one side.
Nevertheless, deleting the first node can produce an inefficient plan. This can lead to very inefficient index scans as shown below.

```
Select * from t1 a, t2 b
where a.col1 = b.col1
    and a.col2 = b.col2

Index of t1 (a,b) NDV(1,100000)
index of t2 (a) NDV (1)
==> Inefficient index scan when t1->t2 order
```

Also, the number of permutations is N!. The number of permutations that remove the first node is (N-first node)*(N-1)! This is not an effective way to reduce the number of permutations.
For this reason, I would like to completely remove the first node-related routines. But here, I modify the first node not to delete up to 10 nodes.


src/optimizer/query_planner.c
```diff
@@ -8622,6 +8633,48 @@ qo_search_partition_join (QO_PLANNER * planner, QO_PARTITION * partition, BITSET
...
+			    }
+			}
+		      else
+			{
+			  /* edge between f_node and other_node */
+			  found_other_edge = true;
+			}
+		    }
+
+		  if (!found_f_edge || found_other_edge)
```
ctshim, line 8673: Maybe found_r_edge should come instead of found_f_edge?

ctshim: Maybe found_r_edge should come instead of found_f_edge?

ctshim: If the newly added code is applied, the` if (cmp == PLAN_COMP_LT) block` and the `if (cmp == PLAN_COMP_GT) block` are almost identical, so extracting it as a function would be better for readability and subsequent maintenance.

shparkcubrid:
```suggestion
		  if (!found_r_edge || found_other_edge)
```

shparkcubrid: I fixed it. thanks.
Rewriting as a function is a good idea. However, I don't want to make any further modifications to this routine which can be fully removed.

beyondykk9 (approved):

ctshim (approved):

hyunikn (approved):

youngjinj (approved):

hgryoo (approved):

PR #3565, opened 2022-04-29
[LETS-288] replicate assigned mvccid completion

cristiarg:
http://jira.cubrid.org/browse/LETS-288

During replication, if LOG_ASSIGNED_MVCCID is found, save the (trid, MVCCID) pair for MVCC completion.
Upon COMMIT/ABORT, if found, also complete the mvccid for that transaction.

### DIFF
--- a/src/transaction/log_replication.cpp
+++ b/src/transaction/log_replication.cpp
@@ -249,7 +249,7 @@ namespace cublog
 	  case LOG_COMMIT:
 	    if (m_replicate_mvcc)
 	      {
-		m_replicator_mvccid->complete_mvccid (header.trid, true);
+		m_replicator_mvccid->complete_mvcc (header.trid, replicator_mvcc::COMMITTED);
 	      }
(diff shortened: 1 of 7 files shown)


src/transaction/log_replication.cpp
```diff
@@ -233,10 +247,18 @@ namespace cublog
 	    break;
 	  }
 	  case LOG_COMMIT:
+	    if (m_replicate_mvcc)
+	      {
+		m_replicator_mvccid->complete_mvccid (header.trid, true);
```
adrianbrab: Maybe add `const bool commited_transaction = true` or a comment so it is a bit easier to understand what the second argument stands for.

adrianbrab (approved):

src/transaction/log_replication.cpp
cristiarg: Correct, there should be some constants. Done.

razvanradulescu (approved):

PR #3567, opened 2022-05-01
[CBRD-24321] after merged to synonym, some test cases occurred to fail.

youngjinj:
http://jira.cubrid.org/browse/CBRD-24321

Prevent ER_LC_UNKNOWN_CLASSNAME from occurring in sm_find_synonym ().

### DIFF
--- a/src/object/schema_manager.c
+++ b/src/object/schema_manager.c
@@ -5406,21 +5406,23 @@ sm_find_class_with_purpose (const char *name, bool for_update)
       return class_mop;
     }
 
-  /* backup error */
-  error = er_errid ();
-
-  synonym_mop = sm_find_synonym (realname);
-  if (synonym_mop)
+  /* class_mop == NULL */
+  if (er_errid () == ER_LC_UNKNOWN_CLASSNAME)
     {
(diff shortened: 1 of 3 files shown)


hgryoo (approved):

shparkcubrid (approved):

hyunikn (approved):

src/object/schema_manager.c
```diff
@@ -5448,15 +5460,10 @@ sm_find_synonym (const char *name)
   AU_DISABLE (save);
   synonym_obj = obj_find_unique (synonym_class_obj, "unique_name", &value, AU_FETCH_READ);
   AU_ENABLE (save);
-  if (synonym_obj == NULL)
+
```
ctshim, line 5465: When obj_find_unique() returns null, not all cases come with er_set() performed.
Therefore, as before, it is necessary to check whether synonym_obj is null.

ctshim (approved):

beyondykk9 (approved):

src/object/schema_manager.c
youngjinj: When synonym_obj is null, it's okay to not set an error.

PR #3570, opened 2022-05-02
[LETS-423] Separate log replicator for page server and passive transaction server - Integrate the atomic_replicator_helper class

adrianbrab:

In order for the atomic replication to work on PTS a new type replicator has to be created. The new `atomic_replicator` will replace the replicator for PTS. 

The integration of the `atomic_replication_helper` is the first step towards the functioning new replicator.
The following PRs will differentiate the replicator types further.


cubrid/CMakeLists.txt
```diff
@@ -309,6 +309,7 @@ set(METHOD_SOURCES
 
 set(TRANSACTION_SOURCES
   ${TRANSACTION_DIR}/atomic_replication_helper.cpp
+  ${TRANSACTION_DIR}/atomic_replicator.cpp
```
cristiarg: rename both files to start with `log_replication_..`

cristiarg: I see the files have actually been renamed. So the code does not compile ...

src/transaction/log_replication_atomic.hpp
```diff
@@ -0,0 +1,52 @@
...
+      atomic_replicator (const atomic_replicator &) = delete;
+      atomic_replicator (atomic_replicator &&) = delete;
+
+      ~atomic_replicator () = default;
+
+      atomic_replicator &operator= (const atomic_replicator &) = delete;
+      atomic_replicator &operator= (atomic_replicator &&) = delete;
+
+    private:
+      void redo_upto (cubthread::entry &thread_entry, const log_lsa &end_redo_lsa);
```
cristiarg: `override`

cubrid/CMakeLists.txt
```diff
@@ -335,6 +335,7 @@ set(TRANSACTION_SOURCES
   ${TRANSACTION_DIR}/log_recovery_redo.cpp
   ${TRANSACTION_DIR}/log_recovery_redo_parallel.cpp
   ${TRANSACTION_DIR}/log_replication.cpp
+  ${TRANSACTION_DIR}/log_replication_atomic.cpp
```
cristiarg, line 338: also rename `atomic_replication_helper` to have it sit near the others; rename class too

src/transaction/log_replication_atomic.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+#include "atomic_replication_helper.hpp"
+#include "log_replication.hpp"
+
+namespace cublog
+{
+  class atomic_replicator : replicator
+  {
+    public:
+      atomic_replicator () = default;
```
cristiarg: the ctor is used in `passive_tran_server::start_log_replicator` and while some params can be internalized, `start_lsa` must remain a parameter; since `cublog::replicator` does not have a default ctor, the code will not compile

src/transaction/log_replication_atomic.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+      atomic_replicator (const atomic_replicator &) = delete;
+      atomic_replicator (atomic_replicator &&) = delete;
+
+      ~atomic_replicator () = default;
```
cristiarg: `override`; I think `cublog::replicator` does not have a virtual dtor; big oopsie

src/transaction/log_replication_atomic.cpp
```diff
@@ -0,0 +1,173 @@
...
+	    m_most_recent_trantable_snapshot_lsa.store (m_redo_lsa);
+	    break;
+	  case LOG_START_ATOMIC_REPL:
+	  case LOG_SYSOP_ATOMIC_START:
+	    if (m_atomic_helper.is_part_of_atomic_replication (header.trid))
+	      {
+		// nested atomic replication
+		assert (false);
+	      }
+	    m_atomic_helper.start_new_atomic_replication_sequence (header.trid);
```
cristiarg: function returns error code but it is ignored; either treat error code or don't return

src/transaction/atomic_replication_helper.cpp
```diff
@@ -125,26 +110,19 @@ namespace cublog
...
 
-  atomic_replication_helper::atomic_replication_unit::~atomic_replication_unit ()
+  atomic_replication_helper::atomic_replication_unit::atomic_replication_unit (
+	  atomic_replication_helper::atomic_replication_unit &&that)
+    : m_record_lsa {std::move (that.m_record_lsa)}
+    , m_vpid {std::move (that.m_vpid)}
+    , m_record_index {std::move (that.m_record_index)}
   {
-    PGBUF_CLEAR_WATCHER (&m_watcher);
+    assert (m_page_ptr == nullptr);
```
cristiarg: comment
```
ctor only used during construction; not allowed while operation is in progress
```

src/transaction/atomic_replication_helper.hpp
```diff
@@ -67,7 +69,7 @@ namespace cublog
 	  atomic_replication_unit (log_lsa lsa, VPID vpid, LOG_RCVINDEX rcvindex);
 
 	  atomic_replication_unit (const atomic_replication_unit &) = delete;
-	  atomic_replication_unit (atomic_replication_unit &&) = delete;
+	  atomic_replication_unit (atomic_replication_unit &&);
```
cristiarg: name parameter

src/transaction/atomic_replication_helper.hpp
```diff
@@ -80,12 +82,13 @@ namespace cublog
 	  int fix_page (THREAD_ENTRY *thread_p);
 	  void unfix_page (THREAD_ENTRY *thread_p);
 
-	  VPID m_vpid;
+	  const VPID m_vpid;
```
cristiarg: I do not see where this needs to be public.

src/transaction/atomic_replication_helper.hpp
```diff
@@ -96,6 +99,49 @@ namespace cublog
...
+
+  template <typename T>
+  int atomic_replication_helper::add_atomic_replication_unit (THREAD_ENTRY *thread_p, TRANID tranid, log_lsa record_lsa,
+      LOG_RCVINDEX rcvindex, VPID vpid, log_rv_redo_context &redo_context, const log_rv_redo_rec_info<T> &record_info)
+  {
+#if !defined (NDEBUG)
+    if (!VPID_ISNULL (&vpid) && !check_for_page_validity (vpid, tranid))
+      {
+	// the page is no longer relevant
+	assert (false);
```
cristiarg: either return error code and process it; or don't return

src/transaction/log_replication_atomic.cpp
```diff
@@ -0,0 +1,173 @@
...
+    log_replication_update_header_mvcc_vacuum_info (mvccid, prev_rec_lsa, rec_lsa, m_bookkeep_mvcc_vacuum_info);
+
+    // Redo b-tree stats differs from what the recovery usually does. Get the recovery index before deciding how to
+    // proceed.
+    const LOG_RCVINDEX rcvindex = log_rv_get_log_rec_data (record_info.m_logrec).rcvindex;
+    const VPID log_vpid = log_rv_get_log_rec_vpid<T> (record_info.m_logrec);
+    if (m_atomic_helper.is_part_of_atomic_replication (trid))
+      {
+	m_atomic_helper.add_atomic_replication_unit (&thread_entry, trid, rec_lsa, rcvindex, log_vpid, m_redo_context,
+	    record_info);
```
cristiarg: if log record is added to an atomic replication, nothing further should be done

src/transaction/log_replication_atomic.cpp
```diff
@@ -0,0 +1,173 @@
...
+      }
+
+    if (rcvindex == RVBT_LOG_GLOBAL_UNIQUE_STATS_COMMIT)
+      {
+	read_and_redo_btree_stats (thread_entry, record_info);
```
cristiarg, line 205: what if this log record - which has a particular handling - is also part of an atomic replication sequence? will that be handled correctly (if at all)?

src/transaction/log_replication_atomic.cpp
```diff
@@ -0,0 +1,173 @@
...
+	// to accurately track progress and avoid clients to wait for too long, notify each change
+	m_redo_lsa_condvar.notify_all ();
+
+	m_perfmon_redo_sync.track_and_start ();
+      }
+  }
+
+  template <typename T>
+  void
+  atomic_replicator::read_and_redo_record_with_atomic_consideration (cubthread::entry &thread_entry, LOG_RECTYPE rectype,
```
cristiarg: what if the same old name `read_and_redo_record` is left in place?

src/transaction/log_replication_atomic.hpp
cristiarg: also, add a proper dtor implementation with just the comment:
```
an active transaction server is a "transient" server instance; it does not store any data
and, thus, does not need to be left in consistent state; thus, no check as to the consistent
termination state for atomic replication is needed
```

src/transaction/log_replication_atomic.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+#ifndef _ATOMIC_REPLICATOR_HPP_
+#define _ATOMIC_REPLICATOR_HPP_
+
+#include "atomic_replication_helper.hpp"
+#include "log_replication.hpp"
+
+namespace cublog
+{
+  class atomic_replicator : replicator
```
cristiarg: extensive comment description about what the class does

src/transaction/log_replication_atomic.cpp
```diff
@@ -0,0 +1,173 @@
...
+	    break;
+	  case LOG_ABORT:
+	    calculate_replication_delay_or_dispatch_async<log_rec_donetime> (
+		    thread_entry, m_redo_lsa);
+	    break;
+	  case LOG_DUMMY_HA_SERVER_STATE:
+	    calculate_replication_delay_or_dispatch_async<log_rec_ha_server_state> (
+		    thread_entry, m_redo_lsa);
+	    break;
+	  case LOG_TRANTABLE_SNAPSHOT:
```
cristiarg: `LOG_TRANTABLE_SNAPSHOT` does not need to be handled on PTS; replace handling with a comment; thus `m_most_recent_trantable_snapshot_lsa` can be private in base class

cristiarg (commented):
Partial review.
Also, build fails.

src/transaction/log_replication_atomic.cpp
```diff
@@ -0,0 +1,173 @@
...
+ */
+
+#include "atomic_replicator.hpp"
+
+#include "log_recovery_redo_parallel.hpp"
+
+namespace cublog
+{
+
+  void atomic_replicator::redo_upto (cubthread::entry &thread_entry, const log_lsa &end_redo_lsa)
```
razvanradulescu: This function looks very similar to one from parent class: replicator::redo_upto.
Perhaps splitting redo_upto in parent class and then overring only parts of inner switch-case may avoid code duplication.

src/transaction/log_replication_atomic.cpp
razvanradulescu: This function is also very similar to the one from parent class: read_and_redo_record.
Did you considered an alternative design: instead of atomic_replicator "is a" new kind of "replicator", just leave replicator and in it "have a" special kind of behavior by templating it or provide the "redo" function as parameter ?

src/transaction/atomic_replication_helper.hpp
```diff
@@ -96,6 +99,49 @@ namespace cublog
...
+#endif
+
+    m_atomic_sequences_map[tranid].emplace_back (atomic_replication_unit (record_lsa, vpid, rcvindex));
+    int error_code = m_atomic_sequences_map[tranid].back ().fix_page (thread_p);
+    if (error_code != NO_ERROR)
+      {
+	return error_code;
+      }
+
+    m_atomic_sequences_map[tranid].back ().apply_log_redo (thread_p, redo_context, record_info);
```
razvanradulescu: This function is called "add_atomic_replication_unit", but here we also "apply" the replication unit.
My understanding is that the atomic sequence is composed of atomic unit (a kind of wrapper on top of log record), the entire atomic sequence needs to be applied at once. For this, I thought we need to fix all pages first, then apply the records, then unfix all pages.

This code seems to apply the logs as the pages are fixed; this comes with a possibility that a page fix (with ordered fix) may trigger an unfix in a previously fixed page from the sequence, allowing other transaction to change it.

src/transaction/log_replication_atomic.cpp
cristiarg: We considered this approach and, although it could be an option, we favored inheritance. More differences will occur which might make the templated approach cluttered.

src/transaction/atomic_replication_helper.hpp
```diff
@@ -96,6 +99,49 @@ namespace cublog
...
+      log_rv_redo_context &redo_context, const log_rv_redo_rec_info<T> &record_info)
+  {
+    LOG_RCV rcv;
+    if (m_page_ptr != nullptr)
+      {
+	rcv.pgptr = m_page_ptr;
+      }
+    else
+      {
+	rcv.pgptr = m_watcher.pgptr;
```
razvanradulescu, line 157: Here an assert would help.
assert (m_watcher.page_was_unfixed == 0)

src/transaction/log_replication_atomic.cpp
adrianbrab: for the atomic replicator there is one more arg required `TRANID trid`

src/transaction/log_replication_atomic.cpp
adrianbrab: I am confused about the comment on the replicator:
```
	    // save the LSA of the last transaction table snapshot that can be found in the log
	    // only needed on the passive transaction server
```

cristiarg: Yes, the comment is somewhat contrived. "Needed" refers to the final usage which is indeed on the PTS.
It should read:
```
// save the LSA of the last transaction table snapshot that can be found in the log
// transaction table snapshots are added to the transactional log by the active transaction server
// the LSA of the most recent is saved/bookkept by the page server (this section)
// and, finally, this LSA is retrieved and used by a booting up passive transaction server
```

src/transaction/log_replication_atomic.hpp
```diff
@@ -0,0 +1,58 @@
...
+{
+  /*
+   * Replicator derived class responsible for handling the log redo only on Passive Transaction Servers
+   * In addition to the standard replicator, the atomic variant also adds support for atomic log replication
+   * with the help of the atomic_replication_helper class.
+   */
+  class atomic_replicator : replicator
+  {
+    public:
+      atomic_replicator (const log_lsa &start_redo_lsa, PAGE_FETCH_MODE page_fetch_mode);
```
cristiarg: `PAGE_FETCH_MODE` is constant for this replicator flavor; internalize it as a constant member

src/transaction/atomic_replication_helper.cpp
cristiarg: assert should read `assert (that.m_page_ptr == nullptr)`
also, `this.m_page_ptr` should be initialized with `nullptr`
`m_watcher` should also be initialized in ctor; not assert for being clean

src/transaction/atomic_replication_helper.cpp
```diff
@@ -125,26 +123,20 @@ namespace cublog
...
-      {
-	rcv.pgptr = m_page_ptr;
-      }
-    else
-      {
-	rcv.pgptr = m_watcher.pgptr;
-      }
-    rcv.reference_lsa = m_record_lsa;
-    log_rv_redo_record_sync_apply (thread_p,redo_context, record_info, m_vpid, rcv);
+    PGBUF_CLEAR_WATCHER (&m_watcher);
```
cristiarg: IMHO
```suggestion
    assert (PGBUF_IS_CLEAN_WATCHER (&m_watcher));
```
although I am not so sure because a PTS can "go down" at any given time, therefore the dtor can occur at any given moment in the state of replication; better to add the assert here also with a comment that it might not hold; and we will see during testing

src/transaction/atomic_replication_helper.hpp
cristiarg: @razvanradulescu http://jira.cubrid.org/browse/LETS-460 has been added to adress this; thanks

src/transaction/log_replication_atomic.cpp
```diff
@@ -0,0 +1,207 @@
...
+    log_replication_update_header_mvcc_vacuum_info (mvccid, prev_rec_lsa, rec_lsa, m_bookkeep_mvcc_vacuum_info);
+
+    // Redo b-tree stats differs from what the recovery usually does. Get the recovery index before deciding how to
+    // proceed.
+    const LOG_RCVINDEX rcvindex = log_rv_get_log_rec_data (record_info.m_logrec).rcvindex;
+    const VPID log_vpid = log_rv_get_log_rec_vpid<T> (record_info.m_logrec);
+
+    if (rcvindex == RVBT_LOG_GLOBAL_UNIQUE_STATS_COMMIT)
+      {
+	read_and_redo_btree_stats (thread_entry, record_info);
```
cristiarg, line 205: add an assert that this log record is not part of an atomic replication;
if it does not hold, we'll re-think

src/transaction/log_replication_atomic.cpp
```diff
@@ -0,0 +1,207 @@
...
+      }
+
+    if (m_atomic_helper.is_part_of_atomic_replication (trid))
+      {
+	m_atomic_helper.add_atomic_replication_unit (&thread_entry, trid, rec_lsa, rcvindex, log_vpid, m_redo_context,
+	    record_info);
+      }
+    else if (rcvindex != RVBT_LOG_GLOBAL_UNIQUE_STATS_COMMIT)
```
cristiarg: `if` is not needed

src/transaction/log_replication_atomic.hpp
```diff
@@ -0,0 +1,58 @@
...
+      atomic_replicator (atomic_replicator &&) = delete;
+
+      ~atomic_replicator ();
+
+      atomic_replicator &operator= (const atomic_replicator &) = delete;
+      atomic_replicator &operator= (atomic_replicator &&) = delete;
+
+    private:
+      void redo_upto (cubthread::entry &thread_entry, const log_lsa &end_redo_lsa) override;
+      // TODO: change name
```
cristiarg: Try with the same name as in the base class `read_and_redo_record`.

src/transaction/log_replication_atomic.hpp
```diff
@@ -0,0 +1,58 @@
...
+   */
+  class atomic_replicator : public replicator
+  {
+    public:
+      atomic_replicator (const log_lsa &start_redo_lsa, PAGE_FETCH_MODE page_fetch_mode);
+
+      atomic_replicator (const atomic_replicator &) = delete;
+      atomic_replicator (atomic_replicator &&) = delete;
+
+      ~atomic_replicator ();
```
cristiarg: `override`

src/transaction/atomic_replication_helper.hpp
```diff
@@ -96,6 +101,50 @@ namespace cublog
...
+      {
+	return error_code;
+      }
+
+    m_atomic_sequences_map[tranid].back ().apply_log_redo (thread_p, redo_context, record_info);
+    return NO_ERROR;
+  }
+
+  template <typename T>
+  void atomic_replication_helper::atomic_replication_unit::apply_log_redo (THREAD_ENTRY *thread_p,
```
cristiarg: The implementation is not complete. It lacks the part where the just fixed page's LSA is checked against the log record's `start_lsa`:
- if log record's LSA is samller or equal to the page's LSA, applying redo is not needed
- check implementation and use of `log_rv_fix_page_and_check_redo_is_needed`

To fix:
- change function `log_rv_fix_page_and_check_redo_is_needed` to only take `pgptr` instead of `log_rcv &rcv` argument (for simplicity)
- call the function in `apply_log_redo` and only apply redo if function returns true
- if function returns false, do not apply redo

Please implement this remark as a separate PR and then, after approval&merge, integrate into this one.

src/transaction/atomic_replication_helper.hpp
```diff
@@ -79,13 +82,15 @@ namespace cublog
 			       const log_rv_redo_rec_info<T> &record_info);
 	  int fix_page (THREAD_ENTRY *thread_p);
 	  void unfix_page (THREAD_ENTRY *thread_p);
+	  LOG_LSA get_lsa () const;
 
-	  VPID m_vpid;
 	private:
-	  log_lsa m_record_lsa;
+	  const VPID m_vpid;
```
cristiarg: I think this implementation is nor generic enough nor, as a consequence, correct.
It does not cover the situation when, during an atomic replication sequence, there is more than one log record that modifies the same page.
The code needs to be refactored:
- a structure needs to keep a shared pool of `pgptr`'s
- an atomic replication unit must only keep a "weak reference" to a page ptr (eg, more than one atomic replication unit can reference the same `pgptr` kept in this separate structure
- when a new replication unit is about to be added, it will check, based on `vpid` whether the page has already been fixed and, if yes, take the `pgptr` from the shared structure
- if not, will fix the `pgptr` and add it to the shared structure/collection for use by other atomic replication units
I will add a separate issue for this to be implemented separately and then integrated in this PR.

src/transaction/atomic_replication_helper.cpp
cristiarg: @adrianbrab 
This comment has not been fully addressed.

src/transaction/log_replication_atomic.cpp
cristiarg: This comment has not been addressed.

src/transaction/log_replication_atomic.cpp
cristiarg: agreed, we'll think of a better name

src/transaction/log_replication_atomic.cpp
cristiarg: Please change the comment in `log_replication::redo_upto`.

cristiarg (commented):
Another partial review. There are some comments which should be taken as separate prerequisite PR's to this one.

src/transaction/log_replication_atomic.cpp
adrianbrab: Can we just overload the original and use the same name?

src/transaction/log_replication_atomic.cpp
```diff
@@ -142,15 +142,10 @@ namespace cublog
...
-		// if the atomic replication sequence start lsa is higher or equal to the sysop
-		// end parent lsa, then the atomic sequence can be ended (commited & released)
-		if (m_atomic_helper.can_end_atomic_sequence (header.trid, log_rec.lastparent_lsa))
-		  {
-		    m_atomic_helper.unfix_atomic_replication_sequence (&thread_entry, header.trid);
-		  }
-	    //  }
+	    if (m_atomic_helper.can_end_atomic_sequence (header.trid, log_rec.lastparent_lsa))
+	      {
+		m_atomic_helper.unfix_atomic_replication_sequence (&thread_entry, header.trid);
```
cristiarg, line 149: The comment should remain:
```
// if the atomic replication sequence start lsa is higher or equal to the syso
// end parent lsa, then the atomic sequence can be ended (commited & released)
```

cristiarg (approved):

PR #3572, opened 2022-05-02
[CBRD-23842] Fix the bug when excute the trigger statement in CDC

joohok:

Purpose

There is the bug when creating a trigger that has no target class. 
For example, `create trigger ... execute print` statement has no target class, but existing code try to access the target class.
So, the logic for checking if the target class exists is added

### DIFF
--- a/src/parser/parse_tree.h
+++ b/src/parser/parse_tree.h
@@ -3870,8 +3870,7 @@ enum cdc_ddl_type
   CDC_ALTER,
   CDC_DROP,
   CDC_RENAME,
-  CDC_TRUNCATE,
-  CDC_TRUNCATE_CASCADE
+  CDC_TRUNCATE
 };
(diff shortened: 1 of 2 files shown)


eido5 (approved):

src/parser/parse_tree.h
```diff
@@ -3870,8 +3870,7 @@ enum cdc_ddl_type
   CDC_ALTER,
   CDC_DROP,
   CDC_RENAME,
-  CDC_TRUNCATE,
-  CDC_TRUNCATE_CASCADE
```
wrlawodms: Why is this removed?

joohok: It is unrelated change. 
I removed it because `TRUNCATE_CASCADE` is obsolete.
I will revert it, because it will be treated In other issue that add some more ddl types.

wrlawodms (approved):

hornetmj (approved):

PR #3576, opened 2022-05-03
[CBRD-24326] Support user related ddl statements in CDC

joohok:
http://jira.cubrid.org/browse/CBRD-24326

Purpose

CDC was designed to support only DML and DDL that change data.

However, as the user schema function is added, 
statements to CREATE/ALTER/DROP users must be copied to the target database.

Implementation

When make a supplemental log for DDL, the ddl target object types for user related statements are added
* object type
   * USER

### DIFF
--- a/src/parser/parse_tree.h
+++ b/src/parser/parse_tree.h
@@ -3870,9 +3870,7 @@ enum cdc_ddl_type
   CDC_ALTER,
   CDC_DROP,
   CDC_RENAME,
-  CDC_TRUNCATE,
-  CDC_GRANT,
-  CDC_REVOKE
+  CDC_TRUNCATE
 };
 typedef enum cdc_ddl_type CDC_DDL_TYPE;
 
--- a/src/parser/parse_tree.h
+++ b/src/parser/parse_tree.h
@@ -3871,7 +3871,8 @@ enum cdc_ddl_type
   CDC_DROP,
   CDC_RENAME,
   CDC_TRUNCATE,
-  CDC_TRUNCATE_CASCADE
+  CDC_GRANT,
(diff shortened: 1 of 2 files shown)


wrlawodms:
1. To deal with the issue you described, I think that CREATE/DROP/ALTER USER seem enough. GRANT is related to the privilege of tables which has existed since before the user-schema feature. Do you wanna handle the privilege issue here?
2. It seems that `ALTER ... OWNER TO` should be also handled because it changes the owner of a table and in turn the full name of the table. Or is it dealt with in another path?

joohok:
> * To deal with the issue you described, I think that CREATE/DROP/ALTER USER seem enough. GRANT is related to the privilege of tables which has existed since before the user-schema feature. Do you wanna handle the privilege issue here?

Yes, I agree with you. There is no need to support GRANT or REVOKE. Thanks.  

> * It seems that `ALTER ... OWNER TO` should be also handled because it changes the owner of a table and in turn the full name of the table. Or is it dealt with in another path?

The DDLs for changing the owner are already handled by the existing implementation. 
For example, 
`ALTER TABLE .. OWNER TO` -> DDL type : CDC_ALTER / Object type : CDC_TABLE
`ALTER TRIGGER .. OWNER TO` -> DDL type : CDC_ALTER / Object type : CDC_TRIGGER
`ALTER VIEW ... OWNER TO` -> DDL type : CDC_ALTER / Object type : CDC_VIEW
...

eido5 (approved):

hornetmj (approved):
Please apply the feedback from @wrlawodms.

wrlawodms (approved):

PR #3578, opened 2022-05-04
[CBRD-24327] When checking a synonym before checking a class in the semantic check, if the synonym does not exist, the error is cleared.

youngjinj:
http://jira.cubrid.org/browse/CBRD-24327

### When checking a synonym before checking a class in the semantic check, if the synonym does not exist, the error is cleared.
- The db_find_synonym () and sm_find_synonym () functions set an ER_SYNONYM_NOT_EXIST error when a synonym does not exist.

### DIFF
--- a/src/parser/semantic_check.c
+++ b/src/parser/semantic_check.c
@@ -4740,6 +4740,7 @@ pt_check_alter (PARSER_CONTEXT * parser, PT_NODE * alter)
       for_update = true;
     }
 
+  /* We cannot change the schema of a class by using synonym names. */
   if (db_find_synonym (cls_nam))
     {
       PT_ERRORmf (parser, alter->info.alter.entity_name, MSGCAT_SET_PARSER_SEMANTIC,
@@ -4749,7 +4750,7 @@ pt_check_alter (PARSER_CONTEXT * parser, PT_NODE * alter)
   else
     {
       /* db_find_synonym () == NULL */
-      if (er_errid () == ER_OBJ_OBJECT_NOT_FOUND)
+      if (er_errid () == ER_SYNONYM_NOT_EXIST)
 	{
 	  er_clear ();
 	}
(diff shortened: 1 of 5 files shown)


src/query/execute_schema.c
```diff
@@ -2623,11 +2623,20 @@ do_rename (PARSER_CONTEXT * parser, PT_NODE * statement)
...
 	  PT_ERRORmf (parser, statement, MSGCAT_SET_PARSER_SEMANTIC, MSGCAT_SEMANTIC_CLASS_DOES_NOT_EXIST, old_name);
 	  goto error_exit;
 	}
+      else
+	{
+	  /* db_find_synonym () == NULL */
+	  if (er_errid () == ER_SYNONYM_NOT_EXIST)
+	    {
+	      er_clear ();
+	    }
```
ctshim: db_find_synonym() returned NULL.
If it is due to an error other than ER_SYNONYM_NOT_EXIST, shouldn't it be handled as an error without ignoring it?

 ```
     if (er_errid () == ER_SYNONYM_NOT_EXIST)
	    {
	      er_clear ();
	    }
     else
	    {
	      // PT_ERRORmf( ... )
	      goto error_exit;
	    }    
```

File: src/parser/semantic_check.c [hunk 2]
```diff
@@ -9486,11 +9541,21 @@ pt_check_truncate (PARSER_CONTEXT * parser, PT_NODE * node)
       if ((name = temp->info.spec.entity_name) != NULL && name->node_type == PT_NAME
 	  && (cls_nam = name->info.name.original) != NULL)
 	{
+	  /* We cannot change the schema of a class by using synonym names. */
```
ctshim, line 9676: Can't truncate with synonym names?

ctshim: I got an explanation from the author. The question has been resolved.

hyunikn (approved):

hgryoo (approved):

src/query/execute_schema.c
youngjinj: Added error handling for cases other than ER_SYNONYM_NOT_EXIST errors.

beyondykk9 (approved):

ctshim (approved):

shparkcubrid (approved):

PR #3581, opened 2022-05-09
[CBRD-24333] Fix the core due to accessing invalid LSA in flashback

joohok:
http://jira.cubrid.org/browse/CBRD-24333

Purpose

When it try to find a LSA with a time, 
it finds a log record that have a time value which is near the input time, and returns the LSA of the next log record. 

But if LSA of the next log record is equal to log_Gl.hdr.append_lsa, which means next log record has not been appended, 
flashback or CDC can access the invalid log record because next log record is not always added at log_Gl.hdr.append_lsa. 

Implementation

* Returns LSA of current log record instead of LSA of next log record

### DIFF
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -14366,8 +14366,9 @@ cdc_get_lsa_with_start_point (THREAD_ENTRY * thread_p, time_t * time, LOG_LSA * 
 	      *time = donetime->at_time;
 
 	      /* It returns LSA of next log record because LSA of current log record is useless in CDC or flashback.
-	       * But if LSA of next log record (forw_lsa) is returned before next log record is appended,
-	       * CDC or flashback can access the wrong log LSA.
+	       * But if LSA of next log record (forw_lsa) is returned
+	       * before next log record is appended (forw_lsa == log_Gl.hdr.append_lsa),
+	       * CDC or flashback can access the invalid log record.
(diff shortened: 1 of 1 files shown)


hornetmj (approved):

eido5 (approved):

src/transaction/log_manager.c
```diff
@@ -14350,6 +14351,8 @@ cdc_get_lsa_with_start_point (THREAD_ENTRY * thread_p, time_t * time, LOG_LSA *
   while (!LSA_ISNULL (&process_lsa))
     {
       log_rec_header = LOG_GET_LOG_RECORD_HEADER (log_page_p, &process_lsa);
+
+      LSA_COPY (&cur_log_rec_lsa, &process_lsa);
```
wrlawodms: 1. `cur_lsa` or `current_lsa` looks better because so do others and it's simple and enough.
2. Why don't you use `process_lsa`, but use a new variable?

joohok: 1. OK, it seems good
2. When `process_lsa` is need to be returned, process_lsa does not point to the starting point of the log record header. This is because process_lsa moves forward as it reads the log record header and reads the data part of the log. (`LOG_READ_ADD_ALIGN`)

wrlawodms: oh, you're right. I missed that. Thank you.

wrlawodms (approved):

PR #3583, opened 2022-05-09
[CBRD-24334] Add NULL string exception when processing a column data

joohok:
http://jira.cubrid.org/browse/CBRD-24334

Purpose

if user executes SQL with NULL data, core has occurred when flashback tries to make SQL statement. 

NULL data is always transferred as a string type, 
and flashback utility checks whether the data is NULL to make below statements 
* `insert into tbl values (null);`
* `delete from tbl where [col1] is null limit 1;`
* `update tbl set [col1] = null where [col1] = 10 limit 1;`
* `update tbl set [col1] = 10 where [col1] is null limit 1;`

### DIFF
--- a/src/transaction/flashback_cl.c
+++ b/src/transaction/flashback_cl.c
@@ -335,7 +335,7 @@ typedef enum
 } FLASHBACK_PACK_FUNC_TYPE;
 
 static int
-flashback_process_column_data (char **data, char **sql, int *max_sql_size, DB_TYPE type)
+flashback_process_column_data (char **data, char **sql, int *max_sql_size, DB_TYPE type, bool is_condition_column)
 {
   char *ptr = *data;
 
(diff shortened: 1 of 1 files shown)


hornetmj (approved):

src/transaction/flashback_cl.c
```diff
@@ -414,7 +414,33 @@ flashback_process_column_data (char **data, char **sql, int *max_sql_size, DB_TY
     case PACK_STRING:
       ptr = or_unpack_string_nocopy (ptr, &s_data);
 
-      if (IS_STRING_TYPE (type))
+      if (s_data == NULL)
+	{
+	  if (is_condition_column == true)
```
wrlawodms, line 419: How about just checking `sql - 2 == '='` without the `is_condition_column` parameter? it seems to me that the variable doesn't provide any flexibility because if the query format is changed, we should change how to deal with it too.

wrlawodms: or feel free to tell me the advantage of using the parameter that I missed. :)

eido5 (approved):

src/transaction/flashback_cl.c
joohok: it seems simpler to check if it is a condition column like your suggestion, instead of using parameter. 
it is easier to check `sql -2 == '='`, but as you mentioned, it could provide no flexibility and readability. 
Maybe, I think it is better to check if there is `where` in the sql statement.

wrlawodms (approved):

PR #3584, opened 2022-05-09
[CBRD-23845] In the error message when the CREATE syntax is incorrect, "CLASS or TABLE" is recommended before "SYNONYM".

youngjinj:
http://jira.cubrid.org/browse/CBRD-23845

In the error message when the CREATE syntax is incorrect, "CLASS or TABLE" is recommended before "SYNONYM".

### DIFF
--- a/src/parser/csql_grammar.y
+++ b/src/parser/csql_grammar.y
@@ -971,13 +971,8 @@ int g_original_buffer_len;
 %type <node> dblink_expr
 %type <node> dblink_conn
 %type <node> dblink_conn_str
-%type <c2> dblink_identifier_col_attrs  
 %type <node> dblink_column_definition_list
 %type <node> dblink_column_definition
-%type <c10> connect_info
-%type <c2>  connect_item
-%type <c10> alter_server_list
-%type <c2>  alter_server_item
 /*}}}*/
 
 /* define rule type (cptr) */
@@ -992,6 +987,8 @@ int g_original_buffer_len;
 /*{{{*/
 %type <c10> opt_serial_option_list
 %type <c10> serial_option_list
+%type <c10> connect_info
+%type <c10> alter_server_list
 
 %type <c4> isolation_level_spec
 %type <c4> opt_constraint_attr_list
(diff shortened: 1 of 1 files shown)


src/parser/csql_grammar.y
```diff
@@ -24133,25 +24135,85 @@ dblink_column_definition
...
       DBG_PRINT}}
     ;
 
+/*
+ * To prevent the change of the error message below,
+ * the rules of opt_or_replace and opt_access_modifier are combined into opt_create_synonym and used.
+ * 
+ * Query: CREATE PARTITION TABLE;
+ * - AS-IS: Syntax error: unexpected 'partition', expecting CLASS or TABLE
+ * - TO-BE: Syntax error: unexpected 'partition', expecting SYNONYM
```
shparkcubrid: Did you change it to show 'CLASS or TABLE'?
The contents of the comments and Jira's comments are different.
```
AS-IS
csql> create partition table;
 *** Token) [create]
 *** Rule) stmt::= :
 *** Token) [ ]
 *** Token) [partition]
 *** Rule) create_stmt::= : CREATE
 *** Rule) opt_hint_list::= :

In line 1, column 8,

ERROR: In line 1, column 8 before ' table; '
Syntax error: unexpected 'partition', expecting CLASS or TABLE

TO-BE (NOT WANT)
csql> create partition table;
 *** Token) [create]
 *** Rule) stmt::= :
 *** Token) [ ]
 *** Token) [partition]
 *** Rule) opt_or_replace::= :
 *** Rule) opt_access_modifier::= :

In line 1, column 8,

ERROR: In line 1, column 8 before ' table; '
Syntax error: unexpected 'partition', expecting SYNONYM
```

ctshim: Isn't AS-IS and TO-BE changed?

ctshim (approved):

src/parser/csql_grammar.y
youngjinj: This is a change to prevent it from being changed to TO-BE.
```
TO-BE (NOT WANT)
```
There seems to be some confusion in the comments, so I'll replace AS-IS and TO-BE.

youngjinj: There seems to be some confusion in the comments, so I'll replace AS-IS and TO-BE.

hyunikn (approved):

shparkcubrid (approved):

beyondykk9 (approved):

hgryoo (approved):

PR #3586, opened 2022-05-09
[CBRD-24336] When the target of the synonym is changed, the related plan must be deleted.

youngjinj:
http://jira.cubrid.org/browse/CBRD-24336

### When the target of the synonym is changed, the related plan must be deleted.
- Even if the target of a synonym changes, the OID of that target remains in the xasl cache. If a synonym is used, use the OID remaining in the xasl cache rather than using the changed target. Since it is not known whether a synonym is being used in the xasl cache, all xasl using the same OID must be deleted from the cache.


ctshim (approved):

beyondykk9 (approved):

src/communication/network_interface_cl.c
```diff
@@ -7452,6 +7452,51 @@ serial_decache (OID * oid)
...
+  reply = OR_ALIGNED_BUF_START (a_reply);
+
+  or_pack_oid (request, oid);
+
+  req_error =
+    net_client_request (NET_SERVER_SYNONYM_REMOVE_XASL_BY_OID, request, OR_ALIGNED_BUF_SIZE (a_request), reply,
+			OR_ALIGNED_BUF_SIZE (a_reply), NULL, 0, NULL, 0);
+  if (!req_error)
+    {
+      or_unpack_int (reply, &status);
```
hyunikn, line 7485: This line seems redundant because status is not used below.

hyunikn (approved):

shparkcubrid (approved):

src/communication/network_interface_cl.c
youngjinj: The serial_decache function is also related, so it seems to be solved together.
If the execution method of synonyms is changed, it can be removed, so I will not change it.

PR #3589, opened 2022-05-10
[LETS-423] Check before appling redo log 

adrianbrab:

Before appling a redo log it is important to check against the `start_lsa` in order to see if operation is still needed.

### DIFF
--- a/src/transaction/atomic_replication_helper.cpp
+++ b/src/transaction/atomic_replication_helper.cpp
@@ -145,15 +145,12 @@ namespace cublog
 	rcv.pgptr = m_watcher.pgptr;
       }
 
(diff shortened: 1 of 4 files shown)


src/transaction/atomic_replication_helper.cpp
```diff
@@ -143,6 +144,14 @@ namespace cublog
...
       }
+
+    if (!log_rv_fix_page_and_check_redo_is_needed (thread_p, m_vpid, rcv.pgptr, record_info.m_start_lsa,
+	redo_context.m_end_redo_lsa, redo_context.m_page_fetch_mode))
+      {
+	/* nothing else needs to be done, see explanation in function */
+	assert (rcv.pgptr == nullptr);
+	return;
+      }
     rcv.reference_lsa = m_record_lsa;
```
cristiarg: Why is this here? I think it is not needed.

cristiarg (approved):
Approved with one required change.

PR #3590, opened 2022-05-10
[CBRD-24337] Users cannot create tables and views with the system class name.

youngjinj:
http://jira.cubrid.org/browse/CBRD-24337

Users cannot create tables and views with the system class name.
Users cannot rename the system class name.


ctshim (approved):

hyunikn (approved):

beyondykk9 (approved):

shparkcubrid (approved):

hgryoo (approved):

shparkcubrid (approved):

beyondykk9 (approved):

src/parser/parser_support.c
```diff
@@ -10196,70 +10196,106 @@ pt_set_user_specified_name (PARSER_CONTEXT * parser, PT_NODE * node, void *arg,
...
+	PT_SYNONYM_OWNER_NAME (node) = pt_name (parser, pt_get_qualifier_name (parser, PT_SYNONYM_NAME (node)));
+	if (pt_get_qualifier_name (parser, PT_SYNONYM_TARGET_NAME (node)) == NULL
+	    && sm_check_system_class_by_name (PT_NAME_ORIGINAL (PT_SYNONYM_TARGET_NAME (node))) == true)
+	  {
+	    PT_SYNONYM_TARGET_OWNER_NAME (node) = pt_name (parser, "dba");
+	  }
+	else
+	  {
+	    PT_SYNONYM_TARGET_OWNER_NAME (node) =
+	      pt_name (parser, pt_get_qualifier_name (parser, PT_SYNONYM_TARGET_NAME (node)));
```
hyunikn, line 10260: Is it possible that pt_get_qualifier_name (parser, PT_SYNONYM_TARGET_NAME (node)) == NULL at this line? If so, isn't it a problem?

src/parser/parser_support.c
```diff
@@ -10196,70 +10196,106 @@ pt_set_user_specified_name (PARSER_CONTEXT * parser, PT_NODE * node, void *arg,
...
 	  else
 	    {
-	      PT_SYNONYM_TARGET_OWNER_NAME (node) =
-		pt_name (parser, pt_get_qualifier_name (parser, PT_SYNONYM_TARGET_NAME (node)));
+	      assert (PT_IS_NAME_NODE (node->info.expr.arg1));
+
+	      PT_NODE *name = node->info.expr.arg1;
+
+	      original_name = name->info.name.original;
+	      resolved_name = name->info.name.resolved;
```
hyunikn: [Q] is resolved_name always NULL at this line?

hyunikn (approved):

src/parser/parser_support.c
youngjinj: Don't set name->info.name.resolved before. So it's probably NULL.
It seems to be left over from the old code.

I will check more and remove if unnecessary.

src/parser/parser_support.c
youngjinj: This is code that cannot be NULL.
I'm going to add an assert to check for NULL.

PR #3592, opened 2022-05-11
[CBRD-24339] If the last row is fetched, is_fetch_completed should be 1.

airnet73:
http://jira.cubrid.org/browse/CBRD-24339

Purpose
When the size of a row is 16kb or more, is_fetch_completed is transmitted as 0 when data is transmitted to the client after fetching the last row.
When is_fetch_completed is 0, the client determines that there is more data to request, and "dblink - Invalid cursor position" occurs when requesting data.
If the last data is fetched, is_fetch_completed must be sent to 1 so that the client does not request any more data.

Implementation
N/A

Remarks
N/A


kisoo-han (approved):

kisoo-han:
Fix jira refef: CBRD-XXXX to CBRD-24339

PR #3600, opened 2022-05-13
[CBRD-24240] Removed the not null constraint of _db_class and db_serial for compatibility.

youngjinj:
http://jira.cubrid.org/browse/CBRD-24240

Removed the not null constraint of _db_class and db_serial for compatibility.


beyondykk9 (approved):

youngjinj:
http://jira.cubrid.org/browse/CBRD-24240

Removed the not null constraint of _db_class and db_serial for compatibility.

PR #3601, opened 2022-05-13
[CBRD-24349] index scan enhancement for correlated subquery.

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-24349

I removed routines that could reduce the load regardless of the severity of the load. It is as below.

1. remove unnessary prm_get_XXX_value ()
2. remove pgbuf_check_page_ptype () for debug (related to : CUBRIDSUS-12543)
3. remove db_page_size ()
4. remove unnecessary btree_compare
5. move malloc for key_val_range from scan_open_scan() to xasl_generation
6. remove getting type in btid_int
7. remove malloc in qfile_open_list()
8. remove malloc for list_id of cov index

About 15% performance improvement by deleting unnecessary routines


src/storage/disk_manager.c
```diff
@@ -3073,7 +3073,7 @@ disk_verify_volume_header (THREAD_ENTRY * thread_p, PAGE_PTR pgptr)
   DISK_VOLUME_HEADER *vhdr;
 
   assert (pgptr != NULL);
-  (void) pgbuf_check_page_ptype (thread_p, pgptr, PAGE_VOLHEADER);
+  /* (void) pgbuf_check_page_ptype (thread_p, pgptr, PAGE_VOLHEADER); */
```
hgryoo: This line is commented out instead of making it run only in debug mode like other code. Can I know the reason?

src/storage/btree.c
```diff
@@ -24114,32 +24161,13 @@ xbtree_find_unique (THREAD_ENTRY * thread_p, BTID * btid, SCAN_OPERATION_TYPE sc
...
        * Acquire snapshot in RR if not already acquired. This is needed since
        * the transaction need to know the actual visible objects - before
        * instance locking. In this way future commands of current transaction
        * may correctly detect visible objects.
        */
-      if (!need_skip_mvcc_snapshot)
-	{
-	  (void) logtb_get_mvcc_snapshot (thread_p);
-	}
+      (void) logtb_get_mvcc_snapshot (thread_p);
```
beyondykk9: why did you remove the need_skip_mvcc_snapshot?

src/parser/xasl_generation.c
```diff
@@ -12007,8 +12007,14 @@ pt_to_index_info (PARSER_CONTEXT * parser, DB_OBJECT * class_, PRED_EXPR * where
       if (indx_infop->coverage)
 	{
 	  qo_check_coll_optimization (index_entryp, &collation_opt);
-
 	  indx_infop->coverage = collation_opt.allow_index_opt;
+	  /* alloc list file id for covering index */
+	  regu_alloc (indx_infop->cov_list_id);
+	  if (indx_infop == NULL)
```
beyondykk9:
```suggestion
	  if (indx_infop->cov_list_id == NULL)
```

src/query/stream_to_xasl.c
```diff
@@ -4533,6 +4567,22 @@ stx_build_key_info (THREAD_ENTRY * thread_p, char *ptr, KEY_INFO * key_info)
...
+  ptr = or_unpack_int (ptr, &offset);
+  if (offset == 0 || key_info->key_cnt == 0)
+    {
+      key_info->key_vals = NULL;
+    }
+  else
+    {
+      key_info->key_vals =
+	stx_restore_key_val_array (thread_p, &xasl_unpack_info->packed_xasl[offset], key_info->key_cnt);
+      if (key_info->key_ranges == NULL)
```
beyondykk9:
```suggestion
      if (key_info->key_vals == NULL)
```

src/storage/disk_manager.c
shparkcubrid: I'll fix it. thanks.

src/parser/xasl_generation.c
shparkcubrid: I'll fix it. thanks.

src/query/stream_to_xasl.c
shparkcubrid: I'll fix it. thanks.

src/storage/btree.c
shparkcubrid: I'll fix it. thanks.

src/storage/btree_load.c
```diff
@@ -279,10 +279,11 @@ btree_get_node_header (THREAD_ENTRY * thread_p, PAGE_PTR page_ptr)
 
   assert (page_ptr != NULL);
 
+/*
 #if !defined(NDEBUG)
   (void) pgbuf_check_page_ptype (thread_p, page_ptr, PAGE_BTREE);
 #endif
-
+*/
```
shparkcubrid: need to fix

src/storage/file_manager.c
```diff
@@ -9293,7 +9293,7 @@ file_tempcache_put (THREAD_ENTRY * thread_p, FILE_TEMPCACHE_ENTRY * entry)
   /* lock temporary cache */
   file_tempcache_lock ();
 
-  if (file_Tempcache->ncached_not_numerable + file_Tempcache->ncached_numerable < file_Tempcache->ncached_max)
+  if (file_Tempcache->ncached_not_numerable + file_Tempcache->ncached_numerable <= file_Tempcache->ncached_max)
```
shparkcubrid: need to fix

src/storage/disk_manager.c
shparkcubrid: Well, it's been a while so I don't remember exactly.. I think I was trying to check the performance change in debug mode.

hyunikn (approved):

src/query/list_file.c
```diff
@@ -1166,14 +1166,22 @@ qfile_finalize (void)
...
   SORT_LIST *src_sort_list_p, *dest_sort_list_p;
   size_t type_list_size;
 
-  list_id_p = (QFILE_LIST_ID *) malloc (sizeof (QFILE_LIST_ID));
+  if (existing_list_id != NULL)
+    {
+      list_id_p = existing_list_id;
+    }
+  else
+    {
```
hgryoo, line 1181: `list_id_p` is allocated by `malloc()` now. The scope of this variable is in a thread task. (reference count of qlist_file will be managed by qfile_update_qlist_count () at line 1275 and checked at the end of the task). so I think we can use the private allocator here.

ctshim:
If pgbuf_check_page_ptype() is checked only in DEBUG mode throughout the source code, why don't you modify it in page_buffer.h as follows?

```
#ifdef NDEBUG
#  define pgbuf_check_page_ptype(thread, pg, type) 
#else
extern bool pgbuf_check_page_ptype (THREAD_ENTRY * thread_p, PAGE_PTR pgptr, PAGE_TYPE ptype);
#endif
```

ctshim (approved):

beyondykk9 (approved):

src/query/list_file.c
shparkcubrid: There are a lot of routines that call qfile_open_list() , so it was not easy to fix. I'll deal with this later in the CBRD-24397 issue.

shparkcubrid:
@ctshim 
About half of it was already written like I did. And I think the way I did it was clearer to prevent the function from being called.

hgryoo (approved):

src/query/query_executor.c
```diff
@@ -12962,31 +12965,19 @@ qexec_start_mainblock_iterations (THREAD_ENTRY * thread_p, xasl_node * xasl, xas
 	      }
 
 	    QFILE_SET_FLAG (ls_flag, QFILE_FLAG_ALL);
-	    t_list_id = qfile_open_list (thread_p, &type_list, NULL, xasl_state->query_id, ls_flag);
-	    if (t_list_id == NULL)
+	    xasl->list_id = qfile_open_list (thread_p, &type_list, NULL, xasl_state->query_id, ls_flag, xasl->list_id);
```
youngjinj, line 12967: t_list_id is deprecated. How about removing it?

youngjinj (approved):

src/query/query_executor.c
shparkcubrid: I fixed it.

PR #3604, opened 2022-05-14
[CBRD-24330] After user schema support, the table name used by the utility must include the owner name.

youngjinj:
http://jira.cubrid.org/browse/CBRD-24330

### After user schema support, the table name used by the utility must include the owner name.
- Before user schema support, the owner name was not used in the table name used by the utility. After user schema support, the same table name may exist depending on the owner, so the table name used by the utility must include the owner name.
- If the table name used by loaddb and unloaddb does not include the owner name, the name of the user who executed the utility is used as the owner name.


src/executables/checksumdb.c
```diff
@@ -2122,19 +2125,57 @@ checksumdb (UTIL_FUNCTION_ARG * arg)
...
+
+      /* Check the length when "_schema" is added. */
+      if (snprintf (NULL, 0, "%s%s", chksum_result_Table_name, CHKSUM_SCHEMA_TABLE_SUFFIX) >= SM_MAX_IDENTIFIER_LENGTH)
+	{
+	  PRINT_AND_LOG_ERR_MSG (msgcat_message
+				 (MSGCAT_CATALOG_UTILS, MSGCAT_UTIL_SET_GENERIC,
+				  MSGCAT_UTIL_GENERIC_CLASSNAME_EXCEED_MAX_LENGTH), SM_MAX_USER_LENGTH,
+				 SM_MAX_IDENTIFIER_LENGTH - SM_MAX_USER_LENGTH);
+	  util_log_write_errid (MSGCAT_UTIL_GENERIC_CLASSNAME_EXCEED_MAX_LENGTH);
+	  return ER_FAILED;
```
beyondykk9, line 2152: it's redundant.

src/executables/checksumdb.c
```diff
@@ -2122,19 +2125,57 @@ checksumdb (UTIL_FUNCTION_ARG * arg)
...
 
-  if (snprintf (chksum_schema_Table_name, SM_MAX_IDENTIFIER_LENGTH - 1, "%s%s", chksum_result_Table_name,
-		CHKSUM_SCHEMA_TABLE_SUFFIX) < 0)
-    {
-      assert (false);
-      goto error_exit;
+      if (snprintf (chksum_schema_Table_name, SM_MAX_IDENTIFIER_LENGTH - 1, "%s%s", chksum_result_Table_name,
+		    CHKSUM_SCHEMA_TABLE_SUFFIX) < 0)
+	{
+	  assert (false);
```
beyondykk9, line 2176: it's redundant.

shparkcubrid (approved):

src/executables/checksumdb.c
youngjinj: If the length of chksum_result_Table_name is 254, adding CHKSUM_SCHEMA_TABLE_SUFFIX(7) will exceed SM_MAX_IDENTIFIER_LENGTH(255), so it cannot be stored in chksum_schema_Table_name.

Even if the length of chksum_result_Table_name is less than SM_MAX_IDENTIFIER_LENGTH(255), if the length of class_name is 222, it exceeds 223 bytes when CHKSUM_SCHEMA_TABLE_SUFFIX(7) is added.

src/executables/checksumdb.c
youngjinj: When the if condition is true, the length of chksum_schema_Table_name should be checked, but in the case of else, it is not necessary.

hyunikn (approved):

beyondykk9 (approved):

PR #3605, opened 2022-05-16
[CBRD-24350] Fix the core dump due to allocating less memory than needed in CDC

joohok:
http://jira.cubrid.org/browse/CBRD-24350

Purpose

When allocating memory for log item, it calculate the length of the record by recdes->length.

If the record (string) is compressed, the exact length of the record cannot be known through `recdes->length`. So, I added a function to know the length of the attributes according to the data type.

Implementation

* cdc_get_attribute_size()
  * Get size of data with db_value
  * if the data is required to be converted to string type, then get the string size

### DIFF
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -12503,7 +12503,7 @@ cdc_get_attribute_size (DB_VALUE * value)
       {
 	/* size of the data converted into bit string include "X''"
 	 * e.g. 17 = B'10001' = X'11' */
-	size = ((db_get_string_length (value) + 3) / 4) + 4;
+	size = ((db_get_string_length (value) + 3) / 4) + 3;
 	break;
       }
     case DB_TYPE_CHAR:
@@ -12514,7 +12514,7 @@ cdc_get_attribute_size (DB_VALUE * value)
     case DB_TYPE_VARNCHAR:
       /* size of string "N''" is 4
        * e.g. N'string' */
-      size = db_get_string_size (value) + 4;
+      size = db_get_string_size (value) + 3;
       break;
     case DB_TYPE_TIME:
       /* precision in data types related to DATE/TIME means the size the string converted from the date/time data */
@@ -12575,14 +12575,7 @@ cdc_get_attribute_size (DB_VALUE * value)
 	DB_ELO *elo;
 	elo = db_get_elo (value);
 
-	if (elo != NULL)
-	  {
-	    size = (int) strlen (elo->locator);
-	  }
-	else
-	  {
-	    size = 0;
-	  }
+	size = elo == NULL ? 0 : (int) strlen (elo->locator);
       }
       break;
     case DB_TYPE_NULL:
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -12471,6 +12471,144 @@ cdc_check_if_schema_changed (RECDES * recdes, HEAP_CACHE_ATTRINFO * attr_info)
   return or_rep_id (recdes) != attr_info->last_classrepr->id;
 }
 
+static int
+cdc_get_attribute_size (DB_VALUE * value)
+{
+  assert (!DB_IS_NULL (value));
+
+  int size = 0;
+
+  switch (DB_VALUE_TYPE (value))
+    {
+    case DB_TYPE_INTEGER:
+      size = sizeof (DB_C_INT);
+      break;
+    case DB_TYPE_BIGINT:
+      size = sizeof (DB_C_BIGINT);
+      break;
+    case DB_TYPE_SHORT:
+      size = sizeof (DB_C_SHORT);
+      break;
+    case DB_TYPE_FLOAT:
+      size = sizeof (DB_C_FLOAT);
+      break;
+    case DB_TYPE_DOUBLE:
+      size = sizeof (DB_C_DOUBLE);
+      break;
+    case DB_TYPE_NUMERIC:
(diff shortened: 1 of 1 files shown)


src/transaction/log_manager.c
```diff
@@ -12471,6 +12471,144 @@ cdc_check_if_schema_changed (RECDES * recdes, HEAP_CACHE_ATTRINFO * attr_info)
   return or_rep_id (recdes) != attr_info->last_classrepr->id;
 }
 
+static int
+cdc_get_attribute_size (DB_VALUE * value)
+{
+  assert (!DB_IS_NULL (value));
+
+  int size = 0;
```
eido5, line 12479: Is the variable 'size' related with attribute size?
If then, What about replace `attr_size` instead of `size`?

(Not critical, and while reading through the code, I can guess 'size' is `attr_size`)

eido5 (approved):

src/transaction/log_manager.c
```diff
@@ -12471,6 +12471,144 @@ cdc_check_if_schema_changed (RECDES * recdes, HEAP_CACHE_ATTRINFO * attr_info)
...
+      break;
+    case DB_TYPE_NUMERIC:
+      size = DB_NUMERIC_BUF_SIZE;
+      break;
+    case DB_TYPE_BIT:
+    case DB_TYPE_VARBIT:
+      {
+	/* size of the data converted into bit string include "X''"
+	 * e.g. 17 = B'10001' = X'11' */
+	size = ((db_get_string_length (value) + 3) / 4) + 4;
```
wrlawodms: Could you tell me the meaning of `/4) + 4`? I think `(db_get_string_length (value) + 3)` means the string + `X''`, and?

src/transaction/log_manager.c
```diff
@@ -12471,6 +12471,144 @@ cdc_check_if_schema_changed (RECDES * recdes, HEAP_CACHE_ATTRINFO * attr_info)
...
+      }
+    case DB_TYPE_CHAR:
+    case DB_TYPE_VARCHAR:
+      size = db_get_string_size (value);
+      break;
+    case DB_TYPE_NCHAR:
+    case DB_TYPE_VARNCHAR:
+      /* size of string "N''" is 4
+       * e.g. N'string' */
+      size = db_get_string_size (value) + 4;
```
wrlawodms: Why is it not 3, but 4? Does it include `\0` or something?

src/transaction/log_manager.c
```diff
@@ -12471,6 +12471,144 @@ cdc_check_if_schema_changed (RECDES * recdes, HEAP_CACHE_ATTRINFO * attr_info)
...
+	elo = db_get_elo (value);
+
+	if (elo != NULL)
+	  {
+	    size = (int) strlen (elo->locator);
+	  }
+	else
+	  {
+	    size = 0;
+	  }
```
```suggestion
	size = elo == NULL ? 0 : (int) strlen (elo->locator);
```

src/transaction/log_manager.c
```diff
@@ -12471,6 +12471,144 @@ cdc_check_if_schema_changed (RECDES * recdes, HEAP_CACHE_ATTRINFO * attr_info)
...
+    case DB_TYPE_OBJECT:
+    case DB_TYPE_SET:
+    case DB_TYPE_MULTISET:
+    case DB_TYPE_SEQUENCE:
+    case DB_TYPE_ELO:
+    case DB_TYPE_JSON:
+    case DB_TYPE_POINTER:
+    case DB_TYPE_ERROR:
+      /* Not supported */
+      size = 0;
```
wrlawodms, line 12594: How about `assert (false)`?

src/transaction/log_manager.c
joohok: Oh.. yes, it includes `\0'
I will change it to 3

src/transaction/log_manager.c
joohok: Following the spec, CDC doesn't support above data types but has to send the data to the user as a NULL.
So I set the size to the 0, and this data will be sent to the user as NULL.

src/transaction/log_manager.c
joohok: Thanks.

src/transaction/log_manager.c
joohok: Since there is only one variable in the simple getter function, I don't think it's necessary to specify the size for what.

src/transaction/log_manager.c
joohok: For the data `B'10001'`, `10001` is stored in DB_VALUE. 

db_get_string_length (value) = 5    //  number of 1, 0 in B'10001'

and when it convert value from B'10001' to X'11' (For reducing the size, it make binary to hex)  
((5 + 3) / 4 ) = 2           / number of 1, 0 in X'11'

And `(db_get_string_length (value) + 3) / 4) + 4` includes `\0`, so I will change it to 
`(db_get_string_length (value) + 3) / 4) + 3`

src/transaction/log_manager.c
eido5: Ok. I understand it

src/transaction/log_manager.c
wrlawodms: Okay

src/transaction/log_manager.c
wrlawodms: Aha, 4 digits to 1 digit, OK, I got it.

wrlawodms (approved):

hornetmj (approved):

PR #3608, opened 2022-05-17
[CBRD-24346] Remove vargrind warning caused by CDC

joohok:
http://jira.cubrid.org/browse/CBRD-24346

Purpose

Remove vargrind warnings caused by CDC
1. Variable initialization
2. Invalid read of size 2 (cdc_get_recdes(), log_manager.c: 11520)
   * The insert record is saved without the mvcc header, and in order to interpret this record, 
   attaching the mvcc header must be preceded. At this time, the memory size calculation was wrong
3. Source and destination overlap (CDC_UPDATE_TEMP_LOGPAGE . log_impl.h:317)
   * A function that stores LOG PAGE in a temporary buffer. 
   * If the pageid of the LSA to be interpreted (process_lsa) is different from the pageid stored in the temporary buffer, 
   the page should be fetched and stored. 
   * However, in the existing implementation, the problem occurred while the same page was continuously stored 
   in the temporary buffer.

### DIFF
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -10560,12 +10560,14 @@ logtb_tran_update_stats_online_index_rb (THREAD_ENTRY * thread_p, void *data, vo
 static int
 cdc_log_extract (THREAD_ENTRY * thread_p, LOG_LSA * process_lsa, CDC_LOGINFO_ENTRY * log_info_entry)
 {
-  LOG_LSA cur_log_rec_lsa, next_log_rec_lsa;
+  LOG_LSA cur_log_rec_lsa = LSA_INITIALIZER;
+  LOG_LSA next_log_rec_lsa = LSA_INITIALIZER;
 
   LOG_PAGE *log_page_p = NULL;
   char log_pgbuf[IO_MAX_PAGE_SIZE + MAX_ALIGNMENT];
 
-  LOG_RECORD_HEADER *log_rec_header, *nx_rec_header;
+  LOG_RECORD_HEADER *log_rec_header = NULL;
+  LOG_RECORD_HEADER *nx_rec_header = NULL;
 
   char *tran_user = NULL;
   int trid;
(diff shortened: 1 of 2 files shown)


wrlawodms (approved):

src/transaction/log_manager.c
```diff
@@ -10678,7 +10679,7 @@ cdc_log_extract (THREAD_ENTRY * thread_p, LOG_LSA * process_lsa, CDC_LOGINFO_ENT
 	/*supplemental log info types : time, tran_user, undo image */
 	LOG_REC_SUPPLEMENT *supplement;
 	int supplement_length;
-	SUPPLEMENT_REC_TYPE rec_type;
+	SUPPLEMENT_REC_TYPE rec_type = LOG_SUPPLEMENT_LARGER_REC_TYPE;
```
wrlawodms, line 10682: A question:
Why does only `rec_type` has to be initialized, not `supplement` and the other? I don't mean for you to fix it. I actually don't think rec_type has to be initialized. I'd just like to know they Valgrind complains only about this.

joohok: Yes, Valgrind checks the initialization of variables which are used for conditional jump (switch (rec_type) case)

wrlawodms: thank you for the information :)

eido5:
I just wonder, does memory leak actually happen? or are just messages printed?

joohok:
> I just wonder, does memory leak actually happen? or are just messages printed?

A memory leak occurs only when, to be precise, valgrind gives a "definitely lost" warning. This issue only removes the warnings not memory leak. Please refer to it when you deal with vargrind issues in the future :(

eido5:
> > I just wonder, does memory leak actually happen? or are just messages printed?
> 
> A memory leak occurs only when, to be precise, valgrind gives a "definitely lost" warning. This issue only removes the warnings not memory leak. Please refer to it when you deal with vargrind issues in the future :(

Oh, I see thank you!

eido5 (approved):

hornetmj (approved):

PR #3613, opened 2022-05-17
[LETS-210] logging b-tree operations for atomic replication: btree_key_remove_object_and_keep_visible_first

cristiarg:

Change the order of b-tree operations to achieve a top-down approach/first-to-last approach:
- the leaf page is modified before the other overflow pages are modified (in order)

Start atomic sysops for btree operations that modify multiple pages.

Other:
- minor modification to a switch statement to prevent accidental overlook if new enum member is ever added

Note:
As re-using the atomic sysop mechanism beyond its initial scope might not be handled correctly in all cases (eg: nested atomic sysops) a different mechanism resulted after extensive discussions. It has been documented in http://jira.cubrid.org/browse/LETS-467 and represents and alternative to the current implementation.


alexstanro (approved):
Ok, let's try in this way at this moment. However, we have the risk to have several nested atomic sysop that currently are not correctly handled by our system. We could reconsider the approach later, if we found such case.

PR #3625, opened 2022-05-19
[LETS-462] Allow for atomic replication sequences with multiple log records affecting the same page

adrianbrab:
http://jira.cubrid.org/browse/LETS-462

During atomic replication the same sequence might have logs that affect the same page. With existing mechanism each log had to fix its page leading to double fixes in the situation previously described. 

To overcome this issue a new class was added called 'atomic_replication_sequence', this will provide an additional map of `VPID` to `PAGE_PTR` to check before fixing if the page is already part of the sequence.

The new class hierarchy:
atomic_replication_helper
 |_______ atomic_replication_sequence
________ |______________ atomic_replication_unit

### DIFF
--- a/src/transaction/atomic_replication_helper.cpp
+++ b/src/transaction/atomic_replication_helper.cpp
@@ -40,11 +40,11 @@ namespace cublog
 	// the page is no longer relevant
 	assert (false);
       }
-    vpid_set_type &vpids = m_atomic_sequences_vpids_map[tranid];
+    vpid_set_type &vpids = m_vpid_sets_map[tranid];
     vpids.insert (vpid);
 #endif
 
-    int error_code = m_atomic_sequences_map[tranid].add_atomic_replication_unit (thread_p, record_lsa, rcvindex, vpid,
+    int error_code = m_sequences_map[tranid].add_atomic_replication_unit (thread_p, record_lsa, rcvindex, vpid,
 		     redo_context, record_info);
     if (error_code != NO_ERROR)
       {
@@ -57,7 +57,7 @@ namespace cublog
 #if !defined (NDEBUG)
   bool atomic_replication_helper::check_for_page_validity (VPID vpid, TRANID tranid) const
   {
-    for (auto const &vpid_sets_iterator : m_atomic_sequences_vpids_map)
+    for (auto const &vpid_sets_iterator : m_vpid_sets_map)
       {
 	if (vpid_sets_iterator.first != tranid)
 	  {
@@ -78,8 +78,8 @@ namespace cublog
 
   bool atomic_replication_helper::is_part_of_atomic_replication (TRANID tranid) const
   {
-    const auto iterator = m_atomic_sequences_map.find (tranid);
-    if (iterator == m_atomic_sequences_map.cend ())
+    const auto iterator = m_sequences_map.find (tranid);
+    if (iterator == m_sequences_map.cend ())
       {
 	return false;
       }
@@ -89,18 +89,18 @@ namespace cublog
 
   void atomic_replication_helper::unfix_atomic_replication_sequence (THREAD_ENTRY *thread_p, TRANID tranid)
   {
-    auto iterator = m_atomic_sequences_map.find (tranid);
-    if (iterator == m_atomic_sequences_map.end ())
+    auto iterator = m_sequences_map.find (tranid);
+    if (iterator == m_sequences_map.end ())
       {
 	assert (false);
 	return;
       }
 
     iterator->second.unfix_sequence (thread_p);
-    m_atomic_sequences_map.erase (iterator);
+    m_sequences_map.erase (iterator);
 
 #if !defined (NDEBUG)
-    m_atomic_sequences_vpids_map.erase (tranid);
+    m_vpid_sets_map.erase (tranid);
 #endif
   }
   /****************************************************************************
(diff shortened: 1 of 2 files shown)


src/transaction/atomic_replication_helper.hpp
```diff
@@ -57,40 +58,64 @@ namespace cublog
...
+	  };
+
+	  std::vector<atomic_replication_unit> m_atomic_replication_unit_vector;
+	  std::map<VPID, PAGE_PTR> m_atomic_sequence_pages_map;
       };
 
-      using atomic_replication_sequence_type = std::vector<atomic_replication_unit>;
-      std::map<TRANID, atomic_replication_sequence_type> m_atomic_sequences_map;
-
+      std::map<TRANID, atomic_replication_sequence> m_atomic_sequences_map;
```
cristiarg: Nit: rename to `m_sequences` or `m_sequence_map`. Everything is about "atomic" in these classes.

src/transaction/atomic_replication_helper.hpp
```diff
@@ -57,40 +58,64 @@ namespace cublog
...
+
+	      VPID m_vpid;
+	    private:
+	      log_lsa m_record_lsa;
+	      PAGE_PTR m_page_ptr;
+	      PGBUF_WATCHER m_watcher;
+	      LOG_RCVINDEX m_record_index;
+	  };
+
+	  std::vector<atomic_replication_unit> m_atomic_replication_unit_vector;
```
cristiarg: Nit: rename to `m_units` or `m_unit_vec`.

src/transaction/atomic_replication_helper.hpp
```diff
@@ -57,40 +58,64 @@ namespace cublog
...
+	  std::map<VPID, PAGE_PTR> m_atomic_sequence_pages_map;
```
cristiarg: Nit: rename to `m_pages` or `m_page_map`.

src/transaction/atomic_replication_helper.cpp
```diff
@@ -98,24 +96,60 @@ namespace cublog
...
+  {
+    for (size_t i = 0; i < m_atomic_replication_unit_vector.size (); i++)
+      {
+	auto iterator = m_atomic_sequence_pages_map.find (m_atomic_replication_unit_vector[i].m_vpid);
+	if (iterator != m_atomic_sequence_pages_map.end ())
+	  {
+	    m_atomic_replication_unit_vector[i].unfix_page (thread_p);
+	    m_atomic_sequence_pages_map.erase (iterator);
+	  }
+      }
```
cristiarg, line 143: Unfixes need to happen in the order given by `m_atomic_replication_unit_vector`. Which is what the code does and it is correct.
However, there is no need to check the page and remove it from `m_atomic_sequence_pages_map` because, at the end of the outer iteration, all pgptr's which are also in the `m_atomic_sequence_pages_map` should have been cleared - assuming the two containers are consistent, which they should be.
Thus, the inner part should only be done as a debug-only check at most. If not, a simple vector.clear after all unfixes should be enough.

src/transaction/atomic_replication_helper.cpp
```diff
@@ -215,4 +250,18 @@ namespace cublog
...
+  PAGE_PTR atomic_replication_helper::atomic_replication_sequence::atomic_replication_unit::get_page_ptr ()
+  {
+    if (m_page_ptr != nullptr)
+      {
+	return m_page_ptr;
+      }
+    return m_watcher.pgptr;
+  }
+
+  void atomic_replication_helper::atomic_replication_sequence::atomic_replication_unit::set_page_ptr (PAGE_PTR &ptr)
```
cristiarg:
```suggestion
  void atomic_replication_helper::atomic_replication_sequence::atomic_replication_unit::set_page_ptr (const PAGE_PTR &ptr)
```
maybe?

src/transaction/atomic_replication_helper.hpp
cristiarg: typedefs for containers

src/transaction/atomic_replication_helper.cpp
```diff
@@ -98,24 +96,60 @@ namespace cublog
...
+	  {
+	    return error_code;
+	  }
+	m_atomic_sequence_pages_map.emplace (vpid,  m_atomic_replication_unit_vector.back ().get_page_ptr ());
+      }
+    else
+      {
+	m_atomic_replication_unit_vector.back ().set_page_ptr (iterator->second ());
+      }
+    m_atomic_replication_unit_vector.back ().apply_log_redo (thread_p, redo_context, record_info);
```
razvanradulescu: I don't know if this is the final version of the code or not:
but having both page fix and apply redo in the same phase may lead to undesired page unfixes of previously fixed pages during current ordered fix.
If a future refactoring task will separate into two phases, please ignore this.

cristiarg: It will not remain like this. It will be a separate modification - http://jira.cubrid.org/browse/LETS-460 .

src/transaction/atomic_replication_helper.cpp
adrianbrab: I made this way in order to prevent double unfix on a page, after the VPID is removed from the map if another unit comes with the same VPID it will not perform the unfix.

cristiarg: You're right. Mea culpa.

cristiarg (approved):

PR #3630, opened 2022-05-22
[LETS-465] complete transaction mvccid on passive transaction server

cristiarg:
http://jira.cubrid.org/browse/LETS-465

Register and complete mvccid's on passive transaction server when transaction commits/aborts.

Sub-transactions (implemented as sysops) which might also allocate mvccid's will be implemented separately.

Implementation:
- mvccid extraction from log records was already present in `replicator::read_and_redo_record`
- reused the `m_replicator_mvccid` implementation already used for `LOG_ASSIGNED_MVCCID` to keep track of tranid - mvccid pairs until the occurence of either a commit or an abort
- only one `mvccid` per transaction is allowed (assert in `replicator_mvcc::new_assigned_mvccid`) - sub-transactions will be supported by a subsequent patch

### DIFF
--- a/src/transaction/log_replication.cpp
+++ b/src/transaction/log_replication.cpp
@@ -216,29 +216,29 @@ namespace cublog
 	switch (header.type)
 	  {
 	  case LOG_REDO_DATA:
-	    read_and_redo_record<log_rec_redo> (thread_entry, header, m_redo_lsa);
+	    read_and_redo_record<LOG_REC_REDO> (thread_entry, header, m_redo_lsa);
 	    break;
 	  case LOG_MVCC_REDO_DATA:
-	    read_and_redo_record<log_rec_mvcc_redo> (thread_entry, header, m_redo_lsa);
+	    read_and_redo_record<LOG_REC_MVCC_REDO> (thread_entry, header, m_redo_lsa);
 	    break;
 	  case LOG_UNDOREDO_DATA:
 	  case LOG_DIFF_UNDOREDO_DATA:
-	    read_and_redo_record<log_rec_undoredo> (thread_entry, header, m_redo_lsa);
+	    read_and_redo_record<LOG_REC_UNDOREDO> (thread_entry, header, m_redo_lsa);
 	    break;
 	  case LOG_MVCC_UNDOREDO_DATA:
 	  case LOG_MVCC_DIFF_UNDOREDO_DATA:
-	    read_and_redo_record<log_rec_mvcc_undoredo> (thread_entry, header, m_redo_lsa);
+	    read_and_redo_record<LOG_REC_MVCC_UNDOREDO> (thread_entry, header, m_redo_lsa);
 	    break;
 	  case LOG_RUN_POSTPONE:
-	    read_and_redo_record<log_rec_run_postpone> (thread_entry, header, m_redo_lsa);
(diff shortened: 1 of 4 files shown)


src/transaction/log_replication.cpp
```diff
@@ -353,17 +353,21 @@ namespace cublog
...
-    const log_rv_redo_rec_info<T> record_info (rec_lsa, rectype,
+    const log_rv_redo_rec_info<T> record_info (rec_lsa, rec_header.type,
 	m_redo_context.m_reader.reinterpret_copy_and_add_align<T> ());
 
     // only mvccids that pertain to redo's are processed here
     const MVCCID mvccid = log_rv_get_log_rec_mvccid (record_info.m_logrec);
     assert_correct_mvccid (record_info.m_logrec, mvccid);
-    log_replication_update_header_mvcc_vacuum_info (mvccid, prev_rec_lsa, rec_lsa, m_bookkeep_mvcc_vacuum_info);
+    log_replication_update_header_mvcc_vacuum_info (mvccid, rec_header.back_lsa, rec_lsa, m_bookkeep_mvcc_vacuum_info);
+    if (m_replicate_mvcc && MVCCID_IS_NORMAL (mvccid))
```
cristiarg, line 367: `MVCC_IS_NORMAL` or `MVCCID_IS_VALID` ?
I suppose `MVCCID_IS_VALID` is more correct.

src/transaction/log_replication_mvcc.cpp
```diff
@@ -33,9 +32,19 @@ namespace cublog
   void
   replicator_mvcc::new_assigned_mvccid (TRANID tranid, MVCCID mvccid)
   {
-    assert (m_mapped_mvccids.find (tranid) == m_mapped_mvccids.cend ());
+    MVCCID_IS_NORMAL (mvccid);
```
razvanradulescu:
```suggestion
    assert (MVCCID_IS_NORMAL (mvccid));
```

src/transaction/log_replication.cpp
razvanradulescu: MVCCID_IS_VALID contains the special value MVCCID_ALL_VISIBLE (I think is used to replace an MVCCID when is not possible to change the record size - to remove the MVCCID, but we still want to set the data record as visible).
I think here, we don't deal with data records, but with log records and we don't expect to have the value MVCCID_ALL_VISIBLE.
Any reason not to have an assert instead of if (.. MVCCID_IS_NORMAL (mvccid)) ?

razvanradulescu (approved):

src/transaction/log_replication.cpp
cristiarg: Ok, thanks.
The function `read_and_redo_record` is typed for both non-mvcc and mvcc-containing log records (`LOG_REDO_DATA`, `LOG_MVCC_REDO_DATA`, `LOG_UNDOREDO_DATA`, ..). Thus, the function `log_rv_get_log_rec_mvccid` is also a template function which will, depending on implementation return either `MVCCID_NULL` or a valid one from the log record. That's why we need to have the check.
Maybe, inside the if an assert that the mvccid is not `MVCCID_ALL_VISIBLE` - but it's maybe superfluous.

src/transaction/log_replication_mvcc.cpp
cristiarg: Mea culpa. C&P error. Some compilers should warn against no-side-effect statement

PR #3634, opened 2022-05-23
[CBRD-24078] support CDC APIs on Windows

joohok:
http://jira.cubrid.org/browse/CBRD-24078

Purpose

Support window style server connection (new style connection) in CDC
CDC requires a connection to the server, and the method of connecting to the server on Windows is different from on Linux.
On Windows, the master process does not directly establish a connection with the server, but delivers the server's port id to the client. It needs to connect to the server again with this port id.

### DIFF
--- a/src/api/cubrid_log.c
+++ b/src/api/cubrid_log.c
@@ -663,15 +663,15 @@ cubrid_log_error:
       free_and_init (recv_data);
     }
 
-  queue_entry = css_find_queue_entry (g_conn_entry->buffer_queue, rid);
-  if (queue_entry != NULL)
-    {
-      queue_entry->buffer = NULL;
-      css_queue_remove_header_entry_ptr (&g_conn_entry->buffer_queue, queue_entry);
-    }
-
   if (g_conn_entry != NULL)
     {
+      queue_entry = css_find_queue_entry (g_conn_entry->buffer_queue, rid);
+      if (queue_entry != NULL)
+	{
+	  queue_entry->buffer = NULL;
+	  css_queue_remove_header_entry_ptr (&g_conn_entry->buffer_queue, queue_entry);
+	}
+
       css_free_conn (g_conn_entry);
       g_conn_entry = NULL;
     }
@@ -1896,15 +1896,15 @@ cubrid_log_error:
       free_and_init (recv_data);
     }
 
-  queue_entry = css_find_queue_entry (g_conn_entry->buffer_queue, rid);
-  if (queue_entry != NULL)
-    {
-      queue_entry->buffer = NULL;
-      css_queue_remove_header_entry_ptr (&g_conn_entry->buffer_queue, queue_entry);
-    }
-
   if (g_conn_entry != NULL)
     {
+      queue_entry = css_find_queue_entry (g_conn_entry->buffer_queue, rid);
+      if (queue_entry != NULL)
+	{
+	  queue_entry->buffer = NULL;
+	  css_queue_remove_header_entry_ptr (&g_conn_entry->buffer_queue, queue_entry);
+	}
+
       css_free_conn (g_conn_entry);
       g_conn_entry = NULL;
     }
--- a/src/api/cubrid_log.c
+++ b/src/api/cubrid_log.c
@@ -615,12 +615,12 @@ cubrid_log_connect_server_internal (char *host, int port, char *dbname)
       if (css_receive_data (g_conn_entry, rid, &recv_data, &recv_data_size, g_connection_timeout * 1000) != NO_ERRORS)
 	{
 	  CUBRID_LOG_ERROR_HANDLING (CUBRID_LOG_FAILED_CONNECT,
-				     "Failed to receive data while making new connection with server.\n");
+				     "Failed to receive the server port id from the master.\n");
 	}
 
       if (recv_data != NULL && recv_data_size == sizeof (int))
 	{
-	  int port id = ntohl (*((int *) recv_data));
+	  int port_id = ntohl (*((int *) recv_data));
 	  css_close_conn (g_conn_entry);
 
 	  if (recv_data != (char *) &reason)
--- a/src/api/cubrid_log.c
+++ b/src/api/cubrid_log.c
@@ -638,6 +638,12 @@ cubrid_log_connect_server_internal (char *host, int port, char *dbname)
 					 "Failed to connect to the server with new port id (%d)\n", port_id);
 	    }
 	}
+      else
+	{
+	  CUBRID_LOG_ERROR_HANDLING (CUBRID_LOG_FAILED_CONNECT,
+				     "Failed to get server port id (recv_data = %s , recv_data_size = %d) \n",
+				     recv_data, recv_data_size);
+	}
     }
   else
     {
--- a/src/api/cubrid_log.c
+++ b/src/api/cubrid_log.c
@@ -561,21 +561,112 @@ cubrid_log_set_extraction_user (char **user_arr, int arr_size)
 static int
 cubrid_log_connect_server_internal (char *host, int port, char *dbname)
 {
+  unsigned short rid = 0;
+
+  char *recv_data = NULL;
+  int recv_data_size;
+
+  int reason;
+
+  CSS_QUEUE_ENTRY *queue_entry;
   int err_code;
 
 #if defined (WINDOWS)
   (void) css_windows_startup ();
 #endif
 
-  g_conn_entry = css_connect_to_cubrid_server (host, dbname);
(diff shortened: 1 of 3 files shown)


src/api/cubrid_log.c
```diff
@@ -567,6 +571,10 @@ cubrid_log_connect_server_internal (char *host, int port, char *dbname)
   CSS_QUEUE_ENTRY *queue_entry;
   int err_code;
 
+#if defined (WINDOWS)
+  (void) css_windows_startup ();
```
hornetmj: Doesn't it need error handling?

src/api/cubrid_log.c
```diff
@@ -1930,6 +1977,10 @@ cubrid_log_finalize (void)
       CUBRID_LOG_ERROR_HANDLING (CUBRID_LOG_FAILED_DISCONNECT, NULL);
     }
 
+#if defined (WINDOWS)
+  (void) css_windows_startup ();
```
hornetmj:
```suggestion
  (void) css_windows_shutdown ();
```

hornetmj (approved):

src/api/cubrid_log.c
joohok: OK, I will add it. Thanks

src/api/cubrid_log.c
joohok: Thanks, I will fix it

src/api/cubrid_log.c
```diff
@@ -599,10 +610,45 @@ cubrid_log_connect_server_internal (char *host, int port, char *dbname)
...
+    {
+      css_queue_user_data_buffer (g_conn_entry, rid, sizeof (int), (char *) &reason);
+
+      if (css_receive_data (g_conn_entry, rid, &recv_data, &recv_data_size, g_connection_timeout * 1000) != NO_ERRORS)
+	{
+	  CUBRID_LOG_ERROR_HANDLING (CUBRID_LOG_FAILED_CONNECT,
+				     "Failed to receive the server port id from the master.\n");
+	}
+
+      if (recv_data != NULL && recv_data_size == sizeof (int))
```
wrlawodms: What if `recv_data_size` is not `sizeof (int)`? Is it not a error case?

src/api/cubrid_log.c
```diff
@@ -599,10 +610,45 @@ cubrid_log_connect_server_internal (char *host, int port, char *dbname)
...
+	{
+	  int port_id = ntohl (*((int *) recv_data));
+	  css_close_conn (g_conn_entry);
+
+	  if (recv_data != (char *) &reason)
```
wrlawodms: `reason` is a local variable and d`recv_data` seems to be allocated in a receive function. How are their addresses the same? and let me know what it means if possible.

src/api/cubrid_log.c
```diff
@@ -631,6 +677,10 @@ cubrid_log_connect_server_internal (char *host, int port, char *dbname)
       g_conn_entry = NULL;
     }
 
+#if defined (WINDOWS)
+  (void) css_windows_shutdown ();
```
wrlawodms, line 680: In no `cubrid_log_error` case (when successful), it also seems to have to be done.

src/api/cubrid_log.c
joohok: As I know, css_queue_user_data_buffer () (616L) stores `(char *) &reason` as a buffer in the queue.

when css_receive_data () (618L) is called with argument `recv_data`, 
internally css_receive_data () 1)  get a buffer (`&reason`) from the queue. 2) stores its address into `recv_data`, 3) return to the caller.

src/api/cubrid_log.c
joohok: Oh, yes. you are right. Thanks.

src/api/cubrid_log.c
joohok: `css_windows_startup` should be maintained until the CDC API is terminated (cubrid_log_finalize()). 
Because, after connected to the server, it communicates with the server through the API ( get lsa from the server , get log items from the server, ...).

src/api/cubrid_log.c
joohok: Oh, then this shouldn't be free. Thank you for your question.

src/api/cubrid_log.c
wrlawodms: Oh, I see. I didn't know the lifecycle. Then, another question: if `css_windows_startup()` failes, `cubrid_log_finalize()` is never called?

src/api/cubrid_log.c
joohok: If ,for some reason, the buffer (`&reason`) stored in the queue cannot be retrieved, free is necessary because internally malloc is performed on `recv_data`.

src/api/cubrid_log.c
joohok: yes, if css_windows_startup() fails, cubrid_log_connect_server () will return error to the client (error handling will be added for `css_windows_startup()` according to @hornetmj 's review) 

If user gets an error from `cubrid_log_connect_server()`, user have to try again with `cubrid_log_connect_server()`. 
Nothing can be called without `cubrid_log_connect_server()`'s success

wrlawodms: Thank you for the elaboration.

eido5 (approved):

src/api/cubrid_log.c
joohok: Entry is created with the key (request id : `rid`) input from css_queue_user_data_buffer (), and store the buffer in the entry.
If there is no buffer (`&reason`) matching request_id (rid) in the queue internally, memory is allocated and returned (`recv_data`). 
Since it may not be possible to create an entry to store the buffer inside css_queue_user_data_buffer(), it seems appropriate to free it(`recv_data`) here as defense code.

src/api/cubrid_log.c
```diff
@@ -599,21 +608,60 @@ cubrid_log_connect_server_internal (char *host, int port, char *dbname)
...
+	  if (g_conn_entry == NULL)
+	    {
+	      CUBRID_LOG_ERROR_HANDLING (CUBRID_LOG_FAILED_CONNECT,
+					 "Failed to connect to the server with new port id (%d)\n", port_id);
+	    }
+
+	  free_and_init (recv_data);
+	}
+      else
+	{
```
joohok, line 640: @wrlawodms 
The reason I didn't do css_close_conn() here is that 
when an error occurs, the label (`cubrid_log_error:`) cleans up the connection through css_free_conn().

src/api/cubrid_log.c
```diff
@@ -610,33 +608,36 @@ cubrid_log_connect_server_internal (char *host, int port, char *dbname)
 
   reason = ntohl (*(int *) recv_data);
 
+  if (recv_data != NULL)
```
wrlawodms: `reason = ntohl (*(int *) recv_data);` above means `recv_data` cannot be NULL.

joohok: Yes, I was fixing it now. You are really fast. Thanks

wrlawodms: haha.. you haven't requested a review yet. I just replied to the email notification. I'll wait. thank you.

joohok: haha.. All test has been done.

src/api/cubrid_log.c
```diff
@@ -599,21 +608,57 @@ cubrid_log_connect_server_internal (char *host, int port, char *dbname)
 
   reason = ntohl (*(int *) recv_data);
 
-  if (reason != SERVER_CONNECTED)
+  free_and_init (recv_data);
```
wrlawodms, line 611: If `recv_data != NULL`, but` recv_data_size != sizeof (int)`, memory leak will occur.

src/api/cubrid_log.c
wrlawodms: OK. However, there seems to be a problem in the code you mentioned. `queue_entry = css_find_queue_entry (g_conn_entry->buffer_queue, rid);` assumes that `g_conn_entry` is not NULL, but it is checked if it is NULL below.

src/api/cubrid_log.c
joohok: then `recv_data` will be deallocated at `cubrid_log_error:` label (661L)

src/api/cubrid_log.c
joohok: Oh, Thanks. I will fix it. 
it should check if `g_conn_entry` is NULL before clear the `queue_entry`.

src/api/cubrid_log.c
wrlawodms: I see. Thx.

wrlawodms (approved):

PR #3636, opened 2022-05-25
[CBRD-24372] Add new client type for loaddb to resolve migration issue by user schema.

youngjinj:
http://jira.cubrid.org/browse/CBRD-24372

Resolve this issue by removing the system parameter no_user_specified_name and setting DB_CLIENT_TYPE_ADMIN_LOADDB_COMPAT instead of DB_CLIENT_TYPE_ADMIN_UTILITY as the client type in the loaddb utility.


src/loaddb/load_server_loader.cpp
```diff
@@ -124,16 +124,17 @@ namespace cubload
 	return found;
       }
 
+#if defined(SERVER_MODE)
     /* This is the case when the loaddb utility is executed with the --no-user-specified-name option as the dba user. */
```
hgryoo, line 128: It seems that this comment should be changed as well.

src/object/trigger_manager.c
```diff
@@ -4254,8 +4254,7 @@ tr_find_trigger (const char *name)
       if (object == NULL)
 	{
 	  /* This is the case when the loaddb utility is executed with the --no-user-specified-name option as the dba user. */
```
hgryoo, line 4256: It seems that this comment should be changed as well.

src/loaddb/load_db.c
```diff
@@ -527,9 +527,9 @@ loaddb_internal (UTIL_FUNCTION_ARG * arg, int dba_mode)
   /* login */
   if (!args.user_name.empty () || !dba_mode)
     {
-      if (strcasecmp (args.user_name.c_str (), "DBA") == 0)
+      if (strcasecmp (args.user_name.c_str (), "DBA") == 0 && args.no_user_specified_name)
 	{
-	  db_set_client_type (DB_CLIENT_TYPE_ADMIN_UTILITY);
+	  db_set_client_type (DB_CLIENT_TYPE_ADMIN_LOADDB_COMPAT);
 	}
```
hgryoo, line 533: What `client_type` is set if "strcasecmp (args.user_name.c_str (), "DBA") == 0" is true and "args.no_user_specified_name" is false?

beyondykk9 (approved):

src/loaddb/load_server_loader.cpp
```diff
@@ -124,16 +124,17 @@ namespace cubload
 	return found;
       }
 
+#if defined(SERVER_MODE)
```
ctshim, line 127: In case of SA_MODE, is there no need to call location_class_for_all_users()?

src/loaddb/load_db.c
youngjinj: The default client type is DB_CLIENT_TYPE_DEFAULT.

Before adding the --no-user-specified-name option, there was no code to set the client type to DB_CLIENT_TYPE_ADMIN_UTILITY.

So, even if the user is a DBA before version 11.2, the client type is DB_CLIENT_TYPE_DEFAULT.

hgryoo (approved):

src/loaddb/load_server_loader.cpp
youngjinj: server_class_installer::locate_class is only called in server mode.
In sa mode, functions in the sa_class_installer namespace are called.

In css_conn_entry, client_type is defined only in SERVER_MODE,
so I also specified SERVER_MODE in server_class_installer::locate_class.
```
/* src/connection/connection_defs.h */

/*
 * This data structure is the interface between the client and the
 * communication software to identify the data connection.
 */
typedef struct css_conn_entry CSS_CONN_ENTRY;
struct css_conn_entry
{
  ...
#if defined(SERVER_MODE)
  ...
  BOOT_CLIENT_TYPE client_type;
  ...
#else
  ...
#endif
...
```

ctshim (approved):

shparkcubrid (approved):

hyunikn (approved):

PR #3641, opened 2022-05-27
[LETS-289] new mvcc log records field - parent_mvccid

cristiarg:

A subtransaction can have it's own mvccid. If that is the case, the transaction will also have a "main" mvccid (this is visible, among others, in the way the function `logtb_get_current_mvccid` is implemented).

With existing information it is not possible to know across replication boundaries (ie: on either page server or passive transaction server) whether the mvccid transmitted in log records is either a "main" mvccid or sub-mvccid.

Added a new `parent_mvccid` on log mvcc records.
If transaction has an mvccid allocated by a sub-transaction, this new field will contain the transaction's "main" mvccid (which, as indicated in implementations in `logtb_get_new_subtransaction_mvccid` and `logtb_get_current_mvccid`, must be valid).
Otherwise, the new field is `MVCCID_NULL`.

The purpose of this field is twofold:
- to discerne the nature of the mvccid across replication on passive transaction server and also - as either 'main' mvccid or sub-mvccid
- and to allow completion of both id's during replication

Only populating the new field is present in this patch. Using it across replication boundary will be a separate patch.


cristiarg:
Closing this PR in favor of a refactored one.

PR #3643, opened 2022-05-27
[LETS-460] [Re]organize page fix, apply and page unfix

adrianbrab:
http://jira.cubrid.org/browse/LETS-460

Applying the log right after the fix could lead to problems as the records are fixed one by one as they come to be read by the PTS and some might be unfixed and refixed after the apply procedure leading to inconsistency.

To solve this a new order of operations has been added, the pages are now fixed one by one as necessary and then when the atomic sequence end log record appears all the records in the sequence are applied sequentially and then the pages are all unfixed. To provide the required context information for the apply the `atomic_replication_sequence` now has a new member in the form of the `log_rv_redo_context m_redo_context` and the logic follows the model of the `replicatior` class.

### DIFF
--- a/src/transaction/atomic_replication_helper.cpp
+++ b/src/transaction/atomic_replication_helper.cpp
@@ -237,8 +237,7 @@ namespace cublog
       }
 
     redo_context.m_reader.advance_when_does_not_fit (sizeof (T));
-    const log_rv_redo_rec_info<T> record_info (m_record_lsa, rectype,
-	redo_context.m_reader.reinterpret_copy_and_add_align<T> ());
+    const log_rv_redo_rec_info<T> record_info (m_record_lsa, rectype, *redo_context.m_reader.reinterpret_cptr<T> ());
     if (log_rv_check_redo_is_needed (rcv.pgptr, record_info.m_start_lsa, redo_context.m_end_redo_lsa))
       {
 	rcv.reference_lsa = m_record_lsa;
--- a/src/transaction/atomic_replication_helper.cpp
+++ b/src/transaction/atomic_replication_helper.cpp
@@ -29,10 +29,13 @@ namespace cublog
   /*********************************************************************
    * atomic_replication_helper function definitions                    *
    *********************************************************************/
+  void atomic_replication_helper::add_atomic_replication_sequence (TRANID trid, log_rv_redo_context redo_context)
+  {
+    m_sequences_map.emplace (trid, redo_context);
+  }
 
-  template <typename T>
   int atomic_replication_helper::add_atomic_replication_unit (THREAD_ENTRY *thread_p, TRANID tranid, log_lsa record_lsa,
-      LOG_RCVINDEX rcvindex, VPID vpid, log_rv_redo_context &redo_context, const log_rv_redo_rec_info<T> &record_info)
+      LOG_RCVINDEX rcvindex, VPID vpid)
   {
 #if !defined (NDEBUG)
     if (!VPID_ISNULL (&vpid) && !check_for_page_validity (vpid, tranid))
@@ -44,8 +47,13 @@ namespace cublog
     vpids.insert (vpid);
 #endif
 
-    int error_code = m_sequences_map[tranid].add_atomic_replication_unit (thread_p, record_lsa, rcvindex, vpid,
-		     redo_context, record_info);
+    auto iterator = m_sequences_map.find (tranid);
+    if (iterator == m_sequences_map.cend ())
+      {
+	return ER_FAILED;
+      }
+
+    int error_code = iterator->second.add_atomic_replication_unit (thread_p, record_lsa, rcvindex, vpid);
     if (error_code != NO_ERROR)
       {
 	return error_code;
@@ -103,13 +111,17 @@ namespace cublog
     m_vpid_sets_map.erase (tranid);
 #endif
   }
-  /****************************************************************************
-   * atomic_replication_helper::atomic_replication_unit function definitions  *
-   ****************************************************************************/
-  template <typename T>
+  /********************************************************************************
+   * atomic_replication_helper::atomic_replication_sequence function definitions  *
+   ********************************************************************************/
+  atomic_replication_helper::atomic_replication_sequence::atomic_replication_sequence (log_rv_redo_context redo_cotext)
+    : m_redo_context { redo_cotext }
+  {
+
+  }
+
   int atomic_replication_helper::atomic_replication_sequence::add_atomic_replication_unit (THREAD_ENTRY *thread_p,
-      log_lsa record_lsa, LOG_RCVINDEX rcvindex, VPID vpid, log_rv_redo_context &redo_context,
-      const log_rv_redo_rec_info<T> &record_info)
+      log_lsa record_lsa, LOG_RCVINDEX rcvindex, VPID vpid)
   {
     m_units.emplace_back (record_lsa, vpid, rcvindex);
     auto iterator = m_page_map.find (vpid);
@@ -124,14 +136,24 @@ namespace cublog
       }
     else
       {
-	m_units.back ().set_page_ptr (iterator->second ());
+	m_units.back ().set_page_ptr (iterator->second);
       }
-    m_units.back ().apply_log_redo (thread_p, redo_context, record_info);
     return NO_ERROR;
+  }
+
+  void atomic_replication_helper::atomic_replication_sequence::apply_all_log_redos (THREAD_ENTRY *thread_p)
+  {
+    for (size_t i = 0; i < m_units.size (); i++)
+      {
+	m_units[i].apply_log_redo (thread_p, m_redo_context);
+      }
   }
 
   void atomic_replication_helper::atomic_replication_sequence::unfix_sequence (THREAD_ENTRY *thread_p)
   {
+    // sequenceally apply each log redo of the sequence before unfixing
+    apply_all_log_redos (thread_p);
+
     for (size_t i = 0; i < m_units.size (); i++)
       {
 	auto iterator = m_page_map.find (m_units[i].m_vpid);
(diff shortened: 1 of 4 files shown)


src/transaction/atomic_replication_helper.cpp
```diff
@@ -179,11 +235,14 @@ namespace cublog
 	rcv.pgptr = m_watcher.pgptr;
```
razvanradulescu, line 241: m_page_ptr and m_watcher.pgptr are mutually exclusive when handling pages. Adding asserts when setting and reading page would make this aspect more clear.

src/transaction/atomic_replication_helper.cpp
```diff
@@ -179,11 +235,14 @@ namespace cublog
...
       }
 
+    redo_context.m_reader.advance_when_does_not_fit (sizeof (T));
+    const log_rv_redo_rec_info<T> record_info (m_record_lsa, rectype,
+	redo_context.m_reader.reinterpret_copy_and_add_align<T> ());
     if (log_rv_fix_page_and_check_redo_is_needed (thread_p, m_vpid, rcv.pgptr, record_info.m_start_lsa,
```
razvanradulescu: From the PR description, I was expecting a split in phases : i) all pages in sequence are fixed, ii) all records are applied iii) all pages are unfixed
This function seems to fix  page (log_rv_fix_page_and_check_redo_is_needed)  and then apply (log_rv_redo_record_sync_apply).

adrianbrab: My mistake, I called the wrong function, changed now

cristiarg:
@adrianbrab Windows build fails.

src/transaction/atomic_replication_helper.hpp
```diff
@@ -62,7 +62,8 @@ namespace cublog
       class atomic_replication_sequence
       {
 	public:
-	  atomic_replication_sequence () = default;
+	  atomic_replication_sequence () = delete;
+	  atomic_replication_sequence (log_rv_redo_context redo_cotext);
```
cristiarg: make this ctor explicit and adapt; it makes code more clear
also, with this form, I suppose the `log_rv_redo_context` copy ctor is called too many times; which leads to too many `log_zip_realloc_if_needed` function calls which in turn allocate/deallocate too many times with no purpose resulting in poor performance

src/transaction/atomic_replication_helper.cpp
```diff
@@ -103,13 +111,17 @@ namespace cublog
     m_vpid_sets_map.erase (tranid);
 #endif
   }
-  /****************************************************************************
-   * atomic_replication_helper::atomic_replication_unit function definitions  *
-   ****************************************************************************/
-  template <typename T>
+  /********************************************************************************
```
cristiarg, line 115: +empty line

src/transaction/atomic_replication_helper.cpp
```diff
@@ -144,9 +166,9 @@ namespace cublog
   }
```
cristiarg:
```suggestion
```

src/transaction/atomic_replication_helper.cpp
```diff
@@ -103,13 +111,17 @@ namespace cublog
...
 #endif
   }
-  /****************************************************************************
-   * atomic_replication_helper::atomic_replication_unit function definitions  *
-   ****************************************************************************/
-  template <typename T>
+  /********************************************************************************
+   * atomic_replication_helper::atomic_replication_sequence function definitions  *
+   ********************************************************************************/
+  atomic_replication_helper::atomic_replication_sequence::atomic_replication_sequence (log_rv_redo_context redo_cotext)
```
cristiarg: typo
```suggestion
  atomic_replication_helper::atomic_replication_sequence::atomic_replication_sequence (log_rv_redo_context redo_context)
```

src/transaction/atomic_replication_helper.cpp
```diff
@@ -165,9 +187,43 @@ namespace cublog
...
+  atomic_replication_helper::atomic_replication_sequence::atomic_replication_unit::apply_log_redo (THREAD_ENTRY *thread_p,
+      log_rv_redo_context &redo_context)
+  {
+    redo_context.m_reader.set_lsa_and_fetch_page (m_record_lsa, log_reader::fetch_mode::FORCE);
+    const log_rec_header header = redo_context.m_reader.reinterpret_copy_and_add_align<log_rec_header> ();
+
+    switch (header.type)
+      {
+      case LOG_REDO_DATA:
+	apply_log_by_type<log_rec_redo> (thread_p, redo_context, header.type);
```
cristiarg: please use the upper case typedefs as used elsewhere; makes code harder to navigate/search (I know I also used the lowercase, but I'm also sticking to using the upper case typedefs).

src/transaction/atomic_replication_helper.cpp
```diff
@@ -179,11 +235,14 @@ namespace cublog
 	rcv.pgptr = m_watcher.pgptr;
       }
 
+    redo_context.m_reader.advance_when_does_not_fit (sizeof (T));
+    const log_rv_redo_rec_info<T> record_info (m_record_lsa, rectype,
+	redo_context.m_reader.reinterpret_copy_and_add_align<T> ());
```
cristiarg: I reckon in this case the `_and_add_align` part is not needed because the reader will not be used to advance to the a next page. T
hus, the reader will, is some cases - could be many cases - will needlessly advance and bring into view a new log page which might never be needed. Investigate using `reinterpret_cptr` only.

src/transaction/atomic_replication_helper.cpp
cristiarg: > From the PR description, I was expecting a split in phases : i) all pages in sequence are fixed, ii) all records are applied iii) all pages are unfixed This function seems to fix page (log_rv_fix_page_and_check_redo_is_needed) and then apply (log_rv_redo_record_sync_apply).

Has this comment been addressed?

src/transaction/atomic_replication_helper.cpp
```diff
@@ -124,14 +136,24 @@ namespace cublog
...
 
+  void atomic_replication_helper::atomic_replication_sequence::apply_all_log_redos (THREAD_ENTRY *thread_p)
+  {
+    for (size_t i = 0; i < m_units.size (); i++)
+      {
+	m_units[i].apply_log_redo (thread_p, m_redo_context);
+      }
+  }
+
   void atomic_replication_helper::atomic_replication_sequence::unfix_sequence (THREAD_ENTRY *thread_p)
```
cristiarg: Rename to include the "apply" action.

cristiarg (commented):
Partial review.

razvanradulescu (approved):

src/transaction/atomic_replication_helper.cpp
```diff
@@ -124,14 +138,24 @@ namespace cublog
...
   {
+    for (size_t i = 0; i < m_units.size (); i++)
+      {
+	m_units[i].apply_log_redo (thread_p, m_redo_context);
+      }
+  }
+
+  void atomic_replication_helper::atomic_replication_sequence::apply_and_unfix_sequence (THREAD_ENTRY *thread_p)
+  {
+    // sequenceally apply each log redo of the sequence before unfixing
```
cristiarg:
```suggestion
    // sequentially apply each log redo of the sequence before unfixing
```

src/transaction/atomic_replication_helper.cpp
```diff
@@ -96,20 +104,26 @@ namespace cublog
...
-  template <typename T>
+
+  /********************************************************************************
+   * atomic_replication_helper::atomic_replication_sequence function definitions  *
+   ********************************************************************************/
+
+  atomic_replication_helper::atomic_replication_sequence::atomic_replication_sequence (log_rv_redo_context redo_context)
+    : m_redo_context { redo_context }
+  {
+
```
cristiarg:
```suggestion
```

src/transaction/atomic_replication_helper.cpp
```diff
@@ -124,14 +138,24 @@ namespace cublog
...
+      {
+	m_units[i].apply_log_redo (thread_p, m_redo_context);
+      }
+  }
+
+  void atomic_replication_helper::atomic_replication_sequence::apply_and_unfix_sequence (THREAD_ENTRY *thread_p)
+  {
+    // sequenceally apply each log redo of the sequence before unfixing
+    apply_all_log_redos (thread_p);
+
```
cristiarg, line 160: add a comment explaining why this order is needed

src/transaction/atomic_replication_helper.cpp
```diff
@@ -165,25 +188,62 @@ namespace cublog
...
+	apply_log_by_type<LOG_REC_MVCC_UNDOREDO> (thread_p, redo_context, header.type);
+	break;
+      case LOG_RUN_POSTPONE:
+	apply_log_by_type<LOG_REC_RUN_POSTPONE> (thread_p, redo_context, header.type);
+	break;
+      case LOG_COMPENSATE:
+	apply_log_by_type<LOG_REC_COMPENSATE> (thread_p, redo_context, header.type);
+	break;
+      default:
+	break;
```
cristiarg, line 224: Assert that default should not happen since all log record types that need to be avoided are already avoided in the replicator.

src/transaction/atomic_replication_helper.cpp
```diff
@@ -165,25 +188,62 @@ namespace cublog
...
   template <typename T>
-  void atomic_replication_helper::atomic_replication_sequence::atomic_replication_unit::apply_log_redo (
-	  THREAD_ENTRY *thread_p, log_rv_redo_context &redo_context, const log_rv_redo_rec_info<T> &record_info)
+  void atomic_replication_helper::atomic_replication_sequence::atomic_replication_unit::apply_log_by_type (
+	  THREAD_ENTRY *thread_p, log_rv_redo_context &redo_context, LOG_RECTYPE rectype)
   {
     LOG_RCV rcv;
     if (m_page_ptr != nullptr)
       {
+	assert (m_page_ptr != nullptr && m_watcher.pgptr == nullptr);
```
cristiarg:
```suggestion
	assert (m_watcher.pgptr == nullptr);
```

src/transaction/atomic_replication_helper.cpp
```diff
@@ -165,25 +188,62 @@ namespace cublog
...
 	rcv.pgptr = m_page_ptr;
       }
     else
       {
+	assert (m_page_ptr == nullptr && m_watcher.pgptr != nullptr);
 	rcv.pgptr = m_watcher.pgptr;
```
cristiarg: I would suggest:
```suggestion
    else if (m_watcher.pgptr != nullptr)
      {
	assert (m_page_ptr == nullptr);
	rcv.pgptr = m_watcher.pgptr;
      }
    else
      {
        // assert_release here ...
      }
```

cristiarg (approved):

PR #3644, opened 2022-05-31
[LETS-289] new mvcc log record field - parent_mvccid

cristiarg:

A subtransaction can have it's own mvccid. If that is the case, the transaction will also have a "main" mvccid (this is visible, among others, in the way the function `logtb_get_current_mvccid` is implemented).

With existing information it is not possible to know across replication boundaries (ie: on either page server or passive transaction server) whether the mvccid transmitted in log records is either a "main" mvccid or sub-mvccid.

Added a new `parent_mvccid` in the MVCC-related union member of `LOG_REC_SYSOP_END`.
If transaction has an mvccid allocated by a sub-transaction, this new field will contain the transaction's "main" mvccid (which, as indicated in implementations in `logtb_get_new_subtransaction_mvccid` and `logtb_get_current_mvccid`, must be valid).
Otherwise, the new field should be ignored.

The purpose of this field is twofold:
- to discerne the nature of the mvccid across replication on passive transaction server and also - as either 'main' mvccid or sub-mvccid
- and to allow completion of both id's during replication

Implementation:
- introduced new anonymous struct in `LOG_REC_SYSOP_END` as union member for mvcc and  added field `parent_mvccid`
- adapted use all over
- only populated the new field where relevant in `log_sysop_end_logical_undo` with newly added function `logtb_get_current_mvccid_and_parent_mvccid`
- logged new field in dump functions where relevant

Only populating the new field is present in this patch. Using it across replication boundary will be a separate patch.

### DIFF
--- a/src/transaction/log_append.cpp
+++ b/src/transaction/log_append.cpp
@@ -1008,7 +1008,7 @@ prior_lsa_gen_undoredo_record_from_crumbs (THREAD_ENTRY *thread_p, LOG_PRIOR_NOD
 	    {
 	      assert (tdes->mvccinfo.sub_ids.size () == 1);
 	      *mvccid_p = tdes->mvccinfo.sub_ids.back ();
-	      assert (MVCCID_IS_VALID (tdes->mvccinfo.id));
+	      assert (MVCCID_IS_NORMAL (tdes->mvccinfo.id));
 	    }
 	  else
 	    {
@@ -1521,7 +1521,7 @@ prior_lsa_next_record_internal (THREAD_ENTRY *thread_p, LOG_PRIOR_NODE *node, LO
 
 	  /* Read from mvcc_undo structure */
 	  mvcc_undo = & ((LOG_REC_SYSOP_END *) node->data_header)->mvcc_undo_info.mvcc_undo;
-	  vacuum_info = &mvcc_undo ->vacuum_info;
+	  vacuum_info = &mvcc_undo->vacuum_info;
 	  mvccid = mvcc_undo->mvccid;
 
 	  /* Reset
--- a/src/transaction/log_append.cpp
+++ b/src/transaction/log_append.cpp
@@ -718,7 +718,6 @@ prior_lsa_gen_undoredo_record_from_crumbs (THREAD_ENTRY *thread_p, LOG_PRIOR_NOD
   int ulength, rlength, *data_header_ulength_p = NULL, *data_header_rlength_p = NULL;
   int total_length;
   MVCCID *mvccid_p = NULL;
-  MVCCID *parent_mvccid_p = NULL;
   LOG_TDES *tdes = NULL;
   char *data_ptr = NULL, *tmp_ptr = NULL;
   char *undo_data = NULL, *redo_data = NULL;
@@ -919,7 +918,6 @@ prior_lsa_gen_undoredo_record_from_crumbs (THREAD_ENTRY *thread_p, LOG_PRIOR_NOD
 
       /* Must also fill MVCCID field */
       mvccid_p = &mvcc_undo_p->mvccid;
-      parent_mvccid_p = &mvcc_undo_p->parent_mvccid;
 
     /* Fall through */
     case LOG_UNDO_DATA:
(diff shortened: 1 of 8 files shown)


src/transaction/log_append.cpp
```diff
@@ -1008,6 +1008,7 @@ prior_lsa_gen_undoredo_record_from_crumbs (THREAD_ENTRY *thread_p, LOG_PRIOR_NOD
 	    {
 	      assert (tdes->mvccinfo.sub_ids.size () == 1);
 	      *mvccid_p = tdes->mvccinfo.sub_ids.back ();
+	      assert (MVCCID_IS_VALID (tdes->mvccinfo.id));
```
razvanradulescu: Although I found other instances where we check IS_VALUE on mvcc_id from tdes, I expect this should only contain "NORMAL".
MVCCID_ALL_VISIBLE should be used only as special value on records (data/log), or as return value from visibility functions, not as mvccid in a transaction.

src/transaction/log_append.cpp
```diff
@@ -1512,8 +1520,8 @@ prior_lsa_next_record_internal (THREAD_ENTRY *thread_p, LOG_PRIOR_NODE *node, LO
 	  assert (sysop_end->type == LOG_SYSOP_END_LOGICAL_MVCC_UNDO);
 
 	  /* Read from mvcc_undo structure */
-	  mvcc_undo = & ((LOG_REC_SYSOP_END *) node->data_header)->mvcc_undo;
-	  vacuum_info = &mvcc_undo->vacuum_info;
+	  mvcc_undo = & ((LOG_REC_SYSOP_END *) node->data_header)->mvcc_undo_info.mvcc_undo;
+	  vacuum_info = &mvcc_undo ->vacuum_info;
```
razvanradulescu: unwanted empty space ?

src/transaction/log_manager.c
```diff
@@ -4301,14 +4301,19 @@ log_sysop_end_logical_undo (THREAD_ENTRY * thread_p, LOG_RCVINDEX rcvindex, cons
...
-      log_record.mvcc_undo.mvccid = logtb_get_current_mvccid (thread_p);
-      log_record.mvcc_undo.vacuum_info.vfid = *vfid;
-      LSA_SET_NULL (&log_record.mvcc_undo.vacuum_info.prev_mvcc_op_log_lsa);
+      log_record.mvcc_undo_info.mvcc_undo.undo.data.offset = NULL_OFFSET;
+      log_record.mvcc_undo_info.mvcc_undo.undo.data.volid = NULL_VOLID;
+      log_record.mvcc_undo_info.mvcc_undo.undo.data.pageid = NULL_PAGEID;
+      log_record.mvcc_undo_info.mvcc_undo.undo.data.rcvindex = rcvindex;
+      log_record.mvcc_undo_info.mvcc_undo.undo.length = undo_size;
+
+      MVCCID temp_mvccid = MVCCID_NULL, temp_parent_mvccid = MVCCID_NULL;
```
razvanradulescu: Nit: assignments on separate lines

src/transaction/log_record.hpp
```diff
@@ -219,7 +219,10 @@ typedef struct log_rec_mvcc_undoredo LOG_REC_MVCC_UNDOREDO;
 struct log_rec_mvcc_undoredo
 {
   LOG_REC_UNDOREDO undoredo;	/* Undoredo information */
-  MVCCID mvccid;		/* MVCC Identifier for transaction */
+  MVCCID mvccid;		/* MVCC Identifier for transaction. If log record is for a sub-transaction, the
```
razvanradulescu: Are the comments relevant in this context ?

razvanradulescu (approved):

src/transaction/log_append.cpp
cristiarg: Corrected

src/transaction/log_manager.c

src/transaction/log_record.hpp
cristiarg: Reverted

src/transaction/log_append.cpp
cristiarg: Corrected

PR #3648, opened 2022-06-02
[LETS-289] complete sub-transactions (sysops) mvccid's

cristiarg:

Sub-transactions have their own MVCCID's that are completed when the sub-transaction is committed. The MVCCID is also completed on PTS.

Implementation:
- the mechanism for assigned mvccid is reused (introduced with http://jira.cubrid.org/browse/LETS-288)
- when a sub-transaction (sysop) mvccid is encountered - `LOG_SYSOP_END_LOGICAL_MVCC_UNDO` - it is recorded for future completion (by either `LOG_SYSOP_END_COMMIT` or `LOG_SYSOP_END_ABORT`)
- also added debug-only code to check that a `LOG_SYSOP_START_POSTPONE` never has a valid mvccid - `replicate_sysop_start_postpone` - not sure whether this invariant holds
- in function `replicator_mvcc::new_assigned_mvccid` provided checking for a situation that happens during "click counter" (selupd) scenarios where a first appearing mvccid turns out to actually be the sub-mvvcid of the transaction
- in function `replicator_mvcc::new_assigned_sub_mvccid_or_mvccid` the same investigative approach is taken to check whether the mvccid is either the main transaction mvccid or a new sub-mvccid
- for `complete_mvcc` the system transaction is being used; as noted in the comment, the approach is temporary and subject to further tests/verification (an alternative solution is to use another reserved transaction index/id in the transaction table specifically for replication; if needed, will be implemented separately)

Other:
- used newly added debug parameter to dump mvccid completion replication


src/transaction/log_record.hpp
```diff
@@ -219,7 +219,10 @@ typedef struct log_rec_mvcc_undoredo LOG_REC_MVCC_UNDOREDO;
 struct log_rec_mvcc_undoredo
 {
   LOG_REC_UNDOREDO undoredo;	/* Undoredo information */
-  MVCCID mvccid;		/* MVCC Identifier for transaction */
+  MVCCID mvccid;		/* MVCC Identifier for transaction. If log record is for a sub-transaction, the
```
cristiarg: I will remove these comments, they were removed but re-appeared after merge.

src/transaction/log_replication.cpp
```diff
@@ -272,15 +272,39 @@ namespace cublog
 	    m_most_recent_trantable_snapshot_lsa.store (m_redo_lsa);
 	    break;
 	  case LOG_MVCC_UNDO_DATA:
-	    read_and_bookkeep_mvcc_vacuum<LOG_REC_MVCC_UNDO> (header.type, header.back_lsa, m_redo_lsa, true);
+	  {
+	    m_redo_context.m_reader.advance_when_does_not_fit (sizeof (LOG_REC_MVCC_UNDO));
```
razvanradulescu: Is the code for handling alignment and record not fitting particular to this change ?
If not, I suggest have separate PR as a fix.
What about other records ?

Later edit : I noticed this is moved code and it is not a consitent rule for handling it.
It looks like a general problem which could be handled before switch-case.
It looks we already do "reinterpret_copy_and_add_align", we need to add  "advance_when_does_not_fit" before it, and get rid of this code for each record in switch-case.

src/transaction/log_replication.cpp
```diff
@@ -422,16 +444,96 @@ namespace cublog
...
   }
 
+  void replicator::replicate_sysop_end (TRANID tranid, const log_lsa &rec_lsa, const LOG_REC_SYSOP_END &log_rec)
+  {
+    assert (m_replicate_mvcc);
+
+    LOG_SYSOP_END_TYPE_CHECK (log_rec.type);
+    // TODO: according to implementation in log_sysop_commit_internal, these asserts should hold
+    // for other than sysop commit
+    //assert (!LSA_ISNULL (&log_rec.lastparent_lsa));
```
razvanradulescu: Still need these ?

src/transaction/log_replication.cpp
```diff
@@ -422,16 +444,96 @@ namespace cublog
...
+    //assert (LSA_LT (&log_rec.lastparent_lsa, &rec_lsa));
+
+    if (log_rec.type == LOG_SYSOP_END_LOGICAL_MVCC_UNDO)
+      {
+	assert (!LSA_ISNULL (&log_rec.lastparent_lsa));
+	assert (LSA_LT (&log_rec.lastparent_lsa, &rec_lsa));
+
+	// mvccid might be valid or not
+	if (MVCCID_IS_VALID (log_rec.mvcc_undo_info.mvcc_undo.mvccid))
```
razvanradulescu: MVCC_IS_NORMAL

src/transaction/log_replication.cpp
```diff
@@ -422,16 +444,96 @@ namespace cublog
...
+	  {
+	    _er_log_debug (ARG_FILE_LINE, "[REPLICATOR_MVCC] %s tranid=%d\n",
+			   log_sysop_end_type_string (log_rec.type), tranid);
+	  }
+	// only complete sub-ids, if found
+	m_replicator_mvccid->complete_sub_mvcc (tranid);
+      }
+    else
+      {
+	// nothing
```
razvanradulescu, line 424: Is an assert possible here ?

src/transaction/log_replication_mvcc.cpp
```diff
@@ -37,13 +38,75 @@ namespace cublog
...
+    if (prm_get_bool_value (PRM_ID_ER_LOG_PTS_REPL_DEBUG))
+      {
+	_er_log_debug (ARG_FILE_LINE, "[REPLICATOR_MVCC] new_assigned_mvccid tranid=%d mvccid=%llu\n",
+		       tranid, (unsigned long long)mvccid);
+	dump_map ();
+      }
+  }
+
+  void
+  replicator_mvcc::new_assigned_mvccid (TRANID tranid, MVCCID mvccid, MVCCID parent_mvccid)
```
razvanradulescu: Nit : a new function name would make the code more readable.

src/transaction/log_replication_mvcc.cpp
```diff
@@ -37,13 +38,75 @@ namespace cublog
...
+
+	    // TODO: a new parent mvccid appeared out of the blue;
+	    // not sure whether this is a valid scenario
+	    // however, it will be commited/aborted at the end of the transaction
+	    new_assigned_mvccid (tranid, parent_mvccid);
+	  }
+
+	if (found_it != m_mapped_mvccids.cend ())
+	  {
+	    assert (found_it->second.id == mvccid);
```
razvanradulescu: Shouldn't this be  ?
```suggestion
	    assert (found_it->second.id == parent_mvccid);
```

src/transaction/log_replication_mvcc.cpp
```diff
@@ -37,13 +38,75 @@ namespace cublog
...
+	    assert (found_it->second.sub_ids.empty ());
+	    found_it->second.sub_ids.push_back (mvccid);
+	  }
+	else
```
razvanradulescu: Is this else needed ?

src/transaction/log_replication_mvcc.cpp
```diff
@@ -37,13 +38,75 @@ namespace cublog
...
+		dump_map ();
+	      }
+
+	    // TODO: a new parent mvccid appeared out of the blue;
+	    // not sure whether this is a valid scenario
+	    // however, it will be commited/aborted at the end of the transaction
+	    new_assigned_mvccid (tranid, parent_mvccid);
+	  }
+
+	if (found_it != m_mapped_mvccids.cend ())
```
razvanradulescu: Nit : else instead of "if" and assert inside the else {} scope

src/transaction/log_replication.cpp
cristiarg: It cannot be implemented before switch because it depends on the type of log record. But you are right that the usage of both alignment and advancement functions is not consistent. I will move them to a separate patch.

src/transaction/log_replication.cpp
cristiarg: Done

src/transaction/log_replication.cpp
cristiarg: I think the `else` can be hit; just that there is nothing to be done. The else and logging is temporary anyway.

src/transaction/log_replication_mvcc.cpp
cristiarg: Corrected

src/transaction/log_replication_mvcc.cpp
```diff
@@ -37,13 +44,87 @@ namespace cublog
...
+	//  - both a new main mvccid and sub-mvccid are allocated at the same time (see implementation
+	//    of logtb_get_new_subtransaction_mvccid function)
+	//  - during the selupd implementation and transaction logging, first the sub-mvccid appears
+	//    as part of log records and it is registered as such
+	//  - at some point both the main mvccid and the sub-mvccid appear as part of a
+	//    LOG_SYSOP_END_LOGICAL_MVCC_UNDO record; in which case we re-assign the - previously thought -
+	//    main mvccid as sub-mvccid and the new main mvccid as proper main
+	//  - subsequently, MVCC log records are replicated with the sub-mvccid figuring as 'main' again
+	//
+	// TODO: might this situation actually be a transactional logging bug?
```
cristiarg, line 63: This `TODO` might need special scrutiny.

src/transaction/log_replication.cpp
```diff
@@ -366,6 +375,80 @@ namespace cublog
...
+
+    m_redo_context.m_reader.advance_when_does_not_fit (sizeof (LOG_REC_SYSOP_START_POSTPONE));
+    const LOG_REC_SYSOP_START_POSTPONE log_rec =
+	    m_redo_context.m_reader.reinterpret_copy_and_add_align<LOG_REC_SYSOP_START_POSTPONE> ();
+
+    LOG_SYSOP_END_TYPE_CHECK (log_rec.sysop_end.type);
+    // TODO: this assert does not hold, to repro:
+    //  - execute the scenario from http://jira.cubrid.org/browse/LETS-289
+    //  - wait some time - probabil for the vacuum to be executed - which will end up here
+    //assert (!LSA_ISNULL (&log_rec.sysop_end.lastparent_lsa));
```
razvanradulescu, line 446: Probably a sys op executed in context of a vacuum thread does not have a parent ?

src/transaction/mvcc_table.cpp
```diff
@@ -495,6 +495,7 @@ mvcctable::complete_mvcc (int tran_index, MVCCID mvccid, bool committed)
        *
        * It will be set to NULL after LOG_COMMIT
        */
+      assert (tran_index < m_transaction_lowest_visible_mvccids_size);
```
razvanradulescu: Nit: I think is wiser to isolate the changes from this file in separate PR.

razvanradulescu (approved):

src/transaction/mvcc_table.cpp
cristiarg: Reverted, will be done separately.

PR #3655, opened 2022-06-09
[LETS-472] complete mvccid without transaction interference

cristiarg:
http://jira.cubrid.org/browse/LETS-472

On passive transaction server complete mvccid's without interfering with transaction level bookkeeping.

Implementation:
- created a new version of the `complete_mvcc` function which does not update transaction lowest visibible mvccid
- called the new function from passive transaction server mvcc replication logic

A different possible implementation:
- consider that the log to be replicated is just a separate transaction
- currently there is only one reserved entry in the transaction table (`LOG_SYSTEM_TRAN_INDEX`, `LOG_SYSTEM_TRANID`);
- and there is a mechanism to reserve a number of the first transactions in the transaction table - see `trantable::num_assigned_indices` - for certain fixed purposes - see calls to `logtb_set_number_of_assigned_tran_indices` where the first transaction is the system transaction
- the "replication transaction" could be modelled as a second fixed transaction - maybe, only on passive transaction server
- and the completion of mvccid could be done using this second reserved transaction such that not to interfere with (read-only) transactions generated on the passive transaction server

### DIFF
--- a/cubrid/CMakeLists.txt
+++ b/cubrid/CMakeLists.txt
@@ -335,6 +335,8 @@ set(TRANSACTION_SOURCES
   ${TRANSACTION_DIR}/log_recovery_redo.cpp
   ${TRANSACTION_DIR}/log_recovery_redo_parallel.cpp
   ${TRANSACTION_DIR}/log_replication.cpp
+  ${TRANSACTION_DIR}/log_replication_jobs.cpp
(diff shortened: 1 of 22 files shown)


cristiarg:
Update.
Further testing revealed that the currently proposed solution does not work with long transactions.

PR #3656, opened 2022-06-09
[LETS-473] allow reallocating page during replication on passive transaction server

cristiarg:
http://jira.cubrid.org/browse/LETS-473

On passive transaction server, a page might be de-allocated as part of the normal transactional log replication. At some future time, again as part of replication, it needs to be accessed again to be allocated.
Provide a switch case for the `OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT` fetch mode under following conditions:
- only on passive transaction server
- only if the access is request for writing - that is, from replication; it is absolutely forbidden for read-only transactions (those generated by clients connecting to passive transaction server) to access non-existing pages

### DIFF
--- a/src/storage/page_buffer.c
+++ b/src/storage/page_buffer.c
@@ -2160,16 +2160,6 @@ try_again:
 	case RECOVERY_PAGE:
 	  /* fixing deallocated page is expected. fall through to return it. */
 	  break;
-
-	case OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT:
-	  /* page was allocated and present in passive transaction server's page buffer;
-	   * was deallocated (by replication) and is now requested to be allocated again (also
-	   * by replication); hence the assert for write access */
-	  assert (is_passive_transaction_server ());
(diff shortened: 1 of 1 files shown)


src/storage/page_buffer.c
```diff
@@ -2171,6 +2171,16 @@ pgbuf_fix_release (THREAD_ENTRY * thread_p, const VPID * vpid, PAGE_FETCH_MODE f
 	  PGBUF_BCB_CHECK_MUTEX_LEAKS ();
 	  pgbuf_unfix (thread_p, pgptr);
 	  return NULL;
+	case OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT:
```
razvanradulescu, line 2174: Nit : can these two last cases be moved first ? ( I always expect "default" to be at the end of switch.)

razvanradulescu (approved):

src/storage/page_buffer.c
cristiarg: I would like that too. But I prefer to keep changes in existing files to a minimum to ease merging from `develop`.

PR #3660, opened 2022-06-13
[CBRD-24386] remove unused code in btree

ctshim:
http://jira.cubrid.org/browse/CBRD-24386

In this issue, we refactoring the following matters.

There are structures and enum that are not used in the btree-related source, so it is annotated.
BTREE_RANGE_SEARCH_HELPER
RECINS_STRUCT

Change the range of use of the function to static.
btree_check_tree
btree_check_by_btid
btree_unpack_mvccinfo

Adjusts the order of comparison of OIDCMP() macros.

### DIFF
--- a/src/storage/btree_load.h
+++ b/src/storage/btree_load.h
@@ -92,9 +92,9 @@
 
 /* compare two object identifiers */
 #define OIDCMP(n1, n2) \
-  ((n1).volid == (n2).volid \
-   && (n1).pageid == (n2).pageid \
-   && (n1).slotid == (n2).slotid)
+  ((n1).pageid == (n2).pageid \
+   && (n1).slotid == (n2).slotid \
+   && (n1).volid == (n2).volid)
 
 /* Header (Oth) record of the page */
 #define HEADER 0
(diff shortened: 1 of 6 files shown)


shparkcubrid (approved):

beyondykk9 (approved):

src/storage/btree_load.h
```diff
@@ -92,9 +92,9 @@
 
 /* compare two object identifiers */
 #define OIDCMP(n1, n2) \
-  ((n1).volid == (n2).volid \
-   && (n1).pageid == (n2).pageid \
-   && (n1).slotid == (n2).slotid)
+  ((n1).pageid == (n2).pageid \
```
hyunikn, line 95: Why did you do this reordering?

ctshim: This macro only checks if the OID is the same.
A total of 3 comparisons (volid, pageid, slotid) are required, and it is efficient to first compare those that are unlikely to match.
For this reason, we checked the pageid, which has the widest range of values, first.

youngjinj (approved):

hgryoo (approved):

PR #3661, opened 2022-06-14
[CBRD-24392] After an index is created, filtering index is created, but not vice versa.

beyondykk9:
http://jira.cubrid.org/browse/CBRD-24392

In creating sequence an index on a table including filtered index, a bug is found as below.

```
csql> create index idx1 on nation(name);
Execute OK. (0.016340 sec)

csql> create index idx2 on nation(name) where name > 'A';
Execute OK. (0.007594 sec)

csql> drop index idx1 on nation;
Execute OK. (0.001760 sec)

csql> create index idx1 on nation(name);
In the command from line 1,

ERROR: Index "idx2" already defined for class "nation".
```

To create the filtered index and non-filtered index should be checked more thoroughly.


shparkcubrid (approved):

ctshim (approved):

youngjinj (approved):

beyondykk9:
The failed test cases in sql-test should be modified to pass the test because creating index is not failed even though creating multiple filtered index

hgryoo (approved):

hyunikn (approved):

PR #3664, opened 2022-06-14
[CBRD-24378] Fix the restoredb not to go beyond the given time (-d)

wrlawodms:
http://jira.cubrid.org/browse/CBRD-24378

**Purpose**
To prevent the restoredb from doing recovery beyond the given time with the -d option.

**Implementation**
- A new log record for the sysop commit is added. It includes the `donetime` like the LOG_COMMIT. This time information is used for the time-specific restore.
- For the backward compatibility, the old log record type still remains.
- The new log record type use the name of the old log record type name: LOG_COMMIT_WITH_POSTPONE.
- The name of the old record type is changed to LOG_COMMIT_WITH_POSTPONE_OBSOLETE.
- The macro number which the new log record type use was for LOG_WILL_COMMIT, which had been obsolete for a long. I believe no one has a log including this type.

### DIFF
--- a/src/transaction/log_recovery.c
+++ b/src/transaction/log_recovery.c
@@ -1132,8 +1132,12 @@ log_rv_analysis_compensate (THREAD_ENTRY * thread_p, int tran_id, LOG_LSA * log_
  * return: error code
  *
  *   tran_id(in):
- *   lsa(in/out):
+ *   log_lsa(in/out):
  *   log_page_p(in/out):
+ *   lsa(in/out):
+ *   is_media_crash(in):
+ *   stop_at(in):
+ *   did_incom_recovery(in/out):
  *
  * Note:
  */
--- a/src/transaction/log_recovery.c
+++ b/src/transaction/log_recovery.c
@@ -69,9 +69,11 @@ static int log_rv_analysis_postpone (THREAD_ENTRY * thread_p, int tran_id, LOG_L
 static int log_rv_analysis_run_postpone (THREAD_ENTRY * thread_p, int tran_id, LOG_LSA * log_lsa, LOG_PAGE * log_page_p,
 					 LOG_LSA * check_point);
 static int log_rv_analysis_compensate (THREAD_ENTRY * thread_p, int tran_id, LOG_LSA * log_lsa, LOG_PAGE * log_page_p);
-static int log_rv_analysis_commit_with_postpone (THREAD_ENTRY * thread_p, int tran_id, LOG_LSA * log_lsa, LOG_PAGE * log_page_p, LOG_LSA * prev_lsa, bool is_media_crash, time_t * stop_at, bool * did_incom_recovery);
+static int log_rv_analysis_commit_with_postpone (THREAD_ENTRY * thread_p, int tran_id, LOG_LSA * log_lsa,
+						 LOG_PAGE * log_page_p, LOG_LSA * prev_lsa, bool is_media_crash,
+						 time_t * stop_at, bool * did_incom_recovery);
 static int log_rv_analysis_commit_with_postpone_obsolete (THREAD_ENTRY * thread_p, int tran_id, LOG_LSA * log_lsa,
-						 LOG_PAGE * log_page_p);
+							  LOG_PAGE * log_page_p);
 static int log_rv_analysis_sysop_start_postpone (THREAD_ENTRY * thread_p, int tran_id, LOG_LSA * log_lsa,
 						 LOG_PAGE * log_page_p);
 static int log_rv_analysis_atomic_sysop_start (THREAD_ENTRY * thread_p, int tran_id, LOG_LSA * log_lsa);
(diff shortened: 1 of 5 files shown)


src/transaction/log_recovery.c
```diff
@@ -1379,9 +1380,42 @@ log_rv_analysis_complete (THREAD_ENTRY * thread_p, int tran_id, LOG_LSA * log_ls
...
        */
       log_lsa->pageid = NULL_PAGEID;
-      log_recovery_resetlog (thread_p, &record_header_lsa, prev_lsa);
       *did_incom_recovery = true;
 
+      /*
+       * We can't use LOG_FIND_TDES (tran_index) here because the compelete log record
+       * may be the first log record seen for the transaction.
+       */
+      auto tdes = logtb_rv_find_allocate_tran_index (thread_p, tran_id, log_lsa);
```
joohok: How about removing the `tran_index = logtb_find_tran_index (thread_p, tran_id);` in 1346L ? 
and move this line(1389L) to 1346L ? 

because, LOG_FIND_TDES () is called inside the `logtb_find_tran_index()`, and It may results in traversing the transaction table twice. 

if you do so, you also modify the `end` label which uses tran_index. (replace `tran_index` with `tdes->tran_index`)

wrlawodms: `logtb_rv_find_allocate_tran_index` may allocate the `tdes` object if it is the first time to call the function. In a normal case, not `is_media_crash`, if the `tdes` is already allocated, it is released to prevent it from undoing (by making it not active). If changed as you said, a `tdes` is allocated and released for nothing when the complete log record is the only record found in the log analysis phase.

wrlawodms: In the normal case, if the complete log record is the only log record found during the analysis, `logtb_find_tran_index` returns null tran id, and just finishes. However, If `logtb_rv_find_allocate_tran_index` is called, a transaction descriptor will be always allocated and released.

joohok: Oh, I understand. Useless tdes can be allocated. Thanks for explain.

joohok (approved):

src/transaction/log_recovery.c
```diff
@@ -1379,9 +1380,42 @@ log_rv_analysis_complete (THREAD_ENTRY * thread_p, int tran_id, LOG_LSA * log_ls
...
+	{
+	  /*
+	   * When there were postpone operations, prevent redoing them
+	   * and restore undo info starting from LOG_COMMIT_WITH_POSTPONE.
+	   */
+	  assert (tdes->state == TRAN_UNACTIVE_COMMITTED_WITH_POSTPONE);
+
+	  log_recovery_resetlog (thread_p, &tdes->rcv.tran_start_postpone_lsa, &NULL_LSA);
+
+	  auto header = (LOG_RECORD_HEADER *) ((char *) LOG_APPEND_PTR ());
```
hornetmj: What happens if the LOG_COMMIT record and LOG_COMMIT_WITH_POSTPONE record exist on the different log page?

src/transaction/log_recovery.c
```diff
@@ -1379,9 +1380,42 @@ log_rv_analysis_complete (THREAD_ENTRY * thread_p, int tran_id, LOG_LSA * log_ls
...
+
+      if (tdes->rcv.tran_start_postpone_lsa.is_null () == false)
+	{
+	  /*
+	   * When there were postpone operations, prevent redoing them
+	   * and restore undo info starting from LOG_COMMIT_WITH_POSTPONE.
+	   */
+	  assert (tdes->state == TRAN_UNACTIVE_COMMITTED_WITH_POSTPONE);
+
+	  log_recovery_resetlog (thread_p, &tdes->rcv.tran_start_postpone_lsa, &NULL_LSA);
```
hornetmj: tx1 - LOG_COMMIT_WITH_POSTPONE
...
tx2 - LOG_COMMIT
...
// stopat time

tx1 - LOG_COMMIT

Can the above scenario occur? If yes, the Tx2 has to be redone completely. But I think this change will make the Tx2 be aborted. Please give me your opinion.

src/transaction/log_recovery.c
hornetmj: log_recovery_resetlog() -> logpb_fetch_start_append_page() -> logpb_locate_page()

@joohok explains this.

wrlawodms: > @joohok explains this.

@hornetmj Does it mean no more elaboration is required?

src/transaction/log_recovery.c
wrlawodms: @hornetmj You're correct. I'll cover the case. Thank you so much.

src/transaction/log_recovery.c
hornetmj: @wrlawodms No, @joohok teaches me enough.

wrlawodms:
Another scenario to consider:
LOG_COMMIT_WITH_POSTPONE
LOG_RUN_POSTPONE (the page affected by this log is **flushed**)
//backuptime
//stopat
LOG_COMMIT (this transaction is on-going during the online backup)

If a page affected by a LOG_RUN_POSTPONE log record is flushed already, there seems no way to undo it. I didn't handle this case because the restoration process starts from the data volume in the backup volume so that the case doesn't happen. However, considering the online backup, it could happen, though unlikely.

wrlawodms:
I have changed to another approach: a new log record type for the sysop commit with a donetime. Please see the PR description.

joohok (approved):
LGTM

hornetmj (approved):

PR #3665, opened 2022-06-16
[CBRD-24400] The problem of comparing plan fixed and variable cost respectively

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-24400

```
if (af <= bf && aa <= ba)
    return PLAN_COMP_LT
if (bf <= af && ba <= aa)
    return PLAN_COMP_GT;
return PLAN_COMP_UNK;
```
An unknown case occurs because of the comparison as above. I modify it to compare with the sum of each.


src/optimizer/query_planner.c
```diff
@@ -4217,18 +4217,17 @@ qo_plan_cmp (QO_PLAN * a, QO_PLAN * b)
     {
       return PLAN_COMP_EQ;
     }
-  if (af <= bf && aa <= ba)
+  if (af + aa <= bf + ba)
```
ctshim, line 4220: Is it correct that the case of returning the PLAN_COMP_UNK disappears? It is a structure that is determined unconditionally by either side.

In the current comparison method, `af (fixed cost)` is treated the same as `aa (variable cost)`. Don't you need to weigh it?

ctshim (approved):

src/optimizer/query_planner.c
shparkcubrid: Did you read my comments on Teams? Please ask specific questions about what you do not understand or have questions about.

hgryoo (approved):

beyondykk9 (approved):

hyunikn (approved):

youngjinj (approved):

PR #3666, opened 2022-06-17
[CBRD-24401] In case of using UNION in an IN clause in an inline view, a segment fault occurs

beyondykk9:
http://jira.cubrid.org/browse/CBRD-24401

In case of using UNION in an IN clause in an inline view, a segment fault occurs when the number of columns does not match.

```
SELECT 1
FROM db_class
WHERE class_name IN (SELECT 1 FROM db_root UNION SELECT 1, 2 FROM db_root)
```

The above query results segmentation fault.


ctshim (approved):

shparkcubrid (approved):

src/parser/query_result.c
```diff
@@ -484,7 +484,10 @@ pt_get_select_list (PARSER_CONTEXT * parser, PT_NODE * query)
 
 	  if (cnt1 != cnt2)
 	    {
-	      PT_ERRORmf2 (parser, arg1, MSGCAT_SET_PARSER_SEMANTIC, MSGCAT_SEMANTIC_ARITY_MISMATCH, cnt1, cnt2);
+	      if (!pt_has_error (parser))
```
hgryoo: Why checking error for parser context is needed here?

youngjinj: I also think error checking is unnecessary.

beyondykk9: If no error cheking, error message can be printed twice or more because the sub-query is checked repleatedly as below.
```
SELECT * 
FROM (
    SELECT 1
    FROM db_class
    WHERE class_name IN (SELECT 1 FROM db_root UNION SELECT 1, 2 FROM db_root)
);
```

beyondykk9: @hgryoo 
This is a part of semantic checking. So, the parser context is necessary for prining line numer.

hyunikn (approved):

src/parser/query_result.c
hgryoo: the semantic check will be done from the deeper block. so If error handling works well, the error message should not be printed twice.
Previously, the error already set was overwritten by PT_ERRORmf2 if the parser has an error. 
Now, If there was no error, it sets MSGCAT_SEMANTIC_ARITY_MISMATCH. And if an error was set, the error previously set is not cleared.
I'd like to know that the behavior of error handling is changed here, Is there any possibility of inconsistent error display?

beyondykk9: I don't know why the error message is printed twice but it's printed only once by this modification.
I guessed the reason because the parser has the error messages for each query level including sub-query.

youngjinj (approved):

hgryoo (approved):

PR #3670, opened 2022-06-17
[CBRD-24389] do broker conf parameter values consistency check

kisoo-han:
http://jira.cubrid.org/browse/CBRD-24389

**Purpose**
* cubrid broker conf parameter value consistency check, if failed stop
  * value of **MAX_NUM_APPL_SERVER** < **MIN_NUM_APPL_SERVER**
  * value of **MIN_NUM_APPL_SERVER** <= 0
* create file/directory for **ADMIN_LOG_FILE**, if it is not exists

**Implementation**
N/A

**Remarks**


src/broker/broker_admin_pub.c
```diff
@@ -270,6 +272,13 @@ admin_start_cmd (T_BROKER_INFO * br_info, int br_num, int master_shm_id, bool ac
   broker_create_dir (get_cubrid_file (FID_SQL_LOG_DIR, path, BROKER_PATH_MAX));
   broker_create_dir (get_cubrid_file (FID_SLOW_LOG_DIR, path, BROKER_PATH_MAX));
   broker_create_dir (get_cubrid_file (FID_CUBRID_ERR_DIR, path, BROKER_PATH_MAX));
+
+  if (admin_log_file != NULL)
+    {
+      admin_log2 = strdup (admin_log_file);
+      broker_create_dir (dirname (admin_log2));
+    }
```
mhoh3963, line 284: I think that admin_log2 is allocated memory by strdup() so should be freed.

src/broker/broker_config.c
```diff
@@ -652,19 +655,21 @@ broker_config_read_internal (const char *conf_file, T_BROKER_INFO * br_info, int
...
 	}
 
+      br_info[num_brs].appl_server_num = br_info[num_brs].appl_server_min_num;
+
       br_info[num_brs].appl_server_max_num =
-	ini_getuint (ini, sec_name, "MAX_NUM_APPL_SERVER", DEFAULT_AS_MAX_NUM, &lineno);
-      if (br_info[num_brs].appl_server_max_num > APPL_SERVER_NUM_LIMIT)
+	ini_getint (ini, sec_name, "MAX_NUM_APPL_SERVER", DEFAULT_AS_MAX_NUM, &lineno);
+      if (br_info[num_brs].appl_server_max_num > APPL_SERVER_NUM_LIMIT ||
+	  br_info[num_brs].appl_server_max_num < br_info[num_brs].appl_server_min_num)
```
mhoh3963: How about returning PARAM_BAD_RANGE instead of PARAM_BAD_MAX_APPL_SERVER if app_server_man_num is greater than APPL_SERVER_NUM_LIMIT ?
When user meet that error, I think that user can mis-understand.

airnet73: How about handling the error by dividing the condition of the if statement into two?
If appl_server_max_num > APPL_SERVER_NUM_LIMIT, return PARAM_BAD_RANGE.
If appl_server_max_num < appl_server_min_num, return PARAM_BAD_MAX_APPL_SERVER.

src/broker/broker_admin_pub.c
kisoo-han: It's better to free memory by strdup (), thank you.

src/broker/broker_config.c
kisoo-han: It's a little bit complicated. My suggestion is:

1. if appl_server_max_num > APPL_SERVER_NUM_LIMIT or appl_server_max_num < 1
   return PARAM_BAD_RANGE
2. appl_server_max_num < appl_server_min_num
   return PARAM_BAD_MAX_APPL_SERVER
   with the message:

   "Value is out of range (MAX_NUM_APPL_SERVER must be greater than MIN_NUM_APPL_SERVER)."

Please comment!

airnet73: I think this suggestion is good.

airnet73 (approved):

mhoh3963 (approved):

PR #3671, opened 2022-06-17
[LETS-476] Do not compare std::function instances to nullptr

adrianbrab:
http://jira.cubrid.org/browse/LETS-476

Removed comparison with `nullptr` for `std::function`, as it will also be removed in C++20

### DIFF
--- a/src/communication/request_sync_client_server.hpp
+++ b/src/communication/request_sync_client_server.hpp
@@ -156,7 +156,7 @@ namespace cubcomm
     assert (a_incoming_request_handlers.size () > 0);
     for (const auto &pair: a_incoming_request_handlers)
       {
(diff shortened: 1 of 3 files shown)


src/communication/request_sync_client_server.hpp
```diff
@@ -156,7 +156,7 @@ namespace cubcomm
     assert (a_incoming_request_handlers.size () > 0);
     for (const auto &pair: a_incoming_request_handlers)
       {
-	assert (pair.second != nullptr);
+	assert (pair.second);
```
cristiarg, line 159: better to use `static_cast<bool>(...)` all over for explicitness

cristiarg (approved):
Approved with comments.

PR #3673, opened 2022-06-17
[LETS-423] reorganize code to allow reuse of common template functions for normal and atomic replication classes

cristiarg:

Reorganize code to allow reuse of templated member function between base `cublog::replicator` and `cublog::atomic_replicator`:
- separate needed jobs in separate file `log_replication_jobs`
- separate member templated functions in file `log_replication.cpp.hpp` and included it in both non-atomic and atomic compilation units
- clean includes

### DIFF
--- a/cubrid/CMakeLists.txt
+++ b/cubrid/CMakeLists.txt
@@ -335,6 +335,7 @@ set(TRANSACTION_SOURCES
   ${TRANSACTION_DIR}/log_recovery_redo.cpp
   ${TRANSACTION_DIR}/log_recovery_redo_parallel.cpp
   ${TRANSACTION_DIR}/log_replication.cpp
+  ${TRANSACTION_DIR}/log_replication_jobs.cpp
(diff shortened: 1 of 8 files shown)


cubrid/CMakeLists.txt
```diff
@@ -374,6 +375,8 @@ set(TRANSACTION_HEADERS
   ${TRANSACTION_DIR}/log_recovery_redo.hpp
   ${TRANSACTION_DIR}/log_recovery_redo_parallel.hpp
   ${TRANSACTION_DIR}/log_replication.hpp
+  ${TRANSACTION_DIR}/log_replication_jobs.hpp
+  ${TRANSACTION_DIR}/log_replication.cpp.hpp
```
adrianbrab, line 379: this file has both `.cpp` and `.hpp`

adrianbrab (approved):

cubrid/CMakeLists.txt
cristiarg: that's on purpose; it contains template implementations

PR #3674, opened 2022-06-21
[CBRD-24396] Ignore hints if there is an invalid index name in the USING INDEX clause

ctshim:
http://jira.cubrid.org/browse/CBRD-24396

In hints that use USE INDEX and USING INDEX syntax, ignore the error when the table name or index name is invalid.
In addition, in the case of invisible indexes or online indexes, ignore them and continue without processing errors.

The error message is left in the WARNING level log.
This log is left in the ERROR_LOG_DIR setup path in csql.err or cubic_broker.conf.

In order for the WARNING level log to be recorded in the log file, the error_log_warning=yes setting must be performed in the cubid.conf.


msg/ko_KR.utf8/cubrid.msg
```diff
@@ -1431,8 +1431,9 @@ $ LOADDB
 1350 잘못된 Synonym 값입니다.
```
youngjinj: ko_KR.utf8 should be displayed normally, but characters are broken after 1350.

src/base/error_code.h
```diff
@@ -1725,8 +1725,9 @@
 #define ER_SYNONYM_INVALID_VALUE                    -1350
 
 #define ER_SM_INVALID_FUNCTION_EXPRESSION_LENGTH    -1351
+#define ER_USING_INDEX_HINT_IGNORE                  -1352
```
youngjinj: It's trivial.

Index hints used in the query syntax are "use index" and "using index".
Wouldn't it be confusing because the name of "ER_USING_INDEX_HINT_IGNORE" only contains "using index"?

hgryoo: The error message of this error code is too general.

src/parser/name_resolution.c
```diff
@@ -6995,6 +6995,31 @@ pt_resolve_hint (PARSER_CONTEXT * parser, PT_NODE * node)
   return ER_FAILED;
 }
 
+static void
+pt_write_warning (PARSER_CONTEXT * parser, PT_NODE * index, int lineNo, int erSetNo, int msgNo)
```
hgryoo: code convention: no camal case
```suggestion
pt_write_warning (PARSER_CONTEXT * parser, PT_NODE * index, int line_no, int er_set_no, int msg_no)
```

src/base/error_code.h
ctshim: The SELECT syntax description in the manual also deals with USING only.
It's not a problem, so I'll keep it as it is.
Thanks.

msg/ko_KR.utf8/cubrid.msg
ctshim: The cause is unknown, but it seems to be a problem with the screen display.
If you receive the actual file and look at it, it's normal.

src/parser/view_transform.c
```diff
@@ -5687,22 +5687,49 @@ mq_translate_local (PARSER_CONTEXT * parser, PT_NODE * statement, void *void_arg
...
+  indexp = using_index ? *using_index : NULL;
   if (indexp != NULL && spec != NULL)
     {
-      for (; indexp; indexp = indexp->next)
+      bool is_ignore = false;
+      PT_NODE *prev = NULL;
+      while (indexp)
 	{
-	  if (pt_resolve_using_index (parser, indexp, spec) == NULL)
+	  if (pt_resolve_using_index (parser, indexp, spec, &is_ignore))
```
youngjinj: I would like to know the purpose of using is_ignore.
And I would like this description added as a comment.

Is_ignore in "else if" is for comparing the value set after the pt_resolve_using_index call?
It can be efficient because it does not compare when pt_resolve_using_index does not return null.
However, the flow is not easily understood.

shparkcubrid (approved):

src/parser/name_resolution.c
ctshim: I will fix it, thanks.

src/parser/name_resolution.c
```diff
@@ -7087,76 +7114,73 @@ pt_resolve_using_index (PARSER_CONTEXT * parser, PT_NODE * index, PT_NODE * from
...
 	    {
-	      PT_INTERNAL_ERROR (parser, "resolution");
-	      return NULL;
+	      continue;
 	    }
-
-	  range = spec->info.spec.range_var;
-	  entity = spec->info.spec.entity_name;
-	  if (range != NULL && entity != NULL && entity->info.name.original != NULL)
+	  if (au_fetch_class (classop, &class_, AU_FETCH_READ, AU_SELECT) != NO_ERROR)
```
youngjinj, line 7143: Shouldn't the error also be printed as pt_write_warning?

src/parser/name_resolution.c
```diff
@@ -7059,26 +7091,21 @@ pt_resolve_using_index (PARSER_CONTEXT * parser, PT_NODE * index, PT_NODE * from
 		      PT_INTERNAL_ERROR (parser, "resolution");
```
youngjinj, line 7091: Shouldn't the error also be printed as pt_write_warning?

youngjinj, line 7090: Shouldn't the error also be printed as pt_write_warning?

ctshim: The nature of the error is judged to be different.
Changed only for MSGCAT_SET_PARSER_SEMANTIC of the existing code.

src/parser/name_resolution.c
ctshim: Changed only for MSGCAT_SET_PARSER_SEMANTIC of the existing code.

youngjinj (approved):

hyunikn (approved):

src/base/error_code.h
ctshim: I corrected the error code and message with the help of hgryoo.

beyondykk9 (approved):

hgryoo (approved):

beyondykk9 (approved):

PR #3676, opened 2022-06-21
[CBRD-24275] seperate externalproject build/install msg from main stream

kisoo-han:
http://jira.cubrid.org/browse/CBRD-24275

**Purpose**
* make message output from build/install of **externalproject** to use it own log file instead of stdout
*  the configure, build, and install logs of external projects will be saved as separate log files in 
   ${CMAKE_CURRENT_BINARY_DIR}/external/...
* this will reduce 2800+ lines in CUBRID build log.

**Implementation**
N/A

**Remarks**
* LOG_CONFIGURE is not set to true in **libedit** because the **CUBRID build fails.**
* LOG_CONFIGURE of libexpat was excluded because off and on did not significantly affect the amount of the logfile output.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1108,6 +1108,7 @@ if(WITH_LIBOPENSSL STREQUAL "EXTERNAL")
     #e.g. https://www.openssl.org/source/openssl-1.1.1f.tar.gz
     externalproject_add(${LIBOPENSSL_TARGET}
       URL                  ${WITH_LIBOPENSSL_URL}
+      LOG_CONFIGURE        TRUE
       LOG_BUILD            TRUE
       LOG_INSTALL          TRUE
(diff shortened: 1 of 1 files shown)


airnet73:
How about adding LOG_CONFIGURE as well?
Adding LOG_CONFIGURE also reduces the configuration log.

kisoo-han:
I also considered adding LOG_CONFIGURE. 
However, in the case of libedit, CUBRID build failed when LOG_CONFIGURE was set to true.
This is why I left out LOG_CONFIGURE. 
Anyway, as you requested, I will set LOG_CONFIGURE to true for the remaining 5 external projects. 
Then, the log of about 2800 lines can be reduced from the main stream.

hgryoo (approved):

airnet73 (approved):

PR #3683, opened 2022-06-22
[CBRD-24411] change shell distribution pkg for sh/csh/zsh

kisoo-han:
http://jira.cubrid.org/browse/CBRD-24411

**Purpose**

* revise shell distribution package to accomodate

  * cubrid.csh to reflect creating ncurses links

  * add/create sh scripts for zsh/csh/sh to start CUBRID environments, .zshrc, .cshrc and .profile respectively

**Implementation**
N/A

**Remarks**


kisoo-han:
Because the change of the issue, it's better close this PR and open another one.

PR #3686, opened 2022-06-23
[CBRD-24410] Incorrect number of affected records when INSERT ON DUPLICATE KEY is executed

beyondykk9:
http://jira.cubrid.org/browse/CBRD-24410

When insert ON DUPLICATE KEY is performed, there is 1 duplicate record, and 1 record has been modified as a result. However, it says that there are two records affected.

Replace is also displayed in two cases in the same way.

csql> select * from nation where code like 'cd%';

```
  code                  name            continent             capital
========================================================================================
  'cd0'                 'old'           NULL                  NULL
  'cd1'                 'old1'          NULL                  NULL
  'cd2'                 'old2'          NULL                  NULL
  'cd3'                 'old3'          NULL                  NULL

4 rows selected. (0.003354 sec) Committed.

csql> INSERT INTO nation(code, name) VALUES ('cd1', 'new')
csql>   ON DUPLICATE KEY UPDATE name = 'new1', capital = 'cap1'
csql> ;
2 rows affected. (0.001580 sec) Committed.1 command(s) successfully processed. 
```

### DIFF
--- a/src/query/query_executor.c
+++ b/src/query/query_executor.c
@@ -11432,7 +11432,7 @@ qexec_execute_insert (THREAD_ENTRY * thread_p, XASL_NODE * xasl, XASL_STATE * xa
 		    {
 		      assert (force_count == 1);
 
-		      xasl->list_id->tuple_cnt += force_count * 2;
+		      xasl->list_id->tuple_cnt += force_count;
 		      continue;
 		    }
 		}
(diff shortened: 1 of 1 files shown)


beyondykk9:
This is not a bug. This is for compatibility with MySQL.

PR #3688, opened 2022-06-24
[LETS-480] btree assert force restart from root

cristiarg:
http://jira.cubrid.org/browse/LETS-480

On Passive Transaction Server (PTS), due to page desynchronization - pages retrieved from Page Server (PS) are "newer" than the current progress of the replication on PTS - sometimes the btree searches need to be interrupted with `force_restart_from_root` (eg: in btree_range_scan_descending_fix_prev_leaf) in the hope that, upon next re-entry into the search, the PTS replication would have advanced sufficiently as for the error to not happen again.

For the current patch, this force restart from root bumps into an existing assert which assumes that the scan must iterate the index in a descending order.
Disable the assert only on PTS.

### DIFF
--- a/src/storage/btree.c
+++ b/src/storage/btree.c
@@ -25289,7 +25289,9 @@ btree_range_scan (THREAD_ENTRY * thread_p, BTREE_SCAN * bts, BTREE_RANGE_SCAN_PR
 	  if (bts->force_restart_from_root)
 	    {
 	      /* Couldn't advance. Restart from root. */
-	      assert (bts->use_desc_index);
+	      /* TODO: not sure why this assert is present here. However, on passive transaction
+	       * server, it is hit when a page desynchronization situation is encountered */
+	      assert (bts->use_desc_index || is_passive_transaction_server ());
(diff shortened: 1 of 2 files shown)


src/storage/btree.c
```diff
@@ -25289,7 +25289,9 @@ btree_range_scan (THREAD_ENTRY * thread_p, BTREE_SCAN * bts, BTREE_RANGE_SCAN_PR
 	  if (bts->force_restart_from_root)
 	    {
 	      /* Couldn't advance. Restart from root. */
-	      assert (bts->use_desc_index);
+	      /* TODO: not sure why this assert is present here. However, on passive transaction
+	       * server, it is hit when a page desynchronization situation is encountered */
+	      assert (bts->use_desc_index || is_passive_transaction_server ());
```
razvanradulescu, line 25294: Before page ahead context, restart from root event occurred only in context of descending index scan.

razvanradulescu (approved):

src/storage/btree.c
cristiarg: Added a, hopefully, clarifying comment.

PR #3692, opened 2022-06-27
[LETS-424] Integrate atomic replicator in passive transaction server replication

adrianbrab:
http://jira.cubrid.org/browse/LETS-424

Integrated the `atomic_replicator` with the new changes that occurred on the normal replicator.
Also implemented the `lowest_unaplied_lsa` functionality that was first introduced in a previous issue.

### DIFF
--- a/src/transaction/atomic_replication_helper.cpp
+++ b/src/transaction/atomic_replication_helper.cpp
@@ -132,6 +132,25 @@ namespace cublog
     return atomic_sequence.can_end_atomic_sequence (sysop_parent_lsa);
   }
 
+  log_lsa atomic_replication_helper::get_the_lowest_start_lsa ()
+  {
+    log_lsa min_lsa = MAX_LSA;
+
+    for (auto const &sequence_map_iterator : m_sequences_map)
+      {
+	if (sequence_map_iterator.second.get_start_lsa () < min_lsa)
+	  {
+	    min_lsa = sequence_map_iterator.second.get_start_lsa ();
+	  }
+      }
+    return min_lsa;
+  }
+
+  bool atomic_replication_helper::is_any_atomic_sequence_open () const
+  {
+    return !m_sequences_map.empty ();
+  }
+
   /********************************************************************************
    * atomic_replication_helper::atomic_replication_sequence function definitions  *
    ********************************************************************************/
@@ -203,6 +222,11 @@ namespace cublog
     assert (!LSA_ISNULL (&sysop_parent_lsa));
 
     return m_start_lsa >= sysop_parent_lsa;
+  }
+
+  log_lsa atomic_replication_helper::atomic_replication_sequence::get_start_lsa () const
+  {
+    return m_start_lsa;
   }
 
   /*********************************************************************************************************
--- a/src/transaction/atomic_replication_helper.cpp
+++ b/src/transaction/atomic_replication_helper.cpp
@@ -146,11 +146,6 @@ namespace cublog
     return min_lsa;
   }
 
-  bool atomic_replication_helper::is_any_atomic_sequence_open () const
-  {
-    return !m_sequences_map.empty ();
-  }
-
   /********************************************************************************
    * atomic_replication_helper::atomic_replication_sequence function definitions  *
    ********************************************************************************/
@@ -361,10 +356,5 @@ namespace cublog
   void atomic_replication_helper::atomic_replication_sequence::atomic_replication_unit::set_page_ptr (const PAGE_PTR &ptr)
   {
     m_page_ptr = ptr;
-  }
-
-  LOG_LSA atomic_replication_helper::atomic_replication_sequence::atomic_replication_unit::get_lsa () const
-  {
-    return m_record_lsa;
   }
 }
--- a/src/transaction/log_replication.cpp
+++ b/src/transaction/log_replication.cpp
@@ -519,18 +519,10 @@ namespace cublog
     return m_redo_lsa;
   }
 
-  log_lsa replicator::get_lowest_unapplied_lsa () const
+  log_lsa
+  replicator::get_lowest_unapplied_lsa () const
   {
-    // TODO: needs to be refactored to work with the new replicators flavors
-    if (m_parallel_replication_redo == nullptr)
-      {
-	// sync
-	return get_highest_processed_lsa ();
-      }
-
-    // a different value will return from here when the atomic replicator is added
-    // for now this part should not be reached
-    assert (false);
+    return get_highest_processed_lsa ();
   }
 
   /*********************************************************************
--- a/src/transaction/log_replication.cpp
+++ b/src/transaction/log_replication.cpp
@@ -522,6 +522,9 @@ namespace cublog
   log_lsa
   replicator::get_lowest_unapplied_lsa () const
   {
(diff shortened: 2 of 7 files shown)


src/transaction/log_replication.cpp
```diff
@@ -519,18 +519,10 @@ namespace cublog
     return m_redo_lsa;
   }
 
-  log_lsa replicator::get_lowest_unapplied_lsa () const
+  log_lsa
+  replicator::get_lowest_unapplied_lsa () const
```
cristiarg, line 523: I think this function should not be called because, on PTS it will be superseeded by the derived one.
But, for now, just add a TODO to replace the function with an assert.
Am I correct?

src/transaction/log_replication_atomic.cpp
```diff
@@ -119,6 +119,7 @@ namespace cublog
 		// nested atomic replication
 		assert (false);
 	      }
+	    m_lowest_unapplied_lsa = m_redo_lsa;
```
cristiarg: I do not think this is correct because there can be concurrent atomic sequences - from other transactions that the transaction for which the current log record is - whose un-applied log records can have LSA's less that the current log record.
Please correct me if I'm wrong.
I think that the new "lowest unapplied lsa" is the minimum between all currently in progress atomic sequences **and** `m_redo_lsa`.

src/transaction/log_replication_atomic.cpp
```diff
@@ -220,4 +244,10 @@ namespace cublog
 	  }
       }
   }
+
+  log_lsa
+  atomic_replicator::get_lowest_unapplied_lsa () const
+  {
+    return m_lowest_unapplied_lsa;
```
cristiarg, line 254: Access to this variable needs to be synchronized with a mutex. Similar to `m_redo_lsa_mutex`.

cristiarg: What is the response in case there was never an atomic sequence processed?
What is the response if there was an atomic sequence present long ago - in relative replication times - in the past and, in the meantime, the replication has clearly advanced enough for that value to not be relevant at all?

src/transaction/log_replication.cpp
adrianbrab: Yes, I will add an assert

src/transaction/log_replication_atomic.cpp
adrianbrab: You are correct, I though this might be the case, I will implement the solution.

src/transaction/log_replication_atomic.cpp
adrianbrab: The value is updated to `m_redo_lsa` if no replication is in progress, it is also initialized in the same way as `m_redo_lsa`. I will add an access mutex.

src/transaction/atomic_replication_helper.cpp
```diff
@@ -205,6 +224,11 @@ namespace cublog
     return m_start_lsa >= sysop_parent_lsa;
   }
 
+  log_lsa atomic_replication_helper::atomic_replication_sequence::get_start_lsa () const
```
cristiarg, line 222: inline this function

src/transaction/log_replication_atomic.cpp
```diff
@@ -248,6 +254,14 @@ namespace cublog
...
   {
+    std::lock_guard<std::mutex> lockg (m_lowest_unapplied_lsa_mutex);
     return m_lowest_unapplied_lsa;
   }
+
+  void
+  atomic_replicator::set_lowest_unapplied_lsa (log_lsa value)
+  {
+    std::lock_guard<std::mutex> lockg (m_lowest_unapplied_lsa_mutex);
+    m_lowest_unapplied_lsa = value;
```
cristiarg: Perform the reconcile calculation between the value returned by the helper and `m_redo_lsa` here. Pseudo:
```
assert (m_redo_lsa is not NULL_LSA);
const LOG_LSA helper_lowest_unapplied_lsa = ....from helper
m_lowest_unapplied_lsa = std::min ( m_redo_lsa, helper_lowest_unapplied_lsa;
```
This code is executed in the - one - replication thread. So, only access to `m_lowest_unapplied_lsa` needs to be synched. The calculation itself can be done before the lock:
```
// .. calculation
{ // .. lock scope
  // lock
  // set
} // .. end lock scope
```

src/transaction/atomic_replication_helper.cpp
```diff
@@ -132,6 +132,25 @@ namespace cublog
...
+      {
+	if (sequence_map_iterator.second.get_start_lsa () < min_lsa)
+	  {
+	    min_lsa = sequence_map_iterator.second.get_start_lsa ();
+	  }
+      }
+    return min_lsa;
+  }
+
+  bool atomic_replication_helper::is_any_atomic_sequence_open () const
```
cristiarg: No need to have this function.
If there are no sequences open, `get_the_lowest_start_lsa` can answer with a guard value - `MAX_LSA` and the caller will reconcile with the value from the helper (either valid value or guard value) with the one from the atomic replicator.

src/transaction/log_replication_atomic.cpp
```diff
@@ -220,4 +246,24 @@ namespace cublog
...
+  {
+    std::lock_guard<std::mutex> lockg (m_lowest_unapplied_lsa_mutex);
+    return m_lowest_unapplied_lsa;
+  }
+
+  void
+  atomic_replicator::set_lowest_unapplied_lsa ()
+  {
+    assert (m_redo_lsa != NULL_LSA);
+    const LOG_LSA helper_lowest_unapplied_lsa = m_atomic_helper.get_the_lowest_start_lsa ();
```
cristiarg, line 261:
```suggestion
    const LOG_LSA helper_lowest_unapplied_lsa = m_atomic_helper.get_the_lowest_start_lsa ();
    assert (!LSA_ISNULL (helper_lowest_unapplied_lsa));
```

src/transaction/log_replication_atomic.cpp
```diff
@@ -220,4 +246,24 @@ namespace cublog
...
+    const LOG_LSA value_to_change = std::min (m_redo_lsa, helper_lowest_unapplied_lsa);
```
cristiarg: I might have mistaken with the suggestion. Not sure how `std::min` is compiler. Maybe the arguments are first promoted to `int64_t` via the `log_lsa` operator. Better use the ternary comparison operator.

src/transaction/log_replication_atomic.hpp
```diff
@@ -42,14 +42,20 @@ namespace cublog
...
       void redo_upto (cubthread::entry &thread_entry, const log_lsa &end_redo_lsa) override;
       template <typename T>
       void read_and_redo_record (cubthread::entry &thread_entry, const LOG_RECORD_HEADER &rec_header,
 				 const log_lsa &rec_lsa);
+      void set_lowest_unapplied_lsa ();
 
     private:
       atomic_replication_helper m_atomic_helper;
+      log_lsa m_lowest_unapplied_lsa;
+      bool m_should_update_lowest_lsa = true;
```
cristiarg: leftover?

cristiarg (approved):
Approved with outstanding remarks.

PR #3697, opened 2022-07-01
[CBRD-24407] Changes for lock_timeout, isolation_level in csql are reflected

hyahong:
http://jira.cubrid.org/browse/CBRD-24407

Purpose
When change lock_timeout or isolation_level through "set system parameters" in csql, it is not reflected in actual processing and utilities (i.e. lockdb).

Implementation
N/A

Remarks
N/A


src/transaction/log_tran_table.c
```diff
@@ -1473,6 +1475,48 @@ logtb_free_tran_mvcc_info (LOG_TDES * tdes)
...
+
+  /* isolation level */
+  param = session_get_session_parameter (thread_p, PRM_ID_LOG_ISOLATION_LEVEL);
+  if (param)
+    {
+      tdes->isolation = (TRAN_ISOLATION) param->value.i;
+    }
+
+  /* lock timeout */
+  param = session_get_session_parameter (thread_p, PRM_ID_LK_TIMEOUT);
```
wrlawodms:
```suggestion
  param = session_get_session_parameter (thread_p, PRM_ID_LK_TIMEOUT);
```
It seems that we can just use `prm_get_integer_value ();`, which calls `session_get_session_parameter` if needed and seems more general.

wrlawodms: And the following conversion code can be removed because `prm_`interface seems to do the conversion internally using `prm_msec_to_sec()`. Please check it.

src/transaction/log_tran_table.c
```diff
@@ -1588,6 +1632,10 @@ logtb_clear_tdes (THREAD_ENTRY * thread_p, LOG_TDES * tdes)
   LSA_SET_NULL (&tdes->rcv.atomic_sysop_start_lsa);
   LSA_SET_NULL (&tdes->rcv.analysis_last_aborted_sysop_lsa);
   LSA_SET_NULL (&tdes->rcv.analysis_last_aborted_sysop_start_lsa);
+
+#if defined (SERVER_MODE)
```
wrlawodms: Why does this only on SERVER_MODE? It doesn't matter with `csql -S` (Standalone mode)?

src/transaction/log_tran_table.c
wrlawodms: If we use `prm_` interface, it'd be better to remove this wrapping function, `logtb_set_session_tdes`. See the other usage of the parameters. They just set these like the below.:
```
 549   tran_isolation = (TRAN_ISOLATION) prm_get_integer_value (PRM_ID_LOG_ISOLATION_LEVEL);
 550   tran_lock_wait_msecs = prm_get_integer_value (PRM_ID_LK_TIMEOUT_SECS);
 ```

**Review by wrlawodms — COMMENTED** [review]
I'm glad to see you again :). Let me have your opinion on my review.

File: src/transaction/log_tran_table.c (hunk 1 above)
Comment (hyahong) [reply src/transaction/log_tran_table.c]: I understood that the `prm_` interface like `prm_get_integer_value ()` concerns the `prm_Def` which is global to the server.

`set system parameters` (in csql) is applied for each session, and in order to fix this issue, the transaction must reflect the session after the commit occurs.

So I used `session_get_session_parameter ()` to get a value of session parameter `PRM_ID_LK_TIMEOUT` and `PRM_ID_LOG_ISOLATION_LEVEL` from the specific session.

Comment (hyahong) [reply src/transaction/log_tran_table.c]: > And the following conversion code can be removed because `prm_`interface seems to do the conversion internally using `prm_msec_to_sec()`. Please check it.

If `wait_msecs` is LK_ZERO_WAIT (0), LK_INFINITE_WAIT (-1), LK_FORCE_ZERO_WAIT (-2), it should not be changed. 
Then, even if i replace `tdes->wait_msecs = wait_msecs * 1000` with `prm_sec_to_msec ()`, the condition (in wrapping function) seems to be necessary.

Also, the `prm_sec_to_msec ()` function is declared static and cannot be called from outside. So I think it's okay to change the value as `* 1000` conversion.

What do you think about this?

**Comment by hyahong** [comment]
Thank you. I'm glad to be able to contribute again, too!

File: src/transaction/log_tran_table.c (hunk 2 above)
Comment (hyahong) [reply src/transaction/log_tran_table.c]: Oh... It's a mistake. I hadn't thought of SA_MODE. There were no problem with the execution, but i could see that the changes were not reflected in the transaction.

I will modify the code to make transaction reflect the value of `prm_Def` using `prm_get_integer_value ()`

File: src/transaction/log_tran_table.c (hunk 1 above)
Comment (wrlawodms) [reply src/transaction/log_tran_table.c]: > `set system parameters` (in csql) is applied for each session, and in order to fix this issue, the transaction must reflect the session after the commit occurs.

@Yechan0815 in `prm_get_integer_value()` -> `prm_get_value()`, it calls `session_get_session_parameter()` if needed. Didn't it work for this case?

Comment (wrlawodms) [reply src/transaction/log_tran_table.c]: > > And the following conversion code can be removed because `prm_`interface seems to do the conversion internally using `prm_msec_to_sec()`. Please check it.
> 
> If `wait_msecs` is LK_ZERO_WAIT (0), LK_INFINITE_WAIT (-1), LK_FORCE_ZERO_WAIT (-2), it should not be changed. Then, even if i replace `tdes->wait_msecs = wait_msecs * 1000` with `prm_sec_to_msec ()`, the condition (in wrapping function) seems to be necessary.
> 
> Also, the `prm_sec_to_msec ()` function is declared static and cannot be called from outside. So I think it's okay to change the value as `* 1000` conversion.
> 
> What do you think about this?

I didn't mean to use `prm_sec_to_msec()` or `prm_msec_to_sec()` directly, but I expected they will be called internally in `prm_` interface. However, I was wrong and the conversion seems needed in this case! Sorry for confusing you.

File: src/transaction/log_tran_table.c [hunk 3]
```diff
@@ -1473,6 +1475,48 @@ logtb_free_tran_mvcc_info (LOG_TDES * tdes)
...
+	}
+      else if (wait_msecs == LK_ZERO_WAIT || wait_msecs == LK_INFINITE_WAIT || wait_msecs == LK_FORCE_ZERO_WAIT)
+	{
+	  tdes->wait_msecs = wait_msecs;
+	}
+      else
+	{
+	  assert_release (0);
+	}
+    }
```
wrlawodms:
```suggestion
  if (param)
    {
      tdes->wait_msecs = param->value.i;
      if (tdes->wait_msecs > 0)
	{
	 tdes->wait_msecs = tdes->wait_msecs * 1000;
	}
    }
```

wrlawodms: or 
```suggestion
  if (param)
    {
       wait_msecs = param->value.i;
       tdes->wait_msecs = wait_msecs > 0 ? wait_msecs * 1000 : wait_msecs; 
    }
```

wrlawodms: The `PRM_ID_LK_TIMEOUT` can't be under -1, so `assert_release (0);` doesn't seem needed. The suggestions above are just suggestions.

src/transaction/log_tran_table.c
hyahong: > > `set system parameters` (in csql) is applied for each session, and in order to fix this issue, the transaction must reflect the session after the commit occurs.
> 
> @Yechan0815 in `prm_get_integer_value()` -> `prm_get_value()`, it calls `session_get_session_parameter()` if needed. Didn't it work for this case?

You're right! I fixed the code.
Only `SERVER_MODE` refers to the session value, in other cases `prm_Def` (and `prm_` interface) is referenced. `SA_MODE` is also resolved this way.

src/transaction/log_tran_table.c
hyahong: I changed param condition (NULL check) to assert because I thought the session paramter `PRM_ID_LOG_ISOLATION_LEVEL`, `PRM_ID_LK_TIMEOUT` should be unconditionally valid.

It's right. `assert_release (0)` seems unnecessary. I will remove `assert_release (0)`.

src/transaction/log_tran_table.c
wrlawodms: Please re-request the review if you're done fxing.

hyahong: I'm sorry. I'll remember it.

src/transaction/log_tran_table.c
```diff
@@ -1473,6 +1475,46 @@ logtb_free_tran_mvcc_info (LOG_TDES * tdes)
...
+
+#if defined (SERVER_MODE)
+  session_parameter = session_get_session_parameter (thread_p, PRM_ID_LOG_ISOLATION_LEVEL);
+  assert (session_parameter != NULL);
+  isolation = (TRAN_ISOLATION) session_parameter->value.i;
+
+  session_parameter = session_get_session_parameter (thread_p, PRM_ID_LK_TIMEOUT);
+  assert (session_parameter != NULL);
+  wait_msecs = session_parameter->value.i;
+#else /* SERVER_MODE */
```
wrlawodms: The same question before: `prm_get_integer_value()` calls `session_get_session_parameter()` internally if needed (if on SERVER_MODE and it's sessions variable). Doesn't it?
```suggestion
```

wrlawodms: in `prm_get_value()`:

```
11123   if (PRM_SERVER_SESSION (prm_id) && BO_IS_SERVER_RESTARTED ())
11124     {
11125       SESSION_PARAM *sprm;
11126       thread_p = thread_get_thread_entry_info ();
11127       sprm = session_get_session_parameter (thread_p, prm_id);
11128       if (sprm)
11129         {
11130           return &(sprm->value);
11131         }
11132     }
```

hyahong: I misunderstood because `prm_get_integer_value ()` referenced `prm_Def`. But it's all handled internally. Sorry.
The code becomes as simple as below.

```
/* isolation level */
tdes->isolation = (TRAN_ISOLATION) prm_get_integer_value (PRM_ID_LOG_ISOLATION_LEVEL);

/* lock timeout */
tdes->wait_msecs = prm_get_integer_value (PRM_ID_LK_TIMEOUT);
if (tdes->wait_msecs > 0)
{
  tdes->wait_msecs = tdes->wait_msecs * 1000;
}
```

I'm thinking about the wrapping function `logtb_set_session_tdes ()` which is detached, because it reflects the session.
Is function `logtb_set_session_tdes ()` necessary? What do you think.

wrlawodms: Oh, you're already mentioning the next thing I wanted to talk about! Yes, I don't think `logtb_set_session_tdes` is needed and it's not even an only-session-related change. And the code snippet looks simple and good.

hyahong: It's good! I'll fix it now.

wrlawodms (approved):
This solution is based on the assumption that a session is mapped to a LOG_TDES permanently (1:1 map). Otherwise, a session will make a transaction in which isolation level and wait_msecs are initialized from another session. Please check it. I think @joohok will help.

joohok:
> This solution is based on the assumption that a session is mapped to a LOG_TDES permanently (1:1 map). Otherwise, a session will make a transaction in which isolation level and wait_msecs are initialized from another session. Please check it. I think @joohok will help.

@Yechan0815 has completed the analysis on the matter @wrlawodms described above. Thanks for you opinion

src/transaction/log_tran_table.c
```diff
@@ -1588,6 +1588,14 @@ logtb_clear_tdes (THREAD_ENTRY * thread_p, LOG_TDES * tdes)
   LSA_SET_NULL (&tdes->rcv.atomic_sysop_start_lsa);
   LSA_SET_NULL (&tdes->rcv.analysis_last_aborted_sysop_lsa);
   LSA_SET_NULL (&tdes->rcv.analysis_last_aborted_sysop_start_lsa);
+
+  /* reflect the session (or prm_Def) */
+  tdes->isolation = (TRAN_ISOLATION) prm_get_integer_value (PRM_ID_LOG_ISOLATION_LEVEL);
+  tdes->wait_msecs = prm_get_integer_value (PRM_ID_LK_TIMEOUT);
```
joohok, line 1594: how about using `PRM_ID_LK_TIMEOUT_SECS` instead of `PRM_ID_LK_TIMEOUT` ? 

In other codes, only `PRM_ID_LK_TIMEOUT_SECS` is used. 
When the `PRM_ID_LK_TIMEOUT` parameter is set by the user, the value is converted into `second` units. This value is also used for `PRM_ID_LK_TIMEOUT_SECS`. If you use this parameter, user can clearly understand that the unit of the timeout value is `second`.

joohok (approved):

joohok:
Please, merge the latest develop branch to solve the ci test failure.

src/transaction/log_tran_table.c
wrlawodms: @joohok In my opinion, `PRM_ID_LK_TIMEOUT ` is better now because `PRM_ID_LK_TIMEOUT_SECS` is deprecated.

wrlawodms: Or, is `PRM_DEPRECATED` just to prevent users from using it, not in the code?

hyahong:
I merged it. Thank you!

src/transaction/log_tran_table.c
joohok: > Or, is `PRM_DEPRECATED` just to prevent users from using it, not in the code?

Maybe this is right. Because  only the `PRM_ID_LK_TIMEOUT_SECS` is used in the code when the lock timeout is set in client initialization step (boot_initialize_client ()) , but this parameter is not opened to the user.

hornetmj (approved):

src/transaction/log_tran_table.c
```diff
@@ -1588,6 +1588,14 @@ logtb_clear_tdes (THREAD_ENTRY * thread_p, LOG_TDES * tdes)
...
+  if (tdes->wait_msecs > 0)
+    {
+      tdes->wait_msecs = tdes->wait_msecs * 1000;
```
hornetmj, line 1597: @Yechan0815 The manual say that 'the max value of lock_timeout is INT_MAX'. But there is no upper limit of this system parameter. And the type of tdes->wait_msecs is INT. If user set the lock_timeout=INT_MAX, then the behavior of lock_timeout work unexpectedly. Do you consider it?

hyahong:
I definitely checked, and the test went well. However, there were two issue below.... 

1. Since `;set` is not committed, it is reflected at the next transaction. (even in SA mode)
2. `prm_get_integer_value ()` gets the value of `PRM_ID_LOG_ISOLATION_LEVEL`, `PRM_ID_LK_TIMEOUT` from prm_Def, not session parameter because of its static flag.

src/transaction/log_tran_table.c
Yechan0815: @hornetmj 

Unfortunately, it was not considered.....

While processing this issue, I didn't see any exception handling related to `set transaction`, `set system parameters`. Therefore, it was not considered extra.
But, both the 'set transaction' and the modified 'set system parameters' have an error.

hyahong: @hornetmj 

If the value is not reflected on the server when `tdes->wait_msecs * 1000` exceeds INT_MAX, the server and client will not be in sync state.

To solve this, we need to complete the error handling part. Currently, error handling of `NET_SERVER_PRM_SET_PARAMETERS` is not used, only NO_ERROR is sent to the client.

Also, this should apply to both `set system parameters` and `set transaction`.

So I think this range considering is out of this issue.
What do you think?

hyahong:
> I definitely checked, and the test went well. However, there were two issue below....
> 
> 1. Since `;set` is not committed, it is reflected at the next transaction. (even in SA mode)
> 2. `prm_get_integer_value ()` gets the value of `PRM_ID_LOG_ISOLATION_LEVEL`, `PRM_ID_LK_TIMEOUT` from prm_Def, not session parameter because of its static flag.

Solved all of the above !

src/transaction/log_tran_table.c
hornetmj: @Yechan0815 I think there is other choice for it. The lower or upper bound check is executed on each client or server side when the value is set. And you make that user can set the max value of the lock_timeout to the INT_MAX/1000(but, I think it is not a good choice). One of choices is that you save the the value of the lock_timeout as seconds. And the value will be converted to the msec when used.

There may exists other options for you. The purpose of this issue is that the user feel that the change of some parameter works well including lower and upper limitation or printing out the correct error. If you think you need to error handling before this is done, then you just share your idea through PR or jira issue or Teams(internal) and make an issue for it. And it will block this issue.

joohok:
> I definitely checked, and the test went well. However, there were two issue below....
> 
> 1. Since `;set` is not committed, it is reflected at the next transaction. (even in SA mode)

I don't think we need to solve the 1 above . `;set ` can be called in the middle of active transaction, and it seems weird for me that transaction timeout and isolation can be changed while transaction is active. 
For example, Isolation level for transaction A was repeatable read. And if isolation level is changed to read committed by `;set` command, users knows that isolation level of A is read committed but A acts as repeatable read internally.

hyahong:
@joohok 

It's a tricky issue...
if isolation level is changed by `;set` and user enters the query, the transaction will not work with the changed values until commit occurs once, even user know isolation level changed.

wrlawodms:
> if isolation level is changed by ;set and user enters the query, the transaction will not work with the changed values until commit occurs once, even user know isolation level changed.

I've heard that this is intended behavior. The changed value will be applied from the next transaction. Isn't it? Or, was there another design discussion?
 For your information, Postgresql doesn't allow to change the isolation level in the middle of a transaction. I don't know it in mysql, tell me if you know.

hyahong:
@joohok @wrlawodms 

Oh, I understood. Sorry for my confused.

src/transaction/log_tran_table.c
hyahong: @hornetmj 

It took me a while to check some.

For system parameters, the server does not handle an exception separately after the client handles an exception for the value. So, it seems like needed to set upper of the value.
Also, a condition should be added to the `check_timeout_value()` function that checks the range of values at `set transaction timeout`.

Then, there is not much to modify, I think it can be fixed in this issue.

what do you think about this?

joohok (approved):

src/transaction/log_tran_table.c
wrlawodms: The default unit of `lock_timeout` is "ms", and it seems to be converted to in "sec" unit  by `prm_msec_to_sec()` internally. It means that users can set it to "INT_MAX/1000 seconds" at most and in turn, this conversion causes no overflow issue. `lock_timeout_secs` may have this issue, but this is deprecated. I'm not sure how it works if a user sets it to "INT_MAXs" or "INT_MAXm", though.

hornetmj: @Yechan0815 Please go a head. And please check the feedback from @wrlawodms.

src/transaction/log_tran_table.c
```diff
@@ -1588,6 +1588,14 @@ logtb_clear_tdes (THREAD_ENTRY * thread_p, LOG_TDES * tdes)
   LSA_SET_NULL (&tdes->rcv.atomic_sysop_start_lsa);
   LSA_SET_NULL (&tdes->rcv.analysis_last_aborted_sysop_lsa);
   LSA_SET_NULL (&tdes->rcv.analysis_last_aborted_sysop_start_lsa);
+
+  /* reflect the session (or prm_Def) */
+  tdes->isolation = (TRAN_ISOLATION) prm_get_integer_value (PRM_ID_LOG_ISOLATION_LEVEL);
```
hornetmj, line 1593: Does the following case work well?
```
# in the session-1: csql
change isolation_level = 6 # serializable
execute transaction and commit

# in the session-2: csql
change isolation_level = 4 # read committed
execute transaction and commit

# in the session-1: csql
execute transaction # this transaction work serializable but I worry about it work read committed
```

src/base/system_parameter.c
```diff
@@ -2849,7 +2849,7 @@ static SYSPRM_PARAM prm_Def[] = {
    (DUP_PRM_FUNC) NULL},
   {PRM_ID_LOG_ISOLATION_LEVEL,
    PRM_NAME_LOG_ISOLATION_LEVEL,
-   (PRM_FOR_CLIENT | PRM_USER_CHANGE | PRM_FOR_SESSION),
+   (PRM_FOR_CLIENT | PRM_FOR_SERVER | PRM_USER_CHANGE | PRM_FOR_SESSION),
```
hornetmj, line 2853: As I know PRM_FOR_SERVER means that cub_server follows the value in the cubrid.conf. If dba set the isolation_level=6(serializable) in the cubrid.conf and start server, then cub_server is configured the isolation_level to the 6 instead of 4(default, read committed). Currently this property is only used on the client side when I check the code. And the manual also said it is a client parameter. Please let me know the reason.

src/transaction/log_tran_table.c
Yechan0815: @wrlawodms 

'lock_timeout_secs' seems to be deprecated.

I couldn't find the part where `PRM_ID_LK_TIMEOUT` changes the value of `PRM_ID_LK_TIMEOUT_SECS`.
`set_dup` (`prm_msec_to_sec ( ... )`) and `get_dup` (`prm_sec_to_msec ( ... )`) of `PRM_ID_LK_TIMEOUT` were also not called when sysprm set/get occurs.
And when I check, the value of `PRM_ID_LK_TIMEOUT_SECS` is not actually reflected and it was still -1 (INFINITE).

Also, `PRM_ID_LK_TIMEOUT` is stored in `sec` even though `PRM_ID_LK_TIMEOUT_SECS` exists.

src/base/system_parameter.c
hyahong: @hornetmj 

When changing the value with `set system parameter`, server determined with macro `PRM_IS_FOR_SESSION` whether to change the value of `prm_Def` or `session parameter`.
At this time, the true condition of macro `PRM_IS_FOR_SESSION` is `PRM_FOR_SESSION` flag.

And `prm_get_value ()` called by `prm_get_integer_value ()` determined with macro `PRM_SERVER_SESSION` whether the value is read from `prm_Def` or `session parameter`.
At this time, the true condition of macro `PRM_SERVER_SESSION` is `PRM_FOR_SESSION & PRM_IS_FOR_SERVER & !PRM_CLIENT_SESSION_ONLY` flag.

However, `PRM_ID_LOG_ISOLATION_LEVEL` and `PRM_ID_LK_TIMEOUT` do not have `PRM_IS_FOR_SERVER` flag, So these will get the value from `prm_Def`, not sesson.

I added `PRM_FOR_SERVER` flag to solve this.



To solve this problem without adding the flag, i think that need to make an exception in the conditional statement below.

```
if ((PRM_SERVER_SESSION (prm_id) || (something...)) && BO_IS_SERVER_RESTARTED())
{
        SESSION_PARAM *sprm;
        thread_p = thread_get_thread_entry_info();
        sprm = session_get_session_parameter(thread_p, prm_id);
        If (sp.m)
        {
                return &(sprm->value);
        }

       .....
}
```

src/transaction/log_tran_table.c
hyahong: @hornetmj 

It seems to be working fine.

session-1 (csql)
```
csql> set system parameters 'isolation_level=6';
csql> ;autocommit off
csql> select * from tbl; ------- (1)

csql> select * from tbl; ------- (3)

csql> commit;
csql> select * from tbl;
```
session-2 (csql)
```
csql> insert into tbl values (20); ------- (2)
```

I used the above test.

(1) and (3) show the same result without `20`. And i can see a value `20` inserted by session 2 after a commit.

src/transaction/log_tran_table.c
hyahong: @wrlawodms 

If a value over INT_MAX is input, overflow occurs and the value is stored as it is.
Therefore, I think it is better to set the upper limit about it.

wrlawodms: When I set `lock_timeout` to INT_MAX + 1 (2147483647+1), `prm_msec_to_sec()` is called and 2147484 is stored. it works on the CS and Standalone mode, and even when the parameter is set by the config file. The following is the call stack on the CS mode.

```
#0  prm_msec_to_sec (out_val=0x7fffffff6018, out_type=PRM_INTEGER, in_val=0x7fffffff6010, in_type=PRM_BIGINT) at /home/cubrid/cubrid/dev/src/base/system_parameter.c:7493
#1  0x00007ffff5c95e2f in prm_check_range (prm=0x7ffff6c7bed0 <prm_Def+2288>, value=0x7fffffff6098) at /home/cubrid/cubrid/dev/src/base/system_parameter.c:9131
#2  0x00007ffff5c9654c in sysprm_generate_new_value (prm=0x7ffff6c7bed0 <prm_Def+2288>, value=0x7fffffff61ad "400000", check=true, new_value=0x4e817f8)
    at /home/cubrid/cubrid/dev/src/base/system_parameter.c:9435
#3  0x00007ffff5c92f37 in sysprm_validate_change_parameters (data=0x7fffffff6a30 "lock_timeout=400000", check=true, assignments_ptr=0x7fffffff69f0)
    at /home/cubrid/cubrid/dev/src/base/system_parameter.c:7897
#4  0x00007ffff5bc2669 in db_set_system_parameters (data=0x7fffffff6a30 "lock_timeout=400000") at /home/cubrid/cubrid/dev/src/compat/db_admin.c:2718
#5  0x00007ffff5b7ce7e in csql_set_sys_param (arg_str=0x4e95005 "lock_timeout=400000") at /home/cubrid/cubrid/dev/src/executables/csql.c:2348
```

hyahong: @wrlawodms 

I made a mistake in the description. Sorry for the confusion.

When `lock_timeout` is set to `2147484000`, the value is stored as `2147484`. This value does not exceed INT_MAX now, but when it is multiplied by 1000 to apply to `tdes` later, overflow occurs and the overflowed value is applied.

So an upper limit was needed.

(Do you know why is `set system parameters lock_timeout` inputted in msec and stored in sec, and `set transaction` inputed in sec and stored in msec? This is just a curiosity.)

wrlawodms: Oh, I misunderstood the issue. Thank you for elaborating.

> (Do you know why is set system parameters lock_timeout inputted in msec and stored in sec, and set transaction inputed in sec and stored in msec? This is just a curiosity.)

I don't know exactly. I guess the developer wanted to share the same variable with `lock_timeout_in_secs`, which is deprecated now and in the second unit. Maybe for the compatibility.

wrlawodms (approved):

src/base/system_parameter.c
hyahong: @hornetmj 

In the manual book `시작하세요! 큐브리드`, I found that `lock_timeout` and `isolation_level` can be set by `cubrid.conf`. Isn't it valid method now? (the first edition of this book is dated 2015.)

hornetmj: @Yechan0815 

> @hornetmj
> 
> In the manual book `시작하세요! 큐브리드`, I found that `lock_timeout` and `isolation_level` can be set by `cubrid.conf`. Isn't it valid method now? (the first edition of this book is dated 2015.)

It is true. The client or server or client/server system paramters can be set in the cubrid.conf.

hornetmj: From my point of view, the following two currently feel strange:

1. For the same session parameter(e.g. lock_timeout), why should setter and getter decide access to prm_Def or session parameter through different conditions?
    - Wouldn't it be the right approach to modify the setter and getter to use the same condition when accessing the session parameter?
2. The lock_timeout and isolation_level have no effect on server (overall) operation. It only has a limited effect on a particular session. The manual describes the server parameter as follows.
    - A server parameter affects the behaviors of the server processes.

I understand that your solution solves the reported problem. However, for the above reasons, I would like to hear more from you and the reviewers. (@joohok, @wrlawodms)

hyahong: @hornetmj 

1.
It feels strange to me that the conditions of getter and setter are different. But I can't explain this part because I haven't figured out the designer's intentions yet...

2.
Both the `tdes` initial value and the `sessoin` initial value use the value sent by the client at boot. The client sends `-1 (infinite)` and `4 (read commited)` as shown below.
```c
tran_isolation = TRAN_DEFAULT_ISOLATION_LEVEL ();
tran_lock_wait_msecs = TRAN_LOCK_INFINITE_WAIT;
```
In fact, even if `prm_Def` is set by `cubrid.conf`, it does not effect on  `tdes` or `session`.

wrlawodms: @hornetmj 

> The lock_timeout and isolation_level have no effect on server (overall) operation. It only has a limited effect on a particular session. The manual describes the server parameter as follows.
A server parameter affects the behaviors of the server processes.

By the definition of `PRM_SERVER_SESSION()` macro and its usage in `prm_get_value()`, I guess that these kinds of parameters **have different values with each session** and are **used in the server process**. I translate "A server parameter affects the behaviors of the server processes." as "The parameter is used by server processes". In this respect, I think this makes enough sense.

Not `PRM_SERVER_SESSION()`, but `PRM_FOR_SESSION` examples:
`optimization_level`, `csql_history_num`, `csql_auto_commit`, and `multi_range_optimization_limit`.

These are only used on the client process like CAS, Csql, or JSP. On the other hand `PRM_SERVER_SESSION()` parameters are used also on the server process. The examples are `create_table_reuseoid`, `timezone`, and `alter_table_change_type_strict`.

However, there are some exceptions to my guess such as `no_backslash_escapes`, which seems to be used only on the client process but has `PRM_FOR_SERVER`. I think this flag setting is wrong, though.

This is just my opinion based on the code and the usage. What do you think?

wrlawodms: > For the same session parameter(e.g. lock_timeout), why should setter and getter decide access to prm_Def or session parameter through different conditions?

Regarding this, I don't know exactly what you mean by the setter's condition. Could you elaborate a little more?

wrlawodms: @Yechan0815 
> In fact, even if prm_Def is set by cubrid.conf, it does not effect on tdes or session.

If the `lock_timeout` is set by cubrid.conf, they are applied when a session is connected in the `boot_restart_client()` and work well. Aren't they? Or did I misunderstand what you mean?

wrlawodms: @Yechan0815 On top of that, I have a question. If we set PRM_FOR_SERVER, so make it use the session variable, the `lock_timeout` on cubrid.conf doesn't work? It's about all server session variables. They can't set by cubrid.conf?

hyahong: @wrlawodms 

The book states that cubrid.conf affects the 'server'. But it works fine for csql.
(It was also mentioned above. Sorry for the confusion.)

hyahong: > @Yechan0815 On top of that, I have a question. If we set PRM_FOR_SERVER, so make it use the session variable, the `lock_timeout` on cubrid.conf doesn't work? It's about all server session variables. They can't set by cubrid.conf?

It works fine. I'm sorry.

wrlawodms: @Yechan0815 no need to sorry. I just wonder if the server session parameter can't be set by cubrid.conf. Have you tried and confirmed it works?

hyahong: @wrlawodms 

> > For the same session parameter(e.g. lock_timeout), why should setter and getter decide access to prm_Def or session parameter through different conditions?
> 
> Regarding this, I don't know exactly what you mean by the setter's condition. Could you elaborate a little more?

It seems that getter is `prm_get_value ()` and setter is `sysprm_set_value ()`.

The condition that the getter `prm_get_value ()` accesses the session is `PRM_SERVER_SESSION ()`, and the condition that the setter `sysprm_set_value ()` accesses the session is `PRM_IS_FOR_SESSION ()`.
The conditions for getter and setter are different as above.

hyahong: @wrlawodms 

> @Yechan0815 no need to sorry. I just wonder if the server session parameter can't be set by cubrid.conf. Have you tried and confirmed it works?

Yes it works fine.

hornetmj: > @hornetmj
> 
> > The lock_timeout and isolation_level have no effect on server (overall) operation. It only has a limited effect on a particular session. The manual describes the server parameter as follows.
> > A server parameter affects the behaviors of the server processes.
> 
> By the definition of `PRM_SERVER_SESSION()` macro and its usage in `prm_get_value()`, I guess that these kinds of parameters **have different values with each session** and are **used in the server process**. I translate "A server parameter affects the behaviors of the server processes." as "The parameter is used by server processes". In this respect, I think this makes enough sense.
> 
> Not `PRM_SERVER_SESSION()`, but `PRM_FOR_SESSION` examples: `optimization_level`, `csql_history_num`, `csql_auto_commit`, and `multi_range_optimization_limit`.
> 
> These are only used on the client process like CAS, Csql, or JSP. On the other hand `PRM_SERVER_SESSION()` parameters are used also on the server process. The examples are `create_table_reuseoid`, `timezone`, and `alter_table_change_type_strict`.
> 
> However, there are some exceptions to my guess such as `no_backslash_escapes`, which seems to be used only on the client process but has `PRM_FOR_SERVER`. I think this flag setting is wrong, though.
> 
> This is just my opinion based on the code and the usage. What do you think?

I agree with you and @joohok(off-line) feedback in aspects of the implemenation of this module. However, I still have doubts about the question posed above. I approve of this resolution given the examples you provided and in terms of addressing the problem.

hornetmj (approved):

src/base/system_parameter.c
wrlawodms: It also looks weird to me that the condition for setting and getting is different. @Yechan0815 Please hold the issue a little bit more.

wrlawodms: @Yechan0815 I've taken a look at the system parameter source code to know a reasonable rule of its flags, but there seems to be some inconsistent use and behavior to build a consistent rule, at least to me. To get a precise concept of that, I think we should use more time and effort to analyze it. Regarding this issue, I suppose It would be better just to resolve it here and analyze it when a problem occurs, or it could be the following job after merging this issue. Anyway, I approved merging this PR.

hyahong: @wrlawodms 
I keep reading, but it's difficult. Thank you!

PR #3700, opened 2022-07-04
[LETS-406] log_global.m_prior_sender must only be initialized if/when needed

adrianbrab:
http://jira.cubrid.org/browse/LETS-406

`log_global.m_prior_sender` is curently a value member.
This patch transforms it to a pointer - similar to `m_prior_recver` to be only initialized when/if needed.
Also it is changed to be only under SERVER_MODE.

### DIFF
--- a/src/base/server_type.cpp
+++ b/src/base/server_type.cpp
@@ -137,6 +137,7 @@ int init_server_type (const char *db_name)
       if (is_active_transaction_server ())
 	{
 	  ts_Gl.reset (new active_tran_server ());
+	  log_Gl.initialize_log_prior_sender ();
 	}
       else if (is_passive_transaction_server ())
 	{
@@ -158,6 +159,7 @@ int init_server_type (const char *db_name)
     {
       // page server needs a log prior receiver
       log_Gl.initialize_log_prior_receiver ();
+      log_Gl.initialize_log_prior_sender ();
     }
   else
     {
@@ -203,12 +205,18 @@ void finalize_server_type ()
 	  assert (pts_Gl != nullptr);
 	  pts_Gl = nullptr;
 	}
+
       ts_Gl->disconnect_page_server ();
+      if (is_active_transaction_server ())
+	{
+	  log_Gl.finalize_log_prior_sender ();
+	}
       ts_Gl.reset (nullptr);
     }
   else if (get_server_type () == SERVER_TYPE_PAGE)
     {
       ps_Gl.disconnect_all_tran_server ();
+      log_Gl.finalize_log_prior_sender ();
     }
   else
     {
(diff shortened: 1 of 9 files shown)


src/server/active_tran_server.cpp
```diff
@@ -72,7 +72,11 @@ active_tran_server::on_boot ()
 	  std::bind (&active_tran_server::push_request, std::ref (*this), tran_to_page_request::SEND_LOG_PRIOR_LIST,
 		     std::placeholders::_1);
 
-  log_Gl.m_prior_sender.add_sink (m_prior_sender_sink_hook_func);
+  if (log_Gl.m_prior_sender == nullptr)
```
cristiarg: Add functions to initialize/finalize must be added to `log_global` (similar to `log_global::..._log_prior_receiver ()`) and these need to be called properly:
- log prior sender must be initialized for ATS (to dispatch transactional log prior lists towards PS - `active_tran_server::on_boot`)
- and PS (to forward transactional log lists towards PTS's - see usage in `page_server::connection_handler::prior_sender_sink_hook ` and `log_pack_log_boot_info`)
Do the initialization/finalization in `server_type.cpp`.
Also, fill in with finalize of prior sender wherever needed similar to prior receiver (see `xboot_shutdown_server`).

src/base/server_type.cpp
```diff
@@ -158,6 +159,7 @@ int init_server_type (const char *db_name)
     {
       // page server needs a log prior receiver
       log_Gl.initialize_log_prior_receiver ();
+      log_Gl.initialize_log_prior_sender ();
```
cristiarg, line 161: it would be better to have the sender be initialized before the receiver; conceptually - although in practice it might not make a difference as it depends when is a sink actually registered with the sender - it is better to have the sender already be available when the receiver starts receiving & dispatching

src/base/server_type.cpp
```diff
@@ -203,12 +205,18 @@ void finalize_server_type ()
 	  assert (pts_Gl != nullptr);
 	  pts_Gl = nullptr;
 	}
+
       ts_Gl->disconnect_page_server ();
+      if (is_active_transaction_server ())
```
cristiarg, line 208: finalize log prior sender before  disconnecting the page server; conceptually, after sending receiving the disconnect message/response (`tran_to_page_request::SEND_DISCONNECT_MSG`) no other message can be sent from a transaction server to the page server

src/transaction/log_global.c
```diff
@@ -131,12 +130,24 @@ log_global::initialize_log_prior_receiver ()
...
 void
 log_global::finalize_log_prior_receiver ()
 {
   m_prior_recver.reset (nullptr);
 }
 
+void
+log_global::finalize_log_prior_sender ()
+{
+  m_prior_sender.reset (nullptr);
```
cristiarg, line 149: `prior_sender` must have a dtor which asserts that all sinks have been removed prior to destruction.
This will actually tell us if we have all un-init order right. On the happy path it will work. In edge cases, we'll see ... :)

cristiarg:
Superseded by #4537 .

PR #3705, opened 2022-07-08
[CBRD-24420] dblink: correct oracle syntax error in rewritten query.

beyondykk9:
http://jira.cubrid.org/browse/CBRD-24420

The rewritten query by "predicate push down" at DBLINK would not be executed at ORACLE.
We should modify the rewritten routine not to ban the grammar of ORACLE.

```
-- create table at remote DB (oracle)
create table tbl1 (a int, b varchar, c int);

-- make a record of the table
insert into tbl1 values (1, 'abc', 2);

-- query to dblink
select *
from dblink (srv, 'select a, b, c from tbl1')
where a > 0;
```

The error arise as like "ORA-00933: SQL command not properly ended"

### DIFF
--- a/src/parser/parse_tree_cl.c
+++ b/src/parser/parse_tree_cl.c
@@ -1246,7 +1246,7 @@ parser_copy_term_dblink (PARSER_CONTEXT * parser, PT_NODE * tree)
 {
   if (tree)
     {
-      tree = parser_walk_tree (parser, tree, copy_node_in_term_dblink, NULL, copy_node_in_tree_post, NULL);
+      tree = parser_walk_tree (parser, tree, copy_node_in_term_dblink, NULL, NULL, NULL);
     }
 
   return tree;
--- a/src/parser/parse_tree_cl.c
+++ b/src/parser/parse_tree_cl.c
@@ -139,6 +139,7 @@ static PT_NODE *pt_lambda_check_reduce_eq (PARSER_CONTEXT * parser, PT_NODE * tr
 					   int *continue_walk);
 static PT_NODE *pt_lambda_node (PARSER_CONTEXT * parser, PT_NODE * tree_or_name, void *void_arg, int *continue_walk);
 static PT_NODE *pt_find_id_node (PARSER_CONTEXT * parser, PT_NODE * tree, void *void_arg, int *continue_walk);
+static PT_NODE *copy_node_in_term_dblink (PARSER_CONTEXT * parser, PT_NODE * old_node, void *arg, int *continue_walk);
 static PT_NODE *copy_node_in_tree_pre (PARSER_CONTEXT * parser, PT_NODE * old_node, void *arg, int *continue_walk);
 static PT_NODE *copy_node_in_tree_post (PARSER_CONTEXT * parser, PT_NODE * new_node, void *arg, int *continue_walk);
 static PT_NODE *free_node_in_tree_pre (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int *continue_walk);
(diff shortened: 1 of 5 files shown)


src/parser/view_transform.c
```diff
@@ -3779,16 +3779,17 @@ pt_copypush_terms (PARSER_CONTEXT * parser, PT_NODE * spec, PT_NODE * query, PT_
       /* copy terms */
       query->info.dblink_table.pushed_pred = parser_copy_tree_list (parser, term_list);
```
hgryoo, line 3808: how about copying the exact node (inside of the CAST expression node) instead of defining a custom print.

beyondykk9: The tree traverse and copy exact node seems complexed some. So, I keep the code as is.

src/parser/parse_tree_cl.c
```diff
@@ -755,17 +755,33 @@ pt_find_id_node (PARSER_CONTEXT * parser, PT_NODE * tree, void *void_arg, int *c
...
+  if (new_node == NULL)
+    {
+      new_node = parser_new_node (parser, old_node->node_type);
+      if (new_node == NULL)
+	{
+	  PT_INTERNAL_ERROR (parser, "allocate new node");
+	  return NULL;
+	}
+      *new_node = *old_node;
+    }
```
hgryoo: `copy_node_in_tree_pre()` is used many places so it seems to affect a lot.

hgryoo: How about just copying the `term_list` and then rewriting the `term_list` sub-tree by the private walk function which eliminates the wrapped CAST by the query compiler?

src/parser/parse_tree_cl.c
```diff
@@ -1191,6 +1234,24 @@ parser_copy_tree (PARSER_CONTEXT * parser, const PT_NODE * tree)
...
+ *   return:
+ *   parser(in):
+ *   tree(in):
+ */
+PT_NODE *
+parser_copy_term_dblink (PARSER_CONTEXT * parser, PT_NODE * tree)
+{
+  if (tree)
+    {
+      tree = parser_walk_tree (parser, tree, copy_node_in_term_dblink, NULL, copy_node_in_tree_post, NULL);
```
hgryoo: `copy_node_in_tree_post()` is only for PT_SPEC. is it necessary?

src/parser/view_transform.c
beyondykk9: Ok, I will try it.

src/parser/parse_tree_cl.c
beyondykk9: ok, I removed it.

src/parser/parse_tree_cl.c
beyondykk9: I will try it.

src/parser/parse_tree_cl.c
```diff
@@ -743,6 +744,48 @@ pt_find_id_node (PARSER_CONTEXT * parser, PT_NODE * tree, void *void_arg, int *c
...
+ * copy_node_in_term_dblink () - copies exactly a term node excluding cast node for dblink,
+     and returns a pointer to the copy. It is eligible for a walk "pre" function
+ *   return:
+ *   parser(in):
+ *   old_node(in):
+ *   arg(in):
+ *   continue_walk(in):
+ */
+static PT_NODE *
+copy_node_in_term_dblink (PARSER_CONTEXT * parser, PT_NODE * old_node, void *arg, int *continue_walk)
```
hgryoo: suggestion of method name: pt_copy_term_for_dblink

the name like `copy_node_...` functions are for copying all kinds of nodes of tree structures.

src/parser/view_transform.c
```diff
@@ -3777,7 +3777,7 @@ pt_copypush_terms (PARSER_CONTEXT * parser, PT_NODE * spec, PT_NODE * query, PT_
       break;
     case PT_DBLINK_TABLE:
       /* copy terms */
-      query->info.dblink_table.pushed_pred = parser_copy_tree_list (parser, term_list);
+      query->info.dblink_table.pushed_pred = parser_copy_term_dblink (parser, term_list);
```
hgryoo: Wrapping by `parser_copy_term_dblink()` seems redundant.
`query->info.dblink_table.pushed_pred = parser_walk_tree (parser, term_list, copy_node_in_term_dblink, NULL, NULL, NULL);`

beyondykk9: Ok, good suggestion, thanks.

src/parser/parse_tree_cl.c
beyondykk9: Ok, good point. thanks

hgryoo (approved):

ctshim (approved):

youngjinj (approved):

src/parser/view_transform.c
```diff
@@ -3725,6 +3726,48 @@ pt_check_pushable_term (PARSER_CONTEXT * parser, PT_NODE * term, FIND_ID_INFO *
...
+    {
+      if (old_node->info.expr.op == PT_CAST && PT_EXPR_INFO_IS_FLAGED (old_node, PT_EXPR_INFO_CAST_WRAP))
+	{
+	  new_node = parser_new_node (parser, old_node->info.expr.arg1->node_type);
+	  if (new_node == NULL)
+	    {
+	      PT_INTERNAL_ERROR (parser, "allocate new node");
+	      return NULL;
+	    }
+	  *new_node = *old_node->info.expr.arg1;
```
shparkcubrid: nvl(col1) in cast( nvl(col1) as int) cannot be copied. It is recommended to make a full copy and remove cast().

shparkcubrid: old_node (cast () ) is needed to free.

beyondykk9: please show the full query as an incorrect rewrite result.
the below query is executed well:
`select  * from dblink(srv, 'select a, b, c from tbl') as t(col1 int, col2 varchar, col3 int) where cast( nvl(col1,1) as int) = 9999;`

the rewritten query is as below:
`SELECT * FROM (select a, b, c from tbl) cublink(col1, col2, col3) WHERE  cast(nvl(col1, 1) as integer)=9999
`

hyunikn (approved):

src/parser/view_transform.c
beyondykk9: I added the routin "pt_remove_cast_wrap_for_dblink"

shparkcubrid (approved):

PR #3719, opened 2022-07-14
[CBRD-24426] The cub_admin is dead abnomally when executing compactdb utility in the SA-mode

hornetmj:
http://jira.cubrid.org/browse/CBRD-24426

### Purpose
* fix coredump

### Implementation
* N/A

### Remarks
* N/A

### DIFF
--- a/src/executables/compactdb.c
+++ b/src/executables/compactdb.c
@@ -63,7 +63,7 @@ static void process_object (THREAD_ENTRY * thread_p, DESC_OBJ * desc_obj, OID * 
 static int process_set (THREAD_ENTRY * thread_p, DB_SET * set);
 static int process_value (THREAD_ENTRY * thread_p, DB_VALUE * value);
 static DB_OBJECT *is_class (OID * obj_oid, OID * class_oid);
-static int disk_update_instance (MOP classop, DESC_OBJ * obj, OID * oid);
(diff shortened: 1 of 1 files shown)


src/executables/compactdb.c
```diff
@@ -700,7 +700,7 @@ disk_update_instance (MOP classop, DESC_OBJ * obj, OID * oid)
 
   heap_create_update_context (&update_context, hfid, oid, WS_OID (classop), Diskrec, NULL,
 			      UPDATE_INPLACE_CURRENT_MVCCID);
-  if (heap_update_logical (NULL, &update_context) != NO_ERROR)
+  if (heap_update_logical (thread_get_thread_entry_info (), &update_context) != NO_ERROR)
```
joohok: I was worried that the thread entry could be fetched in an uninitialized state, but I saw that db_restart() was performed in compactdb. Since it was confirmed that the thread entry is initialized at the time of db_restart(), there seems to be no problem.

And maybe it can get `thread_p` from the caller (process_object()), but this is trivial one.

joohok (approved):

src/executables/compactdb.c
hornetmj: @joohok I applied your feedback. Thanks.

PR #3720, opened 2022-07-15
[LETS-486] fix unit tests after merge from develop

cristiarg:
http://jira.cubrid.org/browse/LETS-486

unit tests were overlooked during the actual merge window from develop


razvanradulescu (approved):

cristiarg:
This change is incorrect therefore, no longer needed.

PR #3725, opened 2022-07-18
[CBRD-24428] dblink: correct error message for repeated query execution

beyondykk9:
http://jira.cubrid.org/browse/CBRD-24428

The error message is not the same as prior query execution's one at error case.

```
-- remote_t is a table of remote server as created as below
CREATE TABLE remote_t (
    col1 numeric(10, 5)
);
```

```
-- First
SELECT *
FROM dblink (dblink-url or server,
             'SELECT col1 FROM remote_t') AS t(col1 time);

-- Second
SELECT *
FROM dblink (dblink-url or server,
             'SELECT col1 FROM remote_t') AS t(col1 time);
 
```

Expected Result:

```
-- First Result
ERROR: Cannot coerce value of domain "numeric" to domain "time".

-- Second Result
ERROR: Cannot coerce value of domain "numeric" to domain "time".
```
 

Actual Result:

```
-- First Result
ERROR: Cannot coerce value of domain "numeric" to domain "time".

-- Second Result
ERROR: Cannot coerce value of domain "numeric" to domain "*NULL*".
```

### DIFF
--- a/src/query/dblink_scan.c
+++ b/src/query/dblink_scan.c
@@ -843,6 +843,7 @@ dblink_scan_next (DBLINK_SCAN_INFO * scan_info, val_list_node * val_list)
       else
 	{
 	  TP_DOMAIN dom;
+	  TP_DOMAIN_STATUS status;
 
 	  tp_domain_init (&dom, (DB_TYPE) db_value_domain_type (valptrp->val));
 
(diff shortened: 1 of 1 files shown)


hyunikn (approved):

shparkcubrid (approved):

youngjinj (approved):

src/query/dblink_scan.c
```diff
@@ -853,6 +853,11 @@ dblink_scan_next (DBLINK_SCAN_INFO * scan_info, val_list_node * val_list)
 
 	  if (db_value_coerce (&cci_value, valptrp->val, &dom) != DOMAIN_COMPATIBLE)
 	    {
+	      /*
+	         in repeated error case, in order to display correct error message,
+	         we need to reinitialize the domain of valptrp->val
+	       */
+	      db_value_domain_init (valptrp->val, (DB_TYPE) dom.type->id, dom.precision, dom.scale);
```
hgryoo: To preserve domain of `valptrp->val`, you can use `tp_value_cast_preserve_domain ()` instead of `db_value_coerce()`. By the way, in many places using `db_value_coerce()`, the destination value is temp or new allocated value.

beyondykk9: The function **tp_value_cast_preserve_domain** does not set the error. To preserve the domain it needs to add a function to set a error. I chose the way not to modify serveral routines or add a new routine.

hgryoo: the error can be set by `tp_domain_status_er_set()`. object domain is already called above in this code. (`tp_domain_init()`).

hgryoo: Also, for `VARCHAR` type, `codeset` and `collation` is not restored too by using the saved `TP_DOMAIN dom` above.

beyondykk9: However, **tp_value_cast_preserve_domain** does not call the tp_value_cast_internal with "**TP_FORCE_COERCION**".
whereas **db_value_coerce** calls the **tp_value_cast_internal** with "**TP_FORCE_COERCION**".

hgryoo (approved):

src/query/dblink_scan.c
beyondykk9: Ok, I fixed it with **tp_value_cast_preserved_domain**

PR #3733, opened 2022-07-19
[CBRD-24429] Not allowed multiple foreign keys on a column attribute

beyondykk9:
http://jira.cubrid.org/browse/CBRD-24429

The foreign should be defined only one on a key column, however, the query parser allow the multiple foreign keys on the same column attributes.

```
CREATE TABLE a_tbl (
        id INT NOT NULL DEFAULT 0 PRIMARY KEY,
        phone VARCHAR(10)
);

CREATE TABLE b_tbl (
        id INT NOT NULL, name VARCHAR (10) NOT NULL,
        CONSTRAINT pk_id PRIMARY KEY (id),
        CONSTRAINT fk_id FOREIGN KEY (id) REFERENCES a_tbl (id)
);
```
```
alter table b_tbl add CONSTRAINT fk_id_2 FOREIGN KEY (id) REFERENCES a_tbl (id) ON DELETE CASCADE ON UPDATE RESTRICT;
alter table b_tbl add CONSTRAINT fk_id_3 FOREIGN KEY (id) REFERENCES a_tbl (id);
alter table b_tbl add CONSTRAINT fk_id_4 FOREIGN KEY (id) REFERENCES a_tbl (id) on delete no action;
```
The above "ALTER TABLE" statments should be failed.


beyondykk9:
sql fail cases from CI is because of multiple foreign key.

ctshim:
Is there no need to modify the compatibility check matrix in the comment above the classobj_check_index_comability() function?

src/object/class_object.c
```diff
@@ -7986,15 +8013,6 @@ classobj_check_index_compatibility (SM_CLASS_CONSTRAINT * constraints, const DB_
 	  ret = SM_CREATE_NEW_INDEX;
 	  return ret;
```
ctshim:
```suggestion
	  return SM_CREATE_NEW_INDEX;
```

src/object/class_object.c
```diff
@@ -7986,15 +8013,6 @@ classobj_check_index_compatibility (SM_CLASS_CONSTRAINT * constraints, const DB_
...
-	  ret = SM_SHARE_INDEX;
-	  if (existing_con->filter_predicate != NULL || existing_con->func_index_info != NULL)
-	    {
-	      ret = SM_CREATE_NEW_INDEX;
-	    }
-	  return ret;
-	}
     }
 
   if (constraint_type == DB_CONSTRAINT_INDEX && existing_con->type == SM_CONSTRAINT_FOREIGN_KEY)
```
ctshim, line 8018:
```suggestion
else  if (constraint_type == DB_CONSTRAINT_INDEX && existing_con->type == SM_CONSTRAINT_FOREIGN_KEY)
```

ctshim:
```suggestion
  else if (constraint_type == DB_CONSTRAINT_INDEX && existing_con->type == SM_CONSTRAINT_FOREIGN_KEY)
```

ctshim (approved):

src/object/class_object.c
beyondykk9: Ok, good point.

src/object/class_object.c
beyondykk9: Ok, good

src/object/class_object.c
```diff
@@ -7986,15 +8013,6 @@ classobj_check_index_compatibility (SM_CLASS_CONSTRAINT * constraints, const DB_
...
 	}
-      else if (SM_IS_SHARE_WITH_FOREIGN_KEY (existing_con->type))
-	{
-	  ret = SM_SHARE_INDEX;
-	  if (existing_con->filter_predicate != NULL || existing_con->func_index_info != NULL)
-	    {
-	      ret = SM_CREATE_NEW_INDEX;
-	    }
-	  return ret;
-	}
```
hgryoo: Is it safe to remove this routine? `SM_IS_SHARE_WITH_FOREIGN_KEY` checks not only `SM_CONSTRAINT_FOREIGN_KEY` but `SM_CONSTRAINT_INDEX`. `SM_CONSTRAINT_INDEX` is added for partition constraints and I found some SQL regressions related to partitions.

src/object/class_object.c
```diff
@@ -4069,6 +4069,33 @@ classobj_find_constraint_by_attrs (SM_CLASS_CONSTRAINT * cons_list, DB_CONSTRAIN
   const char **namep;
   int i, len, order;
 
+  /* for foreign key, need to check redundancy first */
+  if (new_cons == SM_CONSTRAINT_FOREIGN_KEY)
+    {
+      for (cons = cons_list; cons; cons = cons->next)
+	{
+	  /* check foreign key only */
+	  if (new_cons != cons->type || !cons->attributes || !namep)
```
hyunikn: (1) namep can be read without being initialized on this line.
     Or, maybe you mean attr_names, not namep.
(2) new_cons is SM_CONSTRAINT_FOREIGN_KEY on this line.
     Using constants can improve performance, though minor in this case.

src/object/class_object.c
```diff
@@ -4069,6 +4069,33 @@ classobj_find_constraint_by_attrs (SM_CLASS_CONSTRAINT * cons_list, DB_CONSTRAIN
   const char **namep;
   int i, len, order;
 
+  /* for foreign key, need to check redundancy first */
+  if (new_cons == SM_CONSTRAINT_FOREIGN_KEY)
+    {
+      for (cons = cons_list; cons; cons = cons->next)
```
youngjinj, line 4075: I suggest another way.

If SM_CONSTRAINT_PRIMARY_KEY is found in classobj_find_constraint_by_attrs() function and it is not returned immediately, SM_CONSTRAINT_FOREIGN_KEY can be found.
> I wonder why if SM_CONSTRAINT_PRIMARY_KEY is found in classobj_find_constraint_by_attrs() function, it returns immediately without looking for other constraints.

If the existing_con returned by the classobj_find_constraint_by_name() function is SM_CONSTRAINT_FOREIGN_KEY, the SM_CONSTRAINT_FOREIGN_KEY can be checked in the classobj_check_index_compatibility() function.

If SM_CONSTRAINT_FOREIGN_KEY already exists, SM_SHARE_INDEX is returned instead of SM_CREATE_NEW_INDEX.

So, how about changing it like below?
1. In the classobj_find_constraint_by_attrs() function, check all constraints on the column with the same name.
2. The classobj_check_index_compatibility() function returns SM_NOT_SHARE_INDEX_AND_WARNING instead of SM_SHARE_INDEX if SM_CONSTRAINT_FOREIGN_KEY already exists.
3. If SM_NOT_SHARE_INDEX_AND_WARNING is returned after calling classobj_check_index_compatibility() function, an error will occur.

src/object/class_object.c
beyondykk: In this routine, existing_con->type is always SM_CONSTRAINT_FOREIGN_KEY because the classobj_find_constraint_by_attrs is changed to return constraint SM_CONSTRAINT_FOREIGN_KEY for DB_CONSTRAINT_FOREIGN_KEY.

src/object/class_object.c
beyondykk9: Good point, thanks.

src/object/class_object.c
beyondykk9: Ok, I confused.

hyunikn (approved):

src/object/class_object.c
hyunikn, line 8021: Line 7942: the table cell for FK-FK should be updated from shared to error.

src/object/class_object.c
beyondykk: I'm gonna change the routine as below:
`else if (existing_con->type == SM_CONSTRAINT_INDEX)
  {
    AS_IS
  }`

beyondykk: It looks better to modify of **classobj_find_constraint_by_attrs** to return NULL when it can't find existing foreign key.
And this classobj_check_index_exist checks just only existing index, so I think that it's good to return SM_CREATE_INDEX at exising_con = NULL in foreign key.

src/object/class_object.c
beyondykk9: Ok, good point, thanks.

src/object/class_object.c
beyondykk9: the routine is changed as below:
```
     else if (existing_con->type == SM_CONSTRAINT_INDEX)
	{
            return NEW or SHARE as compatibility specification.
        }
```

src/object/class_object.c
```diff
@@ -4069,6 +4069,33 @@ classobj_find_constraint_by_attrs (SM_CLASS_CONSTRAINT * cons_list, DB_CONSTRAIN
   const char **namep;
   int i, len, order;
 
+  /* for foreign key, need to check redundancy first */
+  if (new_cons == DB_CONSTRAINT_FOREIGN_KEY)
+    {
+      for (cons = cons_list; cons; cons = cons->next)
+	{
+	  /* check foreign key only */
+	  if (!cons->attributes || cons->type != SM_CONSTRAINT_FOREIGN_KEY)
```
youngjinj, line 4078: How about checking only cons->type?

youngjinj (approved):

src/object/class_object.c
```diff
@@ -7932,14 +7959,14 @@ classobj_make_descriptor (MOP class_mop, SM_CLASS * classobj, SM_COMPONENT * com
  * |               | /UK(asc) | /UK(desc) |         |          |  /R-Idx   |
  * |               |          |   /R-UK   |         |          |           |
  * +---+-----------+----------+-----------+---------+----------+-----------+
- * |   |   PK(asc) |  share   |  new idx  |  error  |  error   |  new idx  |
+ * |   |   PK(asc) |  error   |  new idx  |  error  |  error   |  new idx  |
```
shparkcubrid: trivial : PK(desc) / UK(desc) <== > PK(desc) / UK(desc) is 'share'. it seems to be rule for PK and UK.

shparkcubrid (approved):

hgryoo (approved):

src/object/class_object.c
beyondykk9: I'd like to keep the code because **cons->attributes** is checked first.

PR #3737, opened 2022-07-20
[LETS-484] start pts with more than one data volume

cristiarg:
http://jira.cubrid.org/browse/LETS-484

On transaction server with remote storage, there is a separate bookkeeping of the number of volumes (which are actually created remotely, on the page server) that must be maintained.
On such a server, the check for volume id validity is done against this separate bookkeeping - `disk_is_valid_volid`.
This patch only maintains the separate bookkeeping for active transaction server. For passive transaction server, a separate patch will do the update of the value upon replicating the `RVDK` log records.

### DIFF
--- a/src/storage/disk_manager.c
+++ b/src/storage/disk_manager.c
@@ -206,6 +206,8 @@ struct disk_cache
 
 static DISK_CACHE *disk_Cache = NULL;
 static DKNVOLS disk_Page_server_perm_volume_count = 0;	// used for transaction server with remote storage
+							// must be maintained during replication on
+							// passive transaction server (with remote storage)
 
 static DKNSECTS disk_Temp_max_sects = -2;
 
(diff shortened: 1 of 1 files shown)


src/storage/disk_manager.c
```diff
@@ -1316,6 +1318,15 @@ disk_rv_undo_format (THREAD_ENTRY * thread_p, const LOG_RCV * rcv)
 	  disk_Cache->perm_purpose_info.extend_info.volid_extend = NULL_VOLID;
 	}
 
+      if (is_tran_server_with_remote_storage ())
```
razvanradulescu, line 1321: Nit : there are common patterns for increase/decrease number of permanent volumes.
it maybe extracted to one or two common functions.
Please consider after passive implementation.

razvanradulescu (approved):

PR #3738, opened 2022-07-21
[CBRD-24425] Add overflow page information when showing the index capacity

ctshim:
http://jira.cubrid.org/browse/CBRD-24425

When displaying Index capacity information, additional information about the Overflow OID page is displayed.
In existing information, only information that used to represent the total number of pages has been changed. In other words, the total number of pages appears to include the number of Overflow pages.

Also, Change the property of the item "Total_value" in the SHOW INDEX CAPACITY syntax from int to bigint.

The effect of this fix affects the following commands or tools:
* cubrid diagdb -d 4
* show index capacity of
* show all indexes capacity of

### DIFF
--- a/src/parser/show_meta.c
+++ b/src/parser/show_meta.c
@@ -444,9 +444,10 @@ metadata_of_index_capacity (SHOW_ONLY_ALL flag)
     {"Index_name", "varchar(256)"},
     {"Btid", "varchar(64)"},
     {"Num_distinct_key", "int"},
-    {"Total_value", "int"},
+    {"Total_value", "bigint"},
     {"Avg_num_value_per_key", "int"},
     {"Num_leaf_page", "int"},
+    {"Num_Ovfl_page", "bigint"},
     {"Num_non_leaf_page", "int"},
     {"Num_total_page", "int"},
     {"Height", "int"},
@@ -456,7 +457,11 @@ metadata_of_index_capacity (SHOW_ONLY_ALL flag)
     {"Total_used_space", "varchar(64)"},
     {"Total_free_space", "varchar(64)"},
     {"Avg_num_page_key", "int"},
-    {"Avg_page_free_space", "varchar(64)"}
+    {"Avg_page_free_space", "varchar(64)"},
+    {"Percentage_free_space_Ovfl_page", "int"},
+    {"Average_free_space_per_Ovfl_page", "int"},
+    {"Average_Ovfl_page_count_per_key", "int"},
+    {"Max_Ovfl_page_count_in_a_key", "bigint"}
   };
 
   static const SHOWSTMT_COLUMN_ORDERBY orderby[] = {
--- a/src/parser/show_meta.c
+++ b/src/parser/show_meta.c
@@ -447,7 +447,7 @@ metadata_of_index_capacity (SHOW_ONLY_ALL flag)
     {"Total_value", "bigint"},
     {"Avg_num_value_per_key", "int"},
     {"Num_leaf_page", "int"},
-    {"Num_Ovfl_page", "bigint"},
+    {"Num_Ovfl_page", "int"},
     {"Num_non_leaf_page", "int"},
     {"Num_total_page", "int"},
     {"Height", "int"},
@@ -461,7 +461,7 @@ metadata_of_index_capacity (SHOW_ONLY_ALL flag)
     {"Percentage_free_space_Ovfl_page", "int"},
     {"Average_free_space_per_Ovfl_page", "int"},
     {"Average_Ovfl_page_count_per_key", "int"},
-    {"Max_Ovfl_page_count_in_a_key", "bigint"}
+    {"Max_Ovfl_page_count_in_a_key", "int"}
   };
 
   static const SHOWSTMT_COLUMN_ORDERBY orderby[] = {
--- a/src/storage/btree.c
+++ b/src/storage/btree.c
@@ -8854,7 +8854,7 @@ btree_dump_capacity (THREAD_ENTRY * thread_p, FILE * fp, BTID * btid)
   fprintf (fp, "Total Page Count: %d\n", cpc.tot_pg_cnt + cpc.ovfl_oid_pg.tot_pg_cnt);
   fprintf (fp, "Leaf Page Count: %d\n", cpc.leaf_pg_cnt);
   fprintf (fp, "NonLeaf Page Count: %d\n", cpc.nleaf_pg_cnt);
-  fprintf (fp, "Overflow Page Count: %lld\n", cpc.ovfl_oid_pg.tot_pg_cnt);
+  fprintf (fp, "Overflow Page Count: %d\n", cpc.ovfl_oid_pg.tot_pg_cnt);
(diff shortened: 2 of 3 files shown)


src/storage/btree.c
```diff
@@ -8657,6 +8645,10 @@ btree_get_subtree_capacity (THREAD_ENTRY * thread_p, BTID_INT * btid, PAGE_PTR p
 	  ovfl_vpid = leaf_pnt.ovfl;
 	  if (!VPID_ISNULL (&ovfl_vpid))
 	    {			/* overflow pages exist */
+	      oid_cnt_ovfl = 0;
```
beyondykk9, line 8648: The varibles as below looks better to be defined at this point because the variables is not used outside.
```
      int free_space_ovfl = 0;
      int oid_cnt_ovfl = 0;
      int pg_cnt_per_key = 0;
```

src/storage/btree.c
```diff
@@ -8612,27 +8600,27 @@ btree_get_subtree_capacity (THREAD_ENTRY * thread_p, BTID_INT * btid, PAGE_PTR p
...
-      cpc->tot_space += DB_PAGESIZE;
-      cpc->tot_used_space += (DB_PAGESIZE - free_space);
-      cpc->avg_key_len = ((cpc->dis_key_cnt > 0) ? ((int) (cpc->sum_key_len / cpc->dis_key_cnt)) : 0);
-      cpc->avg_rec_len = ((cpc->dis_key_cnt > 0) ? ((int) (cpc->sum_rec_len / cpc->dis_key_cnt)) : 0);
-      cpc->avg_pg_key_cnt = ((cpc->leaf_pg_cnt > 0) ? ((int) (cpc->dis_key_cnt / cpc->leaf_pg_cnt)) : 0);
-      cpc->avg_pg_free_sp = ((cpc->tot_pg_cnt > 0) ? (cpc->tot_free_space / cpc->tot_pg_cnt) : 0);
     }
   else
     {				/* a leaf page */
+      int free_space_ovfl = 0;
```
beyondykk9, line 8617: The below definition of variables looks better to move down into "if clause"
```
      int free_space_ovfl = 0;
      int oid_cnt_ovfl = 0;
      int pg_cnt_per_key = 0;
```

beyondykk9 (approved):

hyunikn (approved):

src/storage/btree.c
ctshim: I will fix it.  thank you.

src/storage/btree.c
ctshim: OK, I will fix  it.

shparkcubrid (approved):

src/storage/btree.c
```diff
@@ -22293,10 +22319,13 @@ btree_scan_for_show_index_capacity (THREAD_ENTRY * thread_p, DB_VALUE ** out_val
...
   idx++;
 
+  db_make_bigint (out_values[idx], cpc.ovfl_oid_pg.tot_pg_cnt);
+  idx++;
+
   db_make_int (out_values[idx], cpc.nleaf_pg_cnt);
   idx++;
 
-  db_make_int (out_values[idx], cpc.tot_pg_cnt);
+  db_make_int (out_values[idx], cpc.tot_pg_cnt + cpc.ovfl_oid_pg.tot_pg_cnt);
```
youngjinj, line 22328: If "cpc.ovfl_oid_pg.tot_pg_cnt" is bigint then "cpc.tot_pg_cnt + cpc.ovfl_oid_pg.tot_pg_cnt" must also be bigint.

src/storage/btree.c
```diff
@@ -22284,7 +22310,7 @@ btree_scan_for_show_index_capacity (THREAD_ENTRY * thread_p, DB_VALUE ** out_val
   db_make_int (out_values[idx], cpc.dis_key_cnt);
   idx++;
 
-  db_make_int (out_values[idx], cpc.tot_val_cnt);
+  db_make_bigint (out_values[idx], cpc.tot_val_cnt);
```
youngjinj, line 22313: How about adding a comment for the column name above each out_values?

src/parser/show_meta.c
```diff
@@ -444,9 +444,10 @@ metadata_of_index_capacity (SHOW_ONLY_ALL flag)
     {"Index_name", "varchar(256)"},
     {"Btid", "varchar(64)"},
     {"Num_distinct_key", "int"},
-    {"Total_value", "int"},
+    {"Total_value", "bigint"},
     {"Avg_num_value_per_key", "int"},
     {"Num_leaf_page", "int"},
+    {"Num_Ovfl_page", "bigint"},
```
youngjinj: If the number of pages exceeds INT_MAX, it means that only the index is using 31TiB.
Wouldn't it suffice to use INT yet?

src/storage/btree.h
```diff
@@ -351,11 +351,22 @@ struct btree_checkscan
   BTREE_ISCAN_OID_LIST oid_list;	/* Data area to store OIDs */
 };				/* B+tree <key-oid> check scan structure */
 
+struct btree_ovfl_oid_capacity
+{
+  int max_pg_cnt_per_key;	/* Distinct key count with overflow oid page) */
+  int dis_key_cnt;		/* Distinct key count (with ovfl pages) */
```
youngjinj, line 357: An overflow page occurs when there are too many oids with the same key to store in one leaf node.

In overflow page, oids have the same key value. What does distinct key mean?

src/parser/show_meta.c
```diff
@@ -456,7 +457,11 @@ metadata_of_index_capacity (SHOW_ONLY_ALL flag)
     {"Total_used_space", "varchar(64)"},
     {"Total_free_space", "varchar(64)"},
     {"Avg_num_page_key", "int"},
-    {"Avg_page_free_space", "varchar(64)"}
+    {"Avg_page_free_space", "varchar(64)"},
+    {"Percentage_free_space_Ovfl_page", "int"},
+    {"Average_free_space_per_Ovfl_page", "int"},
+    {"Average_Ovfl_page_count_per_key", "int"},
+    {"Max_Ovfl_page_count_in_a_key", "bigint"}
```
youngjinj: Wouldn't it suffice to use INT yet?

hgryoo (approved):

src/storage/btree.c
ctshim: OK, I will add comments.

src/storage/btree.h
ctshim: Duplicate keys do not always have overflow oid pages.
This variable counts the number of distinct keys with overflow oid pages.

src/parser/show_meta.c
ctshim: That's a good point.
(If the page size is 4K, it will be 8TiB.)
I agree with you. I'll fix it.

src/parser/show_meta.c
ctshim: Thank you

src/storage/btree.c
ctshim: All variables are int.

youngjinj (approved):

PR #3746, opened 2022-07-26
[CBRD-24436] MAX / MIN performance enhancement

beyondykk9:
http://jira.cubrid.org/browse/CBRD-24436

In v9.3, in the case of min(index-column) and max(index-column) with no predicates, the minimum and maximum values of the index are quickly found and processed. However, in 11.2, the entire index is scanned.


src/storage/btree.c
```diff
@@ -1792,6 +1792,76 @@ static DISK_ISVALID btree_check_tree (THREAD_ENTRY * thread_p, const OID * class
...
+  bool dummy_clear_key;
+
+  *is_visible = false;
+
+  if (mvcc_snapshot == NULL)
+    {
+      /* Early out. */
+      *is_visible = true;
+      return ret;
+    }
```
ctshim: I think it would be better to replace it with an asert statement.

assert(mvcc_snapshot != NULL);

src/parser/xasl_generation.c
```diff
@@ -4179,7 +4179,7 @@ pt_to_aggregate_node (PARSER_CONTEXT * parser, PT_NODE * tree, void *arg, int *c
       aggregate_list->flag_agg_optimize = false;
       BTID_SET_NULL (&aggregate_list->btid);
       if (info->flag_agg_optimize
-	  && (aggregate_list->function == PT_COUNT_STAR || aggregate_list->function == PT_COUNT
+	  && (aggregate_list->function == PT_COUNT_STAR
```
ctshim, line 4182: PT_COUNT is mentioned in line 4188.
Is there a reason why you removed the PT_COUNT that was in the existing code from this line location?

src/parser/xasl_generation.c
```diff
@@ -4205,6 +4205,19 @@ pt_to_aggregate_node (PARSER_CONTEXT * parser, PT_NODE * tree, void *arg, int *c
 		  aggregate_list->flag_agg_optimize = true;
 		}
 	    }
+	  else
```
ctshim: Does PT_MIN, PT_MAX only enter the else clause? I'm asking you in connection with the PT_COUNT that I already asked you above.

src/parser/xasl_generation.c
beyondykk9: PT_COUNT is not considered and PT_COUNT will be handled in the future. So, I remove the PT_COUNT in this point.

src/parser/xasl_generation.c
beyondykk9: Ok, I'm gonna fix it as you commented.

src/storage/btree.c
beyondykk9: this routine is called with mvcc_snapshot (i. e. it's always true)

ctshim: That's why I think these codes are unnecessary.

src/parser/xasl_generation.c
ctshim: It would be better to treat PT_COUNT as annotation in line 4188.

if (aggregate_list->function == PT_COUNT_STAR /* || aggregate_list->function == PT_COUNT */)

beyondykk9: Ok, I'm gonna fix it.

src/storage/btree.c
beyondykk9: it looks so, I'm gonna fix it.

src/query/query_executor.c
```diff
@@ -23684,7 +23684,8 @@ qexec_evaluate_partition_aggregates (THREAD_ENTRY * thread_p, ACCESS_SPEC_TYPE *
 	  continue;
 	}
 
-      if (agg_ptr->function == PT_COUNT_STAR && *is_scan_needed)
+      if ((agg_ptr->function == PT_COUNT_STAR || agg_ptr->function == PT_MAX || agg_ptr->function == PT_MIN)
+	  && *is_scan_needed)
```
shparkcubrid: This code does not exist in version 9. What is it for?

src/query/query_executor.c
```diff
@@ -23807,7 +23808,8 @@ qexec_evaluate_aggregates_optimize (THREAD_ENTRY * thread_p, AGGREGATE_TYPE * ag
     {
       if (agg_ptr->flag_agg_optimize)
 	{
-	  if (agg_ptr->function == PT_COUNT_STAR && *is_scan_needed)
+	  if ((agg_ptr->function == PT_COUNT_STAR || agg_ptr->function == PT_MAX || agg_ptr->function == PT_MIN)
+	      && *is_scan_needed)
```
shparkcubrid: This code does not exist in version 9. What is it for?

src/storage/btree.c
```diff
@@ -1792,6 +1792,76 @@ static DISK_ISVALID btree_check_tree (THREAD_ENTRY * thread_p, const OID * class
...
+ *  btid(in): B-tree info.
+ *  pg_ptr(in):	Page pointer.
+ *  mvcc_snapshot(in): The MVCC snapshot.
+ *  slot_id(in) : Slot id to be looked for.
+ *  is_visible(out): True or False
+ *
+ *  return: error code if any error occurs.
+ */
+static int
+btree_is_key_visible (THREAD_ENTRY * thread_p, BTID_INT * btid, PAGE_PTR pg_ptr, MVCC_SNAPSHOT * mvcc_snapshot,
```
shparkcubrid, line 1811: The clear unit test case with Isolation Level is needed to validate this function.

src/object/schema_manager.c
```diff
@@ -5767,6 +5767,74 @@ sm_att_info (MOP classop, const char *name, int *idp, TP_DOMAIN ** domainp, int
...
+ * sm_find_key_index()
+ *   return: Pointer to B-tree ID variable
+ *           has "att_name" as first key attribute.
+ *   classop(in): class object
+ *   att_names(in):
+ *   btid(out):
+ */
+
+BTID *
+sm_find_key_index (MOP classop, char *att_name, BTID * btid)
```
shparkcubrid: Because of the num_atts of sm_find_index() , it seems that there is a parameter to check only a specific column. If it doesn't work as intended, wouldn't it be better to fix a bug in the existing sm_find_index() rather than adding sm_find_key_index()?

src/storage/btree.c
```diff
@@ -16424,8 +16493,6 @@ btree_find_min_or_max_key (THREAD_ENTRY * thread_p, BTID * btid, DB_VALUE * key,
 
   pgbuf_unfix_and_init (thread_p, root_page_ptr);
 
-  assert (tp_valid_indextype (TP_DOMAIN_TYPE (BTS->btid_int.key_type)));
```
shparkcubrid, line 16427: Why did you delete it?

src/storage/btree.c
```diff
@@ -16450,25 +16517,55 @@ btree_find_min_or_max_key (THREAD_ENTRY * thread_p, BTID * btid, DB_VALUE * key,
       goto exit_on_error;
     }
 
-  if (!BTREE_END_OF_SCAN (BTS))
+  if ((mvcc_snapshot = logtb_get_mvcc_snapshot (thread_p)) == NULL)
     {
-      assert (BTS->slot_id > 0);
-      if (spage_get_record (thread_p, BTS->C_page, BTS->slot_id, &rec, PEEK) != S_SUCCESS)
+      ret = er_errid ();
```
hgryoo, line 16522: logtb_get_mvcc_snapshot seems not to set error code. er_errid () returns NO_ERROR

src/query/query_executor.c
beyondykk9: I added the code to fix some sql fail cases, but I missed to revert it.

src/storage/btree.c
beyondykk9: This checks if the index is sing-key type, we should permmit multi-key index. So, I remove the check routine.

src/query/query_executor.c
beyondykk9: I added the code to fix some sql fail cases, but I missed to revert it.

src/object/schema_manager.c
beyondykk: The purpose of sm_find_index() is to find an index includes specific attribute which is given by function argument.
The routine finds an index having attribute as first and also second column. This is not a bug. It's just a purpose of the function. We can fix or refactor the function to add a new purpose to get an index having attribute as first column only for multi-key index but This function is called from many routine. I think it would be better to make a new function.

src/storage/btree.c
beyondykk9: will be handled at **exit_on_error** as like other error case.

src/object/schema_manager.c
beyondykk: There is posted a test case which is written by cci-driver on CBRD-24436 as like count(*). The code hase 3 tranaction group having 50 thread running , which are DML (insert/delete), MIN (select min), MAX (select MAX).

src/storage/btree.c
beyondykk9: There is posted a test case which is written by cci-driver on CBRD-24436 as like count(*). The code hase 3 tranaction group having 50 thread running , which are DML (insert/delete), MIN (select min), MAX (select MAX).

src/object/schema_manager.c
beyondykk9: Ok, I fixed it as you commented.

src/object/schema_manager.c
```diff
@@ -5861,20 +5867,21 @@ sm_find_index (MOP classop, char **att_names, int num_atts, bool unique_index_on
 	    }
 
 	  att2 = classobj_find_attribute (class_, att_names[i], 0);
-	  if (att2 == NULL || att1->id != att2->id)
+	  if (att2 && att1->id == att2->id)
```
hyunikn: The condition for break is negated. Is this OK?
The original logic seems roughly that it breaks if att1->id == att2->id for ALL i < num_atts, 
But the new one seems that it breaks if att1->id == att2->id for SOME i < num_atts.

beyondykk9: Yes, a new logic finds first index has key attribute name. In addition, the origin logic has not been reached.

ctshim (approved):

src/object/schema_manager.c
```diff
@@ -5861,20 +5867,21 @@ sm_find_index (MOP classop, char **att_names, int num_atts, bool unique_index_on
...
 	    {
+	      found = true;
 	      break;
 	    }
 	}
 
-      if ((i == num_atts) && con->attributes[i] == NULL)
+      if (found)
```
shparkcubrid: how about?
if ((i == num_atts) && con->attributes[i] == NULL)
==> if (i == num_atts)

hgryoo (approved):

hyunikn (approved):

src/object/schema_manager.c
beyondykk9: it looks better to write the logic "if (i == 0 || i == num_atts)" instead of using "found" variable.

shparkcubrid (approved):

youngjinj (approved):

PR #3751, opened 2022-07-28
[CBRD-24439] Change pre-condition to assert in pgbuf_is_temporary_volume()

hyahong:
http://jira.cubrid.org/browse/CBRD-24439

Purpose
N/A

Implementation
N/A

Remarks
N/A

### DIFF
--- a/src/storage/page_buffer.c
+++ b/src/storage/page_buffer.c
@@ -4920,7 +4920,7 @@ pgbuf_is_temporary_volume (VOLID volid)
     {
       return false;
     }
-  return (LOG_DBFIRST_VOLID <= volid && xdisk_get_purpose (NULL, volid) == DB_TEMPORARY_DATA_PURPOSE);
+  return xdisk_get_purpose (NULL, volid) == DB_TEMPORARY_DATA_PURPOSE;
 }
 
 /*


wrlawodms (approved):

hornetmj:
@Yechan0815 You need to always register @hornetmj, @joohok and @wrlawodms as reviewers.

src/storage/page_buffer.c
```diff
@@ -4920,7 +4920,8 @@ pgbuf_is_temporary_volume (VOLID volid)
     {
       return false;
     }
-  return (LOG_DBFIRST_VOLID <= volid && xdisk_get_purpose (NULL, volid) == DB_TEMPORARY_DATA_PURPOSE);
+  assert (LOG_DBFIRST_VOLID <= volid);
```
hornetmj: Considering the purpose of this function, it seems safe to remove assert.

hornetmj (approved):

hyahong:
@hornetmj 
I could not register the reviewers because I don't have permission to register.

src/storage/page_buffer.c
hyahong: It's corrected.

wrlawodms: The range of VOLID is not only positive. Auxiliary files have negative values and it can be given if a user(developer) uses this function wrongly. So, it seems to me that the assertion works as a pre-condition for this function.

hornetmj: @wrlawodms The condition 'LOG_DBFIRST_VOLID <= volid' is checked in the xdisk_get_purpose() function by if condition. Since the pgbuf_is_temporary_volume() function only needs to determine the purpose of temp regardless of whether the volid is positive or negative, it is considered safe to remove the assert.

wrlawodms: @hornetmj OK. Now, I agree. Thanks for the elaboration.

joohok (approved):

PR #3752, opened 2022-07-29
[CBRD-24440] Do not consider stop_at in log_recovery_redo()

wrlawodms:
http://jira.cubrid.org/browse/CBRD-24440


src/transaction/log_recovery.c
```diff
@@ -3126,7 +3122,7 @@ log_recovery_redo (THREAD_ENTRY * thread_p, const LOG_LSA * start_redolsa, const
       LSA_COPY (&log_lsa, &lsa);
       if (logpb_fetch_page (thread_p, &log_lsa, LOG_CS_FORCE_USE, log_pgptr) != NO_ERROR)
 	{
-	  if (end_redo_lsa != NULL && (LSA_ISNULL (end_redo_lsa) || LSA_GT (&lsa, end_redo_lsa)))
+	  if (end_redo_lsa != NULL && (LSA_ISNULL (end_redo_lsa) || LSA_GE (&lsa, end_redo_lsa)))
```
hornetmj: We can make simple. The end_redo_lsa != NULL condition can be removed because the end_redo_lsa variable never be NULL.

hornetmj: And the LSA_ISNULL (end_redo_lsa) condition can be also removed. I think that it cannot have the null value(NULL_LOG_PAGEID, NULL_LOG_OFFSET).

wrlawodms: @hornetmj I guess it's the spec of this function: end_redo_lsa can be NULL and the NULL value even though now no one is using this function with NULL value.

hornetmj: @wrlawodms Could you explain a little more in terms of spec? In my opinion, if the code is not likely to be used in the future, I would expect that removing it would make the code more readable.

wrlawodms: @hornetmj The way I see it, the spec. of `log_recovery_redo()` is:
"Redo all redo log records to either the `end_redo_lsa` if it's given or the end of the log". Now, we don't use NULL `end_redo_lsa` because `log_recovery_analysis` always set the `end_redo_lsa`, but it can be if it's changed. If the `log_recovery_redo()` is changed to not allow NULL `end_redo_lsa`, it adds a new restriction to others that `end_redo_lsa` has to be initialized.

wrlawodms: However, I also think you're making sense. Removing not occurring cases can improve readability. Between the two options, I preferred the one that has fewer changes. What do you think?

wrlawodms:
I have reverted checking `end_redo_lsa` as discussed in the Jira issue with @joohok.

joohok (approved):

src/transaction/log_recovery.c
hornetmj: @wrlawodms I also usually choose the one with less change. However, in this case, unnecessary end_redo_lsa checks are occurring multiple times, which reduces readability. I think it would be better to remove it.

wrlawodms: @hornetmj Done!

hornetmj: @wrlawodms Thanks and please add assert for it :D

hornetmj (approved):

PR #3754, opened 2022-08-02
[CBRD-24451] Remove libcas for the legacy Java SP

hgryoo:
http://jira.cubrid.org/browse/CBRD-24451

**Motivation**
The legacy Java SP prior to 11.2 needed the libcas to call CAS function API by the server-side JDBC. But It is redundant because Java SP routines of CAS side (related to query compile and OID functionalities) are moved inside of the cubridcs and cubrid libraries

**Implementation**
*  Remove #if blocks that is LIBCAS_FOR_JSP mode
* Remove linking the libcas from cubridcs and cubridsa
* Remove libcas build scripts (cas/CMakeLists.txt)


beyondykk9 (approved):

src/base/porting.h
```diff
@@ -43,13 +43,15 @@
 #endif
 
 #if defined (WINDOWS)
-#define IMPORT_VAR 	__declspec(dllimport)
-#define EXPORT_VAR 	__declspec(dllexport)
-#include <WinBase.h>
+#ifdef _EXPORTING
```
mhoh3963, line 46: This seem like to be modified at CBRD-21754.
please check it.

mhoh3963 (approved):

ctshim (approved):

src/base/porting.h
hgryoo: I'm going to merge CBRD-21754 first.

PR #3761, opened 2022-08-08
[CBRD-24457] Remove meaningless conditional macros in "src/storage", "src/transaction"

hyahong:
http://jira.cubrid.org/browse/CBRD-24457

Purpose
N/A

Implementation
N/A

Remarks
```
#if 0				/* reserved for future use */
  disk_repr_p->repr_reserved_1 = OR_GET_INT (rec_p + CATALOG_DISK_REPR_RESERVED_1_OFF);
#endif
```

```
#if 0				/* TODO: do not delete me */
      assert (false);
#endif
```

As above, there are some cases where it is not deleted by referring to the comments even if the code is covered with `#if 0` - `#endif`.


src/storage/btree.c
```diff
@@ -15539,12 +15464,6 @@ btree_find_next_index_record (THREAD_ENTRY * thread_p, BTREE_SCAN * bts)
   first_page = bts->C_page;	/* init */
 
   ret_val = btree_find_next_index_record_holding_current (thread_p, bts, NULL);
-#if 0				/* TODO - need to check return value */
```
hornetmj: Please leave it as before. It seems that we need to check the return value after analysis.

src/transaction/log_record.hpp
```diff
@@ -37,33 +37,19 @@ enum log_rectype
   /* In order of likely of appearance in the log */
   LOG_SMALLER_LOGREC_TYPE = 0,	/* A lower bound check */
 
-#if 0
```
hornetmj: Please leave this file as before. The 'Obsolete' and other comments on each member of enum are information for developers.

hornetmj (approved):

hyahong:
@hornetmj 
Two changes have been reverted to before the change.

src/storage/external_sort.c
```diff
@@ -1581,9 +1580,8 @@ sort_validate (char **vector, long size, SORT_CMP_FUNC * compare, void *comp_arg
   assert (size >= 1);
   assert (compare != NULL);
 
-#if 1				/* TODO - for QAF, do not delete me */
+  /* TODO - for QAF, do not delete me */
```
wrlawodms: If you're certain that no validation is needed and the return is enough, I think you should remove the validation code below too.

wrlawodms: Or, I think it'd be better to leave it as it is.

src/storage/page_buffer.c
```diff
@@ -10370,13 +10370,12 @@ pgbuf_check_page_ptype_internal (PAGE_PTR pgptr, PAGE_TYPE ptype, bool no_error)
       return false;
     }
 
-#if 1				/* TODO - do not delete me */
+/* TODO - do not delete me */
```
wrlawodms, line 10373: Does this have to be left?

src/transaction/lock_manager.c
```diff
@@ -8032,184 +8032,6 @@ lock_detect_local_deadlock (THREAD_ENTRY * thread_p)
 #endif /* !SERVER_MODE */
 }
 
-#if 0				/* NOT_USED */
-/*
- */
-
-/*
- * lk_global_deadlock_detection: RUN THE GLOBAL DEADLOCK DETECTION
```
wrlawodms: I think we should not remove it until we know what this function was for. This code could be a good example to refer to later. Of course, I don't really know what even the global deadlock here is, but that's why I think we can't remove these kinds of functions without analysis. For example, CUBRId now doesn't support sharding, but code for sharding must remain somewhere and we can refer to them later when recovering the feature. @hornetmj

src/transaction/log_manager.c
```diff
@@ -3275,16 +3233,6 @@ log_skip_logging_set_lsa (THREAD_ENTRY * thread_p, LOG_DATA_ADDR * addr)
 void
 log_skip_logging (THREAD_ENTRY * thread_p, LOG_DATA_ADDR * addr)
```
wrlawodms: The change makes `log_skip_logging` do nothing. Of course, now it's doing nothing, but to confirm this is not needed anymore and change it permanently, I think we should check unlogged pages doesn't generate warning anymore. Furthermore, if it's confirmed, I think we should remove this function itself and all code calling it together. @hornetmj

src/transaction/log_writer.h
```diff
@@ -199,10 +199,6 @@ extern bool logwr_force_shutdown (void);
 extern int logwr_copy_log_file (const char *db_name, const char *log_path, int mode, INT64 start_page_id);
 extern LOG_PHY_PAGEID logwr_to_physical_pageid (LOG_PAGEID logical_pageid);
 extern const char *logwr_log_ha_filestat_to_string (enum LOG_HA_FILESTAT val);
-#if 0
```
wrlawodms, line 202: Please refer to the definition of these functions and change it 
```if 0```
to
``` #ifdef UNSTABLE_TDE_FOR_REPLICATION_LOG```.

File: src/transaction/log_page_buffer.c [hunk 8]
```diff
@@ -7333,11 +7326,6 @@ logpb_checkpoint (THREAD_ENTRY * thread_p)
 
   LOG_CS_EXIT (thread_p);
 
-#if 0
-  /* have to sync log vol, data vol */
-  fileio_synchronize_all (thread_p, true /* include_log */ );
```
wrlawodms: I'm not sure if it has to be done or not though it's not done now. It could be a temporary fix..

src/transaction/log_page_buffer.c
```diff
@@ -5826,13 +5826,6 @@ logpb_archive_active_log (THREAD_ENTRY * thread_p)
   /* Flush the log header to reflect the archive */
   logpb_flush_header (thread_p);
 
-#if 0
-  if (prm_get_integer_value (PRM_ID_SUPPRESS_FSYNC) != 0)
-    {
-      fileio_synchronize (thread_p, log_Gl.append.vdes, FILEIO_SYNC_ONLY);
```
wrlawodms: I'm not sure if it has to be done or not though it's not done now. It could be a temporary fix..

src/storage/page_buffer.c
hyahong: For now, I have left a TODO comment. hmm

hyahong:
Conversation that did not receive a answer were reverted and others were corrected according to the comments.

src/storage/page_buffer.c
hornetmj: Good. Just remove un-necessary #if 1, #endif.

src/transaction/lock_manager.c
hornetmj: @wrlawodms Good point. Leave it according to the opinion of @wrlawodms.

src/transaction/log_manager.c
hornetmj: @wrlawodms There is no difference in the behavior of the log_skip_logging() function before and after the change. It would be nice to complete the fix here for the purpose of this issue. And the removal of the log_skip_logging() completely would be nice to handle other issue because the purpose of this issue is to remove #if 1 or #if 0 macro. Please give me any opinion if you have.

hornetmj (approved):

src/transaction/log_writer.h
wrlawodms: @Yechan0815 Please follow what I said. The declaration has to be in the same macro as its definitions.

src/transaction/log_manager.c
wrlawodms: @hornetmj The way I see it, removing the `#if 0` part means we are gonna remove `log_skip_logging()`, but I'm not sure about this. In addition, If we remove `log_skip_logging()` later, the macro will be removed along.

src/transaction/log_writer.h
hyahong: rf

@wrlawodms 
I'm not sure why, but the changes aren't showing up properly here...

wrlawodms: Oh, now I can see. Thank you.

src/transaction/log_manager.c
hornetmj: @wrlawodms OK.
@Yechan0815 Please leave a comment that we need to decide whether or not remove log_skip_logging_set_lsa() function at the http://jira.cubrid.org/browse/CBRD-24398.

hyahong: @hornetmj  I added!
Do I need to revert the deletion of macro in function `log_skip_logging ()`?

hornetmj: @Yechan0815 Yes, please.

hornetmj (approved):

wrlawodms (approved):

PR #3762, opened 2022-08-09
[CBRD-24456] Add NULL guards after allocating memory in "src/transaction/lock_manager.c".

hyahong:
http://jira.cubrid.org/browse/CBRD-24456

Purpose
In funcion `lock_initialize_tran_lock_table ()` (line 1071), there is no part to check that the memory allocation is done properly.
Add allocation check and error handling.

```
1069      for (j = 0; j < LOCK_TRAN_LOCAL_POOL_MAX_SIZE; j++)
1070	{
1071	  entry = (LK_ENTRY *) malloc (sizeof (LK_ENTRY));
                  /* need null guard */
1072	  lock_initialize_entry (entry);
```

Implementation
N/A

Remarks
N/A

### DIFF
--- a/src/transaction/lock_manager.c
+++ b/src/transaction/lock_manager.c
@@ -1053,9 +1053,7 @@ lock_initialize_tran_lock_table (void)
   lk_Gl.tran_lock_table = (LK_TRAN_LOCK *) malloc (SIZEOF_LK_TRAN_LOCK * lk_Gl.num_trans);
   if (lk_Gl.tran_lock_table == NULL)
     {
-      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1,
-	      (size_t) (SIZEOF_LK_TRAN_LOCK * lk_Gl.num_trans));
-      return ER_OUT_OF_VIRTUAL_MEMORY;
+      goto exit_on_error;
     }
 
   /* initialize all the entries of transaction lock table */
@@ -1069,6 +1067,10 @@ lock_initialize_tran_lock_table (void)
       for (j = 0; j < LOCK_TRAN_LOCAL_POOL_MAX_SIZE; j++)
 	{
 	  entry = (LK_ENTRY *) malloc (sizeof (LK_ENTRY));
+	  if (entry == NULL)
+	    {
+	      goto exit_on_error;
+	    }
 	  lock_initialize_entry (entry);
 	  entry->next = tran_lock->lk_entry_pool;
 	  tran_lock->lk_entry_pool = entry;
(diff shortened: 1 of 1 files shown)


joohok:
Please briefly state the `purpose` so that reviewers/developers can easily understand what the PR is for with it.

src/transaction/lock_manager.c
```diff
@@ -1077,6 +1079,33 @@ lock_initialize_tran_lock_table (void)
...
+      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1,
+	      (size_t) (SIZEOF_LK_TRAN_LOCK * lk_Gl.num_trans));
+
+      return ER_OUT_OF_VIRTUAL_MEMORY;
+    }
+
+  free_and_init (lk_Gl.tran_lock_table);
+  for (int k = 0; k <= i; k++)
+    {
+      tran_lock = &lk_Gl.tran_lock_table[k];
```
hornetmj: 1. lk_Gl.tran_lock_table is already de-alloced in the 1092 line
2. The same behavior is in the lock_finalize() function. How about making this behavior a separate function?

hyahong:
@joohok 

Added a brief! Thank you!

src/transaction/lock_manager.c
hyahong: @hornetmj 

I changed the overlapping part to a separate function `lock_finalize_tran_lock_table ()`. It definitely looks cleaner.

src/transaction/lock_manager.c
```diff
@@ -5696,6 +5674,43 @@ lock_initialize (void)
...
+  int i;
+
+  /* remove resources */
+  if (lk_Gl.tran_lock_table != NULL)
+    {
+      for (i = 0; i < lk_Gl.num_trans; i++)
+	{
+	  tran_lock = &lk_Gl.tran_lock_table[i];
+	  pthread_mutex_destroy (&tran_lock->hold_mutex);
+	  pthread_mutex_destroy (&tran_lock->non2pl_mutex);
```
wrlawodms, line 5703: It works only if all `lk_Gl.tran_lock_table` elements are initialized. Let's say `lk_Gl.num_trans` is 10 and `lock_finalize_tran_lock_table()` is called from `lock_initialize_tran_lock_table`() while initializing `lk_Gl.num_trans[5]`. Then, `tran_lock->hold_mutex` and `tran_lock->non2pl_mutex` of k_Gl.num_trans[6,7, ..] are not yet initialized but are tried to destroy.

wrlawodms (changes requested):

src/transaction/lock_manager.c
hyahong: @wrlawodms 
I considered it, but I missed it as I turned the structure into a separate function. It's corrected!

src/transaction/lock_manager.c
Yechan0815: Different from the current code, I also thought about increasing the value of `lk_Gl.num_trans` for each loop, but I stopped because I was afraid I would use a lot of macros `MAX_NTRANS` in the code.
what do you think about it?

src/transaction/lock_manager.c
hyahong: Unlike the current code of using index (as parameter), I also thought about increasing the value of `lk_Gl.num_trans` for each loop, but I quitted because I was afraid I would use a lot of macros `MAX_NTRANS` in the code.
what do you think about it?

hyahong: Eventually changed it. I used variable instead of macro.

wrlawodms (approved):

src/transaction/lock_manager.c
wrlawodms: Looks good to me! Could you tell me why do you use a logical variable, `num_trans`, not just use the macro, `MAX_NTRANS`?

hyahong: @wrlawodms 

The 'MAX_NTRANS' macro cannot be changed because it is used elsewhere. If for some reason you need to change the maximum number of transactions only within this function, you must do so one by one (of course not many in this function). So I used local variables.

wrlawodms: I mean the value of `int num_trans` is not changed through the function. Isn't it?

wrlawodms: > Looks good to me! Could you tell me why do you use a logical variable, `num_trans`, not just use the macro, `MAX_NTRANS`?

not logical.. sorry -> local

hyahong: > I mean the value of `int num_trans` is not changed through the function. Isn't it?

It is right! But I thought it would be a little more suitable from a maintenance standpoint when only those values change.

wrlawodms: Makes sense. Thx.

joohok (approved):
LGTM

hornetmj (approved):

PR #3763, opened 2022-08-09
[LETS-483] log recovery analysis process MVCC information in log records and, on PTS, complete mvccid's

cristiarg:
http://jira.cubrid.org/browse/LETS-483

During log recovery analysis, process mvcc information in log records:
- populate transaction descriptor mvccid from MVCC log records;
- complete mvcc when `LOG_COMPLETE` and `LOG_ABORT` log records are parsed - this is done only in the context of passive transaction server

Other:
- correct log reading in `log_rv_analysis_assigned_mvccid` with a missing advance and align

### DIFF
--- a/src/transaction/log_recovery_analysis.cpp
+++ b/src/transaction/log_recovery_analysis.cpp
@@ -562,11 +562,7 @@ log_rv_analysis_undo_redo (THREAD_ENTRY *thread_p, int tran_id, LOG_LSA *log_lsa
       assert (log_page_p != nullptr);
 
       // move read pointer past the log header which is actually read upper in the stack
-      _er_log_debug (ARG_FILE_LINE, "crsdbg: an_mvcc_complete 01 trid = %d log_lsa = %lld|%d\n",
-		     tran_id, LSA_AS_ARGS (log_lsa));
       LOG_READ_ADD_ALIGN (thread_p, sizeof (LOG_RECORD_HEADER), log_lsa, log_page_p);
-      _er_log_debug (ARG_FILE_LINE, "crsdbg: an_mvcc_complete 02 trid = %d log_lsa = %lld|%d\n",
-		     tran_id, LSA_AS_ARGS (log_lsa));
       switch (log_type)
 	{
 	case LOG_MVCC_UNDOREDO_DATA:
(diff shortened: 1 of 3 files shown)


src/transaction/log_recovery_analysis.cpp
```diff
@@ -554,6 +557,45 @@ log_rv_analysis_undo_redo (THREAD_ENTRY *thread_p, int tran_id, LOG_LSA *log_lsa
   if (is_mvcc)
     {
       tdes->last_mvcc_lsa = *log_lsa;
+
+      // assign transaction mvccid from log record to transaction descriptor
+      assert (log_page_p != nullptr);
+
+      // move read pointer past the log header which is actually read upper in the stack
+      LOG_READ_ADD_ALIGN (thread_p, sizeof (LOG_RECORD_HEADER), log_lsa, log_page_p);
+      switch (log_type)
```
razvanradulescu: Nit : Can this code be templated on log record data struct ?

src/transaction/log_recovery_analysis.cpp
```diff
@@ -1045,8 +1103,21 @@ log_rv_analysis_complete (THREAD_ENTRY *thread_p, int tran_id, LOG_LSA *log_lsa,
       // Transaction is completed.
       if (tran_index != NULL_TRAN_INDEX)
 	{
+	  // newer quick fix on top of older quick fix: mark the mvccid as completed
```
razvanradulescu: This piece of code (to complete the transaction) is identical to the one in line 1046.

src/transaction/log_recovery_analysis.cpp
```diff
@@ -554,6 +557,45 @@ log_rv_analysis_undo_redo (THREAD_ENTRY *thread_p, int tran_id, LOG_LSA *log_lsa
...
+      switch (log_type)
+	{
+	case LOG_MVCC_UNDOREDO_DATA:
+	case LOG_MVCC_DIFF_UNDOREDO_DATA:
+	{
+	  // align to read the specific record info
+	  LOG_READ_ADVANCE_WHEN_DOESNT_FIT (thread_p, sizeof (LOG_REC_MVCC_UNDOREDO), log_lsa, log_page_p);
+	  const LOG_REC_MVCC_UNDOREDO *const log_rec
+	    = (const LOG_REC_MVCC_UNDOREDO *) ((char *)log_page_p->area + log_lsa->offset);
+	  tdes->mvccinfo.id = log_rec->mvccid;
```
razvanradulescu: Does a code exist (and is executed after all log analysis is complete) where each TDES is checked if has a valid mvccid ?

razvanradulescu (approved):

src/transaction/log_recovery_analysis.cpp
cristiarg: It can be done but it would be more verbose.
However, I separated non-mvcc and mvcc analysis functions.

src/transaction/log_recovery_analysis.cpp
cristiarg: There is no such check.
Is it needed?
The way I understood it is that the vast majority of transactions do have an mvccid, but I also encountered transactions that will not have an mvccid (eg: I think the checkpoint transaction but I'm not sure).
There was an assert that the mvccid must be valid in the `log_rv_analysis_complete` function but it was raised when such a transaction commited.

src/transaction/log_recovery_analysis.cpp
cristiarg: extracted function

PR #3764, opened 2022-08-11
[LETS-483] [LETS-495] correct mvcc table initialization on PTS

cristiarg:
http://jira.cubrid.org/browse/LETS-483
http://jira.cubrid.org/browse/LETS-495

Correctly initialize mvcc table on passive transaction server:
- when PTS starts, it receives boot information from the page server (PS)
- this info will contain the LSA of the most recent transaction table snapshot and `log_Gl.hdr.append_lsa` from page server (which is the LSA up to which PS is able to return transactional log pages)
- PTS then proceeds to initialization a valid mvcc table
- it first loads the transactions from the snapshot into the transaction table
- with the information up to this point (the transaction table snapshot point, as it were), it reconstructs a correct MVCC table
- afterwards, it continues by analysing the log up to the point specified by the value of `log_Gl.hdr.append_lsa` received from PS
- while analysing the log, it will maintain MVCC information in the transaction table and, when encountering transaction commits or aborts, it will also complete present mvccid's accordingly
- after finishing log analysis, the MVCC table is in a state consistent to be picked up by the - soon to be started - transactional log replication

Implementation:
- `log_initialize_passive_tran_server`
    - pass `replication_start_redo_lsa` to `log_recovery_analysis_from_trantable_snapshot` to allow `log_recovery_analysis` to know when to stop analysing the log
    - remove call to `logtb_define_trantable_log_latch` which, after log recovery analysis, will incorrectly reset an already initialized transaction table **and** the already initialized MVCC table (TODO: there is a discussion whether, after log recovery/initialization, the transaction table needs to be cleared up or not, but will be dealt with separately)
- `log_recovery_analysis_from_trantable_snapshot`:
    - do `log_recovery_build_mvcc_table_from_trantable` right after constructing the transaction table out of the information in transaction table snapshot; this will provide a basis for log recovery analysis to process log records and maintain&update the MVCC information (see Remark1 below)
    - call `log_recovery_analysis_internal` but request to stop before the LSA at which point log replication will take over
- `log_recovery_build_mvcc_table_from_trantable`: also receives the parameters `replicated_mvcc_next_id` with which, in the absence of any active transactions in the transaction table snapshot (and, consequently, in the recovered transaction table), it will initialize the MVCC table by setting `log_Gl.hdr.mvcc_next_id` and initializing the MVCC table (`log_Gl.mvcc_table.reset_start_mvccid`)
- `log_recovery_analysis_internal`:
    - separated new internal function which takes extra parameters: `reset_mvcc_table` whether to [re]initialize or not the MVCC table; `stop_before_lsa` which - if present - specifies the LSA before which the analysis should stop (this should only be the case for passive transaction server initialization as the same function is used for monolithic server recovery after a crash)
- `mvcctable::complete_mvcc`: added an assert that, mvccid's which are completed during passive transaction server's initialization do not re-appear later during replication (extensive comment in code with details) (see Remark1 below).

Remark1:
`log_recovery_build_mvcc_table_from_trantable` will leave the mvccid table in an inconsistent/unsyched state
  - will be synched after `log_recovery_analysis_internal` is executed
  - while parsing the recovered transaction table, some MVCCIDs will be found which will not be a continuous series
  - these found MVCCIDs will be recorded (left) as active in the MVCC table and will be completed later on (either during log recovery analysis or during the regular replica transaction server log replication)
  - the missing MVCCIDs (the "gaps") will contain MVCCIDs which can be in one of the following states:
    1. they belong to transactions which have already been completed (commited/aborted);
    2. they belong to transactions which have not been already completed; these can be transactions that have reserved an MVCCID but have not yet managed to insert a log record with the MVCCID to the transactional log (for this purpose, see also the case with `LOG_ASSIGNED_MVCCID` log record which, just before concluding a transaction, will insert an artificial log records to the transactional log; this is needed to ensure that **all** MVCCIDs make it from the active transaction server to the replica transaction servers for a correct mvcc table)
  - these missing MVCCIDs will be kept in a collection for completion after `log_recovery_analysis_internal` is executed
  - `log_recovery_analysis_internal` will also process the encountered MVCCID and will complete them according to the log (`LOC_COMMIT`, `LOG_ABORT` log records)
  - after `log_recovery_analysis_internal`, the missing MVCCIDs will be either (corresponding with the 1. and 2. states described above):
    1. completed because the transaction is *considered* as having already been completed in the past (before the point where the transaction snapshot was taken)
    2. ignored, because `log_recovery_analysis_internal` would have already completed the MVCCIDs as part of log records processing
  - the latter happens in function `mvcctable::complete_mvccids_if_still_active`
  - Caveat: the transactions whose MVCCIDs are completed are *considered* as having either commited or aborted in the past (before the time when the transaction table snapshot was taken on the active transaction server) although there is no guaranteed that this has happened; in practice, there can be very long running transactions which might register MVCC log records even after the `log_recovery_analysis_internal` has been executed and the regular log replication daemon is started and processes log records; this issue will be dealt with separately

### DIFF
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -1577,6 +1577,8 @@ error:
  *
  * return: nothing
  *
+ * NOTE: this function assumes that the trantable is initialized (as of now, it is done in the function
+ *  caling this one)
  */
 void
 log_initialize_passive_tran_server (THREAD_ENTRY * thread_p)
@@ -1650,6 +1652,8 @@ log_initialize_passive_tran_server (THREAD_ENTRY * thread_p)
     }
     // prior lists from page server are being received now
 
+    assert (LSA_LE (&most_recent_trantable_snapshot_lsa, &replication_start_redo_lsa));
+
     // NOTE: following situations can happen with regard to most recent transaction table snapshot lsa:
     //  1. it is null here at destination on PTS:
     //    1.1. the PTS has been started too fast after the ATS and did not get to generate and relay at least
@@ -1666,7 +1670,8 @@ log_initialize_passive_tran_server (THREAD_ENTRY * thread_p)
     //        - during log analysis, the PTS will ignore (skip) the newer transaction table snapshot log records
     //
     assert (!most_recent_trantable_snapshot_lsa.is_null ());
-    log_recovery_analysis_from_trantable_snapshot (thread_p, most_recent_trantable_snapshot_lsa);
+    log_recovery_analysis_from_trantable_snapshot (thread_p, most_recent_trantable_snapshot_lsa,
+						   replication_start_redo_lsa);
   }
   // prior lists are consumed and flushed to log pages
 
@@ -1676,13 +1681,8 @@ log_initialize_passive_tran_server (THREAD_ENTRY * thread_p)
 
   pts_ptr->start_log_replicator (replication_start_redo_lsa);
 
-  err_code = logtb_define_trantable_log_latch (thread_p, -1);
-  if (err_code != NO_ERROR)
-    {
-      logpb_fatal_error (thread_p, true, ARG_FILE_LINE,
-			 "log_initialize_passive_tran_server: error initializing transaction table");
-      return;
-    }
+  // NOTE: do not re-define trantable here; already called in boot_restart_server
+  // calling again here, will reset all info in already initialized transaction table
 
   // not other purpose on passive transaction server than to conform various checks that rely on the recovery state
   // one such example: the check for "is temporary volume" in page buffer
--- a/src/transaction/log_recovery_analysis.cpp
+++ b/src/transaction/log_recovery_analysis.cpp
@@ -2379,6 +2379,8 @@ log_recovery_build_mvcc_table_from_trantable (THREAD_ENTRY *thread_p, MVCCID rep
 	    }
 	}
     }
+  assert ((max_possible_mvccid == smallest_mvccid && min_possible_mvccid == largest_mvccid)
+	  || (max_possible_mvccid != smallest_mvccid && min_possible_mvccid != largest_mvccid));
 
   if (smallest_mvccid == max_possible_mvccid && largest_mvccid == min_possible_mvccid)
     {
@@ -2418,8 +2420,6 @@ log_recovery_build_mvcc_table_from_trantable (THREAD_ENTRY *thread_p, MVCCID rep
 
       log_Gl.hdr.mvcc_next_id = largest_mvccid + 1;
     }
-
-  const MVCCID oldest_visible_mvccid = log_Gl.mvcc_table.update_global_oldest_visible ();
 }
 
 /* log_recovery_analysis_from_trantable_snapshot - perform recovery for a passive transaction server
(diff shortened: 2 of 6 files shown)


src/transaction/log_manager.c
```diff
@@ -1676,13 +1681,8 @@ log_initialize_passive_tran_server (THREAD_ENTRY * thread_p)
...
 
-  err_code = logtb_define_trantable_log_latch (thread_p, -1);
-  if (err_code != NO_ERROR)
-    {
-      logpb_fatal_error (thread_p, true, ARG_FILE_LINE,
-			 "log_initialize_passive_tran_server: error initializing transaction table");
-      return;
-    }
+  // NOTE: do not re-define trantable here; already called in boot_restart_server
+  // calling again here, will reset all info in already initialized transaction table
```
cristiarg, line 1685: I am sure that at this point the already initialized mvcc table should not be reset (which `logtb_define_trantable_log_latch` will also do).
What I am not sure of, is if we actually do need the transaction table initalization, but without touching the mvcc table.

razvanradulescu: I believe in theory we do need it (reinit tran table without touching mvcc table): tdes table was used during recovery process only, and after this TDES should be cleaned-up.
But, practically, I don't think it matters: TDES table will be used only by replication thread(s), each new transaction is assigned a clean TDES, there is no logging on PTS, we don't care about tran_table checkpoint.

src/transaction/log_recovery_analysis.cpp
```diff
@@ -2335,30 +2373,53 @@ log_recovery_build_mvcc_table_from_trantable (THREAD_ENTRY *thread_p)
...
+	      // TODO: for later, mvcc sub ids
+	      assert (tdes->mvccinfo.sub_ids.empty ());
 	    }
 	}
     }
-  log_Gl.hdr.mvcc_next_id = smallest_mvccid;
-  log_Gl.mvcc_table.reset_start_mvccid ();
 
-  if (!present_mvccids.empty ())
+  if (smallest_mvccid == max_possible_mvccid && largest_mvccid == min_possible_mvccid)
```
razvanradulescu, line 2388:
```suggestion
  if (smallest_mvccid == max_possible_mvccid || largest_mvccid == min_possible_mvccid)
```

razvanradulescu (approved):

src/transaction/log_recovery_analysis.cpp
cristiarg: In practice, if either of them is initialized, the other is initialized as well. So the conditions are - somewhat - equivalent. However, I added an assert along these lines:
```
  assert ((max_possible_mvccid == smallest_mvccid && min_possible_mvccid == largest_mvccid)
	  || (max_possible_mvccid != smallest_mvccid && min_possible_mvccid != largest_mvccid));
```

src/transaction/log_manager.c
cristiarg: I tried also this approach and I have an [ugly] patch that I will try to see whether it is absolutely needed or not.

src/transaction/log_recovery_analysis.cpp
```diff
@@ -2415,7 +2512,7 @@ log_recovery_analysis_from_trantable_snapshot (THREAD_ENTRY *thread_p,
     return true;
   });
 
-  LOG_SET_CURRENT_TRAN_INDEX (thread_p, sys_tran_index);
+  log_Gl.mvcc_table.complete_mvccids_if_still_active (LOG_SYSTEM_TRAN_INDEX, in_gaps_mvccids, true);
```
razvanradulescu: I think here we need a TODO.
There could be transactions without logs encountered during recovery phase. If we complete them here, we may encounter log records from these transactions later, during normal replication.

razvanradulescu: Why do we use committed=true here ?
Since the transaction is active and we didn't encounter a "COMMIT" log, I expect the transaction should be considered aborted.

cristiarg: Agreed, although from the MVCCID completion point of view does not make a difference. Changed.

PR #3769, opened 2022-08-16
[CBRD-23945] Fix to destroy overflow files on rollback

wrlawodms:
http://jira.cubrid.org/browse/CBRD-23945

Please see the Jira comment for the cause and the design consideration. Welcome to even another design for this issue.

**Implementation**:
- Two log records for undo are added: RVBT_CREATE_BTREE, RVHF_CREATE_HEAP. I hope that they will be used for doing logical jobs for undoing the file creation though it only cleans its overflow file now.
- RCV_IS_LOGICAL_COMPENSATE_MANUAL(): means not the basic CLR pattern, even not the default logical CLR pattern. It means the log record does the logical CLR in the undo function. See the log_rollback_record().
- The reason they has to be the RCV_IS_LOGICAL_COMPENSATE_MANUAL() is that they leave CLR conditionally.


wrlawodms:
It's been a long time. I close and get it back to the backlog.

PR #3772, opened 2022-08-18
[CBRD-24187] Append undo log record only when it is needed in catalog_drop_representation_helper

joohok:
http://jira.cubrid.org/browse/CBRD-24187

Purpose

Undo log record for spage_update() is appended whether or not it is needed.
So, it is fixed to add undo log record only when it is needed.

### DIFF
--- a/src/storage/system_catalog.c
+++ b/src/storage/system_catalog.c
@@ -1624,13 +1624,13 @@ catalog_drop_representation_helper (THREAD_ENTRY * thread_p, PAGE_PTR page_p, VP
 
   spage_get_record (thread_p, page_p, CATALOG_HEADER_SLOT, &record, COPY);
 
-  log_append_undo_recdes2 (thread_p, RVCT_UPDATE, &catalog_Id.vfid, page_p, CATALOG_HEADER_SLOT, &record);
-
   overflow_vpid.pageid = CATALOG_GET_PGHEADER_OVFL_PGID_PAGEID (record.data);
   overflow_vpid.volid = CATALOG_GET_PGHEADER_OVFL_PGID_VOLID (record.data);
 
   if (overflow_vpid.pageid != NULL_PAGEID)
     {
+      log_append_undo_recdes2 (thread_p, RVCT_UPDATE, &catalog_Id.vfid, page_p, CATALOG_HEADER_SLOT, &record);
+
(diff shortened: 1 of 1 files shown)


hornetmj (approved):

src/storage/system_catalog.c
```diff
@@ -1624,13 +1624,13 @@ catalog_drop_representation_helper (THREAD_ENTRY * thread_p, PAGE_PTR page_p, VP
...
   spage_get_record (thread_p, page_p, CATALOG_HEADER_SLOT, &record, COPY);
 
-  log_append_undo_recdes2 (thread_p, RVCT_UPDATE, &catalog_Id.vfid, page_p, CATALOG_HEADER_SLOT, &record);
-
   overflow_vpid.pageid = CATALOG_GET_PGHEADER_OVFL_PGID_PAGEID (record.data);
   overflow_vpid.volid = CATALOG_GET_PGHEADER_OVFL_PGID_VOLID (record.data);
 
   if (overflow_vpid.pageid != NULL_PAGEID)
     {
+      log_append_undo_recdes2 (thread_p, RVCT_UPDATE, &catalog_Id.vfid, page_p, CATALOG_HEADER_SLOT, &record);
```
wrlawodms: If the undo is appended on the same condition with the redo of it, why don't we use `log_append_undoredo_recdes2`?

wrlawodms (commented):
Not related to this issue, but one question for curiosity:
When spage_delete() fails on 1616L with the log for it is appended already on 1614L, what happens? The undo is tried even though the update failed?

src/storage/system_catalog.c
joohok: Yes, you are right. 

I also wanted to use the `log_append_undoredo_recdes2` function, but I followed the above method because undo and redo were left separately as above in all places logging using RVCT_UPDATE. 

After analysis, I think it is better to use `log_append_undoredo_recdes2` as you mentioned.

joohok:
> Not related to this issue, but one question for curiosity:
> When spage_delete() fails on 1616L with the log for it is appended already on 1614L, what happens? The undo is tried even though the update failed?

Even if the record in spage is not deleted (there is data in the slot_id), 
it deletes the data at the slot_id and inserts the data for undo during recovery.

it's okay, but it's a bit odd.

wrlawodms (approved):

PR #3775, opened 2022-08-24
[CBRD-24461] Fix the core that occurs when finding user information in cdc

joohok:
http://jira.cubrid.org/browse/CBRD-24461

Background
 - supplemental log for transaction user is logged at the begin of the transaction and at the end of the transaction
 - In CDC, when transaction user log is not found at the begin of the transaction, it search the transaction user info which is logged at the end of the transaction

Purpose
1. When a supplemental log related to serial is logged, the log which contains transaction user info may be omitted at the begin of the transaction. (serial for auto_increment case)
2. The core occurs while traversing the log records to find the tran user information that was recorded at the end of the transaction. This is because the page at the time it was copied to the local buffer does not have the latest record at this time.
3. When referring to the past log, the transaction user information recorded at the commit time is used only when the transaction start point is truncated.
4. So, if requested transaction is not committed, then cdc_find_user() returns fail


src/storage/heap_file.c
```diff
@@ -16725,6 +16725,19 @@ heap_set_autoincrement_value (THREAD_ENTRY * thread_p, HEAP_CACHE_ATTRINFO * att
 	  if (prm_get_integer_value (PRM_ID_SUPPLEMENTAL_LOG) > 0)
 	    {
 	      OID serial_obj_oid = att->auto_increment.serial_obj.load ().oid;
+
+	      LOG_TDES *tdes = LOG_FIND_CURRENT_TDES (thread_p);
+	      if (tdes)
+		{
+		  if (!tdes->has_supplemental_log)
```
```suggestion
	      if (tdes && !tdes->has_supplemental_log)
```

hornetmj: How about adding assert(tdes != NULL) If the tdes is always not NULL? Or it would better to add exception handling code if the tdes is NULL.

src/transaction/log_manager.c
```diff
@@ -13370,8 +13370,28 @@ cdc_find_user (THREAD_ENTRY * thread_p, LOG_PAGE * log_page, LOG_LSA process_lsa
   LOG_REC_SUPPLEMENT *supplement;
   char *data;
 
+  int tran_index = logtb_find_tran_index (thread_p, trid);
+  if (tran_index != NULL_TRAN_INDEX)
+    {
+      LOG_TDES *tdes = LOG_FIND_TDES (tran_index);
```
hornetmj: This thread_p is not the worker thread which processes the transaction. Does there exist the protection mechanism about the tdes after the worker thread finishes the transaction? (I mean, can we find the tdes after the transaction is completed?)

src/transaction/log_manager.c
```diff
@@ -13370,8 +13370,28 @@ cdc_find_user (THREAD_ENTRY * thread_p, LOG_PAGE * log_page, LOG_LSA process_lsa
...
+
+      /* if transaction is active or aborted, it can not find the user information.
+       * because transaction user is logged right before commit record */
+
+      if (!LOG_ISTRAN_COMMITTED (tdes))
+	{
+	  /* TODO: set appropriate error and return */
```
hornetmj: Do you have a plan to handle it?

src/transaction/log_manager.c
joohok: - there is a critical section for transaction table (TR_TABLE_CS_ENTER_READ_MODE () in login_find_tran_index()) 
--  It is possible to access and read tdes that are working in another thread.
- If it is a completed transaction, the tran_index will not be found using trid. This block is the code to filter active transactions. we don't need to handle the logs for active transaction.

src/transaction/log_manager.c
joohok: Sure, it is a refactoring factor for CDC.

src/storage/heap_file.c
joohok: I'm not sure that `(tdes != NULL)` is always true. because the caller for this function (heap_set_autoincrement_value()) check whether tdes is null or not. 

When a thread is allocated and initialized to perform a transaction, log_tdes and tran_index are assigned, so there is no case where the tran_index corresponding to the thread cannot be found (tdes==null). It seems nice to add an assert.

src/transaction/log_manager.c
wrlawodms: Do you have an issue with it? or why do you leave it as future work?

src/transaction/log_manager.c
```diff
@@ -13359,7 +13359,7 @@ cdc_make_timer_loginfo (time_t at_time, int trid, char *user, CDC_LOGINFO_ENTRY
 }
 
 static int
-cdc_find_user (THREAD_ENTRY * thread_p, LOG_PAGE * log_page, LOG_LSA process_lsa, int trid, char **user)
+cdc_find_user (THREAD_ENTRY * thread_p, LOG_LSA process_lsa, int trid, char **user)
```
wrlawodms, line 13344: I hope that there is a function comment which describes its usage and restriction. It took me time to clarify what's this function for. I am still quite confused and guessing. Please clarify the parameters and the situation it is used.

wrlawodms: Why don't we need to know in the case of committed trxs?

joohok: OK, I will add a function comment.

As the CDC code was written in haste, I am sorry that there are many unkind parts to reviewers. There is a lot of part to refactor and add. 

we need the committed transaction, and don't need to care about active transaction. I'll describe why and where this function is called at function comment.

src/transaction/log_manager.c
joohok: For now, there is no CDC error code related to this situation. 

There is still work to be done to fix, remove and add CDC error codes. 
I'll do this when I refactor that error codes.

wrlawodms: OK, please make the issue if it doesn't exist yet. I can't find a related issue in Jira.

wrlawodms (commented):
Is it OK when the completed trx is aborted? The log pages are flushed when it commits in log_change_tran_as_completed(), but it flushes pages conditionally if it aborts. It seems that the newly appended log record still can't be accessed when it aborts.

joohok:
> Is it OK when the completed trx is aborted? The log pages are committed when it commits in log_change_tran_as_completed(), but it flushes pages conditionally if it aborts. It seems that the newly appended log record still can't be accessed when it aborts.

OK, I understand what you mean.. 
 hmm... I'll have to think a bit more about how to deal with it. Thank you.

src/transaction/log_manager.c
joohok: Removed!

src/transaction/log_manager.c
joohok: Checking transaction status is removed! 

CDC thread can see the outdated LOG_TDES, because Transaction worker thread can modify the LOG_TDES. And it is not reasonable that cdc thread blocks the transaction worker from changing the LOG_TDES.

hornetmj (approved):

wrlawodms (approved):

PR #3787, opened 2022-08-31
[LETS-511] Atomic replication for updating a heap record and its prev_version_lsa

wrlawodms:

Add the redo-only log record, RVHF_SET_PREV_VERSION_LSA, and append it with the atomic replication mechanism. Some heap operations already use the mechanism, so I add it in case it is not used.

See the Jira issue for more information about the issue and approach. 

Log records to be applied atomically for each update case:
- **REC_HOME** -> * : For any case, it's enough to set prev_version_lsa and update the home record atomically.
- **REC_RELOCATION**:
-> **REC_HOME, REC_BIGONE, and REC_RELOCATION on the same page**: The home record is updated, and it has to be applied atomically with setting prev version lsa. Actually, for all cases, LOG_START_ATOMIC_REPL is already appended, but I have added it conditionally for safety.
-> **REC_RELOCATION on another page**: The home record is not updated, so the forward record has to be updated atomically with setting prev version lsa.
- **REC_BIGONE** -> * : The BIGONE redo record already has the previous version lsa, so we don't need to care about this case.

Note. I have tested it with the REC_HOME records with the scenario in the Jira issue and see it works and gets a consistent result. However, I haven't tested with every heap log type transition such as REC_HOME -> REC_RELOCATION and REC_NEWHOME.


src/storage/heap_file.c
```diff
@@ -25231,8 +25306,9 @@ heap_get_visible_version_internal (THREAD_ENTRY * thread_p, HEAP_GET_CONTEXT * c
  * Note: It is expected to have the home page fixed and also the forward page in case of relocation.
  */
 static int
-heap_update_set_prev_version (THREAD_ENTRY * thread_p, const OID * oid, PGBUF_WATCHER * home_pg_watcher,
-			      PGBUF_WATCHER * fwd_pg_watcher, LOG_LSA * prev_version_lsa)
+heap_update_set_prev_version (THREAD_ENTRY * thread_p, const HFID * hfid, const OID * oid,
+			      PGBUF_WATCHER * home_pg_watcher, PGBUF_WATCHER * fwd_pg_watcher,
+			      LOG_LSA * prev_version_lsa)
```
cristiarg: Nit:
```suggestion
			      const LOG_LSA * prev_version_lsa)
```
The signature of `or_mvcc_set_log_lsa_to_record` must also be changed to take a `const LOG_LSA *lsa`.
Up to you if you want to make the change or not.

src/storage/heap_file.c
```diff
@@ -25231,8 +25306,9 @@ heap_get_visible_version_internal (THREAD_ENTRY * thread_p, HEAP_GET_CONTEXT * c
  * Note: It is expected to have the home page fixed and also the forward page in case of relocation.
  */
 static int
-heap_update_set_prev_version (THREAD_ENTRY * thread_p, const OID * oid, PGBUF_WATCHER * home_pg_watcher,
-			      PGBUF_WATCHER * fwd_pg_watcher, LOG_LSA * prev_version_lsa)
+heap_update_set_prev_version (THREAD_ENTRY * thread_p, const HFID * hfid, const OID * oid,
```
cristiarg, line 25309: Does the function's description and note need updating?

src/storage/heap_file.c
```diff
@@ -25304,6 +25388,16 @@ heap_update_set_prev_version (THREAD_ENTRY * thread_p, const OID * oid, PGBUF_WA
...
       error_code = or_mvcc_set_log_lsa_to_record (&forward_recdes, prev_version_lsa);
 
+      if (heap_ovf_find_vfid (thread_p, hfid, &ovf_vfid, false, PGBUF_UNCONDITIONAL_LATCH) == NULL)
+	{
+	  assert (false);
+	  error_code = ER_FAILED;
+	  goto end;
+	}
+
+      log_append_redo_data2 (thread_p, RVHF_SET_PREV_VERSION_LSA, &ovf_vfid, overflow_pg_watcher.pgptr, -1,
```
cristiarg: Nit (all new occurences):
```suggestion
      log_append_redo_data2 (thread_p, RVHF_SET_PREV_VERSION_LSA, &ovf_vfid, overflow_pg_watcher.pgptr, NULL_OFFSET,
```

cristiarg (approved):

src/storage/heap_file.c
wrlawodms: I don't think I should change it because what the comment says is not changed. It doesn't mention logging, which I have added.

src/storage/heap_file.c
wrlawodms: Thanks. I have also changed to use NULL_OFFSET in the recovery function.

alexstanro (approved):
I think that the new log record is required only for scalability. Also atomic.

src/storage/heap_file.c
```diff
@@ -25308,7 +25308,7 @@ heap_get_visible_version_internal (THREAD_ENTRY * thread_p, HEAP_GET_CONTEXT * c
 static int
 heap_update_set_prev_version (THREAD_ENTRY * thread_p, const HFID * hfid, const OID * oid,
 			      PGBUF_WATCHER * home_pg_watcher, PGBUF_WATCHER * fwd_pg_watcher,
-			      LOG_LSA * prev_version_lsa)
+			      const LOG_LSA * prev_version_lsa)
```
cristiarg, line 25311: Build fails.
The signature of the function must be changed everywhere.
Also the signature of or_mvcc_set_log_lsa_to_record must also be changed everywhere to:
```
int or_mvcc_set_log_lsa_to_record (RECDES * record, const LOG_LSA * lsa)
```
because the function does not change the data `lsa` points to.

wrlawodms: You're correct, I should've checked. :joy:

wrlawodms: Corrected.

alexstanro:
The issue with adding other transaction log entries is that it affects the performance. Thus, there is high contention on the prior mutex, that is acquired every time when a new transaction log entry is added. More, a transaction has to wait more to commit, since more transactional log has to be written on log pages and that log pages has to be written on disk, up to transaction commit LSA. Also, the log page buffer may fill faster, leading to other performance issue. Finally, the log written on disk must be cleaned - archive removed. All these leads to conclusion that if is not required, adding additional transactional logging must be avoided, to not slow down the overall execution. That's why the transactional log used only for scalability, must be added only in that case.

wrlawodms:
@alexstanro Thank you for the detailed advice. I totally agree. From what I know, by the way, we are not going to support CUBRID without scalability. So, I didn't add any conditions to append the new transaction log. Please tell me if I miss something or something to do. 

> That's why the transactional log used only for scalability, must be added only in that case.   

I think it's always the case.

hornetmj (approved):

alexstanro:
> @alexstanro Thank you for the detailed advice. I totally agree. From what I know, by the way, we are not going to support CUBRID without scalability. So, I didn't add any conditions to append the new transaction log. Please tell me if I miss something or something to do.
> 
> > That's why the transactional log used only for scalability, must be added only in that case.
> 
> I think it's always the case.

alexstanro:
> @alexstanro Thank you for the detailed advice. I totally agree. From what I know, by the way, we are not going to support CUBRID without scalability. So, I didn't add any conditions to append the new transaction log. Please tell me if I miss something or something to do.
> 
> > That's why the transactional log used only for scalability, must be added only in that case.
> 
> I think it's always the case.

Well, I think that the impact of that transactional log must be measured, even if only the scalability is supported.  At least, we must know how serious is the impact. If the impact with additional logging vs. no additional logging is serious (measure the time spent in prior mutex lock, wait for log flush...). If you conclude that the impact in performance is serious, that issue must be fixed and we have some idea about how to reduce the impact.

cristiarg:
@wrlawodms
The build fails because of an unit test I overlooked to adapt at the proper time but which is fixed by https://github.com/CUBRID/cubrid/pull/3779 which is waiting for an approve.

joohok (approved):
LGTM. It seems to me that all record conversion cases are considered. However, test cases are still necessary.

PR #3792, opened 2022-09-02
[CBRD-24438] Hide the password in the log file

ctshim:
http://jira.cubrid.org/browse/CBRD-24438

* Hide the password in the log file
    - Change the password string to be output as '***' in the "ddl_audit.log", "*.sql.log, *.slow.log" files.
    - Target phrase to hide password
        - add_user(), set_password(), login() methods
        - create user / alter user / create server / alter server
    - Even if there is no password, '****' is output in the omitted location.

* Differences from previous versions of ddl_audit.log
   - Unlike previous versions, it is written to the log even when CallableStatement is used.
   - When using ExecuteUpdate(), only the first statement is executed, even if there are multiple statements. At this time, unlike previous versions, only what was actually performed is logged in ddl_audit.log.
   - In previous versions, when using PreparedStatement / executeBatch(), there was a problem in which logs were not recorded even if DDL statements were present.
   - When executing multiple statements simultaneously, in the previous version, the entire statement was written to the log file regardless of whether it was a DDL statement, but this was modified so that only the DDL statement is written to the log.


src/base/password_log.cpp
```diff
@@ -0,0 +1,863 @@
...
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+
+
+/*
+ * password_log.c -
```
beyondykk9: How about rename the file name to hide_password.cpp for class name "hide_password".

ctshim: OK, I will change the file name.

beyondykk9 (approved):

src/base/password_log.cpp
```diff
@@ -0,0 +1,895 @@
+/*
+ * Copyright 2008 Search Solution Corporation
```
kisoo-han: Only the cubrid copyright is described for the newly created file.

src/base/password_log.cpp
```diff
@@ -0,0 +1,895 @@
...
+//#include "parse_tree.h"
+//#include "system_parameter.h"
+//#include "environment_variable.h"
+//#include "broker_config.h"
+//#include "util_func.h"
+
+//extern PARSER_CONTEXT *this_parser;
+//==========================================================================
+
+static unsigned char isSpace[0x100] =
```
kisoo-han: trivial comment.
In CUBRID, C code, not Java code, we generally use hyphens rather than Camel Case notation.
isSpace => is_space

Is there any reason that use of 0x100 rather than 256.

src/parser/parse_tree.h
```diff
@@ -3837,6 +3837,8 @@ typedef struct remote_cols
 typedef int (*PT_CASECMP_FUN) (const char *s1, const char *s2);
 typedef int (*PT_INT_FUNCTION) (PARSER_CONTEXT * c);
 
+#include "hide_password.h"
```
hgryoo: Is there any reason for this include directives to place here instead of on the top of the file?

src/base/hide_password.h
```diff
@@ -0,0 +1,77 @@
...
+ *  limitations under the License.
+ *
+ */
+
+
+/*
+ * hide_password.h -
+ */
+
+#ifndef	_PASSWORD_LOG_H_
```
kisoo-han: You changed entity name from password_log  to hide_password, then what about change the macro **\_PASSWORD_LOG_H\_** to \_HIDE_PASSWORD\_ as well.

ctshim: OK, I will fix it.
Thank you.

src/parser/parse_tree.h
ctshim: Just mistake!
Thank you.

src/base/password_log.cpp
ctshim: The reason 0x100 is used instead of 256 is to express it based on the size of the bit.
It doesn't mean much.

I will fix the variable name.
Thank you.

src/base/hide_password.cpp
```diff
@@ -0,0 +1,863 @@
...
+}
+
+int
+hide_password::snprint_password (char *msg, int size, char *query, HIDE_PWD_INFO_PTR hide_pwd_ptr)
+{
+  char *qryptr = query;
+  char chbk;
+  int pos;
+  int length = 0;
+  EN_ADD_PWD_STRING en_pwd_string;
```
kisoo-han: `en_pwd_string` is not referenced.

ctshim: Thank you.

src/base/password_log.cpp
ctshim: Okay. 
Thanks.

ctshim: @kisoo-han 
If I do as you said, the license check in CI will fail.
So, we'll stick with the original wording again.

src/base/ddl_log.c
```diff
@@ -290,69 +297,99 @@ logddl_set_broker_info (const int index, const char *br_name)
...
-	  ddl_audit_handle.sql_text[len] = '\0';
-	}
+      ddl_audit_handle.sql_text_len += tlen;
     }
-}
 
+  ddl_audit_handle.sql_text[ddl_audit_handle.sql_text_len] = '\0';
+}
 
 bool
```
kisoo-han:
```suggestion
static bool
```

src/base/ddl_log.c
```diff
@@ -290,69 +297,99 @@ logddl_set_broker_info (const int index, const char *br_name)
...
-logddl_set_stmt_type (int stmt_type)
+logddl_set_stmt_type_internal (int stmt_type, PT_NODE * statement)
```
kisoo-han: In general, xxx_internal () is a static function and is usually called from the xxx () function. And, xxx() can be called from a function outside that file. Defining only xxx_internal() without the function xxx() seems a bit strange to me.

src/base/ddl_log.c
```diff
@@ -479,30 +516,6 @@ logddl_set_commit_mode (bool mode)
...
-
-static void
-logddl_set_elapsed_time (long sec, long msec)
-{
-  if (ddl_logging_enabled)
-    {
-      snprintf (ddl_audit_handle.elapsed_time, 20, "elapsed time %ld.%03ld", sec, msec);
-    }
-}
-
```
kisoo-han, line 505: Because you deleted above two functions, member 'jsp_mode' is not used anymore.
What about remove 'jsp_mode' from `struct t_ddl_audit_handle` as well.

src/base/ddl_log.c
```diff
@@ -1307,26 +1299,22 @@ logddl_remove_char (char *string, char ch)
 #endif
 
 bool
```
kisoo-han:
```suggestion
static bool
```

src/base/hide_password.cpp
```diff
@@ -0,0 +1,861 @@
...
+	      break;
+	    }
+	  query++;
+	}
+    }
+  return query;
+}
+
+char *
+hide_password::get_quot_string (char *ps)
```
kisoo-han: '**get_quot_string**' seems to recognize the quoted string itself.
Actually, this function is understood as a function that finds the end of the quote string.
Litte bit confused, how about changing the name, for example `find_end_of_quot ()`

src/base/hide_password.cpp
```diff
@@ -0,0 +1,861 @@
...
+  char *token;
+
+  *method_password_len = -1;
+  token = get_token (ps, len);
+  if (len != 1 || *token != '(')
+    {
+      return NULL;
+    }
+
+  if (strcmp (method_name, "set_password") == 0)
```
kisoo-han:
```suggestion
  if (strcasecmp (method_name, "set_password") == 0)
```

src/base/hide_password.cpp
```diff
@@ -0,0 +1,861 @@
...
+  * CREATE USER user_name
+  * CREATE USER user_name PASSWORD 'password_string'
+  * CREATE USER user_name PASSWORD _utf8'password_string'
+  * CREATE USER user_name PASSWORD _euckr'password_string'
+  * ALTER USER user_name PASSWORD 'password_string'
+  * ALTER SERVER srv1 CHANGE PASSWORD='password string'
+  */
+
+  *password_len = 0;
+  *is_found_pwd_keyword = false;
```
kisoo-han: trivial suggestion:
is_pwd_keyword_found

src/base/hide_password.cpp
```diff
@@ -0,0 +1,861 @@
...
+	}
+    }
+
+  if (*ps == '\0')
+    {
+      // check case "CREATE USER user_name"
+      return (is_create && !is_server) ? ps : NULL;
+    }
+
+  prev = NULL;
```
kisoo-han, line 354: trivial:
what about removing this line and initialize it in decl.
char *token, *prev = NULL;

src/base/hide_password.cpp
```diff
@@ -0,0 +1,861 @@
...
+
+bool hide_password::check_lead_string_in_query (char **query, char **method_name, bool *is_create, bool *is_server)
+{
+  char   *ps;
+  const char *first_cmd_str[] = { "call", "create", "alter", NULL };
+  int  first_cmd_len[] = { 4, 6, 5, -1 };
+  const char *second_cmd_str[] = { "server", "user", NULL };
+  int  second_cmd_len[] = { 6, 4, -1 };
+  static const char *method_name_str[] = { "add_user", "set_password", "login", NULL };
+  int method_name_len[] = { 8, 12, 5, -1 };
```
kisoo-han, line 448: trivial, please ignore:
int method_name_len[] = { 
  strlen (method_name_str[0]),
  strlen (method_name_str[1]),
  strlen (method_name_str[2]),
-1 };

src/base/hide_password.cpp
```diff
@@ -0,0 +1,861 @@
...
+	    {
+	      is_call_stmt = true;
+	    }
+	  else
+	    {
+	      *is_create = (str_ptr == &first_cmd_str[1]) ? true : false;
+	    }
+	  break;
+	}
+    }
```
kisoo-han:
```suggestion
  for (int i = 0 ; *str_ptr; str_ptr++, len_ptr++, i++)
    {
      if (*len_ptr == len && strncasecmp (token, *str_ptr, *len_ptr) == 0)
	{
	  if (i == 0)
	    {
	      is_call_stmt = true;
	    }
	  else
	    {
	      *is_create = (i == 1) ? true : false;
	    }
	  break;
	}
    }
```

kisoo-han: or
```
#define IDX_CALL_STMT 0
#define IDX_CREATE_STMT 1
```

and
```
if (i == IDX_CALL_STMT)
  {
     is_call_stmt = true;
  }

```
or something like that.

src/base/hide_password.cpp
```diff
@@ -0,0 +1,861 @@
...
+	  len_ptr = method_name_len;
+	  str_ptr = method_name_str;
+	}
+      else
+	{
+	  len_ptr = second_cmd_len;
+	  str_ptr = second_cmd_str;
+	}
+
+      for ( /* empty */ ; *str_ptr; str_ptr++, len_ptr++)
```
kisoo-han: if you agreed upon above suggestion, this could be changed to the same style. (but trivial)

src/base/hide_password.cpp
```diff
@@ -0,0 +1,861 @@
...
+{
+  int offset;
+  char chbk;
+
+  while (*query)
+    {
+      offset = strcspn (query, "\r\n");
+      if (offset <= 0)
+	{
+	  cas_fprintf (fp, " ");
```
kisoo-han, line 692: cas_fprintf () in broker/cas_log.c do not check NULL so,
what about check whether fp is not NULL.
```
if (fp != NULL)
  {
    cas_fprintf (fp, " ");
  }

```

kisoo-han (approved):

src/base/ddl_log.c
ctshim: Thank you.

src/base/ddl_log.c
ctshim: OK, I agree.
I will fix it.   And logddl_set_sql_text_internal() also.

src/base/ddl_log.c
ctshim: Good point. 
Thank you.

src/base/ddl_log.c
ctshim: Thank you.

src/base/hide_password.cpp
ctshim: Oh, Thank you.

src/base/hide_password.cpp
```diff
@@ -0,0 +1,861 @@
...
+};
+
+#define SKIP_SPACE_CHARACTERS(p)                \
+  do {                                          \
+        while (is_space[(unsigned char) *p])     \
+	    {                                   \
+	      p++;                              \
+	    }                                   \
+   } while(0)
+
```
hgryoo, line 85: duplicated code.
```
// base/chartype.c
int char_isspace (int c)
{
  return ((c) == ' ' || (c) == '\t' || (c) == '\r' || (c) == '\n' || (c) == '\f' || (c) == '\v');
}

```

src/base/hide_password.cpp
ctshim: What you thought is correct.
I think the function name you suggested is appropriate.

src/base/hide_password.cpp
ctshim: Thank you.

src/base/hide_password.cpp
ctshim: We plan to maintain the status quo for the time being.
I don't want to keep calling the function to skip consecutive spaces.
Additionally, a single function call requires multiple comparisons, which I want to avoid as much as possible.

I think it's necessary to improve functions in the chartype.c,  if possible.

src/base/hide_password.cpp
ctshim: Currently, it is guaranteed that fp is not null during the function call.
We will add assert() in preparation for future modifications.

src/base/hide_password.cpp
ctshim: I made some changes based on your comments.
Thank you.

src/base/hide_password.cpp
ctshim: Thank you.

src/base/hide_password.cpp
ctshim: I'll keep the current code for this part.
Thank you.

src/base/hide_password.cpp
ctshim: I'll keep the current code for this part.
However, I will reinforce some of the comments.

src/base/hide_password.cpp
ctshim: First, I created the issue below.
[CBRD-25043]

hyunikn (approved):

hgryoo (approved):

airnet73 (approved):

PR #3806, opened 2022-09-09
[LETS-527] Deal the NULL previous version LSA with TOO_NEW_FOR_SNAPSHOT

wrlawodms:
http://jira.cubrid.org/browse/LETS-527

I missed a normal case where the previous version can be NULL while the snapshot result is TOO_NEW_FOR_SNAPSHOT.
It means after a snapshot is taken, a record is newly inserted.

### DIFF
--- a/src/storage/heap_file.c
+++ b/src/storage/heap_file.c
@@ -25258,7 +25258,7 @@ heap_get_visible_version_internal (THREAD_ENTRY * thread_p, HEAP_GET_CONTEXT * c
 	  const LOG_LSA *prev_lsa = &MVCC_GET_PREV_VERSION_LSA (&mvcc_header);
 	  if (prev_lsa->is_null ())
 	    {
-	      /* It means the record is newly inserted after the snapshot is taken. */
+	      /* The record is newly inserted after the snapshot is taken. */
 	      scan = S_DOESNT_EXIST;
 	      goto exit;
 	    }
--- a/src/storage/heap_file.c
+++ b/src/storage/heap_file.c
@@ -865,7 +865,7 @@ static void heap_page_rv_chain_update (THREAD_ENTRY * thread_p, PAGE_PTR heap_pa
 static int heap_scancache_add_partition_node (THREAD_ENTRY * thread_p, HEAP_SCANCACHE * scan_cache,
 					      OID * partition_oid);
 static SCAN_CODE heap_get_visible_version_from_log (THREAD_ENTRY * thread_p, RECDES * recdes,
-						    LOG_LSA * previous_version_lsa, HEAP_SCANCACHE * scan_cache,
+						    const LOG_LSA * previous_version_lsa, HEAP_SCANCACHE * scan_cache,
 						    int has_chn);
 static int heap_update_set_prev_version (THREAD_ENTRY * thread_p, const HFID * vfid, const OID * oid,
 					 PGBUF_WATCHER * home_pg_watcher, PGBUF_WATCHER * fwd_pg_watcher,
@@ -24957,7 +24957,7 @@ heap_rv_mvcc_redo_redistribute (THREAD_ENTRY * thread_p, const LOG_RCV * rcv)
  *   scan_cache(in): Heap scan cache.
  */
 static SCAN_CODE
-heap_get_visible_version_from_log (THREAD_ENTRY * thread_p, RECDES * recdes, LOG_LSA * previous_version_lsa,
+heap_get_visible_version_from_log (THREAD_ENTRY * thread_p, RECDES * recdes, const LOG_LSA * previous_version_lsa,
 				   HEAP_SCANCACHE * scan_cache, int has_chn)
 {
   LOG_LSA process_lsa;
--- a/src/storage/heap_file.c
+++ b/src/storage/heap_file.c
@@ -24971,11 +24971,7 @@ heap_get_visible_version_from_log (THREAD_ENTRY * thread_p, RECDES * recdes, con
 
   assert (scan_cache != NULL);
   assert (scan_cache->mvcc_snapshot != NULL);
-
-  if (previous_version_lsa->is_null ())
-    {
-      return S_DOESNT_EXIST;
-    }
+  assert (!previous_version_lsa->is_null ());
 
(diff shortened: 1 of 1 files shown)


cristiarg (approved):
It would be nice to add an extensive comment before `return S_DOESNT..` to explain the scenario. At your choice. Thanks.

wrlawodms:
@cristiarg I have added a comment for the case. On second thought, it seems awkward to get a previous version with the NULL previous LSA, which even always fails and returns S_DOESNT_EXIST. So, I've changed to keep the assertion and filter the case in the caller. 

Plus, I changed the description for the case. The case seems only when a record is newly inserted after the snapshot is taken because the ins_mvccid and prev_version_lsa are cleared together when every active transaction can see it, and in turn the snapshot can't be TOO_NEW_FOR_SNAPSHOT.

src/storage/heap_file.c
```diff
@@ -25255,10 +25255,18 @@ heap_get_visible_version_internal (THREAD_ENTRY * thread_p, HEAP_GET_CONTEXT * c
       snapshot_res = mvcc_snapshot->snapshot_fnc (thread_p, &mvcc_header, mvcc_snapshot);
       if (snapshot_res == TOO_NEW_FOR_SNAPSHOT)
 	{
+	  const LOG_LSA *prev_lsa = &MVCC_GET_PREV_VERSION_LSA (&mvcc_header);
```
cristiarg: Nit:
```suggestion
	  const LOG_LSA *const prev_lsa = &MVCC_GET_PREV_VERSION_LSA (&mvcc_header);
```

cristiarg (approved):

src/storage/heap_file.c
```diff
@@ -25255,10 +25255,18 @@ heap_get_visible_version_internal (THREAD_ENTRY * thread_p, HEAP_GET_CONTEXT * c
...
+	  if (prev_lsa->is_null ())
+	    {
+	      /* The record is newly inserted after the snapshot is taken. */
+	      scan = S_DOESNT_EXIST;
```
hornetmj, line 25262: Is there any reason to use the S_DOESNT_EXIST iso the S_SNAPSHOT_NOT_SATISFIED?

src/storage/heap_file.c
```diff
@@ -25255,10 +25255,18 @@ heap_get_visible_version_internal (THREAD_ENTRY * thread_p, HEAP_GET_CONTEXT * c
       snapshot_res = mvcc_snapshot->snapshot_fnc (thread_p, &mvcc_header, mvcc_snapshot);
       if (snapshot_res == TOO_NEW_FOR_SNAPSHOT)
 	{
+	  const LOG_LSA *prev_lsa = &MVCC_GET_PREV_VERSION_LSA (&mvcc_header);
+	  if (prev_lsa->is_null ())
```
hornetmj, line 25259: In develop branch, the heap_get_visible_version_from_log() can handle when the prev_lsa is NULL. But, this function cannot handle it in the scalability_dev because assert (!previous_version_lsa->is_null ()) condition. If we convince the assert condtion, then it is better not to make this situation when we build the recdes. Isn't it?

hornetmj: But, I think we cannot because the REC_BIGONE type. So, how about remove the assert() and add the if (prev_lsa->is_null ()) condition in the heap_get_visible_version_from_log() function.

cristiarg: @hornetmj @wrlawodms 
Please be aware that there is also a related issue which is a regression coming from QA - http://jira.cubrid.org/browse/LETS-532 - and the test bumped into the assert. So the assert must be evaluated also in the context of this regression (and, possibly others) in the most conservative way.
Me, personally, I would leave the assert if it does not offend any other tests - with the idea that it's best to contain the NULL LSA scenarios as much as possible (ie: be notified if other tests bump into it). If the assert fails again in a legitimate scenario, we can re-evaluate.

src/storage/heap_file.c
wrlawodms: It is what was returned before in the case.

src/storage/heap_file.c
joohok: If the `previous_version_lsa` is null inside the `heap_get_visible_version_from_log()` function, it does nothing(Line 25000) and throws an error (S_DOESNT_EXIST). Rather than adding asserts in `heap_get_visible_version_from_log()`, how about adding exception handling earlier?

wrlawodms: @hornetmj 
>In develop branch, the heap_get_visible_version_from_log() can handle when the prev_lsa is NULL.

I don't think it was handled correctly because it used the NULL previous LSA to compare with:
```if (LSA_LT (&oldest_prior_lsa, previous_version_lsa))```
, which looks unreasonable and triggers unnecessary appending prior lists.

>But, I think we cannot because the REC_BIGONE type.

The big one record can have the NULL previous LSA, and other types also can be considered to have the NULL lsa here (if the flag is off). However, it's only the case when the record is newly inserted. In this case, it seems unreasonable to find the previous version. On top of that, the semantics at all seems awkward: "to find a previous version in the LOG with NULL LSA (N/A location)".

wrlawodms: @joohok You mean the exception handler inside the function like hornetmj's opinion?

wrlawodms: @hornetmj oh sorry to confuse you. It doesn't trigger.
> which looks unreasonable and causes unnecessary appending prior lists.

joohok: > On top of that, the semantics at all seems awkward to find a previous version in the LOG with NULL LSA (N/A location);

I think this is right. Now I agree with your change. I think exception handler is unnecessary.

hornetmj: @wrlawodms @joohok I understand what you mean. However, in my point of view, it is better to handle in the heap_get_visible_version_from_log() from the heap_* API user's point. From a semantic point of view, I think there is no problem. However, I will follow the comments of the two reviewers.

hornetmj (approved):

joohok (approved):

PR #3816, opened 2022-09-18
[LETS-543] move log recovery analysis code back to log_recovery.c to ease merges from develop

cristiarg:
http://jira.cubrid.org/browse/LETS-543

There are no functional changes, only code moved around.

Still, kept the `corruption_checker` class in separate header file.

Other:
- fatal error if log recovery analysis fails to retrieve the page of the very first log record


src/transaction/log_recovery_analysis_corruption_checker.hpp
```diff
@@ -0,0 +1,157 @@
...
+{
+  m_is_page_corrupted = !logpb_page_has_valid_checksum (log_pgptr);
+}
+
+void corruption_checker::find_first_corrupted_block (const LOG_PAGE *log_pgptr)
+{
+  m_first_corrupted_rec_lsa = NULL_LSA;
+  for (size_t block_index = 0; block_index < m_blocks_in_page_count; ++block_index)
+    {
+      if (std::memcmp (get_block_ptr (log_pgptr, block_index), m_null_block.get (), IO_MAX_PAGE_SIZE) == 0)
```
joohok:
```suggestion
      if (std::memcmp (get_block_ptr (log_pgptr, block_index), m_null_block.get (), IO_BLOCK_SIZE) == 0)
```
Oh, this could be refactored in this issue.

cristiarg: done; I hope I got it right

hornetmj (approved):

src/transaction/log_recovery.c
```diff
@@ -994,94 +1051,2176 @@ log_recovery_finish_transactions (THREAD_ENTRY * const thread_p)
...
+	      LSA_COPY (&last_log_tdes->undo_nxlsa, &log_rec->prev_tranlsa);
+	      if (prm_get_bool_value (PRM_ID_LOGPB_LOGGING_DEBUG))
+		{
+		  _er_log_debug (ARG_FILE_LINE, "logpb_recovery_analysis: trid = %d, tail_lsa=%lld|%d\n",
+				 log_rec->trid, last_log_tdes->tail_lsa.pageid, last_log_tdes->tail_lsa.offset);
+		}
+	    }
+	}
+
+      assert (!prev_lsa.is_null ());
```
joohok, line 2931: How about removing this assertion? I think `prev_lsa` can be null in the case below.

1. `prev_lsa` is set to NULL_LSA at the beginning of `log_recovery_analysis_internal ()` ( log_recovery.c:2539L)
2. `logpb_fetch_page()` can fail in `log_recovery_analysis_internal ()` (log_recovery.c:2591L)
3. `log_rv_analysis_handle_fetch_page_fail ()` can be called with `prev_lsa`(NULL_LSA).

or How about throwing it as an exception? 

if it requires much modification, I think it is better to do it in other issue.

joohok (approved):

src/transaction/log_recovery.c
cristiarg: Thrown an exception - fatal. Thanks.

PR #3817, opened 2022-09-19
[LETS-449] Block file I/O operation while replicating disk-related operations in PTS

joohok:
http://jira.cubrid.org/browse/LETS-449

Purpose
Passive transaction server with remote storage should not do file I/O while replicating disk-related operations.
File I/O is performed when adding a new volume (RVDK_NEWVOL) or expanding a volume (RVDK_EXPAND_VOLUME).
PTS with remote storage only needs information about the number of permanent volumes on the PS and does not need to create any files. more detail is described in jira

Implement
1. increase `disk_Page_server_perm_volume_count` in disk_rv_redo_dboutside_newvol () (RVDK_NEWVOL)
2. Block RVDK_EXPAND_VOLUME from being replicated in atomic_replicator::redo_upto ()


src/storage/disk_manager.c
```diff
@@ -1222,16 +1222,38 @@ int
...
+       * disk_rv_redo_format (), which requires disk header page and does update the disk_Page_server_perm_volume_count,
+       * is not called during replication.
+       * Therefore, disk_Page_server_perm_volume_count for PTS is updated here when the new volume is created.
+       *
+       * NOTE : If it is decided to use disk_Cache in PTS, this part can be moved to disk_rv_redo_format ()
+       *        where disk_Cache is updated.
+       */
 
-  if (fileio_find_volume_descriptor_with_label (buf_vol_full_name) == NULL_VOLDES)
+      if (vhdr->type == DB_PERMANENT_VOLTYPE)
```
wrlawodms, line 1239: It doesn't seem to be DB_TEMPORARY_VOLTYPE.
```suggestion
```

src/storage/disk_manager.c
```diff
@@ -1222,16 +1222,38 @@ int
...
+	{
+	  disk_Page_server_perm_volume_count++;
+	}
```
wrlawodms, line 1242:
```suggestion
	  disk_Page_server_perm_volume_count++;
```
It doesn't seem that it can be DB_TEMPORARY_VOLTYPE.

src/storage/disk_manager.c
```diff
@@ -1404,15 +1432,6 @@ disk_rv_redo_format (THREAD_ENTRY * thread_p, const LOG_RCV * rcv)
...
       disk_Cache->perm_purpose_info.extend_info.nsect_total += volheader->nsect_total;
       disk_Cache->perm_purpose_info.extend_info.nsect_max += volheader->nsect_max;
-
-      if (is_tran_server_with_remote_storage ())
-	{
-	  // transaction servers with remote storage:
-	  //  - maintain a separate count of permanent volumes
-	  //  - have their own set of temporary volumes which they maintain using the disk Cache structures
-	  assert ((disk_Page_server_perm_volume_count + 1) == disk_Cache->nvols_perm);
-	  disk_Page_server_perm_volume_count++;
```
wrlawodms, line 1414: In the undo (and CLR) function of RVDK_FORMAT, the disk_Page_server_perm_volume_count is decreased when it's rollbacked. I'm concerned that the corresponding operations are done in different recovery indexes' redo and undo, increasing in REDO of RVDK_NEWVOL and decreasing in UNDO of RVDK_FORMAT.

wrlawodms: In disk_format(), there is a window in which the UNDO log of RVDK_FORMAT is appended, but the REDO log of RVDK_NEWVOL is not appended. If a crash or an error occurs in the window, it seems to be a problem.
 By the way, similarly, there is already a window in which UNDO log of RVDK_FORMAT is appended, but the REDO log of RVDK_FORMAT is not appended. I think it already has the same problem.

hornetmj:
test

src/storage/disk_manager.c
joohok: OK, Thanks

src/storage/disk_manager.c
joohok: OK, I agree. Thanks for detail feedback. 

In case of 
```
RVDK_FORMAT (undo)
RVDK_NEWVOL (redo)

crash!

RVDK_FORMAT (redo)
RVDK_FORMAT (redo)
```
disk_Page_server_perm_count is increased by `RVDK_NEWVOL`  redo log, 
but it can not be decreased by `RVDK_FORMAT` undo log (CLR). 
Because disk volume header is not loaded on PTS yet.. 

Maybe this should be fixed by other way, 
such as
1. changing the fetch mode  from OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT to OLD_PAGE while replicating this kind of log record.
2. or comparing the volume id with disk_Page_server_perm_count when it needs to be increased or decreased .

joohok: 2 can not be a solution, because PTS keeps getting the wrong number of permanent volumes on the page server and tries to fetch the page using it.

src/storage/disk_manager.c
```diff
@@ -1222,16 +1222,38 @@ int
...
+       *        where disk_Cache is updated.
+       */
 
-  if (fileio_find_volume_descriptor_with_label (buf_vol_full_name) == NULL_VOLDES)
+      if (vhdr->type == DB_PERMANENT_VOLTYPE)
+	{
+	  disk_Page_server_perm_volume_count++;
+	}
+    }
+  else
```
cristiarg, line 1244: better return early in `if` and leave the rest of the function as it was before; less noisy for back-porting to `develop` and - although I don't like early returns - this is the custom in cubrid, it seems

src/storage/disk_manager.c
```diff
@@ -1366,6 +1388,12 @@ disk_rv_redo_format (THREAD_ENTRY * thread_p, const LOG_RCV * rcv)
   DKNSECTS nsect_diff;
   LOG_RCV copy_rcv = *rcv;
 
+  /* This redo recovery function is not called in either in ATS or PTS.
+   * - During recovery, REDO phase is skipped in ATS and no recovery process in PTS.
+   * - During replication, it is skipped in PTS because disk volume header page of newly
+   *   added volume is not loaded on PTS. */
+  assert (!(is_tran_server_with_remote_storage ()));
```
cristiarg, line 1395: Nit:
```suggestion
  assert (!is_tran_server_with_remote_storage ());
```

src/storage/disk_manager.c
cristiarg: Would it be a solution to enclose the relevant section in a sysop?
It is used in a few other places in disk manager module.

src/storage/disk_manager.c
cristiarg: Better assert to make intent explicit.

PR #3819, opened 2022-09-22
[LETS-458] Fix the hang when starting PTS if ATS is stopped

wrlawodms:
http://jira.cubrid.org/browse/LETS-458

The cause of the hang: When a page is requested with a target lsa by PTS, the target lsa set to the m_redo_lsa, which is the next one to be applied. PS waits for the replication to go past the target LSA and returns the page. If the last redo record is applied, the target lsa points to after the last log record and the PS can't go past it because no log record is generated without ATS.

**The solution:**
Two solutions are suggested. 
 The one is making PS wait for the target LSA, not past. However, In the case of ATS, the target lsa is the highest evicted LSA and it is expected to be applied.
 The other one is setting the target lsa by PTS to the last applied LSA, not the one that will be applied. This is picked.

**Implementation:**
- m_proccesed_lsa : before the m_redo_lsa. This is added on the atomic replicator on PTS, not the replicator on PS.
- a new contr argument, prev_lsa, for atomic_replicator: as a default value for the m_processed_lsa.
- get_highest_processed_lsa(): return m_processed_lsa instead of m_redo_lsa
- get_lowest_unapplied_lsa(): return m_redo_lsa instead of get_highest_processed_lsa() in the replicator. All others using m_redo_lsa but fetching page from PS uses this. Now that get_highest_processed_lsa() doesn't return m_redo_lsa, it should be changed to return m_redo_lsa. The get_lowest_unapplied_lsa() in replicator already returns value considering m_redo_lsa.

### DIFF
--- a/src/transaction/log_replication.cpp
+++ b/src/transaction/log_replication.cpp
@@ -533,7 +533,6 @@ namespace cublog
   log_lsa
   replicator::get_lowest_unapplied_lsa () const
   {
-    // TODO: needs to be refactored to work with the new replicators flavors
     if (m_parallel_replication_redo == nullptr)
       {
 	std::lock_guard<std::mutex> lockg (m_redo_lsa_mutex);
--- a/src/transaction/log_replication.cpp
+++ b/src/transaction/log_replication.cpp
@@ -66,12 +66,11 @@ namespace cublog
    * replicator - definition
    *********************************************************************/
 
-  replicator::replicator (const log_lsa &start_redo_lsa, const log_lsa &prev_redo_lsa, PAGE_FETCH_MODE page_fetch_mode,
+  replicator::replicator (const log_lsa &start_redo_lsa, PAGE_FETCH_MODE page_fetch_mode,
 			  int parallel_count)
     : m_bookkeep_mvcc { is_page_server () }
     , m_replicate_mvcc { is_passive_transaction_server () }
     , m_redo_lsa { start_redo_lsa }
-    , m_processed_lsa { prev_redo_lsa }
     , m_replication_active { true }
     , m_redo_context { NULL_LSA, page_fetch_mode, log_reader::fetch_mode::FORCE }
     , m_perfmon_redo_sync { PSTAT_REDO_REPL_LOG_REDO_SYNC }
@@ -288,7 +287,6 @@ namespace cublog
 	  // however, this would need one more mutex lock; therefore, suffice to do it here
 	  assert (m_replication_active);
 
-	  m_processed_lsa = m_redo_lsa;
 	  m_redo_lsa = header.forw_lsa;
 	}
 	if (m_parallel_replication_redo != nullptr)
@@ -523,8 +521,7 @@ namespace cublog
   log_lsa
   replicator::get_highest_processed_lsa () const
   {
-    std::lock_guard<std::mutex> lockg (m_redo_lsa_mutex);
-    return m_processed_lsa;
+    assert (false);
   }
 
   log_lsa
--- a/src/transaction/log_replication.cpp
+++ b/src/transaction/log_replication.cpp
@@ -521,6 +521,13 @@ namespace cublog
   log_lsa
   replicator::get_highest_processed_lsa () const
   {
+    /*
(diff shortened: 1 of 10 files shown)


src/transaction/log_replication.cpp
```diff
@@ -286,6 +288,7 @@ namespace cublog
 	  // however, this would need one more mutex lock; therefore, suffice to do it here
 	  assert (m_replication_active);
 
+	  m_processed_lsa = m_redo_lsa;
```
cristiarg: On PS, with the parallel replicator active - by default - one cannot say about the log record with `m_redo_lsa` that "it has been processed". Because the log record might have just been submitted to an async execution - see `read_and_redo_record` -> `log_rv_redo_record_sync_or_dispatch_async`.

Therefore, better to just restrict the `m_processed_lsa` to the atomic replicator class:
- leave `cublog::replicator` implementation untouched
- move `m_processed_lsa` to atomic replication
- mark function `get_highest_processed_lsa` as virtual in base class and override returning `m_processed_lsa` in derived class.
(PS: sorry if it's too much detail for you, my bad)

Since PTS replication is single-threaded (for now), atomic replicator can assume the log record for `m_processed_lsa` has been processed.
Although, if the log record is part of an atomic replication sequence, the log record is not really processed, but at least the page it applies to is fixed/blocked for access from other threads.

wrlawodms: Thank you for the detailed feedback :).

 With m_proccessed_lsa only in the atomic_replicator, I'm concerned that `get_highest_processed_lsa()` will return different value which has different meanings according to its class. The replicator will return m_redo_lsa, which will be processed, and the atomic replicator will return m_processed_lsa, which has been processed. I think the same interface should have the same semantics.

 On top of that, get_highest_processed_lsa() already returns the lsa that has not been processed actually (m_redo_lsa). So, I don't understand why we can't use m_processed_lsa instead of m_redo_lsa. If the name is the problem, we can rename the variable. Do I miss some points or misunderstand something?

cristiarg: You are correct that base and overridden functions should implement the same semantics.
And you are correct that the current implementation is - well - inconsistent to say the least.
What about this:
- `get_highest_processed_lsa` is only called on PTS; never on PS:
  - in `pgbuf_wait_for_replication` on PTS
  - in `pgbuf_read_page_from_file_or_page_server` only for PTS as well
- and still override `get_highest_processed_lsa` in atomic replicator and implement with the newly added member `m_processed_lsa` which will be correct
- this seems to be most constrained to the scenarios that are currently needed

```
  log_lsa
  replicator::get_highest_processed_lsa () const
  {
    // not to be called for this flavor of replicator (aka: on Page Server)
    assert (is_page_server ());
    assert_release (false);
    return MAX_LSA;
  }

  log_lsa
  replicator::get_lowest_unapplied_lsa () const
  {
    if (m_parallel_replication_redo == nullptr)
      {
	// sync
        std::lock_guard<std::mutex> lockg (m_redo_lsa_mutex);
        return m_redo_lsa;
      }

    // not to be called for this flavor of replicator (aka: on Page Server)
    assert_release (false);
  }
```

And I think that the entire `get_lowest_unapplied_lsa` should be asserted against being called as it is only used on PTS.

I'm not completely happy with this neither, but I hope it works for the time being.

The drawback of this is that we cannot use the non-atomic replicator anymore on PTS - when initializing in `passive_tran_server::start_log_replicator` - but I think we are past that stage as atomic replication is working (well, still in progress).

src/transaction/log_replication.cpp
```diff
@@ -530,8 +536,9 @@ namespace cublog
     // TODO: needs to be refactored to work with the new replicators flavors
     if (m_parallel_replication_redo == nullptr)
       {
+	std::lock_guard<std::mutex> lockg (m_redo_lsa_mutex);
```
cristiarg, line 538: remove TODO above

cristiarg (approved):

hornetmj (approved):

joohok (approved):

PR #3821, opened 2022-09-24
[LETS-449] Remove the disk_Page_server_perm_volume_count and block fileIO on PTS

joohok:
http://jira.cubrid.org/browse/LETS-449

Purpose
- Block fileIO on PTS which is not required in TS
- Remove disk_Page_server_perm_volume_count 
  - On ATS, this is used only for booting process
  - On PTS, this is not updated during replication.

more details are in Jira

### DIFF
--- a/src/server/tran_server.cpp
+++ b/src/server/tran_server.cpp
@@ -238,6 +238,9 @@ tran_server::init_page_server_hosts (const char *db_name)
   return exit_code;
 }
 
+
+/* TODO: Remove this boot process because the only information received from PS is the number of permanent volumes.
+ *       The number of permanent volumes is also stored in boot_Db_parm. So this process can be removed */
 int
 tran_server::get_boot_info_from_page_server ()
 {
--- a/src/server/tran_server.cpp
+++ b/src/server/tran_server.cpp
@@ -255,6 +255,9 @@ tran_server::get_boot_info_from_page_server ()
   DKNVOLS nvols_perm;
   std::memcpy (&nvols_perm, response_message.c_str (), sizeof (nvols_perm));
 
+  /* Check the dummay value whether the TS receives the message from PS (receive_boot_info_request) well. */
+  assert (nvols_perm == VOLID_MAX);
+
   return NO_ERROR;
 }
 
--- a/src/server/tran_server.cpp
+++ b/src/server/tran_server.cpp
@@ -238,9 +238,9 @@ tran_server::init_page_server_hosts (const char *db_name)
   return exit_code;
 }
 
-
-/* TODO: Remove this boot process because the only information received from PS is the number of permanent volumes.
- *       The number of permanent volumes is also stored in boot_Db_parm. So this process can be removed */
+/* NOTE : Since TS don't need the information about the number of permanent volume during boot,
+ *        this message has no actual use currently. However, this mechanism will be reserved,
+ *        because it can be used in the future when multiple PS's are supported. */
 int
 tran_server::get_boot_info_from_page_server ()
 {
--- a/src/server/tran_server.cpp
+++ b/src/server/tran_server.cpp
@@ -252,8 +252,6 @@ tran_server::get_boot_info_from_page_server ()
   DKNVOLS nvols_perm;
   std::memcpy (&nvols_perm, response_message.c_str (), sizeof (nvols_perm));
 
-  disk_set_page_server_perm_volume_count (nvols_perm);
-
   return NO_ERROR;
 }
 
--- a/src/storage/disk_manager.c
+++ b/src/storage/disk_manager.c
@@ -5884,8 +5884,14 @@ disk_is_valid_volid (VOLID volid)
     {
       /* Since PTS do not have a disk header page for newly added volume at the time of replication,
        * disk_rv_redo_format (), which requires disk header page and does update the disk_Cache,
-       * is not called during replication.
-       * Therefore, only the range of the volume ID is checked and the validity is checked by PS */
+       * is not called during replication.Thereefore, PTS knows only temporary volumes.
+       * Only the temporary volume ID is checked, and if it is not a temporary volume, only the range of volume id is checked */
+
+      const bool is_valid_temp_volid = (volid > (LOG_MAX_DBVOLID - disk_Cache->nvols_temp));
+      if (is_valid_temp_volid)
+	{
+	  return is_valid_temp_volid;
+	}
 
       return volid >= LOG_DBFIRST_VOLID && volid <= LOG_MAX_DBVOLID;
     }
--- a/src/storage/disk_manager.c
+++ b/src/storage/disk_manager.c
@@ -2660,6 +2660,11 @@ disk_cache_load_all_volumes (THREAD_ENTRY * thread_p)
   assert (disk_Cache != NULL);
   if (is_tran_server_with_remote_storage ())
     {
+      /* TODO: disk_Cache is not required to be initialized on PTS.
+       *       if all the disk_Cache usage parts are blocked on PTS,
+       *       then disk_Cache initialization on PTS can be blocked too.
+       */
+
       const DKNVOLS nvols_perm = xboot_find_number_permanent_volumes (thread_p);
       assert (nvols_perm > 0);
 
@@ -5884,14 +5889,8 @@ disk_is_valid_volid (VOLID volid)
     {
       /* Since PTS do not have a disk header page for newly added volume at the time of replication,
        * disk_rv_redo_format (), which requires disk header page and does update the disk_Cache,
-       * is not called during replication. Thereefore, PTS knows only temporary volumes.
-       * Only the temporary volume ID is checked, and if it is not a temporary volume, only the range of volume id is checked */
-
-      const bool is_valid_temp_volid = (volid > (LOG_MAX_DBVOLID - disk_Cache->nvols_temp));
-      if (is_valid_temp_volid)
-	{
-	  return is_valid_temp_volid;
-	}
+       * is not called during replication. Therefore, PTS knows only temporary volumes.
+       * So, it will check only the range of the volume id on PTS */
 
       return volid >= LOG_DBFIRST_VOLID && volid <= LOG_MAX_DBVOLID;
     }
--- a/src/storage/disk_manager.c
+++ b/src/storage/disk_manager.c
@@ -5884,7 +5884,7 @@ disk_is_valid_volid (VOLID volid)
     {
       /* Since PTS do not have a disk header page for newly added volume at the time of replication,
        * disk_rv_redo_format (), which requires disk header page and does update the disk_Cache,
-       * is not called during replication.Thereefore, PTS knows only temporary volumes.
+       * is not called during replication. Thereefore, PTS knows only temporary volumes.
        * Only the temporary volume ID is checked, and if it is not a temporary volume, only the range of volume id is checked */
 
       const bool is_valid_temp_volid = (volid > (LOG_MAX_DBVOLID - disk_Cache->nvols_temp));
@@ -5896,13 +5896,7 @@ disk_is_valid_volid (VOLID volid)
       return volid >= LOG_DBFIRST_VOLID && volid <= LOG_MAX_DBVOLID;
     }
 
-  const bool is_valid_perm_volid = (volid < disk_Cache->nvols_perm);
-  if (is_valid_perm_volid)
-    {
-      return is_valid_perm_volid;
-    }
-
-  return volid > (LOG_MAX_DBVOLID - disk_Cache->nvols_temp);
+  return volid < disk_Cache->nvols_perm || volid > LOG_MAX_DBVOLID - disk_Cache->nvols_temp;
 }
 
 /*
--- a/src/storage/disk_manager.c
+++ b/src/storage/disk_manager.c
@@ -5889,8 +5889,10 @@ disk_is_valid_volid (VOLID volid)
     {
       /* Since PTS do not have a disk header page for newly added volume at the time of replication,
        * disk_rv_redo_format (), which requires disk header page and does update the disk_Cache,
-       * is not called during replication. Therefore, PTS knows only temporary volumes.
-       * So, it will check only the range of the volume id on PTS */
+       * is not called during replication. Therefore, PTS knows only temporary volumes,
+       * and it stores the temporary volume info in disk_Cache.
+       * So, it will check only the range of the volume id on PTS here,
+       * and actual validation for permananent volume will be done on PS when PTS request a page from PS. */
 
       return volid >= LOG_DBFIRST_VOLID && volid <= LOG_MAX_DBVOLID;
     }
--- a/src/storage/disk_manager.c
+++ b/src/storage/disk_manager.c
@@ -5880,7 +5880,7 @@ disk_dump_goodvol_all (THREAD_ENTRY * thread_p, INT16 volid, void *arg)
 STATIC_INLINE bool
 disk_is_valid_volid (VOLID volid)
 {
-  if (is_tran_server_with_remote_storage () && is_passive_transaction_server ())
+  if (is_passive_transaction_server ())
     {
       /* Since PTS do not have a disk header page for newly added volume at the time of replication,
        * disk_rv_redo_format (), which requires disk header page and does update the disk_Cache,
--- a/src/storage/disk_manager.c
+++ b/src/storage/disk_manager.c
@@ -205,9 +205,6 @@ struct disk_cache
 };
 
 static DISK_CACHE *disk_Cache = NULL;
-static DKNVOLS disk_Page_server_perm_volume_count = 0;	// used for transaction server with remote storage
-							// must be maintained during replication on
-							// passive transaction server (with remote storage)
 
 static DKNSECTS disk_Temp_max_sects = -2;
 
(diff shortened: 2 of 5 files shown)


src/server/tran_server.cpp
```diff
@@ -252,8 +252,6 @@ tran_server::get_boot_info_from_page_server ()
   DKNVOLS nvols_perm;
   std::memcpy (&nvols_perm, response_message.c_str (), sizeof (nvols_perm));
```
hornetmj, line 256: What is the purpose of this function after modification?

src/storage/disk_manager.c
```diff
@@ -5917,19 +5880,22 @@ disk_dump_goodvol_all (THREAD_ENTRY * thread_p, INT16 volid, void *arg)
...
-  const bool is_valid_perm_volid = is_tran_server_with_remote_storage ()?
-    (volid < disk_Page_server_perm_volume_count) : (volid < disk_Cache->nvols_perm);
+  if (is_tran_server_with_remote_storage () && is_passive_transaction_server ())
+    {
+      /* Since PTS do not have a disk header page for newly added volume at the time of replication,
+       * disk_rv_redo_format (), which requires disk header page and does update the disk_Cache,
+       * is not called during replication.
+       * Therefore, only the range of the volume ID is checked and the validity is checked by PS */
+
+      return volid >= LOG_DBFIRST_VOLID && volid <= LOG_MAX_DBVOLID;
```
hornetmj, line 5897: I think that the temp temp volid need to validate correctly on PTS. Because they locate on the local disk.

src/server/tran_server.cpp
joohok: Nothing.
After this PR is merged, it needs to be removed. 
This is the process of getting boot information from the page server, and I thought that cutting out this part was a different topic, so I was going to proceed with the next PR.

src/storage/disk_manager.c
joohok: OK, I agree.

src/storage/disk_manager.c
```diff
@@ -5917,19 +5880,28 @@ disk_dump_goodvol_all (THREAD_ENTRY * thread_p, INT16 volid, void *arg)
...
 disk_is_valid_volid (VOLID volid)
 {
-  // a passive transaction server (which is, implicitly, a transaction server
-  // with remote storage) maintains a separate bookkeeping of permanent data volumes
-  // which is updated when executing corresponding recovery replication functions
-  // as part of the regular replication; this way, it is able to provide a valid
-  // answer via this function
-  const bool is_valid_perm_volid = is_tran_server_with_remote_storage ()?
-    (volid < disk_Page_server_perm_volume_count) : (volid < disk_Cache->nvols_perm);
+  if (is_tran_server_with_remote_storage () && is_passive_transaction_server ())
```
cristiarg: second function implies the first; so only first condition is needed

src/storage/disk_manager.c
```diff
@@ -5917,19 +5880,28 @@ disk_dump_goodvol_all (THREAD_ENTRY * thread_p, INT16 volid, void *arg)
...
+    {
+      /* Since PTS do not have a disk header page for newly added volume at the time of replication,
+       * disk_rv_redo_format (), which requires disk header page and does update the disk_Cache,
+       * is not called during replication.Thereefore, PTS knows only temporary volumes.
```
cristiarg:
```suggestion
       * is not called during replication. Therefore, PTS knows only temporary volumes.
```

src/storage/disk_manager.c
```diff
@@ -5917,19 +5880,28 @@ disk_dump_goodvol_all (THREAD_ENTRY * thread_p, INT16 volid, void *arg)
...
+      const bool is_valid_temp_volid = (volid > (LOG_MAX_DBVOLID - disk_Cache->nvols_temp));
+      if (is_valid_temp_volid)
+	{
+	  return is_valid_temp_volid;
+	}
+
+      return volid >= LOG_DBFIRST_VOLID && volid <= LOG_MAX_DBVOLID;
+    }
+
+  const bool is_valid_perm_volid = (volid < disk_Cache->nvols_perm);
```
cristiarg: Maybe make this check the same as in develop since it serves the "old" purpose - for the "monolithic" server and will make back-merging easier.
```suggestion
    return volid < disk_Cache->nvols_perm || volid > LOG_MAX_DBVOLID - disk_Cache->nvols_temp;
```

joohok: OK, Thanks. applied

src/storage/disk_manager.c
joohok: OK. Applied.

src/storage/disk_manager.c
joohok: Right. However, I know that the first condition includes ATS and PTS.

ATS doesn't have to come into this part, because ATS updates disk_Cache well every time.
So, I think it's better to use only the second condition.

hornetmj: @joohok As I know, the result of the is_passive_transaction_server() function is not guaranteed to use remote storage.

joohok: @hornetmj  Yes, you are right. 
I was confused for a while whether PTS could operate as a single node or not.
I think It is better to leave this condition as it is.

src/server/tran_server.cpp
```diff
@@ -252,8 +252,6 @@ tran_server::get_boot_info_from_page_server ()
...
 
-  disk_set_page_server_perm_volume_count (nvols_perm);
```
wrlawodms, line 255: Even without this, is the caller, `tran_server::get_boot_info_from_page_server()` still needed?

tran_to_page_request::GET_BOOT_INFO message and communication also don't seem to be needed.

src/storage/disk_manager.c
```diff
@@ -2699,8 +2660,10 @@ disk_cache_load_all_volumes (THREAD_ENTRY * thread_p)
   assert (disk_Cache != NULL);
   if (is_tran_server_with_remote_storage ())
     {
-      assert (disk_Page_server_perm_volume_count > 0);
-      for (VOLID volid = 0; volid < disk_Page_server_perm_volume_count; ++volid)
+      const DKNVOLS nvols_perm = xboot_find_number_permanent_volumes (thread_p);
+      assert (nvols_perm > 0);
+
+      for (VOLID volid = 0; volid < nvols_perm; ++volid)
```
wrlawodms, line 2671: From what I heard in person, the disk_Cache on PTS doesn't know about permanent volumes. Why should we initialize it using disk_cache_load_volume()?

src/storage/disk_manager.c
hornetmj: @joohok The passive_tran_server::uses_remote_storage() always returns true.

joohok: Yes, I reduce the number of conditions. Thanks

src/storage/disk_manager.c
```diff
@@ -5917,20 +5880,23 @@ disk_dump_goodvol_all (THREAD_ENTRY * thread_p, INT16 volid, void *arg)
...
+       * is not called during replication. Thereefore, PTS knows only temporary volumes.
+       * Only the temporary volume ID is checked, and if it is not a temporary volume, only the range of volume id is checked */
+
+      const bool is_valid_temp_volid = (volid > (LOG_MAX_DBVOLID - disk_Cache->nvols_temp));
+      if (is_valid_temp_volid)
+	{
+	  return is_valid_temp_volid;
+	}
+
+      return volid >= LOG_DBFIRST_VOLID && volid <= LOG_MAX_DBVOLID;
```
wrlawodms, line 5897: Whether `is_valid_temp_volid` is true or not, this is always true, if the volid is in the valid range. It seems to me that the `is_valid_temp_volid` does nothing.

src/storage/disk_manager.c
joohok: We don't need to initialize disk_Cache on PTS. Yes, it only needs to be initialized on ATS. 
But for now, in this PR, I am going to leave it as it is. 
Because all the disk_Cache usage parts are not blocked in the current PTS, so I tried to fix it after blocking all the disk_Cache usage parts in the next PRs.

I will try to handle the things to be done later in the next PRs. (described TODO section in comment) 
1. Remove the part that received boot_info from PS to set the disk_Page_server_perm_volume_count
2. Block disk_Cache related operations in PTS
3. In PTS, ID of permanent volume and ID of temporary volume should not overlap.

src/transaction/log_replication_atomic.cpp
```diff
@@ -83,17 +83,11 @@ namespace cublog
...
-	  {
-	    m_redo_context.m_reader.advance_when_does_not_fit (sizeof (LOG_REC_DBOUT_REDO));
-	    const LOG_REC_DBOUT_REDO dbout_redo =
-		    m_redo_context.m_reader.reinterpret_copy_and_add_align<LOG_REC_DBOUT_REDO> ();
-	    log_rcv rcv;
-	    rcv.length = dbout_redo.length;
-
-	    log_rv_redo_record (&thread_entry, m_redo_context.m_reader, RV_fun[dbout_redo.rcvindex].redofun, &rcv,
-				&m_redo_lsa, 0, nullptr, m_redo_context.m_redo_zip);
+	    /* Recovery redo for RVDK_NEWVOL and RVDK_EXPAND_VOLUME will be called here,
```
wrlawodms: If you want to skip all redo replication for LOG_DBEXTERN_REDO_DATA. I think it'd be better to leave a comment where `LOG_DBEXTERN_REDO_DATA` is defined: will be not replicated. Although now there are only 3 recv-idx and they are not needed for replication, someone will possibly add a recvindex. Just a suggestion.

wrlawodms: By the way,  is it OK to skip `RVLOG_OUTSIDE_LOGICAL_REDO_NOOP` too?

src/server/tran_server.cpp
joohok: No, the caller becomes useless. 

After this PR is merged, it needs to be removed.
This is the process of getting boot information from the page server, and I thought that removing this part was a different topic, so I was going to proceed with the next PR.

src/storage/disk_manager.c
joohok: OK, I understand. we don't need to check `is_valid_temp_volid`, because the condition of checking the range of volume id will cover it.

src/server/tran_server.cpp
wrlawodms: Oh, I just see the previous resolved conversation. Please make the issue and ignore this.

src/transaction/log_replication_atomic.cpp
joohok: > If you want to skip all redo replication for LOG_DBEXTERN_REDO_DATA. I think it'd be better to leave a comment where LOG_DBEXTERN_REDO_DATA is defined: will be not replicated. Although now there are only 3 recv-idx and they are not needed for replication, someone will possibly add a recvindex. Just a suggestion.

Thanks for suggestion. I think it is better. 

> By the way, is it OK to skip RVLOG_OUTSIDE_LOGICAL_REDO_NOOP too?

recovery redo function for `RVLOG_OUTSIDE_LOGICAL_REDO_NOOP` (log_rv_outside_noop_redo ()) does nothing but returns NO_ERROR during replication. So I skipped this.. but do you know something about this recovery index ?

src/storage/disk_manager.c
wrlawodms: Are you going to work on the following PRs right after this PR? If so, it's Ok. Or, I think at least we should leave a TODO comment.

src/transaction/log_replication_atomic.cpp
wrlawodms: No, I don't know too.

src/storage/disk_manager.c
wrlawodms: I mean not for all TODOs, just for TODO related to what is modified in this PR including get_boot_info_from_page_server().

wrlawodms (approved):

src/storage/disk_manager.c
joohok: I agree, I will leave an TODO comment for it.

src/transaction/log_replication_atomic.cpp
joohok: If so, it seems to be better to block only the rcvindex that does fileio inside. (RVDK_NEWVOL, RVDK_EXPAND_VOLUME).

src/server/tran_server.cpp
```diff
@@ -238,6 +238,9 @@ tran_server::init_page_server_hosts (const char *db_name)
   return exit_code;
 }
 
+
+/* TODO: Remove this boot process because the only information received from PS is the number of permanent volumes.
+ *       The number of permanent volumes is also stored in boot_Db_parm. So this process can be removed */
```
cristiarg: I am skeptical to remove all the code (`tran_server::get_boot_info_from_page_server`, `tran_to_page_request::GET_BOOT_INFO`, `page_server::connection_handler::receive_boot_info_request`, `disk_get_perm_volume_count`) since it might come handy at some future time.
We might find uses for this "boot" message when multiple PS's are supported.
I would say, just remove the volume count from the return message. And replace with returning a dummy value to be asserted on the transaction server. And add comment on both the originating and the receiving side that the message has no actual use currently.

Just a thought.

src/storage/disk_manager.c
```diff
@@ -5917,20 +5885,17 @@ disk_dump_goodvol_all (THREAD_ENTRY * thread_p, INT16 volid, void *arg)
...
-  const bool is_valid_perm_volid = is_tran_server_with_remote_storage ()?
-    (volid < disk_Page_server_perm_volume_count) : (volid < disk_Cache->nvols_perm);
-  if (is_valid_perm_volid)
+  if (is_passive_transaction_server ())
     {
-      return is_valid_perm_volid;
+      /* Since PTS do not have a disk header page for newly added volume at the time of replication,
+       * disk_rv_redo_format (), which requires disk header page and does update the disk_Cache,
+       * is not called during replication. Therefore, PTS knows only temporary volumes.
+       * So, it will check only the range of the volume id on PTS */
```
cristiarg: Please complete the comment that, for permanent volume, the check is "abstained" here and delegated to the actual request that will be later made towards retrieving the heap page from PS (since all of the times the function is called in PTS context is for volume type purpose - `disk_get_voltype` or `disk_get_volpurpose`).

cristiarg (commented):
I still do not understand how the following problem is solved.
On the call path
```
pgbuf_read_page_from_file_or_page_server
  -> pgbuf_is_temporary_volume
  -> xdisk_get_purpose
  -> disk_get_volpurpose
```
the disk cache is still being used.
How will the decision between reading pages of temporary volumes from PTS's local storage and reading pages of permanent volumes from PS will be differentiated.
The current changes are ok, but - maybe - incomplete.
The above call-path works as long as the database works with the initial volumes - it seems that the `disk_Cache` is - somewhere - initialized - maybe from `boot_Db_parm` (orr by some other method).
But once a new volume is added, the disk_Cache will not be updated.
Or am I wrong?

The way I understand is that a new fork in call-path is needed downstream from `pgbuf_read_page_from_file_or_page_server` that, for PTS, will only check whether the volume id a temporary one. If not temporary, the check whether it is a valid permanent volume is delegated to the call on PS (and maybe, on PS, it is needed to augment the check - if no check is already in place on the callpath starting with `pgbuf_respond_data_fetch_page_request`).

src/server/tran_server.cpp
joohok: Since it can be used later, do you mean to maintain the structure of sending and receiving boot info messages? 
OK, I agree. Then I just add a comment for it.

joohok:
> I still do not understand how the following problem is solved. On the call path
> 
> ```
> pgbuf_read_page_from_file_or_page_server
>   -> pgbuf_is_temporary_volume
>   -> xdisk_get_purpose
>   -> disk_get_volpurpose
> ```
> 
> the disk cache is still being used. How will the decision between reading pages of temporary volumes from PTS's local storage and reading pages of permanent volumes from PS will be differentiated. The current changes are ok, but - maybe - incomplete. The above call-path works as long as the database works with the initial volumes - it seems that the `disk_Cache` is - somewhere - initialized - maybe from `boot_Db_parm` (orr by some other method). But once a new volume is added, the disk_Cache will not be updated. Or am I wrong?

Right. Permanent volume information in disk_Cache is not updated on PTS. Therefore, when a new volume is added and disk_Cache->vols[volid].purpose is searched with the ID of the added volume, DISK_UNKNOWN_PURPOSE will be returned. (disk_get_volpurpose). 
However, since the temporary volume information in disk_Cache can be updated on PTS, DB_TEMPORARY_DATA_PURPOSE will be returned properly if disk_Cache->vols[volid].purpose is searched by the ID of the temporary volume.  
Since pgbuf_is_temporary_volume() only matters whether the purpose of the volume is temporary or not, I think there is no problem.

> 
> The way I understand is that a new fork in call-path is needed downstream from `pgbuf_read_page_from_file_or_page_server` that, for PTS, will only check whether the volume id a temporary one. If not temporary, the check whether it is a valid permanent volume is delegated to the call on PS (and maybe, on PS, it is needed to augment the check - if no check is already in place on the callpath starting with `pgbuf_respond_data_fetch_page_request`).

I thought the above method was sufficient. This is because if pgbuf_is_temporary_volume returns a value other than DB_TEMPORARY_DATA_PURPOSE (DB_PERMANENT_DATA_PURPOSE or UNKNOWN, most likely returns UNKNOWN if there are many newly added volumes), PTS will request a page from PS.

src/storage/disk_manager.c
joohok: OK, I agree. I will add some more comment.

cristiarg (approved):

hornetmj (approved):

PR #3822, opened 2022-09-26
[CBRD-24485] Refactoring create_or_drop_index_helper()

ctshim:
http://jira.cubrid.org/browse/CBRD-24485

* create_or_drop_index_helper():  Rewrite with fewer arguments.
* smt_add_index() :   The body of the smt_add_index() function does not exist, but it is declared externally, so it is deleted.

### DIFF
--- a/src/query/execute_schema.c
+++ b/src/query/execute_schema.c
@@ -2777,7 +2777,7 @@ create_or_drop_index_helper (PARSER_CONTEXT * parser, const char *const constrai
 	  if (idx_info->column_names->info.sort_spec.expr)
 	    {
 	      char const *colname = idx_info->column_names->info.sort_spec.expr->info.name.original;
-	      if (sm_att_unique_constrained (obj, colname) || sm_att_fk_constrained (obj, colname))
+	      if (colname && (sm_att_unique_constrained (obj, colname) || sm_att_fk_constrained (obj, colname)))
 		{
 		  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_SM_INDEX_PREFIX_LENGTH_ON_UNIQUE_FOREIGN, 0);
 		  return ER_SM_INDEX_PREFIX_LENGTH_ON_UNIQUE_FOREIGN;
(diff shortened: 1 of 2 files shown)


src/query/execute_schema.c
```diff
@@ -2774,47 +2757,49 @@ create_or_drop_index_helper (PARSER_CONTEXT * parser, const char *const constrai
...
+
+  char *attname_tmp = NULL;
+  if (do_index != DO_INDEX_CREATE)
     {
-      assert (comment->node_type == PT_VALUE);
-      comment_str = (char *) PT_VALUE_GET_BYTES (comment);
+      assert (constraint_name != NULL);
+      nnames = 0;
+      attnames = &attname_tmp;
+      attnames[0] = NULL;
```
beyondykk9, line 2768: it looks not necessary. it is already NULL.

src/query/execute_schema.c
```diff
@@ -2774,47 +2757,49 @@ create_or_drop_index_helper (PARSER_CONTEXT * parser, const char *const constrai
...
+      assert (idx_info);
+      nnames = pt_length_of_list (idx_info->column_names);
+
+      if (nnames == 1 && idx_info->prefix_length)
 	{
-	  colname = n->info.name.original;
+	  if (idx_info->column_names->info.sort_spec.expr)
+	    {
+	      char const *colname = idx_info->column_names->info.sort_spec.expr->info.name.original;
+	      if (sm_att_unique_constrained (obj, colname) || sm_att_fk_constrained (obj, colname))
```
beyondykk9: missed to check if colname is NULL ?

src/query/execute_schema.c
ctshim: It is put in such a way that developers who read the code are clearly aware that the names[0] should be NULL.
Excluding "nnames = 0;" and "attnames [0] = NULL;" is the same, but we want to keep it as it is for readability.

src/query/execute_schema.c
ctshim: Thank you, I missed it.

beyondykk9 (approved):

hgryoo (approved):

shparkcubrid (approved):

hyunikn (approved):

youngjinj (approved):

PR #3824, opened 2022-09-28
[LETS-525] Fixed the core occurred during recovery on PTS

joohok:
http://jira.cubrid.org/browse/LETS-525

Purpose

Added information related to mvcc that is not initialized at the time of commit.
The mvccid of a specific transaction is initialized, but if the address of the last mvcc log record (last_mvcc_lsa) is not initialized, a core occurs when performing recovery in PTS.
This is because PTS sees that the mvcc log of the transaction exists, but there is no mvccid.
more details are in jira.

### DIFF
--- a/src/transaction/log_tran_table.c
+++ b/src/transaction/log_tran_table.c
@@ -4157,6 +4157,8 @@ logtb_complete_mvcc (THREAD_ENTRY * thread_p, LOG_TDES * tdes, bool committed)
 
   logtb_tran_clear_update_stats (&tdes->log_upd_stats);
 
+  tdes->last_mvcc_lsa.set_null ();
+
   if (is_perf_tracking)
     {
       tsc_getticks (&end_tick);
--- a/src/transaction/log_tran_table.c
+++ b/src/transaction/log_tran_table.c
@@ -4133,9 +4133,13 @@ logtb_complete_mvcc (THREAD_ENTRY * thread_p, LOG_TDES * tdes, bool committed)
 	  log_append_assigned_mvccid (thread_p, mvccid);
 	}
       mvcc_table->complete_mvcc (tran_index, mvccid, committed);
+
+      tdes->last_mvcc_lsa.set_null ();
     }
   else
     {
+      assert (tdes->last_mvcc_lsa.is_null ());
+
       if (committed && logtb_tran_update_all_global_unique_stats (thread_p) != NO_ERROR)
 	{
 	  assert (false);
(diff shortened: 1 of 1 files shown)


src/transaction/log_tran_table.c
```diff
@@ -4157,6 +4157,8 @@ logtb_complete_mvcc (THREAD_ENTRY * thread_p, LOG_TDES * tdes, bool committed)
 
   logtb_tran_clear_update_stats (&tdes->log_upd_stats);
 
+  tdes->last_mvcc_lsa.set_null ();
```
cristiarg: Move the line in the `if (MVCCID_IS_VALID (mvccid))` branch and assert that it is null on the else - mvcc not valid - branch.
Also, shouldn't this change be added to all the places where `mvccinfo.reset ()` is called, in:
- `logtb_initialize_system_tdes`
- `logtb_release_tran_index`

joohok: I agree with your first suggestion. I will move it and add assert. 

> logtb_initialize_system_tdes
logtb_release_tran_index

logtb_clear_tdes (), which initialize the last_mvcc_lsa, is called in those functions.

cristiarg: Agree. Indeed, it is called in the first one.

joohok: For logtb_release_tran_index (), 
```
logtb_release_tran_index () 
{
      tdes->mvccinfo.reset ()
      TR_TABLE_CS_ENTER () 
      logtb_free_tran_index ()
         -> logtb_clear_tdes ()
      TR_TABLE_CS_EXIT ()
}
```
For logtb_initialize_system_tdes ()
```
logtb_initialize_system_tdes ()
{
	logtb_clear_tdes ()
	tdes->mvccinfo.reset ()
}
```

I don't think it needs to be added to above places where mvccinfo.reset () is called.

However, last_mvcc_lsa seems to be used whenever dealing with mvcc information, so I wonder if it would be better to move last_mvcc_lsa to tdes->mvccinfo and initialize it together in the mvccinfo.reset() function. Not now, it is just thought.

cristiarg (approved):

src/transaction/log_tran_table.c
wrlawodms: I agree with @joohok that it should be packed together if it's used and initialized together all the time.

wrlawodms (approved):

src/transaction/log_tran_table.c
cristiarg: @joohok
I also thought of this, but did not voice my concern.
Then let's do it now.
Either with this patch, or a new patch on the same issue.
Or, if you prefer, add a new Jira issue.
If this came up as a possible improvement, I don't want to lose track of it.

joohok: OK, I will do it with next patch on the same issue.

hornetmj (approved):

PR #3825, opened 2022-09-28
[LETS-537] refactor atomic replicator helper page bookkeeping

cristiarg:

Refactored internal bookkeeping of an atomic replication sequences to allow for atomic sequences where multiple log records refer to the same VPID.
Thus, it is needed for a "reference counting"-like functionality to be implemented such that the page is only unfixed when the last log record from a sequence that affects that page is applied and done.

Implementation:
- previously a bookkeeping functionality - but without reference counting - was implemented with the help of `vpid_to_page_ptr_map_type m_page_map`
- the functionality of this map was taken over by a new class `page_ptr_bookkeeping` which is instantiated in each atomic replication sequence instance
- this bookkeeping class knows everything about a page: vpid, recovery index, fix type (normal or ordered fix), reference count
- the only result that the class outputs is the page pointer - via the `fix_page`, `unfix_page` functions
- extra page pointer related information that existed in `atomic_log_entry` class was moved to a helper structure - `page_ptr_info` - used by the bookkeeping logic

Other:
- extracted the actual implementation of the fix/unfix in two stand-alone functions - `pgbuf_fix_or_ordered_fix` and `pgbuf_unfix_or_ordered_unfix`

### DIFF
--- a/src/transaction/atomic_replication_helper.cpp
+++ b/src/transaction/atomic_replication_helper.cpp
@@ -268,6 +268,7 @@ namespace cublog
   atomic_replication_helper::atomic_log_sequence::add_atomic_replication_log (THREAD_ENTRY *thread_p,
       log_lsa record_lsa, LOG_RCVINDEX rcvindex, VPID vpid)
   {
+#if (0)
     m_log_vec.emplace_back (record_lsa, vpid, rcvindex);
     const auto page_map_it = m_page_map.find (vpid);
     if (page_map_it == m_page_map.cend ())
@@ -297,6 +298,23 @@ namespace cublog
 	m_log_vec.back ().set_page_ptr (page_map_it->second);
       }
     return NO_ERROR;
+#endif
+    PAGE_PTR page_p = nullptr;
+    // bookkeeping fixes page, keeps all info regarding how the page was fixed (either
+    // regular fix or ordered fix) and only returns back the pointer to the page
+    const int err_code = m_page_ptr_bookkeeping.fix_page (thread_p, vpid, rcvindex, page_p);
+    if (err_code != NO_ERROR)
+      {
+	assert (page_p == nullptr);
+	return err_code;
+      }
+    assert (page_p != nullptr);
+
+    m_log_vec.emplace_back (record_lsa, vpid, rcvindex, page_p);
+    //atomic_log_entry &last = m_log_vec.back ();
+    //last.set_page_ptr (page_p);
+
+    return err_code;
   }
 
   bool
@@ -363,6 +381,7 @@ namespace cublog
     return false;
   }
 
+#if (0)
   void
   atomic_replication_helper::atomic_log_sequence::apply_all_log_redos (THREAD_ENTRY *thread_p)
   {
@@ -371,6 +390,7 @@ namespace cublog
 	m_log_vec[i].apply_log_redo (thread_p, m_redo_context);
       }
   }
+#endif
 
   void
   atomic_replication_helper::atomic_log_sequence::apply_and_unfix_sequence (THREAD_ENTRY *thread_p)
@@ -379,6 +399,7 @@ namespace cublog
     // they come to be read by the PTS and some might be unfixed and refixed after the apply procedure
     // leading to inconsistency. To avoid this situation we sequentially apply each log redo of the sequence
     // when the end sequence log appears and the entire sequence is fixed
+#if (0)
     apply_all_log_redos (thread_p);
 
     for (size_t i = 0; i < m_log_vec.size (); i++)
@@ -389,6 +410,13 @@ namespace cublog
 	    m_log_vec[i].unfix_page (thread_p);
 	    m_page_map.erase (page_map_it);
 	  }
+      }
+#endif
+    for (size_t i = 0; i < m_log_vec.size (); i++)
+      {
+	m_log_vec[i].apply_log_redo (thread_p, m_redo_context);
+	// bookkeeping actually either
+	m_page_ptr_bookkeeping.unfix_page (thread_p, m_log_vec[i].m_vpid);
       }
   }
 
@@ -403,23 +431,29 @@ namespace cublog
    *********************************************************************************************************/
 
   atomic_replication_helper::atomic_log_sequence::atomic_log_entry::atomic_log_entry (
-	  log_lsa lsa, VPID vpid, LOG_RCVINDEX rcvindex)
+	  log_lsa lsa, VPID vpid, LOG_RCVINDEX rcvindex, PAGE_PTR page_ptr)
     : m_vpid { vpid }
     , m_record_lsa { lsa }
     , m_record_index { rcvindex }
-    , m_page_ptr { nullptr }
+    , m_page_ptr { page_ptr }
+#if (0)
     , m_watcher_p { nullptr }
+#endif
   {
     assert (lsa != NULL_LSA);
   }
 
   atomic_replication_helper::atomic_log_sequence::atomic_log_entry::atomic_log_entry (atomic_log_entry &&that)
-    : atomic_log_entry (that.m_record_lsa, that.m_vpid, that.m_record_index)
+    : atomic_log_entry (that.m_record_lsa, that.m_vpid, that.m_record_index, that.m_page_ptr)
   {
+#if (0)
     std::swap (m_page_ptr, that.m_page_ptr);
     std::swap (m_watcher_p, that.m_watcher_p);
+#endif
+    that.m_page_ptr = nullptr;
   }
 
+#if (0)
   atomic_replication_helper::atomic_log_sequence::atomic_log_entry::~atomic_log_entry ()
   {
     if (m_watcher_p != nullptr)
@@ -428,6 +462,7 @@ namespace cublog
 	m_watcher_p.reset ();
       }
   }
+#endif
 
   void
   atomic_replication_helper::atomic_log_sequence::atomic_log_entry::apply_log_redo (
@@ -471,6 +506,7 @@ namespace cublog
       }
   }
 
+#if (0)
   int
   atomic_replication_helper::atomic_log_sequence::atomic_log_entry::fix_page (THREAD_ENTRY *thread_p)
   {
@@ -516,6 +552,7 @@ namespace cublog
       }
 
     return NO_ERROR;
+    //return pgbuf_fix_or_ordered_fix (thread_p, m_vpid, m_record_index, m_watcher_p, m_page_ptr);
   }
 
   void
@@ -541,6 +578,7 @@ namespace cublog
 	pgbuf_unfix (thread_p, m_page_ptr);
 	break;
       }
+    //pgbuf_unfix_or_ordered_unfix (thread_p, m_record_index, m_watcher_p, m_page_ptr);
   }
 
   PAGE_PTR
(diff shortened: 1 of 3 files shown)


src/transaction/atomic_replication_helper.hpp
```diff
@@ -169,36 +166,85 @@ namespace cublog
...
+
+	    public: // methods
+	      int fix_page (THREAD_ENTRY *thread_p, VPID vpid, LOG_RCVINDEX rcv_index, PAGE_PTR &page_ptr_out);
+	      int unfix_page (THREAD_ENTRY *thread_p, VPID vpid);
+
+	    private: // types
+	      using page_ptr_info_map_type = std::map<VPID, page_ptr_info>;
+
+	    private: // variables
+	      page_ptr_info_map_type m_;
```
joohok: It would be better to change it to a more meaningful name. Such as `m_page_ptr_info_map`

src/transaction/atomic_replication_helper.cpp
```diff
@@ -533,32 +628,15 @@ namespace cublog
       case RVHF_UPDATE:
       case RVHF_MVCC_UPDATE_OVERFLOW:
       case RVHF_INSERT_NEWHOME:
-	// sanity asserts inside the function
-	pgbuf_ordered_unfix (thread_p, m_watcher_p.get ());
+	assert (page_ptr == nullptr);
```
joohok, line 640: I think `RVHF_UPDATE_NOTIFY_VACUUM` should be added as a case here.
`RVHF_UPDATE_NOTIFY_VACUUM` is a recovery index for mvcc update. 
```suggestion
      case RVHF_UPDATE_NOTIFY_VACUUM:
	assert (page_ptr == nullptr);
```

src/transaction/atomic_replication_helper.cpp
```diff
@@ -487,29 +578,32 @@ namespace cublog
       case RVHF_MVCC_UPDATE_OVERFLOW:
       case RVHF_INSERT_NEWHOME:
       {
-	assert (m_watcher_p == nullptr);
-	m_watcher_p.reset (new PGBUF_WATCHER ());
+	assert (watcher_uptr == nullptr);
```
joohok, line 589: I think `RVHF_UPDATE_NOTIFY_VACUUM` should be added as a case here too.

src/transaction/atomic_replication_helper.cpp
joohok: Related issue : http://jira.cubrid.org/browse/LETS-422

src/transaction/atomic_replication_helper.cpp
```diff
@@ -268,35 +269,35 @@ namespace cublog
...
-	  }
-	m_page_map.emplace (vpid,  m_log_vec.back ().get_page_ptr ());
+	// failing to fix the page just leaves it unfixed and does not affect overall
+	// functioning of the atomic replication sequence;
+	er_log_debug (ARG_FILE_LINE, "[ATOMIC_REPL] heap page cannot be fixed, cannot add new log record"
+		      "with LSA %lld|%d to atomic sequences started at LSA %lld|%d\n",
+		      LSA_AS_ARGS (&record_lsa), LSA_AS_ARGS (&m_start_lsa));
+
+	// TODO:
+	//  - what happens if there is more than one log record pertaining to the same page
```
razvanradulescu, line 286: Applying only a subset of log records of an atomic change, does leave the DB in an inconsistent state.
Can we manage this case by force evacuating the page from page buffer (and then the replication or user transactions will trigger a  fetch from PS) ?

src/transaction/atomic_replication_helper.cpp
```diff
@@ -268,35 +269,35 @@ namespace cublog
...
+	//    in an atomic sequnce and, for example, for the first log record, the page fails
+	//    to be fixed but succeeds for the second log record
+	//  - what if, while the atomic sequence is in progress with a page having failed to
```
razvanradulescu, line 289: For me, the safe approach would be: 
1. gather all log records of an atomic log seq
2. gather all VPID required
3. fix all pages
4. apply all log records
5. unfix all pages
I think the issue you mention may be caused by apply_and_unfix_sequence completing steps 4+5 at the same time and also if at step 3, if a fix fails we continue with the sequence. 
What if at least one page fix fails during step 3, we give up entirely and force evacuate all VPIDs from page buffer ?

src/transaction/atomic_replication_helper.hpp
cristiarg: Done

src/transaction/atomic_replication_helper.cpp
cristiarg: Done.

src/transaction/atomic_replication_helper.cpp
cristiarg: Done

src/transaction/atomic_replication_helper.cpp
cristiarg: In the first phase I will implement and extension to the bookkeeping to:
- keep track if a page failed to be fixed
- and a subsequent request to the same page - within the same atomic sequence - succeeds

At that point we'll at least throw an assert and, if that ever happens, re-asses at that point.

From what I saw, multiple modifications to the same page do occur when there are nested atomic replications.
But once we implement the simplification of how we handle atomic sequences - in the sense that we'll apply everything before the postpone as a standalone atomic sequence - this should not be a common case anymore.

src/transaction/atomic_replication_helper.cpp
cristiarg: @razvanradulescu 

AFICT, this problem with the replication failing to fix a page while another client transaction might succeed is a possible scenario only in the context of page re-allocation. And there is an issue describing the problem and a possible solution which involves unilaterally aborting client transaction - see http://jira.cubrid.org/browse/LETS-523 .

Currently the atomic replication is implemented as follows:
1. when each log record deemed to be part of an atomic replication is processed, the page is fixed right then&there
  * the bookkeeping takes care whether two log records want to access the same page
2. when the section is concluded, all log records are applied and the page is unfixed *right* after each record is applied
  * again, bookkeeping takes care of when the ref count actually reaches 0 and the page *can* safely be unfixed by the overall sequence

The way I see there is not a fundamental difference between the way described in the previous comment and the implementation.
Except that the implementation does keep the pages fixed for longer. But that can be seen as a feature rather than a problem.
What if, while the replication keeps the page fixed (for a bit longer) an client transaction would need to fix the page for reading in order to read some data which - due to the MVCC functionality - would not collide with the actual data that the replication wants to modify.
Thus, one can argue, the current implementation might impact performance.

hornetmj (approved):

src/transaction/atomic_replication_helper.hpp
```diff
@@ -158,47 +158,87 @@ namespace cublog
...
+
+	    VPID m_vpid = VPID_INITIALIZER;
+	    LOG_RCVINDEX m_rcv_index = RV_NOT_DEFINED;
+	    PAGE_PTR m_page_p = nullptr;
+	    page_ptr_watcher_uptr_type m_watcher_p;
+	    int m_ref_count = -1;
+	  };
+
+	  /*
+	   * Implements a RAII-like reference counted functionality to bokkeep page pointers for
```
wrlawodms:
```suggestion
	   * Implements a RAII-like reference counted functionality to bookkeep page pointers for
```

src/transaction/atomic_replication_helper.hpp
```diff
@@ -158,47 +158,87 @@ namespace cublog
...
+	    page_ptr_bookkeeping (const page_ptr_bookkeeping &) = delete;
+	    page_ptr_bookkeeping (page_ptr_bookkeeping &&) = delete;
+
+	    page_ptr_bookkeeping &operator= (const page_ptr_bookkeeping &) = delete;
+	    page_ptr_bookkeeping &operator= (page_ptr_bookkeeping &&) = delete;
+
+	    int fix_page (THREAD_ENTRY *thread_p, VPID vpid, LOG_RCVINDEX rcv_index, PAGE_PTR &page_ptr_out);
+	    int unfix_page (THREAD_ENTRY *thread_p, VPID vpid);
+
+	    using page_ptr_info_map_type = std::map<VPID, page_ptr_info>;
```
wrlawodms, line 236: Out of curiosity, why don't you use unordered_map though it doesn't have to be ordered?

src/transaction/atomic_replication_helper.cpp
```diff
@@ -403,35 +398,26 @@ namespace cublog
...
-    assert (lsa != NULL_LSA);
+    assert (!VPID_ISNULL (&m_vpid));
+    assert (m_record_lsa != NULL_LSA);
+    assert (0 <= m_record_index && m_record_index <= RV_LAST_LOGID);
+    assert (m_page_ptr != nullptr);
   }
 
   atomic_replication_helper::atomic_log_sequence::atomic_log_entry::atomic_log_entry (atomic_log_entry &&that)
-    : atomic_log_entry (that.m_record_lsa, that.m_vpid, that.m_record_index)
+    : atomic_log_entry (that.m_record_lsa, that.m_vpid, that.m_record_index, that.m_page_ptr)
```
wrlawodms, line 414: Is it OK not to set values in `&&that` to null? (If it's not needed, my bad)

src/transaction/atomic_replication_helper.cpp
```diff
@@ -363,33 +365,26 @@ namespace cublog
...
       {
-	const auto page_map_it = m_page_map.find (m_log_vec[i].m_vpid);
-	if (page_map_it != m_page_map.end ())
-	  {
-	    m_log_vec[i].unfix_page (thread_p);
-	    m_page_map.erase (page_map_it);
-	  }
+	log_entry.apply_log_redo (thread_p, m_redo_context);
+	// bookkeeping actually will either unfix the page or just decrease its reference count
+	m_page_ptr_bookkeeping.unfix_page (thread_p, log_entry.m_vpid);
```
wrlawodms, line 382: How about doing unifx on atomic_log_entry's destructor and fix on its constructor in terms of RAII? It seems that fixing and unfixing always go along with the atomic_log_entry's life cycle.

wrlawodms (approved):
LGTM. Approved with small suggestions and questions.

src/transaction/atomic_replication_helper.hpp
cristiarg: When I learned c++ there was no hash-map in STL. So, out of habit, I use map.
In this scenario, probably that it does not matter since the maps are small and the difference in performance between map and hash map is small (the extra search perfomance of hash-map versus map is compensated by the hash function).

src/transaction/atomic_replication_helper.cpp
cristiarg: It would be needed, but all the members are `const`.

src/transaction/atomic_replication_helper.cpp
cristiarg: On the fix/new entry side, they are decoupled in the sense that an error can interrupt the addition of a new entry - but that is an error case.
On the unfix/dtor side they are one and the same indeed.
I will leave it as it is. The sky is the limit in terms of C++ abstraction anyway and there is no clear line where one should stop or not. :)

wrlawodms: Ok. Thank you for considering it.

PR #3832, opened 2022-09-30
[LETS-547] Passive Transaction Server updates MVCC status information to Page Server

wrlawodms:
http://jira.cubrid.org/browse/LETS-547

- Get the mvccid from `log_Gl.mvcc_table.update_global_oldest_visible()`, which already takes the replicator's and active transactions' into account.
-- if no trx and replication is done, it points to the last replicated trx's mvccid + 1
- Add a daemon sending the mvccid in `class passive_tran_server`. It sends the mvccid periodically only if it's been changed.
- The deamon starts at the end of `log_initialize_passive_tran_server()`.

Please feel free to leave feedback about anything: C++ syntax/semantics, the location of variables, or whatever.

### DIFF
--- a/src/server/passive_tran_server.cpp
+++ b/src/server/passive_tran_server.cpp
@@ -22,6 +22,7 @@
 #include "server_type.hpp"
 #include "system_parameter.h"
 #include "thread_manager.hpp"
+#include "thread_looper.hpp"
 
 passive_tran_server::~passive_tran_server ()
 {
@@ -150,7 +151,22 @@ void passive_tran_server::send_and_receive_stop_log_prior_dispatch ()
   // needs to be consumed (aka: waited to be consumed/serialized to log)
 }
 
-void passive_tran_server::send_oldest_active_mvccid ()
+void passive_tran_server::start_oldest_active_mvccid_sender ()
+{
+  assert (m_oldest_active_mvccid_sender == nullptr);
+
+  /* Now 1s , but it would be a system parameter later to make it tunable. */
+  cubthread::looper loop (std::chrono::milliseconds (1000));
+  auto func_exec = std::bind (&passive_tran_server::send_oldest_active_mvccid, std::ref (*this), std::placeholders::_1);
+  auto sender_entry = new cubthread::entry_callable_task (std::move (func_exec)); /* delete on retire. See the constr */;
+
+  m_oldest_active_mvccid_sender = cubthread::get_manager ()->create_daemon (loop, sender_entry,
+				  "passive_tran_server::oldest_active_mvccid_sender");
+
+  assert (m_oldest_active_mvccid_sender != nullptr);
+}
+
+void passive_tran_server::send_oldest_active_mvccid (cubthread::entry &thread_entry)
 {
   /* TODO dummy function. will be modified corretly. soon  */
   std::string request_message;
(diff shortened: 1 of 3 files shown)


src/server/passive_tran_server.cpp
```diff
@@ -150,12 +153,33 @@ void passive_tran_server::send_and_receive_stop_log_prior_dispatch ()
...
+
+  /* Now 1s , but it would be a system parameter later to make it tunable. */
+  cubthread::looper loop (std::chrono::milliseconds (1000));
+  auto func_exec = std::bind (&passive_tran_server::send_oldest_active_mvccid, std::ref (*this), std::placeholders::_1);
+  auto sender_entry = new cubthread::entry_callable_task (std::move (func_exec)); /* delete on retire. See the constr. */;
+
+  m_oldest_active_mvccid_sender = cubthread::get_manager ()->create_daemon (loop, sender_entry,
+				  "passive_tran_server::oldest_active_mvccid_sender");
+
+  assert (m_oldest_active_mvccid_sender != nullptr); // when create_daemon() fails
```
cristiarg, line 168: maybe propagate fail error code upstream?
and upstream, log fatal error and return?
.. just for completeness and conformity

src/transaction/log_manager.c
```diff
@@ -1703,7 +1703,7 @@ log_initialize_passive_tran_server (THREAD_ENTRY * thread_p)
 
   logpb_initialize_logging_statistics ();
 
-  pts_ptr->send_oldest_active_mvccid ();	/* TODO not here acutally, just for testing. will be removed soon */
+  pts_ptr->start_oldest_active_mvccid_sender ();
```
cristiarg: Let's try to start it sooner, before starting the replicator.
Although in practice it will not make a difference since the message is fire&forget anyway.
Or, better, move the starting of the replicator after this line.

src/server/passive_tran_server.cpp
```diff
@@ -150,12 +153,33 @@ void passive_tran_server::send_and_receive_stop_log_prior_dispatch ()
...
+}
+
+void passive_tran_server::send_oldest_active_mvccid (cubthread::entry &thread_entry)
```
cristiarg: Nit.
Unname the variable if not used:
```suggestion
void passive_tran_server::send_oldest_active_mvccid (cubthread::entry &)
```

hornetmj (approved):
LGTM

src/server/passive_tran_server.cpp
wrlawodms: No error code is defined for this case. I actually was not sure I should put the assert here since no other caller even doesn't check if it succeeds or not, but I did for safety. Do you think we should define an error for this?

src/server/passive_tran_server.cpp
wrlawodms: Oh, I didn't know this way. Thank you.

src/transaction/log_manager.c
wrlawodms: I have moved it before starting the replicator. It seems better in that other daemons start there, and it is right after the PTS log analysis which allows the mvccid-sender to start.

joohok (approved):

src/server/passive_tran_server.cpp
cristiarg: No, just leave it like this.

cristiarg (approved):

PR #3837, opened 2022-10-05
[LETS-548] Page Server collects and aggregates MVCC information received from Passive Transaction Servers

wrlawodms:
http://jira.cubrid.org/browse/LETS-548

- Add a class `pts_mvcc_tracker` in charge of tracking mvcc status from all PTSes.
- The hash map, m_pts_oldest_active_mvccids, has to be protected by a mutex when the container structure can be changed. The operations are thread-safe. It is updated and accessed by multiple threads: 
  - update an entry: each PTS connection handler
  - add an entry: css master thread handling requests
  - delete an entry: each PTS connection handler in a normal case or each PTS's request sender in an abnormal case.
  - retrieve an entry: the ATS connection handler
- I considered the connection handler's destructor as where the entry is deleted because other PS-related resources are cleaned up there. However, I think the disconnection handler is more suitable in that it's done synchronously and the entry doesn't belong to the connection.
- The channel_id used as the key of `m_pts_oldest_active_mvccids` has the same string for all PTS-PS connections, but I use it assuming it has to be different and hoping it will be fixed in another PR.  I have tested with `connection_handler*` as the key and confirmed this PR works. I can use this, but I think we should use the logical one, connection id, rather than the physical one, a pointer.

### DIFF
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -188,9 +188,14 @@ page_server::connection_handler::receive_stop_log_prior_dispatch (tran_server_co
 void
 page_server::connection_handler::receive_oldest_active_mvccid (tran_server_conn_t::sequenced_payload &a_sp)
 {
-  /* TODO dummy function. will be modified corretly. soon  */
   const MVCCID oldest_mvccid = *reinterpret_cast<const MVCCID *const> (a_sp.pull_payload().c_str());
-  er_log_debug (ARG_FILE_LINE, "receive_oldest_active_mvccid(): %d\n", oldest_mvccid);
+
+  assert (m_ps.m_pts_oldest_active_mvccid.find (get_channel_id()) != m_ps.m_pts_oldest_active_mvccid.end());
+  assert (m_ps.m_pts_oldest_active_mvccid[get_channel_id()] < oldest_mvccid);
+
+  /* The mutex is not needed since one entry is only updated by one PTS connection handler.
+   * The mutex is only needed when the conatiner can be changed (eg. rehasing). i.e. when an entry is added or removed. */
+  m_ps.m_pts_oldest_active_mvccid[get_channel_id()] = oldest_mvccid;
 }
 
 void
@@ -199,6 +204,12 @@ page_server::connection_handler::receive_disconnect_request (tran_server_conn_t:
   // if this instance acted as a prior sender sink - in other words, if this connection handler was for a
   // passive transaction server - it should have been disconnected beforehand
   assert (m_prior_sender_sink_hook_func == nullptr);
+
+  {
+    std::lock_guard<std::mutex> lockg { m_ps.m_pts_oldest_active_mvccid_mtx };
+    assert (m_ps.m_pts_oldest_active_mvccid.find (get_channel_id()) != m_ps.m_pts_oldest_active_mvccid.end());
+    m_ps.m_pts_oldest_active_mvccid.erase (get_channel_id());
+  }
 
   m_ps.disconnect_tran_server_async (this);
 }
@@ -238,6 +249,12 @@ page_server::connection_handler::abnormal_tran_server_disconnect (css_error_code
 	  er_log_debug (ARG_FILE_LINE, "abnormal_tran_server_disconnect: ATS disconnected from PS. Error code: %d\n",
 			(int)error_code);
 	}
+
+      {
+	std::lock_guard<std::mutex> lockg { m_ps.m_pts_oldest_active_mvccid_mtx };
+	assert (m_ps.m_pts_oldest_active_mvccid.find (get_channel_id()) != m_ps.m_pts_oldest_active_mvccid.end());
+	m_ps.m_pts_oldest_active_mvccid.erase (get_channel_id());
+      }
 
       m_ps.disconnect_tran_server_async (this);
 
@@ -399,10 +416,19 @@ page_server::set_passive_tran_server_connection (cubcomm::channel &&chn)
   assert (is_page_server ());
 
   chn.set_channel_name ("PTS_PS_comm");
-  er_log_debug (ARG_FILE_LINE, "Passive transaction server connected to this page server. Channel id: %s.\n",
-		chn.get_channel_id ().c_str ());
+
+  const auto channel_id = chn.get_channel_id ().c_str ();
+
+  er_log_debug (ARG_FILE_LINE, "Passive transaction server connected to this page server. Channel id: %s.\n", channel_id);
 
   m_passive_tran_server_conn.emplace_back (new connection_handler (chn, transaction_server_type::PASSIVE, *this));
+
+  {
+    std::lock_guard<std::mutex> lockg { m_pts_oldest_active_mvccid_mtx };
+    assert (m_pts_oldest_active_mvccid.find (channel_id) == m_pts_oldest_active_mvccid.end());
+    m_pts_oldest_active_mvccid[channel_id] =
+	    MVCCID_ALL_VISIBLE; // MVCCID_ALL_VISIBLE means that it hasn't yet received. It will prevent the ATS run vacuum.
+  }
 }
 
 void
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -432,7 +432,8 @@ page_server::set_passive_tran_server_connection (cubcomm::channel &&chn)
 
   const auto channel_id = chn.get_channel_id ();
 
-  er_log_debug (ARG_FILE_LINE, "Passive transaction server connected to this page server. Channel id: %s.\n", channel_id);
+  er_log_debug (ARG_FILE_LINE, "Passive transaction server connected to this page server. Channel id: %s.\n",
+		channel_id.c_str ());
 
   m_passive_tran_server_conn.emplace_back (new connection_handler (chn, transaction_server_type::PASSIVE, *this));
 
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -191,19 +191,19 @@ page_server::connection_handler::receive_oldest_active_mvccid (tran_server_conn_
   const auto oldest_mvccid = *reinterpret_cast<const MVCCID *const> (a_sp.pull_payload().c_str());
 
   const auto channel_id = get_channel_id ();
-  assert (m_ps.m_pts_oldest_active_mvccid.find (channel_id) != m_ps.m_pts_oldest_active_mvccid.end());
-  assert (m_ps.m_pts_oldest_active_mvccid[channel_id] < oldest_mvccid);
+  assert (m_ps.m_pts_oldest_active_mvccids.find (channel_id) != m_ps.m_pts_oldest_active_mvccids.end());
+  assert (m_ps.m_pts_oldest_active_mvccids[channel_id] < oldest_mvccid);
 
   /* The mutex is not needed since one entry is only updated by one PTS connection handler.
    * The mutex is only needed when the conatiner can be changed (eg. rehasing). i.e. when an entry is added or removed. */
-  m_ps.m_pts_oldest_active_mvccid[channel_id] = oldest_mvccid;
+  m_ps.m_pts_oldest_active_mvccids[channel_id] = oldest_mvccid;
 
 #if !defined(NDEBUG)
   std::string msg;
   msg.append ("receive_oldest_active_mvccid: update the oldest active mvccid to " + std::to_string (
 		      oldest_mvccid) + " of " + channel_id);
   msg.append ("\n oldest mvcc ids:");
-  for (const auto &it : m_ps.m_pts_oldest_active_mvccid)
+  for (const auto &it : m_ps.m_pts_oldest_active_mvccids)
     {
       msg.append (" " + std::to_string (it.second));
     }
@@ -219,9 +219,9 @@ page_server::connection_handler::receive_disconnect_request (tran_server_conn_t:
   assert (m_prior_sender_sink_hook_func == nullptr);
 
   {
-    std::lock_guard<std::mutex> lockg { m_ps.m_pts_oldest_active_mvccid_mtx };
-    assert (m_ps.m_pts_oldest_active_mvccid.find (get_channel_id()) != m_ps.m_pts_oldest_active_mvccid.end());
-    m_ps.m_pts_oldest_active_mvccid.erase (get_channel_id());
+    std::lock_guard<std::mutex> lockg { m_ps.m_pts_oldest_active_mvccids_mtx };
+    assert (m_ps.m_pts_oldest_active_mvccids.find (get_channel_id()) != m_ps.m_pts_oldest_active_mvccids.end());
+    m_ps.m_pts_oldest_active_mvccids.erase (get_channel_id());
   }
 
   m_ps.disconnect_tran_server_async (this);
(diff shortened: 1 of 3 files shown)


src/server/page_server.cpp
```diff
@@ -188,9 +188,33 @@ page_server::connection_handler::receive_stop_log_prior_dispatch (tran_server_co
...
+
+  /*
+   * 1. The entry is already created when ths PTS is connected.
+   * 2. It is updated by the PTS only when it move foward.
+   *    Without update, it is MVCCID_ALL_VISIBLE by default, which is lesser than any mvccid assigned.
+   */
+  assert (m_ps.m_pts_oldest_active_mvccids.find (channel_id) != m_ps.m_pts_oldest_active_mvccids.end());
+  assert (m_ps.m_pts_oldest_active_mvccids[channel_id] < oldest_mvccid);
+
+  m_ps.m_pts_oldest_active_mvccids[channel_id] = oldest_mvccid;
```
cristiarg: We can use another approach - such as:
- in each `connection_handler` add a `m_pts_oldest_active_mvccids` whose access is protected by a mutex (it can even be an atomic value if that's more to your linking)
- initialize the value with `MVCCID_ALL_VISIBLE` in the ctor (this initialization does not need to be protected by the mutex as, in `set_passive_tran_server_connection`, the connection handler is first constructed and then added to the vector of connection handlers)
- `receive_oldest_active_mvccid` receives the value and stores it in the connection_handler member (thread safe)
- when the handler for the ATS request to retrieve the aggregated value, the vector for PTS connection handlers is iterated to aggregate the value.

Do we lose anything with this?

I think this approach solves the following:
- no need to have unique channel id's (I agree that those id's are not unique - but it's just a feature that was not needed until now)
- no need to make sure that the entry for that PTS is removed when the PTS either pro-actively disconnects from PS or upon abnormal disconnect to make sure the m_pts_oldest_active_mvccids stays in the correct shape
- no need to deal with active transaction server's connection handler when aggregating the final value as that one is not part of the vector
- a more decoupled approach as only the access to one's PTS's value is locked at one time
  - by the connection handler when receiving a value;
  - by the ATS's connection handler when retrieving the values to construct the aggregated value)
- the performance of the aggregator to calculate the value to be sent to ATS is unlikely to be a problem (but a pro-active calculation can be implemented if needed)

wrlawodms: I didn't consider it thoroughly, but I thought that it seems awkward for the "connection handler" to manage the oldest mvccid of the connected transaction server. I thought it should be in charge of just connecting and communicating. And, the cost for page_server to manage those values would not that expensive. To address it, we can make a wrapper class including the connection handler and the pts information, but I reckon it's too much for this. 
 Anyway, I agree your approach is way simpler. Just my concern. Could you tell me your opinion on it?

cristiarg: You do have a point of polluting connection handler with data.
And a point about the extra class/structure.

A paranthesis:
I think we can make the channel id unique by introducing the `m_socket` into it. With the restriction that the function must not only be called after connect/accept, and after the socket number is ~= `INVALID_SOCKET`. Which I think is the case since `set_passive_tran_server_connection` is called after the call to `server_channel::accept` in `css_process_server_server_connect`.

Thus we can leave the bulk implementation the way you already implemented it.

Or you can add an extra struct/class to handle everything separately - with thread safe access from the side of the PTS connection handlers (update values) and from the side of the ATS connection handler (aggregate results).

src/server/page_server.cpp
```diff
@@ -399,10 +437,22 @@ page_server::set_passive_tran_server_connection (cubcomm::channel &&chn)
...
+
   er_log_debug (ARG_FILE_LINE, "Passive transaction server connected to this page server. Channel id: %s.\n",
-		chn.get_channel_id ().c_str ());
+		channel_id.c_str ());
 
   m_passive_tran_server_conn.emplace_back (new connection_handler (chn, transaction_server_type::PASSIVE, *this));
+
+  {
+    std::lock_guard<std::mutex> lockg { m_pts_oldest_active_mvccids_mtx };
+    /* The entry is already created when ths PTS is connected. */
```
```suggestion
    /* The entry must not already be present. If the same passive transaction server has been connected
     * before, the entry must have been removed when the PTS disconnected or when the connection
     *  to the PTS was aborted.
     */
```

src/server/page_server.cpp
```diff
@@ -188,9 +188,33 @@ page_server::connection_handler::receive_stop_log_prior_dispatch (tran_server_co
...
+  assert (m_server_type == transaction_server_type::PASSIVE);
+
+  const auto oldest_mvccid = *reinterpret_cast<const MVCCID *const> (a_sp.pull_payload().c_str());
+  const auto channel_id = get_channel_id ();
+  std::lock_guard<std::mutex> lockg { m_ps.m_pts_oldest_active_mvccids_mtx };
+
+  /*
+   * 1. The entry is already created when ths PTS is connected.
+   * 2. It is updated by the PTS only when it move foward.
+   *    Without update, it is MVCCID_ALL_VISIBLE by default, which is lesser than any mvccid assigned.
```
cristiarg: `MVCCID_ALL_VISIBLE` is actually a guard value.
At least, as far as I understood, that's how it will be used.

src/server/page_server.cpp
wrlawodms: > Or you can add an extra struct/class to handle everything separately - with thread safe access from the side of the PTS connection handlers (update values) and from the side of the ATS connection handler (aggregate results).

To clarify, you mean a new inner class that belongs to page_server, and a member variable of page_server. Is it correct?

wrlawodms: >I think we can make the channel id unique by introducing the m_socket into it. With the restriction that the function must not only be called after connect/accept, and after the socket number is ~= INVALID_SOCKET. Which I think is the case since set_passive_tran_server_connection is called after the call to server_channel::accept in css_process_server_server_connect.

1. I guess that it will be resolved if the connected PTS sends its IP instead of the page server's IP. Then, we can use the message format as it is, and keep the channel id format. What do you think? 
2. It might as well address this unique channel id problem in another issue. I'll make it if you don't mind.

src/server/page_server.cpp
wrlawodms: Yes, it is. This comment explains why `assert (m_ps.m_pts_oldest_active_mvccids[channel_id] < oldest_mvccid);` is always true. I'll elaborate more on it in the comment when it is initialized (set_passive_tran_server_connection).

src/server/page_server.cpp
wrlawodms: Oh, my bad. This is a Copy&Paste mistake. Thank you.

src/server/page_server.cpp
```diff
@@ -409,6 +377,79 @@ page_server::async_disconnect_handler::disconnect_loop ()
...
+   */
+  assert (m_pts_oldest_active_mvccids.find (pts_channel_id) != m_pts_oldest_active_mvccids.end());
+  assert (m_pts_oldest_active_mvccids[pts_channel_id] < mvccid);
+
+  m_pts_oldest_active_mvccids[pts_channel_id] = mvccid;
+
+#if !defined(NDEBUG)
+  std::string msg;
+  msg.append ("receive_oldest_active_mvccid: update the oldest active mvccid to " + std::to_string (
+		      mvccid) + " of " + pts_channel_id);
```
cristiarg: Use either a `std::stringstream` (not-preferred) or a good ol' `snprintf` in a local buffer (preferred).
Using a `std::string` is the worst approach as each `+` constructs an intermediary `std::string` which is then discarded.

src/server/page_server.cpp
```diff
@@ -371,6 +377,79 @@ page_server::async_disconnect_handler::disconnect_loop ()
...
+   */
+  assert (m_pts_oldest_active_mvccids.find (pts_channel_id) == m_pts_oldest_active_mvccids.end());
+
+  /*
+   * MVCCID_ALL_VISIBLE means that it hasn't yet received. It will prevent the ATS to run vacuum.
+   * This is a guard for the window in which a PTS is connected but has't sent its oldest active mvccid.
+   * In this window, if we vaccum without considering the PTS, we possibly end up cleaning up the data
+   * a read-only transaction on the PTS see.
+   */
+  m_pts_oldest_active_mvccids[pts_channel_id] = MVCCID_ALL_VISIBLE;
```
hornetmj, line 399: Is there any reason to use the initial value as the MVCCID_ALL_VISIBLE instead of the MVCCID_NULL?

src/server/page_server.cpp
```diff
@@ -371,6 +377,79 @@ page_server::async_disconnect_handler::disconnect_loop ()
...
+{
+  std::lock_guard<std::mutex> lockg { m_pts_oldest_active_mvccids_mtx };
+
+  /*
+   * 1. The entry is already created when ths PTS is connected.
+   * 2. It is updated by the PTS only when it move foward.
+   *    Without update, it is MVCCID_ALL_VISIBLE by default, which is lower than any mvccid assigned.
+   */
+  assert (m_pts_oldest_active_mvccids.find (pts_channel_id) != m_pts_oldest_active_mvccids.end());
+  assert (m_pts_oldest_active_mvccids[pts_channel_id] < mvccid);
```
hornetmj, line 414:
```suggestion
  assert (m_pts_oldest_active_mvccids[pts_channel_id] <= mvccid);
```

src/server/page_server.cpp
```diff
@@ -371,6 +377,79 @@ page_server::async_disconnect_handler::disconnect_loop ()
...
+}
+
+MVCCID page_server::pts_mvcc_tracker::get_global_oldest_active_mvccid ()
+{
+  std::lock_guard<std::mutex> lockg { m_pts_oldest_active_mvccids_mtx };
+
+  MVCCID oldest_mvccid = MVCCID_LAST;
+  for (const auto &it : m_pts_oldest_active_mvccids)
+    {
+      if (oldest_mvccid > it.second)
```
hornetmj, line 446: The condition of if statement is always false. So, This function always returns -1(MVCCID_LAST).

src/server/page_server.cpp
wrlawodms: MVCCID_ALL_VISIBLE is already being used in a similar way in mvcctable::build_mvcc_info() and mvcctable::compute_oldest_visible_mvccid (). The local active mvccid is temporarily set as MVCCID_ALL_VISIBLE and during the period, the vacuum master holds back starting the vacuum process.

src/server/page_server.cpp
wrlawodms: I think `assert (m_pts_oldest_active_mvccids[pts_channel_id] < mvccid);` is correct. Please tell me why it's wrong.

src/server/page_server.cpp
wrlawodms: The type of MVCCID is UINT64 so that -1 means the max value of the type. 2^64-1.

wrlawodms: OK, I'll use `std::numeric_limits`to prevent that kind of confusion.

src/server/page_server.cpp
cristiarg: I also think strict comparison - `<` - is correct. Because the update from PTS is only done if the value has changed (increased). Otherwise there is no update.
@wrlawodms If my explanation is correct, please add as a comment for the assert.

hornetmj: I think that a PTS can send same oldest mvccid to PS. But It cannot. (in http://jira.cubrid.org/browse/LETS-547)

Your condition is correct.

src/server/page_server.cpp
```diff
@@ -371,6 +377,79 @@ page_server::async_disconnect_handler::disconnect_loop ()
...
+	{
+	  oldest_mvccid = it.second;
+	}
+    }
+
+  /* it can return either
+   * - MVCCID_LAST: no PTS is being trakced
```
```suggestion
   * - MVCCID_LAST: no PTS is being tracked
```

src/server/page_server.cpp
hornetmj: OK.

src/server/page_server.cpp
```diff
@@ -371,6 +377,79 @@ page_server::async_disconnect_handler::disconnect_loop ()
...
+   * - or MVCCID_ALL_VISIBLE: some PTS are connected, but hasn't update
```
```suggestion
   * - or MVCCID_ALL_VISIBLE: at least one PTS has connected, but hasn't updated yet
```

src/server/page_server.cpp
```diff
@@ -399,10 +478,15 @@ page_server::set_passive_tran_server_connection (cubcomm::channel &&chn)
   assert (is_page_server ());
 
   chn.set_channel_name ("PTS_PS_comm");
+
+  const auto channel_id = chn.get_channel_id ();
```
cristiarg, line 488: @wrlawodms I think this PR can be merged only if a solution for the unique channel id is found.

I suggest to take that as a separate issue using either my suggestion from https://github.com/CUBRID/cubrid/pull/3837#discussion_r989243338 or another solution.

And then come back and test&merge this one.

We can go forward with the current code with only one PTS, but the test infrastructure that we have in place for LETS tests uses 2 PTS's. And a few tests will crash.

Or another option that involves disabling this functionality in the `scalability_dev` branch.
CLARIFICATION: disable it temporary in the `scalability_dev` branch; only enable it locally when developing; in order to be able to merge this PR; and then, immediately come with a solution for the unique channel id

 Your choice.

src/server/page_server.cpp
```diff
@@ -371,6 +377,79 @@ page_server::async_disconnect_handler::disconnect_loop ()
...
+  /*
+   * MVCCID_ALL_VISIBLE means that it hasn't yet received. It will prevent the ATS to run vacuum.
+   * This is a guard for the window in which a PTS is connected but has't sent its oldest active mvccid.
+   * In this window, if we vaccum without considering the PTS, we possibly end up cleaning up the data
+   * a read-only transaction on the PTS see.
+   */
+  m_pts_oldest_active_mvccids[pts_channel_id] = MVCCID_ALL_VISIBLE;
+}
+
+void page_server::pts_mvcc_tracker::update_oldest_active_mvccid (const std::string &pts_channel_id, const MVCCID mvccid)
```
cristiarg, line 402: assert that `mvccid` is valid.

src/server/page_server.cpp
wrlawodms: Thank you for the advice. I've used `std::stringstream` to append elements in the map.

src/server/page_server.cpp
wrlawodms: Have added

src/server/page_server.cpp
wrlawodms: It seems better to get the unique channel id issue done first. I'll create an issue for it.

hornetmj (approved):

src/server/page_server.cpp
```diff
@@ -232,6 +239,8 @@ page_server::connection_handler::abnormal_tran_server_disconnect (css_error_code
 			(int)error_code);
 
 	  remove_prior_sender_sink ();
+
+	  m_ps.m_pts_mvcc_tracker.delete_oldest_active_mvccid (get_channel_id());
```
joohok, line 243: How about moving this part to `page_server::disconnect_tran_server_async ()` ? 

Then we can remove the same part in `page_server::connection_handler::receive_disconnect_request` .

joohok (approved):

src/server/page_server.cpp
wrlawodms: The `disconnect_tran_server_async()` is a different job. To put it together, I think we should make a wrapper function like `cleanup_tran_server()` calling them. It would be better to make that if we get more thing to clean up later in my opinion.

joohok: OK. I just mentioned the above comment because both functions (`abnormal_tran_server_disconnect()`, and `receive_disconnect_request()` ) commonly call the `disconnect_tran_server_async()` function.

PR #3838, opened 2022-10-05
[CBRD-24252] Removes the table of the primary key from the join with the table with the foreign key referencing it.

youngjinj:
http://jira.cubrid.org/browse/CBRD-24252

A table with a primary key is removed from a join with a table with a foreign key that references the primary key. There must be no references to the table to be removed other than join predicates. This is because the relationship between a primary key and a foreign key replaces a data filter of a join. If a table with a primary key is removed, a predicate for IS NOT NULL may be added, as it cannot filter on NULL.

```
e.g.

drop if exists child, parent;
create table parent (c1 int primary key, c2 int);
create table child (c1 int, parent_c1 int references parent (c1), c2 int);

select c.* from child c inner join parent p on p.c1 = c.parent_c1 where c.c2 = 1;
select c.* from child c, parent p where c.parent_c1 = p.c1 and c.c2 = 1;
  -> select c.* from child c where c.c2 = 1 and c.parent_c1 is not null;

select /*+ no_eliminate_join */ c.* from child c inner join parent p on p.c1 = c.parent_c1 where c.c2 = 1;
select /*+ no_eliminate_join */ c.* from child c, parent p where c.parent_c1 = p.c1 and c.c2 = 1;
  -> do not change.

select c.* from child c inner join parent p on p.c1 = c.parent_c1 where p.c2 = 1;
select c.* from child c, parent p where c.parent_c1 = p.c1 and p.c2 = 1;
  -> do not change.
```
```
qo_reduce_joined_tables_referenced_by_foreign_key()
  -> qo_is_exclude_spec()
  -> qo_check_primary_key_referenced_by_foreign_key_in_parent_spec()
  -> qo_check_foreign_keys_referencing_primary_key_in_child_spec()
    -> qo_check_foreign_key_referencing_primary_key_in_child_spec()
  -> qo_reduce_predicate_for_parent_spec()
```


src/parser/parse_tree.h
```diff
@@ -764,11 +800,42 @@ struct json_t;
...
+/* PT_SPEC_INFO */
+#define PT_SPEC_GET_DB_OBJECT(n, r)	\
+	do \
+	  { \
+	    if (PT_SPEC_FLAT_ENTITY_LIST ((n)) != NULL \
+		&& PT_NAME_DB_OBJECT (PT_SPEC_FLAT_ENTITY_LIST ((n))) != NULL) \
+	      { \
+		(r) = PT_NAME_DB_OBJECT (PT_SPEC_FLAT_ENTITY_LIST ((n))); \
+	      } \
+	    if (PT_SPEC_ENTITY_NAME ((n)) != NULL) \
```
ctshim: Can't node->info.spec have flat_entity_list and entity_name at the same time?
Is it intentional to use entity_name if all are present?

youngjinj: I will fix it by changing it to "else if".

ctshim (approved):

src/optimizer/query_rewrite.c
```diff
@@ -3499,6 +3500,651 @@ qo_reduce_outer_joined_tables (PARSER_CONTEXT * parser, PT_NODE * spec, PT_NODE
...
+      return;
+    }
+
+  if (query->info.query.q.select.hint & PT_HINT_NO_MERGE)
+    {
+      return;
+    }
+
+  if (query->info.query.q.select.where != NULL && query->info.query.q.select.where->or_next != NULL)
+    {
```
shparkcubrid, line 3575: it can not check below
```
where col1 = 1 and (col2 = 1 or col3 = 1)
```

src/optimizer/query_rewrite.c
```diff
@@ -3499,6 +3500,651 @@ qo_reduce_outer_joined_tables (PARSER_CONTEXT * parser, PT_NODE * spec, PT_NODE
...
+  PT_NODE *append_pred_list = NULL;
+  PT_NODE *curr_append_pred = NULL, *prev_append_pred = NULL, *next_append_pred = NULL;
+  int i;
+
+  if (parser == NULL || query == NULL || query->node_type != PT_SELECT)
+    {
+      return;
+    }
+
+  if (query->info.query.q.select.hint & PT_HINT_NO_MERGE)
```
hyunikn: What about introducing a new kind of hint other than NO_MERGE.
The user manual says NO_MERGE disables View Merging optimization.

src/optimizer/query_rewrite.c
```diff
@@ -3499,6 +3500,651 @@ qo_reduce_outer_joined_tables (PARSER_CONTEXT * parser, PT_NODE * spec, PT_NODE
...
+	{
+#if defined(CUBRID_DEBUG) || defined(CUBRID_DEBUG_TEST)
+	  printf ("[debug] [------] (-/-) has_reduce: false\n");
+#endif
+
+	  break;
+	}
+    }
+
+exit_on_error:
```
hyunikn: It seems that reaching this line does not always mean an error.
'clean_up' or something else seems more suitable than 'exit_on_error'.

src/optimizer/query_rewrite.c
youngjinj: Below, if it is not PT_EQ, it is skipped.
Since PT_OR is not PT_EQ, it is skipped.

I will add it to the test case.

src/optimizer/query_rewrite.c
youngjinj: Thank you. I agree. I will change it.

src/optimizer/query_rewrite.c
youngjinj: Added NO_ELIMINATE_JOIN hint.
This is a hint used by Oracle.

I will add it to the test case.

hyunikn (approved):

src/optimizer/query_rewrite.c
```diff
@@ -3499,6 +3500,651 @@ qo_reduce_outer_joined_tables (PARSER_CONTEXT * parser, PT_NODE * spec, PT_NODE
...
+
+		      for (i = 0; curr_pk_cons->attributes[i] != NULL && curr_fk_cons->attributes[i] != NULL; i++)
+			{
+			  SM_ATTRIBUTE *pk_cons_attr = curr_pk_cons->attributes[i];
+			  SM_ATTRIBUTE *fk_cons_attr = curr_fk_cons->attributes[i];
+
+			  if (intl_identifier_casecmp (pk_cons_attr->header.name, PT_NAME_ORIGINAL (pk_pred_attr)) == 0)
+			    {
+			      if (intl_identifier_casecmp (fk_cons_attr->header.name,
+							   PT_NAME_ORIGINAL (fk_pred_attr)) == 0)
```
shparkcubrid: Does it work properly if the order of pk and fk is different?

src/optimizer/query_rewrite.c
shparkcubrid: PLZ add or_next check

youngjinj: Added it.

src/optimizer/query_rewrite.c
```diff
@@ -3499,6 +3500,651 @@ qo_reduce_outer_joined_tables (PARSER_CONTEXT * parser, PT_NODE * spec, PT_NODE
...
+
+	  PT_SPEC_GET_DB_OBJECT (curr_pk_spec, curr_pk_mop);
+	  if (curr_pk_mop == NULL)
+	    {
+	      ASSERT_ERROR ();
+	      goto cleanup;
+	    }
+
+#if defined(CUBRID_DEBUG) || defined(CUBRID_DEBUG_TEST)
+	  printf ("[debug] [parent] (1/7) table: %s\n", pt_print_alias (parser, curr_pk_spec));
```
hgryoo: Instead of printf(), `er_log_debug(ARG_FILE_LINE,
		    "[parent] (1/7) table: %s\n", pt_print_alias (parser, curr_pk_spec));`

src/optimizer/query_rewrite.c
```diff
@@ -3499,6 +3500,651 @@ qo_reduce_outer_joined_tables (PARSER_CONTEXT * parser, PT_NODE * spec, PT_NODE
...
+ *         select c.* from child c inner join parent p on p.c1 = c.parent_c1 where c.c2 = 1;
+ *         select c.* from child c, parent p where c.parent_c1 = p.c1 and c.c2 = 1;
+ *           -> select c.* from child c where c.c2 = 1 and c.parent_c1 is not null;
+ *
+ *         select c.* from child c inner join parent p on p.c1 = c.parent_c1 where p.c2 = 1;
+ *         select c.* from child c, parent p where c.parent_c1 = p.c1 and p.c2 = 1;
+ *           -> do not change.
+ */
+static void
+qo_reduce_joined_referenced_tables (PARSER_CONTEXT * parser, PT_NODE * query)
```
hgryoo: Could you divide this function? This function has a long LOC, and the while-loops and for-loops are nested. So It is hard to follow the routine.

src/optimizer/query_rewrite.c
youngjinj: Debug message output became unnecessary, so i removed it.

src/optimizer/query_rewrite.c
youngjinj: I have separated the function as below.
```
qo_reduce_joined_tables_referenced_by_foreign_key()
  -> qo_is_exclude_spec()
  -> qo_check_primary_key_referenced_by_foreign_key_in_parent_spec()
  -> qo_check_foreign_keys_referencing_primary_key_in_child_spec()
    -> qo_check_foreign_key_referencing_primary_key_in_child_spec()
  -> qo_reduce_predicate_for_parent_spec()
```

src/optimizer/query_rewrite.c
youngjinj: Added er_log_debug() again. Thank you.

src/optimizer/query_rewrite.c
youngjinj: The column order of the primary key and foreign key must be the same. The parent table cannot be reduced if it is compared in a different order from the column order of the foreign key referencing the primary key in the predicate.

src/optimizer/query_rewrite.c
```diff
@@ -3499,6 +3523,870 @@ qo_reduce_outer_joined_tables (PARSER_CONTEXT * parser, PT_NODE * spec, PT_NODE
...
+    case PT_JOIN_CROSS:
+      [[fallthrough]];
+    case PT_JOIN_LEFT_OUTER:
+      [[fallthrough]];
+    case PT_JOIN_RIGHT_OUTER:
+      [[fallthrough]];
+    case PT_JOIN_FULL_OUTER:
+      [[fallthrough]];
+    case PT_JOIN_UNION:
+      goto exit_on_fail_with_exclude;
```
beyondykk9, line 3809: no need "default" case ?

src/optimizer/query_rewrite.c
```diff
@@ -3499,6 +3523,870 @@ qo_reduce_outer_joined_tables (PARSER_CONTEXT * parser, PT_NODE * spec, PT_NODE
...
+cleanup_on_fail:
+  parser_free_tree (parser, reduce_pred_point_list);
+  reduce_pred_point_list = NULL;
+
+  goto exit_on_fail;
+
+exit_on_fail_with_exclude:
+  er_log_debug (ARG_FILE_LINE, "%s: exclude. (spec: %s)\n", __func__, pt_print_alias (parser, curr_pk_spec));
+  reduce_reference_info->exclude_pk_spec_point_list =
+    parser_append_node (pt_point (parser, curr_pk_spec), reduce_reference_info->exclude_pk_spec_point_list);
```
beyondykk9, line 3955: exclude "return false" intentionally to fall through?

src/optimizer/query_rewrite.c
```diff
@@ -3499,6 +3523,870 @@ qo_reduce_outer_joined_tables (PARSER_CONTEXT * parser, PT_NODE * spec, PT_NODE
...
+    case PT_JOIN_CROSS:
+      [[fallthrough]];
+    case PT_JOIN_LEFT_OUTER:
+      [[fallthrough]];
+    case PT_JOIN_RIGHT_OUTER:
+      [[fallthrough]];
+    case PT_JOIN_FULL_OUTER:
+      [[fallthrough]];
+    case PT_JOIN_UNION:
+      goto exit_on_fail_with_exclude;
```
beyondykk9, line 4035: no need "default" case?

src/optimizer/query_rewrite.c
```diff
@@ -3499,6 +3523,870 @@ qo_reduce_outer_joined_tables (PARSER_CONTEXT * parser, PT_NODE * spec, PT_NODE
...
+    {
+      /* No foreign key */
+      goto exit_on_fail_with_exclude;
+    }
+
+  goto exit_on_fail;
+
+exit_on_fail_with_exclude:
+  reduce_reference_info->exclude_fk_spec_point_list =
+    parser_append_node (pt_point (parser, curr_fk_spec), reduce_reference_info->exclude_fk_spec_point_list);
```
beyondykk9, line 4081: exclude "return false" intentionally to fall through?

src/optimizer/query_rewrite.c
```diff
@@ -3499,6 +3523,870 @@ qo_reduce_outer_joined_tables (PARSER_CONTEXT * parser, PT_NODE * spec, PT_NODE
...
+	      continue;		/* curr_pk_spec->next */
+	    }
+
+	  reduce_reference_info.pk_spec = curr_pk_spec;
+
+	  if (qo_check_primary_key_referenced_by_foreign_key_in_parent_spec (parser, query, &reduce_reference_info))
+	    {
+	      ;
+	    }
+	  else
```
ctshim:
```suggestion
	 if (qo_check_primary_key_referenced_by_foreign_key_in_parent_spec (parser, query, &reduce_reference_info) == false)
```

ctshim: It would be better to avoid empty sentences with only semicolons as much as possible.

src/optimizer/query_rewrite.c
```diff
@@ -3499,6 +3523,870 @@ qo_reduce_outer_joined_tables (PARSER_CONTEXT * parser, PT_NODE * spec, PT_NODE
...
+{
+  QO_REDUCE_REFERENCE_INFO reduce_reference_info;
+  PT_NODE *exclude_spec_point = NULL, *exclude_spec = NULL;
+  PT_NODE *curr_pk_spec = NULL, *prev_pk_spec = NULL, *next_pk_spec = NULL;
+  PT_NODE *curr_fk_spec = NULL;
+  bool has_reduce = false;
+
+  memset (&reduce_reference_info, 0, sizeof (QO_REDUCE_REFERENCE_INFO));
+
+  if (parser == NULL || query == NULL || query->node_type != PT_SELECT)
```
ctshim:
```suggestion
assert(parser && query);
if (query->node_type != PT_SELECT)
```

src/optimizer/query_rewrite.c
ctshim: @beyondykk
[[failthrough]] is an explicit expression and has no role.
Therefore, it continues to the next case without a return.

src/optimizer/query_rewrite.c
youngjinj: Thank you. I will add it.

src/optimizer/query_rewrite.c
youngjinj: Thank you. I will add it.

src/optimizer/query_rewrite.c
youngjinj: It is intended.
I will add comments below.
```
/* fallthrough */
```

src/optimizer/query_rewrite.c
youngjinj: It is intended.
I will add comments below.
```
/* fallthrough */
```

src/optimizer/query_rewrite.c
youngjinj: Thank you. I will change it.

src/optimizer/query_rewrite.c
youngjinj: Thank you. I will change it.

beyondykk9 (approved):

hyunikn (approved):

src/optimizer/query_rewrite.c
```diff
@@ -3499,6 +3523,874 @@ qo_reduce_outer_joined_tables (PARSER_CONTEXT * parser, PT_NODE * spec, PT_NODE
...
+	}
+      else if (PT_SPEC_ID (curr_pk_spec) == PT_NAME_SPEC_ID (arg2))
+	{
+	  pk_pred_attr = PT_EXPR_ARG2 (curr_pred);
+	}
+      else
+	{
+	  /* Already checked before. */
+	  assert (false);
+	  ERROR_SET_ERROR_ONLY (ER_FAILED);
```
hgryoo: It seems to be a normal case to me if already checked. could you elaborate on it?

youngjinj: This is an impossible case.
And even if we do not set an error, there is no problem if the spec is skipped.
The reason for setting an error and exiting is to reduce inefficiency.

So I will change it to check only in debug mode using assert().

src/optimizer/query_rewrite.c
youngjinj: When creating a child table, even if foreign keys are created in the order of parent_c1 and parent_c2,
the order of the parent's primary key is changed.
```
drop table if exists child, parent;
create table parent (c1 int, c2 int, primary key (c2, c1));
create table child (c1 int primary key, parent_c1 int, parent_c2 int);
alter table child add constraint foreign key (parent_c1, parent_c2) references parent (c1, c2);
```
It is not created like "fk_child_parent_c1_parent_c2 ON dba.child (parent_c1, parent_c2)".
```
csql> ;sc parent
 ...
 <Constraints>

     PRIMARY KEY pk_parent_c2_c1 ON dba.parent (c2, c1)
...
csql> ;sc child

 ...
 <Constraints>

     PRIMARY KEY pk_child_c1 ON dba.child (c1)
     FOREIGN KEY fk_child_parent_c1_parent_c2 ON dba.child (parent_c2, parent_c1) REFERENCES dba.parent ...
...
```
Therefore, there is no problem in comparing the column order of the primary key and the foreign key index.
```
set optimization level 513;

select /*+ recompile */ c.c1 from child as c, parent as p where c.parent_c1 = p.c1 and c.parent_c2 = p.c2;
  -> select c.c1 from child c where c.parent_c1 is not null  and c.parent_c2 is not null

select /*+ recompile */ c.* from child as c, parent as p where c.parent_c1 = p.c2 and c.parent_c2 = p.c1;
  -> do not change.
```

youngjinj: And the creation of a foreign key for a unique index is not allowed.
```
drop table if exists child, parent;
create table parent (c1 int, c2 int, unique (c2, c1));
create table child (c1 int primary key, parent_c1 int, parent_c2 int);
alter table child add constraint foreign key (parent_c1, parent_c2) references parent (c1, c2);
```
```
ERROR: The class 'dba.parent' referred by the foreign key does not have the primary key.
```

shparkcubrid (approved):

hgryoo (approved):

hgryoo:
To remind)
In reviewing this PR, the inheritance behavior that a PK constraint is also inherited to the child table is different from the PostgreSQL or Oracle DBMS. It makes the code very complex.

PR #3839, opened 2022-10-05
[LETS-525] Move the last_mvcc_lsa variable into the mvcc information structure 

joohok:
http://jira.cubrid.org/browse/LETS-525

Purpose

tdes->last_mvcc_lsa is used whenever dealing with mvcc information, 
so it would be better to move tdes->last_mvcc_lsa to tdes->mvccinfo. 
Then it can be initialized or reset together with mvcc related information in mvccinfo.init () or mvccinfo.reset ().

### DIFF
--- a/src/transaction/log_recovery.c
+++ b/src/transaction/log_recovery.c
@@ -1759,7 +1759,6 @@ log_rv_analysis_assigned_mvccid (THREAD_ENTRY * thread_p, int tran_id, LOG_LSA *
 static void
 log_rv_analysis_complete_mvccid (int tran_index, const LOG_TDES * tdes)
 {
-  //TODO: if mvccinfo is reset in complete_mvcc, then move into it
   if (is_passive_transaction_server ())
     {
       if (MVCCID_IS_VALID (tdes->mvccinfo.id))
@@ -1802,8 +1801,8 @@ log_rv_analysis_complete (THREAD_ENTRY * thread_p, int tran_id, LOG_LSA * log_ls
 	  // newer quick fix on top of older quick fix: mark the mvccid as completed
 	  log_rv_analysis_complete_mvccid (tran_index, tdes);
 
-	  // quick fix: reset mvccid.
-	  tdes->mvccinfo.id = MVCCID_NULL;
+	  // quick fix: reset mvcc.
+	  tdes->mvccinfo.reset ();
 	  logtb_free_tran_index (thread_p, tran_index);
 	}
       return NO_ERROR;
--- a/src/transaction/log_recovery.c
+++ b/src/transaction/log_recovery.c
@@ -1801,8 +1801,8 @@ log_rv_analysis_complete (THREAD_ENTRY * thread_p, int tran_id, LOG_LSA * log_ls
 	  // newer quick fix on top of older quick fix: mark the mvccid as completed
 	  log_rv_analysis_complete_mvccid (tran_index, tdes);
 
-	  // quick fix: reset mvcc.
-	  tdes->mvccinfo.reset ();
+	  // quick fix: reset mvccid.
+	  tdes->mvccinfo.id = MVCCID_NULL;
 	  logtb_free_tran_index (thread_p, tran_index);
 	}
       return NO_ERROR;
--- a/src/transaction/log_recovery.c
+++ b/src/transaction/log_recovery.c
@@ -1182,7 +1182,7 @@ log_rv_analysis_mvcc_undo_redo (THREAD_ENTRY * thread_p, int tran_id, LOG_LSA * 
   assert (tdes != nullptr);
 
   // MVCC handling
-  tdes->last_mvcc_lsa = *log_lsa;
+  tdes->mvccinfo.last_mvcc_lsa = *log_lsa;
 
   // assign transaction mvccid from log record to transaction descriptor
   assert (log_page_p != nullptr);
@@ -1744,7 +1744,7 @@ log_rv_analysis_assigned_mvccid (THREAD_ENTRY * thread_p, int tran_id, LOG_LSA *
       return ER_FAILED;
     }
 
-  tdes->last_mvcc_lsa = *log_lsa;
+  tdes->mvccinfo.last_mvcc_lsa = *log_lsa;
 
   // move read pointer past the log header which is actually read upper in the stack
   LOG_READ_ADD_ALIGN (thread_p, sizeof (LOG_RECORD_HEADER), log_lsa, log_page_p);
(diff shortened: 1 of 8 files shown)


hornetmj (approved):
LGTM

src/transaction/log_recovery.c
```diff
@@ -1801,8 +1801,8 @@ log_rv_analysis_complete (THREAD_ENTRY * thread_p, int tran_id, LOG_LSA * log_ls
 	  // newer quick fix on top of older quick fix: mark the mvccid as completed
 	  log_rv_analysis_complete_mvccid (tran_index, tdes);
 
-	  // quick fix: reset mvccid.
-	  tdes->mvccinfo.id = MVCCID_NULL;
+	  // quick fix: reset mvcc.
+	  tdes->mvccinfo.reset ();
```
wrlawodms: How about putting `tdes->mvccinfo.reset ();` into `logtb_clear_tdes()` and removing this line?

src/transaction/log_tran_table.c
```diff
@@ -1526,7 +1526,6 @@ logtb_clear_tdes (THREAD_ENTRY * thread_p, LOG_TDES * tdes)
   LSA_SET_NULL (&tdes->topop_lsa);
   LSA_SET_NULL (&tdes->tail_topresult_lsa);
   LSA_SET_NULL (&tdes->commit_abort_lsa);
-  tdes->last_mvcc_lsa.set_null ();
```
wrlawodms, line 1529: Is mvccinfo.reset() not needed here?

src/transaction/log_tran_table.c
```diff
@@ -1695,7 +1694,6 @@ logtb_initialize_tdes (LOG_TDES * tdes, int tran_index)
...
```
wrlawodms, line 1698: Is mvccinfo.reset() not needed here?

src/transaction/log_tran_table.c
cristiarg: Yes, it would be needed.
But it would also mean a - rather drastic - change in functionality.
Below, in the same function, there is an
```
  assert (tdes->mvccinfo.id == MVCCID_NULL);
```
This assert was introduced with patch 3d49b48ce9 in 2015 for http://jira.cubrid.org/browse/CUBRIDSUS-15914 .

So it seems that the invariant is that the MVCC information must be already cleaned when entering `logtb_clean_tdes`.
I would maintain the existing functionality as ~~we~~ I don't have an overview of what would mean to change this invariant.

So best it would be only clean the `mvcc_info.last_mvcc_lsa`.
It's ugly but risky otherwise.

cristiarg: Other option is to remove this `set_null` as already done.
And add an assert of ours together with the existing one:
```
assert (tdes->mvccinfo.id == MVCCID_NULL);
assert (tdes->mvccinfo.last_mvcc_lsa.is_null ());
```
And we'll see how this fares in tests.

src/transaction/log_tran_table.c
joohok: Because mvccinfo.init () is called at 1739L, mvccinfo.reset () in not needed here

cristiarg: `logtb_initialize_tdes` already calls `mvccinfo.init ()`.
So removing this is ok.

src/transaction/log_recovery.c
joohok: I think it is better to put `tdes->mvccinfo.reset ()` into `logtb_clear_tdes ()`. 

`logtb_clear_tdes ()`, I think, is designed to be called after `logtb_complete_mvcc ()`. 
because mvcc related information is only set in the active transaction, and will be cleared after transaction is completed.

```
log_commit/abort () 
-> log_commit/abort_local () //  calls logtb_complete_mvcc () to reset mvccinfo
-> log_complete ()  // calls logtb_clear_tdes ()
```

However, recovery is not considered.
In `log_rv_undo_abort_complete()` and `log_rv_analysis_complete()`, 
it just set mvccid to null and clear the tdes. 
So, tdes->mvccinfo.last_mvcc_lsa is not reset during recovery, and this will make other problem, since the tdes will be reused by following transaction.

src/transaction/log_tran_table.c
joohok: Thanks for detail comment. 

https://github.com/CUBRID/cubrid/pull/3839/files#r989675406 
this comment could be an answer for the feedback. 

In the process of starting a transaction, performing mvcc operation, finishing mvcc, and clearing tdes, 
the above assert statement will work well. 

However, during recovery, only mvccid is initialized among mvcc information and then tdes is initialized. 
Then, the uninitialized mvcc information (last_mvcc_lsa) is used in the transaction that reused the corresponding tdes, causing a problem. 

So, I replace the place where only mvccid is initialized with mvccinfo.reset (), 
and it seems okay to put this into logtb_clear_tdes ().

cristiarg: @wrlawodms @joohok 
To reach a middle ground in `logtb_clear_tdes`:
- to keep the existing invariant of the assert:
```
assert (tdes->mvccinfo.id == MVCCID_NULL);
```
- to make sure that mvccinfo is still initialized

I would change to the following:
```
assert (tdes->mvccinfo.id == MVCCID_NULL);
tdes->mvccinfo.reset ();
```
~~The root problem is perhaps that the route `logtb_rv_find_allocate_tran_index` -> `logtb_allocate_tran_index` calls `logtb_clear_tdes`. It should, instead, call `logtb_initialize_tdes` which seems more appropriate:~~
- ~~`logtb_initialize_tdes` is to initialize an empty/garbage structure~~
- ~~`logtb_clear_tdes` is to clear an initialized structure~~

Later edit: Disregard the last part of my comment. `logrb_clear_tdes` is correctly used in `logtb_allocate_tran_index`. `logtb_initialize_tdes` is intended to be used only to allocate a newly created tdes. And it is. So my evaluation was incorrect.

src/transaction/log_recovery.c
cristiarg: I think I understand what you mean:
- on the path `log_rv_analysis_complete` -> `logtb_free_tran_index` -> `logtb_clear_tdes`, in function `logtb_clear_tdes` there is a pre-condition that `assert (tdes->mvccinfo.id == MVCCID_NULL)` which I would like us to keep.
- so you say that re-setting the `mvccid` is still needed here - and I agree
- so, let's leave just the `mvccinfo.id = null` as before and still add in the `logdb_clear_tdes` - as I suggested in a previous comment :
```
assert (tdes->mvccinfo.id == MVCCID_NULL);
tdes->mvccinfo.reset ();
```

A similar thing happens in `log_rv_undo_abort_complete`:
- on the path `log_rv_undo_abort_complete` -> `log_complete` -> `logtb_clear_tdes`
- the same precondition from `logtb_clear_tdes` mandates leaving the line
```
  tdes->mvccinfo.id = MVCCID_NULL;
```
in `log_rv_undo_abort_complete`

joohok: What I meant to say was to remove the part that initializes mvccid (`tdes->mvccinfo.id = MVCCID_NULL;`) inside the `log_rv_analysis_complete` and `log_rv_undo_abort_complete` functions.
This part looks like a quick fix to pass the `assert (tdes->mvccinfo.id == MVCCID_NULL);`,
and the above assert statement doesn't seem to take the recovery process into account, so I think it seems unnecessary. 

my opinion : 
1) remove `tdes->mvccinfo.id = MVCCID_NULL;` in `log_rv_analysis_complete` and `log_rv_undo_abort_complete`
2) remove `assert (tdes->mvccinfo.id == MVCCID_NULL);` in `logtb_clear_tdes ()`
3) put `mvccinfo.reset ()` in `log_clear_tdes ()`

But I also understand your opinion of reducing risk. First of all, I will modify according to your opinion. After further analysis, we will proceed with refactoring later.

cristiarg (approved):
Looks good AFAICT.
Let's see how this fares when both shell-debug and LETS tests suites are executed.

wrlawodms (approved):

PR #3840, opened 2022-10-06
[LETS-557] Stop outgoing msgs before disconnecting to the PS

wrlawodms:
http://jira.cubrid.org/browse/LETS-557

- call stop_outgoing_page_server_messages() before disconnecting
- specialize stop_outgoing_page_server_messages() on PTS and destroy the sender daemon here.

### DIFF
--- a/src/server/active_tran_server.cpp
+++ b/src/server/active_tran_server.cpp
@@ -89,3 +89,8 @@ active_tran_server::get_request_handlers ()
 
   return handlers_map;
 }
+
+void
+active_tran_server::stop_outgoing_page_server_messages ()
+{
+}
--- a/src/server/active_tran_server.hpp
+++ b/src/server/active_tran_server.hpp
@@ -39,6 +39,8 @@ class active_tran_server : public tran_server
 
     void receive_saved_lsa (page_server_conn_t::sequenced_payload &a_ip);
 
+    void stop_outgoing_page_server_messages () final override;
+
   private:
     bool m_uses_remote_storage = false;
 
--- a/src/server/passive_tran_server.cpp
+++ b/src/server/passive_tran_server.cpp
@@ -26,9 +26,8 @@
 
 passive_tran_server::~passive_tran_server ()
 {
+  assert (m_oldest_active_mvccid_sender == nullptr);
   assert (m_replicator == nullptr);
-
-  cubthread::get_manager ()->destroy_daemon (m_oldest_active_mvccid_sender);
 }
 
 bool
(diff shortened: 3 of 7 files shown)


src/server/active_tran_server.cpp
```diff
@@ -89,3 +89,8 @@ active_tran_server::get_request_handlers ()
 
   return handlers_map;
 }
+
+void
+active_tran_server::stop_outgoing_page_server_messages ()
+{
+}
```
cristiarg, line 96: Please add the following comment:
- in `xboot_shutdown_server` routine replace with a more elaborate comment for PTS and ATS:
- it should become something like
```
  else if (get_server_type () == SERVER_TYPE_TRANSACTION)
    {
      if (is_passive_transaction_server ())
	{
	  // shutdown order for Passive Transaction Server:
	  //  - send and receive confirmation that Page Server(s) stop dispatching log prior messages
	  //    - done in: shutdown server routine
	  //  - stop log prior receiver on Passive Transaction Server
	  //    - done in: shutdown server routine
	  //  - wait for replication to finish and stop its infrastructurej
	  //    - done in: shutdown server routine
	  //  - finalize log infrastructure
	  //    - done in: shutdown server routine
	  //  - stop sending messages to page server (eg: oldest MVCCID updates)
	  //    - done in:
	  //        shutdown server routine
	  //          -> boot_server_all_finalize
	  //            -> finalize_server_type
	  //              -> and then polymorphically in the transaction server object
	  //                disconnect_page_server routine right before sending
	  //                the final disconnect message to Page Server(s)
	  //  - send the final disconnect message to Page Server(s)
	  //    - done in : tran_server::disconnect_page_server
	  //  - delete the Passive Transaction Server object
	  //    - done in: finalize_server_type

	  passive_tran_server *const pts_ptr = get_passive_tran_server_ptr ();
	  pts_ptr->send_and_receive_stop_log_prior_dispatch ();

	  log_Gl.finalize_log_prior_receiver ();	// stop receiving log before log_final()

	  // NOTE: passive transaction server, regarding replication: even if a
	  // passive transaction server is completely transient - and read-only -
	  // and, thus, does not need to reach a consistent state at shutdown (because it
	  // will pick a consistent state at boot from the page server(s) it connects to), replication
	  // needs to be explicitly terminated gracefully before log infrastructure is finalized
	  pts_ptr->finish_replication_during_shutdown (*thread_p);

	}

      // shutdown order for Active Transaction Server:
      //  - finalize log infrastructure
      //    - done in: shutdown server routine
      //  - the log prior messages will stop being dispatched to the Page Server(s)
      //    by virtue of the client transactions and other daemons being stopped
      //    and, thus, no log will be produced anymore
      //    - log prior messages are being dispatched from Active Transaction Server towards
      //      Page Server(s) semi-synchronously via the prior_sender::send_list routine
      //  - send the final disconnect message to Page Server(s)
      //    - done in:
      //        shutdown server routine
      //          -> boot_server_all_finalize
      //            -> finalize_server_type
      //              -> tran_server::disconnect_page_server
      //  - delete the Active Transaction Server object
      //    - done in: finalize_server_type
    }
```

Just so we have a more complete image of what happens within each server at normal shutdown.
Thanks.

cristiarg (approved):

src/server/active_tran_server.cpp
wrlawodms: Thank you for the work. I've added.

joohok (approved):

hornetmj (approved):

PR #3844, opened 2022-10-11
[CBRD-24504] When an error occurs in a function that uses a buffer to convert an identifier name, the buffer must be initialized.

youngjinj:
http://jira.cubrid.org/browse/CBRD-24504

### When an error occurs in a function that uses a buffer to convert an identifier name, the buffer must be initialized.
- sm_find_class_with_purpose()
- sm_find_synonym()
- sm_user_specified_name() 
- sm_downcase_name()

### DIFF
--- a/src/object/schema_manager.c
+++ b/src/object/schema_manager.c
@@ -2251,6 +2251,9 @@ sm_downcase_name (const char *name, char *buf, int buf_size)
 {
   int error = NO_ERROR;
 
+  assert (buf != NULL);
+  assert (buf_size > 0);
+
   if (name == NULL || name[0] == '\0')
     {
       ERROR_SET_WARNING (error, ER_SM_INVALID_ARGUMENTS);
@@ -2258,7 +2261,6 @@ sm_downcase_name (const char *name, char *buf, int buf_size)
       return NULL;
     }
 
-  assert (buf != NULL);
   assert (intl_identifier_lower_string_size (name) < buf_size);
   intl_identifier_lower (name, buf);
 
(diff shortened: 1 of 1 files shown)


src/object/schema_manager.c
```diff
@@ -2254,6 +2254,7 @@ sm_downcase_name (const char *name, char *buf, int buf_size)
   if (name == NULL || name[0] == '\0')
     {
       ERROR_SET_WARNING (error, ER_SM_INVALID_ARGUMENTS);
+      buf[0] = '\0';
```
ctshim, line 2260: It is recommended to move the inspection of the 2261 line of assert() to the front of this if block.

ctshim (approved):

src/object/schema_manager.c
beyondykk9: Good point, it needs to move.

src/object/schema_manager.c
```diff
@@ -2282,6 +2283,7 @@ sm_user_specified_name (const char *name, char *buf, int buf_size)
   if (name == NULL || name[0] == '\0')
```
beyondykk9, line 2288: it needs to assert for "buf is not null" and "buf_size > 0"

beyondykk9 (approved):

hyunikn (approved):

src/object/schema_manager.c
youngjinj: Thank you. I will change it.

src/object/schema_manager.c
youngjinj: Thank you. I will add it.

hgryoo (approved):

shparkcubrid (approved):

PR #3849, opened 2022-10-11
[CBRD-24467] backport for 10.2.9

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-24467

backport for 10.2.9
[CBRD-24467] Sort-limit optimization does not work when host variable is used in limit clause.
[CBRD-24473] Index scan cannot be used when a stored function is used in a where condition.
[CBRD-23926] Use statistics of pages when checking recompilation threshold.
[CBRD-23732] Optional Statistics Estimation

The default of Optional Statistics Estimation is changed from FALSE to TRUE.

### DIFF
--- a/src/base/system_parameter.c
+++ b/src/base/system_parameter.c
@@ -2273,8 +2273,8 @@ static UINT64 prm_first_log_pageid_lower = 0LL;
 static UINT64 prm_first_log_pageid_upper = LOGPAGEID_MAX;
 static unsigned int prm_first_log_pageid_flag = 0;
 
-bool PRM_USE_STAT_ESTIMATION = false;
-static bool prm_use_stat_estimation_default = false;
+bool PRM_USE_STAT_ESTIMATION = true;
+static bool prm_use_stat_estimation_default = true;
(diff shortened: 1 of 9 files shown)


src/base/system_parameter.c
```diff
@@ -2271,6 +2273,10 @@ static UINT64 prm_first_log_pageid_lower = 0LL;
 static UINT64 prm_first_log_pageid_upper = LOGPAGEID_MAX;
 static unsigned int prm_first_log_pageid_flag = 0;
 
+bool PRM_USE_STAT_ESTIMATION = false;
```
shparkcubrid:
```suggestion
bool PRM_USE_STAT_ESTIMATION = true;
```

src/base/system_parameter.c
```diff
@@ -2271,6 +2273,10 @@ static UINT64 prm_first_log_pageid_lower = 0LL;
...
+static bool prm_use_stat_estimation_default = false;
```
shparkcubrid:
```suggestion
static bool prm_use_stat_estimation_default = true;
```

beyondykk9 (approved):

PR #3867, opened 2022-10-13
[LETS-549] ATS to PS request to retrieve aggregated MVCC information

wrlawodms:
http://jira.cubrid.org/browse/LETS-549

- Add the request and leave a debug logging to confirm it works. (I've tested with 1ATS and (0~2)PTS)
- Add `ats_Gl` and `get_active_tran_server_ptr()` to access the active trans server.

### DIFF
--- a/src/base/server_type.cpp
+++ b/src/base/server_type.cpp
@@ -138,6 +138,7 @@ int init_server_type (const char *db_name)
 
       if (is_active_transaction_server ())
 	{
+	  assert (ats_Gl == nullptr);
 	  ats_Gl = new active_tran_server ();
 	  ts_Gl.reset (ats_Gl);
 	}
--- a/src/base/server_type.cpp
+++ b/src/base/server_type.cpp
@@ -30,7 +30,9 @@
 #include <string>
 
 std::unique_ptr<tran_server> ts_Gl;
+
 // non-owning "shadow" pointer of globally visible ts_Gl
+active_tran_server *ats_Gl = nullptr;
 passive_tran_server *pts_Gl = nullptr;
 
 SERVER_TYPE get_server_type_from_config (server_type_config parameter_value);
@@ -136,7 +138,8 @@ int init_server_type (const char *db_name)
 
       if (is_active_transaction_server ())
 	{
-	  ts_Gl.reset (new active_tran_server ());
+	  ats_Gl = new active_tran_server ();
+	  ts_Gl.reset (ats_Gl);
 	}
       else if (is_passive_transaction_server ())
 	{
@@ -198,8 +201,14 @@ void finalize_server_type ()
 {
   if (get_server_type () == SERVER_TYPE_TRANSACTION)
     {
-      if (is_passive_transaction_server ())
+      if (is_active_transaction_server ())
 	{
+	  assert (ats_Gl != nullptr);
+	  ats_Gl = nullptr;
+	}
+      else
+	{
+	  assert (is_passive_transaction_server ());
 	  assert (pts_Gl != nullptr);
 	  pts_Gl = nullptr;
 	}
@@ -250,6 +259,20 @@ bool is_tran_server_with_remote_storage ()
       return ts_Gl->uses_remote_storage ();
     }
   return false;
+}
+
+active_tran_server *get_active_tran_server_ptr ()
+{
+  if (is_active_transaction_server ())
+    {
+      assert (ats_Gl != nullptr);
+      return ats_Gl;
+    }
+  else
+    {
+      assert (is_active_transaction_server ());
+      return nullptr;
+    }
 }
 
 passive_tran_server *get_passive_tran_server_ptr ()
--- a/src/base/server_type.cpp
+++ b/src/base/server_type.cpp
@@ -271,7 +271,7 @@ active_tran_server *get_active_tran_server_ptr ()
     }
   else
     {
-      assert (is_active_transaction_server ());
+      assert (is_passive_transaction_server ());
       return nullptr;
     }
 }
@@ -285,7 +285,7 @@ passive_tran_server *get_passive_tran_server_ptr ()
     }
   else
     {
-      assert (is_passive_transaction_server ());
+      assert (is_active_transaction_server ());
       return nullptr;
     }
 }
--- a/src/base/server_type.cpp
+++ b/src/base/server_type.cpp
@@ -1,6 +1,6 @@
 /*
  * Copyright 2008 Search Solution Corporation
- * Copyright 2021 CUBRID Corporation
+ * Copyright 2016 CUBRID Corporation
  *
  *  Licensed under the Apache License, Version 2.0 (the "License");
  *  you may not use this file except in compliance with the License.
--- a/src/base/server_type.cpp
+++ b/src/base/server_type.cpp
@@ -202,16 +202,15 @@ void finalize_server_type ()
 {
   if (get_server_type () == SERVER_TYPE_TRANSACTION)
     {
-      if (is_active_transaction_server ())
+      if (is_passive_transaction_server ())
 	{
-	  assert (ats_Gl != nullptr);
-	  ats_Gl = nullptr;
+	  assert (pts_Gl != nullptr);
+	  pts_Gl = nullptr;
 	}
       else
 	{
-	  assert (is_passive_transaction_server ());
-	  assert (pts_Gl != nullptr);
-	  pts_Gl = nullptr;
+	  assert (ats_Gl != nullptr);
+	  ats_Gl = nullptr;
 	}
       ts_Gl->disconnect_page_server ();
       ts_Gl.reset (nullptr);
--- a/src/base/server_type.cpp
+++ b/src/base/server_type.cpp
@@ -270,7 +270,7 @@ active_tran_server *get_active_tran_server_ptr ()
     }
   else
     {
-      assert (is_passive_transaction_server ());
+      assert (false);
       return nullptr;
     }
 }
@@ -284,7 +284,7 @@ passive_tran_server *get_passive_tran_server_ptr ()
     }
   else
     {
-      assert (is_active_transaction_server ());
+      assert (false);
       return nullptr;
     }
 }
--- a/src/query/vacuum.c
+++ b/src/query/vacuum.c
@@ -2984,6 +2984,13 @@ vacuum_master_task::execute (cubthread::entry &thread_ref)
  
   /* TODO temporary logging. The global one will be computed taking both into account, and the vacuum runs */ 
   MVCCID global_pts_oldest_visible_mvccid = get_active_tran_server_ptr ()->get_oldeset_active_mvccid_from_page_server ();
+  if (global_pts_oldest_visible_mvccid == MVCCID_NULL)
+  {
+    assert (false);
+    vacuum_er_log (VACUUM_ER_LOG_MASTER, "%s", "Fail to get the oldest active mvccid across all PTS.");
+    return;
+  }
+
   er_log_debug (ARG_FILE_LINE, "ats oldest_visible = %llu, pts global_oldest_visible = %llu",
       (long long int) m_oldest_visible_mvccid, global_pts_oldest_visible_mvccid);
 
--- a/src/query/vacuum.c
+++ b/src/query/vacuum.c
@@ -2958,6 +2958,8 @@ vacuum_master_task::execute (cubthread::entry &thread_ref)
 {
   PERF_UTIME_TRACKER perf_tracker;
 
+  assert (is_active_transaction_server());
+
   if (prm_get_bool_value (PRM_ID_DISABLE_VACUUM))
     {
       return;
@@ -2974,8 +2976,6 @@ vacuum_master_task::execute (cubthread::entry &thread_ref)
       // check if boot is aborted
       return;
     }
-
-  assert (is_active_transaction_server());
 
   PERF_UTIME_TRACKER_START (&thread_ref, &perf_tracker);
 
--- a/src/query/vacuum.c
+++ b/src/query/vacuum.c
@@ -2983,7 +2983,7 @@ vacuum_master_task::execute (cubthread::entry &thread_ref)
   vacuum_er_log (VACUUM_ER_LOG_MASTER, "update oldest_visible = %lld", (long long int) m_oldest_visible_mvccid);
  
   /* TODO temporary logging. The global one will be computed taking both into account, and the vacuum runs */ 
-  MVCCID global_pts_oldest_visible_mvccid = get_active_tran_server_ptr()->get_oldeset_active_mvccid_from_page_server();
+  MVCCID global_pts_oldest_visible_mvccid = get_active_tran_server_ptr ()->get_oldeset_active_mvccid_from_page_server ();
   er_log_debug (ARG_FILE_LINE, "ats oldest_visible = %llu, pts global_oldest_visible = %llu",
       (long long int) m_oldest_visible_mvccid, global_pts_oldest_visible_mvccid);
 
--- a/src/query/vacuum.c
+++ b/src/query/vacuum.c
@@ -2989,7 +2989,7 @@ vacuum_master_task::execute (cubthread::entry &thread_ref)
   vacuum_er_log (VACUUM_ER_LOG_MASTER, "update oldest_visible = %lld", (long long int) m_oldest_visible_mvccid);
  
   /* TODO temporary logging. The global one will be computed taking both into account, and the vacuum runs */ 
-  MVCCID global_pts_oldest_visible_mvccid = get_active_tran_server_ptr ()->get_oldeset_active_mvccid_from_page_server ();
+  MVCCID global_pts_oldest_visible_mvccid = get_active_tran_server_ptr ()->get_oldest_active_mvccid_from_page_server ();
   if (global_pts_oldest_visible_mvccid == MVCCID_NULL)
   {
     assert (false);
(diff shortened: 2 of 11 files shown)


src/base/server_type.cpp
```diff
@@ -252,6 +262,20 @@ bool is_tran_server_with_remote_storage ()
...
+active_tran_server *get_active_tran_server_ptr ()
+{
+  if (is_active_transaction_server ())
+    {
+      assert (ats_Gl != nullptr);
+      return ats_Gl;
+    }
+  else
+    {
+      assert (is_active_transaction_server ());
```
joohok:
```suggestion
      assert (!is_active_transaction_server ());
```

wrlawodms: ㅇ0ㅇ thanks.

wrlawodms:
It seems that test_*_sql can't start the server. looking into it.

src/base/server_type.cpp
```diff
@@ -198,8 +202,14 @@ void finalize_server_type ()
...
     {
-      if (is_passive_transaction_server ())
+      if (is_active_transaction_server ())
+	{
+	  assert (ats_Gl != nullptr);
+	  ats_Gl = nullptr;
+	}
+      else
 	{
+	  assert (is_passive_transaction_server ());
```
hornetmj: Nit:
I think assert (pts_Gl != nullptr); is enough.

src/base/server_type.cpp
```diff
@@ -252,6 +262,20 @@ bool is_tran_server_with_remote_storage ()
...
+active_tran_server *get_active_tran_server_ptr ()
+{
+  if (is_active_transaction_server ())
+    {
+      assert (ats_Gl != nullptr);
+      return ats_Gl;
+    }
+  else
+    {
+      assert (is_passive_transaction_server ());
```
hornetmj: How about assert(0)?

src/query/vacuum.c
```diff
@@ -2974,10 +2975,19 @@ vacuum_master_task::execute (cubthread::entry &thread_ref)
       return;
     }
 
+  assert (is_active_transaction_server());
```
hornetmj: It would be nice to move the first place(2960 line).

src/base/server_type.cpp
```diff
@@ -1,6 +1,6 @@
 /*
  * Copyright 2008 Search Solution Corporation
- * Copyright 2021 CUBRID Corporation
```
cristiarg, line 3: If this is incorrect, please also change it in `server_type_enum.hpp`. Thanks.

src/server/active_tran_server.cpp
```diff
@@ -36,6 +36,25 @@ active_tran_server::uses_remote_storage () const
...
+      ASSERT_ERROR ();
+      return error_code;
+    }
+
+  MVCCID oldest_mvccid = MVCCID_NULL;
+  std::memcpy (&oldest_mvccid, response_message.c_str (), sizeof (oldest_mvccid));
+
+  assert (MVCCID_IS_VALID (oldest_mvccid));
+
+  return oldest_mvccid;
```
hornetmj, line 60: I think that the exception handling codes are needed. The oldest_mvccid never be less than MVCCID_FIRST.

hornetmj (commented):
My comments are trivials. So, I approved.

hornetmj (approved):

src/base/server_type.cpp
```diff
@@ -198,8 +202,14 @@ void finalize_server_type ()
 {
   if (get_server_type () == SERVER_TYPE_TRANSACTION)
     {
-      if (is_passive_transaction_server ())
+      if (is_active_transaction_server ())
```
cristiarg: Nit.
I would leave the if the other way around - first PTS, and ATS on else.
It is much more probable that the PTS branch on if is executed most of the times - PTS's will be brought up and taken down more of that ATS.

src/base/server_type.cpp
wrlawodms: I followed `get_passive_tran_server_ptr`. I guess this interface allows it to be called from any.

src/query/vacuum.c
wrlawodms: OK

src/base/server_type.cpp
```diff
@@ -261,7 +285,7 @@ passive_tran_server *get_passive_tran_server_ptr ()
     }
   else
     {
-      assert (is_passive_transaction_server ());
+      assert (is_active_transaction_server ());
```
cristiarg: I think the assert was ok. It is intended to fail if it is not PTS. It is not intended to return null. In the sense that the function should not be called if the process is not a PTS.
```suggestion
      assert (is_passive_transaction_server ());
```

src/base/server_type.cpp
cristiarg: Here, same remark as the below comment.
But I might be wrong, as this is a newly added function.
```suggestion
      assert (is_active_transaction_server ());
```

joohok (approved):

src/server/active_tran_server.cpp
wrlawodms: It can be `MVCCID_ALL_VISIBLE`, actually. I did the check with `assert (MVCCID_IS_VALID (oldest_mvccid));`, but you mean we should handle the case?

src/server/active_tran_server.cpp
```diff
@@ -36,6 +36,25 @@ active_tran_server::uses_remote_storage () const
   return m_uses_remote_storage;
 }
 
+MVCCID
+active_tran_server::get_oldeset_active_mvccid_from_page_server () const
```
cristiarg: typo
```suggestion
active_tran_server::get_oldest_active_mvccid_from_page_server () const
```

src/server/active_tran_server.cpp
```diff
@@ -36,6 +36,25 @@ active_tran_server::uses_remote_storage () const
...
+{
+  std::string response_message;
+  const int error_code = send_receive (tran_to_page_request::GET_OLDEST_ACTIVE_MVCCID, std::string (), response_message);
+  if (error_code != NO_ERROR)
+    {
+      ASSERT_ERROR ();
+      return error_code;
```
wrlawodms: Oh, I should handle this properly..

src/query/vacuum.c
cristiarg: Clarification.
I think this assert works because a monolithic server is also an "active transaction server".

src/base/server_type.cpp
hornetmj: Others(PTS, PS) except for the ATS never create(construct) the ats_Gl. I thinks it means that they don't access this. Can we need to open this function to them?

wrlawodms:
> It seems that test_*_sql can't start the server. looking into it.

fixed.

src/server/active_tran_server.cpp
hornetmj: Oh. Sorry I mean the MVCCID_ALL_VISIBLE iso the MVCCID_FIRST. There is a room for 1 and 2. I worry about someone change the PS mechanism to can send 1 or 2 later. At that time this function can not filter even if wrong values.

src/base/server_type.cpp
wrlawodms: Me neither. I'd have done just like `assert (is_active_transaction_server)` without any branch. However, I just respect and follow the existing function 'cause it's not the scope I think. If you'd like to change it, how about asking critsti, who wrote the code?

src/server/active_tran_server.cpp
wrlawodms: > There is a room for 1 and 2.

I understand the concern, then how about `assert (MVCCID_IS_NORMAL (oldest_mvccid) || MVCCID_ALL_VISIBLE (oldest_mvccid))`. I think we can add some comments here for each value, which is described where it is sent, though.

>At that time this function can not filter even if wrong values.

Do you mean we should do more than assert?

src/server/active_tran_server.cpp
cristiarg: error code returned as `MVCCID` ?

src/server/page_server.cpp
```diff
@@ -163,6 +167,19 @@ page_server::connection_handler::receive_data_page_fetch (tran_server_conn_t::se
   push_async_response (&pgbuf_respond_data_fetch_page_request, std::move (a_sp));
 }
 
+void
+page_server::connection_handler::receive_oldest_active_mvccid_request (tran_server_conn_t::sequenced_payload &a_sp)
```
cristiarg: The function also does the handling.
I know that others - like `receive_stop_log_prior_dispatch` - are also mis-named but we'll fix them sometime.
```suggestion
page_server::connection_handler::handle_oldest_active_mvccid_request (tran_server_conn_t::sequenced_payload &a_sp)
```

src/base/server_type.cpp
wrlawodms: So how about just `assert (0)` being discussed in another thread? : https://github.com/CUBRID/cubrid/pull/3867/files#r994220649

src/base/server_type.cpp
wrlawodms: @hornetmj applied with https://github.com/CUBRID/cubrid/pull/3867/files#r994312453.

hornetmj: @cristiarg Please check this conversation.

src/server/active_tran_server.cpp
wrlawodms: I guess you reviewed the outdated code. Could you do it again?

src/server/active_tran_server.cpp
hornetmj: OK. I aggree your suggestion about assert.

> Do you mean we should more than assert?

I think that this function should have an exception handling code(kind of error code, error message) when it receive un-expected value from PS. But it is enough now to add new assert condition. consider this error handling mechanish later.

wrlawodms: OK, we can do as the above error handling:
```
  if (error_code != NO_ERROR)
     {
       return MVCCID_NULL;
     }
```
But, for that, I think we should define the error code. Let's go with just assertion for the time being.

src/query/vacuum.c
```diff
@@ -2974,10 +2977,30 @@ vacuum_master_task::execute (cubthread::entry &thread_ref)
       return;
     }
 
+  if (!ts_Gl->is_page_server_connected())
```
cristiarg: I think this assert will fail in the monolithic scenario.
should be guarded with `is_tran_server_with_remote_storage`.

src/query/vacuum.c
```diff
@@ -2974,10 +2977,30 @@ vacuum_master_task::execute (cubthread::entry &thread_ref)
...
+    return;
+  }
+
   PERF_UTIME_TRACKER_START (&thread_ref, &perf_tracker);
 
   m_oldest_visible_mvccid = log_Gl.mvcc_table.update_global_oldest_visible ();
   vacuum_er_log (VACUUM_ER_LOG_MASTER, "update oldest_visible = %lld", (long long int) m_oldest_visible_mvccid);
+ 
+  /* TODO temporary logging. The global one will be computed taking both into account, and the vacuum runs */ 
+  MVCCID global_pts_oldest_visible_mvccid = get_active_tran_server_ptr ()->get_oldest_active_mvccid_from_page_server ();
```
cristiarg: I also don't know how this should be handled. Because there are three scenarios where vacuum is active:
- monolithic server
- scalability ATS with Local Storage (a configuration mostly used for debugging & validation, and which should still be supported, but which is used *without* PTS's)
- scalability ATS proper with Remote Storage

Given the above, I think we should only request this info from PS when `is_tran_server_with_remote_storage`.
And some [manual] tests should be performed on the monolithic server, preferably before merge if possible.

hornetmj (approved):

src/query/vacuum.c
wrlawodms: Thank you. I didn't consider every scenario. I'll keep those scenarios in mind. After some manual tests, I'll re-request your review. By the way, to be clear, what you mean by `monolithic server` is ATS with Local Storage without PS?

From what I understand, the configurations are:
1. ATS + PS [+ PTS(es)] (normal scalability)
2. (1) + local stroage
3. ATS + PS on the same node (no PTS)
4. ATS + local storage (no PTS, no PS) (monolithic?)

Anything I'm missing? You can just link the related Jira issues.

cristiarg: I decided to call `monolithic server` what is now in `develop` (ie: without scalability/LETS).
I wanted to call it `stand-alone` but in CUBRID standalone means the `-SA` option. :)

My concert is that the `scalability_dev` branch has to be backward compatible with existing tests in various suites. The above changes in `vacuum_master_task::execute` will break **a lot** of existing tests.

cristiarg: >     1. ATS + PS [+ PTS(es)] (normal scalability)
>     2. (1) + local stroage
>     3. ATS + PS on the same node (no PTS)
>     4. ATS + local storage (no PTS, no PS) (monolithic?)

_1_ and _3_ are the same.
In essence `ATS (remote storage) + ....` because it does not matter where the PS executes and stores its data, ATS will still retrieve heap & log pages from another process (PS).

wrlawodms: Thank you for the clarification. One more thing: `(1) + local storage` is not the case? I thought that the local storage can be attached independently with the configuration. The comment from `pgbuf_read_page_from_file_or_page_server ` also says like that, I guess (local, connection to PS):
```
// Server type --- Storage Config Type --- Data Type --- Connection to PS --- Local Read --- Remote Read
//     PS                Local            Perm & Temp          Never              Yes            No
// -----------------------------------------------------------------------------------------------------
//     TS                Local                Perm              Yes               Yes           Yes
//                                                               No               Yes            No
//                                            Temp             Yes/No             Yes            No
//                       Remote               Perm              Yes                No           Yes
//                                            Temp              Yes               Yes            No
```
But, it fails to start a ATS with local storage when there is a PS. 

If the `(1) + local storage` is possible, I think we should use `is_page_server_connected()`.

wrlawodms: And, I guess we should use `is_page_server_connected()` in any case because it works only when the PS is connected. Even when `is_tran_server_with_remote_storage()` is true, the page server possibly hasn't been connected yet depending on the initialization order.

cristiarg: > uess we should use `is_page_server_connected()` in any case because it works only when the PS is connected. Even when `is_tran_server_with_remote_storage()` is true, the page server possibly hasn't been connected yet depending on the initialization

You have a point with this, but I still think `is_tran_server_with_remote_storage` is the more conservative approach:
- indeed, `is_page_server_connected` implies `is_tran_server_with_remote_storage`
- but `is_page_server_connected` is more of a technical prerequisite in the sense that:
  - if `is_tran_server_with_remote_storage` is true and `is_page_server_connected` is false, nothing should work and the TS should just log the error and crash/shutdown/refuse to continue

In a broader sense, `is_page_server_connected` should not even be used outside of `tran_server` (it is more of an implementation detail); visible from the current state where:
  - `is_page_server_connected` is only used 3 times outside `tran_server`
  - `is_tran_server_with_remote_storage` is used ~80 times all over the place

cristiarg: > In a broader sense, `is_page_server_connected` should not even be used outside of `tran_server` (it is more of an implementation detail); visible from the current state where:

Or, better put, `is_page_server_connected` should be a last resort check without which operation cannot continue - like it is now where it is used to check right before an interaction between TS and PS (page retrieval).

In the future, I assume there will be a keep-alive type of communicaction either between servers or between a server and master. And, with that, the state of the `is_page_server_connected` will be updated by the keep-alive state and its check will not be as superfluous as it is now.

wrlawodms: Always thank you for your elaboration.

cristiarg: To summarize. The following scenarios need to be covered:
- monolithic server (what is now in develop, without LETS)
  - vacuum needs to work as it was until now because of tests in various test suites that must continue to pass
- scalability ATS with Local Storage (a configuration mostly used for debugging & validation, and which should still be supported, but which is used without PTS's)
  - this is mostly a debug configuration where ATS is connected to a remote PS but also has its own local storage
  - the configuration is used to validate the transactional log replication mechanism (see functions `pgbuf_read_page_from_file_or_page_server` and `logpb_read_page_from_file_or_page_server`)
  - there are some tests implemented for this configuration but **no PTS's**
  - in this cas vacuum must work as for the monolithic scenario - ie: no requests for oldest active mvccid to the PS
- scalability ATS proper with Remote Storage
  - in this case vacuum must take the value(s) reported by PS into calculation

Given the above, I think `is_tran_server_with_remote_storage` fits the bill.

@wrlawodms Please add some variation of this comment as the rationale for the changes in `vacuum_master_task::execute` function. Thanks

src/query/vacuum.c
```diff
@@ -2978,6 +2981,36 @@ vacuum_master_task::execute (cubthread::entry &thread_ref)
...
+       * 1, 2 don't need to request the mvcc status from other server because there is no PTS,
+       * which are both filtered with is_tran_server_with_remote_storage().
+       */
+
+      /* TODO temporary logging. The global one will be computed taking both into account, and the vacuum runs */ 
+      MVCCID global_pts_oldest_visible_mvccid = get_active_tran_server_ptr ()->get_oldest_active_mvccid_from_page_server ();
+      if (global_pts_oldest_visible_mvccid == MVCCID_NULL)
+        {
+          assert (false);
+          vacuum_er_log (VACUUM_ER_LOG_MASTER, "%s", "Fail to get the oldest active mvccid across all PTS.");
```
cristiarg: It might ease debugging if error is logged just before a possible debug core.
```suggestion
          vacuum_er_log (VACUUM_ER_LOG_MASTER, "%s", "Fail to get the oldest active mvccid across all PTS.");
          assert (false);
```

cristiarg (approved):

PR #3869, opened 2022-10-13
[CBRD-24500] refactoring functions allocate index and classobj put index

ctshim:
http://jira.cubrid.org/browse/CBRD-24500

The allocate_index() and classbj_put_index() functions have many parameters.
But those parameters are the information that belonged to a structure.
Reduce the number of parameters to reconstruct the contents for ease of understanding.

Additionally, the sm_add_index() function was removed because it was deprecated.
The sm_is_global_only_constrain() function also has minor changes.

### DIFF
--- a/src/object/schema_manager.c
+++ b/src/object/schema_manager.c
@@ -437,11 +437,7 @@ static TP_DOMAIN *construct_index_key_domain (int n_atts, SM_ATTRIBUTE ** atts, 
 					      const int *prefix_lengths, int func_col_id, TP_DOMAIN * func_domain);
 static int collect_hier_class_info (MOP classop, DB_OBJLIST * subclasses, const char *constraint_name, int reverse,
 				    int *n_classes, int n_attrs, OID * oids, int *attr_ids, HFID * hfids);
-static int allocate_index (MOP classop, SM_CLASS * class_, DB_OBJLIST * subclasses, SM_ATTRIBUTE ** attrs,
-			   const int *asc_desc, const int *attrs_prefix_length, int unique_pk, int not_null,
-			   int reverse, const char *constraint_name, BTID * index, OID * fk_refcls_oid,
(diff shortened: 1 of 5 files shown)


src/object/schema_manager.c
```diff
@@ -11221,16 +11221,10 @@ allocate_foreign_key (MOP classop, SM_CLASS * class_, SM_CLASS_CONSTRAINT * con,
 	    }
 	}
     }
-  else
+  else if (allocate_index (classop, class_, subclasses, con))
```
hgryoo:
```suggestion
  else if (allocate_index (classop, class_, subclasses, con) != NO_ERROR)
```

src/object/schema_manager.c
```diff
@@ -11128,28 +11146,10 @@ allocate_unique_constraint (MOP classop, SM_CLASS * class_, SM_CLASS_CONSTRAINT
 	  shared_con = classobj_find_constraint_by_name (class_->constraints, con->shared_cons_name);
 	  con->index_btid = shared_con->index_btid;
 	}
-      else
+      else if (allocate_index (classop, class_, local_subclasses, con))
```
hgryoo:
```suggestion
      else if (allocate_index (classop, class_, local_subclasses, con) != NO_ERROR)
```

hgryoo (approved):

beyondykk9 (approved):

hyunikn (approved):

youngjinj (approved):

src/object/schema_manager.c
ctshim: OK, I will fix it.
Thanks.

src/object/schema_manager.c
ctshim: OK, I will fix it.
Thanks.

shparkcubrid (approved):

PR #3873, opened 2022-10-13
[CBRD-24509] Redundant data is sent to Java SP server in method scan

hgryoo:
http://jira.cubrid.org/browse/CBRD-24509

### DIFF
--- a/jsp/CMakeLists.txt
+++ b/jsp/CMakeLists.txt
@@ -53,6 +53,12 @@ add_custom_command(TARGET jsp_build
   COMMENT "Build Java SP Server with Ant ..."
 )
 
+add_custom_target(jsp_unittest)
+add_custom_command(TARGET jsp_unittest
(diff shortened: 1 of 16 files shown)


hgryoo:
I'm considering that maybe we can figure out redundant columns in method scan's list file in advance. In the get_single_tuple() function, they are filtered out, not to read, and assign a null value.

hgryoo:
> I'm considering that maybe we can figure out redundant columns in method scan's list file in advance. In the get_single_tuple() function, they are filtered out, not to read, and assign a null value.

If I filtered out for both C method and Java SP, C method causes regressions. So I've filtered out only for Java SP's argument.

hyunikn (approved):

shparkcubrid (approved):

ctshim (approved):

hgryoo:
I found that after the recent commit 64e4882 is merged, In test_sql, sql/_23_apricot_qa/_00_issues/bug_bts_8166/ tests are extreamly delayed.

hgryoo:
Only cbrd_24482.sql is delayed.

hgryoo:
> Only cbrd_24482.sql is delayed.

The hang is fixed at https://github.com/CUBRID/cubrid-testcases/pull/1317

beyondykk9 (approved):

shparkcubrid (approved):

youngjinj (approved):

PR #3876, opened 2022-10-14
Dummy PR - Test CI

hgryoo:


hgryoo:
cbrd_24473.sql causes hang because the optimization level is not restored, the next cases don't get the result. So all of the following cases take maximum waiting time and return NOK.

PR #3877, opened 2022-10-14
[LETS-561] separate append assigned mvccid from mvcc complete

cristiarg:
http://jira.cubrid.org/browse/LETS-561

It seems that, when the `unittests_snapshot` was written - before LETS developments - there was a separation between the MVCC and the transactional log layers.
During LETS developments, this separation was broken in {{logtb_complete_mvcc}} with the changes for http://jira.cubrid.org/browse/LETS-288.

Implementation:
- separate the part that appends the log record with the assigned MVCCID from the part that does the actual MVCC completion in separate functions
- call both functionalities - `logtb_append_assigned_mvcc_if_needed_and_complete_mvcc` - in all places
- except from the snapshot unittest where only the MVCC complete functionality is needed - `logtb_complete_mvcc`
- create an "internal" function that allows to check correctness of `tdes->mvccinfo->last_mvcc_lsa` via an extra parameter

### DIFF
--- a/src/transaction/log_tran_table.c
+++ b/src/transaction/log_tran_table.c
@@ -4090,6 +4090,38 @@ logtb_get_mvcc_snapshot (THREAD_ENTRY * thread_p)
 
 /*
  * logtb_complete_mvcc () - Called at commit or rollback, completes MVCC info
+ *			    for current transaction. If the transaction used
+ *			    an MVCCID but that id was not added to any of the
+ *			    transaction's log records, adds a specific log record to register the id.
+ *
+ * return	  : Void.
+ * thread_p (in)  : Thread entry.
+ * tdes (in)	  : Transaction descriptor.
+ * committed (in) : True if transaction was committed false if it was aborted.
+ */
+void
+logtb_append_assigned_mvcc_if_needed_and_complete_mvcc (THREAD_ENTRY * thread_p, LOG_TDES * tdes, bool committed)
+{
+  MVCC_INFO *const curr_mvcc_info = &tdes->mvccinfo;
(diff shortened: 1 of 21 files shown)


src/transaction/log_tran_table.c
```diff
@@ -4088,6 +4090,39 @@ logtb_get_mvcc_snapshot (THREAD_ENTRY * thread_p)
 
 }
 
+/*
+ * logtb_complete_mvcc () - Called at commit or rollback, completes MVCC info
```
wrlawodms: It should be modified.

src/transaction/log_tran_table.c
```diff
@@ -4099,6 +4134,13 @@ logtb_get_mvcc_snapshot (THREAD_ENTRY * thread_p)
  */
 void
 logtb_complete_mvcc (THREAD_ENTRY * thread_p, LOG_TDES * tdes, bool committed)
+{
+  constexpr bool check_null_last_mvcc_lsa = true;
```
wrlawodms, line 4140: Just out of curiosity, why don't you pass true directly, but use a temporal constexpr variable? Thanks.

wrlawodms (approved):

joohok (approved):

src/transaction/log_tran_table.c
cristiarg: No other reason than to make the semantics of the code clearer to the reader. With only a `true` passed as argument, the user would have to search the called function to see what that means. It is zero-cost from a performance point of view as a compiler, in release build, will abstract away the parameter altogether.

src/transaction/log_tran_table.c
cristiarg: Done

PR #3883, opened 2022-10-17
[LETS-564] Store channel id when connection between PS and TS is made successfully

joohok:
http://jira.cubrid.org/browse/LETS-564

Purpose

PS need to find and remove the oldest active mvcc id of a specific PTS using the channel ID, but when PS try to get the channel ID while the connection is disconnected, an invalid channel ID is created. This is because the socket value is already initialized to -1.

Therefore, rather than configuring the channel ID every time PS tries to get the channel ID, it is better to use the stored channel id which is stored when it is connected successfully.

### DIFF
--- a/src/communication/communication_channel.cpp
+++ b/src/communication/communication_channel.cpp
@@ -164,7 +164,6 @@ namespace cubcomm
 	return INTERNAL_CSS_ERROR;
       }
 
-    m_type = CHANNEL_TYPE::INITIATOR;
     m_socket = css_tcp_client_open (hostname, port);
 
     er_log_chn_debug ("[%s] Connect to %s:%d socket = %d.\n", get_channel_id ().c_str (), hostname, port, m_socket);
@@ -174,6 +173,7 @@ namespace cubcomm
 	return REQUEST_REFUSED;
       }
 
+    m_type = CHANNEL_TYPE::INITIATOR;
     m_hostname = hostname;
     m_port = port;
 
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -52,13 +52,13 @@ page_server::~page_server ()
 }
 
 page_server::connection_handler::connection_handler (cubcomm::channel &chn, transaction_server_type server_type,
-    page_server &ps, const std::string &underlying_channel_id)
+    page_server &ps)
   : m_server_type { server_type }
-  , m_underlying_channel_id { underlying_channel_id }
+  , m_connection_id { chn.get_channel_id () }
   , m_ps (ps)
   , m_abnormal_tran_server_disconnect { false }
 {
-  assert (!m_underlying_channel_id.empty ());
+  assert (!m_connection_id.empty ());
 
   constexpr size_t RESPONSE_PARTITIONING_SIZE = 1; // Arbitrarily chosen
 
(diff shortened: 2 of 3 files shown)


src/server/page_server.hpp
```diff
@@ -152,6 +153,7 @@ class page_server
 	 * the peer transaction server and the check will no longer be valid
 	 */
 	const transaction_server_type m_server_type;
+	const std::string m_underlying_channel_id;
```
wrlawodms: I think it's not the "channel id" anymore. How about "m_connection_id"? It would be confusing with `m_conn->get_underlying_channel_id ()` in my opinion.

src/server/page_server.cpp
```diff
@@ -485,12 +492,14 @@ page_server::set_passive_tran_server_connection (cubcomm::channel &&chn)
...
   chn.set_channel_name ("PTS_PS_comm");
 
+  assert (chn.is_connection_alive ());
   const auto channel_id = chn.get_channel_id ();
 
   er_log_debug (ARG_FILE_LINE, "Passive transaction server connected to this page server. Channel id: %s.\n",
 		channel_id.c_str ());
 
-  m_passive_tran_server_conn.emplace_back (new connection_handler (chn, transaction_server_type::PASSIVE, *this));
+  m_passive_tran_server_conn.emplace_back (new connection_handler (chn, transaction_server_type::PASSIVE, *this,
```
wrlawodms: I don't we need to pass the channel id explicitly. the connection_handler already got the channel and we can just use ch.get_channel_id() in the contr of the connection_handler.

src/communication/communication_channel.cpp
```diff
@@ -174,6 +173,7 @@ namespace cubcomm
 	return REQUEST_REFUSED;
       }
 
+    m_type = CHANNEL_TYPE::INITIATOR;
```
wrlawodms, line 176: Why is it moved?

src/server/page_server.hpp
wrlawodms: It requires `get_channel_id ()` to be modified along.

src/server/page_server.cpp
```diff
@@ -192,7 +195,7 @@ page_server::connection_handler::receive_oldest_active_mvccid (tran_server_conn_
 
   const auto oldest_mvccid = *reinterpret_cast<const MVCCID *const> (a_sp.pull_payload().c_str());
 
-  m_ps.m_pts_mvcc_tracker.update_oldest_active_mvccid (get_channel_id (), oldest_mvccid);
+  m_ps.m_pts_mvcc_tracker.update_oldest_active_mvccid (m_underlying_channel_id, oldest_mvccid);
```
wrlawodms: Why don't you use the accessor, get_channel_id(), but the variable directly?

src/server/page_server.cpp
joohok: Oh, OK. I understand.
We can just pass the channel (`chn`), so we don't need to pass the channel ID.

src/server/page_server.hpp
joohok: OK, that can be confused. I will follow your suggestion.

src/server/page_server.cpp
joohok: I will replace it with the accessor.

src/communication/communication_channel.cpp
joohok: it seems weird that only `m_type` initialized even if m_socket is failed to open. 

Not related this issue, but just a small refactoring.

wrlawodms (approved):

cristiarg (approved):

hornetmj (approved):

PR #3884, opened 2022-10-17
[CBRD-24502] The problem of updating statistical information when performing DDL.

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-24502

It is modified so that statistics are not updated during DDL. DDL includes index creation and drop, such as 'CREATE INDEX'.
The optimizer sets the *default selectivity* for new statistics that are not updated.

### DIFF
--- a/src/object/schema_manager.c
+++ b/src/object/schema_manager.c
@@ -4303,6 +4303,76 @@ sm_update_statistics (MOP classop, bool with_fullscan)
 }
 
 /*
+ * sm_update_statistics_without_gathering_stats () - fetch the statistics and
+ *    						     cache them with the class.
+ *   return: NO_ERROR on success, non-zero for ERROR
+ *   classop(in): class object
+ *   with_fullscan(in): true iff WITH FULLSCAN
+ *
+ * NOTE: We will delay updating statistics until a transaction is committed
+ *       when it is requested during other processing, such as
+ *       "alter table ..." or "create index ...".
+ */
+int
+sm_update_statistics_without_gathering_stats (MOP classop, bool with_fullscan)
+{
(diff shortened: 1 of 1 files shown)


src/object/schema_manager.c
```diff
@@ -4302,6 +4302,76 @@ sm_update_statistics (MOP classop, bool with_fullscan)
...
+
+  if (classop != NULL && !OID_ISTEMP (WS_OID (classop)))
+    {
+      is_class = locator_is_class (classop, DB_FETCH_QUERY_READ);
+      if (is_class < 0)
+	{
+	  return is_class;
+	}
+    }
+  if (is_class > 0)
```
beyondykk9, line 4335: In "locator_is_class", return 1 for root class and 0 for other class. This processing is correct or valid?

hgryoo (approved):

ctshim (approved):

src/object/schema_manager.c
shparkcubrid: If mop->class_mop is the root class, then mop is the class. It's same routine as sm_update_statistics() .

hyunikn (approved):

src/object/schema_manager.c
beyondykk9: is "is_class >= 0 correct?

shparkcubrid: In "locator_is_class", return 1 for root class. It's wrong. If mop->class_mop is the root class, then the mop is the class. so 1 is only for class.

beyondykk9 (approved):

youngjinj (approved):

PR #3886, opened 2022-10-17
[CBRD-24492] Loading native library in Java SP causes UnsatisfiedLinkError

hgryoo:
http://jira.cubrid.org/browse/CBRD-24492

**Purpose**
Java SP server creates a new classloader per session. A classloader load classes from $CUBRID/java. If .class or .jar file is modified in the path, The new classloader reloads it. (Dynamic Class Loading). It causes UnsatisfiedLinkError: already loaded in another classloader.

**Implementation**
- The StoredProcedureClassLoader is reverted as a singleton
- The class loader only checks the root folder is modified (by Java NIO API). I expect it will show better performance than the previous one, which checked last modified all .class and .jar files in the root folder.
- If any file is modified in the root, the loader is re-instantiated. It could cause the same problem of this issue, but it can be bypassed by restarting the javasp server.

### DIFF
--- a/src/jsp/com/cubrid/jsp/StoredProcedureClassLoader.java
+++ b/src/jsp/com/cubrid/jsp/StoredProcedureClassLoader.java
@@ -49,11 +49,9 @@ public class StoredProcedureClassLoader extends URLClassLoader {
     private FileTime lastModified = null;
 
     /* For singleton */
-    public synchronized static StoredProcedureClassLoader getInstance() {
+    public static synchronized StoredProcedureClassLoader getInstance() {
         if (instance == null) {
-                if (instance == null) {
-                    instance = new StoredProcedureClassLoader();
-                }
+            instance = new StoredProcedureClassLoader();
         }
 
         return instance;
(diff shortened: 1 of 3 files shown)


src/jsp/com/cubrid/jsp/StoredProcedureClassLoader.java
```diff
@@ -31,84 +31,68 @@
...
+    private static final String ROOT_PATH = Server.getSpPath() + "/java/";
+    private static final Path root = Paths.get(ROOT_PATH);
 
-    public StoredProcedureClassLoader() {
+    private FileTime lastModified = null;
+
+    /* For singleton */
+    public synchronized static StoredProcedureClassLoader getInstance() {
+        if (instance == null) {
+                if (instance == null) {
```
hyunikn: same as the above line.

hgryoo: fixed, thank you.

beyondykk9 (approved):

youngjinj (approved):

hyunikn (approved):

shparkcubrid (approved):

ctshim (approved):

PR #3887, opened 2022-10-17
[LETS-537] add a dump function when an atomic sequence is applied

cristiarg:

### DIFF
--- a/src/transaction/atomic_replication_helper.cpp
+++ b/src/transaction/atomic_replication_helper.cpp
@@ -365,6 +365,39 @@ namespace cublog
     return false;
   }
 
+#if !defined (NDEBUG)
+  void
+  atomic_replication_helper::atomic_log_sequence::dump ()
+  {
+    char buf[PATH_MAX];
+    char *buf_ptr = buf;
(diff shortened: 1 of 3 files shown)


src/transaction/atomic_replication_helper.cpp
```diff
@@ -365,6 +365,39 @@ namespace cublog
     return false;
   }
 
+#if !defined (NDEBUG)
+  void
+  atomic_replication_helper::atomic_log_sequence::dump ()
+  {
+    char buf[PATH_MAX];
```
wrlawodms: I'm worried that it seems easy to overflow the buffer of PATH_MAX(256).

wrlawodms (approved):

hornetmj (approved):

src/transaction/atomic_replication_helper.cpp
cristiarg: ~Correct. I added some code to prevent overflow but not complete. I will complete it.~

Please disregard above comment.
I understand what you mean: `PATH_MAX` can have too small a value.

PR #3892, opened 2022-10-18
[CBRD-24516] replace the information which is printed in "show trace" from aggregate function optimization

beyondykk9:
http://jira.cubrid.org/browse/CBRD-24516

The trace information is printed from "show trace" command, however, the aggregate optimzation-related information is not printed.

The information including aggregate optimized should be shown from "show trace" in case of aggregate optimization enabled. The below is an sample information.

```
trace
======================
  '
Query Plan:
  INDEX AGG OPTIMIZED (dba.mx)  
  rewritten query: select max([dba.mx].a) from [dba.mx] [dba.mx]

Trace Statistics:
  SELECT (time: 0, fetch: 12, ioread: 1)
    SCAN (table: dba.mx), (aggregate optimized, time: 0, fetch: 0, ioread: 0)
```
 
The modified code has 2 parts in "agg-optimized" setting.

1) optimizer
    the plan's scan has "plan_un.scan.agg_optimized"

2) scanner
    executor's scanner has "scan_stats.agg_optimized_scan"


beyondykk9:
The failed cases should be modified to reflect the changed TRACE text as following example:
See the "INDEX AGG OPTIMIZED" and "aggregate optimized" carefully.

```
  INDEX AGG OPTIMIZED (dba.mx)  rewritten query: select max([dba.mx].a) from [dba.mx] [dba.mx]Trace Statistics:
    SELECT (time: 0, fetch: 12, ioread: 1)
      SCAN (table: dba.mx), (aggregate optimized, time: 0, fetch: 0, ioread: 0)
```

hyunikn (approved):

src/optimizer/query_planner.c
```diff
@@ -11567,7 +11567,14 @@ qo_plan_scan_print_text (FILE * fp, QO_PLAN * plan, int indent)
...
     case QO_SCANMETHOD_SEQ_SCAN:
-      fprintf (fp, "TABLE SCAN (%s)", class_name);
+      if (plan->plan_un.scan.agg_optimized)
+	{
+	  fprintf (fp, "INDEX AGG OPTIMIZED (%s)", class_name);
+	}
+      else
+	{
+	  fprintf (fp, "TABLE SCAN (%s)", class_name);
+	}
```
shparkcubrid: Since it is determined by the executor, it is not possible to output to the plan.

beyondykk9: I think, however, the plan could be different from executing. Therefore, we need to print both of the plan and scan method  in executing together in order to check whether the plan and scan method is same or different.

ctshim (approved):

hgryoo (approved):

src/optimizer/query_planner.c
shparkcubrid: I don't think so. Inaccurate plan information will only confuse users.

shparkcubrid (approved):

youngjinj (approved):

PR #3896, opened 2022-10-18
[LETS-550] ATS vacuum daemon checks and aggregates MVCC related information from PS

wrlawodms:
http://jira.cubrid.org/browse/LETS-550

- the oldest visible mvccid is updated in `mvcctable::update_global_oldest_visible()`.
- I decided to hand over the PTS-across oldest mvccid as an argument, not get it in the `mvcctable::update_global_oldest_visible()`, because prevents mvcctable from caring about network communication or something. I think it's better in terms of abstraction.
- See the comment section on the Jira issue to see which tests are done manually.

### DIFF
--- a/src/query/vacuum.c
+++ b/src/query/vacuum.c
@@ -2984,6 +2984,13 @@ vacuum_master_task::execute (cubthread::entry &thread_ref)
  
   /* TODO temporary logging. The global one will be computed taking both into account, and the vacuum runs */ 
   MVCCID global_pts_oldest_visible_mvccid = get_active_tran_server_ptr ()->get_oldeset_active_mvccid_from_page_server ();
+  if (global_pts_oldest_visible_mvccid == MVCCID_NULL)
+  {
+    assert (false);
+    vacuum_er_log (VACUUM_ER_LOG_MASTER, "%s", "Fail to get the oldest active mvccid across all PTS.");
+    return;
+  }
+
   er_log_debug (ARG_FILE_LINE, "ats oldest_visible = %llu, pts global_oldest_visible = %llu",
       (long long int) m_oldest_visible_mvccid, global_pts_oldest_visible_mvccid);
 
--- a/src/query/vacuum.c
+++ b/src/query/vacuum.c
@@ -2958,6 +2958,8 @@ vacuum_master_task::execute (cubthread::entry &thread_ref)
 {
   PERF_UTIME_TRACKER perf_tracker;
 
+  assert (is_active_transaction_server());
+
   if (prm_get_bool_value (PRM_ID_DISABLE_VACUUM))
     {
       return;
@@ -2974,8 +2976,6 @@ vacuum_master_task::execute (cubthread::entry &thread_ref)
       // check if boot is aborted
       return;
     }
-
-  assert (is_active_transaction_server());
 
   PERF_UTIME_TRACKER_START (&thread_ref, &perf_tracker);
 
--- a/src/query/vacuum.c
+++ b/src/query/vacuum.c
@@ -2958,6 +2958,8 @@ vacuum_master_task::execute (cubthread::entry &thread_ref)
 {
   PERF_UTIME_TRACKER perf_tracker;
 
+  assert (is_active_transaction_server());
+
   if (prm_get_bool_value (PRM_ID_DISABLE_VACUUM))
     {
       return;
@@ -2975,7 +2977,11 @@ vacuum_master_task::execute (cubthread::entry &thread_ref)
       return;
     }
 
-  assert (is_active_transaction_server());
+  if (!ts_Gl->is_page_server_connected()) 
+  {
+    // have to be connected to get mvcc status of PTSes.
+    return;
+  }
 
   PERF_UTIME_TRACKER_START (&thread_ref, &perf_tracker);
 
@@ -2983,7 +2989,14 @@ vacuum_master_task::execute (cubthread::entry &thread_ref)
   vacuum_er_log (VACUUM_ER_LOG_MASTER, "update oldest_visible = %lld", (long long int) m_oldest_visible_mvccid);
  
   /* TODO temporary logging. The global one will be computed taking both into account, and the vacuum runs */ 
-  MVCCID global_pts_oldest_visible_mvccid = get_active_tran_server_ptr()->get_oldeset_active_mvccid_from_page_server();
+  MVCCID global_pts_oldest_visible_mvccid = get_active_tran_server_ptr ()->get_oldest_active_mvccid_from_page_server ();
+  if (global_pts_oldest_visible_mvccid == MVCCID_NULL)
+  {
+    assert (false);
+    vacuum_er_log (VACUUM_ER_LOG_MASTER, "%s", "Fail to get the oldest active mvccid across all PTS.");
+    return;
+  }
+
   er_log_debug (ARG_FILE_LINE, "ats oldest_visible = %llu, pts global_oldest_visible = %llu",
       (long long int) m_oldest_visible_mvccid, global_pts_oldest_visible_mvccid);
 
--- a/src/query/vacuum.c
+++ b/src/query/vacuum.c
@@ -2983,7 +2983,7 @@ vacuum_master_task::execute (cubthread::entry &thread_ref)
   vacuum_er_log (VACUUM_ER_LOG_MASTER, "update oldest_visible = %lld", (long long int) m_oldest_visible_mvccid);
  
   /* TODO temporary logging. The global one will be computed taking both into account, and the vacuum runs */ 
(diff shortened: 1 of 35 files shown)


hornetmj (approved):

src/transaction/mvcc_table.cpp
```diff
@@ -707,17 +707,35 @@ mvcctable::get_global_oldest_visible () const
 }
 
 MVCCID
-mvcctable::update_global_oldest_visible ()
+mvcctable::update_global_oldest_visible (const MVCCID pts_oldest_visible /* = MVCCID_LAST */)
 {
-  if (m_ov_lock_count == 0)
+  if (pts_oldest_visible != MVCCID_ALL_VISIBLE)
```
joohok, line 738:
```suggestion
  if (pts_oldest_visible != MVCCID_LAST)
```

I think `pts_oldest_visible` can be MVCCID_ALL_VISIBLE,
if PTS is just connected to PS and PTS did not send any mvccid to PS yet.

joohok: So, I think if pts_oldest_visible is `MVCCID_ALL_VISIBLE`, then it needs to be considered.

And if we have to consider the `MVCCID_ALL_VISIBLE`, then the `assert (m_oldest_visible.load () <= ats_oldest_visible);` can be failed.

wrlawodms: It's intended. If it is MVCCID_ALL_VISIBLE, it doesn't update the oldest mvccid and, in turn, the vacuum doesn't go forward and will retry in the next execution. Plus, `MVCCID_LAST` is also intended, which means no PTS exists.

joohok: Oh, OK. Thanks for explain.

joohok (approved):

src/query/vacuum.c
```diff
@@ -2998,19 +2997,17 @@ vacuum_master_task::execute (cubthread::entry &thread_ref)
...
           return;
         }
-
-      er_log_debug (ARG_FILE_LINE, "ats oldest_visible = %llu, pts global_oldest_visible = %llu",
-          (long long int) m_oldest_visible_mvccid, global_pts_oldest_visible_mvccid);
     }
-
-  return;
+  
+  m_oldest_visible_mvccid = log_Gl.mvcc_table.update_global_oldest_visible (global_pts_oldest_visible_mvccid);
```
cristiarg: Nit.
I actually don't really like optional parameters on functions. They are non-explicit and notoriously prone to errors.
Given that `update_global_oldest_visible` is a small function, I would have overloaded the function.

```
{
  MVCCID global_pts_oldest_visible_mvccid = MVCCID_LAST;
  // request to PS
  // new call and logging - with explicit logging whether vacuum will been skipped or not
  // maybe some perfmon update if relevant
}
else
{
  // old call and logging
}
```

My 2cents. Your choice.

src/transaction/mvcc_table.cpp
```diff
@@ -707,17 +707,35 @@ mvcctable::get_global_oldest_visible () const
...
 	{
-	  assert (m_oldest_visible.load () <= oldest_visible);
-	  m_oldest_visible.store (oldest_visible);
+	  MVCCID ats_oldest_visible = compute_oldest_visible_mvccid ();
+	  if (m_ov_lock_count == 0)
+	    {
+	      /*
+	       * The assert below must be met to confirm there is no desynchronizaition.
+	       * But, it's allowed for now.
+	       * It is going to be dealt with soon in http://jira.cubrid.org/browse/LETS-563.
```
cristiarg:
```suggestion
	       * TODO: It is going to be dealt with soon in http://jira.cubrid.org/browse/LETS-563.
```

src/transaction/mvcc_table.cpp
```diff
@@ -707,17 +707,35 @@ mvcctable::get_global_oldest_visible () const
 }
 
 MVCCID
-mvcctable::update_global_oldest_visible ()
+mvcctable::update_global_oldest_visible (const MVCCID pts_oldest_visible /* = MVCCID_LAST */)
```
cristiarg: If you decide to leave the optional param in place, please comment and explain all use cases:
- monolithic (`MVCCID_LAST`) - logic will fall-through and work as before
- scalability: skip vacuum (`MVCCID_ALL_VISIBLE`) - value `MVCCID_ALL_VISIBLE` is received from PS aggregation and means "skip vacuum regardless of MVCCID values because the cluster is in an transitory state with at least one PTS booting-up (or being shut down)"
- scalability: update for vacuum to execute or not (!= `MVCCID_ALL_VISIBLE` and != `MVCCID_LAST`) - the value will be updated depending on the value received from the PS[s] aggregation

I hope I got this right.

src/transaction/mvcc_table.cpp
```diff
@@ -707,17 +707,35 @@ mvcctable::get_global_oldest_visible () const
...
+		  m_oldest_visible.store (ats_oldest_visible);
+		}
+	      else
+		{
+		  m_oldest_visible.store (pts_oldest_visible);
+		}
+
+	    }
 	}
     }
```
cristiarg, line 763:
```suggestion
    }
else
    {
      // some debug logging about vacuum being skipped?
    }
```

src/query/vacuum.c
```diff
@@ -2998,19 +2997,17 @@ vacuum_master_task::execute (cubthread::entry &thread_ref)
...
 
       /* TODO temporary logging. The global one will be computed taking both into account, and the vacuum runs */ 
-      MVCCID global_pts_oldest_visible_mvccid = get_active_tran_server_ptr ()->get_oldest_active_mvccid_from_page_server ();
+      global_pts_oldest_visible_mvccid = get_active_tran_server_ptr ()->get_oldest_active_mvccid_from_page_server ();
       if (global_pts_oldest_visible_mvccid == MVCCID_NULL)
         {
           vacuum_er_log (VACUUM_ER_LOG_MASTER, "%s", "Fail to get the oldest active mvccid across all PTS.");
           assert (false);
           return;
         }
```
cristiarg, line 3005:
```suggestion
        }
        assert (MVCCID_ALL_VISIBLE == global_pts_oldest_visible_mvccid || MVCCID_IS_NORMAL (global_pts_oldest_visible_mvccid));
```

src/transaction/mvcc_table.cpp
```diff
@@ -707,17 +707,35 @@ mvcctable::get_global_oldest_visible () const
...
+	       */
+	      // assert (m_oldest_visible.load () <= pts_oldest_visible);
+	      assert (m_oldest_visible.load () <= ats_oldest_visible);
+	      if (ats_oldest_visible < pts_oldest_visible)
+		{
+		  m_oldest_visible.store (ats_oldest_visible);
+		}
+	      else
+		{
+		  m_oldest_visible.store (pts_oldest_visible);
```
cristiarg, line 758: Should we assert that the value never goes backward? Or should this happen inherently by design?

src/query/vacuum.c
wrlawodms: I'll follow your advice. Thank you.

src/transaction/mvcc_table.cpp
wrlawodms: Some fixes:
> "skip vacuum regardless of MVCCID values because the cluster is in an inconsistent state with at least one PTS booting-up (or being shut down)"

not during shutdown

> scalability: update for vacuum to execute or not (!= MVCCID_ALL_VISIBLE and != MVCCID_LAST)

It is updated even if ==MVCCID_LAST, at this point, just the mvccid of ATS is considered and the vaccum can be run on it. I think you might've said the same one, but rephrase it with my words to be clear.

I'll add a comment on the overloaded one.

src/transaction/mvcc_table.cpp
wrlawodms: No, it must not. The asserts above check that, but one is commented out for now because of [LETS-563](http://jira.cubrid.org/browse/LETS-563). The assert is easy to fail for a while when a new PTS is connected.

cristiarg: Right, I should have noticed.

src/query/vacuum.c
wrlawodms: It's done in `get_oldest_active_mvccid_from_page_server()`, and I'll add in `update_global_oldest_visible(MVCCID)` too.

src/transaction/mvcc_table.cpp
wrlawodms: I have added one in the caller because it's a vacuum error log.
`"PTS-across oldest_visible_mvccid = %lld"`. It doesn't say explicitly that the oldest mvcc id is not updated and the vacuum probably won't run, but I think it's enough to identify what's going on.

cristiarg (approved):

PR #3897, opened 2022-10-19
[CBRD-24522] Shutdown cub_javasp when heartbeat process is stopped

hgryoo:
http://jira.cubrid.org/browse/CBRD-24522

### DIFF
--- a/src/executables/util_service.c
+++ b/src/executables/util_service.c
@@ -270,7 +270,7 @@ static int process_broker (int command_type, int argc, const char **argv, bool p
 static int process_gateway (int command_type, int argc, const char **argv, bool process_window_service);
 static int process_manager (int command_type, bool process_window_service);
 static int process_javasp (int command_type, int argc, const char **argv, bool show_usage, bool suppress_message,
-			   bool process_window_service);
+			   bool process_window_service, bool ha_mode);
 static int process_javasp_start (const char *db_name, bool suppress_message, bool process_window_service);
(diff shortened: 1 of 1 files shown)


src/executables/util_service.c
```diff
@@ -3977,6 +3981,9 @@ us_hb_deactivate (const char *hostname, bool immediate_stop)
       args[opt_idx++] = COMMDB_HB_DEACT_IMMEDIATELY;
     }
 
+  /* stop javasp server */
+  (void) process_javasp (STOP, 0, NULL, false, true, false, false);
```
mhoh3963: Is it right that the last arg is false ?
I think that the arg should be true for checking HA server.

hgryoo:
```suggestion
  (void) process_javasp (STOP, 0, NULL, false, true, false, true);
```

src/executables/util_service.c
```diff
@@ -4025,6 +4032,9 @@ us_hb_process_stop (HA_CONF * ha_conf, const char *db_name)
 
   print_message (stdout, MSGCAT_UTIL_GENERIC_START_STOP_2S, PRINT_HA_PROCS_NAME, PRINT_CMD_STOP);
 
+  /* stop javasp server */
+  (void) process_javasp (STOP, 0, NULL, false, true, false, false);
```
hgryoo:
```suggestion
  (void) process_javasp (STOP, 0, NULL, false, true, false, true);
```

mhoh3963 (approved):

beyondykk9 (approved):

PR #3909, opened 2022-10-24
[CBRD-24517] Records -10006 and -10007 error logs to the cas log file.

airnet73:
http://jira.cubrid.org/browse/CBRD-24517

Purpose
Currently, when an error -10006 or -10007 occurs, it is not recorded in cas.log, and the error code is delivered only to the client.
There is no -10006, -10007 error log recorded in cas.log, so it is difficult to analyze when the error occurred.

Implementation
* Log -10006 and -10007 errors in sql log
* If the environment variable CUBRID_SQL_LOG_TRACE=on is set, the log is also recorded in the error log.

Remarks
N/A


src/broker/cas_error.c
```diff
@@ -52,6 +52,8 @@
 #include "broker_filename.h"
 #include "cas_sql_log2.h"
 
+#define CUBRID_SQL_LOG_TRACE        "CUBRID_SQL_LOG_TRACE"
```
mhoh3963: It seems that the env name could be mistaken for users to trace sql.
How about changing the name to CUBRID_CAS_ERR_TRACE ?

mhoh3963 (approved):

src/broker/cas_error.c
airnet73: Thanks for the advice.
Changed it to CUBRID_CAS_ERR_TRACE.

kisoo-han (approved):
Substitute the correct language for the messages in the message file next time.

PR #3910, opened 2022-10-24
[LETS-545] [LETS-537] [LETS-425] refactor Passive Transaction Server atomic replication with encountered use cases

cristiarg:
http://jira.cubrid.org/browse/LETS-545

Refactor has the purpose of allowing more control of when to actually "draw" fences to apply the log records within an atomic replication sequence within a transaction.
An atomic replication sequence can be introduced by one of the following log records: `LOG_START_ATOMIC_REPL`, `LOG_SYSOP_ATOMIC_START`.
An atomic replication sequence can be concluded by one of the following log records: `LOG_END_ATOMIC_REPL`, `LOG_SYSOP_END`, `LOG_SYSOP_START_POSTPONE`.
These are called, in the context of the modified code, "control log records".
And are added together with the "proper" log records to an atomic sequence.

Dynamics in `atomic_replication_helper`:
- each processed "control" log record which can be part of an atomic replication sequence is added - with either `append_control_log` or `append_control_log_sysop_end`
  - at this moment, there might already exist "proper" log records in the sequence which will be applied and the pages unfixed - call to `atomic_log_sequence::apply_and_unfix`
  - calling `atomic_log_sequence::apply_and_unfix` will process all "proper" log records and also dispose unfix the pages; the implementation in `page_ptr_bookkeeping` will take care whether the page must still be kept fixed or not
- each "proper" log record for which there is an atomic replication sequence open, will be added - `add_atomic_replication_log`
- after a "control" log record is added to the sequence, the sequence will be checked if it can be "purged" - with "atomic_log_sequence::can_purge"
- a sequence can be purged when certain conditions are met - more details about the precise scenarios implemented is described in the comment for the `atomic_replication_helper` - the function `can_purge` implements precisely those scenarios

Relevant changes in `atomic_replicator`:
- a lot of the logic has actually been offloaded to the helper class
- a `TODO` was added for `LOG_ABORT` due to a scenario involving crash and recovery in Active Transaction Server; but more details implementation is needed - a separate issue will be added and a back-link to the issue will be added in code

Other:
- added debug - default off - `er_log_pts_atomic_repl_debug` system parameter specifically for PTS atomic replication
- added relevant `dump` functions for all functional entities to help in debugging and

PS: there is no point in trying to review by comparing new code with the old code; most of the core part related to replication has changed; easiest is to follow above explanation

### DIFF
--- a/src/base/system_parameter.c
+++ b/src/base/system_parameter.c
@@ -708,6 +708,7 @@ static const char sysprm_ha_conf_file_name[] = "cubrid_ha.conf";
 #define PRM_NAME_ER_LOG_COMMIT_CONFIRM "er_log_commit_confirm"
 #define PRM_NAME_ER_LOG_CALC_REPL_DELAY "er_log_calculate_replication_delay"
 #define PRM_NAME_ER_LOG_PTS_REPL_DEBUG "er_log_pts_repl_debug"
+#define PRM_NAME_ER_LOG_PTS_ATOMIC_REPL_DEBUG "er_log_pts_atomic_repl_debug"
 
 #define PRM_NAME_RECOVERY_PARALLEL_COUNT "recovery_parallel_count"
 #define PRM_NAME_RECOVERY_PARALLEL_TASK_DEBUG "recovery_parallel_task_debug"
@@ -2435,6 +2436,10 @@ static unsigned int prm_er_log_calc_repl_delay_flag = 0;
 bool PRM_ER_LOG_PTS_REPL_DEBUG = true;
 static bool prm_er_log_pts_repl_debug_default = false;
 static unsigned int prm_er_log_pts_repl_debug_flag = 0;
+
+bool PRM_ER_LOG_PTS_ATOMIC_REPL_DEBUG = true;
+static bool prm_er_log_pts_atomic_repl_debug_default = false;
+static unsigned int prm_er_log_pts_atomic_repl_debug_flag = 0;
 
 static unsigned int prm_recovery_parallel_count_flag = 0;
 static int prm_recovery_parallel_count_default = 8;
@@ -6346,6 +6351,19 @@ static SYSPRM_PARAM prm_Def[] = {
    (char *) NULL,
    (DUP_PRM_FUNC) NULL,
    (DUP_PRM_FUNC) NULL},
+  {PRM_ID_ER_LOG_PTS_ATOMIC_REPL_DEBUG,
+   PRM_NAME_ER_LOG_PTS_ATOMIC_REPL_DEBUG,
+   (PRM_HIDDEN | PRM_FOR_SERVER | PRM_USER_CHANGE),
+   PRM_BOOLEAN,
+   &prm_er_log_pts_atomic_repl_debug_flag,
+   (void *) &prm_er_log_pts_atomic_repl_debug_default,
+   (void *) &PRM_ER_LOG_PTS_ATOMIC_REPL_DEBUG,
+   (void *) NULL,
+   (void *) NULL,
+   (char *) NULL,
+   (DUP_PRM_FUNC) NULL,
+   (DUP_PRM_FUNC) NULL},
+
   {PRM_ID_RECOVERY_PARALLEL_COUNT,
    PRM_NAME_RECOVERY_PARALLEL_COUNT,
    (PRM_FOR_SERVER),
--- a/src/transaction/atomic_replication_helper.cpp
+++ b/src/transaction/atomic_replication_helper.cpp
@@ -220,7 +220,8 @@ namespace cublog
 	return;
       }
 
-    sequence_it->second.apply_and_unfix_sequence (thread_p);
+    atomic_log_sequence &sequence = sequence_it->second;
+    sequence.apply_and_unfix_sequence (thread_p);
     m_sequences_map.erase (sequence_it);
 
 #if !defined (NDEBUG)
@@ -365,18 +366,63 @@ namespace cublog
   }
 
   void
+  atomic_replication_helper::atomic_log_sequence::dump ()
+  {
+#if !defined (NDEBUG)
+    char buf[PATH_MAX];
+    char *buf_ptr = buf;
+    int written = 0;
+    int left = PATH_MAX;
+
+    written = snprintf (buf_ptr, (size_t)left, "[ATOMIC_REPL] start_lsa = %lld|%d  is_sysop = %d"
+			"  postpone_started = %d  end_pospone_count = %d\n",
+			LSA_AS_ARGS (&m_start_lsa), (int)m_is_sysop,
+			(int)m_postpone_started, m_end_pospone_count);
+    assert (written > 0);
+    buf_ptr += written;
+    assert (left >= written);
+    left -= written;
+
+    for (const atomic_log_entry &log_entry : m_log_vec)
+      {
+	written = snprintf (buf_ptr, (size_t)left, "  LSA = %lld|%d  vpid = %d|%d\n  rcvindex = %s\n",
+			    LSA_AS_ARGS (&log_entry.m_record_lsa),
+			    VPID_AS_ARGS (&log_entry.m_vpid),
+			    rv_rcvindex_string (log_entry.m_record_index));
+	assert (written > 0);
+	buf_ptr += written;
+	assert (left >= written);
+	left -= written;
+      }
+    _er_log_debug (ARG_FILE_LINE, buf);
+#endif
+  }
+
+  void
   atomic_replication_helper::atomic_log_sequence::apply_and_unfix_sequence (THREAD_ENTRY *thread_p)
   {
     // Applying the log right after the fix could lead to problems as the records are fixed one by one as
     // they come to be read by the PTS and some might be unfixed and refixed after the apply procedure
-    // leading to inconsistency. To avoid this situation we sequentially apply each log redo of the sequence
-    // when the end sequence log appears and the entire sequence is fixed
+    // leading to inconsistency.
+    // To avoid this situation each log redo of the sequence is applied when the end sequence log appears
+    // and the entire sequence is already fixed.
+    // Right after applying, unfix and ref-count-down each page. The bookkeeping mechanism will take care
+    // of either unfixing the page or retaining it for a subsequent unfix.
+
+    if (prm_get_bool_value (PRM_ID_ER_LOG_DEBUG))
+      {
+	dump ();
+      }
+
     for (const auto &log_entry : m_log_vec)
       {
 	log_entry.apply_log_redo (thread_p, m_redo_context);
 	// bookkeeping actually will either unfix the page or just decrease its reference count
 	m_page_ptr_bookkeeping.unfix_page (thread_p, log_entry.m_vpid);
       }
+
+    // clear the vector of log records; page pts's might be, at this point, dangling pointers as the page ptr
+    // bookkeeping mechanims might have already unfixed the page
     m_log_vec.clear ();
   }
 
--- a/src/transaction/atomic_replication_helper.cpp
+++ b/src/transaction/atomic_replication_helper.cpp
@@ -268,6 +268,7 @@ namespace cublog
   atomic_replication_helper::atomic_log_sequence::add_atomic_replication_log (THREAD_ENTRY *thread_p,
       log_lsa record_lsa, LOG_RCVINDEX rcvindex, VPID vpid)
   {
+#if (0)
     m_log_vec.emplace_back (record_lsa, vpid, rcvindex);
     const auto page_map_it = m_page_map.find (vpid);
     if (page_map_it == m_page_map.cend ())
@@ -297,6 +298,23 @@ namespace cublog
 	m_log_vec.back ().set_page_ptr (page_map_it->second);
       }
     return NO_ERROR;
+#endif
+    PAGE_PTR page_p = nullptr;
+    // bookkeeping fixes page, keeps all info regarding how the page was fixed (either
+    // regular fix or ordered fix) and only returns back the pointer to the page
+    const int err_code = m_page_ptr_bookkeeping.fix_page (thread_p, vpid, rcvindex, page_p);
+    if (err_code != NO_ERROR)
+      {
+	assert (page_p == nullptr);
+	return err_code;
+      }
+    assert (page_p != nullptr);
+
+    m_log_vec.emplace_back (record_lsa, vpid, rcvindex, page_p);
+    //atomic_log_entry &last = m_log_vec.back ();
+    //last.set_page_ptr (page_p);
+
+    return err_code;
   }
 
   bool
@@ -363,6 +381,7 @@ namespace cublog
     return false;
   }
 
+#if (0)
   void
   atomic_replication_helper::atomic_log_sequence::apply_all_log_redos (THREAD_ENTRY *thread_p)
   {
@@ -371,6 +390,7 @@ namespace cublog
 	m_log_vec[i].apply_log_redo (thread_p, m_redo_context);
       }
   }
+#endif
 
   void
   atomic_replication_helper::atomic_log_sequence::apply_and_unfix_sequence (THREAD_ENTRY *thread_p)
@@ -379,6 +399,7 @@ namespace cublog
     // they come to be read by the PTS and some might be unfixed and refixed after the apply procedure
     // leading to inconsistency. To avoid this situation we sequentially apply each log redo of the sequence
     // when the end sequence log appears and the entire sequence is fixed
+#if (0)
     apply_all_log_redos (thread_p);
 
     for (size_t i = 0; i < m_log_vec.size (); i++)
@@ -389,6 +410,13 @@ namespace cublog
 	    m_log_vec[i].unfix_page (thread_p);
 	    m_page_map.erase (page_map_it);
 	  }
+      }
+#endif
+    for (size_t i = 0; i < m_log_vec.size (); i++)
+      {
+	m_log_vec[i].apply_log_redo (thread_p, m_redo_context);
+	// bookkeeping actually either
+	m_page_ptr_bookkeeping.unfix_page (thread_p, m_log_vec[i].m_vpid);
       }
   }
 
@@ -403,23 +431,29 @@ namespace cublog
    *********************************************************************************************************/
 
   atomic_replication_helper::atomic_log_sequence::atomic_log_entry::atomic_log_entry (
-	  log_lsa lsa, VPID vpid, LOG_RCVINDEX rcvindex)
+	  log_lsa lsa, VPID vpid, LOG_RCVINDEX rcvindex, PAGE_PTR page_ptr)
     : m_vpid { vpid }
     , m_record_lsa { lsa }
     , m_record_index { rcvindex }
-    , m_page_ptr { nullptr }
+    , m_page_ptr { page_ptr }
+#if (0)
     , m_watcher_p { nullptr }
+#endif
   {
     assert (lsa != NULL_LSA);
   }
 
   atomic_replication_helper::atomic_log_sequence::atomic_log_entry::atomic_log_entry (atomic_log_entry &&that)
-    : atomic_log_entry (that.m_record_lsa, that.m_vpid, that.m_record_index)
+    : atomic_log_entry (that.m_record_lsa, that.m_vpid, that.m_record_index, that.m_page_ptr)
   {
+#if (0)
     std::swap (m_page_ptr, that.m_page_ptr);
     std::swap (m_watcher_p, that.m_watcher_p);
+#endif
+    that.m_page_ptr = nullptr;
   }
 
+#if (0)
   atomic_replication_helper::atomic_log_sequence::atomic_log_entry::~atomic_log_entry ()
   {
     if (m_watcher_p != nullptr)
@@ -428,6 +462,7 @@ namespace cublog
 	m_watcher_p.reset ();
       }
   }
+#endif
 
   void
   atomic_replication_helper::atomic_log_sequence::atomic_log_entry::apply_log_redo (
@@ -471,6 +506,7 @@ namespace cublog
       }
   }
 
+#if (0)
   int
   atomic_replication_helper::atomic_log_sequence::atomic_log_entry::fix_page (THREAD_ENTRY *thread_p)
   {
@@ -516,6 +552,7 @@ namespace cublog
       }
 
     return NO_ERROR;
+    //return pgbuf_fix_or_ordered_fix (thread_p, m_vpid, m_record_index, m_watcher_p, m_page_ptr);
   }
 
   void
@@ -541,6 +578,7 @@ namespace cublog
 	pgbuf_unfix (thread_p, m_page_ptr);
 	break;
       }
+    //pgbuf_unfix_or_ordered_unfix (thread_p, m_record_index, m_watcher_p, m_page_ptr);
   }
 
   PAGE_PTR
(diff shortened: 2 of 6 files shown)


src/base/system_parameter.c
```diff
@@ -6346,6 +6351,19 @@ static SYSPRM_PARAM prm_Def[] = {
    (char *) NULL,
    (DUP_PRM_FUNC) NULL,
    (DUP_PRM_FUNC) NULL},
+  {PRM_ID_ER_LOG_PTS_ATOMIC_REPL_DEBUG,
```
razvanradulescu, line 6354: PRM_ID_ER_LOG_PTS_ATOMIC_REPL_DEBUG does not seem to be used outside the debug version.
Maybe is not necessary to have a system parameter, but a toggling macro.
System parameters are supposed to be controlled by user, but in this case, the function is limited to debug info.

src/transaction/atomic_replication_helper.cpp
```diff
@@ -202,73 +128,218 @@ namespace cublog
...
+	    _er_log_debug (ARG_FILE_LINE,
+			   "[ATOMIC_REPL] append_control_log_sysop_end _not_ purged trid = %d\n",
+			   trid);
+#endif
 	  }
       }
-    return min_lsa;
   }
 
+  void atomic_replication_helper::forcibly_remove_idle_sequence (TRANID trid)
```
razvanradulescu: Nit : naming  forcibly_remove_idle_sequence-> forcibly_remove_tran_sequence
Current name suggests it contains some sort of check if sequence is idle (but this is decided exclusively by the caller).

src/transaction/atomic_replication_helper.cpp
```diff
@@ -202,73 +128,218 @@ namespace cublog
...
+
+    // sequence dtor will ensure proper idle state upon destruction
+    m_sequences_map.erase (sequence_it);
+  }
+
+#if !defined (NDEBUG)
+  void
+  atomic_replication_helper::dump (const char *message) const
+  {
+    constexpr int BUF_LEN_MAX = SHRT_MAX;
```
razvanradulescu, line 273: Did you consider using heap memory buffer ?
"string_buffer" manages memory allocation and has variadic append operator.

src/base/system_parameter.c
cristiarg: I want to be able to activate de-activate this debug logging without changes to the binary.
Is there a way to do this without a system param?
The param should be removed as soon as we are confident that atomic replication is realiable. There is a similar param `PRM_ID_ER_LOG_PTS_REPL_DEBUG` only used for MVCC replication (although the name is misleading). I will created a patch to change that name as well.
Actually, maybe my mistake is that I guarded the usage with `NDEBUG`.
I will remove the NDEBUG guards then because the logging might prove useful in release runs as well.

src/transaction/atomic_replication_helper.cpp
cristiarg: Yes, but I thought `snprintf` to stack buffer is faster. Uglier, but faster.
Albeit, it is all arbitrary as this is debug code.

src/transaction/atomic_replication_helper.cpp
cristiarg: Done.

src/transaction/atomic_replication_helper.hpp
```diff
@@ -30,16 +30,32 @@
...
    *
-   * All examples given below are treated as a single atomic sequence.
+   * For all scenarios details below, the following rules apply:
+   * - from the helper's point of view, there are two types of log records:
+   *  - control log records - see 'is_control'
+   *  - proper log records
+   * - control log records are also added to the atomic sequence as atomic log entries
+   *    and they help in deciding for more advanced scenarios (and provide for future extensions)
+   * - in-between control records are the proper log records
+   * - a new new control record is added to the sequence, all proper log records are applied and
```
hornetmj:
```suggestion
   * - a new control record is added to the sequence, all proper log records are applied and
```

src/transaction/log_replication_atomic.cpp
```diff
@@ -116,7 +116,25 @@ namespace cublog
...
+	    //
+	    // for *2*, the issue http://jira.cubrid.org/browse/LETS-572 has been added;
+	    //
+	    // for now, a naive aproach of asserting and forcibly eliminating the atomic sequence
+	    assert (!m_atomic_helper.is_part_of_atomic_replication (header.trid)
+		    || m_atomic_helper.all_log_entries_are_control (header.trid));
+	    if (m_atomic_helper.is_part_of_atomic_replication (header.trid)
+		&& m_atomic_helper.all_log_entries_are_control (header.trid))
+	      {
+		m_atomic_helper.forcibly_remove_sequence (header.trid);
```
wrlawodms, line 136: I think there could be other cases in which an atomic sequence doesn't complete properly without LOG_ABORT, so the sequence has to be cleaned up. As far as I know, there are two cases in which a transaction can be rolled back to some point and continues. It is (1) SAVEPOINT and (2) sysop abort (end with LOG_SYSOP_END_ABORT).
(1) With SAVEPOINT, a user or a developer put a save point and can roll back to that point and continue the transaction. See xtran_server_savepoint().
(2) After a sysop starts, the sysop can abort anytime for some reason. Then, the transaction is rolled back to when the sysop starts and appends LOG_SYSOP_END_ABORT regardless of its intended ending. In this case, the transaction usually are going to be aborted, but it's not guaranteed. The transaction's destiny is just decided by the operation's error handling.

I don't think the SAVEPOINT can cut in an atomic system operation since sysop is used for an internal operation and SAVEPOINT is used by a user (developer or end-user). However, the second can be the case. Even if it always ends up aborting when atomic sysops are aborted for now, I think we should deal with it because there it is allowed architecturally in my opinion.

src/transaction/atomic_replication_helper.cpp
```diff
@@ -301,162 +354,410 @@ namespace cublog
...
-    if (m_end_pospone_count > 0)
+    return std::all_of (m_log_vec.cbegin (), m_log_vec.cend (),
+			[] (const atomic_log_entry &entry)
+    {
+      return entry.is_control ();
+    });
+  }
+
+  bool
+  atomic_replication_helper::atomic_log_sequence::can_purge ()
```
wrlawodms, line 463: Could you summarize this function's purpose and behavior? You don't have to make it so detailed. I just don't want to misunderstand the context.

src/transaction/atomic_replication_helper.cpp
```diff
@@ -202,46 +122,177 @@ namespace cublog
...
+	sequence_it = m_sequences_map.find (trid);
       }
 
+    // TODO: idea, first add control log and then apply and unfix, such that apply and unfix will
+    // be able to make the decisions taking into consideration the last control log;
+    // this can be implemented later and only if needed; works as is right now
     atomic_log_sequence &sequence = sequence_it->second;
-    sequence.apply_and_unfix_sequence (thread_p);
-    m_sequences_map.erase (sequence_it);
+    sequence.apply_and_unfix (thread_p);
```
wrlawodms, line 178: Does this mean kinds of the following sequence are not allowed?
< start atomic sysop 1 > - < redo > - < start atomic sysop2 > - < redo > - <sysop end 1> - **< redo >** - <  sysop end 2 >.
In the above, the atomicity of sysop 1 is broken. The way I see it, there is no restriction to force the way system operations are used even if there is no case for now.

src/transaction/atomic_replication_helper.cpp
```diff
@@ -301,162 +354,410 @@ namespace cublog
...
+    std::swap (m_record_lsa, that.m_record_lsa);
+    std::swap (m_record_index, that.m_record_index);
+    std::swap (m_sysop_end_type, that.m_sysop_end_type);
+    std::swap (m_sysop_end_last_parent_lsa, that.m_sysop_end_last_parent_lsa);
+    std::swap (m_page_ptr, that.m_page_ptr);
+    return *this;
+  }
+
   void
   atomic_replication_helper::atomic_log_sequence::atomic_log_entry::apply_log_redo (
```
wrlawodms, line 824: Could you tell me why only a few record types are handled here? Isn't it true that any record types can be put in an atomic system operation such as LOG_MVCC_DIFF_UNDOREDO_DATA?

wrlawodms (commented):
Sorry for the late review.

src/transaction/atomic_replication_helper.cpp
```diff
@@ -301,162 +354,410 @@ namespace cublog
...
+			[] (const atomic_log_entry &entry)
+    {
+      return entry.is_control ();
+    });
+  }
+
+  bool
+  atomic_replication_helper::atomic_log_sequence::can_purge ()
+  {
+    assert (all_log_entries_are_control ());
```
joohok, line 465: Is this assertion satisfied because it is only called after apply_and_unfix is ​​executed? 

LOG_ATOMIC_SYSOP (1) -> non control log -> LOG_ATOMIC_SYSOP without log  record (2) -> non control log ->  LOG_SYSOP_END (3) -> LOG_SYSOP_END (4) 

When the control log record of `(3) LOG_SYSOP_END` is replicated,
(3) LOG_SYSOP_END is a log record corresponding to LOG_ATOMIC_SYSOP (2). 
In this case, the log records between (1) and (3) will be applied even if the (1) is not ended with commit.

src/transaction/log_replication_atomic.cpp
```diff
@@ -132,27 +150,17 @@ namespace cublog
...
-	      {
-		// atomic sequence is part of a postpone sequence
-		// which is itself part of another, previosly started, atomic sequence
-	      }
-	    else
-	      {
-		m_atomic_helper.start_sysop_sequence (header.trid, m_redo_lsa, m_redo_context);
-		set_lowest_unapplied_lsa ();
-	      }
+	    m_atomic_helper.append_control_log (&thread_entry, header.trid, header.type, m_redo_lsa, m_redo_context);
```
joohok, line 162: I think it needs assertion like LOG_START_ATOMIC_REPL. 
`assert (!m_atomic_helper.is_part_of_atomic_replication (header.trid));` 

LOG_SYSOP_ATOMIC_START is only logged once within a transaction before it is LOG_SYSOP_END with commit.

`LOG_SYSOP_ATOMIC_START - LOG_SYSOP_ATOMIC_START - LOG_SYSOP_END - LOG_SYSOP_END`
above scenario is not allowed, only the first LOG_SYSOP_ATOMIC_START can be logged. 

I'm still not clear on the relationship between LOG_SYSOP_ATOMIC_START and LOG_START_ATOMIC_REPL . Is there a case where two are nested? 
such as
`LOG_SYSOP_ATOMIC_START -> LOG_START_ATOMIC_REPL` or
`LOG_START_ATOMIC_REPL -> LOG_SYSOP_ATOMIC_START` ?

src/transaction/atomic_replication_helper.cpp
cristiarg: Added an extended comment with an example. Let me know if it does the job.

src/transaction/atomic_replication_helper.cpp
cristiarg: They could be overlapped as you describe above (even though I think this will never be the case - it will crash long before replication).
And these overlapping and nonsensical sequences would still be applied. But I think the `can_purge` function would be unable to clean the control log records and therefore crash too.

This is somewhat on purpose because in practice the following "nested" situation is encountered (notice: not the same example as the one above):
```
- < start atomic sysop 1 >  --+
  - ..                        |
  - < redo VPID1 >           (1)
  - ..                        |
- < start atomic sysop2 >   --+
  - ..                        |
  - < redo VPID1 >           (2)
  - ..                        |
- < sysop end 2 >  -----------+
  - ..                        |
  - < redo >                 (3)
  - ..                        |
- < sysop end 1 >  -----------+
```
If the logic were to follow "topologically" and first apply the redos of the inner atomic sequence (2) and then the redo's of the outer atomic sequence - (1) and (3) - the page VPID1 would be inconsistently modified (in debug - most probably crash; in release - could lead to incorrect replication and wrong results).

src/transaction/atomic_replication_helper.cpp
cristiarg: These are the ones that have been encountered in tests. Notice the
```
      default:
	assert (false);
	break;
```
at the end.

src/transaction/atomic_replication_helper.cpp
cristiarg: Yes, this is the intention.
See previous comments and also comments in function `can_purge`.
The direct reason is explained in https://github.com/CUBRID/cubrid/pull/3910#discussion_r1012332005 - same page can be modified by multiple levels of a nested atomic sequence.

Although, on a second thought, there might be a better way to handle this. Given that we can take advantage of the `page_ptr_bookkeeping` functionality which reference-counts fixes/unfixes, thus allowing the both application of nested sequences in same pages and also consistent unfix, and - by consequence - consistent visibility for client transactions.

I will add another issue to re-re-refactor this. :) It is maybe lower prio.

src/transaction/log_replication_atomic.cpp
cristiarg: > I think it needs assertion like LOG_START_ATOMIC_REPL.
assert (!m_atomic_helper.is_part_of_atomic_replication (header.trid));

> LOG_SYSOP_ATOMIC_START is only logged once within a transaction before it is LOG_SYSOP_END with commit.

See section 4. in the comment before `class atomic_replication_helper`.

> LOG_SYSOP_ATOMIC_START - LOG_SYSOP_ATOMIC_START - LOG_SYSOP_END - LOG_SYSOP_END
above scenario is not allowed, only the first LOG_SYSOP_ATOMIC_START can be logged.

I know. This is the ugliest case encountered yet.
It is handled in the `can_purge` function. See:
```
	else if (LOG_END_ATOMIC_REPL == last_entry.m_rectype)
	  {
	    // search backwards, until a start atomic replication log record is met; other 'sysop
	    // ends' are allowed and skipped;
	    // these are the sysops without a matching sysop atomic start - it is the consequence of the
	    // known fact (optimization?): when adding consecutive 'sysop atomic start' log records only
	    // one such log record is added
```

> I'm still not clear on the relationship between LOG_SYSOP_ATOMIC_START and LOG_START_ATOMIC_REPL . Is there a case where two are nested?
> such as
> LOG_SYSOP_ATOMIC_START -> LOG_START_ATOMIC_REPL or
> LOG_START_ATOMIC_REPL -> LOG_SYSOP_ATOMIC_START ?

Yes, there are at least 2 that I encountered with tests:
- one is documented in section 4 of the comment before `class atomic_replication_helper`
- the second one is more recent and has been added with extra atomic sequences added for `RVHF_SET_PREV_VERSION_LSA`

src/transaction/log_replication_atomic.cpp
cristiarg: I took the liberty of pinging back from ~~http://jira.cubrid.org/browse/LETS-572~~ http://jira.cubrid.org/browse/LETS-541 to this comment.
This will be taken up when that is handled.
Or do you think is more urgent and should be handled asap?

src/transaction/atomic_replication_helper.cpp
wrlawodms: I think this is making a restriction on atomic system operation. The way I see it, system operations are designed to be able to contain any kind of operation. I'm quite against that we can ignore their usable cases just because they haven't been used in that way. 
 Do you have a plan to extend this to cover others later?

src/transaction/atomic_replication_helper.cpp
wrlawodms: I know that all REDO history has to be repeated in order. I mean if VPID1 is unfixed after redoing (1), other threads can cut in and fix the page before fixing the page for (2) and (3 - if it also modifies VPID1). It seems to break the atomicity as long as I don't misunderstand.

src/transaction/log_replication_atomic.cpp
wrlawodms: It's not about "urgent", but I meant these cases can happen and they should be handled, even if they don't happen for now. If they will be handled before release and are tracked correctly, it's OK.

src/transaction/atomic_replication_helper.cpp
wrlawodms: > They could be overlapped as you describe above (even though I think this will never be the case - it will crash long before replication).

I don't understand why this sequence is impossible, and it means I may miss something..

Oh, I'm sorry. I mixed up the end log records by mistake. I wanted to say is:
< start atomic sysop 1 > - < redo > - < start atomic sysop2 > - < redo > - <sysop end **_2_**> - < redo > - < sysop end **_1_** >.
And my concern is above with this sequence, which I think is the same as your nested situation example.

src/transaction/atomic_replication_helper.cpp
cristiarg: Of course these can be extended later. It will be the job off different test scenarios to uncover uncovered cases.
I don't see any restriction.

However, and I now see this, The `LOG_RECTYPE`'s that are handled in `atomic_log_entry::apply_log_redo` are the same that are handled via `atomic_replicator::redo_upto` -> `atomic_replicator::read_and_redo_record` -> `atomic_replication_helper::append_log` -> `atomic_log_sequence::append_log`.

And these switches have been taken over from `log_recovery_redo` which was our "single source of truth" when we began replication.

> I'm quite against that we can ignore their usable cases just because they haven't been used in that way.
I don't think I understand.

> Do you have a plan to extend this to cover others later?
Whenever needed.

src/transaction/atomic_replication_helper.cpp
cristiarg: > I know that all REDO history has to be repeated in order. I mean if VPID1 is unfixed after redoing (1), other threads can cut in and fix the page before fixing the page for (2) and (3 - if it also modifies VPID1). It seems to break the atomicity as long as I don't misunderstand.

Yes, this is a possibility.
However:
- in the first version of the atomic replication, the pages were kept fixed by the replication from `< start atomic sysop 1 >` until `< sysop end 1 >`;
- this was deemed too costly for performance; and - based on examples seen then - there was no counter argument as to not apply log records and unfix pages at every occurrence of a "control log record"; hence the current refactoring

I actually do not know which approach is best. As there is no way to know whether future log records within one atomic sequence within one transaction will change same pages.
And a page has to be fixed right from the start when the log record i processed. It is a no-option to, say, first collect all log records from an atomic sequence and then decide how to apply them.

src/transaction/log_replication_atomic.cpp
cristiarg: Agreed, I overlooked, but there was already an issue - http://jira.cubrid.org/browse/LETS-541 - within the same epic that says exactly this. That some of the `SYSOP_END`'s are not handled.

src/transaction/atomic_replication_helper.cpp
wrlawodms: I understand it is too costly, but it is one of the atomic replication's inherent attributes, the way I see it. If an atomic operation gets bigger, it will fix many pages and cost a lot. I don't get the point of why this kind of nested scenario is an exceptional case for which some records are applied and its pages are unfixed. Are they longer than others usually?

src/transaction/log_replication_atomic.cpp
wrlawodms: Oh, I see.

src/transaction/atomic_replication_helper.cpp
wrlawodms: > And these switches have been taken over from log_recovery_redo which was our "single source of truth" when we began replication.

Oh,, then I'm happy with it. I thought it does take over only some parts of that. An example I mentioned, LOG_MVCC_DIFF_UNDOREDO_DATA, is handled, but I missed it. Sorry to bother you with it.

wrlawodms: They only case omitted seems to be LOG_DBEXTERN_REDO_DATA. Is it on purpose?

cristiarg: `LOG_DBEXTERN_REDO_DATA` is handled directly in `atomic_replicator::redo_upto`. It never gets into an atomic sequence. Because it does not change a certain page.
`LOG_DBEXTERN_REDO_DATA` seems to have only 3 rcvindices (as investigated via `log_append_dboutside_redo` -> `prior_lsa_alloc_and_copy_data` -> `prior_lsa_gen_dbout_redo_record`):
- `RVDK_NEWVOL`, `RVDK_EXPAND_VOLUME` - both skipped for atomic replication
- `RVLOG_OUTSIDE_LOGICAL_REDO_NOOP`  whose redo function is no-op

src/transaction/atomic_replication_helper.cpp
cristiarg: > Are they longer than others usually?

I think they are. However I also think that we haven't yet seen the "worst of it".
For example there haven't been any tests with very large records in the schema (table records spanning multiple pages).

I see action possible in the following 3 directions:

1)
However, it is relatively simple to overload the logic to either do it as it is now (piece-wise between "control" log records) or all in one bit "proper" atomic sequence (as it was initially implemented).
For this, added http://jira.cubrid.org/browse/LETS-585 .

2)
However, we're facing with a "chicken & egg" problem:
- currently the implementation for PTS replication is incomplete
- given that, not many tests can be added since they fail

3)
Given this, I propose to
- merge this with this "piece-wise" replication method
- wait for http://jira.cubrid.org/browse/LETS-524 to be implemented - because that will lessen the number of atomic sequence scenarios to be supported
- in the meantime I will add new logging code such that the complete atomic sequence is dumped (text representation of all "control" and "non-control" log records) at the end, before the atomic sequence is destroyed. For this, added http://jira.cubrid.org/browse/LETS-586 .

With this we'll be hopefully able to advance some testing scenarios that we'll allow further analysis.
We might even discover crashes caused precisely by this "piece-wise" atomic replication technique.

src/transaction/atomic_replication_helper.cpp
```diff
@@ -59,120 +52,47 @@ namespace cublog
...
-
-	// safeguard, if the atomic sequence contains a postpone [sub]sequence, that
-	// is more specific case and should have been checked upfront
-	assert (!atomic_sequence.is_postpone_sequence_started ());
-	assert (!atomic_sequence.is_at_least_one_postpone_sequence_completed ());
-
-	return atomic_sequence.can_end_sysop_sequence ();
+	if (sequence_it != m_sequences_map.cend ())
+	  {
+	    const atomic_log_sequence &atomic_sequence = sequence_it->second;
```
hornetmj:
```suggestion
```

src/transaction/atomic_replication_helper.cpp
```diff
@@ -301,160 +354,465 @@ namespace cublog
...
+    // the result of empirical observations of encountered "control" log sequences;
+    // these sequences are laid out in the atomic_replication_helper class comment and
+    // their logic is implemented and acted upon in this function and the function
+    // calling this one
+    while (true)
       {
-	assert (m_is_sysop);
-	assert (m_postpone_started);
-	return true;
+	if (m_log_vec.empty ())
```
hornetmj, line 527: It looks safe-guard. The can_purge() is always called after calling the append_control_log().

src/transaction/atomic_replication_helper.cpp
wrlawodms: OK. Let's move forward with the recognition that this is another problem that needs to be solved in the long term. We will solve it and others together. Thanks.

src/transaction/atomic_replication_helper.cpp
```diff
@@ -202,46 +122,177 @@ namespace cublog
...
+  void atomic_replication_helper::forcibly_remove_sequence (TRANID trid)
+  {
+    auto sequence_it = m_sequences_map.find (trid);
+    if (sequence_it == m_sequences_map.end ())
+      {
+	assert (false);
+	return;
+      }
+
+    atomic_log_sequence &sequence = sequence_it->second;
```
wrlawodms: unused

src/transaction/atomic_replication_helper.cpp
```diff
@@ -301,160 +354,465 @@ namespace cublog
...
+    atomic_log_entry_vector_type::const_iterator first_work_log_it = m_log_vec.end ();
+    --first_work_log_it; // last in vector, must be work
+    assert (!first_work_log_it->is_control ());
+    while (!first_work_log_it->is_control ()) // skip all work entries
+      {
+	--first_work_log_it;
+      }
+    assert (first_work_log_it->is_control ());
+    ++first_work_log_it; // up one to the first actual work
+    assert (!first_work_log_it->is_control ());
```
hornetmj, line 394: These assert()s wrapped in no necessary. I think the while loop is already cover them.

src/transaction/atomic_replication_helper.cpp
```diff
@@ -301,160 +354,465 @@ namespace cublog
...
+    if (rbegin_work_log.is_control ())
+      {
+	assert (all_log_entries_are_control ());
+	return;
+      }
+
+    // search backwards for the first non-control log record entry
+    atomic_log_entry_vector_type::const_iterator first_work_log_it = m_log_vec.end ();
+    --first_work_log_it; // last in vector, must be work
+    assert (!first_work_log_it->is_control ());
```
wrlawodms, line 387: Why don't you check here instead of the above check with rbegin_work_log_it (378L-384L)?
```suggestion
        if (first_work_log_it.is_control ())
      {
	assert (all_log_entries_are_control ());
	return;
      }
```

src/transaction/atomic_replication_helper.cpp
```diff
@@ -301,160 +354,465 @@ namespace cublog
...
+	    dump ("sequence::can_purge - START");
+	  }
+
+	const atomic_log_entry_vector_type::const_iterator last_entry_it = m_log_vec.cend () - 1;
+	const atomic_log_entry &last_entry = *last_entry_it;
+
+	// atomic replication sequence with an already executed postpone sequence that (maybe) contained
+	// - itself - other atomic replication sequences)
+	if (LOG_SYSOP_END == last_entry.m_rectype
+	    && LOG_SYSOP_END_COMMIT == last_entry.m_sysop_end_type)
```
hornetmj, line 549:
```suggestion
	if (last_entry.m_rectype == LOG_SYSOP_END
	    && last_entry.m_sysop_end_type == LOG_SYSOP_END_COMMIT)
```

src/transaction/atomic_replication_helper.cpp
```diff
@@ -301,160 +354,465 @@ namespace cublog
...
+    //  [almost] arbitrary structure to be processed in a controlled way:
+    //
+    // Example:
+    //
+    //  ------------------------------------ transactional log axis ----->
+    //  C1  W1 W2 W3  C2  W4 W5 W6  C3  W7 W8 W9 W10 C4  W11 W12  C5  C6
+    //   |             |             |                |            |   |
+    //  start        start           |              start          |   |
+    //                              end                          end   end
+    //
```
hornetmj, line 485: If my understanding is correct, this modification changes the mechanism that was originally proposed.
1. fix all page
2. apply all and unfix all page.

Found issues are resolved by this fix. (I think)
1. vacuum - postpone dealloc
2. update - alloc new heap page for newhome

No problematic cases were found, but the reflection of W1 W2 W3 and W7 W8 W9 W10 is no longer atomic.

src/transaction/atomic_replication_helper.cpp
```diff
@@ -301,160 +354,465 @@ namespace cublog
...
     assert (m_page_ptr != nullptr);
   }
 
+  atomic_replication_helper::atomic_log_sequence::atomic_log_entry::atomic_log_entry (
+	  LOG_LSA lsa, LOG_RECTYPE rectype)
+    : m_vpid VPID_INITIALIZER
+    , m_rectype { rectype }
+    , m_record_lsa { lsa }
+    , m_record_index { RV_NOT_DEFINED }
+    , m_sysop_end_type { (LOG_SYSOP_END_TYPE)-1 }
```
wrlawodms, line 774: Out of curiosity, (ENUM TYPE)-1 is commonly used for its default value? I even didn't know it succeed to be compiled.

src/transaction/atomic_replication_helper.cpp
```diff
@@ -301,160 +354,465 @@ namespace cublog
...
     , m_record_lsa { lsa }
     , m_record_index { rcvindex }
+    , m_sysop_end_type { (LOG_SYSOP_END_TYPE)-1 }
+    , m_sysop_end_last_parent_lsa { NULL_LSA }
     , m_page_ptr { page_ptr }
   {
     assert (!VPID_ISNULL (&m_vpid));
     assert (m_record_lsa != NULL_LSA);
-    assert (0 <= m_record_index && m_record_index <= RV_LAST_LOGID);
+    assert (0 <= m_record_index &&m_record_index <= RV_LAST_LOGID);
```
wrlawodms:
```suggestion
    assert (0 <= m_record_index && m_record_index <= RV_LAST_LOGID);
```

src/transaction/atomic_replication_helper.cpp
```diff
@@ -301,160 +354,465 @@ namespace cublog
...
    * atomic_replication_helper::atomic_log_sequence::atomic_log_entry function definitions  *
    *********************************************************************************************************/
 
   atomic_replication_helper::atomic_log_sequence::atomic_log_entry::atomic_log_entry (
-	  log_lsa lsa, VPID vpid, LOG_RCVINDEX rcvindex, PAGE_PTR page_ptr)
+	  LOG_LSA lsa, VPID vpid, LOG_RCVINDEX rcvindex, PAGE_PTR page_ptr)
     : m_vpid { vpid }
+    , m_rectype { LOG_LARGER_LOGREC_TYPE }
     , m_record_lsa { lsa }
     , m_record_index { rcvindex }
```
wrlawodms: NIT: I think m_record_rcvidx or m_rcvindex seems better because just 'index' looks vague to me.

src/transaction/atomic_replication_helper.cpp
```diff
@@ -301,160 +354,465 @@ namespace cublog
...
+    : m_vpid VPID_INITIALIZER
+    , m_rectype { LOG_SYSOP_END }
+    , m_record_lsa { lsa }
+    , m_record_index { RV_NOT_DEFINED }
+    , m_sysop_end_type { sysop_end_type }
+    , m_sysop_end_last_parent_lsa { sysop_end_last_parent_lsa }
+    , m_page_ptr { nullptr }
+  {
+    assert (m_record_lsa != NULL_LSA);
+    assert (LOG_SYSOP_END_COMMIT <= sysop_end_type &&sysop_end_type <= LOG_SYSOP_END_LOGICAL_RUN_POSTPONE);
```
wrlawodms:
```suggestion
    assert (LOG_SYSOP_END_COMMIT <= sysop_end_type && sysop_end_type <= LOG_SYSOP_END_LOGICAL_RUN_POSTPONE);
```

src/transaction/atomic_replication_helper.cpp
```diff
@@ -301,160 +354,465 @@ namespace cublog
...
-    : atomic_log_entry (that.m_record_lsa, that.m_vpid, that.m_record_index, that.m_page_ptr)
+    : m_vpid { that.m_vpid }
+    , m_rectype { that.m_rectype }
+    , m_record_lsa { that.m_record_lsa }
+    , m_record_index { that.m_record_index }
+    , m_sysop_end_type { that.m_sysop_end_type }
+    , m_sysop_end_last_parent_lsa { that.m_sysop_end_last_parent_lsa }
+    , m_page_ptr { that.m_page_ptr }
+  {
+  }
```
wrlawodms: Why doesn't the move constr use `swap` to empty its operand while the move operator does?

src/transaction/atomic_replication_helper.cpp
```diff
@@ -499,6 +857,45 @@ namespace cublog
...
   }
 
+  bool
+  atomic_replication_helper::atomic_log_sequence::atomic_log_entry::is_control () const
+  {
+    return (m_rectype == LOG_START_ATOMIC_REPL ||
+	    m_rectype == LOG_END_ATOMIC_REPL ||
+	    m_rectype == LOG_SYSOP_ATOMIC_START ||
+	    m_rectype == LOG_SYSOP_END ||
+	    m_rectype == LOG_SYSOP_START_POSTPONE);
```
wrlawodms: Could you tell me why `LOG_SYSOP_START_POSTPONE` has to be regarded as a control log? There must be a reason, but I feel like it just complicates the design.

wrlawodms (commented):
There could be some kind of silly questions, but please cut me some slack to make our future collaboration easy :).

src/transaction/atomic_replication_helper.cpp
cristiarg: > that needs to be solved in the long term

I would say "short term". :)

src/transaction/atomic_replication_helper.cpp
cristiarg: Added http://jira.cubrid.org/browse/LETS-585 to address both this issue and the issue of "true atomic replication for the entire atomic sequence and sub-sequences".

src/transaction/atomic_replication_helper.cpp
cristiarg: Done

src/transaction/atomic_replication_helper.cpp
cristiarg: Deleted

src/transaction/atomic_replication_helper.cpp
cristiarg: Changed as suggested.

src/transaction/atomic_replication_helper.cpp
cristiarg: > but the reflection of W1 W2 W3 and W7 W8 W9 W10 is no longer atomic

Correct. We will continue to investigate and evaluate what the best approach is. As said, I think we'll be able to better judge with the implementation of http://jira.cubrid.org/browse/LETS-524 (and the addition of http://jira.cubrid.org/browse/LETS-586 logging and extensive testing).

For instance, being able to investigate all cases exposed in http://jira.cubrid.org/browse/LETS-511 will probably reveal some more scenarios which are not even covered by this implementation.

src/transaction/atomic_replication_helper.cpp
cristiarg: Agree. But it can't hurt. I'll leave as it is.

src/transaction/atomic_replication_helper.cpp
cristiarg: It is the only `true` stopping condition. All others result in inconsistent states and return `false`. There is also an empirical idle check at the end of the `while` loop that also asserts.
And the loop is needed in order to be able to purge multiple nested sequences.

src/transaction/atomic_replication_helper.cpp
cristiarg: Typical "C/C++-ism" to prevent mistakes of the type
```
if (last_entry.m_rectype = LOG_SYSOP_END) ...
```
To prevent the compiler's automatic conversion/promotion to boolean true.
I'll leave as it is even though it is harder to read.

src/transaction/atomic_replication_helper.cpp
cristiarg: Poor attempt of making a default value. Since `LOG_SYSOP_END_TYPE` doesn't have a guard value. I didn't even check what it converts to. Internally, all enums are actually compiled to `static const` numeric values. Maybe more recently to `static constexpr`'s that's why in `RelWithDebugInfo` you're not able to see any enum value - 'cause they're all inlined.

src/transaction/atomic_replication_helper.cpp
cristiarg: K

src/transaction/atomic_replication_helper.cpp
cristiarg: k

src/transaction/atomic_replication_helper.cpp
cristiarg: Fixed. Because at some point during the implementation all members of the structure were const.

src/transaction/atomic_replication_helper.hpp
cristiarg: k

src/transaction/atomic_replication_helper.cpp
cristiarg: Done

src/transaction/atomic_replication_helper.cpp
cristiarg: See section (3) of the comment for the `class atomic_replication_helper` - with the layout encountered during a vacuum transaction.

cristiarg:
Addressed most of the comments in one way or the other.

hornetmj (approved):

src/transaction/atomic_replication_helper.cpp
wrlawodms: I see that case (3) happens, but I don't still understand why this case makes LOG_SYSOP_START_POSTPONE one of the control logs.

wrlawodms (approved):

src/transaction/atomic_replication_helper.cpp
cristiarg: To be re-investigated during - http://jira.cubrid.org/browse/LETS-541 - see comment - http://jira.cubrid.org/browse/LETS-541?focusedCommentId=4762726&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-4762726 .

PR #3916, opened 2022-10-27
[CBRD-24530] The replication often fails when replicating large size records in an HA environment

hornetmj:
http://jira.cubrid.org/browse/CBRD-24530

It would be better to read the jira description.

### DIFF
--- a/src/transaction/log_applier.c
+++ b/src/transaction/log_applier.c
@@ -477,7 +477,7 @@ static int la_does_page_exist (LOG_PAGEID pageid);
 static int la_init_repl_lists (bool need_realloc);
 static bool la_is_repl_lists_empty ();
 static LA_APPLY *la_find_apply_list (int tranid);
-static void la_log_copy_fromlog (char *rec_type, char *area, int length, LOG_PAGEID log_pageid, PGLENGTH log_offset,
+static void la_log_copy_fromlog (char *rec_type, char *area, int &length, LOG_PAGEID log_pageid, PGLENGTH log_offset,
 				 LOG_PAGE * log_pgptr);
 static LA_ITEM *la_new_repl_item (LOG_LSA * lsa, LOG_LSA * target_lsa);
 static void la_add_repl_item (LA_APPLY * apply, LA_ITEM * item);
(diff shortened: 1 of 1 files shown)


src/transaction/log_applier.c
```diff
@@ -2962,7 +2962,7 @@ la_add_apply_list (int tranid)
  *   log_pageid, log_offset, and log_pgptr are set as a side effect.
  */
 static void
-la_log_copy_fromlog (char *rec_type, char *area, int length, LOG_PAGEID log_pageid, PGLENGTH log_offset,
+la_log_copy_fromlog (char *rec_type, char *area, int &length, LOG_PAGEID log_pageid, PGLENGTH log_offset,
```
wrlawodms: How does this solve the problem? Could you elaborate on it more?

2993L: `length = length - DB_SIZEOF (INT16);` solve the problem? and is it always done? 
What happens if `rec_type` is NULL?

joohok: I'm not sure if it's better to change `int length` to `int &length`.
Wouldn't it be better to specify it as an (in/out) parameter at function comment and receive it as a pointer type ('int *length) rather than adding a reference type to the c style code?

hornetmj: Whether rec_type is null or not is a kind of condition. If rec_type is null, the size of length passed is read from the log page. If rec_type is not null, rec_type with a length of 2 bytes is read, and then length - 2 is read. In both cases, the read length is the same, but the behavior differs.

The la_log_copy_fromlog() function seems to serve several purposes. It is used to read the format like | rec_type | recdes->data | (redo image) in the case of this issue, but it can also be used to read record headers, such as LOG_REC_REPLICATION.

hornetmj: I will. I had the same thought, but I didn't do it because I had to modify all the callers. I wanted to minimize modifications.

wrlawodms: I guess the modified length would be used to determine its type in salve node from your code and explanation. Correct? I don't really know, but I'm worried some of callers of `la_log_copy_fromlog()` just hand over a local variable for the parameter `length`. Don't these cases matter? Which caller is the case causes the issue? If you have already checked, it's OK.

hornetmj: The following is the callstack you want.

0 la_log_copy_fromlog()
1 la_get_log_data()
2 la_get_recdes()
3 la_apply_insert_log()
...

I think you can ignore la_log_copy_fromlog() callers which pass NULL value of rec_type.

wrlawodms (approved):

src/transaction/log_applier.c
hornetmj: I applied your feedback. Re-review plz.

joohok (approved):

PR #3919, opened 2022-10-28
[CBRD-24537] Improve the performance of MERGE query by using NOT EXISTS sub-query

beyondykk9:
http://jira.cubrid.org/browse/CBRD-24537

The MERGE query's "NOT MATCHED" part is rewritten by using COUNT query for checking "NOT MATCHED", however, the performance of COUNT query is not good compared with NOT EXISTS query.

We will enhance the performance of the MERGE query by using "NOT EXISTS" sub-query instead of COUNT query.


src/parser/xasl_generation.c
```diff
@@ -25217,6 +25204,36 @@ pt_to_merge_insert_query (PARSER_CONTEXT * parser, PT_NODE * select_list, PT_MER
...
+
+  if (subq)
+    {
+      parser_free_tree (parser, subq);
+    }
+
+  if (corr_subq)
+    {
+      parser_free_tree (parser, corr_subq);
+    }
```
ctshim: If you are coming from line 25149 or 25185, you may experience problems with free corr_subq and attempting free again in expr.
Please refer to the comments on line 25197 of the code before the change

beyondykk9: Ok, you are right. I will fixed it to move freeing corr_sub down

ctshim (approved):

shparkcubrid (approved):

hyunikn (approved):

src/parser/xasl_generation.c
```diff
@@ -25217,6 +25204,37 @@ pt_to_merge_insert_query (PARSER_CONTEXT * parser, PT_NODE * select_list, PT_MER
...
+	  parser_free_tree (parser, expr->info.expr.arg1);
+	  corr_subq = NULL;
+	}
+      parser_free_tree (parser, expr);
+    }
+
+  if (corr_subq)
+    {
+      parser_free_tree (parser, corr_subq);
+    }
```
youngjinj, line 25226: How about below?
```suggestion
  if (corr_subq)
    {
      parser_free_tree (parser, corr_subq);
    }
   
  if (expr)
    {
      parser_free_tree (parser, expr);
    }
```

src/parser/xasl_generation.c
```diff
@@ -25111,7 +25111,7 @@ pt_to_merge_update_query (PARSER_CONTEXT * parser, PT_NODE * select_list, PT_MER
 PT_NODE *
 pt_to_merge_insert_query (PARSER_CONTEXT * parser, PT_NODE * select_list, PT_MERGE_INFO * info)
 {
-  PT_NODE *subq, *corr_subq, *expr, *and_expr, *value;
+  PT_NODE *subq, *corr_subq = NULL, *expr = NULL, *and_expr = NULL, *value = NULL;
```
youngjinj, line 25114: It is trivial.
```suggestion
  PT_NODE *subq = NULL, *corr_subq = NULL, *expr = NULL, *and_expr = NULL, *value = NULL;
```

youngjinj (approved):

hgryoo (approved):

src/parser/xasl_generation.c
beyondykk9: **subq** is always assigned some address as new node.

src/parser/xasl_generation.c
beyondykk9: This code sequence is right because the expr could be assigend to **corr_subq**.

PR #3921, opened 2022-10-28
[LETS-568] Synchronize mvcc_next_id of PTS with that of PS when booting PTS

joohok:
http://jira.cubrid.org/browse/LETS-568

Purpose

At the time of PTS booting, the PTS determines mvcc_next_id with only checkpoint information. 
There may be a difference between the mvcc_next_id when the checkpoint is left and the mvcc_next_id when the PTS boots.
A problem may occur when ATS requests vacuuming a mvccid between the mvcc_next_id at the time of PTS booting (latest mvccid at ATS) and the mvcc_next_id at the time of leaving the checkpoint (past).
Therefore, it needs to synchronize log_Gl.hdr.mvcc_next_id with PS. 

Implementation

log_recovery_analysis() traverses log records up to log_Gl.hdr.append_lsa.
The PTS waits for the PS to finish replicating up to log_Gl.hdr.append_lsa before it completes the analysis. 
Therefore, the largest mvccid is updated during analysis phase, and largest mvccid will have the same value as log_Gl.hdr.mvcc_next_id on PS.

### DIFF
--- a/src/transaction/log_recovery.c
+++ b/src/transaction/log_recovery.c
@@ -2911,6 +2911,19 @@ log_recovery_analysis_internal (THREAD_ENTRY * thread_p, INT64 * num_redo_log_re
 	  break;
 	}
 
+      if (is_passive_transaction_server ())
+	{
+	  const int tran_index = logtb_find_tran_index (thread_p, tran_id);
+	  if (tran_index != NULL_TRAN_INDEX)
+	    {
+	      const LOG_TDES *tdes = LOG_FIND_TDES (tran_index);
+	      if (!MVCC_ID_PRECEDES (tdes->mvccinfo.id, context.get_largest_mvccid ()))
+		{
+		  context.set_largest_mvccid (tdes->mvccinfo.id);
+		}
+	    }
+	}
+
       prev_prev_lsa = prev_lsa;
       prev_lsa = context.get_end_redo_lsa ();
 
@@ -3192,8 +3205,6 @@ log_recovery_build_mvcc_table_from_trantable (THREAD_ENTRY * thread_p, MVCCID re
 	    }
 	  prev_mvccid = curr_mvccid;
 	}
-
-      log_Gl.hdr.mvcc_next_id = largest_mvccid + 1;
     }
 }
 
@@ -3318,6 +3329,8 @@ log_recovery_analysis_from_trantable_snapshot (THREAD_ENTRY * thread_p,
   // such situations, but a proper solution is needed
   //
   log_Gl.mvcc_table.complete_mvccids_if_still_active (LOG_SYSTEM_TRAN_INDEX, in_gaps_mvccids, false);
+
+  log_Gl.hdr.mvcc_next_id = log_rcv_context.get_largest_mvccid () + 1;
 
   LOG_SET_CURRENT_TRAN_INDEX (thread_p, sys_tran_index);
 }
--- a/src/transaction/log_recovery.c
+++ b/src/transaction/log_recovery.c
@@ -1230,11 +1230,7 @@ log_rv_analysis_mvcc_undo_redo (THREAD_ENTRY * thread_p, int tran_id, LOG_LSA * 
        * So, largest_mvccid will be updated during log_recovery_analysis () only for PTS,
        * and it will be used to set log_Gl.hdr.mvcc_next_id in log_recovery_analysis_from_trantable_snapshot ().
        */
-
-      if (MVCC_ID_PRECEDES (context.get_largest_mvccid (), tdes->mvccinfo.id))
-	{
-	  context.set_largest_mvccid (tdes->mvccinfo.id);
-	}
+      context.set_largest_mvccid (tdes->mvccinfo.id);
     }
 
   return error_code;
@@ -1773,11 +1769,7 @@ log_rv_analysis_assigned_mvccid (THREAD_ENTRY * thread_p, int tran_id, LOG_LSA *
        * So, largest_mvccid will be updated during log_recovery_analysis () only for PTS,
        * and it will be used to set log_Gl.hdr.mvcc_next_id in log_recovery_analysis_from_trantable_snapshot ().
        */
-
-      if (MVCC_ID_PRECEDES (context.get_largest_mvccid (), tdes->mvccinfo.id))
-	{
-	  context.set_largest_mvccid (tdes->mvccinfo.id);
-	}
+      context.set_largest_mvccid (tdes->mvccinfo.id);
     }
 
   return NO_ERROR;
--- a/src/transaction/log_recovery.c
+++ b/src/transaction/log_recovery.c
@@ -1228,13 +1228,13 @@ log_rv_analysis_mvcc_undo_redo (THREAD_ENTRY * thread_p, int tran_id, LOG_LSA * 
     {
       /* Since there is no recovery redo phase on PTS, PTS does not update log_Gl.hdr.mvcc_next_id.
        * So, largest_mvccid will be updated during log_recovery_analysis () only for PTS,
-       * and it will be used to set log_Gl.hdr.mvcc_next_id.
+       * and it will be used to set log_Gl.hdr.mvcc_next_id in log_recovery_analysis_from_trantable_snapshot ().
        */
 
-      if (!MVCC_ID_PRECEDES (tdes->mvccinfo.id, context.get_largest_mvccid ()))
-        {
-          context.set_largest_mvccid (tdes->mvccinfo.id);
-        }
+      if (MVCC_ID_PRECEDES (context.get_largest_mvccid (), tdes->mvccinfo.id))
+	{
+	  context.set_largest_mvccid (tdes->mvccinfo.id);
+	}
     }
 
   return error_code;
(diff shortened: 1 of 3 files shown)


hornetmj (approved):

src/transaction/log_recovery.c
```diff
@@ -2911,6 +2911,23 @@ log_recovery_analysis_internal (THREAD_ENTRY * thread_p, INT64 * num_redo_log_re
 	  break;
 	}
 
+      if (is_passive_transaction_server ())
```
cristiarg: What if the current log record is not an MVCC record?

This information - the largest encountered mvccid - can be collected also in the following functions:
- --> `log_rv_analysis_record`
  - --> `log_rv_analysis_record_on_tran_server`
    - --> `log_rv_analysis_mvcc_undo_redo` (need to pass `log_recovery_context` instance as an argument, the rest of the information is there)
    - --> `log_rv_analysis_assigned_mvccid` (same, pass `log_recovery_contenxt` as argument)

It is more generic as already needed information is there: proper MVCCID log record, tdes, mvccid value. Only context needs to be supplied from one level up into `log_rv_analysis_mvcc_undo_redo` and `log_rv_analysis_assigned_mvccid`.

I ask to move the collection of the "largest mvccid" because this PR game me an idea of improving the missing mvccid's handling (see comment before `log_Gl.mvcc_table.complete_mvccids_if_still_active`):
- the `log_recovery_context` can be extended to also take the set of "missing mvccids" and substract from them along the way while executing the log recovery analysis
- thus it would minimize even more the uncertainty and would make the logic easier to debug if the assumption described in the aforementioned comment is ever broken
But this is for another time. For now just having the collecting part moved to the lower level functions is enough.

src/transaction/log_recovery.c
```diff
@@ -3319,6 +3334,20 @@ log_recovery_analysis_from_trantable_snapshot (THREAD_ENTRY * thread_p,
   //
   log_Gl.mvcc_table.complete_mvccids_if_still_active (LOG_SYSTEM_TRAN_INDEX, in_gaps_mvccids, false);
 
+  if (!MVCC_ID_PRECEDES (log_rcv_context.get_largest_mvccid (), log_Gl.hdr.mvcc_next_id))
```
cristiarg, line 3339: Nit.
Why not?
I find it more natural this way.
```suggestion
  if (log_Gl.hdr.mvcc_next_id < log_rcv_context.get_largest_mvccid ())
```
Also `!MVCC_ID_PRECEDES` implies also equality, and in case of equality, there's no need to update. If I'm not mistaken.
Or, to keep to the local norms:
```suggestion
  if (MVCC_ID_PRECEDES (log_Gl.hdr.mvcc_next_id, log_rcv_context.get_largest_mvccid ()))
```

src/transaction/log_recovery.c
joohok: > What if the current log record is not an MVCC record?

The mvccid of the transaction that left the log will be used. If there is no mvcc operation within that transaction, then `MVCCID_NULL` will be used. 

> This information - the largest encountered mvccid - can be collected also in the following functions:
> 
> * --> `log_rv_analysis_record`
>   
>   * --> `log_rv_analysis_record_on_tran_server`
>     
>     * --> `log_rv_analysis_mvcc_undo_redo` (need to pass `log_recovery_context` instance as an argument, the rest of the information is there)
>     * --> `log_rv_analysis_assigned_mvccid` (same, pass `log_recovery_contenxt` as argument)
>       It is more generic as already needed information is there: proper MVCCID log record, tdes, mvccid value.

I am aware of the detailed function for parsing mvcc records (`log_rv_analysis_assigned_mvccid`, `log_rv_analysis_mvcc_undo_redo`). However, updating the variables used only for the passive transaction server seems to be out of the scope of the above functions, so I modified it as in the current PR.
 
> I ask to move the collection of the "largest mvccid" because this game me an idea of improving the missing mvccid's handling (see comment before `log_Gl.mvcc_table.complete_mvccids_if_still_active`):
> 
> * the `log_recovery_context` can be extended to also take the set of "missing mvccids" and substract from them along the way while executing the log recovery analysis
> * thus it would minimize even more the uncertainty and would make the logic easier to debug if the assumption described in the aforementioned comment is ever broken
>   But this is for another time.

This is also a fact recognized while analyzing recovery from PTS. If we can remove mvccid from `missing mvccids` during analysis, we can filter out `active mvccid` from the missing mvccid set. But, as you said, this is for another time. I will make another issue about this one.

Summary : 
I'm not sure if it would be better to move the update of largest mvccid inside those function ((`log_rv_analysis_assigned_mvccid`, `log_rv_analysis_mvcc_undo_redo`) . However, I think it will be less expensive than finding log_tdes for all log records and comparing mvccid.

I will fix it. thanks.

cristiarg: The comparison will still be needed.
Howevever it will be one `logtb_find_tran_index` and one `LOG_FIND_TDES` less for each log record.

src/transaction/log_recovery.c
```diff
@@ -3319,6 +3334,20 @@ log_recovery_analysis_from_trantable_snapshot (THREAD_ENTRY * thread_p,
...
+      /* The updated log_Gl.hdr.mvcc_next_id in log_recovery_build_mvcc_table_from_trantable ()
+       * may not be the same as the value of log_Gl.hdr.mvcc_next_id with other servers (ATS, PS)
+       * that have progressed up to log_Gl.hdr.append_lsa.
+       * Since log_Gl.hdr.mvcc_next_id is simply determined with checkpoint information,
+       * if there are some mvcc operation after the checkpoint, those mvccids can not be known on PTS.
+       * However, if ATS requests vacuum for mvccid performed after checkpoint, PTS cannot know about the mvccid, so a crash may occur.
+       * log_Gl.hdr.mvcc_next_id is usually updated in the REDO phase (PS),
+       * but since REDO is not performed on PTS, the largest mvccid obtained in the ANALYSIS phase will be used.
+       */
+      log_Gl.hdr.mvcc_next_id = log_rcv_context.get_largest_mvccid () + 1;
```
```suggestion
      log_Gl.hdr.mvcc_next_id = log_rcv_context.get_largest_mvccid ();
      MVCCID_FORWARD(log_Gl.hdr.mvcc_next_id);
```

wrlawodms (approved):
I agree with cristi's opinion. `logtb_find_tran_index` every time could be expensive.

src/transaction/log_recovery.c
joohok: Even `if (log_Gl.hdr.mvcc_next_id == log_rcv_context.get_larget_mvccid())`,
log_Gl.hdr.mvcc_next_id should be added by 1.

Using macros like MVCC_ID_PRECEDES is also intended to make the code easier to understand, but I used the macro above in case there is a change in the comparison policy for MVCCID values ​​in the future. 
The mvccid values ​​are designed to be assigned circular, but they are not implemented by design, so I think there will be changes in the future.

src/transaction/log_recovery.c
joohok: Oh.. Thanks

src/transaction/log_recovery.c
joohok: Thanks. I fixed.

src/transaction/log_recovery.c
joohok: I applied your suggestion when set_largest_mvccid () is called.

src/transaction/log_recovery.c
```diff
@@ -1753,6 +1767,19 @@ log_rv_analysis_assigned_mvccid (THREAD_ENTRY * thread_p, int tran_id, LOG_LSA *
   auto rec = (const LOG_REC_ASSIGNED_MVCCID *) (log_page_p->area + log_lsa->offset);
   tdes->mvccinfo.id = rec->mvccid;
 
+  if (is_passive_transaction_server ())
```
cristiarg, line 1766: Nit.
Actually, the change in `log_recovery_context` could be done regardless of the type of transaction server.
On an ATS the result will simply not be used.

src/transaction/log_recovery.c
```diff
@@ -1753,6 +1767,19 @@ log_rv_analysis_assigned_mvccid (THREAD_ENTRY * thread_p, int tran_id, LOG_LSA *
...
+    {
+      /* Since there is no recovery redo phase on PTS, PTS does not update log_Gl.hdr.mvcc_next_id.
+       * So, largest_mvccid will be updated during log_recovery_analysis () only for PTS,
+       * and it will be used to set log_Gl.hdr.mvcc_next_id in log_recovery_analysis_from_trantable_snapshot ().
+       */
+
+      if (MVCC_ID_PRECEDES (context.get_largest_mvccid (), tdes->mvccinfo.id))
```
cristiarg: Nit.
This comparison and set should be moved inside `set_largest_mvccid`.
Already the name `largest" suggests that there might be some condition verification inside. :)

cristiarg (approved):

src/transaction/log_recovery.c
joohok: OK, I think it is better

src/transaction/log_recovery.c
joohok: I agree with you. However, I would like to leave this code as is, so that readers can more easily understand the context of this code.

PR #3922, opened 2022-10-31
[CBRD-24475] Refactor memory allocation module

hgryoo:
http://jira.cubrid.org/browse/CBRD-24475

Unify the memory allocation API. The lea heap is used for both workspace (in CS_MODE) and the server's private heap.

- Integrate quick_fit.c into memory_alloc.c to use db_private_XXX interface.
   - For HEAP_ID for workspace (ws_heap_id) is defined and used in CS_MODE and SA_MODE
   - For HEAP_ID for server-side memory in SA_MODE, private_heap_id is used
   - By db_on_server global variable, the allocation API determines which heap_id to be allocated.
- Rename memory allocation-related files to have the same prefix name.
   - fixed_alloc -> memory_alloc_fixed
   - area_alloc -> memory_alloc_area
- Split memory alignment related definitions and functions to memory_alignment.hpp
- Move unrelated functions to misc_string
  - ansisql_strcmp (), ansisql_strcasecmp ()


src/transaction/boot_cl.c
```diff
@@ -1534,6 +1534,7 @@ boot_client_all_finalize (bool is_er_final)
 {
   if (BOOT_IS_CLIENT_RESTARTED () || boot_Is_client_all_final == false)
     {
+#if !defined (SA_MODE)
```
hgryoo, line 1538: Here, the server credential structure is set in xboot_register_client(). In CS_MODE, the structure is unpacked (internally the attributes are allocated by the allocation API). However, In SA_MODE, xboot_register_client() just set global variables (not allocated, the memory exists in the data region) to the server credential

src/base/memory_alignment.hpp
```diff
@@ -0,0 +1,110 @@
...
+    }
+  else if (n >= (int) sizeof (short))
+    {
+      return (int) sizeof (short);
+    }
+  else
+    {
+      return 1;
+    }
+}
```
hgryoo: Hmm... `PTR_ALIGNMENT` is defined according to the `__WORD_SIZE` above. But here, it compares with the `sizeof (void*)`. I'm not sure about this.

src/base/memory_alloc.h
```diff
@@ -137,6 +75,14 @@ extern void db_scramble (void *region, int size);
             (ptr) = NULL; \
           } \
         } while (0)
+
+#define db_ws_free_and_init(ptr) \
+        do { \
+          if ((ptr)) { \
```
youngjinj: It seems to be a duplicate.

```
db_ws_free()
-> db_private_free()
-> db_private_free_debug() or db_private_free_release()
```

```
/* src/base/memory_alloc.c  */

#if !defined(NDEBUG)
void
db_private_free_debug (THREAD_ENTRY * thrd, void *ptr, bool rc_track, const char *caller_file, int caller_line)
#else /* NDEBUG */
void
db_private_free_release (THREAD_ENTRY * thrd, void *ptr, bool rc_track)
#endif				/* NDEBUG */
{
  if (ptr == NULL)	/********** this **********/
    {
      return;
    }
```

src/base/memory_alloc.h
```diff
@@ -155,19 +101,34 @@ extern void db_scramble (void *region, int size);
...
 extern HL_HEAPID private_heap_id;
+extern HL_HEAPID ws_heap_id;
 
 #define os_malloc(size) (malloc (size))
 #define os_free(ptr) (free (ptr))
 #define os_realloc(ptr, size) (realloc ((ptr), (size)))
 
+/* allocation APIs for workspace */
+#define db_ws_alloc(size) \
+        db_private_alloc(NULL, size)
```
youngjinj, line 114: It is trivial.

Isn't our coding convention to put a space after the function name?
It seems to be mixed.
```
#define db_ws_alloc(size) \
        db_private_alloc (NULL, size)
```

src/base/memory_alignment.hpp
ctshim: I recommend "#define PTR_ALIGNMENT sizeof(void*)"
sizeof(void*) will be automatically calculated according to the number of bits.

src/base/memory_alloc.h
hgryoo: thank you. I'll fix it

src/base/memory_alloc.h
hgryoo: In the macro alias, I found that space is not used
```
#if !defined(NDEBUG)
#define db_private_alloc(thrd, size) \
        db_private_alloc_debug(thrd, size, true, __FILE__, __LINE__)
#define db_private_free(thrd, ptr) \
        db_private_free_debug(thrd, ptr, true, __FILE__, __LINE__)
#define db_private_realloc(thrd, ptr, size) \
        db_private_realloc_debug(thrd, ptr, size, true, __FILE__, __LINE__)
...
```

ctshim (approved):

src/base/misc_string.h
```diff
@@ -41,6 +41,9 @@ extern "C"
   extern char *ustr_upper (char *s);
   extern char *ustr_lower (char *s);
 
+  extern int ansisql_strcasecmp (const char *s, const char *t);
+  extern int ansisql_strcmp (const char *s, const char *t);
```
beyondykk9, line 45: ansisql_strcmp is only used at extendible-hash routine in "storage/extendible_hash.c".
therefore, the ansisql_strcmp looks better to move to it as a static function.  and it needs to rename the function name  if it's necessary.
ansisql_strcasecmp is only used at db_query_get_tuple_value_by_name in "compat/db_query.c".
therefore, the ansisql_strcasecmp looks better to move to it as a static function. and it needs to rename the function name if it's necessary.

hgryoo: Is there any possibility that they are used in the other modules? The ansisql_XXX functions deal with trailing spaces. Maybe in parsing or query modules, this function could be used in the future.

beyondykk9 (approved):

youngjinj (approved):

shparkcubrid (approved):

hyunikn (approved):

PR #3923, opened 2022-11-01
[CBRD-24541] Core dump occurs when 'EXISTS' operation or 'count stop' is used for index scan of partitioned table

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-24541

This is an issue that may occur if the scan is stopped in the middle of a partition table index scan by 'count stop', etc.
I solve this problem by adding a routine to initialize partition related information in qexec_close_scan().

### DIFF
--- a/src/query/query_executor.c
+++ b/src/query/query_executor.c
@@ -6833,7 +6833,10 @@ qexec_close_scan (THREAD_ENTRY * thread_p, ACCESS_SPEC_TYPE * curr_spec)
       curr_spec->pruned = false;
 
       /* init btid */
-      BTID_COPY (&curr_spec->indexptr->btid, &curr_spec->btid);
+      if (curr_spec->indexptr)
+	{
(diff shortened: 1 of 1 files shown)


src/query/query_executor.c
```diff
@@ -6844,6 +6823,22 @@ qexec_close_scan (THREAD_ENTRY * thread_p, ACCESS_SPEC_TYPE * curr_spec)
 	  break;
 	}
     }
+
```
beyondykk9, line 6826: This routine can be moved into the prior "switch" clause with just adding "init btid" logic.

hgryoo (approved):

ctshim (approved):

hyunikn (approved):

beyondykk9 (approved):

src/query/query_executor.c
shparkcubrid: No. only if perfmon_is_perf_tracking () is true.

youngjinj (approved):

PR #3924, opened 2022-11-01
[LETS-563] Fix a PTS not to start with mvccid already vacuumed on ATS

wrlawodms:
http://jira.cubrid.org/browse/LETS-563

- Update the active oldest mvccid at the end of log_recovery_analysis_from_trantable_snapshot().
- Set m_transaction_lowest_visible_mvccids[tran_index] only on ATS. LETS-555 is reverted because it doesn't make sense to set the variable while replicating.

Please see the Jira issue for more details.

### DIFF
--- a/src/transaction/log_recovery.c
+++ b/src/transaction/log_recovery.c
@@ -3336,7 +3336,9 @@ log_recovery_analysis_from_trantable_snapshot (THREAD_ENTRY * thread_p,
   //
   log_Gl.mvcc_table.complete_mvccids_if_still_active (LOG_SYSTEM_TRAN_INDEX, in_gaps_mvccids, false);
 
-  /* Update the oldest active mvccid with the currnet mvcctable, which will be taken into account by the global ATS vacuum. */
+  /* Update the oldest active mvccid with the current mvcctable,
+   * which will be taken into account by the global ATS vacuum.
+   */
   log_Gl.mvcc_table.update_oldest_active ();
 
   if (!MVCC_ID_PRECEDES (log_rcv_context.get_largest_mvccid (), log_Gl.hdr.mvcc_next_id))
--- a/src/transaction/log_recovery.c
+++ b/src/transaction/log_recovery.c
@@ -3336,6 +3336,7 @@ log_recovery_analysis_from_trantable_snapshot (THREAD_ENTRY * thread_p,
   //
   log_Gl.mvcc_table.complete_mvccids_if_still_active (LOG_SYSTEM_TRAN_INDEX, in_gaps_mvccids, false);
 
+  /* Update the oldest active mvccid with the currnet mvcctable, which will be taken into account by the global ATS vacuum. */
   log_Gl.mvcc_table.update_oldest_active ();
 
   if (!MVCC_ID_PRECEDES (log_rcv_context.get_largest_mvccid (), log_Gl.hdr.mvcc_next_id))
--- a/src/transaction/log_recovery.c
+++ b/src/transaction/log_recovery.c
@@ -3319,6 +3319,8 @@ log_recovery_analysis_from_trantable_snapshot (THREAD_ENTRY * thread_p,
   //
   log_Gl.mvcc_table.complete_mvccids_if_still_active (LOG_SYSTEM_TRAN_INDEX, in_gaps_mvccids, false);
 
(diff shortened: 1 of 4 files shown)


joohok (approved):

hornetmj (approved):

cristiarg:
@wrlawodms Can you please merge from `scalability_dev`? Thanks.

wrlawodms:
@cristiarg merged.

src/transaction/log_recovery.c
```diff
@@ -3336,6 +3336,9 @@ log_recovery_analysis_from_trantable_snapshot (THREAD_ENTRY * thread_p,
   //
   log_Gl.mvcc_table.complete_mvccids_if_still_active (LOG_SYSTEM_TRAN_INDEX, in_gaps_mvccids, false);
 
+  /* Update the oldest active mvccid with the currnet mvcctable, which will be taken into account by the global ATS vacuum. */
```
cristiarg: line too long; typo

src/transaction/mvcc_table.cpp
```diff
@@ -544,15 +544,21 @@ mvcctable::complete_mvcc (int tran_index, MVCCID mvccid, bool committed)
...
        * - VACUUM clean up transaction modifications
        * - the system crash before LOG_COMMIT of current transaction
        *
        * It will be set to NULL after LOG_COMMIT
+       *
+       * 2. This is only the case if it's active transaction server because only ats generates log records and do vacuum.
+       *
+       * 3. This also prevents ats from vacumming a version on which a RO transaction in PTS may start.
+       *  A newly connected PTS starts replicating from the end of log on PS, and a RO transaction can start with a snapshot of then.
+       *  If VACUUM clean up the modifications before its LOG_COMMIT is flushed on PS, it possibly cleans up data seen by the RO transaction on a PTS.
```
cristiarg: lines should be limited to 120

src/transaction/mvcc_table.cpp
```diff
@@ -544,15 +544,21 @@ mvcctable::complete_mvcc (int tran_index, MVCCID mvccid, bool committed)
   next_tran_status_finish (next_status, next_index);
 
   assert (tran_index < m_transaction_lowest_visible_mvccids_size);
-  if (committed)
+  if (committed && is_active_transaction_server ())
```
cristiarg, line 547: I may not understand completely, but I have a question.
On PTS, the else branch will be the one **always** executed for `LOG_COMMIT` - via `replicator_mvcc::complete_mvcc` -> `mvcctable::complete_mvcc`?
Is this the intention?
Is it because of the remark in the issue?
```
I assume that we don't need to consider the ABORT case because VACUUM the modifications of aborted trxs are already allowed, and it may have no actual effects.
```
If so, I would still prefer to guard the entire if-else in `is_active_transaction_server`:
```
  if (is_active_transaction_server)
  {
    if (commited)
    ....
    else
  {
```
Would it be correct?
Would it express the intent better?

wrlawodms: > Is this the intention?

Yes. I assume `m_transaction_lowest_visible_mvccids[tran_index]` has to be set to NULL because it's reset when aborting and the below condition in 565L says it may be set somewhere.
`if (tran_lowest_active == MVCCID_NULL || MVCC_ID_PRECEDES (tran_lowest_active, mvccid))`

The only thing that could seem awkward is that it will be reset later, even not here. It means we can skip resetting here for Commit. However, I'd like to keep the tracking of `oldest_active_event::COMPLETE_MVCC`, and it's no harm to reset twice.

wrlawodms: > Is it because of the remark in the issue?

No. It's a statement on the ATS side. While all log records are flushed during COMMIT, but not during ABORT, and mvccid on ATS can advance even in the case of ABORT. Someone might think that it is a problem because a new PTS can have a lower mvccid than ATS at the time. The remark  is about that.

src/transaction/log_recovery.c
wrlawodms: fixed

src/transaction/mvcc_table.cpp
wrlawodms: fixed

cristiarg (approved):

PR #3925, opened 2022-11-02
[CBRD-24544] not allow to alter the column type to a type that does not support auto-increment.

beyondykk9:
http://jira.cubrid.org/browse/CBRD-24544

In DB table, the column type can be altered to any DB type, however, if the column has an auto-increment attribute the type altering should be checked severly not to allow non numeric type.
```
CREATE a_tbl (a int AUTO_INCREMENT, b VARCHAR);
ALTER TABLE a_tbl MODIFY COLUMN a VARCHAR;
```
The above "ALTER" command must not be allowed because the auto-increment doesn't work properly.

 If the auto-increment column type is overflown as altering the type, the ALTER command results error as following message:

**ERROR: Start value of an auto increment constraint should be less then max value.** 

Example:

```
CREATE TABLE a_tbl (a INT AUTO_INCREMENT (1, 1000000000), b INT);
INSERT INTO a_tbl (b) VALUES (1);
INSERT INTO a_tbl (b) VALUES (2);
ALTER TABLE a_tbl MODIFY COLUMN a SHORT;

ERROR: Start value of an auto increment constraint should be less then max value. 
```


beyondykk9:
The failed Test cases should be modified as changed specification for modifying column type having auto-increment attribute.

ctshim (approved):

hgryoo (approved):

src/query/execute_schema.c
```diff
@@ -10502,6 +10502,35 @@ do_change_att_schema_only (PARSER_CONTEXT * parser, DB_CTMPL * ctemplate, PT_NOD
       goto exit;
     }
 
+  /* if it is an auto_increment column, check its domain */
+  if (found_att->auto_increment != NULL)
+    {
+      PT_NODE *att;
+      const char *att_name;
+
+      att = attribute->info.attr_def.attr_name;
```
hyunikn: Though minor, it seems better to put these two lines under the default case below
because att and att_name are not used in other cases.

shparkcubrid (approved):

youngjinj (approved):

src/query/execute_schema.c
beyondykk9: Ok, thanks

hyunikn (approved):

PR #3931, opened 2022-11-07
[LETS-554] use OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT fetch mode for atomic replication sequences as well

cristiarg:
http://jira.cubrid.org/browse/LETS-554

For Passive Transaction Server atomic replication, use also `OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT` fetch mode as it is used for the non-atomic replication.

For Passive Transaction Server atomic replication, the behavior is as such: if a page from within an atomic sequence cannot be fixed, it is skipped altogether - see `atomic_replication_helper::atomic_log_sequence::append_log`.
However, the following scenarios are described and discussed:

(1)
- If, while an atomic replication sequences is being processed (log records are processed and pages are fixed),
- a page fails to be fixed (for whatever reason), see next diagram:
```
   START                                        END
     LSA1   LSA2     LSA3  LSA4     LSA5      LSA6
           (VPID1) (VPID2)(VPID3)  (VPID4)
──────┬───────┬───────┬──────┬───────┬─────────┬─────────►
      │       │       │      │       │         │
             (OK)   (OK)   (NOK)
                                  ▲
                                  │
                                 PTS replication here
```
- the atomic replication sequence will just move on - not having fixed the page with VPID3 affected by LSA4 log record
- let's say that, overall, PTS replication is somewhere past LSA4
- if, at this point, a client transaction wants to fix the page with VPID3 and succeeds (the page is requested and fetched from Page Server - e.g. maybe Page Server's replication would have cycled the page VPID3 though a de-allocate, re-allocate sequence (as it often happens)
- when the client transaction will request the page to PS it will do this under the rules of the page synchronization mechanism:
  - the request to PS will be done with a `target_replication_lsa` with the value >LSA4 (for this, see how `replicator::m_redo_lsa` progresses and the logic in `pgbuf_read_page_from_file_or_page_server`
  - upon retrieval of the page from PS, if the page is ahead of PTS replication it will, again, wait for the PTS replication to have progressed past the page's LSA (past the LSA the page had when it was retrieved from Page Server) - (for this, see the mechanism implemented in `pgbuf_fix_old_and_check_repl_desync` and associated functions);
  - the PTS replication's overall progress, when atomic replication sequences are present only moves forward when entire atomic replication sequences are concluded (for this, see the how `atomic_replicator::m_lowest_unapplied_lsa` is handled)
- thus we can conclude that in this case, the client transaction will not see inconsistent data for page VPID3

(2)
- what if, within an atomic sequence, there is more than one log record which changes the same page
- and the first attempt to fix fails while a subsequent one succeeds
- this is deemed to break both the atomicity of the sequence and the consistency of the data
- one option for this is:
  - if one page within one atomic sequence fails to be fixed for modification
  - "throw away" the entire sequence by:
    - first concluding the sequence by fixing all the pages within the sequence
    - secondly, instead of applying the log records, by "force-evicting" all the already fixed pages from the page buffer (using some mechanism that already exists such as `pgbuf_invalidate_bcb`/`pgbuf_invalidate` if appropriate, or some mechanism that must be implemented)
- a solution for this issue is not provided within this patch - will be taken up with http://jira.cubrid.org/browse/LETS-583 .

Other changes:
- comment and small corrections for `pgbuf_request_data_page_from_page_server`

### DIFF
--- a/src/storage/page_buffer.c
+++ b/src/storage/page_buffer.c
@@ -1053,10 +1053,12 @@ static int pgbuf_unlock_page (THREAD_ENTRY * thread_p, PGBUF_BUFFER_HASH * hash_
 			      int need_hash_mutex);
 static PGBUF_BCB *pgbuf_allocate_bcb (THREAD_ENTRY * thread_p, const VPID * src_vpid);
 static PGBUF_BCB *pgbuf_claim_bcb_for_fix (THREAD_ENTRY * thread_p, const VPID * vpid, PAGE_FETCH_MODE fetch_mode);
-static int pgbuf_read_page_from_file_or_page_server (THREAD_ENTRY * thread_p, const VPID * vpid, FILEIO_PAGE * io_page);
+static int pgbuf_read_page_from_file_or_page_server (THREAD_ENTRY * thread_p, const VPID * vpid,
+						     PAGE_FETCH_MODE fetch_mode, FILEIO_PAGE * io_page);
 static int pgbuf_read_page_from_file (THREAD_ENTRY * thread_p, const VPID * vpid, void *io_page);
 #if defined (SERVER_MODE)
-static int pgbuf_request_data_page_from_page_server (const VPID * vpid, log_lsa target_repl_lsa, FILEIO_PAGE * io_page);
+static int pgbuf_request_data_page_from_page_server (const VPID * vpid, PAGE_FETCH_MODE fetch_mode,
+						     log_lsa target_repl_lsa, FILEIO_PAGE * io_page);
 #endif // SERVER_MODE
 static int pgbuf_victimize_bcb (THREAD_ENTRY * thread_p, PGBUF_BCB * bufptr);
 static int pgbuf_bcb_safe_flush_internal (THREAD_ENTRY * thread_p, PGBUF_BCB * bufptr, bool synchronous, bool * locked);
@@ -8027,7 +8029,8 @@ pgbuf_claim_bcb_for_fix (THREAD_ENTRY * thread_p, const VPID * vpid, PAGE_FETCH_
 	  CUBRID_IO_READ_START (query_id);
 	}
 #endif /* ENABLE_SYSTEMTAP */
-      int error_code = pgbuf_read_page_from_file_or_page_server (thread_p, vpid, &bufptr->iopage_buffer->iopage);
+      int error_code = pgbuf_read_page_from_file_or_page_server (thread_p, vpid, fetch_mode,
+								 &bufptr->iopage_buffer->iopage);
 
 #if defined(ENABLE_SYSTEMTAP)
       if (monitored)
@@ -8143,8 +8146,18 @@ pgbuf_claim_bcb_for_fix (THREAD_ENTRY * thread_p, const VPID * vpid, PAGE_FETCH_
   return bufptr;
 }
 
+/*
+ * pgbuf_read_page_from_file_or_page_server () -
+ *
+ *    return : error code
+ *
+ * vpid (in)            : page identifier
+ * fetch_mode (in)      : fetch mode
+ * io_page(out): Address where content of page is stored. Must be of sufficient size.
+ */
 int
-pgbuf_read_page_from_file_or_page_server (THREAD_ENTRY * thread_p, const VPID * vpid, FILEIO_PAGE * io_page)
+pgbuf_read_page_from_file_or_page_server (THREAD_ENTRY * thread_p, const VPID * vpid, PAGE_FETCH_MODE fetch_mode,
+					  FILEIO_PAGE * io_page)
 {
 
 // Server type --- Storage Config Type --- Data Type --- Connection to PS --- Local Read --- Remote Read
@@ -8193,7 +8206,7 @@ pgbuf_read_page_from_file_or_page_server (THREAD_ENTRY * thread_p, const VPID * 
 	  FILEIO_PAGE *second_io_page = reinterpret_cast<FILEIO_PAGE *> (buffer_uptr.get ());
 	  // *INDENT-ON*
 	  // The page returned can be null during recovery in the case that the page was already deallocated on PS
-	  error_code = pgbuf_request_data_page_from_page_server (vpid, target_repl_lsa, second_io_page);
+	  error_code = pgbuf_request_data_page_from_page_server (vpid, fetch_mode, target_repl_lsa, second_io_page);
 	  if (error_code != NO_ERROR)
 	    {
 	      ASSERT_ERROR ();
@@ -8281,7 +8294,7 @@ pgbuf_read_page_from_file_or_page_server (THREAD_ENTRY * thread_p, const VPID * 
 	}
       else
 	{
-	  return pgbuf_request_data_page_from_page_server (vpid, target_repl_lsa, io_page);
+	  return pgbuf_request_data_page_from_page_server (vpid, fetch_mode, target_repl_lsa, io_page);
 	}
     }
   return NO_ERROR;
@@ -8319,14 +8332,18 @@ pgbuf_read_page_from_file (THREAD_ENTRY * thread_p, const VPID * vpid, void *io_
 /*
  * pgbuf_request_data_page_from_page_server () - Sends a request for a page to Page Server.
  *   return: void
+ *
+ * vpid(in) : page identifier
+ * fetch_mode(in) : fetch mode
+ * target_repl_lsa(int) : an LSA sent to the Page Server which the Page Server's replication must have
+ *                        advanced past; part of the replication mechanism
+ * io_page(out) : Address where content of page is stored. Must be of sufficient size.
  */
 static int
-pgbuf_request_data_page_from_page_server (const VPID * vpid, log_lsa target_repl_lsa, FILEIO_PAGE * io_page)
+pgbuf_request_data_page_from_page_server (const VPID * vpid, PAGE_FETCH_MODE fetch_mode,
+					  log_lsa target_repl_lsa, FILEIO_PAGE * io_page)
 {
   assert (get_server_type () == SERVER_TYPE_TRANSACTION);
-
-  // *INDENT-OFF*
-  /* Send a request to Page Server for the Page. */
 
   cubpacking::packer pac;
   size_t size = 0;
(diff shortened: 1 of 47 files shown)


src/storage/page_buffer.c
```diff
@@ -8319,35 +8332,46 @@ pgbuf_read_page_from_file (THREAD_ENTRY * thread_p, const VPID * vpid, void *io_
...
+ * target_repl_lsa(int) : an LSA sent to the Page Server which the Page Server's replication must have
+ *                        advanced past; part of the replication mechanism
+ * io_page(out) : Address where content of page is stored. Must be of sufficient size.
  */
 static int
-pgbuf_request_data_page_from_page_server (const VPID * vpid, log_lsa target_repl_lsa, FILEIO_PAGE * io_page)
+pgbuf_request_data_page_from_page_server (const VPID * vpid, const PAGE_FETCH_MODE fetch_mode,
+					  log_lsa target_repl_lsa, FILEIO_PAGE * io_page)
 {
   assert (get_server_type () == SERVER_TYPE_TRANSACTION);
```
wrlawodms: Nit
```suggestion
  assert (is_transaction_server());
```

src/storage/page_buffer.c
```diff
@@ -8319,35 +8332,46 @@ pgbuf_read_page_from_file (THREAD_ENTRY * thread_p, const VPID * vpid, void *io_
...
+      if (!(fetch_mode == RECOVERY_PAGE || fetch_mode == OLD_PAGE || fetch_mode == OLD_PAGE_PREVENT_DEALLOC))
+	{
+	  er_set (ER_FATAL_ERROR_SEVERITY, ARG_FILE_LINE, ER_GENERIC_ERROR, 0);
+	}
     }
-    else
+  else
     {
-      fetch_mode = OLD_PAGE;
+      if (!(fetch_mode == OLD_PAGE || fetch_mode == OLD_PAGE_MAYBE_DEALLOCATED
```
wrlawodms: When the replicator fixes pages so that it retrieves from PS, isn't it supposed to use OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT?

wrlawodms: I tried it on your branch, but it fails because of an assertion fails: 
```
cub_server: /home/cubrid/cubrid/lets/src/transaction/log_replication_atomic.cpp:109: virtual void cublog::atomic_replicator::redo_upto(cubthread::entry&, const log_lsa&): Assertion `!m_atomic_helper.is_part_of_atomic_replication (header.trid)' failed.
```
So, just asking :).

wrlawodms (commented):
Some questions:
From what I know, the replicator on PTS fixes page with the `OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT` fetch mode. It requests the page from PS only if it's not IN-TRANSIT. If it's in transit, it fails to fix and skips the redo log. So, I thought that the fetch mode propagation would be done when there is no "on-transit" page fixing.

1. However, it doesn't seem propagated when `OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT`. Is it intended that the fetch mode is only propagated when dealing with atomic sequences? 
2. Why don't atomic sequences use `OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT`? Is there no correctness issue?

src/storage/page_buffer.c
```diff
@@ -8319,35 +8332,46 @@ pgbuf_read_page_from_file (THREAD_ENTRY * thread_p, const VPID * vpid, void *io_
...
   pac.set_buffer (buffer.get (), size);
   vpid_utils::pack (pac, *vpid);
   cublog::lsa_utils::pack (pac, target_repl_lsa);
 
-  PAGE_FETCH_MODE fetch_mode;
+  // register a fatal error if an unexpected fetch mode
   if (log_is_in_crash_recovery_and_not_yet_completes_redo ())
     {
-      fetch_mode = RECOVERY_PAGE;
+      if (!(fetch_mode == RECOVERY_PAGE || fetch_mode == OLD_PAGE || fetch_mode == OLD_PAGE_PREVENT_DEALLOC))
```
hornetmj: Nit:

It would be better to use != iso == for readability.

hornetmj:
> Some questions: From what I know, the replicator on PTS fixes page with the `OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT` fetch mode. It requests the page from PS only if it's not IN-TRANSIT. If it's in transit, it fails to fix and skips the redo log. So, I thought that the fetch mode propagation would be done when there is no "on-transit" page fixing.
> 
> 1. However, it doesn't seem propagated when `OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT`. Is it intended that the fetch mode is only propagated when dealing with atomic sequences?
> 2. Why don't atomic sequences use `OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT`? Is there no correctness issue?

I expect @cristiarg to correct the wrong part again.

About OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT, it acts like the followings:
- when the PTS replicator fixes the data page with the OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT fetch mode
  - fix and redo when a data page exists in the data buffer or a data page is requested by other read transaction(in transit)  
  - otherwise, no replication is performed(no fix a data page)
- about atomic replication
  - data pages included in the atomic sequence must be fixed and redoed all together. Therefore, the OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT fetch mode cannot be used

cristiarg:
@hornetmj 

> I expect @cristiarg to correct the wrong part again.
> 
> About OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT, it acts like the followings:
> - when the PTS replicator fixes the data page with the OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT fetch mode
>   - fix and redo when a data page exists in the data buffer or a data page is requested by other read transaction(in transit)  
>   - otherwise, no replication is performed(no fix a data page)
> - about atomic replication
>   - data pages included in the atomic sequence must be fixed and redoed all together. Therefore, the OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT fetch mode cannot be used 

I have updated the PR description with a scenario that tries to answer your concerns. Please check it and follow-up. Thanks.

cristiarg:
@wrlawodms 

> Some questions:
> From what I know, the replicator on PTS fixes page with the `OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT` fetch mode. It requests the page from PS only if it's not IN-TRANSIT. If it's in transit, it fails to fix and skips the redo log. So, I thought that the fetch mode propagation would be done when there is no "on-transit" page fixing.
> 
> 1. However, it doesn't seem propagated when `OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT`. Is it intended that the fetch mode is only propagated when dealing with atomic sequences? 
> 2. Why don't atomic sequences use `OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT`? Is there no correctness issue? 

I have updated the implementation as suggested in the discussion.
I also added the detailed scenario that we discussed in the comment.
Please check it and follow-up.
Thanks.

src/storage/page_buffer.c
```diff
@@ -8388,12 +8397,13 @@ pgbuf_request_data_page_from_page_server (const VPID * vpid, log_lsa target_repl
...
 	{
-	  _er_log_debug (ARG_FILE_LINE, "[READ DATA] Received error: %d \n", error_code);
+	  _er_log_debug (ARG_FILE_LINE, "[READ DATA] Received error from Page Server: %d\n", error_code);
 	}
 
       if (error_code == ER_PB_BAD_PAGEID)
 	{
-	  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_PB_BAD_PAGEID, 2, VPID_AS_ARGS (vpid));
+	  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_PB_BAD_PAGEID, 2, vpid->pageid,
+		  fileio_get_volume_label_with_unknown (vpid->volid));
```
cristiarg, line 8406: Added missing argument

src/transaction/atomic_replication_helper.cpp
```diff
@@ -1092,6 +1092,7 @@ namespace cublog
   pgbuf_fix_or_ordered_fix (THREAD_ENTRY *thread_p, VPID vpid, LOG_RCVINDEX rcvindex,
 			    std::unique_ptr<PGBUF_WATCHER> &watcher_uptr, PAGE_PTR &page_ptr)
   {
+    constexpr PAGE_FETCH_MODE fetch_mode = OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT;
```
cristiarg, line 1095: Actually, it is the same value as used in `atomic_replicator::atomic_replicator` ctor

wrlawodms:
> upon retrieval of the page from PS, if the page is ahead of PTS replication it will, again, wait for the PTS replication to have progressed past the page's LSA (past the LSA the page had when it was retrieved from Page Server) - (for this, see the mechanism implemented in pgbuf_fix_old_and_check_repl_desync and associated functions);

It seems that the desynchronization is checked only for some cases if it's done explicitly such as btree cases. Does it mean all the atomic sequence cases should use the mechanism? Some of them (heap cases) don't seem to use it. Have I got it wrong?

cristiarg:
@wrlawodms 

> It seems that the desynchronization is checked only for some cases if it's done explicitly such as btree cases. Does it mean all the atomic sequence cases should use the mechanism? Some of them (heap cases) don't seem to use it. Have I got it wrong?

Atomic replication already checks if the redo function needs to be applied.
See callpath: `atomic_log_entry::apply_log_redo` -> `atomic_log_entry::apply_log_by_type` -> `log_rv_check_redo_is_needed`.
So, the following scenario is covered:
- a page is fixed (by either PTS client thread or PTS atomic replication thread)
- the page is waaaay ahead of where the atomic replication is at the moment
- if the page was fixed by the atomic replication thread, the "ahead" check is done and none of the redo functions are applied until the replication progresses enough
- if the page is fixed by a client thread/transaction, the replication is waited to progress and pass the LSA of the page (this mechanism is imperfect at the moment IMHO as it is based on continuous retries and an improved mechanism needs to be implemented - maybe based on an pub-sub mechanism: the client transaction halts and subscribes to be notified by the replication thread when the replication has advanced enough for the client transaction to continue)

Does this answer your concern?

About

> Some of them (heap cases) don't seem to use it. Have I got it wrong?

I don't think I understand this part of the question.

wrlawodms:
@cristiarg 
> if the page is fixed by a client thread/transaction, the replication is waited to progress and pass the LSA of the page

I guess this part is done by `pgbuf_fix_old_and_check_repl_desync()` and `pgbuf_wait_for_replication()`, which is described in the description. However, some operations using atomic sequences don't use the mechanism. An example of it is a heap operation modifying multiple pages. That's what I understand and am concerned about.

cristiarg:
@wrlawodms 

> However, some operations using atomic sequences don't use the mechanism. An example of it is a heap operation modifying multiple pages. That's what I understand and am concerned about.

I am sorry I don't understand your statement.
> some operations using atomic sequences don't use the mechanism
Which operations are you describing?
Operations that affect pages as part of the PTS replication?

> An example of it is a heap operation modifying multiple pages
Can you please list a specific example? I am really trying to understand.

Regarding the PTS client transactions, when the mechanism has been implemented, there was some analysis about what client operations need to check for page desynchronization (see http://jira.cubrid.org/browse/LETS-307 and http://jira.cubrid.org/browse/LETS-305 and their parent issue http://jira.cubrid.org/browse/LETS-259). That analysis might be incomplete, I don't know at the moment. And it might be very difficult to know only my examining the code. I assume only stress & consistency tests will be able to detect issues.

Thank you.

wrlawodms:
@cristiarg http://jira.cubrid.org/browse/LETS-305 handles the case I mentioned. I didn't consider the desynchronization in https://github.com/CUBRID/cubrid/pull/3787 so I thought that some operations affecting multiple pages don't use the desynchronization mechanism. I missed the issue (LETS-305). I'm so sorry if it takes you some time. Thanks.

src/transaction/atomic_replication_helper.cpp
wrlawodms: Since the fetch mod is a memeber variable of replicator, how about putting the function `pgbuf_fix_or_ordered_fix` into atomic_replicator as a member function?

wrlawodms (approved):

joohok (approved):

hornetmj (approved):

PR #3935, opened 2022-11-09
[LETS-524] Remove atomic replication for heap operations (Update, Delete)

joohok:
http://jira.cubrid.org/browse/LETS-524

Purpose

Atomic replication is the mechanism used for multi-page replication
When multiple pages need to be replicated at once, it was a proposed mechanism to guarantee atomicity, 
but it seems that it is not necessary to guarantee atomicity in heap operation for most cases, 
so this is a PR to remove the unnecessary atomic replication. 

Implementation

* Remove atomic replication in heap_delete/update_logical ()
* Do not remove the atomic replication when updating overflow pages without update REC_BIGONE record in  heap_update_bigone ()
* Do not remove the atomic replication when it needs to update a record and its previous version lsa (RVHF_SET_PREV_VERSION_LSA) atomically.
  *  heap_update_bigone 
  * heap_update_relocation
  * heap_update_home with MVCC


src/storage/heap_file.c
```diff
@@ -21035,17 +21035,6 @@ heap_delete_relocation (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * contex
 
   LOG_TDES *tdes = NULL;
 
-  // *INDENT-OFF*
-  // To ensure that, if started, the atomic replication area will also end.
-  scope_exit <std::function<void (void)>> log_on_exit ([&thread_p, &atomic_replication_flag]()
-  {
-    if (atomic_replication_flag == true)
```
cristiarg, line 21045: in some of the functions `atomic_replication_flag` is no longer used

src/storage/heap_file.c
```diff
@@ -22067,6 +22026,9 @@ heap_update_bigone (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * context, b
   /* Proceed with the update. the new record is prepared and for mvcc it should have the prev version lsa set */
   if (heap_is_big_length (context->recdes_p->length))
     {
+      log_append_empty_record (thread_p, LOG_START_ATOMIC_REPL, NULL);
+      atomic_replication_flag = true;
```
cristiarg, line 22042: is it a good idea to refactor in `heap_update_bigone` and replace the call to `log_append_empty_record` on the `exit:` label with the RAII mechanism used in the other functions (eg: as it was used in `heap_delete_home` function)?
(yes, that was a question :) )

cristiarg (commented):
I will also execute the scalability tests on this patch branch.

src/storage/heap_file.c
joohok: OK, I removed it

src/storage/heap_file.c
joohok: OK, Thanks. I change it as `heap_delete_home()` does.

src/storage/heap_file.c
```diff
@@ -22472,16 +22424,6 @@ heap_update_relocation (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * contex
...
-      /* 
-       * Updating forward record and setting prev_version_lsa have to be atomic
-       * if the home record (REC_RELOCATION) is not updated.
-       */
-      if (!atomic_replication_flag && is_mvcc_op)
-	{
-	  log_append_empty_record (thread_p, LOG_START_ATOMIC_REPL, NULL);
-	  atomic_replication_flag = true;
-	}
-
```
wrlawodms: We have to update a record and its previous version lsa (RVHF_SET_PREV_VERSION_LSA) atomically. It requires atomic operation. There are some cases. Please see 9cd23bcc8f.

wrlawodms: Although some cases still require atomic operation because of RVHF_SET_PREV_VERSION_LSA, I think we can shrink atomic operation's range even for them.

joohok: OK, I didn't take RVHF_SET_PREV_VERSION_LSA into account. Thanks. I will consider that case and update jira issue.

src/storage/heap_file.c
```diff
@@ -22070,6 +22038,9 @@ heap_update_bigone (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * context, b
   /* Proceed with the update. the new record is prepared and for mvcc it should have the prev version lsa set */
   if (heap_is_big_length (context->recdes_p->length))
     {
+      log_append_empty_record (thread_p, LOG_START_ATOMIC_REPL, NULL);
```
wrlawodms, line 22041: Is it for RVHF_SET_PREV_VERSION_LSA?
I don't remember exactly, but the commit-msg is saying that we don't need it in this case.
> REC_BIGONE -> * : The BIGONE redo record already has the previous version lsa, so we don't need to care about this case.

Please tell me why we must pack it with the atomic operation.

wrlawodms: It seems to me that the record already contains the previous version lsa in 22034L.
`or_mvcc_set_log_lsa_to_record (context->recdes_p, logtb_find_current_tran_lsa (thread_p));`

src/storage/heap_file.c
```diff
@@ -22586,6 +22551,17 @@ heap_update_home (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONTEXT * context, boo
...
+   // To ensure that, if started, the atomic replication area will also end.
+   scope_exit <std::function<void (void)>> log_on_exit ([&thread_p, &atomic_replication_flag]()
+   {
+     if (atomic_replication_flag == true)
+      {
+         log_append_empty_record (thread_p, LOG_END_ATOMIC_REPL, NULL);
+      }
+   });
+   // *INDENT-ON*
+
```
wrlawodms, line 22564: 1. LOG_START_ATOMIC_REPL is never left in heap_update_home().
2. I think we must leave them in some cases. (probably whenever updating a home record..)

wrlawodms:
@joohok Please make sure to use atomic op appropriately based on the commit-msg from https://github.com/CUBRID/cubrid/commit/9cd23bcc8f25378a158b2308bf1fc361d66ddfa1.

>REC_HOME -> * : For any case, it's enough to set prev_version_lsa and update the home record atomically.
REC_RELOCATION:
-> REC_HOME, REC_BIGONE, and REC_RELOCATION on the same page: The home record is updated, and it has to be applied atomically with setting prev version lsa. Actually, for all cases, LOG_START_ATOMIC_REPL is already appended, but I have added it conditionally for safety.
-> REC_RELOCATION on another page: The home record is not updated, so the forward record has to be updated atomically with setting prev version lsa.
REC_BIGONE -> * : The BIGONE redo record already has the previous version lsa, so we don't need to care about this case.

src/storage/heap_file.c
wrlawodms: Oh, it's for the case in which updating a heap record must be atomic! Sorry, my bad.

src/storage/heap_file.c
wrlawodms: Oh, it is left when `is_mvcc_op` is true. My bad, again, sorry.

wrlawodms: I guess we can make change to 22711L
`if (!atomic_replication_flag && is_mvcc_op)`
->
`if (is_mvcc_op)`
And I think we can get rid of `atomic_replication_flag` and just use only `is_mvcc_op`. It's up to you.

wrlawodms (approved):

cristiarg:
Please update the PR description (and the commit description) to state:
- from `was the mechanism` to `is the mechanism`; it is still used in some cases :)
- make a synthesis of all the remaining situations where the mechanism is still used:
  - `heap_update_bigone`
  - `heap_update_relocation`
  - `heap_update_home` with MVCC

Thanks

src/storage/heap_file.c
cristiarg: My opinion is to keep the `_flag` - it's the same elsewhere.
It's just more explicit even if redundant.

Nit.
Using only the `is_mvcc_op` argument is actually prone to bugs.
Since `bool is_mvcc_op` is not `const`, it might be modified within the function (even if accidentally) - so, by the end of the function, when it would be used in the lambda to add the closing log record it's value might not be the same.
I would make that argument `const` but this is outside the scope of the PR.

cristiarg (approved):

hornetmj (approved):

PR #3937, opened 2022-11-10
[CBRD-24554] incorrect result of "NOT EXISTS" and "NOT IN" with sub-query returning null

beyondykk9:
http://jira.cubrid.org/browse/CBRD-24554

The constant folding in query rewriter makes sub-query which returns always "false" to "null". However, the "NOT EXISTS" or "NOT IN" operation in front of the sub-query returns incorrect result by the "null".
```
CREATE TABLE t1 (a INT);
INSERT INTO t1 VALUES (1);
SELECT * FROM t1 WHERE NOT EXISTS (SELECT 1 FROM dual WHERE 'a' = 'b');
SELECT * FROM t1 WHERE 1 NOT IN (SELECT 1 FROM dual WHERE 'a' = 'b');
```
The both of two above queries should be result 1.


src/parser/parse_tree.h
```diff
@@ -3665,6 +3665,7 @@ struct parser_node
     unsigned is_system_generated_stmt:1;	/* is internally generated by system */
     unsigned use_auto_commit:1;	/* use autocommit */
     unsigned done_reduce_equality_terms:1;	/* reduce_equality_terms() is already called */
+    unsigned is_false_where:1;	/* reduced node from false where */
```
shparkcubrid: Adding it to pt_value_info looks better.

ctshim (approved):

src/parser/parse_tree.h
beyondykk9: I fixed as you commented.

shparkcubrid (approved):

hgryoo (approved):

hyunikn (approved):

youngjinj (approved):

PR #3938, opened 2022-11-10
[LETS-522] replace send_int and recv_int in server-server protocol

cristiarg:
http://jira.cubrid.org/browse/LETS-522

Revised the way the command mnemonic is sent to the peer server in the server-server communication protocol.
Trying to fix the situations where a mere `::send` was failing upon first try.
The functions `send_int` and `recv_int` were using a simple `::send` with an all-or-nothing approach and sometimes, especially under heavy load and when server are executing on different physical machines, the "nothing" happened.

Other:
- fixed some compiler warnings

### DIFF
--- a/src/communication/communication_channel.cpp
+++ b/src/communication/communication_channel.cpp
@@ -135,16 +135,15 @@ namespace cubcomm
 
   css_error_code channel::recv_int (int &received)
   {
-    size_t len = sizeof (received);
+    const size_t len = sizeof (received);
 
-    size_t readlen = (size_t) css_readn (m_socket, reinterpret_cast<char *> (&received), (int) len,
-					 m_max_timeout_in_ms);
+    const int readlen = css_readn (m_socket, reinterpret_cast<char *> (&received), (int) len, m_max_timeout_in_ms);
     css_error_code error = NO_ERRORS;
     if (readlen < 0)
       {
 	error = css_error_code::ERROR_ON_COMMAND_READ;
       }
-    else if (readlen != len)
+    else if ((size_t) readlen != len)
       {
 	error = css_error_code::READ_LENGTH_MISMATCH;
       }
--- a/src/communication/communication_channel.hpp
+++ b/src/communication/communication_channel.hpp
@@ -71,6 +71,9 @@ namespace cubcomm
       css_error_code recv (char *buffer, std::size_t &maxlen_in_recvlen_out);
       css_error_code send (const std::string &message);
       css_error_code send (const char *buffer, std::size_t length);
+
+      // WARN: these functions use an all-or-nothing approach; use only if the higher level logic
+      // is able to recover from such a failure (important especially under load)
       css_error_code send_int (int val);
       css_error_code recv_int (int &received);
 
--- a/src/communication/communication_channel.hpp
+++ b/src/communication/communication_channel.hpp
@@ -55,6 +55,7 @@ namespace cubcomm
       constexpr static int INVALID_PORT = -1;
 
     public:
+      channel () = delete;
       channel (int max_timeout_in_ms);
       channel (std::string &&channel_name);
       channel (int max_timeout_in_ms, std::string &&channel_name);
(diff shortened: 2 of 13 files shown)


src/communication/request_client_server.hpp
```diff
@@ -337,16 +337,18 @@ namespace cubcomm
   css_error_code request_server<MsgId>::receive_request_buffer (std::unique_ptr<char[]> &message_buffer,
       size_t &message_size)
   {
-    int ilen = 0;
-    css_error_code err = m_channel.recv_int (ilen);
+    size_t expected_size = 0;
+    std::size_t size_ilen = sizeof (expected_size);
```
joohok: Nit:
I think there is no difference between std::size_t and size_t. How about unifying it as size_t?

joohok (approved):

src/communication/communication_channel.hpp
```diff
@@ -71,6 +71,9 @@ namespace cubcomm
       css_error_code recv (char *buffer, std::size_t &maxlen_in_recvlen_out);
       css_error_code send (const std::string &message);
       css_error_code send (const char *buffer, std::size_t length);
+
+      // WARN: these functions use an all-or-nothing approach; use only if the higher level logic
+      // is able to recover from such a failure (important especially under load)
```
razvanradulescu: These functions (send_int at least) is used in other contexts (connect_to_page_server).
The handling there seems to be escalation (do we retry at higher level in such case ?).

What are the cons with the other approach (to refactor send_int, recv_int to use safe buffer send and recv) ?

cristiarg: @razvanradulescu apparently this does not solve our issue; even with this approach the recurring crash experienced in tests ... re-occurs; still pondering as to the actual cause;
actually I'm open to refactor `send_int`/`recv_int`.

cristiarg:
Since this patch does not fix the actual proble, for now, I will move this PR to draft.

src/communication/communication_channel.hpp
cristiarg: @razvanradulescu I tried to refactor the `send_int` and `recv_int` in terms of re-using the `send` and `recv` member methods but it is not possible.
Because they are not used symmetrically everywhere.
While `send_int` is used in initiating a connection via `cub_master`, `recv_int` is not used on the receiving side.
And implementing the `send/recv_int` by reusing `send/recv` changes what is sent over the network as protocol.
The most I could do is add new `send_int_ex`/`recv_int_ex` versions of the methods based on `send` and `recv`, respectively, but I think that is superfluous.
Aside from their use in the LETS project - which I am now replacing with `send`, `recv` - the rest of the uses are in "handshake" conversations/connection set-up scenarios where it is unlikely that they will ever fail.

Perhaps it is best to add a CBRD issue to address this concern.

src/communication/communication_channel.cpp
```diff
@@ -135,16 +135,15 @@ namespace cubcomm
 
   css_error_code channel::recv_int (int &received)
   {
-    size_t len = sizeof (received);
+    const size_t len = sizeof (received);
 
-    size_t readlen = (size_t) css_readn (m_socket, reinterpret_cast<char *> (&received), (int) len,
-					 m_max_timeout_in_ms);
+    const int readlen = css_readn (m_socket, reinterpret_cast<char *> (&received), (int) len, m_max_timeout_in_ms);
```
cristiarg, line 140: fixed compiler warning of sign mismatch

src/communication/communication_server_channel.cpp
```diff
@@ -77,7 +77,7 @@ namespace cubcomm
      * character to the numer related to the server type enum value, the rest of the
      * buffer space is used to copy the server name.
      */
-    std::string msg { (char) m_server_type + '0' };
+    std::string msg { (char) (m_server_type + '0') };
```
cristiarg, line 80: fixed compiler warning

src/communication/request_client_server.hpp
```diff
@@ -151,7 +151,7 @@ namespace cubcomm
       inline const channel &get_channel () const;		// get underlying channel
 
     private:
-      channel m_channel;					// requests are sent on this channel
+      channel m_send_channel;				// requests are sent on this channel
```
cristiarg: two classes in the same header file had the same member `m_channel`

cristiarg:
> Since this patch does not fix the actual proble, for now, I will move this PR to draft.

Further testing revealed that this patch fixes the problem. Submitting for review.

wrlawodms (approved):

src/communication/request_client_server.hpp
cristiarg: Done

razvanradulescu (approved):

PR #3942, opened 2022-11-11
[CBRD-24502] The problem of updating statistical information when performing DDL.

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-24502

It is modified so that statistics are not updated during DDL. DDL includes index creation and drop, such as 'CREATE INDEX'.
The optimizer sets the default selectivity for new statistics that are not updated.

### DIFF
--- a/src/object/schema_manager.c
+++ b/src/object/schema_manager.c
@@ -4314,7 +4314,7 @@ sm_update_statistics (MOP classop, bool with_fullscan)
  *
  * NOTE: We will delay updating statistics until a transaction is committed
  *       when it is requested during other processing, such as
- *       "alter table ..." or "create index ...".
+ *       "alter table ...." or "create index ...".
  */
 int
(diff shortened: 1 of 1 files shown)


src/object/schema_manager.c
```diff
@@ -4305,6 +4305,76 @@ sm_update_statistics (MOP classop, bool with_fullscan)
...
+/*
+ * sm_update_statistics_without_gathering_stats () - fetch the statistics and
+ *    						     cache them with the class.
+ *   return: NO_ERROR on success, non-zero for ERROR
+ *   classop(in): class object
+ *   with_fullscan(in): true iff WITH FULLSCAN
+ *
+ * NOTE: We will delay updating statistics until a transaction is committed
+ *       when it is requested during other processing, such as
+ *       "alter table ..." or "create index ...".
```
shparkcubrid:
```suggestion
 *       "alter table ...." or "create index ...".
```

PR #3944, opened 2022-11-11
[CBRD-24555] The spage_collect_statistics function should consider records deleted by vacuum when calculating the number of records in a data page

hornetmj:
http://jira.cubrid.org/browse/CBRD-24555

**Description**
- handled the requirements not to count vacuumed records in the spage_collect_statistics function
- however, vacuumed records are not already being counted
```
// spage_vacuum_slot()
slot_p->offset_to_record = SPAGE_EMPTY_OFFSET;

// spage_collect_statistics()
if (slot_p->offset_to_record == SPAGE_EMPTY_OFFSET)
  {
    continue;
  }
```
- but, I modified it for the following reasons
  - more intuitive to use slot_p->record_type instead of slot_p->offset_to_record in the spage_collect_statistics function
  - the spage_mark_deleted_slot_as_reusable function does not set slot_p->offset_to_record to SPAGE_EMPTY_OFFSET
  - defensive code by quick fix - leave 'if (slot_p->offset_to_record == SPAGE_EMPTY_OFFSET)' as it is

### DIFF
--- a/src/storage/slotted_page.c
+++ b/src/storage/slotted_page.c
@@ -1033,11 +1033,8 @@ spage_collect_statistics (PAGE_PTR page_p, int *npages, int *nrecords, int *rec_
     {
       if (slot_p->offset_to_record == SPAGE_EMPTY_OFFSET)
 	{
-	  continue;
-	}
+	  assert (slot_p->record_type == REC_MARKDELETED || slot_p->record_type == REC_DELETED_WILL_REUSE);
 
-      if (slot_p->record_type == REC_MARKDELETED || slot_p->record_type == REC_DELETED_WILL_REUSE)
-	{
 	  continue;
 	}
 
--- a/src/storage/slotted_page.c
+++ b/src/storage/slotted_page.c
@@ -1036,6 +1036,11 @@ spage_collect_statistics (PAGE_PTR page_p, int *npages, int *nrecords, int *rec_
 	  continue;
 	}
 
+      if (slot_p->record_type == REC_MARKDELETED || slot_p->record_type == REC_DELETED_WILL_REUSE)
+	{
+	  continue;
+	}
+
       if (slot_p->record_type == REC_BIGONE)
 	{
 	  pages += 2;


src/storage/slotted_page.c
```diff
@@ -1036,6 +1036,11 @@ spage_collect_statistics (PAGE_PTR page_p, int *npages, int *nrecords, int *rec_
 	  continue;
 	}
 
+      if (slot_p->record_type == REC_MARKDELETED || slot_p->record_type == REC_DELETED_WILL_REUSE)
+	{
+	  continue;
+	}
+
```
wrlawodms: How about this and removing `if (slot_p->offset_to_record == SPAGE_EMPTY_OFFSET)`?
```suggestion
      if (slot_p->record_type == REC_MARKDELETED || slot_p->record_type == REC_DELETED_WILL_REUSE)
	{
	  assert (slot_p->offset_to_record == SPAGE_EMPTY_OFFSET);
	  continue;
	}

```

wrlawodms: > the spage_mark_deleted_slot_as_reusable function does not set slot_p->offset_to_record to SPAGE_EMPTY_OFFSET

Even `spage_mark_deleted_slot_as_reusable()` requires that `slot_p->offset_to_record` is set to `SPAGE_EMPTY_OFFSET`.

wrlawodms: It's in terms of one of the reasons, readability. I don't understand `quick fix`. What is fixed?

hornetmj: I called it a quick fix because it applied only the if condition necessary for the blocked issue. I think there is no problem on my fix when I checked it in the limited time. However I think I doesn't consider all cases enough.

src/storage/slotted_page.c
```diff
@@ -1036,6 +1036,11 @@ spage_collect_statistics (PAGE_PTR page_p, int *npages, int *nrecords, int *rec_
 	  continue;
 	}
 
+      if (slot_p->record_type == REC_MARKDELETED || slot_p->record_type == REC_DELETED_WILL_REUSE)
```
joohok: How about removing `if (slot_p->offset_to_record == SPAGE_EMPTY_OFFSET)` at 1034L, and adding the REC_UNKNOWN to the condition ? 
```suggestion
      if (slot_p->record_type == REC_MARKDELETED || slot_p->record_type == REC_DELETED_WILL_REUSE || slot_p->record_type == REC_UNKNOWN)
```

After inserting a record and before determining the type of the record, the record type can be set to REC_UNKNOWN. REC_UNKNOWN can be detected because it is not latched against spage.
Probably, it seems that it is protected by a lock instead of a latch, so it won't prevent the access of the thread trying to check the statistics.
Below is the call graph example, and REC_UNKNOWN is set in spage_add_new_slot ()

```
heap_insert_logical 
-> heap_get_insert_location_with_lock
-> heap_insert_physical
--> spage_insert_at
---> spage_find_empty_slot_at
----> spage_add_new_slot 
```

wrlawodms: It seems to be latched in every case to me. Is there any way to get a page pointer without pgbuf_fix()? In the case you mentioned, a page is fixed either heap_scan_pb_lock_and_fetch() or heap_vpid_alloc(). or hint_p is given.

joohok: Oh.. yes you are right. then there is no need to check the REC_UNKNOWN type

src/storage/slotted_page.c
hornetmj: About spage_mark_deleted_slot_as_reusable() (re-visit)

The slot_p->offset_to_record should be the SPAGE_EMPTY_OFFSET when this function is called. So, I think it is no need to set the slot_p->offset_to_record to the SPAGE_EMPTY_OFFSET again. The purpose of this function is to change the value of the slot_p->record_type to REC_DELETED_WILL_REUSE.

src/storage/slotted_page.c
hornetmj: @wrlawodms You mean that a transaction(worker thread) never see the record with the REC_UNKNOWN type because it handle under the single page latch cycle. Is it right?

hornetmj: @joohok @wrlawodms I was concerned about one situation where slot_p->record_type could remain REC_UNKNOWN, but there doesn't appear to be such a case in the code.

* worker thread: set slot_p->record_type to REC_UNKNOWN
  * ex) spage_shift_slot_up(), spage_shift_slot_down(), spage_add_new_slot()
* worker thread: fail and return a error before changing the slot_p->record_type to the valid type such as REC_HOME. So, the slot_p->record_type remains as the REC_UNKNOWN
  * this case is currently confirmed to be non-existent. But, adding REC_UNKNOWN in advance seems like a good option to me 

How about modifying it like this:

```
if (slot_p->offset_to_record == SPAGE_EMPTY_OFFSET)
{
    assert (slot_p->record_type == REC_MARKDELETED || slot_p->record_type == REC_DELETED_WILL_REUSE || slot_p->record_type == REC_UNKNOWN)
    // REC_UNKNOWN will be excluded or included depending on your opinion. It doesn't seem necessary at this point
    ...
}
```

joohok: > more intuitive to use slot_p->record_type instead of slot_p->offset_to_record in the spage_collect_statistics function

For the description above, I think it would be better to remove `if (slot_p->offset_to_record == SPAGE_EMPTY_OFFSET)`. 
But, Whether it is checked using `offset` or `record type`, there seems to be no problem in terms of functionality.

wrlawodms: > @wrlawodms You mean that a transaction(worker thread) never see the record with the REC_UNKNOWN type because it handle under the single page latch cycle. Is it right?

I'm not sure about every case since I didn't look into every single case. What I said was only for the case Jooho mentioned.

> worker thread: fail and return a error before changing the slot_p->record_type to the valid type such as REC_HOME. So, the slot_p->record_type remains as the REC_UNKNOWN

I think that other metadata such as `page_header_p->num_slots` has to be reset so that the `REC_UNKNOWN` must not be accessed in such cases.

> worker thread: set slot_p->record_type to REC_UNKNOWN
ex) spage_shift_slot_up(), spage_shift_slot_down(), spage_add_new_slot()

In these cases, the `REC_UNKNOWN` can be accessed? then, I agree we have to cover the case.

It seems to me that it's a kind of 'lazy' way to handle `REC_UNKNOWN` without necessity, which could confuse readers. If you're worried, how about this:
```
assert (slot_p->record_type != REC_UNKNOWN);
if (slot_p->offset_to_record == SPAGE_EMPTY_OFFSET)
{
    assert (slot_p->record_type == REC_MARKDELETED || slot_p->record_type == REC_DELETED_WILL_REUSE)
    // REC_UNKNOWN will be excluded or included depending on your opinion. It doesn't seem necessary at this point
    ...
}
```
but I prefer the below in terms of readability that you mentioned in the description:
> more intuitive to use slot_p->record_type instead of slot_p->offset_to_record in the spage_collect_statistics function

```
assert (slot_p->record_type != REC_UNKNOWN);
 if (slot_p->record_type == REC_MARKDELETED || slot_p->record_type == REC_DELETED_WILL_REUSE)
{
    assert (slot_p->offset_to_record == SPAGE_EMPTY_OFFSET);
}
```

However, in the terms of minimizing changes, the first one is also a good option. It's up to you.

hornetmj: @wrlawodms I want to minimize changes/affections of codes. So, I choose the first option(similar to the my suggestion). However the assert condition 'slot_p->record_type != REC_UNKNOWN' can be covered by the condition check of if(if the slot_p->record_type is REC_UNKNOWN then the slot_p->offset_to_record is SPAGE_EMPTY_OFFSET) and second assert. So, I won't add it.

@joohok I understand your feedback. However, I think I can convey enough information with the assert to be added.

Thanks all.

joohok (approved):

wrlawodms (approved):

PR #3949, opened 2022-11-15
[CBRD-24543] remove compat directory from the CUBRID distribution

kisoo-han:
http://jira.cubrid.org/browse/CBRD-24543

**Purpose**
* **Remove** the $CUBRID/**compat** directory and $CUBRID/compat/* files **from the CUBRID distribution**

**Implementation**
N/A

**Remarks**
* The '**cubrid**' front-end is sufficient to do same work for the binaries in compat/*.
  * cubrid unloaddb ... (compat/unloaddb)
  * cubrid compatdb ... (compat/compactdb)
* If a user includes $CUBRID/compat in his/her command path, 'cubrid unloaddb' can be used as an abbreviation in the form of 'unloaddb'.
* However, this can confuse users.

### DIFF
--- a/util/CMakeLists.txt
+++ b/util/CMakeLists.txt
@@ -344,33 +344,6 @@ if(UNIX)
   target_link_libraries(unittests_bit LINK_PRIVATE cubrid)
 endif(UNIX)
 
-
-# compat utils
-if(UNIX)
-  SET_SOURCE_FILES_PROPERTIES(
-    ${EXECUTABLES_DIR}/util_front.c
-    PROPERTIES LANGUAGE CXX
-  )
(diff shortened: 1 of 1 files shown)


airnet73:
Codes related to COMPAT also exist in cubrid/CMakeLists.txt. (COMPAT_DIR, CUBRID_COMPATDIR, etc.)
Shouldn't this code be deleted too?
What do you think?

kisoo-han:
I think it is not related with $CUBRID/compat/* in CUBRID Distribution.
It is a set of DB APIs used by Engine (defined in cubrid/CMakeLists.txt)
We just want remove $CUBRID/compat directory in the Distribution.

mhoh3963 (approved):

airnet73 (approved):

PR #3953, opened 2022-11-15
[LETS-584] Block ha daemon if there is no need to calculate the replication delay

joohok:
http://jira.cubrid.org/browse/LETS-584

Purpose

`log_Check_ha_delay_info_daemon` is a daemon that continuously appends dummy log records every second.
Dummy log records are used only for calculating replication delay by LETS. 
Therefore, This is to prevent the daemon from running when there is no need to calculate the replication delay time.

### DIFF
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -10889,9 +10889,13 @@ log_clock_daemon_init ()
 void
 log_check_ha_delay_info_daemon_init ()
 {
-  bool do_supplemental_log = prm_get_integer_value (PRM_ID_SUPPLEMENTAL_LOG) > 0 ? true : false;
+  const bool do_supplemental_log = prm_get_integer_value (PRM_ID_SUPPLEMENTAL_LOG) > 0 ? true : false;
+  const bool do_calc_replication_delay = prm_get_bool_value (PRM_ID_ER_LOG_CALC_REPL_DELAY);
+  const bool is_transaction_server = (get_server_type () == SERVER_TYPE_TRANSACTION);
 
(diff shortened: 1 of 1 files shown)


cristiarg (approved):

src/transaction/log_manager.c
```diff
@@ -10889,9 +10889,13 @@ log_clock_daemon_init ()
 void
 log_check_ha_delay_info_daemon_init ()
 {
-  bool do_supplemental_log = prm_get_integer_value (PRM_ID_SUPPLEMENTAL_LOG) > 0 ? true : false;
+  const bool do_supplemental_log = prm_get_integer_value (PRM_ID_SUPPLEMENTAL_LOG) > 0 ? true : false;
+  const bool do_calc_replication_delay = prm_get_bool_value (PRM_ID_ER_LOG_CALC_REPL_DELAY);
+  const bool is_transaction_server = (get_server_type () == SERVER_TYPE_TRANSACTION);
```
```suggestion
  const bool is_transaction_server = is_transaction_server();
```

wrlawodms: or `is_active_transaction_server()` though it is only called when it's not PTS?

wrlawodms (approved):

src/transaction/log_manager.c
joohok: It will be better to use `is_active_transaction_server()` for more readability. 
I used `(get_server_type () == SERVER_TYPE_TRANSACTION)` just because caller checks whether it is passive transaction server or not.

hornetmj (approved):

PR #3954, opened 2022-11-15
[LETS-595] overflow heap page atomic sequence

cristiarg:
http://jira.cubrid.org/browse/LETS-595

Passive Transaction Server atomic replication sequence for page overflow allocation scenarios.

Avoid purging an atomic sequence by an `LOG_SYSOP_END` (with `LOG_SYSOP_END_LOGICAL_UNDO`) when that sequence was not started by a `LOG_SYSOP_ATOMIC_START` (see log sequence synopsis in the comment for `class atomic_replication_helper`).

Also drop the condition regarding `lastparent_lsa` to fit both the cases when it has a valid value and when it has a null value. This is needed to accommodate the situation when `LOG_SYSOP_ATOMIC_START` log record that starts the atomic sequence coincides with the start of the transaction (is the very first log record of a newly occurring transaction in the log).

### DIFF
--- a/src/storage/page_buffer.c
+++ b/src/storage/page_buffer.c
@@ -1881,9 +1881,7 @@ pgbuf_fix_release (THREAD_ENTRY * thread_p, const VPID * vpid, PAGE_FETCH_MODE f
     {
       /* Make sure that the page has been allocated (i.e., is a valid page) */
       /* Suppress errors if fetch mode is OLD_PAGE_IF_IN_BUFFER. */
-      const bool no_error_when = (fetch_mode == OLD_PAGE_IF_IN_BUFFER) ||
-	(is_passive_transaction_server () && (fetch_mode == OLD_PAGE_MAYBE_DEALLOCATED));
-      if (pgbuf_is_valid_page (thread_p, vpid, no_error_when, NULL, NULL) != DISK_VALID)
+      if (pgbuf_is_valid_page (thread_p, vpid, fetch_mode == OLD_PAGE_IF_IN_BUFFER, NULL, NULL) != DISK_VALID)
 	{
 	  return NULL;
 	}
--- a/src/storage/page_buffer.c
+++ b/src/storage/page_buffer.c
@@ -1881,7 +1881,9 @@ pgbuf_fix_release (THREAD_ENTRY * thread_p, const VPID * vpid, PAGE_FETCH_MODE f
     {
       /* Make sure that the page has been allocated (i.e., is a valid page) */
       /* Suppress errors if fetch mode is OLD_PAGE_IF_IN_BUFFER. */
-      if (pgbuf_is_valid_page (thread_p, vpid, fetch_mode == OLD_PAGE_IF_IN_BUFFER, NULL, NULL) != DISK_VALID)
+      const bool no_error_when = (fetch_mode == OLD_PAGE_IF_IN_BUFFER) ||
+	(is_passive_transaction_server () && (fetch_mode == OLD_PAGE_MAYBE_DEALLOCATED));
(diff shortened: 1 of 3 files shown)


joohok (approved):

src/storage/page_buffer.c
```diff
@@ -1881,7 +1881,9 @@ pgbuf_fix_release (THREAD_ENTRY * thread_p, const VPID * vpid, PAGE_FETCH_MODE f
     {
       /* Make sure that the page has been allocated (i.e., is a valid page) */
       /* Suppress errors if fetch mode is OLD_PAGE_IF_IN_BUFFER. */
-      if (pgbuf_is_valid_page (thread_p, vpid, fetch_mode == OLD_PAGE_IF_IN_BUFFER, NULL, NULL) != DISK_VALID)
+      const bool no_error_when = (fetch_mode == OLD_PAGE_IF_IN_BUFFER) ||
+	(is_passive_transaction_server () && (fetch_mode == OLD_PAGE_MAYBE_DEALLOCATED));
+      if (pgbuf_is_valid_page (thread_p, vpid, no_error_when, NULL, NULL) != DISK_VALID)
```
wrlawodms: Does it mean the sector including the page is possibly not reserved when `(is_passive_transaction_server () && (fetch_mode == OLD_PAGE_MAYBE_DEALLOCATED`?

wrlawodms (approved):

hornetmj (approved):
@cristiarg It would be better if the jira issue for this PR, not the QA jira issue, were separated individually. This will make it easier to track issues later.

src/transaction/atomic_replication_helper.cpp
```diff
@@ -649,25 +655,27 @@ namespace cublog
...
 		 && LOG_SYSOP_END_LOGICAL_UNDO == last_entry.m_sysop_end_type)
 	  {
 	    const atomic_log_entry_vector_type::const_iterator last_but_one_entry_it = (last_entry_it - 1);
 	    const atomic_log_entry &last_but_one_entry = *last_but_one_entry_it;
 
-	    if (!LSA_ISNULL (&last_entry.m_sysop_end_last_parent_lsa) &&
-		(last_but_one_entry.m_lsa >= last_entry.m_sysop_end_last_parent_lsa))
+	    if (last_but_one_entry.m_rectype == LOG_SYSOP_ATOMIC_START) //&&
+	      // !LSA_ISNULL (&last_entry.m_sysop_end_last_parent_lsa) &&
+	      // (last_but_one_entry.m_lsa >= last_entry.m_sysop_end_last_parent_lsa))
```
joohok: is this change related to #3955 (which is now closed) ?

cristiarg: In a way, yes.
In absence of the patch in #3955 I decided to drop the `lastparent_lsa` criterium. Thus the two encountered situations where the `lastparent_lsa` either is `(-1|-1)` or a valid value will be satisfied by the remaining condition.

cristiarg:
> @cristiarg It would be better if the jira issue for this PR, not the QA jira issue, were separated individually. This will make it easier to track issues later.

@hornetmj 
I re-directed this PR as a fix for newly added specific issue http://jira.cubrid.org/browse/LETS-595 .
I hope I understood correctly the intention of your comment.

src/storage/page_buffer.c
cristiarg: @wrlawodms
I decided to drop this change from this PR because, as discussed, `OLD_PAGE_MAYBE_DEALLOCATED` fetch mode will be replaced by `OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT` by the patch #3931 .
We'll take it up separately if/when needed.

PR #3955, opened 2022-11-15
[LETS-511] properly fill sysop end lastparent_lsa

cristiarg:

In the situations where the `LOG_SYSOP_ATOMIC_START` log record is the very first transactional log record to be added by a transaction, the corresponding closing `LOG_SYSOP_END` log record's `lastparent_lsa` does not point properly to the starting `LOG_SYSOP_ATOMIC_START` log record lsa.
This patch fixes that precise scenario.

### DIFF
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -3965,6 +3965,9 @@ log_sysop_start_internal (THREAD_ENTRY * thread_p, LOG_TDES * tdes)
   tdes->topops.last++;
   LSA_COPY (&tdes->topops.stack[tdes->topops.last].lastparent_lsa, &tdes->tail_lsa);
   LSA_COPY (&tdes->topop_lsa, &tdes->tail_lsa);
+  /* if this newly added sysop is the very first log record added for this transaction, the
+   * tail_lsa will be empty; the tail lsa will be copied again after the sysop log record is
+   * added and tail lsa is filled in */
 
   LSA_SET_NULL (&tdes->topops.stack[tdes->topops.last].posp_lsa);
 
@@ -4009,6 +4012,16 @@ log_sysop_start_atomic (THREAD_ENTRY * thread_p)
 	}
 
       (void) prior_lsa_next_record (thread_p, node, tdes);
+
+      /* in case this sysop is the very first log record of this transaction, the descriptor's tail lsa was empty;
+       * copy again the sysop_start's lsa to be filled as the sysop_end's lastparent_lsa */
+      if (LSA_ISNULL (&tdes->topops.stack[tdes->topops.last].lastparent_lsa))
+	{
+	  assert (LSA_ISNULL (&tdes->topop_lsa));
+	  assert (!LSA_ISNULL (&tdes->tail_lsa));
+	  LSA_COPY (&tdes->topops.stack[tdes->topops.last].lastparent_lsa, &tdes->tail_lsa);
+	  LSA_COPY (&tdes->topop_lsa, &tdes->tail_lsa);
+	}
     }
   else
     {


src/transaction/log_manager.c
```diff
@@ -4009,6 +4012,16 @@ log_sysop_start_atomic (THREAD_ENTRY * thread_p)
...
       (void) prior_lsa_next_record (thread_p, node, tdes);
+
+      /* in case this sysop is the very first log record of this transaction, the descriptor's tail lsa was empty;
+       * copy again the sysop_start's lsa to be filled as the sysop_end's lastparent_lsa */
+      if (LSA_ISNULL (&tdes->topops.stack[tdes->topops.last].lastparent_lsa))
+	{
+	  assert (LSA_ISNULL (&tdes->topop_lsa));
+	  assert (!LSA_ISNULL (&tdes->tail_lsa));
+	  LSA_COPY (&tdes->topops.stack[tdes->topops.last].lastparent_lsa, &tdes->tail_lsa);
+	  LSA_COPY (&tdes->topop_lsa, &tdes->tail_lsa);
```
wrlawodms, line 4023: They are supposed to point to the last LSA before LOG_SYSOP_ATOMIC_START and some different operations use them with the assumption they do, for example, to get an abort point (`log_sysop_abort()`) and to get a skip range during postpone (`log_get_next_nested_top()`).
Actually, it seems that the above cases don't matter since the `LOG_SYSOP_ATOMIC_START` has no effect on the operations, but I'm worried about changing the meaning of them even only for the atomic sysop case.  I don't look into every case they are used, but I think some operations may misbehave or confuse readers at least. I hope to keep it simple.

wrlawodms: Could you tell me the reason why the lastparent_lsa has to point to the start of a newly created sysop while they are supposed to point to its parent? In the case there is no parent, it seems to make sense to point to NULL.

cristiarg: > Actually, it seems that the above cases don't matter since the LOG_SYSOP_ATOMIC_START has no effect on the operations

That was my reasoning.

> but I'm worried about changing the meaning of them even only for the atomic sysop case

Same concern from my point of view.

> Could you tell me the reason why the lastparent_lsa has to point to the start of a newly created sysop while they are supposed to point to its parent?

Because I am trying to accommodate all the encountered cases of atomic sysops. There are so many permutations and edge cases, it seems neverending. With this change, I was able to fit a newly found case in an already existing one.

Actually, the complication comes from the "optimization" in `log_sysop_start_atomic`: 
```
      /* This must be a nested atomic system operation. If parent is atomic, we'll be atomic too.
       * But, a new sysop atomic start is not added.
```

cristiarg: I looked into `log_sysop_abort` (and the called `log_rollback`).
As before (without this patch), if the sysop is the first that starts the transaction, as per the intention of this patch, it would rollback all log records until the start of the transaction.
With this patch applied, it would rollback the same but without the `LOG_SYSOP_ATOMIC_START` log record - the very first log record of the transaction.
It seems that, in practice, the effect is the same - as in, the same "effective" log records are undone. However, I don't know what is happens with the remaining `LOG_SYSOP_ATOMIC_START` log record of the transaction and what this entails.

cristiarg: Also looked into `log_get_next_nested_top` and it is only used in the context of finding sysops within the postpone regions of the transactions. This implies that its not the very first log record of a transaction.

cristiarg: An alternative to this patch is:
- for PTS atomic replication, we accept that there are transactions for which there are "loose end" atomic replication sequences
- ie: some atomic replication sequences could not reliably be "purged" at the correct time
- we will purge these atomic replication sequences at either `LOG_COMMIT` or `LOG_ABORT` of the transaction (actually, at `LOG_ABORT` this is already done).

However the correctness is maintained as the purging at commit will only remove atomic sequence leftovers containing only "control" log records, all side-effects log records would have been applied at the appropriate time.


However, this means that we loose the determinism of the atomic replication sequences that we currently strive to maintain - ie: we know where it starts and know precisely where it should end and be purged.

wrlawodms: First of all, I still don't understand why the lastparent_lsa of a sysop end must not be NULL when there is no parent. When does it cause a problem?

cristiarg: @wrlawodms 
The reason for this change is explained with the comments in #3954 .

It seems like a correct fix from following points of view:
- the fact that I have encountered other cases in which the `lastparent_lsa` is set to the actual start of the sequence (`LOG_SYSOP_START` or `LOG_SYSOP_ATOMIC_START`)
- and the fact that it looks obvious that, when the start of the atomic sysop coincides with the very first log record of the transaction, the `lastparent_lsa` is missing
- it allows me to nicely fit a new occurring pattern of atomic sequence into an existing one (again see peer patch #3954).

However, I tested further and I think it might work even without this patch.
At least for the scenarios for LETS-511 it will work without this patch.
We'll see if these changes are needed in further development/refining of the atomic replication on PTS.

I will close this PR and only its peer #3954 will remain for now.

Many thanks for turning a critical eye on this change. Much appreciated. :)

cristiarg:
Closing - for now - as - maybe - not needed.

PR #3957, opened 2022-11-17
[CBRD-24545] Can not create a lot of statements within one transaction

beyondykk9:
http://jira.cubrid.org/browse/CBRD-24545

Crash has occurred while making a lot of statements without closing any of them. 
This crash is caused by the constant value QMGR_MAX_QUERY_ENTRY_PER_TRAN.
This constant value seems to represent the maximum number of queries that can be performed within one transaction. However, QMGR_MAX_QUERY_ENTRY_PER_TRAN, which is hard-coded as 100, is too small to perform intensive tests.

In system parameterized value should be limited to SHRT_MAX (32767) because the query manager allocation routine assert to max short integer by static_assert.


src/query/query_manager.c
```diff
@@ -284,9 +282,9 @@ qmgr_allocate_query_entry (THREAD_ENTRY * thread_p, QMGR_TRAN_ENTRY * tran_entry
       tran_entry_p->free_query_entry_list_p = query_p->next;
       pthread_mutex_unlock (&tran_entry_p->mutex);
     }
-  else if (QMGR_MAX_QUERY_ENTRY_PER_TRAN < tran_entry_p->num_query_entries)
+  else if (qmgr_max_query_entry_per_tran < tran_entry_p->num_query_entries)
     {
-      assert (QMGR_MAX_QUERY_ENTRY_PER_TRAN >= tran_entry_p->num_query_entries);
+      assert (qmgr_max_query_entry_per_tran >= tran_entry_p->num_query_entries);
```
ctshim, line 287: It's a meaningless check.

```suggestion     
```

ctshim (approved):

src/query/query_manager.c
beyondykk9: static_assert works at compile time and QMGR_MAX_QUERY_PER_TRAN parameter is already limited to SHRT (32767).

hgryoo (approved):

src/query/query_manager.c
hyunikn: The assert is equivalent to 'assert(false)' because of the if condition in line 285. It seems inappropriate to assert false in this line because this can be reachable when the number of queries exceeds the parameter value.

ctshim: Sorry, 
I misunderstood the assert code.
Assert is a required code.

But, is same assert(false)?

shparkcubrid (approved):

src/query/query_manager.c
beyondykk9: The purpose of static_assert is compiling error.  **qmgr_max_query_entry_per_tran** value is checked not to allow grater than **SHRT_MAX**. In server booting, it will be failed at inappropriate value. Therefore, any assertion is not neccesary in this location at run time.

src/query/execute_schema.c
```diff
@@ -10502,6 +10502,7 @@ do_change_att_schema_only (PARSER_CONTEXT * parser, DB_CTMPL * ctemplate, PT_NOD
       goto exit;
     }
 
+
```
youngjinj: It is trivial.
Why was the empty line added?

src/query/query_manager.c
```diff
@@ -274,7 +272,7 @@ qmgr_allocate_query_entry (THREAD_ENTRY * thread_p, QMGR_TRAN_ENTRY * tran_entry
   int i;
   bool usable = false;
 
-  static_assert (QMGR_MAX_QUERY_ENTRY_PER_TRAN < SHRT_MAX, "Bad query entry count");
+  static int qmgr_max_query_entry_per_tran = prm_get_bool_value (PRM_ID_QMGR_MAX_QUERY_PER_TRAN);
```
youngjinj: I think prm_get_integer_value() should be used.

src/query/query_manager.c
```diff
@@ -1281,6 +1279,8 @@ xqmgr_execute_query (THREAD_ENTRY * thread_p, const XASL_ID * xasl_id_p, QUERY_I
   bool is_xasl_pinned_reference;
   bool do_not_cache = false;
 
+  static int qmgr_max_query_entry_per_tran = prm_get_bool_value (PRM_ID_QMGR_MAX_QUERY_PER_TRAN);
```
youngjinj: I think prm_get_integer_value() should be used.

src/query/query_manager.c
```diff
@@ -1776,6 +1776,8 @@ xqmgr_prepare_and_execute_query (THREAD_ENTRY * thread_p, char *xasl_stream, int
   bool saved_is_stats_on;
   bool xasl_trace;
 
+  static int qmgr_max_query_entry_per_tran = prm_get_bool_value (PRM_ID_QMGR_MAX_QUERY_PER_TRAN);
```
youngjinj: I think prm_get_integer_value() should be used.

src/query/query_manager.c
```diff
@@ -1878,7 +1880,7 @@ xqmgr_prepare_and_execute_query (THREAD_ENTRY * thread_p, char *xasl_stream, int
 
   if (query_p == NULL)
     {
-      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_QM_QENTRY_RUNOUT, 1, QMGR_MAX_QUERY_ENTRY_PER_TRAN);
+      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_QM_QENTRY_RUNOUT, 1, qmgr_max_query_entry_per_tran);
```
youngjinj, line 1883: How about setting the error in qmgr_allocate_query_entry()?

This way, we don't have to get qmgr_max_query_entry_per_tran to print the error message.
It is enough to add "assert(er_errid() != NO_ERROR);"

src/query/execute_schema.c
beyondykk9: It's just a slip. Thanks.

src/query/query_manager.c
beyondykk9: fixed. thanks.

src/query/query_manager.c
beyondykk9: fixed. thanks.

src/query/query_manager.c
beyondykk9: fixed. thanks.

src/query/query_manager.c
beyondykk9: I keep this routine for readability.

hyunikn (approved):

youngjinj (approved):

PR #3959, opened 2022-11-18
[CBRD-24564] reflect reuse-oid option in error message prining at view updating query.

beyondykk9:
http://jira.cubrid.org/browse/CBRD-24564

The updating query with a view can be faced with error when the included table in the view has reuse-oid option. However, the error message does not reflect the option as the below message.

AS-IS:
```
create table t1(c1 int);
create view v3 as select c1 from t1;

insert into v3(c1) values (1);
ERROR: Vclass dba.v3 is not updatable.

```
TO-BE:
`ERROR: Vclass dba.v3 is not updatable, because dba.t1 is reuse_oid table.`


ctshim (approved):

hyunikn (approved):

hgryoo:
Is there any case where the MSGCAT_RUNTIME_VCLASS_NOT_UPDATABE error occurs and is unrelated to the reuse_oid table?

youngjinj:
In addition to msg/en_US/cubrid.msg, other message files must also be changed.

And even if there is more than one reuse_oid table, only one table name is printed in the error message.

```
drop table if exists t1, t2;
create table t1 (c1 int, c2 int);
create table t2 (c1 int, c2 int);
insert into t1 values (1, 1), (2, 2), (3, 3);
insert into t2 values (1, 10), (2, 20), (3, 30);
create or replace view v1 as
  select t1.c1 as c1, t1.c2 as t1_c2, t2.c2 as t2_c2
  from t1, t2
  where t1.c1 = t2.c1;
update v1 set t1_c2 = t1_c2 * 10, t2_c2 = t2_c2 * 10;
```
```
ERROR: Vclass dba.v1 is not updatable, because dba.t1 table is REUSE_OID..
```

I think the purpose of this change is to make the meaning clear to users.
However, it is not clear to tell reuse_oid for only one table.

In addition, the code becomes complicated by continuously passing unnecessary arguments only for error message output.

The reason why users cannot clearly understand the error message is that they do not understand reuse_oid.
I think it's better to add a message like the one below rather than mentioning individual tables.
```
ERROR: Vclass dba.v1 is not updatable, Vclass referencing class that are reuse_oid cannot be updated.
```

src/parser/view_transform.c
```diff
@@ -11717,9 +11732,16 @@ mq_fetch_subqueries_for_update_local (PARSER_CONTEXT * parser, PT_NODE * class_,
...
 	  && (!query_cache->view_cache->vquery_for_partial_update || (fetch_as != PT_PARTIAL_SELECT)) && parser)
 	{
-	  PT_ERRORmf (parser, class_, MSGCAT_SET_PARSER_RUNTIME, MSGCAT_RUNTIME_VCLASS_NOT_UPDATABLE,
-		      /* use function to get name. class_->info.name.original is not always set. */
-		      db_get_class_name (class_object));
+	  char reuse_tbl[300] = "";
+
+	  if (reuse_oid_class_name)
+	    {
+	      sprintf (reuse_tbl, ", because %s table is REUSE_OID.", reuse_oid_class_name);
```
youngjinj: The period is redundant.
```suggestion
	      sprintf (reuse_tbl, ", because %s table is REUSE_OID", reuse_oid_class_name);
```

beyondykk9: Ok, I will fix it.

beyondykk9:
> In addition to msg/en_US/cubrid.msg, other message files must also be changed.
> 
> And even if there is more than one reuse_oid table, only one table name is printed in the error message.
> 
> ```
> drop table if exists t1, t2;
> create table t1 (c1 int, c2 int);
> create table t2 (c1 int, c2 int);
> insert into t1 values (1, 1), (2, 2), (3, 3);
> insert into t2 values (1, 10), (2, 20), (3, 30);
> create or replace view v1 as
>   select t1.c1 as c1, t1.c2 as t1_c2, t2.c2 as t2_c2
>   from t1, t2
>   where t1.c1 = t2.c1;
> update v1 set t1_c2 = t1_c2 * 10, t2_c2 = t2_c2 * 10;
> ```
> 
> ```
> ERROR: Vclass dba.v1 is not updatable, because dba.t1 table is REUSE_OID..
> ```
> 
> I think the purpose of this change is to make the meaning clear to users. However, it is not clear to tell reuse_oid for only one table.
> 
> In addition, the code becomes complicated by continuously passing unnecessary arguments only for error message output.
> 
> The reason why users cannot clearly understand the error message is that they do not understand reuse_oid. I think it's better to add a message like the one below rather than mentioning individual tables.
> 
> ```
> ERROR: Vclass dba.v1 is not updatable, Vclass referencing class that are reuse_oid cannot be updated.
> ```

The message is made from discussion enough of the issue. The reuse_oid table related to the view is minimal information for showing the reason of the fail.

And the error messages will be added as multilingual version.

beyondykk9:
> Is there any case where the MSGCAT_RUNTIME_VCLASS_NOT_UPDATABE error occurs and is unrelated to the reuse_oid table?

The unauthorized table which is not granted (SELECT ONLY) from the user is not updatable.

hgryoo:
> > Is there any case where the MSGCAT_RUNTIME_VCLASS_NOT_UPDATABE error occurs and is unrelated to the reuse_oid table?
> 
> The unauthorized table which is not granted (SELECT ONLY) from the user is not updatable.

If the same error code is specified for multiple cases, IMO, the same detailed level of error message should be displayed.

ctshim (approved):

hgryoo (approved):

shparkcubrid (approved):

youngjinj (approved):

PR #3960, opened 2022-11-18
[LETS-558] [LETS-511] completely purge the page from page buffer upon page de-alloc in atomic replication context

cristiarg:
http://jira.cubrid.org/browse/LETS-558

When a page is de-allocated - via `RVPGBUF_DEALLOC` - all that is done is to mark the page as `PAGE_UNKNOWN` and mark it dirty with the idea that:
- in a "monolithic" server, there will be no other transaction that needs access to the page given that
  - all previous transactions that needed the page concluded - guaranteed by the fact that the transaction that de-allocates the page was allowed to acquire write latch for the page
  - all subsequent transactions will not "see" the page - guaranteed by mechanisms such as MVCC
- being dirty, the page buffer flush daemon will write it to disk upon next run

However, this seems to break down in Passive Transaction Server because the page buffer flush daemon does not flush the permanent pages and remove them from the page buffer (see the mechanism implemented with the help of the flag `PGBUF_BCB_FLUSH_NOT_NEEDED_FLAG`).
The page will remain dirty in the page buffer until it is needed again - most probably until a next re-allocation.
On PTS re-allocation will solely be done by the replicating thread. When the page is tried for a fix, its type will be found as `PAGE_UNKNOWN` and fixing will fail and crash due to an assert.

Completely purge the page from page buffer when applying the redo function for `RVPGBUF_DEALLOC`.
Only do this when the function is used within an Passive Transaction Server atomic replication context.

Implementation:
- The function `pgbuf_invalidate` is used to completely [force-]remove the page from page buffer; in tests, this proved as the correct level of abstraction to use to perform page removal/purging
- there is a debug time check that no other concurrent threads (user transactions) are enqueued to read the page; if there are, and the assert is invalidated it remains to be seen as what to do (ie: it will crash; current implementation is supposed to unblock a set of testing scenarios and http://jira.cubrid.org/browse/LETS-596 has been added to revisit this functionality at a later stage of LETS development);
- given the visibility explanation above this should not be case
- normally, one would expect that `pgbuf_invalidate` would also invalidate and wake those waiters informing them of the change, but no such logic seems to exist (or was not found), hence the assert was deemed as the correct mode to ensure this;
- `RVPGBUF_DEALLOC` was, as of now, only encountered in atomic replication sequences, therefore the atomic replication helper bookkeeping mechanism must not try to unfix the page as `pgbuf_invalidate` already did that - change in `pgbuf_unfix_or_ordered_unfix`

Notes:
There is also an option to not change the function `pgbuf_unfix_or_ordered_unfix` and, instead only perform the modification in the atomic replication specific code (ie: atomic replication helper class and associated functions). However, we want to keep this functionality as limited as possible (again, will be followed up with http://jira.cubrid.org/browse/LETS-596). So, if `RVPGBUF_DEALLOC` occurs outside delimited [sysop] atomic sequences of log records, the PTS will crash in (`log_rv_redo_record_sync` - `unfix_rcv_pgptr`).

### DIFF
--- a/src/storage/page_buffer.c
+++ b/src/storage/page_buffer.c
@@ -8807,7 +8807,11 @@ pgbuf_put_bcb_into_invalid_list (THREAD_ENTRY * thread_p, PGBUF_BCB * bufptr)
   /* the caller is holding bufptr->mutex */
   VPID_SET_NULL (&bufptr->vpid);
   bufptr->latch_mode = PGBUF_LATCH_INVALID;
-  assert ((bufptr->flags & PGBUF_BCB_FLAGS_MASK) == 0);
+
+  assert ((bufptr->flags & PGBUF_BCB_FLAGS_MASK) == 0
+	  || (is_passive_transaction_server ()
+	      && ((bufptr->flags & PGBUF_BCB_FLAGS_MASK) == PGBUF_BCB_FLUSH_NOT_NEEDED_FLAG)));
+
   pgbuf_bcb_change_zone (thread_p, bufptr, 0, PGBUF_INVALID_ZONE);
   pgbuf_bcb_check_and_reset_fix_and_avoid_dealloc (bufptr, ARG_FILE_LINE);
 
@@ -14718,6 +14722,30 @@ pgbuf_rv_dealloc_redo (THREAD_ENTRY * thread_p, const LOG_RCV * rcv)
   pgbuf_set_page_ptype (thread_p, rcv->pgptr, PAGE_UNKNOWN);
   pgbuf_set_tde_algorithm (thread_p, rcv->pgptr, TDE_ALGORITHM_NONE, true);
   pgbuf_set_dirty (thread_p, rcv->pgptr, DONT_FREE);
+
+#if defined(SERVER_MODE)
+  if (is_passive_transaction_server ())
+    {
+      /* on a passive transaction server; upon de-allocation, remove the page
+       * from page buffer altogether; do this to avoid the following scenario:
+       * -
+       */
+#if !defined(NDEBUG)
+      /*
+       */
+      PGBUF_BCB *bufptr;
+      CAST_PGPTR_TO_BFPTR (bufptr, rcv->pgptr);
+      assert (bufptr->next_wait_thrd == nullptr);
+#endif
+      const int error_code = pgbuf_invalidate (thread_p, rcv->pgptr);
+      if (error_code != NO_ERROR)
+	{
+	  assert ("page invalidation failed" == nullptr);
+	  return error_code;
+	}
+    }
+#endif
+
   return NO_ERROR;
 }
 
--- a/src/storage/page_buffer.c
+++ b/src/storage/page_buffer.c
@@ -14727,12 +14727,12 @@ pgbuf_rv_dealloc_redo (THREAD_ENTRY * thread_p, const LOG_RCV * rcv)
   if (is_passive_transaction_server ())
     {
       /* on a passive transaction server; upon de-allocation, remove the page
-       * from page buffer altogether; do this to avoid the following scenario:
-       * -
-       */
+       * from page buffer altogether; needed because to provide a fail-safe mechanism
+       * for page re-allocation: when the same page is re-allocated in a subsequent
+       * transaction, the replication mechanism will fail to fix an 'PAGE_UNKNOWN' page */
 #if !defined(NDEBUG)
-      /*
-       */
+      /* crash if there are other, client, transactions waiting in line to fix the same page;
+       * architecturally, this should not be the case (as in a "monolithic" server as well) */
       PGBUF_BCB *bufptr;
       CAST_PGPTR_TO_BFPTR (bufptr, rcv->pgptr);
       assert (bufptr->next_wait_thrd == nullptr);
--- a/src/transaction/atomic_replication_helper.cpp
+++ b/src/transaction/atomic_replication_helper.cpp
@@ -1148,6 +1148,10 @@ namespace cublog
 	// other sanity asserts inside the function
 	pgbuf_ordered_unfix (thread_p, watcher_uptr.get ());
 	break;
+      case RVPGBUF_DEALLOC:
+	assert (watcher_uptr == nullptr);
+	// TODO: do not unfix the page, it has been already flushed from the page buffer
+	break;
       default:
 	assert (watcher_uptr == nullptr);
 	// other sanity asserts inside the function


src/storage/page_buffer.c
```diff
@@ -14718,6 +14722,30 @@ pgbuf_rv_dealloc_redo (THREAD_ENTRY * thread_p, const LOG_RCV * rcv)
...
+       * for page re-allocation: when the same page is re-allocated in a subsequent
+       * transaction, the replication mechanism will fail to fix an 'PAGE_UNKNOWN' page */
+#if !defined(NDEBUG)
+      /* crash if there are other, client, transactions waiting in line to fix the same page;
+       * architecturally, this should not be the case (as in a "monolithic" server as well) */
+      PGBUF_BCB *bufptr;
+      CAST_PGPTR_TO_BFPTR (bufptr, rcv->pgptr);
+      assert (bufptr->next_wait_thrd == nullptr);
+#endif
+      const int error_code = pgbuf_invalidate (thread_p, rcv->pgptr);
```
joohok, line 14740: I'm not sure `pgbuf_invalidate ()` is safe function to use while transactions are running on PTS.

Can we be sure that no transaction has fixed the page in the PTS at the time the pgbuf_rv_dealloc_redo function is called?
I think it would be fine if the pgbuf_rv_dealloc_redo () was a replication that occurred during the vacuum process, but I am not sure if there are other cases.

wrlawodms:
I'd like it if we could deal with this issue after resolving http://jira.cubrid.org/browse/LETS-458, http://jira.cubrid.org/browse/LETS-560. Becasue this will cause another problem after fixing LETS-458, and be solved in LETS-560. And I think this might be not needed if LETS-560 considers this case (maybe NEW_PAGE fix on PTS). 

Even if the deallocated page is purged forcedly, there are still problematic cases. This will remove the crash that occurred on PTS, but possibly still cause a crash on PS while fixing a deallocated page if the kind of new page initialization log record which sets the page type does not belong to an atomic sequence. I mean if it's fixed with OLD_PAGE on PS. Because the RVBT_GET_NEWPAGE redo job on the PTS requests PS to fix the deallocated page with the OLD_PAGE fetch mode. Actually, the case of RVBT_GET_NEWPAGE doesn't matter because it belongs to an atomic operation and the atomic replicator (will) fixes a page using OLD_PAGE_MAYBE_DEALLOCATED (http://jira.cubrid.org/browse/LETS-554), but other new page initialization log records like RVPGBUF_NEW_PAGE will cause the problem. (It seems that there are cases in which they don't belong to an atomic operation such as in `overflow_update()`).
 
Actually, this is the problem described in http://jira.cubrid.org/browse/LETS-560, which is shadowed by http://jira.cubrid.org/browse/LETS-458. The way I see it, these issues seem correlated. 
1. In the current code, PTS requests a page with the "next" of replicated LSA, and PS waits for the LSA to be processed. [_This means that when a PTS requests a page to apply a redo log record, the page received from PS has already been applied with the redo log record_| **Later edited**: It turns out this can't happen because the replicator doesn't fetch a page from PS]. This is not only awkward but also could cause several problems like a hang during Boot. (http://jira.cubrid.org/browse/LETS-458)
2. If the first one is fixed, another hidden problem is revealed, which is described above and in http://jira.cubrid.org/browse/LETS-560.

Although if all LETS-458/560 are resolved, there will be no problem with this PR, but I think we should come up with a solution that considers all of them. What I suggested in LETS-560 is to propagate a proper fetch mode while applying a new page initialization redo record. If we also deal with a NEW_PAGE fetch mode on PTS, this PR is not needed and it seems a more natural way to me. It doesn't even suffer from the redundant page purge and fetch on the network by this PR.

cristiarg:
@wrlawodms 
Thank a lot for your analysis.
I do agree that we have to fix the actual fundamental root cause.

But let's also apply this patch - temporarily - to unblock:
- allow testing for http://jira.cubrid.org/browse/LETS-524
- in my tests this also unblocks other QA issues regarding PTS functioning which are currently blocked by the same kind of crash
- merge as many tests from http://jira.cubrid.org/browse/LETS-511 (and other QA issue) as possible in the testcases mainline as to have a guard against further patches

We currently have next to nothing as tests for PTS in the testcases.
We will add a new issue to revert this change once http://jira.cubrid.org/browse/LETS-458 and associated issues are fixed.
It seems that there are so many inter-dependencies and we are running around "chasing our own tail".

So, if this change is **not considered fundamentally wrong**, I would still like to apply it **temporarily** for reasons above.

wrlawodms:
@cristiarg Sorry, I misunderstood the machoism to fetch a page from PS. I've realized this during the meeting...
> but possibly still cause a crash on PS while fixing a deallocated page if the kind of new page initialization log record which sets the page type does not belong to an atomic sequence.
...
but other new page initialization log records like RVPGBUF_NEW_PAGE will cause the problem. (It seems that there are cases in which they don't belong to an atomic operation such as in overflow_update()).

It won't happen because the replicator doesn't fetch a page if it doesn't belong to an atomic replication. Right?

> This means that when a PTS requests a page to apply a redo log record, the page received from PS has already been applied with the redo log record

It's only the case with atomic sequences.

**The conclusion is the same, though. We should come up with a solution considering the issues mentioned. I just want to correct my wrong opinion which possibly confuses you. 😢.**

One another question with my current understanding:
> Upon replicating the subsequent RVBT_GET_NEWPAGE log record, I assume that the log replication mechanism fails to fix the page - as expected - and thus, the page continues to remain marked as PAGE_UNKNOWN in the page buffer's bookkeeping.
The replication's fix mechanism is implemented with the help of the new fetch mode OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT and the place this is done starts with log_rv_fix_page_and_check_redo_is_needed.

I think the reason replicating the RVBT_GET_NEWPAGE log record fails is that it returns NULL to fix a deallocated page with the OLD_PAGE_MAYBE_DEALLOCATED fetch mode. Am I wrong? Is it related to the OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT?

cristiarg:
@wrlawodms 
> It won't happen because the replicator doesn't fetch a page if it doesn't belong to an atomic replication. Right?

Correct.

> I think the reason replicating the RVBT_GET_NEWPAGE log record fails is that it returns NULL to fix a deallocated page with the OLD_PAGE_MAYBE_DEALLOCATED fetch mode. Am I wrong? Is it related to the OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT?

~~Hmm. You are right. So, this would not be needed if we'' switch to `OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT` for the atomic replication fixes.~~
~~In that case, let's first implement/test and evaluate that change and then come back to this. Maybe it's not needed anymore at all.~~

The crossed-out above statement is true only for fixes coming from the atomic replication thread itself - which, with the changes in #3931 will use `OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT`.
However, further tests revealed that, without this patch, a PTS would still crash because requests for pages coming from client threads - with `OLD_PAGE/OLD_PAGE_PREVENT_DEALLOC` fetch modes - will still bump into the buffer control block for the page showing `PAGE_UNKNOWN`.
It is actually an uncovered scenario in `pgbuf_fix` which might be solved by explicitly checking for the specific corner case somewhere at the top of `pgbuf_fix` - `is_passive_transaction_server` and fetch mode read - and explicitly invalidate & re-request the page from Page Server.

wrlawodms:
> However, further tests revealed that, without this patch, a PTS would still crash because requests for pages coming from client threads - with OLD_PAGE/OLD_PAGE_PREVENT_DEALLOC fetch modes - will still bump into the buffer control block for the page showing PAGE_UNKNOWN.

Assuming using OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT, problematic cases that I can imagine are:
- On the replicator: The deallocated page is still in buffer or transit, and PTS tries to fix it with OLD_PAGE. It might have to be with NEW_PAGE or something.
- On clients: Clients request a page with OLD_PAGE, but it's been deallocated on PS. I'm not sure this case can happen, but I guess it can since a PTS could request a page with quite outdated information. 

At least, the frist one won't happen with this PR.

src/transaction/atomic_replication_helper.cpp
```diff
@@ -1161,6 +1161,10 @@ namespace cublog
 	// other sanity asserts inside the function
 	pgbuf_ordered_unfix (thread_p, watcher_uptr.get ());
 	break;
+      case RVPGBUF_DEALLOC:
+	assert (watcher_uptr == nullptr);
+	// TODO: do not unfix the page, it has been already flushed from the page buffer
```
wrlawodms, line 1166: Does the RVPGBUF_DEALLOC always belong to an atomic sequence? Don't we skip unfixing in a normal replicator (log_rv_redo_record_sync())?

src/storage/page_buffer.c
```diff
@@ -8807,7 +8807,11 @@ pgbuf_put_bcb_into_invalid_list (THREAD_ENTRY * thread_p, PGBUF_BCB * bufptr)
   /* the caller is holding bufptr->mutex */
   VPID_SET_NULL (&bufptr->vpid);
   bufptr->latch_mode = PGBUF_LATCH_INVALID;
-  assert ((bufptr->flags & PGBUF_BCB_FLAGS_MASK) == 0);
+
+  assert ((bufptr->flags & PGBUF_BCB_FLAGS_MASK) == 0
+	  || (is_passive_transaction_server ()
+	      && ((bufptr->flags & PGBUF_BCB_FLAGS_MASK) == PGBUF_BCB_FLUSH_NOT_NEEDED_FLAG)));
```
wrlawodms, line 8813: Could you explain it briefly if you don't mind?

src/storage/page_buffer.c
wrlawodms, line 14740: The comment says `pgbuf_invalidate` has to be called only in a postponed operation. Is RVPGBUF_DEALLOC always called in it?

src/transaction/atomic_replication_helper.cpp
wrlawodms: > RVPGBUF_DEALLOC was, as of now, only encountered in atomic replication sequences, therefore the atomic replication helper bookkeeping mechanism must not try to unfix the page as pgbuf_invalidate already did that - change in pgbuf_unfix_or_ordered_unfix

Oh.. I see this late. Then, forget it.

src/storage/page_buffer.c
wrlawodms: I did a little digging and it seems right, though.

wrlawodms (approved):

cristiarg:
@wrlawodms 

> Assuming using OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT, problematic cases that I can imagine are:
> - On the replicator: The deallocated page is still in buffer or transit, and PTS tries to fix it with OLD_PAGE. It might have to be with NEW_PAGE or something.
> - On clients: Clients request a page with OLD_PAGE, but it's been deallocated on PS. I'm not sure this case can happen, but I guess it can since a PTS could request a page with quite outdated information. 
> 
> At least, the frist one won't happen with this PR.

Another way to fix this is (or, could be):
- discard the changes proposed in this PR
- adapt `pgbuf_fix` function (or one of the sub-functions working on the buffer control block - bcb) to account for the de-alloc/re-alloc scenario (in PTS context only, of course);
I don't know how. Maybe use a new `bcb` flag, aside from `PGBUF_BCB_DIRTY_FLAG` which is set from `pgbuf_rv_dealloc_redo` (via `pgbuf_set_dirty`).

Invalidating the page is a brute force approach. But it provide us insight about all the cases (ie: where this approach does not fit it will crash or dead-latches/locks will occur).
We can implement this via brute force to sort out what the scenarios are - via further tests. And then, we can refactor this via an extra flag taken which will then be handled in the `pgbuf_fix` function.

What do you think about this approach?

src/transaction/atomic_replication_helper.cpp
cristiarg: If it occurs in a non-atomic sequence context - it will crash in `log_rv_redo_record_sync` - `unfix_rcv_pgptr` functor

wrlawodms:
@cristiarg Sorry to confuse you. I agree with the approach, which means that let's merge this PS and see, right? I just wanted to share some cases I'm concerned and might be taken into account later.

cristiarg:
@wrlawodms 

Ok. Thanks.
I will write a follow-up issue for this anyway as - I think - this patch unblocks some tests but we definitely have not seen the bottom of all is possible.
And, the more tests we have, the better we are covered and the more we are free to try-out new approaches.

I have added http://jira.cubrid.org/browse/LETS-596 as a follow-up to this patch and other related.

src/storage/page_buffer.c
cristiarg: @wrlawodms 

So, I might have gotten a few things wrong:
- I assumed that the flush daemon is **not** executed in PTS context - it is, see `log_daemons_init`
- `PGBUF_BCB_FLUSH_NOT_NEEDED_FLAG` was added specifically for PTS to inhibit flush daemon to flush pages because there is nowhere to flush those pages to
- so, in effect, the pages remain in 
- in effect, I think, the PTS flush daemon only works on temporary page buffer pages

Maybe another way of dealing with this is to actually make the PTS flush daemon perform this purging from the PTs's page buffer.
However, this raise a concern about race conditions:
  - (1) what if PTS replication must replicate a log-record which re-allocates the page
  - (2) but the PTS flush daemon has not flushed (ie: purged) the page yet from the page buffer

(1) If flush daemon purged the page from page buffer, and no other user transactions re-fetched the page from PS, the replication would skip applying the re-allocation log record.
(2) If flush daemon purged the page from page buffer, and if a subsequent user transaction re-fetched the page from PS, the page would go through the page synchronization mechanism and this situation would also be ok.
(3) If flush daemon did not manage to purge the page, and the replication would need to re-apply the log record re-allocating the page, this is the race condition and it will crash.

So the change above is a "cheap" solution for this. Not **the solution**, but a "cheap" one.

Maybe we must add a separate issue to revisit the concept introduced by `PGBUF_BCB_FLUSH_NOT_NEEDED_FLAG`:
- use the flush daemon to perform purge of de-allocated pages on PTS
- PTS replication must invoke this daemon in a more deterministic way

wrlawodms: Overall, I understand why you invalidate BCBs directly, not leveraging the flush daemon. That's what you're explaining, right? 

> I assumed that the flush daemon is not executed in PTS context - it is, see log_daemons_init

1. Do you mean `pgbuf_page_flush_daemon_init()` in `pgbuf_daemons_init()` by the flush daemon? Or do you really mean a log-related daemon in `log_daemons_init()`? It could be the checkpoint daemon.
I'm asking to clarify.

2. The reason to make the chage to the assert is that BCBs were not invalidated so, it didn't consider PGBUF_BCB_FLAGS_MASK turned on in PTS. Do I get it right?

joohok (approved):

hornetmj (approved):

src/storage/page_buffer.c
cristiarg: > 1. Do you mean `pgbuf_page_flush_daemon_init()` in `pgbuf_daemons_init()` by the flush daemon? Or do you really mean a log-related daemon in `log_daemons_init()`? It could be the checkpoint daemon.
> I'm asking to clarify.

You are correct. I am wrong.
Log daemons have nothing to do with this case. It is the page buffer daemons. And they are started in all server flavors (ATS, PTS, PS) regardless of any condition - `pgbuf_daemons_init`.

> 2. The reason to make the chage to the assert is that BCBs were not invalidated so, it didn't consider PGBUF_BCB_FLAGS_MASK turned on in PTS. Do I get it right? 

Yes, buffer control blocks were not invalidated before this change.
And, given your comment, I am starting to think that we will encounter further issues/crashes once we start testing without the entire database fitting into page buffer. I assume that - probably - >~97% of all tests up to now have only encountered the situation where the entire database fitted into memory and the page buffer flush daemons did not have much to do.
Or maybe I get this wrong and they also consider age and they did purge the page buffer (LRU? maybe). I don't know actually.

However, http://jira.cubrid.org/browse/LETS-596 is written to come back to this implementation once we are past a critical point that allows us to test further scenarios.

wrlawodms: > Or maybe I get this wrong and they also consider age and they did purge the page buffer (LRU? maybe). I don't know actually.

I did an analysis of the page buffer pool. From what I know, if there is no available BCB for a new page, a BCB that is basically not dirty is assigned for it (there are some other conditions such as not flushing or something). This is called "victimize". Which BCB is victimized is decided by basically multi-layered LRU lists. To keep enough victimizable BCBs, the page flush daemon, pgbuf_page_flush_daemon_task, periodically flushes BCBs. 
 If all BCB are not dirty, i.e. no victimizable BCB, a thread wanting to allocate a new BCB waits until a BCB gets victimizable by the flush daemon or others. A BCB that is assigned in this way is called a "direct victim".

Basically, I guess a data page is "purged" while victimizing. I hope this information helps you, @cristiarg.

PR #3964, opened 2022-11-25
[LETS-599] fix dead-latch between client transaction thread and replication thread on PTS

cristiarg:
http://jira.cubrid.org/browse/LETS-599

During heap scan, there is a mechanism - which is governed by `heap_scancache::cache_last_fix_page` flag - in which the state where the scan failed is saved in a `heap_get_context::scan_cache::page_watcher` (maybe for a future re-entry?).
On Passive Transaction Server,  when such a heap scan retrieves a page (from Page Server) which is ahead of the current state of replication on PTS, it must halt its processing and wait for the replication to progress enough for it to continue the scan.
Since the replication mechanism can modify any page whatsoever, it can modify the very pages that the heap scan currently holds.
In order to allow the replication to advance, the client transaction heap scan must unfix all pages that it has fixed.
It was the case that the page held by `scan_cache::page_watcher` was not being released, leading to a dead-latch between client transaction thread and replication thread in the page buffer layer.

Implementation:
- in `heap_get_visible_version_with_repl_desync` make sure to also unfix the page held by `context->scan_cache->page_watcher` (if present)
- but keep state information to allow the same page to be re-fixed after the wait for replication thread has finished
- after the wait, re-fix the page with the same parameters as before (same latch mode)
- the re-fix can of course fail (the replication might have deleted the page); leave it up to the next call to `heap_get_visible_version_internal` to sort it out
- after re-fix, make use of the function `heap_init_get_context` to restore the state ready for a new call to `heap_get_visible_version_internal`; the function `heap_init_get_context` will move back fixed page from the watcher `scan_cache::page_watcher` to the home watcher of the scan structure - the following section of the function does this:
```
  if (scan_cache != NULL && scan_cache->cache_last_fix_page && scan_cache->page_watcher.pgptr != NULL)
    {
      /* switch to local page watcher */
      pgbuf_replace_watcher (thread_p, &scan_cache->page_watcher, &context->home_page_watcher);
    }
```
- related change: extended the function `pgbuf_ordered_unfix*` with a new flavor `pgbuf_ordered_unfix_and_save_for_refix*` that outputs information required (vpid, latch mode) to do a re-fix of the same page

Note that `heap_get_visible_version_with_repl_desync` implements a loop because, a scan involves access to multiple pages (at least a home page an a forward page at a time) and both these pages might be "ahead of replication" and both of them might need to be waited by the client transaction.


src/storage/heap_file.c
```diff
@@ -25141,6 +25192,17 @@ heap_get_visible_version_internal (THREAD_ENTRY * thread_p, HEAP_GET_CONTEXT * c
       /* we need class_oid to check if the class is mvcc enabled */
       context->class_oid_p = &class_oid_local;
     }
+  // *INDENT-OFF*
+  scope_exit <std::function<void (void)>> reset_class_out_ftor (
```
cristiarg, line 25230: change not related to the main PR; it seemed like a potential bug

src/storage/heap_file.c
```diff
@@ -25091,19 +25091,70 @@ heap_get_visible_version_with_repl_desync (THREAD_ENTRY * thread_p, HEAP_GET_CON
...
+	      // (most likely, the transactional log replication thread) to advance past a
+	      // certain point;
+	      // during replication, the page might be de-allocated while processing a log record
+	      // thus, failing to re-fix is expected
+	      //
+	      // TODO: in effect, the fetch mode is not correctly propagated to the page server;
+	      // instead - OLD_PAGE - is actually propagated to the
+	      // server (see pgbuf_request_data_page_from_page_server)
+	      // thus, if page has been de-allocated, an error will be retrieved from page server
+	      assert (false);
```
razvanradulescu: Q: the debug version has this assert - does this means we think it should not happen, or is temporary to catch unhandled cases ?
In case of release version, do we continue ? I expected to break loop and return error.

src/storage/heap_file.c
```diff
@@ -25091,19 +25091,70 @@ heap_get_visible_version_with_repl_desync (THREAD_ENTRY * thread_p, HEAP_GET_CON
...
+      // clean - saving state if required
+      //
+      heap_clean_get_context (thread_p, context);
+
+      // also unfix what is in cache watcher; saving the state for a future re-fix after the wait
+      //
+      VPID saved_vpid VPID_INITIALIZER;
+      PGBUF_LATCH_MODE saved_latch_mode = PGBUF_NO_LATCH;
+      bool restore_scan_cache_page_watcher = false;
+      if (is_heap_scan)
```
joohok: Just a question, 
I don't understand why this condition (`is_heap_scan`) is required when unfix the page buffer. It would be very helpful if you could explain why.

src/storage/heap_file.c
cristiarg: If the page cannot be re-fixed the single known cause is that the page has been de-allocated in the meantime. And since the only thing that executes changes on pages on PTS is the transactional log replication, this means that the page has been de-allocated and the replication has processed a `RVPGBUF_DEALLOC` for the page.

With the patch #3960, on PTS, once the `RVPGBUF_DEALLOC` log record is replicated, the page is invalidated from the page buffer (this might be an intermediary brute force solution - but it's what we have currently).
Thus, in such a scenario, if the page would be re-attempted for a fix at this point, the request for fix would transform into a request for fetch from Page Server, which will return an error code (**) as the page has been de-allocated.
In this case the assert is problably incorrect.
I will add a dump to err_log.
A subsequent interation - call to `heap_get_visible_version_internal` will probably come up empty handed and return.

(**)
There is even a worse case when the de-allocated page has already been re-allocated - maybe by another transaction on ATS. In which case, the fix will succeed but the layout of the page will be different and the search will fail - maybe even downright crash.
For this, we don't have an implementation. We might have a solution - described in http://jira.cubrid.org/browse/LETS-523 .

So, in all, I added the assert with the intention of catching one/both of the above described situations.

src/storage/heap_file.c
cristiarg: You might be right.
I also checked and there is no apparent link between `is_heap_scan` and the fact that the home page watcher is saved in `scan_cache`. The save in `scan_cache` happens in `heap_get_visible_version_internal` -> `heap_prepare_get_context` -> `heap_clean_get_context`

src/storage/heap_file.c
cristiarg: You might be right.
I also checked and there is no apparent link between `is_heap_scan` and the fact that the home page watcher is saved in `scan_cache`.
The save in `scan_cache` happens in `heap_get_visible_version_internal` -> `heap_prepare_get_context` -> `heap_clean_get_context`.
Removed the check.

joohok (approved):

src/storage/heap_file.c
```diff
@@ -25090,20 +25093,102 @@ heap_get_visible_version_with_repl_desync (THREAD_ENTRY * thread_p, HEAP_GET_CON
...
+      // wait for replication to catch-up
+      //
       constexpr VPID null_vpid = VPID_INITIALIZER;
       pgbuf_wait_for_replication (thread_p, &null_vpid);
+
+      // re-fix scan cache page watcher, if unfixed previously
+      //
+      if (restore_scan_cache_page_watcher)
+	{
+	  const int error_code = pgbuf_ordered_fix (thread_p, &saved_vpid, OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT,
```
cristiarg, line 25134: @razvanradulescu 
After some more though, I changed to use the `OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT` for the re-fix.
I also added an extensive comment below better explaining the rationale.

src/storage/heap_file.c
```diff
@@ -25090,20 +25093,102 @@ heap_get_visible_version_with_repl_desync (THREAD_ENTRY * thread_p, HEAP_GET_CON
...
-	{
-	  /* Unfix home page. */
-	  pgbuf_ordered_unfix (thread_p, &context->home_page_watcher);
-	}
+      //
 
-      if (context->fwd_page_watcher.pgptr != NULL)
+      // clean - saving state if required
+      //
+      heap_clean_get_context (thread_p, context);
```
wrlawodms, line 25103: I thought that you want to refix not only the home page and the forward page, but also a page saved in `context->scan_cache->page_watcher.pgptr`. However, it seems that `heap_clean_get_context` makes `page_watcher.pgptr` point to the home page so it ends up just refixing only home and forward pages just as before. Do I get it wrong?

wrlawodms: >However, it seems that heap_clean_get_context makes page_watcher.pgptr point to the home page 

in heap_clean_get_context()
```
25485   if (context->scan_cache != NULL && context->scan_cache->cache_last_fix_page
25486       && context->home_page_watcher.pgptr != NULL)
25487     {
25488       /* Save home page (or NULL if it had to be unfixed) to scan_cache. */
25489       pgbuf_replace_watcher (thread_p, &context->home_page_watcher, &context->scan_cache->page_watcher);
25490       assert (context->home_page_watcher.pgptr == NULL);
25491     }
```

cristiarg: Right after `ER_PAGE_AHEAD_OF_REPLICATION` error is detected the intended logic is as follows:
- step `heap_clean_get_context`:
  - if there is a home page watcher, it is saved in `scan_cache` watcher
  - home and forward pages are cleaned (well, at this point only the forward pages might have something in the watcher)
- step `pgbuf_ordered_unfix_and_save_for_refix`:
  - if there was something saved in the `scan_cache` page wather, it is unfixed and the coordinates for refix are saved (latch mode and vpid)
- step: wait for replication
- step: if `restore_scan_cache_page_watcher`
  - if there was a `scan_cache` watcher active it is re-fixed with `OLD_PAGE_IF_IN_BUFFER_OR_IN_TRANSIT`
- step: `heap_init_get_context`
  -  will restore the `scan_cache` watcher (if present) to the home page wather

The purpose is to leave the state as it was as if there was no wait for replication.
With the caveat explained that the page might not be there anymore.
In that case we leave it up to the next call to `heap_get_visible_version_internal` to deal with such state.

However, the forward page is purposefully not re-fixed, only the home will be re-fixed (via the `scan_cache` route as described above).

Does this answer your question?

wrlawodms: I understand every step you explain, but I don't think it still addresses my question.
> It was the case that the page held by scan_cache::page_watcher was not being released, leading to a dead-latch between client transaction thread and replication thread in the page buffer layer.

The purpose of this PR is to release a page that is possibly fixed in scan_cache::page_watcher. Do I get it right? Then, 
> step heap_clean_get_context:
if there is a home page watcher, it is saved in scan_cache watcher
home and forward pages are cleaned (well, at this point only the forward pages might have something in the watcher)

In the above,  scan_cache::page_watcher is overwritten with the home page and the forward page is unfixed in heap_clean_get_context(), and the home page (in scan_cache::page_watcher) is unfixed in the caller. After the home page is refixed after waiting and the forward page will be refixed in the next `heap_get_visible_version_internal` call. 
So, it seems to end up just refixing the home and forward pages just like before. That is my question. Sorry if you already explained this.

wrlawodms: Additional question: 
As far as I know, a thread also possibly keeps some other different pages fixed than the three pages above. Don't they matter for the issue case?

cristiarg: The purpose is to:
- unfix all pages that are left fixed by the call to `heap_get_visible_version_internal`
- wait for replication
- refix only what is saved in `scan_cache` based on the idea that it is a "cache" and is needed to re-attempt to retrieve the data
- looking deeper at the call path `heap_get_visible_version_internal` -> `heap_prepare_get_context` -> `heap_prepare_object_page(.. &context->home_page_watcher ..)` I can see that the logic benefits from having the home page fixed (as restored from `scan_cache` by the last call to `heap_init_get_context`).

I still have a feeling that I don't manage to answer your question.
If you think the patch introduces some inconsistency or does "too much" or "too little" just name it. :)

> As far as I know, a thread also possibly keeps some other different pages fixed than the three pages above. Don't they matter for the issue case?

What other pages? Where?
This patch does address one found case of "keeps other different pages fixed".

wrlawodms: >unfix all pages that are left fixed by the call to heap_get_visible_version_internal 

The home and forward pages were unfixed also in the previous code, right? And you said we missed a page to unfix before waiting. It must be the page stored in `scan_cache::page_watcher`. Do I get it right? 
In the changed code, the way I see it, the `scan_cache::page_watcher` could be overwritten by the home page in `heap_clean_get_context()`, Then, we end up losing track of the page stored in `scan_cache::page_watcher` previously and leaving it unfixed. Do I make it clear? 

in heap_clean_get_context():
```
25485   if (context->scan_cache != NULL && context->scan_cache->cache_last_fix_page
25486       && context->home_page_watcher.pgptr != NULL)
25487     {
25488       /* Save home page (or NULL if it had to be unfixed) to scan_cache. */
25489       pgbuf_replace_watcher (thread_p, &context->home_page_watcher, &context->scan_cache->page_watcher);
25490       assert (context->home_page_watcher.pgptr == NULL);
25491     }
```

P.S. Looking into `pgbuf_replace_watcher()` again, it expects `context->scan_cache->page_watcher` to point to nothing. Then, I guess that pgbuf_replace_watcher() is only called when there is no pgptr in `context->scan_cache->page_watcher`. And I think the above code snippet is not executed for the issue case and that `heap_clean_get_context()` is just another way to unfix the home and forward page. If it's just a short way to unfix them, I'm OK. I was worried that we possibly lose the `context->scan_cache->page_watcher`.

> What other pages? Where?
This patch does address one found case of "keeps other different pages fixed".

Generally speaking, a thread can fix many different pages and I guessed that there may be some pages fixed other than the three pages described. For example, before coming into `heap_get_visible_version_with_repl_desync()`, a heap header or the previous heap page during a heap scan could be fixed (I didn't see the case, I just guessed it generally). But if no pages are guaranteed to remain fixed, just forget it. It would be good if there is an assertion that we can check if no pages are fixed by the thread, though.

hornetmj (approved):

src/storage/heap_file.c
razvanradulescu: @wrlawodms : Indeed, the same transaction can have other pages already fixed from outer scans.
I expect the replication thread, wants to fix a bunch of pages belonging to the same file as part of an "atomic" replication operation (it will never want to fix pages from different heap files).
So, I think the problem can arise only when the outer scan is on the same heap file as the inner scan (as hint to reproduce the issue).
Regardless to repro, a possible solution could be to unfix all ordered fixed pages (all pages having watchers in the current transaction) and then refix all of them.

cristiarg: @wrlawodms 

> The home and forward pages were unfixed also in the previous code, right?

Yes

> And you said we missed a page to unfix before waiting. It must be the page stored in scan_cache::page_watcher. Do I get it right?

Yes

> In the changed code, the way I see it, the scan_cache::page_watcher could be overwritten by the home page in heap_clean_get_context(), Then, we end up losing track of the page stored in scan_cache::page_watcher previously and leaving it unfixed. Do I make it clear?

Yes, the question is clear. But we cannot lose track of whatever was previously fixed in `scan_cache->page_watcher` because, as you also point out, the `scan_cache->page_watcher` is asserted empty in `pgbuf_replace_watcher`. However, this code is safe only in debug mode while in release mode, I suppose an overwrite into an non-empty `scan_cache->page_watcher` could happen.
On the other hand, this scenario could have happened before as well. So wherever this inconsistency might have happened it shoul have already happened (either in tests or in production). The newly introduced code does not introduce anything new in this aspect.

cristiarg: @wrlawodms @razvanradulescu 

I looked further into the matter and I see two aspects of the problem which I can summarize as follows:

> If a PTS client transaction/thread must wait the replication thread to advance, it must also "make way" so the replication thread can advance. "Make way" means: unfix ALL pages that the transaction holds fixed before waiting on the replication thread to advance.

I looked at the code and I see one clear potential issue:
- on the call path `heap_next_internal` -> `heap_scan_get_visible_version` -> `heap_get_visible_version_with_repl_desync`
- in function `heap_next_internal` there are some other watchers - `curr_page_watcher` and `old_page_watcher` - which I don't think are guaranteed to point to nothing
- thus, it is possible that these are holding other pages fixed while the function `heap_get_visible_version_with_repl_desync` is performing a wait on replication thread
- thus, the potential for dead-latches is still there
- for this problem I will add another - separate - issue to implement a fix; at the very least we can add some checks with some `err_set` statements and debug asserts that will at least signal the problem to us;
- in short we'll deal with this in a separate issue

I also can imagine another issue but I know no clear scenario if this can be reproduced:
- is it possible in a client transaction to have outer and inner scans and as @razvanradulescu sais:

> So, I think the problem can arise only when the outer scan is on the same heap file as the inner scan (as hint to reproduce the issue).

- maybe this issue is in fact a more general one of the previous particular description (the one above starting from `heap_next_internal` function)
- I will also describe this and we'll take this in the separate issue

wrlawodms: Now everything is clear to me. Thank you for taking it seriously even without a concrete example.

wrlawodms (approved):

cristiarg:
Added http://jira.cubrid.org/browse/LETS-609 as a follow up issue.

PR #3969, opened 2022-11-28
[CBRD-24569] Reduce unnecessary function calls when getting system parameters.

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-24569

Reduce unnecessary function calls when getting system parameters.
After the modification of CUBRIDSUS-8193, getting the system parameter was changed from a global variable to a function call. I change this to inline function.


src/base/system_parameter.h
```diff
@@ -539,13 +677,6 @@ extern "C"
 
   extern void *prm_get_value (PARAM_ID prm_id);
```
beyondykk9, line 678: By your modification, "prm_get_value" is called in only case of SERVER_MODE. Therefore, you should modifity "prm_get_value" to reflect your modified logic.

The prm_get_value has redundant logic as like below:

```
#if defined (SERVER_MODE)
  THREAD_ENTRY *thread_p;

  assert (prm_id <= PRM_LAST_ID);

  if (PRM_SERVER_SESSION (prm_id) && BO_IS_SERVER_RESTARTED ())
    {
      SESSION_PARAM *sprm;
      thread_p = thread_get_thread_entry_info ();
      sprm = session_get_session_parameter (thread_p, prm_id);
      if (sprm)
	{
	  return &(sprm->value);
	}
    }

  return prm_Def[prm_id].value;
#else /* SERVER_MODE */
  assert (prm_id <= PRM_LAST_ID);

  return prm_Def[prm_id].value;
#endif /* SERVER_MODE */
```

shparkcubrid: There may be cases where only prm_get_value () is called separately, so I did not fix it.

beyondykk9: The "prm_get_value" is called from "css_sockout" and the other calls are only from system_param local funcions as like "prm_get_xxxx_value". The "css_sockout" can be modified simply as calling "prm_get_integer_value". In my opinion, the "prm_get_value" should be modified because it has redundant logics as assert (prm_id <= PRM_LAST_ID) and "PRM_SERVER_SESSION" checking.

hyunikn (approved):

hgryoo (approved):

beyondykk9 (approved):

ctshim (approved):

PR #3971, opened 2022-11-30
[LETS-584] completely suppress replication delay output on PTS when debug logging is not specified

cristiarg:
http://jira.cubrid.org/browse/LETS-584

Supress PTS replication delay debug logging unless both `er_log_debug` and `er_log_calculate_replication_delay` are `true`.
Needed to allow seamless execution of PTS replication thread as debug logging is synchronous and, thus, affects performance.

This also is needed for executing the Scale-Out Demo with a "silent" PTS server.

### DIFF
--- a/src/transaction/log_replication_jobs.cpp
+++ b/src/transaction/log_replication_jobs.cpp
@@ -48,6 +48,11 @@ namespace cublog
   int
   log_rpl_calculate_replication_delay (THREAD_ENTRY *thread_p, time_msec_t a_start_time_msec)
   {
+    // only log if both params are active; needed to completely supress logging because
+    // logging - being synchronous - affects replication performance in release builds
+    const bool output_debug_log = prm_get_bool_value (PRM_ID_ER_LOG_DEBUG) &&
+				  prm_get_bool_value (PRM_ID_ER_LOG_CALC_REPL_DELAY);
+
     // skip calculation if bogus input (sometimes, it is -1);
     // TODO: fix bogus input at the source if at all possible (debugging revealed that
     // it happens for LOG_COMMIT messages only and there is no point at the source where the 'at_time'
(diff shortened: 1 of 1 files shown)


src/transaction/log_replication_jobs.cpp
```diff
@@ -57,12 +62,11 @@ namespace cublog
...
-	// assert (time_diff_msec >= 0);
 	if (time_diff_msec > 0)
 	  {
 	    perfmon_set_stat (thread_p, PSTAT_REDO_REPL_DELAY, static_cast<int> (time_diff_msec), false);
 
-	    if (prm_get_bool_value (PRM_ID_ER_LOG_CALC_REPL_DELAY))
+	    if (output_debug_log)
 	      {
 		_er_log_debug (ARG_FILE_LINE, "[CALC_REPL_DELAY]: %lld msec", time_diff_msec);
 	      }
```
wrlawodms, line 72: How about just using 'er_log_debug' instead? It prints only if `prm_get_bool_value (PRM_ID_ER_LOG_DEBUG)`.
```suggestion
	    if (prm_get_bool_value (PRM_ID_ER_LOG_CALC_REPL_DELAY))
	      {
		er_log_debug (ARG_FILE_LINE, "[CALC_REPL_DELAY]: %lld msec", time_diff_msec);
	      }
```

cristiarg: `PRM_ID_ER_LOG_DEBUG` is already accounted for upon initialization of `output_debug_log`.

wrlawodms: I suggested using `er_log_debug` without `output_debug_log` to write logging only in the debug mode. This function seems suited for your purpose. It's up to you.

wrlawodms (approved):

joohok (approved):

cristiarg:
> I suggested using er_log_debug without output_debug_log to write logging only in the debug mode. This function seems suited for your purpose. It's up to you.

My modification achieves the same thing as you suggest.
And I do know about `er_log_debug` macro.
The idea with the `output_debug_log` is that maybe we'll output it from outside to only read value once from system parameters.

PR #3972, opened 2022-12-01
[CBRD-24478] reduce duplicate keys

ctshim:
http://jira.cubrid.org/browse/CBRD-24478

* Make sure that the index is not skewed by duplicate keys.


src/base/system_parameter.h
```diff
@@ -466,8 +466,16 @@ enum param_id
   PRM_ID_QMGR_MAX_QUERY_PER_TRAN,
   PRM_ID_REGEXP_ENGINE,
   PRM_ID_ORACLE_STYLE_NUMBER_RETURN,
+#if defined(SUPPORT_COMPRESS_MODE)
```
shparkcubrid, line 469: This could be a problem here, how about writing it as a comment?

hgryoo:
Instead of the term 'COMPRESS', how about using another term? It sounds like the index key or data is compressed.

I'd like to suggest 'DEDUPLICATE_KEY'.
```
opt_index_dedup_key_mode ::= [ {DEDUPLICATE_KEY [OFF | HIGH | LOW | MEDIUM opt_index_dedup_key_mod_level } ]        
opt_index_dedup_key_mod_level ::= ['(' UNSIGNED_INTEGER ')' ] 
```

Details:
```
opt_index_dedup_key_mode
              OID = {Volumn ID, Page ID, Slot ID}

               OFF: Default value when the setting is omitted, configured the same as the index of the previous version, use the Overflow pages for OIDs of a duplicated key as it is.
                       (data representation for Index entires: k* = <{k}, [OID1, OID2, OID3, ...]>)
                
               LOW: The duplicated index entries are grouped by the same page id of OID as additional information
                       (data representation for Index entires: k* = <{k, page_id}, [OID1, OID2, OID3, ...]>)

               HIGH: The duplicated index entries are treated as unique keys using OID.
                       (data representation for Index entires: k* = <{k, OID}, []>)
                       * accessed by the covered index..? I'm not sure.

               MEDIUM: The duplicated index entries are grouped by the remainder of OID page id division. 'opt_index_dedup_key_mod_level' in this case.
```

src/base/object_representation_sr.c
```diff
@@ -2234,6 +2246,21 @@ or_install_btids_constraint (OR_CLASSREP * rep, DB_SEQ * constraint_seq, BTREE_T
     {
       assert (DB_VALUE_TYPE (&att_val) == DB_TYPE_INTEGER);
       att_id = db_get_int (&att_val);	/* The first attrID */
+
+#if defined(SUPPORT_DEDUPLICATE_KEY_MODE)
+      if (IS_DEDUPLICATE_KEY_ATTR_ID (att_id))
+	{
+	  /* { btid, reserved_index_attrID, asc_desc, [attrID, asc_desc]+, {fk_info} or {key prefix length}, status, comment} */
+	  assert (seq_size >= 8);
```
beyondykk9, line 2269: needed a comment for constant 3 and 8.

src/object/class_object.c
```diff
@@ -4045,7 +4064,75 @@ classobj_find_cons_index2_col_type_list (SM_CLASS_CONSTRAINT * cons, OID * root_
...
+	{
+	  continue;
+	}
+
+      if (!cons->attributes)
+	{
+	  continue;
+	}
+
+      for (attp = cons->attributes; *attp; attp++)
```
beyondykk9, line 4091:
```suggestion
      for (attp = cons->attributes; *attp && !found; attp++)
```

src/object/class_object.c
```diff
@@ -4045,7 +4064,75 @@ classobj_find_cons_index2_col_type_list (SM_CLASS_CONSTRAINT * cons, OID * root_
...
+	  for (idx = 0, namep = att_names; *namep && idx < cols_non_func; namep++, idx++)
+	    {
+	      if (intl_identifier_casecmp ((*attp)->header.name, *namep) == 0)
+		{
+		  found = true;
+		  break;
+		}
+	    }
+
+	  if (found == false)
```
beyondykk9, line 4104: it can be removed by code modification according to the above suggestion.

src/object/deduplicate_key.c
```diff
@@ -0,0 +1,497 @@
...
+  for (level = DEDUPLICATE_KEY_LEVEL_MIN; level <= DEDUPLICATE_KEY_LEVEL_MAX; level++)
+    {
+      att_id = MK_DEDUPLICATE_KEY_ATTR_ID (level);
+      idx = LEVEL_2_IDX (level);
+      st_or_atts[idx].id = att_id;
+      st_or_atts[idx].domain = get_reserved_index_attr_domain_type (level);
+      st_or_atts[idx].type = st_or_atts[idx].domain->type->id;
+    }
+
+#ifndef NDEBUG
```
beyondykk9, line 78: need comment for the reason of debugging

src/object/deduplicate_key.c
```diff
@@ -0,0 +1,497 @@
...
+				     int deduplicate_key_col_pos, int nnames)
+{
+  int func_no_args = 0;
+
+  assert (!DB_IS_CONSTRAINT_UNIQUE_FAMILY (ctype));
+  if (ctype != DB_CONSTRAINT_FOREIGN_KEY)
+    {
+      assert (asc_desc != NULL);
+    }
+
```
beyondykk9: need assert (nnames >= 1) or assert (nnames > 0)

src/object/deduplicate_key.c
```diff
@@ -0,0 +1,497 @@
...
+      deduplicate_key_col_pos = -1;
+    }
+}
+
+void
+dk_create_index_level_adjust (const PT_INDEX_INFO * idx_info, char **attnames, int *asc_desc,
+			      int *attrs_prefix_length, SM_FUNCTION_INFO * func_index_info, int nnames, bool is_reverse)
+{
+  int deduplicate_key_col_pos;
+
```
beyondykk9, line 334: doesn't need to check nnames value to have valid range?

src/object/class_object.c
ctshim: It was explained verbally.
The code remains as is.

src/object/class_object.c
ctshim: It was explained verbally.
The code remains as is.

src/base/object_representation_sr.c
ctshim: To reach this point, the inside of the set must have at least the following structure.
{ btid, **reserved_index_attrID**, asc_desc, [ **attrID**, asc_desc ]+, {fk_info} or {key prefix length}, status, comment}
That is, the size of this set must be 8 or more, and the 3rd position will be attrID.
Since the first position is reserved_index_attrID, which is virtual information, the third value must be read to obtain actual column information.
Comments will be reinforced.

src/object/deduplicate_key.c
ctshim: Check whether the macro that creates or extracts the attribute name and ID for the deduplicate_key for the specified level operates normally.

I will add a comment.

src/object/deduplicate_key.c
ctshim: OK, I will add assert().

src/object/deduplicate_key.c
ctshim: It is unlikely that the value of nnames is unusual when this function is called.
But let's add the assert() statement.

beyondykk9 (approved):

hyunikn (approved):

src/base/system_parameter.c
```diff
@@ -6228,7 +6250,44 @@ SYSPRM_PARAM prm_Def[] = {
...
+   (void *) &prm_use_deduplicate_key_mode_oid_default,
+   (void *) &PRM_USE_DEDUPLICATE_KEY_MODE_OID,
+   (void *) NULL, (void *) NULL,
+   (char *) NULL,
+   (DUP_PRM_FUNC) NULL,
+   (DUP_PRM_FUNC) NULL},
+  {PRM_ID_DEDUPLICATE_KEY_MODE,
+   PRM_NAME_DEDUPLICATE_KEY_MODE,
+   (PRM_FOR_CLIENT | PRM_USER_CHANGE | PRM_FOR_SESSION),
+   PRM_KEYWORD,
```
hgryoo: PRM_BOOLEAN?

please refer to the `boolean_words` at line 6259.

ctshim: PRM_ID_USE_DEDUPLICATE_KEY_MODE_OID_TEST is PRM_BOOLEAN
PRM_ID_DEDUPLICATE_KEY_MODE                          is PRM_KEYWORD.

I'm not sure exactly what you're talking about.

hgryoo: PRM_BOOLEAN also accepts "on" and "off". It seems that this parameter does not have to be the keyword type parameter.

ctshim: This doesn't seem appropriate because it doesn't accept formats like "deduplicate_key yes" or "deduplicate_key 1".
It would be weird if it looked something like "deduplicate_key 1(16)"

hgryoo: Because it is a system parameter, not a SQL syntax, So I thought we don't need to define the same key values. It is up to you.

ctshim: I just thought so too.
We will reflect your opinion.
Thank you.

src/transaction/boot_cl.c
```diff
@@ -5653,7 +5653,14 @@ boot_define_view_index (void)
...
 	  "[i].[class_of].[class_name] AS [class_name], "
 	  "CAST ([i].[class_of].[owner].[name] AS VARCHAR(255)) AS [owner_name], " /* string -> varchar(255) */
+#if defined(SUPPORT_DEDUPLICATE_KEY_MODE)
+          "NVL2((SELECT 1 FROM _db_index_key [k] WHERE [k].index_of.class_of = [i].class_of "
+                                                   "AND [k].index_of.index_name = [i].[index_name] "
+                                                   "AND [k].key_attr_name LIKE " DEDUPLICATE_KEY_ATTR_NAME_LIKE_PATTERN
+                 "), ([i].[key_count] - 1), [i].[key_count]) AS [key_count], "
+#else
 	  "[i].[key_count] AS [key_count], "
+#endif
```
youngjinj: How about changing like below?
```suggestion
#if defined(SUPPORT_DEDUPLICATE_KEY_MODE)
	  "NVL2 ("
	      "("
		"SELECT 1 "
		"FROM "
		  /* CT_INDEXKEY_NAME */
		  "[%s] [k] "
		"WHERE "
		  "[k].index_of.class_of = [i].class_of "
		  "AND [k].index_of.index_name = [i].[index_name] "
		  "AND [k].key_attr_name LIKE " DEDUPLICATE_KEY_ATTR_NAME_LIKE_PATTERN
	      "), "
	      "([i].[key_count] - 1), "
	      "[i].[key_count]"
	    ") AS [key_count], "
#else
	  "[i].[key_count] AS [key_count], "
#endif
```

The code below must be added when naming the system table.
```
#if defined(SUPPORT_DEDUPLICATE_KEY_MODE)
	CT_INDEXKEY_NAME,
#endif
	CT_INDEX_NAME,
	AU_USER_CLASS_NAME,
	AU_USER_CLASS_NAME,
	CT_CLASSAUTH_NAME,
	AU_USER_CLASS_NAME);
```

src/transaction/boot_cl.c
```diff
@@ -5790,6 +5797,9 @@ boot_define_view_index_key (void)
 	  /* CT_INDEXKEY_NAME */
 	  "[%s] AS [k] "
 	"WHERE "
+#if defined(SUPPORT_DEDUPLICATE_KEY_MODE)
```
youngjinj, line 5813: How about changing like below?
In issue CBRD-24419, it was important to use equally spaced indentation.
And I think the common predicate for all system views should come in last order.
```
#if defined(SUPPORT_DEDUPLICATE_KEY_MODE)
	  "[k].[key_attr_name] NOT LIKE " DEDUPLICATE_KEY_ATTR_NAME_LIKE_PATTERN " "
	  "AND ("
#endif
	      "{'DBA'} SUBSETEQ ("
		  "SELECT "
		    "SET {CURRENT_USER} + COALESCE (SUM (SET {[t].[g].[name]}), SET {}) "
		  "FROM "
		    /* AU_USER_CLASS_NAME */
		    "[%s] AS [u], TABLE ([u].[groups]) AS [t] ([g]) "
		  "WHERE "
		    "[u].[name] = CURRENT_USER"
		") "
	      "OR {[k].[index_of].[class_of].[owner].[name]} SUBSETEQ ("
		  "SELECT "
		    "SET {CURRENT_USER} + COALESCE (SUM (SET {[t].[g].[name]}), SET {}) "
		  "FROM "
		    /* AU_USER_CLASS_NAME */
		    "[%s] AS [u], TABLE ([u].[groups]) AS [t] ([g]) "
		  "WHERE "
		    "[u].[name] = CURRENT_USER"
		") "
	      "OR {[k].[index_of].[class_of]} SUBSETEQ ("
		  "SELECT "
		    "SUM (SET {[au].[class_of]}) "
		  "FROM "
		    /* CT_CLASSAUTH_NAME */
		    "[%s] AS [au] "
		  "WHERE "
		    "{[au].[grantee].[name]} SUBSETEQ ("
			"SELECT "
			  "SET {CURRENT_USER} + COALESCE (SUM (SET {[t].[g].[name]}), SET {}) "
			"FROM "
			  /* AU_USER_CLASS_NAME */
			  "[%s] AS [u], TABLE ([u].[groups]) AS [t] ([g]) "
			"WHERE "
			  "[u].[name] = CURRENT_USER"
		      ") "
		    "AND [au].[auth_type] = 'SELECT'"
#if defined(SUPPORT_DEDUPLICATE_KEY_MODE)
		")"
	    ")",
#else
		")",
#endif
```

src/parser/show_meta.c
```diff
@@ -445,6 +445,9 @@ metadata_of_index_capacity (SHOW_ONLY_ALL flag)
     {"Btid", "varchar(64)"},
     {"Num_distinct_key", "int"},
     {"Total_value", "bigint"},
+#if defined(SUPPORT_DEDUPLICATE_KEY_MODE)
+    {"Deduplicate_distinct_key", "int"},
```
youngjinj, line 449: I think 'Deduplicate_distinct_key' should be output right after 'Num_distinct_key'.

And this may be trivial, but 'Deduplicate_distinct_key' does not contain the word 'Num' .
Consider changing the names for consistency with other columns.

e.g.
  - Num_deduplicate_distinct_key
  - Num_distinct_key_with_deduplicate

util/CMakeLists.txt
```diff
@@ -210,6 +210,7 @@ if(WIN32)
 endif(WIN32)
 add_executable(cub_javasp ${JAVASP_SERVER_SOURCES})
 target_link_libraries(cub_javasp LINK_PRIVATE cubridsa)
+target_include_directories(cub_javasp PRIVATE ${EP_INCLUDES})
```
youngjinj: Is this change related to this issue?

src/compat/db_value_printer.cpp
```diff
@@ -656,3 +656,30 @@ db_sprint_value (const db_value *value, string_buffer &sb)
   db_value_printer printer (sb);
   printer.describe_value (value);
 }
+
+#ifndef NDEBUG
+void
+db_value_print_console (const db_value *value, bool add_newline, char *fmt, ...)
```
youngjinj, line 662: Is there anything I can see in more detail if I run the function below instead of the existing db_fprint_value function?
```
db_fprint_value (stdout , &value)
```

src/base/xserver_interface.h
```diff
@@ -173,7 +173,11 @@ extern bool logtb_has_updated (THREAD_ENTRY * thread_p);
 
 
 extern BTID *xbtree_add_index (THREAD_ENTRY * thread_p, BTID * btid, TP_DOMAIN * key_type, OID * class_oid, int attr_id,
-			       int unique_pk, long long num_oids, long long num_nulls, long long num_keys);
+			       int unique_pk, long long num_oids, long long num_nulls, long long num_keys
+#if defined(SUPPORT_DEDUPLICATE_KEY_MODE)
+			       , int deduplicate_key_attr_pos
```
youngjinj: It's trivial.
The btree_add_index and sbtree_add_index functions use the name `deduplicate_key_pos`,
while xbtree_add_index uses the name `deduplicate_key_attr_pos`.
Is there any reason why they use different names?

src/parser/show_meta.c
ctshim: The order of exposure is at the request of MyungHwan Oh.

src/transaction/boot_cl.c
ctshim: OK,

src/transaction/boot_cl.c
ctshim: OK, thank you

src/compat/db_value_printer.cpp
ctshim: It's not to see more details, but because I want to show it by adding other messages I want to the output.
I'm trying to leave something like where it was printed during debugging.

src/base/xserver_interface.h
ctshim: There is no particular reason.
I will unify it with deduplicate_key_pos.

shparkcubrid (approved):

src/object/deduplicate_key.c
```diff
@@ -0,0 +1,436 @@
+/*
+ * Copyright 2008 Search Solution Corporation
```
hgryoo: This is a new file, only "Copyright 2016 CUBRID Corporation" should exist.

hgryoo: Please check the other files too.

util/CMakeLists.txt
ctshim: Thanks for the check.
It was sorted out with the help of @hgryoo.

src/parser/csql_grammar.y
```diff
@@ -27567,3 +27652,37 @@ pt_ct_check_select (char* p, char *perr_msg)
    sprintf(perr_msg, "Only SELECT statements are supported.");
    return false;
 }
+
+#if defined(SUPPORT_DEDUPLICATE_KEY_MODE)
+
+#include "system_parameter.h"
+static void pt_get_deduplicate_key_mode_level(int mode_level, short* mode, short* level)
```
hgryoo: Including system_paramter.h here seems to make the code complex.
How about refactoring like the following
1) Add the declaration to parser.h
2) Move the body of the pt_get_deduplicate_key_mode_level() in proper file which includes the "parser.h" and "system_parameter.h"

src/object/deduplicate_key.c
ctshim: I will fix the deduplicate_key.h and deduplicate_key.c files.

src/parser/csql_grammar.y
ctshim: First, I will remove only the line of code that includes the "system_parameter.h" file.

hgryoo (approved):

youngjinj (approved):

beyondykk9 (approved):

src/storage/btree.h
```diff
@@ -242,7 +245,9 @@ struct btree_scan
   bool is_btid_int_valid;
   bool is_scan_started;
   bool force_restart_from_root;
-
+#if 1				/* defined(SUPPORT_DEDUPLICATE_KEY_MODE) */
```
beyondykk9, line 248:
```suggestion
#if defined(SUPPORT_DEDUPLICATE_KEY_MODE)
```

ctshim: There are two macros declared in the btree.h file: BTREE_RESET_SCAN and BTREE_INIT_SCAN.
There is a part that is initialized inside this so that the variable is always declared.

This will be adjusted in a future code cleanup for SUPPORT_DEDUPLICATE_KEY_MODE.

ctshim:
change to #4387

PR #3978, opened 2022-12-02
[CBRD-24571] assert() occurs when using DELETE/UPDATE including DBLINK

ctshim:
http://jira.cubrid.org/browse/CBRD-24571

* During DELETE/UPDATE syntax execution that contains DBLINK, an assert() occurs and dies is found.
This is a problem when the specified server name does not exist.

* Test case
```
drop if exists tbl;
create table tbl(id int, val int);

delete b from (select id from dblink(srv, 'select id from tbl') a(id int)) a join tbl b where a.id = b.id;
update (select id from dblink(srv, 'select id from tbl') a(id int)) a inner join tbl b on a.id = b.id set b.val = b.val + 1;


```


beyondykk9 (approved):

shparkcubrid (approved):

hgryoo:
The fix changes the common routine. According to the change, a side-effect could occur. for example, any other error (not related to the DBLINK) may not be handled appropriately at the next step.
```
...
	  else if (table->node_type == PT_DBLINK_TABLE)
	    {
	      assert (spec->info.spec.derived_table_type == PT_DERIVED_DBLINK_TABLE);
	      if (table->info.dblink_table.is_name)
		{
		  if (pt_resolve_dblink_server_name (parser, table) != NO_ERROR)
		    {
		      return;
		    }
		}
...

	  /* must bind any expr types in table. pt_bind_types requires it. */
	  save_donot_fold = bind_arg->sc_info->donot_fold;	/* save */
	  bind_arg->sc_info->donot_fold = true;	/* skip folding */
	  table = pt_semantic_type (parser, table, bind_arg->sc_info);
	  bind_arg->sc_info->donot_fold = save_donot_fold;	/* restore */
	  if (table != NULL)
	    {
	      spec->info.spec.derived_table = table;
	      /* table spec passed bind_names and semantic_type */
	      pt_bind_spec_attrs (parser, spec);
	    }
	  else
	    {
	      assert (pt_has_error (parser));    // **A**
	    }
```
In pt_bind_scope(), if pt_semantic_type() sets an error, the table is NULL. and the pt_has_error() is assumed by assert (name_resolution.c 1028~1041, above code). IMO, if resolving a dblink server's name is not successful, assertion failure should be occurred at the "**A**" above. maybe error handling for dblink table needs to be placed here.

hyunikn (approved):

hgryoo (approved):

beyondykk9 (approved):

shparkcubrid (approved):

hyunikn (approved):

hyunikn (approved):

youngjinj (approved):

PR #3988, opened 2022-12-05
[LETS-605] unit test for atomic replication helper

cristiarg:
http://jira.cubrid.org/browse/LETS-605

Added unit test for `atomic_replication_helper` class which does the "heavy lifting" of atomic replication on PTS.

Implementation:
- the overall idea is to restrain the test as much as possible to the helper class only
- for this, the actual "apply the redo function" was been "hijacked" by means of a newly added dummy log record type `LOG_DUMMY_UNIT_TESTING`
- the definition of a test is with the class `test_spec_type` which mainly does 3 things:
  - holds a vector of log records - `log_record_spec_type` - which should actually define one - or more - sequences of atomic records
  - executes the atomic sequence of records by feeding the definitions to a supplied `cublog::atomic_replication_helper` instance - `test_spec_type::execute`
  - mocks the fix/unfix of page buffer pages - `test_spec_type::fix_page` and `test_spec_type::unfix_page` - the actual page buffer pages are not needed since effectively applying the log records is "hijacked" (see above) but their presence is asserted all over the place so the unit test must supply non-null pointers
- page buffer and log page buffer functions are also mocked with the help of the implementation in the `test_spec_type`
- one important aspect is the mocking of the log page layout that allows the redo-execute call-path to read and actually fake the redo - see implementation of the mocked `logpb_fetch_page` and the corresponding modification in `atomic_log_sequence::atomic_log_entry::apply_log_redo`


hornetmj (approved):

unit_tests/log_repl/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
+#
+#  Copyright 2008 Search Solution Corporation
```
joohok, line 2:
```suggestion
```
This can be removed

joohok (approved):

wrlawodms (approved):

PR #3995, opened 2022-12-09
[CBRD-24580] Set the request from consumer to producer only by the consumer thread

joohok:
http://jira.cubrid.org/browse/CBRD-24580

Purpose

There are two threads. CDC log consumer and producer. 
Consumer and producer pass requests to each other through cdc_Gl. 
However, the problem occurred when the producer also modified the request transmitted from the consumer to the producer. (`cdc_Gl.producer.request`)

`cdc_Gl.producer.request` is only allowed to be set by the consumer, who calls `cdc_wakeup_producer`, `cdc_pause_producer`, and `cdc_kill_producer`.

### DIFF
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -11010,7 +11010,6 @@ cdc_loginfo_producer_execute (cubthread::entry & thread_ref)
 	  cdc_log ("cdc_loginfo_producer_execute : cdc_Gl.producer.state is in CDC_PRODUCER_STATE_WAIT ");
 
 	  cdc_Gl.producer.state = CDC_PRODUCER_STATE_WAIT;
-	  cdc_Gl.producer.request = CDC_REQUEST_PRODUCER_NONE;
 
 	  pthread_mutex_lock (&cdc_Gl.producer.lock);
 	  pthread_cond_wait (&cdc_Gl.producer.wait_cond, &cdc_Gl.producer.lock);
@@ -13900,6 +13899,8 @@ void
 cdc_wakeup_producer ()
 {
   cdc_log ("cdc_wakeup_producer : consumer request the producer to wakeup");
+
+  cdc_Gl.producer.request = CDC_REQUEST_PRODUCER_NONE;
 
   pthread_cond_signal (&cdc_Gl.producer.wait_cond);
 }


src/transaction/log_manager.c
```diff
@@ -11010,7 +11010,6 @@ cdc_loginfo_producer_execute (cubthread::entry & thread_ref)
 	  cdc_log ("cdc_loginfo_producer_execute : cdc_Gl.producer.state is in CDC_PRODUCER_STATE_WAIT ");
 
 	  cdc_Gl.producer.state = CDC_PRODUCER_STATE_WAIT;
-	  cdc_Gl.producer.request = CDC_REQUEST_PRODUCER_NONE;
```
joohok, line 11013: `cdc_kill_producer ()` is called when server is shutting down. 

In `cdc_kill_producer ()`, consumer set `cdc_Gl.producer.request` to `CDC_REQUEST_PRODUCER_TO_BE_DEAD`, and send signal to producer. 

But at this line, `CDC_REQUEST_PRODUCER_TO_BE_DEAD` is overwritten by `CDC_REQUEST_PRODUCER_NONE`. 

So, the producer is not killed by the consumer and starts to make log info entry (fetch log pages starting from the LSA set by the consumer)

Following scenario can be problem :
1. start server with `supplemental_log=1` configuration
2. producer daemon is started and waits until the consumer send the signal
3. without any requests from consumer, producer has no information about LSA to start extraction.
4. while shutting down server, consumer sends signal and set `cdc_Gl.producer.request` (`cdc_kill_producer ()`)

> 3. without any requests from consumer, producer has no information about LSA to start extraction.

Because of this, it will be crashed when it tries to fetch log pages using LSA (to start extraction)

hornetmj: Do you mean that the cdc_Gl.producer.request should be set by the caller of the 'pthread_cond_signal (&cdc_Gl.producer.wait_cond)'? Right? The setter are the consumer thread or main thread.

hornetmj: And the cdc_kill_producer() is called between 11009 and 11012. So, the cdc_Gl.producer.request value is overwrited from CDC_REQUEST_PRODUCER_TO_BE_DEAD to CDC_REQUEST_PRODUCER_NONE. Right?

joohok: Yes, perfect.

hornetmj (approved):

PR #3996, opened 2022-12-09
[LETS-458] Revert the reverted change for LETS-458

wrlawodms:
http://jira.cubrid.org/browse/LETS-458

Revert the reverted PR, #3819. The PR caused the regression issue, [LETS-560](http://jira.cubrid.org/browse/LETS-560). However, it turns out to be from other ones and now it's been fixed so just I'm doing re-apply the previous PR, #3819.
 
Some details: 
The desynchronization in the LETS-560 happens when the replicator on a PTS tries to fix a deallocated page from PS, but the replicator doesn't fix a page anymore if it doesn't exist on the PTS (http://jira.cubrid.org/browse/LETS-554). And the deallocated page is always not on the PTS since http://jira.cubrid.org/browse/LETS-558.


joohok (approved):

joohok:
build fail due to unit test. 
```
[ 12%] Building CXX object unit_tests/log_repl/CMakeFiles/test_log_repl_atomic_helper.dir/test_main_log_repl_atomic_helper.cpp.o
/home/unit_tests/log_repl/test_main_log_repl_atomic_helper.cpp:20:10: fatal error: catch2/catch.hpp: No such file or directory
 #include "catch2/catch.hpp"
          ^~~~~~~~~~~~~~~~~~
compilation terminated.
```

hornetmj (approved):

PR #3998, opened 2022-12-12
[CBRD-24525] Support COMMIT/ROLLBACK in stored routine

hgryoo:
http://jira.cubrid.org/browse/CBRD-24525

**Implementation**
- For backward compatibility, commit or rollback will perform by default connection's property
    - e.g.) jdbc:default:connection::?transaction_control=false
    - The default value of the transaction_control will be false in this PR. It means the Java SP will ignore commit or abort as prior versions.
    - In PL/CSQL compiler, transaction_control=yes is added.
    - To make it consistent for both, It will be handled in the next issue.
- A new system paramter, `pl_transaction_control` is added.
- either the system paramter is on or the default connection's property has transaction_control=true, commit() and rollback() work
- The TCL functions will clear statement objects of servet-side connection in the corresponding Context of Java SP. And It also clears query handlers in method_callback after calling `db_[commit|rollback]_transaction()`.


src/jsp/com/cubrid/plcsql/compiler/ast/Unit.java
```diff
@@ -97,9 +97,14 @@ public String toJavaCode() {
...
-                            "%'AUTONOMOUS-TRANSACTION'%",
-                            autonomousTransaction ? "?autonomous_transaction=true" : "");
+                    tmplGetConn
+                            .replace(
+                                    "%'TRANSACTION-CONTROL'%",
+                                    "transaction_control=true") // In PL/CSQL, enable TCL as a
+                            // default
+                            .replace(
+                                    "%'AUTONOMOUS-TRANSACTION'%",
+                                    autonomousTransaction ? "autonomous_transaction=false" : "");
```
hyunikn: Is the change to "autonomous_transaction=false" intentional?

hgryoo: Yes, we don't support this feature yet.

hyunikn (approved):

src/jsp/com/cubrid/plcsql/compiler/ast/Unit.java
hyunikn: OK.

PR #4001, opened 2022-12-14
[LETS-583] check inconsistent fix result in atomic seq

cristiarg:
http://jira.cubrid.org/browse/LETS-583

Implements a guard against subsequent different results of a call to fix a page:
- if a first call to "fix_page" fails (because the page is not in internal page buffer of the passive transaction server)
- a subsequent call will also be considered as failed

This ensures that the page will not be inconsistently replicated as the following scenario can happen.

Suppose that the replication must replicate an atomic sequence for which there are 2 log records for the same page (VPID1):
  - LSA(i-1)
  - LSA(i)

Chronologically, in a corner case scenario, the following sequence of actions is possible:
- the replication processes the first log record LSA(i-1) and attempts to fix the page VPID1
  - the replication fails to fix the page VPID1 because the page is not in the PTS's page buffer nor in transit from PS to PTS
- at this point the replication reports:
  - m_processed_lsa = LSA(i-2)
  - m_redo_lsa = LSA(i-1)
  - this version supposes that the currently executing instruction on the replication thread is somewhere within function `atomic_replicator::redo_upto` BEFORE the synchronized sequence where `m_processed_lsa` and `m_redo_lsa` are advanced
- a client transaction also tries to fix page VPID1 and succeeds:
  - the page is requested to Page Server to be retrieved having LSA(i-2) already processed (see implementation for `atomic_replicator::get_highest_processed_lsa`)
  - the PS waits for its own replication to have already passed LSA(i-2) (so LSA(i-2) is already applied to its respective page but LSA(i-1) is not guaranteed to have been applied to VPID1)
  - at this point, the PTS client transaction fixes for read page VPID1 with applied LSA(i-2) but not with LSA(i-1) (in the worst case scenario)
- the client transaction does its job and unfixes the page VPID1
- the replication thread steps in and attempts to fix page VPID1 to apply the log at LSA(i)
  - but the page does not contain the modification from LSA(i-1)
- this is the window of opportunity that must be avoided; the window is very small, as it would require that a client transaction would request a page, retrieve it from page server in a smaller time frame than it takes the atomic replication thread to process an atomic sequence; but cannot be ruled out as long as the atomic replication thread also receives the prior log lists info over the same wire from the Page Server
- and this is what this current check guards against; for now, the check is to set a fatal error to see whether this ever happens

Also extended and adapted the unit test to cover the extra logic and expected error setting.

Note: this PR is on top of #3999, so should be merged after that one

Later Note: During discussions for this PR it became obvious that there is a better option, one which even might render this check superfluous. It will be investigated as a separate issue.


hornetmj:
@cristiarg
In the 'see call-path:' on the description of this PR the replicator::get_highest_processed_lsa() is replaced by the atomic_replicator::get_highest_processed_lsa(). right? The replicator::get_highest_processed_lsa() just returns MAX_LSA.

cristiarg:
@hornetmj  On PTS `replicator::get_highest_processed_lsa` is actually `atomic_replicator::get_highest_processed_lsa` by means of inheritance/overriding. I will update the description.

wrlawodms:
@cristiarg 
> see call-path: -> pgbuf_read_page_from_file_or_page_server -> log_lsa passive_tran_server::get_highest_processed_lsa -> atomic_replicator::get_highest_processed_lsa
and the way the value of "m_redo_lsa" is updated in atomic_replicator::redo_upto

1. If I understand it correctly, with `m_redo_lsa`, when processing the log record of LSA(i), the page server returns a page applied with the log record of LSA(i), not LSA(i-1). Because a PTS requests a page with LSA(i) and PS waits until LSA(i) is applied.
2. However, since #3996, `m_redo_lsa` is replaced with `m_processed_lsa`, and the case you described is true. This has been merged quite recently.
3. I guess we can handle the issue by setting `m_processed_lsa` earlier than now, specifically, before fixing a data page for processing a redo record.

With (3), while applying a log record of LSA(i) and after being done with it, the client will request a page applied up to LSA(i). And we can avoid the problematic situation you described. What do you think?

cristiarg:
@wrlawodms 

> If I understand it correctly, with m_redo_lsa, when processing the log record of LSA(i), the page server returns a page applied with the log record of LSA(i), not LSA(i-1). Because a PTS requests a page with LSA(i) and PS waits until LSA(i) is applied.

You are correct. Before the fix for #3996, my explanation was wrong. I will update it adding details for `m_processed_lsa`.

> I guess we can handle the issue by setting m_processed_lsa earlier than now, specifically, before fixing a data page for processing a redo record.

I don't understand this. As it is now, the semantics is correct:
- `m_redo_lsa` is the "to be processed log record"
- `m_processed_lsa` is the "has been processed (ie: either: skipped as the page is not present, not of interest for replication, or the page has been fixed)

Do you suggest:
- setting the value for `m_processed_lsa` earlier such that `m_processed_lsa` and `m_redo_lsa` have the same value for a brief moment?
  - I thing this would be incorrect.
- setting the value of both of them earlier?
  - I think this is also incorrect (as it would be a revert to the previous state when there was only `m_redo_lsa`).

cristiarg:
@wrlawodms

I have thought more on this and drafted the following scenarios out of which Scenario 1 is problematic.
I think it is, in essence the same as described in http://jira.cubrid.org/browse/LETS-560 .
Please give it a thought and let me know your opinion.

Version 1:
-----------
Suppose that the replication must replicate an atomic sequence for which there are 2 log records for the same page (VPID1):
  - LSA(i-1)
  - LSA(i)

Chronologically, in a corner case scenario, the following sequence of actions is possible:
- the replication processes the first log record LSA(i-1) and attempts to fix the page VPID1
  - the replication fails to fix the page VPID1 because the page is not in the PTS's page buffer
- at this point the replication reports:
  - m_processed_lsa = LSA(i-2)
  - m_redo_lsa = LSA(i-1)
  - this version supposes that the currently executing instruction on the replication thread is somewhere within function `atomic_replicator::redo_upto` BEFORE the synchronized sequence where `m_processed_lsa` and `m_redo_lsa` are advanced
- a client transaction also tries to fix page VPID1 and succeeds:
  - the page is requested to Page Server to be retrieved having LSA(i-2) already processed (see implementation for `atomic_replicator::get_highest_processed_lsa`)
  - the PS waits for its own replication to have already passed LSA(i-2) (so LSA(i-2) is already applied to its respective page but LSA(i-1) is not guaranteed to have been applied to VPID1)
  - at this point, the PTS client transaction fixes for read page VPID1 with applied LSA(i-2) but not with LSA(i-1) (in the worst case scenario)
- the client transaction does its job and unfixes the page VPID1
- the replication thread steps in and attempts to fix page VPID1 to apply the log at LSA(i)
  - but the page does not contain the modification from LSA(i-1)
- this is the window of opportunity that must be avoided


Version 2:
-----------
Suppose that the replication must replicate an atomic sequence for which there are 2 log records for the same page (VPID1):
  - LSA(i-1)
  - LSA(i)

Chronologically, in a corner case scenario, the following sequence of actions is possible:
- the replication processes the first log record LSA(i-1) and attempts to fix the page VPID1
  - the replication fails to fix the page VPID1 because the page is not in the PTS's page buffer
- at this point the replication reports:
  - m_processed_lsa = LSA(i-1)
  - m_redo_lsa = LSA(i) (already moved on to forward LSA)
  - this version supposes that the currently executing instruction on the replication thread is somewhere within function `atomic_replicator::redo_upto` AFTER the synchronized sequence where `m_processed_lsa` and `m_redo_lsa` are advanced
- a client transaction also tries to fix page VPID1 and succeeds:
  - the page is requested to Page Server to be retrieved having LSA(i-1) already processed (see implementation for `atomic_replicator::get_highest_processed_lsa`)
  - the PS waits for its own replication to have already passed LSA(i-1) (so LSA(i-1) is already applied onto page VPID1 but LSA(i) is not guaranteed to have been applied to VPID1)
  - at this point, the PTS client transaction fixes for read page VPID1 with applied LSA(i-1) but not with LSA(i) (in the worst case scenario)
- the client transaction does its job and unfixes the page VPID1
- the replication thread steps in and attempts to fix page VPID1 to apply the log at LSA(i)
  - now the page VPID1 properly contains the redo added by LSA(i-1) log record and the replicator could move on and redo the modification at LSA(i)

wrlawodms:
@cristiarg 
> setting the value for m_processed_lsa earlier such that m_processed_lsa and m_redo_lsa have the same value for a brief moment?

Yes, this is what I suggested. The sequence in `redo_upto()`  will be:
1. Read the log page and the log record (`m_redo_context.m_reader.set_lsa_and_fetch_page`)
2. Advance `m_proccessed_lsa`. It is the same with `m_redo_lsa` for a moment since here.
3. Read the log page and the log record. And apply the record.
4. Advance `m_redo_lsa`.

I assume the thing is that a client in PTS can fetch a page with an LSA preceding a skipped log record by the replicator. And this happens in the window where the LSA a client requests with (`m_processed_lsa`) hasn't advanced yet, but the replicator already tries to fix the page and fails. If `m_processed_lsa` is advanced before the replicator tries to fix the page, the client would have requested a page with the advanced LSA and it doesn't happen.
- If the page hasn't been on the PTS, the client requests the page with advanced LSA, and all log records have been applied in PS.
- If the page has been on the PTS, the client fixes the page and unfixes it. Then, the replicator applies a log record.

The way I see it, the version 1 scenario is addressed with this solution. 

Could you tell me the reason it's incorrect?

---

> setting the value of both of them earlier?
I think this is also incorrect (as it would be a revert to the previous state when there was only m_redo_lsa).

What I suggested was the above, but I think we can set both values before fixing a page. I don't think it's a revert to the previous one. Previously, PTS can request a data page with the **next of the last** applied LSA. It means if the last applied LSA is the very last log generated by ATS, we can't fetch a page (->hang). However, for now, `m_processed_lsa` points to the **last** applied LSA, we can retrieve a page with it. However, I'm not 100% sure we can do it, because the `m_redo_lsa` would be used in other contexts with an assumption that it's set after applying the log record of it.

cristiarg:
@wrlawodms 

> Could you tell me the reason it's incorrect?

On a second thought, I think this might work.
I will choose the conservative approach on only advancing `m_processed_lsa` before the page is actually fixed and the log applied.
The main drawback is that we must introduce a new critical section. I will do this by adding a new mutex specifically for `m_processed_lsa`. The performance would be impacted to some degree, I htink.

The only inconsistency (although it is not really an inconsistency) is:
- the replicator is about to replicate LSA(i)
- a client transaction wants to fix the page, it will ask for `m_processed_lsa` which will either be LSA(i-1) or LSA(i) (depending of the moment when the client transaction asks for the value)
- if the page is in the PTS's local page buffer and the `m_processed_lsa` reported by the replicator is at LSA(i) (that is: the `m_processed_lsa` is retrieved after it has been updated - when it's value is equal with `m_redo_lsa`) then the client transaction will actually obliviously use a page which is not at LSA(i) but at LSA(i-1).

As said, this is only conceptually inconsistent.



I will give your suggestion a try. I have added issue http://jira.cubrid.org/browse/LETS-630 that pings back to relevant comments in this PR.

cristiarg:
@wrlawodms 

> I don't think it's a revert to the previous one.

Indeed.

> However, I'm not 100% sure we can do it, because the m_redo_lsa would be used in other contexts with an assumption that it's set after applying the log record of it.

Indeed. The other contexts are:
- `wait_past_target_lsa` - used in "page ahead of replication" scenarios
- `wait_replication_finish_during_shutdown` (this is marginally used)

src/transaction/atomic_replication_helper.cpp
```diff
@@ -974,52 +981,171 @@ namespace cublog
...
+	     *      server and the state of the page's replication is at LSA(i-1) and the passive
+	     *      transaction server's replication would not have applied the log record at LSA(i)
+	     *    - if the atomic sequence contains a new log record for the same page - LSA(i+1) -
+	     *      the replication will fix the page and apply the log record LSA(i+1)
+	     *      but the page will not have contained the log record at LSA(i)
+	     *  - the window of opportunity for this to happen is quite small; for this to happen
+	     *    it would mean that a client transaction would request a page, and retrieve from
+	     *    page server in smaller frame of time than it takes the atomic replication thread
+	     *    to process an atomic sequence; but cannot be ruled out entirely
+	     */
```
wrlawodms: I think it should be revised with m_processed_lsa instead of m_redo_lsa.

src/transaction/atomic_replication_helper.cpp
```diff
@@ -1077,9 +1203,10 @@ namespace cublog
       {
 	const page_ptr_info &info = pair.second;
 	written = snprintf (buf_ptr, (size_t)left, "  m_vpid = %d|%d  rcvindex = %s"
-			    "  page_p = %p  watcher_p = %p  ref_cnt = %d\n",
+			    "  page_p = %p  watcher_p = %p  ref_cnt = %d  fixed=%d\n",
```
wrlawodms:
```suggestion
			    "  page_p = %p  watcher_p = %p  ref_cnt = %d  fixed = %d\n",
```

wrlawodms (commented):
> if the page is in the PTS's local page buffer and the m_processed_lsa reported by the replicator is at LSA(i) (that is: the m_processed_lsa is retrieved after it has been updated - when it's value is equal with m_redo_lsa) then the client transaction will actually obliviously use a page which is not at LSA(i) but at LSA(i-1).

Agreed. I guess we have the same understanding :).

With http://jira.cubrid.org/browse/LETS-630, I think the case this PR wants to filter out is a valid case. A replicator can succeed to fix a page that failed previously in the same atomic operation.  Right?
1) the replicator fails to fix since it's not on transit or the buffer pool 
2) a client fix the page
3) the replicator succeeds to fix the page, which has the change of (1).

Do I get it correct? Then, do we need to block this case? Or are we going to remove it later after some tests?

---
I guess we can check in another way assuming the LETS-630 is done. If it fails to fix a page, it stores the page's info like the code in this PR and, plus, the LSA of a log record then. If it succeeds to fix later in the same atomic sequence, we can make sure the previous log record has been applied by comparing the saved LSA and the page LSA of the page. If the page LSA is   smaller than the saved LSA, it means the previous one hasn't applied and we have a problem. What do you think?

src/transaction/atomic_replication_helper.cpp
cristiarg: Done

src/transaction/atomic_replication_helper.cpp
cristiarg: Reformulated the entire comment with the relevant problematic scenario described above.

cristiarg:
@wrlawodms 

> if the page is in the PTS's local page buffer and the m_processed_lsa reported by the replicator is at LSA(i) (that is: the m_processed_lsa is retrieved after it has been updated - when it's value is equal with m_redo_lsa) then the client transaction will actually obliviously use a page which is not at LSA(i) but at LSA(i-1).

Thought more about it and, I was incorrect.
On PTS, the call path is: `pgbuf_fix` -> `pgbuf_claim_bcb_for_fix` -> `pgbuf_read_page_from_file_or_page_server` -> (get highest processed LSA) -> `pgbuf_request_data_page_from_page_server`
So, as long as the page is in the PTS's page buffer, there's no incorrect assumption regarding the "age" of the page.

> Do I get it correct? Then, do we need to block this case? Or are we going to remove it later after some tests?

I think it's better to add an extra check as suggested by you further - the LSA check. At least temporary. I pinged back with comment http://jira.cubrid.org/browse/LETS-630?focusedCommentId=4763112&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-4763112 to your comment. Thank you.

PR #4002, opened 2022-12-15
[CBRD-24584] Improved to minimize the impact of query execution speed when setting sql_trace_slow

beyondykk9:
http://jira.cubrid.org/browse/CBRD-24584

When setting the sql_trace_slow value in cubrid.conf to a positive number other than -1, the query performance is highly affected compared to 9.x.

The reason for the increased impact on query performance compared to 9.x is that statdump information is calculated before starting a query and after finishing query execution.

Therefore, it is necessary to minimize the impact on query performance by improving it to calculate only the information necessary for setting sql_trace_slow.

### DIFF
--- a/src/base/perf_monitor.c
+++ b/src/base/perf_monitor.c
@@ -983,8 +983,8 @@ perfmon_server_get_stats (THREAD_ENTRY * thread_p)
 }
 
 /*
- *   xperfmon_server_copy_stats - Copy recorded server statistics for the current
- *				  transaction index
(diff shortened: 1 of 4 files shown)


ctshim (approved):

hyunikn (approved):

shparkcubrid (approved):

hgryoo (approved):

youngjinj (approved):

beyondykk:
http://jira.cubrid.org/browse/CBRD-24584

PR #4004, opened 2022-12-15
[CBRD-24585] Explicit commit/rollback statement causes invalid XASL error in SP's query

hgryoo:
http://jira.cubrid.org/browse/CBRD-24585

- Fix SP's statement handler and statement handler cache are not cleared when commit and rollback statements were explicitly executed. It can cause referencing an invalid XASL node that has already been removed from the XASL cache
- Add defense code for query execution failure situation by SP

### DIFF
--- a/src/method/method_callback.cpp
+++ b/src/method/method_callback.cpp
@@ -168,12 +168,26 @@ namespace cubmethod
       {
 	if (handler->execute (request) == NO_ERROR)
 	  {
+	    int error = handler->execute (request);
+	  }
+	if (error == NO_ERROR)
+	  {
 	    /* register query_id for out resultset */
 	    const cubmethod::query_result &qresult = handler->get_result();
 	    if (qresult.stmt_type == CUBRID_STMT_SELECT)
 	      {
 		uint64_t qid = (uint64_t) handler->get_execute_info ().qresult_info.query_id;
 		m_qid_handler_map[qid] = request.handler_id;
+	      }
+	  }
+	else
+	  {
+	    else
+	      {
+		/* XASL cache is not found */
+		m_error_ctx.clear ();
+		handler->prepare_retry ();
+		handler->execute (request);
 	      }
 	  }
       }
--- a/src/method/method_callback.cpp
+++ b/src/method/method_callback.cpp
@@ -166,10 +166,7 @@ namespace cubmethod
       }
     else
       {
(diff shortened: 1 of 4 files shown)


src/method/method_callback.cpp
```diff
@@ -176,6 +177,13 @@ namespace cubmethod
 		m_qid_handler_map[qid] = request.handler_id;
 	      }
 	  }
+	else
```
ctshim, line 180: Is the only reason why handler->execute() fails is if XASL cache is not found?
It would be better if we could accurately determine the cause of the failure with the error code.

ctshim (approved):

src/method/method_callback.cpp
hgryoo: This is a safeguard in case the query execution fails for any reason. Even if there is no code to clear all statement handlers for SP queries (`cubmethod::get_callback_handler ()->free_query_handle_all (true);`), this safeguard fixes the problem.

youngjinj (approved):

shparkcubrid (approved):

src/method/method_query_handler.cpp
```diff
@@ -190,6 +190,16 @@ namespace cubmethod
     return error;
   }
 
+  int
+  query_handler::prepare_retry ()
+  {
+    if (is_prepared ())
```
beyondykk9, line 196: is_prepared means "was prepared" ?

hgryoo: In the query handler, is_prepared() means that the statement handler is created by calling `db_compile_statement()`.

src/method/method_callback.cpp
ctshim: OK, Thank you.

src/method/method_query_handler.cpp
hgryoo: To be clear, the meaning of is_prepared here does not test whether XASL exists on the server side. It means whether the internal structure including PT_NODE was created by parsing the query received from the SP. Therefore, in this function, "is prepared" is checked, not "was prepared".

beyondykk9: Ok, I understood.

beyondykk9 (approved):

hyunikn (approved):

PR #4008, opened 2022-12-20
[CBRD-24588] to print out only the comment string without single quote escape in session command

beyondykk9:
http://jira.cubrid.org/browse/CBRD-24588

CREATE TABLE tbl2 (id INT) COMMENT 'aa''bb';

As a result of this query, ;sc tbl2 has a problem with two single-quotes appearing in succession.

;sc tbl2

<Class Name>

dba.tbl2 COMMENT 'aa''bb'
...

However, the query "SELECT class_name, comment FROM db_class WHERE class_name = 'tbl2'" returns fine.

class_name comment
'tbl2' 'aa'bb'

We need to modify the routine prints out the original comment string in the session command as like ";sc table-name". However, in the "show create table" command, it needs to print out the escaped single quote.


src/object/object_printer.cpp
```diff
@@ -43,6 +43,30 @@
...
+      printer.describe_comment_value (&comment_value);
+    }
+  else
+    {
+      m_buf ("''");
+    }
+
+  pr_clear_value (&comment_value);
+}
+
```
hgryoo, line 69: Most of the lines are duplicated with describe_comment(). How about just adding `class_description::type prt_type` as an argument in describe_comment () and calling describe_comment_value() or describe_value() according to the `prt_type`?

hgryoo (approved):

shparkcubrid (approved):

youngjinj (approved):

ctshim (approved):

hyunikn (approved):

src/object/object_printer.cpp
beyondykk9: I keep this change for simplify.

PR #4011, opened 2022-12-22
[CBRD-24593] replace select () with poll () in the BROKER

kisoo-han:
http://jira.cubrid.org/browse/CBRD-24593

**Purpose**
* To handle up to 4096 connections simultaneously.
* replace select () system call to poll () because it cannot monitor more than 1024 fds naturally.

**Implementation**
N/A

**Remarks**
* This changes are critical to broker's behavior, **careful review is required**.
* By the way, current CCI able to support up to 2048 connections, we have to modify it also.
* An unit test was completed for simultaneous access of 4,000 clients.

### DIFF
--- a/src/broker/broker.c
+++ b/src/broker/broker.c
@@ -3342,12 +3342,12 @@ read_buffer_async (SOCKET sock_fd, char *buf, int size, int timeout)
     }
   while (ret < 0 && errno == EINTR);
 
-  if (ret < 1) /* ERROR OR TIMEOUT */
+  if (ret < 1)			/* ERROR OR TIMEOUT */
     {
       return -1;
     }
 
-  if (po[0].revents & POLLIN) /* RECEIVE NEW REQUEST */
+  if (po[0].revents & POLLOUT)	/* RECEIVE NEW REQUEST */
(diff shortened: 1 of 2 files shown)


src/broker/broker.c
```diff
@@ -3367,3 +3325,65 @@ get_as_slow_log_filename (char *log_filename, int len, char *broker_name, T_APPL
...
+}
+
+static int
+write_buffer_async (SOCKET sock_fd, char *buf, int size, int timeout)
+{
+  int write_len = -1;
+  struct pollfd po[1] = { {0, 0, 0} };
+  int ret;
+
+#ifdef ASYNC_MODE
```
mhoh3963: How about moving "#ifdef ASYNC_MODE" from this function to write_to_client_with_timeout() like  implementing "#ifdef ASYNC_MODE" at read_to_client_with_timeout() ?

kisoo-han: It looks much better.
Thank you for your comment!

mhoh3963 (approved):

airnet73 (approved):

PR #4012, opened 2022-12-23
[CBRD-24594] fix pr_midxkey_add_elements()

ctshim:
http://jira.cubrid.org/browse/CBRD-24594

* The parameter value was misrepresenting when calling pr_midxkey_init_boundbits() in the pr_midxkey_add_element() function.
  As a result, if there are more than nine columns, there is a problem that "boundbits" is initialized only for the preceding eight columns.

### DIFF
--- a/src/object/object_primitive.c
+++ b/src/object/object_primitive.c
@@ -9541,7 +9541,7 @@ pr_midxkey_add_elements (DB_VALUE * keyval, DB_VALUE * dbvals, int num_dbvals, s
   else
     {
       /* bound bits */
-      (void) pr_midxkey_init_boundbits (bound_bits, num_dbvals);
+      MIDXKEY_BOUNDBITS_INIT (bound_bits, bitmap_size);
       or_advance (&buf, bitmap_size);
     }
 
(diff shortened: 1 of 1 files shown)


src/object/object_primitive.c
```diff
@@ -9541,7 +9541,7 @@ pr_midxkey_add_elements (DB_VALUE * keyval, DB_VALUE * dbvals, int num_dbvals, s
   else
     {
       /* bound bits */
-      (void) pr_midxkey_init_boundbits (bound_bits, bitmap_size);
+      (void) pr_midxkey_init_boundbits (bound_bits, num_dbvals);
```
youngjinj: Since we have already calculated bitmap_size,
calculating it once more in pr_midxkey_init_boundbits is redundant.
```suggestion
      MIDXKEY_BOUNDBITS_INIT (bound_bits, bitmap_size);
```

shparkcubrid (approved):

src/object/object_primitive.c
ctshim: Good point!
I will fix it. thanks.

youngjinj (approved):

hgryoo (approved):

beyondykk9 (approved):

hyunikn (approved):

PR #4015, opened 2022-12-27
[CBRD-24598] Memory allocated on the global heap must be freed on the global heap.

youngjinj:
http://jira.cubrid.org/browse/CBRD-24598

### Use of the global heap
-  Generates xasl in the global heap when a clone for the corresponding xasl cannot be found in the xasl cache entry. Before calling stx_map_stream_to_xasl() in xcache_find_xasl_id_for_execute(), set heap_id to 0 to use the global heap.
- In the global heap, db_private_alloc() calls malloc() to allocate memory. In a private heap where heap_id is not 0, hl_lea_alloc() (mspace_malloc()) is called to allocate memory. In the private heap, a problem occurs when trying to find and free memory allocated from the global heap.
- A flag called REGU_VARIABLE_CLEAR_AT_CLONE_DECACHE is set in REGU_VARIABLE to distinguish whether memory is allocated from the global heap. Set when unpacking DB_VALUE in stx_unpack_regu_variable_value().

### Before calling pr_clear_value(), it is necessary to check whether the global heap is used.
- When eval_pred() calls eval_value_rel_cmp(), peek_val2 (dbval2, dest) is using the global heap. Before calling tp_value_coerce(), heap_id must be set to 0.
   ```
   /* src/query/query_evaluator.c */
   
   static DB_LOGICAL
   eval_value_rel_cmp (DB_VALUE * dbval1, DB_VALUE * dbval2, REL_OP rel_operator,
                       const COMP_EVAL_TERM * et_comp)
   {
     ...
     /* add */
     if (REGU_VARIABLE_IS_FLAGED (et_comp->rhs, REGU_VARIABLE_CLEAR_AT_CLONE_DECACHE))
       {
         save_heapid = db_change_private_heap (thread_p, 0);
       }
     
     (void) tp_value_coerce (dbval2, dbval2, dom);
     
     /* add */
     if (save_heapid != 0)
       {
         (void) db_change_private_heap (thread_p, save_heapid);
       }
   ...
   ```

### DIFF
--- a/src/query/query_evaluator.c
+++ b/src/query/query_evaluator.c
@@ -47,11 +47,11 @@
 
 static DB_LOGICAL eval_negative (DB_LOGICAL res);
 static DB_LOGICAL eval_logical_result (DB_LOGICAL res1, DB_LOGICAL res2);
-static DB_LOGICAL eval_value_rel_cmp (DB_VALUE * dbval1, DB_VALUE * dbval2, REL_OP rel_operator,
-				      const COMP_EVAL_TERM * et_comp);
-static DB_LOGICAL eval_some_eval (DB_VALUE * item, DB_SET * set, REL_OP rel_operator);
-static DB_LOGICAL eval_all_eval (DB_VALUE * item, DB_SET * set, REL_OP rel_operator);
-static int eval_item_card_set (DB_VALUE * item, DB_SET * set, REL_OP rel_operator);
+static DB_LOGICAL eval_value_rel_cmp (THREAD_ENTRY * thread_p, DB_VALUE * dbval1, DB_VALUE * dbval2,
+				      REL_OP rel_operator, const COMP_EVAL_TERM * et_comp);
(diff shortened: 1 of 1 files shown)


hyunikn (approved):

hgryoo:
I really love the detailed description of the JIRA issue. The patches proposed in this PR seem enough. :-)
Beyond what is resolving in this PR, The code pattern that changes to the global heap id and restores it also exists in the query executor (+ query aggregate). They don't look lovely.
The query executors (+ aggregate, evaluator) do not have to care how and where the memory was allocated. The same problem can occur anywhere in the XASL structure, where any DB_VALUE already allocated from the global heap id, is cleared for any reason.
I'd like to discuss the issue that I explained above further.

hgryoo (approved):

beyondykk9 (approved):

ctshim (approved):

shparkcubrid (approved):

PR #4020, opened 2022-12-28
[CBRD-24600] Problem with creating the same index with different names

ctshim:
http://jira.cubrid.org/browse/CBRD-24600

* fix classobj_check_index_compatibility() and classobj_find_constraint_by_attrs()

The classobj_find_constraint_by_attrs() function is the role of checking whether an index with the same configuration already exists.
The cause of the problem was that the function did not include function index information in the check target of the same configuration.
classobj_check_index_compatibility() was checking if the function index information was the same.

For example, all three indexes below use column val.
create index idx_a on tbl(val);
create index idx_b on tbl(abs(val));
create index idx_c on tbl(val);

case1) If you attempt to create idx_c after creating it in the order of idx_a and idx_b
The classobj_find_constrain_by_attrs() function returns idx_a.
In classobj_check_index_compatibility(), idx_c and idx_a are determined to have the same configuration and do not generate idx_c.

case2) If you attempt to create idx_c after creating it in the order of idx_b and idx_a
The classobj_find_constrain_by_attrs() function returns idx_b.
In classobj_check_index_compatibility(), idx_c and idx_b are determined to have different configurations and will generate idx_c.

So, here's the correction:

1) classobj_find_constrain_by_attrs() has been changed to find the same function index information as well as filter conditions.
2) classobj_check_index_compatibility() removes checks that are no longer needed.
3) Additionally, I changed the logic to a switch statement so that you can see it intuitively.

### DIFF
--- a/src/object/class_object.c
+++ b/src/object/class_object.c
@@ -7985,6 +7985,18 @@ classobj_check_index_compatibility (SM_CLASS_CONSTRAINT * constraints, const DB_
 {
   SM_CONSTRAINT_COMPATIBILITY ret;
 
+  /* only one primary key is allowed in a table. */
+  if (constraint_type == DB_CONSTRAINT_PRIMARY_KEY)
+    {
+      SM_CLASS_CONSTRAINT *prim_con;
+      prim_con = classobj_find_cons_primary_key (constraints);
+      if (prim_con != NULL)
+	{
+	  *primary_con = prim_con;
+	  return SM_NOT_SHARE_PRIMARY_KEY_AND_WARNING;
+	}
+    }
+
   if (existing_con == NULL)
     {
       return SM_CREATE_NEW_INDEX;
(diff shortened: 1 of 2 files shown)


src/object/class_object.c
```diff
@@ -4112,50 +4116,53 @@ classobj_find_constraint_by_attrs (SM_CLASS_CONSTRAINT * cons_list, DB_CONSTRAIN
...
+	    {
+	      continue;
+	    }
 
-		      if (!filter_predicate->pred_string || !cons->filter_predicate->pred_string)
-			{
-			  continue;
-			}
+	  if (filter_predicate)
+	    {
```
hgryoo, line 4141: Checking `cons->filter_predicate` seems to be removed.

ctshim: Due to the conditions added to the 4105 line,
If filter_predicate is not NULL, cons->filter_predicate is also determined to be NULL.

`(filter_predicate && !cons->filter_predicate) || (!filter_predicate && cons->filter_predicate)`

hgryoo (approved):

hyunikn (approved):

beyondykk9 (approved):

shparkcubrid (approved):

youngjinj (approved):

PR #4025, opened 2023-01-04
[CBRD-24606] refactoring pr_midxkey_compare()

ctshim:
http://jira.cubrid.org/browse/CBRD-24606

* Removes meaningless code lines.
* Try to perform fewer comparisons of if() statements.
* The code was simplified by reducing the components of the function.

### DIFF
--- a/src/object/object_primitive.c
+++ b/src/object/object_primitive.c
@@ -7761,8 +7761,7 @@ clean_up:
 
 DB_VALUE_COMPARE_RESULT
 pr_midxkey_compare (DB_MIDXKEY * mul1, DB_MIDXKEY * mul2, int do_coercion, int total_order, int num_index_term,
-		    int *start_colp, int *result_size1, int *result_size2, int *diff_column, bool * dom_is_desc,
-		    bool * next_dom_is_desc)
+		    int *start_colp, int *diff_column, bool * dom_is_desc, int *result_size)
 {
   DB_VALUE_COMPARE_RESULT c = DB_UNK;
   int i;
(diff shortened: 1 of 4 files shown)


src/object/object_primitive.c
```diff
@@ -7830,50 +7829,41 @@ pr_midxkey_compare (DB_MIDXKEY * mul1, DB_MIDXKEY * mul2, int do_coercion, int t
...
-      if (OR_MULTI_ATT_IS_BOUND (bitptr1, i))
-	{
-	  adv_size1 = pr_midxkey_element_disk_size (mem1, dom1);
-	  mem1 += adv_size1;
-	  size1 += adv_size1;
-	}
+	  if (OR_MULTI_ATT_IS_BOUND (bitptr1, i))
+	    {
+	      adv_size = pr_midxkey_element_disk_size (mem1, dom1);
+	      mem1 += adv_size;
```
youngjinj: How about below?
```suggestion
	      mem1 += pr_midxkey_element_disk_size (mem1, dom1);
```

ctshim: I thought Optimizer would take care of that.
I thought it would be nice to have a variable to put the value in for debugging in case.

youngjinj (approved):

src/object/object_primitive.c
youngjinj: I agree. Thank you.

shparkcubrid (approved):

beyondykk9 (approved):

hyunikn (approved):

src/storage/btree.c
```diff
@@ -18847,11 +18845,14 @@ btree_compare_key (DB_VALUE * key1, DB_VALUE * key2, TP_DOMAIN * key_domain, int
...
+      c =
+	pr_midxkey_compare (db_get_midxkey (key1), db_get_midxkey (key2), do_coercion, total_order, -1, start_colp,
+			    &dummy_diff_column, dom_is_desc, NULL);
       assert_release (c == DB_UNK || (DB_LT <= c && c <= DB_GT));
 
-      if (dom_is_desc)
+      if (dom_is_desc[0])
 	{
 	  c = ((c == DB_GT) ? DB_LT : (c == DB_LT) ? DB_GT : c);
 	}
```
youngjinj, line 18858: Shouldn't it be handled inside pr_midxkey_compare?

ctshim: It's possible, but it's not appropriate because you need to add an agreement.

youngjinj: I found out that this part should not be included because pr_midxkey_compare() is not used only for index sorting.

hgryoo (approved):

PR #4026, opened 2023-01-06
[CBRD-24609] When loading objects files online, if the schema name is not specified, the schema name of the current session must be used.

youngjinj:
http://jira.cubrid.org/browse/CBRD-24609

When loading objects files online, if the schema name is not specified, the schema name of the current session must be used.


src/loaddb/load_server_loader.cpp
```diff
@@ -117,8 +117,28 @@ namespace cubload
     cubthread::entry &thread_ref = cubthread::get_entry ();
     LC_FIND_CLASSNAME found = LC_CLASSNAME_EXIST;
     LC_FIND_CLASSNAME found_again = LC_CLASSNAME_EXIST;
+    const char *dot = strchr (class_name, '.');
```
shparkcubrid: Is there any other way to check? For example checking the flag

hgryoo (approved):

beyondykk9 (approved):

hyunikn (approved):

src/loaddb/load_server_loader.cpp
```diff
@@ -117,8 +117,28 @@ namespace cubload
...
+
+    if (dot != NULL)
```
ctshim:
```suggestion
    if (strchr (class_name, '.'))
```

ctshim (approved):

shparkcubrid (approved):

src/loaddb/load_server_loader.cpp
youngjinj: It doesn't seem to be a problem with loaddb only.
I will think more through other issues.

src/loaddb/load_server_loader.cpp
youngjinj: Thank you. I will change it.

PR #4028, opened 2023-01-09
[CBRD-24615] Result value error when using query cache in partition table

beyondykk9:
http://jira.cubrid.org/browse/CBRD-24615

When using a partition table, the existing query cache does not disappear even if data is inserted or modified.

Even if it looks at the actual ";info qcache", the existing cache should disappear when data is inserted, but it remains and the previous value is retrieved.

The operation without partitioning is good.

### DIFF
--- a/src/transaction/locator_sr.c
+++ b/src/transaction/locator_sr.c
@@ -5159,6 +5159,10 @@ locator_insert_force (THREAD_ENTRY * thread_p, HFID * hfid, OID * class_oid, OID
 			    "locator_insert_force: qexec_clear_list_cache_by_class failed for class { %d %d %d }\n",
 			    real_class_oid.pageid, real_class_oid.slotid, real_class_oid.volid);
 	    }
+	  if (!OID_EQ (&real_class_oid, class_oid))
+	    {
+	      qmgr_add_modified_class (thread_p, class_oid);
+	    }
 	  qmgr_add_modified_class (thread_p, &real_class_oid);
 	}
 #if 0				/* TODO - dead code; do not delete me */
@@ -5329,7 +5333,11 @@ locator_update_force (THREAD_ENTRY * thread_p, HFID * hfid, OID * class_oid, OID
   REPL_INFO repl_info;
   TDE_ALGORITHM tde_algo = TDE_ALGORITHM_NONE;
 
+  OID superclass_oid;
+
   assert (class_oid != NULL && !OID_ISNULL (class_oid));
+
+  OID_SET_NULL (&superclass_oid);
 
   /*
    * While scanning objects, the given scancache does not fix the last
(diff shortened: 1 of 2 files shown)


src/transaction/locator_sr.c
```diff
@@ -5329,8 +5333,12 @@ locator_update_force (THREAD_ENTRY * thread_p, HFID * hfid, OID * class_oid, OID
   REPL_INFO repl_info;
   TDE_ALGORITHM tde_algo = TDE_ALGORITHM_NONE;
 
+  OID superclass_oid;
+
   assert (class_oid != NULL && !OID_ISNULL (class_oid));
 
+  OID_SET_NULL (&superclass_oid);
```
hgryoo, line 5340: For information only
the following is the same.
```
OID superclass_oid = OID_INITIALIZER;
```

hgryoo (approved):

hgryoo (approved):

src/transaction/locator_sr.c
```diff
@@ -6274,12 +6285,35 @@ locator_delete_force_internal (THREAD_ENTRY * thread_p, HFID * hfid, OID * oid,
...
 			    class_oid.pageid, class_oid.slotid, class_oid.volid);
 	    }
+
+	  OID_SET_NULL (&superclass_oid);
+
+	  HFID_COPY (&real_hfid, hfid);
+	  COPY_OID (&real_class_oid, &class_oid);
+
+	  error_code =
+	    partition_prune_update (thread_p, &class_oid, &copy_recdes, NULL, DB_PARTITIONED_CLASS, &real_class_oid,
```
shparkcubrid, line 6305: Can't the routine be executed only in the case of a partition table? like below.
```
locator_update_force()
if (pruning_type != DB_NOT_PARTITIONED_CLASS)
```

beyondykk9: the pruning_type is necessary but the caller doesn't transfer the value. the partition_load_pruning_context called by partition_prun_update  returns null as superclass_oid when the pruning_type is not DB_PARTITION_CLASS. To modify the function call locator_delete_force with pruning_type as an argument needs to modify several routines.
In addition, partition_prune_update is only called at QUERY_CACHE is enabled. Therefore the partition_prune_update 
 calling cost is limited.

src/transaction/locator_sr.c
beyondykk9: Ok, thanks.

hyunikn (approved):

shparkcubrid (approved):

src/transaction/locator_sr.c
```diff
@@ -6274,12 +6285,35 @@ locator_delete_force_internal (THREAD_ENTRY * thread_p, HFID * hfid, OID * oid,
       /* remove query result cache entries which are relevant with this class */
       if (!QFILE_IS_LIST_CACHE_DISABLED)
 	{
+	  OID superclass_oid;
```
ctshim, line 6288: Please apply hgryoo's comments here as well

OID superclass_oid = OID_INITIALIZER;

ctshim (approved):

src/transaction/locator_sr.c
beyondykk9: OID_SET_NULL is better readability?

youngjinj (approved):

PR #4029, opened 2023-01-10
[CBRD-24618] Remove unnecessary hton, ntoh related routines

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-24618

Remove unnecessary hton, ntoh related routines


shparkcubrid:
hopefully, it is fixed later.

PR #4032, opened 2023-01-11
[CBRD-24605] Java package name option for loadjava utility

hgryoo:
http://jira.cubrid.org/browse/CBRD-24605

**Purpose**
In loadjava utility, the package name was not considered. 

**Implementation**
- Add the package name argument in loadjava
- Modernize the utility in C++


shparkcubrid (approved):

ctshim (approved):

src/executables/loadjava.cpp
```diff
@@ -0,0 +1,239 @@
...
+	default:
+	  usage ();
+	  return ER_FAILED;
+	}
+    }
+
+  if (optind + 1 < argc)
+    {
+      Dbname = argv[optind];
+      Src_class = argv[optind + 1];
```
hyunikn, line 121: It will be helpful if we can handle many src classes at once, because compiling a java file of a PL/CSQL program usually generates many class files because of its local classes.

hyunikn (approved):

src/executables/loadjava.cpp
```diff
@@ -0,0 +1,239 @@
...
+static std::string package_path;
+
+static void
+usage (void)
+{
+  fprintf (stderr, "Usage: loadjava [OPTION] database-name java-class-file\n");
+  fprintf (stderr, "Options:\n-y\t%s\n-p\t%s\n",
+	   msgcat_message (MSGCAT_CATALOG_UTILS, MSGCAT_UTIL_SET_LOADJAVA, LOADJAVA_ARG_FORCE_OVERWRITE_HELP),
+	   msgcat_message (MSGCAT_CATALOG_UTILS, MSGCAT_UTIL_SET_LOADJAVA, LOADJAVA_ARG_FORCE_PACKAGE_HELP));
+}
```
youngjinj: How about making a message like below?
```
$set 1 MSGCAT_UTIL_SET_GENERIC
...
44 cubrid utility, version %1$s\n\
usage: %2$s javasp <command> [args]\n\
\n\
Available command:\n\
    start       [database-name]\n\
    stop        [database-name]\n\
    restart     [database-name]\n\
    status      [database-name]\n
```

src/executables/loadjava.cpp
hgryoo: I'd like to add the functionality in the next issue. Thank you for the suggestion.

src/executables/loadjava.cpp
hgryoo: Good point! Let me modify the messages.

src/executables/loadjava.cpp
```diff
@@ -0,0 +1,239 @@
...
+using namespace std::regex_constants;
+
+#define JAVA_DIR                "java"
+#if defined(WINDOWS)
+#define SEPERATOR               "\\"
+#else /* ! WINDOWS */
+#define SEPERATOR               "/"
+#endif /* !WINDOWS */
+
+static const std::string JAVA_PACKAGE_PATTERN = "^([a-z_]{1}[a-z0-9_]*(\\.[a-z_]{1}[a-z0-9_]*)*)$";
```
youngjinj, line 51: Can the package name start with an underscore ( _ ) ?

src/executables/loadjava.cpp
```diff
@@ -0,0 +1,239 @@
...
+create_package_directories (const std::string &java_dir_path)
+{
+  try
+    {
+      if (fs::exists (java_dir_path) == false)
+	{
+	  fs::create_directories (java_dir_path);
+	  fs::permissions (java_dir_path,
+			   fs::perms::owner_all | fs::perms::group_read | fs::perms::others_read,
+			   fs::perm_options::add);	// mkdir (java_dir_path, 0744)
```
youngjinj, line 144: Is there a reason for 744 instead of giving 755?
It seems to be the same as 700 if the execute permission is missing from the directory.

src/executables/loadjava.cpp
youngjinj: I found that it is possible.

youngjinj (approved):

src/executables/loadjava.cpp
hgryoo: I just keep it as it is. the classes in the path are not executable, IMO the permission seems not required.

beyondykk9 (approved):

src/executables/loadjava.cpp
youngjinj: It is trivial.

Usually, it is natural to check the umask and grant permission.

The owner has all the rights, so no problem.
I don't know if the group user or other user should have the permission.

Neither group users nor other users can enter the directory because they do not have execute permission on the directory.
So, I know that directories are given read permission and execute permission together or not together.

hgryoo: I will revisit the loadjava utility while working on PL/CSQL project. I'm going to consider the permission change at then.

PR #4036, opened 2023-01-12
[CBRD-24501] extending dblink dml

beyondykk9:

Test Cases

– local server

create table tbl (a int, b varchar(10));
insert into tbl values (1, 'abc'), (2, 'abcd'), (3, 'abcde'), (4, 'abcdef'), (5, 'abcdefg');

– remote 192.168.2.193 server (srv1)

create table t1 (a int, b varchar(10));
insert into t1 values (1, 'abc'), (2, 'abcd'), (3, 'abcde'), (4, 'abcdef'), (5, 'abcdefg');
create table t2 (a int, b varchar(10));
insert into t2 values (1, 'abc'), (2, 'abcd'), (3, 'abcde'), (4, 'abcdef'), (5, 'abcdefg');

– remote 192.168.2.194 server(srv2)

create table t1 (a int, b varchar(10));
insert into t1 values (1, 'abc'), (2, 'abcd'), (3, 'abcde'), (4, 'abcdef'), (5, 'abcdefg');
create table t2 (a int, b varchar(10));
insert into t2 values (1, 'abc'), (2, 'abcd'), (3, 'abcde'), (4, 'abcdef'), (5, 'abcdefg');

– insert query

insert into tbl select * from t1@srv1 t1 where t1.a > 1; (OK)
insert into tbl select t1.a, t2.b from t1@srv1 t1, t2@srv2 t2 where t1.a = t2.a and t2.a > 1; (OK)
insert into t1@srv1 select * from t1@srv1 t1 where t1.a > 1; (OK)
insert into t1@srv1 select * from t2@srv2 t2 where t2.a > 1; (ERR multi-remote)
insert into t1@srv1 select * from t1@srv1 t1, t2@srv2 t2 where t1.a = t2.a and t2.a > 1; (ERR: multi-remote)
insert into tbl select * from dblink(srv1, 'select a, b from t1') t1(a int, b varchar) (OK)
insert into tbl select * from dblink('192.168.2.194:33000:demodb:dba::', 'select a, b from t1') t1(a int, b varchar) (OK)
insert into t1@srv1 select * from dblink(srv1, 'select a, b from t2') t(a int, b varchar); (ERR: dblink query)

– update query

update tbl join t1@srv1 t1 on tbl.a = t1.a set tbl.b = 'upd' where tbl.a > 1 and t1.b = 'abc'; (OK)
update tbl join (select * from t1@srv1) t1 on tbl.a = t1.a set tbl.b = 'upd' where tbl.a > 1 and t1.b = 'abc'; (OK)
update t1@srv1 t1 set t1.b = 'upd' where t1.a > 1; (OK)
update t1@srv1 t1 join t1@srv2 t2 on t1.a = t2.a set t1.b = 'upd' where t2.a > 1; (ERR: multi-remote)
update tbl join t1@srv1 t1 on tbl.a = t1.a set t1.b = 'upd' where tbl.a > 1; (ERR: mixed update)
update tbl join (select * from dblink(srv1, 'select a, b from t1') t(a int, b varchar)) t1 on tbl.a = t1.a set tbl.b = 'xxx'; (OK)
update tbl join (select * from dblink('192.168.2.194:33000:demodb:dba::', 'select a, b from t1') t(a int, b varchar)) t1 on tbl.a = t1.a set tbl.b = 'xxx'; (OK)
update t1@srv1 t1 join (select * from dblink(srv1, 'select a, b from t2') t(a int, b varchar)) t2 on t1.a = t2.a set t1.b = 'xxx'; (ERR: dblink query)

– delete query

delete tbl from tbl join t1@srv1 t1 on tbl.a = t1.a where tbl.a > 1; (OK)
delete tbl from tbl join (select * from t1@srv1) t1 on tbl.a = t1.a where tbl.a > 1; (OK)
delete from t1@srv1 t1 where t1.a > 1; (OK)
delete t1 from tbl join t1@srv1 t1 on tbl.a = t1.a where t1.a > 1; (ERR: mixed delete)
delete t1 from t1@srv1 t1 join t2@srv2 t2 on t1.a = t2.a where t1.a > 1 and t2.b = 'abc'; (ERR: multi-remote)
delete tbl from tbl join (select * from dblink(srv1, 'select a, b from t1') t(a int, b varchar)) t1 on tbl.a = t1.a; (OK)
delete tbl from tbl join (select * from dblink('192.168.2.194:33000:demodb:dba::', 'select a, b from t1') t(a int, b varchar)) t1 on tbl.a = t1.a; (OK)
delete t2 from t2@srv1 t2 join (select * from dblink(srv1, 'select a, b from t1') t(a int, b varchar)) t1 on t2.a = t1.a where t1.a > 1; (ERR: dblink query)

delete from t1 using t2@srv1 t2, t1@srv1 t1 where t2.a = t1.a; (OK)
delete from t1 using t2@srv1, t1@srv1 where t2.a = t1.a; (ERR: multi-remote)
delete from t1 using t2@srv1, t1@srv1 t1 where t2.a = t1.a; (OK)

– merge query

merge into tbl using t1@srv1 t1 on tbl.a = t1.a
when matched then update set b = 'upd'
when not matched then insert values (999, 'new'); (OK)
merge into tbl using (select * from t1@srv1) t1 on tbl.a = t1.a
when matched then update set b = 'upd'
when not matched then insert values (999, 'new'); (OK)
merge into tbl using (select * from dblink(srv1, 'select a, b from t1') t(a int, b varchar)) t1 on tbl.a = t1.a
when matched then update set b = 'upd'
when not matched then insert values (999, 'new'); (OK)
merge into t1@srv1 t1 using t2@srv1 t2 on t2.a = t1.a
when matched then update set b = 'upd'
when not matched then insert values (999, 'new'); (OK)
merge into t1@srv1 using (select * from t2@srv1) t2 on t2.a = t1.a
when matched then update set b = 'upd'
when not matched then insert values (999, 'new'); (OK)
merge into t1@srv1 using (select * from dblink(srv1, 'select a, b from t2') t(a int, b varchar)) t2 on tbl.a = t1.a
when matched then update set b = 'upd'
when not matched then insert values (999, 'new'); (ERR: dblink query)
merge into t1@srv1 t1 using t2@srv2 t2 on t2.a = t1.a
when matched then update set b = 'upd'
when not matched then insert values (999, 'new'); (ERR: multi-remote)


youngjinj:
In the case of a PR review related to dblink, I think that the target test server information is required.
Is it possible?

beyondykk9:
> In the case of a PR review related to dblink, I think that the target test server information is required. Is it possible?

The broker configuraion is as follows:
* 192.168.2.193 - port 33000, demodb, dba
* 192.168.2.194 - port 33000, demodb, dba

src/compat/db_vdb.c
```diff
@@ -581,6 +581,14 @@ db_compile_statement_local (DB_SESSION * session)
   /* forget about any previous parsing errors, if any */
   pt_reset_error (parser);
 
+  // @server
+  pt_check_server_extension (parser, statement);
```
youngjinj: I think "pt_check_" only checks if it is a server_extension.
However, pt_check_server_extension() calls pt_convert_server() and changes the statement to server_extension.

So I want it to be a clear name like "pt_convert_" rather than "pt_check_".

src/parser/csql_grammar.y
```diff
@@ -5405,9 +5448,21 @@ only_all_class_spec
 		DBG_PRINT}}
 	;
 
+class_name_with_server_name
+        : class_name '@' dblink_server_name
+                {{DBG_TRACE_GRAMMAR(class_name_with_server_name, : class_name '@' dblink_server_name);
+                assert($1 != NULL && $3 != NULL);
+                container_2 ctn;
+                // Do NOT automatically assign a user name. 
+                PT_NAME_INFO_CLEAR_FLAG($1, PT_NAME_INFO_USER_SPECIFIED);
```
youngjinj, line 5505: Since the target_owner of _db_synonym is NOT NULL, the schema name of the current session must be attached.
So the PT_NAME_INFO_USER_SPECIFIED flag must not be removed.

How about attaching the schema name of the current session, and removing it when using it?

src/parser/compile.c
```diff
@@ -431,7 +432,7 @@ PT_NODE *
 pt_class_pre_fetch (PARSER_CONTEXT * parser, PT_NODE * statement)
 {
   PT_CLASS_LOCKS lcks;
-  PT_NODE *node = NULL;
+  PT_NODE *node = NULL, *spec;
```
youngjinj: Not used.

src/parser/parse_tree_cl.c
```diff
@@ -9182,6 +9201,19 @@ pt_print_spec (PARSER_CONTEXT * parser, PT_NODE * p)
 	}
       r1 = pt_print_bytes (parser, p->info.spec.entity_name);
       q = pt_append_varchar (parser, q, r1);
+
+      if (p->info.spec.remote_server_name)
```
youngjinj, line 9205: It comes here only if it is PT_META_CLASS.
Since it is about the Object concept, this code should be removed.

```
create table t1 class attribute (a1 int) (c1 int) dont_reuse_oid;

set optimization level 513;
select /*+ recompile */ * from class t1;
```
```
create table t2 (c1 int) dont_reuse_oid;
create table t3 as subclass of t2 (c1 int) dont_reuse_oid;

set optimization level 513;
select /*+ recompile */ * from all t3;
```

src/parser/parse_tree_cl.c
```diff
@@ -9164,6 +9171,18 @@ pt_print_spec (PARSER_CONTEXT * parser, PT_NODE * p)
       q = pt_append_nulstring (parser, q, "(");
       r1 = pt_print_bytes_l (parser, p->info.spec.entity_name);
       q = pt_append_varchar (parser, q, r1);
+
+      if (p->info.spec.remote_server_name)
```
youngjinj, line 9175: This code only runs in the following cases.
But the result is different from what I expected.

If such syntax should be allowed, it is considered a bug.

This code should be removed if such syntax is not allowed.
And exception handling is required.

**On the remote server**
```
csql> select * from (t1, t2);

           c1
=============
          200
          100
```

**Use DBLink**
```
csql> select * from t1@cubrid_dblink;

           c1
=============
          100

csql> select * from t2@cubrid_dblink;

           c1
=============
          200

csql> select * from (t1@cubrid_dblink, t2@cubrid_dblink);

There are no results.
0 row selected. 

csql> select * from dblink (cubrid_dblink, 'select c1 from (t1, t2)') as t1 (c1 int);

           c1
=============
          200
          100
```

src/parser/csql_grammar.y
```diff
@@ -13963,10 +14036,11 @@ opt_from_clause
 
 			    node->info.query.q.select.with_increment = $10;
 			    node->info.query.id = (UINTPTR) node;
-			  }
 
 			if (node->info.query.all_distinct != PT_ALL)
 			  is_dummy_select = false;	/* not dummy */
+			  }
```
youngjinj, line 14090: Need to indent the if statement above.

src/parser/csql_grammar.y
```diff
@@ -4053,6 +4070,24 @@ only_class_name
 	| user_specified_name
 		{ DBG_TRACE_GRAMMAR(only_class_name, | user_specified_name);
                   $$ = $1; }
+        | class_name_with_server_name
+                {{ DBG_TRACE_GRAMMAR(only_class_name, | class_name_with_server_name);
+
+                        PT_NODE *scs = parser_new_node (this_parser, PT_SPEC);
```
youngjinj, line 4090: "only_class_name" should return PT_NAME.
Otherwise PT_SPEC will come where PT_NAME was expected to come.

Exception handling is required.

```
csql> alter table t1@cubrid_dblink comment 'The use of dblink is not allowed in the ALTER statement.';

ERROR: before '  comment 'The use of dblink is not allowed in the ALTER state...'
Class _ does not exist.
```

src/parser/parse_tree_cl.c
beyondykk9: The blow case also returns error:

select * from (t1, t2);

In line 1, column 22,

ERROR: before ' ); '
There are no attributes in ([dba.t1], [dba.t2]).

youngjinj: image

It runs just fine...

src/parser/compile.c
```diff
@@ -746,12 +756,37 @@ pt_add_lock_class (PARSER_CONTEXT * parser, PT_CLASS_LOCKS * lcks, PT_NODE * spe
...
+
+	  /* for alias */
+	  synonym_name = (char *) spec->info.spec.entity_name->info.name.original;
+	  if ((s = (char *) strchr (synonym_name, '.')) != NULL)
+	    {
+	      synonym_name = s + 1;
+	    }
+
+	  *r = 0;
+	  spec->info.spec.entity_name->info.name.original = pt_append_string (parser, NULL, class_name);
```
youngjinj: If target_owner_name is specified in target_unique_name of synonym, owner_name must be skipped in class_name.
```
char *dot = NULL;
if ((dot = (char *) strchr (class_name, '.')) != NULL)
  {
    class_name = dot + 1;
  }
```

src/parser/csql_grammar.y
youngjinj: I think class_name_with_server_name should not be added to only_class_name, and the parent rule of only_class_name should be added.

e.g.
```
only_class_name_extend
        : only_class_name
        | class_name_with_server_name

only_class_name
        : ONLY user_specified_name
        | user_specified_name
```

src/compat/db_vdb.c
```diff
@@ -658,37 +677,47 @@ db_compile_statement_local (DB_SESSION * session)
 	}
     }
 
-  /* translate views or virtual classes into base classes */
-  statement_result = mq_translate (parser, statement);
-  if (!statement_result || pt_has_error (parser))
+  if ((statement->node_type == PT_INSERT && statement->info.insert.spec->info.spec.remote_server_name)
```
youngjinj: How about checking "statement->flag.has_remote_server_name" instead of "statement->info.insert.spec->info.spec.remote_server_name"?

src/parser/semantic_check.c
```diff
@@ -11429,6 +11430,17 @@ pt_check_with_info (PARSER_CONTEXT * parser, PT_NODE * node, SEMANTIC_CHK_INFO *
       sc_info_ptr->system_class = false;
       node = pt_resolve_names (parser, node, sc_info_ptr);
 
+      if (!pt_has_error (parser))
+	{
+	  if ((node->node_type == PT_INSERT && node->info.insert.spec->info.spec.remote_server_name)
```
youngjinj, line 11435: How about checking "node->flag.has_remote_server_name" instead of "node->info.insert.spec->info.spec.remote_server_name"?

youngjinj, line 11432: How about adding PT_INSERT_DBLINK, PT_DELETE_DBLINK, PT_UPDATE_DBLINK, PT_MERGE_DBLINK?

I think adding a new node_type will make the code more concise.
We can distinguish in the switch statement.

src/parser/compile.c
```diff
@@ -746,6 +756,8 @@ pt_add_lock_class (PARSER_CONTEXT * parser, PT_CLASS_LOCKS * lcks, PT_NODE * spe
   synonym_mop = db_find_synonym (class_name);
   if (synonym_mop != NULL)
     {
+      char *r;
+
```
hgryoo: unused

src/query/dblink_scan.c
```diff
@@ -540,6 +540,64 @@ dblink_bind_param (DBLINK_SCAN_INFO * scan_info, VAL_DESCR * vd, DBLINK_HOST_VAR
...
+	    }
+	}
+
+      ret = cci_execute (stmt_handle, 0, 0, &err_buf);
+      if (ret < 0)
+	{
+	  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_DBLINK, 1, err_buf.err_msg);
+	  goto error_exit;
+	}
+    }
```
hgryoo: After the end of this function, the conn_handle value can not be referenced. Is it ok not to disconnect it?

src/parser/xasl_generation.c
```diff
@@ -18460,6 +18460,49 @@ pt_to_constraint_pred (PARSER_CONTEXT * parser, XASL_NODE * xasl, PT_NODE * spec
...
+  XASL_NODE *xasl = regu_xasl_node_alloc (INSERT_PROC);
+  if (xasl == NULL)
+    {
+      return NULL;
+    }
+
+  PT_NODE *server_spec = spec->info.spec.remote_server_name;
+  PT_DBLINK_INFO *pdblink = &(server_spec->info.dblink_table);
+
+  assert (server_spec->node_type == PT_DBLINK_TABLE_DML);
```
youngjinj, line 18477: The query below throws an assert because PT_DBLINK_TABLE_DML is not.
```
create synonym s1 for t1@cubrid_dblink;
insert into s1 values (1);

csql: src/parser/xasl_generation.c:18477: XASL_NODE* pt_to_xasl_for_dblink(PARSER_CONTEXT*, PT_NODE*): Assertion `server_spec->node_type == PT_DBLINK_TABLE_DML' failed.
Aborted (core dumped)
```
pt_convert_dblink_dml_query() is not called when pt_convert_dblink_insert_query() is called.

hgryoo (changes requested):
I found that this PR is referencing the POC JIRA issue. It should be merged into the feature branch as a POC project because the design and test spec don't clearly define it. Or please update the JIRA content.

src/parser/xasl_generation.c
beyondykk9: error is from create synonym s1 for t1@cubrid_dblink;
and insert fail will be patched.

youngjinj: Same below.
```
insert into t1@cubrid_dblink values (1);
```
I made a synonym like below for testing.
```
create synonym s1 for dba.t1@cubrid_dblink;
```

src/parser/compile.c
beyondykk9: Ok, good point.

src/compat/db_vdb.c
beyondykk9: The added flag of statement, has_remote_server_name was deleted in next commit.

src/parser/compile.c
beyondykk9: This logic was reverted by the next commit.

src/compat/db_vdb.c
beyondykk9: The function logic is convert, so I keep this function name.

src/parser/compile.c
beyondykk9: Good point.

src/parser/xasl_generation.c
beyondykk9: This is resolved by next commit.

src/compat/db_vdb.c
```diff
@@ -658,37 +666,47 @@ db_compile_statement_local (DB_SESSION * session)
...
+      || (statement->node_type == PT_DELETE && statement->info.delete_.spec->info.spec.remote_server_name)
+      || (statement->node_type == PT_UPDATE && statement->info.update.spec->info.spec.remote_server_name)
+      || (statement->node_type == PT_MERGE && statement->info.merge.into->info.spec.remote_server_name))
```
shparkcubrid: I disagree with skipping mq_translate(). Various processing such as VIEW processing will not proceed.

src/parser/csql_grammar.y
beyondykk9: This is removed from next commit.

src/compat/db_vdb.c
beyondykk9: This is only case of remote DML. In general DML case, the mq_translate() is called.

src/query/dblink_scan.c
beyondykk9: Good point. I will fix it.

src/parser/csql_grammar.y
beyondykk9: fixed.

src/parser/csql_grammar.y
beyondykk9: fixed.

src/parser/parse_tree_cl.c
beyondykk9: The query from (table, table) is not support yet as remote query.

src/parser/parse_tree_cl.c
beyondykk9: fixed.

src/compat/db_vdb.c
shparkcubrid: I disagree with skipping mq_translate() even for remote DML.

shparkcubrid:
> I found that this PR is referencing the POC JIRA issue. It should be merged into the feature branch as a POC project because the design and test spec don't clearly define it. Or please update the JIRA content.

There is no Jira content except for the tests. It should contain the contents of the specification and implementation.

src/compat/db_vdb.c
beyondykk9: mq_translate() is executed at remote server side. therefore the mq_tranlate at local side is not necessary.

beyondykk9:
> > I found that this PR is referencing the POC JIRA issue. It should be merged into the feature branch as a POC project because the design and test spec don't clearly define it. Or please update the JIRA content.
> 
> There is no Jira content except for the tests. It should contain the contents of the specification and implementation.

Yes, I modified the JIRA issue CBRD-24501

beyondykk9:
> I found that this PR is referencing the POC JIRA issue. It should be merged into the feature branch as a POC project because the design and test spec don't clearly define it. Or please update the JIRA content.

Yes, I modified the JIRA issue CBRD-24501

src/parser/semantic_check.c
beyondykk9: Good point, but I will keep the logic to prevent from accidental side effect.

src/compat/db_vdb.c
shparkcubrid: There may be cases like the following.
```
update remote@s1 a, local_view b 
     set ....
where a.col1 = b.col1
```

beyondykk9: In this version, this case results error because the remote DML query is not allow the join with a local view yet. The error is handled at pt_convert_dblink_dml_query.

src/parser/csql_grammar.y
```diff
@@ -5341,31 +5395,67 @@ meta_class_spec
...
 
-only_all_class_spec
-	: only_class_name opt_partition_spec
-		{{ DBG_TRACE_GRAMMAR(only_all_class_spec, : only_class_name opt_partition_spec );
+                        $$ = ocs;
+                DBG_PRINT}}
+	;
 
-			PT_NODE *ocs = parser_new_node (this_parser, PT_SPEC);
+only_all_class_spec_with_server
```
youngjinj, line 5426: I think there is no syntactical consistency in meta_class_spec, only_all_class_spec_with_server, only_all_class_spec.

CLASS syntax is allowed in meta_class_spec.
However, only_all_class_spec_with_server does not allow ALL or EXCEPT syntax.

It seems that CLASS syntax is allowed in case of connecting from CUBRID to CUBRID.
However, because it is inconsistent, I think that CLASS syntax should not be supported either, or all grammars should be supported consistently.

In CUBRID, syntax related to object orientation is rarely used.
Even if the user wants to use it, the table must be created with DONT_REUSE_OID.
Considering this situation, it would be better not to support object-oriented syntax in DBLINK.

src/parser/csql_grammar.y
```diff
@@ -5341,31 +5395,67 @@ meta_class_spec
...
+			        ocs->info.spec.partition = $2;
 			      }
+			   }
+
+			$$ = ocs;
+			PARSER_SAVE_ERR_CONTEXT ($$, @$.buffer_pos)
+		DBG_PRINT}}
+	;
+
+only_all_class_spec
```
youngjinj, line 5444: Can't we just add the following rule to only_all_class_spec instead of creating a new only_all_class_spec_with_server rule?
```
only_all_class_spec
	: only_class_name opt_partition_spec
        | only_class_name '@' dblink_server_name opt_partition_spec /* add */
        ...
```
This avoids using container_2 and simplifies the rules.

src/compat/db_vdb.c
shparkcubrid: It would be better that the routine where the query is rewritten is placed in mq_translate() for consistency. And better to make the name clear like rewrite_for_dblink.

src/parser/csql_grammar.y
beyondykk9: It's fixed.

src/compat/db_vdb.c
beyondykk9: Ok, I fixed as rename to pt_rewrite_for_dblink.

hgryoo:
In test_sql CI, some answers' error codes are changed from PT_SEMANTIC to PT_SYNTAX. Could you explain why it is changed?

beyondykk9:
> In test_sql CI, some answers' error codes are changed from PT_SEMANTIC to PT_SYNTAX. Could you explain why it is changed?

In pt_rewrite_for_dblink, the table spec is checked before pre_fetch, so the error code is returned as PT_SYNTAX type.

hgryoo:
> > In test_sql CI, some answers' error codes are changed from PT_SEMANTIC to PT_SYNTAX. Could you explain why it is changed?
> 
> In pt_rewrite_for_dblink, the table spec is checked before pre_fetch, so the error code is returned as PT_SYNTAX type.

pt_convert_dml.. functions seem to perform name binding from remote server information (checking synonyms, remote server names..). It is a part of the semantic check. So IMHO, The error type should be PT_SEMANTIC.

```
static void
pt_convert_dblink_insert_query (PARSER_CONTEXT * parser, PT_NODE * node, char *sql_user_text, SERVER_NAME_LIST * snl)
{
  PT_NODE *insert, *spec;
  int remote_ins = 0;
  bool has_synonym = false;

  parser_walk_tree (parser, node, pt_convert_dblink_synonym, &has_synonym, NULL, NULL);
  if (pt_has_error (parser))
    {
      return;
    }

 ....
```

Or if either PT_SYNTAX or PT_SEMANTIC can occur in the process of executing pt_rewrite_for_dblink, you can define the parser error code in parser_message.h and set PT_SYNTAX or PT_SEMANTIC error according to the parser error code.

e.g.)
```
PT_ERRORm (... MSGCAT_SET_PARSER_SYNTAX or MSGCAT_SET_PARSER_SEMANTIC)
```

src/compat/db_vdb.c
```diff
@@ -581,6 +581,14 @@ db_compile_statement_local (DB_SESSION * session)
   /* forget about any previous parsing errors, if any */
   pt_reset_error (parser);
 
+  // @server
```
shparkcubrid: please comment more

src/compat/db_vdb.c
shparkcubrid: How about 'PT_IS_DML_DBLINK_QUERY' macro function?

src/parser/name_resolution.c
```diff
@@ -2633,8 +2869,10 @@ pt_bind_names (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int *continue
       pt_bind_scope (parser, bind_arg);
 
       (void) pt_resolve_hint (parser, node);
-
-      parser_walk_leaves (parser, node, pt_bind_names, bind_arg, pt_bind_names_post, bind_arg);
+      if (node->info.update.spec->info.spec.remote_server_name == NULL)
```
shparkcubrid, line 2875: Why doesn't db_link do it?

src/parser/name_resolution.c
```diff
@@ -4268,6 +4530,7 @@ pt_flat_spec_pre (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int *conti
 			      pt_short_print_l (parser, node->info.spec.as_attr_list));
 		}
 	    }
+	  /* dblink query */
```
shparkcubrid: Is this case only for db_link?

src/parser/view_transform.c
```diff
@@ -5335,6 +5336,13 @@ mq_translate_insert (PARSER_CONTEXT * parser, PT_NODE * insert_statement)
 	  flat = from->info.spec.flat_entity_list;
 	  if (flat == NULL)
 	    {
+	      if (from_spec->remote_server_name)
```
shparkcubrid, line 5344: Isn't that unnecessary code?

src/query/execute_statement.c
```diff
@@ -5700,17 +5700,24 @@ check_trigger (DB_TRIGGER_EVENT event, PT_DO_FUNC * do_func, PARSER_CONTEXT * pa
 	  if (node->info.spec.flag & PT_SPEC_FLAG_DELETE)
 	    {
 	      flat = node->info.spec.flat_entity_list;
-	      class_ = (flat ? flat->info.name.db_object : NULL);
-	      if (class_ == NULL)
+	      if (node->info.spec.remote_server_name)
```
shparkcubrid, line 5703: How about 'IS_DBLINK_DML' for readability?

shparkcubrid (approved):

src/parser/view_transform.c
beyondykk9: the spec must be remote-server under the flat_entity_list is NULL but I keep it for readability.

src/compat/db_vdb.c
beyondykk9: Ok. good point

src/query/execute_statement.c
beyondykk9: I'd like to keep the logic to check remote server name easily.

src/compat/db_vdb.c
beyondykk9: Ok, I modified it.

src/parser/name_resolution.c
beyondykk9: Ok, I'll remove the comment.

src/parser/name_resolution.c
beyondykk9: remote DML (update/delete/insert/merge) is handled at server-side and the DML query is rewriten at pt_rewrite_dblink.

src/compat/db_vdb.c
```diff
@@ -658,37 +671,48 @@ db_compile_statement_local (DB_SESSION * session)
...
-	    {
-	      (void) pt_add_row_oid (parser, statement);
-	    }
+	  pt_report_to_ersys_with_statement (parser, PT_SYNTAX, statement);
+	  return er_errid ();
 	}
-      else
+
+      /* validate include_oid setting in the session */
+      if (session->include_oid)
```
hgryoo: session->include_oid is not a boolean.
```suggestion
      if (session->include_oid != DB_NO_OIDS)
```

src/parser/name_resolution.c
```diff
@@ -243,7 +251,139 @@ static PT_NODE *pt_get_attr_list_of_derived_table (PARSER_CONTEXT * parser, PT_M
...
+    if (m_nm_buf)
+      {
+	free (m_nm_buf);
+      }
+    if (m_attr)
+      {
+	free (m_attr);
+      }
+  }
+
```
hgryoo, line 340: Make the struct non-copyable.
```suggestion
  remote_tbl_cols (const remote_tbl_cols &) = delete;
  remote_tbl_cols& operator=(const remote_tbl_cols &) = delete;
```

src/parser/name_resolution.c
```diff
@@ -10530,13 +11279,285 @@ pt_resolve_dblink_server_name (PARSER_CONTEXT * parser, PT_NODE * node)
...
+    }
+  return node;
+}
+
+static PT_NODE *
+pt_get_column_name_post (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int *continue_walk)
+{
+  *continue_walk = PT_CONTINUE_WALK;
+  return node;
+}
```
hgryoo, line 11488: Why is this function required? Passing NULL to the post_function arg in `parser_walk_tree()` seems to be enough
```suggestion
(void) parser_walk_tree (parser, node_list,
			   pt_get_column_name_pre, plkcol, NULL, /*plkcol */ NULL);
```

hgryoo (approved):

src/compat/db_vdb.c
beyondykk9: This is not related to this issue, but it looks need to fix for readability.

hyunikn (approved):

src/parser/name_resolution.c
beyondykk9: Ok, good point.

src/parser/name_resolution.c
beyondykk9: Ok.

src/parser/csql_grammar.y
```diff
@@ -3155,6 +3160,24 @@ create_stmt
...
+class_name_for_synonym
+	: class_name
+		{ $$ = $1; }
+	| class_name_with_server_name
+		{
+		  PT_NODE *cname = CONTAINER_AT_0 ($1);
+		  PT_NODE *sname = CONTAINER_AT_1 ($1);
+
+		  cname->info.name.original = pt_append_string (this_parser, cname->info.name.original, "@");
+		  cname->info.name.original = pt_append_string (this_parser, cname->info.name.original, sname->info.name.original);
```
youngjinj, line 3173: I think it would be better to do `parser_free_tree(this_parser, sname);`.

src/parser/csql_grammar.y
```diff
@@ -5341,31 +5395,53 @@ meta_class_spec
...
 		DBG_PRINT}}
 	;
 
-only_all_class_spec
-	: only_class_name opt_partition_spec
-		{{ DBG_TRACE_GRAMMAR(only_all_class_spec, : only_class_name opt_partition_spec );
-
-			PT_NODE *ocs = parser_new_node (this_parser, PT_SPEC);
+only_all_class_spec_with_server
+	: class_name_with_server_name opt_partition_spec
```
youngjinj: I think the `opt_partition_spec` rule should be excluded.

youngjinj:
I understand that the queries below are not supported.
But the below queries succeed.
I think it should be blocked from the syntax so that users do not misunderstand.

```
/* remote server */
create user u1 password 'password';

drop table if exists u1.t1;
create table u1.t1 (c1 int);
insert into t1 values (1);
```
```
/* local server */
create user u2;

drop server cubrid_dblink;
create server cubrid_dblink ( host='192.168.2.206', port=33000, dbname=demodb, user=u1, password='password');

select * from dblink (cubrid_dblink, 'select c1 from t1') as t1 (c1 int);
/*
           c1
=============
            1
*/
```
1. Query succeeds but not working
```
truncate t1@cubrid_dblink; /* Execute OK. */
drop t1@cubrid_dblink; /* Execute OK. */
grant select on t1@cubrid_dblink to u2; /* Execute OK. */
revoke select on t1@cubrid_dblink from u2; /* Execute OK. */
```
2. I know that DML is allowed, but the following fails.
```
delete from t1@cubrid_dblink;
/* Class t1 does not exist. */
```
3. If possible, I think the query below should also be blocked syntactically.
```
scope___ REJECT from on t1@cubrid_dblink; /* Execute OK. */
```

src/parser/csql_grammar.y
ctshim: Good point!
It needs to be fixed.

ctshim (approved):

youngjinj (approved):

src/parser/csql_grammar.y
beyondykk9: It's fixed.

PR #4065, opened 2023-01-19
[CBRD-24599] Add new csql session command 'SingleLine'

eido5:
http://jira.cubrid.org/browse/CBRD-24599

**Purpose**

- In csql, there are two modes to handle sql statements.

        single line (with semicolon, handling sql each line)
        --no-single-line (handling the sql at once with ;xr or ;r session command)
        

- To csql change mode, we must quit the csql and restart with '--no-single-line' option or not. 
So, we need to switch the both modes while csql is working.
This session command will help user use csql more convenient.

**Implementation**

Add new csql session command.

**Remarks**

N/A

### DIFF
--- a/src/executables/csql.c
+++ b/src/executables/csql.c
@@ -1439,6 +1439,17 @@ csql_do_session_cmd (char *line_read, CSQL_ARGUMENT * csql_arg)
 	  fprintf (csql_Error_fp, "Auto trace isn't allowed in SA mode.\n");
 	}
       break;
+
+    case S_CMD_SINGLELINE:
+      if (!strcasecmp (argument, "on"))
+	{
+	  csql_arg->single_line_execution = true;
+	}
+      else if (!strcasecmp (argument, "off"))
+	{
+	  csql_arg->single_line_execution = false;
+	}
+      break;
     }
 
   return DO_CMD_SUCCESS;
--- a/src/executables/csql.c
+++ b/src/executables/csql.c
@@ -1449,8 +1449,10 @@ csql_do_session_cmd (char *line_read, CSQL_ARGUMENT * csql_arg)
 	{
 	  csql_arg->single_line_execution = false;
 	}
-
-      fprintf (csql_Output_fp, "SINGLELINE IS %s\n", (csql_arg->single_line_execution ? "ON" : "OFF"));
+      if (csql_Is_interactive)
+	{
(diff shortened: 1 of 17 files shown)


src/executables/csql.c
```diff
@@ -1439,6 +1439,17 @@ csql_do_session_cmd (char *line_read, CSQL_ARGUMENT * csql_arg)
...
+
+    case S_CMD_SINGLELINE:
+      if (!strcasecmp (argument, "on"))
+	{
+	  csql_arg->single_line_execution = true;
+	}
+      else if (!strcasecmp (argument, "off"))
+	{
+	  csql_arg->single_line_execution = false;
+	}
```
kisoo-han, line 1451: When the csql mode is '**_interactive_**', you have to print out current SINGLELINE MODE,
for example
```
csql> ;single
SINGLELINE IS ON
csql> ;single off
SINGLELINE IS OFF
```
Please refer the session command '**AUTOCOMMIT**'.

kisoo-han (approved):

kisoo-han:
I think its better to include 'SINGLELINE' to the '**csql HELP**', with this PR.

src/executables/csql.c
eido5: Thank you. I will ask you review again after I will add the function which is the status of 'SingleLine' mode

src/executables/csql.c
```diff
@@ -1449,6 +1449,9 @@ csql_do_session_cmd (char *line_read, CSQL_ARGUMENT * csql_arg)
 	{
 	  csql_arg->single_line_execution = false;
 	}
+
+      fprintf (csql_Output_fp, "SINGLELINE IS %s\n", (csql_arg->single_line_execution ? "ON" : "OFF"));
```
kisoo-han: Print only when it is csql interactive (check for the variable **csql_Is_interactive** is **true**).

kisoo-han:
1. update csql help display a line for 'SINGLELINE'.
2. update your repo with base repository to remove circle-ci fails (test_sql, test_sql_ol)

src/executables/csql.c
eido5: I understand. I will apply it.

airnet73 (approved):

kisoo-han (approved):
Ok, go ahead for other languages.

Srltas (approved):

PR #4066, opened 2023-01-19
[CBRD-24632] Change data put, get and support functions using OR_BUF in object_representation.c to inline functions

youngjinj:
http://jira.cubrid.org/browse/CBRD-24632

Change data put, get and support functions using OR_BUF in object_representation.c to inline functions


hgryoo:
To fix the unresolved external symbol error, please try EXPORT_IMPORT at the function declaration.
e.g. 
```
EXPORT_IMPORT EXTERN_INLINE int or_put_offset_internal (OR_BUF * buf, int num, int offset_size) __attribute__ ((ALWAYS_INLINE));
```

src/storage/byte_order.c
```diff
@@ -0,0 +1,197 @@
...
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+
+/*
+ * byte_order.c - Definitions related to disk storage order
```
hgryoo, line 20: CUBRID is currently supporting only LINUX and Windows. <arpa/inet.h> and <winsock2.h> implement the byte ordering functions in Linux and Windows, respectively. And CUBRID's target OS seems not to be changed in the future. Why do we need these declarations and implementations for unsupported OSs?

For reference:
```
#if defined(HPUX) || defined(_AIX) || defined(sparc)
#define OR_BYTE_ORDER OR_BIG_ENDIAN
#else /* HPUX || _AIX || sparc */
#define OR_BYTE_ORDER OR_LITTLE_ENDIAN	/* WINDOWS, LINUX, x86_SOLARIS */
#endif /* HPUX || _AIX || sparc */

#if defined(HPUX) || defined(_AIX) || defined(WINDOWS) || defined(LINUX)
#define OR_HAVE_NTOHS
#define OR_HAVE_NTOHL
#define OR_HAVE_HTONS
#define OR_HAVE_HTONL
#endif /* HPUX || _AIX || WINDOWS || LINUX */
```

shparkcubrid (approved):

src/storage/byte_order.c
youngjinj: It was moved as-is to minimize the impact of code changes.

I think removing code for other OS should be another issue.
I will create a separate issue for this and share it.

youngjinj: I created the issue below.
http://jira.cubrid.org/browse/CBRD-24646

src/base/porting_inline.hpp
```diff
@@ -34,14 +34,17 @@
...
+#define EXTERN_INLINE __forceinline extern
 #define STATIC_INLINE __forceinline static
 #define INLINE __forceinline
 #else
 /* TODO: we have several cases of using INLINE/STATIC_INLINE and adding function definition in headers. This won't
  * work. */
+#define EXTERN_INLINE extern
 #define STATIC_INLINE static
 #define INLINE
 #endif
```
hgryoo: In other code, They only use STATIC_INLINE (dbtype_function.i) even though the functions are exposed. Could you explain why we need EXTERN_INLINE for this case? What is the difference between extern inline and static inline?

src/base/object_representation.h
```diff
@@ -85,167 +85,128 @@ struct setobj;
 #define OR_CHECK_FLOAT_OVERFLOW(i)         ((i) > FLT_MAX || (-(i)) > FLT_MAX)
 #define OR_CHECK_DOUBLE_OVERFLOW(i)        ((i) > DBL_MAX || (-(i)) > DBL_MAX)
 
+#if __WORDSIZE == 32
+#define OR_PTR_SIZE             4
+#define OR_PUT_PTR(ptr, val)    OR_PUT_INT ((ptr), (val))
```
beyondykk9: OR_PUT_INT is defined more belower from this. This could produce unintentional error at build time.

src/base/object_representation.h
```diff
@@ -85,167 +85,128 @@ struct setobj;
...
+#define OR_GET_PTR(ptr)         OR_GET_INT ((ptr))
```
beyondykk9: OR_GET_INT is defined more belower from this. This could produce unintentional error at build time.

src/base/object_representation.h
```diff
@@ -1416,30 +1397,955 @@ extern int or_put_json_schema (OR_BUF * buf, const char *schema);
...
+EXTERN_INLINE DB_BIGINT
+or_get_bigint (OR_BUF * buf, int *error)
+{
+  DB_BIGINT value = 0;
+
+  ASSERT_ALIGN (buf->ptr, INT_ALIGNMENT);
+
+  if ((buf->ptr + OR_BIGINT_SIZE) > buf->endptr)
+    {
+      *error = or_underflow (buf);
```
beyondykk9, line 1864: no need to check null pointer for **error**?

src/base/porting_inline.hpp
youngjinj: I will use static inline consistently.

TMI.
when compiling with g++, there was no difference between static inline, inline and extern inline in assembly code.
However, When compiling with gcc, multiple definition errors occur when extern inline is used.
We are compiling our c files with g++, so no error occurs.

src/base/object_representation.h
youngjinj: Changed the location of the code.

src/base/object_representation.h
youngjinj: Changed the location of the code.

src/base/object_representation.h
```diff
@@ -418,6 +382,42 @@ OR_PUT_DOUBLE (char *ptr, double val)
...
+      } \
+    else if (offset_size == OR_SHORT_SIZE) \
+      { \
+        OR_PUT_SHORT(ptr, val); \
+      } \
+    else if (offset_size == OR_INT_SIZE) \
+      { \
+        OR_PUT_INT(ptr, val); \
+      } \
+  } while (0)
```
ctshim: Let's check the int type first. 
The same applies to OR_GET_OFFSET_INTERNAL().

ctshim:
The following three functions are rarely used:
I think making it inline is more inefficient.

or_overflow, or_underfolw, or_abort

hyunikn (approved):

src/base/object_representation.h
youngjinj: OR_PUT_OFFSET and OR_GET_OFFSET change to direct calls to OR_PUT_BIG_VAR_OFFSET and OR_GET_BIG_VAR_OFFSET respectively.
This way we don't have to change OR_INT_SIZE to compare first.

youngjinj:
As @ctshim commented, the 3 functions below are rarely called.
So I won't change it to inline.
- or_overflow()
- or_underflow()
- or_abort()

beyondykk9 (approved):

hgryoo (approved):

ctshim (approved):

PR #4071, opened 2023-01-19
[LETS-594] reduce page dead latch occurences by reducing fix interval in atomic replication

cristiarg:
http://jira.cubrid.org/browse/LETS-594

- the call path `atomic_replication_helper::append_log` -> `atomic_replication_helper::atomic_log_sequence::append_log` just adds info for the log record without actually fixing a page
- the function `atomic_log_sequence::apply_and_unfix` now does the heavy-lifting of checking what needs to be applied and applying fixes:
  - first fixes all pages as dictated by the log records in the atomic [sub]sequence
  - it uses an `page_ptr_bookkeeping` for this as there can be more than one log records for the same page
  - then, as it progresses with executing `log_entry.apply_log_redo` for each log record, it leverages the page bookkeeping to unfix pages as it progresses

Other:
- added some asserts and had to comment others
- removed one aspect of the `atomic_replication_helper` unit test that deals with checking the functionality that protects against fail-to-fix pages - see `UTEST_LOG_REPL_ATOMIC_TEST_FIX_FAIL`; I found no way to actually maintain this test with the new modifications (it would mean instrumenting the class `atomic_replication_helper` via derivation/overloading, and it just seems to over-complicated); it will be re-activated/re-evaluated if needed


src/transaction/atomic_replication_helper.cpp
```diff
@@ -889,6 +888,13 @@ namespace cublog
     // sysop end parent lsa can also be null
   }
 
+  atomic_replication_helper::atomic_log_sequence::atomic_log_entry::~atomic_log_entry ()
+  {
+    // this assert should be valid; except we're using emplaced constructed objects for which
+    // we're not properly initializing this value
+    //assert (m_page_ptr == nullptr);
```
hornetmj:
```suggestion
    assert (m_page_ptr == nullptr);
```

hornetmj: Or intended?

hornetmj (approved):

src/transaction/atomic_replication_helper.cpp
cristiarg: No, it is intended. The comment explains the cause.
I have tried to alter the code to make the assert pass but I was not able to.
The reason is that these objects are stored in a vector - `std::vector<atomic_log_entry>` - while operating the vector, the contents will be moved move semantics (move ctor, move assignment operator). Although I have added `m_page_ptr == nullptr` initializations all over the place, this assert still failed in the unit tests.

However I have removed the dtor altogether as it became superfluous - being empty.

wrlawodms:
@cristiarg 
I remember we talked (or you talked,  I don't remember) about the cases the delaying fixing would be problematic: 

Log sequence: T2(atomic op start) T2(p1) T1(p1) T2(atomic op end)
(TN: Transaction n, pN: modify page N)

In the above sequence, if fixing p1 for T2(p1) is delayed until the end of atomic op, the order of T1(p1) and T2(p1) is mixed and it results in a different data page image. Actually, T1(p1) is dependent on T2(p1) since it's physical logging. 

Is this kind of problem solved? or considered in this issue?

cristiarg:
> @cristiarg I remember we talked (or you talked, I don't remember) about the cases the delaying fixing would be problematic:
> 
> Log sequence: T2(atomic op start) T2(p1) T1(p1) T2(atomic op end) (TN: Transaction n, pN: modify page N)
> 
> In the above sequence, if fixing p1 for T2(p1) is delayed until the end of atomic op, the order of T1(p1) and T2(p1) is mixed and it results in a different data page image. Actually, T1(p1) is dependent on T2(p1) since it's physical logging.
> 
> Is this kind of problem solved? or considered in this issue?

@wrlawodms 

No, what you describe is not covered by this patch.
What is covered is when such a scenario happens between two atomic sequences. This check is implemented in `vpid_bookeeping` class.

This scenario is already described in http://jira.cubrid.org/browse/LETS-570 .
After the recent modifications in the atomic replication helper (concluded with this patch), I think we can extend the use of `vpid_bookeeping` to check also pages that are not part of an atomic sequence.

Of course, checking is one thing. But that does not solve the actual problem.
To solve the problem properly it is needed to amend the place where the log is generated in the first place - on Active Transaction Server. It has been discussed about it.

So, for the check you suggested, I will continue with the check as part of LETS-570 .

wrlawodms (approved):

PR #4073, opened 2023-01-24
[LETS-425] small refactor; remove functions; TBD

cristiarg:

- functions `get_highest_processed_lsa` and `get_lowest_unapplied_lsa` are only used in the derived `atomic_replicator` class
- removed these from base class `replicator` and adapted code in `passive_tran_server` class
- comments

### DIFF
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -210,7 +210,7 @@ page_server::connection_handler::receive_oldest_active_mvccid (tran_server_conn_
 {
   assert (m_server_type == transaction_server_type::PASSIVE);
 
-  const auto oldest_mvccid = *reinterpret_cast<const MVCCID *const> (a_sp.pull_payload().c_str());
+  const auto oldest_mvccid = *reinterpret_cast<const MVCCID *> (a_sp.pull_payload().c_str());
 
   m_ps.m_pts_mvcc_tracker.update_oldest_active_mvccid (get_connection_id (), oldest_mvccid);
 }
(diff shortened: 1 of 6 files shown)


src/server/page_server.cpp
```diff
@@ -210,7 +210,7 @@ page_server::connection_handler::receive_oldest_active_mvccid (tran_server_conn_
 {
   assert (m_server_type == transaction_server_type::PASSIVE);
 
-  const auto oldest_mvccid = *reinterpret_cast<const MVCCID *const> (a_sp.pull_payload().c_str());
+  const auto oldest_mvccid = *reinterpret_cast<const MVCCID *> (a_sp.pull_payload().c_str());
```
wrlawodms, line 213: Just out of curiosity, would you tell me why this is better? Thanks.

wrlawodms (approved):

src/server/page_server.cpp
cristiarg: It is just a warning reported by gcc9.
Basically, it had no effect as you want to convert  some memory address to some data type. It is only then that you either assign the result of that conversion to a variable with either const-volatile qualifiers. The conversion itself has no role in that assignment. 

In this case, the result of the conversion is even dereferenced and the value copied locally. 

For context, this article explains some aspect which, I think, is also involved in the above:

http://www.dansaks.com/articles/2000-02%20Top-Level%20cv-Qualifiers%20in%20Function%20Parameters.pdf

PR #4080, opened 2023-01-30
[CBRD-24645] When order-by of subquery is copied to main query during view merging, rownum of main query is displayed incorrectly

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-24645

When 'ORDER BY' is copied to the main query during view-merging, I fix so that 'ROWNUM' of the select list of the main query is replaced with 'ORDERBY_NUM'.
Since expression having orderby_num() such as 'orderby_num() + 1' is not supported, expr having rownum such as 'rownum+1' is restricted from view merging.
Hopefully, I'll take care of it later.

### DIFF
--- a/src/parser/parser_support.c
+++ b/src/parser/parser_support.c
@@ -3354,7 +3354,7 @@ pt_has_inst_num (PARSER_CONTEXT * parser, PT_NODE * node)
 }
 
 /*
- * pt_has_inst_num () - check if tree has an EXPR node having INST_NUM in select list
+ * pt_has_expr_of_inst_in_sel_list  () - check if tree has an EXPR node having INST_NUM in select list
  *   return: true if tree has EXPR node having INST_NUM
  *   parser(in):
  *   node(in):
--- a/src/parser/parser_support.c
+++ b/src/parser/parser_support.c
@@ -3354,6 +3354,44 @@ pt_has_inst_num (PARSER_CONTEXT * parser, PT_NODE * node)
 }
 
 /*
+ * pt_has_inst_num () - check if tree has an EXPR node having INST_NUM in select list
+ *   return: true if tree has EXPR node having INST_NUM
+ *   parser(in):
+ *   node(in):
+ */
+bool
+pt_has_expr_of_inst_in_sel_list (PARSER_CONTEXT * parser, PT_NODE * select_list)
+{
+  PT_NODE *save_next, *col;
+
+  /*
+   * FIXME!! : remove this check after expanding syntax related to orderby_num().
+   * This check is needed because the syntax below is not supported.
+   * 'select orderby_num() + 10 ...'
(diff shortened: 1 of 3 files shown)


src/parser/parser_support.c
```diff
@@ -3353,6 +3353,44 @@ pt_has_inst_num (PARSER_CONTEXT * parser, PT_NODE * node)
   return has_inst_num;
 }
 
+/*
+ * pt_has_inst_num () - check if tree has an EXPR node having INST_NUM in select list
```
shparkcubrid:
```suggestion
 * pt_has_expr_of_inst_in_sel_list  () - check if tree has an EXPR node having INST_NUM in select list
```

src/parser/view_transform.c
```diff
@@ -1669,7 +1669,7 @@ static PUSHABLE_TYPE
 mq_is_pushable_subquery (PARSER_CONTEXT * parser, PT_NODE * subquery, PT_NODE * mainquery, PT_NODE * class_spec,
 			 bool is_vclass, PT_NODE * order_by, PT_NODE * class_)
 {
-  PT_NODE *pred, *statement_spec = NULL, *orderby_for;
+  PT_NODE *pred, *statement_spec = NULL, *orderby_for, *select_list;
```
hgryoo, line 1672: How about initializing it ahead?

```suggestion
  PT_NODE *pred, *statement_spec = NULL, *orderby_for, *select_list = NULL;
```

hyunikn: It is unnecessary in case of PT_SELECT.

ctshim (approved):

hyunikn (approved):

src/parser/view_transform.c
```diff
@@ -2162,30 +2168,29 @@ mq_update_order_by (PARSER_CONTEXT * parser, PT_NODE * statement, PT_NODE * quer
 
   statement->info.query.order_by = parser_append_node (order_by, statement->info.query.order_by);
 
-  where = statement->info.query.q.select.where;
-  if (where != NULL && PT_EXPR_INFO_IS_FLAGED (where, PT_EXPR_INFO_ROWNUM_ONLY) && statement->info.query.order_by)
+  /* generate orderby_num(), inst_num() */
+  if (!(ord_num = parser_new_node (parser, PT_EXPR)) || !(ins_num = parser_new_node (parser, PT_EXPR)))
```
hgryoo, line 2172: possibly memory leak if `ord_num = parser_new_node (parser, PT_EXPR)` is true and `ins_num = parser_new_node (parser, PT_EXPR)` is false.

hgryoo (approved):

youngjinj (approved):

beyondykk9 (approved):

src/parser/view_transform.c
shparkcubrid: I fixed it.

PR #4081, opened 2023-01-31
[CBRD-24644] 64bit statdump output for CMS the clients

kisoo-han:
http://jira.cubrid.org/browse/CBRD-24644

**Purpose**
* '**cubrid statdump**' supports 64bit statistics, for example, **Num_query_selects**
* However, CMS read statistical elements by '%u' and assign it to integer before it send to the clients
* For this reason, even 'cubrid statdump' can handle 64bit elements, CMS client will receive negative values if it is greater than I**NTEGER MAX**.

**Implementation**
* Due to CMS uses cm_common library to extract such a value, both CUBRID Engine and CMS should be modified together

**Remarks**

### DIFF
--- a/src/cm_common/cm_mem_cpu_stat.c
+++ b/src/cm_common/cm_mem_cpu_stat.c
@@ -1271,9 +1271,9 @@ extract_db_exec_stat (FILE * fp, const char *dbname, T_CM_ERROR * err_buf)
   while (fgets (linebuf, sizeof (linebuf), fp))
     {
       unsigned int *member_ptr;
-      unsigned int prop_val;
+      uint64_t prop_val;
       memset (prop_name, 0, sizeof (prop_name));
-      sscanf (linebuf, "%99s%*s%u", prop_name, &prop_val);
+      sscanf (linebuf, "%99s%*s%lu", prop_name, &prop_val);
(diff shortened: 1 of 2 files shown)


mhoh3963 (approved):

src/cm_common/cm_mem_cpu_stat.c
```diff
@@ -1271,9 +1271,9 @@ extract_db_exec_stat (FILE * fp, const char *dbname, T_CM_ERROR * err_buf)
   while (fgets (linebuf, sizeof (linebuf), fp))
     {
       unsigned int *member_ptr;
-      unsigned int prop_val;
+      uint64_t prop_val;
       memset (prop_name, 0, sizeof (prop_name));
-      sscanf (linebuf, "%99s%*s%u", prop_name, &prop_val);
+      sscanf (linebuf, "%99s%*s%lu", prop_name, &prop_val);
```
airnet73, line 1276: %lu is 4 bytes on windows and 8 bytes on centos.
If you want to satisfy both windows and centos, %llu seems to be good.

kisoo-han: Thank you for your comment!
By the way, changes to my repo are not reflected in this PR. I don't know why. So I want to close this PR and open it again.

PR #4082, opened 2023-01-31
[LETS-85] Active transaction server sends the log to all page servers

wrlawodms:
http://jira.cubrid.org/browse/LETS-85

- Add prior sinks for each connected page server connection.
- Set `exit_code = NO_ERROR` when a pager server connection fails to start ATS without it.
- Add `push_request_to` to request to a specific server with an index.
- I don't handle the case where some PS are disconnected while ATS is running. It's going to be handled in another issue.

### DIFF
--- a/src/server/active_tran_server.cpp
+++ b/src/server/active_tran_server.cpp
@@ -97,9 +97,10 @@ active_tran_server::on_boot ()
 
   for (size_t i = 0; i < get_connected_page_server_count (); i++)
     {
-      m_prior_sender_sink_hooks.emplace_back (std::bind (static_cast<request_functor_t> (&active_tran_server::push_request),
-					      this, i, tran_to_page_request::SEND_LOG_PRIOR_LIST, std::placeholders::_1));
-      log_Gl.m_prior_sender.add_sink (m_prior_sender_sink_hooks.back ());
+      m_prior_sender_sink_hooks.emplace_back (std::make_unique <cublog::prior_sender::sink_hook_t> (
+	  std::bind (static_cast<request_functor_t> (&active_tran_server::push_request),				      this, i,
+		     tran_to_page_request::SEND_LOG_PRIOR_LIST, std::placeholders::_1)));
+      log_Gl.m_prior_sender.add_sink (*m_prior_sender_sink_hooks.back ());
     }
 }
 
@@ -123,6 +124,6 @@ active_tran_server::stop_outgoing_page_server_messages ()
 {
   for (const auto &sink : m_prior_sender_sink_hooks)
     {
-      log_Gl.m_prior_sender.remove_sink (sink);
+      log_Gl.m_prior_sender.remove_sink (*sink);
     }
 }
--- a/src/server/active_tran_server.cpp
+++ b/src/server/active_tran_server.cpp
@@ -92,11 +92,13 @@ active_tran_server::on_boot ()
 {
   assert (is_active_transaction_server ());
 
-  m_prior_sender_sink_hook_func =
-	  std::bind (&active_tran_server::push_request, std::ref (*this), tran_to_page_request::SEND_LOG_PRIOR_LIST,
-		     std::placeholders::_1);
-
-  log_Gl.m_prior_sender.add_sink (m_prior_sender_sink_hook_func);
+  for (size_t i = 0; i < get_connected_page_server_count(); i++)
+    {
+      m_prior_sender_sink_hooks.emplace_back (std::bind (&active_tran_server::push_request_to, this, i,
+					      tran_to_page_request::SEND_LOG_PRIOR_LIST,
+					      std::placeholders::_1));
+      log_Gl.m_prior_sender.add_sink (m_prior_sender_sink_hooks.back ());
+    }
 }
 
 active_tran_server::request_handlers_map_t
@@ -117,4 +119,8 @@ active_tran_server::get_request_handlers ()
 void
 active_tran_server::stop_outgoing_page_server_messages ()
 {
+  for (const auto &sink : m_prior_sender_sink_hooks)
+    {
+      log_Gl.m_prior_sender.remove_sink (sink);
+    }
 }
--- a/src/server/active_tran_server.cpp
+++ b/src/server/active_tran_server.cpp
@@ -92,14 +92,10 @@ active_tran_server::on_boot ()
 {
   assert (is_active_transaction_server ());
 
-  /* the signature of the push_request */
-  using request_functor_t = void (active_tran_server::*) (size_t, tran_to_page_request, std::string &&);
-
   for (size_t i = 0; i < get_connected_page_server_count (); i++)
     {
-      m_prior_sender_sink_hooks.emplace_back (std::make_unique <cublog::prior_sender::sink_hook_t> (
-	  std::bind (static_cast<request_functor_t> (&active_tran_server::push_request),				      this, i,
-		     tran_to_page_request::SEND_LOG_PRIOR_LIST, std::placeholders::_1)));
+      m_prior_sender_sink_hooks.emplace_back (std::make_unique <cublog::prior_sender::sink_hook_t> (std::bind (
+	  &active_tran_server::push_request_to, this, i, tran_to_page_request::SEND_LOG_PRIOR_LIST, std::placeholders::_1)));
       log_Gl.m_prior_sender.add_sink (*m_prior_sender_sink_hooks.back ());
     }
 }
--- a/src/server/active_tran_server.cpp
+++ b/src/server/active_tran_server.cpp
@@ -92,7 +92,7 @@ active_tran_server::on_boot ()
 {
   assert (is_active_transaction_server ());
 
-  for (size_t i = 0; i < get_connected_page_server_count(); i++)
+  for (size_t i = 0; i < get_connected_page_server_count (); i++)
     {
       m_prior_sender_sink_hooks.emplace_back (std::bind (&active_tran_server::push_request_to, this, i,
 					      tran_to_page_request::SEND_LOG_PRIOR_LIST,
--- a/src/server/active_tran_server.cpp
+++ b/src/server/active_tran_server.cpp
@@ -92,11 +92,13 @@ active_tran_server::on_boot ()
 {
   assert (is_active_transaction_server ());
 
+  /* the signature of the push_request */
+  using request_functor_t = void (active_tran_server::*) (size_t, tran_to_page_request, std::string &&);
+
   for (size_t i = 0; i < get_connected_page_server_count (); i++)
     {
-      m_prior_sender_sink_hooks.emplace_back (std::bind (&active_tran_server::push_request_to, this, i,
-					      tran_to_page_request::SEND_LOG_PRIOR_LIST,
-					      std::placeholders::_1));
+      m_prior_sender_sink_hooks.emplace_back (std::bind (static_cast<request_functor_t> (&active_tran_server::push_request),
+					      this, i, tran_to_page_request::SEND_LOG_PRIOR_LIST, std::placeholders::_1));
       log_Gl.m_prior_sender.add_sink (m_prior_sender_sink_hooks.back ());
     }
 }
--- a/src/server/active_tran_server.hpp
+++ b/src/server/active_tran_server.hpp
@@ -22,7 +22,8 @@
 #include "log_prior_send.hpp"
 #include "tran_server.hpp"
 
-#include <list>
+#include <vector>
+#include <memory>
 
 class active_tran_server : public tran_server
 {
@@ -51,7 +52,7 @@ class active_tran_server : public tran_server
      * sends prior nodes to the page servers.
      * The order can differ from m_page_server_conn_vec
      */
-    std::list<cublog::prior_sender::sink_hook_t> m_prior_sender_sink_hooks;
+    std::vector<std::unique_ptr<cublog::prior_sender::sink_hook_t>> m_prior_sender_sink_hooks;
 };
 
 #endif // !_ACTIVE_TRAN_SERVER_HPP_
--- a/src/server/active_tran_server.hpp
+++ b/src/server/active_tran_server.hpp
@@ -22,6 +22,8 @@
 #include "log_prior_send.hpp"
 #include "tran_server.hpp"
 
+#include <list>
+
 class active_tran_server : public tran_server
 {
   public:
@@ -45,7 +47,11 @@ class active_tran_server : public tran_server
   private:
     bool m_uses_remote_storage = false;
 
-    cublog::prior_sender::sink_hook_t m_prior_sender_sink_hook_func;
+    /*
+     * sends prior nodes to the page servers.
+     * The order can differ from m_page_server_conn_vec
+     */
+    std::list<cublog::prior_sender::sink_hook_t> m_prior_sender_sink_hooks;
 };
 
 #endif // !_ACTIVE_TRAN_SERVER_HPP_
--- a/src/server/tran_server.cpp
+++ b/src/server/tran_server.cpp
@@ -217,7 +217,9 @@ tran_server::init_page_server_hosts (const char *db_name)
     {
       //at least one valid host exists clear the error remaining from previous failing ones
       er_clear ();
+      exit_code = NO_ERROR;
     }
+
   // validate connections vs. config
   //
   if (valid_connection_count == 0 && uses_remote_storage)
@@ -358,14 +360,22 @@ tran_server::uses_remote_storage () const
 }
 
 void
-tran_server::push_request (tran_to_page_request reqid, std::string &&payload)
+tran_server::push_request_to (size_t idx, tran_to_page_request reqid, std::string &&payload)
 {
+  assert (idx < m_page_server_conn_vec.size());
   if (!is_page_server_connected ())
     {
       return;
     }
 
-  m_page_server_conn_vec[0]->push (reqid, std::move (payload));
+  m_page_server_conn_vec[idx]->push (reqid, std::move (payload));
+}
+
+void
+tran_server::push_request (tran_to_page_request reqid, std::string &&payload)
+{
+  // TODO push a request to the "main connection"
+  push_request_to (0, reqid, std::move (payload));
 }
 
 int
@@ -373,6 +383,7 @@ tran_server::send_receive (tran_to_page_request reqid, std::string &&payload_in,
 {
   assert (is_page_server_connected ());
 
+  // TODO push a request to the "main connection"
   const css_error_code error_code = m_page_server_conn_vec[0]->send_recv (reqid, std::move (payload_in), payload_out);
   // NOTE: enhance error handling when:
   //  - more than one page server will be handled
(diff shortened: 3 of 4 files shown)


src/server/active_tran_server.hpp
```diff
@@ -45,7 +47,11 @@ class active_tran_server : public tran_server
   private:
     bool m_uses_remote_storage = false;
 
-    cublog::prior_sender::sink_hook_t m_prior_sender_sink_hook_func;
+    /*
+     * sends prior nodes to the page servers.
+     * The order can differ from m_page_server_conn_vec
+     */
+    std::list<cublog::prior_sender::sink_hook_t> m_prior_sender_sink_hooks;
```
wrlawodms: @cristiarg I don't use `std::vector` because the functor `sink_hook_t` has no `noexcept` move constructor so the vector copies them when it grows (until c++20), for which in turn `prior_sender:: m_sink_hooks` may refer to a removed object.

Please share your opinion if you have a better idea.

cristiarg: maybe you can use `unique_ptr` as is done in `tran_server::m_page_server_conn_vec`

src/server/tran_server.cpp
```diff
@@ -217,7 +217,9 @@ tran_server::init_page_server_hosts (const char *db_name)
     {
       //at least one valid host exists clear the error remaining from previous failing ones
       er_clear ();
+      exit_code = NO_ERROR;
```
cristiarg, line 218: was this a bug?

src/server/active_tran_server.hpp
cristiarg: However, this is more complicated here.

I think we need a design of objects that is able to fulfill the following functional requirements:
- a PTS must be able to connect to multiple PS's, disconnect from some (when those Ps's become unavailable; connect to new ones when those are "advertised" to the cluster; ... and whatnot ..)
- a PTS must be able to fail-over as an ATS
- but, at the same time, a PTS will not need to send prior log data to the connected PS's (this is subjective, the functionality "could" be formally hidden from the PTS but does not need to be - we'll not do this as it over-complicates things even further)
- an ATS must be able to send prior log data to connected PS's
- an ATS (and maybe PTS's too) must be able to interrogate the connected PS's about all kinds of information that will be needed to implement the consensus logic
- ...

We have more than one way to achieve the above:

1. Simple way:
We need a way to identify the same PS across different structures (ie: not by index as it is now)
We have the following options:
- by `cubcomm::channe::get_channel_id` (as is done with `page_server::connection_handler::get_connection_id`)
- by `cubcomm::node` information - "host:port" (although only host might be enough)
Then we can use this id to juggle different maps

2. More formal way:
I think we would better refactor to group connection information to PS's in some sort of `connection_handler` as is done with `page_server::connection_handler`.

@wrlawodms  
This patch is only preliminary and could be merged - up to you.
But we also nee to refactor for future development either with one of the options above or with one proposal of yours.
Let me know your opinion. Thanks.

src/server/tran_server.cpp
```diff
@@ -358,21 +360,30 @@ tran_server::uses_remote_storage () const
...
     }
 
-  m_page_server_conn_vec[0]->push (reqid, std::move (payload));
+  m_page_server_conn_vec[idx]->push (reqid, std::move (payload));
+}
+
+void
+tran_server::push_request (tran_to_page_request reqid, std::string &&payload)
+{
+  // TODO push a request to the "main connection"
```
cristiarg, line 383: this is already done :)
the first one is - by convention - the main connection

in a configuration with many PS's, the PTS's connection list should be:
- PTS1: PS1, PS2 (PS1 is the main connection)
- PTS2: PS1, PS2 ( -----``----)
- PTS3: PS2, PS1 (PS2 is the main connection)
- ...

This way some form of manual-static "load balancing" is obtained. This was the idea from the beginning.

src/server/tran_server.hpp
```diff
@@ -85,6 +86,8 @@ class tran_server
 
   protected:
```
cristiarg:
```suggestion
```
similar extra lines after `private:` below

src/server/tran_server.cpp
```diff
@@ -358,21 +360,30 @@ tran_server::uses_remote_storage () const
 }
 
 void
-tran_server::push_request (tran_to_page_request reqid, std::string &&payload)
+tran_server::push_request_to (size_t idx, tran_to_page_request reqid, std::string &&payload)
 {
+  assert (idx < m_page_server_conn_vec.size());
   if (!is_page_server_connected ())
```
cristiarg: should be segregated in two:
- `are_all_page_server_connected`
- `is_page_server_connected(idx)`

and called according to function - here `is_page_server_connected(idx)`

src/server/tran_server.hpp
```diff
@@ -74,6 +74,7 @@ class tran_server
 
     void disconnect_page_server ();
     bool is_page_server_connected () const;
+    void push_request_to (size_t idx, tran_to_page_request reqid, std::string &&payload);
```
cristiarg, line 78: can have the same name, overloading does its job:
```suggestion
    void push_request (size_t idx, tran_to_page_request reqid, std::string &&payload);
```

src/server/active_tran_server.cpp
```diff
@@ -117,4 +119,8 @@ active_tran_server::get_request_handlers ()
 void
 active_tran_server::stop_outgoing_page_server_messages ()
 {
+  for (const auto &sink : m_prior_sender_sink_hooks)
```
cristiarg, line 121: hmm, this could explain some crashes I was seeing sometime when stopping ATS - some time ago;
however, I have not seen them recently

However, what is the relation between removing the sinks and the transactional log still being produced?
I think the order should be:
- first, the log daemon is stopped so no transactional log can be produced anymore
- secondly, after all prior info has been dispatched to the sinks, the sinks can be removed
Are we sure this is the case?

wrlawodms: > first, the log daemon is stopped so no transactional log can be produced anymore

The removal of sinks just means "stop generating the prior node sending requests when appending prior lists". So, it doesn't matter that the log daemon keeps running even after removing sinks.

> secondly, after all prior info has been dispatched to the sinks, the sinks can be removed

The removal sinks just removes the pointer of them from the prior_sender. The sinks are alive in the `active_tran_server`. It's safe as long as the sinks in `active_tran_server` are not deleted before handling the prior info requests which are already generated and pushed into the sending queue. And in `tran_server::disconnect_page_server`, the send queue , auto_send, and client_server are cleared in the destructor of `request_sync_client_server`. So, I think it's OK.

src/server/tran_server.hpp
wrlawodms: Done. But it makes `bind` hard :(. I've used `using`. Welcome to any suggestion.

src/server/tran_server.cpp
wrlawodms: Yes, I guess. Without it, ATS fails to start if it fails to connect to any PS. It returns not `NO_ERROR`, and in turn the caller's assertion fails.

src/server/tran_server.cpp
wrlawodms: Good point, Thank you.

By the way, there is no use of `are_all_page_server_connected`. Who have used it usually checks if any ps is connected. So, we could just overload it like this:
- is_page_server_connected()      : check if at least one is available. (used as it is)
- is_page_server_connected(idx) : check if the specific one is available.

What do you think?

src/server/tran_server.cpp
wrlawodms: Thank you for the heads-up. I'm aware of this, but I think we will introduce the main connection explicitly in the code when handling failure of PS or a newly connected PS which has more priority. For example:

TS1: _PS1_ (not connected), PS2, PS3 
The vector[0] contains PS2. When the PS1 is connected, it should become the main connection and it would be in the vector[2]. Or we can shift existing ones.

I wanted to note that the `push_request` without a given index is supposed to the main connection.

src/server/tran_server.hpp
wrlawodms: Done

src/server/tran_server.cpp
hornetmj: How about remove the local_exit_code var?

src/server/active_tran_server.hpp
wrlawodms: > More formal way:
I think we would better refactor to group connection information to PS's in some sort of connection_handler as is done with page_server::connection_handler.

1. Actually, this is the way I mentioned in the previous meeting. I am going to deal with disconnection cases in the following issue: http://jira.cubrid.org/browse/LETS-639. How about refactoring by adding connection_handler in the issue? 
2. And I'm planning to only add the connection handler for ATS, not PTS. I doubt that the PTS also requires this kind of class, but we can add later if needed.

Please share your opinion on it.

> maybe you can use unique_ptr as is done in tran_server::m_page_server_conn_vec

I considered it, but unique_ptr wouldn't work for this case, because the prior sender also points to sinks. We can use shared_ptr, but it requires to make changes to the prior_sender too, the way I see it. Anyway, If we go with the above "connection_handler" refactoring, we don't need to worry about it. Then, I'd like to go with `list` for now, and we can change it with the refactoring.

hornetmj (approved):

src/server/tran_server.cpp
wrlawodms: @hornetmj Even if `local_exit_cde` is removed, the `exit_code` is needed and cleared when the last connection fails. 

I assume that use `exit_code` instead of `local_exit_code` used for `connect_to_page_server`. Please tell me know more details of suggestion if it's not you meant.

hornetmj: @wrlawodms I think the local varable local_exit_code is useless. So, I gave you a suggestion.

wrlawodms: @hornetmj Oh, I got it finally. :) and it's done.

src/server/active_tran_server.cpp
cristiarg: > The removal of sinks just means "stop generating the prior node sending requests when appending prior lists". So, it doesn't matter that the log daemon keeps running even after removing sinks.

I think it does matter in the case of graceful stop. It means that log is being legitimately produced but not sent to the Page Servers.

> The sinks are alive in the active_tran_server. It's safe as long as the sinks in active_tran_server are not deleted before handling the prior info requests which are already generated and pushed into the sending queue.

The sending of the prior log is:
- partially synchronous: in the application layer
- partially asynchronous: in the connection layer (or, better put at the edge between the application layer and the server-server connection layer) - the async part starts in `request_sync_client_server<T_OUTGOING_MSG_ID, T_INCOMING_MSG_ID, T_PAYLOAD>::push`
So, you are partially correct.
If the sink is unregistered the link between the synchronous application layer and the asynchronous server-server connection layer is broken and the produced prior log is lost.

However, all is good it seems, please see the extended comment `shutdown order for Active Transaction Server` in `xboot_shutdown_server`.
Can you please extend the comment such as it reads:
```
      // shutdown order for Active Transaction Server:
      //  - finalize log infrastructure
      //    - done in: shutdown server routine (this routine)
      //  - the log prior messages will stop being dispatched to the connected Page Server(s)
      //    by virtue of the client transactions and other daemons being stopped
      //    and, thus, no log will be produced anymore
      //    - log prior messages are being dispatched from Active Transaction Server towards
      //      Page Server(s) semi-synchronously via the prior_sender::send_list routine
      //  - send the final disconnect message to all connected Page Server(s)
      //    - done in:
      //        shutdown server routine (this routine)
      //          -> boot_server_all_finalize
      //            -> finalize_server_type
      //              -> tran_server::disconnect_page_server
      //  - unregister log prior sinks (used to dispatch produced transactional log to connected
      //    Page Servers:
      //    - done in:
      //        shutdown server routine (this routine)
      //          -> boot_server_all_finalize
      //            -> finalize_server_type
      //              -> tran_server::disconnect_page_server
      //                -> active_tran_server::stop_outgoing_page_server_messages
      //  - delete the Active Transaction Server object
      //    - done in:
      //        shutdown server routine (this routine)
      //          -> boot_server_all_finalize
      //            -> finalize_server_type
```
(either in this PR or in a different patch, your choice)

src/server/active_tran_server.hpp
cristiarg: > How about refactoring by adding connection_handler in the issue?

As I said, it can be done separately. This patch is self-contained as it is.

> And I'm planning to only add the connection handler for ATS, not PTS. I doubt that the PTS also requires this kind of class, but we can add later if needed.

Actually you are right.
I was concerned about the PTS fail-over to ATS in the future.
But for the fail-ove, a PTS will have to swap the `passive_tran_server` with the `active_tran_server` structure as a whole.

cristiarg: > I considered it, but unique_ptr wouldn't work for this case, because the prior sender also points to sinks. We can use shared_ptr, but it requires to make changes to the prior_sender too, the way I see it. Anyway, If we go with the above "connection_handler" refactoring, we don't need to worry about it. Then, I'd like to go with list for now, and we can change it with the refactoring.

`unique_ptr` does work :)
The `active_tran_server` instance is the owner. `prior_sender::add_sink` takes as argument a reference (just for the formalism of ensuring that there is a valid instance) which is then referentiated as a pointer. And `prior_sender::add_sink` has non-owning pointers.

wrlawodms: > unique_ptr does work :)

Aha, OK. The prior_sender doesn't need to own this. I got it wrong. Thank you.

src/server/active_tran_server.cpp
wrlawodms: > I think it does matter in the case of graceful stop. It means that log is being legitimately produced but not sent to the Page Servers.

Hmm.. I still doubt it's a problem. Yes, some log records produced legitimately can be lost. However, 
- If the majority of PSes don't receive it, they can be ignored because it's never been committed.
- If the majority of PSes receive it, it will be caught up.

And this is the same to the situation in which a disconnected PS can't receive some legitimately produced log records. I think it's OK in terms of correctness as long as there is no hole among log records. Is it necessary to be careful to stop and wait for all generated logs to be transmitted and clean up resources?

> If the sink is unregistered the link between the synchronous application layer and the asynchronous server-server connection layer is broken and the produced prior log is lost.

I'm not sure I understand you correctly, but this is also saying the above, right?

> Can you please extend the comment such as it reads:

Thanks for your efforts! I'll change this in the next issue as I believe the details might change during the refactoring process with the use of the connection_handler.

cristiarg: > Is it necessary to be careful to stop and wait for all generated logs to be transmitted and clean up resources?

Now that you mention it, I see there is no waiting for the termination in the server-server connection layer.
The server-server communication layer is abstracted via the `request_sync_send_queue` and `request_queue_autosend` classes and there is no explicit wait in the dtor of those class for the buffers to be consumed. (I have added http://jira.cubrid.org/browse/LETS-643 to not lose track of this).

As a guide, the closest to a clean stop is done in `page_server::connection_handler::~connection_handler` where:
- incoming communication is stopped
- the already received data is being processed
- the outgoing communication is stopped

> If the sink is unregistered the link between the synchronous application layer and the asynchronous server-server connection layer is broken and the produced prior log is lost.

Take a look at the following call path:
- `logpb_append_prior_lsa_list`
- `prior_sender::send_list`: the prior lists are sent (and copied, unfortunately - this can be improved) - to all sinks (ie: on ATS, to all PS's); still synchronous
  - the sinks in the list are packaged bound calls to `active_tran_server::push_request` (as registered in `active_tran_server::on_boot`); so, still synchronous
- `tran_server::push_request` calls into `request_sync_client_server::push`; still synchronous
- `request_sync_client_server::push` calls into `request_sync_send_queue::push`; and it is only here that the synchronous call chain is broken as the message is pushed on the queue and the sending thread is notified there is new data to dispatch

What I was explaining was a response to your comment:

> The removal sinks just removes the pointer of them from the prior_sender. The sinks are alive in the active_tran_server. It's safe as long as the sinks in active_tran_server are not deleted before handling the prior info requests which are already generated and pushed into the sending queue.

To rephrase: the sinks hold no state; they are a bridge between the application layer and the server-server communication layer (sending thread) and thus, removing them from the log prior sender only breaks this link. Thus, if log would still be produced (it is not, as shown by the exended comment in `xboot_shutdown_server`) the only effect would be that the `prior_sender::send_list` would not send anything. And now I see that the `prior_sender::send_list` implementation is wrong as it reports sending the log although it might send nothing. So, it would be best to add an experimental assert in `prior_sender::send_list` in case log is not being sent at all and see what happens. (I have added http://jira.cubrid.org/browse/LETS-642 for this.)

src/server/tran_server.hpp
cristiarg: @wrlawodms 
I did not anticipate this. My bad.
And the new implementation of `active_tran_server::on_boot` is hard to parse/understand.
Would it be too much to as to revert is as it was initially?
I did not anticipate it would have this ugly repercussion.

src/server/tran_server.cpp
cristiarg: >     is_page_server_connected() : check if at least one is available. (used as it is)
>     is_page_server_connected(idx) : check if the specific one is available.

Agreed.

cristiarg:
@wrlawodms 
I think only the comments:
- https://github.com/CUBRID/cubrid/pull/4082#discussion_r1094351135 (at your choice)
- https://github.com/CUBRID/cubrid/pull/4082#discussion_r1094353913

need to be addressed.
The rest is just "productive discussion".

I have added http://jira.cubrid.org/browse/LETS-642 and http://jira.cubrid.org/browse/LETS-643 as to follow-up on minor issues.

Thanks.

cristiarg (approved):

joohok (approved):

src/server/tran_server.hpp
wrlawodms: done!

src/server/active_tran_server.cpp
wrlawodms: > Now that you mention it, I see there is no waiting for the termination in the server-server connection layer.

OK, now I can see (I guess). 
My opinion was that it's ok in terms of correctness that some legitimately generated log records are lost. However, in normal cases, we don't need to risk losing. We should stop communication gracefully to make sure that everything is under control. 

> As a guide, the closest to a clean stop is done in page_server::connection_handler::~connection_handler where:

Thank you for the guide. Only one missing seems `m_response_broker.terminate()`. Others are done in each dtor. Let's see it in the following issue.

> So, it would be best to add an experimental assert in prior_sender::send_list in case log is not being sent at all and see what happens. (I have added http://jira.cubrid.org/browse/LETS-642 for this.)

I think it's not problematic in terms of correctness to allow the loss of some log records by removal sinks before sending them as long as there is no hole between received log records. But, regardless of the correctness, I assume that you want to make sure that all log-records-sending-requests have been digested properly before removing sinks. Do I understand you correctly? or do we have misunderstandings? I think it is worth your intention since it's more robust and error-prone. I'll consider it in the following issue. Please let me know if there is some misunderstanding.

PR #4087, opened 2023-02-01
[LETS-425] overloaded dtor for server request responder task; adapt dependent code

cristiarg:

This patch changes nothing functionally, it just contains formalisms:
- add an overloaded dtor for `class server_request_responder::task`
- adapt dependent code:
  - `request_sync_client_server::sequenced_payload::operator= (std::nullptr_t)`
  - unit test

### DIFF
--- a/src/communication/request_sync_client_server.hpp
+++ b/src/communication/request_sync_client_server.hpp
@@ -116,6 +116,8 @@ namespace cubcomm
       sequenced_payload &operator= (sequenced_payload &&other);
       sequenced_payload &operator= (const sequenced_payload &) = delete;
 
+      sequenced_payload &operator= (std::nullptr_t);
+
       void push_payload (T_PAYLOAD &&a_payload);
       T_PAYLOAD pull_payload ();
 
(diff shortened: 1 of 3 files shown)


src/communication/request_sync_client_server.hpp
```diff
@@ -116,6 +116,8 @@ namespace cubcomm
       sequenced_payload &operator= (sequenced_payload &&other);
       sequenced_payload &operator= (const sequenced_payload &) = delete;
 
+      sequenced_payload &operator= (std::nullptr_t);
```
wrlawodms, line 119: Oh.. I didn't know that we can make use of `std::nullptr_t` to reset a variable in this way. 

The following question is just to learn it. Thank you in advance:
- The way I see it, it could be confusing in that std::nullptr_t is assigned to a non-pointer variable. (seemingly assign a pointer value to a non-pointer variable?). It would be ok if it's used widely. Is it used widely?

wrlawodms (approved):

hornetmj (approved):

joohok (approved):

src/communication/request_sync_client_server.hpp
cristiarg: I drew inspiration from `unique_pts`'s:
```
unique_ptr& operator=(std::nullptr_t) ) noexcept;
```
I do agree it is a bit forced but it shows the intention to actually "obliterate" / "invalidate" an object instance. Akin to a "reset" function.

PR #4088, opened 2023-02-02
[CBRD-24622] Create 10 SCHEMA Files for each object type.

airnet73:
http://jira.cubrid.org/browse/CBRD-24623

Purpose
To move data by object without editing schema file, when creating schema file, it is necessary to create by object as follows.
You can create 10 schema files by using options.

* Create a schema file for each object.

> testdb_schema_user
> testdb_schema_serial
> testdb_schema_synonym
> testdb_schema_class
> testdb_schema_vclass
> testdb_schema_procedure
> testdb_schema_fk
> testdb_schema_pk
> testdb_schema_grant
> testdb_schema_server

* Create schema_info file and schema file order
> testdb_schema_user
> testdb_schema_class
> testdb_schema_vclass
> testdb_schema_serial
> testdb_schema_procedure
> testdb_schema_server
> testdb_schema_pk
> testdb_schema_fk
> testdb_schema_grant
> testdb_schema_synonym

Implementation
N/A

Remarks
N/A


kisoo-han:
@airnet73,
please add 'msg/en_US/utils.msg' for review.

kisoo-han:
You used the term '**each_schema_file**', 
for me It's not intuitive, I propose terms, **seperate_schema_files** or **split_schema_files** instead.
Selection is your choice.

airnet73:
> @airnet73, please add 'msg/en_US/utils.msg' for review.

I have plans to add other languages when the Korean review is complete.
Is there anything in ko_KR.utf8/utils.msg that needs to be modified?

src/executables/unload_schema.c
```diff
@@ -180,9 +199,27 @@ static void emit_partition_info (print_output & output_ctx, MOP clsobj);
 static int emit_stored_procedure_args (print_output & output_ctx, int arg_cnt, DB_SET * arg_set);
 static int emit_stored_procedure (print_output & output_ctx);
 static int emit_foreign_key (print_output & output_ctx, DB_OBJLIST * classes);
+static int emit_grant (extract_context & ctxt, print_output & output_ctx, DB_OBJLIST * classes);
 static int create_filename (const char *output_dirname, const char *output_prefix, const char *suffix,
 			    char *output_filename_p, const size_t filename_size);
+static int create_filename (const char *output_dirname, const char *output_prefix, const char *middle,
+			    const char *suffix, char *output_filename_p, const size_t filename_size);
 static int export_server (print_output & output_ctx);
```
kisoo-han, line 210: do you have any reason that use of '**export_server**' instead '**emit_server**'

airnet73: No reason. I just used the existing function name as it is.

src/executables/unload_schema.c
```diff
@@ -3934,3 +4045,691 @@ create_filename (const char *output_dirname, const char *output_prefix, const ch
...
+      return -1;
+    }
+
+  snprintf (output_filename_p, filename_size - 1, "%s/%s%s%s", output_dirname, output_prefix, middle, suffix);
+
+  return 0;
+}
+
+static int
+extract_only_user (extract_context & ctxt)
```
kisoo-han: You used function name like 'extract_only_user', '
extract_only_serial,
extract_only_class,
extract_only_synonym,
extract_only_procedure,
extract_only_server,
extract_only_vclass,
extract_only_pk,
extract_only_fk,
extract_only_grant

I'll take one example with **extract_only_grant**, following is preferred priority for me.
1. if it is applicable '**emit_grant**' use it
2. if 1 is not applicable then,  applicable '**extract_grant**'
3. nothing applicable, '**extract_grant_only**'

kisoo-han:
Why I asked for en_US/utils.msg, I cannot see help message from binary of this PR 
(when I just hit 'cubrid unloaddb', I cannot see help for that.)

src/executables/unload_schema.c
```diff
@@ -180,9 +199,27 @@ static void emit_partition_info (print_output & output_ctx, MOP clsobj);
...
 static int emit_stored_procedure (print_output & output_ctx);
 static int emit_foreign_key (print_output & output_ctx, DB_OBJLIST * classes);
+static int emit_grant (extract_context & ctxt, print_output & output_ctx, DB_OBJLIST * classes);
 static int create_filename (const char *output_dirname, const char *output_prefix, const char *suffix,
 			    char *output_filename_p, const size_t filename_size);
+static int create_filename (const char *output_dirname, const char *output_prefix, const char *middle,
+			    const char *suffix, char *output_filename_p, const size_t filename_size);
 static int export_server (print_output & output_ctx);
+static int get_classes (extract_context & ctxt, print_output & output_ctx);
+static int extract_whole_schema_file (extract_context & ctxt, const char *output_filename);
```
kisoo-han: what about, 
**extract_all_schema_file ()** as the function name

src/executables/unload_schema.c
airnet73: Thanks for the advice.
I will refer to the above.

src/executables/unload_schema.c
airnet73: I'll fix it. thank you.

kisoo-han (approved):
Please check for CI fails.

src/executables/unload_schema.c
```diff
@@ -3934,3 +4045,691 @@ create_filename (const char *output_dirname, const char *output_prefix, const ch
 
   return 0;
 }
+
+static int
+create_filename (const char *output_dirname, const char *output_prefix, const char *middle, const char *suffix,
```
eido5: As the parameter, I understand 'const char *middle' takes 3 types name;  SCHEMA_NAME, TRIGGER_NAME, INDEX_NAME.

I am a little bit confused because what purpose of 'middle'. 

Would you change the parameter name according to the purpose?
Or Would you add the comment above the create_filename function?

Do we need to make this parameter clear?
I suggest parameter name to be more clear because I think this function is created by this 'middle' parameter. 

I can suggest below;
1. file_type
2. expression
3. infix

If my suggestion is not adequate what you design, could you explain about the design?

airnet73: This function is just a function to generate a file by receiving three inputs, and it is not a function with special significance.
So, create_filename is designed to input three values to create a file in the format of [dbname]_[schema_name]_[object_name].
Since the format of the file can be changed at any time, the parameter was written in an ambiguous way. So the function name is also "create_filename" with no special meaning.

eido5: So, in your opinion,is parameter name of 'middle' clear to understand?

airnet73: Thank you for your comments.
I confirmed that the meaning of the word was not used correctly.
I would change it to "infix" instead of "middle".

eido5: @airnet73 
Oh, I understand. 
It seems look so good.

src/executables/unload_schema.c
```diff
@@ -3934,3 +4045,691 @@ create_filename (const char *output_dirname, const char *output_prefix, const ch
...
+    {
+      fclose (output_file);
+      output_file = NULL;
+    }
+
+  return err;
+}
+
+static int
+extract_server (extract_context & ctxt)
```
eido5, line 4421: Could you explain about the **extract_server**?

Or could you give me the link which includes the explanation **extract_server**? because I searched that I couldn't found about that.

eido5 (approved):

src/executables/unload_schema.c
airnet73: Please refer to the link below
https://www.cubrid.org/manual/en/11.2/sql/schema/server_stmt.html

Srltas (approved):

src/executables/unload_schema.c
eido5: Thank you!

kisoo-han (approved):
Well DONE!

PR #4092, opened 2023-02-03
[CBRD-24630] The stop command of cubrid hb (applylogdb/repl) does not work correctly when the build is generated in the QA environment, release mode and devtoolset-9

hornetmj:
http://jira.cubrid.org/browse/CBRD-24630

Purpose
- when the build is created in the QA env, release mode and devtoolset-9, the stop command of cubrid hb (applylogdb or replication) doesn't work correctly
- the reason is not clear. But the problem is related to the strcmp function. The following code block is in the process_is_registered_proc(). And the if condition checks the response from cub_master in order to check the copylogdb or applylogdb is alive
`if (size > 0 && strcmp (reply_buffer, HA_REQUEST_SUCCESS) == 0)` 
- the problem is that the strcmp() returns a positive value even if values of arguments are equal
  - when I try to dump values of arguments as hexa, they have same values with null termination
  - when I try to compare them using memcmp(), strcmp()(again), they return 0
  - when I try to check the length of them using strlen(), they return 1 (it means values are terminated with null)
- now, I don't have any options to handle this problem. But when I change the strcmp() to strncmp(), the problem is disapeared. So I apply this fix for the regression test temparaly(or permenantly if there are no solution)

Implementation
- strcmp() -> strncmp()

Remarks
- the problem is occurred after changing devtoolset-8 to devtoolset-9

### DIFF
--- a/src/executables/commdb.c
+++ b/src/executables/commdb.c
@@ -704,7 +704,7 @@ process_is_registered_proc (CSS_CONN_ENTRY * conn, char *args)
   rid = send_request_one_arg (conn, IS_REGISTERED_HA_PROC, (char *) buffer, len);
   return_string (conn, rid, &reply_buffer, &size);
 
-  if (size > 0 && strcmp (reply_buffer, HA_REQUEST_SUCCESS) == 0)
+  if (size > 0 && strncmp (reply_buffer, HA_REQUEST_SUCCESS, size - 1) == 0)
     {
       error = NO_ERROR;
     }


src/executables/commdb.c
```diff
@@ -704,7 +704,7 @@ process_is_registered_proc (CSS_CONN_ENTRY * conn, char *args)
   rid = send_request_one_arg (conn, IS_REGISTERED_HA_PROC, (char *) buffer, len);
   return_string (conn, rid, &reply_buffer, &size);
 
-  if (size > 0 && strcmp (reply_buffer, HA_REQUEST_SUCCESS) == 0)
+  if (size > 0 && strncmp (reply_buffer, HA_REQUEST_SUCCESS, size - 1) == 0)
```
wrlawodms, line 707: The reason to use "size-1" instead of "size" is to compare until the '\0'? I guess we don't need to do it and "size" looks more readable.

wrlawodms (approved):

joohok (approved):
I can't find the alternative solution for this issue.
I have one question, 
what I know is that all the requests from commdb to cub_master uses this mechanism, such as `process_ha_deregister_by_pid (), process_deactivate_heartbeat() , ..`, do these functions work well? or does it need more regression test for HA ?

mhoh3963 (approved):

src/executables/commdb.c
hornetmj: It is a right reason what you said. Thanks for your feedback but I gonna not to change this code because someone can easily understand the size includes the length of the null termination.

hornetmj:
> I can't find the alternative solution for this issue. I have one question, what I know is that all the requests from commdb to cub_master uses this mechanism, such as `process_ha_deregister_by_pid (), process_deactivate_heartbeat() , ..`, do these functions work well? or does it need more regression test for HA ?

I doesn't check each function-base(using debugger or logging). Instead I do test each command base like cubrid hb start/stop, cubrid hb copylogdb start/stop, cubrid hb applylogdb start/stop and cubrid hb replication start/stop. They works well and I think it is enough.

If you feel it is not enough and more white box test should be executed, then let me know.

joohok:
> > I can't find the alternative solution for this issue. I have one question, what I know is that all the requests from commdb to cub_master uses this mechanism, such as `process_ha_deregister_by_pid (), process_deactivate_heartbeat() , ..`, do these functions work well? or does it need more regression test for HA ?
> 
> I doesn't check each function-base(using debugger or logging). Instead I do test each command base like cubrid hb start/stop, cubrid hb copylogdb start/stop, cubrid hb applylogdb start/stop and cubrid hb replication start/stop. They works well and I think it is enough.
> 
> If you feel it is not enough and more white box test should be executed, then let me know.

Thanks for reply.

It was just curiosity while trying to find the cause of this problem. 
I was curious to see if the same problem occurred with other functions using the same mechanism. 

> cubrid hb start/stop, cubrid hb copylogdb start/stop, cubrid hb applylogdb start/stop and cubrid hb replication start/stop 
 
But, I agreed with you I think above cases are enough.

hornetmj:
@mhoh3963 Please merge this PR forcibly. The ci failures is not related to this PR.

PR #4093, opened 2023-02-03
[LETS-638] implement compress/decompress of heap pages dispatched by page server

cristiarg:
http://jira.cubrid.org/browse/LETS-638

Implement compress/decompress of heap pages dispatched by page server:
  - the existing `cubthread::entry::log_zip_undo` structure is used
  - it is considered safe to reuse this structures for scalability data pages compression/decompression
  - however, there are combined use-cases are as follows:
    - on the Active Transaction Server, a client transaction both requests data pages (thus having to decompress them) and also produces transactional log (thus, having to compress it) but this happens in distinct moments in time
    - on an Active Transaction Server, a client transaction might need to perform a rollback, for which log records must be processed (uncompressed) to restore previous values of the data
    - on any Transaction Server, a transaction might need to access old values of the data (MVCC) which, in the current architecture, means reading (uncompressing) log pages
  - it must be evaluated whether it is safe to re-use structures for the new purpose, or, new structures must be used (eg: by a localized object pool - this option has also been explored and proved to work)
- add system parameter to tune compression on and off
- in case decompression fails, this is considered catastrophic; set fatal error and crash;

The established "protocol" allows for 3 scenarios:
- either send the pages uncompressed, as before - sysparam `replication_compress_pages_transfer` is off
- send the the pages compressed, the new default - sysparam `replication_compress_pages_transfer` is on
- in case the underlying LS4-based compression logic fails - it happens when the output is larger than the input - also the uncompressed raw data page will be sent

NOTE (to reviewers):
- to be evaluated whether the existing structures are can be reused or new, specifically added structures are needed

- only data pages are compressed
- the patch does not contain a measurement of the average compression ratio (only logging); a new perf measurement for average compression ratio will be added in a next patch
- log pages are not compressed (their contents is supposed to be compressed by the existing compression of the transactional log)
- question: should the log pages also be compressed in the same manner as data pages are?


src/storage/page_buffer.c
```diff
@@ -8489,22 +8517,121 @@ pgbuf_respond_data_fetch_page_request (THREAD_ENTRY &thread_r, std::string &payl
...
+      // it is arbitrary which of the thread entry compress structures is used in this context
+      // the thread on which this routine is to be executed is part of a thread
+      // pool which is only used to dispatch responses from the page server to the transaction servers
+      if (thread_r.log_zip_undo == nullptr)
+	{
+	  thread_r.log_zip_undo = log_zip_alloc (IO_PAGESIZE);
+	}
+
+      const bool compressed = log_zip (thread_r.log_zip_undo, data_in_size, data_in);
+      assert (compressed);
```
razvanradulescu: Please see inside of log_zip:
    compressed = false; 
...
      if (log_zip->data_length < length)
	{
	  compressed = true;
	} 
If zipped size is not less	original, it is considered failed.

razvanradulescu (approved):

wrlawodms (approved):

src/storage/page_buffer.c
wrlawodms: What if there is no room to compress sine it's too compact? I mean if it's try to compress and there is no exception, but the compression ratio is 1 (same as original size), this will cause the assertion failure. I think this is a legal case. I'm not sure it's the case, though.

It might be the same as razvan's feedback.

cristiarg: I will adapt the "protocol" to send the uncompressed page if that is the case:
- if compression fails (resulting length same or larger), send the uncompressed page
- on the decompress side (*TS), if the received size is the same as `IO_PAGESIZE`, it means there was no compression

@wrlawodms 
Lz4 can produce outputs larger than input. The check inside `log_zip` is in this direction, if the output is larger, to let the client code know and act accordingly.

Of course there is also the very edge case where the length of the initial data (which is also saved together with the compressed data) + the length of the compressed data is still larger than the initial data's length. In which case compressing/decompressing still makes no sense.
But I think we can ignore this corner case.

joohok (approved):
Please check the unit test failure (test_server_request_responder)
I think it is due to #4087, but not sure.

hornetmj (approved):

cristiarg:
@joohok 
> Please check the unit test failure (test_server_request_responder) I think it is due to #4087, but not sure.

Does not crash in debug mode.
It crashes in release mode.
I will investigate. Thanks for the heads-up.

src/storage/page_buffer.c
cristiarg: Adapted the "protocol". Updated PR description accordingly.

> Of course there is also the very edge case where the length of the initial data (which is also saved together with the compressed data) + the length of the compressed data is still larger than the initial data's length. 

This concern is already accounted for in the implementation of `log_zip` function.

src/storage/page_buffer.c
```diff
@@ -8412,14 +8421,39 @@ pgbuf_request_data_page_from_page_server (const VPID * vpid, log_lsa target_repl
...
+      // all are handled in the decompress function
+      const bool decompressed = pgbuf_decompress_page (thread_r, reinterpret_cast < const void *>(message_buf),
+						       (int) data_length, decompressed_data, decompressed_data_size);
+      assert (decompressed);
+      message_buf += data_length;
+
+      if (!decompressed)
+	{
+	  // TODO: error
+	  assert_release (false);
```
razvanradulescu, line 8447: I think it should early exit ?!
Or do we ignore in release ?

src/storage/page_buffer.c
```diff
@@ -8412,14 +8421,39 @@ pgbuf_request_data_page_from_page_server (const VPID * vpid, log_lsa target_repl
...
+      int decompressed_data_size = 0;
+
+      // decompress data that is sent over; there are 3 cases:
+      //  - compression setting on and data compressed successfully
+      //  - compression setting on but data failled to be compressed
+      //  - compression off
+      // all are handled in the decompress function
+      const bool decompressed = pgbuf_decompress_page (thread_r, reinterpret_cast < const void *>(message_buf),
+						       (int) data_length, decompressed_data, decompressed_data_size);
+      assert (decompressed);
```
razvanradulescu: This is redundant; assert_release (false) bellow can handle the case.

razvanradulescu (approved):

src/storage/page_buffer.c
cristiarg: If it fails, it's catastrophic. Examples:
- if source of request is a client transaction (ATS, PTS): abort transaction
- if source of request is vacuum thread (ATS): it could retry request
- if source of request is replication thread (PTS): it could retry request
- ...

For now I think we should just crash.
I will add an error as `TODO`-ed.

cristiarg: Added error message

PR #4096, opened 2023-02-08
[CBRD-24654] When getting a NUMERIC type value, if the value after the decimal point is 0, it should be improved so that the trailing 0's are not passed.

beyondykk9:
http://jira.cubrid.org/browse/CBRD-24654

If the column type is specified with a digit after the decimal point, such as numeric(38,10), different results are returned than in the case of ORACLE/TIBERO, which causes inconvenience in application modification when not familiar with CUBRID or when converting.

### DIFF
--- a/src/base/system_parameter.c
+++ b/src/base/system_parameter.c
@@ -712,6 +712,8 @@ static const char sysprm_ha_conf_file_name[] = "cubrid_ha.conf";
 
 #define PRM_NAME_REGEXP_ENGINE "regexp_engine"
 
+#define PRM_NAME_ORACLE_STYLE_NUMBER_RETURN "oracle_style_number_return"
+
 #define PRM_VALUE_DEFAULT "DEFAULT"
 #define PRM_VALUE_MAX "MAX"
 #define PRM_VALUE_MIN "MIN"
@@ -2339,6 +2341,10 @@ static int prm_regexp_engine_lower = cubregex::engine_type::LIB_CPPSTD;
 static int prm_regexp_engine_upper = cubregex::engine_type::LIB_RE2;
 static unsigned int prm_regexp_engine_flag = 0;
 /* *INDENT-ON* */
+
+bool PRM_ORACLE_STYLE_NUMBER_RETURN = false;
+static bool prm_oracle_style_number_return_default = false;
+static unsigned int prm_oracle_style_number_return_flag = 0;
 
 typedef int (*DUP_PRM_FUNC) (void *, SYSPRM_DATATYPE, void *, SYSPRM_DATATYPE);
 
@@ -6155,6 +6161,17 @@ SYSPRM_PARAM prm_Def[] = {
    (void *) &PRM_REGEXP_ENGINE,
    (void *) &prm_regexp_engine_upper,
    (void *) &prm_regexp_engine_lower,
+   (char *) NULL,
+   (DUP_PRM_FUNC) NULL,
+   (DUP_PRM_FUNC) NULL},
+  {PRM_ID_ORACLE_STYLE_NUMBER_RETURN,
+   PRM_NAME_ORACLE_STYLE_NUMBER_RETURN,
+   (PRM_FOR_SERVER | PRM_FORCE_SERVER),
+   PRM_BOOLEAN,
+   &prm_oracle_style_number_return_flag,
+   (void *) &prm_oracle_style_number_return_default,
+   (void *) &PRM_ORACLE_STYLE_NUMBER_RETURN,
+   (void *) NULL, (void *) NULL,
    (char *) NULL,
    (DUP_PRM_FUNC) NULL,
    (DUP_PRM_FUNC) NULL},
--- a/src/query/numeric_opfunc.c
+++ b/src/query/numeric_opfunc.c
@@ -3852,6 +3852,8 @@ numeric_db_value_print (const DB_VALUE * val, char *buf)
   bool found_first_non_zero = false;
   int scale = db_value_scale (val);
 
+  static bool oracle_style_number = prm_get_bool_value (PRM_ID_ORACLE_STYLE_NUMBER_RETURN);
+
   assert (val != NULL && buf != NULL);
 
   if (DB_IS_NULL (val))
@@ -3871,8 +3873,7 @@ numeric_db_value_print (const DB_VALUE * val, char *buf)
       /* Add the negative sign */
       if (temp[i] == '-')
 	{
-	  buf[nbuf] = '-';
-	  nbuf++;
+	  buf[nbuf++] = '-';
 	}
 
       /* Add decimal point */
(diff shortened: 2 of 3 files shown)


hgryoo:
It is written in the JIRA issue that ORALCE_STYPE_NUMBER_RETURN is set through cubrid_broker.conf, but PR implements it as a server parameter. Which one is right?

hgryoo:
If it is set as a server parameter, restarting the cubrid server is required to change the parameter. Is It the intended behavior?

beyondykk9:
> If it is set as a server parameter, restarting the cubrid server is required to change the parameter. Is It the intended behavior?

Yes, it is intended.

beyondykk9:
> It is written in the JIRA issue that ORALCE_STYPE_NUMBER_RETURN is set through cubrid_broker.conf, but PR implements it as a server parameter. Which one is right?

It's a server parameter, th JIRA issue is not a confirmation.

hgryoo:
> > It is written in the JIRA issue that ORALCE_STYPE_NUMBER_RETURN is set through cubrid_broker.conf, but PR implements it as a server parameter. Which one is right?
> 
> It's a server parameter, th JIRA issue is not a confirmation.

Dev process:
- Specification should be written correctly in the JIRA issue.
- The JIRA's status should be changed to "REVIEW IN PROGRESS" before requesting code review.

hgryoo:
> > It is written in the JIRA issue that ORALCE_STYPE_NUMBER_RETURN is set through cubrid_broker.conf, but PR implements it as a server parameter. Which one is right?
> 
> It's a server parameter, th JIRA issue is not a confirmation.

If it is supported as a server parameter, compatibility for applications implemented according to our NUMERIC output is broken.

youngjinj:
Perhaps this system parameter is used when a developer does not want to change any code
after migrating an application developed for oracle.

But I am worried that it will cause another bug like oracle_style_empty_string
and become an unused parameter.

NUMERIC and DECIMAL of SQL Server, PostgreSQL, MySQL and MariaDB are the same as ours.
And use the FORMAT() or TRIM_SCALE() function to remove the trailing zeros after the decimal point.

I think we should solve it by adding or extending functions.

**CUBRID**
```
create table t1 (c1 numeric (5,3));
insert into t1 values (5), (5.5), (5.050), (5.005); 

select
    c1 as original,

    /* intended but too long. */
    trim (trailing '.' from (trim (trailing '0' from c1))) as use_trim,

    /* unintended result. (1 or 2 or 3 ?) */
    format (c1, 2) as use_format,

    /* unintended result. */
    to_char (c1, '9.999') as use_to_char
from
    t1;

  original              use_trim              use_format            use_to_char
========================================================================================
  5.000                 '5'                   '5.00'                '5.000'
  5.500                 '5.5'                 '5.50'                '5.500'
  5.050                 '5.05'                '5.05'                '5.050'
  5.005                 '5.005'               '5.01'                '5.005'
```

**SQL Server - FORMAT()** 
```
select c1, format (c1, '0.###') from t1;
```
https://dbfiddle.uk/EMHgbpMV

**PostgreSQL - TRIM_SCALE()**
```
select c1, trim_scale (c1) from t1;
```
https://dbfiddle.uk/01NnilHS

hyunikn (approved):

hyunikn (approved):

src/base/system_parameter.c
```diff
@@ -2340,6 +2342,10 @@ static int prm_regexp_engine_upper = cubregex::engine_type::LIB_RE2;
 static unsigned int prm_regexp_engine_flag = 0;
 /* *INDENT-ON* */
 
+bool PRM_ORACLE_STYLE_NUMBER_RETURN = false;
+static bool prm_oracle_style_number_return_default = false;
```
shparkcubrid, line 2346: I recommend to change the default to true and perform a SQL test at once.
By doing that, I think unexpected problems may be discovered.
I worry that the changes are not tested at all in SQL test.

beyondykk9: Yes, I have already tested and I got the expected results which are just modification by printing out without trailing zeros.

shparkcubrid (approved):

src/query/numeric_opfunc.c
```diff
@@ -3871,15 +3873,36 @@ numeric_db_value_print (const DB_VALUE * val, char *buf)
...
 
       /* Add decimal point */
       if (i == temp_size - scale)
 	{
-	  buf[nbuf] = '.';
-	  nbuf++;
+	  int k = temp_size - 1;
+
+	  if (oracle_style_number)
+	    {
```
youngjinj, line 3885: It would be better to change it like below.
```suggestion
	  if (oracle_style_number)
	    {
	      int k = temp_size - 1;
```

src/query/numeric_opfunc.c
```diff
@@ -3871,15 +3873,36 @@ numeric_db_value_print (const DB_VALUE * val, char *buf)
...
-	  nbuf++;
+	  buf[nbuf++] = '-';
 	}
 
       /* Add decimal point */
       if (i == temp_size - scale)
 	{
-	  buf[nbuf] = '.';
-	  nbuf++;
+	  int k = temp_size - 1;
```
youngjinj, line 3882: This is an out of scope question.

Why is the size 80?
```
char temp[80];
```

I know this is because NUMERIC_MAX_STRING_SIZE is (80 + 1).
However, as a result of the test below, I don't know why 0 to 38 are needed.

Even when 38, the maximum value of scale, is specified, 38 repetitions must be passed meaninglessly.
If the scale is 5, 71 times are repeated meaninglessly.

e.g.
```
create table t1 (c1 numeric (38,0));
insert into t1 values (12345678901234567890123456789012345678.0); // Fail
insert into t1 values (12345678901234567890123456789012345678); // Success

select * from t1;

( '0', '0', '0', '0', '0', '0',    '0',    '0',    '0',    '0'   // 0~9
  '0', '0', '0', '0', '0', '0',    '0',    '0',    '0',    '0'   // 10~19
  '0', '0', '0', '0', '0', '0',    '0',    '0',    '0',    '0'   // 20~29
  '0', '0', '0', '0', '0', '0',    '0',    '0',    '1',    '2'   // 30~39
  '3', '4', '5', '6', '7', '8',    '9',    '0',    '1',    '2'   // 40~49
  '3', '4', '5', '6', '7', '8',    '9',    '0',    '1',    '2'   // 50~59
  '3', '4', '5', '6', '7', '8',    '9',    '0',    '1',    '2'   // 60~69
  '3', '4', '5', '6', '7', '8', '\000', '\000', '\000', '\000' ) // 70~79

create table t2 (c1 numeric (38,38));
insert into t2 values (0.12345678901234567890123456789012345678);

select * from t2;

( '0', '0', '0', '0', '0', '0',    '0',    '0',    '0',    '0'   // 0~9
  '0', '0', '0', '0', '0', '0',    '0',    '0',    '0',    '0'   // 10~19
  '0', '0', '0', '0', '0', '0',    '0',    '0',    '0',    '0'   // 20~29
  '0', '0', '0', '0', '0', '0',    '0',    '0',    '1',    '2'   // 30~39
  '3', '4', '5', '6', '7', '8',    '9',    '0',    '1',    '2'   // 40~49
  '3', '4', '5', '6', '7', '8',    '9',    '0',    '1',    '2'   // 50~59
  '3', '4', '5', '6', '7', '8',    '9',    '0',    '1',    '2'   // 60~69
  '3', '4', '5', '6', '7', '8', '\000', '\000', '\000', '\000' ) // 70~79
  
create table t3 (c1 numeric (38,5));
insert into t3 values (123456789012345678901234567890123.98765);

( '0', '0', '0', '0', '0', '0',    '0',    '0',    '0',    '0'   // 0~9
  '0', '0', '0', '0', '0', '0',    '0',    '0',    '0',    '0'   // 10~19
  '0', '0', '0', '0', '0', '0',    '0',    '0',    '0',    '0'   // 20~29
  '0', '0', '0', '0', '0', '0',    '0',    '0',    '1',    '2'   // 30~39
  '3', '4', '5', '6', '7', '8',    '9',    '0',    '1',    '2'   // 40~49
  '3', '4', '5', '6', '7', '8',    '9',    '0',    '1',    '2'   // 50~59
  '3', '4', '5', '6', '7', '8',    '9',    '0',    '1',    '2'   // 60~69
  '3', '9', '8', '7', '6', '5', '\000', '\000', '\000', '\000' ) // 70~79
```

youngjinj: Wouldn't it be better to run the loop from the back?

ctshim (approved):

youngjinj (approved):

hgryoo (approved):

PR #4097, opened 2023-02-08
[CBRD-24614] Integrate SQL formatter to ‘edit’ csql session command

Srltas:
http://jira.cubrid.org/browse/CBRD-24614

**Purpose**
Want to simplify QUERY tuning and SQL analysis in console without SQL tools and Editing tools.
SQL formatter is available through the ';edit' csql session command option 'format / fmt'.

**Implementation**
N/A

**Remarks**
N/A


src/executables/csql.c
```diff
@@ -117,6 +117,7 @@ int (*csql_text_console_to_utf8) (const char *, const int, char **, int *) = NUL
 
 int csql_Row_count;
 int csql_Num_failures;
+int csql_return_status;
```
eido5: Could you verify that is used as global variable?

If then, I think one character of variable should be the upper character.
Like, csql_Row_count.(csql_row_count)

eido5:
[CBRD-23943](http://jira.cubrid.org/browse/CBRD-23943) this issue noted the indentation about the C based code.
But I'm not sure `indent -l120 -lc120` command is applied in VS or not.

And if you apply this indentation, code-style error will be removed.

or,

// *INDENT-OFF*
/*cpp code*/
// *INDENT-ON*

This code would be needed to remove the code style error for cpp code.

src/executables/csql.c
```diff
@@ -913,7 +919,28 @@ csql_do_session_cmd (char *line_read, CSQL_ARGUMENT * csql_arg)
...
+	}
+      else
+	{
+	  csql_Error_code = CSQL_ERR_FORMAT_OPTION;
+	  nonscr_display_error (csql_Scratch_text, SCRATCH_TEXT_LEN);
+	  break;
+	}
+
+      csql_return_status = csql_invoke_system_editor (csql_Editor_format);
+      if (csql_return_status == CSQL_ERR_FORMAT)
```
eido5: Do you have any special purposes this statement form?

I mean, like killtran, checkpoint case usually compares its return value with expression in the if-statement.

` case S_CMD_CHECKPOINT:
if (csql_arg->sysadm && au_is_dba_group_member (Au_user))`
` case S_CMD_KILLTRAN:
if (csql_arg->sysadm && au_is_dba_group_member (Au_user))`
` case S_CMD_ROLLBACK:
      if (db_abort_transaction () < 0)
`
So, my opinion is not breaking the unity of the code form.
But If you have any reasons that they should be different in this code, could you tell me about that?

msg/en_US/csql.msg
```diff
@@ -153,7 +155,7 @@ $ csql help messages
    ;EXit                        - exit program.\n\
 \n\
    ;CLear                       - clear command buffer.\n\
-   ;EDIT                        - invoke system editor with command buffer.\n\
+   ;EDIT   [format/fmt]         - invoke system editor with command buffer.\n\
```
kisoo-han: ;EDIT   [format/fmt]         - Invoke system editor [after formatter] with command buffer.\n\

msg/en_US/csql.msg
```diff
@@ -164,10 +166,11 @@ $ csql help messages
    ;AUtocommit [ON|OFF]         - enable/disable auto commit mode.\n\
    ;REStart                     - restart database.\n\
 \n\
-   ;SHELL_Cmd    [shell-cmd]    - set default shell, editor, print and pager\n\
+   ;SHELL_Cmd    [shell-cmd]    - set default shell, editor, print, pager and formatter\n\
    ;EDITOR_Cmd   [editor-cmd]     command to new one, or display the current\n\
    ;PRINT_Cmd    [print-cmd]      one, respectively.\n\
    ;PAger_cmd    [pager-cmd]\n\
+   ;FORMATTER_Cmd[formatter-cmd]\n\
```
kisoo-han: add blank or tab between ';**FORMATTER_Cmd**' and '**[formatter-cmd]** '
;FORMATTER_Cmd [formatter-cmd] set formatter.\n\

src/executables/csql.c
kisoo-han: * I propose this case as (agree with @eido5)
```
if (csql_invoke_system_editor ((argument[0] == '\0') ? NULL : argument) != CSQL_SUCCESS)
  {
    return DO_CMD_FAILURE;
  }
break;
```

- and delete variable 'csql_return_status'
- and process rest code in function csql_invoke_system_editor (const char *argument) in csql_support.c
 * check argument is null, 'format', 'fmt' procecess error ...
 * set csql_Error_code = CSQL_ERR_FORMAT in csql_invoke_system_editor ()

src/executables/csql.c
Srltas: Decided not to use global variable. 
Thank you.

src/executables/csql.c
Srltas: I understand. I will modify it to be similar to the existing code.
Thank you.

src/executables/csql.c
eido5: Ok, I got it!

src/executables/csql.c
```diff
@@ -187,6 +189,7 @@ static int csql_column_width_info_list_size = 0;
 static int csql_column_width_info_list_index = 0;
 
 static bool csql_Query_trace = false;
+static bool csql_Editor_format = false;
```
Srltas: not used

kisoo-han (approved):
Finish the trivial things well.
Thanks for your first contribution!

msg/en_US/csql.msg
Srltas: All right, I'll fix it like that.

msg/en_US/csql.msg
Srltas: It's not aligned as below, is it okay?
```
;PAger_cmd    [pager-cmd]
;FOrmatter_cmd [formatter-cmd]
```

src/executables/csql.c
Srltas: That's how the modification was completed. Thank you.

src/executables/csql_support.c
```diff
@@ -239,17 +239,97 @@ csql_invoke_system (const char *command)
...
+  auto[after_filename, after_fileptr] = filesys::open_temp_file ("aft_fmt_");
+  if (after_fileptr == NULL)
+    {
+      csql_Error_code = CSQL_ERR_OS_ERROR;
+      return CSQL_FAILURE;
+    }
+  filesys::auto_delete_file after_file_del (after_filename.c_str ());
+  filesys::auto_close_file after_file (after_fileptr);
+
+  char *cmd = csql_get_tmp_buf (strlen (csql_Formatter_cmd) + 1 + before_filename.size () + 1 + after_filename.size ());
```
eido5:
```suggestion
csql_invoke_formatter ()

int size;
char *cmd
.
.
.


size = strlen (csql_Formatter_cmd) + 1 + before_filename.size () + 1 + after_filename.size ();
  if ((cmd = csql_get_tmp_buf (size)) != NULL);
```

what about this code?

1. It is not easy to figure out what csql_get_tmp_buf has the argument.
2. Without initialized locally, I think variables should be initialized below the first paragraph.(Could you refer the legacy code?)

src/executables/csql_support.c
```diff
@@ -239,17 +239,97 @@ csql_invoke_system (const char *command)
...
+  if (cmd == NULL)
+    {
+      nonscr_display_error (csql_Scratch_text, SCRATCH_TEXT_LEN);
+      return CSQL_FAILURE;
+    }
+  fclose (after_file.release ());
+  sprintf (cmd, "%s %s > %s", csql_Formatter_cmd, before_filename.c_str (), after_filename.c_str ());
+
+  before_file.reset (fopen (before_filename.c_str (), "r"));
+  if (!before_file)
```
eido5, line 306: which value **before_file** have?
if it is 0, what about exchange code like 

> if (system (cmd) != 0)

 (line:302)
 
 In my opinion, the code unity makes other reviewers and other users easy to figure out the code.

src/executables/csql_support.c
Srltas: Consistency to match legacy code.
Legacy codes for reference below.

```
char *cmd = csql_get_tmp_buf (strlen (csql_Editor_cmd) + 1 + filename.size ());
(csql_support.c)

char *cmd = csql_get_tmp_buf (1 + strlen (csql_Print_cmd) + 3 + filename.size ());
(csql.c)
```

src/executables/csql_support.c
Srltas: It is the same as the csql_invoke_system_editor function.
it's like this
```
  file.reset (fopen (filename.c_str (), "r"));
  if (!file)
    {
      csql_Error_code = CSQL_ERR_OS_ERROR;
      nonscr_display_error (csql_Scratch_text, SCRATCH_TEXT_LEN);
      return CSQL_FAILURE;
    }
```

However, if the code is difficult to understand, I will also change the csql_invoke_system_editor function in regacy code.

src/executables/csql_support.c
eido5: Ok. I see. Could you add comment above this line?
As you know, legacy code explains what it is.

`/* invoke the print command */`
`char *cmd = csql_get_tmp_buf (1 + strlen (csql_Print_cmd) + 3 + filename.size ()); (csql.c)`

src/executables/csql_support.c
eido5: I think instead of changing legacy code, the adding comment will be better or could you write down the value in the jira issue?
@kisoo-han How do you think?

We cannot expect what error occur because this is legacy code.

Or if you think this is necessary to change, another issue should be handled this problem.

src/executables/csql_support.c
```diff
@@ -239,17 +239,97 @@ csql_invoke_system (const char *command)
...
+  after_file.reset (fopen (after_filename.c_str (), "r"));
+  if (!after_file)
+    {
+      csql_Error_code = CSQL_ERR_OS_ERROR;
+      return CSQL_FAILURE;
+    }
+
+  if (system (cmd) != 0)
+    {
+      csql_Error_code = CSQL_ERR_FORMAT;
```
airnet73: free_and_init (cmd) is also required when an error occurs in system().

airnet73 (approved):

src/executables/csql_support.c
Srltas: Ok. I'll add comment.
Thank you.

src/executables/csql_support.c
Srltas: Ok. I understand. I'll add comment.
Thank you.

src/executables/csql_support.c
Srltas: That part is missing. I'll correct it.
Thank you.

eido5 (approved):

PR #4098, opened 2023-02-09
[LETS-645] Refactor connection and communication in transaction server

wrlawodms:
http://jira.cubrid.org/browse/LETS-645

To identify which PS a message comes from and to release connection-specific resources, data and functions related to a page server should be packaged in a class: connection_handler. The tran_server has the inner class, _connection_handler_, and active_tran_server and passive_tran_server specialize it in their own way.

It's a refactoring issue, which doesn't make any functional changes.
- **tran_server::connection_handler**
    - includes all connection and communication functions.
    - includes all request handlers.
    - disconnect(): clean up resources and send SNED_DISCONNECT_MSG to a PS. It's created to support per-connection clean-up.
    - get_request_handler(): Unlike page_server:connection_handler, its child classes have their own request handler, so this is used for Polymorphism. The request handlers are defined in connection_handler, so each handler is able to know which connection the msg comes from.
    - no destructor: may be added the following issue to support graceful ps-disconnection.
- **active_tran_server::connection_handler**
    - specialize tran_server::connection_handler for ATS.
    - it includes its sink_hook.
    - disconnect(): Remove sink_hook from prior_sender is moved here from active_tran_server::stop_outgoing_page_server_messages().
- **passive_tran_server::connection_handler**
    - specialize tran_server::connection_handler for PTS.

Note
- The methods using requests internally in *tran_server are still in them, not in connection_handler (e.g. active_tran_server::get_oldest_active_mvccid_from_page_server())
- tran_server::create_connection_handler(): It is for creating its own connection_handler of PTS and ATS in tran_server. (Polymorphism)  
- Registering sink_hooks is moved from tran_server::on_boot to connection_handler's ctor because it is a connection-specific job and makes sense to start when the connection is created.


hornetmj (approved):

src/server/tran_server.hpp
```diff
@@ -72,21 +72,62 @@ class tran_server
...
+
+	void push_request (tran_to_page_request reqid, std::string &&payload);
+	int send_receive (tran_to_page_request reqid, std::string &&payload_in, std::string &payload_out) const;
+
+	virtual void disconnect ();
+
+	const std::string get_channel_id () const;
+
+      protected:
+	virtual request_handlers_map_t get_request_handlers ();
```
cristiarg, line 107: this can be a virtual pure function;
at least one virtual pure function would be needed to make the code explicit and prevent accidental instantiation of `tran_server::connection_handler`

src/server/tran_server.hpp
```diff
@@ -72,21 +72,62 @@ class tran_server
...
+	connection_handler (const connection_handler &) = delete;
+	connection_handler (connection_handler &&) = delete;
+
+	connection_handler &operator= (const connection_handler &) = delete;
+	connection_handler &operator= (connection_handler &&) = delete;
+
+	void push_request (tran_to_page_request reqid, std::string &&payload);
+	int send_receive (tran_to_page_request reqid, std::string &&payload_in, std::string &payload_out) const;
+
+	virtual void disconnect ();
```
cristiarg, line 101: Nit.
This can also be a virtual pure function; the base implementation is small enough to be worth a duplication; the implementation in `active_tran_server` is non-explicit:
```
void
active_tran_server::connection_handler::disconnect ()
{
  remove_prior_sender_sink ();
  tran_server::connection_handler::disconnect ();
}
```
Your choice.

src/server/tran_server.hpp
```diff
@@ -72,21 +72,62 @@ class tran_server
...
-    using request_handlers_map_t = std::map<page_to_tran_request, page_server_conn_t::incoming_request_handler_t>;
+    class connection_handler
+    {
+      public:
+	using page_server_conn_t = cubcomm::request_sync_client_server<tran_to_page_request, page_to_tran_request, std::string>;
+	using request_handlers_map_t = std::map<page_to_tran_request, page_server_conn_t::incoming_request_handler_t>;
+
+	connection_handler () = delete;
+
+	connection_handler (cubcomm::channel &&chn, tran_server &ts, request_handlers_map_t &&request_handlers);
```
cristiarg: I think these ctors can be made protected. They are only used from the derived definitions.
In fact everything can be made protected except the API - `push_request`, `send_receive`, `disconnect`, ...

src/server/active_tran_server.cpp
```diff
@@ -68,7 +68,48 @@ active_tran_server::get_remote_storage_config ()
 }
 
 void
-active_tran_server::receive_saved_lsa (page_server_conn_t::sequenced_payload &a_ip)
+active_tran_server::on_boot ()
```
cristiarg: `on_boot` seems to serve no specific purpose anymore - except the assert, which is superfluous; can it be removed?

src/server/tran_server.hpp
```diff
@@ -72,21 +72,62 @@ class tran_server
 
     int boot (const char *db_name);
 
-    void disconnect_page_server ();
-    bool is_page_server_connected () const;
-    bool is_page_server_connected (const size_t idx) const;
-    void push_request_to (size_t idx, tran_to_page_request reqid, std::string &&payload);
+    /* send request to a specific connection */
+    void push_request (size_t idx, tran_to_page_request reqid, std::string &&payload);
+    int send_receive (size_t idx, tran_to_page_request reqid, std::string &&payload_in, std::string &payload_out) const;
```
cristiarg: These two functions not used publicly. Consider making them private. To be exposed publicly if/when needed.

src/server/tran_server.hpp
```diff
@@ -72,21 +72,62 @@ class tran_server
...
     int send_receive (tran_to_page_request reqid, std::string &&payload_in, std::string &payload_out) const;
 
+    void disconnect_all_page_servers ();
+    bool is_page_server_connected () const;
     virtual bool uses_remote_storage () const;
 
   protected:
-    using page_server_conn_t = cubcomm::request_sync_client_server<tran_to_page_request, page_to_tran_request, std::string>;
-    using request_handlers_map_t = std::map<page_to_tran_request, page_server_conn_t::incoming_request_handler_t>;
+    class connection_handler
```
cristiarg, line 84: Nit.
I would take the `connection_handler` hierarchy outside the `*tran_server` hierarchy.
Your choice.

cristiarg (commented):
Good job. Mostly minor comments and nits.

src/server/tran_server.cpp
```diff
@@ -138,6 +138,33 @@ tran_server::boot (const char *db_name)
...
+{
+  assert (idx < m_page_server_conn_vec.size());
+  m_page_server_conn_vec[idx]->push_request (reqid, std::move (payload));
+}
+
+int
+tran_server::send_receive (size_t idx, tran_to_page_request reqid, std::string &&payload_in,
+			   std::string &payload_out) const
+{
+  assert (idx < m_page_server_conn_vec.size());
```
joohok, line 150: Nit.
I think above assertion means `is_page_server_connected ()` which was used previously in `push_request_to ()`,
and I think it's better to keep using the function as previous because it is easier to understand.

P.S

Currently, the size of that vector does not consider the connection with the PS that was disconnected in the middle.
(the size of the `m_page_server_conn_vec` does not change if TS is initially connected to 10 PSs and disconnected from 2 PSs in the middle.)

Perhaps, I understand that this issue will simply proceed with refactoring, and the issue above (disconnected abnormally) will be dealt with later.

I think in the future, a function that can check whether the connection (idx) is disconnected will be added or 
meaning of the size of the vector will be changed.

I think it would be better to leave `is_page_server_connected ()` for this case too, so that the amount of modification later is small.

joohok (approved):

src/server/tran_server.cpp
cristiarg: > Perhaps, I understand that this issue will simply proceed with refactoring, and the issue above (disconnected abnormally) will be dealt with later.

This is also my understanding. This refactoring should contain no functional changes.
In that regard it only deals with starting all PS's once when ATS is started and disconnecting all PS's when ATS is stopped (via dtors mainly).

wrlawodms: > I think above assertion means is_page_server_connected () which was used previously in push_request_to (),
and I think it's better to keep using the function as previous because it is easier to understand. 

In fact, `is_page_server_connected ()` has no meaning after refactoring. Previously, the communication interface like `push_request()` doesn't know a connection has been established.   But, now they are inside the connection so if they can be called, it means the connection already has been established.

> Perhaps, I understand that this issue will simply proceed with refactoring, and the issue above (disconnected abnormally) will be dealt with later.

Nevertheless, yes, some more handling is needed. But I'm not sure the size would be shrunk. I guess we can just let it empty, and put a new or reconnection connection to it. Let's see then.

src/server/tran_server.hpp
wrlawodms: Good. Done.

src/server/tran_server.hpp
wrlawodms: hmm, do you think it's more readable? Is it the reason? or it could make developers make mistakes?

src/server/tran_server.hpp
wrlawodms: The request handlers in common are going to be into `tran_server::get_request_handlers`. For example, a handler for the disconnection msg will be added here in the following issue.

wrlawodms: > at least one virtual pure function would be needed to make the code explicit and prevent accidental instantiation of tran_server::connection_handler

But, it's a good point. How about 3f3fd72? I've deleted all public ctors of `tran_server::connection_handler` to avoid it.

src/server/active_tran_server.cpp
wrlawodms: I just left it to respect the writer, who may think a callback for booting is needed. I think we can remove it for now.

wrlawodms: done.

src/server/tran_server.hpp
wrlawodms: I'd like to keep it since more common behaviors could be added and it makes sense to wrap the common behaviors in one place.

src/server/tran_server.hpp
wrlawodms: done.

src/server/tran_server.cpp
cristiarg: > Nevertheless, yes, some more handling is needed. But I'm not sure the size would be shrunk. I guess we can just let it empty, and put a new or reconnection connection to it. Let's see then.

I don't think I understand fully @joohok 's concern:

> Currently, the size of that vector does not consider the connection with the PS that was disconnected in the middle.
(the size of the m_page_server_conn_vec does not change if TS is initially connected to 10 PSs and disconnected from 2 PSs in the middle.)

But here are my 2cents.
There are a few ways in which a connection handler can break down once it is established:
- the other side (PS) does not respond anymore (crashed, hanged ...) and the send buffers in the network layer - not even in the application, but down from the OS upwards, become full and some error is returned back from somewhere in cubrid's css code; the connection handler will need a way to cope with this - take a look how it is done in page_server/connection_handler - each connection handler has a reference to the page_server object instance an can basically call upwards with "remove me: I'm dead/invalid/hanged/cannot work anymore"
- it can be removed by the owner/parent - tran_server object instance as a result of some higher-up decision (eg: cluster management software deciding that one PS is to be de-activated; or some error in the higher application layers)
- what not ...

There are also transient errors which the connection handler (+ the infrastructure below) needs to cope with (eg: like a single request fails with some error code; or with a timeout - we don't currently have timeouts implemented, but we'll need to add them).

None are implemented here with this patch.

cristiarg (approved):

src/server/tran_server.cpp
joohok: OK, I understand. 
Thanks for elaboration!

joohok (approved):

PR #4104, opened 2023-02-11
[LETS-644] add perf stat to allow calculation of data pages compression ratio

cristiarg:
http://jira.cubrid.org/browse/LETS-644

Used two accumulators and a ratio to measure:
- the amount of data that is be sent compressed
- the amount of data that would be sent uncompressed
- the calculated ratio between them - in percent


razvanradulescu (approved):
If this is not temporary, and you intend to keep it, I think it will be misleading to display as PSTAT_COUNTER_TIMER_VALUE.
If we give up on max compress ratio (it is not relevant since we don't have any info on distribution), a proper way (but with more code) could be two PSTAT_ACCUMULATE_SINGLE_VALUE (for amount compressed + uncompressed) and one PSTAT_COMPUTED_RATIO_VALUE.

cristiarg:
> If this is not temporary, and you intend to keep it, I think it will be misleading to display as PSTAT_COUNTER_TIMER_VALUE. If we give up on max compress ratio (it is not relevant since we don't have any info on distribution), a proper way (but with more code) could be two PSTAT_ACCUMULATE_SINGLE_VALUE (for amount compressed + uncompressed) and one PSTAT_COMPUTED_RATIO_VALUE.

@razvanradulescu 
I think it will be in the code - for a while at least.
I don't mind the code. I also thought about using 2 values and a 3rd computed one.
The max gives some information - but not much, as you also say.
Yet another option I considered is extending the statistics mechanism with smth similar with `PSTAT_COUNTER_TIMER_VALUE` but which is able to store and compute with floating point values.

wrlawodms (approved):
For the most part, I agree with @razvanradulescu.
but with two PSTAT_ACCUMULATE_SINGLE_VALUE:
- amount compressed
- num pages to be fetched to TS
and one PSTAT_COMPUTED_RATIO_VALUE.

I think the two single values above can achieve the same goal and also may be helpful to investigate the cluster's behavior. Just opinion.

joohok (approved):

hornetmj (approved):

PR #4106, opened 2023-02-13
[CBRD-24659] Add linux style absolute path as windows style also

kisoo-han:
http://jira.cubrid.org/browse/CBRD-24659

**Purpose**

* To add Linux style absolute file path to absolute path for Windows

* ASIS

  * C:\tmp/admin/adm.log is **absolute** path for WIndows

  * C:/tmp/admin/adm.log is relative path for WIndows 
* TOBE
 * C:\tmp/admin/adm.log is **absolute** path for WIndows
 * C:/tmp/admin/adm.log is **absolute** path for WIndows as well
 
**Implementation**

**Remarks**

### DIFF
--- a/src/base/porting.h
+++ b/src/base/porting.h
@@ -354,7 +354,7 @@ check_is_array (const T & a)
 
 #if defined (WINDOWS)
 #define IS_ABS_PATH(p) IS_PATH_SEPARATOR((p)[0]) \
-	|| (isalpha((p)[0]) && (p)[1] == ':' && IS_PATH_SEPARATOR((p)[2]))
+	|| (isalpha((p)[0]) && (p)[1] == ':' && (IS_PATH_SEPARATOR((p)[2]) || (p)[2] == '/'))
 #else /* WINDOWS */
 #define IS_ABS_PATH(p) IS_PATH_SEPARATOR((p)[0])
 #endif /* WINDOWS */
--- a/src/base/porting.h
+++ b/src/base/porting.h
@@ -350,11 +350,15 @@ check_is_array (const T & a)
 #endif /* WINDOWS */
 #define PATH_CURRENT    '.'
 
+#if defined (WINDOWS)
+#define IS_PATH_SEPARATOR(c) ((c) == PATH_SEPARATOR || (c) == '/')
+#else
 #define IS_PATH_SEPARATOR(c) ((c) == PATH_SEPARATOR)
+#endif
 
 #if defined (WINDOWS)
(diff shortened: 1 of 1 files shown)


src/base/porting.h
```diff
@@ -354,7 +354,7 @@ check_is_array (const T & a)
 
 #if defined (WINDOWS)
 #define IS_ABS_PATH(p) IS_PATH_SEPARATOR((p)[0]) \
-	|| (isalpha((p)[0]) && (p)[1] == ':' && IS_PATH_SEPARATOR((p)[2]))
+	|| (isalpha((p)[0]) && (p)[1] == ':' && (IS_PATH_SEPARATOR((p)[2]) || (p)[2] == '/'))
```
mhoh3963: I guess you need to check that p[0] == '/' is also an absolute path.
And if it's a window, how about not using '/' directly, but adding a variable defined like this:
#define LINUX_PATH_SEPARATOR '/'

kisoo-han: Yes, you right, but I want to minimize impact of modification so conservative correction was made.
Basically, I agree upon your suggestion, I will take a look for the impact after modification.

mhoh3963: OK, after checking impact of modification, I want to you consider my suggestion.

mhoh3963 (approved):

airnet73 (approved):

src/base/porting.h
kisoo-han: My intention was to take your comment and take modification to decide following cases below to absolute paths in the Windows, with careful investigation for side-effects:
1. /abc/tmp.log
2. c:/abc/tmp.log (extended by this PR)
3. c:\abc/tmp.log (or c:\abc\tmp.log)

I'll ask for your opinion again.

1. keep the current revision of this PR for **IS_ABS_PATH ()**
2. includes all three cases above for the Windows absolute path 

Of course, this modification will have a greater impact than the previous one, but it is necessary when considering the environment such as MinGW.

kisoo-han: @mhoh3963, @airnet73
I think we have to take **option 2** and pushed the codes for it. 
- If you agree with my opinion and accept the changes, just approve or leave a comment. 
- Or please leave comments, then I'll reflect your opinion including back to previous commit. 
- This push [2b2fa3](https://github.com/CUBRID/cubrid/pull/4106/commits/2b2fa34adca884ab35381a4e9e45bdc2758b19e6) is just for your choice for the options

Thank you.

mhoh3963 (approved):
LGTM^^

airnet73 (approved):

PR #4118, opened 2023-02-17
[CBRD-24665] removing ncurses dependency from broker_monitor

kisoo-han:
http://jira.cubrid.org/browse/CBRD-24665

**Purpose**
* To remove ncurses dependency on **broker_monitor**
* In Linux, **ncurses** function will not be called,
* Instead, Linux command watch will be called with given arguments (-s number will be changed to -s 0)
* User Invoked:
`cubrid broker status -s 1 -f`
* Actual action
`watch -t -n 1 cubrid broker status -s 0 -f`

**Implementation**

**Remarks**
* dummy **ncurses** functions such as endwin () are added, 
* if not, we have to refactor this code that remove all ncurses reference for the Linux.
* addstr (), getch () is called only when refresh_sec > 0 && !tty_mode in Linux,
* this situation will not be happen if (refresh_sec > 0 && !tty_mode) then 
  system call to execute watch will be called, and broker_monitor will be terminated in **Linux**.

### DIFF
--- a/src/broker/broker_monitor.c
+++ b/src/broker/broker_monitor.c
@@ -514,7 +514,12 @@ main (int argc, char **argv)
 	}
 
       free (br_vector);
-      err = system (cmd);
+
+      if (system (cmd) != 0)
+	{
+	  fprintf (stderr, "%s: -s option require Linux command: '%s'\n", argv[0], WATCH_CMD);
+	  fprintf (stderr, "Install the '%s' command to use this option.\n", WATCH_CMD);
+	}
 
       return 0;
 #endif
--- a/src/broker/broker_monitor.c
+++ b/src/broker/broker_monitor.c
@@ -515,7 +515,7 @@ main (int argc, char **argv)
 
       free (br_vector);
 
-      if (system (cmd) != 0)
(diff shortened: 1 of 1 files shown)


kisoo-han:
@mhoh3963, @airnet73 ,
Please comments!

src/broker/broker_monitor.c
```diff
@@ -519,9 +505,18 @@ main (int argc, char **argv)
 	}
 //  FillConsoleOutputCharacter(h_console, ' ', scr_info.dwSize.X * scr_info.dwSize.Y, top_left_pos, &size);
 #else
-      win = initscr ();
-      timeout (refresh_sec * 1000);
-      noecho ();
+      snprintf (cmd, PATH_MAX, "%s -t -n %d", WATCH_CMD, refresh_sec);
```
mhoh3963, line 508: if there is not a watch command in the running machine, you need to output a message such as "please install watch command for using this option..".

mhoh3963 (approved):

mhoh3963 (approved):

airnet73 (approved):

PR #4119, opened 2023-02-17
[LETS-649] A race condition when connecting/disconnecting TS to PS

wrlawodms:
http://jira.cubrid.org/browse/LETS-649

To prevent the race condition among TS connection and disconnection, a mutex for them is added.

### DIFF
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -489,6 +489,8 @@ page_server::set_active_tran_server_connection (cubcomm::channel &&chn)
   er_log_debug (ARG_FILE_LINE, "Active transaction server connected to this page server. Channel id: %s.\n",
 		channel_id.c_str ());
 
+  std::lock_guard lk_guard (m_conn_mutex);
+
   if (m_active_tran_server_conn != nullptr)
     {
       // When [A]TS crashes there are two possibilities:
@@ -515,7 +517,11 @@ page_server::set_passive_tran_server_connection (cubcomm::channel &&chn)
   er_log_debug (ARG_FILE_LINE, "Passive transaction server connected to this page server. Channel id: %s.\n",
 		channel_id.c_str ());
 
-  m_passive_tran_server_conn.emplace_back (new connection_handler (chn, transaction_server_type::PASSIVE, *this));
+  {
+    std::lock_guard lk_guard (m_conn_mutex);
+
+    m_passive_tran_server_conn.emplace_back (new connection_handler (chn, transaction_server_type::PASSIVE, *this));
+  }
 
   m_pts_mvcc_tracker.init_oldest_active_mvccid (channel_id);
 }
@@ -540,14 +546,18 @@ void
 page_server::disconnect_tran_server_async (const connection_handler *conn)
 {
   assert (conn != nullptr);
+
+  std::lock_guard lk_guard (m_conn_mutex);
+
   if (conn == m_active_tran_server_conn.get ())
     {
+      er_log_debug (ARG_FILE_LINE, "Page server disconnected from active transaction server with channel id: %s.\n",
+		    conn->get_connection_id ().c_str ());
       m_async_disconnect_handler.disconnect (std::move (m_active_tran_server_conn));
       assert (m_active_tran_server_conn == nullptr);
     }
   else
     {
-      bool passive_tran_server_found { false };
       for (auto it = m_passive_tran_server_conn.begin (); it != m_passive_tran_server_conn.end (); ++it)
 	{
 	  if (conn == it->get ())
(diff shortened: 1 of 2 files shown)


src/server/page_server.cpp
```diff
@@ -557,17 +567,17 @@ page_server::disconnect_tran_server_async (const connection_handler *conn)
...
 	    }
 	}
-      assert (passive_tran_server_found);
     }
 }
 
 void
 page_server::disconnect_all_tran_server ()
 {
+  std::lock_guard lk_guard (m_conn_mutex);
```
wrlawodms, line 587: I think the scope for this is too large. Especially, the dtor for a connection_handler is quite a long job, which even waits for joining internal threads. However, I just wrap them all now because they are going to be changed in the following issue ([LETS-639](http://jira.cubrid.org/browse/LETS-639))

src/server/page_server.cpp
```diff
@@ -557,17 +567,17 @@ page_server::disconnect_tran_server_async (const connection_handler *conn)
 	      m_async_disconnect_handler.disconnect (std::move (*it));
 	      assert (*it == nullptr);
 	      m_passive_tran_server_conn.erase (it);
-	      passive_tran_server_found = true;
 	      break;
 	    }
 	}
-      assert (passive_tran_server_found);
```
wrlawodms, line 564: Now, it's possible. If the connection is not in the container, it means another thread has also done it.

joohok (approved):

joohok:
I think build failure can be solved if the branch is merged upto the commit (#4103 )

hornetmj (approved):

src/server/page_server.cpp
cristiarg: This function is only called explicitly from parts of the program where the Page Server is already in a shutdown sequence.
Thus, the mutex should not have any negative performance implications.
What could have actually a negative effect, is, if at this point - while the server still has all its inbound/outbound network communication infrastructure up - a new connection from a TS is attempted (can happen from a PTS).
It would be needed to limit such a situation with a `m_shutdown` flag which should avoid accepting new connections while the PS is going down.
What is your opinion about this?

src/server/page_server.cpp
```diff
@@ -515,7 +517,11 @@ page_server::set_passive_tran_server_connection (cubcomm::channel &&chn)
   er_log_debug (ARG_FILE_LINE, "Passive transaction server connected to this page server. Channel id: %s.\n",
 		channel_id.c_str ());
 
-  m_passive_tran_server_conn.emplace_back (new connection_handler (chn, transaction_server_type::PASSIVE, *this));
+  {
+    std::lock_guard lk_guard (m_conn_mutex);
```
cristiarg, line 529: Nit.
For the functions `set_active_tran_server_connection` and `set_passive_tran_server_connection` the mutex would actually not be needed because these functions are executed both on the master connection loop (`css_master_thread`). The usage of the mutex is to synchronize with the disconnects which can happen more asynchronously.

Maybe it's useful to add this comment somewhere.

wrlawodms: As you said, these new connections are handled by css_master_thread, and the disconnection request is handled by the request handler threads for each connection. So, I think they have to be exclusive. I don't understand `the mutex would actually not be needed`. The new connection is rare, but I don't think it means they don't need the mutex.

src/server/page_server.cpp
wrlawodms: I didn't know that a new connection can be established even after `finalize_server_type()`. If it's the case, we should block it. It may cause the new connection to suffer from communication errors right after establishing the connection. I think it should be dealt with in another issue.

src/server/page_server.cpp
cristiarg: @wrlawodms 
My comment is not negative.
It was only an explanation for why it is needed to guard both `set_active_tran_server_connection` and `set_passive_tran_server_connection` together with the disconnect functionality even if these two functions are called from the same thread.

The idea was to draw a comment in the code from my comment:
```
// Even if the functions set_active_tran_server_connection and set_passive_tran_server_connection are
// called from the same thread (master-server connection handler) the mutex is actually needed.
// The usage of the mutex is to synchronize with the disconnects which are trigered from each connection
// handler's connection threads (inbound or outbound).
```

wrlawodms: Aha, OK. I got it wrong. I'll add the comment in both functions.

src/server/page_server.cpp
cristiarg: `finalize_server_type` server type is called after `disconnect_all_tran_server` (see `xboot_shutdown_server`).
There is a small window of opportunity. In practice it may never be a problem.
But, as you say, a PTS booting up in this window of opportunity will most probably have to deal with it.
Lets forget about this for now.
If you want, you can add a new issue about this: a shutting down `page_server` must block new connections to be made.
Up to you.

cristiarg (approved):

src/server/page_server.cpp
wrlawodms: Ok, let's forget about this for now.

PR #4120, opened 2023-02-17
[CBRD-24624] Add ability to load multiple schema files from loaddb

airnet73:
http://jira.cubrid.org/browse/CBRD-24624

Purpose
You can load schema files that are separated into several in loaddb one by one, but if you load them one by one, errors may occur because the order of loading is incorrect.
To reduce these inconveniences and errors, a file in which the loading order is written is provided, and schema files can be loaded in order through this file.

Implementation
cubrid loaddb -u user --multiload-schema-file=schema_file_name testdb

Remarks
N/A


msg/en_US/utils.msg
```diff
@@ -489,32 +489,33 @@ $set 12 MSGCAT_UTIL_SET_LOADDB
...
+  -s, --schema-file=FILE[:LINE]      create schema from LINE with FILE\n\
+  -i, --index-file=FILE[:LINE]       create index from LINE with FILE\n\
+      --trigger-file=FILE[:LINE]     create trigger from LINE with FILE\n\
+  -d, --data-file=FILE               load data with FILE\n\
+  -t, --table=TABLE                  table name that is substituted for missing class header in data file\n\
+      --error-control-file=FILE      FILE to control error(s) during loading\n\
+      --ignore-class-file=FILE       input file of class names that skip load\n\
+      --no-user-specified-name       Find classes, serials, and triggers by their object names without their owner names.\n\
+                                     This is to load files generated by cubrid before version    11.2.\n\
+      --multiload-schema-file=FILE   File for loading multiple schema files, The content of this file must be a schema file created with the --split-schema-file option of unloaddb.\n
```
kisoo-han: --schema-file-list=FILE  The content of this file is a list of schema file names to be used in loaddb. loaddb is executed sequentially from the top.

msg/ko_KR.utf8/utils.msg
```diff
@@ -488,32 +488,33 @@ $set 12 MSGCAT_UTIL_SET_LOADDB
...
+  -s, --schema-file=FILE[:LINE]      파일[:라인] 정보로 스키마 생성\n\
+  -i, --index-file=FILE[:LINE]       파일[:라인] 정보로 인덱스 생성\n\
+      --trigger-file=FILE[:LINE]     파일[:라인] 정보로 트리거 생성\n\
+  -d, --data-file=FILE               적재할 데이터 파일\n\
+  -t, --table=TABLE                  데이터를 적재할 테이블 이름; 데이터 파일에 테이블 정보가 없는 경우 사용\n\
+      --error-control-file=FILE      적재 시 발생하는 에러에 대한 제어 파일\n\
+      --ignore-class-file=FILE       적재하지 않을 클래스 이름이 있는 파일\n\
+      --no-user-specified-name       소유자 이름 없이 객체 이름으로 CLASS, SERIAL 및 TRIGGER를 찾음.\n\
+                                     CUBRID 11.2 전 버전에서 생성된 파일을 불러오기 위함.\n\
+      --multiload-schema-file=FILE   여러개 schema file을 로드하기 위한 파일, 해당 파일의 내용은 unloaddb의--split-schema-file 옵션으로 생성된 schema file 이여야 한다.\n
```
kisoo-han: --schema-file-list=FILE 이 파일의 내용은 loaddb에서 사용할 스키마 파일 이름 목록이다. loaddb는 위에서부터 순차적으로 실행된다.

src/loaddb/load_db.c
```diff
@@ -50,6 +50,13 @@ static FILE *loaddb_log_file;
 int interrupt_query = false;
 bool load_interrupted = false;
 
+typedef struct t_mulitload_schema_file_info T_MULTILOAD_SCHEMA_FILE_INFO;
+struct t_mulitload_schema_file_info
+{
+  char schema_file_name[100];
```
kisoo-han: 100 may not be enough.
`char schema_file_name[PATH_MAX];`

kisoo-han:
The result of the following command should be clear, either usage or manual.

assume that content of the file xy.txt is:
`demodb_schema_vclass`

`cubrid loaddb -s demodb_schema_class --multiload-schema-file=xy testdb`

1. Whether to allow use of -s and --multiload-schema-file options together
2. If allowed, which option has higher priority

I think these options should be used exclusively

msg/en_US/utils.msg
airnet73: thanks for your comment.
I'll change it.

msg/ko_KR.utf8/utils.msg
airnet73: thanks for your comment.
I'll change it.

src/loaddb/load_db.c
airnet73: thanks for your comment.
I'll change it.

airnet73:
> The result of the following command should be clear, either usage or manual.
> 
> assume that content of the file xy.txt is: `demodb_schema_vclass`
> 
> `cubrid loaddb -s demodb_schema_class --multiload-schema-file=xy testdb`
> 
> 1. Whether to allow use of -s and --multiload-schema-file options together
> 2. If allowed, which option has higher priority
> 
> I think these options should be used exclusively

airnet73:
> > The result of the following command should be clear, either usage or manual.
> > assume that content of the file xy.txt is: `demodb_schema_vclass`
> > `cubrid loaddb -s demodb_schema_class --multiload-schema-file=xy testdb`
> > 
> > 1. Whether to allow use of -s and --multiload-schema-file options together
> > 2. If allowed, which option has higher priority
> > 
> > I think these options should be used exclusively

-s and --multiload-schema-file can be used together.
And, the order of execution is --multiload-schema-file, then -s.
There are overlapping actions, but it also seems to have the advantage of being diversified in how to use them.
What do you think? tell me your opinions.

src/loaddb/load_db.c
```diff
@@ -685,68 +706,45 @@ loaddb_internal (UTIL_FUNCTION_ARG * arg, int dba_mode)
...
+		{
+		  goto error_return;
+		}
 
-      /* update catalog statistics */
-      AU_DISABLE (au_save);
-      sm_update_all_catalog_statistics (STATS_WITH_FULLSCAN);
-      AU_ENABLE (au_save);
+	      fclose (schema_file_list[i]->schema_fp);
+	      schema_file_list[i]->schema_fp = NULL;
```
kisoo-han: You used this style statements frequently, what about define macro in front.

```
#define fclose_and_init (ptr) \
        do { \
          if ((ptr)) { \
            fclose (((ptr)); \
            (ptr) = NULL; \
          } \
        } while (0)
```

kisoo-han (approved):
Well DONE !
I agree with your modification.

kisoo-han:
Please CHECK CI fails.

kisoo-han (approved):

src/loaddb/load_db.c
```diff
@@ -1361,3 +1372,196 @@ load_object_file (load_args * args, int *exit_status)
...
+
+  error_code = ER_FAILED;
+  num_files = 0;
+  return NULL;
+}
+
+static void
+ldr_free_and_fclose (T_SCHEMA_FILE_LIST_INFO ** file_list, int num)
+{
+  int i;
```
eido5, line 1484: Not serious problem. 

Could we think of idx instead of i?
It seems so general name. Of course we are easy to know that is file list index.
But to me, It is seems so general. 

But it is up to you, it is very subjective idea.

eido5 (approved):

Srltas (approved):

src/loaddb/load_db.c
airnet73: Thank you for your comments.
I will fixed it.

PR #4129, opened 2023-02-21
[CBRD-24671] Disable deprecated functions in lock manager.

hyahong:
http://jira.cubrid.org/browse/CBRD-24671

**Purpose**

Cover a deprecated function, lock_remove_object_lock with macro `ENABLE_UNUSED_FUNCTION`, not to be reflected in compilatoin.

**Implementation**
N/A

**Remarks**
N/A


hornetmj (approved):

hornetmj:
@Yechan0815 How about handling the lock_has_lock_transaction() together?

hyahong:
@hornetmj 
Ah... I wrote something wrong in JIRA. The function, lock_has_lock_transaction was already processed.

hornetmj:
@Yechan0815 OK.

wrlawodms (approved):
Why don't remove it, but leave it out? It's just a wrapper function so there seems no upside in keeping it.

hyahong:
@wrlawodms 
There is a function above that work similar to `lock_remove_object_lock ()`, so I have kept this function for comparison in future maintenance.
But.. I think removing it would be fine too. I don't know which one is better in terms of maintenance.
Can I get some advice?

joohok (approved):

wrlawodms:
@Yechan0815 If I were you, I would remove it as I said above. But I don't know well about the context of lock manager. At your choice.

hyahong:
@wrlawodms 
Thanks for your advice! It was helpful to me to decide whether to remove it or not.
I was determined to get rid of it from the code.

PR #4130, opened 2023-02-21
[CBRD-24628] Change the csql CLI library not to use ncurses in the Linux.

eido5:
http://jira.cubrid.org/browse/CBRD-24628

**Purpose**

On the csql execution, when the user type the 'tab' key, the csql shows the current directory elements.(directories, files, etc)
To user, it seems unexpected execution. So, it should not be executed.
We try to eliminate this function, and then the CLI(command-line interface) should be changed that the reason of unexpected execution occurred from typing 'tab' key.
Without tab key, every function must be worked.

The entry of CLI library is below;
[simple editline](https://github.com/troglobit/editline#api)

**Implementation**

el_prev_hist() : return the ptr of history string from the latest one.
el_next_hist () :  return the ptr of history string from the oldest one.
csql_edit_contents_append() : store a history entry in the edit buffer.
add_history() : add a history in the CLi library buffer

**Remarks**

N/A


airnet73:
A build error has occurred. Please check.

eido5:
> 

Thank you I will do rebase onto this pr again.

kisoo-han:
@eido5 ,
Would you describe following in description:
1. Remove ncurses dependency from csql
2. change of history related library from libedit to editline

eido5:
Build test will not pass because I must change the path of CLi.
It will be fixed soon. (about 1 or 2 days)
I should create other repo and update modified CLi. So, It takes times a little bit.

@kisoo-han 
Could we ignore for a while till CLi will be modified?

mhoh3963 (commented):
I suggest a few thing.

(1) The code complexity seems to have increased.
      so, how about making history_get() with el_prev_hist() and wl_next_hist() ?
     
(2) I knew that the replaced library support saving and reading a history string in the file.
      How about adding  new feature to read and save in the history file using this function ?

src/executables/csql_support.c
```diff
@@ -1336,3 +1337,11 @@ csql_errmsg (int code)
       return (csql_get_message (CSQL_E_UNKNOWN_TEXT));
     }
 }
+
```
kisoo-han, line 1340: #if !defined(WINDOWS)

src/executables/csql_support.c
```diff
@@ -1336,3 +1337,11 @@ csql_errmsg (int code)
...
+void
+stifle_history (int max_history_size)
+{
+  el_hist_size = max_history_size;
+
+  return;
+}
```
kisoo-han:
```suggestion
#if !defined(WINDOWS)
void
stifle_history (int max_history_size)
{
  el_hist_size = max_history_size;

  return;
}
#endif /* !WINDOWS */
```

eido5: I will apply it. 

Thank you.

src/executables/csql.c
```diff
@@ -1806,6 +1842,13 @@ csql_execute_statements (const CSQL_ARGUMENT * csql_arg, int type, const void *s
       logddl_set_csql_input_type (CSQL_INPUT_TYPE_EDITOR);
     }
 
+  strcpy (prompt_cpy, stmts);
```
airnet73: stmts may be larger than 100 bytes.

eido5: Right. Its size expands till the memory lack.

I should consider about it. Thank you.

src/executables/csql.c
```diff
@@ -1756,6 +1790,8 @@ csql_execute_statements (const CSQL_ARGUMENT * csql_arg, int type, const void *s
   bool do_abort_transaction = false;	/* flag for transaction abort */
   PT_NODE *statement = NULL;
   char sql_text[DDL_LOG_BUFFER_SIZE] = { 0 };
+  char prompt_cpy[100];		/*static char csql_Prompt[100] */
```
eido5: necessary to fix

src/executables/csql.c
eido5: add_history is needed to be fixed.

PR #4132, opened 2023-02-21
[LETS-639] Refactor async_disconnect_handler to use it in tran_server too

wrlawodms:
http://jira.cubrid.org/browse/LETS-639

We need to add the "disconnection request" from PS to TS. In the request handler, it will try to delete connection_handler object, but it ends up kind of deadlock. The request handler is called by the internal request receiver thread in connection_handler. If it tries to delete itself, it tries to join itself in the dtor (~connection_handler). To address this issue, we can pass the deletion job to another thread. At first, I come up with using a detached thread, but the `async_disconnect_handler` has been introduced seemingly for this purpose in #3407 and a more robust way. So, I make it a template class to be used in not only page_server but also tran_server.

- No functional change
- Make the async_disconnect_handler a template class to be used in common for tran_server and page_server.
- An prior issue for the main PR for LETS-639.


src/server/async_disconnect_handler.hpp
```diff
@@ -0,0 +1,63 @@
+/*
+ * Copyright 2008 Search Solution Corporation
```
joohok: I think we don't need this line
```suggestion

```

joohok (approved):

hornetmj (approved):

src/server/async_disconnect_handler.cpp
```diff
@@ -0,0 +1,102 @@
+/*
+ * Copyright 2008 Search Solution Corporation
```
wrlawodms:
```suggestion
```

src/server/async_disconnect_handler.cpp
```diff
@@ -0,0 +1,101 @@
+/*
+ * Copyright 2016 CUBRID Corporation
```
joohok:
```suggestion
  *
  * Copyright 2016 CUBRID Corporation
```

Sorry for wrong suggestion...
I followed the license in schema_class_truncator.hpp

src/server/async_disconnect_handler.hpp
```diff
@@ -0,0 +1,62 @@
+/*
+ * Copyright 2016 CUBRID Corporation
```
wrlawodms, line 3:
```suggestion
 *
 * Copyright 2016 CUBRID Corporation
```

src/server/async_disconnect_handler.cpp
wrlawodms: I didn't check it too. Thank you. :)

src/server/async_disconnect_handler.hpp
```diff
@@ -0,0 +1,63 @@
...
+#include <mutex>
+#include <atomic>
+#include <queue>
+#include <condition_variable>
+
+/* helper class with the task of destroying connnection handlers and, by this,
+ * also waiting for the receive and transmit threads inside the handlers to terminate
+ */
+template <typename T_CONN_HANDLER>
+class async_disconnect_handler
```
cristiarg, line 33: A TODO here would be needed. Or a separate Jira issue.
This class also needs to have a timeout functionality. Based on a system parameter, the class must wait up to a certain time interval to attempt to delete an object. If the delete is not fulfilled in that interval - being that the actual delete of an object depents of finishing some I/O rountrips to the peer server - it is a catastrophic error. (something that works similar to `pgbuf_timed_sleep`)
What do you think?

cristiarg (approved):

src/server/async_disconnect_handler.hpp
cristiarg: I have added a new epic - http://jira.cubrid.org/browse/LETS-655 - to track this issue. It is not urgent though.

PR #4133, opened 2023-02-22
[LETS-653] [LETS-528] Initialize ps_Gl only when it is needed

wrlawodms:
http://jira.cubrid.org/browse/LETS-653
http://jira.cubrid.org/browse/LETS-528 

Use unique_ptr for `ps_Gl` and initialize it only when needed. See the Jira issue.

### DIFF
--- a/src/server/page_server.hpp
+++ b/src/server/page_server.hpp
@@ -249,7 +249,4 @@ class page_server
     pts_mvcc_tracker m_pts_mvcc_tracker;
 };
 
-extern std::unique_ptr<page_server> ps_Gl;
-
 #endif // !_PAGE_SERVER_HPP_
-
--- a/src/server/page_server.hpp
+++ b/src/server/page_server.hpp
@@ -249,7 +249,7 @@ class page_server
     pts_mvcc_tracker m_pts_mvcc_tracker;
 };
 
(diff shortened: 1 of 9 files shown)


hornetmj (approved):
LGTM.

joohok (approved):

src/server/page_server.hpp
```diff
@@ -249,7 +249,7 @@ class page_server
     pts_mvcc_tracker m_pts_mvcc_tracker;
 };
 
-extern page_server ps_Gl;
+extern std::unique_ptr<page_server> ps_Gl;
```
cristiarg: Could we move this in `server_type.h` - similar to where `ts_Gl` is?
The include clauses would be leaner.

cristiarg:
@wrlawodms 
Please also reference http://jira.cubrid.org/browse/LETS-528 .
Both in the title of the PR and in the comment section:
```
[LETS-653][LETS-528] Initialize ps_Gl only when it is needed
```

src/server/page_server.hpp
wrlawodms: Thank you, I missed it.
It's moved in server_type.hpp/cpp inside SERVER_MODE.

PR #4134, opened 2023-02-22
[CBRD-24620] Add new csql session command 'connect'

eido5:
http://jira.cubrid.org/browse/CBRD-24620

**Description**

The user typing that the below command to connect other users is not intuitive, while the CSQL session is connected;

CALL login ('user_name', 'password') ON CLASS db_user;
So, we should add the session command user to use more intuitive while the CSQL session is connected.

;connect user_name db_name@host_name

This can helps user to connect more intuitive while the CSQL session is connected

**Implementation**

- Make new switch condition 'S_CMD_CONNECT:' In csql_do_session_cmd().
- Parse the argument to connect user, db and host
- Connect through db_restart_ex() function.
- Print the 'Connected.' to ensure the connection.

**Remarks**

N/A


eido5:
I am fixing the process when the db_name, host_name, user_name are is NULL.
So, could you review except these case?

src/executables/csql.c
```diff
@@ -1455,6 +1459,195 @@ csql_do_session_cmd (char *line_read, CSQL_ARGUMENT * csql_arg)
 	}
 
       break;
+
+    case S_CMD_CONNECT:
+      /*dbname_cpy can be stored DB name+ @ + remote hostname + \0 */
+      char dbname_cpy[MAXLOGNAME + CUB_MAXHOSTNAMELEN + 2];
```
airnet73: MAXLOGNAME define is not appropriate.

eido5: @airnet73 
I used MAXLOGNAME because when we create db, the name length is limited by MAXLOGNAME.
So, do we need to make other macro like below?
`#define DB_NAME_LEN MAXLOGNAME`

msg/en_US/csql.msg
```diff
@@ -113,6 +113,7 @@ $ These 3 should be blank
 196 An undefined error has occurred.
 197 ';edit' is not allowed when output has been redirected.
 198 Format failed.
+199 Invalid username/password/dbname/hostname; login denied.
```
eido5: @kisoo-han  @airnet73  @Srltas 
Could you review this sentence first?
I need to check this sentence to translate either other languages.

src/executables/csql.c
airnet73: Yes, it would be nice to add a macro.
I'm not sure if 64 is the right value for DB_NAME.
#define DB_NAME_LEN (64)

src/executables/csql.c
```diff
@@ -1460,6 +1464,205 @@ csql_do_session_cmd (char *line_read, CSQL_ARGUMENT * csql_arg)
...
+      char *save_ptr_strtok = NULL;
+      char *db_name = NULL;
+      char *user_name = NULL;
+      char *host_name = NULL;
+      char *p = NULL;
+      CSQL_ARGUMENT csql_new_arg;
+
+      /*find @hostname */
+      host_name = strchr (argument, '@');
+      if (host_name != NULL)
```
airnet73: How about this code?

```
if (host_name != NULL)
  {
    if (*(host_name + 1) == '\0')
      {
        host_name = NULL;
        csql_Error_code = CSQL_ERR_CONNECT;
        return DO_CMD_FAILURE;
      }

      if (strcasecmp ("127.0.0.1", host_name +1 ) == 0 || strcasecmp ("localhost", host_name +1) == 0
            || strcasecmp(boot_get_host_name(), host_name+1) == 0 || strcasecmp(boot_get_ip(), host_name+1) == 0)
        {
          *(host_name ) = '\0';
        }
    }

```

src/executables/csql.c
```diff
@@ -1460,6 +1464,205 @@ csql_do_session_cmd (char *line_read, CSQL_ARGUMENT * csql_arg)
...
+		{
+		  p = getpass ((char *) csql_get_message (CSQL_PASSWD_PROMPT_TEXT));
+		  error_code = au_login (user_name, p, false);
+		}
+	    }
+	  else
+	    {
+	      error_code = au_login (user_name, NULL, false);
+	    }
+	  if (error_code == NO_ERROR)
```
airnet73: It is recommended to put a space or two for readability.

src/executables/csql.c
```diff
@@ -1460,6 +1464,205 @@ csql_do_session_cmd (char *line_read, CSQL_ARGUMENT * csql_arg)
...
+	    {
+	      error_code = clogin_user (user_name);
+	    }
+	  else
+	    {
+	      return DO_CMD_FAILURE;
+	    }
+	  if (error_code == NO_ERROR)
```
airnet73: It is recommended to put a space or two for readability.

src/executables/csql.c
```diff
@@ -1460,6 +1464,205 @@ csql_do_session_cmd (char *line_read, CSQL_ARGUMENT * csql_arg)
...
+	    {
+	      csql_new_arg.user_name = strdup (user_name);
```
airnet73: If you use strdup() you should do FREE.
If you log in repeatedly, a memory leak may occur.
I can't find anywhere to memory free it.

src/executables/csql.c
eido5: 1. It keeps the unity with 1520 line `db_name += 1;`
2. I think it is difference from the expression.
3. Could you explain why you suggest this one?

src/executables/csql.c
eido5: As we knew that MAXLOGNAME is 16. So, we set the DB_NAME_LEN is 16.

Thank you. I will apply it.

src/executables/csql.c
eido5: I will apply it.

src/executables/csql.c
eido5: I will apply it.

src/executables/csql.c
eido5: @airnet73 
There is the complicated problem.

1. CSQL_ARGUMENT is initialized in the main function.
2. csql_arg.user_name is made by strdup.
3. when we de-allocate memory, if the address of memory is changed, we cannot de-allocate the memory.
4. CSQL_ARGUMENT is concatenated from the command.(ex) csql -u dba demodb)
5. So, when it is initialized, already the memory address is not fixed.
6. csql_launcher.c cannot be in the header file.
[reference](https://blog.naver.com/PostView.nhn?isHttpsRedirect=true&blogId=r2adne&logNo=120203848656)

src/executables/csql.c
airnet73: The syntax handling host_name starts at line 1483 (if (host_name != NULL)). I think handling host_name in this one if statement is good for readability.
The decision is yours.

msg/en_US/csql.msg
kisoo-han: ok! Go ahead!

src/executables/csql.c
```diff
@@ -1460,6 +1464,205 @@ csql_do_session_cmd (char *line_read, CSQL_ARGUMENT * csql_arg)
...
+		{
+		  csql_new_arg.passwd = (char *) NULL;	/* to fit into db_login protocol */
+		}
+	      else
+		{
+		  csql_new_arg.passwd = strdup (p);
+		}
+
+	      memcpy (csql_arg, &csql_new_arg, sizeof (CSQL_ARGUMENT));
+	      fprintf (csql_Output_fp, "CONNECTED : %s %s\n", csql_arg->user_name, csql_arg->db_name);
```
kisoo-han: I think
`fprintf (csql_Output_fp, "Connected.\n");`
is sufficient
or
`fprintf (csql_Output_fp, "Connected to '%s'.\n", csql_arg->db_name);`

msg/en_US/csql.msg
eido5: @kisoo-han 
Thank you!

src/executables/csql.c
```diff
@@ -1460,6 +1464,205 @@ csql_do_session_cmd (char *line_read, CSQL_ARGUMENT * csql_arg)
...
+		      au_sysadm_disable ();
+		    }
+		  csql_Database_connected = true;
+
+		  if (csql_arg->trigger_action_flag == false)
+		    {
+		      db_disable_trigger ();
+		    }
+
+		  fprintf (csql_Output_fp, "CONNECTED : %s %s\n", csql_arg->user_name, csql_arg->db_name);
```
kisoo-han: `fprintf (csql_Output_fp, "Connected.\n")`;

is sufficient.

src/executables/csql.c
```diff
@@ -1460,6 +1464,205 @@ csql_do_session_cmd (char *line_read, CSQL_ARGUMENT * csql_arg)
...
+		  au_sysadm_disable ();
+		}
+	      csql_Database_connected = true;
+
+	      if (csql_arg->trigger_action_flag == false)
+		{
+		  db_disable_trigger ();
+		}
+
+	      fprintf (csql_Output_fp, "CONNECTED : %s %s\n", csql_arg->user_name, csql_arg->db_name);
```
kisoo-han: fprintf (csql_Output_fp, "Connected.\n");
is sufficient

src/executables/csql.c
```diff
@@ -1460,6 +1464,205 @@ csql_do_session_cmd (char *line_read, CSQL_ARGUMENT * csql_arg)
...
+	    {
+	      p = getpass ((char *) csql_get_message (CSQL_PASSWD_PROMPT_TEXT));
+
+	      /* try again */
+	      if (db_restart_ex (UTIL_CSQL_NAME, db_name, user_name, p, NULL, db_get_client_type ()) != NO_ERROR)
+		{
+		  csql_Error_code = CSQL_ERR_SQL_ERROR;
+		  csql_check_server_down ();
+
+		  fprintf (csql_Output_fp, "SQL session is disconnected\n");
```
kisoo-han: fprintf (csql_Output_fp, "Warning: current CSQL session is disconnected.\n");

src/executables/csql.c
```diff
@@ -1460,6 +1464,205 @@ csql_do_session_cmd (char *line_read, CSQL_ARGUMENT * csql_arg)
 	}
 
       break;
+
+    case S_CMD_CONNECT:
+      /*dbname_cpy can be stored DB name+ @ + remote hostname + \0 */
```
kisoo-han: I think too many lines inside case, 194 lines inside single CASE. 
What about simpify the case itself by making function, like csql_set_plan_dump ()
For example,

```
case S_CMD_CONNECT:
  error_code = csql_connect ((argument[0] == '\0') ? NULL : argument);
  if (error_code != NO_ERROR)
    {
       return error_code;
    }

```... near the end ...
```
int
csql_connect (const char *arg_str)
{
 ...
}
```

File: src/executables/csql.c [hunk 12]
```diff
@@ -1460,6 +1464,205 @@ csql_do_session_cmd (char *line_read, CSQL_ARGUMENT * csql_arg)
...
+      user_name = strtok_r (username_cpy, delim, &save_ptr_strtok);
+      if (user_name == NULL)
+	{
+	  csql_Error_code = CSQL_ERR_CONNECT;
+	  return DO_CMD_FAILURE;
+
+	}
+
+      memset (&csql_new_arg, 0, sizeof (CSQL_ARGUMENT));
+      csql_arg_copy (&csql_new_arg, csql_arg);
```
kisoo-han: `memcpy (&csql_new_arg, csql_arg, sizeof (CSQL_ARGUMENT));`

And delete function **csql_arg_copy ()**

src/executables/csql.c
```diff
@@ -1460,6 +1464,205 @@ csql_do_session_cmd (char *line_read, CSQL_ARGUMENT * csql_arg)
 	}
 
       break;
+
+    case S_CMD_CONNECT:
+      /*dbname_cpy can be stored DB name+ @ + remote hostname + \0 */
+      char dbname_cpy[MAXLOGNAME + CUB_MAXHOSTNAMELEN + 2];
+      char username_cpy[DB_MAX_USER_LENGTH + 1];
+      char host_cpy[DB_MAX_USER_LENGTH + 1];
```
kisoo-han: I don't feel comfortable with variable names of style **username_cpy**. What about following
* without dash: array name
* with dash: pointer name

as char array
username [DB_MAX_USER_LENGTH + 1]
hostname [DB_MAX_USER_LENGTH + 1]
dbname [MAXLOGNAME + CUB_MAXHOSTNAMELEN + 2]

and as a char pointer (as you used)
char *user_name;
char *db_name;
char *host_name;

kisoo-han:
@eido5 ,
Please extend csql HELP for ';**CONnect**' session command.

src/executables/csql.c
```diff
@@ -1460,6 +1464,205 @@ csql_do_session_cmd (char *line_read, CSQL_ARGUMENT * csql_arg)
...
+      const char *delim = " @\n";
+      char *save_ptr_strtok = NULL;
+      char *db_name = NULL;
+      char *user_name = NULL;
+      char *host_name = NULL;
+      char *p = NULL;
+      CSQL_ARGUMENT csql_new_arg;
```
airnet73: How about using 3 variables (db_name, user_name, passwd) instead of csql_new_arg?
How about directly assigning the three variables to csql_arg after login?

src/executables/csql.c
```diff
@@ -1460,6 +1464,205 @@ csql_do_session_cmd (char *line_read, CSQL_ARGUMENT * csql_arg)
...
+	}
+
+      if (host_name != NULL)
+	{
+	  if (strcasecmp ("127.0.0.1", host_name) == 0 || strcasecmp ("localhost", host_name) == 0
+	      || strcasecmp (boot_get_host_name (), host_name) == 0 || strcasecmp (boot_get_ip (), host_name) == 0)
+	    {
+	      *(host_name - 1) = '\0';
+	    }
+	}
```
kisoo-han: @eido5 ,
Do you have any reaon that removing hostname for localhost?

src/executables/csql.c
eido5: It works, I will replace  csql_arg_copy() with memcpy(). Thank you.

src/executables/csql.c
eido5: Ok. thank you.

src/executables/csql.c
eido5: Ok. thank you.

src/executables/csql.c
eido5: Ok. thank you.

src/executables/csql.c
eido5: Ok. thank you.

src/executables/csql.c
eido5: Thank you. I will apply it.

src/executables/csql.c
```diff
@@ -1460,6 +1464,205 @@ csql_do_session_cmd (char *line_read, CSQL_ARGUMENT * csql_arg)
...
+	  if (csql_Database_connected)
+	    {
+	      csql_Database_connected = false;
+	      db_end_session ();
+	      db_shutdown ();
+	    }
+	  er_init ("./csql.err", ER_NEVER_EXIT);
+	  if (db_restart_ex (UTIL_CSQL_NAME, db_name, user_name, NULL, NULL, db_get_client_type ()) != NO_ERROR)
+	    {
+	      p = getpass ((char *) csql_get_message (CSQL_PASSWD_PROMPT_TEXT));
```
kisoo-han: You should check error code, and if error_code is not **ER_AU_IN VALID_PASSWORD** getpass is meaningless.
Even If you don't have running database server **testdb**, following session command will ask you the password.

`csql> ;connect public testdb`

src/executables/csql.c
eido5: @airnet73 
I decide this macro because digits (16) is not easy to figure out the reason.
But below macro is easier than 16.
#define DB_NAME_LEN MAXLOGNAME

src/executables/csql.c
eido5: To me mine is better to read. and even unity because demo version style was like yours but it is so hard to read.

src/executables/csql.c
eido5: - error case
;Connect public demodb@localhost

- if not remove,
call db_restart_ex()

src/executables/csql.c
eido5: Right. This is my mistake. thank you.

src/executables/csql.c
kisoo-han: Could you check it again?
What you mean, db_restart_ex () do not accept dbname@host style?

src/executables/csql.c
```diff
@@ -3330,3 +3351,254 @@ get_host_ip (unsigned char *ip_addr)
...
+	    {
+	      csql_new_arg.passwd = strdup (p);
+	    }
+
+	  FREE_MEM ((char *) csql_arg->user_name);
+	  FREE_MEM ((char *) csql_arg->db_name);
+	  FREE_MEM ((char *) csql_arg->passwd);
+
+	  memcpy (csql_arg, &csql_new_arg, sizeof (CSQL_ARGUMENT));
+	  fprintf (csql_Output_fp, "CONNECTED : %s %s\n", csql_arg->user_name, csql_arg->db_name);
```
eido5: @kisoo-han  @airnet73 @Srltas 
Could you ignore this function?
This is very useful to verify testcase.
I will replace this line with one below line before merged.

src/executables/csql.c
eido5: No, I mean when we login with locahost user like `;Connect public demodb@localhost`, function flow will go into the db_restart_ex () (line +3505)

src/executables/csql.c
```diff
@@ -1460,6 +1464,205 @@ csql_do_session_cmd (char *line_read, CSQL_ARGUMENT * csql_arg)
...
+	      db_name = strdup (csql_arg->db_name);
+	    }
+
+	  if (csql_Database_connected)
+	    {
+	      csql_Database_connected = false;
+	      db_end_session ();
+	      db_shutdown ();
+	    }
+	  er_init ("./csql.err", ER_NEVER_EXIT);
```
kisoo-han: is it required (er_init())?

src/executables/csql.c
```diff
@@ -1460,6 +1464,205 @@ csql_do_session_cmd (char *line_read, CSQL_ARGUMENT * csql_arg)
...
+	}
+
+      memset (&csql_new_arg, 0, sizeof (CSQL_ARGUMENT));
+      csql_arg_copy (&csql_new_arg, csql_arg);
+
+      /*connect to same DB user or not */
+      if ((db_name == NULL || strcasecmp (csql_arg->db_name, db_name) == 0) && csql_Database_connected == TRUE)
+	{
+
+	  if (!au_is_dba_group_member (Au_user))
```
kisoo-han: The new session command ';CONnect' does not require dba privileges. Is it required to check dba_group_member?
I think else-block is not required, and don't need to check for '**au_is_dba_group_member()'**. Let me know your intension.

src/executables/csql.c
eido5: I copied the code area of ;restart.
If we proof whether it is required or not, we would need analysis the db_restart() function.

How do you think?

src/executables/csql.c
eido5: Yes, dba group member do not need to ask password of a user when a user tries to login.
So, If I am a dba group user (ex : dba), there is necessary to ask no password of any other.

`dba -> not dba group user` => not need password
not vice versa.

src/executables/csql.c
kisoo-han: Users will be rather confused if the connection they try gives different results for local and remote. Personally I think it should yield the same result whether it is local or remote, even you could sustain current session for changing user on same database server. Whether or not the attempted connection fails or success, it would be good to clearly give the message that the current connection will be closed anyway.

Anyway, pls ask @mhoh3963 for it, I think it is a user requirements.

src/executables/csql.c
kisoo-han: So, is it mendatory? I don't think so. I think else-block is not required and don't need to check for dba_group_member. If a user is a dba_group_memner he willl pass **au_login ().** Most of the csql code checks the dba member because it needs dba privileges to execute requested actions.

src/executables/csql.c
kisoo-han: I tested it, and er_init () has no effect!
What is your exact intension of 
`er_init ("./csql.err", ER_NEVER_EXIT);`
it has no effect.

src/executables/csql.c
eido5: Right. I tested your suggestion and it works. 
I didn't know that checking dba group is for the dba privileges.
And session command is not for the dba privileges.

Thank you. I will apply it.

src/executables/csql.c
eido5: 1. I copied `er_init ("./csql.err", ER_NEVER_EXIT);` because almost it is used before db_restart_ex() is used.
2. I remember that it might affect implicit effects.

Can I verify it more? I think I should know why it is used in legacy code.

src/executables/csql.c
eido5: I understand what you wrote comment.
From the my point of view, localhost is not necessary to call db_restart_ex as view of cost.

**But** from the user's point of view, as your suggest, I could add session end with this code.
`      if (csql_Database_connected)
        {
          csql_Database_connected = false;
          db_end_session ();
          db_shutdown ();
        }
`
And user always take same results.
Ok. I will ask him about our opinion. Thank you.

src/executables/csql.c
airnet73: I suggest the below.
- Connected
- Connection success
- Connection was successful.

src/executables/csql.c
```diff
@@ -3330,3 +3351,246 @@ get_host_ip (unsigned char *ip_addr)
...
+	  csql_new_arg.db_name = strdup (db_name);
+	  csql_new_arg.passwd = (char *) NULL;
+
+	  FREE_MEM ((char *) csql_arg->user_name);
+	  FREE_MEM ((char *) csql_arg->db_name);
+	  FREE_MEM ((char *) csql_arg->passwd);
+
+	  memcpy (csql_arg, &csql_new_arg, sizeof (CSQL_ARGUMENT));
+
+	  if (csql_arg->sysadm && au_is_dba_group_member (Au_user))
```
airnet73: After connecting with the --sysadm option, is there no need to au_enable() when connecting to another DB as a general user?
I'm curious if there is a scenario like this.

eido5: Thank you for your comment.
This is copy from legacy code. So, I do not understand how it works in --sysadm option.

But for a few hours ago, our CTO and we decided to not to allow --sysadm in the CONNECT session command.

src/executables/csql.c
eido5: @airnet73 
"Connected." is already chosen. in 3492
did you mean that two messages are entries?

src/executables/csql.c
eido5: Because csql_arg.user_name is const * char, so we must use same structure and then copy it.

I tried what you suggest but it didn't work.

src/executables/csql.c
eido5: Could we  change ***user_name**;  as char ***user_name_ptr**?

kisoo-han: Sure, **user_name_ptr** style is good.

src/executables/csql.c
```diff
@@ -79,6 +83,14 @@
...
 #endif /* defined (SUPPRESS_STRLEN_WARNING) */
 
+#define DB_NAME_LEN MAXLOGNAME
+#define FREE_MEM(PTR)           \
+        do {                    \
+          if (PTR) {            \
+            free(PTR);          \
+          }                     \
+        } while (0)
+
```
kisoo-han: It is good to define FREE_MEM macro, but csql already uses macro '**free_and_init** ()'  with the same meaning (defined in base/memory_alloc.h). 
How about not defining FREE_MEM and just using the **free_and_init**() macro like any other csql code?

src/executables/csql.c
```diff
@@ -3330,3 +3351,254 @@ get_host_ip (unsigned char *ip_addr)
 
   return 0;
 }
+
+/*
+int
+csql_arg_copy (CSQL_ARGUMENT * dest, CSQL_ARGUMENT * src)
```
kisoo-han: Delete csql_arg_copy () function, including comment.

src/executables/csql.c
```diff
@@ -3330,3 +3351,254 @@ get_host_ip (unsigned char *ip_addr)
...
+  return 0;
+}
+*/
+static int
+csql_connect (char *argument, CSQL_ARGUMENT * csql_arg)
+{
+  /*dbname_cpy can be stored DB name+ @ + remote hostname + \0 */
+  char dbname_cpy[MAXLOGNAME + CUB_MAXHOSTNAMELEN + 2];
+  char username_cpy[DB_MAX_USER_LENGTH + 1];
+  char host_cpy[DB_MAX_USER_LENGTH + 1];
```
kisoo-han: Do not use varible name of style dbname_cpy [];

src/executables/csql.c
eido5: @kisoo-han 
Thank you so much!

src/executables/csql.c
eido5: I encounter the error when the PTR is set to NULL.
(I mean I also user free_and_init() and then error occured)

src/executables/csql.c
```diff
@@ -3330,3 +3358,161 @@ get_host_ip (unsigned char *ip_addr)
...
+  const char *delim = " @\n";
+  char *save_ptr_strtok = NULL;
+  char *db_name_ptr = NULL;
+  char *user_name_ptr = NULL;
+  char *host_name_ptr = NULL;
+  char *p = NULL;
+  int error_code;
+  CSQL_ARGUMENT csql_new_arg;
+
+  /*verify @hostname */
```
kisoo-han: Check whether argument is not NULL first, if omitted it could be coredump when user typed just
;connect without arguments (I think I give you the comment for this)
csql> ;connect

```
if (argument == NULL)
  {
    ...
  }
```

eido5: Right. It occured core dump. I will find any solution about this.
Thank you.
(error msg, function flow, etc)

src/executables/csql.c
```diff
@@ -3330,3 +3358,161 @@ get_host_ip (unsigned char *ip_addr)
...
+  host_name_ptr = strchr (argument, '@');
+  if (host_name_ptr != NULL)
+    {
+      if (*(host_name_ptr + 1) == '\0')
+	{
+	  host_name_ptr = NULL;
+	  csql_Error_code = CSQL_ERR_CONNECT;
+	  return DO_CMD_FAILURE;
+	}
+    }
```
kisoo-han: I think two ifs can be merged into one, and **host_name_ptr = NULL;** is not required.: 
 if (host_name_ptr != NULL && *(host_name_ptr + 1) == '\0')
  {
    csql_Error_code = CSQL_ERR_CONNECT;
    return DO_CMD_FAILURE;
   }

src/executables/csql.c
```diff
@@ -3330,3 +3358,161 @@ get_host_ip (unsigned char *ip_addr)
...
+  /*find db name following the user name */
+  db_name_ptr = strrchr (dbname, ' ');
+  if (db_name_ptr != NULL)
+    {
+      if (*(db_name_ptr + 1) == '\0')
+	{
+	  db_name_ptr = NULL;
+	  csql_Error_code = CSQL_ERR_CONNECT;
+	  return DO_CMD_FAILURE;
+	}
```
kisoo-han: Same with previous comment for hostname:

```
if (db_name_ptr != NULL && (db_name_ptr != NULL)
 {
   csql_Error_code = CSQL_ERR_CONNECT;
   return DO_CMD_FAILURE;
  }

db_name_ptr += 1;
```
...

src/executables/csql.c
```diff
@@ -3330,3 +3358,161 @@ get_host_ip (unsigned char *ip_addr)
...
+
+  user_name_ptr = strtok_r (username, delim, &save_ptr_strtok);
+  if (user_name_ptr == NULL)
+    {
+      csql_Error_code = CSQL_ERR_CONNECT;
+      return DO_CMD_FAILURE;
+    }
+
+  memset (&csql_new_arg, 0, sizeof (CSQL_ARGUMENT));
+  memcpy (&csql_new_arg, csql_arg, sizeof (CSQL_ARGUMENT));
```
kisoo-han: You don't need to **memset** (), **memcpy** () will overide it.

src/executables/csql.c
eido5: Thank you. It will be better.

src/executables/csql.c
eido5: Thank you. It will be better.

src/executables/csql.c
eido5: Right, I worried about any trash value (ex: 0x5348798378)

It is not necessary to use memset(). I will remove it thank you!

src/executables/csql.c
```diff
@@ -3352,98 +3359,54 @@ get_host_ip (unsigned char *ip_addr)
...
-  strcpy (dbname_cpy, argument);
-  strcpy (username_cpy, argument);
+  strcpy (dbname, argument);
+  strcpy (username, argument);
 
   /*find db name following the user name */
-  db_name = strrchr (dbname_cpy, ' ');
-  if (db_name != NULL)
+  db_name_ptr = strrchr (dbname, ' ');
+  if (db_name_ptr != NULL)
```
airnet73: How about processing db_name_ptr in one if statement?  

```
  if (db_name_ptr != NULL)
    {
      if (*(db_name_ptr + 1) == '\0')
	{
	  db_name_ptr = NULL;
	  csql_Error_code = CSQL_ERR_CONNECT;
	  return DO_CMD_FAILURE;
	}
      else
	{
	  db_name_ptr += 1;
	}
    }
  else
    {
      db_name_ptr = strdup (csql_arg->db_name);
    }
```

kisoo-han: Assignment to local variable db_name_ptr to NULL before return is not required.

```
  if (db_name_ptr != NULL)
    {
      if (*(db_name_ptr + 1) == '\0')
	{
	  csql_Error_code = CSQL_ERR_CONNECT;
	  return DO_CMD_FAILURE;
	}
      else
	{
	  db_name_ptr += 1;
	}
    }
  else
    {
      db_name_ptr = strdup (csql_arg->db_name);
    }

```

kisoo-han:
@eido5 ,
Resolve conflicts

**msg/ko_KR.euckr/csql.msg**

src/executables/csql.c
eido5: Thank you it will be better.

src/executables/csql.c
eido5: Thank you. It will be better.

src/executables/csql.c
```diff
@@ -3330,3 +3358,168 @@ get_host_ip (unsigned char *ip_addr)
...
+  memcpy (&csql_new_arg, csql_arg, sizeof (CSQL_ARGUMENT));
+
+#if defined(CS_MODE)
+  memset (boot_Host_connected, 0, sizeof (boot_Host_connected));
+#endif /* CS_MODE */
+
+  /*Failed to access other host or db and then access formal db_name */
+  if (db_name_ptr == NULL)
+    {
+      db_name_ptr = strdup (csql_arg->db_name);
```
airnet73: The memory of db_name_ptr needs to be freed.
However, be careful when freeing (see lines 3425 and 3393).
So, it seems necessary to consider whether memory should be allocated to db_name_ptr.

src/executables/csql.c
```diff
@@ -3330,3 +3358,168 @@ get_host_ip (unsigned char *ip_addr)
...
+
+      if (csql_arg->trigger_action_flag == false)
+	{
+	  db_disable_trigger ();
+	}
+
+//            fprintf (csql_Output_fp, "Connected.\n");
+      fprintf (csql_Output_fp, "CONNECTED : %s %s\n", csql_arg->user_name, csql_arg->db_name);
+    }
+
```
kisoo-han, line 3487: Line 3438 - 3523 contains same codes except csql_new_arg.passwd is NULL or not.
So, how about following code (this is just example):

```
er_init ("./csql.err", ER_NEVER_EXIT);
csql_new_arg.passwd = (char *) NULL;

 if (db_restart_ex (UTIL_CSQL_NAME, db_name_ptr, user_name_ptr, NULL, NULL, db_get_client_type ()) != NO_ERROR)
    {
      if (csql_Is_interactive && db_error_code () == ER_AU_INVALID_PASSWORD)
        {
          p = getpass ((char *) csql_get_message (CSQL_PASSWD_PROMPT_TEXT));

          /* try again */
          if (db_restart_ex (UTIL_CSQL_NAME, db_name_ptr, user_name_ptr, p, NULL, db_get_client_type ()) != NO_ERROR)
            {
              csql_Error_code = CSQL_ERR_SQL_ERROR;
              csql_check_server_down ();

              return DO_CMD_FAILURE;
            }
          else
            {
              /*login success with password */
              csql_new_arg.user_name = strdup (user_name_ptr);
              csql_new_arg.db_name = strdup (db_name_ptr);
            }

          csql_new_arg.passwd = (p[0] == NULL) ? (char *) NULL : strdup (p);
        }
    }

 csql_new_arg.user_name = strdup (user_name_ptr);
 csql_new_arg.db_name = strdup (db_name_ptr);

 FREE_MEM ((char *) csql_arg->user_name);
 FREE_MEM ((char *) csql_arg->db_name);
 FREE_MEM ((char *) csql_arg->passwd);

 memcpy (csql_arg, &csql_new_arg, sizeof (CSQL_ARGUMENT));
 if (csql_arg->sysadm && au_is_dba_group_member (Au_user))
   {
     au_sysadm_disable ();
   }
 csql_Database_connected = true;

 if (csql_arg->sysadm && au_is_dba_group_member (Au_user))
   {
     au_sysadm_disable ();
   }
 csql_Database_connected = true;

 if (csql_arg->trigger_action_flag == false)
   {
     db_disable_trigger ();
   }

//            fprintf (csql_Output_fp, "Connected.\n");
  fprintf (csql_Output_fp, "CONNECTED : %s %s\n", csql_arg->user_name, csql_arg->db_name);

return DO_CMD_SUCCESS;
```

src/executables/csql.c
eido5: I think 
```suggestion
      strcpy (db_name_ptr, csql_arg->db_name);
```
 And then, there will be necessary to user() free.
 @airnet73 How do you think about this?

src/executables/csql.c
eido5: Thank you. I will apply it.

airnet73: After executing the ;connect command, only user_name, db_name, and passwd are changed. The code below is the same as before. So I think it's unnecessary code.

```
 if (csql_arg->sysadm && au_is_dba_group_member (Au_user))
   {
     au_sysadm_disable ();
   }

 if (csql_arg->trigger_action_flag == false)
   {
     db_disable_trigger ();
   }
```

eido5: @airnet73 
At first, this is from the legacy code. 

And I analyzed a little bit.
I think "Au_disable" can be removed till 'csql_arg->sysadm' is not allowed in the ;connect.

But db_disable_trigger() is less related with whether trigger_action_flag is on/off.
(recommend to be called after db_restart is called is written in the function description.)

Do we need more analyze about this function?

src/executables/csql.c
airnet73: db_name_ptr is of char * type. Also, strcpy cannot be performed because db_name_ptr is null.

src/executables/csql.h
```diff
@@ -171,7 +173,8 @@ extern "C"
     CSQL_ERR_CANT_EDIT,
     CSQL_ERR_INFO_CMD_HELP,
     CSQL_ERR_CLASS_NAME_MISSED,
-    CSQL_ERR_FORMAT
+    CSQL_ERR_FORMAT,
+    CSQL_ERR_CONNECT
```
kisoo-han: error msg, **CSQL_ERR_CONNECT** is defined but not used.

src/executables/csql.c
```diff
@@ -1460,6 +1473,21 @@ csql_do_session_cmd (char *line_read, CSQL_ARGUMENT * csql_arg)
...
 
       break;
+
+    case S_CMD_CONNECT:
+      if (csql_arg->sysadm != true)
+	{
+	  error_code = csql_connect ((argument[0] == '\0') ? NULL : argument, csql_arg);
+	  if (error_code != NO_ERROR)
+	    {
+	      return error_code;
```
kisoo-han, line 1485: I think it's better to give warning to user only when ';**connect**' is failed.

eido5: I think so. Thank you!

src/executables/csql.c
eido5: Right, I will de-allocate this variable. Thank you.

src/executables/csql.h
eido5: Ok. It will be removed thank you!

src/executables/csql.h
```diff
@@ -374,6 +377,8 @@ extern "C"
   extern char *string_to_string (const char *string_value, char string_delimiter, char string_introducer, int length,
 				 int *result_length, bool plain_string, bool change_single_quote);
 
+  extern int csql_arg_copy (CSQL_ARGUMENT * src, CSQL_ARGUMENT * dest);
```
airnet73: Please remove unused codes

eido5: Thank you! I will apply it.

src/executables/csql.c
```diff
@@ -3330,3 +3358,131 @@ get_host_ip (unsigned char *ip_addr)
...
+  /*dbname can be stored DB name+ @ + remote hostname + \0 */
+  char buf[DB_MAX_USER_LENGTH + MAXLOGNAME + CUB_MAXHOSTNAMELEN + 1];
+  const char *delim = " \n";
+  char *save_ptr_strtok = NULL;
+  const char *db_name_ptr = NULL;
+  char *user_name_ptr = NULL;
+  char *host_name_ptr = NULL;
+  char *p = NULL;
+  const char *err_msg;
+  int error_code;
```
airnet73: Please remove unused codes (error_code)

src/executables/csql.c
```diff
@@ -3330,3 +3358,131 @@ get_host_ip (unsigned char *ip_addr)
...
+    }
+
+  /*verify @hostname */
+  if ((host_name_ptr = strchr (argument, '@')) != NULL && (*(host_name_ptr + 1) == '\0'))
+    {
+      csql_Error_code = CSQL_ERR_SQL_ERROR;
+      return DO_CMD_FAILURE;
+    }
+
+  strcpy (buf, argument);
```
airnet73: It is recommended to use strncpy() to prevent buffer overflow.

src/executables/csql.c
```diff
@@ -3330,3 +3358,131 @@ get_host_ip (unsigned char *ip_addr)
...
+  db_name_ptr = strtok_r (NULL, delim, &save_ptr_strtok);
+  if (db_name_ptr == NULL)
+    {
+      db_name_ptr = csql_arg->db_name;
+    }
+
+  memcpy (&csql_new_arg, csql_arg, sizeof (CSQL_ARGUMENT));
+
+#if defined(CS_MODE)
+  memset (boot_Host_connected, 0, sizeof (boot_Host_connected));
```
airnet73, line 3415: Is this code necessary?

eido5: Yes, If not, It cannot works to connect other hosts.

src/executables/csql.c
eido5: Thank you. I will remove it.

src/executables/csql.c
eido5: Right. It will be better. Thank you.

src/executables/csql.c
```diff
@@ -3330,3 +3358,129 @@ get_host_ip (unsigned char *ip_addr)
...
+    }
+
+  /*verify @hostname */
+  if ((host_name_ptr = strchr (argument, '@')) != NULL && (*(host_name_ptr + 1) == '\0'))
+    {
+      csql_Error_code = CSQL_ERR_SQL_ERROR;
+      return DO_CMD_FAILURE;
+    }
+
+  strncpy (buf, argument, sizeof (argument));
```
airnet73: Did you test the code after fixing it?
This is the code with the error.

eido5: Right. I will fix it. Thank you.

src/executables/csql.c
```diff
@@ -3330,3 +3358,129 @@ get_host_ip (unsigned char *ip_addr)
 
   return 0;
 }
+
+static int
+csql_connect (char *argument, CSQL_ARGUMENT * csql_arg)
+{
+  /*dbname can be stored DB name+ @ + remote hostname + \0 */
+  char buf[DB_MAX_USER_LENGTH + MAXLOGNAME + CUB_MAXHOSTNAMELEN + 1];
```
kisoo-han: what about:
`  char buf[DB_MAX_USER_LENGTH + MAXLOGNAME + CUB_MAXHOSTNAMELEN + 1] = { 0, };`

src/executables/csql.c
```diff
@@ -3330,3 +3358,129 @@ get_host_ip (unsigned char *ip_addr)
...
+      return DO_CMD_SUCCESS;
+    }
+
+  /*verify @hostname */
+  if ((host_name_ptr = strchr (argument, '@')) != NULL && (*(host_name_ptr + 1) == '\0'))
+    {
+      csql_Error_code = CSQL_ERR_SQL_ERROR;
+      return DO_CMD_FAILURE;
+    }
+
```
kisoo-han: I recommend delete these lines.

src/executables/csql.c
kisoo-han: I think
`  strcpy (buf, argument);`

kisoo-han: @airnet73 recommend,

`strncpy (buf, argument, sizeof(buf) - 1);`

kisoo-han (approved):
LGTM,
Thanks!!

airnet73 (approved):

msg/en_US/csql.msg
```diff
@@ -92,6 +92,7 @@ $ These 3 should be blank
 81 sysadm>
 82 %1$d transactions %2$s.
 83 %1$d transaction %2$s.
+84 Too few arguments given.\n
```
eido5, line 95: It will be added as other language soon.

Srltas (approved):

src/executables/csql.c
```diff
@@ -1484,7 +1484,7 @@ csql_do_session_cmd (char *line_read, CSQL_ARGUMENT * csql_arg)
 	}
       else
 	{
-	  fprintf (csql_Output_fp, "CONNECT session command cannot support --sysadm mode\n");
+	  fprintf (csql_Output_fp, "CONNECT session command cannot support --SA-mode and --sysadm mode\n");
```
kisoo-han: CONNECT session command **does not support** --SA-mode and --sysadm mode

I propose a classification of error messages, for example:
```
          fprintf (csql_Output_fp, "CONNECT session command does not support %s %s.\n",
                csql_arg->sa_mode ? "--SA-mode" : "", csql_arg->sysadm ? "(--sysadm mode)" : "");
 
```

eido5: It is better to apply each case.(sa and sysadm)
I will apply it thank you.

kisoo-han:
Please append following line after line#581 of csql.c

ASIS
`581           fputs ("!", csql_Output_fp);`
TOBE
```
581           fputs ("!", csql_Output_fp);
582           fflush (csql_Output_fp);
```

kisoo-han:
We need to completely hand over screen control to libedit for csql. If the database is offline, readline should be called by changing the prompt to '!csql> ' instead of outputting '!' and passing it to readline with prompt "csql> ".

Please refer following diff as an example:
==============================================


```
diff --git a/src/executables/csql.c b/src/executables/csql.c
index 8dca6862f..a3e847013 100644
--- a/src/executables/csql.c
+++ b/src/executables/csql.c
@@ -161,6 +161,7 @@ char csql_Scratch_text[SCRATCH_TEXT_LEN];
 int csql_Error_code = NO_ERROR;

 static char csql_Prompt[100];
+static char csql_Prompt_offline[100];
 static char csql_Name[100];

 /*
@@ -575,12 +576,7 @@ start_csql (CSQL_ARGUMENT * csql_arg)

   for (line_no = 1;; line_no++)
     {
-      if (db_Connect_status != DB_CONNECTION_STATUS_CONNECTED)
-       {
-         csql_Database_connected = false;
-         fputs ("!", csql_Output_fp);
-       }
-
+      char *prompt = db_Connect_status == DB_CONNECTION_STATUS_CONNECTED ? csql_Prompt : csql_Prompt_offline;
       read_whole_line = false;

       memset (line_buf, 0, LINE_BUFFER_SIZE);
@@ -592,7 +588,7 @@ start_csql (CSQL_ARGUMENT * csql_arg)
          fputs (csql_Prompt, csql_Output_fp);  /* display prompt */
          line_read = fgets ((char *) line_buf, LINE_BUFFER_SIZE, csql_Input_fp);
 #else
-         if ((line_read = readline (csql_Prompt)) != NULL)
+         if ((line_read = readline (prompt)) != NULL)
            {
              if (line_read_alloced != NULL)
                {
@@ -2800,6 +2796,7 @@ csql (const char *argv0, CSQL_ARGUMENT * csql_arg)
     {
       strncat (csql_Prompt, " ", avail_size);
     }
+  snprintf (csql_Prompt_offline, sizeof (csql_Prompt_offline), "!%s", csql_Prompt);
   strncpy_bufsize (csql_Name, csql_get_message (CSQL_NAME));

   /* as we must use db_open_file_name() to open the input file, it is necessary to be opening csql_Input_fp at this

```

eido5:
SA mode handling. 
Only using by username can login.

src/executables/csql.c
```diff
@@ -1460,6 +1475,21 @@ csql_do_session_cmd (char *line_read, CSQL_ARGUMENT * csql_arg)
...
+	  error_code = csql_connect ((argument[0] == '\0') ? NULL : argument, csql_arg);
+	  if (error_code != NO_ERROR)
+	    {
+	      return error_code;
+	    }
+	  break;
+	}
+      else
+	{
+	  fprintf (csql_Output_fp, "CONNECT session command cannot supports --sysadm mode\n");
```
kisoo-han: cannot supports ==> "does not support"

src/executables/csql.c
```diff
@@ -3331,3 +3362,133 @@ get_host_ip (unsigned char *ip_addr)
...
+  db_name_ptr = strtok_r (NULL, delim, &save_ptr_strtok);
+  if (db_name_ptr == NULL)
+    {
+      db_name_ptr = csql_arg->db_name;
+    }
+  else
+    {
+      if (csql_arg->sa_mode == true)
+	{
+	  fprintf (csql_Output_fp, "CONNECT session command allows the login with only user name in the --SA-mode.\n");
```
kisoo-han: Just opinion,
`fprintf (csql_Output_fp, "Cannot connect to another DB in the SA mode.\n");`

src/executables/csql.c
eido5: Ok. I will apply it. thank you.

src/executables/csql.c
eido5: Thank you. 
I will apply as "Cannot connect to other DB in the SA mode."

kisoo-han (approved):
Now, everything is ok.
Thank you for your efforts!

PR #4135, opened 2023-02-23
[LETS-639] Shutdown a page server gracefully

wrlawodms:
http://jira.cubrid.org/browse/LETS-639

Please see the protocol described in the Jira issue and Implementation section for what's done in the code.

**Details**
**_Page Server_**
- Now that the disconnection only starts from TS and PS just requests it to disconnect, all disconnections are handled in the request handler. In disconnect_all_tran_servers(), now it just requests and waits until all are done using m_conn_cv
- ~connection_handler() accesses page_server::m_responder which is released in ps_Gl->finalize_request_responder(). So, it has to make sure all connections are destroyed before calling finalize_request_responder(). 
  -  For this, async_disconnect_handler::terminate() now joins its thread so make sure all connection_handlers are destroyed. It's called at the end of disconnect_all_tran_servers(). The dtor of it just makes sure it's been terminated beforehand.
- Reorder some resource-releasing jobs when PS shuts down considering their dependency.

**_Tran Server_**
- Use async_disconnect_handler as page_server does to destroy a connection_handler in its request_handler.
  - The ~connection_handler() joins the request handling thread internally, and if the request handler destroys the connection_handler itself, it ends up trying to join itself and a deadlock.
- The `shared_mutex` is used for m_page_server_conn_vec. The communication functions like push_request() have to be exclusive with disconnection, but they can access the connection concurrently.
- ~connection_handler(): send the SEND_DISCONNECT_MSG to a PS, stop all internal thread, and join them.
- Remove the connection-specified push_reuqest and send_receive. The upper layer doesn't need to know the connection's index and it makes it hard to manage them.


src/server/active_tran_server.hpp
```diff
@@ -33,7 +33,7 @@ class active_tran_server : public tran_server
     }
 
     bool uses_remote_storage () const final override;
-    MVCCID get_oldest_active_mvccid_from_page_server () const;
+    MVCCID get_oldest_active_mvccid_from_page_server ();
```
cristiarg, line 37: Nit.
Function's `const` signature could have been preserved by making
```
mutable std::shared_mutex m_page_server_conn_vec_mtx
```

but I think we would run into other complications later anyway; so it's ok

src/server/tran_server.cpp
```diff
@@ -137,30 +138,40 @@ tran_server::boot (const char *db_name)
...
-			   std::string &payload_out) const
-{
-  assert (idx < m_page_server_conn_vec.size());
-  return m_page_server_conn_vec[idx]->send_receive (reqid, std::move (payload_in), payload_out);
-}
+  {
+    std::shared_lock<std::shared_mutex> s_lock (m_page_server_conn_vec_mtx);
+    if (m_page_server_conn_vec.empty())
+      {
+	return; // All connections have been disconnected already
```
cristiarg: I would argue that we need some sort of mitigation here.
What happens on an ATS if its single/last PS goes down? It should go down as well.
If this is not for this patch, at least add an `assert (false)` for now.

src/server/tran_server.cpp
```diff
@@ -386,14 +399,55 @@ tran_server::connection_handler::connection_handler (cubcomm::channel &&chn, tra
...
+				    std::placeholders::_1);
+
+  handlers_map.insert (std::make_pair (page_to_tran_request::SEND_DISCONNECT_REQUEST_MSG, disconnect_request_handler));
+
   return handlers_map;
 }
+void
+tran_server::connection_handler::receive_disconnect_request (page_server_conn_t::sequenced_payload &a_ip)
+{
+  std::shared_ptr<connection_handler> conn_sptr;
```
cristiarg: Chronologically, I have to admit that from here on I don't understand.

The thing that triggered my skepticism is the complication introduced by `shared_ptr` usage. Which does not address the aspect of waiting for a `connection_handler` to finish processing en-route requests.

By removing the connection handler from the list - guarded by an exclusive mutex - new requests cannot be sent any further - so we're good up to here.
But en-route requests are not waited for.
This is not the fault of the current patch. This was missing even before.
There is something but it is only halfway done (the connection handler has an instance of `request_sync_client_server` which, in turn, has an instance of `response_broker`; on the destruction path, the broker has a `terminate` method but it is not blocking; nor does it wait for the requests to be fully services - it just aborts - see implementation of `response_broker::bucket::get_response`; the connection handler can be made to wait for its underlying requests to be answered if this `terminate` method from the broker is made blocking - even if, as it is now, it only aborts the existing requests instead of actually fulfilling them).

Coming back to the `shared_ptr` can you please elaborate why it is needed?

```
While disconnecting a connection by m_async_disconnect_handler, a connection can be accessed to communicate in push_request and send_receive. To prevent them from accessing a destroyed connection, the shared_ptr is used.
```
This is not correct, IMHO. If the connection handler is removed from `m_page_server_conn_vec`, then the functions `tran_server::push_request` and `tran_server::send_receive` can't get to that connection handler anymore.

Thus, the way I see the chronology:
- remove connection handler from the list - already does
- wait for the connection handler to service existing requests - by making `response_broker<T_PAYLOAD, T_ERROR>::terminate` blocking and surfacing it as explained above
- then the connection handler can be moved to the async disconnect handler which will only handle the tearing down of the outgoing/incoming connection threads - as already `connection_handler::~connection_handler` does.

Please let me know where I got things wrong.

cristiarg (commented):
Partial review.

hornetmj (approved):

src/server/async_disconnect_handler.cpp
```diff
@@ -18,37 +18,31 @@
...
-    {
-      m_thread.join ();
-    }
-  else
-    {
-      assert (m_terminate.load ());
-    }
-
-  assert (m_disconnect_queue.empty ());
+  // Call terminate() before the resources that can be aceessed during destroying T_CONN_HANDLER_PTR is released. For example, page_server::~connection_handler accesses page_server::m_responder, which is released during shutdown.
```
joohok:
```suggestion
  // Call terminate() before the resources that can be accessed during destroying T_CONN_HANDLER_PTR are released. For example, page_server::~connection_handler accesses page_server::m_responder, which is released during shutdown.
```
typo

joohok (commented):
~~Are `disconnect_tran_server_async ()` and `disconnect_tran_server_async ()` not required to be changed ?~~

Sorry, those functions are called only when the connection is disconnected abnormally.

src/server/tran_server.cpp
```diff
@@ -341,13 +342,18 @@ tran_server::disconnect_all_page_servers ()
...
   assert_is_tran_server ();
 
-  for (const auto &conn : m_page_server_conn_vec)
-    {
-      er_log_debug (ARG_FILE_LINE, "Transaction server disconnected from page server with channel id: %s.\n",
-		    conn->get_channel_id ().c_str ());
-      conn->disconnect ();
-    }
-  m_page_server_conn_vec.clear ();
+  // We assumes that no threads are waiting for response at this moment so that m_conn->stop_response_broker() is not needed.
```
wrlawodms, line 347: Can we really assume it? I assume it because all outgoing msg generators stop and all transactions abort before disconnecting connections (`ts_Gl->stop_outgoing_page_server_messages` and `log_abort_all_active_transaction()`) . However, I'm not sure they have a way out from blocking on `send_receive()`.

hornetmj (approved):

src/server/page_server.cpp
```diff
@@ -525,14 +542,41 @@ page_server::disconnect_all_tran_server ()
...
+	  m_passive_tran_server_conn[i]->push_disconnection_request ();
 	}
-      m_passive_tran_server_conn.clear ();
     }
+
+  er_log_debug (ARG_FILE_LINE,
+		"disconnect_all_tran_server: Wait until all connections are disconnected.\n");
+
+  /*
+   *  m_active_tran_server_conn == nullptr : The ATS has been disconnected or disconncted in progress (by async_disconnect_handler)
```
```suggestion
   *  m_active_tran_server_conn == nullptr : The ATS has been disconnected or disconnect in progress (by async_disconnect_handler)
```

src/server/page_server.cpp
```diff
@@ -525,14 +542,41 @@ page_server::disconnect_all_tran_server ()
...
+   *  If some disconnections are underway, they will be waited for at the m_async_disconnect_handler.terminate () below.
+   */
+  {
+    auto no_conn_alive_pred = [this]
+    {
+      return m_active_tran_server_conn == nullptr && m_passive_tran_server_conn.empty();
+    };
+
+    std::unique_lock ulock { m_conn_mutex };
+    m_conn_cv.wait (ulock, no_conn_alive_pred);
```
cristiarg: I think it's better to `wait_for` here with a small time delay.

cristiarg: However, here we have the problem that - while the PS is in the process of wanting to disconnect from the cluster, another [passive] TS can request to connect to it.

One way to solve this is with a boolean flag, also guarded by `m_conn_mutex` which will prevent `set_active_tran_server_connection` and `set_passive_tran_server_connection` from registering new connections. It is better to take this in a separate issue because an extension is also needed further down the line - `css_process_server_server_connect` - and for the peer server to recognize that the connection failed - ``tran_server::connect_to_page_server`.

src/server/page_server.cpp
```diff
@@ -45,8 +45,7 @@ page_server::~page_server ()
   // when/if needed connections can be moved here to the disconnect handler and that one waited for
   assert (m_active_tran_server_conn == nullptr);
   assert (m_passive_tran_server_conn.size () == 0);
-
-  m_async_disconnect_handler.terminate ();
+  assert (m_async_disconnect_handler.is_terminated());
```
cristiarg: Assert not needed.
There already is an assert in `async_dysconnect_handler` with the same end effect
I suppose `is_terminated` is not needed either

src/server/tran_server.cpp
```diff
@@ -384,15 +390,55 @@ tran_server::connection_handler::connection_handler (cubcomm::channel &&chn, tra
...
+
+  handlers_map.insert (std::make_pair (page_to_tran_request::SEND_DISCONNECT_REQUEST_MSG, disconnect_request_handler));
+
   return handlers_map;
 }
 
+void
+tran_server::connection_handler::receive_disconnect_request (page_server_conn_t::sequenced_payload &a_ip)
+{
+  m_conn->stop_response_broker (); // wake up threads waiting for a response and tell them it won't be served.
```
cristiarg, line 442: (1) - see related remarks

I am ok with this.

src/communication/request_sync_client_server.hpp
```diff
@@ -184,7 +193,7 @@ namespace cubcomm
 
     // if the receiving thread terminated, there will not be any more responses
     // unblock all waiting client thread - they will receive errors
-    m_response_broker.notify_terminate_and_wait ();
+    stop_response_broker ();
```
cristiarg, line 196: (1)
Since we're doing this outside explicitly, maybe we should remove it from here.
And remove the `if (m_terminate)` check in `response_broker::notify_terminate_and_wait`

src/communication/response_broker.hpp
```diff
@@ -216,7 +216,11 @@ namespace cubcomm
     {
       std::lock_guard<std::mutex> lk_guard (m_mutex);
 
-      assert (!m_terminate);
+      if (m_terminate)
```
cristiarg, line 219: By my understanding, this check goes hand-in-hand with the following line below:

```
      // NOTE: when terminate is invoked there is the possibility to implement a configurable behaviour:
      //  - either abort and return error - as done below  [[1]]
      //  - or continue waiting - with a timeout - until the request is serviced by the peer server [[2]]

      // still under lock, erase the entry from the container to validate termination condition
      m_response_payloads.erase (found_it);

```
Currently the broker does [[1]].
When requested to terminate, the broker immediately terminates, returns errors and - while doing this - some requests might still be "in flight".
And these newly added checks are needed.

If the broker were modified to behave like [[2]], this check would not be needed because all "in-flight" requests are waited for.

Thus, I would advise adding a comment explaining this.

src/server/tran_server.cpp
```diff
@@ -384,15 +390,55 @@ tran_server::connection_handler::connection_handler (cubcomm::channel &&chn, tra
...
+      {
+	if (it->get () == this)
+	  {
+	    m_ts.m_async_disconnect_handler.disconnect (std::move (*it));
+	    assert (*it == nullptr);
+	    conn_vec.erase (it);
+	    break;
+	  }
+      }
+  }
```
This should be a function on `tran_server` - `m_ts.disconnect_me(this)` - since it works will all the objects that are member of `tran_server`.
```suggestion
  {
    std::lock_guard<std::shared_mutex> lk_guard (m_ts.m_page_server_conn_vec_mtx);
    auto &conn_vec = m_ts.m_page_server_conn_vec;

    auto it = conn_vec.begin ();
    for (; it != conn_vec.end (); it++)
      {
	if (it->get () == this)
	  {
	    m_ts.m_async_disconnect_handler.disconnect (std::move (*it));
	    assert (*it == nullptr);
	    conn_vec.erase (it);
	    break;
	  }
      }
  }
```

src/communication/response_broker.hpp
```diff
@@ -236,7 +240,11 @@ namespace cubcomm
     {
       std::lock_guard<std::mutex> lockg (m_mutex);
 
-      assert (!m_terminate);
+      if (m_terminate)
```
cristiarg, line 244: Same comment as for `if (m_terminate)` check in `register_response`.

src/communication/response_broker.hpp
```diff
@@ -313,6 +321,10 @@ namespace cubcomm
   {
     {
       std::lock_guard<std::mutex> lockg (m_mutex);
+      if (m_terminate)
```
cristiarg, line 326: (1)

src/server/tran_server.cpp
cristiarg: Indeed, they don't.
I think we will have to add some timeouts in a few places.
I did a simple thing with this patch - started ATS, PS, PTS1, PTS2 and the `cubrid server stop` PS. The result: PTS's disconnected cleanly (they were idle anyway). PTS also disconnected cleanly - it seemed to me - and upon the first connection from a client, it crashed, as expected. However, PS hanged because there was still some message in one of its outgoing connection - in `wait_not_empty_and_send_all`.

src/communication/request_sync_client_server.hpp
wrlawodms: `disconnect_all_page_servers` doesn't call the `stop_response_broker()` explicitly in advance, so there is a case it's needed. Likewise, `notify_terminate_and_wait` also needs that check.

wrlawodms: If we call `disconnect_all_page_servers()` in the `disconnect_all_page_servers`, which I tried actually, we need to guard it with `m_page_server_conn_vec_mtx` since it iterates `conn_vec`. However, a thread can sleep with the mutex acquired in `send_receive`, which end up deadlock. We should come up with a way to deal with it if we want to call `stop_response_broker()` in `disconnect_all_page_servers`, but it's not necessary as I commented. So, I've chosen this way.

src/communication/response_broker.hpp
wrlawodms: Ok, I've added.

src/server/page_server.cpp
wrlawodms: done. :)

src/server/page_server.cpp
wrlawodms: > I think it's better to wait_for here with a small time delay.

I'm not against it, but I'd like to know the advantage of `wait for` instead of `wait`, which I asked before but not replied. Is it for a future extension like timeout? or, does it have another advantage itself?

wrlawodms: > However, here we have the problem that - while the PS is in the process of wanting to disconnect from the cluster, another [passive] TS can request to connect to it.

You're right. Thank you for pointing out. Another way would be to stop `css_master_thread()` before this function, or prevent a new connection is created in the function with a flag at least. Do you want me to create the issue? I think it can be done after discussing new regression tests.

src/server/tran_server.cpp
wrlawodms: > I think we will have to add some timeouts in a few places.

I'll create the issue if you don't mind, or you do it if you have a more detailed image.

> it crashed, as expected

You mean the crash because there is no connection in ATS by the added assertion, right? I ran into it, but I thought it's just undefined behavior. But, yes, the PS should get out of the hang at least. I'll look into it.

wrlawodms: > However, PS hung because there was still some message in one of its outgoing connection - in wait_not_empty_and_send_all.

**Short story:**
Having looked into it, I've discovered a race condition during disconnection. So, I've widened the lock scope, and the hang doesn't occur: https://github.com/CUBRID/cubrid/pull/4135/commits/5f5c6c2d2c0d1dab6d6e9a8b7958aba7aef2cae0.

**Long story:**
It didn't always happen. When I ran into hanged PS with your configuration, there was one connection left in `m_passive_tran_server_conn` so the wait() in `disconnect_all_tran_servers` couldn't exit.

The race condition scenario:
1. PS: send DISCONNECT_REQ_MSG to PTS1 iterating the connection_handler vector
2. PTS1:  receive the msg and reply with DISCONNECT_REQ_MSG to PS
3. PS: remove the connection_handler for PTS1 from the vector and push it to the asyns_disconnection_handler. (by the request handler)
4. PS: send  DISCONNECT_REQ_MSG to PTS2 iterating the connection_handler vector
5. PTS2:  receive the msg and reply with DISCONNECT_REQ_MSG to PS
6. PS: remove the connection_handler for PTS1 from the vector and push it to the asyns_disconnection_handler. (by the request handler)

There is a race condition between (3) and (4). While it's being iterated, the vector can be modified. 
However, I doubt it can happen in practice since the room for the race condition is too small. It's just a timeframe from after sending DISC_MSG and going to the next iterator. I wanted to make sure it's the cause by leaving a debug log, but then it didn't happen. By the way, definitely there is a race condition, so I prevent it and the hang doesn't occur now. 

Q. 
> However, PS hanged because there was still some message in one of its outgoing connection - in wait_not_empty_and_send_all.

Did you see the hang ocuured by `wait_not_empty_and_send_all()`? Where I see is in `disconnect_all_tran_servers()` as described above.

src/server/tran_server.cpp
wrlawodms: Thank you for the pointer. https://github.com/CUBRID/cubrid/pull/4135/commits/f24b2e1ccb42f59c7ffb869482735b33a5ba6941

src/server/page_server.cpp
cristiarg: > I'm not against it, but I'd like to know the advantage of wait for instead of wait, which I asked before but not replied. Is it for a future extension like timeout? or, does it have another advantage itself?

No, no advantage other than being able to add a timeout afterwards.
In my opinion, I also think that `wait` is the more conservative to actually use but I saw that `wait_for` is used a lot in the codebase.
So, in truth, I do not have a good reason to push for `wait_for` other than being able to easily add a timeout in the future. But, for now, we don't need that timeout so - `wait` it is.

cristiarg: @wrlawodms 

> You're right. Thank you for pointing out. Another way would be to stop css_master_thread() before this function, or prevent a new connection is created in the function with a flag at least. Do you want me to create the issue? I think it can be done after discussing new regression tests.

yes, separate issue

src/communication/request_sync_client_server.hpp
cristiarg: I think you meant:

> If we call **->stop_response_broker()** in the disconnect_all_page_servers, which I tried actually, we need to

Yes. Fair enough. I wanted for it to be a bit simpler. But, we will see once we have more experience with this new protocol or we see that something does not work.

cristiarg (approved):
There is still a blocking for this issue - http://jira.cubrid.org/browse/LETS-678.

joohok (approved):

src/server/page_server.cpp
wrlawodms: added: http://jira.cubrid.org/browse/LETS-679

src/server/tran_server.cpp
wrlawodms: > Can we really assume it? I assume it because all outgoing msg generators stop and all transactions abort before disconnecting connections (`ts_Gl->stop_outgoing_page_server_messages` and `log_abort_all_active_transaction()`) . However, I'm not sure they have a way out from blocking on `send_receive()`.

http://jira.cubrid.org/browse/LETS-687

PR #4136, opened 2023-02-23
[CBRD-24633] Delete owner information from schema, index, trigger, and object files of general users.

airnet73:
http://jira.cubrid.org/browse/CBRD-24633

Purpose
If there is owner information in the schema file, an error may occur when moving data from user_a to user_b. In this case, it is necessary to delete the owner information or change the onwer information in the schema file. In order to remove this inconvenience, it is necessary to delete the owner information.

Implementation
* Change File
Schema file, Index file, Trigger file, Object file

AS_IS
> CREATE CLASS **[user2].**[test_t] REUSE_OID, COLLATE utf8_bin;
> ALTER CLASS **[user2].**[test_t] ADD ATTRIBUTE
>         [col] integer NOT NULL;
> 
> CREATE PRIVATE SYNONYM **[user1].**[user1_s1] FOR [user2].[test_t];

TO_BE
> CREATE CLASS [test_t] REUSE_OID, COLLATE utf8_bin;
> ALTER CLASS [test_t] ADD ATTRIBUTE
>         [col] integer NOT NULL;
>
> CREATE PRIVATE SYNONYM [user1_s1] FOR [user2].[test_t];

* Add option 
--same-as-dba : If the login user is a member of the DBA group, extract the same schema file as the DBA.

Remarks
N/A


msg/en_US/utils.msg
```diff
@@ -554,7 +554,8 @@ valid options:\n\
   -S, --SA-mode               stand-alone mode execution\n\
   -C, --CS-mode               client-server mode execution\n\
       --datafile-per-class    create a object file for each class; default: disabled\n\
-      --split-schema-files    split schema information by object and generate each file; Default: generate one schema file with all objects\n
+      --split-schema-files    split schema information by object and generate each file; Default: generate one schema file with all objects\n\
+      --same-as-dba           If the login user is a member of the DBA group, extract the same schema file as the DBA.\n
```
kisoo-han: what about:
--as-dba

src/executables/unload_schema.c
```diff
@@ -636,7 +646,7 @@ get_ordered_classes (print_output & output_ctx, MOP * class_table)
  *    class(in): class MOP
  */
 static void
-emit_class_owner (print_output & output_ctx, MOP class_)
+emit_class_owner (extract_context & ctxt, print_output & output_ctx, MOP class_)
```
eido5: I found that in this function `extract_context & ctxt` parameter is not used.
Is this for function unity or is there any probability that this parameter will be used?

src/executables/unload_schema.c
```diff
@@ -1099,18 +1201,23 @@ extract_schema (extract_context & ctxt, print_output & schema_output_ctx)
 	}
     }
 
+  if (ctxt.is_dba_user == false && ctxt.is_dba_group_member == false)
+    {
+      filter_user_classes (&ctxt.classes, ctxt.login_user);
```
eido5, line 1207: Should we `ctxt.login_user` verify whether NULL or not?
I found `ctxt.classes` is checked. But `ctxt.login_user` is not found.

src/executables/unload_object.c
```diff
@@ -661,13 +669,24 @@ extract_objects (const char *exec_name, const char *output_dirname, const char *
 
 	  if (!datafile_per_class && (!required_class_only || IS_CLASS_REQUESTED (i)))
 	    {
-	      SPLIT_USER_SPECIFIED_NAME (sm_ch_name ((MOBJ) class_ptr), owner_name, class_name);
-	      if (text_print
-		  (obj_out, NULL, 0, "%cid %s%s%s.%s%s%s %d\n", '%', PRINT_IDENTIFIER (owner_name),
-		   PRINT_IDENTIFIER (class_name), i) != NO_ERROR)
+	      if (ctxt.is_dba_user || ctxt.is_dba_group_member)
```
kisoo-han: Even I'm not sure it is effieient but I'll propose a solution to simplify, it up to you its just idea:
1. define a function like:
```
char *
print_owner_name (char * owner_name, bool print_owner_name)
{
  char buf[DB_MAX_USER_LENGTH+4] = "";

  if (print_owner_name)
    {
      sprintf (buf, "%s%s%s%s", PRINT_IDENTIFIER (owner_name), ".")
    }

  return buf;
}
```
2. Next re-write following if-else-block as follow:
```
if (text_print (obj_out, NULL, 0, "%cid %s%s%s%s %d\n", '%', print_owner_name (owner_name, ctxt.is_dba_user || ctxt.is_dba_group_member), PRINT_IDENTIFIER (class_name), ) != NO_ERROR)
  {
    status = 1;
    goto end;
  }
```

kisoo-han (approved):

src/executables/unload_schema.c
airnet73: Thank you for your comment.

Srltas (approved):

PR #4137, opened 2023-02-23
[CBRD-24662] Add a LoggingThread to access javasp's log file by only one thread

hgryoo:
http://jira.cubrid.org/browse/CBRD-24662

Java FileHandler creates a new file with an increasing number when there exist conflicts to a file. Because of this, If an exception occurs very quickly by many threads in the Java SP server, Many log files are generated by the FileHandler's behavior.
To avoid this behavior, This PR implements the LoggingThread to access a log file with a single thread. Other threads will put their exception into LinkedBlockingQueue. And the LoggingThread will take the log string of exception without conflict.

### DIFF
--- a/src/jsp/com/cubrid/jsp/LoggingThread.java
+++ b/src/jsp/com/cubrid/jsp/LoggingThread.java
@@ -1,0 +1,79 @@
+/*
+ *
+ * Copyright (c) 2016 CUBRID Corporation.
+ *
+ * Redistribution and use in source and binary forms, with or without modification,
+ * are permitted provided that the following conditions are met:
+ *
+ * - Redistributions of source code must retain the above copyright notice,
+ *   this list of conditions and the following disclaimer.
+ *
+ * - Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ *
+ * - Neither the name of the <ORGANIZATION> nor the names of its contributors
+ *   may be used to endorse or promote products derived from this software without
+ *   specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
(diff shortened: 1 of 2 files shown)


src/jsp/com/cubrid/jsp/LoggingThread.java
```diff
@@ -0,0 +1,79 @@
...
+                String logString = logQueue.take();
+                logger.log(logginLevel, logString);
+            } catch (InterruptedException e) {
+                break;
+            }
+        }
+
+        if (logHandler != null) {
+            try {
+                logHandler.close();
```
ctshim, line 66: If the queue is not empty, shouldn't we print it out and close()?

src/jsp/com/cubrid/jsp/Server.java
```diff
@@ -74,6 +76,16 @@ private Server(
...
+            String logFilePath =
+                    rootPath
+                            + File.separatorChar
+                            + LOG_DIR
+                            + File.separatorChar
+                            + serverName
+                            + "_java.log";
+            loggingThread = new LoggingThread(logFilePath);
+            loggingThread.start();
+
```
ctshim, line 88: Is the file name determined only when it is first driven?
Isn't there a time when we need to make a new file like when it's over a certain size?

beyondykk9: The log file size is no limit?

shparkcubrid (approved):

hyunikn (approved):

youngjinj (approved):

src/jsp/com/cubrid/jsp/Server.java
hgryoo: As a specification that is already defined, the file name is determined. In logging.properties in the jsp folder, you can set an option to create a new log file when the log file grows to a certain size. This is not controlled in code.

src/jsp/com/cubrid/jsp/LoggingThread.java
hgryoo: This line can be reached in the exit routine before the server exits via exit(0). At this time, all other threads have already terminated, so the producer of the log no longer exists.

beyondykk9 (approved):

ctshim (approved):

PR #4138, opened 2023-02-24
[CBRD-24675] A meaningless assignment occurs in btree_get_stats_key.

hyahong:
http://jira.cubrid.org/browse/CBRD-24675

Purpose
Move the definition to the front of function.

Implementation
N/A

Remarks
N/A

### DIFF
--- a/src/storage/btree.c
+++ b/src/storage/btree.c
@@ -6507,15 +6507,15 @@ btree_get_stats_key (THREAD_ENTRY * thread_p, BTREE_STATS_ENV * env, MVCC_SNAPSH
 
   if (mvcc_snapshot != NULL)
     {
-      bts = &(env->btree_scan);
+      BTS = &(env->btree_scan);
 
-      if (bts->C_page == NULL)
+      if (BTS->C_page == NULL)
 	{
 	  goto exit_on_error;
 	}
 
-      assert (bts->slot_id > 0);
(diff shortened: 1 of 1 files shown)


hornetmj:
@Yechan0815 I leave a comment about the changing BTS to bts in https://github.com/CUBRID/cubrid/pull/4139#issuecomment-1448403293.

**Suggestion**
When you make a PR, you have to focus on the purpose of the issue. If you feel that some point of code need to refactor, then you need to create other issue(or sub-issue) for it or make different PR if the refactoring is related to the same issue.

hornetmj (approved):

joohok (approved):

src/storage/btree.c
```diff
@@ -6498,6 +6498,7 @@ btree_get_stats_key (THREAD_ENTRY * thread_p, BTREE_STATS_ENV * env, MVCC_SNAPSH
   LEAF_REC leaf_pnt;
   bool clear_key = false;
   int offset;
+  int max_visible_oids, num_visible_oids;
```
wrlawodms, line 6501: Nit. I would put the declaration in the scope it's used since it's more readable.

wrlawodms (approved):

src/storage/btree.c
hyahong: I think that declaring the variables in one place is easy to read. Personally, it is convenient to see the declarations at a glance.

PR #4139, opened 2023-02-24
[CBRD-24676] Remove the assert in btree_get_stats_with_AR_sampling.

hyahong:
http://jira.cubrid.org/browse/CBRD-24676

Purpose
Remove the meaningless assert and condition in function btree_get_stats_with_AR_sampling ().

Implementation
N/A

Remarks
N/A

### DIFF
--- a/src/storage/btree.c
+++ b/src/storage/btree.c
@@ -6663,7 +6663,7 @@ exit_on_error:
 static int
 btree_get_stats_with_AR_sampling (THREAD_ENTRY * thread_p, BTREE_STATS_ENV * env)
 {
-  BTREE_SCAN *BTS;
+  BTREE_SCAN *bts;
   int n, i;
   bool found;
   int key_cnt;
@@ -6676,8 +6676,8 @@ btree_get_stats_with_AR_sampling (THREAD_ENTRY * thread_p, BTREE_STATS_ENV * env
   assert (env != NULL);
   assert (env->stat_info != NULL);
 
-  BTS = &(env->btree_scan);
-  BTS->use_desc_index = 0;	/* init */
+  bts = &(env->btree_scan);
+  bts->use_desc_index = 0;	/* init */
 
   for (n = 0; n < STATS_SAMPLING_THRESHOLD; n++)
     {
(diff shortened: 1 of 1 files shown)


hornetmj:
@Yechan0815 It would be nice to make another jira issue for changing lowercase to uppercase of BTS. When I count BTS using grep command, there exists other BTS.

$ grep 'BTS->' src/storage/btree.c | wc -l
51

$ grep 'BTS,' src/storage/btree.h | wc -l
1

Additionally, I agree your suggestion BTS -> bts. But It is not a mandatory. So, I suggest that you need to make an agreement about that in this PR first and create a jira issue for it.

hyahong:
@hornetmj 
Capitalization changes are written in JIRA. But you mean, should an issue be written for one purpose at the time?

hornetmj:
@Yechan0815 It is a suggestion for you and for reviewers. And capitalization changes of BTS in this PR is not cover all. So, It would be nice to change all together in the other jira issue or PR.

hyahong:
@hornetmj 
Aha! I will deal with each of these issues separately. Let me write a new issue for capitalization.

src/storage/btree.c
```diff
@@ -6706,29 +6705,26 @@ btree_get_stats_with_AR_sampling (THREAD_ENTRY * thread_p, BTREE_STATS_ENV * env
 	  assert (header->node_level == 1);	/* BTREE_LEAF_NODE */
 #endif
 
-	  if (key_cnt > 0)
```
hornetmj: The key_cnt of leaf node can be 0. If you remove this line, then the behavior is changed when the key_cnt is zero. Please explain more detail that the found is always false when the key_cnt of leaf node is zero. If you feel that the explanation of it is difficult then I recommand you to change just assert_release().

hyahong: I thought key_cnt was unconditionally bigger than zero, but I think I got confused with other issues. I turned it back.

hornetmj (approved):

src/storage/btree.c
hornetmj: Please change the summary of this PR.

joohok (approved):
PR title is required to be changed.
`unnecessary condition` can be removed

wrlawodms (approved):

PR #4140, opened 2023-02-24
[LETS-652] async_disconnect_handler has to guarantee the all requests are digested

wrlawodms:
http://jira.cubrid.org/browse/LETS-652

Without this #4135 runs into an assertion failure at `assert (m_disconnect_queue.empty ());` during shutdown.

- expend the lock scope to include the termination check in `async_disconnect_handler<T_CONN_HANDLER>::disconnect()`. Otherwise, a request could be added after terminating.


joohok (approved):

src/server/async_disconnect_handler.cpp
```diff
@@ -79,7 +79,7 @@ async_disconnect_handler<T_CONN_HANDLER>::disconnect_loop ()
   constexpr std::chrono::seconds one_second { 1 };
 
   std::queue<connection_handler_uptr_t> disconnect_work_buffer;
-  while (!m_terminate.load ())
+  while (!m_terminate.load () || !m_disconnect_queue.empty())
```
cristiarg: To be completely pedantic, the `m_disconnect_queue` is not guarded by lock.

After `m_terminate` is set, we're sure no one touches `m_disconnect_queue` anymore.
So it is better to add a second loop after the existing one where `m_disconnect_queue` can be accessed unguarded:
```
      while (!m_disconnect_queue.empty ())
	{
	  connection_handler_uptr_t &front = m_disconnect_queue.front ();
	  front.reset (nullptr);
	  m_disconnect_queue.pop ();
	}
```
Or, better, extract this segment in a separate function and call it twice:
```
  while (!m_terminate.load ())
    {
      // lock scope ...

      empty_queue(disconnect_work_buffer);
    }
  empty_queue(m_disconnect_queue);
```

cristiarg: Just realized. It will work as it is as well due to shorthand evaluation of condition.
I would still change it as described above to avoid having to wait `one_second` in absence of a notification from the producing side via `m_queue_cv`.

hornetmj: @cristiarg @wrlawodms

The m_terminate is set to true in the page_server dtor(page_server::~page_server ()). As I know it means that the memory area for it(page_server) is returned to the system. But ps_Gl is a global variable so there is no problem now. However I think that the related resources like m_disconnect_queue, m_queue_mtx, m_thread, etc should be destroyed or cleaned up before calling the page_server dtor or at least in it. (I mean that the m_disconnect_queue handling should be finished before calling the dtor.)

hornetmj (approved):

src/server/async_disconnect_handler.cpp
cristiarg: @hornetmj 
No, there is no problem regarding `page_server` dtor.
While the dtor is being executed, all infrastructure proper to the `page_server` instance (object) is still in its own place (regardless whether the object would be created static global (also maybe somewhere on the heap), dinamically on the heap or on the stack of some thread). The purpose of the dtor is to destroy what has been manually allocated or - as done in this case - to just dispatch a message to a nested object.
The chronology of what happens in the `page_server` dtor is:
- starting with `m_async_disconnect_handler.terminate` - this is a non-blocking function, it just enters sets the value notifies and exists
- the body of the `page_server::~page_server` ends
- now the compiler generated code added as part of the `page_server` dtor will call dtors of `page_server` dtor member variables in the reverse order they are declared in the header (that's what I know - the objects are constructed in the order they are declared in the header and are destroyed in the reverse order)
- thus, when it's `m_async_disconnect_handler`'s turn to be destroyed, `async_disconnect_handler::~async_disconnect_handler` is called
- and the dtor will just join `m_thread` thus waiting for the `disconnect_loop` function to terminate

src/server/async_disconnect_handler.cpp
```diff
@@ -92,11 +92,12 @@ async_disconnect_handler<T_CONN_HANDLER>::disconnect_loop ()
 	m_disconnect_queue.swap (disconnect_work_buffer);
       }
 
-      while (!disconnect_work_buffer.empty ())
-	{
-	  connection_handler_uptr_t &front = disconnect_work_buffer.front ();
-	  front.reset (nullptr);
```
wrlawodms, line 98: Q: Do we really need to set it to nullptr explicitly?

src/server/async_disconnect_handler.cpp
```diff
@@ -50,9 +50,9 @@ template <typename T_CONN_HANDLER>
 void
 async_disconnect_handler<T_CONN_HANDLER>::disconnect (connection_handler_uptr_t &&handler)
 {
+  std::unique_lock<std::mutex> ulock { m_queue_mtx };
   if (!m_terminate.load ())
     {
-      std::unique_lock<std::mutex> ulock { m_queue_mtx };
```
wrlawodms, line 55: To avoid a race condition. With the previous code, the `m_terminate` can be true after the lock is released. It means it could push the disconnect handler after terminating.

src/server/async_disconnect_handler.cpp
hornetmj: @cristiarg
> regardless whether the object would be created static global (also maybe somewhere on the heap), dinamically on the heap or on the stack of some thread

Does this mean that the access of instance memory is guaranteed after calling dtor even if the instance don't use global variable?

cristiarg: > Does this mean that the access of instance memory is guaranteed after calling dtor even if the instance don't use global variable?

In a way. As I understand what we see when we look at a dtor
```
page_server::~page_server ()
{
  // ...
  m_async_disconnect_handler.terminate ();
  // human-written portion of the dtor ends here :)

  // "smarter human"-written (aka: compiler developer) portion of dtor starts here :)
  // ... other dtors called by compiler generated code
  // m_async_disconnect_handler dtor called somewhere around here by compiler generated code
  // ... other dtors called by compiler generated code
}
```
is not all a dtor is.
Thus, in a way, the object is not destroyed after the human-written part of the dtor is executed at runtime.
There are other instructions added by the compiler to call the destructors of member variables.

-----
Side note.
When I think of the way the objects are constructed/destructed I make an analogy with pre-order parsing of a [binary] tree - https://www.geeksforgeeks.org/tree-traversals-inorder-preorder-and-postorder/ :
- at construction: the object tree is traversed in pre-order and constructors are called - either explicit constructors or default or no-ops ctors (in the case of POD types)
- at destruction: the object tree is traversed in reverse pre-order (again, respecting the header file declaration of types) and dtors are called - either explicit destructors or default or no-op dtors (in case of POD types)
[ somebody please correct me if my assumption is incorrect ]
-------

In this case there is no access of an object's memory after destruction. The `m_async_disconnect_handler` has it's dtor called and that dtor waits for another thread to end - thus releasing not only the `async_disconnect_handler::m_thread` object but also all the resources entitled by a thread (stack, OS resources used for that thread, whatever that thread allocated as private memory, aso ..).

src/server/async_disconnect_handler.cpp
cristiarg: fair enough

src/server/async_disconnect_handler.cpp
cristiarg: not really; but the new code - albeit more correct - is also more cryptic :)

cristiarg (approved):

src/server/async_disconnect_handler.cpp
```diff
@@ -92,11 +92,12 @@ async_disconnect_handler<T_CONN_HANDLER>::disconnect_loop ()
...
-	{
-	  connection_handler_uptr_t &front = disconnect_work_buffer.front ();
-	  front.reset (nullptr);
-	  disconnect_work_buffer.pop ();
-	}
+      disconnect_work_buffer = {}; // clear
+    }
+
+  // clear requests added after swapped to m_disconnect_queue before termination.
+  if (!m_disconnect_queue.empty ())
```
cristiarg, line 101: maybe just add a comment that the queue can be accessed here unlocked as `disconnect` cannot be accessed by other thread because `m_terminate` is already set.

wrlawodms: Actually, it can. Because during 95L`disconnect_work_buffer = {};`, the lock is not being held, so `disconnect()` can acquire the lock and interleave (m_terminate is true here). When it's about to push the disconnect request, it's terminated and this clearing `m_disconnect_queue` job can start. This also has to be guarded..

wrlawodms: I've made the change to also acquire the lock during clearing left requests.

cristiarg: @wrlawodms 
I think that the problem is the atomic `m_terminate`.
The code would be more free of possible race conditions if the `m_terminate` is a simple bool that will be guarded by the same mutex as the container. And `disconnect_loop` function reworked a little.
But since it is not super-critical code - ie: it is used occasionally when a xTS will disconnect - let's leave it like this.

cristiarg (approved):

joohok (approved):

hornetmj (approved):
LGTM.

PR #4141, opened 2023-02-24
[LETS-654] Do disconnect_all_tran_server() only in xboot_shutdown_server

wrlawodms:
http://jira.cubrid.org/browse/LETS-654

- Delete ps_Gl->disconnect_all_tran_server (); in finalize_server_type(), which is called redundantly.
- Move ps_Gl->disconnect_all_tran_server (); to xboot_shutdown_server too for code consistency.
  - Update the comment about the shutdown sequence.
- Without this, #4135 runs into an assertion failure during shutdown.

### DIFF
--- a/src/base/server_type.cpp
+++ b/src/base/server_type.cpp
@@ -215,7 +215,6 @@ void finalize_server_type ()
 	  assert (ats_Gl != nullptr);
 	  ats_Gl = nullptr;
 	}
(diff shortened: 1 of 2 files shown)


joohok (approved):

cristiarg (approved):
LGTM.
We need to wait for an execution of the test suite to validate this.

PR #4145, opened 2023-02-27
[CBRD-24678] Disable the deprecated function lock_initialize_resource () in lock manager.

hyahong:
http://jira.cubrid.org/browse/CBRD-24678

**Purpose**
The enum type LOCK_RESOURCE_OBJECT is not used and the function lock_initialize_resource (), which only uses the enum LOCK_RESOURCE_OBJECT, is deprecated now.
Also, if LOCK_RESOURCE_OBJECT exists during processing, call the assert.

**Implementation**
N/A

**Remarks**
N/A

### DIFF
--- a/src/transaction/lock_manager.h
+++ b/src/transaction/lock_manager.h
@@ -154,7 +154,7 @@ typedef enum
   LOCK_RESOURCE_INSTANCE,	/* An instance resource */
   LOCK_RESOURCE_CLASS,		/* A class resource */
   LOCK_RESOURCE_ROOT_CLASS,	/* A root class resource */
-  LOCK_RESOURCE_OBJECT		/* An object resource */
+  LOCK_RESOURCE_OBJECT		/* An object resource: used when the type of locking resource is unknown or undetermined. */
 } LOCK_RESOURCE_TYPE;
 
 /*
--- a/src/transaction/lock_manager.h
+++ b/src/transaction/lock_manager.h
@@ -154,7 +154,7 @@ typedef enum
   LOCK_RESOURCE_INSTANCE,	/* An instance resource */
   LOCK_RESOURCE_CLASS,		/* A class resource */
   LOCK_RESOURCE_ROOT_CLASS,	/* A root class resource */
-  LOCK_RESOURCE_OBJECT		/* An object resource: used when the type of locking resource is unknown or undetermined. */
+  LOCK_RESOURCE_OBJECT		/* An object resource. Obsolete */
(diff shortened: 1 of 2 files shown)


src/transaction/lock_manager.h
```diff
@@ -154,7 +154,7 @@ typedef enum
   LOCK_RESOURCE_INSTANCE,	/* An instance resource */
   LOCK_RESOURCE_CLASS,		/* A class resource */
   LOCK_RESOURCE_ROOT_CLASS,	/* A root class resource */
-  LOCK_RESOURCE_OBJECT		/* An object resource */
+  LOCK_RESOURCE_OBJECT		/* An object resource: used when the type of locking resource is unknown or undetermined. */
```
joohok: If it is not used anywhere, how about checking it as an obsolete one ?
```suggestion
  LOCK_RESOURCE_OBJECT		/* An object resource. Obsolete */
```

hyahong: It's simple. looks like good!

wrlawodms (approved):

src/transaction/lock_manager.h
```diff
@@ -154,7 +154,7 @@ typedef enum
   LOCK_RESOURCE_INSTANCE,	/* An instance resource */
   LOCK_RESOURCE_CLASS,		/* A class resource */
   LOCK_RESOURCE_ROOT_CLASS,	/* A root class resource */
-  LOCK_RESOURCE_OBJECT		/* An object resource */
+  LOCK_RESOURCE_OBJECT		/* An object resource. Obsolete */
```
hornetmj, line 157: How about change the following too?

case LOCK_RESOURCE_OBJECT: -> assert(); in the lock_res_key_copy()
```
839     case LOCK_RESOURCE_OBJECT:
840       /* nothing, it's a free object */
841       break;
842
843     default:
844       assert (false);
845       return ER_FAILED;
```

hyahong: When the function lock_res_key_copy() is called, if the type is LOCK_RESOURCE_OBJECT, then something is wrong. Because this function is called to insert RESOURCE BLOCK.
so It looks good to use fall through for assert.

```
     case LOCK_RESOURCE_OBJECT:
     default:
       assert (false);
       return ER_FAILED;
```

hornetmj: @Yechan0815 Do it please.

hornetmj (approved):

joohok (approved):

PR #4146, opened 2023-02-27
[CBRD-24681] Change the lock mode that will be released on error after the function file_tracker_interruptable_iterate () is called.

hyahong:
http://jira.cubrid.org/browse/CBRD-24681

**Purpose**
```
lock_unlock_object (thread_p, &class_oid, oid_Root_class_oid, "requested lock mode", "force");
```
If the parameter force is true when the function lock_unlock_object is called, the lock on the resource is released, even if the requested lock mode is different from the actual lock mode.
So requesting to release the lock different from the actual lock didn't make problems.

**Implementation**
N/A

**Remarks**
N/A

### DIFF
--- a/src/storage/btree.c
+++ b/src/storage/btree.c
@@ -8258,7 +8258,7 @@ btree_repair_prev_link (THREAD_ENTRY * thread_p, OID * oid, BTID * index_btid, b
 	  break;
 	}
     }
-  /* if the break statement occurs due to an error, the lock acquired immediately before should be released. */
+  /* if the break statement occurs due to an error, the acquired lock should be released. */
   if (!OID_ISNULL (&class_oid))
     {
       lock_unlock_object (thread_p, &class_oid, oid_Root_class_oid, IX_LOCK, true);
@@ -8319,7 +8319,7 @@ btree_check_all (THREAD_ENTRY * thread_p)
 
   if (!OID_ISNULL (&class_oid))
     {
-      lock_unlock_object (thread_p, &class_oid, oid_Root_class_oid, SCH_S_LOCK, true);
+      lock_unlock_object (thread_p, &class_oid, oid_Root_class_oid, IX_LOCK, true);
     }
   return allvalid;
 }
--- a/src/storage/btree.c
+++ b/src/storage/btree.c
@@ -8261,7 +8261,7 @@ btree_repair_prev_link (THREAD_ENTRY * thread_p, OID * oid, BTID * index_btid, b
   /* if the break statement occurs due to an error, the lock acquired immediately before should be released. */
   if (!OID_ISNULL (&class_oid))
     {
-      lock_unlock_object (thread_p, &class_oid, oid_Root_class_oid, FILE_GET_TRACKER_LOCK_MODE (FILE_BTREE), true);
+      lock_unlock_object (thread_p, &class_oid, oid_Root_class_oid, IX_LOCK, true);
     }
 
   return valid;
--- a/src/storage/btree.c
+++ b/src/storage/btree.c
@@ -8258,9 +8258,10 @@ btree_repair_prev_link (THREAD_ENTRY * thread_p, OID * oid, BTID * index_btid, b
 	  break;
 	}
     }
+  /* if the break statement occurs due to an error, the lock acquired immediately before should be released. */
   if (!OID_ISNULL (&class_oid))
     {
-      lock_unlock_object (thread_p, &class_oid, oid_Root_class_oid, SCH_S_LOCK, true);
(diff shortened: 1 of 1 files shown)


joohok:
Please check the build failure `ci/circleci:build`

hyahong:
@joohok 
Thank you! I checked

src/storage/btree.c
```diff
@@ -8258,9 +8258,10 @@ btree_repair_prev_link (THREAD_ENTRY * thread_p, OID * oid, BTID * index_btid, b
 	  break;
 	}
     }
+  /* if the break statement occurs due to an error, the lock acquired immediately before should be released. */
   if (!OID_ISNULL (&class_oid))
     {
-      lock_unlock_object (thread_p, &class_oid, oid_Root_class_oid, SCH_S_LOCK, true);
+      lock_unlock_object (thread_p, &class_oid, oid_Root_class_oid, IX_LOCK, true);
```
wrlawodms, line 8264: > So requesting to release the lock different from the actual lock didn't make problems.

- What is the upside of using IX_LOCK instead of SCH_S_LOCK to unlock this? Please elaborate on it. 
- Is it locked with SCH_S_LOCK?

hyahong: No. It is locked with IX_LOCK, but lock_unlock_object requested SCH_S_LOCK.
so It should be fixed.

hornetmj, line 8264: It looks like if the file_tracker_interruptable_iterate() returns !NO_ERROR then the caller releases the class lock if needed. There exists four position which calls this function and uses the same lock release handling when I checked. Can we have a chance to move this if phrase into the file_tracker_interruptable_iterate()?

wrlawodms: Oh, I guess I forgot to check the Jira description. Sorry.

wrlawodms:
```
lock_unlock_object (thread_p, class_oid, oid_Root_class_oid, FILE_GET_TRACKER_LOCK_MODE (desired_ftype), true);
```
Based on the above code in the `file_tracker_interruptable_iterate()`, IX_LOCK seems only for BTREE_FILE and only two (`btree_check_all()` and `btree_repair_prev_link()`) cases among the four are with BTREE_FILE.

wrlawodms (approved):

src/storage/btree.c
hyahong: oh. btree_check_all () also looks like it needs to be fixed. Is it okay to handle the modification of btree_check_all () togatehr in this issue?

wrlawodms: I think it's OK with the description of the issue and PR revised more generally. After communicating with @hornetmj, by the way. I guess he means handling unlocking inside the `file_tracker_interruptable_iterate()`.

hyahong: Even if `file_tracker_interruptable_iterate ()` operates normally and returns NO_ERROR, unlock is required if an error occurs in below.

```
      if (file_tracker_interruptable_iterate (thread_p, FILE_BTREE, &btid.vfid, &class_oid) != NO_ERROR)
	{
	  ASSERT_ERROR ();
	  valid = valid == DISK_VALID ? DISK_ERROR : valid;
	  break;
	}
      if (VFID_ISNULL (&btid.vfid))
	{
	  /* no more b-trees */
	  break;
	}

      if (btree_get_btid_from_file (thread_p, &btid.vfid, &btid) != NO_ERROR)
	{
	  ASSERT_ERROR ();
	  valid = valid == DISK_VALID ? DISK_ERROR : valid;
	  break;
	}

      /* get the index name of the index key */
      if (heap_get_indexinfo_of_btid (thread_p, &class_oid, &btid, NULL, NULL, NULL, NULL, &index_name, NULL)
	  != NO_ERROR)
	{
	  ASSERT_ERROR ();
	  valid = valid == DISK_VALID ? DISK_ERROR : valid;
	  break;
	}

      valid = btree_repair_prev_link_by_btid (thread_p, &btid, repair, index_name);
      if (valid == DISK_ERROR)
	{
	  ASSERT_ERROR ();
	  break;
	}
```

When the statement `break` is called because of an error, the unlocking is needed too.
So, I think it is difficult to move `lock_unlock_object ()` into `file_tracker_interruptable_iterate ()`.

hyahong: @hornetmj 
If you think it's okay, I will revise the issue and description of PR (and JIRA) more comprehensively.

hornetmj: @Yechan0815 Go ahead!

joohok (approved):

hornetmj (approved):

PR #4148, opened 2023-02-27
[LETS-640] Quorum commit with the majority of PSes

wrlawodms:
http://jira.cubrid.org/browse/LETS-640

- Each `connection_handler` has the `m_saved_lsa` for its corresponding page server.
- Change `m_max_ps_flushed_lsa` which is the maximum saved_lsa among page servers to `m_ps_consensus_flushed_lsa` which is the lsa until which the majority of PSes confirm to flush.
  - This is calculated in `active_tran_server::compute_consensus_lsa ()` by gathering all saved LSA from all `page_server_node`s. 
  - When a connection_handler receives saved_lsa from its server, it request log_Gl to update `m_ps_consensus_flushed_lsa`. It gets the value from active_tran_server.
- Add some debugging msg to check if everything goes well.


src/transaction/log_page_buffer.c
```diff
@@ -4499,10 +4499,17 @@ logpb_flush_pages (THREAD_ENTRY * thread_p, const LOG_LSA * flush_lsa)
       // *INDENT-OFF*
       if (get_server_type () == SERVER_TYPE_TRANSACTION && ts_Gl->is_page_server_connected ())
 	{
-	  log_Gl.wait_flushed_lsa (*flush_lsa);
 	  if (prm_get_bool_value (PRM_ID_ER_LOG_COMMIT_CONFIRM))
 	    {
-	      _er_log_debug (ARG_FILE_LINE, "Page server committed LSA = %lld|%d.\n", LSA_AS_ARGS (&log_Gl.m_max_ps_flushed_lsa));
+	      _er_log_debug (ARG_FILE_LINE, "[COMMIT CONFIRM] Wait until consenesus LSA gets larger than or equal to %lld|%d.\n", LSA_AS_ARGS (flush_lsa));
```
joohok: logpb_flush_pages () is also called by the checkpoint function.
(ATS checkpoint : logpb_checkpoint_trantable () -> logpb_flush_pages () )
(In order to flush the LOG_TRANTABLE_SNAPSHOT log record for PTS recovery)

When ATS do checkpoint, I think this debug log message can be logged. 

if the checkpoint also need the consensus, how about just removing `[COMMIT CONFIRM]`  ?

hornetmj (approved):
LGTM.

IMO:
It would be nice the page_server_node has all about page server related connection, variables, etc. With this, tran_server(active_tran_server, passive_tran_server) use the page_server_node interface only when they do something related to the page server.

src/transaction/log_page_buffer.c
wrlawodms: It makes sense. While considering your suggestion, a concern came up that some regression tests would use the existing debug logging, and it turned out to be the case. So, I'll leave existing errors as it is, and just add some logging for this PR without "COMMIT CONFIRM". Thanks.

src/server/tran_server.hpp
```diff
@@ -81,6 +82,34 @@ class tran_server
     virtual bool uses_remote_storage () const;
 
   protected:
+    /*
+     * The permanent data of each page server is stored and served regardless of connection.
+     */
+    class page_server_node
```
cristiarg: As this is now, I do not see any reason to have it as a new member in the base class.

The implementation is [over]complicated and involves `dynamic_cast` (which is, again - in my view, a red-flag when it comes to design) (you'll notice that is not used anywhere in the codebase except in unit tests).

Consideration (1)
We can consider `m_connection_list` as being static information (see also suggestion comment below).
While `m_saved_lsa` is dynamic information - linked to the list of active connection handlers. We already intend to maintain `connection_handler` instances as de-facto objects for handling connections with peer servers in the cluster.
For instance, in the future we can use the `m_connection_list` to poll to re-connect to page servers (or, what not)
So, I would not mix the two concepts.
What i would do though, is duplicate the `cubcomm::conn` info also inside the `tran_server::connection_handler` base class (because it seems that this is needed - in the very least for logging).

Consideration (2)
Consideration (1) being said, I think it's easiest to refactor the current patch as:
- extend `active_tran_server::conn_handler` with information about `m_saved_lsa`
- in `active_tran_server::conn_handler::receive_saved_lsa` just store the received value in `conn_handler::m_saved_lsa`:
- and then call a newly added virtual function on the `tran_server`/`active_tran_server` hierarchy - `compute_consensus_lsa` (or whatever you want to call it)
  -  [[ the call to `compute_consensus_lsa` can even be implemented asynchronously if need be - `std::async` - but we'll see ]]
  - the overridden implementation `active_tran_server::compute_consensus_lsa` will do the actual computation - that is done now in `active_tran_server::connection_handler::receive_saved_lsa` (the consensus calculation does not belong conceptually at the level of a connection handler - but to a higher level)
    - this way it is also ensured that when a Page Server disconnects (for whatever reason) it's corresponding `m_saved_lsa` is not used anymore; but we'll have to decide whether - when a PS disconnects - the quorum is reduced or remains the same
  - the overridden implementation `passive_tran_server::compute_consensus_lsa` will just assert (will not be called on a passive transaction server)

I think it can be done simpler. Please comment if I got anything incorrect.

src/server/tran_server.hpp
```diff
@@ -141,7 +174,6 @@ class tran_server
     int parse_page_server_hosts_config (std::string &hosts);
 
   private:
-    std::vector<cubcomm::node> m_connection_list;
```
cristiarg, line 144: Add comment for `m_connection_list`:
```
Static information about available page server connection peers.
For now, this information is static. In the future this can be maintained dinamically (eg: via cluster
management sofware).
```

src/server/tran_server.hpp
```diff
@@ -102,11 +131,13 @@ class tran_server
...
 
       protected:
-	connection_handler (cubcomm::channel &&chn, tran_server &ts, request_handlers_map_t &&request_handlers);
+	connection_handler (cubcomm::channel &&chn, page_server_node &node, tran_server &ts,
+			    request_handlers_map_t &&request_handlers);
 
 	virtual request_handlers_map_t get_request_handlers ();
 
       protected:
+	page_server_node &m_node; // in tran_server::m_node_vec;
```
cristiarg: with the new approach (as described above) this will be a
```suggestion
	cubcomm::node m_conn_node;
```
or
```suggestion
	const cubcomm::node &m_conn_node;
```

..a const reference to the entry in `tran_server::m_connection_list`

src/server/active_tran_server.cpp
```diff
@@ -104,20 +105,52 @@ active_tran_server::connection_handler::disconnect ()
...
+      log_Gl.update_ps_consensus_flushed_lsa (consensus_lsa);
     }
 
   if (prm_get_bool_value (PRM_ID_ER_LOG_COMMIT_CONFIRM))
     {
-      _er_log_debug (ARG_FILE_LINE, "[COMMIT CONFIRM] Received LSA = %lld|%d.\n", LSA_AS_ARGS (&saved_lsa));
+      std::stringstream ss;
+      const log_lsa consensus_lsa = log_Gl.get_ps_consensus_flushed_lsa ();
+
+      ss << "[COMMIT CONFIRM] Received saved LSA = " << saved_lsa.pageid << "|" << saved_lsa.offset;
```
cristiarg: Nit.
I would really use a stack fixed buffer + `snprintf`; it is probably faster; and this is not debug-only code; I know iostream is "fancier" but ... :)

src/server/active_tran_server.cpp
```diff
@@ -104,20 +105,52 @@ active_tran_server::connection_handler::disconnect ()
...
+      collected_saved_lsa.emplace_back (node->get_saved_lsa ());
+    }
+  std::sort (collected_saved_lsa.begin (), collected_saved_lsa.end ());
+
+  const log_lsa consensus_lsa = collected_saved_lsa[total_node_cnt - quorum];
+  if (log_Gl.get_ps_consensus_flushed_lsa () < consensus_lsa)
     {
-      log_Gl.update_max_ps_flushed_lsa (saved_lsa);
+      log_Gl.update_ps_consensus_flushed_lsa (consensus_lsa);
     }
```
cristiarg: this verification can be done inside `log_global::update_max_ps_flushed_lsa` with only one lock of the mutex; this version causes 2 locks of the mutex;
also, `get_ps_consensus_flushed_lsa` is not needed because the effective value (either set or the old value) can be returned back by the `set_...` function and used for logging below (again, saving one more mutex lock)

src/transaction/log_global.c
```diff
@@ -103,25 +103,32 @@ log_global::~log_global ()
...
   {
     std::unique_lock<std::mutex> lock (m_ps_lsa_mutex);
-    m_max_ps_flushed_lsa = lsa;
+    m_ps_consensus_flushed_lsa = lsa;
   } // lock.unlock ();
   m_ps_lsa_cv.notify_all ();
 }
 
+log_lsa
+log_global::get_ps_consensus_flushed_lsa ()
```
cristiarg: maybe not needed?
or, if needed, restrained to logging needs (eg: assert that some logging sys param is set)
function should also be `const`

src/server/active_tran_server.cpp
```diff
@@ -60,6 +60,64 @@ active_tran_server::get_oldest_active_mvccid_from_page_server () const
...
 
+log_lsa
+active_tran_server::compute_consensus_lsa () const
+{
+  const int total_node_cnt = m_connection_list.size ();
+  const int quorum = total_node_cnt / 2 + 1; // For now, it's fixed to the number of the majority.
+  int cur_node_cnt;
+  std::vector<log_lsa> collected_saved_lsa;
+
+  // TODO The next block has to be exclusive with connection and disconnection
```
wrlawodms, line 71: It'll be addressed in #4135, which is introducing the mutex for this.

src/server/passive_tran_server.cpp
```diff
@@ -56,6 +56,12 @@ passive_tran_server::connection_handler::get_request_handlers ()
   return handlers_map;
 }
 
+log_lsa
+passive_tran_server::connection_handler::get_saved_lsa () const
+{
+  assert (false); // Only used in active transaction server
```
wrlawodms, line 62: I have no idea but defining `get_saved_lsa()` in tran_server to access it through `tran_server::connection_handler` in `active_tran_server::compute_consensus_lsa()`. Please tell me if you have a better idea.

src/server/tran_server.hpp
wrlawodms: Overall, I've accepted your suggestion. Thank you.
The difference: 
- compute_consensus_lsa in active_tran_sever, not in tran_sever.
- receive_saved_lsa just requests log_Gl to update the LSA and it tries to get the latest one from active_tran_server.
- No cubcomm::node in the connection_handler, which is not necessary.

src/server/active_tran_server.cpp
wrlawodms: done :)

src/transaction/log_global.c
wrlawodms: removed.

wrlawodms:
Regarding the cppcheck failure, it's a false-positive due to LSA_AS_ARGS().

cristiarg:
> Regarding the cppcheck failure, it's a false-positive due to LSA_AS_ARGS().

can be suppressed with something like:
```
  // cppcheck-suppress [wrongPrintfScanfArgNum]
```
already used in unit test code - `unit_tests/log_repl/test_main_log_repl_atomic_helper.cpp` .

src/transaction/log_page_buffer.c
```diff
@@ -4499,10 +4499,17 @@ logpb_flush_pages (THREAD_ENTRY * thread_p, const LOG_LSA * flush_lsa)
...
-	      _er_log_debug (ARG_FILE_LINE, "Page server committed LSA = %lld|%d.\n", LSA_AS_ARGS (&log_Gl.m_max_ps_flushed_lsa));
+	      _er_log_debug (ARG_FILE_LINE, "Wait until consenesus LSA gets larger than or equal to %lld|%d.\n", LSA_AS_ARGS (flush_lsa));
+	    }
+	  
+          log_Gl.wait_flushed_lsa (*flush_lsa);
+	  
+          if (prm_get_bool_value (PRM_ID_ER_LOG_COMMIT_CONFIRM))
+	    {
+              std::lock_guard<std::mutex> lock_guard (log_Gl.m_ps_lsa_mutex);
+	      _er_log_debug (ARG_FILE_LINE, "Page server committed LSA = %lld|%d.\n", LSA_AS_ARGS (&log_Gl.m_ps_consensus_flushed_lsa));
```
```suggestion
	      _er_log_debug (ARG_FILE_LINE, "Page server consensus LSA = %lld|%d.\n", LSA_AS_ARGS (&log_Gl.m_ps_consensus_flushed_lsa));
```
maybe it is better to change "committed" to "consensus" or something.

src/server/active_tran_server.cpp
```diff
@@ -60,6 +60,68 @@ active_tran_server::get_oldest_active_mvccid_from_page_server () const
...
+		    "compute_consensus_lsa - total node count = %d, current node count = %d, quorum = %d, consensus LSA = %lld|%d\n",
+		    total_node_cnt, cur_node_cnt, quorum, LSA_AS_ARGS (&consensus_lsa));
+      n += snprintf (msg_buf + n, BUF_SIZE - n, "Collected saved lsa list = [ ");
+      for (const auto &lsa : collected_saved_lsa)
+	{
+	  // cppcheck-suppress [wrongPrintfScanfArgNum]
+	  n += snprintf (msg_buf + n, BUF_SIZE - n, "%lld|%d ", LSA_AS_ARGS (&lsa));
+	}
+      snprintf (msg_buf + n, BUF_SIZE - n, "]\n");
+      assert (n < BUF_SIZE);
```
joohok, line 113:
```suggestion
	  n += snprintf (msg_buf + n, BUF_SIZE - n, "%lld|%d, ", LSA_AS_ARGS (&lsa));
	}
      snprintf (msg_buf + n - 2, BUF_SIZE - n, " ]\n");
      assert (n < BUF_SIZE);
```
Nit.
How about putting a comma between LSA values? 

it may looks like this :
```
compute_consensus_lsa - total node count = 0, current node count = 0, quorum = 0, consensus LSA = 0|0
Collected saved lsa list = [ 0|1, 1|2, 2|3, 3|4, 4|5 ]
```

joohok (commented):
I started by looking at the minor details. I will take a closer look at the overall structure of the code.

src/transaction/log_page_buffer.c
wrlawodms: Hmm.. my concern was some regression tests would already use this msg like "COMMIT CONFIRM". However, I can't find "server committed" string in tests using `grep`. So, I guess I can revise it.

src/server/active_tran_server.cpp
wrlawodms: Hmm.. In my opinion, It's not worth making the arguments less readable. Thanks.

src/transaction/log_global.c
```diff
@@ -143,5 +135,29 @@ log_global::get_log_prior_receiver ()
...
+
+void
+log_global::update_ps_consensus_flushed_lsa ()
+{
+  assert (is_active_transaction_server ());
+
+  const log_lsa consensus_lsa = get_active_tran_server_ptr ()->compute_consensus_lsa ();
+  if (consensus_lsa == NULL_LSA)
+  {
+    return; // The number of connected node are less than the quorum
```
cristiarg: This effectively means that we leave the transactions waiting on this locked? Because we're not notifying the waiting transactions if the consensus LSA cannot be calculated.
I mean we do not have to solve this now with this PR. The question is to just have a discussion on this?

The logical outcome would be to abort the waiting (and, implicitly, all future transactions)?

For the time being, I think a temporary `debug_release` somewhere is advised.

src/server/active_tran_server.cpp
```diff
@@ -110,15 +173,27 @@ active_tran_server::connection_handler::receive_saved_lsa (page_server_conn_t::s
...
-    }
+  assert (saved_lsa > get_saved_lsa ()); // increasing monotonically
+  set_saved_lsa (saved_lsa);
 
   if (prm_get_bool_value (PRM_ID_ER_LOG_COMMIT_CONFIRM))
     {
       _er_log_debug (ARG_FILE_LINE, "[COMMIT CONFIRM] Received LSA = %lld|%d.\n", LSA_AS_ARGS (&saved_lsa));
     }
+
+  log_Gl.update_ps_consensus_flushed_lsa ();
```
cristiarg: Consensus calculation has the potential of being a costly calculation (mutex and all..)
Here we have the option to:
- calculate the consensus on the communication thread (as done now) - and then notify the waiting client threads
- just notify the waiting client threads and let one of them (the first one able to acquire that mutex) do it

Which one do you think is the best option?

src/transaction/log_page_buffer.c
cristiarg: We can adapt the regression test. I think it is more important - for the long term - to have meaningful /debugerror messages. :)
So, if it's more proper to reword the message, we can do it.

src/server/passive_tran_server.cpp
cristiarg: It is ok like this.
I understand it is a bit contrived but I have no better idea ... (It is the price to pay for not using a `dynamic_cast`)

cristiarg: However, you still need to return a `NULL_LSA`.
If you look into the compile output from gcc you'll see a warning for not returning a value.

src/transaction/log_global.c
```diff
@@ -32,6 +32,8 @@
 #include "mvcc_table.hpp"
 #include "porting.h"
 #include "storage_common.h"
+#include "server_type.hpp"
```
cristiarg, line 35: order of includes - alphabetical (as far as I know)

src/transaction/log_global.c
```diff
@@ -143,5 +135,29 @@ log_global::get_log_prior_receiver ()
   assert (m_prior_recver != nullptr);
   return *m_prior_recver;
 }
+
+void
+log_global::update_ps_consensus_flushed_lsa ()
```
cristiarg: Nit.
You changed the order of the functions - presumably - to reflect the order in the header. (this is how I understand your reason)
However, this is not a must because it adds unwanted noise in the diffs.
And some poor guy (maybe still one of us) coming back to this change in a year or two, will have some ugly words to say about the change trying to trace the modification causing a nasty bug. :)

cristiarg (approved):
Approved with two outstanding changes needed:
- return `NULL_LSA`
- and order of includes

Thanks.

joohok (approved):

src/server/passive_tran_server.cpp
wrlawodms: Thank you for the advice.

src/transaction/log_global.c
wrlawodms: done

src/transaction/log_global.c
wrlawodms: make sense. I'll follow that. :)

src/transaction/log_global.c
wrlawodms: > This effectively means that we leave the transactions waiting on this locked? 

Yes, it is. 
However, in my opinion, I don't think we should abort waiting (and future) transactions. The way I see it, it's an acceptable case that the quorum is not met temporarily and transactions wait until the quorum number is met again. It's one of the advantages of keeping going with some temporarily unavailable nodes (not only with a node lagging behind). This case would be added as a test case with the mechanism of catch-up and attaching a new node in the next epic.
 For the long term, we may need a timeout mechanism here to deal with a long period of unsatisfying quorum. For now, it's better to leave it as it is.

src/server/active_tran_server.cpp
wrlawodms: The second seems better in that the waiting threads are just waiting and the request handler can be busy. I'll try it. Thank you.

src/transaction/log_page_buffer.c
wrlawodms: OK, I'll revise debug loggings. Let's adapt test cases if needed.

hornetmj (approved):

wrlawodms:
@cristiarg @joohok 
Changes:
- logging
  - Change the log message: [COMMIT CONFIRM] -> [QUORUM CONSENSUS] since it's not only for commit
  - quourm_consenesus_er_log() since more logs are added.
  - The name of the system parameter er_log_commit_confim -> er_log_quorum_consensus
- computing the quorum lsa
  - Done by a waiter for the consensus flushed lsa, not by the request handler.
  - The handler received the saved_lsa just wakes up the waters and the waiter who sees the change first updates the consensus lsa
  - Why's m_ps_lsa_up_to_date atomic, not protected by the same mutex? To get the request handler to avoid lock contention (in wakeup_ps_flush_waiters()).

src/server/active_tran_server.cpp
```diff
@@ -110,15 +169,24 @@ active_tran_server::connection_handler::receive_saved_lsa (page_server_conn_t::s
...
-    {
-      _er_log_debug (ARG_FILE_LINE, "[COMMIT CONFIRM] Received LSA = %lld|%d.\n", LSA_AS_ARGS (&saved_lsa));
-    }
+  quourm_consenesus_er_log ("Received saved LSA = %lld|%d.\n", LSA_AS_ARGS (&saved_lsa));
+
+  log_Gl.wakeup_ps_flush_waiters ();
+}
+
+void
+active_tran_server::connection_handler::set_saved_lsa (const log_lsa lsa)
```
cristiarg: I think `set_saved_lsa` is not needed - only used internally.

src/transaction/log_impl.h
```diff
@@ -986,6 +986,10 @@ extern bool cdc_Logging;
 #define LOG_THREAD_TRAN_ARGS(thread_p) thread_get_current_entry_index (), LOG_FIND_CURRENT_TDES (thread_p)
 #endif /* SERVER_MODE */
 
+#define quourm_consenesus_er_log(msg, ...) \
```
cristiarg: Typo

src/transaction/log_global.c
```diff
@@ -102,29 +106,41 @@ log_global::~log_global ()
...
+log_global::wait_for_ps_flushed_lsa (const log_lsa &flush_lsa)
 {
-  if (m_max_ps_flushed_lsa >= flush_lsa)
+  std::unique_lock<std::mutex> ulock (m_ps_lsa_mutex);
+  while (m_ps_consensus_flushed_lsa < flush_lsa)
     {
-      // already flushed
-      return;
+      // Only who first notices it's out-dated updates the consensus lsa.
+      if (!m_ps_lsa_up_to_date.exchange (true))
```
cristiarg, line 125: I don't think the `std::atomic<bool> m_ps_lsa_up_to_date` is needed.
The mutex and the condition variable is enough - at least that's the way I see it:
- when the connection thread notifies all waiting client threads (transactions) about a new saved lsa
- the first waiting thread that is woken up by the notification, will lock the mutex, calculate the consensus LSA and return up its stack
- on their own, the other threads will be woken up and also found the value has been calculated, and then unblock/return the value up their stack

Is this the case, or do I get it wrong?
The way I see it, slight upgrade from the previous version:
```
  std::unique_lock<std::mutex> ulock (m_ps_lsa_mutex);
  m_ps_lsa_cv.wait (ulock, [flush_lsa, this]
    {
      // Mandatory: first check and return if the consensus LSA value is enough - it
      //       might have already been calculated by another thread - because all
      //       waiting client threads/transactions execute the same waiting routine;

      // if not, call consensus LSA calculation function
      // this also covers the corner case where a saved LSA is confirmed by one
      //       page server and the notification is not picked up by any waiting thread - this
      //       should only happen in the case where the client transaction that should
      //       have waited for this commit quorum was abnormally aborted - maybe
      //       even impossible to happen, but we're covered anyway

      // if NULL_LSA or does not satisfy flush_lsa, return false - thus continuing to wait
      // return true otherwise, ending the wait for the current thread
    });
    // might even be a good idea to switch to a `condition_variable::wait_for` instead of
    // the current `condition_variable::wait`
  ulock.unlock();
  // Question: `m_ps_lsa_cv.notify_all` for other waiting threads??!! not sure if needed
```

src/server/active_tran_server.hpp
```diff
@@ -34,6 +35,7 @@ class active_tran_server : public tran_server
 
     bool uses_remote_storage () const final override;
     MVCCID get_oldest_active_mvccid_from_page_server () const;
+    log_lsa compute_consensus_lsa () const;
```
cristiarg: this function will drop the `const` upon #4135 because of the addition of the mutex

src/transaction/log_impl.h
```diff
@@ -986,6 +986,10 @@ extern bool cdc_Logging;
...
+  if (prm_get_bool_value (PRM_ID_ER_LOG_QUORUM_CONSENSUS)) \
+    _er_log_debug (ARG_FILE_LINE, "[QUORUM CONSENSUS] " msg, __VA_ARGS__)
```
cristiarg: Nit. I usually make a mnemonic such as `QUORUM CONSENSUS` as `QUORUM_CONSENSUS` - one word. It makes it easier to search for in logs.

src/transaction/log_global.c
wrlawodms: Your approach would work correctly, but some waste can occur when the condition is not met:
- The saved_lsa is received, so It wakes up all transactions. However, no transaction is happy with the newly calculated consensus lsa. It makes all transactions do the same calculations and sleep again.
- Every new transaction calculates the LSA if the LSA hasn't been met for the transaction.

To avoid the useless calculation, which can occur whenever a saved_lsa is received, the condition variable is needed. Do you think it's better to ignore it?

> might even be a good idea to switch to a `condition_variable::wait_for` instead of

Do we have something to do periodically in the middle of waiting? Or is it for the future like timeout?

src/server/active_tran_server.hpp
wrlawodms: Ok, removed in advance.

src/transaction/log_impl.h
wrlawodms: I'll follow it.

src/transaction/log_global.c
cristiarg: > The saved_lsa is received, so It wakes up all transactions. However, no transaction is happy with the newly calculated consensus lsa. It makes all transactions do the same calculations and sleep again.

The way I see it:
- all transactions that are waiting are doing so by the `m_ps_lsa_cv` and guarded by `m_ps_lsa_mutex`
- and  only one transaction is able to acquire the mutex
- and execute the condition lambda body:
  - check value
  - if not happy with the calculated value, do the calculation
  - then check value again
    - either happy and return
      - after unlocking the unique lock, it should notify again all other potentially waiting transactions which will then first check the up-to-date calculated value and return or do the calculation again (indeed here is the potential that each transaction that is unhappy with the value will end up calculating the value again, but I see this as more of a corner case)
    - or not happy and continue waiting

Let's leave the code as you proposed and see how it behaves.
We'll change it later with another refactoring if needed.
My idea was to avoid introducing yet another flag variable, but maybe it is forced. :)

src/transaction/log_impl.h
```diff
@@ -720,7 +721,8 @@ struct log_global
 
   std::mutex m_ps_lsa_mutex;
   std::condition_variable m_ps_lsa_cv;
-  LOG_LSA m_max_ps_flushed_lsa = NULL_LSA;
+  std::atomic<bool> m_ps_lsa_up_to_date;
+  LOG_LSA m_ps_consensus_flushed_lsa; // The quorum (number of the majority) of PS have done flushing log recrods until this.
```
cristiarg: I think these can be private
```suggestion
private:
  std::atomic<bool> m_ps_lsa_up_to_date;
  LOG_LSA m_ps_consensus_flushed_lsa; // The quorum (number of the majority) of PS have done flushing log recrods 
  until this.
public:
```

src/transaction/log_page_buffer.c
```diff
@@ -4499,11 +4496,9 @@ logpb_flush_pages (THREAD_ENTRY * thread_p, const LOG_LSA * flush_lsa)
...
       if (get_server_type () == SERVER_TYPE_TRANSACTION && ts_Gl->is_page_server_connected ())
 	{
-	  log_Gl.wait_flushed_lsa (*flush_lsa);
-	  if (prm_get_bool_value (PRM_ID_ER_LOG_COMMIT_CONFIRM))
-	    {
-	      _er_log_debug (ARG_FILE_LINE, "Page server committed LSA = %lld|%d.\n", LSA_AS_ARGS (&log_Gl.m_max_ps_flushed_lsa));
-	    }
+          quorum_consenesus_er_log ("Wait until page server flushed LSA gets larger than or equal to %lld|%d.\n", LSA_AS_ARGS (flush_lsa));
+	  
+          log_Gl.wait_for_ps_flushed_lsa (*flush_lsa);
```
cristiarg, line 4501: I am not sure the previous "Page server committed LSA" is reflected in the new code.
There is the detailed log message in `compute_consensus_lsa`. But someone looking into the debug log, would have to corroborate the message logged here with the one logged in `compute_consensus_lsa` to deduce whether the thread was able to move on or not.

cristiarg (approved):

src/transaction/log_page_buffer.c
wrlawodms: It is moved into "wait_for_ps_flushed_lsa()" to not hold the mutex.
`quorum_consenesus_er_log ("Quorum satisfied: page server flushed LSA = %lld|%d.\n", LSA_AS_ARGS (&log_Gl.m_ps_consensus_flushed_lsa));`

PR #4151, opened 2023-02-28
[CBRD-24614] Fix csql msg file caused by CBRD-24614

Srltas:
http://jira.cubrid.org/browse/CBRD-24614


kisoo-han (approved):
I confirmed that this file was encoded with CP51949.

PR #4152, opened 2023-02-28
[LETS-570] [LETS-663] crosscheck vpid inside and outside atomic replication sequence

cristiarg:
http://jira.cubrid.org/browse/LETS-570
http://jira.cubrid.org/browse/LETS-663

Use existing vpid bookkeeping structure inside atomic replication helper to cross check for out-of-order application of log records between log records which are part of atomic replications sequences and log records which are not part of.

Other:
- replaced the use of `log_rv_redo_record_sync_or_dispatch_async` for atomic replication with a more direct function; as a consequence, in `replicator` made some members private (this change goes naturally with the addition of the check in `atomic_replicator::read_and_redo_record`)


joohok (approved):

wrlawodms (approved):
Q. When is the macro `ATOMIC_REPL_PAGE_BELONGS_TO_SINGLE_ATOMIC_SEQUENCE_CHECK` set? Is it turned on with all tests?

cristiarg:
> Q. When is the macro `ATOMIC_REPL_PAGE_BELONGS_TO_SINGLE_ATOMIC_SEQUENCE_CHECK` set? Is it turned on with all tests?

It is hard-coded in the header:
```
#define ATOMIC_REPL_PAGE_BELONGS_TO_SINGLE_ATOMIC_SEQUENCE_CHECK
```
For now, it is always enabled because it has effect even in release builds - due to `assert_release` in `vpid_bookeeping::check_vpid_not_part_of_any_sequence`. 
In the future it can be disabled when no longer needed.
As explained previously, I do not have too much faith that this approach is the correct one.
I expected us to hit into these `assert_release` in `vpid_bookeeping` when we start testing with large data or specific scenarios that involve larger atomic sequences.
And when those asserts are hit, the only reason is to re-visit the way the atomic sequences are generated on the Active Transaction Server by really enforcing the atomic sequences when log is generated.

PR #4156, opened 2023-03-03
[CBRD-24542] Disable schema loading in loaddb online

airnet73:
http://jira.cubrid.org/browse/CBRD-24542

Purpose
When running online mode loaddb (cubrid loaddb -C) in HA environment DDL replication does not occur to slave nodes.
For this reason, the online mode loaddb is modified so that only object loading is possible.

Implementation
In CS Mode, the following options are not available.
> --load-only
> -i, --index-file
> -s, --schema-file
> --schema-file-list
> -t, --trigger-file

Remarks
N/A


src/loaddb/load_db.c
```diff
@@ -186,7 +186,10 @@ ldr_validate_object_file (const char *argv0, load_args * args)
...
     }
 
-  if (args->schema_file.empty () == false && args->schema_file_list.empty () == false)
+  if (args->cs_mode && (args->load_only == true
+			|| args->index_file.empty () == false
+			|| args->schema_file.empty () == false
+			|| args->schema_file_list.empty () == false || args->trigger_file.empty () == false))
     {
       util_log_write_errid (MSGCAT_UTIL_GENERIC_INVALID_ARGUMENT);
       load_usage (argv0);
```
kisoo-han: 1. I think you shouldn't treat these errors together. 
2. If an option other than -d is used in loaddb of CS mode, the following message should be displayed. 
3. Since the syntax is not wrong, there is no need to print the usage.
`ERROR: In loaddb CS mode, only object loading is possible.`
Sample Code:
```
if (args->schema_file.empty () == false && args->schema_file_list.empty () == false)
  {
    util_log_write_errid (MSGCAT_UTIL_GENERIC_INVALID_ARGUMENT);
    load_usage (argv0);
    return ER_FAILED;
  }

if (args->cs_mode && (args->load_only == true
 			|| args->index_file.empty () == false
 			|| args->schema_file.empty () == false
 			|| args->schema_file_list.empty () == false || args->trigger_file.empty () == false))
  {
    fprintf (stderr, "ERROR: In loaddb CS mode, only object loading is possible.\n"); // this is just sample
    return ER_FAILED;
  }
```

kisoo-han (approved):

src/loaddb/load_db.c
airnet73: Thank you for your advice.

@kisoo-han 
I wrote how to use CS mode in Usage. What do you think?

Usage (utils.msg file)
-C, --cs-mode   client-server mode execution \ n \
                          **Warning: in Client-Server Mode, Only Data File (-D, -Data-file = file) can be loaded.** \n\

kisoo-han (approved):

eido5 (approved):

PR #4158, opened 2023-03-03
[CBRD-24626] Supports an alternative way when the ping command is blocked in the HA

hornetmj:
http://jira.cubrid.org/browse/CBRD-24626

Purpose
* ping check doesn't work if ICMP protocol is disabled in user environment
* in this case, additional option(candidate) is neccessary
* triage decides to support tcp ping for this

Implementation
* implement based on the ping check implementation

Remarks
* see the test scenarios attached in the jira if you needed

### DIFF
--- a/src/base/system_parameter.c
+++ b/src/base/system_parameter.c
@@ -379,6 +379,8 @@ static const char sysprm_ha_conf_file_name[] = "cubrid_ha.conf";
 
 #define PRM_NAME_HA_TCP_PING_HOSTS "ha_tcp_ping_hosts"
 
+#define PRM_NAME_HA_PING_TIMEOUT "ha_ping_timeout"
+
 #define PRM_NAME_HA_APPLYLOGDB_RETRY_ERROR_LIST "ha_applylogdb_retry_error_list"
 
 #define PRM_NAME_HA_APPLYLOGDB_IGNORE_ERROR_LIST "ha_applylogdb_ignore_error_list"
@@ -1455,6 +1457,12 @@ static unsigned int prm_ha_ping_hosts_flag = 0;
 const char *PRM_HA_TCP_PING_HOSTS = "";
 static const char *prm_ha_tcp_ping_hosts_default = NULL;
 static unsigned int prm_ha_tcp_ping_hosts_flag = 0;
+
+int PRM_HA_PING_TIMEOUT = 1000;
+static int prm_ha_ping_timeout_default = 1000;
+static int prm_ha_ping_timeout_upper = INT_MAX;
+static int prm_ha_ping_timeout_lower = 0;
+static unsigned int prm_ha_ping_timeout_flag = 0;
 
 int *PRM_HA_APPLYLOGDB_RETRY_ERROR_LIST = int_list_initial;
 static bool *prm_ha_applylogdb_retry_error_list_default = NULL;
@@ -6189,6 +6197,18 @@ SYSPRM_PARAM prm_Def[] = {
    (void *) &prm_ha_tcp_ping_hosts_default,
    (void *) &PRM_HA_TCP_PING_HOSTS,
    (void *) NULL, (void *) NULL,
+   (char *) NULL,
+   (DUP_PRM_FUNC) NULL,
+   (DUP_PRM_FUNC) NULL},
+  {PRM_ID_HA_PING_TIMEOUT,
+   PRM_NAME_HA_PING_TIMEOUT,
+   (PRM_FOR_CLIENT | PRM_RELOADABLE | PRM_FOR_HA | PRM_HIDDEN),
+   PRM_INTEGER,
+   &prm_ha_ping_timeout_flag,
+   (void *) &prm_ha_ping_timeout_default,
+   (void *) &PRM_HA_PING_TIMEOUT,
+   (void *) &prm_ha_ping_timeout_upper,
+   (void *) &prm_ha_ping_timeout_lower,
    (char *) NULL,
    (DUP_PRM_FUNC) NULL,
    (DUP_PRM_FUNC) NULL}
--- a/src/base/system_parameter.c
+++ b/src/base/system_parameter.c
@@ -377,6 +377,8 @@ static const char sysprm_ha_conf_file_name[] = "cubrid_ha.conf";
 
 #define PRM_NAME_HA_PING_HOSTS "ha_ping_hosts"
 
+#define PRM_NAME_HA_TCP_PING_HOSTS "ha_tcp_ping_hosts"
+
 #define PRM_NAME_HA_APPLYLOGDB_RETRY_ERROR_LIST "ha_applylogdb_retry_error_list"
 
 #define PRM_NAME_HA_APPLYLOGDB_IGNORE_ERROR_LIST "ha_applylogdb_ignore_error_list"
@@ -1449,6 +1451,10 @@ static unsigned int prm_ha_max_heartbeat_gap_flag = 0;
 const char *PRM_HA_PING_HOSTS = "";
 static const char *prm_ha_ping_hosts_default = NULL;
 static unsigned int prm_ha_ping_hosts_flag = 0;
+
+const char *PRM_HA_TCP_PING_HOSTS = "";
+static const char *prm_ha_tcp_ping_hosts_default = NULL;
+static unsigned int prm_ha_tcp_ping_hosts_flag = 0;
 
 int *PRM_HA_APPLYLOGDB_RETRY_ERROR_LIST = int_list_initial;
 static bool *prm_ha_applylogdb_retry_error_list_default = NULL;
@@ -6175,6 +6181,17 @@ SYSPRM_PARAM prm_Def[] = {
    (char *) NULL,
    (DUP_PRM_FUNC) NULL,
    (DUP_PRM_FUNC) NULL},
+  {PRM_ID_HA_TCP_PING_HOSTS,
+   PRM_NAME_HA_TCP_PING_HOSTS,
+   (PRM_FOR_CLIENT | PRM_RELOADABLE | PRM_FOR_HA),
+   PRM_STRING,
+   &prm_ha_tcp_ping_hosts_flag,
+   (void *) &prm_ha_tcp_ping_hosts_default,
+   (void *) &PRM_HA_TCP_PING_HOSTS,
+   (void *) NULL, (void *) NULL,
+   (char *) NULL,
+   (DUP_PRM_FUNC) NULL,
+   (DUP_PRM_FUNC) NULL}
 };
 
 static int num_session_parameters = 0;
--- a/src/base/system_parameter.c
+++ b/src/base/system_parameter.c
@@ -1458,9 +1458,9 @@ const char *PRM_HA_TCP_PING_HOSTS = "";
 static const char *prm_ha_tcp_ping_hosts_default = NULL;
 static unsigned int prm_ha_tcp_ping_hosts_flag = 0;
 
-int PRM_HA_PING_TIMEOUT = 1000;
-static int prm_ha_ping_timeout_default = 1000;
-static int prm_ha_ping_timeout_upper = INT_MAX;
+int PRM_HA_PING_TIMEOUT = PRM_TCP_CONNECTION_TIMEOUT;
+static int prm_ha_ping_timeout_default = prm_tcp_connection_timeout_default;	/* NOTE: It is difficult to determine an accurate default value for TCP connection timeout, so the default value of the connection_time system parameter is followed. */
+static int prm_ha_ping_timeout_upper = INT_MAX / 1000;	/* divided by msecs */
 static int prm_ha_ping_timeout_lower = 0;
 static unsigned int prm_ha_ping_timeout_flag = 0;
 
--- a/src/executables/master_heartbeat.c
+++ b/src/executables/master_heartbeat.c
@@ -986,7 +986,7 @@ hb_cluster_job_check_ping (HB_JOB_ARG * arg)
     {
       for (ping_host = hb_Cluster->ping_hosts; ping_host; ping_host = ping_host->next)
 	{
-	  if (ping_host->port == 0)	// prm_get_string_value (PRM_ID_HA_PING_HOSTS) != NULL
+	  if (prm_get_string_value (PRM_ID_HA_PING_HOSTS))
 	    {
 	      ping_result = hb_check_ping (ping_host->host_name);
 	    }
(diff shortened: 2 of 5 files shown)


src/executables/master_heartbeat.c
```diff
@@ -970,7 +986,14 @@ hb_cluster_job_check_ping (HB_JOB_ARG * arg)
     {
       for (ping_host = hb_Cluster->ping_hosts; ping_host; ping_host = ping_host->next)
 	{
-	  ping_result = hb_check_ping (ping_host->host_name);
+	  if (ping_host->port == 0)	// prm_get_string_value (PRM_ID_HA_PING_HOSTS) != NULL
```
joohok:
```suggestion
	  if (prm_get_string_value (PRM_ID_HA_PING_HOSTS))
```

I think it is more readable

joohok: Also, this is for consistency with other code such as 1241L in `hb_cluster_job_demote ()`

src/executables/master_heartbeat.c
```diff
@@ -1363,7 +1410,14 @@ hb_cluster_check_valid_ping_server (void)
 
   for (ping_host = hb_Cluster->ping_hosts; ping_host; ping_host = ping_host->next)
     {
-      ping_host->ping_result = hb_check_ping (ping_host->host_name);
+      if (ping_host->port == 0)	// prm_get_string_value (PRM_ID_HA_PING_HOSTS) != NULL
```
joohok:
```suggestion
      if (prm_get_string_value (PRM_ID_HA_PING_HOSTS))
```

src/executables/master_heartbeat.c
```diff
@@ -4676,6 +4848,26 @@ hb_cluster_initialize (const char *nodes, const char *replicas)
...
     }
 
+  if (hb_Cluster->num_ping_hosts == 0)
+    {
+      /* The ha_tcp_ping_hosts can be used instead of the ha_ping_hosts in case the ICMP protocol is disabled.
+         However, both system parameters cannot be set at the same time and the ha_tcp_ping_hosts works only when the ha_ping_hosts is not set. */
+      hb_Cluster->num_ping_hosts = hb_cluster_load_tcp_ping_host_list (prm_get_string_value (PRM_ID_HA_TCP_PING_HOSTS));
+      hb_Cluster->ping_timeout = prm_get_integer_value (PRM_ID_HA_PING_TIMEOUT);
+
+      if (hb_cluster_check_valid_ping_server () == false)
```
joohok, line 4908: Inside the `hb_cluster_check_valid_ping_server()`, 
`hb_check_ping ()` should be replaced with `hb_check_tcp_ping ()` if it is called by tcp ping context.

hornetmj: Explain more for me. It already distinguish which version is called according to 'ping_host->port == 0'.

joohok: Oh.. Sorry, I just grep with `hb_cluster_check_valid_ping_server` and didn't find out that function has been modified.

src/base/system_parameter.c
```diff
@@ -1450,6 +1454,16 @@ const char *PRM_HA_PING_HOSTS = "";
 static const char *prm_ha_ping_hosts_default = NULL;
 static unsigned int prm_ha_ping_hosts_flag = 0;
 
+const char *PRM_HA_TCP_PING_HOSTS = "";
+static const char *prm_ha_tcp_ping_hosts_default = NULL;
+static unsigned int prm_ha_tcp_ping_hosts_flag = 0;
+
+int PRM_HA_PING_TIMEOUT = 1000;
```
joohok: Nit.
Other timeout-related parameters are set in seconds. 
To avoid confusion, I think it may be helpful to add a suffix like `IN_MSEC` to the parameter name to indicate that it is in milliseconds.

joohok (approved):

src/base/system_parameter.c
hornetmj: In the manual, it is a little bit inconsistent about you commented on as following: (see the lock_timeout and _seconds suffix)

- lock_timeout (msec)
- thread_connection_timeout_seconds (sec)
- thread_worker_timeout_seconds (sec)
- session_state_timeout (sec) 
- checkpoint_sleep_msecs (msec)
- group_commit_interval_in_msecs (msec)

But, I agree with your comments and I will apply it. And I want to change the default value from 1000 msec to 2000 msecs for tcp connection. Do you have any opinion about this?

joohok: I thought the reason that default value of this parameter is set to 1 second was because the timeout of the existing ping behavior was hard-coded to 1 second (PING_COMMAND_FORMAT : `"ping -w 1 -c 1 %s >/dev/null 2>&1; "`). However, if there is no need to maintain this, wouldn't it be better to change it to 5 seconds? I mean how about following the PRM_ID_TCP_CONNECTION_TIMEOUT?

src/executables/master_heartbeat.h
```diff
@@ -219,6 +219,7 @@ struct hb_ping_host_entry
   HB_PING_HOST_ENTRY **prev;
 
   char host_name[CUB_MAXHOSTNAMELEN];
+  int port;
```
wrlawodms: How about adding a comment that says it's just for TCP ping?

src/executables/master_heartbeat.h
```diff
@@ -261,6 +262,7 @@ struct hb_cluster
 
   HB_PING_HOST_ENTRY *ping_hosts;
   int num_ping_hosts;
+  int ping_timeout;
```
wrlawodms: How about adding a comment that says it's just for TCP ping?

src/executables/master_heartbeat.c
```diff
@@ -4676,6 +4848,26 @@ hb_cluster_initialize (const char *nodes, const char *replicas)
       return ER_FAILED;
     }
 
+  if (hb_Cluster->num_ping_hosts == 0)
+    {
+      /* The ha_tcp_ping_hosts can be used instead of the ha_ping_hosts in case the ICMP protocol is disabled.
+         However, both system parameters cannot be set at the same time and the ha_tcp_ping_hosts works only when the ha_ping_hosts is not set. */
```
wrlawodms: No need to leave an error logging when both are set? I guess it's worth leaving MASTER_ER_LOG_DEBUG at least.

src/base/system_parameter.c
hornetmj: @joohok The ha_ping_timeout will be used for the ping command some day. But now, it is good idea to me to follow the connection_timeout for the default value of tcp ping time.

src/executables/master_heartbeat.c
```diff
@@ -2182,6 +2280,80 @@ hb_cluster_load_ping_host_list (char *ha_ping_host_list)
...
+	  do
+	    {
+	      port_p++;
+	    }
+	  while (*port_p == ' ');
+
+	  if (*port_p == '-')
+	    {
+	      break;
+	    }
```
wrlawodms: What's it for?

src/executables/master_heartbeat.c
```diff
@@ -2182,6 +2280,80 @@ hb_cluster_load_ping_host_list (char *ha_ping_host_list)
...
+	}
+
+      if (strcmp (host_p, "0.0.0.0") == 0)
+	{
+	  snprintf (buf, sizeof (buf), "We do not allow 0.0.0.0 as a tcp ping hosts, excluded");
+	  MASTER_ER_SET (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_HB_NODE_EVENT, 1, buf);
+	}
+      else
+	{
+	  if (hb_add_tcp_ping_host (host_p, atoi (port_p)) == NULL)
```
wrlawodms: What happens if a user inputs a port as a string by accident like 1.1.1.1:3a3?

src/executables/master_heartbeat.c
```diff
@@ -6389,6 +6687,48 @@ hb_check_ping (const char *host)
...
+      if (are_hostnames_equal (host, node->host_name))
+	{
+	  /* PING Host is same as cluster's host name */
+	  snprintf (buf, sizeof (buf), "Useless TCP PING host name %s", host);
+	  MASTER_ER_SET (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_HB_NODE_EVENT, 1, buf);
+	  return HB_PING_USELESS_HOST;
+	}
+    }
+
+  sfd = css_tcp_client_open_with_timeout (host, port, timeout);
```
wrlawodms, line 6763: Doesn't it need a TCP host that listens and accepts on the port?

src/executables/master_heartbeat.c
```diff
@@ -2091,6 +2145,50 @@ hb_add_ping_host (char *host_name)
     {
       strncpy (p->host_name, host_name, sizeof (p->host_name) - 1);
       p->host_name[sizeof (p->host_name) - 1] = '\0';
+      p->port = 0;
```
wrlawodms: Nit.
```suggestion
      p->port = -1;
```
Because the TCP port ranges from 0 to 65535 even though the 0 is reserved.

wrlawodms (commented):
I'm rather asking questions than digging further because of the short notice of PR. :)

src/base/system_parameter.c
hornetmj: When I check the system parameters which is related to the connection time out (connection_timeout, thread_connection_timeout_seconds, etc) It would be better to set ping_timeout in sec units rather than msecs. In this case I could think of ha_ping_timeout_seconds but I prefer not to append suffix. please comment on.

src/executables/master_heartbeat.h
hornetmj: OK.

src/executables/master_heartbeat.h
hornetmj: OK.

src/executables/master_heartbeat.c
hornetmj: I consider it.

I designed ha_tcp_ping_hosts to be configured along with ha_ping_hosts. However, the ha_ping_hosts has high priority than the ha_tcp_ping_hosts and the ha_tcp_ping_hosts is ignored in this case. For reference, ping and tcp ping can be changed at runtime with the cubrid hb reload command.

src/executables/master_heartbeat.c
hornetmj: The purpose is to filter out negative port numbers like -22. However, in the process of reviewing, I found points for improvement. That part will be removed.

src/executables/master_heartbeat.c
hornetmj: 1.1.1.1:3a3 becomes host_p="1.1.1.1", port_p="3a3", and the port number becomes 3 by the atoi() function. If you look at the attached test scenario in jira, this is the intended behavior. However, in the process of reviewing, I found points for improvement. It is going to be corrected.

src/executables/master_heartbeat.c
hornetmj: The tcp ping(ping) check acts as a client. There is a server that allows tcp connection to the outside(I tested 22 ssh port), and cubrid operates as a client. Also, check the network status by connecting/disconnecting rather than checking by message communication.

src/executables/master_heartbeat.c
hornetmj: OK.

As you already know, 0 also checked by if (port < 1 || port > USHRT_MAX) in the hb_add_tcp_ping_host().

src/base/system_parameter.c
joohok: Agreed. 
I agree to use seconds as the unit of parameter, and I think there's no need to add the suffix to the parameter name.

src/executables/master_heartbeat.c
hornetmj: I will also edit the comment to reduce misunderstanding.

src/executables/master_heartbeat.c
```diff
@@ -2182,6 +2275,136 @@ hb_cluster_load_ping_host_list (char *ha_ping_host_list)
...
+	      goto error;
+	    }
+	}
+      else if (*port_p == ' ')
+	{
+	  i++;
+
+	  if (num_count > 1)	/* atoi("80 80") returns 80. It needs an exception handling */
+	    {
+	      is_space_after_numbers = true;
```
wrlawodms, line 2314: it's not used

src/executables/master_heartbeat.c
hornetmj: I just edit the comment. I thinks other log messages show enough information what kind of the system parameters are used. If you feel it is needed to leave a explicit error log when the both system parameters are set, then let me know. I will add it.

src/executables/master_heartbeat.c
wrlawodms: It means to just connect to any kind of TCP server listening on the port. I didn't think of this way. Thank you. 
However, this way depends on a random external application that listens on the port. In other words, the heartbeat is not reliable itself. It's different from ICMP in that the ICMP is designed for the purpose. Is it acceptable?

wrlawodms: If I'm correct, it seems pretty vulnerable so It should be used carefully. This should be described as a warning in the manual.

wrlawodms (approved):
Please check the code-style error.

src/executables/master_heartbeat.c
hornetmj: @wrlawodms The TCP ip/port is selected by the dba. If the unreliable ip/port is selected, then the system believes the unreliable result of a TCP ping even if there are no network issues. I will leave a commet(NOTE) in the manual.

src/executables/master_heartbeat.c
hornetmj: Do you mean is_space_after_numbers variable?

joohok:
@mhoh3963 Please merge it. Test failure is due to CBRD-24638 that is forced to be merged before. 
@ssihil  Please refer this PR, when you build 11.2.4 version.

src/executables/master_heartbeat.c
```diff
@@ -2182,6 +2275,136 @@ hb_cluster_load_ping_host_list (char *ha_ping_host_list)
...
+  bool is_space_after_numbers = false;
+
+  if (port_p == NULL || *port_p == '\0')
+    {
+      goto error;
+    }
+
+  while (*(port_p + i) != '\0')
+    {
+      if ((*port_p >= '0' && *port_p <= '9'))
```
hornetmj:
```suggestion
      if (*port_p >= '0' && *port_p <= '9')
```

hornetmj:
@mhoh3963 Please merge it. Test failure is due to CBRD-24638 that is forced to be merged before.

kisoo-han:
This PR also include backport of #4035 (execlude 0.0.0.0 from ping hosts) to release/11.2

PR #4161, opened 2023-03-06
[LETS-659] response broker blocking terminate

cristiarg:
http://jira.cubrid.org/browse/LETS-659

Response broker termination function is extended into a blocking call - `notify_terminate_and_wait`.

In order to allow for this extension to work correctly it was needed to keep track of en-route requests to the peer server. Up until now, an en-route request was not marked in any way. Extended the broker with a `register_request` call which actually add and entry in the appropriate broker bucket before actually dispatching the roundtrip request, thus, allowing the terminating logic to avoid race conditions.
Adapted and extended the unit test as a result of this behavioral change.

The sync client server is modified to use the blocking call.

Currently this function is only explicitly invoked on the Page Server side. With subsequent refactorings, the functionality will be used on Transaction Servers as well.

Other:
- reworded some comments

### DIFF
--- a/src/communication/response_broker.hpp
+++ b/src/communication/response_broker.hpp
@@ -57,12 +57,14 @@ namespace cubcomm
       response_broker &operator = (const response_broker &) = delete;
       response_broker &operator = (response_broker &&) = delete;
 
+      void register_request (response_sequence_number a_rsn);
       void register_response (response_sequence_number a_rsn, T_PAYLOAD &&a_payload);
       void register_error (response_sequence_number a_rsn, T_ERROR &&a_error);
 
       std::tuple<T_PAYLOAD, T_ERROR> get_response (response_sequence_number a_rsn);
 
       void terminate ();
+      void terminate_blocking ();
 
     private:
       struct bucket
@@ -76,18 +78,21 @@ namespace cubcomm
 	  bucket &operator = (const bucket &) = delete;
 	  bucket &operator = (bucket &&) = delete;
 
+	  void register_request (response_sequence_number a_rsn);
 	  void register_response (response_sequence_number a_rsn, T_PAYLOAD &&a_payload);
 	  void register_error (response_sequence_number a_rsn, T_ERROR &&a_error);
 
 	  std::tuple<T_PAYLOAD, T_ERROR> get_response (response_sequence_number a_rsn);
 
-	  void terminate ();
+	  void notify_terminate ();
+	  void notify_terminate_and_wait ();
 
 	private:
 	  struct payload_or_error_type
 	  {
 	    T_PAYLOAD m_payload;
 	    T_ERROR m_error;
+	    bool m_response_or_error_present = false;
 	  };
 
 	  using response_payload_container_type = std::unordered_map<response_sequence_number, payload_or_error_type>;
@@ -96,6 +101,7 @@ namespace cubcomm
 	  const T_ERROR m_no_error;
 	  const T_ERROR m_error;
 
+	  // mutex and condition variable protecting the response payloads container
 	  std::mutex m_mutex;
 	  std::condition_variable m_condvar;
 	  response_payload_container_type m_response_payloads;
@@ -133,6 +139,13 @@ namespace cubcomm
 
   template <typename T_PAYLOAD, typename T_ERROR>
   void
+  response_broker<T_PAYLOAD, T_ERROR>::register_request (response_sequence_number a_rsn)
+  {
+    get_bucket (a_rsn).register_request (a_rsn);
+  }
+
+  template <typename T_PAYLOAD, typename T_ERROR>
+  void
   response_broker<T_PAYLOAD, T_ERROR>::register_response (response_sequence_number a_rsn, T_PAYLOAD &&a_payload)
   {
     get_bucket (a_rsn).register_response (a_rsn, std::move (a_payload));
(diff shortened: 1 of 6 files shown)


src/communication/response_broker.hpp
```diff
@@ -179,8 +212,10 @@ namespace cubcomm
       std::lock_guard<std::mutex> lk_guard (m_mutex);
```
wrlawodms, line 218:
```suggestion
assert (m_response_payloads.find (a_rsn) != m_response_payloads.cend ());
```
Nit. to make sure that it's registered in advance.

src/communication/response_broker.hpp
```diff
@@ -195,61 +230,66 @@ namespace cubcomm
...
   response_broker<T_PAYLOAD, T_ERROR>::bucket::get_response (response_sequence_number a_rsn)
   {
     constexpr std::chrono::milliseconds millis_100 { 100 };
 
     {
-      typename response_payload_container_type::iterator found_it { m_response_payloads.end () };
+      std::unique_lock<std::mutex> ulock (m_mutex);
+
+      // nobody else but us should access this payload container but must be accessed locked
+      typename response_payload_container_type::iterator found_it = m_response_payloads.find (a_rsn);
```
wrlawodms, line 261:
```suggestion
      typename response_payload_container_type::iterator found_it = m_response_payloads.find (a_rsn);
      assert (found_it != m_response_payloads.cend ());
```
Nit.

src/communication/response_broker.hpp
```diff
@@ -195,61 +230,66 @@ namespace cubcomm
...
 	      return payload_or_error;
 	    }
+
+	  // NOTE: when terminate is invoked there is the possibility to implement a configurable behaviour:
+	  //  - either abort and return error - as done below
+	  //  - or continue waiting - with a timeout - until the request is serviced by the peer server
 	}
     }
 
+    // if here, terminate was specified and there is at most one thread to notify
```
wrlawodms, line 303: Don't we need to do `m_response_payloads.erase()` for the entry for which it waited?
The waiter waits until the `m_response_payloads` is empty, and it seems that an entry is not erased when terminating.

src/communication/response_broker.hpp
cristiarg: This is kind of given. There can't be any other thread/entity which touches that response payload. If there is - the whole server should come crashing down right away. So the assert adds no further value.

src/communication/response_broker.hpp
cristiarg: Same comment as above.

src/communication/response_broker.hpp
cristiarg: Yes. Thank you for this!
I have also adapted and extended the unit test for the addition of the `register_request` function.
Indeed, without emptying the bucket container, it would not work.
This is easy to see with the unit test - that's why I added an explicit unit test for this - "Test response broker without responses and with terminate/abort".

While developing, I tried to replicate this behavior by instrumenting the PS code with a call to `request_sync_client_server::stop_incoming_communication_thread` and logging - but I was not able to.

cristiarg:
For some reason, build fails because of missing `lz4.h`. It happened before....

src/communication/response_broker.hpp
```diff
@@ -166,11 +185,28 @@ namespace cubcomm
...
+    //    error is returned to the waiting thread
     assert (m_response_payloads.empty ());
   }
 
+  template <typename T_PAYLOAD, typename T_ERROR>
+  void
+  response_broker<T_PAYLOAD, T_ERROR>::bucket::register_request (response_sequence_number a_rsn)
+  {
+    {
+      std::lock_guard<std::mutex> lk_guard (m_mutex);
```
wrlawodms, line 201: What happens if `register_request()` is called while `notify_terminate_and_wait()` is waiting with the lock of m_mutex until the last request is digested? It seems that this will acquire the lock and register the request after `notify_terminate_and_wait()` confirms the container is empty and exits. Then, `m_response_payloads` ends up having some requests left. If it's the case.. , 
```suggestion
      std::lock_guard<std::mutex> lk_guard (m_mutex);
      if (m_terminate)
      {
        return;
      }
```

cristiarg: Bad things can happen.
However, I do not think it is the responsibility of this lower application layer to deal with this.
It would be the job of the connection handlers (instantiated by `page_server` and `tran_server` objects) to deal with the order of tearing down a connection.
Somewhere we have to draw a line. Not all layers have to deal with all aspects.
For consistency though, we can at least add an assert.
What do you think?

wrlawodms: Fair enough. Let's go with just an assert. 
> Somewhere we have to draw a line. Not all layers have to deal with all aspects. 

Good pointer. Thank you.

wrlawodms (approved):

hgryoo:
> For some reason, build fails because of missing `lz4.h`. It happened before....

I found that LZ4 was not set in the EP_TARGETS variable in CMake script. It can cause compile failures when you build in parallel (-j). The fix was merged into develop branch but not in the scalability_dev branch.
Please refer to the following issue link.
http://jira.cubrid.org/browse/CBRD-24627

I didn't test it in this branch. I hope this helps.

cristiarg:
@hgryoo 

Thanks for the heads-up. It certainly helps.
We'll get it in `scalability_dev` after our next merge from develop.

hornetmj (approved):

PR #4162, opened 2023-03-07
[CBRD-24636] Result error of order by limit statement in join query using USE_DESC_IDX hint

youngjinj:
http://jira.cubrid.org/browse/CBRD-24636

The following rules already exist.
However, `plan_un.join.outer` and `plan_un.join.inner` were not checked when `plan_type` was `QO_PLANTYPE_JOIN`.
So, the existing rule was changed to check `plan_un.join.outer` and `plan_un.join.inner`.

Add two rules to remove the `PT_HINT_USE_IDX_DESC` hint.
1. If MRO (Multi Range Optimization) is used and `plan->use_iscan_descending` is false, the `PT_HINT_USE_IDX_DESC` hint cannot be applied.
   ```
   create index i1_desc on t2 (c1, c2 desc, c3);

   select /*+ recompile use_desc_idx */ t.c2, t.c3
   from t0 inner join t2 as t on t0.c1 = t.c1
   where t.c1 in (1, 2)
   order by t.c2 desc, c3
   limit 4;
   ```
   ```
   <rewrite>
   select t.c2, t.c3
   from t0 t0, t2 t
   where ((t.c1= ?:0 ) or (t.c1= ?:1 )) and t0.c1=t.c1
   order by 1 desc , 2 for orderby_num()<= ?:2
   ```
2. Even if indexes are not scanned in descending order, if `oderby_skip` is true, the `PT_HINT_USE_IDX_DESC` hint should not be applied.
   ```
   create index i1_desc on t2 (c1, c2 desc);

   select /*+ recompile use_desc_idx */ t.c2
   from t0 inner join t2 as t on t0.c1 = t.c1
   where t.c1 = 1
   order by t.c2 desc;
   ```
   ```
   <rewrite>
   /* ---> skip ORDER BY */
   select t.c2
   from t0 t0, t2 t
   where t0.c1= ?:0  and t.c1= ?:1  and t0.c1=t.c1
   order by 1 desc
   ```


src/optimizer/query_planner.c
```diff
@@ -1005,8 +1005,27 @@ qo_top_plan_new (QO_PLAN * plan)
...
+			       * and the column order of order by are the same without reverse order.
+			       * So the PT_HINT_USE_IDX_DESC hint should not be applied.
+			       */
+
+			      PT_HINT_ENUM *hint = &(tree->info.query.q.select.hint);
+			      bool has_hint = (*hint & PT_HINT_USE_IDX_DESC) > 0;
+
+			      if (has_hint)
+				{
+				  *hint = (PT_HINT_ENUM) (*hint & ~PT_HINT_USE_IDX_DESC);
```
hgryoo: Equivalant with `tree->info.query.q.select.hint = (PT_HINT_ENUM) (tree->info.query.q.select.hint & ~PT_HINT_USE_IDX_DESC);`

youngjinj: Is the intention to reduce the use of temporary variables?

hgryoo: Maybe defining macro functions to perform bitwise operations for the PT_HINT_ENUM type would be better. But it is trivial. the current implementation seems enough.

youngjinj: I think we need to change the common code together through different issues.
I will think more and proceed.

shparkcubrid (approved):

ctshim (approved):

hgryoo (approved):

hyunikn (approved):

beyondykk9 (approved):

PR #4174, opened 2023-03-13
[LETS-660] Stop out-going msg generators outside explicitly.

wrlawodms:
http://jira.cubrid.org/browse/LETS-660

**Cause**
 The cause of the regression issue is calling the virtual function `stop_outgoing_page_server_messages` inside the connection_handler's destructor. It results in: ~tran_server() -> disconnect_all_page_servers() -> ~connection_handler() -> stop_outgoing_page_server_messages() so it ends up calling a virtual function in its destructor, which is problematic.
 
**Solution**
Call `stop_outgoing_page_server_messages()` not in the dtor, but explicitly outside before calling disconnect_all_page_servers().
- It's also better regarding code consistency because other threads are explicitly destroyed in this way, such as `finish_replication_during_shutdown()` and `finalize_request_responder()`.
- On abnormal finalization not through `xboot_shutdown_server()`, this also has to be called in probably `boot_server_all_finalize()`, but others are also not destroyed explicitly in this case. It should be dealt in another issue.

### DIFF
--- a/src/server/active_tran_server.hpp
+++ b/src/server/active_tran_server.hpp
@@ -34,7 +34,6 @@ class active_tran_server : public tran_server
 
     bool uses_remote_storage () const final override;
     MVCCID get_oldest_active_mvccid_from_page_server () const;
-    void stop_outgoing_page_server_messages () final override;
 
   private:
     class connection_handler : public tran_server::connection_handler
@@ -65,6 +64,7 @@ class active_tran_server : public tran_server
   private:
     bool get_remote_storage_config () final override;
 
+    void stop_outgoing_page_server_messages () final override;
     connection_handler *create_connection_handler (cubcomm::channel &&chn,
 	tran_server &ts) const final override;
 
--- a/src/server/active_tran_server.hpp
+++ b/src/server/active_tran_server.hpp
@@ -34,6 +34,7 @@ class active_tran_server : public tran_server
 
     bool uses_remote_storage () const final override;
     MVCCID get_oldest_active_mvccid_from_page_server () const;
+    void stop_outgoing_page_server_messages () final override;
 
   private:
     class connection_handler : public tran_server::connection_handler
@@ -64,7 +65,6 @@ class active_tran_server : public tran_server
   private:
     bool get_remote_storage_config () final override;
 
-    void stop_outgoing_page_server_messages () final override;
     connection_handler *create_connection_handler (cubcomm::channel &&chn,
 	tran_server &ts) const final override;
 
--- a/src/server/passive_tran_server.hpp
+++ b/src/server/passive_tran_server.hpp
@@ -44,7 +44,6 @@ class passive_tran_server : public tran_server
     log_lsa get_lowest_unapplied_lsa () const;
     void finish_replication_during_shutdown (cubthread::entry &thread_entry);
     void wait_replication_past_target_lsa (LOG_LSA lsa);
-    void stop_outgoing_page_server_messages () final override;
 
   private:
     class connection_handler : public tran_server::connection_handler
@@ -75,6 +74,7 @@ class passive_tran_server : public tran_server
     bool uses_remote_storage () const final override;
     bool get_remote_storage_config () final override;
 
+    void stop_outgoing_page_server_messages () final override;
     connection_handler *create_connection_handler (cubcomm::channel &&chn,
 	tran_server &ts) const final override;
 
--- a/src/server/passive_tran_server.hpp
+++ b/src/server/passive_tran_server.hpp
@@ -44,6 +44,7 @@ class passive_tran_server : public tran_server
     log_lsa get_lowest_unapplied_lsa () const;
     void finish_replication_during_shutdown (cubthread::entry &thread_entry);
     void wait_replication_past_target_lsa (LOG_LSA lsa);
+    void stop_outgoing_page_server_messages () final override;
 
   private:
     class connection_handler : public tran_server::connection_handler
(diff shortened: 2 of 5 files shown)


joohok (approved):

src/server/passive_tran_server.hpp
```diff
@@ -44,6 +44,7 @@ class passive_tran_server : public tran_server
     log_lsa get_lowest_unapplied_lsa () const;
     void finish_replication_during_shutdown (cubthread::entry &thread_entry);
     void wait_replication_past_target_lsa (LOG_LSA lsa);
+    void stop_outgoing_page_server_messages () final override;
```
cristiarg: Nit.
This could also be left private. Only the one in the base class must be made public.

src/server/active_tran_server.hpp
```diff
@@ -34,6 +34,7 @@ class active_tran_server : public tran_server
 
     bool uses_remote_storage () const final override;
     MVCCID get_oldest_active_mvccid_from_page_server () const;
+    void stop_outgoing_page_server_messages () final override;
```
cristiarg: Nit.
This could also be left private. Only the one in the base class must be made public.

src/server/tran_server.hpp
```diff
@@ -80,6 +80,9 @@ class tran_server
     bool is_page_server_connected () const;
     virtual bool uses_remote_storage () const;
 
+    // Before disconnecting page server, make sure no message is being sent anymore to the page server.
+    virtual void stop_outgoing_page_server_messages () = 0;
```
cristiarg, line 84: Only this one needs to be made public.

src/server/tran_server.cpp
cristiarg, line 1: @wrlawodms 
I would still remove from `tran_server::~tran_server` the call to `disconnect_all_page_servers` with this change.
Or is this the part that you were going to separate in a different issue/patch?

cristiarg (commented):
@wrlawodms 
I am still not convinced that the problem in http://jira.cubrid.org/browse/LETS-660 is fixed by this.
Have you tested?

wrlawodms:
> Have you tested? 

I didn't run the failed test case to test, but I arbitrarily made a situation "boot_restart_server" fails after "init_server_type" in it. It runs into the problem and causes the same core. It doesn't happen with this patch. The thing is that connections are established but not destroyed explicitly since `xboot_shutdown_server` is not called. It ends up trying to destroy the connections in finalize_server_type()->~tran_server() and causing the core.

src/server/tran_server.hpp
wrlawodms: Thanks!

src/server/tran_server.cpp
wrlawodms: Yes, it is. Initially, I considered removing it in the destructor and moving the call to `xboot_shutdown_server` to `finalize_server_type`, as connections are established in `init_server_type`, and I want to call it only once during shutdown. However, this approach has a potential problem because the `connection_handler` can access `log_Gl.m_prior_recver` in `receive_log_prior_list`, which is finalized in `xboot_shutdown_server`. This means that `disconnect_all_page_servers` must be called before `finalize_log_prior_receiver`, but we can't move `finalize_log_prior_receiver()` into `finalize_server_type` because it has another dependency on `log_final()`. With this PR, the issue is not fixed either, but I think it should be addressed in another issue.

Given the above problem, we should consider some additional factors:

- PS has the same dependency problem.
- PS's `disconnect_all_tran_server` has an additional dependency on `finalize_request_responder()` because `~connection_handler()` accesses the request responder.
- All threads created during boot have to be explicitly destroyed, even if it is not through `xboot_shutdown_server()`, including `disconnect_all_tran_server`.

Now that the topic has been brought up, I think it would be better to create a separate issue after discussing the matter. What do you think?

cristiarg: Indeed.
The purpose of this PR is to just fix the tests.

And I do know that there is a "chicken & egg" problem between xTS and PS servers regarding tearing down connection.

As a rule-of-thumb, I think we have either of the two solutions in the medium term:
- either implement a keep-alive connection ping, which, if not received in a timely fashion will tear down the connection to the unresponsive peer server from the local one (this has the potential to be superseeded by the future implementation of the cluster management software/heartheat implementation)
- decide on a protocol:
  - xTS is the sole initiator of the disconnection routine (even when requested pro-actively by the PS); because, as of now, PS can work without any xTS while xTS cannot work without at least one PS
  - on the PS side, we can, as a last resort, add a time-out (minutes/ tens of minutes) after which the PS will just delete all infrastructure related with communication with a xTS (again subject of being superseeded by the cluster management software/heartbeat implementation)

That is why this is not urgent - even it is not urgent as a separate issue. If we can continue to develop all features with a minimal teardown infrastructure, we will do so. And once we have a clearer image of what is needed - with hearbeat implementation, at first - it is simple to upgrade/downgrade the existing protocol.

cristiarg:
> I didn't run the failed test case to test, but I arbitrarily made a situation "boot_restart_server" fails after "init_server_type" in it. It runs into the problem and causes the same core. It doesn't happen with this patch. The thing is that connections are established but not destroyed explicitly since xboot_shutdown_server is not called. It ends up trying to destroy the connections in finalize_server_type()->~tran_server() and causing the core.

I will run the tests and update here. ...

cristiarg:
Later update.
I ran the tests on the feature branch and the cores are no more.

cristiarg (approved):

hornetmj (approved):

src/server/tran_server.cpp
wrlawodms: OK. Let me just get this straight. The current implementation has potential problems, but this has a dependency on future implementation and design, so would be changed. So, for now, let's just fix what's needed at least for tests, and revisit later. Do I get it right? Then, I'm not going to create a separate issue and move on with this in mind.

PR #4184, opened 2023-03-17
[CBRD-24701] Modify packer class not to call abort in check_range() function

beyondykk9:
http://jira.cubrid.org/browse/CBRD-24701

It is related to [CBRD-24696](http://jira.cubrid.org/browse/CBRD-24696). Changing the behavior of check_range() n packer/unpacker raises abort is difficult. However, At least a FATAL ERROR message should be written to let users know why cub_server is shut down.


src/base/packer.cpp
```diff
@@ -42,8 +42,40 @@ namespace cubpacking
...
+    const char *msglog;
+    char msg[LINE_MAX];
+
+    va_list ap;
+
+    va_start (ap, fmt);
+
+    /* call er_set() to print call stack to the log */
+    vsnprintf (msg, LINE_MAX, fmt, ap);
+    er_set (ER_FATAL_ERROR_SEVERITY, ARG_FILE_LINE, ER_NET_DATASIZE_MISMATCH, 1, fmt);
```
hgryoo: ER_NET_DATASIZE_MISMATCH requires two integer sizes.

beyondykk9: I changed error code in the next commit.

hgryoo (approved):

PR #4189, opened 2023-03-17
[CBRD-24643] fix the replication failure when a large number of records are inserted by loaddb with -C option

hornetmj:
http://jira.cubrid.org/browse/CBRD-24643

Purpose
* fix the replication failure when a large number of records are inserted by loaddb with -C
* some of the records aren't replicated from master to slave
  * the applylogdb is crashed by assert on debug mode
* the reason for this is described on the jira issue
```
locator_multi_insert_force()
-> locator_insert_force (use_bulk_logging==true)
// each physical log is replaced by page log (see pgbuf_log_redo_new_page)
// in this case, the replication log cannot have lsa for the physical log for each insert
```

Implementation
* loaddb should not be call the locator_multi_insert_force()
  * because this function prevents proper logging of replication log

Remarks
* N/A

### DIFF
--- a/src/loaddb/load_server_loader.cpp
+++ b/src/loaddb/load_server_loader.cpp
@@ -700,7 +700,7 @@ namespace cubload
 
     insert_errors_filtered = has_errors_filtered_for_insert (m_session.get_args().m_ignored_errors);
 
-    if (insert_errors_filtered)
+    if (insert_errors_filtered || !HA_DISABLED ())
       {
 	// In case of possible errors filtered for insert we disable the unique optimization
 	for (size_t i = 0; i < m_recdes_collected.size (); i++)


joohok (approved):

wrlawodms:
It holds BU_LOCK which is more conservative than IX_LOCK which is sufficient for record insertion so it blocks all concurrent record reads and does operations like it's not mvcc ops. It's occurred because it's filtered with exceptions while trying to bulk insertion. I think it would be the more proper way not to initiate bulk insertion in the first place if it's possible. What do you think? Is it complicated just to trigger normal insertion with HA mode?

wrlawodms (commented):
It seems to hold the BU_LOCK which is more conservative than IX_LOCK which is sufficient for record insertion so it blocks all concurrent record reads and does operations like it's not mvcc ops. It's occurred because it's filtered with exceptions while trying to bulk insertion. I think it would be the more proper way not to initiate bulk insertion in the first place if it's possible. What do you think? Is it complicated just to trigger normal insertion in the case of HA mode?

hornetmj:
@wrlawodms I thinks it is a design problem(choice). If BU_LOCK is changed to IX_LOCK, then the system get the concurrency. But, loaddb(transaction) and user transaction should contend to hold the page latch. And any transactions which have old snapshot should read (many) pages/records inserted by loaddb even though the records are obviously unreadable. So, in my opinion, it would be better to leave it as it is now to fit loaddb's bulky insert.

wrlawodms:
@hornetmj 
> @wrlawodms I thinks it is a design problem(choice). If BU_LOCK is changed to IX_LOCK, then the system get the concurrency. But, loaddb(transaction) and user transaction should contend to hold the page latch. And any transactions which have old snapshot should read (many) pages/records inserted by loaddb even though the records are obviously unreadable. So, in my opinion, it would be better to leave it as it is now to fit loaddb's bulky insert.

agreed.

src/loaddb/load_server_loader.cpp
```diff
@@ -700,7 +700,7 @@ namespace cubload
 
     insert_errors_filtered = has_errors_filtered_for_insert (m_session.get_args().m_ignored_errors);
 
-    if (insert_errors_filtered)
+    if (insert_errors_filtered || !HA_DISABLED ())
```
wrlawodms, line 703: I'd leave a comment for the case of `!HA_DISABLED ()`.

wrlawodms (approved):

src/loaddb/load_server_loader.cpp
hornetmj: Where is the comment?

wrlawodms: It's a suggestion. How about leaving a comment for this?

hornetmj: Oh. I see.

But I think it is too long story to explain using comments. It would better read the description of jira issue. I will think about the short comment before merging. Thanks.

PR #4195, opened 2023-03-22
[CBRD-24711] fix to recognize case for database name used as section name in cubrid.conf

kisoo-han:
http://jira.cubrid.org/browse/CBRD-24711

**Purpose**
* CUBRID recognizes **_demodb_** and **_DEMODB_** as different database names.
* database names can be used as a section name in cubrid.conf in the form of
`[@db_name]`
* ini parser, a MIT License library parser, converts all section names to lower case when it starts
* ini parser, a MIT License library parser, compare with ignore case option for the keys

**Implementation**

**Remarks**
* ini_parser.c, ini_parser.h **cannot pass CI License check**
* I think **force merge** is required due to CI License check.

### DIFF
--- a/src/base/ini_parser.c
+++ b/src/base/ini_parser.c
@@ -787,7 +787,7 @@ ini_hassec (const char *key)
  * Note:
  */
 int
-ini_seccmp (const char *key1, const char *key2)
+ini_seccmp (const char *key1, const char *key2, bool ignore_case)
 {
   const char *s1 = strchr (key1, ':');
   const char *s2 = strchr (key2, ':');
@@ -816,7 +816,12 @@ ini_seccmp (const char *key1, const char *key2)
       return 0;
     }
 
-  if (strncasecmp (key1, key2, key1_sec_len) == 0)
+  if (ignore_case && strncasecmp (key1, key2, key1_sec_len) == 0)
+    {
+      return key1_sec_len;
+    }
+
+  if (ignore_case == false && strncmp (key1, key2, key1_sec_len) == 0)
     {
       return key1_sec_len;
     }
--- a/src/base/ini_parser.c
+++ b/src/base/ini_parser.c
@@ -517,7 +517,11 @@ ini_parse_line (char *input_line, char *section, char *key, char *value)
 	{
 	  sprintf (section, "%c%s", leading_char, ini_str_trim (section + 1));
 	}
-      strcpy (section, ini_str_lower (section));
+
+      if (leading_char != '@')
+	{
+	  strcpy (section, ini_str_lower (section));
+	}
       status = LINE_SECTION;
     }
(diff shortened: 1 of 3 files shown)


airnet73 (approved):

src/base/ini_parser.c
```diff
@@ -517,7 +517,11 @@ ini_parse_line (char *input_line, char *section, char *key, char *value)
 	{
 	  sprintf (section, "%c%s", leading_char, ini_str_trim (section + 1));
 	}
-      strcpy (section, ini_str_lower (section));
+
```
mhoh3963, line 520: If leading_char is %, it mean [%hostname].
I think that hostname also need to be case sensitive.
So, you should also check that leading_char is not %.

src/base/system_parameter.c
```diff
@@ -7186,33 +7187,33 @@ prm_read_and_parse_ini_file (const char *prm_file_name, const char *db_name, con
...
     {
-      error = prm_load_by_section (ini, "service", false, reload, prm_file_name, load_flags);
+      error = prm_load_by_section (ini, "service", false, reload, prm_file_name, load_flags, true);
     }
   if (error == NO_ERROR && !SYSPRM_LOAD_IS_IGNORE_HA (load_flags) && PRM_HA_MODE != HA_MODE_OFF
       && GETHOSTNAME (host_name, CUB_MAXHOSTNAMELEN) == 0)
     {
       snprintf (sec_name, LINE_MAX, "%%%s|*", host_name);
-      error = prm_load_by_section (ini, sec_name, true, reload, prm_file_name, load_flags);
+      error = prm_load_by_section (ini, sec_name, true, reload, prm_file_name, load_flags, true);
```
mhoh3963, line 7212: You should change to false,because the hostname of section is case-sensitive.

src/base/system_parameter.c
```diff
@@ -7186,33 +7187,33 @@ prm_read_and_parse_ini_file (const char *prm_file_name, const char *db_name, con
...
       if (error == NO_ERROR && getlogin_r (user_name, CUB_MAXHOSTNAMELEN) == 0)
 	{
 	  snprintf (sec_name, LINE_MAX, "%%%s|%s", host_name, user_name);
-	  error = prm_load_by_section (ini, sec_name, true, reload, prm_file_name, load_flags);
+	  error = prm_load_by_section (ini, sec_name, true, reload, prm_file_name, load_flags, true);
```
mhoh3963, line 7216: You should change to false,because the hostname of section is case-sensitive.

src/base/ini_parser.c
kisoo-han: I think hostname itself is case insensitive, localhost or LOCALhost are the same host.
Should I consider '%'?

mhoh3963: You are right. so you don't need to consider it.

mhoh3963 (approved):

PR #4196, opened 2023-03-22
[CBRD-24643] fix the replication failure when a large number of records are inserted by loaddb with -C option (#4189)

hornetmj:
http://jira.cubrid.org/browse/CBRD-24643

backport: #4189, #4206 (cherry-pick)

### DIFF
--- a/src/loaddb/load_server_loader.cpp
+++ b/src/loaddb/load_server_loader.cpp
@@ -700,7 +700,7 @@ namespace cubload
 
     insert_errors_filtered = has_errors_filtered_for_insert (m_session.get_args().m_ignored_errors);
 
-    if (insert_errors_filtered)
+    if (insert_errors_filtered || !HA_DISABLED ())
       {
 	// In case of possible errors filtered for insert we disable the unique optimization
 	for (size_t i = 0; i < m_recdes_collected.size (); i++)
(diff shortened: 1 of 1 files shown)


wrlawodms (approved):
I still think the comment for the new case is needed. The Jira issue could be inaccessible later and hard to track when more changes happen.

hornetmj:
> I still think the comment for the new case is needed. The Jira issue could be inaccessible later and hard to track when more changes happen.

I understand what you mean. However when we follow that approach, we will leave many comments to explain the reason even if there are small change in every issue. It could reduce code readability.

Anyway, I accept your feedback. I will leave the following comment:

> In HA mode, the LSA of the log record created by insert(loaddb) should be accurately delivered when creating the replication log record. The replication to the slave may fail if the locator_multi_insert_force() function is called. Because it sometimes creates the data page-based log record instead of the record-based log record. It means that the replication log cannot have an accurate LSA of each insert.

Do you agree with the comment?

wrlawodms:
> However when we follow that approach, we will leave many comments to explain the reason even if there are small change in every issue. It could reduce code readability.

I don't mean that comments are needed for every single change. I think the best is a self-describing code without any comments. However, in my opinion, this change needs a comment because no context surrounding it describes it. The surrounding code finds ignorable errors and filters them and the if-condition even describes the case. `!HA_DISABLED()` seems to come out of the sky.

> Do you agree with the comment?

You know better. I'm even happy with a shorter one.
Thank you for pondering my opinion.

PR #4201, opened 2023-03-22
[CBRD-24613] Implicit type conversion and typechecking with it

hyunikn:


hgryoo:
Please move the classes in serverapi to the parent folder to avoid git conflict.

hyunikn:
> Please move the classes in serverapi to the parent folder to avoid git conflict.

Conflicts usually happen when coworking. 
Is it especially hard to resolve this case of conflict?
It seems good to collect classes related to server api into a package.

hyunikn:
This PR is closed and reopend for another issue https://github.com/CUBRID/cubrid/pull/4203

PR #4222, opened 2023-03-27
[LETS-670] Allow the same saved_lsa as before

wrlawodms:
http://jira.cubrid.org/browse/LETS-670

In some cases, a PS can send the same saved_lsa as before. (logpb_write_append_pages_to_disk() -> logpb_send_flushed_lsa_to_ats()). PS sends the saved_lsa whenever it flushes log pages and there seem to be some cases in which log pages are flushed but the saved_lsa is not updated. I've opted in to allowing the same saved_lsa rather than to filtering out possibly complicated cases.

- Allow the same saved_lsa as before in `active_tran_server::connection_handler::receive_saved_lsa`
- Request to update the consensus flushed LSA only if It's updated.
- Revise the debug logging to identify who updates the saved_lsa.

### DIFF
--- a/src/server/active_tran_server.cpp
+++ b/src/server/active_tran_server.cpp
@@ -169,12 +169,15 @@ active_tran_server::connection_handler::receive_saved_lsa (page_server_conn_t::s
   assert (sizeof (log_lsa) == message.size ());
   std::memcpy (&saved_lsa, message.c_str (), sizeof (log_lsa));
 
-  assert (saved_lsa > get_saved_lsa ()); // increasing monotonically
-  m_saved_lsa.store (saved_lsa);
+  assert (saved_lsa >= get_saved_lsa ()); // PS can send the same saved_lsa as before in some cases
+  if (saved_lsa > get_saved_lsa ())
+    {
+      m_saved_lsa.store (saved_lsa);
+      log_Gl.wakeup_ps_flush_waiters ();
+    }
 
-  quorum_consenesus_er_log ("Received saved LSA = %lld|%d.\n", LSA_AS_ARGS (&saved_lsa));
-
-  log_Gl.wakeup_ps_flush_waiters ();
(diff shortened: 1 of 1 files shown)


joohok (approved):
I thought I had found the cause of the issue (partial append of log records), but after checking the code, it turned out that partial append was not the culprit. 

It is likely that the issue mentioned earlier (the case where the LSA that needs to be written to disk (flush_lsa) is the same as nxio_lsa, which causes the log not to be written to disk) is the real cause. 
If we can resolve this issue, I believe this problem will also be resolved. 

I will create an issue and proceed with it.

wrlawodms:
@joohok 
> If we can resolve this issue, I believe this problem will also be resolved.
I will create an issue and proceed with it.

I don't fully understand you. Do you mean that it happens because of a known issue and will be fixed there? and this PR is not needed?

joohok:
> @joohok
> 
> > If we can resolve this issue, I believe this problem will also be resolved.
> > I will create an issue and proceed with it.
> 
> I don't fully understand you. Do you mean that it happens because of a known issue and will be fixed there? and this PR is not needed?

Yes, I thought so. 
However, in conclusion, even if we solve the known issue, 
the issue mentioned in this PR will not be resolved. Sorry for the confusion ...🥲

-------------------
The following content is an inference, so you can skip it if you want.

- `saved_lsa` is only sent to the ATS when log pages are flushed. 
- Log pages must have at least one to-be-flushed log page (num_toflush) for log page flush to occur. 
- However, the situation where saved_lsa (= nxio_lsa) is not changed seems equivalent to saying that no log records have been added but they have been flushed.

While there are cases where added log records are not flushed (partial append, the known issue), I cannot find any cases where log records are not added but are still flushed.

cristiarg:
@joohok 
Thanks for your input.
I have added http://jira.cubrid.org/browse/LETS-671 that pings back here and describes a possible fix for this.
In my opinion we can consider the fix as done here for the regression.

cristiarg (approved):

hornetmj (approved):

PR #4225, opened 2023-03-27
[CBRD-24723] Allow special characters in shard password of cubrid_broker.conf

kisoo-han:
http://jira.cubrid.org/browse/CBRD-24723

Description
* allow special character for '**shard_db_password**'

Implementation
* ini_parse_line () at ini_parser.c parse cubrid.conf
* commonly, it discard after character '#'
* before return, if the subject keyword is 'shard_db_password', reparsing value using
`sscanf (line, "%[^=] = %[^ \t]", dummy, value)`
* it means, accept a string until it encounters the blank or tab key.

Remarks

### DIFF
--- a/src/base/ini_parser.c
+++ b/src/base/ini_parser.c
@@ -559,6 +559,16 @@ ini_parse_line (char *input_line, char *section, char *key, char *value)
       /* Generate syntax error */
       status = LINE_ERROR;
     }
+
+  if (strcmp (key, "shard_db_password") == 0 && strlen (value) > 0)
+    {
+      char dummy[512];
+
+      if (sscanf (line, "%[^=] = %[^ \t]", dummy, value) != 2)
+	{
+	  status = LINE_ERROR;
+	}
+    }
   return status;
 }
 


src/base/ini_parser.c
```diff
@@ -559,6 +559,16 @@ ini_parse_line (char *input_line, char *section, char *key, char *value)
       /* Generate syntax error */
       status = LINE_ERROR;
     }
+
+  if (strcmp (key, "shard_db_password") == 0 && strlen (value) > 0)
```
mhoh3963, line 563: It seems like a not good way to compare a specific key value (shard_db_password) in the function. 
Please consider modify it by the way ignoring the # inside single quotes.

kisoo-han:
**ini_parser** already accepts special characters. If you want to enter special characters, wrap them with " or ' character. For example:

SHARD_DB_USER           ='kshan2'
SHARD_DB_PASSWORD       ="shard1234!#@$"

* if you want use string s"hard1234!#@$ as password use `'s"hard1234!#@$'`
* if you want use string s'hard1234!#@$ as password use `"s'hard1234!#@$"`

No signle character escape is not supported using '\'. If you set password as
's\hard1234!#@$', actual password string will be `s\hard1234!#@$`

PR #4230, opened 2023-03-27
[CBRD-24724] retry proxy server connect 3 times, up to 1200 msec

kisoo-han:
http://jira.cubrid.org/browse/CBRD-24724

**Description**
* when a client want to connect proxy thru broker, it may fails because shard proxy is not ready at the moment
* This is because shard proxy need some time to connect to the database server

**Implementation**
* retry when available proxy connection is not found.
* retry unto 3 times, retry interval is 400 ms

**Remarks**
* I measured proxy availability using 200ms timeout.
* first 3 tries were failed, and 4th try to find proxy was successful, 
* On an average machine the proxy needs at least 600ms at boot time.

### DIFF
--- a/src/broker/broker.c
+++ b/src/broker/broker.c
@@ -206,6 +206,8 @@
 #define SOCKET_TIMEOUT_SEC	2
 #endif
 
+#define SHARD_CAS_WARMUP_MSEC	900
+
 /* server state */
 enum SERVER_STATE
 {
@@ -815,7 +817,7 @@ receiver_thr_f (void *arg)
 #if defined(LINUX)
   if (br_shard_flag == ON)
     {
-      SLEEP_MILISEC (0, 700);
+      SLEEP_MILISEC (0, SHARD_CAS_WARMUP_MSEC);
     }
 
   timeout = 5;
--- a/src/broker/broker.c
+++ b/src/broker/broker.c
@@ -206,8 +206,6 @@
 #define SOCKET_TIMEOUT_SEC	2
 #endif
 
-#define SHARD_CAS_WARMUP_MSEC	900
-
 /* server state */
 enum SERVER_STATE
 {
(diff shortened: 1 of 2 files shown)


src/broker/broker.c
```diff
@@ -813,6 +813,11 @@ receiver_thr_f (void *arg)
 #endif
 
 #if defined(LINUX)
+  if (br_shard_flag == ON)
```
mhoh3963: How about sleeping in broker.c:575 in main(),  rather than sleeping in the receiver_thr_f() function ?
And I would like to know why it sleeps 700 ms only in Linux environment.

kisoo-han: my intention is:
start receiver_thr_f, but it will sleep few msecs and shard_dispatch_thr will start immediately. If we put sleep at line 575, this will postpone both threads for few msecs and start shard_dispatch_thr_f as soon as receiver_thr_f starts

For the 700ms, it has no special meaning. It can be predicted that it will take some time as it has to run db login. It was desired that the value be less than 1 second. The actual application results were also positive, so it was used.

This process is necessary for very few special applications, and sleep itself may not be necessary.

Windows does not support SHARD, so I put the sleep code for Linux.

kisoo-han: I changed it as a macro, and changed value to 900. Because it only wait once at the shard broker startup,
it could not be burden. If you have opinion, comments, let me know it.

#define SHARD_CAS_WARMUP_MSEC 900

src/broker/broker_proxy_conn.c
```diff
@@ -371,6 +376,14 @@ broker_find_available_proxy (T_SHM_PROXY * shm_proxy_p)
 	  min_cur_client = cur_client;
 	}
     }
+
+  if (shm_proxy_p->num_proxy > 0 && fd < 0 && retry_count++ < PROXY_SVR_CON_RETRY_COUNT)
+    {
+      pthread_mutex_unlock (&proxy_conn_mutex);
```
mhoh3963, line 383: pthread_mutex_lock() only be used at not window.
So, you should use #if !defined(WINDOWS) .. #endif.

mhoh3963 (approved):
LGTM

src/broker/broker_proxy_conn.c
kisoo-han: You right! I did it.
Thank you for your comment!

PR #4239, opened 2023-03-29
[CBRD-24708] Fix type binding of parameterized PL/CSQL variable in Static SQL

hgryoo:
http://jira.cubrid.org/browse/CBRD-24708

Host variables, not character types, have a null expected domain, so the type binding of parameterized PL/CSQL variable name is not properly resolved.

In the following, the code is an integer type, so the expected domain is set as tp_null (null domain). In the db_push_values (setting host variable value when executing the query), If the domain is null, It just tries to coerce the value at the time. So I've changed the routine to bind host variable's type in PL/CSQL compiler module
```
create or replace function test_typebinding() return string as
    n char(3) := 'KOR';

    g varchar;
    c int := 10615;
    m varchar;
begin
    select gender, name into g, m from athlete where nation_code = n and code = c;
    return m;

exception
when no_data_found then
    return 'no_data_found';
when too_many_rows then
    return 'too_many_rows ';
when others then
    return 'others';
end;
```

### DIFF
--- a/src/jsp/com/cubrid/plcsql/compiler/SymbolStack.java
+++ b/src/jsp/com/cubrid/plcsql/compiler/SymbolStack.java
@@ -290,6 +290,10 @@ public class SymbolStack {
         return getDecl(DeclProc.class, name);
     }
 
+    DeclVar getDeclVar(String name) {
+        return getDecl(DeclVar.class, name);
+    }
+
     DeclFunc getDeclFunc(String name) {
         DeclFunc ret = getDecl(DeclFunc.class, name);
         if (ret == null) {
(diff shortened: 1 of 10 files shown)


hyunikn (approved):

src/jsp/com/cubrid/plcsql/compiler/SymbolStack.java
```diff
@@ -290,6 +290,10 @@ DeclProc getDeclProc(String name) {
         return getDecl(DeclProc.class, name);
     }
 
+    DeclVar getDeclVar(String name) {
```
hyunikn, line 293: This implementation of getDeclVar() cannot be used because an Id expression can also be a function call expression without any arguments (Oracle and CUBRID syntax allow it) For an Id, we have to look-up symbol table both for Ids and for functions. Exisiting getDeclForIdExpr() is for this purpose.
For example, we have to raise an semantic error in the following case:

```
...
    i int;
begin
    declare
        function i ...
    begin
        select ... where col = i;    -- error: i cannot be a function call in Static SQL in Oracle and CUBRID
    end;
    ...
end;
```

PR #4243, opened 2023-03-30
[CBRD-24721] Improve the sql logging in HA

hornetmj:
http://jira.cubrid.org/browse/CBRD-24721

Purpose
* make a limitation of the number of the sql log files generated by applylogdb when ha_enable_sql_logging=y
  * no limitation now
* support a way to set the path of this files
* for this, two system parameters ha_sql_log_path and ha_sql_log_keep_count are introduced

Implementation
* see the changed code

Remarks
* the case which the sql log file id overflows the maximum integer range will be handled in another PR in sub-task
* If reviewers see the attached test scenarios in the jira, then they can easily understand the changed behaviors of sql logging

### DIFF
--- a/src/base/system_parameter.c
+++ b/src/base/system_parameter.c
@@ -391,6 +391,10 @@ static const char sysprm_ha_conf_file_name[] = "cubrid_ha.conf";
 
 #define PRM_NAME_HA_SQL_LOG_MAX_SIZE_IN_MB "ha_sql_log_max_size_in_mbytes"
 
+#define PRM_NAME_HA_SQL_LOG_PATH "ha_sql_log_path"
+
+#define PRM_NAME_HA_SQL_LOG_KEEP_COUNT "ha_sql_log_keep_count"
+
 #define PRM_NAME_HA_COPY_LOG_MAX_ARCHIVES "ha_copy_log_max_archives"
 
 #define PRM_NAME_HA_COPY_LOG_TIMEOUT "ha_copy_log_timeout"
@@ -1482,6 +1486,16 @@ static unsigned int prm_ha_applylogdb_log_wait_time_in_secs_flag = 0;
 bool PRM_HA_SQL_LOGGING = false;
 static bool prm_ha_sql_logging_default = false;
 static unsigned int prm_ha_sql_logging_flag = 0;
+
+const char *PRM_HA_SQL_LOG_PATH = "";
+static char *prm_ha_sql_log_path_default = NULL;
+static unsigned int prm_ha_sql_log_path_flag = 0;
+
+int PRM_HA_SQL_LOG_KEEP_COUNT = 2;
+static int prm_ha_sql_log_keep_count_default = 2;
+static int prm_ha_sql_log_keep_count_upper = INT_MAX;
+static int prm_ha_sql_log_keep_count_lower = 0;
+static unsigned int prm_ha_sql_log_keep_count_flag = 0;
 
 int PRM_HA_SQL_LOG_MAX_SIZE_IN_MB = INT_MIN;
 static int prm_ha_sql_log_max_size_in_mb_default = 50;
@@ -6226,6 +6240,29 @@ SYSPRM_PARAM prm_Def[] = {
    (void *) &prm_statdump_force_add_int_max_default,
    (void *) &PRM_STATDUMP_FORCE_ADD_INT_MAX,
    (void *) NULL, (void *) NULL,
+   (char *) NULL,
+   (DUP_PRM_FUNC) NULL,
+   (DUP_PRM_FUNC) NULL},
+  {PRM_ID_HA_SQL_LOG_PATH,
+   PRM_NAME_HA_SQL_LOG_PATH,
+   (PRM_FOR_CLIENT | PRM_FOR_HA | PRM_HIDDEN),
+   PRM_STRING,
+   &prm_ha_sql_log_path_flag,
+   (void *) &prm_ha_sql_log_path_default,
+   (void *) &PRM_HA_SQL_LOG_PATH,
+   (void *) NULL, (void *) NULL,
+   (char *) NULL,
+   (DUP_PRM_FUNC) NULL,
+   (DUP_PRM_FUNC) NULL},
+  {PRM_ID_HA_SQL_LOG_KEEP_COUNT,
+   PRM_NAME_HA_SQL_LOG_KEEP_COUNT,
+   (PRM_FOR_CLIENT | PRM_FOR_HA | PRM_HIDDEN),
+   PRM_INTEGER,
+   &prm_ha_sql_log_keep_count_flag,
+   (void *) &prm_ha_sql_log_keep_count_default,
+   (void *) &PRM_HA_SQL_LOG_KEEP_COUNT,
+   (void *) &prm_ha_sql_log_keep_count_upper,
+   (void *) &prm_ha_sql_log_keep_count_lower,
    (char *) NULL,
    (DUP_PRM_FUNC) NULL,
    (DUP_PRM_FUNC) NULL}
--- a/src/base/system_parameter.c
+++ b/src/base/system_parameter.c
@@ -1493,8 +1493,8 @@ static unsigned int prm_ha_sql_log_path_flag = 0;
 
 int PRM_HA_SQL_LOG_KEEP_COUNT = 2;
 static int prm_ha_sql_log_keep_count_default = 2;
-static int prm_ha_sql_log_keep_count_upper = INT_MAX;
-static int prm_ha_sql_log_keep_count_lower = 0;
+static int prm_ha_sql_log_keep_count_upper = 5;
+static int prm_ha_sql_log_keep_count_lower = 2;
 static unsigned int prm_ha_sql_log_keep_count_flag = 0;
 
 int PRM_HA_SQL_LOG_MAX_SIZE_IN_MB = INT_MIN;
@@ -6245,7 +6245,7 @@ SYSPRM_PARAM prm_Def[] = {
    (DUP_PRM_FUNC) NULL},
   {PRM_ID_HA_SQL_LOG_PATH,
    PRM_NAME_HA_SQL_LOG_PATH,
-   (PRM_FOR_CLIENT | PRM_FOR_HA | PRM_HIDDEN),
+   (PRM_FOR_CLIENT | PRM_FOR_HA),
    PRM_STRING,
    &prm_ha_sql_log_path_flag,
    (void *) &prm_ha_sql_log_path_default,
@@ -6256,7 +6256,7 @@ SYSPRM_PARAM prm_Def[] = {
    (DUP_PRM_FUNC) NULL},
   {PRM_ID_HA_SQL_LOG_KEEP_COUNT,
    PRM_NAME_HA_SQL_LOG_KEEP_COUNT,
-   (PRM_FOR_CLIENT | PRM_FOR_HA | PRM_HIDDEN),
+   (PRM_FOR_CLIENT | PRM_FOR_HA),
    PRM_INTEGER,
    &prm_ha_sql_log_keep_count_flag,
    (void *) &prm_ha_sql_log_keep_count_default,
--- a/src/base/system_parameter.c
+++ b/src/base/system_parameter.c
@@ -393,7 +393,7 @@ static const char sysprm_ha_conf_file_name[] = "cubrid_ha.conf";
 
 #define PRM_NAME_HA_SQL_LOG_PATH "ha_sql_log_path"
 
-#define PRM_NAME_HA_SQL_LOG_KEEP_COUNT "ha_sql_log_keep_count"
+#define PRM_NAME_HA_SQL_LOG_MAX_COUNT "ha_sql_log_max_count"
 
 #define PRM_NAME_HA_COPY_LOG_MAX_ARCHIVES "ha_copy_log_max_archives"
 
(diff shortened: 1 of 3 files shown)


src/transaction/log_applier_sql_log.c
```diff
@@ -642,6 +686,10 @@ create_dir (const char *new_dir)
 	{
 	  if (mkdir (path, 0777) < 0)
 	    {
+	      er_stack_push ();
+	      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_ES_INVALID_PATH, 1, new_dir);
```
joohok: I think `ER_ES_INVALID_PATH` is used in external storage code (storage/es.c). So how about defining additional error code ? such as `ER_HA_LA_INVALID_PATH` ?

src/transaction/log_applier_sql_log.c
```diff
@@ -546,6 +562,11 @@ sl_write_sql (string_buffer & query, string_buffer * select)
   if (ftell (log_fp) >= SL_LOG_FILE_MAX_SIZE)
     {
       log_fp = sl_open_next_file (log_fp);
+
+      if (sql_log_keep_cnt != 0 && sl_Info.curr_file_id >= sql_log_keep_cnt)
```
joohok: I think `sql_log_keep_cnt` can not be 0, because system parameter will be set to the value larger than 1.
how about just adding the assert ? ( assert (sql_log_keep_cnt > 0))

hornetmj: sql_log_keep_cnt != 0 is for the previous spec. I will remove it. But I think that assert() is not needed because the lower and upper range check are already done in the system paramter module.

src/transaction/log_applier_sql_log.c
hornetmj: You pointed out the right thing. But I really don't want to add new error code. OTL

I will change it.

hornetmj:
@joohok @wrlawodms I didn't check the case when the relative path is set to the ha_sql_log_path. I will check it and update code and tc if needed.

src/transaction/log_applier_sql_log.c
hornetmj: How about use ER_HA_GENERIC_ERROR? SQL logger already use this like "HA generic: Failed to initialize SQL logger.".

joohok: If only the debug logs are well written, I think it would be okay to use ER_HA_GENERIC. If ER_HA_GENERIC is used, the return value below should also be changed. (return ER_HA_GENERIC_ERROR)

Also, in the function that calls create_dir, it does not check whether the directory has been created (in sl_init ()). I think it would be better to add the code to check if the directory has been created. 
```
if (create_dir (tmp_log_path) != NO_ERROR) ...
```

However, after checking other codes in log_applier_sql_log.c, 
it seems that the `sl_* ` functions do not set error code when error occur. 
I think it is appropriate to set the error where it occurs, but it seems that they all return ER_FAILED and then er_set is done in the caller. 
In this situation, I am wondering whether it is appropriate to only use `er_log_debug` to write debugging logs, or to set the error using `er_set`. What do you think?

src/transaction/log_applier_sql_log.c
```diff
@@ -171,15 +173,27 @@ sl_init (const char *db_name, const char *repl_log_path)
...
+      snprintf (sql_catalog_path, PATH_MAX, "%s/%s_applylogdb.sql.info", sql_log_path, db_name);
+    }
+  else
+    {
+      snprintf (tmp_log_path, PATH_MAX, "%s/sql_log", repl_log_path);
+      snprintf (sql_log_base_path, PATH_MAX, "%s/%s.sql.log", tmp_log_path, basename (basename_buf));
+      snprintf (sql_catalog_path, PATH_MAX, "%s/%s_applylogdb.sql.info", repl_log_path, db_name);
+    }
+
+  create_dir (tmp_log_path);
```
wrlawodms:
```suggestion

  sql_log_path = prm_get_string_value (PRM_ID_HA_SQL_LOG_PATH);
  if (sql_log_path == NULL)
    {
      sql_log_path = repl_log_path;
    }
    
    snprintf (tmp_log_path, PATH_MAX, "%s/sql_log", sql_log_path);
    snprintf (sql_log_base_path, PATH_MAX, "%s/%s.sql.log", tmp_log_path, basename (basename_buf));
    snprintf (sql_catalog_path, PATH_MAX, "%s/%s_applylogdb.sql.info", sql_log_path, db_name);
```

src/base/system_parameter.c
```diff
@@ -391,6 +391,10 @@ static const char sysprm_ha_conf_file_name[] = "cubrid_ha.conf";
 
 #define PRM_NAME_HA_SQL_LOG_MAX_SIZE_IN_MB "ha_sql_log_max_size_in_mbytes"
 
+#define PRM_NAME_HA_SQL_LOG_PATH "ha_sql_log_path"
+
+#define PRM_NAME_HA_SQL_LOG_KEEP_COUNT "ha_sql_log_keep_count"
```
wrlawodms:
```suggestion
#define PRM_NAME_HA_SQL_LOG_KEEP_COUNT "ha_sql_log_max_count"
```
NIT. "max" seems to me more intuitive and broadly used. at your choice.

src/transaction/log_applier_sql_log.c
```diff
@@ -546,6 +562,11 @@ sl_write_sql (string_buffer & query, string_buffer * select)
   if (ftell (log_fp) >= SL_LOG_FILE_MAX_SIZE)
     {
       log_fp = sl_open_next_file (log_fp);
+
+      if (sql_log_keep_cnt != 0 && sl_Info.curr_file_id >= sql_log_keep_cnt)
+	{
+	  sl_remove_oldest_file ();
```
wrlawodms:
```suggestion
	  (void) sl_remove_oldest_file ();
```
It might cause a warning. `sl_remove_oldest_file()` really needs to return an error code, by the way?

src/transaction/log_applier_sql_log.c
```diff
@@ -612,6 +633,29 @@ sl_open_next_file (FILE * old_fp)
...
+ * Note:
+ *   This function is related to the ha_sql_log_keep_count system parameter.
+ *   This system parameter can be set from 2 to 5 and only that number of sql log files are kept.
+ */
+static int
+sl_remove_oldest_file (void)
+{
+  int oldest_file_id;
+  char oldest_file_path[PATH_MAX];
+
```
wrlawodms, line 641:
```suggestion
assert (sl_Info.curr_file_id >= sql_log_keep_cnt);
```
As a pre-condition of this function.

src/transaction/log_applier_sql_log.c
wrlawodms: Or, if we should keep it returning an error code, I think we should do error handling even though now `sl_remove_oldest_file` always returns NO_ERROR. In my opinion, the caller should not expect or assume the return value of a callee.
```suggestion
	  if (sl_remove_oldest_file () != NO_ERROR)
	  {
	    return ER_FAILED;
	  }
```

src/transaction/log_applier_sql_log.c
```diff
@@ -642,6 +686,10 @@ create_dir (const char *new_dir)
 	{
 	  if (mkdir (path, 0777) < 0)
 	    {
+	      er_stack_push ();
```
wrlawodms: Is it for only printing the error without any side effects? Can't we just er_set() and return the error instead of ER_FAILED below?

src/transaction/log_applier_sql_log.c
hornetmj: When ha_sql_log_path is set, the failure of create_dir() is checked in the sl_read_catalog(). So, the error logs are written like:
```
Path for external storage '/root/sql_log' is invalid.
Cannot open SQL catalog file: Permission denied
HA generic: Failed to initialize SQL logger.
```
But, it would be better add an error checking code at the create_dir().

And your opinion looks like every error case should have explict error code. right? I don't aggree with it. Some error situations like logical operation(create dir) is considered as one error code even if there are many reason for the failure like invalid path, permission or non-directory. It seems to depend on how detailed the problem situation is to be shown depending on the importance of the error.

src/transaction/log_applier_sql_log.c
hornetmj: Not always. I will describe in the sub-task(http://jira.cubrid.org/browse/CBRD-24738).

src/transaction/log_applier_sql_log.c
hornetmj: Yes, that reason. But I don't completely understand the er_set() mechanism. I just follow the following codes and check that the error message is well written in the error log. I don't want to disturb main error stream.
```
8031   if (prm_get_bool_value (PRM_ID_HA_SQL_LOGGING))
8032     {
8033       if (sl_init (la_slave_db_name, log_path) != NO_ERROR)
8034         {
8035           er_stack_push ();
8036           er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_HA_GENERIC_ERROR, 1, "Failed to initialize SQL logger");
8037           er_stack_pop ();
8038         }
8039       else
8040         {
8041           la_enable_sql_logging = true;
8042         }
8043     }
```

src/transaction/log_applier_sql_log.c
```diff
@@ -171,15 +173,45 @@ sl_init (const char *db_name, const char *repl_log_path)
...
-  snprintf (sql_log_base_path, PATH_MAX, "%s/sql_log/%s.sql.log", repl_log_path, basename (basename_buf));
-  snprintf (sql_catalog_path, PATH_MAX, "%s/%s_applylogdb.sql.info", repl_log_path, db_name);
+
+  sql_log_path = prm_get_string_value (PRM_ID_HA_SQL_LOG_PATH);
+  if (sql_log_path == NULL)
+    {
+      sql_log_path = repl_log_path;
+    }
+
+  if (!IS_ABS_PATH (sql_log_path))
```
joohok: How about using `realpath()` ?

hornetmj: I explained why I don't use the realpath(). (The limitation of realpath() is that the path set in the ha_sql_log_path should be existed.) I consider one more time there is a good way to handle this issue.

hornetmj:
@wrlawodms I will re-request after test termination.

wrlawodms:
> @wrlawodms I will re-request after test termination.

Just click the re-request button when it's ready, please. I can see it.

src/transaction/log_applier_sql_log.c
```diff
@@ -612,44 +625,124 @@ sl_open_next_file (FILE * old_fp)
...
+	  if (access (path_buf, F_OK) < 0)
 	    {
-	      return ER_FAILED;
+	      if (mkdir (path_buf, 0777) < 0)
+		{
+		  snprintf (er_msg, sizeof (er_msg), "Failed to create SQL log directory \'%s\' (%s)", path_buf,
+			    strerror (errno));
+
+		  er_stack_push ();
+		  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_HA_GENERIC_ERROR, 1, er_msg);
```
wrlawodms:
```suggestion
		  snprintf (er_msg, sizeof (er_msg), "Failed to create SQL log directory \'%s\'", path_buf);

		  er_stack_push ();
		  er_set_with_oserror (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_HA_GENERIC_ERROR, 1, er_msg);
```
This is the CUBRID way and probably works.

src/transaction/log_applier_sql_log.c
```diff
@@ -612,44 +625,124 @@ sl_open_next_file (FILE * old_fp)
...
+  if (log_path == path_buf)
+    {
+      path_len = strlen (log_path);
+
+      snprintf (path_buf + path_len, path_buf_size - path_len, "%s%s", FILEIO_PATH_SEPARATOR (log_path), "sql_log");
+    }
+  else
+    {
+      snprintf (path_buf, path_buf_size, "%s%s%s", log_path, FILEIO_PATH_SEPARATOR (log_path), "sql_log");
     }
```
wrlawodms:
```suggestion
  snprintf (path_buf, path_buf_size, "%s%s%s", log_path, FILEIO_PATH_SEPARATOR (log_path), "sql_log");
```

wrlawodms: It'd probably work, and I think the simplicity and readability are worth the double-copy in this case.

src/transaction/log_applier_sql_log.c
```diff
@@ -612,44 +625,124 @@ sl_open_next_file (FILE * old_fp)
...
 
-  strcpy (path, new_dir);
+  if (strlen (path_buf) >= path_buf_size - 1)
```
wrlawodms: This won't happen since `snprintf` cut the argument's string and attach the NULL at the end of string if the size of format exceeds the given size.

wrlawodms: I guess this error handling should be placed before snprintf (path_buf, path_buf_size, "%s%s%s", log_path, FILEIO_PATH_SEPARATOR (log_path), "sql_log") with the calculated size;

src/transaction/log_applier_sql_log.c
hornetmj: I already do test it. It doesn't work.

src/transaction/log_applier_sql_log.c
hornetmj: I think that '>' makes you confusing. I wanted to verify the case where the path string created in path_buf exceeded PATH_MAX and was truncated.

src/transaction/log_applier_sql_log.c
hornetmj: This is a simple test. It prints out " World" not "Hello World".

```
#include <stdio.h>

int main()
{
        char buf[20] = "Hello";
        char *p;

        p = buf;

        snprintf(buf, 10, "%s%s", p, " World");

        printf ("%s\n", buf);

        return 0;
}
```

src/transaction/log_applier_sql_log.c
hornetmj: Apply it. Thanks.

wrlawodms (approved):

wrlawodms (approved):
discussed offline.

hornetmj:
> Just click the re-request button when it's ready, please. I can see it.

Yes, I will.

PR #4244, opened 2023-03-31
[CBRD-24096] A core dump occurs when cas writes a log after receiving a sigkill signal.

airnet73:
http://jira.cubrid.org/browse/CBRD-24096

Purpose
A core dump occurs when cas writes a log after receiving a sigkill signal.
The cause of a core dump is that a collision occurs while process and signal use one file descriptor at the same time.
To avoid conflicts, we log using a new file descriptor.

Implementation
N/A

Remarks
N/A

### DIFF
--- a/src/broker/cas.c
+++ b/src/broker/cas.c
@@ -1721,7 +1721,7 @@ cas_free (bool from_sighandler)
 	    }
 	  else
 	    {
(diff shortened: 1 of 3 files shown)


kisoo-han (approved):

kisoo-han:
include jira# into PR title

PR #4250, opened 2023-04-02
[LETS-666] add extra perfmon metric wether slotted pages need compacting

cristiarg:

Add extra information to the existing `PERF_SERVED_COMPR_` perf extended statistics for whether the the slotted pages need also compacting.

Implementation:
- added enum value `PERF_SERVED_COMPR_SLOTTED_NEEDS_COMPACT`
- when a page is served from Page Server to a Transaction Server, check whether the page is slotted, and - if it is - whether it needs to be compacted and - if specified via the `PERFMON_ACTIVATION_FLAG_SERVED_COMPR_PAGE_TYPE` - also log this information
- extended perfmon functions to also add this information

Context:
In the monolithic server, it is the job off the vacuum daemons to ensure that the slotted pages remain in an optimal state. The reason for this might be that, chronologically, the vacuum is the last entity to touch a modified page after all the other (the modifying transaction, potential older concurrent transactions that might also need older values, ...).
In the scalability architecture, the vacuum threads compacting the page - although currently still done - it is useless, as the Active Transaction Server pages, compacted or not, will just be evicted from the page buffer after a certain time.
In the scalability architecture, in order to have the on-permanent-storage data pages also in an optimal state, it is needed that the replication threads on Page Server perform this action because these threads are the only ones that actually modify the data pages.

### DIFF
--- a/src/base/perf_monitor.h
+++ b/src/base/perf_monitor.h
@@ -282,7 +282,7 @@ typedef enum
   PERF_SERVED_COMPR_COUNT = 0,
   /* per page type, the average ratio of pages compressed and served by page server */
   PERF_SERVED_COMPR_RATIO,
-  /* per page type, whether the page being server needs to be compacted or not; for non-slotted
+  /* per page type, whether the page being served needs to be compacted or not; for non-slotted
    * pages, the value is zero and irrelevant */
   PERF_SERVED_COMPR_SLOTTED_NEEDS_COMPACT,
 
--- a/src/base/perf_monitor.h
+++ b/src/base/perf_monitor.h
@@ -278,8 +278,12 @@ typedef enum
 
 typedef enum
 {
-  PERF_SERVED_COMPR_COUNT = 0,	/* per page type, this many pages have been compressed and served by page server */
-  PERF_SERVED_COMPR_RATIO,	/* per page type, the average ratio of pages compressed and served by page server */
+  /* per page type, this many pages have been compressed and served by page server */
+  PERF_SERVED_COMPR_COUNT = 0,
+  /* per page type, the average ratio of pages compressed and served by page server */
+  PERF_SERVED_COMPR_RATIO,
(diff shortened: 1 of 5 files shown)


src/base/perf_monitor.h
```diff
@@ -278,8 +278,13 @@ typedef enum
 
 typedef enum
 {
-  PERF_SERVED_COMPR_COUNT = 0,	/* per page type, this many pages have been compressed and served by page server */
-  PERF_SERVED_COMPR_RATIO,	/* per page type, the average ratio of pages compressed and served by page server */
+  /* per page type, this many pages have been compressed and served by page server */
+  PERF_SERVED_COMPR_COUNT = 0,
+  /* per page type, the average ratio of pages compressed and served by page server */
+  PERF_SERVED_COMPR_RATIO,
+  /* per page type, whether the page being server needs to be compacted or not; for non-slotted
```
hornetmj:
```suggestion
  /* per page type, whether the page being served needs to be compacted or not; for non-slotted
```

hornetmj (approved):

src/storage/page_buffer.c
```diff
@@ -8480,20 +8480,24 @@ void
...
   VPID vpid;
   vpid_utils::unpack (message_upk, vpid);
 
   LOG_LSA target_repl_lsa;
   cublog::lsa_utils::unpack (message_upk, target_repl_lsa);
+  // *INDENT-ON*
   int packed_fetch_mode;
-  message_upk.unpack_int (packed_fetch_mode);
-  const PAGE_FETCH_MODE fetch_mode = static_cast<PAGE_FETCH_MODE> (packed_fetch_mode);
+    message_upk.unpack_int (packed_fetch_mode);
```
wrlawodms:
```suggestion
  message_upk.unpack_int (packed_fetch_mode);
```

wrlawodms (approved):

PR #4257, opened 2023-04-05
[CBRD-24743] change broker_monitor.c - from ncurses to tinfo with dlopen

kisoo-han:
http://jira.cubrid.org/browse/CBRD-24743

**Purpose**
* removing ncurses dependancy from CUBRID binaries
* same way as we did for csql
* same behavior with previous 'cubrid broker status -s' but do not use ncurses and uses tinfo

**Implementation**
* run time binding of **tinfo** functions
  * tgoto (), tgetent (), tgetstr (), tputs ()
* rewrite curses to tinfo + alpha for tty handling

**Remarks**
* due to ncurses itself calls other functions, 
* it is not possible to convert current ncurses function calls of broker monitot to dlopen style, we tried
* we have an experience of converting libedit library for csql to dlopen style (libtinfo)

### DIFF
--- a/src/broker/broker_monitor.c
+++ b/src/broker/broker_monitor.c
@@ -545,7 +545,12 @@ main (int argc, char **argv)
 	}
 //  FillConsoleOutputCharacter(h_console, ' ', scr_info.dwSize.X * scr_info.dwSize.Y, top_left_pos, &size);
 #else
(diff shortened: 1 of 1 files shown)


kisoo-han:
will be reopen after rollback of #4118 for broker_monitor.c

PR #4261, opened 2023-04-05
[LETS-666] page server replication slotted page compaction

cristiarg:

Page Server replication also does slotted page compaction.

Implementation:
- added parameter to drive whether compaction is done or not (it is expected that the slotted page compaction will remain in the page server replication pipeline; the sysparam is temporary for investigative purposes)
- extended `log_rv_redo_context` - which is also used for replication as well as for recovery - to parameterize whether the compaction is needed or not
- specifically disable compaction for: recovery (in `log_recovery_redo`) and passive transaction server replication (in `atomic_replicator::atomic_replicator`)
- only activate slotted page compaction for page server replication - `page_server::start_log_replicator`
- actual implementation is performed upon page unfix - in `log_rv_redo_record_sync` (this function is called in 3 contexts: log recovery, page server replication, passive transaction server replication and each case is treated as described)


hornetmj (approved):

cristiarg:
Implemented approach is incorrect because it introduces changes which are not reflected by the transactional log and associated REDO functions. Thus it can introduce discrepancies which can lead to the following issues:
- a redo log record in a slotted page on page server can be done - because of sideline page compaction performed as part of replication
- the same redo log record replicated on a passive transaction server cannot be done - there can be situations where the physical layout of the slotted page is more fragmented and, thus, there is not enough continuous space to write the data

Reworking implementation.

cristiarg:
This patch is deemed incorrect because it introduces out-of-transactional-log compaction which causes different layouts of a slotted page across the same page being present in different servers' page buffers.

PR #4266, opened 2023-04-06
[CBRD-24743] change broker_monitor.c - from ncurses to tinfo functions

kisoo-han:
http://jira.cubrid.org/browse/CBRD-24743

**Purpose**
* removing ncurses version dependancy from CUBRID binaries
* same way as we did for csql
* same behavior with previous 'cubrid broker status -s' but do not use ncurses and uses tinfo

**Implementation**
* rewrite following curses funtions using tinfo functions
  * move ()
  * clrtobot ()
  * clrtoeol ()
  * endwin ()
  * addstr ()
* implementing the gtch() function without using ncurses
* dynamic binding of following tinfo functions at run time
  * tgetent ()
  * tgoto ()
  * tgetstr ()
  * tputs ()

**Remarks**
* CUBRID binaries will search libtinfo.so.x from the user's LIB paths
* CMakeLists.txt is just rolled back from previous 81046b
* **addstr** () will print as many lines as **LINES** from the beginning.
* Even if you resize your terminal, the command below will output the actual number of LINES and COLUMNS.
`$ echo $LINES $COLUMNS`
* And, tgetnum ("li") returns the actual line number of the current terminal.
* this PR can respond to **LINE** and **COLUMNS**
* this PR can respond to **terminal resize**.

### DIFF
--- a/src/broker/broker_monitor.c
+++ b/src/broker/broker_monitor.c
@@ -360,18 +360,23 @@ static void timeout (int delay);
 static void addstr (const char *);
 static int tty_noblock (void);
 static int get_timeout (void);
+static int num_newlines (const char *);
 
 typedef char *(*tgoto_func_t) (const char *cap, int col, int row);
 typedef int (*tgetent_func_t) (char *bp, const char *name);
 typedef char *(*tgetstr_func_t) (char *id, char **area);
 typedef int (*tputs_func_t) (const char *str, int affcnt, int (*putc) (int));
+typedef int (*tgetnum_func_t) (char *id);
 
 tgoto_func_t tgoto;
 tgetent_func_t tgetent;
 tgetstr_func_t tgetstr;
 tputs_func_t tputs;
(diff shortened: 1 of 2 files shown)


mhoh3963 (approved):

src/broker/broker_monitor.c
```diff
@@ -1943,6 +1985,279 @@ endwin ()
...
+
+  return fgetc (stdin);
+}
+
+static void
+clear ()
+{
+  currentY = 0;
+  tputs (cl, 1, putchar);
+  fflush (stdout);
```
mhoh3963, line 2103: How about adding code "tty_Lines = tgetnum ("li");" for changing line height when user resize terminal ?

kisoo-han: Thank you for your comment!
I will accept your comment.

airnet73 (approved):

mhoh3963 (approved):
LGTM

PR #4267, opened 2023-04-06
[CBRD-24746] Query with non-existent shard_id causes coredump

airnet73:
http://jira.cubrid.org/browse/CBRD-24746

Purpose
In the Debug version, performing a query with a non-existent shard_id causes a coredump.
And, in the Release version, "Request timed out" occurs.

Implementation
N/A

Remarks
N/A


kisoo-han:
use jira# in PR title

kisoo-han (approved):

PR #4281, opened 2023-04-11
[CBRD-24756] Fix the crash while reading a log record for CDC/Flashback

joohok:
http://jira.cubrid.org/browse/CBRD-24756

Purpose

If (offset + log record size) is greater than the log page size, 
the log record to be read is located on the next page. 
To read the corresponding log record,  the pointer to read is required to be adjusted.

Implementation
1. Add macro to read appropriate log record header
   -  `LOG_READ_ADVANCE_WHEN_DOESNT_FIT`
2.  Stored log page buffer which contains the log record header at redo/undo lsa
    -  `cdc_get_overflow_recdes()` requires a log page that contains the log record header pointed to by the redo_lsa
    -   https://github.com/CUBRID/cubrid/pull/4281#discussion_r1163623777
    -   Stored existing log page only if needed (`cdc_log_read_advance_and_preserve_if_needed ()`)
3. Fix trivial bugs
   -  `redo_recdes`, `undo_recdes` can be null in `error` label (`cdc_get_recdes ()`)
   -  handle the rcvindex RVOVF_CHANGE_LINK and  RVOVF_NEWPAGE_LINK
      -  https://github.com/CUBRID/cubrid/pull/4281#discussion_r1168359155

### DIFF
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -12230,8 +12230,7 @@ cdc_get_overflow_recdes (THREAD_ENTRY * thread_p, LOG_PAGE * log_page_p, RECDES 
 
   if (((current_log_record->type == LOG_UNDO_DATA) && (rcvindex == RVOVF_PAGE_UPDATE) && is_redo)
       || ((current_log_record->type == LOG_REDO_DATA) && (rcvindex == RVOVF_PAGE_UPDATE) && !is_redo)
-      || rcvindex == RVOVF_CHANGE_LINK
-      || rcvindex == RVOVF_NEWPAGE_LINK)
+      || rcvindex == RVOVF_CHANGE_LINK || rcvindex == RVOVF_NEWPAGE_LINK)
     {
       /* start to traverse with prev_transla log record */
       LSA_COPY (&current_lsa, &current_log_record->prev_tranlsa);
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -11205,6 +11205,42 @@ cdc_get_undo_record (THREAD_ENTRY * thread_p, LOG_PAGE * log_page_p, LOG_LSA lsa
   return scan_code;
 }
 
+/*
+ * cdc_log_read_advance_and_preserve_if_needed () - Fetch the next log page and preserve the existing one,
+ *                                                  if the (lsa.offset + size) exceeds the log page size
+ *
+ * return                     : error code
+ * thread_p (in)              : thread entry
+ * size (in)                  : size to read in log page
+ * lsa (in/out)               : log address to read
+ * log_page_p (in/out)        : fetch the next log page if needed
+ * preserved  (in/out)        : preserve existing log page if needed
+ */
+
+static int
+cdc_log_read_advance_and_preserve_if_needed (THREAD_ENTRY * thread_p, size_t size, LOG_LSA * lsa, LOG_PAGE * log_page_p,
+					     LOG_PAGE * preserved)
+{
+  if (lsa->offset + size >= (int) LOGAREA_SIZE)
+    {
+      /* Before fetching the next page, current log page is required to be preserved */
+      memcpy (preserved, log_page_p, IO_MAX_PAGE_SIZE);
+
+      lsa->pageid++;
+
+      if (logpb_fetch_page (thread_p, lsa, LOG_CS_FORCE_USE, log_page_p) != NO_ERROR)
+	{
+	  logpb_fatal_error (thread_p, true, ARG_FILE_LINE,
+			     "log page fetch in cdc_log_read_advance_and_preserve_if_needed ()");
+	  return ER_FAILED;
+	}
+
+      lsa->offset = 0;
+    }
+
+  return NO_ERROR;
+}
+
 int
 cdc_get_recdes (THREAD_ENTRY * thread_p, LOG_LSA * undo_lsa, RECDES * undo_recdes, LOG_LSA * redo_lsa,
 		RECDES * redo_recdes, bool is_flashback)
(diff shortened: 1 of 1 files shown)


src/transaction/log_manager.c
```diff
@@ -11904,14 +11911,16 @@ cdc_get_recdes (THREAD_ENTRY * thread_p, LOG_LSA * undo_lsa, RECDES * undo_recde
...
+
 	    redo = (LOG_REC_REDO *) (log_page_p->area + process_lsa.offset);
 	    rcvindex = redo->data.rcvindex;
 
 	    if (rcvindex == RVOVF_NEWPAGE_INSERT || rcvindex == RVOVF_PAGE_UPDATE)
 	      {
 		/* GET OVF REDO IMAGE */
 		if ((error_code =
-		     cdc_get_overflow_recdes (thread_p, log_page_p, redo_recdes, *redo_lsa, rcvindex,
+		     cdc_get_overflow_recdes (thread_p, reserve_log_page_p, redo_recdes, *redo_lsa, rcvindex,
```
joohok: In some cases, the page_id of `redo_lsa` may differ from the page_id of the `log_page_p` due to the `LOG_READ_ADVANCE_WHEN_DOESNT_FIT (process_lsa, log_page_p)` macro. This macro can cause `log_page_p` to point to the next page. Therefore, the log page (`reserve_log_page_p`) referenced by the `redo_lsa` is temporarily stored and used from here. 

NOTE about `cdc_get_overflow_recdes()`
* It requires a `log page` that contains the log record header pointed to by the `redo_lsa`
  * In order to reduce the cost of fetching the log page buffer and just re-use the fetched one

src/transaction/log_manager.c
```diff
@@ -11360,7 +11365,7 @@ cdc_get_recdes (THREAD_ENTRY * thread_p, LOG_LSA * undo_lsa, RECDES * undo_recde
 	      {
```
wrlawodms, line 11409: `undo_length = undo->length;` doesn't seem to be needed. The same goes for all in `if (undo_lsa != NULL)`.

wrlawodms (approved):

src/transaction/log_manager.c
joohok: Thanks, I will check.

src/transaction/log_manager.c
```diff
@@ -11263,6 +11266,8 @@ cdc_get_recdes (THREAD_ENTRY * thread_p, LOG_LSA * undo_lsa, RECDES * undo_recde
 	    }
 	}
 
+      memcpy (reserve_log_page_p, log_page_p, IO_MAX_PAGE_SIZE);
```
hornetmj: This memcpy() is only needed when the log record exists in the separated log page. How about executing memcpy() only in that case?

hornetmj: The forw_lsa in the LOG_RECORD_HEADER is a good source for checking this. (I think)

joohok: OK, It seems reasonable. Whenever it tries to find `recdes` for each log record, it seems unnecessary to perform the memcpy() operation. Thank you. I will try to find a solution.

src/transaction/log_manager.c
joohok: done.

src/transaction/log_manager.c
```diff
@@ -11335,9 +11377,12 @@ cdc_get_recdes (THREAD_ENTRY * thread_p, LOG_LSA * undo_lsa, RECDES * undo_recde
 	      }
 	    else if (rcvindex == RVOVF_CHANGE_LINK || rcvindex == RVOVF_NEWPAGE_LINK)
 	      {
+		/* if the next log page is fetched into log_page_p, then use the preserved_log_page_p */
+		log_page_p = process_lsa.pageid == undo_lsa->pageid ? log_page_p : preserved_log_page_p;
+
 		/* GET OVF UNDO IMAGE */
 		if ((error_code =
-		     cdc_get_overflow_recdes (thread_p, log_page_p, undo_recdes, prev_lsa, RVOVF_PAGE_UPDATE,
+		     cdc_get_overflow_recdes (thread_p, log_page_p, undo_recdes, *undo_lsa, rcvindex,
```
joohok, line 11384: When rcvindex is RVOVF_CHANGE_LINK or RVOVF_NEWPAGE_LINK, 
the log records were traversed in reverse starting from prev_tranlsa to obtain the overflow record using cdc_get_overflow_recdes function with prev_tranlsa and RVOVF_PAGE_UPDATE as arguments.

However, it cannot be guaranteed that prev_lsa.pageid is the same as log_page_p's pageid. Even with the current and previous log pages (log_page_p, preserved_log_page_p), the log at prev_tranlsa could be stored on the previous log page.

Therefore, RVOVF_CHANGE_LINK and RVOVF_NEWPAGE_LINK will be handled as exceptions within cdc_get_overflow_recdes().

src/transaction/log_manager.c
```diff
@@ -11205,6 +11205,42 @@ cdc_get_undo_record (THREAD_ENTRY * thread_p, LOG_PAGE * log_page_p, LOG_LSA lsa
...
+ *                                                  if the (lsa.offset + size) exceeds the log page size
+ *
+ * return                     : error code
+ * thread_p (in)              : thread entry
+ * size (in)                  : size to read in log page
+ * lsa (in/out)               : log address to read
+ * log_page_p (in/out)        : fetch the next log page if needed
+ * preserved  (in/out)        : preserve existing log page if needed
+ */
+
```
joohok:
```suggestion
```

src/transaction/log_manager.c
```diff
@@ -11213,7 +11248,9 @@ cdc_get_recdes (THREAD_ENTRY * thread_p, LOG_LSA * undo_lsa, RECDES * undo_recde
   int tmpbuf_index;
 
   char *log_pgbuf[IO_MAX_PAGE_SIZE + MAX_ALIGNMENT];
-  LOG_PAGE *log_page_p = NULL;
+  char *preserved_log_pgbuf[IO_MAX_PAGE_SIZE + MAX_ALIGNMENT];
```
hornetmj, line 11251: I think you need to use LOG_PAGESIZE iso IO_MAX_PAGE_SIZE. Or any reason for this?

joohok: Done. LOG_PAGESIZE is right. 

When developing the CDC module, I used IO_MAX_PAGE_SIZE as the constant to represent the page size throughout the module. This was because the log manager module also predominantly used IO_MAX_PAGE_SIZE, so I followed its lead when developing the CDC module.

I think, changing the constant for page size to `LOG_PAGESIZE` throughout the CDC module would need to be addressed in a separate issue. For now, I think it would be better to modify the constants in the functions touched by this PR. What do you think?

hornetmj: @joohok I understand why IO_MAX_PAGE_SIZE is used. If it is, I recommend just leaving it as is or handling it all together in another issue.

joohok: OK, I will make another issue

wrlawodms (approved):

hornetmj (approved):

PR #4283, opened 2023-04-12
[CBRD-24705] When attempting a covered index scan in a connect by query, encountering 'Execute: Query execution failure' (#4237)

youngjinj:
http://jira.cubrid.org/browse/CBRD-24705

backport #4237


youngjinj:
CBRD-24705 is not included in 10.2 Patch 11. So I'll close it.

PR #4285, opened 2023-04-12
[LETS-672] Change the main connection when the previous one is disconnected.

wrlawodms:
http://jira.cubrid.org/browse/LETS-672

- In `tran_server::send_receive`, if the main connection is being disconnected, it waits until a new connection becomes the main connection and retries with it.
- When a TS receives the disconnect msg, it sets conn->m_disconnecting before removing the connection_handler from its container to get the waiters to know the main connection will be changed soon.

**Design choice**
1. Open the main connection concept to upper layers (who uses `tran_server`) and throw the main connection changing err msg to them to deal with it in their own way.
2. Hide the main connection concept to upper layers and make `send_receive` communicate with the main connection in one way or another.

I've chosen option (2) because the upper layers don't need to know the details and there are no other meaningful reactions in the case that the main connection is being changed. Welcome to any argument.


src/server/tran_server.hpp
```diff
@@ -126,6 +127,7 @@ class tran_server
 
       private:
 	std::unique_ptr<page_server_conn_t> m_conn;
+	std::atomic<bool> m_is_disconnecting;
```
cristiarg, line 130: I think it's time we introduce a state-machine mechanism here. Not with this patch, but with a refactoring. An enum with "state" values and some valid/invalid transitions between these states. Eg: UNINITIALIZED, CONNECTING, CONNECTED, DISCONNECTING, DISCONNECTED.

src/server/tran_server.cpp
```diff
@@ -362,20 +381,47 @@ tran_server::disconnect_all_page_servers ()
...
-  for (auto it = conn_vec.begin (); it != conn_vec.end (); it++)
+  auto conn_it = std::find_if (conn_vec.begin (), conn_vec.end (),
+			       [&conn] (std::unique_ptr<connection_handler> &conn_uptr)
+  {
+    return conn_uptr.get () == conn;
+  });
+
+  if (conn_it == conn_vec.end ())
+    {
+      return; // the connection is already cleared in disconnect_all_page_servers()
```
cristiarg, line 397: Not sure. Maybe this supposition is not true anymore and we should assert this situation.
I can't remember when this "the connection is already cleared in disconnect_all_page_servers()" situation occurs.

src/server/tran_server.cpp
```diff
@@ -156,14 +156,32 @@ tran_server::push_request (tran_to_page_request reqid, std::string &&payload)
...
+   *  In this case, it waits and retries with the expectation that the main connection will be changed.
+   *  - Normal disconnection: in receive_disconnect_request() and disconnect_all_page_servers()
+   *  - Abnormal disconnection: in the error handler, here, or somewhere else (TODO)
+   */
 
-  return m_page_server_conn_vec[0]->send_receive (reqid, std::move (payload_in), payload_out);
+  // TODO: exits when a thread waiting on it (transaction or whatever) stops. It should wake up periodically and check it.
+  m_main_conn_cv.wait_for (s_lock, timeout_1m, [&] ()
+  {
+    if (m_page_server_conn_vec.empty())
```
cristiarg, line 172: What if here the connection handler is already disconnecting. We would enter the `send_receive` sequence just to remain stuck.
However, I think it is better as currently written because - on the happy path - there are minimal checks/waits/locks. :)

src/server/tran_server.cpp
```diff
@@ -156,14 +156,32 @@ tran_server::push_request (tran_to_page_request reqid, std::string &&payload)
...
-    }
+  /*
+   *  send_receive () on a connection could fail when it is disconnected while waiting for its response.
+   *  In this case, it waits and retries with the expectation that the main connection will be changed.
+   *  - Normal disconnection: in receive_disconnect_request() and disconnect_all_page_servers()
+   *  - Abnormal disconnection: in the error handler, here, or somewhere else (TODO)
+   */
 
-  return m_page_server_conn_vec[0]->send_receive (reqid, std::move (payload_in), payload_out);
+  // TODO: exits when a thread waiting on it (transaction or whatever) stops. It should wake up periodically and check it.
```
cristiarg, line 169: Don't understand this `TODO`.

cristiarg (approved):

joohok (commented):
Is it not necessary to go through a process of selecting a PS as the main connection when changing it? 
For example, I think it may be necessary to check if the PS has flushed the logs up to the latest LSA (consensus LSA?) because some PSs may lag behind (due to not a member of majority), and ATS needs to show the latest data to users. 

If any PS can be selected as the main connection, it seems that a catch-up process for that PS would be necessary after changing the main connection. 

what do you think?

wrlawodms:
@joohok That is one of the TODOs. We'll calculate the cost by taking some factors into account such as priority, how far behind, latency, etc. The factors are TBD, yet. 

> If any PS can be selected as the main connection, it seems that a catch-up process for that PS would be necessary after changing the main connection.

No explicit catch-up process is required. When a TS request to a new main connection that is lagged behind, it will just be blocked until it catches up like a normal case (ATS: up to the last evicted LSA, PTS: up to the last replicated LSA)

joohok:
> No explicit catch-up process is required. When a TS request to a new main connection that is lagged behind, it will just be blocked until it catches up like a normal case (ATS: up to the last evicted LSA, PTS: up to the last replicated LSA)

OK, I understand. Thanks

joohok (approved):

src/server/tran_server.cpp
wrlawodms: One of what we discussed. When shutdown a server, any transaction or thread should have a way out of this wait().

src/server/tran_server.cpp
wrlawodms: > What if here the connection handler is already disconnecting.

In this case, the `connection_handler::send_receive()` returns immediately and it gets out because the response_broerk inside this has been already terminated, or will be terminated very soon.

src/server/tran_server.cpp
wrlawodms: When the lock is held in `disconnect_all_page_servers()`, but it hasn't cleared the vector yet, the disconnect request can arrive and this function tries to acquire the lock and, in turn, is blocked. When it gets the lock since `disconnect_all_page_servers` releases, the connection_handler has already been removed.

src/server/tran_server.hpp
wrlawodms: totally agreeee. Let's do that in the right next time we need this kind of thing, or we can create an issue in advance. :)

hornetmj (approved):

PR #4289, opened 2023-04-14
[CBRD-24726] In demodb_schema, add a synonym for the public user's tables to the dba user

youngjinj:
http://jira.cubrid.org/browse/CBRD-24726

This is for test convenience purpose.
Avoid specifying the public schema every time the dba user runs a query that uses public user tables.

### DIFF
--- a/demo/demodb_schema
+++ b/demo/demodb_schema
@@ -26,7 +26,7 @@ call change_owner('code', 'PUBLIC') on class db_root;
 CREATE CLASS nation dont_reuse_oid;
 call change_owner('nation', 'PUBLIC') on class db_root;
 
-CREATE CLASS event dont_reuse_oid;
+CREATE CLASS [event] dont_reuse_oid;
 call change_owner('event', 'PUBLIC') on class db_root;
 
 CREATE CLASS athlete dont_reuse_oid;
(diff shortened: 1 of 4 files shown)


beyondykk9 (approved):

hgryoo (approved):

shparkcubrid (approved):

hyunikn (approved):

demo/demodb_schema
```diff
@@ -48,9 +48,19 @@ CREATE CLASS history dont_reuse_oid;
...
+CREATE PRIVATE SYNONYM history FOR public.history;
+CREATE PRIVATE SYNONYM nation FOR public.nation;
+CREATE PRIVATE SYNONYM olympic FOR public.olympic;
+CREATE PRIVATE SYNONYM participant FOR public.participant;
+CREATE PRIVATE SYNONYM record FOR public.record;
+CREATE PRIVATE SYNONYM stadium FOR public.stadium;
 
 
-ALTER CLASS stadium ADD ATTRIBUTE
+ALTER CLASS public.stadium ADD ATTRIBUTE
```
ctshim, line 63: You have already created a SYNONYM called "stadium" above.
So, can't we just keep this content the same as before without editing it?

ctshim (approved):

demo/demodb_schema
youngjinj: This statement should work even if there is no synonym.
I think it is clearer to access directly by specifying "PUBLIC" than indirectly through synonym.

And when unloading in version 11.2 or higher, the alter statement explicitly specifies the schema.

PR #4313, opened 2023-04-19
[LETS-685] add dump function for RVBT_LOG_GLOBAL_UNIQUE_STATS_COMMIT log record

cristiarg:
http://jira.cubrid.org/browse/LETS-685


src/storage/btree.c
```diff
@@ -22459,6 +22459,38 @@ btree_rv_data_pack_btid_and_stats (const BTID * btid, long long nulls, long long
...
+
+  OR_GET_BIGINT (data_ptr, &stats.num_oids);
+  data_ptr += OR_BIGINT_SIZE;
+
+  OR_GET_BIGINT (data_ptr, &stats.num_keys);
+  data_ptr += OR_BIGINT_SIZE;
+
+  assert (CAST_BUFLEN (data_ptr - (const char *)data) == length);
+
+  fprintf (fp, "OID = page_id=%d|vol_id=%d|file_id=%d   NEW_STATS = %lld keys, %lld objects, %lld nulls:  %d\n",
```
hornetmj: How about replace OID = page_id=%d|vol_id=%d|file_id=%d with BTID = (%d, %d|%d) or BTREE_ID_MSG macro?

Nit:
And when I checked another debug messages, they use "nulls=%lld, oids=%lld, keys=%lld" style.

hornetmj: And the last %d is not used.

hornetmj (approved):

hornetmj:
@cristiarg Please check the code-style violation.

src/storage/btree.c
```diff
@@ -22459,6 +22459,38 @@ btree_rv_data_pack_btid_and_stats (const BTID * btid, long long nulls, long long
...
+  OR_GET_BIGINT (data_ptr, &stats.num_nulls);
+  data_ptr += OR_BIGINT_SIZE;
+
+  OR_GET_BIGINT (data_ptr, &stats.num_oids);
+  data_ptr += OR_BIGINT_SIZE;
+
+  OR_GET_BIGINT (data_ptr, &stats.num_keys);
+  data_ptr += OR_BIGINT_SIZE;
+
+  assert (CAST_BUFLEN (data_ptr - (const char *)data) == length);
```
wrlawodms: Why don't you use `btree_rv_data_unpack_btid_and_stats`?

wrlawodms: oh.. we don't have LOG_RCV in dummy functions... OK.  But isn't it better to create a LOG_RCV arbitrarily to make use of the existing function?

wrlawodms (approved):
approved with the correction from @hornetmj.

src/storage/btree.c
cristiarg: Thanks, I will do this.
When I get back to this issue. :)

src/storage/btree.c
cristiarg: Fixed

PR #4314, opened 2023-04-19
[CBRD-24756] Fix coredump due to trying to find log record using wrong rcvindex

joohok:
http://jira.cubrid.org/browse/CBRD-24756

Describe here

In `cdc_get_overflow_recdes ()`, 
when `rcvindex` is `RVOVF_CHANGE_LINK` or `RVOVF_NEWPAGE_LINK`, 
it needs to find the previous log record and start traversing in reverse order from that record.

It can obtain data using the LSA and `rcvindex` of the previous log record, however,
a problem occurred because it still considers the previous log record's rcvindex to be `RVOVF_CHANGE_LINK` or `RVOVF_NEWPAGE_LINK`. Previous log record's rcvindex is always 'RVOVF_PAGE_UPDATE'.
 
Other modifications have been made to address cases where it may reference incorrect log pages:

- When calling `cdc_get_undo_record()`, the pageid passed as an argument may be different from the pageid of `log_page_p`.
- When calculating diff between undo and redo, temporary undo data is retrieved, which can cause `log_page_p`'s pageid to change. This changed `log_page_p` can affect the operation after calculating the difference, so a temporary log page is used.

### DIFF
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -11253,7 +11253,6 @@ cdc_get_recdes (THREAD_ENTRY * thread_p, LOG_LSA * undo_lsa, RECDES * undo_recde
   LOG_PAGE *preserved_log_page_p = NULL;	// log_page for redo/undo lsa
 
   LOG_LSA process_lsa = LSA_INITIALIZER;
-  LOG_LSA current_logrec_lsa = LSA_INITIALIZER;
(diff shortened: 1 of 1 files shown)


joohok:
Passed all the CTP test cases for cdc/flashback

hornetmj (approved):

PR #4321, opened 2023-04-21
[CBRD-24533] fix the problem of increasing the number of pages in a table where data is frequently inserted and deleted

hornetmj:
http://jira.cubrid.org/browse/CBRD-24533

### Purpose
* fixed the problem of the increasing number of heap pages used when repeating the delete and insert operations
* all cases have not been taken into account, and the processing of reproduction scenarios reported in the Jira issue
* CBRD-24778 and 24779, registered as related Jira issues, are scheduled to handle other cases and refactor related code
  * proceed immediately after this PR

### Implementation
* consider registering in bestspace after vacuuming to the forward pages

### Remarks
* N/A

### DIFF
--- a/src/query/vacuum.c
+++ b/src/query/vacuum.c
@@ -2385,6 +2385,24 @@ vacuum_heap_record (THREAD_ENTRY * thread_p, VACUUM_HEAP_HELPER * helper)
 
       spage_vacuum_slot (thread_p, helper->forward_page, helper->forward_oid.slotid, true);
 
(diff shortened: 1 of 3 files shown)


joohok (approved):

hornetmj:
@joohok @wrlawodms @Rudeus This PR almost solves the issue reported in Jira. However, in the process of analyzing the issue, matters that need to be investigated and corrected were found, and this revision will be reviewed again with those related issues. Therefore, after merging, it will be reviewed again in the PRs of those issues.

PR #4322, opened 2023-04-21
[CBRD-24613] Implicit type conversion and typechecking with it

hyunikn:
http://jira.cubrid.org/browse/CBRD-24613

- Three kinds of implicit type conversion (see Coerce.java)
  - Identity : Actually no type conversion. Using value as it is.
  - Cast: corresponds to Java type cast. Used for two cases 
    - typecast null (e.g. (Timestamp) null ) to resolve ambiguous function overloads
    - typecast Object (e.g. i := (Integer) r.foo; for a result r of a dynamic SQL) to a type required in its place
  - Conversion: value change by a function call. (using one of convXXXtoYYY functions in SpLib)
- See Coerce::getCoerce for which coercion of the above three are used for combinations of a source type and a target type

- Initialization of operator overloads for implicit type conversion
  - See static initializer 78 ~ 122 of SymbolStack.java
  - collect static methods of SpLib whose names begin with 'op', and  
  - put their name, parameter types, return type, and coercion scheme annotation into a FuncOverloads class object.
  - a FuncOverload (an inner class of SymbolStack) object stores overloaded versions of a single operator

- CoercionScheme (see CoercionScheme.java)
  - Each and every operator has its own coercion scheme which indicates how to get operator's common type
  - Input to the coercion scheme is (argument types, operator name) and the output is (parameter types to pick an overload, coercions for given argument types)
  - Different categories of operators use different coercion schemes
    - CompOp (comparison operator) : =, <=>, !=, <. >. <=. >=
    - NAryCompOp(n-ary comparison operator) : between, in
    - ArithOp (arithmetic operator) : +, -, *, /, unary -
    - IntArithOp (integer arithmetic operator) : mod(%), div, ~
    - LogicalOp : and, or, xor, not
    - StringOp : ||, like
    - BitOp (bitwise operator) : <<, >>, &, ^, |
    - ObjectOp : is-null
  
- Applying implicit type conversion to operator usage 
  - See symbolStack.getOperator() invocation in visitor/TypeChecker.java
  - Get appropriate overload version of the operator and coercions, and typecheck with them

- Applying implicit type conversion in case expressions/statements
  - selector and selector values in simple case expressoins/statements
    - similar to implicit type conversion in In operator 
  - result type of case expressions
    - calculating the common type of expressions of all the branches (cases) , and coerce each expression to the common type
  
- Fix segfault of numeric type case
  - `StoredProcedure.java`: For DB_NUMERIC type, NumericValue should be created instead of DoubleValue when making a return value (makeReturnValue() in StoredProcedure). **This change mainly fixes the segfault.**
  - `CUBRIDPacker.java`: Use BigDecimal.toPlainString() instead of BigDecimal.toString() not to display exponential notations.
  - `method_struct_value.cpp`: For numeric type, unpack_string_to_memblock () is not needed. unpack_string_to_memblock() is used to allocate a new private memory block, not to release the block by the destructor. (see DB_TYPE_VARCHAR: block) However, the coerced value will be copied in numeric_coerce_string_to_num (), so it changed to use unpack_string().


src/jsp/jsp_sr.c
```diff
@@ -747,7 +747,7 @@ jsp_server_port (void)
  *   return: if jsp is disabled return -2 (JAVASP_PORT_DISABLED)
  *           else if jsp is UDS mode return -1
  *           else return a port (TCP mode)
- * 
+ *
```
hgryoo: please remove the unrelated change

src/jsp/com/cubrid/plcsql/compiler/ast/DeclFunc.java
```diff
@@ -63,4 +63,5 @@ public String kind() {
     // --------------------------------------------------
     // Private
     // --------------------------------------------------
+
```
hgryoo: redudant

src/jsp/com/cubrid/plcsql/compiler/CoercionScheme.java
```diff
@@ -0,0 +1,744 @@
...
+            assert argTypes.size() == 2;
+
+            TypeSpec commonTy =
+                    getCommonTypeInner(argTypes.get(0), argTypes.get(1), compOpCommonType);
+            if (commonTy == null) {
+                return null; // not applicable to this argument types
+            }
+
+            if (commonTy.equals(TypeSpecSimple.NULL)) {
+                commonTy = TypeSpecSimple.OBJECT;
```
hgryoo, line 61: a comment is required

src/jsp/jsp_sr.c
hyunikn: updated

src/jsp/com/cubrid/plcsql/compiler/CoercionScheme.java
hyunikn: updated

src/jsp/com/cubrid/plcsql/compiler/ast/DeclFunc.java
hyunikn: removed

hgryoo (approved):

PR #4324, opened 2023-04-25
[CBRD-24501] revised: extending dblink to DML

beyondykk9:

The following DML is not handled properly, so it needs to modify some logic.

- A test case of multiple select on insert query
insert into a_tbl1 values (select a from dblink(srv1, 'select a from t1 where a =99') t(a int), select b from t1@srv1 where a = 1, default);
- A test case of nested sub-query for update
update a_tbl1 join (select a from (select * from t1@srv1 t1 where t1.a > 0) where a > 10) t1 on a_tbl1.a = t1.a set a_tbl1.b = 'xyxyz';

### DIFF
--- a/src/parser/parser_support.c
+++ b/src/parser/parser_support.c
@@ -11504,6 +11504,50 @@ pt_convert_dblink_update_query (PARSER_CONTEXT * parser, PT_NODE * node, char *s
   return;
 }
 
+static PT_NODE *pt_convert_select (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int *continue_walk);
+
+static void
+pt_check_sub_query_spec (PARSER_CONTEXT * parser, PT_NODE * node, SERVER_NAME_LIST * snl)
+{
+  PT_NODE *list = node, *spec;
+  PT_NODE *sub_sel = NULL;
+
+  while (list)
+    {
+      if (list->info.spec.derived_table)
+	{
+	  sub_sel = list->info.spec.derived_table;
+	}
+
+      if (list->info.spec.remote_server_name)
+	{
+	  sub_sel = parser_new_node (parser, PT_SELECT);
+	  sub_sel->info.query.q.select.list = parser_new_node (parser, PT_VALUE);
+	  sub_sel->info.query.q.select.list->type_enum = PT_TYPE_STAR;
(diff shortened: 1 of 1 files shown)


src/parser/parser_support.c
```diff
@@ -11594,7 +11617,7 @@ pt_convert_dblink_dml_query (PARSER_CONTEXT * parser, PT_NODE * node, char *sql_
 
   if (snl->local_cnt > 0 && remote_upd > 0)
     {
-      PT_ERROR (parser, upd_spec, "dblink: local mixed remote DML is not allowed");
+      PT_ERROR (parser, upd_spec ? upd_spec : into_spec, "dblink: local mixed remote DML is not allowed");
```
ctshim, line 11627: In the case of PT_MERGE, you can have both upd_spec and into_spec.

src/parser/parser_support.c
```diff
@@ -11513,16 +11557,27 @@ pt_convert_dblink_dml_query (PARSER_CONTEXT * parser, PT_NODE * node, char *sql_
...
   switch (node->node_type)
     {
     case PT_INSERT:
       into_spec = node->info.insert.spec;
       sub_sel = node->info.insert.value_clauses;
-      if (sub_sel->info.node_list.list && sub_sel->info.node_list.list->node_type == PT_SELECT)
+      for (list = sub_sel->info.node_list.list; list; list = list->next)
 	{
-	  upd_spec = sub_sel->info.node_list.list->info.query.q.select.from;
+	  if (list->node_type == PT_SELECT)
```
ctshim: insert into x values (5 + (select count(*) from dual@srv1) );

If you perform the query as above, it seems that PT_SELECT alone will not solve it.

ctshim (approved):

youngjinj:
Why is `\n` needed in pt_print_dblink_table_dml function?
```
static PARSER_VARCHAR *
pt_print_dblink_table_dml (PARSER_CONTEXT * parser, PT_NODE * p)
{
    ...
    /* For Query-cache:
     * Separate comments have been added 
     * for cases where there is no change in the query but information on the server has changed. */
    q = pt_append_nulstring (parser, q, "\n /* DBLINK(");
...
```

PR #4325, opened 2023-04-25
[CBRD-24781] Added csql option for catalog rebuild

youngjinj:
http://jira.cubrid.org/browse/CBRD-24781

Add an option to change the catalog only when the csql utility is executed in standalone mode and the `command` or `input-file` option is used.


mhoh3963 (approved):
LGTM

PR #4331, opened 2023-04-27
[CBRD-24521] Make PL/CSQL registerable with CREATE PROCEDURE/FUNCTION

hgryoo:
http://jira.cubrid.org/browse/CBRD-24521

**Implementation**
- Revise the syntax error text for CREATE PROCEDURE/FUNCTION in message files
- Add a new syntax for PL/CSQL (in csql_grammar.y)
- Store the Java signature of translated PL/CSQL program in the CompileInfo and Return it to CAS. The Java signature is written in db_stored_procedure's target column (catalog)
- Add a new routine to register PL/CSQL program in jsp_create_stored_procedure(). The new routine calls plcqsl_transfer_file (), which communicates with the Java SP server.
- [update lexer and parser rules to identify PL/CSQL texts](https://github.com/CUBRID/cubrid/pull/4331/commits/460115e7edc40b00d555bb1b826c88aa79fe51ba)

**TODO in another PRs**
- Syntax error is not properly handled for the new routine
- Checking connection with Java SP server is not considered

**Remarks**

For test)
---
1. normal case

```
csql -u public demodb
```

csql>
```
;server-output on

create or replace procedure demo_hello() as 
begin
    put_line('Hello CUBRID PL/CSQL!');
end;

call demo_hello ();
```
Result
```
Hello CUBRID PL/CSQL!
Execute OK. (0.003685 sec) Committed. (0.000115 sec)
```
---
2. Semantic error
```
csql -u public demodb
```

csql>
```
create or replace procedure poo(i int) as
    i int;
    s string;
begin
    // cpp comment
    -- sql comment
    /* c comment
       **
     */
    null;
end;
```
Result
```
In the command from line 11,
ERROR: Stored procedure compile error: I has already been declared in the same scope at 1.
```


hgryoo:
@hyunikn Please help me to extract the string between AS ... END to get rid of the restriction of having to wrap the SP body as a quoted string.

src/parser/csql_lexer.l
```diff
@@ -132,7 +140,12 @@ int is_dblink_query_string = 0;
 [aA][rR][iI][aA]							{ begin_token(yytext);   
                                                                                 csql_yylval.cptr = pt_makename(yytext);
                                                                                 return ARIA; }
-[aA][sS]								{ begin_token(yytext);   return AS; }
+[aA][sS]								{ begin_token(yytext);
+                                                                                if (expecting_pl_lang_spec) {
+                                                                                        expecting_pl_lang_spec = 0;
+                                                                                        BEGIN(PL_LANG_SPEC);
+                                                                                }
+                                                                                return AS; }
```
hgryoo, line 148: `AS` can be used when specifying column alias.
```
SELECT a AS 'alias' FROM tbl;
```
Is it ok to move to PL_LANG_SPEC mode without any consideration?

hyunikn: It happens when expecting_pl_lang_spec flag is on. 
The flag is turned on by the parser when CREATE (OR REPALCE) PROCEDURE/FUNCTION tokens are found in this order.

hyunikn: Anyway, I realized that the flag is not turned off for a invalid statement having a syntax error of no IS or AS after the CREATE (OR REPALCE) PROCEDURE/FUNCTION token stream. I will fix it.

hyunikn:
> ERROR: Stored procedure compile error: I has already been declared in the same scope at 1.
. How about dropping 'Stored procedure compile error: ' in the error message?
. We need to discuss about how to show line and column number of PL/CSQL compile errors precisely

hyunikn (approved):

PR #4332, opened 2023-04-28
[CBRD-24782] extracting Static SQL texts without SqlParser.g4

hyunikn:
http://jira.cubrid.org/browse/CBRD-24782

Implementation

- Identify beginnings of Static SQLs by keywords WITH, SELECT, INSERT, UPDATE, DELETE, REPLACE, MERGE, and TRUNCATE, and change the lexer's mode to STATIC_SQL.
- In STATIC_SQL lexer mode, only semicolons, left&right parentheses, quoted strings, white spaces are recognized, and all the other character chunks are recognized as, namely, NON_STR.
- STATIC_SQL lexer mode is ended when a semicolon is encountered, or for the SELECT statement in a for loop, when a non-matching right parenthesis is encountered.
- SqlParser.g4 and no more used lexer tokens in PcsLexer.g4 were removed.
- '(', ')', and ';' in parser rules are replaced with their lexer token name LPAREN, RPAREN, and SEMICOLON, respectively due to the ambiguity introduced by new tokens SS_LPAREN, SS_RPAREN, and SS_SEMICOLON.
- parser rules data_manipulation_language_statements and s_select_statement are replaced by a new one named static_sql.


hyunikn (commented):
OK.

src/jsp/antlr4/PcsLexer.g4
```diff
@@ -26,256 +26,120 @@ lexer grammar PcsLexer;
...
+private int staticSqlParenMatch = -1;
+}
+
+// keywords that starts Static SQL
+WITH:                         W I T H           { staticSqlParenMatch++; mode(STATIC_SQL); };
+SELECT:                       S E L E C T       { staticSqlParenMatch++; mode(STATIC_SQL); };
+INSERT:                       I N S E R T       { staticSqlParenMatch++; mode(STATIC_SQL); };
+UPDATE:                       U P D A T E       { staticSqlParenMatch++; mode(STATIC_SQL); };
+DELETE:                       D E L E T E       { staticSqlParenMatch++; mode(STATIC_SQL); };
+REPLACE:                      R E P L A C E     { staticSqlParenMatch++; mode(STATIC_SQL); };
```
hgryoo, line 39: Does this PR consider the REPLACE built-in function without assigning expression (:=)? (that we discussed)
```
REPLACE('12345abcdeabcde','abcde','ABCDE');
```

We do not support CUBRID built-in function calls currently, IMO, we don't have to consider the problem in the issue.

hgryoo (approved):

src/jsp/antlr4/PcsLexer.g4
hyunikn: Yes. It seems OK to handle it together with built-in functions support.
I believe it can be handled by immediately returning to the default lexer mode at a left parenthesis to the right of a REPLACE token.

PR #4334, opened 2023-04-28
[CBRD-24620] Replace db name copy method not to occur memory leak

eido5:
http://jira.cubrid.org/browse/CBRD-24620

**Description**

The user typing that the below command to connect other users is not intuitive, while the CSQL session is connected;

CALL login ('user_name', 'password') ON CLASS db_user;
So, we should add the session command user to use more intuitive while the CSQL session is connected.

;connect user_name db_name@host_name
This can helps user to connect more intuitive while the CSQL session is connected.

**Remarks**
* This PR is to avoid possible memory leak of #4134 pointed out by valgrind.

### DIFF
--- a/src/executables/csql.c
+++ b/src/executables/csql.c
@@ -127,6 +127,7 @@ int (*csql_text_console_to_utf8) (const char *, const int, char **, int *) = NUL
 
 int csql_Row_count;
 int csql_Num_failures;
+char csql_Db_name[512];
 
 /* command editor lines */
 int csql_Line_lwm = -1;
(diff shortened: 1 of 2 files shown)


src/executables/csql.c
```diff
@@ -3384,6 +3384,7 @@ csql_connect (char *argument, CSQL_ARGUMENT * csql_arg)
   char *p = NULL;
   const char *err_msg;
   CSQL_ARGUMENT csql_new_arg;
+  char csql_db_name[DB_NAME_LEN];
```
kisoo-han: Since the use of this variable's address is outside the scope of this function, it is better to make it as a global variable.

kisoo-han (approved):

src/executables/csql.c
eido5: Right. If not, It may occur memory problem.
After I change it as global variable and then Can I merge this pr?

kisoo-han: Yes, I already approved this PR.

eido5: Ok! Thank you!

PR #4342, opened 2023-05-04
[CBRD-24501] rev: (bug-fix) extending dblink to DML

beyondykk9:

Process dml statements on remote DB tables in local DB triggers
```
A trigger is created in the "before INSERT ON" statement, but an error occurs when the trigger occurs.
A trigger cannot be created in the "before UPDATE ON" statement.
A trigger cannot be created in the "before DELETE ON" statement.

```
Error when searching for one constant value or function in dblink select statement
```
select '1' from db_class@srv1 limit 1;
select 1 from db_class@srv1 limit 1;
```

The following quries work well successfully.
```
select sysdate from db_class@srv1 limit 1;
select sysdate from db_class@srv1 limit 1;
```


src/parser/csql_grammar.y
```diff
@@ -5505,6 +5508,11 @@ class_name_with_server_name
                 PT_NAME_INFO_CLEAR_FLAG($1, PT_NAME_INFO_USER_SPECIFIED);
                 SET_CONTAINER_2 (ctn, $1, $3);
                 $$ = ctn;
+		if (is_in_create_trigger)
+		  {
+		    PT_ERROR(this_parser, $1, "check syntax: Not supported @!!!");
```
ctshim: Shouldn't the error message be modified to be more appropriate?

ctshim (approved):

src/parser/csql_grammar.y
beyondykk9: Fixed.

src/parser/parser_support.c
```diff
@@ -10968,39 +10968,198 @@ pt_get_server_name_list (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int
...
+	    }
+	  ps++;
+	}
+    }
+
+  return ps;
+}
+
+static char *
+find_circle_at_char (bool ansi_quotes, bool no_escape, char *ps)
```
youngjinj, line 11063: I need some test cases to understand what this code is for.

Is it for exception handling when the '@' character is included in the identifier?
If so, I think it is efficient to receive `PT_NODE *` as an argument rather than `char *sql_user_text`.

src/parser/parser_support.c
```diff
@@ -10968,39 +10968,198 @@ pt_get_server_name_list (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int
...
+}
+
 static PARSER_VARCHAR *
 pt_make_remote_query (PARSER_CONTEXT * parser, char *sql_user_text, SERVER_NAME_LIST * snl)
 {
   PARSER_VARCHAR *pvc = NULL;
   char *ps, *pt, *t;
+  int adjust_pos = 1;
+  bool ansi_quotes = prm_get_bool_value (PRM_ID_ANSI_QUOTES);
+  bool no_escape = prm_get_bool_value (PRM_ID_NO_BACKSLASH_ESCAPES);
```
youngjinj, line 11129: I think it is more efficient to copy `PT_NODE *` here, process the necessary parts, and then use the `parser_print_tree` function to make `char *sql_user_text`.

youngjinj:
It is not reproducible, so you can ignore it.

---

An error occurs when ansi_quotes is no.
```
$ vi $CUBRID/conf/cubrid.conf

ansi_quotes=no
```

```
INSERT INTO ddl_0001@srv1 VALUES( ddl_0001_01.next_value, 'insert'||'1', col3, col4 );
/*
1 row affected.
*/
```

It succeeds when ansi_quotes is yes, but fails when no.
```
INSERT INTO ddl_0001@srv1 VALUES( ddl_0001_01.next_value, 'insert', col3, col4 );
/*
ERROR: dblink - Semantic: before ' , col3, col4 );'
Attribute "" was not found. insert into [dba.ddl_0001] [dba.ddl_0001] ([dba.ddl_0001].co...
*/
```

beyondykk9:
> INSERT INTO ddl_0001@srv1 VALUES( ddl_0001_01.next_value, 'insert'||'1', col3, col4 );

INSERT INTO ddl_0001@"srv1" VALUES( ddl_0001_01.next_value, 'insert'||'1', col3, col4 );

This is normal behavior.
This is because double and single quotes are recognized as string handling symbols when ansi_quotes is no.

src/parser/parser_support.c
beyondykk9: It should come out the same as the user's original query string, and it should include all hint strings, conditions, etc.
However, the node query may have been partially rewritten, and the hint string may not be output as it is.

src/parser/parser_support.c
ctshim: The reason for using sql_user_text is as follows.
1) Here we need the user's original query that has not been rewritten in the middle.
2) It is not easy to get the top root node of the tree.

src/parser/parser_support.c
ctshim: This function removes the @server information from the query entered by the user and creates a query string to be transmitted to the remote server.

The reason for using sql_user_text is as follows.
1) Here we need the user's original query that has not been rewritten in the middle.
2) It is not easy to get the top root node of the tree.

youngjinj:
> > INSERT INTO ddl_0001@srv1 VALUES( ddl_0001_01.next_value, 'insert'||'1', col3, col4 );
> 
> INSERT INTO ddl_0001@"srv1" VALUES( ddl_0001_01.next_value, 'insert'||'1', col3, col4 );
> 
> This is normal behavior. This is because double and single quotes are recognized as string handling symbols when ansi_quotes is no.

It happened yesterday, but I can not reproduce it again.
It is not reproducible, so you can ignore it.

src/parser/parser_support.c
beyondykk9: This function will not be used, because I change the routine to use parser_print_tree or pt_print_bytes to add the custom print flag "PT_SUPPRESS_SERVER_NAME".

src/parser/parser_support.c
beyondykk9: This function will not be used, because I change the routine to use parser_print_tree or pt_print_bytes to add the custom print flag "PT_SUPPRESS_SERVER_NAME".

ctshim (approved):

src/compat/db_vdb.c
```diff
@@ -677,7 +677,12 @@ db_compile_statement_local (DB_SESSION * session)
    */
   if (PT_IS_DBLINK_DML_QUERY (statement))
     {
-      ;
+      /*
+         The remote DML query can not use the cached plan,
+         because the atuo-variable holder can not be composed during compiling.
+         Instead the remote query is generated by parser_print_tree.
+       */
+      statement->flag.recompile = 1;
```
youngjinj, line 685: I think there should be an example in the comments of the code or in the comments of issue CBRD-24501 for understanding.

src/parser/parser_support.c
```diff
@@ -11669,12 +11830,10 @@ pt_convert_dblink_dml_query (PARSER_CONTEXT * parser, PT_NODE * node, char *sql_
   val->type_enum = PT_TYPE_CHAR;
   val->info.value.string_type = ' ';
 
-  assert (sql_user_text && sql_user_text[0]);
-  if (sql_user_text)
-    {
-      val->info.value.data_value.str = pt_make_remote_query (parser, sql_user_text, snl);
-      PT_NODE_PRINT_VALUE_TO_TEXT (parser, val);
-    }
+  parser->custom_print |= PT_PRINT_SUPPRESS_SERVER_NAME;
```
youngjinj: In the `pt_convert_dblink_dml_query` function, `sql_user_text` seems to be no longer needed.
And where the `pt_convert_dblink_merge_query` function is called, it seems that the code below is no longer needed either.

src/parser/parser_support.c
```diff
@@ -10968,39 +10968,199 @@ pt_get_server_name_list (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int
   return node;
 }
 
+#if defined (ENABLE_UNUSED_FUNCTION)
```
youngjinj, line 10971: For readability, how about setting `ENABLE_UNUSED_FUNCTION` separately for each function?

youngjinj:
Why is `\n` needed in pt_print_dblink_table_dml function?
```
static PARSER_VARCHAR *
pt_print_dblink_table_dml (PARSER_CONTEXT * parser, PT_NODE * p)
{
    ...
    /* For Query-cache:
     * Separate comments have been added 
     * for cases where there is no change in the query but information on the server has changed. */
    q = pt_append_nulstring (parser, q, "\n /* DBLINK(");
...
```

src/parser/parser_support.c
youngjinj: If it is an oracle dblink query, reserved words must be enclosed in double quotes, not square brackets.
- AS-IS
   ```
   insert into desc@srv1 values ('invalid table name');
     -> 'insert into [desc] values (''invalid table name'')'
   ```

youngjinj: Is it okay if the LIMIT statement is written together?
Oracle does not support LIMIT statements.
- AS-IS
   ```
   insert into [desc]@srv1 select * from [desc]@srv1 limit 1;
     -> 'insert into [desc] select * from [desc] limit 1'
   ```

youngjinj:
It's trivial.
I found that I had to put the server name in db_root or dual as well.
```
merge into t1@srv1 as a using db_root b on (a.c1 = 2)
when matched then update set a.c2 = 2
when not matched then insert (a.c1, a.c2) values (2, 2);

/*
dblink: multi-remote DML is not allowed
*/
```

```
merge into t1@srv1 as a using db_root@srv1 b on (a.c1 = 2)
when matched then update set a.c2 = 2
when not matched then insert (a.c1, a.c2) values (2, 2);

merge into t1@srv1 as a using dual@srv1 b on (a.c1 = 2)
when matched then update set a.c2 = 2
when not matched then insert (a.c1, a.c2) values (2, 2);

/*
1 row affected.
*/
```

src/parser/parser_support.c
youngjinj: This can be a problem when using sequences in oracle dblonk queries.
The `serial_next_value` method supports only CUBRID.
- AS-IS
   ```
   insert into t1@srv1 values (sequence1.nextval);
     -> 'insert into t1 values (serial_next_value(sequence1, 1))'
   ```

beyondykk9: Hmm, it needs to suppress rewriting serial.nextval?

beyondykk9:
> It's trivial. I found that I had to put the server name in db_root or dual as well.
> 
> ```
> merge into t1@srv1 as a using db_root b on (a.c1 = 2)
> when matched then update set a.c2 = 2
> when not matched then insert (a.c1, a.c2) values (2, 2);
> 
> /*
> dblink: multi-remote DML is not allowed
> */
> ```
> 
> ```
> merge into t1@srv1 as a using db_root@srv1 b on (a.c1 = 2)
> when matched then update set a.c2 = 2
> when not matched then insert (a.c1, a.c2) values (2, 2);
> 
> merge into t1@srv1 as a using dual@srv1 b on (a.c1 = 2)
> when matched then update set a.c2 = 2
> when not matched then insert (a.c1, a.c2) values (2, 2);
> 
> /*
> 1 row affected.
> */
> ```

db_root and dual is also table name and it has to specifiy the server name with '@' to identify remote side.

src/parser/parser_support.c
beyondykk9: 'limit' syntax will arise error in ORACLE. It's responsiblity of user for the wrong syntax.

beyondykk9: It's responsiblity of user for the wrong specifying a name.

src/parser/parser_support.c
beyondykk9: OK, fixed.

src/parser/parser_support.c
beyondykk9: Ok, fixed.

beyondykk9:
> Why is `\n` needed in pt_print_dblink_table_dml function?
> 
> ```
> static PARSER_VARCHAR *
> pt_print_dblink_table_dml (PARSER_CONTEXT * parser, PT_NODE * p)
> {
>     ...
>     /* For Query-cache:
>      * Separate comments have been added 
>      * for cases where there is no change in the query but information on the server has changed. */
>     q = pt_append_nulstring (parser, q, "\n /* DBLINK(");
> ...
> ```

It's a trivial for readability of rewriting dblink query.

src/compat/db_vdb.c
beyondykk9: Ok, I commented on the JIRA issue CBRD-24501.

src/parser/parser_support.c
youngjinj: It's not that I used the wrong name.

Both Oracle and CUBRID can use reserved words as identifiers. Oracle uses reserved words by enclosing them in double quotations(" "), and CUBRID uses them by enclosing them in square brackets([ ]), double quotations(" "), and backquotes(\` \`).

The problem is that the parser_print_tree function only uses square brackets([ ]) when printing reserved words. An error will occur if the result output by the parser_print_tree function is used as an oracle dblink query.

youngjinj: When the @ interface is used, the oracle dblink query is made in CUBRID. So, the responsibility for making a mistake should be seen as CUBRID.

Limit should be removed when writing dblink query to oracle.
If this is difficult, an error should occur when LIMIT is used in a query that only uses dblink.

For now, users can bypass with rownum.
This is something that needs to be improved in the future.

youngjinj: sequence1.nextval is changed to the serial_next_value method in the parser.
We need to make sure the parser doesn't change it.

src/compat/db_vdb.c
youngjinj: I saw it. Thank you.

youngjinj:
> > It's trivial. I found that I had to put the server name in db_root or dual as well.
> > ```
> > merge into t1@srv1 as a using db_root b on (a.c1 = 2)
> > when matched then update set a.c2 = 2
> > when not matched then insert (a.c1, a.c2) values (2, 2);
> > 
> > /*
> > dblink: multi-remote DML is not allowed
> > */
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> > merge into t1@srv1 as a using db_root@srv1 b on (a.c1 = 2)
> > when matched then update set a.c2 = 2
> > when not matched then insert (a.c1, a.c2) values (2, 2);
> > 
> > merge into t1@srv1 as a using dual@srv1 b on (a.c1 = 2)
> > when matched then update set a.c2 = 2
> > when not matched then insert (a.c1, a.c2) values (2, 2);
> > 
> > /*
> > 1 row affected.
> > */
> > ```
> 
> db_root and dual is also table name and it has to specifiy the server name with '@' to identify remote side.

Is there any way to exclude db_root and dual from being checked as they are remote tables? It seems strange that the @ interface is attached to db_root and dual.

youngjinj:
> \n

I feel that '\n' is rather less readable. How about removing it?

youngjinj:
> > Why is `\n` needed in pt_print_dblink_table_dml function?
> > ```
> > static PARSER_VARCHAR *
> > pt_print_dblink_table_dml (PARSER_CONTEXT * parser, PT_NODE * p)
> > {
> >     ...
> >     /* For Query-cache:
> >      * Separate comments have been added 
> >      * for cases where there is no change in the query but information on the server has changed. */
> >     q = pt_append_nulstring (parser, q, "\n /* DBLINK(");
> > ...
> > ```
> 
> It's a trivial for readability of rewriting dblink query.

I feel that `\n` is rather less readable. How about removing it?

src/parser/parser_support.c
beyondykk9: I'm going to handle this issue as a separate issue.

beyondykk9: I'm going to add PT_SUPPRESS_SERIAL_CONV custom print flag.

youngjinj: It seems we will use the `PT_PRINT_SUPPRESS_SERVER_NAME`, `PT_SUPPRESS_SERIAL_CONV` flags only for outputting DBLink queries.
How about doing both using just one flag?

src/parser/parser_support.c
```diff
@@ -11262,7 +11431,10 @@ pt_convert_dblink_synonym (PARSER_CONTEXT * parser, PT_NODE * spec, void *has_sy
 	      spec->info.spec.range_var = parser_new_node (parser, PT_NAME);
 	      spec->info.spec.range_var->info.name.original = pt_append_string (parser, NULL, synonym_name);
 	    }
-	  *(bool *) has_synonym = true;
+	  if (has_synonym)
```
youngjinj: Why do we still need to use has_synonym?
It is only the pt_convert_dblink_insert_query function that passes has_synonym when calling pt_convert_dblink_synonym.

The reason is to call parser_free_node function for 'node->info.insert.spec->info.spec.range_var' added in pt_convert_dblink_synonym.
Is it correct to call the parser_free_node function when has_synonym is true?

src/parser/parser_support.c
```diff
@@ -11792,19 +11947,19 @@ pt_convert_dml (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int *continu
   switch (node->node_type)
     {
     case PT_INSERT:
-      pt_convert_dblink_insert_query (parser, node, node->sql_user_text, snl);
+      pt_convert_dblink_insert_query (parser, node, snl);
```
youngjinj, line 11938: `snl` does not require a return value from the pt_convert_dml function.
pt_convert_dblink_insert_query, pt_convert_dblink_delete_query, and pt_convert_dblink_update_query also do not use `snl`.
It seems only needed inside the pt_convert_dblink_dml_query function.

The pt_convert_dblink_merge_query function uses `snl`, but I don't know if I should use it only here.

Can we change it to use `snl` only in the pt_convert_dblink_dml_query function?

src/parser/parser_support.c
beyondykk9: Ok, I aggree with you. I fixed it.

src/parser/parser_support.c
beyondykk9: snl is tranferred to pt_convert_dblink_dml_query from pt_convert_dblink_insert_query, pt_convert_dblink_delete_query, and pt_convert_dblink_update_query, and pt_convert_dblink_merge_query.
The snl is used to check if the query is local or remote.
The snl come from pt_convert_dml by pt_rewrite_for_dblink which calls parser_walk_tree (parser, stmt, NULL, NULL, pt_convert_dml, &snl);

youngjinj: I was mistaken. There seems to be no problem with `snl`.

youngjinj: Get `snl` from pt_rewrite_for_dblink function.
But `snl` is empty until we pass it to the pt_convert_dblink_dml_query function.
```
pt_rewrite_for_dblink (parser, statement);
  -> SERVER_NAME_LIST snl;
  -> memset (&snl, 0x00, sizeof (SERVER_NAME_LIST)); /********** `snl` is empty. /**********/
  -> parser_walk_tree (parser, stmt, NULL, NULL, pt_convert_dml, &snl);
       -> pt_convert_dml
            -> pt_convert_dblink_insert_query (parser, node, snl); /********** Until here, `snl` is not used. /**********/
                 -> pt_convert_dblink_dml_query (parser, node, (remote_ins == 0), remote_ins, snl);
            -> pt_convert_dblink_delete_query (parser, node, snl); /********** Until here, `snl` is not used. /**********/
                 -> pt_convert_dblink_dml_query (parser, node, local_del, remote_del, snl);
            -> pt_convert_dblink_update_query (parser, node, snl); /********** Until here, `snl` is not used. /**********/
                 -> pt_convert_dblink_dml_query (parser, node, local_upd, remote_upd, snl);
            -> pt_convert_dblink_merge_query (parser, node, snl);
                 -> if (remote_target) 
                      {
                        parser_walk_tree (parser, source, pt_get_server_name_list, snl, NULL, NULL);
                        if (snl->local_cnt > 0)
                          {
                            /* not allowed to merge multi-server */
                            PT_ERROR (parser, source, "dblink: multi-remote DML is not allowed");
                          }
                        else
                          {
                            /* only remote merge */
                            assert (snl->server_cnt > 0);
                          }
                      }
                 -> pt_convert_dblink_dml_query (parser, node, (int) (remote_target == false), (int) (remote_target == true), snl);
```

Only the pt_convert_dblink_merge_query function is using `snl`.

`snl` does not require a return value from the pt_rewrite_for_dblink function.

No matter how much I think about it, I think `snl` continues to be passed unnecessarily.

youngjinj (approved):

PR #4343, opened 2023-05-09
[CBRD-24204] Change print function to keep printing order

eido5:
http://jira.cubrid.org/browse/CBRD-24204

**Description**
 I changed print function to keep printing order.

### DIFF
--- a/src/executables/util_cs.c
+++ b/src/executables/util_cs.c
@@ -3286,7 +3286,8 @@ applyinfo (UTIL_FUNCTION_ARG * arg)
 	{
 	  db_clear_host_connected ();
 
-	  printf ("\n *** Applied Info. *** \n");
+	  fprintf (stdout, "\n *** Applied Info. *** \n");
+	  fflush (stdout);
 	  if (check_database_name (local_database_name))
 	    {
 	      goto check_applied_info_end;
@@ -3347,7 +3348,8 @@ applyinfo (UTIL_FUNCTION_ARG * arg)
 
       if (check_master_info_temp)
 	{
-	  printf ("\n ***  Active Info. *** \n");
+	  fprintf (stdout, "\n ***  Active Info. *** \n");
+	  fflush (stdout);
 	  memset (master_database_name, 0x00, CUB_MAXHOSTNAMELEN);
 	  strcpy (master_database_name, database_name);
 	  strcat (master_database_name, "@");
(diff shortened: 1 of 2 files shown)


hornetmj:
@eido5 I don't quite understand how the explanation in the description was solved with fflush. I would like to ask for a little more detailed explanation.

src/executables/util_cs.c
```diff
@@ -3286,7 +3286,8 @@ applyinfo (UTIL_FUNCTION_ARG * arg)
 	{
 	  db_clear_host_connected ();
 
-	  printf ("\n *** Applied Info. *** \n");
+	  fprintf (stdout, "\n *** Applied Info. *** \n");
```
kisoo-han, line 3289: Because, printf (...) is same with fprintf (stdout, ...), you can keep original _printf_ statements.
For example,

```
printf ("\n *** Applied Info. *** \n");
fflush (stdout);
```

kisoo-han (approved):

eido5:
> @eido5 I don't quite understand how the explanation in the description was solved with fflush. I would like to ask for a little more detailed explanation.

I think I should explain overall explanation.

0. Before applying fflush(), *** Active Info. *** is printed not correct order. I would guess that is derived from repeated print and buffer is not flushed.
1. This commit does not occur any problems itself.
2. After commit 6091c7c is applied (Jooho's commit), it occurred incorrect order printing.
3. I heard that Jooho told me 6091c7c does not seem to be reason.
4. So, I can guess there is the something wrong with print buffer  (it looks transparent.)

Would this explanation help you?

src/executables/util_cs.c
eido5: Thank you. I will test and apply it.

joohok:
Hi, 
I read the description and comments because I was mentioned above.. 

At first, the commit you mentioned is related to appending log records only when necessary, so that issue has no impact on the print function. 

> When applyinfo executes on the slave node, slave node stops and then -a option related prints are also stopped.
When we restart ha slave node, It does not work still.
So, we should make applyinfo utility works normal again when ha slave node works again 

and I don't understand this part, because fflush() only flushes the buffers, and it does not have any impact on the applyinfo utility.

hornetmj:
@eido5 As I know it works as single-thread and every printf function uses same buffer. So, I'm not convinced that calling fflush() solved the problem. Is there any reference that fflush can solve this problem?

eido5:
@joohok 
Above mention is aim to be helpful CBRD-24204. But It makes confusing. I will remove that.

hornetmj (approved):
@eido5 @kisoo-han It's not an issue I'm managing, so I approve it first. However, it seems that sufficient consideration is needed. I don't understand why this problem is solved by calling fflush function. Therefore, it is not possible to determine if this correction is an accurate or not. Any references or explanations on this would be appreciated.

kisoo-han:
My conclusion is that this PR is not necessary. I think the TestCase needs to be modified.
/home/kshan/cubrid-testcases-private/HA/shell/_12_bts_issue/cbrd_24204/cases
Refer comment on CBRD-24204.

eido5:
The reason of this issue is not code fix. So, this PR is closed.

PR #4344, opened 2023-05-10
[CBRD-24542] Prevent schema loading with loaddb in HA Mode.

airnet73:
http://jira.cubrid.org/browse/CBRD-24542

Purpose
* When running online mode loaddb (cubrid loaddb -C) in HA environment DDL replication does not occur to slave nodes.
* For this reason, the online mode loaddb is modified so that only object loading is possible.
* If schema loading is attempted, an error is output.
* In case of using loaddb -C option in HA Mode, the following options cannot be used.
--load-only
-i, --index-file
-s, --schema-file
--schema-file-list
--trigger-file=FILE

Implementation
N/A

Remarks
N/A

### DIFF
--- a/src/loaddb/load_db.c
+++ b/src/loaddb/load_db.c
@@ -198,7 +198,7 @@ ldr_validate_object_file (const char *argv0, load_args * args)
 		   || args->schema_file_list.empty () == false || args->trigger_file.empty () == false))
 		{
 		  PRINT_AND_LOG_ERR_MSG
-		    ("loaddb: In case of using loaddb CS mode (-C, --CS-mode) in HA Mode, only OBJECT can be loaded.\n");
+		    ("In loaddb CS (-C, --CS-mode) in HA mode, only OBJECT loading is possible.\n");
 		  return ER_FAILED;
 		}
 	    }
--- a/src/loaddb/load_db.c
+++ b/src/loaddb/load_db.c
@@ -186,13 +186,28 @@ ldr_validate_object_file (const char *argv0, load_args * args)
       return ER_FAILED;
     }
 
-  if (args->cs_mode && (args->load_only == true
-			|| args->index_file.empty () == false
-			|| args->schema_file.empty () == false
-			|| args->schema_file_list.empty () == false || args->trigger_file.empty () == false))
+  if (args->cs_mode)
     {
-      fprintf (stderr, "In loaddb CS mode (-C, --CS-mode), only object loading is possible.\n");
-      return ER_FAILED;
+      if (sysprm_load_and_init (args->volume.c_str (), NULL, SYSPRM_IGNORE_INTL_PARAMS) == NO_ERROR)
+	{
+	  if (prm_get_integer_value (PRM_ID_HA_MODE))
+	    {
+	      if ((args->load_only == true
+		   || args->index_file.empty () == false
+		   || args->schema_file.empty () == false
+		   || args->schema_file_list.empty () == false || args->trigger_file.empty () == false))
+		{
+		  PRINT_AND_LOG_ERR_MSG
+		    ("loaddb: In case of using loaddb CS mode (-C, --CS-mode) in HA Mode, only OBJECT can be loaded.\n");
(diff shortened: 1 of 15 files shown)


src/loaddb/load_db.c
```diff
@@ -186,13 +186,28 @@ ldr_validate_object_file (const char *argv0, load_args * args)
...
+	{
+	  if (prm_get_integer_value (PRM_ID_HA_MODE))
+	    {
+	      if ((args->load_only == true
+		   || args->index_file.empty () == false
+		   || args->schema_file.empty () == false
+		   || args->schema_file_list.empty () == false || args->trigger_file.empty () == false))
+		{
+		  PRINT_AND_LOG_ERR_MSG
+		    ("loaddb: In case of using loaddb CS mode (-C, --CS-mode) in HA Mode, only OBJECT can be loaded.\n");
```
kisoo-han:
```suggestion
		    ("In loaddb CS (-C, --CS-mode) in HA mode, only OBJECT loading is possible.\n");
```

kisoo-han (approved):

src/loaddb/load_db.c
```diff
@@ -186,13 +186,27 @@ ldr_validate_object_file (const char *argv0, load_args * args)
...
-			|| args->index_file.empty () == false
-			|| args->schema_file.empty () == false
-			|| args->schema_file_list.empty () == false || args->trigger_file.empty () == false))
+  if (args->cs_mode)
     {
-      fprintf (stderr, "In loaddb CS mode (-C, --CS-mode), only object loading is possible.\n");
-      return ER_FAILED;
+      if (sysprm_load_and_init (args->volume.c_str (), NULL, SYSPRM_IGNORE_INTL_PARAMS) == NO_ERROR)
+	{
+	  if (prm_get_integer_value (PRM_ID_HA_MODE))
```
eido5, line 193: what about changing this expression?
I knew it is referenced from legacy code. But I remember that we usually use expression in the if statement.
```suggestion          
	  if (prm_get_integer_value (PRM_ID_HA_MODE) == true)
```

src/loaddb/load_db.c
```diff
@@ -186,13 +186,27 @@ ldr_validate_object_file (const char *argv0, load_args * args)
       return ER_FAILED;
     }
 
-  if (args->cs_mode && (args->load_only == true
-			|| args->index_file.empty () == false
-			|| args->schema_file.empty () == false
-			|| args->schema_file_list.empty () == false || args->trigger_file.empty () == false))
+  if (args->cs_mode)
```
eido5, line 189:
```suggestion
  if  (args->cs_mode == true)
```
the reason is also written in 193 line comment.

src/loaddb/load_db.c
eido5: And what about use in one if statement below two statements?
Could you tell me why you choose this way?
```suggestion
	  if ((prm_get_integer_value (PRM_ID_HA_MODE)==true) &&  (args->load_only == true)
	  ||
```

airnet73: The two doors have different meanings, so they are separated.

eido5: > The two doors have different meanings, so they are separated.
I see.
in case 'prm_get_integer_value (PRM_ID_HA_MODE)' returns nothing but args->load_only == true returns something.

I can guess the case of 'prm_get_integer_value (PRM_ID_HA_MODE) == false'  not the return case.

airnet73: The prm_get_integer_value () function returns TREU or FALSE. I made it so that only OBJECT can be loaded when it is TRUE. If FALSE, there is no need to check the data type. so it was separated
The reason is to prevent schema loading when in HA Mode. See "Purpose" above for details.

src/loaddb/load_db.c
airnet73: I'm usually doing a comparison only if it's FALSE.

eido5 (approved):

PR #4346, opened 2023-05-10
[LETS-666] zero-out unused parts of a slotted page

cristiarg:

This patch contains a small part that is considered correct out of the more extensive modification in #4261 .
The rest - meaning out-of-transactional-log compaction is most probably incorrect - this is still to be investigated by http://jira.cubrid.org/browse/LETS-703.

It only zeroes-out free slotted page space in only two situations:
- when a slot is in-place modified with shorted data
- after a complete page compation, the contiguous free space in page is zeroed-out

This zeroing-out of unused slotted page space will help in the event that page compaction is employed when a slotted page is served by a page servrer.

As a side note. There are probably many other situations where it makes sense to zero-out unused slotted page areas, but these are the low hanging ones.

### DIFF
--- a/src/storage/slotted_page.c
+++ b/src/storage/slotted_page.c
@@ -2432,8 +2432,8 @@ spage_update_record_in_place (PAGE_PTR page_p, SPAGE_HEADER * page_header_p, SPA
   // zero-out the remaining free space
   if (slot_p->record_length < old_record_length)
     {
-      memset ((void *) page_p + slot_p->offset_to_record + slot_p->record_length,
-              0, (old_record_length - slot_p->record_length));
+      memset ((void *) page_p + slot_p->offset_to_record + slot_p->record_length, 0,
+	      (old_record_length - slot_p->record_length));
     }
 
   page_header_p->total_free -= space;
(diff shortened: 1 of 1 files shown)


src/storage/slotted_page.c
```diff
@@ -1266,6 +1266,9 @@ spage_compact (THREAD_ENTRY * thread_p, PAGE_PTR page_p)
   page_header_p->offset_to_free_area = to_offset;
   ASSERT_ALIGN ((char *) page_p + page_header_p->offset_to_free_area, page_header_p->alignment);
 
+  // zero-out the remaining free space
+  memset ((void *) page_p + page_header_p->offset_to_free_area, 0, page_header_p->total_free);
```
wrlawodms: Nit. the same value, but the meaning seems more correct to me.
```suggestion
  memset ((void *) page_p + page_header_p->offset_to_free_area, 0, page_header_p->cont_free);
```

wrlawodms (approved):
Please check the code-style warning.

src/storage/slotted_page.c
cristiarg: Agree. Done.

joohok (approved):

PR #4352, opened 2023-05-11
[CBRD-24802] [PL/CSQL] Showing line and column information of compile time errors in CSQL

hgryoo:
http://jira.cubrid.org/browse/CBRD-24802

- Send the CREATE statement text unchanged from that the user wrote including comments and newlines and blanks (instead of using parser_print_tree ())
- Add err_column attribute in compile_info struct to store the error column number information

Example)
```
csql -u public demodb
```
```
/* comment line 1 */
// comment line 2
/* comment line 3 */
create or replace procedure poo(i int) as
    i int;
    s string;
begin
    // cpp comment
    -- sql comment
    /* c comment
       **
     */
    null;
end;
```
Result
```
ERROR: In line 5, column 10
Stored procedure compile error: I has already been declared in the same scope
```


hgryoo:
LGTM!

hyunikn (approved):

PR #4354, opened 2023-05-11
[CBRD-24741] Provides a way to revive the cub_server process in case of abnormal shutdown

hornetmj:
http://jira.cubrid.org/browse/CBRD-24741

### Purpose
* support the functionality to revive the cub_server when it dies abnormally

### Implementation
* restart cub_server if the following signals are received and cub_server is abnormally terminated
  * SIGABRT
  * SIGILL
  * SIGFPE
  * SIGBUS
  * SIGSEGV
  * SIGSYS
* other signals were not considered in this PR
* please see the jira issue if you wonder about si_code

### Remarks
* about SIGFPE
  * signal handler of SIGFPE is set to SIG_IGN in the css_setup_server_loop(). However, I think that this signal should not be ignored when it occurs. Please share your thoughts on this.

### DIFF
--- a/src/connection/server_support.c
+++ b/src/connection/server_support.c
@@ -324,7 +324,7 @@ css_setup_server_loop (void)
       (void) os_set_signal_handler (SIGFPE, SIG_IGN);
     }
 #else /* LINUX || x86_SOLARIS || HPUX */
-  (void) os_set_signal_handler (SIGFPE, SIG_IGN);
+  // (void) os_set_signal_handler (SIGFPE, SIG_IGN);
 #endif /* LINUX || x86_SOLARIS || HPUX */
 
   if (!IS_INVALID_SOCKET (css_Pipe_to_master))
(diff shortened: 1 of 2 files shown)


joohok:
Oh.. existing crash handler was already reviving the server ! 

I have a question: what if the server keeps repeating the behavior of dying and reviving due to the same reason, by continuously triggering the signal (SIGABRT)? 
This would indicate an environment where the server cannot operate properly, so it would be better to stop reviving it. 

If the cub_master only manages the recent time of server revival, this problem could be resolved. Just like what HA does.
( | current time - recent revival | <= threshold ?  , then do not revive server process)  

What do you think about this?

hornetmj:
@joohok It is a good question. Your suggestion was one of my considerations. But this approach gave up the handling of that case. Because it needs additional permanent information about the previous restart time. And I think that such cases occur very rarely.

Anyway, a new requirement is received. So, drop this PR. Thanks for your feedback.

PR #4370, opened 2023-05-18
[CBRD-24516] rev: replace the information which is printed in "show trace" from aggregate function optimization

beyondykk9:
http://jira.cubrid.org/browse/CBRD-24516

Trace Statistics:
SELECT (time: 0, fetch: 12, ioread: 0)
SCAN (table: dba.t1), (aggregate optimized, time: 0, fetch: 0, ioread: 0, readrows: 0, rows: 0)
-->
SCAN (table: dba.t1), (noscan time: 0, fetch: 0, ioread: 0, readrows: 0, rows: 0, agl:pk_t1_id)

### DIFF
--- a/src/query/query_executor.c
+++ b/src/query/query_executor.c
@@ -23880,13 +23880,15 @@ qexec_evaluate_aggregates_optimize (THREAD_ENTRY * thread_p, AGGREGATE_TYPE * ag
 
 	    }
 
-	  /* agg_index_name is set to show it in the trace information */
-	  spec->s_id.scan_stats.agg_index_name = NULL;
-	  error = heap_get_indexinfo_of_btid (thread_p, &ACCESS_SPEC_CLS_OID (spec), &agg_ptr->btid,
-					      NULL, NULL, NULL, NULL, &spec->s_id.scan_stats.agg_index_name, NULL);
-	  if (error != NO_ERROR)
+	  if (thread_is_on_trace (thread_p))
 	    {
-	      return error;
+	      /* agg_index_name is set to show it in the trace information */
+	      error = heap_get_indexinfo_of_btid (thread_p, &ACCESS_SPEC_CLS_OID (spec), &agg_ptr->btid,
+						  NULL, NULL, NULL, NULL, &spec->s_id.scan_stats.agg_index_name, NULL);
+	      if (error != NO_ERROR)
+		{
+		  return error;
+		}
 	    }
 	}
     }
--- a/src/query/query_executor.c
+++ b/src/query/query_executor.c
@@ -7860,7 +7860,6 @@ qexec_intprt_fnc (THREAD_ENTRY * thread_p, XASL_NODE * xasl, XASL_STATE * xasl_s
 
 	  if (!is_scan_needed)
 	    {
-	      xasl->spec_list->s_id.scan_stats.agg_optimized_scan = true;
 	      return S_SUCCESS;
 	    }
 
@@ -23871,6 +23870,14 @@ qexec_evaluate_aggregates_optimize (THREAD_ENTRY * thread_p, AGGREGATE_TYPE * ag
 		{
 		  error = er_errid ();
 		  return (error == NO_ERROR ? ER_FAILED : error);
+		}
+
+	      spec->s_id.scan_stats.agg_index_name = NULL;
+	      error = heap_get_indexinfo_of_btid (thread_p, &ACCESS_SPEC_CLS_OID (spec), &agg_ptr->btid,
+						  NULL, NULL, NULL, NULL, &spec->s_id.scan_stats.agg_index_name, NULL);
+	      if (error != NO_ERROR)
+		{
+		  return error;
 		}
 	    }
 	}
--- a/src/query/query_executor.c
+++ b/src/query/query_executor.c
@@ -23878,13 +23878,15 @@ qexec_evaluate_aggregates_optimize (THREAD_ENTRY * thread_p, AGGREGATE_TYPE * ag
 		  return (error == NO_ERROR ? ER_FAILED : error);
 		}
 
-	      spec->s_id.scan_stats.agg_index_name = NULL;
-	      error = heap_get_indexinfo_of_btid (thread_p, &ACCESS_SPEC_CLS_OID (spec), &agg_ptr->btid,
-						  NULL, NULL, NULL, NULL, &spec->s_id.scan_stats.agg_index_name, NULL);
-	      if (error != NO_ERROR)
-		{
-		  return error;
-		}
+	    }
+
+	  /* agg_index_name is set to show it in the trace information */
+	  spec->s_id.scan_stats.agg_index_name = NULL;
+	  error = heap_get_indexinfo_of_btid (thread_p, &ACCESS_SPEC_CLS_OID (spec), &agg_ptr->btid,
+					      NULL, NULL, NULL, NULL, &spec->s_id.scan_stats.agg_index_name, NULL);
+	  if (error != NO_ERROR)
+	    {
+	      return error;
 	    }
 	}
     }
--- a/src/query/query_executor.c
+++ b/src/query/query_executor.c
@@ -1802,6 +1802,12 @@ qexec_clear_access_spec_list (THREAD_ENTRY * thread_p, XASL_NODE * xasl_p, ACCES
   pg_cnt = 0;
   for (p = list; p; p = p->next)
     {
+      /* aggregate optimize related should be free */
+      if (p->s_id.scan_stats.agg_index_name != NULL)
+	{
+	  free (p->s_id.scan_stats.agg_index_name);
+	}
+
       memset (&p->s_id.scan_stats, 0, sizeof (SCAN_STATS));
 
       if (p->parts != NULL)
--- a/src/query/scan_manager.c
+++ b/src/query/scan_manager.c
@@ -7732,7 +7732,8 @@ scan_print_stats_json (SCAN_ID * scan_id, json_t * scan_stats)
 	{
 	  if (scan_id->scan_stats.agg_index_name)
 	    {
-	      json_object_set_new (scan_stats, "aggregate optimized,", scan);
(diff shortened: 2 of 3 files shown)


src/query/scan_manager.c
```diff
@@ -7879,16 +7880,27 @@ scan_print_stats_text (FILE * fp, SCAN_ID * scan_id)
...
   switch (scan_id->type)
     {
     case S_HEAP_SCAN:
     case S_LIST_SCAN:
-      fprintf (fp, ", readrows: %llu, rows: %llu)", (unsigned long long int) scan_id->scan_stats.read_rows,
+      fprintf (fp, ", readrows: %llu, rows: %llu", (unsigned long long int) scan_id->scan_stats.read_rows,
 	       (unsigned long long int) scan_id->scan_stats.qualified_rows);
+      if (scan_id->scan_stats.agg_index_name)
+	{
+	  fprintf (fp, ", agl=%s", scan_id->scan_stats.agg_index_name);
```
youngjinj: I think it is consistent and natural for the index name to come immediately after SCAN, as shown below.
```
SCAN (index: participant.fk_participant_host_year), (btree time: 1, ...) (lookup time: 0, rows: 14)
```
And the output should also use ':', not '='.
```
SCAN (agl: dba.t1.pk_t1_id), (btree time: 1, ...) (lookup time: 0, rows: 14)
```

beyondykk9: btree is not a scan method in this case because the it just provides statistics in its header. Therefore, btree-printed should be excluded as a scan method. How about "SCAN (table: dba.t1), (time: 0, fetch: 0, ioread: 0, readrows: 0, rows: 0, agl: pk_t1_id)" ?

src/query/query_executor.c
```diff
@@ -23872,6 +23871,14 @@ qexec_evaluate_aggregates_optimize (THREAD_ENTRY * thread_p, AGGREGATE_TYPE * ag
 		  error = er_errid ();
 		  return (error == NO_ERROR ? ER_FAILED : error);
 		}
+
+	      spec->s_id.scan_stats.agg_index_name = NULL;
+	      error = heap_get_indexinfo_of_btid (thread_p, &ACCESS_SPEC_CLS_OID (spec), &agg_ptr->btid,
```
youngjinj, line 23886: Below is memory allocation with strdup. But it is not free.
```
(SCAN_ID * )->scan_stats.agg_index_name
```
It is being initialized with memset in the qexec_clear_access_spec_list function.
```
static int
qexec_clear_access_spec_list (THREAD_ENTRY * thread_p, XASL_NODE * xasl_p, ACCESS_SPEC_TYPE * list, bool is_final)
{
  ...
  for (p = list; p; p = p->next)
    {
      memset (&p->s_id.scan_stats, 0, sizeof (SCAN_STATS));
...
```

src/query/scan_manager.c
youngjinj: Is there any way below?
```
SCAN (table: dba.t1, agl: pk_t1_id), (time: 0, fetch: 0, ioread: 0, readrows: 0, rows: 0)
```

youngjinj: At least we should use '=' instead of ':'.

src/query/scan_manager.c
```diff
@@ -7730,9 +7730,10 @@ scan_print_stats_json (SCAN_ID * scan_id, json_t * scan_stats)
 
       if (scan_id->type == S_HEAP_SCAN)
 	{
-	  if (scan_id->scan_stats.agg_optimized_scan)
+	  if (scan_id->scan_stats.agg_index_name)
 	    {
-	      json_object_set_new (scan_stats, "aggregate optimized,", scan);
+	      json_object_set_new (scan, "alg_index", json_string (scan_id->scan_stats.agg_index_name));
+	      json_object_set_new (scan_stats, "agl", scan);
```
youngjinj: Is it correct that "agl" is output in the json result?
```
{
  "Trace Statistics": {
    "SELECT": {
      "time": 0,
      "fetch": 0,
      "ioread": 0,
      "SCAN": {
        "access": "table (dba.t1)",
        "agl": { /********** this **********/
          "time": 0,
          "fetch": 0,
          "ioread": 0,
          "readrows": 0,
          "rows": 0,
          "alg_index": "pk_t1_id"
        }
      }
    }
  }
}
```
When printing as text, "agl" does not appear first. I think there should be consistency between the two output formats.
```
Trace Statistics:
  SELECT (time: 0, fetch: 0, ioread: 0)
    SCAN (table: dba.t1), (time: 0, fetch: 0, ioread: 0, readrows: 0, rows: 0, agl=pk_t1_id)
```

youngjinj: No new scan method has been added.
I think it is one of the optimization methods during HEAP scan.
If so, shouldn't "heap" rather than "agl" be output as it is?

youngjinj: It is not an index scan, but it is correct to use an index.
I think it's okay to display the bree time since we need to display the index name anyway.
Instead, all values will be zero.
```
Trace Statistics:
  SELECT (time: 0, fetch: 2, ioread: 0)
    SCAN (index: dba.t1.pk_t1_id), (btree time: 0, fetch: 0, ioread: 0, readkeys: 0, filteredkeys: 0, rows: 0, covered: true, count_only: true)
```

I think it's okay to stick with "covered: true, count_only: true" as it is.
But if we really need to display "agl: true" then I don't think this is too bad.
```
Trace Statistics:
  SELECT (time: 0, fetch: 2, ioread: 0)
    SCAN (index: dba.t1.pk_t1_id), (btree time: 0, fetch: 0, ioread: 0, readkeys: 0, filteredkeys: 0, rows: 0, agl: true)
```

beyondykk9: How about the below ?

```
Trace Statistics:
  SELECT (time: 0, fetch: 2, ioread: 0)
    SCAN (btree-header time: 0, fetch: 0, ioread: 0, readkeys: 0, filteredkeys: 0, rows: 0, agl: dba.t1.pk_t1_id)
```
or
```
Trace Statistics:
  SELECT (time: 0, fetch: 2, ioread: 0)
    SCAN (btree-agl time: 0, fetch: 0, ioread: 0, readkeys: 0, filteredkeys: 0, rows: 0, agl: dba.t1.pk_t1_id)
```

youngjinj:
```
Trace Statistics:
  SELECT (time: 0, fetch: 0, ioread: 0)
    SCAN (table: public.t1), (heap time: 0, fetch: 0, ioread: 0, readrows: 0, rows: 0)
```

Information about the object used is separately written in `(table: public.t1)`.
The cumulative values and option information while scanning are written in `(heap time: 0, fetch: 0, ioread: 0, readrows: 0, rows: 0)`.

It is an opinion that `agl: dba.t1.pk_t1_id` should be written together with `(table: public.t1)` because it is the object information used.
And since it is index information, I think it should be written as `(index: public.t1.pk_t1_id)` instead of `(agl: public.t1.pk_t1_id)`.

`agl: true` is the option information to be used while scanning.
So I think it should be written as `(heap time: 0, fetch: 0, ioread: 0, readrows: 0, rows: 0)`.

The scan method is heap scan, but it uses index information.
So I think `btree time` is correct, not `heap time`.

I disagree with `btree-header` or `btree-agl` before `time`.
Since it is optional information used during scanning, I think it should be added at the end.
And if `btree time` is printed, `btree` does not have to be output redundantly in `btree-agl`.
We just need to print `agl`.

```
Trace Statistics:
  SELECT (time: 0, fetch: 0, ioread: 0)
    SCAN (index: dba.t1.pk_t1_id), (btree time: 0, fetch: 0, ioread: 0, readkeys: 0, filteredkeys: 0, rows: 0, agl: true)
```

youngjinj: I think `header_only: true` or `use_header: true` instead of `agl: true` is intuitive.

beyondykk9: fianally trace print, **"noscan"** shown
```
Trace Statistics:
    SELECT (time: 0, fetch: 2, ioread: 0)
        SCAN (noscan time: 0, fetch: 0, ioread: 0, readkeys: 0, filteredkeys: 0, rows: 0, agl: dba.t1.pk_t1_id)
```
Json format
```
"Trace Statistics": {
    "SELECT": {
      "time": 0,
      "fetch": 2,
      "ioread": 0,
      "SCAN": {
        "access": "table (dba.t1)",
        "noscan": {
          "time": 0,
          "fetch": 0,
          "ioread": 0,
          "readrows": 0,
          "rows": 0,
          "agl": "dba.pk_t1_id"
        }
      }
    }
  }  
```

src/query/query_executor.c
beyondykk9: I fixed next commit.

src/query/query_executor.c
```diff
@@ -23872,6 +23877,16 @@ qexec_evaluate_aggregates_optimize (THREAD_ENTRY * thread_p, AGGREGATE_TYPE * ag
...
+	    }
+
+	  /* agg_index_name is set to show it in the trace information */
+	  spec->s_id.scan_stats.agg_index_name = NULL;
+	  error = heap_get_indexinfo_of_btid (thread_p, &ACCESS_SPEC_CLS_OID (spec), &agg_ptr->btid,
+					      NULL, NULL, NULL, NULL, &spec->s_id.scan_stats.agg_index_name, NULL);
+	  if (error != NO_ERROR)
+	    {
+	      return error;
 	    }
```
youngjinj: This is unnecessary if no trace is output.
```suggestion
	  if (thread_is_on_trace (thread_p))
	    {
	      /* agg_index_name is set to show it in the trace information */
	      spec->s_id.scan_stats.agg_index_name = NULL;
	      error = heap_get_indexinfo_of_btid (thread_p, &ACCESS_SPEC_CLS_OID (spec), &agg_ptr->btid,
						  NULL, NULL, NULL, NULL, &spec->s_id.scan_stats.agg_index_name, NULL);
	      if (error != NO_ERROR)
		{
		  return error;
		}
	    }
```

beyondykk9: OK, I aggree.

youngjinj (approved):

PR #4371, opened 2023-05-18
[CBRD-24810] Avoid repeated memory reallocation in the parser for a big PL/CSQL programs

hyunikn:
http://jira.cubrid.org/browse/CBRD-24810

avoided memory allocation and copy by just figuring out the length of the PL/CSQL text and copying text of that length from original buffer

### DIFF
--- a/src/parser/csql_grammar.y
+++ b/src/parser/csql_grammar.y
@@ -309,7 +309,7 @@ typedef enum
 
 FUNCTION_MAP *keyword_offset (const char *name);
 
-static PT_NODE* get_string_literal_node(const char* str);
+static PT_NODE* get_string_literal_node(const char* str, const int opt_str_size);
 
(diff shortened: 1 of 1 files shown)


src/parser/csql_grammar.y
```diff
@@ -27854,3 +27789,38 @@ pt_ct_check_select (char* p, char *perr_msg)
    sprintf(perr_msg, "Only SELECT statements are supported.");
    return false;
 }
+
+static PT_NODE*
+get_string_literal_node(const char* str, const int opt_str_size) {
```
hgryoo: naming: how about get_sting_literal_node () -> pt_create_string_literal_node_w_charset_coll ()?

hyunikn: OK. thank you

hgryoo (approved):

PR #4375, opened 2023-05-19
[CBRD-24815] The precision of the NUMERIC type is set differently for SMALLINT, INT, and BIGINT types

youngjinj:
http://jira.cubrid.org/browse/CBRD-24815

This is about how to set the precision and scale when changing to NUMERIC type. 

1. Precision
   - AS-IS
      ```
      /* src/parser/type_checking.c - pt_coerce_expression_argument */

      bigint, int, smallint -> numeric (15, ?) /* DB_DEFAULT_NUMERIC_PRECISION */ 
      ```
   - TO-BE
      ```
      /* src/parser/type_checking.c - pt_coerce_expression_argument */

      bigint   -> numeric (19, ?) /* DB_BIGINT_PRECISION */
      int      -> numeric (10, ?) /* DB_INTEGER_PRECISION */
      smallint -> numeric (5, ?)  /* DB_SMALLINT_PRECISION */
      ```
2. Use macro variables instead of magic numbers.


src/parser/parse_tree_cl.c
```diff
@@ -6592,7 +6592,8 @@ pt_print_attr_def (PARSER_CONTEXT * parser, PT_NODE * p)
       if (p->data_type)
 	{
 	  /* only show non-default parameter */
-	  if (p->data_type->info.data_type.precision != 15 || p->data_type->info.data_type.dec_precision != 0)
+	  if (p->data_type->info.data_type.precision != DB_DEFAULT_NUMERIC_PRECISION
+	      || p->data_type->info.data_type.dec_precision != DB_DEFAULT_NUMERIC_SCALE)
```
ctshim, line 6596: The condition of this if statement is strange.
Isn't that always true?

youngjinj: This is true only for NUMERIC (15, 0) type.

src/parser/parse_tree_cl.c
```diff
@@ -8352,7 +8353,8 @@ pt_print_datatype (PARSER_CONTEXT * parser, PT_NODE * p)
 
     case PT_TYPE_NUMERIC:
       q = pt_append_nulstring (parser, q, pt_show_type_enum (p->type_enum));
-      if (p->info.data_type.precision != 15 || p->info.data_type.dec_precision != 0)
+      if (p->info.data_type.precision != DB_DEFAULT_NUMERIC_PRECISION
+	  || p->info.data_type.dec_precision != DB_DEFAULT_NUMERIC_SCALE)
```
ctshim, line 8357: Always true if DB_DEFAULT_NUMERIC_PRECISION and DB_DEFAULT_NUMERIC_SCALE are not equal.

src/parser/parse_tree_cl.c
youngjinj: False only if NUMERIC (15, 0). And the intent of the code is correct.

src/parser/parse_tree_cl.c
youngjinj: False only if NUMERIC (15, 0). And the intent of the code is correct.

beyondykk9 (approved):

src/parser/parse_tree_cl.c
ctshim: All right. I was looking at the code wrong
You were comparing different variables with precision and dec_precision.
I mistakenly thought it was the same variable.

src/parser/parse_tree_cl.c
ctshim: All right. I was looking at the code wrong
You were comparing different variables with precision and dec_precision.
I mistakenly thought it was the same variable.

ctshim (approved):

src/parser/parse_tree_cl.c
youngjinj: It is confusing because it is not called scale here, but dec_precision.

src/parser/parse_tree_cl.c
youngjinj: It is confusing because it is not called scale here, but dec_precision.

hyunikn (approved):

hgryoo (approved):

shparkcubrid (approved):

PR #4379, opened 2023-05-22
[CBRD-24820] Mandate 'DBMS_OUTPUT.' before put_line() usages.

hyunikn:
http://jira.cubrid.org/browse/CBRD-24820

. updates PL/CSQL syntax so that it mandates 'DBMS_OUTPUT.' before usages of put_line() for Oracle compatibility.
. Currently, DBMS_OUTPUT is not an actual package. PL/CSQL does not have the concept of package at least for its first release version.
DBMS_OUTPUT is only allowed before put_line procedure call and is just a syntactic "ornament" only to ease migration from Oracle.

### DIFF
--- a/demo/plcsql/demo_hello.sql
+++ b/demo/plcsql/demo_hello.sql
@@ -1,4 +1,4 @@
 create or replace procedure demo_hello() as
 begin
-    put_line('Hello CUBRID PL/CSQL!');
+    DBMS_OUTPUT.put_line('Hello CUBRID PL/CSQL!');
 end;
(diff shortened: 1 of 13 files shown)


hgryoo:
DBMS_OUTPUT.put(), get_line(), get_lines(), enable(), disable() are also required.

hyunikn:
> DBMS_OUTPUT.put(), get_line(), get_lines(), enable(), disable() are also required.

get_lines() cannot be called from PL/CSQL side because there is no "array" type (array of string in this case) in the current Spec.
Other four procedures seems OK to add.

hgryoo (approved):

PR #4383, opened 2023-05-23
[LETS-721] Set system parameters for transaction server automatically when ha_mode is on

joohok:
http://jira.cubrid.org/browse/LETS-721

Purpose

Automatically setting the system parameters for LETS using legacy system parameters set in cubrid_ha.conf and cubrid.conf
page_server_hosts and remote_storage will be set in this issue, and transaction_server_type will be set later.

### DIFF
--- a/src/base/server_type.cpp
+++ b/src/base/server_type.cpp
@@ -38,6 +38,7 @@ passive_tran_server *pts_Gl = nullptr;
 SERVER_TYPE get_server_type_from_config (server_type_config parameter_value);
 transaction_server_type get_transaction_server_type_from_config (transaction_server_type_config parameter_value);
 void setup_tran_server_params_on_single_node_config ();
+void setup_tran_server_params_on_ha_mode ();
 
 static SERVER_TYPE g_server_type = SERVER_TYPE_UNKNOWN;
 static transaction_server_type g_transaction_server_type = transaction_server_type::ACTIVE;
@@ -127,10 +128,16 @@ int init_server_type (const char *db_name)
       g_server_type = get_server_type_from_config (server_type_from_config);
     }
 
-  if (g_server_type == SERVER_TYPE_TRANSACTION && server_type_from_config == server_type_config::SINGLE_NODE)
+  // if ha_mode, then set page_server_hosts, remote_storage, and transaction_server_type
+  if (g_server_type == SERVER_TYPE_TRANSACTION && !HA_DISABLED ())
+    {
+      setup_tran_server_params_on_ha_mode ();
+    }
+  else if (g_server_type == SERVER_TYPE_TRANSACTION && server_type_from_config == server_type_config::SINGLE_NODE)
     {
       setup_tran_server_params_on_single_node_config ();
     }
+
 #if !defined(NDEBUG)
   g_server_type_initialized = true;
 #endif
@@ -199,6 +206,46 @@ void setup_tran_server_params_on_single_node_config ()
   sprintf (page_hosts_new_value, "localhost:%d", prm_get_master_port_id ());
   prm_set_string_value (PRM_ID_PAGE_SERVER_HOSTS, page_hosts_new_value);
   prm_set_bool_value (PRM_ID_REMOTE_STORAGE, true);
+}
+
+void setup_tran_server_params_on_ha_mode ()
+{
+  assert (!HA_DISABLED ());
+
+  char *page_hosts_new_value;
+  constexpr size_t MAX_BUFSIZE = 4096;
+
+  char ha_node_list[MAX_BUFSIZE];
+  char *str, *savep;
+
+  int port_id = prm_get_master_port_id ();
+
+  page_hosts_new_value = (char *) malloc (MAX_BUFSIZE); // free is called by sysprm_final()
+
+  strncpy_bufsize (ha_node_list, prm_get_string_value (PRM_ID_HA_NODE_LIST));
+
+  str = strtok_r (ha_node_list, "@", &savep); // dbname@host1:host2:...
+  str = strtok_r (NULL, ",:", &savep);
+  while (str)
+    {
+      char tmp[MAX_BUFSIZE] = {0};
+      sprintf (tmp, "%s:%d,", str, port_id);
+      strcat (page_hosts_new_value, tmp);
+      str = strtok_r (NULL, " ,:", &savep);
+    }
+
+  page_hosts_new_value[strlen (page_hosts_new_value) - 1] = '\0';
+
+  prm_set_string_value (PRM_ID_PAGE_SERVER_HOSTS, page_hosts_new_value);
+  prm_set_bool_value (PRM_ID_REMOTE_STORAGE, true);
+
+  /* TODO:
+   * *transaction_server_type* has to be determined.
+   * To determine the transaction_server_type, we need to know the node's state,
+   * and to know the node's state, we need to query the cub_master for the node's state.
+   * Therefore, in order to determine the transaction_server_type,
+   * communication with the cub_master must precede the init_server_type() step.
+   */
 }
 
 void finalize_server_type ()
--- a/src/base/server_type.cpp
+++ b/src/base/server_type.cpp
@@ -38,7 +38,7 @@ passive_tran_server *pts_Gl = nullptr;
 SERVER_TYPE get_server_type_from_config (server_type_config parameter_value);
 transaction_server_type get_transaction_server_type_from_config (transaction_server_type_config parameter_value);
 void setup_tran_server_params_on_single_node_config ();
-void setup_tran_server_params_on_ha_mode ();
+int setup_tran_server_params_on_ha_mode ();
 
 static SERVER_TYPE g_server_type = SERVER_TYPE_UNKNOWN;
 static transaction_server_type g_transaction_server_type = transaction_server_type::ACTIVE;
@@ -131,7 +131,7 @@ int init_server_type (const char *db_name)
   // if ha_mode, then set page_server_hosts, remote_storage, and transaction_server_type
   if (g_server_type == SERVER_TYPE_TRANSACTION && !HA_DISABLED ())
     {
-      setup_tran_server_params_on_ha_mode ();
+      er_code = setup_tran_server_params_on_ha_mode ();
     }
   else if (g_server_type == SERVER_TYPE_TRANSACTION && server_type_from_config == server_type_config::SINGLE_NODE)
     {
@@ -208,12 +208,13 @@ void setup_tran_server_params_on_single_node_config ()
   prm_set_bool_value (PRM_ID_REMOTE_STORAGE, true);
 }
 
-void setup_tran_server_params_on_ha_mode ()
+int setup_tran_server_params_on_ha_mode ()
 {
   assert (!HA_DISABLED ());
 
   char *page_hosts_new_value = NULL;
   constexpr size_t MAX_BUFSIZE = 4096;
+  int new_value_size = 0;
 
   char ha_node_list[MAX_BUFSIZE];
   char *str, *savep;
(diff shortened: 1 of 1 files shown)


src/base/server_type.cpp
```diff
@@ -201,6 +208,51 @@ void setup_tran_server_params_on_single_node_config ()
...
+
+  char ha_node_list[MAX_BUFSIZE];
+  char *str, *savep;
+
+  int port_id = prm_get_master_port_id ();
+
+  page_hosts_new_value = (char *) calloc (MAX_BUFSIZE, sizeof (char)); // free is called by sysprm_final()
+  if (page_hosts_new_value == NULL)
+    {
+      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, MAX_BUFSIZE);
```
hornetmj, line 229: When is this error handled?

src/base/server_type.cpp
```diff
@@ -201,6 +208,51 @@ void setup_tran_server_params_on_single_node_config ()
...
+  strncpy_bufsize (ha_node_list, prm_get_string_value (PRM_ID_HA_NODE_LIST));
+
+  str = strtok_r (ha_node_list, "@", &savep); // dbname@host1:host2:...
+  str = strtok_r (NULL, ",:", &savep);
+  while (str)
+    {
+      char tmp[MAX_BUFSIZE] = {0};
+      sprintf (tmp, "%s:%d,", str, port_id);
+      strcat (page_hosts_new_value, tmp);
+      str = strtok_r (NULL, " ,:", &savep);
```
hornetmj: Is intended? 233 line has not space delimiter but, this line have it.

src/base/server_type.cpp
```diff
@@ -201,6 +208,51 @@ void setup_tran_server_params_on_single_node_config ()
...
+    }
+
+  page_hosts_new_value[strlen (page_hosts_new_value) - 1] = '\0';
```
hornetmj: Does not truncate the last character?

hornetmj: And overflow check is needed.

src/base/server_type.cpp
joohok: It seems that I forgot to include error handling in the process of getting the code from `setup_tran_server_params_on_single_node_config()`. I will fix it.

src/base/server_type.cpp
joohok: Oh.. Thanks. space delimiter is also needed in 233L

src/base/server_type.cpp
joohok: I intentionally truncated the last character because, while looping through above code, there would likely be a ',' at the end. 

I will add the code to handle buffer overflow.

hornetmj: Related to this "%s:%d,". I understand.

joohok: Done.

src/base/server_type.cpp
joohok: error code will be handled at the caller (init_server_type ())

src/base/server_type.cpp
```diff
@@ -201,6 +208,73 @@ void setup_tran_server_params_on_single_node_config ()
...
+  new_value_size = MAX_BUFSIZE;
+
+  strncpy_bufsize (ha_node_list, prm_get_string_value (PRM_ID_HA_NODE_LIST));
+
+  str = strtok_r (ha_node_list, "@", &savep); // dbname@host1:host2:...
+  str = strtok_r (NULL, " ,:", &savep);
+  while (str)
+    {
+      char tmp[MAX_BUFSIZE] = {0};
+      sprintf (tmp, "%s:%d,", str, port_id);
```
wrlawodms: Does it mean that all PSes have to use the same port as the ATS's?

src/base/server_type.cpp
```diff
@@ -201,6 +208,73 @@ void setup_tran_server_params_on_single_node_config ()
...
+    {
+      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, MAX_BUFSIZE);
+      return ER_OUT_OF_VIRTUAL_MEMORY;
+    }
+  new_value_size = MAX_BUFSIZE;
+
+  strncpy_bufsize (ha_node_list, prm_get_string_value (PRM_ID_HA_NODE_LIST));
+
+  str = strtok_r (ha_node_list, "@", &savep); // dbname@host1:host2:...
+  str = strtok_r (NULL, " ,:", &savep);
```
wrlawodms: The manual says that multiple hosts are given with either `;` or `,`. Can the space also separate them?

src/base/server_type.cpp
```diff
@@ -201,6 +208,73 @@ void setup_tran_server_params_on_single_node_config ()
...
+  while (str)
+    {
+      char tmp[MAX_BUFSIZE] = {0};
```
wrlawodms: I think two `tmp` should have a different name since it's confusing to read. like `host` and `new_page_hosts` or whatever.

src/base/server_type.cpp
```diff
@@ -201,6 +208,73 @@ void setup_tran_server_params_on_single_node_config ()
...
+
+  page_hosts_new_value = (char *) calloc (MAX_BUFSIZE, sizeof (char)); // free is called by sysprm_final()
+  if (page_hosts_new_value == NULL)
+    {
+      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, MAX_BUFSIZE);
+      return ER_OUT_OF_VIRTUAL_MEMORY;
+    }
+  new_value_size = MAX_BUFSIZE;
+
+  strncpy_bufsize (ha_node_list, prm_get_string_value (PRM_ID_HA_NODE_LIST));
```
wrlawodms, line 234: Q. What if `PRM_ID_HA_NODE_LIST` is not set and `ha_mode` is enabled?

wrlawodms (commented):
What about others described in the issue? Are they just ignored? 
- ha_db_list
- ha_port_id
- ha_ping_hosts

src/base/server_type.cpp
joohok: `localhost@localhost` is set by default.
So, page server hosts may be set to `localhost:1523`.

src/base/server_type.cpp
joohok: Oh.. you are right.
According to the manual, spaces should not be included in the delimiter. 

However, most of the legacy code that separates ha_node_list using the delimiter " :,".  (`hb_cluster_load_group_and_node_list ()`, `util_split_ha_node ()`, ...)

I will make a note of this and make sure to modify it together when the corresponding part is updated by newHA.

src/base/server_type.cpp
joohok: OK..!

src/base/server_type.cpp
joohok: Yes, it can be a restriction.
In oldHA, it was also required for all nodes to have the same value for the port ID (same master port id)
While this issue could be improved in newHA, but I think it is better to not make any modifications for now.

joohok:
> What about others described in the issue? Are they just ignored?
> 
> * ha_db_list
> * ha_port_id
> * ha_ping_hosts

System parameters you mentioned are set by users and will be used in newHA.

Other than the parameters specified in the issue, they will not be used. 
These parameters are mostly used in `copylogdb` and `applylogdb`, and instead of removing the system parameters from the code, the intention is to eliminate the sections where the system parameters are used, effectively ignoring them.

wrlawodms:
> These parameters are mostly used in copylogdb and applylogdb, and instead of removing the system parameters from the code, the intention is to eliminate the sections where the system parameters are used, effectively ignoring them.

OK, I'v got it. I think we should make an issue to deprecate them later.

src/base/server_type.cpp
wrlawodms: the default value of it, `prm_ha_node_list_default`, is NULL, but it's set to `localhost` somewhere?

wrlawodms (approved):

src/base/server_type.cpp
joohok: In `prm_tune_parameters ()`, if `ha_mode` is set to true, then it set the value of `ha_node_list` to `localhost@localhost`.
`prm_tune_parameters ()` is called in `sysprm_load_and_init ()`, and `sysprm_load_and_init ()` is the function that load the system parameters.

src/base/server_type.cpp
```diff
@@ -201,6 +212,73 @@ void setup_tran_server_params_on_single_node_config ()
   prm_set_bool_value (PRM_ID_REMOTE_STORAGE, true);
 }
 
+int setup_tran_server_params_on_ha_mode ()
+{
+  assert (!HA_DISABLED ());
+
```
hornetmj:
```suggestion
```

src/base/server_type.cpp
```diff
@@ -201,6 +212,73 @@ void setup_tran_server_params_on_single_node_config ()
...
+  strncpy_bufsize (ha_node_list, prm_get_string_value (PRM_ID_HA_NODE_LIST));
+
+  str = strtok_r (ha_node_list, "@", &savep); // dbname@host1:host2:...
+  str = strtok_r (NULL, ",:", &savep);
+  while (str)
+    {
+      char page_server_host[MAX_BUFSIZE] = {0};
+      sprintf (page_server_host, "%s:%d,", str, port_id);
+
+      if (strlen (page_server_host) + strlen (page_server_host_list) > list_size)
```
hornetmj:
```suggestion
      if (strlen (page_server_host) + strlen (page_server_host_list) >= list_size)
```

hornetmj: for null termination.

src/base/server_type.cpp
```diff
@@ -201,6 +212,73 @@ void setup_tran_server_params_on_single_node_config ()
...
+	  char *tmp = (char *) realloc (page_server_host_list, list_size + MAX_BUFSIZE);
+	  if (tmp == NULL)
+	    {
+	      free_and_init (page_server_host_list);
+
+	      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, list_size + MAX_BUFSIZE);
+	      return ER_OUT_OF_VIRTUAL_MEMORY;
+	    }
+
+	  memset (tmp + list_size, 0, MAX_BUFSIZE);
```
hornetmj: It is not necessary.

src/base/server_type.cpp
joohok: Done!

src/base/server_type.cpp
joohok: Removed!

src/base/server_type.cpp
joohok: Done!

hornetmj:
> > These parameters are mostly used in copylogdb and applylogdb, and instead of removing the system parameters from the code, the intention is to eliminate the sections where the system parameters are used, effectively ignoring them.
> 
> OK, I'v got it. I think we should make an issue to deprecate them later.

Please let us know which jira issue is handle this.

hornetmj (approved):

joohok:
> > > These parameters are mostly used in copylogdb and applylogdb, and instead of removing the system parameters from the code, the intention is to eliminate the sections where the system parameters are used, effectively ignoring them.
> > 
> > 
> > OK, I'v got it. I think we should make an issue to deprecate them later.
> 
> Please let us know which jira issue is handle this.

I will handle this in http://jira.cubrid.org/browse/LETS-745

PR #4384, opened 2023-05-24
[CBRD-24824] change /etc/profile.d/cubrid.csh to 644

kisoo-han:
http://jira.cubrid.org/browse/CBRD-24484

**Description**
* Change file protection mode of **/etc/profile.d/cubrid.csh** from 755 to **644**


kisoo-han:
incorrect jira issue number referenced, I will close this.

PR #4389, opened 2023-05-24
[LETS-698] Fail to register request once response_broker terminated

wrlawodms:
http://jira.cubrid.org/browse/LETS-698

- register_request () returns the error when the response broker has been terminated.
- It can be seen as what was missed in d7c5db015c.

### DIFF
--- a/src/communication/response_broker.hpp
+++ b/src/communication/response_broker.hpp
@@ -57,7 +57,7 @@ namespace cubcomm
       response_broker &operator = (const response_broker &) = delete;
       response_broker &operator = (response_broker &&) = delete;
 
-      void register_request (response_sequence_number a_rsn);
+      T_ERROR register_request (response_sequence_number a_rsn);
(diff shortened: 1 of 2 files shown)


cristiarg (approved):

joohok (approved):

src/communication/response_broker.hpp
```diff
@@ -194,17 +194,22 @@ namespace cubcomm
...
+  T_ERROR
   response_broker<T_PAYLOAD, T_ERROR>::bucket::register_request (response_sequence_number a_rsn)
   {
     {
       std::lock_guard<std::mutex> lk_guard (m_mutex);
 
-      assert (!m_terminate);
+      if (m_terminate)
+	{
+	  return m_error;
```
hornetmj, line 205: Does it return ERROR_ON_WRITE as initial value?

wrlawodms: Yes, it does. It's the same as what get_response() below returns if it spots the termination.

hornetmj (approved):

PR #4390, opened 2023-05-24
[LETS-719] [LETS-720] Refactor to decouple [dis]connection and the main connection

wrlawodms:
http://jira.cubrid.org/browse/LETS-719

http://jira.cubrid.org/browse/LETS-720

Note that this PR includes both inflating the connection_handler vector in advance and changing where to change the main connection. These both are for decoupling. I was going to split them into two PRs, but one required adding logic and code which are bound to be removed in the other and ensuring it works, so I've combined them. For LETS-720 in this PR, only `push_request/send_receive` are changed and the main connection change is removed when handling the disconnect msg.

**Implementation**
- Inflating the conneciton_handler vector before connecting to page servers and they are not transient, which means each connection_handler lasts even the connection is closed. 
  - The main connection is not the first element in the vector any more, but the `m_main_conn` points to the main connection.
  - The entries are added before `connect_to_page_server ()` and the connection is established and set using `set_connection()` later.
  - The ctor of connection_handlers and `create_connection_handler` are modified accordingly.
  - Some functions have been modified accordingly: `compute_consensus_lsa(), is_page_server_connected()`
- Use `std::async` instead of async_disconnect_handler to do more jobs asynchronously. It gives more flexibility and I think it's manageable enough, we can ensure a job is done by `std::future`.
  - future.get() (like join()) waits for the async job to be done in the dtor of connection_handler. Other existing jobs, stopping threads, have been removed since they are called in the dtor of each class suing the threads.
  - remove `disconnect_page_server_async`
  - use the async disconnection even in the `disconnect_all_page_servers ` to go through one routine for disconnection.
- Every connection_handler has a mutex to avoid destroying m_conn (disconnection) while using it.
- Change the main connection when not the previous one is disconnected, but a communication function, push_request or send_receive, runs into the disconnected connection_handler. See LETS-720.
  - Now, tran_server::push_request also changes the main connection and retries. For that, connection_handler::push_request can return an error.

It makes git diff unreadable to add and remove some functions. I don't know how to make it pretty :(. Tell me you know how.


src/server/tran_server.cpp
```diff
@@ -142,45 +140,53 @@ tran_server::boot (const char *db_name)
...
 {
-  std::shared_lock<std::shared_mutex> s_lock (m_page_server_conn_vec_mtx);
-  if (m_page_server_conn_vec.empty ())
+  int err_code = NO_ERROR;
+  auto slock = std::shared_lock<std::shared_mutex> { m_main_conn_mtx };
+  while (true)
     {
-      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_CONN_NO_PAGE_SERVER_AVAILABLE, 0);
-      return; // All connections have been disconnected already
+      err_code = m_main_conn->push_request (reqid, std::move (payload));
```
cristiarg: I don't think this approach will work with more than one iteration.
Once the `payload` is moved once into a `push_request` it is lost, a second `std::move` will actually move garbage.

Copying the `payload` can induce a performance hazard but - at least for now - we have to do it.
We will also refactor this to use a more pointer-y approach (rather than moving the underlying data).

The problem that this code tries to avoid is that the connection can initiate a shutdown (as a response coming from one PS) at any moment - even during when this request is in-flight (I think the problem is the same as described in http://jira.cubrid.org/browse/LETS-730).

My opinion is: can we withhold processing disconnect requests from a PS as long as a connection handler has in-flight requests? Or does this overcomplicate things?

src/server/tran_server.cpp
```diff
@@ -362,75 +370,56 @@ void
...
-    return conn_uptr.get () == conn;
+    return conn->is_connected ();
   });
 
-  if (conn_it == conn_vec.end ())
+  if (main_conn_cand == conn_vec.cend())
     {
-      return; // the connection is already cleared in disconnect_all_page_servers()
+      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_CONN_NO_PAGE_SERVER_AVAILABLE, 0);
+      return ER_CONN_NO_PAGE_SERVER_AVAILABLE;
```
cristiarg, line 401: Shouldn't `m_main_conn` be set to `nullptr`?
Not setting `m_main_conn` to null, such as to have a dead-end in case no PS's are still alive, can end up in busy loops in `tran_server::push_request` which will constantly try to reset the main connection.

src/server/tran_server.cpp
```diff
@@ -362,75 +370,56 @@ void
...
-  m_main_conn_cv.notify_all ();
-
-  // Wait until all disconnection reuqests are handled.
-  m_async_disconnect_handler.terminate ();
+  for (auto &conn : m_page_server_conn_vec)
+    {
+      constexpr bool with_disconnect_msg = true;
+      conn->disconnect_async (with_disconnect_msg);
+    }
+  m_page_server_conn_vec.clear ();
```
cristiarg: `reset_main_connection` somewhere around here?

src/server/tran_server.cpp
cristiarg: Also, given that `m_main_conn` might become `nullptr` in extreme cases (see comment below), verification for null before using it?

src/server/tran_server.cpp
```diff
@@ -362,75 +370,56 @@ void
...
 bool
 tran_server::is_page_server_connected () const
 {
   assert_is_tran_server ();
-
-  return !m_page_server_conn_vec.empty ();
+  return std::any_of (m_page_server_conn_vec.cbegin (), m_page_server_conn_vec.cend (),	[] (const auto &conn)
+  {
+    return conn->is_connected ();
+  });
```
cristiarg, line 426:
```suggestion
  return std::any_of (m_page_server_conn_vec.cbegin (), m_page_server_conn_vec.cend (),	[] (const auto &conn)
  {
    return conn->is_connected ();
  }) && (m_main_conn != nullptr);
```
or at least an assert

src/server/tran_server.cpp
```diff
@@ -461,12 +459,35 @@ tran_server::connection_handler::connection_handler (cubcomm::channel &&chn, tra
...
+tran_server::connection_handler::disconnect_async (bool with_disc_msg)
+{
+  auto lockg = std::lock_guard <std::mutex> { m_disconn_mtx };
+
+  if (m_disconn_future.valid ())
+    {
+      return; // already done by other
+    }
+
+  m_disconn_future = std::async (std::launch::async, [this, &with_disc_msg]
```
cristiarg: nice :)

cristiarg (commented):
partial review

src/server/tran_server.cpp
wrlawodms: Agreed. How about just `m_main_conn = nullptr;`?

src/server/tran_server.cpp
wrlawodms: I think `return ER_CONN_NO_PAGE_SERVER_AVAILABLE` will break the loop, but I agree to set the `m_main_conn` to nullptr for safety.

src/server/tran_server.cpp
wrlawodms: > Also, given that m_main_conn might become nullptr in extreme cases (see comment below), verification for null before using it?

Do you mean the case where `reset_main_connection` set it to nullptr, and someone calles `push_request` again without handling the `ER_CONN_NO_PAGE_SERVER_AVAILABLE` error?

wrlawodms: > Copying the payload can induce a performance hazard but - at least for now - we have to do it.
We will also refactor this to use a more pointer-y approach (rather than moving the underlying data).

Oh, I missed it. Thank you. We may copy the string in this function for now.

> My opinion is: can we withhold processing disconnect requests from a PS as long as a connection handler has in-flight requests? Or does this overcomplicate things?

Actually, I considered a similar approach: Change the disconnection protocol to "All responses are served before handling disconnection request". It may remove the `stop_response_broker` and keep logic simpler. It might be a better approach in terms of safety and simplicity but also has downsides.
- It can be delayed to process the disconnection request.
- There can be a starvation of disconnection if requests are pushed continuously, and it amplifies the delay problem.

So, I've just chosen to keep the existing protocol to make use of existing logic. And about "Copying the payload" problem, it will be still needed to deal with the abnormal disconnection. The `send_recv` can fail abnormally, and it has to be retried somewhere, which requires the `payload` again. 

I lean forward to keep the existing code to make progress. Of course, it has to be fixed if flawed, though. What do you think?

src/server/tran_server.cpp
wrlawodms: On second thought, it could be a way to not set the `m_main_conn` to `nullptr` in any case after it's set while booting. Then, we don't need to consider if it's `nullptr` or not. What do you think? It also means that your feedback in `is_page_server_connected` and `disconnect_all_page_servers` is not applied.

hornetmj (approved):
LGTM.

joohok (approved):

src/server/tran_server.cpp
```diff
@@ -142,45 +140,53 @@ tran_server::boot (const char *db_name)
...
 {
-  std::shared_lock<std::shared_mutex> s_lock (m_page_server_conn_vec_mtx);
-  if (m_page_server_conn_vec.empty ())
+  int err_code = NO_ERROR;
+  auto slock = std::shared_lock<std::shared_mutex> { m_main_conn_mtx };
+  while (true)
     {
-      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_CONN_NO_PAGE_SERVER_AVAILABLE, 0);
-      return; // All connections have been disconnected already
+      err_code = m_main_conn->push_request (reqid, std::string (payload));
```
cristiarg, line 147: in the case of ATS + serialized log prior info - where the messages can actually be very large - there will be a number of (2 * number of page server) copies of the serialized data:
- one for each PS in `prior_sender::send_list`
```
    for (auto &sink : m_sink_hooks)
      {
	(*sink) (std::string (message));
      }
```
- and again, for each PS in `tran_server::push_request` - here

Added http://jira.cubrid.org/browse/LETS-744 to keep track of this issue.

src/server/tran_server.cpp
cristiarg: yes, at least that

src/server/tran_server.cpp
```diff
@@ -469,16 +470,22 @@ tran_server::connection_handler::~connection_handler ()
 void
 tran_server::connection_handler::disconnect_async (bool with_disc_msg)
 {
-  auto lockg = std::lock_guard <std::mutex> { m_disconn_mtx };
+  auto lockg = std::lock_guard <std::shared_mutex> { m_state_mtx };
 
-  if (m_disconn_future.valid ())
+  assert (m_state != state::CONNECTING);
```
joohok: Q. is it possible to call this function with the state value as `state::CONNECTING`? 
It seems possible that disconnecting the TS-PS connection which is being established. 
(For instance, A situation where TS is started and immediately stopped.)

If so, separate handling might be required specifically for `state::CONNECTING`.
For example, it might be necessary to wait until state::CONNECTED is reached.

wrlawodms: It seems impossible because the OPEN->CONNECTING->CONNECTED transition happens at once with m_state_mtx acquired. Do you happen to see a room for that to happen?

wrlawodms: see `set_connection`

joohok: Ah, I understand. I got a little confused about shared_mutex

src/server/tran_server.hpp
```diff
@@ -103,22 +106,58 @@ class tran_server
...
+	 * | CONNECTING    | X            | O           | X            | △          | X          |
+	 * | CONNECTED     | X            | O           | O            | O          | O          |
+	 * | DISCONNECTING | X            | O           | X            | △          | X          |
+	 * +---------------+--------------+-------------+--------------+------------+------------+
+	 *
+	 * m_state and m_conn are coupled and mutexes for them are managed carefully to provide above rules.
+	 */
+	enum class state
+	{
+	  OPEN,
```
cristiarg: Nit.
Isn't it better to use `IDLE` or `EMPTY` instead of `OPEN`?

src/server/tran_server.cpp
```diff
@@ -452,21 +443,81 @@ tran_server::connection_handler::connection_handler (cubcomm::channel &&chn, tra
...
+
+  if (m_state != state::CONNECTED)
+    {
+      return; // already disconnected by other
+    }
+
+  m_state = state::DISCONNECTING;
+
+  m_disconn_future = std::async (std::launch::async, [this, &with_disc_msg]
```
cristiarg: don't pass `&with_disk_msg` to the lambda; which, I think, passes by address;
the address can become dangling (the stack might be gone/re-wound) until the future executes the lambda;
pass by value
that's my understanding

src/server/tran_server.hpp
```diff
@@ -127,11 +166,16 @@ class tran_server
 
       private:
 	std::unique_ptr<page_server_conn_t> m_conn;
-	std::atomic<bool> m_is_disconnecting;
+	std::shared_mutex m_conn_mtx;
+
+	state m_state;
+	std::shared_mutex m_state_mtx;
```
cristiarg, line 171: I don't want to "whistle in the church". I appreciate your effort on this but.

By looking at both `m_conn_mtx` and `m_state_mtx` at this point they are almost parallel:
- `connection_handler::set_connection`: write lock both of them, state_mtx outer, conn_mtx inner
- `connection_handler::disconnect_async`: write lock lock on state_mtx
- `connection_handler::disconnect_async` - async lambda: write lock both of them, state_mtx outer, conn_mtx inner
- `connection_handler::wait_async_disconnection`: write lock state_mtx
  - btw: this should have been a shared_lock because value is only read
- `connection_handler::push_request` and `connection_handler::send_receive`: shared (read-only) lock both of them, state_mtx outer, conn_mtx inner

By synthetically looking at this, I would simplify and make it only one mutex.

I think you can go forward with this PR and simplify in a separate one. Or... your choice.

src/server/tran_server.cpp
```diff
@@ -487,55 +538,93 @@ tran_server::connection_handler::get_request_handlers ()
...
+  // Make sure that m_conn is not nullptr
   return m_conn->get_underlying_channel_id ();
 }
 
 bool
-tran_server::connection_handler::is_disconnecting () const
+tran_server::connection_handler::is_connected () const
 {
-  return m_is_disconnecting.load ();
+  return m_state == state::CONNECTED;
```
cristiarg, line 609: shared lock when reading state?

src/server/tran_server.cpp
```diff
@@ -487,55 +538,93 @@ tran_server::connection_handler::get_request_handlers ()
...
+  if (m_state != state::CONNECTED)
+    {
+      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_CONN_PAGE_SERVER_CANNOT_BE_REACHED, 0);
+      return ER_CONN_PAGE_SERVER_CANNOT_BE_REACHED;
+    }
+
+  // state::CONNECTED guarantees that the internal m_node is not nullptr.
+  auto slock_conn = std::shared_lock<std::shared_mutex> { m_conn_mtx };
+  assert (m_conn != nullptr); // because m_conn is set with both m_state_mtx and m_conn_mtx locked.
+  slock_state.unlock ();
```
cristiarg, line 565:
```suggestion
  slock_conn.unlock ();
  slock_state.unlock ();
```
I don't think that without there is some inconsistency, but just so it "looks" nice. :)
Same in `send_receive`.

src/server/tran_server.cpp
```diff
@@ -487,55 +538,93 @@ tran_server::connection_handler::get_request_handlers ()
...
+  auto slock_state = std::shared_lock<std::shared_mutex> { m_state_mtx };
+
+  if (m_state != state::CONNECTED)
+    {
+      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_CONN_PAGE_SERVER_CANNOT_BE_REACHED, 0);
+      return ER_CONN_PAGE_SERVER_CANNOT_BE_REACHED;
+    }
+
+  // state::CONNECTED guarantees that the internal m_node is not nullptr.
+  auto slock_conn = std::shared_lock<std::shared_mutex> { m_conn_mtx };
```
cristiarg, line 564: If `m_conn_mtx` is locked just for the assert, the assert is `nop` in release mode. Then the lock is useless (and consumes resources).
Same remark for `send_receive`.

src/server/tran_server.cpp
```diff
@@ -452,21 +443,81 @@ tran_server::connection_handler::connection_handler (cubcomm::channel &&chn, tra
...
+
+void
+tran_server::connection_handler::wait_async_disconnection ()
+{
+  if (m_disconn_future.valid ())
+    {
+      m_disconn_future.get ();
+    }
+
+  auto lockg = std::lock_guard<std::shared_mutex> { m_state_mtx };
```
cristiarg: read only lock?

src/server/tran_server.cpp
```diff
@@ -487,55 +538,93 @@ tran_server::connection_handler::get_request_handlers ()
...
+  push_request_internal (reqid, std::move (payload));
+
+  return NO_ERROR;
+}
+
+int
+tran_server::connection_handler::send_receive_internal (tran_to_page_request reqid, std::string &&payload_in,
+    std::string &payload_out)
+{
+  // make sure the m_conn is not nullptr;
```
cristiarg: Remove comment?
Is this a TODO? I do not understand the comment.
```suggestion
```

src/server/tran_server.cpp
```diff
@@ -487,55 +538,93 @@ tran_server::connection_handler::get_request_handlers ()
...
-  m_ts.disconnect_page_server_async (this);
+  constexpr bool with_disconnect_msg = true;
+  disconnect_async (with_disconnect_msg);
 }
 
 void
-tran_server::connection_handler::push_request (tran_to_page_request reqid, std::string &&payload)
+tran_server::connection_handler::push_request_internal (tran_to_page_request reqid, std::string &&payload)
 {
+  // make sure the m_conn is not nullptr;
```
cristiarg: Is this a TODO? I do not understand the comment.

src/server/tran_server.hpp
```diff
@@ -103,22 +106,58 @@ class tran_server
...
+	 *
+	 * The allowed operations for each state are:
+	 * +---------------+--------------+-------------+--------------+------------+------------+
+	 * |     state     | accept a new | request     | request      | m_conn     | main       |
+	 * |               | connection   | from inside | from outside | != nullptr | connection |
+	 * +---------------+--------------+-------------+--------------+------------+------------+
+	 * | OPEN          | O            | X           | X            | X          | X          |
+	 * | CONNECTING    | X            | O           | X            | △          | X          |
+	 * | CONNECTED     | X            | O           | O            | O          | O          |
+	 * | DISCONNECTING | X            | O           | X            | △          | X          |
```
cristiarg: What do `O`, `△` mean?

src/server/tran_server.cpp
wrlawodms: I should've done. Thank you 👍

src/server/tran_server.cpp
wrlawodms: Oh, by mistake. Thank you so much.

src/server/tran_server.cpp
wrlawodms: fixed.

src/server/tran_server.cpp
wrlawodms: It's a pre-condition. it's warning to a user to use this internal method. I think `assert (m_conn != nullptr)` is no good since it's checked right below while accessing `m_conn`.

src/server/tran_server.cpp
wrlawodms: No, it's not just for the assert. If it's not locked, the `m_conn` can be `nullptr` since it can be disconnected anytime.

src/server/tran_server.cpp
wrlawodms: `slock_state` must be unlocked to allow `disconnect_async` which also locks `m_state_mtx`. It can be called by the request handler, and if this `send_receive` is sleeping with the mutex locked, a deadlock will happen.
```
client A:                       send_receive(m_state_mtx locked) and sleep
the request handler:   
   - SEND_DISCONNECT_REQUEST_MSG: disconenct_async(): blocked by m_state_mtx
   - RESPONSE: serve to client A (will wake up client A, but can't server this for good)
```
And `slock_conn` must be locked, or `m_node` can be set to `nullptr` by disconnecting.

src/server/tran_server.cpp
wrlawodms: the same answered for the `push_request_internal`.

src/server/tran_server.cpp
wrlawodms: hmmm.. I remember that I thought the lock could be acquired outside.. but I can't find that usage. It's better to acquire the lock. Thanks

src/server/active_tran_server.cpp
```diff
@@ -63,24 +63,27 @@ active_tran_server::get_oldest_active_mvccid_from_page_server ()
...
 active_tran_server::compute_consensus_lsa ()
 {
-  const int total_node_cnt = m_connection_list.size ();
+  const int total_node_cnt = m_page_server_conn_vec.size ();
   const int quorum = total_node_cnt / 2 + 1; // For now, it's fixed to the number of the majority.
-  int cur_node_cnt;
+  int cur_node_cnt = 0;
   std::vector<log_lsa> collected_saved_lsa;
 
   {
```
hornetmj:
```suggestion
```

src/server/active_tran_server.cpp
```diff
@@ -63,24 +63,27 @@ active_tran_server::get_oldest_active_mvccid_from_page_server ()
...
       {
 	quorum_consenesus_er_log ("compute_consensus_lsa - Quorum unsatisfied: total node count = %d, curreunt node count = %d, quorum = %d\n",
 				  total_node_cnt, cur_node_cnt, quorum);
 	return NULL_LSA;
       }
-    for (const auto &conn : m_page_server_conn_vec)
-      {
-	collected_saved_lsa.emplace_back (conn->get_saved_lsa ());
-      }
   }
```
hornetmj:
```suggestion
```

src/server/tran_server.hpp
wrlawodms: I really wanted to use only one mutex, but I can't in the end. Where they're not parallel is described in the above answer for `push_request` and `send_receive`. Shortly, it's to allow both `disconnect_async` and ensure `m_conn` is not set to nullptr while `send_receive` waiting for its response. Do you happen to have an idea to solve it with one mutex?

PS. 
> connection_handler::disconnect_async - async lambda: write lock both of them, state_mtx outer, conn_mtx inner

It's also modified to avoid a deadlock, which is described in the code comment, `tran_server::connection_handler::disconnect_async`.

src/server/tran_server.cpp
```diff
@@ -452,21 +443,81 @@ tran_server::connection_handler::connection_handler (cubcomm::channel &&chn, tra
   // Transaction server will use message specific error handlers.
   // Implementation will assert that an error handler is present if needed.
 
-  m_conn.reset (new page_server_conn_t (std::move (chn), std::move (request_handlers), tran_to_page_request::RESPOND,
+  auto lockg = std::lock_guard<std::shared_mutex> { m_state_mtx };
+  assert (m_state == state::OPEN);
+
+  m_state = state::CONNECTING;
```
hornetmj: The transition from CONNECTING to CONNECTED is very short. Separating the states means increasing the complexity, so why not combine the two into one state? If you dis-agree, please explain a bit more about the benefits of splitting the two.

src/server/tran_server.hpp
wrlawodms: Agree. I'll go with `IDLE`.

src/server/tran_server.cpp
```diff
@@ -362,75 +370,60 @@ void
...
-  if (conn_it == conn_vec.end ())
+  if (main_conn_cand == conn_vec.cend())
     {
-      return; // the connection is already cleared in disconnect_all_page_servers()
+      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_CONN_NO_PAGE_SERVER_AVAILABLE, 0);
+      return ER_CONN_NO_PAGE_SERVER_AVAILABLE;
     }
 
-  if (conn_it == conn_vec.begin ())
+  auto ulock = std::unique_lock<std::shared_mutex> { m_main_conn_mtx };
```
hornetmj, line 404: Question:
Can consecutive push_request() or send_receive() calls make a starvation state?

src/server/tran_server.hpp
wrlawodms: O: Allowed, X: not allowed, △: not certain.
I have added the legend about them.

src/server/tran_server.cpp
wrlawodms: Actually, this is about `send_receive`, not about `push_request`. In the case of `push_request`, it is just for code consistency.

src/server/active_tran_server.cpp
wrlawodms: done.

src/server/tran_server.cpp
wrlawodms: All they're bound to want to reset the main connection in the end, and it will be done. So, I don't think the starvation for resetting the main connection happens.

src/server/tran_server.cpp
wrlawodms: Now, yes, the CONNECTING is not open to the outside. It's just defining the state and the `CONNECTING` will be used in the near future when the catch-up process comes in. We can add the state then, but I've just added it while I'm at it.

src/server/tran_server.cpp
wrlawodms: @joohok On second thought, it'd be better to deal with `state::CONNECTING` here for the near future. It's worth it when the catch-up process comes in and the transition, OPEN->CONNECTING->CONNECTED, doesn't happen at once. Thank you.

src/server/tran_server.cpp
cristiarg: By "starvation state" I understand that consecutive/continuous `shared_lock`s on the `m_main_conn_mtx` prevent resetting the main connection.
Yes, but very unlikely.
By comparison to the push/send methods, this one is a veeery rare event.

If the case, then we can use a `std::timed_mutex` and implement a loop here in `reset_main_connection`. This will, at least, allow us to log that starvation has happened.

src/server/tran_server.cpp
cristiarg: I would leave the `CONNECTING` state. That was the idea -  to have a clear image of what state is the handler is.

src/server/tran_server.cpp
wrlawodms: Oh,, true. It can be. Hmm..

wrlawodms: @hornetmj  I think we can leave it as it is now even without an issue to track this. It would be very rare, and we can easily notice it if the problem arises since every request keeps failing. What do you think?

wrlawodms:
To all, the transition has been refined a little bit. just allow CONNECTING to disconnect. It makes more sense, and is needed soon.

	/*
 	 * The internal state of connection_handler. A connection_handler must be in one of those states.
 	 *
 	 *    IDLE ---> CONNECTING ---> CONNECTED ---> DISCONNECTING ---> IDLE
 	 *                   |                     |
 	 *                   +---------------------+
 	 * The allowed operations for each state are:
 	 * +---------------+--------------+--------------+--------------+------------+-------------+
 	 * |     state     | accept a new | send request | send request | m_conn     | set as main |
 	 * |               | connection   | from inside  | from outside | != nullptr | connection  |
 	 * +---------------+--------------+--------------+--------------+------------+-------------+
 	 * | IDLE          | O            | X            | X            | X          | X           |
 	 * | CONNECTING    | X            | O            | X            | O          | X           |
 	 * | CONNECTED     | X            | O            | O            | O          | O           |
 	 * | DISCONNECTING | X            | O            | X            | △          | X           |
 	 * +---------------+--------------+--------------+--------------+------------+-------------+
 	 *

src/server/tran_server.hpp
cristiarg: Fair enough, let's leave it as is.
Profiling should tell us if there are problems.

src/server/tran_server.cpp
cristiarg: Side note.
To enforce the usage of of a function under lock, you can follow this pattern from here:
```
  inline void
  redo_parallel::redo_task::set_unapplied_log_lsa_from_push (const log_lsa &a_log_lsa,
      const std::lock_guard<std::mutex> &)
```
In that function the last argument is a locked mutex. The function does not do anything with it (parameter unnamed) but it enfonces whoever calls the function to supply a locker.
In release code generation, I assume the compiler will optimize out the dispatching of any unused parameters, so there is even no performance impact.

src/server/tran_server.cpp
cristiarg: Ah, now I understand. Sorry I did not look in more detail.

Then I think there are 2 options:
- either do away with functions `..._internal` and just inline their contents
- or apply pattern described in comment https://github.com/CUBRID/cubrid/pull/4390#discussion_r1238261057

src/server/tran_server.cpp
cristiarg: For both of them, the `m_conn_mtx` is needed.
I overlooked the call to `*_internal` functions when I reviewed. Sorry.
- for `push_request` the code translates to :
```
  auto slock_conn = std::shared_lock<std::shared_mutex> { m_conn_mtx };
  assert (m_conn != nullptr); // because m_conn is set with both m_state_mtx and m_conn_mtx locked.
  slock_state.unlock ();

  //push_request_internal (reqid, std::move (payload));
  // make sure the m_conn is not nullptr;
  m_conn->push (reqid, std::move (payload)); // <<------------- slock_conn needed here

```
- for `send_receive` the code transalates to:
```
  auto slock_conn = std::shared_lock<std::shared_mutex> { m_conn_mtx };
  assert (m_conn != nullptr); // because m_conn is set with both m_state_mtx and m_conn_mtx locked.
  slock_state.unlock (); // to allow disconnect_async ()

  //return send_receive_internal (reqid, std::move (payload_in), payload_out);
  // make sure the m_conn is not nullptr;
  // <<------------------ slock_conn needed from here on
  const css_error_code error_code = m_conn->send_recv (reqid, std::move (payload_in), payload_out);
  if (error_code != NO_ERRORS)
    {
      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_CONN_PAGE_SERVER_CANNOT_BE_REACHED, 0);
      return ER_CONN_PAGE_SERVER_CANNOT_BE_REACHED;
    }

  return NO_ERROR;

```

To make it more explicit, from my side, same remark as in comment https://github.com/CUBRID/cubrid/pull/4390/files#r1238267018 .

cristiarg (approved):

cristiarg:
> To all, the transition has been refined a little bit. just allow CONNECTING to disconnect. It makes more sense, and is needed soon.
> 
> ```
> /*
>  * The internal state of connection_handler. A connection_handler must be in one of those states.
>  *
>  *    IDLE ---> CONNECTING ---> CONNECTED ---> DISCONNECTING ---> IDLE
>  *                   |                     |
>  *                   +---------------------+
>  * The allowed operations for each state are:
>  * +---------------+--------------+--------------+--------------+------------+-------------+
>  * |     state     | accept a new | send request | send request | m_conn     | set as main |
>  * |               | connection   | from inside  | from outside | != nullptr | connection  |
>  * +---------------+--------------+--------------+--------------+------------+-------------+
>  * | IDLE          | O            | X            | X            | X          | X           |
>  * | CONNECTING    | X            | O            | X            | O          | X           |
>  * | CONNECTED     | X            | O            | O            | O          | O           |
>  * | DISCONNECTING | X            | O            | X            | △          | X           |
>  * +---------------+--------------+--------------+--------------+------------+-------------+
>  *
> ```

Nit.
In the state diagram, there should be an arrow:
```
	 *    IDLE ---> CONNECTING ---> CONNECTED ---> DISCONNECTING ---> IDLE
	 *                   |                     |^
	 *                   +---------------------+

```

src/server/tran_server.cpp
wrlawodms: I think better to opt for the second once since it will be used in following issues.

src/server/tran_server.cpp
wrlawodms: Oh, exactly what I wanted t know. Thank you for the tip 👍

src/server/tran_server.cpp
wrlawodms: However, the pattern can't handle some issues so I end up doing away the internal functions.
- the lock can be either shared_lock or unique_lock according to its context. No type seems to accept both.
- Different from lock_guard, shared_lock and unique_lock doesn't guarantee that it's locked. It seems to be abled to be checked  with `owns_lock`, though.

PR #4394, opened 2023-05-30
[CBRD-24828] Executor error when using udf() and bind variable in update query

beyondykk9:
http://jira.cubrid.org/browse/CBRD-24828

An error occurred in execution after binding a value in a prepared query including a stored procedure.

```
create table t_user (
        u_no int primary key,
        email varchar(100),
        yn char(1)
);

insert into t_user set
        u_no=123,
        email='info@cubrid.com',
        yn='Y';

prepare st from 'update t_user set email=(select encryptFunc(?)), yn=? where u_no = ?';

execute st using 'info@cubrid.com', 'Y', 123;
```
expected successful execution.

### DIFF
--- a/src/parser/type_checking.c
+++ b/src/parser/type_checking.c
@@ -12120,7 +12120,7 @@ pt_upd_domain_info (PARSER_CONTEXT * parser, PT_NODE * arg1, PT_NODE * arg2, PT_
 	{
 	case PT_TYPE_CHAR:
 	  dt->info.data_type.precision = ((dt->info.data_type.precision > DB_MAX_CHAR_PRECISION)
-					  ? DB_MAX_CHAR_PRECISION : dt->info.data_type.precision);
+					  ? DB_DEFAULT_PRECISION : dt->info.data_type.precision);
 	  break;
 
 	case PT_TYPE_VARCHAR:
@@ -12130,7 +12130,7 @@ pt_upd_domain_info (PARSER_CONTEXT * parser, PT_NODE * arg1, PT_NODE * arg2, PT_
 
 	case PT_TYPE_NCHAR:
 	  dt->info.data_type.precision = ((dt->info.data_type.precision > DB_MAX_NCHAR_PRECISION)
-					  ? DB_MAX_NCHAR_PRECISION : dt->info.data_type.precision);
+					  ? DB_DEFAULT_PRECISION : dt->info.data_type.precision);
 	  break;
 
 	case PT_TYPE_VARNCHAR:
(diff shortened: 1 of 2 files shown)


src/parser/type_checking.c
```diff
@@ -12120,7 +12120,7 @@ pt_upd_domain_info (PARSER_CONTEXT * parser, PT_NODE * arg1, PT_NODE * arg2, PT_
 	{
 	case PT_TYPE_CHAR:
 	  dt->info.data_type.precision = ((dt->info.data_type.precision > DB_MAX_CHAR_PRECISION)
-					  ? DB_MAX_CHAR_PRECISION : dt->info.data_type.precision);
```
hgryoo, line 12123: I'd like to confirm that it is related to db_value_domain_init() sets the precision according to the type and DB_DEFAULT_PRECISION. Am I understanding correctly?
If so, this behavior is tough to connect this concept if you don't know the context about it. Describing this behavior by a comment seems to be required here.

hgryoo:
Also, I'm not sure the NCHAR and VARNCHAR is dead code. But In case I don't know, this part should be updated in the same way before NCHAR nad VARNCHAR routine is removed.

src/parser/type_checking.c
beyondykk9: Ok, I will comment the context in the JIRA issue.

beyondykk9:
> Also, I'm not sure the NCHAR and VARNCHAR is dead code. But In case I don't know, this part should be updated in the same way before NCHAR nad VARNCHAR routine is removed.

Maybe this modified code might be useless, however, I expect that the code will be removed at code refactoring.

hyunikn:
The CREATE TABLE statement causes an error as shown below:
```
csql> create table t_user (
csql>         no int primary key,
csql>         email varchar(100),
csql>         yn char(1)
csql> );

In line 2, column 9,

ERROR: In line 2, column 9 before ' int primary key,
        email varchar(100),
        yn char(...'
Syntax error: unexpected 'no', expecting LIKE

```
Is 'no' a keyword?

hyunikn:
The above reproduction scenario (with updating 'no' to 'n'), and the following one in the related issue CBRD-24734 do not pass the PT_TYPE_CHAR case which is updated in this issue. Does the update has any effect on the scenarios?
```
CREATE OR REPLACE FUNCTION SP_FUNC() RETURN STRING
 AS LANGUAGE JAVA
 NAME 'SP_FUNC.SP_FUNC()  RETURN java.lang.String';
PREPARE st FROM '
 SELECT CASE WHEN ? = 1
             THEN ''1234567890'' || NVL2(?, ?, ''A'')
             ELSE ( SELECT sp_func() FROM dual )
        END AS p
 FROM dual
 ';
EXECUTE st USING 1, null, null;
```

hyunikn (approved):

ctshim (approved):

hgryoo (approved):

src/parser/type_checking.c
hgryoo: Please add a comment in the source file.

PR #4398, opened 2023-05-31
[LETS-724] Define the process types for newHA

joohok:
http://jira.cubrid.org/browse/LETS-724

Purpose

Add process type managed by cub_master
- Transaction Server : HB_PTYPE_TRAN_SERVER
- Page Server : HB_PTYPE_PAGE_SERVER

After removing or replacing all legacy code, 
then I will remove the HB_PTYPE_SERVER, HB_PTYPE_COPYLOGDB, and HB_PTYPE_APPLYLOGDB.

### DIFF
--- a/src/connection/heartbeat.c
+++ b/src/connection/heartbeat.c
@@ -105,6 +105,10 @@ hb_process_type_string (int ptype)
 {
   switch (ptype)
     {
+    case HB_PTYPE_TRAN_SERVER:
+      return HB_PTYPE_TRAN_SERVER_STR;
+    case HB_PTYPE_PAGE_SERVER:
+      return HB_PTYPE_PAGE_SERVER_STR;
     case HB_PTYPE_SERVER:
       return HB_PTYPE_SERVER_STR;
     case HB_PTYPE_COPYLOGDB:
--- a/src/connection/heartbeat.h
+++ b/src/connection/heartbeat.h
@@ -59,12 +59,18 @@
 /* heartbeat resource process type */
 enum hb_proc_type
 {
-  HB_PTYPE_SERVER = 0,
-  HB_PTYPE_COPYLOGDB = 1,
-  HB_PTYPE_APPLYLOGDB = 2,
+  HB_PTYPE_PAGE_SERVER = 0,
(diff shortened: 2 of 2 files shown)


hornetmj (approved):

cristiarg (approved):

wrlawodms (commented):
Are the related codes with the process type changed later or just missed?
`hb_type_to_str()`, `hb_pack_server_name()`, `ha_argv_to_args()`, and more (I'm not sure).

joohok:
> Are the related codes with the process type changed later or just missed?
> `hb_type_to_str()`, `hb_pack_server_name()`, `ha_argv_to_args()`, and more (I'm not sure).

I understand what you mean. 

I plan to remove/change the codes related to existing process types at a later stage. 
Since they are still deeply intertwined with the untouched code, I have refrained from making any changes for now. 

It seems that the existing codes will be removed after the following steps are completed:

1. Adding process types. (This PR)
2. Executing the added process types. (Next PR)
   - `hb_pack_server_name ()` may be changed here. (this is called during starting up the process)
3. Implementing the jobs that the process types should perform. (Later)

wrlawodms (approved):
OK, it makes sense. 
It seems better to change here some codes not for behavior, but for definitions like: `hb_process_type_string()` and `hb_type_to_str()`. at your choice.

wrlawodms (approved):
OK, it makes sense. 
It seems better to change here some codes not for behavior, but for definitions like: `hb_process_type_string()` and `hb_type_to_str()`, though. at your choice.

joohok:
> OK, it makes sense.
> It seems better to change here some codes not for behavior, but for definitions like: `hb_process_type_string()` and `hb_type_to_str()`, though. at your choice.

OK, thanks.
`hb_type_to_str()` is only called when copylogdb and applylogdb is started. 

I want to focus on where the  `HB_PTYPE_SERVER` is used because it is the part relevant to  the issues. When it becomes possible to remove the parts related to `HB_PTYPE_APPLYLOGDB/COPYLOGDB`, I will proceed with removing the  `hb_type_to_str()`

PR #4400, opened 2023-06-01
[LETS-726] Start page server and transaction server within one node

joohok:
http://jira.cubrid.org/browse/LETS-726

Purpose

Both PS and TS should be started within one node when `cubrid hb start` command is executed. 

Implementation
- start PS and TS when 'cubrid hb start' command is executed
- us_hb_process_start () is called when `cubrid hb start` is executed
  - call stack : us_hb_process_start () -> process_server () -> process_servers_start_within_node ()
- process_servers_start_within_node ()
  - Use the existing process_server_start_single_node() and rename the function
  - if ha_mode is enabled, then start both PS and TS within one node
- is_server_running ()
  - requests information from commdb to determine if the server is running
  - commdb, in turn, requests cub_master to print out the status of all servers
    - define HA_PAGE_SERVER_FORMAT_STRING, HA_TRAN_SERVER_FORMAT_STRING in master_request.c
    - css_process_all_list_info (), prints out the servers currently running. 
  - commdb ~~checks~~ returns the output to the caller (check_server ())
  - check_server () checks if there is a server running from the information returned by commdb
- cub_server connects to cub_master
  - when server starts, cub_server tries to connect to cub_master after recovery in css_init ()
  - css_pack_message_to_master () 
    - message format : server type (Transaction or Page) + # (appended only if ha_mode is on) + db_name + ...


joohok:
ci tests are failed due to the enabled ha_mode in test_medium, sql_test

src/base/server_type.cpp
```diff
@@ -127,7 +127,14 @@ int init_server_type (const char *db_name)
       g_server_type = get_server_type_from_config (server_type_from_config);
     }
 
-  if (g_server_type == SERVER_TYPE_TRANSACTION && server_type_from_config == server_type_config::SINGLE_NODE)
+  if (g_server_type == SERVER_TYPE_TRANSACTION && !HA_DISABLED ())
+    {
+      /* TODO: Temporary implementation.
+       * This part will be addressed in LETS-721, and the system parameters will be automatically set.
+       */
+      setup_tran_server_params_on_single_node_config ();
```
joohok: I temporarily use this function to establish a connection between TS and PS.

src/executables/util_service.c
```diff
@@ -273,7 +273,7 @@ static int parse_arg (UTIL_SERVICE_OPTION_MAP_T * option, const char *arg);
 static int process_service (int command_type, bool process_window_service);
 static int process_server (int command_type, int argc, char **argv, bool show_usage, bool check_ha_mode,
 			   bool process_window_service);
-static int process_server_start_single_node (char *server_name, int command_type);
+static int process_servers_within_node (char *server_name, int command_type, bool ha_mode);
```
wrlawodms: as described
```suggestion
static int process_servers_start_within_node (char *server_name, int command_type, bool ha_mode);
```

src/executables/util_service.c
```diff
@@ -1770,10 +1770,13 @@ process_server (int command_type, int argc, char **argv, bool show_usage, bool c
 		{
 		  int pid;
 
-		  if (server_type_config::SINGLE_NODE ==
+		  if (util_get_ha_mode_for_sa_utils () != HA_MODE_OFF
```
wrlawodms, line 1774: Q. Why can't we use HA_DISABLED  or HA_GET_MODE ()?  I'm asking since `util_get_ha_mode_for_sa_utils` seems to mean it's for SA mode. I have to say that I don't really know the relevant code.

src/executables/util_service.c
```diff
@@ -1931,19 +1934,19 @@ process_server (int command_type, int argc, char **argv, bool show_usage, bool c
...
+process_servers_within_node (char *server_name, int command_type, bool ha_mode)
 {
   int status = NO_ERROR;
   int pid = 0;
 
   const char *args_page[] = { UTIL_CUBRID_NAME, "-t", "page", server_name, NULL };
   status = proc_execute (UTIL_CUBRID_NAME, args_page, false, false, false, &pid);
 
-  if (status == NO_ERROR && !is_server_running (CHECK_PAGE_SERVER, server_name, pid))
+  if (status == NO_ERROR && !is_server_running (ha_mode ? CHECK_HA_PAGE_SERVER : CHECK_PAGE_SERVER, server_name, pid))
```
wrlawodms, line 1957: Why do we have to use CHECK_HA_PAGE_SERVER, not reuse the CHECK_PAGE_SERVER? Why do we have to differentiate them?

src/connection/server_support.c
```diff
@@ -1821,19 +1826,20 @@ css_pack_message_to_master (const char *server_name)
     + strlen (pid_string) + 1;
   /* in order to prepend '#' or server type */
   message_size++;
+  if (!HA_DISABLED())
+    {
+      message_size++;
```
wrlawodms: How about adding a comment about what it's for?

src/connection/server_support.c
```diff
@@ -1821,19 +1826,20 @@ css_pack_message_to_master (const char *server_name)
     + strlen (pid_string) + 1;
   /* in order to prepend '#' or server type */
```
wrlawodms: should be revised. seems that it's now only for server type.

src/executables/master.c
```diff
@@ -444,26 +444,9 @@ receive_server_info (CSS_CONN_ENTRY * conn, unsigned short rid, std::string & db
...
-	{
-	  // HA mode on; only transaction servers connect.
-	  type = SERVER_TYPE_TRANSACTION;
-	  // Include '#' in the dbname; legacy requirement
-	  dbname = std::string (buffer, buffer_length);
-	}
-      else if (first_char == '$' || first_char == '%')
-	{
-	  // not really a server, it is copylogdb or applylogdb
-	  type = SERVER_TYPE_UNKNOWN;
```
wrlawodms, line 458: hmm.. I think you should keep it if you want to keep copylogdb and applyogdb code. Or you should delete others like in `css_process_all_list_info`. What do you think?

src/executables/commdb.c
```diff
@@ -548,7 +553,7 @@ process_server_info_pid (CSS_CONN_ENTRY * conn, const char *server, int server_t
 
 	      if (pid == 0)
 		{
-		  sprintf (search_pattern, "HA-Server %s (", server);
+		  sprintf (search_pattern, "HA-Tran-Server %s (", server);
```
wrlawodms: The comment above should be revised too.

src/executables/commdb.c
```diff
@@ -536,6 +536,11 @@ process_server_info_pid (CSS_CONN_ENTRY * conn, const char *server, int server_t
 	    case SERVER_TYPE_PAGE:
 	      sprintf (search_pattern, "Page-Server %s (", server);
 	      pid = search_server_pid (server_info, search_pattern);
+	      if (pid == 0)
```
wrlawodms: Q. What does it mean? and what's this for?

src/base/server_type.cpp
hornetmj: Nit:

for readability

```
if (g_server_type == SERVER_TYPE_TRANSACTION)
{
        if (!HA_DISABLED ())
        {
                ...
        }

        if (server_type_from_config == server_type_config::SINGLE_NODE)
        {
                ...
        }
}
```

hornetmj: And if ha_mode turns on, then the server_type cannot be set to single_node. It should be handle one of the following options:
1. explicit error
2. forcely update PTS

And it would be nice to add assert for this.

src/connection/server_support.c
hornetmj: If it is for the type, then it would be message_size += sizeof (type).

... I see the '(char) type' below. But this line is still not good readablilty.

src/connection/server_support.c
```diff
@@ -1821,19 +1826,20 @@ css_pack_message_to_master (const char *server_name)
...
   message_size++;
+  if (!HA_DISABLED())
+    {
+      message_size++;
+    }
 
   message.reserve (message_size);
 
-  if (!HA_DISABLED ())
+  message.append (1, '0' + ((char) type));
```
hornetmj, line 1832: ;ignore because I finally understand what you do.

Is there any reason to change the protocol format in the No-HA mode?

The previous code also send the server type in HA.

src/executables/commdb.c
hornetmj: Is there no way to identify HA-mode?

```
if (no-HA)
{
        call "Page-Server" style
}
else
{
        call "HA-Page-Server" style
}
```

src/executables/commdb.c
hornetmj: The comment above should be revised too.

src/executables/master.c
hornetmj: Nit:

How about enclose the codes related to the copylogdb and applylogdb to the your defined macro? For example

```
#if defined (DISABLE_COPYLOGDB_APPLYLOGDB)

#endif
```

Later you can easily delete this codes all together.

src/executables/master_request.c
```diff
@@ -408,7 +410,15 @@ css_process_all_list_info (CSS_CONN_ENTRY * conn, unsigned short request_id)
...
 	    {
 	    case '#':
-	      required_size += strlen (HA_SERVER_FORMAT_STRING);
+	      if (temp->server_type == SERVER_TYPE_TRANSACTION)
+		{
+		  required_size += strlen (HA_TRAN_SERVER_FORMAT_STRING);
+		}
+	      else
+		{
+		  /* SERVER_TYPE_PAGE */
```
hornetmj:
```suggestion
```

src/executables/util_service.c
hornetmj: Or change the description because consistency other process_* functions.

src/executables/utility.h
```diff
@@ -919,6 +919,8 @@ typedef struct _ha_config
 
 #define CHECK_SERVER              "Server"
 #define CHECK_HA_SERVER           "HA-Server"
+#define CHECK_HA_PAGE_SERVER      "HA-Page-Server"
+#define CHECK_HA_TRAN_SERVER      "HA-Tran-Server"
```
hornetmj: Nit:

make consistency between CHECK_HA_TRAN_SERVER and CHECK_TRANSACTION_SERVER.
1. CHECK_HA_TRAN_SERVER -> CHECK_HA_TRANSACTION_SERVER
2. or viceversa

hornetmj:
Don't spend too much time to solve the ci failures(to add a temp code for this). This project introduces new architecture. So, It is OK to block some part of regression in some period.

src/base/server_type.cpp
joohok: OK, I will handle it in #4383.

src/executables/commdb.c
joohok: server_info = "HA-page-server tdb (rel 12.0, pid 5523)" 
if pid is 0, then it fails to find pattern "Page-Server tdb (", and it means it can not find a pid. 
I'm now looking for more better way, because I just used the legacy code.

src/connection/server_support.c
joohok: OK, I will revise the comment. 
1827L was the comment for this. (`/* in order to prepend '#' or server type */`) 

protocol is,
server type (1 byte) + # (1 byte, for HA server) + dbname+ ..

src/executables/commdb.c
joohok: OK, Thanks.

src/executables/master.c
joohok: Hmm..I agree. I'll leave this code.
@hornetmj I think it is good idea, I will consider it

src/executables/util_service.c
joohok: All the part that check `ha_mode` in util_servce.c uses `util_get_ha_mode_for_sa_utils ()`.
I used everything to get ha_mode, such as `prm_get_integer_values (PRM_ID_HA_MODE)`, HA_DISABLED, HA_GET_MODE,.. but they always returns false (ha_mode=off). 

I am now analyzing the cause, and I will inform you once I have the results.

src/executables/util_service.c
joohok: `is_server_running()` function checks if there is a pattern matching `CHECK_PAGE_SERVER ("Page-Server")` in the process status information provided by cub_master, in order to determine whether the server is running or not.
And the status information is used to print out to the user.

Reusing the `CHECK_PAGE_SERVER` pattern means that cub_master should output the status information without distinguishing between HA and non-HA servers. Similar to how `oldHA` previously differentiated between the `server tdb (rel, pid)` and the `HA-server tdb (rel, pid)` in the output, the I distinguish between `CHECK_PAGE_SERVER` and `CHECK_HA_PAGE_SERVER` to provide the same functionality.

src/executables/utility.h
joohok: OK..!

src/executables/util_service.c
joohok: Done

joohok:
> Don't spend too much time to solve the ci failures(to add a temp code for this). This project introduces new architecture. So, It is OK to block some part of regression in some period.

I've confirmed that the CI tests are functioning properly with a simple modification. This issue will be resolved in http://jira.cubrid.org/browse/LETS-741.

src/executables/util_service.c
joohok: Sorry! Those are just information exchanged between `cub_master` and `comdb`. They are not shown to the user.

it seems better to follow your suggestion and not distinguish between HA and non-HA servers.

Analysis : 
I apologize for any confusion. From my understanding, even before the modifications, cub_master did differentiate between HA and non-HA servers and communicated this information to commdb. commdb then used this information to determine whether HA servers or non-HA servers were running.

If there is a scenario where both HA and non-HA servers can be running simultaneously, it would indeed be appropriate to make such distinctions. However, based on the current architecture, it seems that HA and non-HA servers cannot run simultaneously.

It is not possible to configure ha_mode on a per-database basis (`[@testdb]`), and it would also be strange to start an HA server and then turn off ha_mode to start another server. This is because it would require changing the statically configured configuration.

src/executables/util_service.c
joohok: Analysis : 
cubrid utility works as SA_MODE. (util_service.c)
So, if `ha_mode` is set to `ON`, then it sets `PRM_ID_HA_MODE_FOR_SA_UTILS_ONLY` to `ON` and use this parameter. (`util_get_ha_mode_for_sa_utils ()` can get the value for this parameter)

src/executables/commdb.c
joohok: Removed.

src/executables/util_service.c
wrlawodms: Thank you for the information!

wrlawodms (commented):
The token, '#', representing HA-Node, is really needed? Every code seems to ignore the token, at least in the modified code.

src/executables/util_service.c
hornetmj: Nit:
For the code consistency and save your time to analysis the code, I think you just follow the same approach like oldHA. The reason is I think you can find same problem in the next issues. It consumes un-expected time and make slower the progression. I don't mean this process is not worthfull(rather, it is very important. But we need to focus on more important things.). I think it would be better comment on about what to do and handle it later.

src/base/server_type.cpp
```diff
@@ -127,7 +127,14 @@ int init_server_type (const char *db_name)
...
+    }
+  else if (g_server_type == SERVER_TYPE_TRANSACTION && server_type_from_config == server_type_config::SINGLE_NODE)
```
hornetmj: Is it needed to check g_server_type == SERVER_TYPE_TRANSACTION twice?

src/connection/server_support.c
```diff
@@ -1797,6 +1799,9 @@ css_pack_message_to_master (const char *server_name)
   char pid_string[16];
   std::string message;
 
+  SERVER_TYPE type = get_server_type ();
+  assert (type == SERVER_TYPE_TRANSACTION || type == SERVER_TYPE_PAGE);
```
hornetmj, line 1804: How about move this assert in the set_server_type()?

src/connection/server_support.c
```diff
@@ -1819,21 +1824,25 @@ css_pack_message_to_master (const char *server_name)
   size_t message_size =
     strlen (server_name) + 1 + strlen (rel_major_release_string ()) + 1 + strlen (env_name) + 1
     + strlen (pid_string) + 1;
-  /* in order to prepend '#' or server type */
+
+  /* in order to prepend a server type (1 byte char)*/
```
hornetmj: How about comment the payload format on together? like

| server_type (1-byte) | #db_name (1-byte + strlen(db_name) | ...

And remove "In order to prepend ..." comments.

src/executables/util_service.c
joohok: @hornetmj  
Thanks. I understand what you mean. 

~~I think it would be better to proceed with these modifications just for now, since this part is limited to communication between `cub_master` and `commdb`. When we start handling the information between cub_master and cub_server during communication (probably from the next PR), I would like to follow your opinion and focus on the key aspects as much as possible. What do you think?~~

joohok:
> The token, '#', representing HA-Node, is really needed? Every code seems to ignore the token, at least in the modified code.

The '#' token is still being used in cub_master. It is used to differentiate HA servers among the servers registered in cub_master, particularly when killing servers or terminating slave servers. (IS_MASTER_CONN_NAME_HA_SERVER())

src/executables/util_service.c
joohok: TO SUMMARY! 

It seems appropriate to proceed with @hornetmj  suggestion. Initially, I removed the distinction by removing `CHECK_HA_PAGE_SERVER` since I thought that non-HA servers and HA servers cannot be running simultaneously. However, I discovered *exceptional situations where both servers can be running simultaneously. It is possible that these situations were already considered in the oldHA, or it might be a new scenario. Instead of analyzing these exceptions, it might be better to focus on implementing the required functionality.

*Exceptional situations:

1. Session 1 (Terminal 1) and Session 2 (Terminal 2) have different `$CUBRID` 
    - by different commit number or something
2. In each session, both HA server and non-HA server are executed.
3. When checking the status of the Cubrid server, it is observed that both servers are running.
   - Same cub_master manages both servers

@hornetmj  @wrlawodms  what do you think?

src/base/server_type.cpp
joohok: This is refactored in #4383.

Above implementation is just temporary implementation to avoid runtime error.

src/connection/server_support.c
joohok: then it needs refactoring I think. 

The g_server_type is not only changed using the set_server_type() function, but there are also direct modifications to it. That is why the assert statement is placed here.

src/connection/server_support.c
joohok: OK, it will be more clear

src/executables/util_service.c
hornetmj: @joohok I sorry to make you confution. The guideline is not for collecting other opinions. I just want for you to save the time and focus more important and difficult tasks like supporting fail-over and split-brain mechanism perfectly. Additionally, the sooner the newha function is supported, the sooner QA testing and user-level test verification can be introduced, and the more effective it can be to derive and find overall problems. Therefore, I ask that you minimize unnecessary analysis and proceed with a focus on key features. However, please leave a clear comment on the part that needs to be reviewed later.

joohok: OK, then I will 
1) rollback these changes, 
2) leave a comment about what to do, (remove distinction between non-HA server and HA-server)
3) and make an issue and link to this issue to track.

hornetmj: Nit:

About 3. it depends on you. I think it is enough to leave comments with a keywords like 'JOHOK TODO'. Because you can open a issue grouping several comments later.

joohok: OK, I agree.

src/connection/server_support.c
```diff
@@ -1816,24 +1821,27 @@ css_pack_message_to_master (const char *server_name)
     */
 
   sprintf (pid_string, "%d", getpid ());
+
+  /* | server_type (1 byte) | '#' if HA-mode enabled (1 byte) | server_name | release string | env_name | pid | */
```
wrlawodms, line 1826: The comment should match the order of definition of `message_size`.
```suggestion
  /* | server_type (1 byte) | server_name | '#' if HA-mode enabled (1 byte) | release string | env_name | pid | */
```

wrlawodms: Or the order in the definition of `message_size` according to the order of those are appended.

src/executables/util_service.c
wrlawodms: I understand that old-HA distinguishes the non-HA server and HA-server, but what is it for? Is it valid for NewHA?

src/executables/master_request.c
```diff
@@ -66,6 +66,12 @@
 
 #define SERVER_FORMAT_STRING " %s-Server %s (rel %s, pid %d)\n"
 #define HA_SERVER_FORMAT_STRING " HA-Server %s (rel %s, pid %d)\n"
+/* TODO by Joohok
+ * if there is no need to distinguish between HA-server and nonHA-Server,
```
wrlawodms: When is the time? What is the requirements? It can be handled by other and should be understandable.

src/connection/server_support.c
joohok: I will make sure to align 
 the actual order in which the components are appended, 
 the order in which the component sizes are added in the `message_size`, 
and the order described in the comments.

src/executables/master_request.c
joohok: OK

src/executables/util_service.c
joohok: @wrlawodms 

> I understand that old-HA distinguishes the non-HA server and HA-server, but what is it for?

The current implementation distinguishes between nonHA servers and HA servers for only one reason.

It is only used for users to determine whether the currently running server is an HA server or a nonHA server based on the output of the `cub_commdb -O` command. There doesn't seem to be any other purpose or utility for it.

I think it would be valid for newHA as well, 
In terms of 
1) providing the same user experience and 
2) being able to catch bugs like the one below analysis 

Analysis :
HA servers and nonHA servers with the same name cannot be started simultaneously.
An example of a problem scenario when not distinguishing between HA servers and non-HA servers is as follows.
(used exceptional case mentioned in https://github.com/CUBRID/cubrid/pull/4400/files#discussion_r1224092830 )

**Situation :** 
Terminal 1: Create a nonHA server named testdb.
Terminal 2: Create an HA server named testdb.
1. In terminal 1, Start testdb as nonHA.
2. In ternminal 2, Execute "cubrid hb start" and the node's state becomes Master.
   - Without distinguishing between HA and nonHA, it is assumed to provide HA services because HB utility assumes that the HA server is already running..
3. Connect to the server in Terminal 2 and perform transactions. 
   - Since they use the same cub_master, it is possible to connect to the server started in Terminal 1.

**Issue:**
The user assumes that they connected to the HA server and performed the transaction, believing that replication has taken place. However, in reality, they connected to the non-HA server and performed the transaction.

**conclusion:**
By distinguishing between HA servers and nonHA servers, such bugs can be addressed. It can also be applied to newHA.
For example, when executing "cubrid hb start," it can check if a server with the same name already exists and, if so, kill the existing server and start the HA server instead.

src/executables/master_request.c
joohok: I'll revise this comment when the https://github.com/CUBRID/cubrid/pull/4400/files#r1227358257 
discussion is concluded

src/executables/util_service.c
hornetmj: @joohok Disccussed off-line, I think customers have no experience the scenario you mentioned(but, need to check). So, there is other chance like block starting non-ha server in ha_mode=on. Let's consider it later.

src/connection/server_support.c
hornetmj: only one place init_server_type().

joohok: ~~OK, then I will move this assertion to set_server_type ()
In init_server_type (), it also add assertion (at get_server_type_from_config ()) when setting the g_server_type.~~

I believe there should be an assertion here. This is because someone may accidentally call this function (`css_pack_message_to_master ()`) before `g_server_type` being set to PAGE_SERVER or TRANSACTION_SERVER (default is TYPE UNKNOWN).

src/connection/server_support.c
wrlawodms: GOOD

joohok: Done, by adding `ha_token_size`

src/executables/util_service.c
wrlawodms: @joohok Thank you for the elaboration. 
> providing the same user experience and

Agreed.

> being able to catch bugs like the one below analysis

I don't think we need to address that kind of potential wrong behavior which has been since the develop branch.

wrlawodms (approved):

src/transaction/boot_sr.c
```diff
@@ -2263,13 +2263,6 @@ boot_restart_server (THREAD_ENTRY * thread_p, bool print_restart, const char *db
       goto error;
     }
 
-  if (get_server_type () == SERVER_TYPE_PAGE && !HA_DISABLED ())
-    {
-      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_INCOMPATIBLE_SERVER_TYPE_HA_CONFIG, 0);
```
cristiarg, line 2268: Nit.
I think that `ER_INCOMPATIBLE_SERVER_TYPE_HA_CONFIG` is not used anywhere else.

src/executables/utility.h
```diff
@@ -919,6 +919,12 @@ typedef struct _ha_config
 
 #define CHECK_SERVER              "Server"
 #define CHECK_HA_SERVER           "HA-Server"
+/* TODO by Joohok
+ * if there is no need to distinguish between HA-server and nonHA-Server,
+ * remove CHECK_HA_PAGE_SERVER and CHECK_HA_TRANSACTION_SERVER
+ */
+#define CHECK_HA_PAGE_SERVER      "HA-Page-server"
+#define CHECK_HA_TRANSACTION_SERVER      "HA-Transaction-server"
 #define CHECK_PAGE_SERVER         "Page-Server"
```
cristiarg, line 924: We have the following working modes intended for Cubrid server - these are possible in CS (client-server) mode only:
1. the monolithic server, non-HA
    - this should be kept
2. the monolithic server, (old)HA
    - this is to be replaced by (new)HA based on 3; and this one only works in "single-node" configuration - one xTS and one PS under the same `cub_master` and the same OS-user with the same database (out of which only PS actually touches the database files, and the xTS can only create some temporary volumes and a meta-log - to be seen if this is in place yet, I think not)
3. scalability mode, non-HA - this will be the [infinitely] horizontally scalable architecture (of course, once we manage to validate it)
    - newly introduced and what we are currently using and running to prove the architecture

And these modes are in no way composable.
I am not aware of others.

So, in fact the NewHA is actually a combination of "Transaction-Server" and "Page-Server" + plus the "single node" boolean flag.
I think there should be some functions which just provide identification strings such as "HA-Page-server" based on the existing `CHECK_PAGE_SERVER` and `CHECK_TRANSACTION_SERVER` macros, plus a new
```
#define CHECK_NEW_HA_MNEMONIC "(HA)"
```
mnemonic which is to be defined. The function will take the `single_node` flag to either assemble the identification strings for non-HA scalability (when false) or for new-HA (when true).
And `CHECK_HA_SERVER` must be deleted.
And also `CHECK_HA_PAGE_SERVER` and `CHECK_HA_TRANSACTION_SERVER` are not needed.

src/executables/util_service.c
cristiarg: My 2cents.
The way I understand the various configurations in which CUBRID can be run, they are almost endless.

For example, for Scale-Out tests we've been running with multiple servers on the same machine (ATS, PSs, PTSs) but each server runs under a different user with its own `cub_master` and they are separating by separating the `cub_master` ports and application server ports (for the xTSs).

Also the general tendency is to migrate to running in virtual machines/containers where each server will be isolated via OS-level "borders" (ie: kernel namespaces) from other servers running on the same physical machine. So we should not stumble upon these details.

The scenarios described above - where the same `testdb` server - is run in both non-HA and HA should not be possible.
This is because in `databases.txt` the unique server identification is the database name.
```
Terminal 1: Create a nonHA server named testdb.
Terminal 2: Create an HA server named testdb.

In terminal 1, Start testdb as nonHA.
In ternminal 2, Execute "cubrid hb start" and the node's state becomes Master.
Without distinguishing between HA and nonHA, it is assumed to provide HA services because HB utility assumes that the HA server is already running..
Connect to the server in Terminal 2 and perform transactions.
Since they use the same cub_master, it is possible to connect to the server started in Terminal 1.
```
This scenario should not be possible. Step 2 - `cubrid hb start` - should already result in an error.

Single node was added specifically to "break" this rule and allow 2 servers (a xTS and a PS) to execute under the same `cub_master` and the same OS user with the same database name.

src/executables/util_service.c
cristiarg: I will refer to http://jira.cubrid.org/browse/LETS-139.
Single node was introduced with no other apparent purpose than to allow a simple execution of both ATS (at that time there was no PTS) and PS under the same roof (same database and same OS-user).

However, experience showed that, with the addition of PTS this "single node" mode did not receive too much use. Actually we have not used it in practice. We have some tests though.

Why don't we adopt (with a rename later maybe) this "single node" mode of execution specifically for NewHA?

Would that not simplify things? (see also comment about different working modes from the current review).
I mean:
- monollithic: no-scalability, no single node ==>> classical server
- scalability per se: scalability, no single node ==>> well, scalability
- NewHA, scalability, single node ==> HA

The only drawback I see is that HA will only be able to run with a xTS+PS on the same physical machine. But that is the case even now with the HA+monolithic server.
If the user wants more "scaling", there is Scalability mode where each server can be run under its own physical machine.

cristiarg (approved):
I do not want to add more fuel to the fire.
I approved but I also added some questions which I think are relevant. Thank you.

src/connection/server_support.c
hornetmj: @joohok I agreed.

hornetmj (approved):

src/executables/util_service.c
joohok: Thank you for providing detailed explanations. 
> This scenario should not be possible. Step 2 - cubrid hb start - should already result in an error.

I completely agree. The scenario should not be possible. It requires modifications in the develop branch to block above scenario.

src/transaction/boot_sr.c
joohok: Thanks, then it can be removed

src/executables/utility.h
joohok: It seems that the goal is to assemble an identification string for the NewHA TS and PS based on the currently used mode and the single_node flag.
Transaction-Server + single_node flag = `CHECK_HA_TRANSACTION_SERVER` 
Page-Server + single_node flag = `CHECK_HA_PAGE_SERVER`

This identification string is used in the utility process (`cubrid`), commdb process, and cub_master. It's uncertain whether it's better to assemble this string every time in each process or not.

Based on the feedback provided in https://github.com/CUBRID/cubrid/pull/4400#discussion_r1227850948 and this comment, it seems necessary to consider the usage of the single-node flag again. This should be carefully considered and I'll make an issue to be linked to sub-task.

Thanks for your elaboration

PR #4402, opened 2023-06-02
[LETS-731] reuse extensible block to avoid re-allocating memory at each call

cristiarg:
http://jira.cubrid.org/browse/LETS-731

Use `cubmem::extensible_block` for:
- the send buffer in `request_client`
- the receive buffer in `request_server` and `request_client_server`
- the send buffer in `request_client_server`

Will avoid alloc/de-alloc buffers for each message.

### DIFF
--- a/src/communication/request_client_server.hpp
+++ b/src/communication/request_client_server.hpp
@@ -244,6 +244,7 @@ namespace cubcomm
   template <typename MsgId>
   request_client<MsgId>::request_client (channel &&chn)
     : m_channel (std::move (chn))
+    , m_send_extensible_block { cubmem::CSTYLE_BLOCK_ALLOCATOR }
   {
   }
 
@@ -405,6 +406,7 @@ namespace cubcomm
   template <typename ClientMsgId, typename ServerMsgId>
   request_client_server<ClientMsgId, ServerMsgId>::request_client_server (channel &&chn)
     : request_server<ServerMsgId>::request_server (std::move (chn))
+    , m_send_extensible_block { cubmem::CSTYLE_BLOCK_ALLOCATOR }
   {
   }
 
--- a/src/communication/request_client_server.hpp
+++ b/src/communication/request_client_server.hpp
@@ -151,7 +151,7 @@ namespace cubcomm
       inline const channel &get_channel () const;		// get underlying channel
 
     protected:
-      cubmem::extensible_block m_reused_extensible_block;
+      cubmem::extensible_block m_send_extensible_block;
     private:
       channel m_channel;					// requests are sent on this channel
   };
@@ -194,7 +194,6 @@ namespace cubcomm
 
     protected:
       channel m_channel;	  // request are received on this channel
-      cubmem::extensible_block m_reused_extensible_block;
 
     private:
       std::thread m_thread;				// thread that loops and handles requests
@@ -220,11 +219,14 @@ namespace cubcomm
 
       template <typename ... PackableArgs>
       css_error_code send (ClientMsgId msgid, const PackableArgs &... args);
+
+    private:
+      cubmem::extensible_block m_send_extensible_block;
   };
 
   // Helper function that packs MsgId & PackableArgs and sends them on chn
   template <typename MsgId, typename ... PackableArgs>
-  css_error_code send_client_request (channel &chn, cubmem::extensible_block &m_ext_block,
+  css_error_code send_client_request (channel &chn, cubmem::extensible_block &send_ext_block,
 				      MsgId msgid, const PackableArgs &... args);
 
   // Err logging functions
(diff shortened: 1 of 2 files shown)


src/communication/request_client_server.hpp
```diff
@@ -349,10 +358,11 @@ namespace cubcomm
       }
     assert (size_ilen == sizeof (expected_size));
 
-    message_buffer.reset (new char [expected_size]);
+    // will re-alloc until the block size gets big enough that will not neet to re-alloc anymore
```
wrlawodms, line 361:
```suggestion
    // will re-alloc until the block size gets big enough that will not need to re-alloc anymore
```

wrlawodms (commented):
1. Is the overhead by `__memcpy_ssse3_back` disappeared with this patch? The copy seems still to be there. Does it work just to remove initialization?
2. the extensible_blocks used during send/recv can be extended, but are not shrunk. Is it OK?

hornetmj (commented):
LGTM.

Could you please share the performance improvement with this fix?

src/communication/request_client_server.hpp
cristiarg: > Is the overhead by __memcpy_ssse3_back disappeared with this patch? The copy seems still to be there. Does it work just to remove initialization?

No. We'll be able to re-evaluate whether the memcpy overhead is reduced with the next patch - see #4405 .

> the extensible_blocks used during send/recv can be extended, but are not shrunk. Is it OK?

Never shrunk. Will be de-allocated upon destruction of the object. This basically means when the server (ATS/PS/PTS) goes down.
Can this be a problem?

While investigating this, I have also come across the following - which I described in http://jira.cubrid.org/browse/LETS-743 : ATS can end up sending huge messages over the network, this has the potential to cause congestion in the uploading thread.

Technically, via a carefully implemented protocol - ie not what is described in http://jira.cubrid.org/browse/LETS-743 - we can limit the allocated size to a reasonable value:
- upload from ATS to PS: whatwever we decide is the biggest size that a packed log prior list can be; there can be problems here in the case of large UNDO-REDO log records, but I assume there is an upper limit
- download from PS to TS's: technically the page server can never serve something bigger data a storage page (16K); PS will also use the same stream to dispatch log prior lists to PTS's, but that comes from the first limitation described above

cristiarg:
@hornetmj 
> Could you please share the performance improvement with this fix?

I expect this change to have a unnoticeable impact. The following patch - see #4405 - should be measured.

src/communication/request_client_server.hpp
wrlawodms: > Can this be a problem? 

No, it's not for now, I think. What I wanted to talk about is whether it's worth it. 
1. If the performance gets improved enough, and it's necessary, we should do this and create an issue for the potential problem. Or we can add some decsiprtion on http://jira.cubrid.org/browse/LETS-743 about the huge msg will keep the huge memory.
2. If the advantage is not that good, it'd be better to leave it as it is.

wrlawodms (approved):

cristiarg:
> If the performance gets improved enough, and it's necessary, we should do this and create an issue for the potential problem. Or we can add some description on http://jira.cubrid.org/browse/LETS-743 about the huge msg will keep the huge memory.

LETS-743 will be done as well (I think, rather soon). So, after that, the reused extensible block will not be grown to a very large size - as of now it can grow in the megabytes and stay there - and this is not acceptable in the long run.
However, since there is a recurring pattern of needing a buffer for repeated same operations, it seems logical to me to only allocate once and re-use rather than allocated many times and de-allocate each time.

> If the advantage is not that good, it'd be better to leave it as it is.

You'll also notice that with this patch, the extensible block's allocator is changed:
- from the default `STANDARD_BLOCK_ALLOCATOR` - which does - take a look in `standard_alloc` function - `new` and `memcpy` on extensions
- to `CSTYLE_BLOCK_ALLOCATOR` - which does - see `cstyle_alloc` functions - `malloc` and ` realloc` - which "might" elide a `memcpy` when extended (as per `realloc` specification)

Other considerations:

1. From past experience we also hit this overhead of `new`/`delete` when we perf-ed the parallel replication and found out that new-ing/delete-ing tasks had a significant performance impact (ie: it appeared high in the perf call graph). Hence we implemented and integrated the `cublog::reusable_jobs_stack` to hold a bunch of tasks available for replication threads.

2. With code that is intermingled with I/O it is very hard (at least, I find in very hard) to easily pinpoint where performance optimizations need to occur because performances is overshadowed by I/O - in this case network - which takes orders of magnitue more time than the actual "optimized" bit.
But every little bit adds up in the end.

Notifying @hornetmj as well :)

hornetmj (approved):

PR #4405, opened 2023-06-06
[LETS-757] forward payload to avoid copying

cristiarg:
http://jira.cubrid.org/browse/LETS-757

There is a mix of
```
    template <typename ... PackableArgs>
    css_error_code send (MsgId msgid, const PackableArgs &... args);	
```
and
```
    template <typename ExtBlk, typename ... Args>
    void set_buffer_and_pack_all (ExtBlk &eb, Args &&... args);
```
variadic arguments in the functions declarations of some classes: `request_sync_client_server`, `reqquest_client_server`, `cubpacking::packer`.

Important note about performance.
This change does not eliminate any actual copies of the data (because data is actually being passed by reference in the callstack of the same thread). But it is worthwhile to make this change not only from a pure cosmetic reason but to allow further reafactorings (such as offloading the packing/unpacking of data asynchronously).

Implementation:
- change `const PackableArgs &... args` with `PackableArgs &&... args`
- and properly forward `PackableArgs &&... args` arguments with `std::forward` in `request_client`, `request_client_server`

Other:
- some clarifying changes in unit tests where the a payload was being pulled twice:
  - once in `test_two_request_sync_client_server_env::handle_req_and_respond`
  - and again in the called `mock_check_expected_id_sync`


hornetmj:
@hornetmj For me.

Check the performacne improvement of the following PR:
https://github.com/CUBRID/cubrid/pull/4402

wrlawodms:
@cristiarg 
To clarify, could you tell me where exactly the copy happens?

The journey of the payload seems like this to me: 
```
connection_handler::push_request 
(Universal ref, Move) ->  request_sync_client_server::push
(Universal ref, Move) -> request_sync_send_queue::push
(Universal ref, Move) -> m_request_queue (std::queue)

-> in request_sync_send_queue::sned_queue
(Lvalue ref) -> request_client_server::send
(Lvalue ref) -> request_client_server::send_client_request
(Universal ref, lvalue) -> cubpacking::packer::set_buffer_and_pack_all  <— copy the contents regardless of it’s rvalue or lvalue
```
I can't spot the redundant copy. I'd love it if you tell me what I'm missing..

cristiarg:
@wrlawodms

Hm. Good question.
Now that you asked, I see indeed that there is no copy eliminated as the argument "travels" in the form of a reference down the callstack of the same thread - the send_queue thread.

It is only in between the client (transaction) thread and the sending queue thread that these moves do have to happen. And they did so even before this change - by `std::move`s in `request_sync_client_server::send_recv` and `request_sync_client_server::push`.

So this change brings nothing to the table other than cosmetics.

Maybe this explains why, in my profiling, I did not see any reduction in `memcpy` percentages. Which I was expecting. :(

Up to you if you still consider this PR worthwhile.

cristiarg:
Now this raises a possible optimization question.
Currently, the packing of the message to be sent over the network happens at the expense of the send queue thread (in `request_sync_send_queue::send_queue`).
It can be an investigation point if it is better to do the packing (if possible) at the expense of the client (transaction) thread.
I think it's a simple enough change.

I added a bullet about this in http://jira.cubrid.org/browse/LETS-636 .

wrlawodms (approved):
> Up to you if you still consider this PR worthwhile.

I agree with the PR in terms of code consistency (to use only universal references).

hornetmj:
> Check the performacne improvement of the following PR:
> #4402

The changes of this PR doesn't look like related to the performace improvement of #4402. Please share the way to check the improvement of it.

The followings might be answer. (no gain)
- https://github.com/CUBRID/cubrid/pull/4405#issuecomment-1600476968

PR #4407, opened 2023-06-07
[LETS-731] error manager debug function

cristiarg:
http://jira.cubrid.org/browse/LETS-731

I have written this function many times while debugging various issues.
I am adding this here in `!NDEBUG` mode only.
Documented with usage example in code.

Note: on top of #4405


cristiarg:
Superseeded by #4409

PR #4408, opened 2023-06-07
[LETS-742] streamline scalability communication protocol

cristiarg:
http://jira.cubrid.org/browse/LETS-742

The connection support layer (`css_*` functions) already send the length and the data (as an application level protocol).
In the network code introduced by LETS (`communication_channel`, `request_client_server`), the sending of length was done ome more time.
This translated in 4 calls to underlying `send`/`writev` functions:
1. send/recv "length of length"
2. send/recv actual length as data
3. send/recv "length of data"
4. send/recv actual data

This was done in the beginning of development for speed/simplicity of development rather than performance of the data transfer.
It is easy to see that 1. and 2. are superfluous.
The benefit of having 1. and 2. is that they are both messages of fixed/known length and they allowed the receiving side to know exactly how much data it would receive in step 4.
Removing them has the downside that the receiving side's application layer never knows in advance how much data it receives in step 4.
But also, removing them, has the benefit of uncluttering the dialog between peer servers.
To remove threm, the connection support layer has been extended to allow a two step receive of unknown side data (see functions `css_net_recv_allow_truncated` and `css_net_recv_remainder` - patch 8253662fb) such as to allow proper consumption of data in step 4.
In practice, in the receiving side's application layer - with the reuse of a memory buffer (in this case `cubmem::extensible_buffer`) the calls to the second function will be very rare (if ever).

Implementation:
- `channel::recv_allow_truncated`: calls into `css_net_recv_allow_truncated`
  - if all data fits into supplied buffer it will return a `NO_ERRORS` code and output a 0 `remaining_len` (same as `channel::recv` does)
  - if not all data fits into supplied buffer, it will return a `RECORD_TRUNCATED` code and a non-zero value for `remaining_len`
- if the application layer receives `RECORD_TRUNCATED` it is supposed to follow-up with a call to `channel::recv_remainder` and a properly sized buffer as to consume all remaining data as dictated by the previous call to `recv_allow_truncated`
- `request_server::receive_request_buffer` is the function that actually implements this simplification - the `recv_allow_truncated` is followed up by a call to `recv_remainder` when/if needed
  - the signature of `receive_request_buffer` is unchanged so nothing has changed for the upper layers
- adapted the relevant unit test to also exploit/test the newly added mechanism

Other:
- the signature difference for function `send_client_request` seems to be a merge error from a previous patch (31803efd1)


src/communication/request_client_server.hpp
```diff
@@ -344,33 +346,48 @@ namespace cubcomm
...
+	// there is more data to retrieve
+	// extend the block and supply the new pointer
 
-    // will re-alloc until the block size gets big enough that will not neet to re-alloc anymore
-    m_recv_extensible_block.extend_to (expected_size);
+	assert (recv_size == m_recv_extensible_block.get_size ());
+	assert (rem_size > 0);
+
+	// TODO: maybe it should be extended to a multiple of page size?
+	m_recv_extensible_block.extend_by (rem_size);
```
cristiarg, line 368: I suppose it would be better to extend to a multiple of the page size?

src/communication/communication_channel.cpp
```diff
@@ -95,16 +95,57 @@ namespace cubcomm
...
+
+    assert (m_type != NO_TYPE);
+    assert (actual_remaining_len == 0);
+
+    const css_error_code rc = css_net_recv_allow_truncated (m_socket, buffer, &actual_recv_maxlen,
+			      &actual_remaining_len, m_max_timeout_in_ms);
+    er_log_chn_debug ("[%s] Receive%s buffer of size = %d, max_size = %zu, remain = %d, result = %d",
+		      get_channel_id ().c_str (), ((actual_remaining_len > 0) ? " truncated" : " full"),
+		      actual_recv_maxlen, maxlen_in_recvlen_out,
+		      actual_remaining_len, (int)rc);
```
wrlawodms, line 124:
```suggestion
    er_log_chn_debug ("[%s] Receive %s buffer of size = %d, max_size = %zu, remain = %d, result = %d",
		      get_channel_id ().c_str (), ((actual_remaining_len > 0) ? " truncated" : " full"),
		      actual_recv_maxlen, maxlen_in_recvlen_out,
		      actual_remaining_len, (int) rc);
```

src/communication/request_client_server.hpp
wrlawodms: I don't really know, but it seems that the communication buffer is irrelevant with the page size.

src/communication/request_client_server.hpp
```diff
@@ -433,16 +450,7 @@ namespace cubcomm
...
-    const css_error_code css_size_err = chn.send (
-	reinterpret_cast<const char *> (&packer_current_size), sizeof (packer_current_size));
-    if (css_size_err != NO_ERRORS)
-      {
-	er_log_send_fail (chn, css_size_err);
-	return css_size_err;
-      }
-
-    const css_error_code css_err = chn.send (send_ext_block.get_ptr (), packer.get_current_size ());
+    const css_error_code css_err = chn.send (send_ext_block.get_ptr (), packer_current_size);
```
wrlawodms, line 457: Nit. 
I don't think `packer_current_size` is needed anymore since it's for sending the size, which requires its address.
```suggestion
    const css_error_code css_err = chn.send (send_ext_block.get_ptr (), packer.get_current_size ());
```

wrlawodms (approved):
LGTM.

src/communication/request_client_server.hpp
```diff
@@ -267,6 +267,8 @@ namespace cubcomm
     : m_channel (std::move (chn))
     , m_recv_extensible_block { cubmem::CSTYLE_BLOCK_ALLOCATOR }
   {
+    // arbitrary initial size; will be grown upon need
+    m_recv_extensible_block.extend_to (IO_MAX_PAGE_SIZE * 4);
```
joohok: Nit.
I'm not sure how `IO_MAX_PAGE_SIZE` is related to network communication. 
If it is intended to consider the exchange of pages between servers, it might be a good idea to define a separate constant for that purpose. 

And that constant could be utilized when resizing the block. (`m_recv_extensible_block.extend_by(constant value)`)

joohok (approved):

src/communication/request_client_server.hpp
cristiarg: As commented, it is an arbitrary value.
Could even be set to 1 and the logic will adapt to the actually required size.

src/communication/request_client_server.hpp
cristiarg: I know, but it is more explicit like this. Helps for debug/cores.

wrlawodms: I've got it!

PR #4414, opened 2023-06-08
[CBRD-24834] Utility also works with invalid cubrid_hosts.conf file configuration

eido5:
http://jira.cubrid.org/browse/CBRD-24834

**Description**
* in addition to #3740,
* fix to do same style action of cub_master with glibc host lookup when cubrid_hosts.conf looks like:
```
127.0.0.1 localhost
192.168.240
```

**Remarks**


eido5:
@kisoo-han 
`(hostent_Cache[cache_idx] = hostent_alloc (ipaddr, hostname)) == NULL`
This code and inside **hostent_alloc ()** occurs NULL return value when the malloc() is failed.
So, this is not necessary to be fixed.

Is it right what your comment is?

src/connection/host_lookup.c
```diff
@@ -346,65 +345,35 @@ load_hosts_file ()
...
-	}
     }
-  fclose (fp);
-
-  return LOAD_SUCCESS;
 
-load_fail_phase:
+load_end_phase:
+  if (fp != NULL)
+    fclose (fp);
```
kisoo-han:
```suggestion
  if (fp != NULL)
    {
      fclose (fp);
    }      
```

eido5: I will fix it. Thank you.

airnet73 (approved):

src/connection/host_lookup.c
```diff
@@ -303,37 +304,35 @@ load_hosts_file ()
...
+		  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_UHOST_INVALID_FORMAT, 4, "Hostname", token, line_num,
 			  USER_HOSTS_FILE);
 
-		  user_host_Map.clear ();
-		  fclose (fp);
-		  goto load_fail_phase;
+		  continue;
 		}
-	      else if (ip_format_check (temp_token) == true)
+	      else if (hostname_format_check (temp_token, str_len) == false)
```
kisoo-han: what about change function name as (is_valid_hostname): _trivial_
else if (**is_valid_hostname** (temp_token, str_len) == false)

src/connection/host_lookup.c
```diff
@@ -462,6 +433,46 @@ ip_format_check (char *ip_addr)
...
+ *
+ * return   : true if host name is valid format, false otherwise
+ */
+static bool
+hostname_format_check (char *hostname, int str_len)
+{
+
+  int char_num = 0;
+
+  if (!((hostname[0] >= 'A' && hostname[0] <= 'Z') ^ (hostname[0] >= 'a' && hostname[0] <= 'z')))
```
kisoo-han: For the readability, I propose:

`if (!isalpha (hostname[0]))`

src/connection/host_lookup.c
```diff
@@ -462,6 +433,46 @@ ip_format_check (char *ip_addr)
...
+    {
+      return false;
+    }
+
+
+  if (!((hostname[str_len - 1] >= 'A' && hostname[str_len - 1] <= 'Z') ^
+	(hostname[str_len - 1] >= 'a' && hostname[str_len - 1] <= 'z') ^
+	(hostname[str_len - 1] >= '0' && hostname[str_len - 1] <= '9')))
```
kisoo-han: For the same reason,
if (!isalnum (hostname[str_len - 1]))

src/connection/host_lookup.c
eido5: It will be better. I will apply it.

src/connection/host_lookup.c
eido5: If then, I will change "ip_format_check" to "is_valid_ip" either.

src/connection/host_lookup.c
```diff
@@ -462,6 +433,46 @@ ip_format_check (char *ip_addr)
...
+    {
+      return false;
+    }
+
+  for (char_num = 1; char_num < str_len - 1; char_num++)
+    {
+      if (!((hostname[char_num] >= 'A' && hostname[char_num] <= 'Z') ^
+	    (hostname[char_num] >= 'a' && hostname[char_num] <= 'z') ^
+	    (hostname[char_num] >= '0' && hostname[char_num] <= '9') ^
+	    (hostname[char_num] == '-') ^ (hostname[char_num] == '.')))
```
kisoo-han: if (!(isalnum (hostname[char_num]) || hostname[char_num] == '-' || hostname[char_num] == '.'))

kisoo-han:
It is recommended to divide the message for hostname lookup error into two types according to the value of the **use_user_hosts** system parameter as follows. Reference of "/etc/hosts" should be modified in case of value of use_user_hosts is true.
* Cannot find ("(unknown)") in /etc/hosts. Please check permissions of the file or if there is the hostname.
* Cannot find ("(unknown)") in **$CUBRID/conf/cubrid_hosts.conf**. Please check permissions of the file or if there is the hostname.

eido5:
I edited error message format only in the "msg/en_US/cubrid.msg".
If it is edited well, please reply on this comment. and then I will edit other languages one.
@kisoo-han

kisoo-han:
Good. Go ahead!

iOS용 Outlook<https://aka.ms/o0ukef> 다운로드
________________________________
보낸 사람: eido5 ***@***.***>
보낸 날짜: Thursday, June 15, 2023 8:46:00 PM
받는 사람: CUBRID/cubrid ***@***.***>
참조: 한기수 ***@***.***>; Mention ***@***.***>
제목: Re: [CUBRID/cubrid] [CBRD-24834] Utility also works with invalid cubrid_hosts.conf file configuration (PR #4414)


I edited error message format only in the "msg/en_US/cubrid.msg".
If it is edited well, please reply on this comment. and then I will edit other languages one.
@kisoo-han<https://github.com/kisoo-han>

src/connection/host_lookup.c
```diff
@@ -346,74 +345,46 @@ load_hosts_file ()
       if (hostname[0] && ipaddr[0])
 	{
 	  /*not duplicated hostname, IP address or not duplicated hostname and duplicated IP address */
-	  if ((user_host_Map.find (hostname) == user_host_Map.end ()))
+	  if ((user_host_Map.find (hostname) == user_host_Map.end ())
+	      && (user_host_Map.find (ipaddr) == user_host_Map.end ()))
```
kisoo-han, line 349: This function can replace alias. I'd like to keep the original.

```
/etc/hosts
192.168.2.34 dev1 dev-1
```

```
cubrid_hosts.conf
192.168.2.34 dev1
192.168.2.34 dev-1
```

eido5: I knew that etc/hosts saves first of one if there are duplicated ip address.
But in this case, this hash map function saves latest one. So, I wrote above code.

So, when you find '192.168.2.34' and dev-1 is returned.

src/connection/host_lookup.c
```diff
@@ -298,42 +299,40 @@ load_hosts_file ()
...
 		  hostent_flag = INSERT_HOSTNAME;
 		}
 	    }
 	  else
 	    {
-	      if (strlen (token) > HOSTNAME_LEN - 1)
+	      str_len = strlen (temp_token);
+	      if (str_len > HOSTNAME_LEN - 1)
 		{
-		  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_UHOST_HOST_NAME_TOO_LONG, 3, token, line_num,
```
eido5, line 322: Should I roll back this coded?
Because user cannot know which one is the problem.
@kisoo-han

kisoo-han: Ok, let's keep change.

eido5: Ok. I will recover it.

src/connection/host_lookup.c
kisoo-han: Ok, let's keep change.

src/connection/host_lookup.c
kisoo-han: Considering compatibility with glibc, we don't need to give these error messages at library level. If necessary, it is better to do it in the function that call host lookup.
So, do not roll back.

eido5: It does not print error into the stdout. Just record in the log. 
If you mean that log recording is also banned. I will keep it.

eido5:
I think we should finish task about what we are requested or we fixed about the error message in this PR.
(at least **direct related with** host look up)

1. fprintf (stderr, "unknown host : %s\n", hostname); This message is refered in the last meeting. So, we need fix it because we fixed the "er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_GAI_ERROR, 1, hostname, HOSTS_FILE);" 
2. Error log recording (**not printed to the stdout**) should be fixed about the over 256 length hostname.

I think above two things what we did similar in this PR, So we need to finish these.
If not, It remains the code integrity violation.
And we can decide beyond master or broker error message can be handled this fig release.

kisoo-han (approved):
please merge and mark CBRD-24834 as resolved.

PR #4416, opened 2023-06-12
[LETS-741] Temporary fix for CI tests

joohok:
http://jira.cubrid.org/browse/LETS-741

Purpose

use oldHA mechanism to make TS writable
CI test will be run successfully after #4400 is merged

Implementation

Use  `hb_resource_job_change_mode ()` to change the Server state to writable.

### DIFF
--- a/src/executables/master_heartbeat.c
+++ b/src/executables/master_heartbeat.c
@@ -3695,9 +3695,9 @@ hb_resource_job_change_mode (HB_JOB_ARG * arg)
 	   * Transaction Server can not execute the write transaction.
 	   * The Transaction Server state need to be changed to HB_PSTATE_REGISTERED_AND_ACTIVE by this job in order to
 	   * execute the write transaction, which internally changes the db_Disable_modification to false
-           * in css_change_ha_server_state () -> logtb_enable_update ().
+	   * in css_change_ha_server_state () -> logtb_enable_update ().
 	   * When the TS state is defined and this job is re-defined for newHA,
-           * these modified parts will also be addressed.
+	   * these modified parts will also be addressed.
 	   */
 
 	  continue;
--- a/src/executables/master_heartbeat.c
+++ b/src/executables/master_heartbeat.c
@@ -3688,7 +3688,18 @@ hb_resource_job_change_mode (HB_JOB_ARG * arg)
     {
       if (proc->type != HB_PTYPE_TRAN_SERVER)
 	{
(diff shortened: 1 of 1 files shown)


cristiarg (approved):

hornetmj (approved):

src/executables/master_heartbeat.c
```diff
@@ -3686,8 +3686,9 @@ hb_resource_job_change_mode (HB_JOB_ARG * arg)
   rv = pthread_mutex_lock (&hb_Resource->lock);
   for (proc = hb_Resource->procs; proc; proc = proc->next)
     {
-      if (proc->type != HB_PTYPE_SERVER)
+      if (proc->type != HB_PTYPE_TRAN_SERVER)
 	{
+	  // TODO: Temporary fix for CI test failure, this can be replaced when all the states for TS are defined.
```
wrlawodms: I think we should write a comment about what makes the CI fail and how this change fixes it temporarily. It's not certain even fo me now.

wrlawodms: and when it's reverted. I mean the requirements to revert it.

joohok: OK, I agree. It would be easier to track (understand) the issues behind it when handling the TODO section.

src/executables/master_heartbeat.c
```diff
@@ -3686,8 +3686,20 @@ hb_resource_job_change_mode (HB_JOB_ARG * arg)
   rv = pthread_mutex_lock (&hb_Resource->lock);
   for (proc = hb_Resource->procs; proc; proc = proc->next)
     {
-      if (proc->type != HB_PTYPE_SERVER)
-	{
+      if (proc->type != HB_PTYPE_TRAN_SERVER)
```
wrlawodms, line 3689: Q1. How does this change make the server writable?
Q2. Does it allow all PTS to write? (It's OK, I just want to understand. or it needs to be described?)

joohok: A1. This job is invoked during failover. 
 When a node becomes the master node, this part finds the TS process among the processes registered in cub_master and sends a request to enable write transactions. TS receives this request and changes the value of the global variable (db_Disable_modification) that was blocking write transactions.

With the removal of the HB_PTYPE_SERVER type, this job was not executed.

A2. No, only the PTS in the master node.

wrlawodms (approved):
with the what-to-do comment

PR #4418, opened 2023-06-12
[CBRD-24843] Eliminate joins must not occur unless all columns of the primary key and foreign key are used as join predicates

youngjinj:
http://jira.cubrid.org/browse/CBRD-24843

There are three changes.
1. Eliminate joins must not occur unless all columns of the primary key and foreign key are used as join predicates.
   ```
   e.g.
   
   drop table if exists child, parent;
   create table parent (c1 int, c2 int, primary key (c1, c2));
   create table child (c1 int, c2 int);
   alter table child add constraint foreign key (c1, c2) references parent (c1, c2);
   
   select c.* from child c, parent p where c.c1 = p.c1 and c.c2 = p.c2;
   select c.* from child c, parent p where c.c1 = p.c1 and c.c2 = p.c2 and c.c1 = p.c1 and c.c2 = p.c2;
     -> select c.* from child c where c1 is not null and c2 is not null;
   
   select c.* from child c, parent p where c.c1 = p.c1;
   select c.* from child c, parent p where c.c2 = p.c2;
   select c.* from child c, parent p where c.c1 = p.c1 and c.c1 = p.c1;
     -> do not change.
   ```
2. Some predicates are non-join predicates, but can be reduced by predicate fulfillment.
   - 'p.c1 = 1' is a reference to the parent, but can be reduced if 'c.c1 = 1' exists. In the rewrite query, since 'c.c1 = 1' is added by predicate fulfillment, eliminate join is possible even if 'p.c1 = 1' exists.
   - Predicate candidates that can be reduced by predicate fulfillment are added to parent_pred_point_list.
   - This check is performed by the qo_check_reduce_predicate_for_parent_spec function.
   ```
   e.g.
   
   drop table if exists child, parent;
   create table parent (c1 int, c2 int, primary key (c1, c2));
   create table child (c1 int, c2 int);
   alter table child add constraint foreign key (c1, c2) references parent (c1, c2);
   
   select c.* from child c, parent p where c.c1 = p.c1 and c.c2 = p.c2 and p.c1 = 1;
   
   -- rewritten query
   select c.* from child c, parent p where c.c1 = p.c1 and c.c2 = p.c2 and p.c1 = 1 and c.c1 = 1;

   -- execute query
   select c.* from child c where c.c1 = 1;
   ```
3. Rename reduce_pred_point_list to join_pred_point_list.


xmilex-git (approved):

src/optimizer/query_rewrite.c
```diff
@@ -3846,10 +3865,39 @@ qo_check_primary_key_referenced_by_foreign_key_in_parent_spec (PARSER_CONTEXT *
...
+	  if (curr_pred->or_next == NULL && PT_NODE_IS_EXPR (curr_pred) && PT_EXPR_OP (curr_pred) == PT_EQ)
 	    {
-	      reduce_pred_point_list = parser_append_node (pt_point (parser, curr_pred), reduce_pred_point_list);
+	      if (pt_is_attr (PT_EXPR_ARG1 (curr_pred)) && pt_is_attr (PT_EXPR_ARG2 (curr_pred)))
+		{
+		  reduce_pred_point_list = parser_append_node (pt_point (parser, curr_pred), reduce_pred_point_list);
+		  continue;
+		}
+
+	      /* This is to ignore the predicate of the parent added by the pt_lambda_node function.
```
shparkcubrid: pt_lambda_node() is called in view-merge, push-predicate. I think it doesn't seem to be a problem. However, since the lambda_xxx() function is a commonly used function, I am concerned that there may be an exception case.
Instead of this, how about excluding the predicate having same column as the column of join condition as below?
```
select b.* from parent a, child b where a.fk = b.fk and a.fk =3  -- a can be removed
```

youngjinj: Fixed it. Please review again.

Added qo_check_reduce_predicate_for_parent_spec function.
And this function checks the parent_pred_point_list.

shparkcubrid (approved):

PR #4419, opened 2023-06-12
[LETS-676] Add unit tests for async_disconnect_handler

wrlawodms:
http://jira.cubrid.org/browse/LETS-676

Add the unit test for the async_disconnect_handler. Thank you for the detailed guide @cristiarg.
The main goal of this unit test is that the handler clears connections properly and cleanly. To test it, 
- The `connection_handler_mock` sleeps for a while during its destruction. 
- Count the number of connection_handler and make sure all of them are cleared at the end of tests.

There are two tests. 
- Add a simple test to add a connection_handler and disconnect it.
- Add several (50) connection_handlers with a random destruction delay and disconnect them.


unit_tests/server/test_async_disconnect_handler_main.cpp
```diff
@@ -0,0 +1,98 @@
...
+{
+  async_disconnect_handler<connection_handler_mock> disconn_handler;
+  std::atomic<int> instance_counter = 0;
+
+  std::random_device rd;
+  std::mt19937 gen (rd ());
+  std::uniform_int_distribution<int> dis (10, 20000); // [10, 20000] usec
+
+  constexpr int cnt_conn =
+	  50; // time consumed for disconnecting all = cnt_conn * dis[] = 10*50 ~ 20000*50 = 0.5ms ~ 1000ms
```
wrlawodms, line 82: @cristiarg It can consume 1s at most. Is it OK?

wrlawodms: Actually, I considered adding another test like
- Add and disconnect a connection that takes 1s to destroy.
- sleep (0.5s)
- Add and disconnect more connections and terminate the async_disconnect_handler.

to make sure that `async_disconnect_handler` terminates well when there are connections left in its internal queue after termination (the async_disconnect_handler takes all connections and clears them periodically (1s)).

However, it takes at least one more second to sleep and it depends on the internal constant for the period (1s) in the `async_disconnect_handler` and seems implement-specific. So I'm hesitant to add this test. What do you think?

unit_tests/server/test_async_disconnect_handler_main.cpp
```diff
@@ -0,0 +1,98 @@
...
+#define CATCH_CONFIG_MAIN
+#include "catch2/catch.hpp"
+
+#include "async_disconnect_handler.hpp"
+
+#include <thread>
+#include <random>
+#include <chrono>
+#include <atomic>
+
```
joohok, line 30:
```suggestion
#include <array>

```
I think `<array>` is needed to fix window build error

hornetmj (approved):
LGTM.

joohok (approved):

unit_tests/server/test_async_disconnect_handler_main.cpp
wrlawodms: Thank you. I missed it.

unit_tests/server/test_async_disconnect_handler_main.cpp
cristiarg: It is ok as done.
A unit test is much more than `async_disconnect_handler` could have hoped for. :)
Thanks.

cristiarg (approved):

PR #4428, opened 2023-06-16
[CBRD-24409] Add NDV of columns to catalog info

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-24409

- Add NDV to catalog information of class
- Since the query-based module is in the client, NDV is obtained from the client-side and transmitted to the server-side for processing.
- Update statistics
-- Call stats_get_ndv_by_query() to get NDV and send it to cub_server. The NDV is stored in cub_server.
- update statistics on all classes
  -- Changed from server-side execution to client-side execution.
- loaddb (C/S mode)
  -- Receive class OID from the server-side and update statistics on the client-side.
- partitioned table
  -- When updating partition index statistics, The statistics is changed from average to sum.


src/object/schema_manager.c
```diff
@@ -4446,6 +4446,10 @@ sm_update_all_catalog_statistics (bool with_fullscan)
     CT_METHFILE_NAME, CT_QUERYSPEC_NAME, CT_INDEX_NAME,
     CT_INDEXKEY_NAME, CT_CLASSAUTH_NAME, CT_DATATYPE_NAME,
     CT_COLLATION_NAME, CT_CHARSET_NAME, CT_SYNONYM_NAME,
+    CT_STORED_PROC_NAME, CT_STORED_PROC_ARGS_NAME, CT_PARTITION_NAME,
+    CT_SERIAL_NAME, CT_COLLATION_NAME, CT_USER_NAME,
+    CT_TRIGGER_NAME, CT_PASSWORD_NAME, CT_CHARSET_NAME,
+    CT_DB_SERVER_NAME, CT_SYNONYM_NAME,
```
youngjinj: The following is redundant.
- CT_COLLATION_NAME
- CT_CHARSET_NAME
- CT_SYNONYM_NAME

And was it intentional to exclude the following?
- CT_AUTHORIZATION_NAME
- CT_HA_APPLY_INFO_NAME

src/storage/statistics_cl.c
```diff
@@ -342,152 +340,148 @@ stats_dump (const char *class_name_p, FILE * file_p)
...
 	  break;
 
 	case DB_TYPE_MONETARY:
-	  fprintf (file_p, "DB_TYPE_MONETARY\n");
+	  fprintf (file_p, "MONETARY)\n");
 	  break;
 
 	case DB_TYPE_DATE:
-	  fprintf (file_p, "DB_TYPE_DATE\n");
+	  fprintf (file_p, "DB_TYPE_DATE)\n");
```
youngjinj: I think it should be DATE, right?

src/storage/statistics_cl.c
```diff
@@ -342,152 +340,148 @@ stats_dump (const char *class_name_p, FILE * file_p)
...
       name_p = sm_get_att_name (class_mop, attr_stats_p->id);
-      fprintf (file_p, " Attribute: %s\n", (name_p ? name_p : "not found"));
-      fprintf (file_p, "    id: %d\n", attr_stats_p->id);
-      /*
-         fprintf (file_p, "    ndv: %ld\n", attr_stats_p->ndv);
-       */
-      fprintf (file_p, "    Type: ");
+      fprintf (file_p, " Attribute: %s (", (name_p ? name_p : "not found"));
 
       switch (attr_stats_p->type)
```
youngjinj: Instead of a switch statement, what about the following?
```
fprintf (file_p, "%s)\n", pr_type_name (attr_stats_p->type));
```
AS-IS will break consistency when new types are added.
For example, there is no `DB_TYPE_JSON` in the switch statement.
In src/object/object_primitive.c, see after line 949.

Even if this method is not applied, `DB_TYPE_JSON` must be added.

src/storage/statistics_cl.c
```diff
@@ -542,7 +534,7 @@ stats_ndv_dump (const char *class_name_p, FILE * file_p)
     }
 
   /* get NDV by query */
-  if (stats_get_ndv_by_query (class_mop, &class_attr_ndv, file_p) != NO_ERROR)
+  if (stats_get_ndv_by_query (class_mop, &class_attr_ndv, file_p, 0) != NO_ERROR)
```
youngjinj: This is the magic number. I think we should use the following.
```suggestion
  if (stats_get_ndv_by_query (class_mop, &class_attr_ndv, file_p, STATS_WITH_SAMPLING) != NO_ERROR)
```

src/storage/statistics_cl.c
```diff
@@ -614,7 +606,14 @@ stats_get_ndv_by_query (const MOP class_mop, CLASS_ATTR_NDV * class_attr_ndv, FI
     }
 
   /* create sampling SQL statement */
-  query = "SELECT /*+ SAMPLING_SCAN */ %s FROM [%s]";
+  if (with_fullscan)
```
youngjinj: It is trivial.
I think it's better for readability to compare `with_fullscan` with `STATS_WITH_FULLSCAN` and `STATS_WITH_SAMPLING`.

src/storage/statistics_cl.c
```diff
@@ -530,8 +524,6 @@ stats_ndv_dump (const char *class_name_p, FILE * file_p)
 {
   MOP class_mop;
   int error = NO_ERROR;
```
youngjinj: It is trivial.
Not used. But if it is used later, keep it,

src/communication/network_interface_cl.c
```diff
@@ -5682,36 +5736,205 @@ stats_update_all_statistics (int with_fullscan)
...
+      free_and_init (request);
+    }
+
+  return (error == DB_CURSOR_END) ? NO_ERROR : error;
+
 #else /* CS_MODE */
-  int success;
+  int error = NO_ERROR;
+  CLASS_ATTR_NDV class_attr_ndv;
+  const char *query = "SELECT class_name FROM db_class";
```
youngjinj: I think it should be the same as the query in CS_MODE.

src/object/schema_manager.c
shparkcubrid: I fixed it.

src/storage/statistics_cl.c
shparkcubrid: Good point. I fixed it.

src/storage/statistics_cl.c
shparkcubrid: I fixed it.

src/storage/statistics_cl.c
shparkcubrid: I fixed it.

src/communication/network_interface_cl.c
shparkcubrid: I fixed it.

xmilex-git (approved):

src/storage/statistics_cl.c
shparkcubrid: I fixed it.

src/communication/network_interface_sr.c
```diff
@@ -10350,28 +10335,59 @@ sloaddb_interrupt (THREAD_ENTRY * thread_p, unsigned int rid, char *request, int
...
+  char *reply = OR_ALIGNED_BUF_START (a_reply);
   load_session *session = NULL;
   int error_code = session_get_load_session (thread_p, session);
-  if (error_code == NO_ERROR)
-    {
-      assert (session != NULL);
+  std::vector < const cubload::class_entry * >class_entries;
 
-      session->update_class_statistics (*thread_p);
+  /* check disable_statistics */
```
youngjinj, line 10353: NULL check for session is required.
AS-IS assumes that session is not NULL when it is NO_ERROR.

youngjinj (approved):

src/communication/network_interface_sr.c
shparkcubrid: I fixed it.

PR #4432, opened 2023-06-19
[CBRD-24850] Enhancing group_concat() Performance by Optimizing String Size and Length Calculation Methods

xmilex-git:
http://jira.cubrid.org/browse/CBRD-24850

This pull request introduces the following changes to enhance the performance of the group_concat() function:

1. A **_medium.length_** component has been added to DB_CHAR, which is one of the constituents of DB_VALUE (db_value->data.ch.medium.length). Code has been added to **initialize this component to -1** when executing db_value_domain_default() and db_make_db_char() functions. (src/compat/dbtype_def.h, src/compat/db_macro.c, src/compat/dbtype_function.i)
2. When copying DB_VALUE, the pr_clone_value() function is executed, which in turn carries out mr_setval_string() on the destination. The code has been modified to also copy the **_length_** value during this process. (src/object/object_primitive.c)
3. A code modification has been included that checks if a **_length_** value is valid when executing the db_get_string_length() function. If the **_length_** value is valid, it is used. If not, a value obtained through intl_char_count() is stored in **_length_**.  (src/query/string_opfunc.c)
4. The code within qstr_concatenate has been updated to minimize the execution of the intl_char_count() and intl_char_size() functions. Instead, these values are now pre-received as arguments. (src/query/string_opfunc.c)

### DIFF
--- a/src/compat/dbtype_def.h
+++ b/src/compat/dbtype_def.h
@@ -1010,6 +1010,7 @@ extern "C"
       const char *buf;
       int compressed_size;
       char *compressed_buf;
+      int length;
     } medium;
     struct
     {
--- a/src/compat/dbtype_def.h
+++ b/src/compat/dbtype_def.h
@@ -1010,7 +1010,7 @@ extern "C"
       const char *buf;
       int compressed_size;
       char *compressed_buf;
-      int length;		//Only Use for group_concat() now
+      int length;		/* Only Use for group_concat() now */
     } medium;
     struct
     {
--- a/src/compat/dbtype_def.h
+++ b/src/compat/dbtype_def.h
@@ -1010,7 +1010,7 @@ extern "C"
       const char *buf;
       int compressed_size;
       char *compressed_buf;
-      int length;
+      int length;		//Only Use for group_concat() now
     } medium;
     struct
     {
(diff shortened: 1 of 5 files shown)


src/compat/dbtype_def.h
```diff
@@ -1010,6 +1010,7 @@ extern "C"
       const char *buf;
       int compressed_size;
       char *compressed_buf;
+      int length;		//Only Use for group_concat() now
```
youngjinj: It is trivial.
I want it to be a comment format similar to the surroundings.
```suggestion
      int length;		/* Only Use for group_concat() now */
```

src/query/string_opfunc.c
```diff
@@ -8076,7 +8079,12 @@ db_get_string_length (const DB_VALUE * value)
   str = value->data.ch.medium.buf;
   length = size = value->data.ch.medium.size;
   codeset = (INTL_CODESET) value->data.ch.medium.codeset;
+  DB_VALUE *p = (DB_VALUE *) value;
 
+  if (p->data.ch.medium.length != -1)
+    {
+      return p->data.ch.medium.length;
+    }
```
youngjinj: How about this?
```
    int
    db_get_string_length (const DB_VALUE * value)
    {
      DB_CONST_C_CHAR str;
      int size;
      INTL_CODESET codeset;
      int length = 0;

(+)   if (value->data.ch.medium.length != -1)
(+)     {
(+)       return value->data.ch.medium.length;
(+)     }

      str = value->data.ch.medium.buf;
      length = size = value->data.ch.medium.size;
      codeset = (INTL_CODESET) value->data.ch.medium.codeset;
      ...
```

shparkcubrid (approved):

youngjinj (approved):

src/query/string_opfunc.c
xmilex-git: Based on this opinion, the code was modified: The commit message is
'(DB_VALUE*)p->data => value->data'

PR #4435, opened 2023-06-20
[LETS-725] Register the cub_server process to cub_master when it is created

joohok:
http://jira.cubrid.org/browse/LETS-725

Purpose

In newHA, it is necessary for cub_master to manage the cub_server processes before their recovery is completed.
Therefore, it is intended to establish a connection between cub_master and the server before the recovery starts, and register the TS and PS processes to cub_master.

### DIFF
--- a/src/connection/server_support.c
+++ b/src/connection/server_support.c
@@ -1290,9 +1290,10 @@ css_start_shutdown_server ()
 }
 
 int
-css_ha_register (const char *server_name)
+css_register_ha_server (const char *server_name)
 {
   assert (server_name != NULL);
+  assert (!HA_DISABLED ());
 
   CSS_CONN_ENTRY *conn;
   std::string message_to_master = css_pack_message_to_master (server_name);
--- a/src/connection/server_support.c
+++ b/src/connection/server_support.c
@@ -1289,29 +1289,41 @@ css_start_shutdown_server ()
   css_Server_shutdown_inited = true;
 }
 
-void
+int
 css_ha_register (const char *server_name)
 {
+  assert (server_name != NULL);
+
   CSS_CONN_ENTRY *conn;
   std::string message_to_master = css_pack_message_to_master (server_name);
+
   conn =
     css_connect_to_master_server (prm_get_integer_value (PRM_ID_TCP_PORT_ID), message_to_master.c_str (),
 				  message_to_master.size ());
   if (conn != NULL)
     {
-      int status =
-	hb_register_to_master (conn,
-			       get_server_type () ==
-			       SERVER_TYPE_TRANSACTION ? HB_PTYPE_TRAN_SERVER : HB_PTYPE_PAGE_SERVER);
-      css_Master_conn = conn;
+      int status = hb_register_to_master (conn,
+					  get_server_type () ==
+					  SERVER_TYPE_TRANSACTION ? HB_PTYPE_TRAN_SERVER : HB_PTYPE_PAGE_SERVER);
 
       if (status != NO_ERROR)
 	{
 	  fprintf (stderr, "failed_to_hearbeat register");
 	  css_close_connection_to_master ();
+
+	  return status;
+	}
+      else
+	{
+	  css_Master_conn = conn;
+	  return NO_ERROR;
 	}
     }
-
+  else
+    {
+      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ERR_CSS_ERROR_DURING_SERVER_CONNECT, 1, server_name);
+      return ERR_CSS_ERROR_DURING_SERVER_CONNECT;
+    }
 }
 
 /*
--- a/src/connection/server_support.c
+++ b/src/connection/server_support.c
@@ -1289,6 +1289,31 @@ css_start_shutdown_server ()
   css_Server_shutdown_inited = true;
 }
 
+void
+css_ha_register (const char *server_name)
+{
+  CSS_CONN_ENTRY *conn;
+  std::string message_to_master = css_pack_message_to_master (server_name);
+  conn =
+    css_connect_to_master_server (prm_get_integer_value (PRM_ID_TCP_PORT_ID), message_to_master.c_str (),
+				  message_to_master.size ());
+  if (conn != NULL)
+    {
+      int status =
+	hb_register_to_master (conn,
+			       get_server_type () ==
+			       SERVER_TYPE_TRANSACTION ? HB_PTYPE_TRAN_SERVER : HB_PTYPE_PAGE_SERVER);
+      css_Master_conn = conn;
+
+      if (status != NO_ERROR)
+	{
+	  fprintf (stderr, "failed_to_hearbeat register");
+	  css_close_connection_to_master ();
+	}
+    }
+
+}
+
 /*
  * css_init() -
  *   return:
@@ -1361,7 +1386,16 @@ css_init (THREAD_ENTRY * thread_p, const char *server_name, int port_id)
 
   css_Server_connection_socket = INVALID_SOCKET;
 
-  conn = css_connect_to_master_server (port_id, message_to_master.c_str (), (int) message_to_master.size ());
+  if (!HA_DISABLED ())
+    {
+      // css_Master_conn is already set by css_ha_register ()
+      conn = css_Master_conn;
+    }
+  else
+    {
+      conn = css_connect_to_master_server (port_id, message_to_master.c_str (), (int) message_to_master.size ());
+    }
+
   if (conn != NULL)
     {
       /* insert conn into active conn list */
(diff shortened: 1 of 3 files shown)


src/connection/server_support.c
```diff
@@ -1361,7 +1401,16 @@ css_init (THREAD_ENTRY * thread_p, const char *server_name, int port_id)
 
   css_Server_connection_socket = INVALID_SOCKET;
 
-  conn = css_connect_to_master_server (port_id, message_to_master.c_str (), (int) message_to_master.size ());
+  if (!HA_DISABLED ())
+    {
+      // css_Master_conn is already set by css_ha_register ()
```
hornetmj:
```suggestion
      // css_Master_conn is already set by css_register_ha_server ()
```

src/connection/server_support.c
```diff
@@ -1372,19 +1421,6 @@ css_init (THREAD_ENTRY * thread_p, const char *server_name, int port_id)
...
-	    hb_register_to_master (css_Master_conn,
-				   is_transaction_server ()? HB_PTYPE_TRAN_SERVER : HB_PTYPE_PAGE_SERVER);
-	  if (status != NO_ERROR)
-	    {
-	      fprintf (stderr, "failed to heartbeat register.\n");
-	    }
-	}
-#endif
-
       if (status == NO_ERROR)
```
hornetmj: Checking the status is for the deleted(previous) code.

src/connection/server_support.c
```diff
@@ -1372,19 +1421,6 @@ css_init (THREAD_ENTRY * thread_p, const char *server_name, int port_id)
       css_Pipe_to_master = conn->fd;
       css_Master_conn = conn;
```
hornetmj: duplicated.

src/connection/server_support.c
```diff
@@ -1289,6 +1289,46 @@ css_start_shutdown_server ()
...
 }
 
+int
+css_register_ha_server (const char *server_name)
+{
+  assert (server_name != NULL);
+  assert (!HA_DISABLED ());
+
+  CSS_CONN_ENTRY *conn;
+  std::string message_to_master = css_pack_message_to_master (server_name);
```
hornetmj, line 1305: error handling. (NULL check)

src/connection/server_support.c
```diff
@@ -1289,6 +1289,46 @@ css_start_shutdown_server ()
...
+      else
+	{
+	  // established connection will be re-used in css_init () after server recovery.
+	  css_Master_conn = conn;
+	  return NO_ERROR;
+	}
+    }
+  else
+    {
+      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ERR_CSS_ERROR_DURING_SERVER_CONNECT, 1, server_name);
```
hornetmj, line 1339: I think the following error message is not enough to describe this un-expected result. It seems like you just follow the error handling in the css_init(). Share your opinion.
> ERR_CSS_ERROR_DURING_SERVER_CONNECT -368 Communication error during connect for %1$s.

src/connection/server_support.c
```diff
@@ -1289,6 +1289,46 @@ css_start_shutdown_server ()
...
+				  message_to_master.size ());
+  if (conn != NULL)
+    {
+      int status = hb_register_to_master (conn,
+					  get_server_type () ==
+					  SERVER_TYPE_TRANSACTION ? HB_PTYPE_TRAN_SERVER : HB_PTYPE_PAGE_SERVER);
+
+      if (status != NO_ERROR)
+	{
+	  fprintf (stderr, "failed to hearbeat register to master \n");
```
hornetmj:
```suggestion
	  fprintf (stderr, "failed to hearbeat register to master\n");
```

src/connection/server_support.c
hornetmj: And how about inserting assert(css_Master_conn != NULL)?

src/connection/server_support.c
hornetmj: HA_MODE is not supported on the Windows platform. So, how about considering this case or just adding assert(false) temparally?

joohok: return data type of `css_pack_message_to_master ()` is `std::string`.
but `css_pack_message_to_master ()` returns NULL for exception case, and there will be crash in that case. 
It should not return NULL using data type `std::string`. 

So, I think `css_pack_message_to_master ()` should be refactored first, and then we can handle the error case.

joohok: Does it mean using "#if defined (WINDOWS)" to handle the case?

src/connection/server_support.c
joohok: It is true that it is the same as the error situation in css_init(). 
If an error occurs during the connection to cub_master from a server with a `server_name`, 
the above error message is printed out. 

I agree that the error message itself is not clear in explaining the situation. 

Changing it to 
'Communication error during connect to cub_master from %1$s.' 
would be more appropriate, but it goes beyond the scope of the current issue.

src/connection/server_support.c
joohok: Done

src/connection/server_support.c
joohok: Done.

src/connection/server_support.c
joohok: Move it below the 
`conn = css_connect_to_master_server (); `

src/connection/server_support.c
joohok: Deleted.

src/connection/server_support.c
joohok: Considering the Windows environment is already taken into account in the functions being called, 
it seems straightforward to add assert(false) using #if defined(WINDOWS). 
what do you think? 

In the `hb_register_to_master()` function, an error will be returned if it is a Windows environment. When a `hb_register_to_master()` request is received in the cub_master thread in Windows, cub_master thread will send an error to the requester.

src/connection/server_support.c
```diff
@@ -1289,6 +1289,46 @@ css_start_shutdown_server ()
...
+  CSS_CONN_ENTRY *conn;
+  std::string message_to_master = css_pack_message_to_master (server_name);
+
+  // connection is established before server recovery, only when ha_mode is turned on.
+  conn =
+    css_connect_to_master_server (prm_get_integer_value (PRM_ID_TCP_PORT_ID), message_to_master.c_str (),
+				  message_to_master.size ());
+  if (conn != NULL)
+    {
+      int status = hb_register_to_master (conn,
```
wrlawodms: Nit.
```suggestion
      const int status = hb_register_to_master (conn,
```

src/transaction/boot_sr.c
```diff
@@ -2263,6 +2263,18 @@ boot_restart_server (THREAD_ENTRY * thread_p, bool print_restart, const char *db
       goto error;
     }
 
+#if defined(SERVER_MODE)
+  if (!HA_DISABLED ())
+    {
+      // This must be called after init_server_type () because cub_master have to know which type of server it is.
+      error_code = css_register_ha_server (db_name);
```
wrlawodms, line 2282: Need a comment shortly about why it's needed here. At least the mention that it has to be before recovery.

src/connection/server_support.c
```diff
@@ -1370,26 +1423,9 @@ css_init (THREAD_ENTRY * thread_p, const char *server_name, int port_id)
       css_Master_server_name = strdup (server_name);
       css_Master_port_id = port_id;
       css_Pipe_to_master = conn->fd;
```
wrlawodms: How about putting them into the above `else` clause and setting them `css_register_ha_server()` in the case of HA? I know they are the same value here and there finally, but I think it's better if it's set by the same source at the same time it's used, which is more safe and open to change.
```suggestion
      css_Master_server_name = strdup (server_name);
      css_Master_port_id = port_id;
      css_Pipe_to_master = conn->fd;
```

wrlawodms (commented):
Looks good overall.

src/connection/server_support.c
joohok: I added `#if !defined (WINDOWS)` at the caller, similar to existing code in `css_init ()` where it calls `hb_register_to_master ()`.

src/connection/server_support.c
```diff
@@ -1361,41 +1407,38 @@ css_init (THREAD_ENTRY * thread_p, const char *server_name, int port_id)
...
     {
-      /* insert conn into active conn list */
-      css_insert_into_active_conn_list (conn);
-
-      css_Master_server_name = strdup (server_name);
-      css_Master_port_id = port_id;
-      css_Pipe_to_master = conn->fd;
-      css_Master_conn = conn;
+      // These global variables are set in css_register_ha_server ()
+      assert (css_Master_port_id != -1);
```
wrlawodms, line 1421:
```suggestion
      assert (css_Master_server_name != NULL);
      assert (css_Master_port_id != -1);
```

wrlawodms (approved):

src/connection/server_support.c
wrlawodms: > return data type of css_pack_message_to_master () is std::string.
but css_pack_message_to_master () returns NULL for exception case, and there will be crash in that case.
It should not return NULL using data type std::string.

I think we can just make an issue to track it and go ahead.

joohok: I added, http://jira.cubrid.org/browse/LETS-760

src/connection/server_support.c
hornetmj: The error code ERR_CSS_TCP_CANNOT_CONNECT_TO_MASTER you mentioned in the comment can be set in the callstack you use (css_connect_to_master_server() -> css_common_connect()).

Let's leave it you are written for now.

hornetmj (approved):

PR #4436, opened 2023-06-20
[LETS-742] expand communication support functions to allow partial read and completion

cristiarg:
http://jira.cubrid.org/browse/LETS-742

The existing function `css_net_recv` used throughout the codebase behaves the following way:
- expects a buffer of a specified length
- reads up to the maximum size of that buffer
- if the data to be read is larger, it just reads the rest of the data and throws it away while returning `RECORD_TRUNCATED` to the caller

This way of working impedes implementation of a flexible receiving mechanism as the application layer must be sure that the supplied buffer is sufficiently large or data will be lost.

Implementations:
- added function `css_net_recv_allow_truncated` - simiar to `css_net_recv` which also returns `RECORD_TRUNCATED` (maybe a new return value would be more appropriate - there is core `INTERRUPTED_READ` which is not used) but does not read and throw away the remaining data
- upon receiving `RECORD_TRUNCATED` from a call to `css_net_recv_allow_truncated`, the caller is expected to follow up with a call to `css_net_recv_remainder` which will complete the read of the data
- `css_net_recv_remainder` assumes that the supplied buffer is exactly the same size as the remainder data to be read, otherwise error code will be returned

Code has been tested and in use in follow-up patch #4408.

Note:
This is a split - for ease of reviewing - from the full development of this feature which is visible in draft #4408 .

### DIFF
--- a/src/connection/connection_support.c
+++ b/src/connection/connection_support.c
@@ -790,7 +790,7 @@ css_net_recv_remainder (SOCKET fd, char *buffer, int *maxlen, int timeout)
       return ERROR_ON_READ;
     }
 
-  if (nbytes_read_data != *maxlen)
+  if (nbytes_read_data > *maxlen)
     {
       // this should never occur
 #ifdef CUBRID_DEBUG
--- a/src/connection/connection_support.c
+++ b/src/connection/connection_support.c
@@ -773,7 +773,8 @@ css_net_recv_allow_truncated (SOCKET fd, char *buffer, int *maxlen, int *remlen,
  *   buffer(out): buffer for data be read
  *   maxlen(in/out):
  *          in - length of buffer, must be exactly the same as the number of remaining bytes to be read;
- *          out - count of bytes that were read; exactly the same value as input
+ *          out - count of bytes that were read; normally, exactly the same value as input; in case of
+ *                error, outputs the read number of bytes
  *   timeout(in): timeout value in milli-second
  */
 css_error_code
@@ -783,7 +784,8 @@ css_net_recv_remainder (SOCKET fd, char *buffer, int *maxlen, int timeout)
   const int nbytes_read_data = css_readn (fd, buffer, *maxlen, timeout);
   if (nbytes_read_data < *maxlen)
     {
-      // expecting to be able to read exactly as expected
+      // expecting to be able to read exactly as specified
+      *maxlen = nbytes_read_data;
 #ifdef CUBRID_DEBUG
       er_log_debug (ARG_FILE_LINE, "css_net_recv_remainder: ERROR_ON_READ bytes %d\n", nbytes_read_data);
 #endif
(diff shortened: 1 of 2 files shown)


src/connection/connection_support.c
```diff
@@ -645,6 +645,164 @@ css_net_recv (SOCKET fd, char *buffer, int *maxlen, int timeout)
...
+      return RECORD_TRUNCATED;
+    }
+
+  if (nbytes_read_data != templen)
+    {
+#ifdef CUBRID_DEBUG
+      er_log_debug (ARG_FILE_LINE, "css_net_recv: returning READ_LENGTH_MISMATCH bytes %d\n", nbytes_read_data);
+#endif
+      return READ_LENGTH_MISMATCH;
+    }
```
hornetmj, line 759: I think it never happen.

src/connection/connection_support.c
```diff
@@ -645,6 +645,164 @@ css_net_recv (SOCKET fd, char *buffer, int *maxlen, int timeout)
...
+
+  if (nbytes_read_data != *maxlen)
+    {
+      // this should never occur
+#ifdef CUBRID_DEBUG
+      er_log_debug (ARG_FILE_LINE, "css_net_recv_remainder: READ_LENGTH_MISMATCH bytes %d\n", nbytes_read_data);
+#endif
+      assert (false);
+      return READ_LENGTH_MISMATCH;
+    }
```
hornetmj: The condition 'nbytes_read_data != *maxlen' is actually 'nbytes_read_data > *maxlen' because 'nbytes_read_data < *maxlen' is already checked in the previous if condition. I don't think it can happen. What is it for?

hornetmj:
@cristiarg I don't know why this PR separates css_net_recv() into css_net_recv_allow_truncated() and css_net_recv_remainder() exactly. Because remaining recv data in the socket buffer should always be consumed. I expect I will understand the purpose of this PR in the #4408.

cristiarg:
@hornetmj 

> @cristiarg I don't know why this PR separates css_net_recv() into css_net_recv_allow_truncated() and css_net_recv_remainder() exactly. Because remaining recv data in the socket buffer should always be consumed. I expect I will understand the purpose of this PR in the https://github.com/CUBRID/cubrid/pull/4408.

I added the extra functions to not touch existing functionality of `css_net_recv` which is used across the codebase.
I agree with you that all data must be consumed.

The proper way to handle this is to refactor `css_net_recv` in `develop` by replacing it with the two functions introduced here and adapting the calls as appropriate:
- either implementing the completion mechanism
- or `release_assert`ing that it can never happen.

And yes, the purpose is revealed in the next PR. It boils down to this:
- the connection support layer (`css_*` functions) already send the length and the data (as an application level protocol)
- in the network code introduced by LETS, we were doing this sending of length again
- this translated in 4 calls to "send"
  - 1. send "length of length"
  - 2. send actual length as data
  - 3. send length of data
  - 4. send actual data
- it is easy to see that 1. and 2. are useless
- the benefit of having 1. and 2. is that they both messages of fixed length and allowed the receiving side to know exactly how much data it would receive in step 4.
- removing them has the downside that the receiving side never knows in advance how much data it receives (after #4408 only steps 3. and 4. will remain)
- so these two functions allow for a mechanism for the receiving side to be able to read any length of data:
  - it can start with a smaller buffer - in call to first function
  - and it can increase the buffer as needed to read all data - in call to second function
- in practice, with the reusing of a `cubmem::extensible_buffer` the calls to the second function will be very rare (if ever)

src/connection/connection_support.c
cristiarg: I also think that, but I wanted to leave it similar to `css_net_recv`.

wrlawodms (approved):
LGTM : )))))))

PR #4438, opened 2023-06-20
[LETS-759] Temporarily speed up CI tests by using localhost rather than hostname

joohok:
http://jira.cubrid.org/browse/LETS-759

Purpose

There was a performance degradation when accessing the local Page Server through the network layer in newHA. 
As a temporary solution, if the hostname of the Page Server matches the local hostname, we intend to set the page_server_hosts to localhost:port_id.

### DIFF
--- a/src/base/server_type.cpp
+++ b/src/base/server_type.cpp
@@ -26,7 +26,7 @@
 #include "log_impl.h"
 #include "page_server.hpp"
 #include "system_parameter.h"
-#include "tcp.h"
+#include "util_func.h"
 
 #include <string>
 
@@ -40,7 +40,6 @@ SERVER_TYPE get_server_type_from_config (server_type_config parameter_value);
 transaction_server_type get_transaction_server_type_from_config (transaction_server_type_config parameter_value);
 void setup_tran_server_params_on_single_node_config ();
 int setup_tran_server_params_on_ha_mode ();
-bool is_localhost (const char *hostname);
 
 static SERVER_TYPE g_server_type = SERVER_TYPE_UNKNOWN;
 static transaction_server_type g_transaction_server_type = transaction_server_type::ACTIVE;
@@ -218,14 +217,14 @@ int setup_tran_server_params_on_ha_mode ()
 {
   char *page_server_host_list = NULL;
   constexpr size_t MAX_BUFSIZE = 4096;
-  int list_size = 0;
+  size_t list_size = 0;
 
   char ha_node_list[MAX_BUFSIZE];
   char *str, *savep;
 
   int port_id = prm_get_master_port_id ();
 
-  const char *localhost_str = "localhost";
+  constexpr const char *localhost_str = "localhost";
 
   page_server_host_list = (char *) calloc (MAX_BUFSIZE, sizeof (char)); // free is called by sysprm_final()
   if (page_server_host_list == NULL)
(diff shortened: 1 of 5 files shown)


src/base/server_type.cpp
```diff
@@ -277,6 +289,26 @@ int setup_tran_server_params_on_ha_mode ()
   return NO_ERROR;
 }
 
+bool is_localhost (const char *hostname)
```
cristiarg: why not reuse `util_is_localhost` -> `are_hostnames_equal` ?

src/base/server_type.cpp
```diff
@@ -223,6 +225,8 @@ int setup_tran_server_params_on_ha_mode ()
 
   int port_id = prm_get_master_port_id ();
 
+  const char *localhost_str = "localhost";
```
cristiarg: something with `constexpr`;
maybe even hardcode the string where needed

src/base/server_type.cpp
joohok: Thanks, I replace this function with `util_is_localhost`.

src/base/server_type.cpp
joohok: OK, done.

hornetmj (approved):

src/base/util_func.c
```diff
@@ -867,3 +873,64 @@ util_get_second_and_ms_since_epoch (time_t * secs, int *msec)
...
+
+/*
+ * util_is_localhost -
+ *
+ * return:
+ *
+ * NOTE:
+ */
+bool
+util_is_localhost (char *host)
```
cristiarg: Since you moved the functions, it is also worthwhile to "fix" them. :)
```suggestion
util_is_localhost (const char *host)
```
because `host` is passed to a `const char*`

src/base/util_func.c
```diff
@@ -45,6 +45,12 @@
 #include "system_parameter.h"
 #include "environment_variable.h"
 
+#if defined (WINDOWS)
```
cristiarg, line 48: this bit is still included in `util_common.c`
check if still needed; if not, cleanup

src/base/util_func.c
```diff
@@ -867,3 +873,64 @@ util_get_second_and_ms_since_epoch (time_t * secs, int *msec)
   assert (*msec < 1000);
 }
 // *INDENT-ON*
+
+/*
+ * util_is_localhost -
```
cristiarg: Some more description

src/base/util_func.h
```diff
@@ -79,4 +79,7 @@ extern void util_get_second_and_ms_since_epoch (time_t * secs, int *msec);
 extern int64_t util_get_time_as_ms_since_epoch ();
 extern time_t util_msec_to_sec (int64_t msec);
 
+extern bool util_is_localhost (char *host);
```
cristiarg:
```suggestion
extern bool util_is_localhost (const char *host);
```

cristiarg (approved):

wrlawodms (approved):

src/base/util_func.c
joohok: Done.

src/base/util_func.c
joohok: Done.

src/base/util_func.h
joohok: Done.

src/base/util_func.c
joohok: They are still used in other function.  Thanks.

PR #4440, opened 2023-06-20
[LETS-738] Add information about whether TS, PS exists to the node information

joohok:
http://jira.cubrid.org/browse/LETS-738

Describe here

Required to know whether TS and PS are currently running within one of the nodes in the cluster. 
This information will be used for failover (score calculation). 
In the process of defining the newHA node, the information within cub_master needs to be predefined.

Implementation
- Insert TS and PS variables into the data managed per node (HB_NODE_ENTRY)
- Initialize the variable values during initialization and re-initialization.

Note
- The part where these variables are updated will be done during the process of exchanging heartbeats

### DIFF
--- a/src/executables/master_heartbeat.c
+++ b/src/executables/master_heartbeat.c
@@ -2077,6 +2077,8 @@ hb_add_node_to_cluster (char *host_name, unsigned short priority)
       p->state = HB_NSTATE_UNKNOWN;
       p->score = 0;
       p->heartbeat_gap = 0;
+      p->has_transaction_server = false;
+      p->has_page_server = false;
       p->last_recv_hbtime.tv_sec = 0;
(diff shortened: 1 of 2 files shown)


hornetmj (approved):

wrlawodms (approved):
Sorry for the late review.

I approve it, but I think this PR is too small. Actually, I'm not sure these flags are necessary with this PR. It'd better to combine declaration and their usage into one issue if the declaration is not that huge. :)

joohok:
> Sorry for the late review.
> 
> I approve it, but I think this PR is too small. Actually, I'm not sure these flags are necessary with this PR. It'd better to combine declaration and their usage into one issue if the declaration is not that huge. :)

After creating this issue, I regretted not making the change in the place where the heartbeat is modified (where this flag is used)... Thank you for your  opinion 🥲

PR #4442, opened 2023-06-21
[CBRD-24862] Missing column name extraction of scalar subquery when using extended DBLINK

ctshim:
http://jira.cubrid.org/browse/CBRD-24862

* There was a problem of not getting the list of columns to be obtained from the remote table in the "inline view" and "connect by" statements.

`create table t1(id int, v1 int);
create table t2(id int, v2 int);
insert into t1 values(1, 1), (2,2), (3,10);
insert into t2 values (1, 10), (2, 100);
create server localsrv(host='localhost', port=33000, dbname=demodb, user=dba);

select (select id from t1@localsrv where v1=v2)   from t2@localsrv;
select (select id from t1@localsrv where v1=x.v2) from t2@localsrv x;`

`CREATE TABLE tree(ID INT, MgrID INT, Name VARCHAR(32), BirthYear INT);
INSERT INTO tree VALUES (1,NULL,'Kim', 1963);
INSERT INTO tree VALUES (2,NULL,'Moy', 1958);

SELECT name FROM tree@localsrv  CONNECT BY PRIOR id=mgrid ORDER BY id;`

### DIFF
--- a/src/parser/name_resolution.c
+++ b/src/parser/name_resolution.c
@@ -11447,19 +11447,16 @@ pt_get_column_name_pre (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int 
   S_LINK_COLUMNS *plkcol = (S_LINK_COLUMNS *) arg;
   PT_NODE *name = NULL;
 
-  if (node->node_type == PT_SELECT)
+  switch (node->node_type)
     {
-      ;				//*continue_walk = PT_STOP_WALK;
-    }
-  else if (node->node_type == PT_DOT_)
-    {				// case: tbl.col      
+    case PT_DOT_:
       check_for_already_exists
 	(parser, plkcol, node->info.dot.arg1->info.name.original, node->info.dot.arg2->info.name.original);
 
       *continue_walk = PT_LIST_WALK;
-    }
-  else if (node->node_type == PT_NAME)
-    {
+      break;
+
+    case PT_NAME:
       if (node->type_enum == PT_TYPE_STAR)
 	{			// case:  tbl.*
 	  check_for_already_exists (parser, plkcol, node->info.name.original, NULL);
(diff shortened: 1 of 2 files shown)


src/parser/name_resolution.c
```diff
@@ -11449,7 +11449,7 @@ pt_get_column_name_pre (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int
 
   if (node->node_type == PT_SELECT)
     {
-      *continue_walk = PT_STOP_WALK;
+      ;				//*continue_walk = PT_STOP_WALK;
```
hgryoo: Why is it commented out?

src/parser/name_resolution.c
```diff
@@ -11484,26 +11484,62 @@ pt_get_column_name_pre (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int
...
 }
 
+#if 0
 static PT_NODE *
 pt_get_column_name_post (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int *continue_walk)
 {
   *continue_walk = PT_CONTINUE_WALK;
   return node;
 }
+#endif
```
hgryoo: how about removing this function if it is not used.

ctshim: OK,  I will do it.
Thanks.

src/parser/name_resolution.c
ctshim: select (select id from t1@localsrv where v1=v2) from t2@localsrv;
In a query like the one in the example above, the select list could have a new select statement, so it needed to continue uninterrupted.

In this case, the if statement is empty, so let's clean it up.

hgryoo (approved):

hyunikn (approved):

beyondykk9 (approved):

PR #4443, opened 2023-06-21
[CBRD-24858] Create structures to be used for network communication

Rudeus:
http://jira.cubrid.org/browse/CBRD-24858

Add structures to be used for memmon utility / Memory Monitoring Manager in new header memory_monitoring_common.h

The added structures are used to store the memory usage information tracked by the server.

### DIFF
--- a/cubrid/CMakeLists.txt
+++ b/cubrid/CMakeLists.txt
@@ -150,12 +150,12 @@ set (BASE_HEADERS
   ${BASE_DIR}/mem_block.hpp
   ${BASE_DIR}/memory_reference_store.hpp
   ${BASE_DIR}/memory_private_allocator.hpp
+  ${BASE_DIR}/memory_monitor_common.h
   ${BASE_DIR}/msgcat_set_log.hpp
   ${BASE_DIR}/packable_object.hpp
   ${BASE_DIR}/packer.hpp
   ${BASE_DIR}/perf.hpp
   ${BASE_DIR}/perf_def.hpp
-  ${BASE_DIR}/memory_monitor_common.h
   ${BASE_DIR}/pinning.hpp
   ${BASE_DIR}/pinnable_buffer.hpp
   ${BASE_DIR}/porting_inline.hpp
--- a/cubrid/CMakeLists.txt
+++ b/cubrid/CMakeLists.txt
@@ -155,6 +155,7 @@ set (BASE_HEADERS
   ${BASE_DIR}/packer.hpp
   ${BASE_DIR}/perf.hpp
   ${BASE_DIR}/perf_def.hpp
+	${BASE_DIR}/memory_monitor_common.hpp
   ${BASE_DIR}/pinning.hpp
   ${BASE_DIR}/pinnable_buffer.hpp
   ${BASE_DIR}/porting_inline.hpp
--- a/cubrid/CMakeLists.txt
+++ b/cubrid/CMakeLists.txt
@@ -155,7 +155,7 @@ set (BASE_HEADERS
   ${BASE_DIR}/packer.hpp
   ${BASE_DIR}/perf.hpp
   ${BASE_DIR}/perf_def.hpp
-  ${BASE_DIR}/memory_monitor_common.hpp
+  ${BASE_DIR}/memory_monitor_common.h
   ${BASE_DIR}/pinning.hpp
   ${BASE_DIR}/pinnable_buffer.hpp
   ${BASE_DIR}/porting_inline.hpp
--- a/cubrid/CMakeLists.txt
+++ b/cubrid/CMakeLists.txt
@@ -155,7 +155,7 @@ set (BASE_HEADERS
   ${BASE_DIR}/packer.hpp
   ${BASE_DIR}/perf.hpp
   ${BASE_DIR}/perf_def.hpp
-	${BASE_DIR}/memory_monitor_common.hpp
+  ${BASE_DIR}/memory_monitor_common.hpp
   ${BASE_DIR}/pinning.hpp
   ${BASE_DIR}/pinnable_buffer.hpp
   ${BASE_DIR}/porting_inline.hpp
--- a/src/base/memory_monitor_common.h
+++ b/src/base/memory_monitor_common.h
@@ -1,5 +1,5 @@
 /*
- * Copyright 2008 Search Solution Corporation
+ *
  * Copyright 2016 CUBRID Corporation
  *
  *  Licensed under the Apache License, Version 2.0 (the "License");
--- a/src/base/memory_monitor_common.h
+++ b/src/base/memory_monitor_common.h
@@ -17,7 +17,7 @@
  */
 
 /*
- * memory_monitor_common.hpp - common structures for cubrid memory monitoring module
+ * memory_monitor_common.hpp - common structures for memory monitoring module
  */
 
 #ifndef _MEMORY_MONITOR_COMMON_H_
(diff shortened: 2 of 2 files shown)


hornetmj:
@Rudeus The header file newly introduced should be added to the cs/CMakeLists.txt and cubrid/CMakeLists.txt.
- cs/CMakeLists.txt - for the libcubridcs.so library, for the client
- cubrid/CMakeLists.txt - for the libcubrid.so library, for the server

src/base/memory_monitor_common.h
```diff
@@ -0,0 +1,79 @@
...
+  MEMMON_SUBCOMP_INFO *subcomp_info;
+} MEMMON_COMP_INFO;
+
+typedef struct memmon_module_info
+{
+  MEMMON_SERVER_INFO server_info;
+  char *name;
+  MEMMON_MEM_STAT stat;
+  uint32_t num_comp;
+  MEMMON_COMP_INFO **comp_info;
```
hornetmj: Is the double-pointer type intended? I just ask because the single-pointer is used for the subcomp_info in the MEMMON_COMP_INFO.

src/base/memory_monitor_common.h
```diff
@@ -0,0 +1,79 @@
...
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+
+/*
+ * memory_monitor_common.hpp - common structures for cubrid memory monitoring module
```
hornetmj:
```suggestion
 * memory_monitor_common.hpp - common structures for memory monitoring module
```

hornetmj: The previous comment is Nit :D

I think we don't mention 'CUBRID' prefix because everybody knows it.

src/base/memory_monitor_common.h
Rudeus: Since the variable MEMMON_SUBCOMP_INFO subcomp_info[1] in MEMMON_COMP_INFO was not a pointer, it introduced variability in allocating memory for that structure. However, since subcomp_info has been changed to MEMMON_SUBCOMP_INFO* type, I will modify it to a single pointer.

src/base/memory_monitor_common.h
Rudeus: The change has been made accordingly :)

cubrid/CMakeLists.txt
```diff
@@ -155,6 +155,7 @@ set (BASE_HEADERS
   ${BASE_DIR}/packer.hpp
   ${BASE_DIR}/perf.hpp
   ${BASE_DIR}/perf_def.hpp
+  ${BASE_DIR}/memory_monitor_common.hpp
```
joohok: I think it need to be sorted alphabetically

Rudeus:
@hornetmj 
There was an error in the past when adding cs/CMakeLists.txt during prototype work. As a result, I only added it to cubrid/CMakeLists.txt.

src/base/memory_monitor_common.h
```diff
@@ -0,0 +1,79 @@
+/*
+ * Copyright 2008 Search Solution Corporation
```
joohok:
```suggestion
 *
```
No need to add this license for newly created files

src/base/memory_monitor_common.h
```diff
@@ -0,0 +1,79 @@
...
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+
+/*
+ * memory_monitor_common.hpp - common structures for memory monitoring module
```
joohok:
```suggestion
 * memory_monitor_common.h - common structures for memory monitoring module
```

src/base/memory_monitor_common.h
```diff
@@ -0,0 +1,79 @@
...
+{
+  uint64_t init_stat;
+  uint64_t cur_stat;
+  uint64_t peak_stat;
+  uint32_t expand_count;
+} MEMMON_MEM_STAT;
+
+typedef struct memmon_server_info
+{
+  char *name;
```
joohok: I'm not sure how this string will be used, but typically, it is used with a char array. How about using it like 'char name[DB_MAX_IDENTIFIER_LENGTH]'?

src/base/memory_monitor_common.h
```diff
@@ -0,0 +1,79 @@
...
+ */
+
+/*
+ * memory_monitor_common.h - common structures for memory monitoring module
+ */
+
+#ifndef _MEMORY_MONITOR_COMMON_H_
+#define _MEMORY_MONITOR_COMMON_H_
+
+#include <string>
```
wrlawodms: seems unused.

cubrid/CMakeLists.txt
Rudeus: I modified it with new commit acd441a

src/base/memory_monitor_common.h
Rudeus: Changed *name -> name[DB_MAX_IDENTIFIER_LENGTH]

joohok (approved):

wrlawodms (commented):
Are they used client-side or server-side? or both?

Rudeus:
> Are they used client-side or server-side? or both?

they are used to both client-side and server-side

src/base/memory_monitor_common.h
Rudeus: Deleted at c942924

wrlawodms:
> they are used to both client-side and server-side

@Rudeus Then, I think they should be also added in cs/CMakefiles.txt, or we probably can't use them on the client side. Isn't it?

Rudeus:
> > they are used to both client-side and server-side
> 
> @Rudeus Then, I think they should be also added in cs/CMakefiles.txt, or we probably can't use them on the client side. Isn't it?

It seems that adding it only to cubrid/CMakeLists.txt was sufficient for both the client and server to work correctly. If any issues arise during further implementation that require adding it to cs/CMakeLists.txt, I will analyze the problem and add it at that time.

wrlawodms:
@Rudeus OK, I might be wrong. Let's see.

wrlawodms (approved):

hornetmj:
@Rudeus @wrlawodms

FYI

cubrid utility does not link to libcubridcs.so. And cub_admin utility loads libcubridcs.so dynamically when the utility_type in the ua_Utility_Map is set to CS_ONLY.

```
$ ldd cub_cas | grep libcubridcs
	libcubridcs.so.11.3 => /home/hornetmj/work/cubrid/build_x86_64_debug/_install/CUBRID/lib/libcubridcs.so.11.3 (0x00007fa1362c8000)

$ ldd cub_master | grep libcubridcs
	libcubridcs.so.11.3 => /home/hornetmj/work/cubrid/build_x86_64_debug/_install/CUBRID/lib/libcubridcs.so.11.3 (0x00007f1942f4a000)

$ ldd cubrid | grep libcubridcs

$ ldd cub_admin | grep libcubridcs
```

hornetmj (approved):

PR #4444, opened 2023-06-22
[CBRD-24864] POC: extend query cache to CTE

beyondykk9:
http://jira.cubrid.org/browse/CBRD-24864

Before applying the result cache to a subquery, proceed with a POC that applies it to the CTE query. The query cache hint will be extended to CTE queries.

First of all, we will check whether the CTE query with hints uses the result cache well, and whether the result cache is successfully added.

Based on the results of this POC, we plan to decide what to do with the next POC.
In particular, we must ensure that there are no problems with transaction processing in the process of handling the subquery result cache.


beyondykk9:
This issue will be continued at https://github.com/CUBRID/cubrid/pull/4727

PR #4445, opened 2023-06-22
[LETS-758] Disable replication logs generated for applylogdb

joohok:
http://jira.cubrid.org/browse/LETS-758

Purpose

In newHA implementation, there is no applylogdb to apply replication log records so that replication logs are no longer required.

Implementation

- comment out the `log_does_allow_replication ()`
- `log_does_allow_replication()` is used in the context of generating replication logs when performing DML and DDL statements.

### DIFF
--- a/src/transaction/log_comm.c
+++ b/src/transaction/log_comm.c
@@ -269,6 +269,13 @@ log_dump_log_info (const char *logname_info, bool also_stdout, const char *fmt, 
 bool
 log_does_allow_replication (void)
 {
+  return false;
+
+#if 0
+/* TODO - newHA
+ * if copylogdb and applylogdb is re-used for HA, then this commented out code should be enabled.
+ */
+
 #if defined(WINDOWS) || defined(SA_MODE)
   return false;
 
@@ -309,5 +316,6 @@ log_does_allow_replication (void)
 #else /* SERVER_MODE */
 
   return false;
+#endif
 #endif
 }
--- a/src/transaction/replication.c
+++ b/src/transaction/replication.c
@@ -299,6 +299,8 @@ repl_log_insert (THREAD_ENTRY * thread_p, const OID * class_oid, const OID * ins
   char *ptr;
   int error = NO_ERROR, strlen;
 
+  assert (log_does_allow_replication () == true);
+
   tran_index = LOG_FIND_THREAD_TRAN_INDEX (thread_p);
   tdes = LOG_FIND_TDES (tran_index);
   if (tdes == NULL)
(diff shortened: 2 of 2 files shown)


hornetmj (approved):
LGTM.

src/transaction/log_comm.c
```diff
@@ -269,6 +269,13 @@ log_dump_log_info (const char *logname_info, bool also_stdout, const char *fmt,
 bool
 log_does_allow_replication (void)
 {
+  return false;
+
+#if 0
+/* TODO - newHA
```
wrlawodms, line 275: Why is it TODO? Is this TODO for the NewHA?

wrlawodms (commented):
OK, forcing `log_does_allow_replication()` to set to false, the replication log records are not generated. But some pre-requisite jobs are done for them and they also should be removed. Am I right? An issue has to be created if it will be dealt with another issue.
    - For example, in do_statement(), locator_all_flush() and db_set_suppress_repl_on_transaction() are done.
    - Another example could be creating the replication log volume. 
    - and more, I'm not sure, which should be investigated.

Please let me have your plan.

joohok:
> - For example, in do_statement(), locator_all_flush() and db_set_suppress_repl_on_transaction() are done. 

OK, I will make an issue, and treat it in that issue. Those should be removed as they are unnecessary codes, because they make unnecessary network I/O costs. 

>- Another example could be creating the replication log volume. - and more, I'm not sure, which should be investigated.

Well, when the replication log is generated, it is stored in the log volume. The replica log volume is only created when the `copylogdb` is running, and it retrieves logs from other nodes. Therefore, it seems unnecessary to deal with the creation of the replica log volume.
 
> Please let me have your plan.

Currently, I am trying to block only unnecessary processes, such as running copylogdb, logging, etc. (copylogdb, applylogdb , replication log generation). In the future, it may be possible to provide Disaster Recovery (DR) functionality using replication logs, copylogdb, and applylogdb, so for now, I want to block them minimally.

In the part where replication logs are generated, it seems appropriate to remove the unnecessary **preliminary tasks** you mentioned. I will create a separate issue to address this. It would be best to find all the places where replication logs can be generated and remove the unnecessary part at once. 

Thanks!

src/transaction/log_comm.c
joohok: I had to leave it as a TODO because the DR concept was not decided. 
This is because generating this replication log may become necessary again for DR (what CTO said). 
I left it as a TODO, considering it part of completing the newHA, which includes implementing the DR concept

wrlawodms (approved):
Yep, Please create and share the issue here before merging. Thank you.

joohok:
created http://jira.cubrid.org/browse/LETS-761 , for removing unnecessary codes when generating replication log records

PR #4448, opened 2023-06-26
[CBRD-24870] Increase the number of sampling pages when collecting btree statistics

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-24870

- Change the number related to sampling statistics for b-tree.
```
STATS_SAMPLING_THRESHOLD : 50 to 1000. If this value is not exceeded, the entire page is full-scanned.
STATS_SAMPLING_LEAFS_MAX : 7 to 500. The number of samples extracted when collecting statistics on sampling.
```
Since the number of leaf pages in the b-tree is smaller than the number of pages in the heap, the number of b-tree samples is set to 500. The number of sampling pages in the heap is 1000. These values will be adjusted later in optimizer enhancement.
- Change sample weights from int to double for accurate calculation.
- Delete routine that do not update statistics on the same number of pages to prevent the problem that statistics are not updated.
- Delete routine for acceptance probability. Since the number of sampling pages is increased, it becomes unnecessary.


src/storage/btree.c
```diff
@@ -14603,12 +14588,11 @@ btree_find_AR_sampling_leaf (THREAD_ENTRY * thread_p, BTID * btid, VPID * pg_vpi
   int est_page_size, free_space;
   int key_cnt = 0;
   int root_level = 0, depth = 0;
-  double prob = 1.0;		/* Acceptance probability */
 
   assert (stat_info_p != NULL);
   assert (found_p != NULL);
 
-  *found_p = false;		/* init */
+  *found_p = true;		/* init */
```
youngjinj, line 14595: It is trivial.

`btree_find_AR_sampling_leaf` is only called by `btree_get_stats_with_AR_sampling`.
How about not using `found` in `btree_get_stats_with_AR_sampling` and `btree_find_AR_sampling_leaf`?

youngjinj (approved):

xmilex-git (approved):

src/storage/btree.c
shparkcubrid: Okay. I'll fix it on next PR.

PR #4449, opened 2023-06-26
[LETS-756] Try to connect to disconnected page servers periodically

wrlawodms:
http://jira.cubrid.org/browse/LETS-756

Add a ps_connector in tran_server, which tries to reconnect to all disconnected page servers. 
- It tries to all `connection_handler` that are in `IDLE` state. As described in a comment, later on, it is going to connect to only PSes that are ready to connect by an upper layer.
- The internal thread is implemented with cubrid daemon since it uses `er_set` internally.
- It's started on `tran_server::boot()` and terminated in `tran_server::disconnect_all_page_servers ()`.

Put `connect_to_page_server` into connection_handler as `connect`
- Mainly, because the creating channel has to be done inside the `m_state` lock scope.
- And it seems more make sense with the current codebase after several refactoring.

### DIFF
--- a/src/server/tran_server.cpp
+++ b/src/server/tran_server.cpp
@@ -324,7 +324,7 @@ tran_server::get_boot_info_from_page_server ()
 int
 tran_server::connection_handler::connect (const cubcomm::node &node)
 {
-  auto ps_conn_error_lambda = [&node, this] ()
+  auto ps_conn_error_lambda = [&node, this] (const std::lock_guard<std::shared_mutex> &)
   {
     m_state = state::IDLE;
 
@@ -349,22 +349,22 @@ tran_server::connection_handler::connect (const cubcomm::node &node)
 				   CMD_SERVER_SERVER_CONNECT);
   if (comm_error_code != css_error_code::NO_ERRORS)
     {
-      return ps_conn_error_lambda ();
+      return ps_conn_error_lambda (lockg_state);
     }
 
   if (srv_chn.send_int (static_cast<int> (m_ts.m_conn_type)) != NO_ERRORS)
     {
-      return ps_conn_error_lambda ();
+      return ps_conn_error_lambda (lockg_state);
     }
 
   int returned_code;
   if (srv_chn.recv_int (returned_code) != css_error_code::NO_ERRORS)
     {
-      return ps_conn_error_lambda ();
+      return ps_conn_error_lambda (lockg_state);
     }
   if (returned_code != static_cast<int> (m_ts.m_conn_type))
     {
-      return ps_conn_error_lambda ();
+      return ps_conn_error_lambda (lockg_state);
     }
 
   // NOTE: only the base class part (cubcomm::channel) of a cubcomm::server_channel instance is
--- a/src/server/tran_server.cpp
+++ b/src/server/tran_server.cpp
@@ -540,30 +540,59 @@ tran_server::connection_handler::is_disconnecting () const
 
 tran_server::ps_connector::ps_connector (tran_server &ts)
   : m_ts { ts }
+  , m_terminate { false }
 {
-  m_thread = std::thread (&tran_server::ps_connector::connect_loop, std::ref (*this));
 }
 
 tran_server::ps_connector::~ps_connector ()
 {
+  assert (m_terminate.load () == true);
 }
 
 void
 tran_server::ps_connector::start ()
 {
+  m_thread = std::thread (&tran_server::ps_connector::connect_loop, std::ref (*this));
 }
 
 void
 tran_server::ps_connector::terminate ()
 {
+  m_terminate.store (true);
+
+  if (m_thread.joinable ())
+    {
+      m_thread.join ();
+    }
+  else
+    {
+      assert (false);
+    }
 }
 
 void
 tran_server::ps_connector::connect_loop ()
 {
+  constexpr std::chrono::seconds one_sec { 1 };
+
+  /* Assume that they stores PS information in the same order */
+  assert (m_ts.m_connection_list.size () == m_ts.m_page_server_conn_vec.size());
+
+  while (true)
+    {
+      for (int i=0; i < m_ts.m_page_server_conn_vec.size (); i++)
+	{
+	  // 흠 state 상태를 체크하고, connect를 수행하기까지의 lock을 잡아야 하지 않나..? 밖에서 어케 잡지
+	  // connect 요청하고 내부에서 에러처리하는 방식으로 해야하나
+	  if (m_ts.m_page_server_conn_vec[i].get_state () ==
+	      tran_server::connection_handler::state::OPEN))
+	    {
+	    }
+
+	}
+      this_thread::sleep_for (one_sec);
+    }
 }
-
-
 
 void
 assert_is_tran_server ()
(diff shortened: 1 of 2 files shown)


src/server/tran_server.cpp
```diff
@@ -609,6 +615,70 @@ tran_server::connection_handler::is_connected ()
...
+tran_server::ps_connector::terminate ()
+{
+  m_terminate.store (true);
+  cubthread::get_manager ()->destroy_daemon (m_daemon);
+}
+
+void
+tran_server::ps_connector::connect_if_idle (cubthread::entry &)
+{
+  /* Assume that they stores PS information in the same order */
```
wrlawodms: I'll create a refactoring PR putting `m_node` into `connection_handler` as the following PR.

src/server/tran_server.cpp
```diff
@@ -609,6 +615,74 @@ tran_server::connection_handler::is_connected ()
...
+   * TODO: combine two vectors */
+  assert (m_ts.m_connection_list.size () == m_ts.m_page_server_conn_vec.size());
+
+  for (size_t i = 0; i < m_ts.m_page_server_conn_vec.size (); i++)
+    {
+      if (m_ts.m_page_server_conn_vec[i]->is_idle ())
+	{
+	  /*
+	   * TODO It can be too verbose now since it always complain to fail to connect when a PS has been stopped.
+	   * Later on, this job is going to be tirggered by a cluster manager or cub_mster when a PS is ready to connect.
```
hornetmj:
```suggestion
	   * Later on, this job is going to be triggered by a cluster manager or cub_master when a PS is ready to connect.
```

src/server/tran_server.cpp
```diff
@@ -609,6 +615,74 @@ tran_server::connection_handler::is_connected ()
...
+  if (m_terminate.exchange (true) == false)
+    {
+      cubthread::get_manager ()->destroy_daemon (m_daemon);
+    }
+}
+
+void
+tran_server::ps_connector::try_connect_to_all_ps (cubthread::entry &)
+{
+  /* Assume that they stores PS information in the same order.
```
hornetmj:
```suggestion
  /* Assume that they store PS information in the same order.
```

hornetmj (approved):
LGTM

joohok (approved):

src/server/tran_server.cpp
```diff
@@ -318,19 +322,26 @@ tran_server::get_boot_info_from_page_server ()
 }
 
 int
-tran_server::connect_to_page_server (connection_handler &conn_handler, const cubcomm::node &node, const char *db_name)
+tran_server::connection_handler::connect (const cubcomm::node &node)
 {
-  auto ps_conn_error_lambda = [&node] ()
+  auto ps_conn_error_lambda = [&node, this] ()
```
To formally ensure that the lock is in effect.
```suggestion
  auto ps_conn_error_lambda = [&node, this] (const std::lock_guard<std::shared_mutex>&)
```
I know it makes the code ugly. :)

src/server/tran_server.cpp
```diff
@@ -341,7 +352,7 @@ tran_server::connect_to_page_server (connection_handler &conn_handler, const cub
       return ps_conn_error_lambda ();
```
```suggestion
      return ps_conn_error_lambda (lockg_state);
```

src/server/tran_server.cpp
```diff
@@ -341,7 +352,7 @@ tran_server::connect_to_page_server (connection_handler &conn_handler, const cub
...
     }
 
-  if (srv_chn.send_int (static_cast<int> (m_conn_type)) != NO_ERRORS)
+  if (srv_chn.send_int (static_cast<int> (m_ts.m_conn_type)) != NO_ERRORS)
     {
       return ps_conn_error_lambda ();
```
```suggestion
      return ps_conn_error_lambda (lockg_state);
```

src/server/tran_server.cpp
```diff
@@ -351,14 +362,18 @@ tran_server::connect_to_page_server (connection_handler &conn_handler, const cub
...
```
```suggestion
      return ps_conn_error_lambda (lockg_state);
```

src/server/tran_server.cpp
```diff
@@ -351,14 +362,18 @@ tran_server::connect_to_page_server (connection_handler &conn_handler, const cub
...
     }
-  if (returned_code != static_cast<int> (m_conn_type))
+  if (returned_code != static_cast<int> (m_ts.m_conn_type))
     {
       return ps_conn_error_lambda ();
```
```suggestion
      return ps_conn_error_lambda (lockg_state);
```

src/server/tran_server.cpp
```diff
@@ -443,25 +462,12 @@ tran_server::connection_handler::set_connection (cubcomm::channel &&chn)
   // Transaction server will use message specific error handlers.
   // Implementation will assert that an error handler is present if needed.
 
-  auto lockg_state = std::lock_guard<std::shared_mutex> { m_state_mtx };
+  assert (m_conn == nullptr);
+  m_conn.reset (new page_server_conn_t (std::move (chn), get_request_handlers (), tran_to_page_request::RESPOND,
+					page_to_tran_request::RESPOND, RESPONSE_PARTITIONING_SIZE, std::move (no_transaction_handler)));
+  assert (m_conn != nullptr);
```
cristiarg: Kind of pointless assert.
```suggestion
```
I know it existed before, but for what?

src/server/tran_server.cpp
```diff
@@ -435,6 +453,7 @@ tran_server::uses_remote_storage () const
 void
 tran_server::connection_handler::set_connection (cubcomm::channel &&chn)
 {
+  auto lockg_conn = std::lock_guard<std::shared_mutex> { m_conn_mtx };
```
cristiarg: What's the point to move the lock here? Should remain just before the `assert`.

src/server/tran_server.cpp
```diff
@@ -609,6 +615,74 @@ tran_server::connection_handler::is_connected ()
...
+}
+
+void
+tran_server::ps_connector::try_connect_to_all_ps (cubthread::entry &)
+{
+  /* Assume that they store PS information in the same order.
+   * TODO: combine two vectors */
+  assert (m_ts.m_connection_list.size () == m_ts.m_page_server_conn_vec.size());
+
+  for (size_t i = 0; i < m_ts.m_page_server_conn_vec.size (); i++)
```
cristiarg, line 668: Nit:
```suggestion
  for (auto & connection_handler : m_ts.m_page_server_conn_vec)
```
something like that.

cristiarg (approved):

src/server/tran_server.cpp
wrlawodms: We need the index to access `m_ts.m_connection_list[i]`. It will be removed anyway in the following PR for combining `m_ts.m_connection_list` and `m_ts.m_page_server_conn_vec`.

src/server/tran_server.cpp
wrlawodms: I thought that this location of lock says that the whole job in this function needs the lock acquired, and the lock scope is the same.

src/server/tran_server.cpp
wrlawodms: Seems better to me. Thanks.

src/server/tran_server.cpp
wrlawodms: I don't know either. It even doesn't throw an exception. Let's remove it.

src/server/tran_server.cpp
cristiarg: fair enough

wrlawodms: modified as you said, though. :)))

PR #4451, opened 2023-06-27
[CBRD-24684] revised: divide operation result into NUMERIC type

beyondykk9:
http://jira.cubrid.org/browse/CBRD-24684

Under the parameter ORACLE_DIVIDE_STYLE on

```
csql> select rownum from db_class limit 5/4 offset 0;
                rownum
======================
                     1

csql> select rownum from db_class limit 5/4 offset 1/2;
                rownum
======================
                     1
                     2
```

limit with offset represents "\<limit\> from \<offset\> to \<offset\> + \<limit\>" and the query is rewritten as below:

select rownum from db_class limit 5/4 offset 1/2;
==> rewritten
select rownum from db_class where inst_num() > offset and inst_num() <= offset + limit;

The above two query results should be the same as in Oracle.

### DIFF
--- a/src/parser/parser_support.c
+++ b/src/parser/parser_support.c
@@ -4456,7 +4456,9 @@ pt_limit_to_numbering_expr (PARSER_CONTEXT * parser, PT_NODE * limit, PT_OP_TYPE
 
       sum->data_type->type_enum = PT_TYPE_NUMERIC;
       sum->data_type->info.data_type.precision = DB_MAX_NUMERIC_PRECISION;
-      sum->data_type->info.data_type.dec_precision = 0;
+
+      /* It is necessary to prepare for the division operation in advance.*/
+      sum->data_type->info.data_type.dec_precision = DB_DEFAULT_NUMERIC_DIVISION_SCALE; /* 9 */
 
       sum->info.expr.arg1 = parser_copy_tree (parser, limit);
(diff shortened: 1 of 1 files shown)


ctshim (approved):

hyunikn (approved):

hgryoo (approved):

src/parser/parser_support.c
```diff
@@ -4458,6 +4460,12 @@ pt_limit_to_numbering_expr (PARSER_CONTEXT * parser, PT_NODE * limit, PT_OP_TYPE
       sum->data_type->info.data_type.precision = DB_MAX_NUMERIC_PRECISION;
       sum->data_type->info.data_type.dec_precision = 0;
 
+      if (oracle_style_divide)
+	{
+	  /* It is necessary to prepare for the division operation in advance. */
+	  sum->data_type->info.data_type.dec_precision = DB_DEFAULT_NUMERIC_DIVISION_SCALE;	/* 9 */
+	}
+
```
hgryoo, line 4468:
```suggestion
      sum->data_type->info.data_type.dec_precision = oracle_style_divide ? DB_DEFAULT_NUMERIC_DIVISION_SCALE : 0;
```

beyondykk9: I'd like to keep for readability

PR #4452, opened 2023-06-27
[LETS-739] Define heartbeat message

joohok:
http://jira.cubrid.org/browse/LETS-739

Purpose

The heartbeat message included the aliveness of each node's TS and PS.

### DIFF
--- a/src/executables/master_heartbeat.c
+++ b/src/executables/master_heartbeat.c
@@ -1633,13 +1633,15 @@ hb_cluster_send_heartbeat_internal (struct sockaddr_in *saddr, socklen_t saddr_l
 {
   HBP_HEADER *hbp_header;
   char buffer[HB_BUFFER_SZ], *p;
-  size_t hb_len;
+  const size_t hb_body_len = OR_INT_SIZE * 3;	//heartbeat message body length
+  const size_t hb_len = sizeof (HBP_HEADER) + hb_body_len;	//heartbeat message length including header and body
+
   int send_len;
 
   memset ((void *) buffer, 0, sizeof (buffer));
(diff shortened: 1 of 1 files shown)


wrlawodms (approved):
How did you test it works well?

joohok:
> How did you test it works well?

I run a cluster consists of 2 nodes, and see the Master error log to check if there is an error. 

I also tried debugging manually and also added temporary debugging messages to check.
Since the heartbeat is sent/received every second, I decided not to include the debugging messages as they would have made it too verbose.

wrlawodms:
@joohok That makes sense. Thank you.

src/executables/master_heartbeat.c
```diff
@@ -1633,22 +1633,24 @@ hb_cluster_send_heartbeat_internal (struct sockaddr_in *saddr, socklen_t saddr_l
...
     {
       return error_code;
     }
 
   p = (char *) (hbp_header + 1);
   p = or_pack_int (p, hb_Cluster->state);
-
-  hb_len = sizeof (HBP_HEADER) + OR_INT_SIZE;
+  p = or_pack_int (p, hb_Cluster->myself->is_tran_server_alive);
+  p = or_pack_int (p, hb_Cluster->myself->is_page_server_alive);
```
hornetmj, line 1653: Can you reduce the size of the message to transfer to the network? This kind of information of all is enough 32-byte. It would be good to handle later for optimization.

hornetmj (approved):

src/executables/master_heartbeat.c
joohok: It is possible to fit two flags in a four-byte integer, but I will optimize it later.

PR #4455, opened 2023-06-27
[LETS-697] Handle the abnormal disconnection of a page server.

wrlawodms:
http://jira.cubrid.org/browse/LETS-697

An error during communication is not spotted immediately. It's noticed when a request is pushed to the connection. The disconnection can be found when
- (A) push_request() : it's reported through the error handler given when setting the connection (`connection_handler::set_connection`).  
- send_receive()      : 
  - (B) while pushing a request: it's reported through the error handler given in the `request_sync_client_server::send_recv`
  - (C) while getting its response: it's never reported now.
 
When the connection is disconnected, the push() fails, its internal socket gets invalid (`channel:send`), and throws the error. The error code, in this case, is `ERROR_ON_WRITE`, but I set it to `CONNECTION_CLOSED` when the connection gets invalid to let the outside know that it's not just a write error, but the socket has been closed. And in each case:
- (A): If the error code is `CONNECTION_CLOSED`, clean up the connection_handler.
- (B): The current error handler set the error code and make it's returned to `get_response()`. So, I leave it as it is and deal with the case it's returned in `connection_handler::send_receive`. It's because `request_sync_client_server::send_recv` doesn't know connection_handler.
- (C): If push() succeeded and the connection fails while waiting for its response, there is no way for the waiter to know it. The subsequent request will fail, clean the connection_handler, and, in turn, let the waiter know the connection_handler got wrong so it's closed. (http://jira.cubrid.org/browse/LETS-762). In this case, the default error code, `ERROR_ON_WRITE` is returned. This is a walkaround and this should be addressed with a timeout for send_receive or something in a separate issue.

### DIFF
--- a/src/server/tran_server.cpp
+++ b/src/server/tran_server.cpp
@@ -606,6 +606,18 @@ tran_server::connection_handler::send_receive (tran_to_page_request reqid, std::
   const css_error_code error_code = m_conn->send_recv (reqid, std::move (payload_in), payload_out);
   if (error_code != NO_ERRORS)
     {
+      if (!m_conn->is_underlying_channel_alive ())
+	{
+	  er_log_debug (ARG_FILE_LINE,
+			"sned_receive: an abnormal disconnection has been detected. error code: %d, channel id: %s.\n", error_code,
+			get_channel_id ().c_str ());
+
+	  slock_conn.unlock (); /* disconnect_async requires that m_conn_mtx and m_state_mtx are unlocked */
+
+	  constexpr auto with_disc_msg = false;
+	  disconnect_async (with_disc_msg);
+	}
+
       er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_CONN_PAGE_SERVER_CANNOT_BE_REACHED, 0);
       return ER_CONN_PAGE_SERVER_CANNOT_BE_REACHED;
     }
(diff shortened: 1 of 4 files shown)


joohok (approved):

hornetmj (approved):

src/server/tran_server.cpp
```diff
@@ -457,15 +457,31 @@ tran_server::connection_handler::set_connection (cubcomm::channel &&chn)
   // TODO: to reduce contention as much as possible, should be equal to the maximum number
   // of active transactions that the system allows (PRM_ID_CSS_MAX_CLIENTS) + 1
 
-  cubcomm::send_queue_error_handler no_transaction_handler { nullptr };
-  // Transaction server will use message specific error handlers.
-  // Implementation will assert that an error handler is present if needed.
+  cubcomm::send_queue_error_handler default_error_handler = [this]
```
The handler would be more visible in code if it were a function and passed with `std::bind`.

src/server/tran_server.cpp
```diff
@@ -457,15 +457,31 @@ tran_server::connection_handler::set_connection (cubcomm::channel &&chn)
...
+      (css_error_code error_code, bool &abort_further_processing)
+  {
+    // Remove the connection_handler if the internal socket is closed. It's been disconnected abnormally.
```
cristiarg: Nit. Just to make it explicit.

```suggestion
    abort_further_processing = false;
    // Remove the connection_handler if the internal socket is closed. It's been disconnected abnormally.
```

cristiarg (approved):

src/server/tran_server.cpp
wrlawodms: Not a Nit, I hope I understand you correctly: [2935cb9](https://github.com/CUBRID/cubrid/pull/4455/commits/2935cb9ec9431f868c7801351da7c73abe8ba706)

src/server/tran_server.cpp
wrlawodms: done

PR #4457, opened 2023-06-27
[LETS-501] Print the status of TS and PS when cubrid hb status is executed

joohok:
http://jira.cubrid.org/browse/LETS-501

Purpose

As the Page Server and Transaction Server operate on the same node, it is necessary to be able to output their status information when `cubrid hb status` command is executed.

### DIFF
--- a/src/executables/master_heartbeat.c
+++ b/src/executables/master_heartbeat.c
@@ -296,6 +296,10 @@ static HB_JOB_FUNC hb_resource_jobs[] = {
 	" HA-Process Info (master %d, state %s)\n"
 #define HA_SERVER_PROCESS_FORMAT_STRING  \
 	"   Server %s (pid %d, state %s)\n"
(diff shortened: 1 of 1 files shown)


src/executables/master_heartbeat.c
```diff
@@ -6051,7 +6055,7 @@ hb_get_process_info_string (char **str, bool verbose_yn)
   required_size += HB_NSTATE_STR_SZ;	/* length of node state */
   buf_size += required_size;
 
-  required_size = strlen (HA_APPLYLOG_PROCESS_FORMAT_STRING);
+  required_size = strlen (HA_TRAN_SERVER_PROCESS_FORMAT_STRING);
```
joohok, line 6056: This value is set based on the length of the longest FORMAT.

hornetmj (approved):

wrlawodms (approved):

PR #4462, opened 2023-06-29
[CBRD-24873] Set NDV extracted from heap as selectivity.

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-24873

The NDV extracted from the heap is used first, and then the NDV from the btree is used. If the precision of the char type of the column exceeds 4000, NDV is not extracted from the heap.

Number of object is counted using [select --+sampling_scan count(*) ...]


xmilex-git (approved):

src/optimizer/query_graph.c
```diff
@@ -5187,6 +5188,7 @@ qo_get_attr_info (QO_ENV * env, QO_SEGMENT * seg)
       cum_statsp->key_type = NULL;
       cum_statsp->pkeys_size = 0;
       cum_statsp->pkeys = NULL;
+      attr_infop->ndv = 0;
```
youngjinj, line 5191: What about initializing on the next line after malloc?

src/storage/statistics_sr.c
```diff
@@ -241,16 +241,8 @@ xstats_update_statistics (THREAD_ENTRY * thread_p, OID * class_id_p, bool with_f
...
-  if (estimated_nobjs == -1)
-    {
-      /* cannot get estimates from the heap, use old info */
-      assert (cls_info_p->ci_tot_objects >= 0);
-    }
-  else
-    {
-      cls_info_p->ci_tot_objects = estimated_nobjs;
-    }
+  /* use value from "select --+ sampling count(*) ..." */
```
youngjinj, line 244: It seems that `estimated_nobjs` is no longer necessary.

youngjinj (approved):

src/storage/statistics_sr.c
shparkcubrid: Okay. I'll fix it later.

PR #4464, opened 2023-06-30
[LETS-763] [Refactoring] Putting cubcomm::node into connection_handler in tran_server

wrlawodms:
http://jira.cubrid.org/browse/LETS-763

- connection_handler contains cubcomm::node.
- connection_handlers are created when parsing host_node.

### DIFF
--- a/src/server/tran_server.cpp
+++ b/src/server/tran_server.cpp
@@ -74,8 +74,7 @@ tran_server::create_connection_with_server_host (const std::string &host)
   m_ps_hostname = host.substr (0, col_pos);
   er_log_debug (ARG_FILE_LINE, "Page server hosts: %s port: %d\n", m_ps_hostname.c_str (), port);
 
-  cubcomm::node conn{port, m_ps_hostname};
-  m_connection_list.push_back (conn);
+  m_page_server_conn_vec.emplace_back (create_connection_handler (*this, {port, m_ps_hostname}));
 
   return NO_ERROR;
 }
@@ -236,7 +235,7 @@ tran_server::init_page_server_hosts ()
     }
 
   int exit_code = create_connections_with_server_hosts_config (hosts);
-  if (m_connection_list.empty ())
+  if (m_page_server_conn_vec.empty ())
     {
       // no valid hosts
       int exit_code = ER_HOST_PORT_PARAMETER;
@@ -254,11 +253,9 @@ tran_server::init_page_server_hosts ()
   //
   int valid_connection_count = 0;
   bool failed_conn = false;
-  for (const cubcomm::node &node : m_connection_list)
+  for (const auto &conn : m_page_server_conn_vec)
     {
-      /* create a empty connection_handler specialized for each tran serve type */
-      m_page_server_conn_vec.emplace_back (create_connection_handler (*this));
-      exit_code = m_page_server_conn_vec.back ()->connect (node);
+      exit_code = conn->connect ();
       if (exit_code == NO_ERROR)
 	{
 	  ++valid_connection_count;
@@ -266,7 +263,6 @@ tran_server::init_page_server_hosts ()
       else
 	{
 	  failed_conn = true;
-	  er_log_debug (ARG_FILE_LINE, "Failed to connect to host: %s port: %d\n", node.get_host ().c_str (), node.get_port ());
 	}
     }
   reset_main_connection ();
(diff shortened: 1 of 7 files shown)


src/server/tran_server.cpp
```diff
@@ -254,19 +253,16 @@ tran_server::init_page_server_hosts ()
...
-      exit_code = m_page_server_conn_vec.back ()->connect (node);
+      exit_code = conn->connect ();
       if (exit_code == NO_ERROR)
 	{
 	  ++valid_connection_count;
 	}
       else
 	{
 	  failed_conn = true;
-	  er_log_debug (ARG_FILE_LINE, "Failed to connect to host: %s port: %d\n", node.get_host ().c_str (), node.get_port ());
```
wrlawodms, line 269: tran_server doesn't know the node, and a similar log is written in `tran_server::connection_handler::connect`.

joohok (approved):

src/server/tran_server.hpp
```diff
@@ -122,8 +122,9 @@ class tran_server
 	virtual log_lsa get_saved_lsa () const = 0; // used in active_tran_server
 
       protected:
-	connection_handler (tran_server &ts)
+	connection_handler (tran_server &ts, cubcomm::node &&node)
 	  : m_ts { ts }
+	  , m_node { std::move (node) }
```
cristiarg, line 127: Nit.
If a `cubcomm::node` is mostly moved around and not stored in a vector anymore, I suppose some of its copy ctor/operators can be deleted.

src/server/tran_server.hpp
```diff
@@ -168,6 +169,8 @@ class tran_server
 	void receive_disconnect_request (page_server_conn_t::sequenced_payload &&a_sp);
 
       private:
+	cubcomm::node m_node;
```
cristiarg: maybe even
```suggestion
	const cubcomm::node m_node;
```
?

src/server/tran_server.hpp
cristiarg: Actually, the whole `cubcomm::node` class is a bit superfluous now.

cristiarg (approved):

src/server/tran_server.hpp
```diff
@@ -233,8 +235,8 @@ class tran_server
     int get_boot_info_from_page_server ();
     int reset_main_connection ();
 
-    int parse_server_host (const std::string &host);
-    int parse_page_server_hosts_config (std::string &hosts);
+    int create_connection_with_server_host (const std::string &host);
+    int create_connections_with_server_hosts_config (std::string &hosts);
```
hornetmj: When I see the code, the physical connectiones are not created in the scope of them. However, "create_connection" make us confuse to believe the connections establish in them. So, the previous name would be better. Or how about making more appropriate name?

wrlawodms: In this function, the connection_handler is created. And "parsing" doesn't indicate it. 
> Or how about making more appropriate name?

Actually, I struggled to come up with a proper name for this. I changed it several times and even thought that "Should I change the code flow if I can't come up with a proper name?", but I concluded it would take more than enough. Do you happen to have a good name for this?  I'd love it if you could suggest it.

src/server/tran_server.hpp
wrlawodms: I'v deleted copy things in this PR.

src/server/tran_server.hpp
wrlawodms: Good. Thanks. Done

src/server/tran_server.hpp
hornetmj: I thinks that create_connection_handler() can take over the role of parse_server_host(). And other jobs like parsing and using host info(_with_server_host) doesn't need to show in the function name. Other option from me is make_connection_handler().

wrlawodms: How about [497194a](https://github.com/CUBRID/cubrid/pull/4464/commits/497194a6d7b48ac13011be64559c653cab32c84e)? I agree that we can hide parsing, but we already have another `create_connection_hander()` function. They can be overloaded, but I don't want to do that since they have different return type and usage.

Another one I considered is 
create_connection_handlers() -> setup_connection_handlers(). Give me your thought.

hornetmj: @wrlawodms off-line discussion was done.

hornetmj (approved):

PR #4466, opened 2023-07-03
[LETS-765] Start ps_conenctor only with remote storage

wrlawodms:
http://jira.cubrid.org/browse/LETS-765

The ps_connector has to run only when there are remote storages to connect.

### DIFF
--- a/src/server/tran_server.cpp
+++ b/src/server/tran_server.cpp
@@ -134,9 +134,9 @@ tran_server::boot (const char *db_name)
 	  ASSERT_ERROR ();
 	  return error_code;
 	}
-    }
 
-  m_ps_connector.start ();
+      m_ps_connector.start ();
+    }
 
   return NO_ERROR;
 }
@@ -625,7 +625,7 @@ tran_server::connection_handler::is_idle ()
 tran_server::ps_connector::ps_connector (tran_server &ts)
   : m_ts { ts }
   , m_daemon { nullptr }
-  , m_terminate { false }
+  , m_terminate { true }
 {
 }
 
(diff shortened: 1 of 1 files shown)


src/server/tran_server.cpp
```diff
@@ -625,7 +625,7 @@ tran_server::connection_handler::is_idle ()
 tran_server::ps_connector::ps_connector (tran_server &ts)
   : m_ts { ts }
   , m_daemon { nullptr }
-  , m_terminate { false }
+  , m_terminate { true }
```
hornetmj, line 628: Nit:

Leave the initialize value as false and add 'if (uses_remote_storage())' conditon before calling the tran_server::ps_connector::terminate () for code consistency. There is no rules but other kind of m_terminate_xxx member value usually set to false as initial value.

hornetmj (approved):

src/server/tran_server.cpp
wrlawodms: Some of the classes do it like that if their internal thread is not started automatically when creating an object. 
See `request_queue_autosend<ReqQueue>::start_thread ()` and `request_server<MsgId>::start_thread ()`.

joohok (approved):

cristiarg (approved):

src/server/tran_server.cpp
hornetmj: @wrlawodms Do you mean m_shutdown flag? The previous comment is 'Nit'. So, it is your choice. I just leave a comment in point of code consistency about m_terminate flag and ps_connector is only created when uses_remote_storage().

wrlawodms: I know you left it at my choice :). I just wanted to clarify it and inform you. Anyway, on second thought, I agree with you in that the ps_connector start only when `if(uses_remote_storage())`, so it should be terminated only in the same condition. Thanks.

wrlawodms: We've decided to go as it is with http://jira.cubrid.org/browse/LETS-767.

PR #4467, opened 2023-07-03
[CBRD-24874] Minimize rule based optimization and delete heuristic factor.

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-24874

 Delete heuristic factor
- Delete statistics estimation
- Delete first node routines. 
- Delete heuristic in calculation formula
-- Remove zero cost calculation for unique index.
-- Remove IO count calculation routine by selectivity, data buffer and the number of inner pages.

Delete RBO
- RBO is used only when cost is within 1.x difference.
- Generate seq_scan even if index scan can be generated. In RBO, prefer index scan compared to sequence scan.

ETC.
- add routines to check plan cost for partial search
- Fixed a bug that used the cumulative selectivity of the next column when calculating the select limit


shparkcubrid:
SQL errors have been reviewed. A detailed description of the SQL error will be written in the next issue.

src/optimizer/query_planner.c
```diff
@@ -3733,18 +3585,15 @@ qo_plan_cmp (QO_PLAN * a, QO_PLAN * b)
 	  return PLAN_COMP_GT;
 	}
 
-      if (!qo_is_index_iss_scan (a) && !qo_is_index_loose_scan (a))
+      if (a->plan_un.scan.index && a->plan_un.scan.index->head->groupby_skip)
```
youngjinj, line 3588: Is it okay not to check `qo_is_iscan ()` ?
Even in AS-IS, it is not restricted to be executed only when `qo_is_iscan ()` is true.

However, since `(QO_PLAN *)->plan_un` is a union,
I think there may be garbage values in `(QO_PLAN *)->plan_un.scan.index` even though it is not an index scan.
Would not it be dangerous to move to `(QO_PLAN *)->plan_un.scan.index` when garbage values are stored?

src/optimizer/query_planner.c
```diff
@@ -3042,94 +2966,10 @@ qo_nljoin_cost (QO_PLAN * planp)
...
       guessed_result_cardinality = (outer->info)->cardinality;
     }
-  inner_cpu_cost = guessed_result_cardinality * (double) QO_CPU_WEIGHT;
-  /* join cost */
-
-  if (qo_is_iscan (inner) && inner->plan_un.scan.index_equi == true)
-    {
-      /* correlated index equi-join */
-      inner_cpu_cost += inner->variable_cpu_cost;
-      if (qo_is_seq_scan (outer) && prm_get_integer_value (PRM_ID_MAX_OUTER_CARD_OF_IDXJOIN) != 0
```
youngjinj, line 3052: If the `PRM_ID_MAX_OUTER_CARD_OF_IDXJOIN` system parameter is no longer used, set the `PRM_DEPRECATED` flag.

src/optimizer/query_planner.c
```diff
@@ -3042,94 +2966,10 @@ qo_nljoin_cost (QO_PLAN * planp)
...
-		      inner_io_cost += diff_cost + 0.1;
-		    }
-		}
-	    }
-	}
-
-      if (planp->plan_un.join.join_type != JOIN_INNER)
-	{
-	  /* outer join leads nongrouped scan overhead */
-	  inner_cpu_cost += ((outer->info)->cardinality * pages * NONGROUPED_SCAN_COST);
```
youngjinj, line 3103: Remove the NONGROUPED_SCAN_COST macro variable if it is no longer used.

src/optimizer/query_planner.c
youngjinj: I thought again that there is no problem because `a->plan_type == QO_PLANTYPE_SCAN` is checked first.

youngjinj (approved):

xmilex-git (approved):

src/optimizer/query_planner.c
shparkcubrid: Okay. I'll fix it on next issue.

PR #4470, opened 2023-07-05
[LETS-766] Add/Remove sink_hook on connecting/disconnecting

wrlawodms:
http://jira.cubrid.org/browse/LETS-766

The sink_hook for sending prior lists to PS was added on the ctor and removed on the dtor of the connection_handler. But now the connection_handler is destroyed with the server. It must be added and removed on connect and disconnect.

- Add pure virtual function: on_connecting() and on_disconnecting(). It's called between the state transition:
  -  on_connecting:     CONNECTING -> (*) -> CONNECTED
  -  on_disconnecting:  DISCONNECTING -> (*) -> IDLE
- Each server type defines its own preliminary jobs for connecting and disconnecting those functions. The add/remove are there.

### DIFF
--- a/src/server/active_tran_server.cpp
+++ b/src/server/active_tran_server.cpp
@@ -186,6 +186,20 @@ active_tran_server::connection_handler::get_saved_lsa () const
 }
 
 void
+active_tran_server::connection_handler::on_connecting ()
+{
+  m_prior_sender_sink_hook_func = std::bind (&tran_server::connection_handler::push_request, this,
+				  tran_to_page_request::SEND_LOG_PRIOR_LIST, std::placeholders::_1);
+  log_Gl.m_prior_sender.add_sink (m_prior_sender_sink_hook_func);
+}
+
+void
+active_tran_server::connection_handler::on_disconnecting ()
+{
+  remove_prior_sender_sink ();
+}
+
+void
 active_tran_server::connection_handler::remove_prior_sender_sink ()
 {
   /*
--- a/src/server/active_tran_server.cpp
+++ b/src/server/active_tran_server.cpp
@@ -135,14 +135,11 @@ active_tran_server::connection_handler::connection_handler (tran_server &ts, cub
   : tran_server::connection_handler (ts, std::move (node))
   ,m_saved_lsa { NULL_LSA }
 {
-  m_prior_sender_sink_hook_func = std::bind (&tran_server::connection_handler::push_request, this,
-				  tran_to_page_request::SEND_LOG_PRIOR_LIST, std::placeholders::_1);
-  log_Gl.m_prior_sender.add_sink (m_prior_sender_sink_hook_func);
 }
 
 active_tran_server::connection_handler::~connection_handler ()
 {
-  remove_prior_sender_sink ();
+  assert (m_prior_sender_sink_hook_func == nullptr);
 }
 
 active_tran_server::connection_handler::request_handlers_map_t
@@ -197,12 +194,6 @@ active_tran_server::connection_handler::on_connecting ()
 
 void
 active_tran_server::connection_handler::on_disconnecting ()
-{
-  remove_prior_sender_sink ();
-}
-
-void
-active_tran_server::connection_handler::remove_prior_sender_sink ()
 {
   if (m_prior_sender_sink_hook_func != nullptr)
     {
--- a/src/server/active_tran_server.cpp
+++ b/src/server/active_tran_server.cpp
@@ -188,6 +188,8 @@ active_tran_server::connection_handler::get_saved_lsa () const
 void
 active_tran_server::connection_handler::on_connecting ()
 {
+  assert (m_prior_sender_sink_hook_func == nullptr);
+
   m_prior_sender_sink_hook_func = std::bind (&tran_server::connection_handler::push_request, this,
 				  tran_to_page_request::SEND_LOG_PRIOR_LIST, std::placeholders::_1);
   log_Gl.m_prior_sender.add_sink (m_prior_sender_sink_hook_func);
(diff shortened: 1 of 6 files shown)


src/server/active_tran_server.cpp
```diff
@@ -186,13 +183,19 @@ active_tran_server::connection_handler::get_saved_lsa () const
 }
 
 void
-active_tran_server::connection_handler::remove_prior_sender_sink ()
+active_tran_server::connection_handler::on_connecting ()
```
hornetmj, line 186: If the job processing in this function is going to heavy later by someone then, the prior_sender could be blocked some period time by m_state_mtx mutex(And other connection handler... ATS -> PS(es)). So, how about checking m_state first before getting the shared mutex m_state_mtx in the tran_server::connection_handler::push_request().

wrlawodms: I understand the prior_sender can get blocked since m_state_mtx is locked exclusively during on_creating() and prior_sender iterates sink_hooks sequentially. However, I don't understand the solution. We can't trust the value of m_state without m_state_mtx. Do I get it right?

Anyway, the problem is valid and we should avoid that situation. I think we can avoid that with releasing m_state_mtx before on_creating() and re-acquiring the mutex after the function. What do you think?

hornetmj: I need to time to check your suggestion. I will come back soon hear. My suggestion is the following:
```
if (m_state != CONNECTED)
{
        // give-up this time, try next call
}

acquire m_state_mtx mutex

// check again
if (m_state != CONNECTED)
{
        // give-up
}
```

hornetmj: @wrlawodms Before start review again ..

Is there any reason the prior_sender has to know(check) the state of conection hander? Please remind me. I expect that prior_sender doesn't check the state of connection handler like CONNECTING, CONNECTED. Only CONNECTED connection is added to the prior_sender and prior_sender just call the function(push_request) without checking the state.

wrlawodms: `push_request` checks the state and only sends requests when "CONNECTED". It acquires the mutex to check it. By the way, now it has a problem. The prior_sender has to be able to send the prior list even during CONNECTING because log records have to be sent during the catch-up and the state of the PS is CONNECTING during the job, I realized it just yesterday. It will be fixed with http://jira.cubrid.org/browse/LETS-768.

All in all,
- the prior_sender will use another function as the prior sink hook, not push_request(). it will acquire the m_conn_mtx only, but not m_state_mtx.
- In this case, on_creating with m_state_mtx doesn't matter for the problem you mentioned. But, I think I should release m_state_mtx before on_connecting() and on_disconnecting(). It will reduce the lock scope.

joohok (approved):

src/server/active_tran_server.cpp
hornetmj: > push_request checks the state and only sends requests when "CONNECTED". It acquires the mutex to check it.

I know it aready. I told that connection is added to prior_sender after changing the state to the CONNECTED. It mean that prior_sender only have CONNECTED state connections. And prior_sender doesn't consider the state of the connection_hander.

wrlawodms: I might miss something, but the prior sink hook is added before the state is set to CONNECTED in `connect()`. It has to be done for the catch-up or some preliminary communication before being set to CONNECTED.

hornetmj: OK. Let's see in the LETS-768 :D

hornetmj (approved):

src/server/tran_server.cpp
```diff
@@ -474,7 +478,7 @@ tran_server::connection_handler::~connection_handler ()
 void
 tran_server::connection_handler::disconnect_async (bool with_disc_msg)
 {
-  auto lockg = std::lock_guard <std::shared_mutex> { m_state_mtx };
+  auto ulock = std::unique_lock <std::shared_mutex> { m_state_mtx };
```
cristiarg: make a scope:
```suggestion
  {
    auto ulock = std::unique_lock <std::shared_mutex> { m_state_mtx };
```

src/server/tran_server.cpp
```diff
@@ -484,6 +488,10 @@ tran_server::connection_handler::disconnect_async (bool with_disc_msg)
   assert (m_state == state::CONNECTING || m_state == state::CONNECTED);
   m_state = state::DISCONNECTING;
 
+  ulock.unlock();
```
```suggestion
  }
```

src/server/tran_server.cpp
```diff
@@ -484,6 +488,10 @@ tran_server::connection_handler::disconnect_async (bool with_disc_msg)
...
+
+  on_disconnecting (); // server-type specific jobs before disconnect
```
cristiarg: Honest question.
Would it be better to also call this in the async functor below?
This would lead to a more unified approach. The function `disconnect_async` just sets the state to `DISCONNECTING` synchronously, while deferring everything to the async functor.
Later on, it might even be needed to alter the order where the `on_disconnecting` is called within the disconnect functor as a whole.

Moreover, as we have these new functions now, their semantics is more like:
- `after_connecting`
- `before_disconnecting`

That's how it looks to me. Is it worthwhile to alter their name?

wrlawodms: > The function disconnect_async just sets the state to DISCONNECTING synchronously, while deferring everything to the async functor.

Hmm, that's what I did, to begin with. now, it doesn't matter so I moved here since it just looks more natural. However, I'll move it to the async functor again, because it has less constraint in the location, on second thought.

> after_connecting
before_disconnecting

I regard what's done are on_connecting and on_disconnecting as the connecting/disconnecting jobs. Without it, there is nothing but transition in that state. What I'm thinking of for the states in detail:
- IDLE -> CONNECTING: establish the internal connection.
- on CONNECTING: preliminary jobs before opening the connection to the outside such as the catch-up, adding prior_sink_hooks
- on DISCONNECTING: preliminary jobs before closing the connection such as removing sink hooks.
- DISCONNECTING -> IDLE: destroy the internal connection. 

Doesn't it make sense?

cristiarg: Yes, it makes sense.
That's how the structure is now in `tran_server::connection_handler::connect` and `tran_server::connection_handler::disconnect_async` and based on this, the `after/before_disconnecting` makes more sense to me.

Maybe I have a bias from Delphi where there is something like [AfterConstruction and BeforeDestruction](https://docwiki.embarcadero.com/Libraries/Sydney/en/System.TObject_Methods) :)
But, [what's in a name](https://writingtips.org/whats-in-a-name/) ?

cristiarg (approved):

PR #4471, opened 2023-07-05
[LETS-754] Change the main connection if a node with a higher priority comes

wrlawodms:
http://jira.cubrid.org/browse/LETS-754

- Reset the main connection on every connect(). This allows the dynamic main connection change on boot and by ps_conenctor.

### DIFF
--- a/src/server/tran_server.cpp
+++ b/src/server/tran_server.cpp
@@ -381,7 +381,7 @@ tran_server::connection_handler::connect ()
   /*
    *  Now, the main connection is reset whenever a new connection is established.
    *  The main connection information is used in the following connection_handler's connection.
-   *  Sepcifically, the target PS to catch up with during CONNECTING is the main connection.
+   *  Specifically, the target PS to catch up with during CONNECTING is the main connection.
    *
    *  TODO: later, when the ATS recovery with multi-PS comes in, a leader PS elected will be the target PS,
    *  and the main connection doesn't need to be set here, but in more appropriate location.
@@ -716,7 +716,7 @@ tran_server::ps_connector::try_connect_to_all_ps (cubthread::entry &)
 	   * TODO It can be too verbose now since it always complain to fail to connect when a PS has been stopped.
 	   * Later on, this job is going to be triggered by a cluster manager or cub_master when a PS is ready to connect.
 	   */
-	  conn->connect ();
+	  (void) conn->connect ();
 	}
 
       if (m_terminate.load ())
--- a/src/server/tran_server.cpp
+++ b/src/server/tran_server.cpp
@@ -367,6 +367,8 @@ tran_server::connection_handler::connect ()
     // moved as argument below
     set_connection (std::move (srv_chn));
 
+
+    // TODO initailizing jobs will be triggered such as the catch-up process. state::CONNECTED will be set asynchronously when it's done.
     m_state = state::CONNECTED;
 
     er_log_debug (ARG_FILE_LINE, "Transaction server successfully connected to the page server. Channel id: %s.\n",
--- a/src/server/tran_server.cpp
+++ b/src/server/tran_server.cpp
@@ -367,8 +367,8 @@ tran_server::connection_handler::connect ()
     // moved as argument below
     set_connection (std::move (srv_chn));
 
-
     // TODO initailizing jobs will be triggered such as the catch-up process. state::CONNECTED will be set asynchronously when it's done.
+
     m_state = state::CONNECTED;
 
     er_log_debug (ARG_FILE_LINE, "Transaction server successfully connected to the page server. Channel id: %s.\n",
--- a/src/server/tran_server.cpp
+++ b/src/server/tran_server.cpp
@@ -435,11 +435,7 @@ tran_server::reset_main_connection ()
     }
 
   auto ulock = std::unique_lock<std::shared_mutex> { m_main_conn_mtx };
-  if (m_main_conn == main_conn_cand->get ())
-    {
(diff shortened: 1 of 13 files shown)


joohok (approved):

src/server/tran_server.cpp
```diff
@@ -330,49 +329,56 @@ tran_server::connection_handler::connect ()
...
+    m_state = state::CONNECTED;
 
-  m_state = state::CONNECTED;
+    er_log_debug (ARG_FILE_LINE, "Transaction server successfully connected to the page server. Channel id: %s.\n",
+		  srv_chn.get_channel_id ().c_str ());
+  }
 
-  er_log_debug (ARG_FILE_LINE, "Transaction server successfully connected to the page server. Channel id: %s.\n",
-		srv_chn.get_channel_id ().c_str ());
+  if (m_ts.reset_main_connection () != NO_ERROR)
```
hornetmj, line 390: How about reverting changes and add reset_main_connection() at the following position?
```
tran_server::ps_connector::try_connect_to_all_ps (cubthread::entry &)
{
  for (const auto &conn : m_ts.m_page_server_conn_vec)
    {
      if (conn->is_idle ())
        {
          /*
           * TODO It can be too verbose now since it always complain to fail to connect when a PS has been stopped.
           * Later on, this job is going to be triggered by a cluster manager or cub_master when a PS is ready to connect.
           */
          conn->connect ();
--> this position (after new conection is established)
        }

      if (m_terminate.load ())
        {
          break;
        }
    }
}
```

wrlawodms: It also must be done during boot() because the catch-up information containing main connection's info has to be sent during connect().

hornetmj: Is it enough to wait if there is no main connection for a while on boot when ATS receives a catch-up request from PS?

wrlawodms: Please clarify it. ATS doesn't "receive" the request, it sends. and it doesn't need to wait for it. Shortly, it sends dummy information to indicate "you're the target PS" when there is no PS, and sends the main connection information to others. This will be addressed in http://jira.cubrid.org/browse/LETS-746 actually.

hornetmj: Discuss off-line. Please add comment about to-do.

hornetmj (approved):

src/server/tran_server.cpp
```diff
@@ -420,11 +434,7 @@ tran_server::reset_main_connection ()
     }
 
   auto ulock = std::unique_lock<std::shared_mutex> { m_main_conn_mtx };
```
Wrap this mutex in a scope.

src/server/tran_server.cpp
```diff
@@ -373,10 +372,25 @@ tran_server::connection_handler::connect ()
...
+  /*
+   *  Now, the main connection is reset whenever a new connection is established.
+   *  The main connection information is used in the following connection_handler's connection.
+   *  Sepcifically, the target PS to catch up with during CONNECTING is the main connection.
+   *
+   *  TODO: later, when the ATS recovery with multi-PS comes in, a leader PS elected will be the target PS,
+   *  and the main connection doesn't need to be set here, but in more appropriate location.
+   *  For example, it can be set after establishing all connection.
+   */
+  if (m_ts.reset_main_connection () != NO_ERROR)
```
cristiarg, line 390: My opinion is that this call should be taken out of here.
And:
- in `tran_server::init_page_server_hosts` be called only once after all connections are established - as it was before 
- in `ps_connector::try_connect_to_all_ps` be called after each successful connection 

I see a clear semantic difference between
- connection_handler::connect to a peer server
- main connection reset, handled by whoever calls the connection_handler::connect (a higher hierarchical entity than the connection_handler)

Or, in other words, "do one thing, and do it well. :)
Feel free to ignore this suggestion. :)

Also:
BTW `ps_connector::try_connect_to_all_ps` does not interpret the code from `connection_handler::connect` - either change to
```
(void) conn->connect ();
```
or handle the error code.

cristiarg (approved):
Sorry for the very late review. I will try to be more prompt.

src/server/tran_server.cpp
wrlawodms: > My opinion is that this call should be taken out of here.

That's the same as what @hornetmj pointed out and I agree. However, the main connection should be set during the first `connect()` for the following `connect()`. See the comment:
```
  *  Now, the main connection is reset whenever a new connection is established.
    *  The main connection information is used in the following connection_handler's connection.
    *  Sepcifically, the target PS to catch up with during CONNECTING is the main connection.
```

I can probably get it done in another way. For example, we can send the catch-up request after all `connect()` are done, and do the state transition to CONNECTED. However, the whole connection sequence will be modified in the next epic (ATS recovery). It's the simplest fix before then.

src/server/tran_server.cpp
wrlawodms: Although this lock is kept at the end of this function?

cristiarg: The code is correct as it is.
With the extra scope it would just be more explicit.
```
{
  // unique_lock
  // if (main ) { .. }  else { .. }
}

return NO_ERROR.
```

Similar to how you added a new scope in `tran_server::connection_handler::connect` because you added code after where the `m_state_mtx` should have ended.

src/server/tran_server.cpp
cristiarg: Fair enough.

src/server/tran_server.cpp
wrlawodms: OK, I get it.

PR #4472, opened 2023-07-05
[CBRD-24684] bug-fix: divide operation result into NUMERIC type

beyondykk9:
http://jira.cubrid.org/browse/CBRD-24684

If limit and offset include a division operation, the between logic is created for inst_num() as shown below.

`limit 7/4 offset 1/2 --> inst_num() > 1/2 and inst_num() <= 1/2 + 7/4`
But in this calculation logic, 1/2 + 7/4 = 2.25 returns wrong result.
The query result below should have 1 row, but a wrong result occurs that brings 2 rows.

`select rownum from db_class limit 7/4 offset 1/2; `
To correct this, we need to treat 1/2 as floor(1/2).

`select rownum from db_class limit 7/4 offset 1/2`
What this query means is to get 7/4 results from offset 1/2.
That is, since we need to get 1.25 results from offset 0.5, we will get 1 result. Because 1.75 is less than 2.
When processing this in a query, it is processed as getting a result between 0.5 and (0.5 + 1.75) of the results. Here, 1/2 or 0.5 offset is not a problem, but the latter part becomes 0.5 + 1.75 = 2.25, so the number of results is 2. That is, 0.5 + 1.75 should be calculated as 0 + 1.75.

### DIFF
--- a/src/parser/parser_support.c
+++ b/src/parser/parser_support.c
@@ -4479,7 +4479,7 @@ pt_limit_to_numbering_expr (PARSER_CONTEXT * parser, PT_NODE * limit, PT_OP_TYPE
       else
 	{
 	  sum->data_type->info.data_type.dec_precision = 0;
-	  sum->info.expr.arg2 = parser_copy_tree (parser, limit);
+	  sum->info.expr.arg1 = parser_copy_tree (parser, limit);
 	}
 
       sum->info.expr.arg2 = parser_copy_tree (parser, limit->next);
--- a/src/parser/parser_support.c
+++ b/src/parser/parser_support.c
@@ -4458,15 +4458,30 @@ pt_limit_to_numbering_expr (PARSER_CONTEXT * parser, PT_NODE * limit, PT_OP_TYPE
 
       sum->data_type->type_enum = PT_TYPE_NUMERIC;
       sum->data_type->info.data_type.precision = DB_MAX_NUMERIC_PRECISION;
-      sum->data_type->info.data_type.dec_precision = 0;
 
       if (oracle_style_divide)
 	{
-	  /* It is necessary to prepare for the division operation in advance. */
+	  PT_NODE *fl_limit;
+
+	  /* It is necessary to prepare for the division operation in advance.
+	     the arg1 and arg2 of the SUM function require a floor operation,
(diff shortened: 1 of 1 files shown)


hgryoo:
Could you please explain the calculation that uses the floor() for the offset? I would appreciate it if you could provide a detailed explanation.

src/parser/parser_support.c
```diff
@@ -4458,15 +4458,30 @@ pt_limit_to_numbering_expr (PARSER_CONTEXT * parser, PT_NODE * limit, PT_OP_TYPE
...
+	  fl_limit->type_enum = PT_TYPE_NUMERIC;
+	  fl_limit->info.expr.arg1 = limit;
+
 	  sum->data_type->info.data_type.dec_precision = DB_DEFAULT_NUMERIC_DIVISION_SCALE;	/* 9 */
+	  sum->info.expr.arg1 = parser_copy_tree (parser, fl_limit);
+	}
+      else
+	{
+	  sum->data_type->info.data_type.dec_precision = 0;
+	  sum->info.expr.arg1 = parser_copy_tree (parser, limit);
```
hgryoo, line 4482: It seems to affect current behavior too.

hyunikn (approved):

src/parser/parser_support.c
```diff
@@ -4458,15 +4458,30 @@ pt_limit_to_numbering_expr (PARSER_CONTEXT * parser, PT_NODE * limit, PT_OP_TYPE
...
 	{
-	  /* It is necessary to prepare for the division operation in advance. */
+	  PT_NODE *fl_limit;
+
+	  /* It is necessary to prepare for the division operation in advance.
+	     the arg1 and arg2 of the SUM function require a floor operation,
+	     but it can apply the floor function only to arg1.
+	     This is because arg2 can respond to SUM operation by setting only the scale. */
+
+	  fl_limit = parser_new_node (parser, PT_EXPR);
```
hgryoo, line 4471: what if parser_new_node() returns NULL?

ctshim (approved):

beyondykk9:
The issue is closed and the related issue is opened at https://github.com/CUBRID/cubrid/pull/4478

PR #4474, opened 2023-07-06
[LETS-746] Add the request to send the target PS and catch-up LSA

wrlawodms:
http://jira.cubrid.org/browse/LETS-746

**catchup_lsa: `unsent_lsa` in LOG_PRIOR_LSA_INFO.** 
- It is the prior_lsa when the prior list is sent, from which new prior nodes will be appended and sent to PSes after sending the current ones. This is the LSA of the first prior node which hasn't been sent yet.
- It's updated when sending prior lists so that the new PS can know the first LSA it will receive when adding its sink_hook.
  - It is taken when prior lists are removed to send in `logpb_prior_lsa_append_all_list`.
  - It is set in log_Gl.m_prior_sender() when sending.
  - It is retrieved and used as the catchup_lsa while adding sink_hook in `on_connecting`.

**The `SEND_START_CATCH_UP` request**
- Send (target ps info { hostname, port }, catchup_lsa)   
- It's sent after adding the sink_hook in `on_connecting`.
- dummy info is sent while the ATS is booting. See the **Note**.

**Miscellaneous**
- cubcomm::node::port - long -> int32_t, a more concrete and precise integer type to pack it.
- Adapt a unit test.
- Some helper functions.

**Note**
- We don't deal with sending the right information in the booting-up case. It will be `receive_start_catch_up: hostname: N/A, port : -1, LSA (-1|-1)`.
- On second thought, I should've split this into two PRs of unsent_lsa and SEND_START_CATCH_UP. Please tell me if it's too hard to review.

### DIFF
--- a/src/server/active_tran_server.cpp
+++ b/src/server/active_tran_server.cpp
@@ -138,6 +138,8 @@ active_tran_server::connection_handler::connection_handler (tran_server &ts)
   m_prior_sender_sink_hook_func = std::bind (&tran_server::connection_handler::push_request, this,
 				  tran_to_page_request::SEND_LOG_PRIOR_LIST, std::placeholders::_1);
   log_Gl.m_prior_sender.add_sink (m_prior_sender_sink_hook_func);
+
+  send_start_catch_up_request ();
 }
 
 active_tran_server::connection_handler::~connection_handler ()
@@ -182,14 +184,14 @@ active_tran_server::connection_handler::receive_saved_lsa (page_server_conn_t::s
 void
 active_tran_server::connection_handler::send_start_catch_up_request ()
 {
-  packing_packer packer;
+  cubpacking::packer packer;
   cubmem::extensible_block ext_blk;
 
-  LOG_LSA lsa;
+  log_lsa lsa = { 1, 2 };
+  long port = 3363L;
   std::string str1 = "str1";
-  std::string str2 = "str2";
 
-  packer.set_buffer_and_pack_all (ext_blk, str1, str2, lsa);
+  packer.set_buffer_and_pack_all (ext_blk, static_cast<int64_t> (lsa), port, str1);
 
   push_request (tran_to_page_request::SEND_START_CATCH_UP, ext_blk.get_ptr ());
 }
--- a/src/server/active_tran_server.cpp
+++ b/src/server/active_tran_server.cpp
@@ -132,8 +132,8 @@ active_tran_server::stop_outgoing_page_server_messages ()
 {
 }
 
-active_tran_server::connection_handler::connection_handler (tran_server &ts)
-  : tran_server::connection_handler (ts)
+active_tran_server::connection_handler::connection_handler (tran_server &ts, cubcomm::node &&node)
+  : tran_server::connection_handler (ts, std::move (node))
   ,m_saved_lsa { NULL_LSA }
 {
   m_prior_sender_sink_hook_func = std::bind (&tran_server::connection_handler::push_request, this,
@@ -227,8 +227,8 @@ active_tran_server::connection_handler::remove_prior_sender_sink ()
 }
 
 active_tran_server::connection_handler *
-active_tran_server::create_connection_handler (tran_server &ts) const
+active_tran_server::create_connection_handler (tran_server &ts, cubcomm::node &&node) const
 {
   // active_tran_server::connection_handler
-  return new connection_handler (ts);
+  return new connection_handler (ts, std::move (node));
 }
--- a/src/server/active_tran_server.cpp
+++ b/src/server/active_tran_server.cpp
@@ -67,7 +67,6 @@ active_tran_server::compute_consensus_lsa ()
   const int total_node_cnt = m_page_server_conn_vec.size ();
   const int quorum = total_node_cnt / 2 + 1; // For now, it's fixed to the number of the majority.
   int cur_node_cnt = 0;
-
   std::vector<log_lsa> collected_saved_lsa;
 
   for (const auto &conn : m_page_server_conn_vec)
--- a/src/server/active_tran_server.cpp
+++ b/src/server/active_tran_server.cpp
@@ -217,14 +217,19 @@ active_tran_server::connection_handler::on_connecting ()
   std::string hostname = "N/A";
   int32_t port = -1;
 
-  if (m_ts.get_main_connection_info (hostname, port) == false)
+  // TODO: We should make two paths to set a connection_handler to CONNECTED:
+  //    - While booting: a target PS is determined after communicating with multiple PSes and
+  //      send the catch-up request based on it.
+  //    - While running: the unsent_lsa from prior_sender is the LSA of the first log records to send.
+  if (!unsent_lsa.is_null ())
     {
-      // there is no main connection yet. it's booting up.
-      // TODO: We should make two paths to set a connection_handler to CONNECTED:
-      //    - While booting: a target PS is determined after communicating with multiple PSes and
-      //      send the catch-up request based on it.
-      //    - While running: the unsent_lsa from prior_sender is the LSA of the first log records to send.
-      unsent_lsa.set_null ();
+      auto res = m_ts.get_main_connection_info (hostname, port);
+      assert (res);
+    }
+  else
+    {
+      // It's booting up and before log_initialize (). There is no main connection yet. For now, it just sends NULL_LSA.
+      // TODO: While booting up, the catchup_lsa is not unsent_lsa, but determined by communicating with PSes.
     }
 
   send_start_catch_up_request (std::move (hostname), port, std::move (unsent_lsa));
--- a/src/server/active_tran_server.cpp
+++ b/src/server/active_tran_server.cpp
@@ -184,16 +184,16 @@ active_tran_server::connection_handler::send_start_catch_up_request (std::string
   cubpacking::packer packer;
   size_t size = 0;
 
-  size += cublog::lsa_utils::get_packed_size (packer, size); // catchup_lsa
-  size += packer.get_packed_int_size (size); // port
   size += packer.get_packed_string_size (host, size); // host
+  size += packer.get_packed_int_size (size); // port
+  size += cublog::lsa_utils::get_packed_size (packer, size); // catchup_lsa
 
   std::unique_ptr < char[] > buffer (new char[size]);
   packer.set_buffer (buffer.get (), size);
 
-  cublog::lsa_utils::pack (packer, catchup_lsa);
-  packer.pack_int (port);
   packer.pack_string (host);
+  packer.pack_int (port);
+  cublog::lsa_utils::pack (packer, catchup_lsa);
 
   push_request_regardless_of_state (tran_to_page_request::SEND_START_CATCH_UP, std::string (buffer.get (), size));
 }
--- a/src/server/active_tran_server.cpp
+++ b/src/server/active_tran_server.cpp
@@ -182,6 +182,8 @@ active_tran_server::connection_handler::send_start_catch_up_request (LOG_LSA &&c
 {
   cubpacking::packer packer;
   size_t size = 0;
+
+  /* TODO: It will be the target PS's info. */
   const auto hostname = m_node.get_host ();
   const auto port = m_node.get_port ();
 
--- a/src/server/active_tran_server.cpp
+++ b/src/server/active_tran_server.cpp
@@ -67,6 +67,7 @@ active_tran_server::compute_consensus_lsa ()
   const int total_node_cnt = m_page_server_conn_vec.size ();
   const int quorum = total_node_cnt / 2 + 1; // For now, it's fixed to the number of the majority.
   int cur_node_cnt = 0;
+
   std::vector<log_lsa> collected_saved_lsa;
 
   for (const auto &conn : m_page_server_conn_vec)
@@ -183,9 +184,14 @@ active_tran_server::connection_handler::send_start_catch_up_request (LOG_LSA &&c
   cubpacking::packer packer;
   size_t size = 0;
 
-  /* TODO: It will be the target PS's info. */
-  const auto hostname = m_node.get_host ();
-  const auto port = m_node.get_port ();
+  std::string hostname;
+  int32_t port = -1;
+
+  if (m_ts.get_main_connection_info (hostname, port) == false)
+    {
+      // there is no main connection yet. you will be the main connection and don't catch-up
+      catchup_lsa.set_null ();
+    }
 
   size += cublog::lsa_utils::get_packed_size (packer, size); // catchup_lsa
   size += packer.get_packed_int_size (size); // port
--- a/src/server/active_tran_server.cpp
+++ b/src/server/active_tran_server.cpp
@@ -186,6 +186,20 @@ active_tran_server::connection_handler::get_saved_lsa () const
 }
 
 void
+active_tran_server::connection_handler::on_connecting ()
+{
+  m_prior_sender_sink_hook_func = std::bind (&tran_server::connection_handler::push_request, this,
+				  tran_to_page_request::SEND_LOG_PRIOR_LIST, std::placeholders::_1);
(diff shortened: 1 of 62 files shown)


src/server/tran_server.hpp
```diff
@@ -190,6 +192,8 @@ class tran_server
 	std::shared_mutex m_state_mtx;
 
 	std::future<void> m_disconn_future; // To delete m_conn asynchronously and make sure there is only one m_conn at a time.
+
+
```
joohok:
```suggestion
```

joohok (approved):

src/server/tran_server.cpp
```diff
@@ -47,7 +47,7 @@ tran_server::register_connection_handler (const std::string &host)
 {
   std::string m_ps_hostname;
   auto col_pos = host.find (":");
```
cristiarg, line 49:
```suggestion
  const auto col_pos = host.find (":");
```

src/transaction/log_prior_send.cpp
```diff
@@ -44,19 +44,23 @@ namespace cublog
       }
 
     std::unique_lock<std::mutex> ulock (m_sink_hooks_mutex);
+    assert (m_unsent_lsa == head->start_lsa);
```
cristiarg: If this assert is to be trusted always, why the need for the `unsent_lsa` argument?

src/transaction/log_prior_send.hpp
```diff
@@ -46,15 +47,18 @@ namespace cublog
...
-      void add_sink (const sink_hook_t &fun);                     // add a hook for a new sink
+      LOG_LSA add_sink (const sink_hook_t &fun);                     // add a hook for a new sink
       void remove_sink (const sink_hook_t &fun);                  // add a hook for a new sink
+      void reset_unsent_lsa (const LOG_LSA &lsa);                    // reset only when prior_lsa is reset
 
     private:
       // non-owning pointers
       std::vector<const sink_hook_t *> m_sink_hooks;              // hooks for sinks
       std::mutex m_sink_hooks_mutex;                              // protect access on sink hooks
+      LOG_LSA m_unsent_lsa;                                          // lsa log records to send, in other word, unsent
```
cristiarg, line 67: initialized in the ctor?
I know it has a default ctor, but that one initializes - I think to `0|0`. Instead of the invalid `-1|-1`. And I think we don't what to change the default ctor for legacy reasons.

src/transaction/log_append.cpp
```diff
@@ -184,6 +184,7 @@ LOG_RESET_APPEND_LSA (const LOG_LSA *lsa)
   // todo - concurrency safe-guard
   log_Gl.hdr.append_lsa = *lsa;
   log_Gl.prior_info.prior_lsa = *lsa;
+  log_Gl.m_prior_sender.reset_unsent_lsa (*lsa);
```
cristiarg: Don't think the function `reset_unsent_lsa` is needed.
Is it that the combination of `reset_unsent_lsa` and the checks (assert) inside `prior_sender::send_list` is intended to ensure than no log prior lists are "lost" between the initialization moment (`LOG_RESET_APPEND_LSA`) and the actual first call to `send_list`?

Because I think that this can be solved in a different manner: enforce that `send_list` is not called without having sinks hooks registered.

I debugged and the chronology of the ATS boot-up is somethink like this:
- `prior_sender` ctor is being called
- connection to all PSs in the static configuration is done
  - the log prior sink hooks to dispatch log prior lists to PS are registered
- the ATS choses the "main connection" (for now, the first one active)
  - but in the future a "leader PS" will be elected
- `tran_server::get_boot_info_from_page_server`
- ATS requests the log header page from main PS - `logpb_fetch_header_from_page_server`
- ATS requests last log page - I think - from main PS - another call to `logpb_fetch_header_from_page_server`
- `LOG_RESET_APPEND_LSA` is called

src/server/active_tran_server.cpp
```diff
@@ -189,7 +211,23 @@ active_tran_server::connection_handler::on_connecting ()
 
   m_prior_sender_sink_hook_func = std::bind (&active_tran_server::connection_handler::prior_sender_sink_hook, this,
 				  std::placeholders::_1);
-  log_Gl.m_prior_sender.add_sink (m_prior_sender_sink_hook_func);
+
+  auto unsent_lsa = log_Gl.m_prior_sender.add_sink (m_prior_sender_sink_hook_func);
```
cristiarg: I am struggling to understand the actual value of this.
So please bear with me if this sounds silly.

To my eyes it looks like this value is destined to tell the booting up PS if it needs to do catch-up or not:
- for cold start the value of `unsent_lsa` will be exactly that from `log_Gl.prior_info.prior_lsa` which is, in turn, received from the chosen leader PS
- for PS hot-plug-ins (hot-restarts) the value will be "some value" which will be advanced enough that the booting-up PS will decide for itself that it needs to do catch-up

Are these assumptions correct?
If they are, I don't think that we need to store this `m_unsent_lsa` in the log prior sender and have all this mechanics of keeping it up to date. We can just pull the value from `log_Gl.prior_info.prior_lsa`.

src/server/active_tran_server.cpp
```diff
@@ -189,7 +211,23 @@ active_tran_server::connection_handler::on_connecting ()
...
+
+  std::string hostname = "N/A";
+  int32_t port = -1;
+
+  if (m_ts.get_main_connection_info (hostname, port) == false)
+    {
+      // there is no main connection yet. it's booting up.
```
cristiarg: agreed with this statement

src/server/page_server.cpp
```diff
@@ -184,6 +188,26 @@ page_server::connection_handler::handle_oldest_active_mvccid_request (tran_serve
...
+
+  std::string host;
+  int32_t port;
+  LOG_LSA catchup_lsa;
+
+  unpacker.unpack_string (host);
+  unpacker.unpack_int (port);
+  cublog::lsa_utils::unpack (unpacker, catchup_lsa);
+
+  er_log_debug (ARG_FILE_LINE, "receive_start_catch_up: hostname: %s, port : %d, LSA (%lld|%d)\n", host.c_str (), port,
```
```suggestion
  er_log_debug (ARG_FILE_LINE, "receive_start_catch_up: hostname = %s, port = %d, LSA = (%lld|%d)\n", host.c_str (), port,
```
makes it easier to search in log

src/server/tran_server.cpp
```diff
@@ -454,6 +454,21 @@ tran_server::uses_remote_storage () const
   return false;
 }
 
+bool tran_server::get_main_connection_info (std::string &host_out, int32_t &port_out)
+{
+  auto slock = std::shared_lock<std::shared_mutex> { m_main_conn_mtx };
+  if (m_main_conn == nullptr)
+    {
+      return false;
```
cristiarg, line 471: Actually, more consistent to set the invalid host/port values here - in the same place where the valid values are set.
```suggestion
      host_out = "N/A";
      port_out = -1;
      return false;
```

src/server/tran_server.cpp
```diff
@@ -454,6 +454,21 @@ tran_server::uses_remote_storage () const
...
+    }
+
+  auto &node = m_main_conn->get_node ();
```
cristiarg, line 474: Nit.
Although maybe superfluous as `auto` == `const cubcomm::node` already. Not sure. Some `decltype` debugging should tell you if superfluous or not.
```suggestion
  const auto &node = m_main_conn->get_node ();
```

src/server/tran_server.cpp
```diff
@@ -667,6 +683,12 @@ tran_server::connection_handler::is_idle ()
   return m_state == state::IDLE;
 }
 
+const cubcomm::node &
+tran_server::connection_handler::get_node () const
```
hornetmj, line 696: Nit:

How about tran_server::get_main_connection_info() -> tran_server::connection_handler::get_connection_info()?

src/transaction/log_prior_send.cpp
```diff
@@ -26,7 +26,7 @@
 namespace cublog
 {
   void
-  prior_sender::send_list (const log_prior_node *head)
+  prior_sender::send_list (const log_prior_node *head, const LOG_LSA *unsent_lsa)
```
hornetmj, line 37: About the naming of the unsent_lsa, it is hard to understand. How about just catchup_lsa or ps_catchup_lsa? (m_catchup_lsa)

src/server/active_tran_server.cpp
```diff
@@ -189,7 +211,23 @@ active_tran_server::connection_handler::on_connecting ()
...
+      // TODO: We should make two paths to set a connection_handler to CONNECTED:
+      //    - While booting: a target PS is determined after communicating with multiple PSes and
+      //      send the catch-up request based on it.
+      //    - While running: the unsent_lsa from prior_sender is the LSA of the first log records to send.
+      unsent_lsa.set_null ();
```
wrlawodms: Others besides the leader PS also can do the "cold-start". It'd be better to send just LSA and let the PS decide if it's cold-start or not.

src/server/active_tran_server.cpp
wrlawodms: You have a point, are not silly at all. Let me share the design choice and have your thought on it. 

What I wanted is, basically, to send the "exact" LSA of the log record which the PS will start to receive.
Let's say this the log on ATS:    `===(A)===[unsent_lsa]-----(B)----->[prior_lsa]`
 The thing is that log prior lists are cut from the list in the prior sender and sent asynchronously. Therefore, there is a case where the (A) range is cut and not sent, and the (B) range is in the list. In this case, a PS connected at this moment may receive (A) or may not, and will receive (B). 

 Must we avoid this situation? No, I don't think so. The PS can just throw it out if the received log records are in the past than the catch-up LSA. However, it does not only cost, but I think it can also hide problem which can happen during development. If some not intended log records in the past are sent to PS for some reason, it would be just thrown out and ignored.

src/server/tran_server.cpp
wrlawodms: Thank you for the pointer. I'll keep it in mind.

src/server/tran_server.cpp
wrlawodms: If it's not reference, the const would be cut off, but it's read-only as it is. I was not sure, but I've tested it.

src/server/tran_server.cpp
wrlawodms: Then, we should make the `m_main_conn` public. The protected `m_main_conn` is open to the `active_tran_server`, but open to `active_tran_server::connection_handler`, so `active_tran_server::connection_handler::on_connecting` can't access it.

src/server/tran_server.hpp
wrlawodms: thanks.

src/transaction/log_prior_send.cpp
wrlawodms: It's checking the previous one is the same to `head->start_lsa`. We don't know the new value.

src/transaction/log_prior_send.hpp
wrlawodms: I agree that we should do it on the actor.
> I think to 0|0. Instead of the invalid -1|-1

Why do you think 0|0 is better? I'd set it -1|-1 which is defined as initial value, LSA_INITIALIZER.
> And I think we don't what to change the default ctor for legacy reasons.

I don't understand this. Is there an issue to make the default ctor like
`prior_sender () : m_unsent_lsa { -1, -1}`

src/transaction/log_append.cpp
wrlawodms: It's not intended for that purpose. I should've described it. It's for the case where the `send_list` hasn't been called yet, but a PS requests the catch-up LSA not on booting. I presume that it's possible when the ATS is done starting without recovery since there is nothing to recover. If there were no recovery/checkpoint/any updates, it hasn't send any prior list. And, a PS attaches at this moment, it will get the uninitialized catch-up LSA without this code.

wrlawodms:
I'll resolve the merge conflict after a [dicussion](https://github.com/CUBRID/cubrid/pull/4474#discussion_r1269141811) that affects the conflicted code.

hornetmj (approved):

src/transaction/log_prior_send.hpp
cristiarg: I do not say that `0|0` is better.
I just say that - without and explicit initialization - like `LSA_INITIALIZER`, the `LOG_LSA` ctor initializes the values to `0|0` which is a valid value.
And we would want to initialize to some invalid value - `-1|-1`.

The legacy ctor of `log_lsa` is wrong and - indeed - we are by no means going to fix it now because .. legacy.

`prior_sender () : m_unsent_lsa { -1, -1}` is definitely the correct way.

I am sorry for my message. I realize now that it is confusing. I will try to be more precise.

cristiarg: I do not say that `0|0` is better.
I just say that - without and explicit initialization - like `LSA_INITIALIZER`, the `LOG_LSA` ctor initializes the values to `0|0` which is a valid value.
And we would want to initialize to some invalid value - `-1|-1`.

The legacy ctor of `log_lsa` is wrong and - indeed - we are by no means going to fix it now because .. legacy.

`prior_sender () : m_unsent_lsa { -1, -1}` is definitely the correct way.

I am sorry for my message. I realize now that it is confusing. I will try to be more precise.

src/server/active_tran_server.cpp
cristiarg: > Must we avoid this situation? No, I don't think so.

I understand your reasoning.
But I still remain to my idea that we might make it work without this. It's also possible that I miss something from your design.

Let's move forward and we'll correct/refactor later when we have a clearer picture of the mechanism - if needed. :)

src/transaction/log_append.cpp
cristiarg: > hasn't been called yet, but a PS requests the catch-up LSA not on booting.

Isn't this a practically impossible situation?

cristiarg (approved):

src/transaction/log_append.cpp
wrlawodms: It seems allowed by design, and I presume it's possible. If you start cub_server after a normal shutdown with develop, it doesn't do recovery and only `LOG_DUMMY_HA_SERVER_STATE` is appended, which I don't think is necessary for a database.

unit_tests/log/test_main_prior_list_serialize.cpp
```diff
@@ -464,11 +464,6 @@ namespace cublog
   {
     assert_release (!m_thread.joinable ());
   }
-
-  prior_sender::~prior_sender ()
-  {
-    assert_release (false);
-  }
```
wrlawodms, line 471: Now, we add "${TRANSACTION_DIR}/log_prior_send.cpp" to use `m_sender.reset_unsent_lsa` in the test, it's duplicated.

src/transaction/log_impl.h
```diff
@@ -737,7 +737,12 @@ struct log_global
   void initialize_log_prior_sender ();
   void finalize_log_prior_sender ();
   // *INDENT-OFF*
-  cublog::prior_sender &get_log_prior_sender ();
+  cublog::prior_sender &
+  get_log_prior_sender()
+  {
+    assert (m_prior_sender != nullptr);
+    return *m_prior_sender;
+  }
```
wrlawodms, line 745: Now `LOG_RESET_APPEND_LSA` use this function that a test fails to build with the msg:
```
CMakeFiles/test_prior_list_serialize.dir/__/__/src/transaction/log_append.cpp.o: In function `LOG_RESET_APPEND_LSA(log_lsa const*)':
/home/cubrid/cubrid/lets/src/transaction/log_append.cpp:188: undefined reference to `log_global::get_log_prior_sender()'
```

wrlawodms: I could put the implementation, `log_global.c`, but causes more errors since it has more dependencies on other source files.

cristiarg: It is ok like this.

cristiarg (approved):

PR #4476, opened 2023-07-07
[CBRD-24860] Implement functions related to memory tracking for Memory Monitoring Manager Module

Rudeus:
http://jira.cubrid.org/browse/CBRD-24860

Implementation of memory tracking in memory monitoring manager module

description of memory_monitor::add/sub_stat

1. get module_idx, comp_map_idx through stat_id
2. if LOG_ISRESTARTED() == 0, allocation_at_init = true
3. get component index(comp_idx) and subcomponent index(subcomp_idx) using comp_idx_map in module
4. update m_total_mem_usage
5. call m_module[module_idx]->add_stat (..)
    add/sub_stat (..) in mmon_module and mmon_component
        1. update init_stat(if allocation/deallocation_at_init == true), cur_stat, peak_stat(if cur_stat > peak_stat) of module
        2. (only add_stat) update expand_count if is_expanded == true
        3. if comp_idx or subcomp_idx != mmon_module::MAX_COMP_IDX, call mmon_component::add/sub_stat (..) or mmon_subcomponent::add/sub_cur_stat (..)
6. if transaction exists, update tdes->cur_mem_usage

resize_stat and move_stat just use memory_monitor::add_stat/sub_stat (..)

### DIFF
--- a/src/base/memory_monitor_sr.cpp
+++ b/src/base/memory_monitor_sr.cpp
@@ -465,11 +465,8 @@ namespace cubperf
 
     if (thread_p)
       {
-	tdes = LOG_FIND_TDES (thread_p->tran_index);
-	if (tdes != NULL)
-	  {
-	    tdes->cur_mem_usage += size;
-	  }
+	tdes = LOG_FIND_CURRENT_TDES (thread_p);
+	tdes->cur_mem_usage += size;
       }
   }
 
@@ -492,11 +489,8 @@ namespace cubperf
 
     if (thread_p)
       {
-	tdes = LOG_FIND_TDES (thread_p->tran_index);
-	if (tdes != NULL)
-	  {
-	    tdes->cur_mem_usage -= size;
-	  }
+	tdes = LOG_FIND_CURRENT_TDES (thread_p);
+	tdes->cur_mem_usage -= size;
       }
   }
 
--- a/src/base/memory_monitor_sr.cpp
+++ b/src/base/memory_monitor_sr.cpp
@@ -25,8 +25,9 @@
 #include <vector>
 #include <cassert>
 
-#include "perf.hpp"
 #include "error_manager.h"
+#include "system_parameter.h"
+#include "perf.hpp"
 #include "memory_monitor_sr.hpp"
 #include "log_impl.h"
 
@@ -64,8 +65,7 @@ namespace cubperf
       mmon_subcomponent &operator = (mmon_subcomponent &&) = delete;
 
       const char *get_name ();
-      void add_cur_stat (uint64_t size);
-      void sub_cur_stat (uint64_t size);
+      void add_cur_stat (int64_t size);
 
     private:
       std::string m_subcomp_name;
@@ -86,9 +86,9 @@ namespace cubperf
       mmon_component &operator = (mmon_component &&) = delete;
 
       const char *get_name ();
-      void add_stat (uint64_t size, int subcomp_idx, bool allocation_at_init);
-      void sub_stat (uint64_t size, int subcomp_idx, bool deallocation_at_init);
+      void add_stat (int64_t size, int subcomp_idx);
       void add_expand_count ();
+      void end_init_phase ();
       int get_subcomp_index (const char *subcomp_name);
       int add_subcomponent (const char *name);
 
@@ -117,12 +117,17 @@ namespace cubperf
       mmon_module &operator = (mmon_module &&) = delete;
 
       virtual int aggregate_stats (const MMON_MODULE_INFO &info);
-      void add_stat (uint64_t size, int comp_idx, int subcomp_idx, bool allocation_at_init);
-      void sub_stat (uint64_t size, int comp_idx, int subcomp_idx, bool deallocation_at_init);
-      void add_expand_count (int comp_idx);
+      void add_stat (MMON_STAT_ID stat_id, int64_t size);
+      void add_expand_count (MMON_STAT_ID stat_id);
+      void end_init_phase ();
       int add_component (const char *name);
-      int get_comp_idx_map_val (int idx);
       inline int make_comp_idx_map_val (int comp_idx, int subcomp_idx);
+
+    private:
+      void get_comp_and_subcomp_idx (MMON_STAT_ID stat_id, int &comp_idx, int &subcomp_idx);
+      inline int get_comp_map_idx (MMON_STAT_ID stat_id);
+      inline int get_comp_idx (int comp_idx_map_val);
+      inline int get_subcomp_idx (int comp_idx_map_val);
 
     private:
       std::string m_module_name;
@@ -145,14 +150,16 @@ namespace cubperf
       memory_monitor &operator = (const memory_monitor &) = delete;
       memory_monitor &operator = (memory_monitor &&) = delete;
 
-      void add_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, uint64_t size);
-      void sub_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, uint64_t size);
+      void add_stat (MMON_STAT_ID stat_id, int64_t size);
+      void add_tran_stat (THREAD_ENTRY *thread_p, int64_t size);
       void resize_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, uint64_t old_size, uint64_t new_size);
       void move_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID src, MMON_STAT_ID dest, uint64_t size);
+      void end_init_phase ();
       int aggregate_module_info (MMON_MODULE_INFO *info, int module_index);
       int aggregate_tran_info (MMON_TRAN_INFO *info, int tran_count);
-      inline int get_comp_idx_from_comp_idx_map (int comp_idx_map_val);
-      inline int get_subcomp_idx_from_comp_idx_map (int comp_idx_map_val);
+
+    private:
+      inline int get_module_idx (MMON_STAT_ID stat_id);
 
     private:
       class aggregater
@@ -188,12 +195,21 @@ namespace cubperf
     return (comp_idx << 16 | subcomp_idx);
   }
 
-  inline int memory_monitor::get_comp_idx_from_comp_idx_map (int comp_idx_map_val)
+  inline int mmon_module::get_comp_map_idx (MMON_STAT_ID stat_id)
+  {
+    return (((int) stat_id) & MMON_PARSE_MASK);
+  }
+  inline int memory_monitor::get_module_idx (MMON_STAT_ID stat_id)
+  {
+    return (((int) stat_id) >> 16);
+  }
+
+  inline int mmon_module::get_comp_idx (int comp_idx_map_val)
   {
     return (comp_idx_map_val >> 16);
   }
 
-  inline int memory_monitor::get_subcomp_idx_from_comp_idx_map (int comp_idx_map_val)
+  inline int mmon_module::get_subcomp_idx (int comp_idx_map_val)
   {
     return (comp_idx_map_val & MMON_PARSE_MASK);
   }
@@ -203,14 +219,9 @@ namespace cubperf
     return m_subcomp_name.c_str ();
   }
 
-  void mmon_subcomponent::add_cur_stat (uint64_t size)
+  void mmon_subcomponent::add_cur_stat (int64_t size)
   {
     m_cur_stat += size;
-  }
-
-  void mmon_subcomponent::sub_cur_stat (uint64_t size)
-  {
-    m_cur_stat -= size;
   }
 
   const char *mmon_component::get_name ()
@@ -218,13 +229,8 @@ namespace cubperf
     return m_comp_name.c_str ();
   }
 
-  void mmon_component::add_stat (uint64_t size, int subcomp_idx, bool allocation_at_init)
+  void mmon_component::add_stat (int64_t size, int subcomp_idx)
   {
-    if (allocation_at_init)
-      {
-	m_stat.init_stat += size;
-      }
-
     m_stat.cur_stat += size;
 
     if (m_stat.cur_stat > m_stat.peak_stat)
@@ -232,32 +238,22 @@ namespace cubperf
 	m_stat.peak_stat = m_stat.cur_stat.load ();
       }
 
-    /* If there is a subcomponent info */
+    /* If there is a component info */
     if (subcomp_idx != mmon_module::MAX_COMP_IDX)
       {
+	assert (subcomp_idx < m_subcomponent.size ());
 	m_subcomponent[subcomp_idx]->add_cur_stat (size);
-      }
-  }
-
-  void mmon_component::sub_stat (uint64_t size, int subcomp_idx, bool deallocation_at_init)
-  {
-    if (deallocation_at_init)
-      {
-	m_stat.init_stat -= size;
-      }
-
-    m_stat.cur_stat -= size;
-
-    /* If there is a subcomponent info */
-    if (subcomp_idx != mmon_module::MAX_COMP_IDX)
-      {
-	m_subcomponent[subcomp_idx]->sub_cur_stat (size);
       }
   }
 
   void mmon_component::add_expand_count ()
   {
     m_stat.expand_count++;
+  }
+
+  void mmon_component::end_init_phase ()
+  {
+    m_stat.init_stat = m_stat.cur_stat.load ();
   }
 
   int mmon_component::get_subcomp_index (const char *subcomp_name)
@@ -342,12 +338,21 @@ namespace cubperf
     return error;
   }
 
-  void mmon_module::add_stat (uint64_t size, int comp_idx, int subcomp_idx, bool allocation_at_init)
+  void mmon_module::get_comp_and_subcomp_idx (MMON_STAT_ID stat_id, int &comp_idx, int &subcomp_idx)
   {
-    if (allocation_at_init)
-      {
-	m_stat.init_stat += size;
-      }
+    int comp_map_idx = get_comp_map_idx (stat_id);
+    int idx_map_val;
+
+    idx_map_val = m_comp_idx_map[comp_map_idx];
+    comp_idx = get_comp_idx (idx_map_val);
+    subcomp_idx = get_subcomp_idx (idx_map_val);
+  }
+
+  void mmon_module::add_stat (MMON_STAT_ID stat_id, int64_t size)
+  {
+    int comp_idx, subcomp_idx;
+
+    get_comp_and_subcomp_idx (stat_id, comp_idx, subcomp_idx);
 
     m_stat.cur_stat += size;
 
@@ -359,34 +364,34 @@ namespace cubperf
     /* If there is a component info */
     if (comp_idx != mmon_module::MAX_COMP_IDX)
       {
-	m_component[comp_idx]->add_stat (size, subcomp_idx, allocation_at_init);
+	assert (comp_idx < m_component.size ());
+	m_component[comp_idx]->add_stat (size, subcomp_idx);
       }
   }
 
-  void mmon_module::sub_stat (uint64_t size, int comp_idx, int subcomp_idx, bool deallocation_at_init)
+  void mmon_module::add_expand_count (MMON_STAT_ID stat_id)
   {
-    if (deallocation_at_init)
-      {
-	m_stat.init_stat -= size;
-      }
+    int comp_idx, subcomp_idx;
 
-    m_stat.cur_stat -= size;
-
-    /* If there is a component info */
-    if (comp_idx != mmon_module::MAX_COMP_IDX)
-      {
-	m_component[comp_idx]->sub_stat (size, subcomp_idx, deallocation_at_init);
-      }
-  }
-
-  void mmon_module::add_expand_count (int comp_idx)
-  {
     m_stat.expand_count++;
 
+    get_comp_and_subcomp_idx (stat_id, comp_idx, subcomp_idx);
+
     /* If there is a component info */
     if (comp_idx != mmon_module::MAX_COMP_IDX)
       {
+	assert (comp_idx < m_component.size ());
 	m_component[comp_idx]->add_expand_count ();
+      }
+  }
+
+  void mmon_module::end_init_phase ()
+  {
+    m_stat.init_stat = m_stat.cur_stat.load ();
+
+    for (auto i = 0; i < m_component.size (); i++)
+      {
+	m_component[i]->end_init_phase ();
       }
   }
 
@@ -394,11 +399,6 @@ namespace cubperf
   {
     m_component.push_back (std::make_unique<mmon_component> (name));
     return (m_component.size () - 1);
-  }
-
-  int mmon_module::get_comp_idx_map_val (int idx)
-  {
-    return m_comp_idx_map[idx];
   }
 
   memory_monitor::aggregater::aggregater (memory_monitor *mmon)
@@ -421,72 +421,49 @@ namespace cubperf
     return 0;
   }
 
-  void memory_monitor::add_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, uint64_t size)
+  void memory_monitor::add_stat (MMON_STAT_ID stat_id, int64_t size)
   {
-    LOG_TDES *tdes;
-    TRANID trid = NULL_TRANID;
-    int module_idx = MMON_GET_MODULE_IDX_FROM_STAT_ID (stat_id);
-    int comp_map_idx = MMON_GET_COMP_MAP_IDX_FROM_STAT_ID (stat_id);
-    int idx_map_val, comp_idx, subcomp_idx;
-    bool allocation_at_init = !LOG_ISRESTARTED ();
-
-    idx_map_val = m_module[module_idx]->get_comp_idx_map_val (comp_map_idx);
-    comp_idx = get_comp_idx_from_comp_idx_map (idx_map_val);
-    subcomp_idx = get_subcomp_idx_from_comp_idx_map (idx_map_val);
+    int module_idx = get_module_idx (stat_id);
 
     m_total_mem_usage += size;
 
-    m_module[module_idx]->add_stat (size, comp_idx, subcomp_idx, allocation_at_init);
-
-    tdes = LOG_FIND_CURRENT_TDES (thread_p);
-    assert (tdes != NULL);
-    tdes->cur_mem_usage += size;
+    m_module[module_idx]->add_stat (stat_id, size);
   }
 
-  void memory_monitor::sub_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, uint64_t size)
+  void memory_monitor::add_tran_stat (THREAD_ENTRY *thread_p, int64_t size)
   {
-    LOG_TDES *tdes;
-    TRANID trid = NULL_TRANID;
-    int module_idx = MMON_GET_MODULE_IDX_FROM_STAT_ID (stat_id);
-    int comp_map_idx = MMON_GET_COMP_MAP_IDX_FROM_STAT_ID (stat_id);
-    int idx_map_val, comp_idx, subcomp_idx;
-    bool deallocation_at_init = !LOG_ISRESTARTED ();
-
-    idx_map_val = m_module[module_idx]->get_comp_idx_map_val (comp_map_idx);
-    comp_idx = get_comp_idx_from_comp_idx_map (idx_map_val);
-    subcomp_idx = get_subcomp_idx_from_comp_idx_map (idx_map_val);
-
-    m_total_mem_usage -= size;
-
-    m_module[module_idx]->sub_stat (size, comp_idx, subcomp_idx, deallocation_at_init);
-
-    tdes = LOG_FIND_CURRENT_TDES (thread_p);
+    LOG_TDES *tdes = LOG_FIND_CURRENT_TDES (thread_p);
     assert (tdes != NULL);
-    tdes->cur_mem_usage -= size;
+    tdes->cur_mem_usage += size;
   }
 
   void memory_monitor::resize_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, uint64_t old_size,
 				    uint64_t new_size)
   {
-    int module_idx = MMON_GET_MODULE_IDX_FROM_STAT_ID (stat_id);
-    int comp_map_idx = MMON_GET_COMP_MAP_IDX_FROM_STAT_ID (stat_id);
-    int idx_map_val, comp_idx, subcomp_idx;
+    int module_idx = get_module_idx (stat_id);
 
-    sub_stat (thread_p, stat_id, old_size);
-    add_stat (thread_p, stat_id, new_size);
+    add_stat (stat_id, - (old_size));
+    add_tran_stat (thread_p, - (old_size));
+    add_stat (stat_id, new_size);
+    add_tran_stat (thread_p, new_size);
     if (new_size - old_size > 0) /* expand */
       {
-	idx_map_val = m_module[module_idx]->get_comp_idx_map_val (comp_map_idx);
-	comp_idx = get_comp_idx_from_comp_idx_map (idx_map_val);
-	subcomp_idx = get_subcomp_idx_from_comp_idx_map (idx_map_val);
-	m_module[module_idx]->add_expand_count (comp_idx);
+	m_module[module_idx]->add_expand_count (stat_id);
       }
   }
 
   void memory_monitor::move_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID src, MMON_STAT_ID dest, uint64_t size)
   {
-    sub_stat (thread_p, src, size);
-    add_stat (thread_p, dest, size);
+    add_stat (src, - (size));
+    add_stat (dest, size);
+  }
+
+  void memory_monitor::end_init_phase ()
+  {
+    for (int i = 0; i < MMON_MODULE_LAST; i++)
+      {
+	m_module[i]->end_init_phase ();
+      }
   }
 
   int memory_monitor::aggregate_module_info (MMON_MODULE_INFO *info, int module_index)
@@ -501,43 +478,76 @@ namespace cubperf
 } // namespace cubperf
 
 /* APIs */
+// TODO: API parameter check (after error code setting)
+using namespace cubperf;
+
 int mmon_initialize (const char *server_name)
 {
   int error = NO_ERROR;
-  cubperf::mmon_Gl = new cubperf::memory_monitor (server_name);
 
-  if (cubperf::mmon_Gl == nullptr)
+  assert (server_name != NULL);
+
+  if (prm_get_bool_value (PRM_ID_MEMORY_MONITORING))
     {
-      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, sizeof (cubperf::memory_monitor));
-      error = ER_OUT_OF_VIRTUAL_MEMORY;
-    }
+      mmon_Gl = new memory_monitor (server_name);
 
+      if (mmon_Gl == nullptr)
+	{
+	  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, sizeof (memory_monitor));
+	  error = ER_OUT_OF_VIRTUAL_MEMORY;
+	}
+    }
   return error;
+}
+
+void mmon_notify_server_start ()
+{
+  if (prm_get_bool_value (PRM_ID_MEMORY_MONITORING))
+    {
+      mmon_Gl->end_init_phase ();
+    }
 }
 
 void mmon_finalize ()
 {
-  delete cubperf::mmon_Gl;
+  if (prm_get_bool_value (PRM_ID_MEMORY_MONITORING))
+    {
+      delete mmon_Gl;
+    }
 }
 
-void mmon_add_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, uint64_t size)
+void mmon_add_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, int64_t size)
 {
-  cubperf::mmon_Gl->add_stat (thread_p, stat_id, size);
+  if (prm_get_bool_value (PRM_ID_MEMORY_MONITORING))
+    {
+      mmon_Gl->add_stat (stat_id, size);
+      mmon_Gl->add_tran_stat (thread_p, size);
+    }
 }
 
-void mmon_sub_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, uint64_t size)
+void mmon_sub_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, int64_t size)
 {
-  cubperf::mmon_Gl->sub_stat (thread_p, stat_id, size);
+  if (prm_get_bool_value (PRM_ID_MEMORY_MONITORING))
+    {
+      mmon_Gl->add_stat (stat_id, - (size));
+      mmon_Gl->add_tran_stat (thread_p, - (size));
+    }
 }
 
-void mmon_move_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID src, MMON_STAT_ID dest, uint64_t size)
+void mmon_move_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID src, MMON_STAT_ID dest, int64_t size)
 {
-  cubperf::mmon_Gl->move_stat (thread_p, src, dest, size);
+  if (prm_get_bool_value (PRM_ID_MEMORY_MONITORING))
+    {
+      mmon_Gl->move_stat (thread_p, src, dest, size);
+    }
 }
 
-void mmon_resize_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, uint64_t old_size, uint64_t new_size)
+void mmon_resize_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, int64_t old_size, int64_t new_size)
 {
-  cubperf::mmon_Gl->resize_stat (thread_p, stat_id, old_size, new_size);
+  if (prm_get_bool_value (PRM_ID_MEMORY_MONITORING))
+    {
+      mmon_Gl->resize_stat (thread_p, stat_id, old_size, new_size);
+    }
 }
 
 int mmon_aggregate_module_info (MMON_MODULE_INFO *info, int module_index)
--- a/src/base/memory_monitor_sr.cpp
+++ b/src/base/memory_monitor_sr.cpp
@@ -86,8 +86,8 @@ namespace cubperf
       mmon_component &operator = (mmon_component &&) = delete;
 
       const char *get_name ();
-      void add_stat (uint64_t size, int subcomp_idx, bool init, bool expand);
-      void sub_stat (uint64_t size, int subcomp_idx, bool init);
+      void add_stat (uint64_t size, int subcomp_idx, bool allocation_at_init, bool expand);
+      void sub_stat (uint64_t size, int subcomp_idx, bool deallocation_at_init);
       int get_subcomp_index (const char *subcomp_name);
       int add_subcomponent (const char *name);
 
@@ -100,6 +100,12 @@ namespace cubperf
   class mmon_module
   {
     public:
+      /* max index of component or subcomponent is 0x0000FFFF
+       * if some stats have max index of component or subcomponent,
+       * we don't have to increase it */
+      static constexpr int MAX_COMP_IDX = 0x0000FFFF;
+
+    public:
       mmon_module (const char *module_name, const MMON_STAT_INFO *info);
       mmon_module (const mmon_module &) = delete;
       mmon_module (mmon_module &&) = delete;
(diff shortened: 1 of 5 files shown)


joohok:
Is there a reason why the "class definition", "implementation of method", "API definition', and "API implementation" are all written in one file? It seems like it would be better to separate the **memory_monitor class** and the **API** so that users only need to focus on the API part or memory_monitor

In my opinion, it would be good to divide them as follows:
Just an example
- Class definition - memory_monitor_def.hpp
- Implementation of method - memory_monitor_def.cpp
- API definition - memory_monitor_api.h
- API implementation - memory_monitor_api.c

src/transaction/log_impl.h
```diff
@@ -602,6 +603,7 @@ struct log_tdes
   log_postpone_cache m_log_postpone_cache;
 
   bool has_supplemental_log;	/* Checks if supplemental log has been appended within the transaction */
+    std::atomic < uint64_t > cur_mem_usage;
```
wrlawodms: It has to be initialized to 0 explicitly, probably in `logtb_clear_tdes` and `logtb_clear_tdes`. There might be more. Please check it.

src/base/memory_monitor_sr.cpp
```diff
@@ -358,25 +439,88 @@ namespace cubperf
...
+    subcomp_idx = get_subcomp_idx_from_comp_idx_map (idx_map_val);
+
+    m_total_mem_usage -= size;
+
+    m_module[module_idx]->sub_stat (size, comp_idx, subcomp_idx, deallocation_at_init);
+
+    if (thread_p)
+      {
+	tdes = LOG_FIND_TDES (thread_p->tran_index);
+	if (tdes != NULL)
```
wrlawodms: Do you assume tdes can be NULL when calling this function? Could you give me an example?

src/base/memory_monitor_sr.cpp
```diff
@@ -281,61 +315,108 @@ namespace cubperf
...
 
   int mmon_module::aggregate_stats (const MMON_MODULE_INFO &info)
   {
-    return 0;
+    int error = NO_ERROR;
+    return error;
   }
 
-  void mmon_module::add_stat (uint64_t size, int comp_idx, int subcomp_idx, bool init, bool expand)
+  void mmon_module::add_stat (uint64_t size, int comp_idx, int subcomp_idx, bool allocation_at_init, bool is_expanded)
```
wrlawodms: Could you explain the meaning of "expand" with an example"? I think I misunderstood it. I'd like to make sure it should be given as a parameter.

src/base/memory_monitor_sr.cpp
```diff
@@ -281,61 +315,108 @@ namespace cubperf
...
+      {
+	m_stat.peak_stat = m_stat.cur_stat.load ();
+      }
+
+    if (is_expanded)
+      {
+	m_stat.expand_count++;
+      }
+
+    if (comp_idx != mmon_module::MAX_COMP_IDX)
```
wrlawodms, line 350: What if `comp_idx == mmon_module::MAX_COMP_IDX`? If it's impossible, I'd use `assert`.

src/base/memory_monitor_sr.cpp
```diff
@@ -281,61 +315,108 @@ namespace cubperf
...
 		  {
-		    subcomp_idx = m_component[comp_idx]->add_subcomponent (info[cnt].subcomp_name);
+		    subcomp_idx = m_component[comp_idx]->add_subcomponent (info[idx].subcomp_name);
 		  }
-		/* XXX: add error if subcomp_idx > MAX_SUBCOMP_IDX */
+		assert (subcomp_idx < MAX_COMP_IDX);
 	      }
+	    else
+	      {
+		assert (info[idx].subcomp_name == NULL);
```
wrlawodms: What's the point of this assertion? it's determinate.

src/base/memory_monitor_sr.cpp
```diff
@@ -281,61 +315,108 @@ namespace cubperf
...
+	      }
+	  }
+	else
+	  {
+	    assert (info[idx].comp_name == NULL);
```
wrlawodms: What I suggested was `assert (info[idx].subcomp_name == NULL);`.  Do you have a different point?

src/base/memory_monitor_sr.cpp
```diff
@@ -358,25 +439,88 @@ namespace cubperf
...
+    TRANID trid = NULL_TRANID;
+    int module_idx = MMON_GET_MODULE_IDX_FROM_STAT_ID (stat_id);
+    int comp_map_idx = MMON_GET_COMP_MAP_IDX_FROM_STAT_ID (stat_id);
+    int idx_map_val, comp_idx, subcomp_idx;
+    bool allocation_at_init = false;
+
+    if (!LOG_ISRESTARTED ())
+      {
+	allocation_at_init = true;
+      }
```
```suggestion
    bool allocation_at_init = !LOG_ISRESTARTED ();
```

src/base/memory_monitor_sr.cpp
wrlawodms: Is it just for `move_stat`? Or is there a meaningful use case for a user?

src/base/memory_monitor_sr.cpp
```diff
@@ -358,25 +439,88 @@ namespace cubperf
...
+
+    idx_map_val = m_module[module_idx]->get_comp_idx_map_val (comp_map_idx);
+    comp_idx = get_comp_idx_from_comp_idx_map (idx_map_val);
+    subcomp_idx = get_subcomp_idx_from_comp_idx_map (idx_map_val);
+
+    m_total_mem_usage += size;
+
+    m_module[module_idx]->add_stat (size, comp_idx, subcomp_idx, allocation_at_init, is_expanded);
+
+    if (thread_p)
```
joohok: Is it possible for `thread_p` to have a NULL value?
if not, assertion is required for this. 
`assert (thread_p != NULL);`

src/base/memory_monitor_sr.cpp
Rudeus: is_expanded is used to update expand_count. It can be occurred in resize_stat.

src/base/memory_monitor_sr.cpp
joohok: or use `LOG_FIND_THREAD_TRAN_INDEX()`

src/base/memory_monitor_sr.cpp
Rudeus: comp_idx == mmon_module::MAX_COMP_IDX means this stat_id have no mapping component(or subcomponent in mmon_component::add_stat) to update. So if comp_idx == mmon_module::MAX_COMP_IDX, we just do not call mmon_component/subcomponent::add_stat. It's not an assert case.

src/base/memory_monitor_sr.cpp

src/base/memory_monitor_sr.cpp
Rudeus: I get this code lines from logtb_find_tranid(int tran_index). And also many CUBRID codes which call `LOG_FIND_TDES(tran_index)` have this checking code. I think there is some cases that a tran_index has no mapping transaction structure(LOG_TDES).

src/transaction/log_impl.h
Rudeus: Done in `logtb_clear_tdes` and `logtb_initialize_tdes`.

src/base/memory_monitor_sr.cpp
```diff
@@ -393,6 +537,7 @@ namespace cubperf
 /* APIs */
 int mmon_initialize (const char *server_name)
 {
+  int error = NO_ERROR;
   cubperf::mmon_Gl = new cubperf::memory_monitor (server_name);
 
   if (cubperf::mmon_Gl == nullptr)
```
joohok: if you would use variable `error`, then it would be better to set error here. 
`error = ER_...`

src/base/memory_monitor_sr.cpp

src/base/memory_monitor_sr.cpp

src/transaction/log_impl.h
joohok:
```suggestion
// *INDENT-OFF*
  std::atomic < uint64_t > cur_mem_usage;
// *INDENT-ON*
```

src/base/memory_monitor_sr.cpp
Rudeus: If there is some cases that memory allocation with no thread_p parameter in function.

src/transaction/log_impl.h
Rudeus: Done.

src/base/memory_monitor_sr.cpp

src/base/memory_monitor_sr.cpp
```diff
@@ -393,49 +523,42 @@ namespace cubperf
...
+      error = ER_OUT_OF_VIRTUAL_MEMORY;
+      return error;
     }
 
   /* normal return */
-  return 0;
+  return error;
 }
 
 void mmon_finalize ()
```
wrlawodms, line 518: Please call this in the appropriate locations. Now it causes a memory leak.

src/base/memory_monitor_sr.cpp
```diff
@@ -393,49 +523,42 @@ namespace cubperf
...
 int mmon_initialize (const char *server_name)
 {
+  int error = NO_ERROR;
   cubperf::mmon_Gl = new cubperf::memory_monitor (server_name);
 
   if (cubperf::mmon_Gl == nullptr)
     {
-      /* error case */
-      return -1;
+      error = ER_OUT_OF_VIRTUAL_MEMORY;
```
joohok: `er_set` should be called where the error is set.
example is as below :  
```suggestion
      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, sizeof (cubperf::memory_monitor));
      error = ER_OUT_OF_VIRTUAL_MEMORY;
```

src/base/memory_monitor_sr.cpp
```diff
@@ -393,49 +523,42 @@ namespace cubperf
...
+      return error;
```
```suggestion
```

error will be returned at the end.

src/base/memory_monitor_sr.cpp
joohok: What are some situations where it increase or decrease memory usage without thread_p? 
Hmm, if this module is measuring server memory usage... I don't think there would be a situation without thread_p, so I'm wondering about that. Could you explain more about it ?

src/base/memory_monitor_sr.cpp
wrlawodms: Yes, some skip jobs if it's NULL, and some don't assuming it's not null. We should know when it's NULL and decide if we can allow this to happen for this function. I mean we should "allow" this case knowing the context, not just "follows" other code without any background. 
For example, if it's null when it's called out of any transaction, we probably accept the case. (I'm not sure, it's up to you). Of if there is no case it's NULL when it's called, you can just assert().

src/base/memory_monitor_sr.cpp
wrlawodms: I've got it. It's for the { "HEAP", NULL, NULL } case. OK!
I'd leave a comment like `If there is a component info`. at your choice :).

src/base/memory_monitor_sr.cpp
wrlawodms: I can see that it happens in `resize_stat`. What I'm wondering is whether it can't be counted automatically. It seems that every add_stat should be called with `is_expanded = true` except for the move_stat. So, I'd like to know another case that a user can do. I want to relate to the need of this.

src/base/memory_monitor_sr.cpp
Rudeus: Done.

src/base/memory_monitor_sr.cpp
Rudeus: This code means that when `thread_p` is given as a parameter, if there is a transaction associated with that thread, it increases/decreases the memory occupancy of the transaction. The presence of a transaction associated with `thread_p` is not necessary, so I think there seems to be no reason to assert it.

src/transaction/boot_sr.c
```diff
@@ -2360,6 +2361,14 @@ boot_restart_server (THREAD_ENTRY * thread_p, bool print_restart, const char *db
 	}
     }
 
+#if defined(SERVER_MODE)
+  error_code = mmon_initialize (db_name);
```
hornetmj, line 2365: Check whether the "memory_monitoring" is on or off.

+ at mmon_finalize() too.

src/base/memory_monitor_sr.cpp
hornetmj: This module should be define error messages to describe each error case correctly.

src/base/memory_monitor_sr.cpp
Rudeus: It is correct that `is_expanded`, which checks the increase of `expand_count`, should only become conditionally true within the `resize_stat` function. This was a method to increment and decrement the stat within a unified function. However, I agree with your opinion that checking this condition in other operations for the sake of `resize_stat` can be detrimental. Therefore, I have added a separate function to increase expand_count.

src/base/memory_monitor_sr.cpp
hornetmj: And dtor is nothing to do.

~memory_monitor () {}

src/base/memory_monitor_sr.cpp

src/base/memory_monitor_sr.cpp

src/base/memory_monitor_sr.cpp
Rudeus: I've changed it recent commit. Done.

src/base/memory_monitor_sr.cpp
Rudeus: For example, in heap_classrepr_initialize_cache(), there is no THREAD_ENTRY but memory allocation(malloc) occurred. In my opinion, because thread_p parameter is just about transaction memory occupancy, I think it's okay to thread_p == NULL.

src/base/memory_monitor_sr.cpp
```diff
@@ -393,49 +523,42 @@ namespace cubperf
...
   delete cubperf::mmon_Gl;
 }
 
-int mmon_add_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, uint64_t size)
+void mmon_add_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, uint64_t size)
 {
-  int error = NO_ERROR;
-
-  return error;
+  cubperf::mmon_Gl->add_stat (thread_p, stat_id, size, false);
```
hornetmj: If you declare using namespace cubperf in the top of this file, then you can skip the cubperf::. I recommand it.

src/base/memory_monitor_sr.cpp
joohok: Even the function you mentioned can retrieve `thread_p` because it can find the `thread_p` on which it is currently executing. 
Hmm... I have a question. 
Are you planning to track memory usage only for user transactions? If memory usage tracking is not required for system transactions, I agree to allow `thread_p` to have a NULL value. And if `thread_p` is provided, we would need to filter out cases where the `thread_p` represents a system transaction.

src/transaction/log_impl.h
```diff
@@ -602,6 +603,9 @@ struct log_tdes
   log_postpone_cache m_log_postpone_cache;
 
   bool has_supplemental_log;	/* Checks if supplemental log has been appended within the transaction */
+  // *INDENT-OFF*
+  std::atomic < uint64_t > cur_mem_usage;
```
joohok:
```suggestion
  std::atomic<uint64_t> cur_mem_usage;
```

src/base/memory_monitor_sr.cpp
Rudeus: OK. I change LOG_FIND_TDES to LOG_FIND_CURRENT_TDES.

src/base/memory_monitor_sr.cpp
Rudeus: I confused it. I add `mmon_finalize()` at boot / shutdown.

src/base/memory_monitor_sr.cpp
wrlawodms: OK. Fair enough.

src/base/memory_monitor_sr.cpp
```diff
@@ -358,25 +446,82 @@ namespace cubperf
...
+
+    idx_map_val = m_module[module_idx]->get_comp_idx_map_val (comp_map_idx);
+    comp_idx = get_comp_idx_from_comp_idx_map (idx_map_val);
+    subcomp_idx = get_subcomp_idx_from_comp_idx_map (idx_map_val);
+
+    m_total_mem_usage += size;
+
+    m_module[module_idx]->add_stat (size, comp_idx, subcomp_idx, allocation_at_init);
+
+    if (thread_p)
```
```suggestion
```

Then we don't need to check if `thread_p` is NULL or not

src/base/memory_monitor_sr.cpp
```diff
@@ -358,25 +446,82 @@ namespace cubperf
...
+      {
+	tdes = LOG_FIND_CURRENT_TDES (thread_p);
+	tdes->cur_mem_usage += size;
```
joohok: how about using ER_LOG_UNKNOWN_TRANINDEX

src/base/memory_monitor_sr.cpp
```diff
@@ -358,25 +446,82 @@ namespace cubperf
...
+      }
   }
 
-  int memory_monitor::sub_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, uint64_t size)
+  void memory_monitor::sub_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, uint64_t size)
```
wrlawodms: I think `add_stat` and `sub_stat` can be integrated into one function. It has the exact same logic, but the sign.
```
add_stat (thread_p, -, -, 100); // add stat
add_stat (thread_p, -, -, -100); // sub stat
```
If you'd like keep the meaning, we can still do share the code like
```
void memory_monitor::sub_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, uint64_t size)
{
  add_stat (thread_p, stat_id, -size);.
}
```

Then, we loose 1 bit to represent the size, but I don't think it's necessary.
It can be applied to all `add/sub_*` functions.

src/base/memory_monitor_sr.cpp
```diff
@@ -358,25 +446,82 @@ namespace cubperf
...
+    if (new_size - old_size > 0) /* expand */
+      {
+	this->add_expand_count (stat_id);
+      }
+  }
+
+  void memory_monitor::move_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID src, MMON_STAT_ID dest, uint64_t size)
+  {
+    this->sub_stat (thread_p, src, size);
+    this->add_stat (thread_p, dest, size);
```
wrlawodms: Nit. to all others too.
```suggestion
    sub_stat (thread_p, src, size);
    add_stat (thread_p, dest, size);
```

src/base/memory_monitor_sr.cpp
```diff
@@ -202,38 +206,71 @@ namespace cubperf
...
+    if (allocation_at_init)
+      {
+	m_stat.init_stat += size;
+      }
+
+    m_stat.cur_stat += size;
+
+    if (m_stat.cur_stat > m_stat.peak_stat)
+      {
+	m_stat.peak_stat = m_stat.cur_stat.load ();
```
wrlawodms, line 242: Nit. As fas as I know, it's the same.
```suggestion
	m_stat.peak_stat = m_stat.cur_stat;
```

src/base/memory_monitor_sr.cpp
```diff
@@ -142,10 +145,11 @@ namespace cubperf
       memory_monitor &operator = (const memory_monitor &) = delete;
       memory_monitor &operator = (memory_monitor &&) = delete;
 
-      int add_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, uint64_t size, bool expand);
-      int sub_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, uint64_t size);
-      int resize_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, uint64_t old_size, uint64_t new_size);
-      int move_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID src, MMON_STAT_ID dest, uint64_t size);
+      void add_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, uint64_t size);
+      void sub_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, uint64_t size);
+      void add_expand_count (MMON_STAT_ID stat_id);
```
wrlawodms: It has to be private it only happens in `resize_stat()`.

src/base/memory_monitor_sr.cpp
```diff
@@ -86,8 +86,9 @@ namespace cubperf
       mmon_component &operator = (mmon_component &&) = delete;
 
       const char *get_name ();
-      void add_stat (uint64_t size, int subcomp_idx, bool init, bool expand);
-      void sub_stat (uint64_t size, int subcomp_idx, bool init);
+      void add_stat (uint64_t size, int subcomp_idx, bool allocation_at_init);
```
wrlawodms: Nit. to me, `allocation` seems redundant.
```suggestion
      void add_stat (uint64_t size, int subcomp_idx, bool init);
```

src/base/memory_monitor_sr.cpp
```diff
@@ -202,38 +206,71 @@ namespace cubperf
...
     /* description of add_stat(..).
-     * 1) if init == true, add init_stat
+     * 1) if allocation_at_init == true, add init_stat
      * 2) then, add cur_stat
      * 3) compare with peak_stat
      * 4) if cur_stat(new) > peak_stat, update peak_stat
-     * 5) if expand == true, expand_count++
-     * 6) call subcomponent->add_cur_stat(size) */
-    return;
+     * 5) call subcomponent->add_cur_stat(size) */
```
wrlawodms: I think the code explains it well enough. We can remove this comment.

src/base/memory_monitor_sr.cpp
```diff
@@ -358,25 +446,82 @@ namespace cubperf
...
+    LOG_TDES *tdes;
+    TRANID trid = NULL_TRANID;
+    int module_idx = MMON_GET_MODULE_IDX_FROM_STAT_ID (stat_id);
+    int comp_map_idx = MMON_GET_COMP_MAP_IDX_FROM_STAT_ID (stat_id);
+    int idx_map_val, comp_idx, subcomp_idx;
+    bool allocation_at_init = !LOG_ISRESTARTED ();
+
+    idx_map_val = m_module[module_idx]->get_comp_idx_map_val (comp_map_idx);
+    comp_idx = get_comp_idx_from_comp_idx_map (idx_map_val);
+    subcomp_idx = get_subcomp_idx_from_comp_idx_map (idx_map_val);
```
wrlawodms: We need to leave a comment about thess bit operations. Maybe, one month later, we got confused :). Here or somewhere, maybe around the bit operation macros.
like:
```
// purpose : ...
// 32bit MMON_STAT_ID: [module_idx:16 | com_idx_map_idx:16],
// m_comp_idx_map[com_idx_map_idx] -> [comp_idx:16 | subcomp_idx:16 ].
```

Do I get right? I already get a little confused.. lol.

wrlawodms (commented):
probably the last review

src/base/memory_monitor_sr.cpp

src/base/memory_monitor_sr.cpp

src/base/memory_monitor_sr.hpp
```diff
@@ -37,7 +37,7 @@
 #define MMON_PARSE_MASK 0x0000FFFF
 #define MMON_MAKE_INIT_STAT_ID_FOR_MODULE(module_idx) ((module_idx) << 16)
 #define MMON_GET_MODULE_IDX_FROM_STAT_ID(stat) ((stat) >> 16)
-#define MMON_GET_COMP_INFO_IDX_FROM_STAT_ID(stat) ((stat) & MMON_PARSE_MASK)
+#define MMON_GET_COMP_MAP_IDX_FROM_STAT_ID(stat) ((stat) & MMON_PARSE_MASK)
```
hornetmj: 1. Move them to the .cpp file except MMON_MAKE_INIT_STAT_ID_FOR_MODULE
2. How about changing MMON_MAKE_INIT_STAT_ID_FOR_MODULE(module_idx) to MMON_MAKE_STAT_ID(module_id)?
3. from: MMON_GET_MODULE_IDX_FROM_STAT_ID
    - to: MMON_GET_MODULE_ID (stat_id)

src/base/memory_monitor_sr.cpp
Rudeus: You mean making size parameter of add_stat uint64_t to int64_t?

src/base/memory_monitor_sr.cpp

src/base/memory_monitor_sr.cpp
```diff
@@ -393,49 +538,41 @@ namespace cubperf
...
   delete cubperf::mmon_Gl;
 }
 
-int mmon_add_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, uint64_t size)
+void mmon_add_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, uint64_t size)
 {
-  int error = NO_ERROR;
-
-  return error;
+  cubperf::mmon_Gl->add_stat (thread_p, stat_id, size);
```
hornetmj: use 'using namespace cubperf;'. Then you can remove cubperf::. (cubperf::mmon_Gl to mmon_Gl)

src/base/memory_monitor_sr.cpp
Rudeus: It makes error because of atomic::atomic(atomic &)

src/base/memory_monitor_sr.cpp

src/base/memory_monitor_sr.cpp

src/base/memory_monitor_sr.cpp
Rudeus: You are right.
I add a comment for it.

src/base/memory_monitor_sr.cpp
Rudeus: For clear understanding, I want to remain this parameter name.

src/base/memory_monitor_sr.cpp
```diff
@@ -358,25 +446,82 @@ namespace cubperf
     return 0;
   }
 
-  int memory_monitor::add_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, uint64_t size, bool expand)
+  void memory_monitor::add_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, uint64_t size)
   {
-    return 0;
+    LOG_TDES *tdes;
+    TRANID trid = NULL_TRANID;
+    int module_idx = MMON_GET_MODULE_IDX_FROM_STAT_ID (stat_id);
```
hornetmj: 1. avoid using macro and replace it to inline function
2. MMON_GET_MODULE_IDX_FROM_STAT_ID
    - int memory_monitor::get_module_index (MMON_STAT_ID stat_id)
3. idx_map_val, comp_idx, subcomp_idx are not needed in here
    - so, change the mmon_module::add_stat() signiture
        - from (uint64_t size, int comp_idx, int subcomp_idx, bool allocation_at_init)
        - to (MMON_STAT_ID stat_id, uint64_t size, bool allocation_at_init)
    - and get the necessary index in there
 4. remove allocation_at_init mechanishm (I aready mentioned it in the previous PR)
    - introduce a new API function to notify(inform) end of the initialization state to memory_monitor
    - for example
        ```
        int mmon_notify_end_of_boot(void)
        {
            // interate modules and componts
            // copy cur_stat to init_stat
            // etc. something you need to handle
        }
        ```

src/base/memory_monitor_sr.cpp
hornetmj: About tracking the transaction memory usage
1. it would be better introduce a new function for this
    - for example
        ```
        int memory_monitor::add_tran_stat (THREAD_ENTRY *thread_p, unit64_t size);
        {
            // put the code in here
        }
        ```
2. consider this job(transaction memory usage) handling in the upper layer; like
    ```
    mmon_add_stat()
    {
        mmon_Gl->add_stat(); // responsibility for tracking the module memory usages
        mmon_Gl->add_tran_stat(); // responsibility for tracking the transaction memory usages
    }
    ```

src/base/memory_monitor_sr.cpp
```diff
@@ -393,49 +538,41 @@ namespace cubperf
...
 int mmon_initialize (const char *server_name)
 {
+  int error = NO_ERROR;
   cubperf::mmon_Gl = new cubperf::memory_monitor (server_name);
 
   if (cubperf::mmon_Gl == nullptr)
     {
-      /* error case */
-      return -1;
+      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, sizeof (cubperf::memory_monitor));
```
hornetmj: Need to define the error codes for the memory monitor module first. And use them.

src/base/memory_monitor_sr.cpp
```diff
@@ -281,61 +319,111 @@ namespace cubperf
...
+
+    m_stat.cur_stat += size;
+
+    if (m_stat.cur_stat > m_stat.peak_stat)
+      {
+	m_stat.peak_stat = m_stat.cur_stat.load ();
+      }
+
+    /* If there is a component info */
+    if (comp_idx != mmon_module::MAX_COMP_IDX)
```
hornetmj, line 350: Based on the comment in the memory_monitor::add_stat()
1. introduce a new function to get comp and subcomp index
    - for example
        ```
        int mmon_module::get_comp_and_subcomp_id (MMON_STAT_ID stat_id, int &comp_id, int &subcump_id)
        ```
    - with this, I expect to get rid of the MMON_GET_COMP_MAP_IDX_FROM_STAT_ID macro
2. move the condition check into the newly introduced function and if the index is decided to the unexpected range, then reture error code and handle it in this function

hornetmj (commented):
Please consider my feedback. After the discussion is done, I will review again based on that.

src/transaction/boot_sr.c
Rudeus: Done.

src/base/memory_monitor_sr.cpp
Rudeus: New error code for memory monitoring manager module will be set next story - implement aggregate functions -.

src/transaction/boot_sr.c
Rudeus: Done.

src/base/memory_monitor_sr.cpp

src/base/memory_monitor_sr.cpp
hornetmj: 5. based on the definition of the init_stat as I know, the LOG_ISRESTARTED() checking is not enough; because the log_Gl.rcv_phase is set to LOG_RESTARTED in the log_initialize() function stack
    - BOOT_SERVER_UP is the better option for you

src/base/memory_monitor_sr.cpp
wrlawodms: oh, I see.

src/base/memory_monitor_sr.cpp
wrlawodms: yes, with the change.

src/base/memory_monitor_sr.cpp
joohok: Then `BO_IS_SERVER_RESTARTED ()` can be used instead

src/base/memory_monitor_sr.hpp
```diff
@@ -34,10 +34,13 @@
 #include "thread_compat.hpp"
 #include "memory_monitor_common.h"
 
+/* 32-bit MMON_STAT_ID: [module_idx: 16 | comp_idx_map_idx: 16]
+ * 32-bit m_comp_idx_map[comp_idx_map_idx] value: [comp_idx: 16 | subcomp_idx: 16]
```
```suggestion
 * 32-bit mmon_module::m_comp_idx_map[comp_idx_map_idx] value: [comp_idx: 16 | subcomp_idx: 16]
```

src/base/memory_monitor_sr.hpp

src/base/memory_monitor_sr.cpp

src/base/memory_monitor_sr.cpp

src/base/memory_monitor_sr.cpp
Rudeus: Done.

src/base/memory_monitor_sr.cpp

src/base/memory_monitor_sr.cpp

src/base/memory_monitor_sr.cpp
```diff
@@ -391,51 +478,76 @@ namespace cubperf
...
 
 /* APIs */
+// TODO: API parameter check (after error code setting)
+using namespace cubperf;
+
 int mmon_initialize (const char *server_name)
 {
-  cubperf::mmon_Gl = new cubperf::memory_monitor (server_name);
+  int error = NO_ERROR;
+
```
wrlawodms, line 493:
```suggestion
assert (mmon_Gl == nullptr);
```

src/base/memory_monitor_sr.cpp
```diff
@@ -358,25 +421,49 @@ namespace cubperf
...
+    m_module[module_idx]->add_stat (stat_id, size);
   }
 
-  int memory_monitor::sub_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, uint64_t size)
+  void memory_monitor::add_tran_stat (THREAD_ENTRY *thread_p, int64_t size)
   {
-    return 0;
+    LOG_TDES *tdes = LOG_FIND_CURRENT_TDES (thread_p);
+    assert (tdes != NULL);
```
wrlawodms: I think this assert can be removed since if it's NULL the right below line spit a segfault. Nothing undefined happens with the NULL tdes.

src/base/memory_monitor_sr.cpp
```diff
@@ -142,14 +150,16 @@ namespace cubperf
...
+      void resize_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, uint64_t old_size, uint64_t new_size);
+      void move_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID src, MMON_STAT_ID dest, uint64_t size);
+      void end_init_phase ();
       int aggregate_module_info (MMON_MODULE_INFO *info, int module_index);
       int aggregate_tran_info (MMON_TRAN_INFO *info, int tran_count);
-      inline int get_comp_idx_from_comp_idx_map (int comp_idx_map_val);
-      inline int get_subcomp_idx_from_comp_idx_map (int comp_idx_map_val);
+
+    private:
+      inline int get_module_idx (MMON_STAT_ID stat_id);
```
```suggestion
      inline int get_module_idx (MMON_STAT_ID stat_id) const;
```
As many const as possible. It represents immutability explicitly and prevents mistakes. I think many of methods can be const including the `aggregate_*` functions.

src/base/memory_monitor_sr.cpp
```diff
@@ -358,25 +421,49 @@ namespace cubperf
...
-    return 0;
+    LOG_TDES *tdes = LOG_FIND_CURRENT_TDES (thread_p);
+    assert (tdes != NULL);
+    tdes->cur_mem_usage += size;
   }
 
-  int memory_monitor::resize_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, uint64_t old_size,
-				   uint64_t new_size)
+  void memory_monitor::resize_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, uint64_t old_size,
+				    uint64_t new_size)
```
```suggestion
  void memory_monitor::resize_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, int64_t old_size,
				    int64_t new_size)
```
better to use the same type for the `size`s with other functions. You may miss it to change?

src/base/memory_monitor_sr.cpp
```diff
@@ -391,51 +478,76 @@ namespace cubperf
...
+    }
 }
 
-int mmon_add_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, uint64_t size)
+void mmon_add_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, int64_t size)
 {
-  int error = NO_ERROR;
-
-  return error;
+  if (prm_get_bool_value (PRM_ID_MEMORY_MONITORING))
```
```suggestion
  if (mmon_Gl != nullptr)
```
It's less coupled and faster only to use a module-related variable and to check the system parameter only in `mmon_initialize`.

src/base/memory_monitor_sr.cpp
```diff
@@ -200,40 +219,41 @@ namespace cubperf
...
 
-  void mmon_component::sub_stat (uint64_t size, int subcomp_idx, bool init)
+  void mmon_component::end_init_phase ()
   {
-    /* description of sub_stat(size, init).
-     * 1) if init == true, sub init_stat
-     * 2) then, sub cur_stat
-     * 3) call component->sub_stat(size, subcomp_idx, init) */
-    return;
+    m_stat.init_stat = m_stat.cur_stat.load ();
```
wrlawodms, line 261:
```suggestion
    assert (m_stat.init_stat == 0);
    m_stat.init_stat = m_stat.cur_stat.load ();
```

src/base/memory_monitor_sr.cpp
```diff
@@ -281,61 +302,103 @@ namespace cubperf
...
+  void mmon_module::end_init_phase ()
   {
-    /* description of sub_stat(size, init).
-     * 1) if init == true, sub init_stat
-     * 2) then, sub cur_stat
-     * 3) call component->sub_stat(size, subcomp_idx, init) */
-    return;
+    m_stat.init_stat = m_stat.cur_stat.load ();
+
+    for (auto i = 0; i < m_component.size (); i++)
```
wrlawodms: The `for-each` is better if possible. It's open to changing its container and probably faster. 
```suggestion
    for (const &auto comp :  m_component)
```

src/base/memory_monitor_sr.cpp
Rudeus: Done.

src/base/memory_monitor_sr.cpp
Rudeus: Done.

src/base/memory_monitor_sr.cpp

src/base/memory_monitor_sr.cpp

src/base/memory_monitor_sr.cpp
Rudeus: I missed it. Done.

src/base/memory_monitor_sr.cpp

src/base/memory_monitor_sr.cpp

wrlawodms (approved):
LGTM mostly.

All methods that can be const should be const. There seem to be many including:
```
inline int make_comp_idx_map_val (int comp_idx, int subcomp_idx);

     private:
       void get_comp_and_subcomp_idx (MMON_STAT_ID stat_id, int &comp_idx, int &subcomp_idx);
       inline int get_comp_map_idx (MMON_STAT_ID stat_id);
       inline int get_comp_idx (int comp_idx_map_val);
       inline int get_subcomp_idx (int comp_idx_map_val);
```

src/base/memory_monitor_sr.cpp
```diff
@@ -110,16 +117,17 @@ namespace cubperf
...
 
       virtual int aggregate_stats (const MMON_MODULE_INFO &info);
-      void add_stat (uint64_t size, int comp_idx, int subcomp_idx, bool init, bool expand);
-      void sub_stat (uint64_t size, int comp_idx, int subcomp_idx, bool init);
+      void add_stat (MMON_STAT_ID stat_id, int64_t size);
+      void add_expand_count (MMON_STAT_ID stat_id);
+      void end_init_phase ();
       int add_component (const char *name);
-      inline int make_stat_idx_map (int comp_idx, int subcomp_idx);
+      inline int make_comp_idx_map_val (int comp_idx, int subcomp_idx);
```
joohok: can this function be moved to private section?

joohok (approved):

src/base/memory_monitor_sr.cpp

src/base/memory_monitor_sr.hpp
```diff
@@ -34,23 +34,25 @@
 #include "thread_compat.hpp"
 #include "memory_monitor_common.h"
 
+/* 32-bit MMON_STAT_ID: [module_idx: 16 | comp_idx_map_idx: 16]
+ * 32-bit mmon_module::m_comp_idx_map[comp_idx_map_idx] value: [comp_idx: 16 | subcomp_idx: 16]
+ */
 #define MMON_PARSE_MASK 0x0000FFFF
```
hornetmj, line 37: Move it to cpp file because of unused in here.

src/base/memory_monitor_sr.cpp
```diff
@@ -358,25 +424,48 @@ namespace cubperf
...
+    m_total_mem_usage += size;
+
+    m_module[module_idx]->add_stat (stat_id, size);
   }
 
-  int memory_monitor::sub_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, uint64_t size)
+  void memory_monitor::add_tran_stat (THREAD_ENTRY *thread_p, int64_t size)
   {
-    return 0;
+    LOG_TDES *tdes = LOG_FIND_CURRENT_TDES (thread_p);
```
hornetmj, line 436: Is it always valid that the tdes is not null?

src/base/memory_monitor_sr.cpp
```diff
@@ -358,25 +424,48 @@ namespace cubperf
...
+    tdes->cur_mem_usage += size;
```
hornetmj, line 443: and for safety reason ... after update stat(not only this) .. it would be better to check if the stat variable has negative value using assert().

src/base/memory_monitor_sr.cpp
```diff
@@ -252,26 +274,27 @@ namespace cubperf
...
+    while (info[idx].id != MMON_STAT_LAST)
       {
 	bool comp_skip = false;
 	bool subcomp_skip = false;
 	int comp_idx = mmon_module::MAX_COMP_IDX, subcomp_idx = mmon_module::MAX_COMP_IDX;
 
-	if (info[cnt].comp_name)
+	if (info[idx].comp_name)
 	  {
 	    for (size_t i = 0; i < m_component.size (); i++)
```
hornetmj: 1. It would be better use same approach of get_subcomp_index() like mmon_module::get_comp_index()
2. The better recommandation from me is to group this kind of jobs (finding index -> add comp/subcomp) to singe function. (Nit: your choise)
    ```
    mmon_comp::add_subcomponent (...)
    {
        1. check if subcomp is exist or not
        2. if doesn exit
            2-1. add it
    }
    ```

src/base/memory_monitor_sr.cpp
```diff
@@ -281,61 +304,104 @@ namespace cubperf
...
+		assert (subcomp_idx < MAX_COMP_IDX);
 	      }
 	  }
+	else
+	  {
+	    assert (info[idx].subcomp_name == NULL);
+	  }
 
-	m_comp_idx_map.push_back (make_stat_idx_map (comp_idx, subcomp_idx));
+	m_comp_idx_map.push_back (make_comp_idx_map_val (comp_idx, subcomp_idx));
```
hornetmj, line 309: Duplicated infomation is added to this vector when comp_skip == true and subcomp_skip == true condition is met.

src/base/memory_monitor_sr.cpp
```diff
@@ -358,25 +424,48 @@ namespace cubperf
...
-    return 0;
+    int module_idx = get_module_idx (stat_id);
+
+    add_stat (stat_id, - (old_size));
+    add_tran_stat (thread_p, - (old_size));
+    add_stat (stat_id, new_size);
+    add_tran_stat (thread_p, new_size);
+    if (new_size - old_size > 0) /* expand */
+      {
+	m_module[module_idx]->add_expand_count (stat_id);
```
hornetmj: Just sharing my thought about definition of expand.

A number of object(for example, HEAP_STATS_ENTRY) is 2 in a pool(may be bestspace). After some time it is going to 4. I think this is expantion (logical expantion). In this case can we track this using resize_stat() function?

src/base/memory_monitor_sr.hpp
```diff
@@ -34,23 +34,25 @@
...
-#define MMON_MAKE_INIT_STAT_ID_FOR_MODULE(module_idx) ((module_idx) << 16)
-#define MMON_GET_MODULE_IDX_FROM_STAT_ID(stat) ((stat) >> 16)
-#define MMON_GET_COMP_INFO_IDX_FROM_STAT_ID(stat) ((stat) & MMON_PARSE_MASK)
+#define MMON_MAKE_INIT_STAT_ID(module_idx) ((module_idx) << 16)
```
hornetmj:
```suggestion
#define MMON_MAKE_STAT_ID(module_idx) ((module_idx) << 16)
```

hornetmj: There is no 'INIT'_STAT_ID because it never change after setting.

src/base/memory_monitor_sr.cpp
```diff
@@ -180,60 +190,72 @@ namespace cubperf
...
+  void mmon_component::add_stat (int64_t size, int subcomp_idx)
+  {
+    m_stat.cur_stat += size;
+
+    if (m_stat.cur_stat > m_stat.peak_stat)
+      {
+	m_stat.peak_stat = m_stat.cur_stat.load ();
+      }
+
+    /* If there is a component info */
```
```suggestion
    /* If there is a sub-component info */
```

src/base/memory_monitor_sr.cpp
hornetmj: If it is, it would be better NULL check like if (tdes == NULL).

src/base/memory_monitor_sr.cpp

src/base/memory_monitor_sr.hpp
Rudeus: Done.

src/base/memory_monitor_sr.cpp
Rudeus: I accepted the review comment from "joohok" and made the transition to the corresponding function. At that time, there was an assert(tdes != NULL); code, but based on the review comment from "wrlawodms", it was mentioned that removing it would be fine since if tdes is NULL, it would cause a SIGSEGV on the next line anyway. I would like to know your opinion on this matter.

src/base/memory_monitor_sr.cpp
Rudeus: Thank you for reminding me of something I had thought about and forgotten. I have added `assert(size <= cur_stat);` for the case when size is negative, and for transactions, I have added `assert(size <= tdes->cur_mem_usage);`. Especially regarding the transaction `tdes` mentioned in the previous review comment, it seems that if `tdes == NULL`, it would trigger in this part.

src/base/memory_monitor_sr.hpp

src/base/memory_monitor_sr.cpp
Rudeus: In my opinion, "expand" means physical expansion such as `realloc`. In your example, I think resize_stat() will be called when a hash table which manages HEAP_STATS_ENTRY has to grow up because of full of capacity.

src/base/memory_monitor_sr.cpp
Rudeus: I think that case you explain is error case. I add an assert.

src/base/memory_monitor_sr.cpp

hornetmj (approved):

PR #4478, opened 2023-07-07
[CBRD-24881] division operation bug for limit and offset

beyondykk9:
http://jira.cubrid.org/browse/CBRD-24881

According to the manual, CUBRID allows only unsigned integers when expressing limits and offsets in queries. However, in the process of calculating the limit and offset, they are being calculated as NUMERIC types.

We need to change the type of LIMIT and OFFSET to BIGINT because the number of rows represent with BIGINT.

The test case sql/_12_mysql_compatibility/_09_table_related/cases/_016_limit.sql is modified to error case because the number 18446744073709551615 is not a BIGINT.


ctshim (approved):

hgryoo:
18446744073709551615 is the maximum value of BIGINT UNSIGNED. 
In sql/_12_mysql_compatibility/_09_table_related/cases/_016_limit.sql, Previously, the LIMIT clause accepted beyond the value that exceeded the maximum limit for the INTEGER without causing an error. Consequently, the query was executed successfully despite this.
However, It is returning a semantic error. Could you explain why it changes?

hgryoo:
select /*+ recompile */ i,j,k,l,m,n,avg(n) from t where n > 10 and n < 110 group by n having n > 100 limit 2,10;

Here, "limit 2, 0" is rewritten to GROUPBY_NUM() expression.
=> groupby_num()<=cast(? as BIGINT) and groupby_num()>?
```
-select t.i, t.j, t.k, t.l, t.m, t.n, avg(t.n) from t t where (t.n> ?:?  and t.n< ?:? ) group by t.n having groupby_num()<=? and groupby_num()>?
+select t.i, t.j, t.k, t.l, t.m, t.n, avg(t.n) from t t where (t.n> ?:?  and t.n< ?:? ) group by t.n having groupby_num()<=cast(? as BIGINT) and groupby_num()>?
```

Can I know why only one predicate casting to BIGINT?

hyunikn:
To my surprise, "SELECT rownum FROM db_class LIMIT 7/4 OFFSET 1/2;" returns only 1 record before this change, that is, in the develop branch. I cannot figure out why.

beyondykk9:
> 18446744073709551615

In CUBRID, BIGINT max is 9223372036854775807, not 18446744073709551615.
In previous version, the limit and offset was calculated as a NUMERIC type.

beyondykk9:
> groupby_num

The limit 10, 10 is rewritten as an expression, that is, PT_LE (less than) with groupby_num() (PT_GROUPBY_NUM) as an argument and PT_GT (greater than) with SUM as an argument.
PT_GROUPBY_NUM is set as INTEGER and SUM is set as BIGINT in **pt_limit_to_numbering_expr**.
This would be resolved by setting a BIGINT type at groupby_num() (PT_GROUPBY_NUM).

beyondykk9:
> To my surprise, "SELECT rownum FROM db_class LIMIT 7/4 OFFSET 1/2;" returns only 1 record before this change, that is, in the develop branch. I cannot figure out why.

In develop version, the query's predicate is rewritten to "inst_num() > 1/2 and inst_num() <= 1/2 + 7/4". 
Both of 1/2 and 7/4 are applied as NUMERIC and they are calculated to 0.5 and 2.25 each other.

note: please, test the query with **oracle_style_divide** parameter "ON"

hyunikn:
> note: please, test the query with **oracle_style_divide** parameter "ON"

Sorry. I forgot this.

hyunikn (approved):

hgryoo (approved):

PR #4480, opened 2023-07-10
[LETS-772] Send prior lists regardless of connection_handler’s state

wrlawodms:
http://jira.cubrid.org/browse/LETS-772

- Add a sink_hook function using m_conn directly.
- Make m_conn protected.

### DIFF
--- a/src/server/active_tran_server.cpp
+++ b/src/server/active_tran_server.cpp
@@ -187,8 +187,8 @@ active_tran_server::connection_handler::on_connecting ()
 {
   assert (m_prior_sender_sink_hook_func == nullptr);
 
-  m_prior_sender_sink_hook_func = std::bind (&tran_server::connection_handler::push_request, this,
-				  tran_to_page_request::SEND_LOG_PRIOR_LIST, std::placeholders::_1);
+  m_prior_sender_sink_hook_func = std::bind (&active_tran_server::connection_handler::prior_sender_sink_hook, this,
+				  std::placeholders::_1);
   log_Gl.m_prior_sender.add_sink (m_prior_sender_sink_hook_func);
 }
 
@@ -200,6 +200,16 @@ active_tran_server::connection_handler::on_disconnecting ()
       log_Gl.m_prior_sender.remove_sink (m_prior_sender_sink_hook_func);
       m_prior_sender_sink_hook_func = nullptr;
     }
+}
+
+void
+active_tran_server::connection_handler::prior_sender_sink_hook (std::string &&message)
+{
+  assert (message.size () > 0);
+
+  auto slock_conn = std::shared_lock<std::shared_mutex> { m_conn_mtx };
+  m_conn->push (tran_to_page_request::SEND_LOG_PRIOR_LIST, std::move (message));
+
 }
 
 active_tran_server::connection_handler *
--- a/src/server/active_tran_server.cpp
+++ b/src/server/active_tran_server.cpp
@@ -208,8 +208,7 @@ active_tran_server::connection_handler::prior_sender_sink_hook (std::string &&me
   assert (message.size () > 0);
 
   auto slock_conn = std::shared_lock<std::shared_mutex> { m_conn_mtx };
-  m_conn->push (tran_to_page_request::SEND_LOG_PRIOR_LIST, std::move (message));
-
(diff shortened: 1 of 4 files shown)


src/server/active_tran_server.cpp
```diff
@@ -202,6 +202,16 @@ active_tran_server::connection_handler::on_disconnecting ()
     }
 }
 
+void
+active_tran_server::connection_handler::prior_sender_sink_hook (std::string &&message)
```
hornetmj, line 206: How about override tran_server::connection_handler::push_request?

hornetmj (approved):

src/server/active_tran_server.cpp
hornetmj: L Not a good idea.

joohok (approved):

src/server/tran_server.hpp
```diff
@@ -142,6 +142,9 @@ class tran_server
       protected:
 	tran_server &m_ts;
 
+	std::unique_ptr<page_server_conn_t> m_conn;
```
cristiarg: Nit.
Rather than exposing these, expose a protected function to be called from `active_tran_server`. It is "somewhat" safer.

src/server/active_tran_server.cpp
```diff
@@ -202,6 +202,16 @@ active_tran_server::connection_handler::on_disconnecting ()
...
+{
+  assert (message.size () > 0);
```
cristiarg, line 208: Assert that the state is at least `CONNECTING`?
I know it is by design as long as this handler is only registered in `on_connecting`.

cristiarg (approved):

src/server/active_tran_server.cpp
wrlawodms: It's registered on CONNECTING, but this function is called whenever a prior list is sent on `send_list`. Do I get it wrong?

src/server/tran_server.hpp
wrlawodms: Do you mean a getter for it like `const page_server_conn_t* get_internal_conn()`?
or
a function using it like `push_request_internal()`?

src/server/active_tran_server.cpp
cristiarg: I know. That's why I said that it is "by design". It is registered when the state is `CONNECTING` and it executes:
1. during `CONNECTING` - after registration
2. during `CONNECTED` - the whole duration of the state
3. during `DISCONNECTING` - for a very short while as seen in function `connection_handler::disconnect_async`:
```
  assert (m_state == state::CONNECTING || m_state == state::CONNECTED);
  m_state = state::DISCONNECTING; // <<=================== (1)

  m_disconn_future = std::async (std::launch::async, [this, with_disc_msg]
  {
    on_disconnecting (); // server-type specific jobs before disconnect // <<==================== (2)
```
Between (1) and (2) - where the handler is unregistered log prior can still be sent.

My suggestion was to just add an assert to formalize these inter-dependencies towards the state.
```suggestion
  assert (m_state >= state::CONNECTING && m_state <= state::DISCONNECTING);
```
But now, looking back at it, the assert will be reading the state unguarded by mutex. I guess it's fine being only debug code.

I leave it up to you.

src/server/tran_server.hpp
cristiarg: a protected function eg: `push_request_regardless_of_state()` next to the existing `tran_server::connection_handler::push_request`.

Conceptually, if the connection and associated mutex are exposed to the derived code, this opens the gate to more "improvisations" in the future when other developers (or older versions of us :) ) come and need to implement/extend/change stuff.

My suggestion if loosely based from what I still remember from https://en.wikipedia.org/wiki/SOLID . :)

Feel free to ignore any comment of mine that starts with "Nit".

src/server/active_tran_server.cpp
wrlawodms: Thank you for elaborating and I've got it. In this case, it'd be just for debugging but should open the m_state to derived classes. I leave it as it is, but I'll keep in mind the formalization with an assertion.

src/server/tran_server.hpp
wrlawodms: I thought that what's different to the previous `push_request_internal()` which is removed since it has no point, but it's different. Thanks.

PR #4481, opened 2023-07-11
[CBRD-24879] core dump at btree_split_node()

ctshim:
http://jira.cubrid.org/browse/CBRD-24879

* fix core dump at btree_split_node

When a root or leaf node is split, adjustments are made to secure space for new keys to be added after splitting.
However, when a non-leaf node is split, this space is not secured, which causes a problem.

### DIFF
--- a/src/storage/btree.c
+++ b/src/storage/btree.c
@@ -12001,8 +12001,6 @@ btree_find_split_point (THREAD_ENTRY * thread_p, BTID_INT * btid, PAGE_PTR page_
       stop_at--;
     }
 
-
-
   if (node_type == BTREE_LEAF_NODE)
     {
       new_ent_size = btree_get_max_new_data_size (thread_p, btid, page_ptr, node_type, key_len, helper, found);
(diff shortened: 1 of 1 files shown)


hyunikn:
At line 12054, we have "left_min_size = tot_rec - right_max_size;". 
It seems that left_min_size can be smaller than this. In other words, this line should be 
"left_min_size = tot_rec - new_ent_size - right_max_size;" 
as the newly inserted key can go to the right split.

ctshim:
@hyunikn
   It doesn't seem necessary.
  Above, right_max_size gets adjusted.
  Also left_min_size is not sensitive.

hyunikn (approved):

hgryoo:
In develop branch, I couldn't reproduce the scenario written in JIRA.
Instead, I've faced the following error. 
```
...
1 row affected. (0.000391 sec) 
1 row affected. (0.000327 sec) 
1 row affected. (0.000452 sec) 
1 row affected. (0.000791 sec)

In the command from line 10200,

ERROR: Execute: Query execution failure #11637.
```

hgryoo (approved):

beyondykk9 (approved):

PR #4483, opened 2023-07-11
[CBRD-24883] combine with oracle_style_number_return parameter

beyondykk9:
http://jira.cubrid.org/browse/CBRD-24883

It needs to merge two system parameters **oracle_style_number_return** and **oracle_style_divide** into one oracle_style_number_return.

There is no need to maintain two oracle-style number-related parameters. That is, the two parameters **oracle_style_divide** and **oracle_style_number_return** are unlikely to be used separately, so it is better to combine them into one.


beyondykk9:
I will reopen the issue to refine codes.

PR #4484, opened 2023-07-11
[CBRD-24883] combine with oracle_style_number_return parameter

beyondykk9:
http://jira.cubrid.org/browse/CBRD-24883

It needs to merge two system parameters oracle_style_number_return and oracle_style_divide into one oracle_style_number_return.

There is no need to maintain two oracle-style number-related parameters. That is, the two parameters oracle_style_divide and oracle_style_number_return are unlikely to be used separately, so it is better to combine them into one.


ctshim (approved):

hgryoo (approved):

hyunikn (approved):

hgryoo:
We need to discuss renaming "oracle_style_number_return" to "oracle_compat_number_behavior" (just an example) due to the parameter's changing effects when combined with other parameters.

hgryoo (changes requested):
Also, please update the JIRA about Specification changes and A/C

beyondykk9:
> We need to discuss renaming "oracle_style_number_return" to "oracle_compat_number_behavior" (just an example) due to the parameter's changing effects when combined with other parameters.

I applied the parameter as your opinion.

hgryoo (approved):

PR #4486, opened 2023-07-12
[CBRD-24877] Shard Invalid argument error occurs only in debug build.

airnet73:
http://jira.cubrid.org/browse/CBRD-24877

Purpose
When copying the string "shard_id(xxx)*/" to an array, overflow occurred and the value of the "error" variable was changed.

Implementation
Changing the size and initialization of string arrays, and string copying methods

Remarks
N/A


kisoo-han (approved):
very nice!

PR #4488, opened 2023-07-17
[CBRD-24867] rev: regression bug-fix related to DBLINK - alias check, DELETE query check

beyondykk9:

In **pt_check_dblink_column_alias**, the **strcmp** should be replaced with **strcasecmp**.

Changed:
The pt_dblink_table_get_column_defs is called from two cases:
- SELECT ... FROM table-name@server-name
- SELECT ... FROM DBLINK(server-name, 'SELECT ... FROM ... ') ...
In first case, pt_dblink_table_get_column_defs uses the query "SELECT * FROM table-name" through cci_prepare
In second case, it uses the user's query 'SELECT ... FROM ...' in defined at DBLINK clause.

1) for alias check
   *  I added a REMOTE_COLS dblink_remote meber to PARSER_CONTEXT to keep remote columns (S_REMOTE_TBL_COLS) of the tables
      the remot columns kept in **dblink_remote** is freeed at **parser_free_parser** by calling **pt_free_dblink_remote_cols**
   * **pt_check_dblink_column_alias** is called from pt_copypush_terms for dblink query. 
 
      In case of predicate pushing, it needs to check that the column aliases are the same as column definitions.
```
     SELECT t.a, bb 
     FROM dblink(srv1, 'select a, b || ''abc'' as p from t1') t(a int, bb varchar)  JOIN tbl ON t.a = tbl.a 
     WHERE bb > ''
```
     This query is written for remote server (Oracle or MySQL or etc) as below:
`     SELECT * FROM (select a, b || 'abc' as p from t1) cublink WHERE bb > ''`
      However, the bb is not matched "p" in the subquery. The query results error in Oracle.

   * pt_dblink_table_get_column_defs modified
      separating user-name and table-name from "table_name"
      make upper case for user-name and table-name
2) for other DBMS's DELETE query
    the cubrid syntax is "DELETE target-name FROM target, ... WHERE ..." but the other DBMS like Oracle has syntax as like "DELETE FROM target ... WHERE ... "
    So, I added PT_PRINT_SUPPRESS_DELETE_TARGET


ctshim (approved):

src/parser/name_resolution.c
```diff
@@ -1203,19 +1203,25 @@ pt_bind_scope (PARSER_CONTEXT * parser, PT_BIND_NAMES_ARG * bind_arg)
...
+		  S_REMOTE_TBL_COLS *rmt_tbl_cols;
+		  REMOTE_COLS *remote;
+		  PT_DBLINK_INFO *dblink_table = &table->info.dblink_table;
+
 		  if (pt_resolve_dblink_server_name (parser, table, NULL) != NO_ERROR)
 		    {
 		      return;
 		    }
+
+		  rmt_tbl_cols = new S_REMOTE_TBL_COLS;
```
ctshim:
```suggestion
		  rmt_tbl_cols = new (std::nothrow)S_REMOTE_TBL_COLS;
```

ctshim: assert_release(rmt_tbl_cols != NULL)   ?

src/parser/name_resolution.c
```diff
@@ -5125,7 +5151,29 @@ pt_dblink_table_get_column_defs (PARSER_CONTEXT * parser, PT_NODE * dblink, S_RE
...
+      char user_name[256];
+
+      /* make upper case for Oracle */
+      for (i = 0; i < len; i++)
+	{
+	  if (table_name[i] >= 'a' && table_name[i] <= 'z')
+	    {
+	      table_name[i] -= ('a' - 'A');
+	    }
+	}
```
ctshim: As a result of this operation, if the string contained in dblink_table->remote_table_name is changed to uppercase, does it have no effect?

beyondykk9: It doesn't matter. The table name and user name is case insensitive.

src/parser/name_resolution.c
```diff
@@ -5125,7 +5151,29 @@ pt_dblink_table_get_column_defs (PARSER_CONTEXT * parser, PT_NODE * dblink, S_RE
...
-      sprintf (sql_text, "SELECT * FROM %s", table_name);
+      int len = strlen (table_name);
+      char user_name[256];
+
+      /* make upper case for Oracle */
+      for (i = 0; i < len; i++)
+	{
+	  if (table_name[i] >= 'a' && table_name[i] <= 'z')
+	    {
+	      table_name[i] -= ('a' - 'A');
```
ctshim: In pt_check_dblink_column_alias(), intl_identifier_casecmp() was used.
Wouldn't it be a problem to switch only the English character area?

Instead of handling the strchr() in the code below, I think you can get the "find" variable here.

PR #4489, opened 2023-07-17
[CBRD-24878] Implement aggregate functions for Memory Monitoring Manager Module

Rudeus:
http://jira.cubrid.org/browse/CBRD-24878

Implement aggregating functions

**Purpose:**
- The aggregating functions are implemented with the purpose of collecting and delivering memory usage information of registered modules that the Memory Monitoring Manager module has been monitoring, to the user upon their request.

**Additional Info:**
- Add mmon_aggregate_module_info_summary(..) and sub-functions for "brief" and "default" case of -m option

### DIFF
--- a/src/base/memory_monitor_cl.hpp
+++ b/src/base/memory_monitor_cl.hpp
@@ -28,9 +28,9 @@
 
 typedef enum mmon_module_option
 {
-  MMON_MODULE_DETAIL,
-  MMON_MODULE_BRIEF,
-  MMON_MODULE_DEFAULT
+  MMON_MODULE_DETAIL,   /* -m <module_index> or <module_name> */
+  MMON_MODULE_BRIEF,    /* -m brief */
+  MMON_MODULE_DEFAULT   /* default case (with no command option) */
 } MMON_MODULE_OPTION;
 
 constexpr char *module_names[] =
--- a/src/base/memory_monitor_cl.hpp
+++ b/src/base/memory_monitor_cl.hpp
@@ -26,6 +26,13 @@
 
 #include "memory_monitor_common.h"
 
+typedef enum mmon_module_option
+{
+  MMON_MODULE_DETAIL,
+  MMON_MODULE_BRIEF,
+  MMON_MODULE_DEFAULT
+} MMON_MODULE_OPTION;
+
 constexpr char *module_names[] =
 {
   ""				/* It will be changed to "heap" */
--- a/src/base/memory_monitor_cl.hpp
+++ b/src/base/memory_monitor_cl.hpp
@@ -26,13 +26,6 @@
 
 #include "memory_monitor_common.h"
 
-typedef enum mmon_module_option
-{
-  MMON_MODULE_DETAIL,   /* -m <module_index> or <module_name> */
-  MMON_MODULE_BRIEF,    /* -m brief */
-  MMON_MODULE_DEFAULT   /* default case (with no command option) */
-} MMON_MODULE_OPTION;
-
 constexpr char *module_names[] =
 {
   ""				/* It will be changed to "heap" */
--- a/src/base/memory_monitor_common.h
+++ b/src/base/memory_monitor_common.h
@@ -26,12 +26,6 @@
 
 #include <cstdint>
 
-typedef enum mmon_module_option
-{
-  MMON_MODULE_BRIEF_OPTION,
-  MMON_MODULE_DEFAULT_OPTION
-} MMON_MODULE_OPTION;
-
 typedef enum mmon_module_id
 {
   //MMON_MODULE_HEAP = 1,
@@ -68,7 +62,6 @@ typedef struct mmon_comp_info
 
 typedef struct mmon_module_info
 {
-  MMON_SERVER_INFO server_info;
   char name[DB_MAX_IDENTIFIER_LENGTH];
   MMON_OUTPUT_MEM_STAT stat;
   uint32_t num_comp;
@@ -83,7 +76,6 @@ typedef struct mmon_tran_stat
 
 typedef struct mmon_tran_info
 {
-  MMON_SERVER_INFO server_info;
   uint32_t num_tran;
   MMON_TRAN_STAT *tran_stat;
 } MMON_TRAN_INFO;
--- a/src/base/memory_monitor_common.h
+++ b/src/base/memory_monitor_common.h
@@ -25,6 +25,7 @@
 #include "dbtype_def.h"
 
 #include <cstdint>
+#include <vector>
 
 typedef enum mmon_module_id
 {
@@ -57,7 +58,7 @@ typedef struct mmon_comp_info
   char name[DB_MAX_IDENTIFIER_LENGTH];
   MMON_OUTPUT_MEM_STAT stat;
   uint32_t num_subcomp;
-  MMON_SUBCOMP_INFO *subcomp_info;
+    std::vector < MMON_SUBCOMP_INFO > subcomp_info;
 } MMON_COMP_INFO;
 
 typedef struct mmon_module_info
@@ -65,7 +66,7 @@ typedef struct mmon_module_info
   char name[DB_MAX_IDENTIFIER_LENGTH];
   MMON_OUTPUT_MEM_STAT stat;
   uint32_t num_comp;
-  MMON_COMP_INFO *comp_info;
+    std::vector < MMON_COMP_INFO > comp_info;
 } MMON_MODULE_INFO;
 
 typedef struct mmon_tran_stat
@@ -77,7 +78,7 @@ typedef struct mmon_tran_stat
 typedef struct mmon_tran_info
 {
   uint32_t num_tran;
-  MMON_TRAN_STAT *tran_stat;
+    std::vector < MMON_TRAN_STAT > tran_stat;
 } MMON_TRAN_INFO;
 
 #endif // _MEMORY_MONITOR_COMMON_H_
--- a/src/base/memory_monitor_common.h
+++ b/src/base/memory_monitor_common.h
@@ -58,7 +58,9 @@ typedef struct mmon_comp_info
   char name[DB_MAX_IDENTIFIER_LENGTH];
   MMON_OUTPUT_MEM_STAT stat;
   uint32_t num_subcomp;
-    std::vector < MMON_SUBCOMP_INFO > subcomp_info;
+  // *INDENT-OFF*
+  std::vector < MMON_SUBCOMP_INFO > subcomp_info;
+  // *INDENT-ON*
 } MMON_COMP_INFO;
 
 typedef struct mmon_module_info
@@ -66,7 +68,9 @@ typedef struct mmon_module_info
   char name[DB_MAX_IDENTIFIER_LENGTH];
   MMON_OUTPUT_MEM_STAT stat;
   uint32_t num_comp;
-    std::vector < MMON_COMP_INFO > comp_info;
+  // *INDENT-OFF*
+  std::vector < MMON_COMP_INFO > comp_info;
+  // *INDENT-ON*
 } MMON_MODULE_INFO;
 
 typedef struct mmon_tran_stat
@@ -78,7 +82,9 @@ typedef struct mmon_tran_stat
 typedef struct mmon_tran_info
 {
   uint32_t num_tran;
-    std::vector < MMON_TRAN_STAT > tran_stat;
+  // *INDENT-OFF*
+  std::vector < MMON_TRAN_STAT > tran_stat;
+  // *INDENT-ON*
 } MMON_TRAN_INFO;
 
 #endif // _MEMORY_MONITOR_COMMON_H_
--- a/src/base/memory_monitor_common.h
+++ b/src/base/memory_monitor_common.h
@@ -26,8 +26,11 @@
 
 #include <cstdint>
 
-#define MMON_MODULE_BRIEF_OPTION 1
-#define MMON_MODULE_DEFAULT_OPTION 2
+typedef enum mmon_module_option
+{
+  MMON_MODULE_BRIEF_OPTION,
+  MMON_MODULE_DEFAULT_OPTION
+} MMON_MODULE_OPTION;
 
 typedef enum mmon_module_id
 {
--- a/src/base/memory_monitor_common.h
+++ b/src/base/memory_monitor_common.h
@@ -58,7 +58,9 @@ typedef struct mmon_comp_info
   char name[DB_MAX_IDENTIFIER_LENGTH];
   MMON_OUTPUT_MEM_STAT stat;
   uint32_t num_subcomp;
-    std::vector < MMON_SUBCOMP_INFO > subcomp_info;
+  // *INDENT-OFF*
+  std::vector<MMON_SUBCOMP_INFO> subcomp_info;
+  // *INDENT-ON*
 } MMON_COMP_INFO;
 
 typedef struct mmon_module_info
@@ -66,7 +68,9 @@ typedef struct mmon_module_info
   char name[DB_MAX_IDENTIFIER_LENGTH];
   MMON_OUTPUT_MEM_STAT stat;
   uint32_t num_comp;
-    std::vector < MMON_COMP_INFO > comp_info;
+  // *INDENT-OFF*
+  std::vector<MMON_COMP_INFO> comp_info;
+  // *INDENT-ON*
 } MMON_MODULE_INFO;
 
 typedef struct mmon_tran_stat
@@ -78,7 +82,9 @@ typedef struct mmon_tran_stat
 typedef struct mmon_tran_info
 {
   uint32_t num_tran;
-    std::vector < MMON_TRAN_STAT > tran_stat;
+  // *INDENT-OFF*
+  std::vector<MMON_TRAN_STAT> tran_stat;
+  // *INDENT-ON*
 } MMON_TRAN_INFO;
 
 #endif // _MEMORY_MONITOR_COMMON_H_
--- a/src/base/memory_monitor_common.h
+++ b/src/base/memory_monitor_common.h
@@ -26,6 +26,9 @@
 
 #include <cstdint>
 
+#define MMON_MODULE_BRIEF_OPTION 1
+#define MMON_MODULE_DEFAULT_OPTION 2
+
 typedef enum mmon_module_id
 {
   //MMON_MODULE_HEAP = 1,
--- a/src/base/memory_monitor_common.h
+++ b/src/base/memory_monitor_common.h
@@ -37,7 +37,7 @@ typedef struct mmon_output_mem_stat
   uint64_t init_stat;
   uint64_t cur_stat;
   uint64_t peak_stat;
-  uint32_t expand_count;
+  uint32_t expand_resize_count;
 } MMON_OUTPUT_MEM_STAT;
 
 typedef struct mmon_server_info
--- a/src/base/memory_monitor_sr.cpp
+++ b/src/base/memory_monitor_sr.cpp
@@ -87,7 +87,7 @@ namespace cubperf
       mmon_component &operator = (mmon_component &&) = delete;
 
       const char *get_name ();
-      void add_stat (int64_t size, int subcomp_idx);
+      void add_stat (int subcomp_idx, int64_t size);
       void add_expand_resize_count ();
       void end_init_phase ();
       int get_stat (MMON_COMP_INFO &info) const;
@@ -242,7 +242,7 @@ namespace cubperf
     return m_comp_name.c_str ();
   }
 
-  void mmon_component::add_stat (int64_t size, int subcomp_idx)
+  void mmon_component::add_stat (int subcomp_idx, int64_t size)
   {
     assert ((size >= 0) || ((uint64_t) (-size) <= m_stat.cur_stat));
 
@@ -369,32 +369,33 @@ namespace cubperf
 	info.stat.cur_stat = m_stat.cur_stat.load ();
 	info.stat.peak_stat = m_stat.peak_stat.load ();
 	info.stat.expand_resize_count = m_stat.expand_resize_count.load ();
-      }
 
-    info.num_comp = m_component.size ();
-    if (info.num_comp != 0)
-      {
-	info.comp_info = (MMON_COMP_INFO *)malloc (sizeof (MMON_COMP_INFO) * info.num_comp);
 
-	if (info.comp_info == NULL)
+	info.num_comp = m_component.size ();
+	if (info.num_comp != 0)
 	  {
-	    er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, sizeof (MMON_COMP_INFO) * info.num_comp);
-	    error =  ER_OUT_OF_VIRTUAL_MEMORY;
+	    info.comp_info = (MMON_COMP_INFO *)malloc (sizeof (MMON_COMP_INFO) * info.num_comp);
+
+	    if (info.comp_info == NULL)
+	      {
+		er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, sizeof (MMON_COMP_INFO) * info.num_comp);
+		error =  ER_OUT_OF_VIRTUAL_MEMORY;
+	      }
+
+	    for (uint32_t i = 0; i < info.num_comp; i++)
+	      {
+		error = m_component[i]->get_stat (info.comp_info[i]);
+	      }
 	  }
 
-	for (uint32_t i = 0; i < info.num_comp; i++)
+	if (error == ER_OUT_OF_VIRTUAL_MEMORY)
 	  {
-	    error = m_component[i]->get_stat (info.comp_info[i]);
+	    for (uint32_t i = 0; i < info.num_comp; i++)
+	      {
+		free_and_init (info.comp_info[i].subcomp_info);
+	      }
+	    free_and_init (info.comp_info);
 	  }
-      }
-
-    if (error == ER_OUT_OF_VIRTUAL_MEMORY)
-      {
-	for (uint32_t i = 0; i < info.num_comp; i++)
-	  {
-	    free_and_init (info.comp_info[i].subcomp_info);
-	  }
-	free_and_init (info.comp_info);
       }
 
     return error;
@@ -432,7 +433,7 @@ namespace cubperf
     if (comp_idx != mmon_module::MAX_COMP_IDX)
       {
 	assert ((size_t)comp_idx < m_component.size ());
-	m_component[comp_idx]->add_stat (size, subcomp_idx);
+	m_component[comp_idx]->add_stat (subcomp_idx, size);
       }
   }
 
--- a/src/base/memory_monitor_sr.cpp
+++ b/src/base/memory_monitor_sr.cpp
@@ -391,11 +391,6 @@ namespace cubperf
 
 		    if (error != NO_ERROR)
 		      {
-			for (uint32_t j = 0; j < i; j++)
-			  {
-			    free_and_init (info.comp_info[j].subcomp_info);
-			  }
-			free_and_init (info.comp_info);
 			break;
 		      }
 		  }
--- a/src/base/memory_monitor_sr.cpp
+++ b/src/base/memory_monitor_sr.cpp
@@ -157,7 +157,7 @@ namespace cubperf
       void end_init_phase ();
       void aggregate_server_info (MMON_SERVER_INFO &info) const;
       void aggregate_module_info (MMON_MODULE_INFO *&info, int module_index) const;
-      void aggregate_module_info_summary (MMON_MODULE_INFO *&info, bool sorted_by_mem_usage) const;
+      void aggregate_module_info_summary (MMON_MODULE_INFO *&info) const;
       void aggregate_tran_info (MMON_TRAN_INFO &info, int tran_count) const;
 
     private:
@@ -178,7 +178,7 @@ namespace cubperf
 
 	  void get_server_info (MMON_SERVER_INFO &info) const;
 	  void get_module_info (MMON_MODULE_INFO *&info, int module_index) const;
-	  void get_module_info_summary (MMON_MODULE_INFO *&info, bool sorted_by_mem_usage) const;
+	  void get_module_info_summary (MMON_MODULE_INFO *&info) const;
 	  void get_transaction_info (MMON_TRAN_INFO &info, int tran_count) const;
 
 	private:
@@ -467,6 +467,12 @@ namespace cubperf
 	  {
 	    m_mmon->m_module[idx]->aggregate_stats (info[idx], false);
 	  }
+
+	const auto &comp = [] (const auto& module_info1, const auto& module_info2)
+	{
+	  return module_info1.stat.cur_stat > module_info2.stat.cur_stat;
+	};
+	std::sort (info, info + MMON_MODULE_LAST, comp);
       }
     else
       {
@@ -474,21 +480,18 @@ namespace cubperf
       }
   }
 
-  void memory_monitor::aggregater::get_module_info_summary (MMON_MODULE_INFO *&info, bool sorted_by_mem_usage) const
+  void memory_monitor::aggregater::get_module_info_summary (MMON_MODULE_INFO *&info) const
   {
     for (int i = 0; i < MMON_MODULE_LAST; i++)
       {
 	m_mmon->m_module[i]->aggregate_stats (info[i], true);
       }
 
-    if (sorted_by_mem_usage)
-      {
-	const auto &comp = [] (const auto& module_info1, const auto& module_info2)
-	{
-	  return module_info1.stat.cur_stat > module_info2.stat.cur_stat;
-	};
-	std::sort (info, info + MMON_MODULE_LAST, comp);
-      }
+    const auto &comp = [] (const auto& module_info1, const auto& module_info2)
+    {
+      return module_info1.stat.cur_stat > module_info2.stat.cur_stat;
+    };
+    std::sort (info, info + MMON_MODULE_LAST, comp);
   }
 
   void memory_monitor::aggregater::get_transaction_info (MMON_TRAN_INFO &info, int tran_count) const
@@ -505,14 +508,14 @@ namespace cubperf
     else
       {
 	info.num_tran = std::min (tran_count, (int) tran_info.size ());
-
-	// sort by memory usage in descending order
-	const auto &comp = [] (const auto& tran_pair1, const auto& tran_pair2)
-	{
-	  return tran_pair1.second > tran_pair2.second;
-	};
-	std::sort (tran_info.begin (), tran_info.end (), comp);
       }
+
+    // sort by memory usage in descending order
+    const auto &comp = [] (const auto& tran_pair1, const auto& tran_pair2)
+    {
+      return tran_pair1.second > tran_pair2.second;
+    };
+    std::sort (tran_info.begin (), tran_info.end (), comp);
 
     info.tran_stat.resize (info.num_tran);
 
@@ -585,9 +588,9 @@ namespace cubperf
     m_aggregater.get_module_info (info, module_index - 1);
   }
 
-  void memory_monitor::aggregate_module_info_summary (MMON_MODULE_INFO *&info, bool sorted_by_mem_usage) const
+  void memory_monitor::aggregate_module_info_summary (MMON_MODULE_INFO *&info) const
   {
-    m_aggregater.get_module_info_summary (info, sorted_by_mem_usage);
+    m_aggregater.get_module_info_summary (info);
   }
 
   void memory_monitor::aggregate_tran_info (MMON_TRAN_INFO &info, int tran_count) const
@@ -684,11 +687,11 @@ void mmon_aggregate_module_info (MMON_MODULE_INFO *&info, int module_index)
   mmon_Gl->aggregate_module_info (info, module_index);
 }
 
-void mmon_aggregate_module_info_summary (MMON_MODULE_INFO *&info, bool sorted_by_mem_usage)
+void mmon_aggregate_module_info_summary (MMON_MODULE_INFO *&info)
 {
   assert (mmon_Gl != nullptr);
 
-  mmon_Gl->aggregate_module_info_summary (info, sorted_by_mem_usage);
+  mmon_Gl->aggregate_module_info_summary (info);
 }
 
 void mmon_aggregate_tran_info (MMON_TRAN_INFO &info, int tran_count)
--- a/src/base/memory_monitor_sr.cpp
+++ b/src/base/memory_monitor_sr.cpp
@@ -498,13 +498,12 @@ namespace cubperf
 
   int memory_monitor::aggregater::get_module_info (MMON_MODULE_INFO *&info, int module_index) const
   {
-    int idx = 0;
     int error = NO_ERROR;
 
     if (module_index == -1)
       {
 	// aggregate all detail memory information of modules
-	for (idx; idx < MMON_MODULE_LAST; idx++)
+	for (int idx = 0; idx < MMON_MODULE_LAST; idx++)
 	  {
 	    error = m_mmon->m_module[idx]->aggregate_stats (info[idx], false);
 	    if (error != NO_ERROR)
@@ -515,7 +514,7 @@ namespace cubperf
       }
     else
       {
-	error =  m_mmon->m_module[module_index]->aggregate_stats (info[idx], false);
+	error =  m_mmon->m_module[module_index]->aggregate_stats (info[0], false);
       }
 
     return error;
--- a/src/base/memory_monitor_sr.cpp
+++ b/src/base/memory_monitor_sr.cpp
@@ -564,7 +564,7 @@ namespace cubperf
       {
 	info.num_tran = std::min (tran_count, (int) tran_info.size ());
 
-	// sort memory usage in descending order
+	// sort by memory usage in descending order
 	const auto &comp = [] (const auto& tran_pair1, const auto& tran_pair2)
 	{
 	  return tran_pair1.second > tran_pair2.second;
--- a/src/base/memory_monitor_sr.cpp
+++ b/src/base/memory_monitor_sr.cpp
@@ -391,9 +391,9 @@ namespace cubperf
 
 		    if (error != NO_ERROR)
 		      {
-			for (uint32_t i = 0; i < info.num_comp; i++)
+			for (uint32_t j = 0; j < i; j++)
 			  {
-			    free_and_init (info.comp_info[i].subcomp_info);
+			    free_and_init (info.comp_info[j].subcomp_info);
 			  }
 			free_and_init (info.comp_info);
 			break;
--- a/src/base/memory_monitor_sr.cpp
+++ b/src/base/memory_monitor_sr.cpp
@@ -116,7 +116,7 @@ namespace cubperf
       mmon_module &operator = (const mmon_module &) = delete;
       mmon_module &operator = (mmon_module &&) = delete;
 
-      virtual void aggregate_stats (MMON_MODULE_INFO &info, bool get_summary) const;
+      virtual void aggregate_stats (bool get_summary, MMON_MODULE_INFO &info) const;
       void add_stat (MMON_STAT_ID stat_id, int64_t size);
       void add_expand_resize_count (MMON_STAT_ID stat_id);
       void end_init_phase ();
@@ -156,9 +156,9 @@ namespace cubperf
       void move_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID src, MMON_STAT_ID dest, int64_t size);
       void end_init_phase ();
       void aggregate_server_info (MMON_SERVER_INFO &info) const;
-      void aggregate_module_info (MMON_MODULE_INFO *&info, int module_index) const;
-      void aggregate_module_info_summary (MMON_MODULE_INFO *&info) const;
-      void aggregate_tran_info (MMON_TRAN_INFO &info, int tran_count) const;
+      void aggregate_module_info (int module_index, MMON_MODULE_INFO *info) const;
+      void aggregate_module_info_summary (MMON_MODULE_INFO *info) const;
+      void aggregate_tran_info (int tran_count, MMON_TRAN_INFO &info) const;
 
     private:
       inline int get_module_idx (MMON_STAT_ID stat_id) const;
@@ -177,9 +177,9 @@ namespace cubperf
 	  aggregater &operator = (aggregater &&) = delete;
 
 	  void get_server_info (MMON_SERVER_INFO &info) const;
-	  void get_module_info (MMON_MODULE_INFO *&info, int module_index) const;
-	  void get_module_info_summary (MMON_MODULE_INFO *&info) const;
-	  void get_transaction_info (MMON_TRAN_INFO &info, int tran_count) const;
+	  void get_module_info (int module_index, MMON_MODULE_INFO *info) const;
+	  void get_module_info_summary (MMON_MODULE_INFO *info) const;
+	  void get_transaction_info (int tran_count, MMON_TRAN_INFO &info) const;
 
 	private:
 	  memory_monitor *m_mmon;
@@ -340,7 +340,7 @@ namespace cubperf
       }
   }
 
-  void mmon_module::aggregate_stats (MMON_MODULE_INFO &info, bool get_summary) const
+  void mmon_module::aggregate_stats (bool get_summary, MMON_MODULE_INFO &info) const
   {
     int error = NO_ERROR;
 
@@ -458,14 +458,14 @@ namespace cubperf
     info.total_mem_usage = m_mmon->m_total_mem_usage.load ();
   }
 
-  void memory_monitor::aggregater::get_module_info (MMON_MODULE_INFO *&info, int module_index) const
+  void memory_monitor::aggregater::get_module_info (int module_index, MMON_MODULE_INFO *info) const
   {
     if (module_index == -1)
       {
 	// aggregate all detail memory information of modules
 	for (int idx = 0; idx < MMON_MODULE_LAST; idx++)
 	  {
-	    m_mmon->m_module[idx]->aggregate_stats (info[idx], false);
+	    m_mmon->m_module[idx]->aggregate_stats (false, info[idx]);
 	  }
 
 	const auto &comp = [] (const auto& module_info1, const auto& module_info2)
@@ -476,15 +476,15 @@ namespace cubperf
       }
     else
       {
-	m_mmon->m_module[module_index]->aggregate_stats (info[0], false);
+	m_mmon->m_module[module_index]->aggregate_stats (false, info[0]);
       }
   }
 
-  void memory_monitor::aggregater::get_module_info_summary (MMON_MODULE_INFO *&info) const
+  void memory_monitor::aggregater::get_module_info_summary (MMON_MODULE_INFO *info) const
   {
     for (int i = 0; i < MMON_MODULE_LAST; i++)
       {
-	m_mmon->m_module[i]->aggregate_stats (info[i], true);
+	m_mmon->m_module[i]->aggregate_stats (true, info[i]);
       }
 
     const auto &comp = [] (const auto& module_info1, const auto& module_info2)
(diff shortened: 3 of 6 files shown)


src/base/memory_monitor_sr.cpp
```diff
@@ -261,6 +286,35 @@ namespace cubperf
...
+
+    info.stat.init_stat = m_stat.init_stat.load ();
+    info.stat.cur_stat = m_stat.cur_stat.load ();
+    info.stat.peak_stat = m_stat.peak_stat.load ();
+    info.stat.expand_resize_count = m_stat.expand_resize_count.load ();
+
+    info.num_subcomp = m_subcomponent.size ();
+    if (info.num_subcomp != 0)
+      {
+	info.subcomp_info = (MMON_SUBCOMP_INFO *)malloc (sizeof (MMON_SUBCOMP_INFO) * info.num_subcomp);
```
joohok: where it is freed?

src/base/memory_monitor_sr.cpp
```diff
@@ -312,10 +366,33 @@ namespace cubperf
...
+
+    info.stat.init_stat = m_stat.init_stat.load ();
+    info.stat.cur_stat = m_stat.cur_stat.load ();
+    info.stat.peak_stat = m_stat.peak_stat.load ();
+    info.stat.expand_resize_count = m_stat.expand_resize_count.load ();
+
+    info.num_comp = m_component.size ();
+    if (info.num_comp != 0)
+      {
+	info.comp_info = (MMON_COMP_INFO *)malloc (sizeof (MMON_COMP_INFO) * info.num_comp);
```
joohok: where is this memory freed?

src/base/memory_monitor_sr.cpp
```diff
@@ -557,16 +694,26 @@ void mmon_resize_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, int64_t old
...
+
   return error;
 }
 
-int mmon_aggregate_tran_info (MMON_TRAN_INFO *info, int tran_count)
+int mmon_aggregate_tran_info (MMON_TRAN_INFO &info, int tran_count)
 {
   int error = NO_ERROR;
 
+  if (mmon_Gl != nullptr)
```
wrlawodms: I think it must not be called if memmon is not initialized. It has to be handled as an error on the upperside or set an error here. So, `assert (mmon_Gl != nullptr)` or it should return an error. `NO_ERROR` with the empty `info` seems bad.

src/base/memory_monitor_sr.cpp
```diff
@@ -402,19 +479,69 @@ namespace cubperf
...
+	    info.num_tran = tran_info.size ();
+	  }
+	else
+	  {
+	    info.num_tran = tran_count;
+	  }
+	qsort (&tran_info.front (), tran_info.size (), sizeof (tran_info[0]), comp);
+      }
+
+    info.tran_stat = (MMON_TRAN_STAT *)malloc (sizeof (MMON_TRAN_STAT) * info.num_tran);
```
joohok: I think all the malloc parts needs `free ()`

src/base/memory_monitor_sr.cpp
```diff
@@ -154,8 +156,8 @@ namespace cubperf
       void resize_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, int64_t old_size, int64_t new_size);
       void move_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID src, MMON_STAT_ID dest, int64_t size);
       void end_init_phase ();
-      int aggregate_module_info (MMON_MODULE_INFO *info, int module_index);
-      int aggregate_tran_info (MMON_TRAN_INFO *info, int tran_count);
+      int aggregate_module_info (MMON_MODULE_INFO &info, int module_index);
+      int aggregate_tran_info (MMON_TRAN_INFO &info, int tran_count);
```
wrlawodms: It should be const
```suggestion
      int aggregate_module_info (MMON_MODULE_INFO &info, int module_index) const;
      int aggregate_tran_info (MMON_TRAN_INFO &info, int tran_count) const;
```

src/base/memory_monitor_sr.cpp
```diff
@@ -115,7 +117,7 @@ namespace cubperf
       mmon_module &operator = (const mmon_module &) = delete;
       mmon_module &operator = (mmon_module &&) = delete;
 
-      virtual int aggregate_stats (const MMON_MODULE_INFO &info);
+      virtual int aggregate_stats (MMON_MODULE_INFO &info);
```
```suggestion
      virtual int aggregate_stats (MMON_MODULE_INFO &info) const;
```

src/base/memory_monitor_sr.cpp
```diff
@@ -89,6 +90,7 @@ namespace cubperf
       void add_stat (int64_t size, int subcomp_idx);
       void add_expand_resize_count ();
       void end_init_phase ();
+      int get_stat (MMON_COMP_INFO &info);
```
```suggestion
      int get_stat (MMON_COMP_INFO &info) const;
```

src/base/memory_monitor_sr.cpp
```diff
@@ -66,6 +66,7 @@ namespace cubperf
 
       const char *get_name ();
       void add_cur_stat (int64_t size);
+      void get_stat (MMON_SUBCOMP_INFO &info);
```
```suggestion
      void get_stat (MMON_SUBCOMP_INFO &info) const;
```

src/base/memory_monitor_sr.cpp
```diff
@@ -173,9 +175,26 @@ namespace cubperf
 	  aggregater &operator = (const aggregater &) = delete;
 	  aggregater &operator = (aggregater &&) = delete;
 
-	  int get_server_info (const MMON_SERVER_INFO &info);
-	  int get_module_info (const MMON_MODULE_INFO &info, int module_index);
-	  int get_transaction_info (const MMON_TRAN_INFO &info, int tran_count);
+	  void get_server_info (MMON_SERVER_INFO &info);
+	  int get_module_info (MMON_MODULE_INFO &info, int module_index);
+	  int get_transaction_info (MMON_TRAN_INFO &info, int tran_count);
```
```suggestion
	  void get_server_info (MMON_SERVER_INFO &info) const;
	  int get_module_info (MMON_MODULE_INFO &info, int module_index) const;
	  int get_transaction_info (MMON_TRAN_INFO &info, int tran_count) const;
```

src/base/memory_monitor_sr.cpp
```diff
@@ -402,19 +479,69 @@ namespace cubperf
...
+      {
+	tdes = trantable[i];
+	if (tdes != NULL && tdes->trid != NULL_TRANID && tdes->state == TRAN_ACTIVE)
+	  {
+	    tran_info.push_back (std::make_pair (tdes->trid, tdes->cur_mem_usage.load ()));
+	  }
+      }
+
+    if (tran_count == 0)
+      {
```
```suggestion
      {
      // 0: all by default
```

src/base/memory_monitor_sr.cpp
```diff
@@ -402,19 +479,69 @@ namespace cubperf
...
+	if (tran_count > tran_info.size ())
+	  {
+	    info.num_tran = tran_info.size ();
+	  }
+	else
+	  {
+	    info.num_tran = tran_count;
+	  }
+	qsort (&tran_info.front (), tran_info.size (), sizeof (tran_info[0]), comp);
```
wrlawodms: I suggest std::sort, and to define comp() only in this context.
```suggestion
        std::sort (tran_info.begin (), tran_info.end (), [] (const auto& tran_pair1, const auto& tran_pair2)
            {
              return tran_pair1.second > tran_pair2.second; // mem usage comparison
            });
```

wrlawodms: or 
```
        const auto& comp = [] (const auto& tran_pair1, const auto& tran_pair2)
        {
          return tran_pair1.second > tran_pair2.second; // mem usage comparison
        };
        std::sort (tran_info.begin (), tran_info.end (), comp);
```

src/base/memory_monitor_sr.cpp
```diff
@@ -402,19 +479,69 @@ namespace cubperf
...
   {
-    return 0;
+    int error = NO_ERROR;
+
+    error = m_mmon->m_module[module_index]->aggregate_stats (info);
+    return error;
   }
 
-  int memory_monitor::aggregater::get_transaction_info (const MMON_TRAN_INFO &info, int tran_count)
+  int memory_monitor::aggregater::get_transaction_info (MMON_TRAN_INFO &info, int tran_count)
```
```suggestion
  int memory_monitor::aggregater::get_transaction_info (MMON_TRAN_INFO &info, int max_tran_count)
```

wrlawodms (commented):
The memmon-related code generates a lot of compile warnings. Please create an issue to deal with them and make sure no additional warnings are added.
```
/home/cubrid/cubrid/dev/src/base/memory_monitor_sr.cpp:262:38: warning: comparison of integer expressions of different signedness: ‘int64_t’ {aka ‘long int’} and ‘std::[360/1824]
ase<long unsigned int>::__int_type’ {aka ‘long unsigned int’} [-Wsign-compare]
     assert ((size >= 0) || (- (size) <= m_stat.cur_stat));
                             ~~~~~~~~~^~~~~~~~~~~~~~~~~~
/home/cubrid/cubrid/dev/src/base/memory_monitor_sr.cpp:274:22: warning: comparison of integer expressions of different signedness: ‘int’ and ‘std::vector<std::unique_ptr<cubperf:
:mmon_subcomponent> >::size_type’ {aka ‘long unsigned int’} [-Wsign-compare]
  assert (subcomp_idx < m_subcomponent.size ());
          ~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~
/home/cubrid/cubrid/dev/src/base/memory_monitor_sr.cpp: In member function ‘int cubperf::mmon_component::get_stat(MMON_COMP_INFO&)’:
/home/cubrid/cubrid/dev/src/base/memory_monitor_sr.cpp:311:23: warning: comparison of integer expressions of different signedness: ‘int’ and ‘uint32_t’ {aka ‘unsigned int’} [-Wsi
gn-compare]
     for (int i = 0; i < info.num_subcomp; i++)
                     ~~^~~~~~~~~~~~~~~~~~
/home/cubrid/cubrid/dev/src/base/memory_monitor_sr.cpp: In member function ‘virtual int cubperf::mmon_module::aggregate_stats(MMON_MODULE_INFO&)’:
/home/cubrid/cubrid/dev/src/base/memory_monitor_sr.cpp:391:23: warning: comparison of integer expressions of different signedness: ‘int’ and ‘uint32_t’ {aka ‘unsigned int’} [-Wsi
gn-compare]
     for (int i = 0; i < info.num_comp; i++)
                     ~~^~~~~~~~~~~~~~~
In file included from /home/cubrid/cubrid/dev/src/transaction/log_impl.h:66,
                 from /home/cubrid/cubrid/dev/src/base/memory_monitor_sr.cpp:33:
/home/cubrid/cubrid/dev/src/base/memory_monitor_sr.cpp: In member function ‘void cubperf::mmon_module::add_stat(MMON_STAT_ID, int64_t)’:
/home/cubrid/cubrid/dev/src/base/memory_monitor_sr.cpp:416:38: warning: comparison of integer expressions of different signedness: ‘int64_t’ {aka ‘long int’} and ‘std::__atomic_b
ase<long unsigned int>::__int_type’ {aka ‘long unsigned int’} [-Wsign-compare]
     assert ((size >= 0) || (- (size) <= m_stat.cur_stat));
                             ~~~~~~~~~^~~~~~~~~~~~~~~~~~
/home/cubrid/cubrid/dev/src/base/memory_monitor_sr.cpp:430:19: warning: comparison of integer expressions of different signedness: ‘int’ and ‘std::vector<std::unique_ptr<cubperf:
:mmon_component> >::size_type’ {aka ‘long unsigned int’} [-Wsign-compare]
  assert (comp_idx < m_component.size ());
          ~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
/home/cubrid/cubrid/dev/src/base/memory_monitor_sr.cpp: In member function ‘void cubperf::mmon_module::add_expand_resize_count(MMON_STAT_ID)’:
/home/cubrid/cubrid/dev/src/base/memory_monitor_sr.cpp:446:19: warning: comparison of integer expressions of different signedness: ‘int’ and ‘std::vector<std::unique_ptr<cubperf:
:mmon_component> >::size_type’ {aka ‘long unsigned int’} [-Wsign-compare]
  assert (comp_idx < m_component.size ());
          ~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
/home/cubrid/cubrid/dev/src/base/memory_monitor_sr.cpp: In member function ‘int cubperf::memory_monitor::aggregater::get_transaction_info(MMON_TRAN_INFO&, int)’:
/home/cubrid/cubrid/dev/src/base/memory_monitor_sr.cpp:521:17: warning: comparison of integer expressions of different signedness: ‘int’ and ‘std::vector<std::pair<int, long unsi
gned int> >::size_type’ {aka ‘long unsigned int’} [-Wsign-compare]
  if (tran_count > tran_info.size ())
```

src/base/memory_monitor_sr.hpp
```diff
@@ -50,7 +50,7 @@ void mmon_add_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, int64_t size);
 void mmon_sub_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, int64_t size);
 void mmon_move_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID src, MMON_STAT_ID dest, int64_t size);
 void mmon_resize_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, int64_t old_size, int64_t new_size);
-int mmon_aggregate_module_info (MMON_MODULE_INFO *info, int module_index);
-int mmon_aggregate_tran_info (MMON_TRAN_INFO *info, int tran_count);
+int mmon_aggregate_module_info (MMON_MODULE_INFO &info, int module_index);
+int mmon_aggregate_tran_info (MMON_TRAN_INFO &info, int tran_count);
```
hornetmj: MMON_MODULE_INFO and MMON_TRAN_INFO have MMON_SERVER_INFO. It must be same information. How about extracting MMON_SERVER_INFO from them and make addtional API and protocol etc for it?

src/base/memory_monitor_sr.cpp
```diff
@@ -557,16 +694,26 @@ void mmon_resize_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, int64_t old
     }
 }
 
-int mmon_aggregate_module_info (MMON_MODULE_INFO *info, int module_index)
+int mmon_aggregate_module_info (MMON_MODULE_INFO &info, int module_index)
```
hornetmj: What is your plan to check if the module_index is valid or not?

src/base/memory_monitor_sr.cpp
```diff
@@ -312,10 +366,33 @@ namespace cubperf
...
+	  {
+	    er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, sizeof (MMON_COMP_INFO) * info.num_comp);
+	    return ER_OUT_OF_VIRTUAL_MEMORY;
+	  }
+      }
+
+    for (int i = 0; i < info.num_comp; i++)
+      {
+	m_component[i]->get_stat (info.comp_info[i]);
+      }
```
hornetmj: Move this for loop to the previous if block to grouping 'get component' action.

src/base/memory_monitor_sr.cpp
```diff
@@ -312,10 +366,33 @@ namespace cubperf
       }
   }
 
-  int mmon_module::aggregate_stats (const MMON_MODULE_INFO &info)
+  int mmon_module::aggregate_stats (MMON_MODULE_INFO &info)
   {
-    int error = NO_ERROR;
-    return error;
+    strcpy (info.name, m_module_name.c_str ());
```
hornetmj: Use strncpy or memcpy() for safety reason. Please consider all strcpy together.

src/base/memory_monitor_sr.cpp
hornetmj: Apply same approach to mmon_component::get_stat.

src/base/memory_monitor_sr.cpp

src/base/memory_monitor_sr.cpp

src/base/memory_monitor_sr.cpp

src/base/memory_monitor_sr.cpp

src/base/memory_monitor_sr.cpp

src/base/memory_monitor_sr.cpp
Rudeus: It will be freed at network function after packing information to reply buffer.

src/base/memory_monitor_sr.cpp
Rudeus: It will be freed at network function after packing information to reply buffer.

src/base/memory_monitor_sr.cpp

src/base/memory_monitor_sr.cpp
Rudeus: What is the reason for wanting to change tran_count to max_tran_count?

src/base/memory_monitor_sr.cpp
Rudeus: I added comments with branching based on the options because the explanation could be confusing. Default option of memmon has to show 5 transaction in descending order (that means, tran_count =  5).

src/base/memory_monitor_sr.cpp
Rudeus: I was not proficient in using C++ STL and was unaware of better approaches. Thank you for the advice.

src/base/memory_monitor_sr.cpp
Rudeus: It will be freed at network function after packing information to reply buffer.

src/base/memory_monitor_sr.cpp
Rudeus: 'module_index' will be checked in the utility function 'memmon' by verifying against MMON_MODULE_ID. The only error case will be when the user provides an invalid module index, not a module name. Module name will be checked using "char **module_names" in 'memory_monitor_cl.hpp'.

Rudeus: But I agree about that some ways to check parameter to prevent some corner-cases, for example, the value of the parameter is changed because of memory overflow of other section. I will add it.

src/base/memory_monitor_sr.cpp

src/base/memory_monitor_sr.cpp
Rudeus: I agree. I change it to assertion.

src/base/memory_monitor_sr.hpp
Rudeus: Considering the aspect of potential duplicated information, I believe your opinion is correct. However, for the information to actually be duplicated, both options (modules and transactions) need to be invoked. Also going through the network protocol again for the server information, which is a very small amount of information, is not ideal in situations where the utility may be frequently invoked.

Rudeus (commented):
Remove warnings and add API and functions for getting brief information of modules

src/base/memory_monitor_common.h
```diff
@@ -26,6 +26,9 @@
 
 #include <cstdint>
 
+#define MMON_MODULE_BRIEF_OPTION 1
+#define MMON_MODULE_DEFAULT_OPTION 2
```
wrlawodms: Why don't you use enum or enum class? It's safer and more readable.

src/base/memory_monitor_sr.cpp
```diff
@@ -312,10 +351,41 @@ namespace cubperf
...
   {
-    int error = NO_ERROR;
-    return error;
+    strncpy (info.name, m_module_name.c_str (), m_module_name.size ());
+
+    if (get_brief_info)
+      {
+	info.stat.cur_stat = m_stat.cur_stat.load ();
+	return NO_ERROR;
+      }
```
wrlawodms: How about making another method for this? It even doesn't seem aggregation to me.

src/base/memory_monitor_sr.cpp
```diff
@@ -402,30 +472,110 @@ namespace cubperf
...
   }
 
-  int memory_monitor::aggregater::get_module_info (const MMON_MODULE_INFO &info, int module_index)
+  int memory_monitor::aggregater::get_module_info (MMON_MODULE_INFO &info, int module_index) const
   {
-    return 0;
+    int error = NO_ERROR;
+
+    error = m_mmon->m_module[module_index]->aggregate_stats (info, false);
+    return error;
```
```suggestion
    return m_mmon->m_module[module_index]->aggregate_stats (info, false);
```

src/base/memory_monitor_sr.cpp
```diff
@@ -402,30 +472,110 @@ namespace cubperf
...
-  int memory_monitor::aggregater::get_server_info (const MMON_SERVER_INFO &info)
+  void memory_monitor::aggregater::get_server_info (MMON_SERVER_INFO &info) const
   {
-    return 0;
+    strncpy (info.name, m_mmon->m_server_name.c_str (), m_mmon->m_server_name.size ());
+    info.total_mem_usage = m_mmon->m_total_mem_usage.load ();
   }
 
-  int memory_monitor::aggregater::get_module_info (const MMON_MODULE_INFO &info, int module_index)
+  int memory_monitor::aggregater::get_module_info (MMON_MODULE_INFO &info, int module_index) const
```
wrlawodms: Nit. input-first seems better to me.
```suggestion
  int memory_monitor::aggregater::get_module_info (int module_index, MMON_MODULE_INFO &info) const
```

src/base/memory_monitor_sr.cpp
```diff
@@ -402,30 +472,110 @@ namespace cubperf
...
   {
-    return 0;
+    int error = NO_ERROR;
+
+    for (int i = 0; i < MMON_MODULE_LAST; i++)
+      {
+	error = m_mmon->m_module[i]->aggregate_stats (info[i], true);
+	if (error != NO_ERROR)
+	  {
+	    return error;
```
wrlawodms: Don't we need to free the partially allocated memory here?

src/base/memory_monitor_sr.cpp
```diff
@@ -557,16 +724,44 @@ void mmon_resize_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, int64_t old
...
 
+  assert (mmon_Gl != nullptr);
+
+  if (option == MMON_MODULE_BRIEF_OPTION || option == MMON_MODULE_DEFAULT_OPTION)
+    {
+      error = mmon_Gl->aggregate_module_brief_info (info, option);
+    }
+  else
+    {
+      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_GENERIC_ERROR, 0);
```
wrlawodms: Are you going to define one later? Then, comment TODO.

src/base/memory_monitor_sr.cpp
```diff
@@ -402,30 +472,110 @@ namespace cubperf
...
+    for (int i = 0; i < MMON_MODULE_LAST; i++)
+      {
+	error = m_mmon->m_module[i]->aggregate_stats (info[i], true);
+	if (error != NO_ERROR)
+	  {
+	    return error;
+	  }
+      }
+
+    if (MMON_MODULE_DEFAULT_OPTION)
```
wrlawodms: 1. What is the default?
2. The default could be changed. Let's use more specific conditions in the code.

src/base/memory_monitor_sr.cpp
```diff
@@ -557,16 +724,44 @@ void mmon_resize_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, int64_t old
...
+  return error;
+}
+
+int mmon_aggregate_module_brief_info (MMON_MODULE_INFO *&info, int option)
 {
   int error = NO_ERROR;
 
+  assert (mmon_Gl != nullptr);
+
+  if (option == MMON_MODULE_BRIEF_OPTION || option == MMON_MODULE_DEFAULT_OPTION)
```
wrlawodms: Please elaborate on the need of `option`. `MMON_MODULE_BRIEF_OPTION` is not used under `aggregate_module_brief_info`, I can't even guess it.

src/base/memory_monitor_common.h
Rudeus: I have been considering that point, and I chose to use define. However, I think the method you suggested is better. I change it.

src/base/memory_monitor_sr.cpp
Rudeus: Why isn't it aggregation? I think getting name and current stat also can be aggregation.

src/base/memory_monitor_sr.cpp

src/base/memory_monitor_sr.cpp
Rudeus: Declaration of MMON_MODULE_INFO occurs in network function. Stored information through mmon has to pack to reply buffer, so structures which is alloced in these aggregating functions will be freed after all information is packed in network function.

src/base/memory_monitor_sr.cpp
Rudeus: In memmon utility, there is an option which is named default(cubrid memmon <server-name>, with no option command) that shows top5 memory using modules and transactions. MMON_MODULE_DEFAULT_OPTION is for that.

src/base/memory_monitor_sr.cpp
Rudeus: I changed defining options to use enum type, so it doesn't need anymore. I remove it.

src/base/memory_monitor_sr.cpp
Rudeus: In memmon utility, there is an option which is named default(cubrid memmon , with no option command) that shows top5 memory using modules and transactions. MMON_MODULE_DEFAULT_OPTION is for that. There is also an module option named brief(cubrid memmon -m <b or brief>) that shows total memory usage of each module. MMON_MODULE_BRIEF_OPTION is for that.

Rudeus: They are included in spec.

src/base/memory_monitor_sr.cpp
```diff
@@ -221,19 +225,25 @@ namespace cubperf
 
   void mmon_subcomponent::add_cur_stat (int64_t size)
   {
-    assert ((size >= 0) || (- (size) <= m_cur_stat));
+    assert ((size >= 0) || ((uint64_t) (-size) <= m_cur_stat));
```
joohok, line 228: `(uint64_t) (-size)`
when the `size` has a positive value, `(uint64_t) (-size)` can have a very large value.

src/base/memory_monitor_sr.cpp
joohok: So, is memory allocated and deallocated for each request? 
if I understand correctly, I think we can remove/reduce the memory alloc/dealloc process, because we can know the number of subcomponent (it maybe determined when server boot up). 

There can be a method in which 
1. the information (info) to be passed to the network is pre-allocated and reused continuously. 
2. Alternatively, using an array within the `info` structure without memory allocation can be considered. 

Reducing the frequency of memory allocation and deallocation processes can indeed decrease risks and make management easier.

joohok (commented):
I think ctor for aggregator (`aggregater (memory_monitor *mmon);`) can be declared as
`aggregater (memory_monitor *mmon) : m_mmon (mmon) {};`

src/base/memory_monitor_sr.hpp
hornetmj: Design and implementation require a generalized (unified) approach. As you said, it is possible to do some optimizations to reduce communication costs. However, having a redundant data structure (MMON_SERVER_INFO) in the server for this purpose is not a clear reason. Since this information only needs to be collected once between the start and end of a utility request, there are plenty of ways to meet both requirements from you and me.

src/base/memory_monitor_sr.cpp
```diff
@@ -115,7 +117,7 @@ namespace cubperf
       mmon_module &operator = (const mmon_module &) = delete;
       mmon_module &operator = (mmon_module &&) = delete;
 
-      virtual int aggregate_stats (const MMON_MODULE_INFO &info);
+      virtual int aggregate_stats (MMON_MODULE_INFO &info, bool get_brief_info) const;
```
```suggestion
      virtual int aggregate_stats (MMON_MODULE_INFO &info, bool is_brief) const;
```

src/base/memory_monitor_sr.cpp
```diff
@@ -154,8 +156,9 @@ namespace cubperf
       void resize_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID stat_id, int64_t old_size, int64_t new_size);
       void move_stat (THREAD_ENTRY *thread_p, MMON_STAT_ID src, MMON_STAT_ID dest, int64_t size);
       void end_init_phase ();
-      int aggregate_module_info (MMON_MODULE_INFO *info, int module_index);
-      int aggregate_tran_info (MMON_TRAN_INFO *info, int tran_count);
+      int aggregate_module_info (MMON_MODULE_INFO &info, int module_index) const;
+      int aggregate_module_brief_info (MMON_MODULE_INFO *&info, MMON_MODULE_OPTION option) const;
```
hornetmj: Referring to the specifications of CBRD-24635, the following three cases need to be considered when collecting module information.
1. default behavior - specific memory usage of specified modules
2. brief - summary of unsorted memory usage for all modules
3. default - top 5 modules in memory usage sorted

Accordingly, changing aggregate_module_info() as follows can be an option.
```
int aggregate_module_info (MMON_MODULE_INFO &info, int module_index, int brief, int count) const;
// brief == true && count == 0 -> generating a summary of unsorted all module memory usages
// brief == true && count > 0 -> generate a summary of sorted memory usage of count number of modules
```

It is up to you to decide the variable names and specific behavior specifications.

hornetmj: The method you suggested is also good for me. It would be nice if the aggregate_module_brief_info() function could handle the default case as well.

src/base/memory_monitor_sr.cpp
hornetmj: The positive value is fillter out by (size >= 0) condition.

src/base/memory_monitor_sr.cpp
```diff
@@ -221,19 +225,25 @@ namespace cubperf
...
+    strncpy (info.name, m_subcomp_name.c_str (), m_subcomp_name.size ());
+    info.cur_stat = m_cur_stat.load ();
+  }
+
   const char *mmon_component::get_name ()
   {
     return m_comp_name.c_str ();
   }
 
   void mmon_component::add_stat (int64_t size, int subcomp_idx)
```
hornetmj: Nit:

Arguments are usually ordered naturally, first by target, then by size. - (int subcomp_idx, int64_t size)

src/base/memory_monitor_sr.cpp
```diff
@@ -312,10 +351,41 @@ namespace cubperf
       }
   }
 
-  int mmon_module::aggregate_stats (const MMON_MODULE_INFO &info)
+  int mmon_module::aggregate_stats (MMON_MODULE_INFO &info, bool get_brief_info) const
```
```suggestion
  int mmon_module::aggregate_stats (MMON_MODULE_INFO &info, bool is_brief) const
```

src/base/memory_monitor_sr.cpp
```diff
@@ -312,10 +351,41 @@ namespace cubperf
...
   {
-    int error = NO_ERROR;
-    return error;
+    strncpy (info.name, m_module_name.c_str (), m_module_name.size ());
+
+    if (get_brief_info)
+      {
+	info.stat.cur_stat = m_stat.cur_stat.load ();
+	return NO_ERROR;
```
hornetmj: Code that returns in the case of if and does not return in the else is less readable.

```
if (is_brief)
{
    // gather cur_stat
}
else
{
    // gather all
    // gather comps and sub-comps
}

return return_code(error_code);
```

src/base/memory_monitor_sr.cpp
hornetmj: @wrlawodms I agree with you if you mean input params followed by ouput params. @Rudeus Isn't the previous version correct in this case?

src/base/memory_monitor_sr.cpp
hornetmj: The related code can be reviewed after the decision on this opinion is complete.

src/base/memory_monitor_sr.cpp
```diff
@@ -312,9 +352,52 @@ namespace cubperf
...
+
+	    if (info.comp_info == NULL)
+	      {
+		er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, sizeof (MMON_COMP_INFO) * info.num_comp);
+		error =  ER_OUT_OF_VIRTUAL_MEMORY;
+	      }
+
+	    for (uint32_t i = 0; i < info.num_comp; i++)
+	      {
+		error = m_component[i]->get_stat (info.comp_info[i]);
```
wrlawodms: A positive `error` will be overwritten by following one. For example
I = 0 : error
I = 1 : no_error
-> It has to realize when I == 0.

I think we should check the error every get_stat() and clean up resources we have allocated at that point.

src/base/memory_monitor_sr.cpp
```diff
@@ -312,9 +352,52 @@ namespace cubperf
...
+	      }
+	  }
+
+	if (error == ER_OUT_OF_VIRTUAL_MEMORY)
```
wrlawodms: We know there is only one error, ER_OUT_OF_VIRTUAL_MEMORY, but it's more open to change for additional errors. Actually, in this case, this function doesn't need to know which error is returned for get_stat() since it has only one way to deal with an error, just cleaning up and stopping.
```suggestion
	if (error != NO_ERROR)
```

src/base/memory_monitor_sr.cpp
```diff
@@ -312,9 +352,52 @@ namespace cubperf
...
+	  {
+	    for (uint32_t i = 0; i < info.num_comp; i++)
+	      {
+		free_and_init (info.comp_info[i].subcomp_info);
+	      }
+	    free_and_init (info.comp_info);
```
wrlawodms: Some may have succeeded to allocate and some may not.

src/base/memory_monitor_sr.cpp
```diff
@@ -402,29 +485,116 @@ namespace cubperf
...
+  }
+
+  int memory_monitor::aggregater::get_module_info (MMON_MODULE_INFO *&info, int module_index) const
+  {
+    int idx = 0;
+    int error = NO_ERROR;
+    if (module_index == -1)
+      {
+	// aggregate all detail memory information of modules
+	for (idx; idx < MMON_MODULE_LAST; idx++)
```
wrlawodms: nit.
```suggestion
	for (int idx = 0; idx < MMON_MODULE_LAST; idx++)
```

src/transaction/log_tran_table.c
```diff
@@ -1457,6 +1458,28 @@ xlogtb_dump_trantable (THREAD_ENTRY * thread_p, FILE * out_fp)
...
+ */
+void
+logtb_get_tran_memory_info_nolatch (std::vector < std::pair < int, uint64_t >> &tran_info)
+{
+  LOG_TDES *tdes;
+
+  for (int i = 0; i < NUM_TOTAL_TRAN_INDICES; i++)
+    {
+      tdes = log_Gl.trantable.all_tdes[i];
+      if (tdes != NULL && tdes->trid != NULL_TRANID && tdes->state == TRAN_ACTIVE)
```
hornetmj, line 1476: After COMMIT/ABORT, tdes could have same condition. (Need to check)
- log_complete() -> logtb_clear_tdes()

src/base/memory_monitor_sr.cpp
```diff
@@ -402,29 +485,116 @@ namespace cubperf
...
+	  {
+	    error = m_mmon->m_module[idx]->aggregate_stats (info[idx], false);
+	    if (error != NO_ERROR)
+	      {
+		return error;
```
wrlawodms: Q. I wonder it is really necessary to get all module information. What I was told is that a client will request several time even when it want know about all module. Then it's not needed.

If this is really needed, we should come up with a way to notify until which index it succeeds to aggregate.

src/base/memory_monitor_sr.cpp
```diff
@@ -402,29 +485,116 @@ namespace cubperf
...
+    else
+      {
+	error =  m_mmon->m_module[module_index]->aggregate_stats (info[idx], false);
+      }
+
+    return error;
   }
 
-  int memory_monitor::aggregater::get_module_info (const MMON_MODULE_INFO &info, int module_index)
+  int memory_monitor::aggregater::get_module_info_summary (MMON_MODULE_INFO *&info, bool sorted_result) const
```
```suggestion
  int memory_monitor::aggregater::get_module_info_summary (MMON_MODULE_INFO *&info, bool sorted_by_mem_usage) const
```
It seems better to note how sorted they are.

src/base/memory_monitor_cl.hpp
```diff
@@ -26,6 +26,13 @@
 
 #include "memory_monitor_common.h"
 
+typedef enum mmon_module_option
+{
+  MMON_MODULE_DETAIL,
+  MMON_MODULE_BRIEF,
+  MMON_MODULE_DEFAULT
```
hornetmj: For other developer who will develop this module, add description(comment) for each case.

src/base/memory_monitor_sr.cpp
```diff
@@ -402,29 +485,116 @@ namespace cubperf
...
+
+    if (tran_count == 0)
+      {
+	// transaction option with no tran-count option
+	info.num_tran = tran_info.size ();
+      }
+    else
+      {
+	// 1) transaction option with tran-count option
+	// 2) default option (cubrid memmon <server-name>)
```
wrlawodms: We shouldn't mention the "default" here.
We can describe here that It's just get the max value between the request tran count and the active tran count. I think we even don't need the comment with the `std::max`.
 ``` info.num_tran = std::max (tran_count, (int) tran_info.size ());```

src/base/memory_monitor_sr.cpp
```diff
@@ -402,29 +485,116 @@ namespace cubperf
...
+	    info.num_tran = tran_info.size ();
+	  }
+	else
+	  {
+	    info.num_tran = tran_count;
+	  }
+
+	const auto &comp = [] (const auto& tran_pair1, const auto& tran_pair2)
+	{
+	  return tran_pair1.second > tran_pair2.second;
```
wrlawodms: Please comment by which it is sorted. pair.second it unreadable. 
```// sorted in descending order by mem usage 
or something.

File: src/base/memory_monitor_sr.cpp (hunk 31 above)
Comment (wrlawodms) [reply src/base/memory_monitor_sr.cpp]: If it's too picky to track allocation and do delicate according to error cases, I think you can use unique_ptr<A[]>. or you can just use std::vector.

File: src/base/memory_monitor_cl.hpp (hunk 38 above)
Comment (Rudeus) [reply src/base/memory_monitor_cl.hpp]: Done.

File: src/base/memory_monitor_sr.cpp (hunk 31 above)
Comment (Rudeus) [reply src/base/memory_monitor_sr.cpp]: I think it can solve with adding some condition. Please check my latest commit and give additional feedback if it is not good.

File: src/base/memory_monitor_sr.cpp (hunk 32 above)

File: src/base/memory_monitor_sr.cpp (hunk 33 above)
Comment (Rudeus) [reply src/base/memory_monitor_sr.cpp]: `free_and_init()` function checks nullity and only frees not null pointers.

File: src/base/memory_monitor_sr.cpp (hunk 34 above)
Comment (Rudeus) [reply src/base/memory_monitor_sr.cpp]: `idx` is initialized with 0 when declared, and I intend to use this variable both for aggregating entire module information and aggregating specific modules with `module_index`.

File: src/base/memory_monitor_sr.cpp (hunk 36 above)
Comment (Rudeus) [reply src/base/memory_monitor_sr.cpp]: Server (memory monitoring manager) has to give detailed information of all registered modules when memmon gets the '-a' command from a user.  And due to the display specification of this utility, the atomicity of displaying information is limited by category, such as module, transaction, etc.

File: src/base/memory_monitor_sr.cpp (hunk 37 above)

File: src/base/memory_monitor_sr.cpp (hunk 39 above)
Comment (Rudeus) [reply src/base/memory_monitor_sr.cpp]: I agree. But I think we have to use `std::min`.

File: src/base/memory_monitor_sr.cpp (hunk 40 above)
Comment (Rudeus) [reply src/base/memory_monitor_sr.cpp]: I add comment.

File: src/transaction/log_tran_table.c (hunk 35 above)
Comment (Rudeus) [reply src/transaction/log_tran_table.c:1476]: Upon checking the mentioned content, it appears that the state of transactions is generally being properly managed during commit, but there seem to be some vulnerabilities during checkpoint situations or certain system transaction processes. Therefore, merely checking `tdes->state == TRAN_ACTIVE` is not sufficient to determine if the transaction is currently active. However, it is believed that this can be overcome by checking `tdes->trid != NULL_TRANID`, as confirmed in the dump_trantable() function, where output is determined based on such conditions.

File: src/base/memory_monitor_sr.cpp [hunk 41]
```diff
@@ -402,29 +490,109 @@ namespace cubperf
...
+    if (tran_count == 0)
+      {
+	// transaction option with no tran-count option
+	info.num_tran = tran_info.size ();
+      }
+    else
+      {
+	info.num_tran = std::min (tran_count, (int) tran_info.size ());
+
+	// sort memory usage in descending order
```
```suggestion
	// sort by memory usage in descending order
```

src/base/memory_monitor_sr.cpp
```diff
@@ -312,9 +354,55 @@ namespace cubperf
...
+		for (uint32_t i = 0; i < info.num_comp; i++)
+		  {
+		    error = m_component[i]->get_stat (info.comp_info[i]);
+
+		    if (error != NO_ERROR)
+		      {
+			for (uint32_t i = 0; i < info.num_comp; i++)
+			  {
+			    free_and_init (info.comp_info[i].subcomp_info);
+			  }
```
wrlawodms: Even if `free_and_init` checks whether it's null, it'd be better to free what's necessary in terms of readability and even performance (which it trivial, though. but why not?)).
```suggestion
			for (uint32_t j = 0; j < i; j++)
			  {
			    free_and_init (info.comp_info[j].subcomp_info);
			  }
```

src/base/memory_monitor_sr.cpp
wrlawodms: :)

src/base/memory_monitor_sr.cpp
wrlawodms: I'm good with yours. I just mentioned another way (seems easier to me :) ).

src/base/memory_monitor_sr.cpp
wrlawodms: I don't understand your answer. I mean, with `-a`, a client will request 3 times to get each module's information when there are 3 modules. I was told that it's the design since we should print as many as we can with even an error. And, this means we don't need for `get_module_info` to collect all module's information. That's my understanding.

src/base/memory_monitor_sr.cpp
wrlawodms: Why not just using 0 instead of idx for a request for one module? it's more obvious to use the constant.

Rudeus: Ok. I fix it.

src/base/memory_monitor_sr.cpp
```diff
@@ -402,29 +490,108 @@ namespace cubperf
...
-    return 0;
+    int error = NO_ERROR;
+
+    if (module_index == -1)
+      {
+	// aggregate all detail memory information of modules
+	for (int idx = 0; idx < MMON_MODULE_LAST; idx++)
+	  {
+	    error = m_mmon->m_module[idx]->aggregate_stats (info[idx], false);
+	    if (error != NO_ERROR)
```
wrlawodms: Don't we have to free allocated memory of succeeded aggregation? For example, 
m_mmon->m_module[0]->aggregate_stats // success
m_mmon->m_module[1]->aggregate_stats // success
m_mmon->m_module[2]->aggregate_stats // error

Or are they gonna addressed in the outside?

wrlawodms: If the above is the case, depending on the last parameter of `aggregated_stats`, `get_summary`, the caller decides if it must free memory or not. I think it's bad because the side-effect that the called should deal with is inconsistent and it can be only warned by a comment. 
To resolve it, it seems that there are two ways in handy:
1. Split the function into `aggregate_summary_stats` and `aggregate_stats`.
2. Use a smart pointer so leave it to RAII.

Sorry if I get it wrong.

Rudeus: Inside `mmon_module::aggregate_stats()`, there is a deallocation phase for the allocated memory in case of an error. However, this deallocation is for a single module processing, and in the situation where errors occur while aggregating information for multiple modules, deallocation cannot be performed for the previously successful aggregation before the error. To solve this situation, I modify the logic of deallocation that network function, where first allocation of `MMON_XXX_INFO` was occurred, handles all deallocations. With this updated logic, there is no need to perform deallocation for a single module within `mmon_module::aggregate_stats()`, so that part has been removed.

src/base/memory_monitor_common.h
```diff
@@ -57,16 +58,15 @@ typedef struct mmon_comp_info
...
 } MMON_COMP_INFO;
 
 typedef struct mmon_module_info
 {
-  MMON_SERVER_INFO server_info;
   char name[DB_MAX_IDENTIFIER_LENGTH];
   MMON_OUTPUT_MEM_STAT stat;
   uint32_t num_comp;
-  MMON_COMP_INFO *comp_info;
+    std::vector < MMON_COMP_INFO > comp_info;
```
wrlawodms: I guess you can remove `num_comp` and `num_subcomp`.

wrlawodms: Use `/*INDENT-OFF*/` and `/*INDENT-ON*/` for the cpp code in .c file or change its file extension to cpp. We use the different style checkers for cpp and c, and the checker for C doesn't recognize this and may make it ugly. I think it'd be better to use .cpp extension to make use of the cpp checker.

Rudeus: I think memory monitoring manager needs `num_comp` and `num_comp` because utility needs to know how many components/subcomponents exists for unpacking phase.

Rudeus: And I fix indent. Thanks

wrlawodms: > I think memory monitoring manager needs num_comp and num_comp because utility needs to know how many components/subcomponents exists for unpacking phase.

I can't find the usage.. maybe will be used in one of next PRs?
I wonder it can't be done with `comp_info.size()`.

When packing or unpacking, the count value is needed. Ok, but I don't understand this structure itself needs the count value. In both side (client and server), the infos are stored in this structure and the structure can know the count with `comp_info.size()`. Isn't it the case? Is there a case where `num_comp` has value, but `comp_info` doesn't have infos?

joohok (approved):

src/base/memory_monitor_common.h
Rudeus: I agree that using the `size()` function of the vector allows us to determine the number of elements inside the vector without needing `num_comp` or `num_subcomp`. However, during the process of unpacking packed information at the client, the client needs to know the count of elements, so it is indeed necessary to include the count information in the sending buffer of server when packing the information. If the struct does not have this count information separated internally, developers may find it inconvenient to pack additional information separately from the information contained in the struct. Since the total amount of data transmitted through the network remains unchanged, I believe it would be more advantageous in terms of readability and manageability to keep this count information.

wrlawodms: hm.. I don't agree that it's "necessary", but the advantage of removing it is not that serious, and I see you have a point. I'll follow your choice :)

src/transaction/log_tran_table.c
```diff
@@ -1457,6 +1458,28 @@ xlogtb_dump_trantable (THREAD_ENTRY * thread_p, FILE * out_fp)
...
+	{
+	  tran_info.push_back (std::make_pair (tdes->trid, tdes->cur_mem_usage.load ()));
```
wrlawodms, line 1478: Nit. you can shoten this :).
```suggestion
	  tran_info.push_back ({tdes->trid, tdes->cur_mem_usage.load ()});
```
or
```suggestion
	  tran_info.emplace_back (tdes->trid, tdes->cur_mem_usage.load ());
```

src/base/memory_monitor_sr.cpp
```diff
@@ -402,29 +452,82 @@ namespace cubperf
...
   {
-    int module_idx = get_module_idx (stat_id);
+    std::vector <std::pair <int, uint64_t>> tran_info;
+
+    logtb_get_tran_memory_info_nolatch (tran_info);
+
+    if (tran_count == 0)
+      {
+	// transaction option with no tran-count option
+	info.num_tran = tran_info.size ();
```
wrlawodms: > I added comments with branching based on the options because the explanation could be confusing. Default option of memmon has to show 5 transaction in descending order (that means, tran_count = 5).

1. When tran_count is not given, I was told that only 5 trxs are shown, wasn't it? I barely remember, but the comment above said that.. so I'm asking.
2. Aren't it needed to be sorted?

wrlawodms (approved):

src/base/memory_monitor_sr.cpp
```diff
@@ -402,29 +452,85 @@ namespace cubperf
...
+      return module_info1.stat.cur_stat > module_info2.stat.cur_stat;
+    };
+    std::sort (info, info + MMON_MODULE_LAST, comp);
   }
 
-  void memory_monitor::add_stat (MMON_STAT_ID stat_id, int64_t size)
+  void memory_monitor::aggregater::get_transaction_info (MMON_TRAN_INFO &info, int tran_count) const
   {
-    int module_idx = get_module_idx (stat_id);
+    std::vector <std::pair <int, uint64_t>> tran_info;
```
wrlawodms: Nit
```suggestion
    assert (tran_count >= 0);
    std::vector <std::pair <int, uint64_t>> tran_info;
```

wrlawodms (approved):

joohok (approved):

src/transaction/log_impl.h
```diff
@@ -1201,6 +1201,9 @@ extern bool logtb_has_deadlock_priority (int tran_index);
 /* For Debugging */
 extern void xlogtb_dump_trantable (THREAD_ENTRY * thread_p, FILE * out_fp);
 
+/* For memory monitoring */
+extern void logtb_get_tran_memory_info_nolatch (std::vector < std::pair < int, uint64_t >> &tran_info);
```
hornetmj:
```suggestion
/* For memory monitoring */
// *INDENT-OFF*
using LOG_TRAN_MEM_INFO std::vector <std::pair <int, uint64_t>>
// *INDENT-ON*

extern void logtb_get_tran_memory_info_nolatch (LOG_TRAN_MEM_INFO &tran_info);
```

src/base/memory_monitor_sr.cpp
```diff
@@ -115,7 +116,7 @@ namespace cubperf
       mmon_module &operator = (const mmon_module &) = delete;
       mmon_module &operator = (mmon_module &&) = delete;
 
-      virtual int aggregate_stats (const MMON_MODULE_INFO &info);
+      virtual void aggregate_stats (bool get_summary, MMON_MODULE_INFO &info) const;
```
```suggestion
      virtual void aggregate_stats (bool is_summary, MMON_MODULE_INFO &info) const;
```

src/transaction/log_tran_table.c
Rudeus: @joohok checked and fix it.

src/transaction/log_impl.h
Rudeus: Done.

src/base/memory_monitor_sr.cpp

src/transaction/log_tran_table.c
```diff
@@ -1457,6 +1458,30 @@ xlogtb_dump_trantable (THREAD_ENTRY * thread_p, FILE * out_fp)
...
 }
 
+/*
+ * logtb_get_tran_memory_info_nolatch - get the transaction memory information
+ *                                      with nolatch
+ *
+ * return: nothing
+ *
+ */
+// *INDENT-OFF*
```
hornetmj:
```suggestion
```

src/transaction/log_tran_table.c
```diff
@@ -1457,6 +1458,30 @@ xlogtb_dump_trantable (THREAD_ENTRY * thread_p, FILE * out_fp)
...
+void
+logtb_get_tran_memory_info_nolatch (LOG_TRAN_MEM_INFO &tran_info)
+// *INDENT-ON*
```
hornetmj:
```suggestion
```

hornetmj (approved):

PR #4493, opened 2023-07-17
[LETS-754] Reset main connection after connecting PSes

wrlawodms:
http://jira.cubrid.org/browse/LETS-754

I thought that "resetting the main connection at the end of `connect()`" would allow the catchup during ATS boot, but it turns out its' not the case because the catch-up process also requires prior_lsa which hasn't initialized when making connection_handler.
So, It'd be better to reset the main connection after connecting all PSes.

- Reset the main connection after establishing all connections on `boot()`
- Reset the main connection after checking and reconnecting all PSes by `ps_connector`.
- Fix `reset_main_connection` not to mix up the main connection by widening the lock scope. It was that a main connection elected first can be set later.

Sidenote:
When the catch-up process is implemented, the resetting should be also revised. We will not be able to set a new connection as the main connection until the asynchronous catch-up process is done. My plan for now is to extend `ps_connector` to also check and reset the main connection periodically.

### DIFF
--- a/src/server/tran_server.cpp
+++ b/src/server/tran_server.cpp
@@ -414,6 +414,8 @@ int
 tran_server::reset_main_connection ()
 {
   auto &conn_vec = m_page_server_conn_vec;
+  auto ulock = std::unique_lock<std::shared_mutex> { m_main_conn_mtx };
+
   /* the priority to select the main connection is the order in the container */
   const auto main_conn_cand = std::find_if (conn_vec.cbegin (), conn_vec.cend (),
 			      [] (const auto &conn)
@@ -427,15 +429,12 @@ tran_server::reset_main_connection ()
       return ER_CONN_NO_PAGE_SERVER_AVAILABLE;
     }
 
-  {
-    auto ulock = std::unique_lock<std::shared_mutex> { m_main_conn_mtx };
-    if (m_main_conn != main_conn_cand->get ())
-      {
-	m_main_conn = main_conn_cand->get ();
-	er_log_debug (ARG_FILE_LINE, "The main connection is set to %s.\n",
-		      m_main_conn->get_channel_id ().c_str ());
-      }
-  }
+  if (m_main_conn != main_conn_cand->get ())
+    {
+      m_main_conn = main_conn_cand->get ();
+      er_log_debug (ARG_FILE_LINE, "The main connection is set to %s.\n",
+		    m_main_conn->get_channel_id ().c_str ());
+    }
 
   return NO_ERROR;
 }
--- a/src/server/tran_server.cpp
+++ b/src/server/tran_server.cpp
@@ -413,17 +413,16 @@ tran_server::disconnect_all_page_servers ()
 int
 tran_server::reset_main_connection ()
 {
-  auto &conn_vec = m_page_server_conn_vec;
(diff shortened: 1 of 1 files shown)


hornetmj (approved):
LGTM

src/server/tran_server.cpp
```diff
@@ -725,14 +719,27 @@ tran_server::ps_connector::try_connect_to_all_ps (cubthread::entry &)
...
+	    }
 	}
 
       if (m_terminate.load ())
 	{
 	  break;
 	}
     }
+
+  if (newly_connected)
```
joohok, line 733:
```suggestion
  if (newly_connected && !m_terminate.load ())
```
I'm not sure if this `reset_main_connection ()` is required when `m_terminate` is true, what do you think? 

I don't think there will be any functional issues, but I just wonder if there is a need to change the main connection while disconnecting.

src/server/tran_server.cpp
```diff
@@ -420,6 +414,8 @@ int
 tran_server::reset_main_connection ()
 {
   auto &conn_vec = m_page_server_conn_vec;
```
cristiarg: Why is `conn_vec` still needed?
And why is it outside the lock?

src/server/tran_server.cpp
```diff
@@ -420,6 +414,8 @@ int
...
+  auto ulock = std::unique_lock<std::shared_mutex> { m_main_conn_mtx };
+
   /* the priority to select the main connection is the order in the container */
   const auto main_conn_cand = std::find_if (conn_vec.cbegin (), conn_vec.cend (),
```
```suggestion
  const auto main_conn_cand_it = std::find_if (conn_vec.cbegin (), conn_vec.cend (),
```
because it is an iterator.

src/server/tran_server.cpp
```diff
@@ -725,14 +719,27 @@ tran_server::ps_connector::try_connect_to_all_ps (cubthread::entry &)
...
-	  (void) conn->connect ();
+	  if (conn->connect () == NO_ERROR)
+	    {
+	      newly_connected = true;
+	    }
 	}
 
       if (m_terminate.load ())
 	{
 	  break;
```
Why not `return` in case of `m_terminate`?

cristiarg (approved):

src/server/tran_server.cpp
wrlawodms: It was the same, but with this change, it gets different. I'll make it return.

src/server/tran_server.cpp
wrlawodms: It's just making a reference. I think it is safe without the lock, isn't it?

I think I can remove it, by the way.

src/server/tran_server.cpp
wrlawodms: done.

src/server/tran_server.cpp
wrlawodms: You're right. It's been also fixed with the change from the `break` to `return` above.

Actually, there is still room for it to be called with your suggestion and the current change because `m_terminate` can be set after checking the condition. However, I think it is acceptable.

joohok: OK, I understand.

joohok (approved):

PR #4497, opened 2023-07-18
[CBRD-24842] Rewrite error messages for hostname lookup

eido5:
http://jira.cubrid.org/browse/CBRD-24842

**Description**
- in addition to #3740 ,
- fix below:

1. Delete some er_set() functions.
2. Delete deprecated error codes.
3. Fix error message flow before broker shared memory allocated.
4. Fix the err_set() function according to #4414 error message edited.

**Remarks**


msg/en_US/cubrid.msg
```diff
@@ -1438,15 +1438,10 @@ Check the path of the key file (_keys) and make sure that it includes proper key
 1353 Internal CAS error : %1$d 
 1354 Internal CAS error : %1$d, %2$s
 
-1355 %1$s "%2$s" is invalid format at line %3$d of the file "%4$s".
-1356 Hostname "%1$s" is too long, it should be less than or equal to 256 at line %2$d of the file "%3$s".
-1357 Hostname "%1$s" already exists, a hostname cannot be duplicated with other IP addresses at line %2$d of the file "%3$s".
-1358 Both, the IP address and hostname, must be written completely at line %1$d of the file "%2$s".
-1359 Unable to look up "%1$s" in the cubrid_hosts.conf.
```
eido5, line 1445: I deleted all the error message to make same result using by etc/hosts.
Instead of using these, edited legacy error code will be used as shown in #4414 .

src/broker/broker_shm.h
```diff
@@ -694,7 +694,11 @@ int uw_sem_post (sem_t * sem_t);
 int uw_sem_destroy (sem_t * sem_t);
 #endif
 T_SHM_BROKER *broker_shm_initialize_shm_broker (int master_shm_id, T_BROKER_INFO * br_info, int br_num, int acl_flag,
-						char *acl_file);
+						char *acl_file, unsigned char *ip_addr);
 T_SHM_APPL_SERVER *broker_shm_initialize_shm_as (T_BROKER_INFO * br_info_p, T_SHM_PROXY * shm_proxy_p);
 
+#if !defined(WINDOWS)
+int get_host_ip (unsigned char *ip_addr);
```
eido5: To use get_host_ip() function in the broker_shm.c, I moved the get_host_ip() to the broker_shm.h.
And then the paramater of broker_shm_initialize_shm_broker() is changed to send unsigned char *ip_addr

eido5:
I tested whether error message is removed.

These messages are removed:

- unknown host : eido5-53
- failed to initialize broker shared memory
- failed to initialize gateway shared memory
- "Hostname "[host name]" is invalid format at line [n] of the file "$CUBRID/conf/cubrid_hosts.conf"."

This message is changed:

- ERROR: Unable to restart/initialize the database server. Cannot find ("(unknown)") in "etc/hosts". Please check permissions of the file or if there is the hostname.

src/broker/broker_admin_pub.c
```diff
@@ -322,8 +323,14 @@ admin_start_cmd (T_BROKER_INFO * br_info, int br_num, int master_shm_id, bool ac
     }
   chdir (envvar_bindir_file (path, BROKER_PATH_MAX, ""));
 
+  /* cannot execute broker unless find host name */
+  if (get_host_ip (ip_addr) < 0)
+    {
+      return -1;
+    }
+
```
kisoo-han: To minimize modifications, I propose the alternative:

1. add 
`#include "tcp.h"`
under 
#include "porting.h
2. add following in the decl. part.
`char hostname [CUB_MAXHOSTNAMELEN];`
4. replace get_host_ip () to the following
```
  if (GETHOSTNAME (hostname, CUB_MAXHOSTNAMELEN) != 0)
    {
      return -1;
    }

```

eido5: I see. I will be better. Thank you.

kisoo-han (approved):

mhoh3963 (approved):

src/broker/broker_admin_pub.c
```diff
@@ -322,6 +324,12 @@ admin_start_cmd (T_BROKER_INFO * br_info, int br_num, int master_shm_id, bool ac
     }
   chdir (envvar_bindir_file (path, BROKER_PATH_MAX, ""));
 
+  /* cannot execute broker initialize unless success host look-up */
+  if (GETHOSTNAME (hostname, CUB_MAXHOSTNAMELEN) != 0)
+    {
+      return -1;
```
airnet73, line 330: Wouldn't an error message be necessary?
For reference, admin_start_cmd() This function executes broker and gateway. And the cubrid_utility.log log is recorded.

eido5: Yes, it seems unnecessary.

css_gethostname() function record error message when the host look up is failed. 

So, in this code, we try to check whether host look up is executed, error message is not necessary.

airnet73: all right.

airnet73 (approved):

PR #4503, opened 2023-07-20
[CBRD-24896] Problem when accesing system table/views with search condition comparing set values

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-24896

- Modified to determine that index scan is impossible in the following cases
```
set = set
{col1,col2} = {1,2}
```
- fix so that only collection types written in parentheses can index-scan.


src/optimizer/query_graph.c
```diff
@@ -2195,20 +2195,14 @@ qo_analyze_term (QO_TERM * term, int term_type)
 		  lhs_indexable = true;
 		}
 	    }
-	  else if (pt_is_multi_col_term (lhs_expr))
+	  else if (pt_is_multi_col_term (lhs_expr) && lhs_expr->flag.is_paren)
 	    {
 	      /* multi column case (attr,attr,...) is indexable for RANGE, EQ operation */
```
youngjinj: It is trivial.
The comment about 'EQ operation' should be removed.

shparkcubrid:
```suggestion
	      /* multi column case (attr,attr,...) is indexable for RANGE operation */
```

shparkcubrid: I fixed it.

src/optimizer/query_graph.c
```diff
@@ -2249,7 +2243,7 @@ qo_analyze_term (QO_TERM * term, int term_type)
 			}
 		      segs++;
 		    }
-		  if (!is_find_local_name)
+		  if (!is_find_local_name || segs <= 1)
```
youngjinj, line 2246: Are there test scenarios that require this change?

youngjinj (approved):

xmilex-git (approved):

PR #4504, opened 2023-07-20
[LETS-777] small refactoring of log prior sender

cristiarg:
http://jira.cubrid.org/browse/LETS-777

Extract the code that actually sends the packaged message in a function - which will be reused afterwards from the outside.

Small optimization to:
- copy messages to all but the last sink hook
- move the message into the last sink hook as it is of no use afterwards

### DIFF
--- a/src/transaction/log_prior_send.cpp
+++ b/src/transaction/log_prior_send.cpp
@@ -32,21 +32,35 @@ namespace cublog
       {
 	return;
       }
-    const std::string message = prior_list_serialize (head);
+    std::string message = prior_list_serialize (head);
 
     if (prm_get_bool_value (PRM_ID_ER_LOG_PRIOR_TRANSFER))
       {
 	const log_prior_node *tail;
 	for (tail = head; tail->next != nullptr; tail = tail->next);
 	_er_log_debug (ARG_FILE_LINE,
-		       "[LOG_PRIOR_TRANSFER] Sending. Head lsa %lld|%d. Tail lsa %lld|%d. Message size = %zu.\n",
+		       "[LOG_PRIOR_TRANSFER] Sending. head_lsa = %lld|%d tail_lsa = %lld|%d. Message size = %zu.\n",
 		       LSA_AS_ARGS (&head->start_lsa), LSA_AS_ARGS (&tail->start_lsa), message.size ());
       }
 
+    send_serialized_message (std::move (message));
+  }
+
+  void
+  prior_sender::send_serialized_message (std::string &&message)
+  {
     std::unique_lock<std::mutex> ulock (m_sink_hooks_mutex);
-    for (auto &sink : m_sink_hooks)
+    // copy the message into every sink but the last..
+    for (int index = 0; index < ((int)m_sink_hooks.size () - 1); ++index)
       {
-	(*sink) (std::string (message));
+	const sink_hook_t *const sink_p = m_sink_hooks[index];
+	(*sink_p) (std::string (message));
+      }
+    // ..and optimize by moving the message into the last sink, because it is of no use afterwards
+    if (m_sink_hooks.size () > 0)
+      {
+	const sink_hook_t *const sink_p = m_sink_hooks[m_sink_hooks.size () - 1];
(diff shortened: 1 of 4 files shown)


src/transaction/log_prior_send.cpp
```diff
@@ -32,21 +32,35 @@ namespace cublog
...
+    for (int index = 0; index < ((int)m_sink_hooks.size () - 1); ++index)
+      {
+	const sink_hook_t *const sink_p = m_sink_hooks[index];
+	(*sink_p) (std::string (message));
+      }
+    // ..and optimize by moving the message into the last sink, because it is of no use afterwards
+    if (m_sink_hooks.size () > 0)
       {
-	(*sink) (std::string (message));
+	const sink_hook_t *const sink_p = m_sink_hooks[m_sink_hooks.size () - 1];
```
hornetmj, line 62:
```suggestion
	const sink_hook_t *const sink_p = m_sink_hooks[index];
```

src/transaction/log_prior_send.cpp
```diff
@@ -32,21 +32,35 @@ namespace cublog
...
     std::unique_lock<std::mutex> ulock (m_sink_hooks_mutex);
-    for (auto &sink : m_sink_hooks)
+    // copy the message into every sink but the last..
+    for (int index = 0; index < ((int)m_sink_hooks.size () - 1); ++index)
+      {
+	const sink_hook_t *const sink_p = m_sink_hooks[index];
+	(*sink_p) (std::string (message));
+      }
+    // ..and optimize by moving the message into the last sink, because it is of no use afterwards
+    if (m_sink_hooks.size () > 0)
```
hornetmj, line 60:
```suggestion
```

hornetmj (approved):

src/transaction/log_prior_send.cpp
cristiarg: I am not sure that the function cannot be called with an empty `m_sink_hooks`.

Actually there is this issue - http://jira.cubrid.org/browse/LETS-642 - which should deal with the empty list issue.

The issue described by LETS-642 is not that important.
It's more important to have an as optimal as possible packing and dispatch of the information on the happy path.
By this, I mean when there are no sinks in the list, the packing of the list into a string is useless. But this situation should happen only in corner-case scenarios or, at best, be a transitory state of the system, thus not the "happy path".

src/transaction/log_prior_send.cpp
cristiarg: If:
- sink hook list is empty: the `index` resulting from the previous `for` remains `0` --> out of bounds access
- sink hook list has at least one element: the `index` value resulting from the previous loop remains at the desired value - `m_sink_hook.size () -1` - so, OK

In fact this change is mutually exclusive with the previous one - removing the check `if (m_sink_hooks.size () > 0)`.
I prefer to have them both for explicitness.

wrlawodms (commented):
Don't you say that we should change the parameter for requests to a lvalue reference, not rvalue? I remember that we talked about it and this comes from the retrial of push_request and send_receive on the main connection. If we are going to make all the request interfaces get lvalue parameter, do we need this fix?

cristiarg:
> Don't you say that we should change the parameter for requests to a lvalue reference, not rvalue? I remember that we talked about it and this comes from the retrial of push_request and send_receive on the main connection. If we are going to make all the request interfaces get lvalue parameter, do we need this fix?

I do remember about it - it's written in http://jira.cubrid.org/browse/LETS-744 .
This PR is a small preliminary part of http://jira.cubrid.org/browse/LETS-777 .
The subsequent PR of LETS-777 will implement the PS relaying the log prior messages asynchronously to the connected PTSs without unpackage-repackage .
As such this implementation and LETS-744 are independent in my view - one does not exclude the other.

Yes, you could say that the small bit of optimization with the copy/move is useless - indeed.
Maybe it would have been simpler if I had provided the full implementation of LETS-777 in one PR. :)
But I wanted to make it easier to digest. :)

wrlawodms (approved):

PR #4505, opened 2023-07-21
[CBRD-24899] check push predicate and quoted name for dblink

beyondykk9:

First, in the dblink query, the function must be excluded from "predicate push".
Second, in the case of other DBMS such as Oracle, it is necessary to enclose the table name or column name in double quotes in the query to be transmitted to dblink.


src/parser/type_checking.c
```diff
@@ -18958,6 +18958,12 @@ pt_fold_const_expr (PARSER_CONTEXT * parser, PT_NODE * expr, void *arg)
       return expr;
     }
 
+  /* if dblink query, do not constant fold */
+  if (parser->dblink_remote)
+    {
+      return expr;
+    }
```
hgryoo: When is dblink_remote set? It seems to affect the expressions unrelated to dblink pushable terms.

beyondykk9: It is not easy to determine whether it was a dblink-related expr, so I handled it simply.

hgryoo: Does it affect the SELECT query as well? If so, the results of the SELECT query may be inconsistent depending on whether or not the DBLINK spec exists.

beyondykk9: I added the function to check dblink-related expr.

hgryoo (approved):

src/parser/type_checking.c
```diff
@@ -18916,6 +18916,34 @@ pt_evaluate_db_value_expr (PARSER_CONTEXT * parser, PT_NODE * expr, PT_OP_TYPE o
   return 0;
 }
 
+/*
+ * pt_fold_const_expr () - check dblinke-related expression
```
hyunikn: This comment should be updated.

hyunikn (approved):

src/parser/view_transform.c
```diff
@@ -4186,8 +4236,18 @@ mq_copypush_sargable_terms_helper (PARSER_CONTEXT * parser, PT_NODE * statement,
   is_outer_joined = mq_is_outer_join_spec (parser, spec);
 
   /* term(predicate) check */
+  in_spec = statement->info.query.q.select.from;
   for (term = statement->info.query.q.select.where; term; term = term->next)
     {
+      /* check for dblink's function term */
+      if (in_spec->info.spec.derived_table_type == PT_DERIVED_DBLINK_TABLE)
```
ctshim, line 4243: Can't there be multiple tables in the From clause?
Wouldn't it be necessary to check in_spec->next (or_next?)?

src/parser/type_checking.c
beyondykk9: OK, good point.

src/parser/view_transform.c
beyondykk9: This routine is called for each dblinked spec. So, it's no need for inspection of all specs.

PR #4507, opened 2023-07-24
[CBRD-24867] rev: regression bug-fix related to DBLINK - CCI_U_TYPE, no serve name

beyondykk9:

buf-fix list
1) DATETIMELTZ was not processed for ext-type
    ext-type should be converted to CCI_U_TYPE
2) SELECT DBLINK ('url', 'SELECT ... FROM ...') was not processed.
    we should consider the query with URL instead server-name
3) alias match check
    not need to check for table@server (i.e. SELECT * FROM ...)
    added **m_attr_star** to S_REMOTE_TBL_COLS to identified (SELECT * FROM...)
4) remake select list
   only when the column-def list not exists
5) quoted name is processed properly
    the quoted name like "From" is transmitted to "From" as a quoted name
    because the remote query should be executed on other DBMS like Oracle.
    The ["From"] and \`"From"\` and """From""" is will be "From" at remote server even though the DBMS is CUBRID or not.
6) the functions including generic-function should passed to remote server
    the parser set the flag cannot_prepare not to prepare for genric-function
    we should clear the flag for remote DML.


src/parser/name_resolution.c
```diff
@@ -1201,74 +1213,78 @@ pt_bind_scope (PARSER_CONTEXT * parser, PT_BIND_NAMES_ARG * bind_arg)
...
 		  if (pt_resolve_dblink_server_name (parser, table, NULL) != NO_ERROR)
 		    {
 		      return;
 		    }
+		}
 
-		  /* remote table's column list */
-		  rmt_tbl_cols = new (std::nothrow) S_REMOTE_TBL_COLS;
+	      /* remote table's column list */
+	      rmt_tbl_cols = new (std::nothrow) S_REMOTE_TBL_COLS;
```
ctshim, line 1233: You should check for memory allocation failures.
Please add at least assert_release(rmt_tbl_cols != NULL)

ctshim, line 1233: If it returns without passing "parser->dblink_remote = remote;" in line 1286, you must delete it.

beyondykk9: I fixed it.

beyondykk9: I fixed it.

hgryoo:
Can you give test cases for each regression? They would help me to understand the changes in this PR.

beyondykk9:
> Can you give test cases for each regression? They would help me to understand the changes in this PR.

see the last comment on JIRA issue CBRD-24867.

hgryoo (approved):

src/parser/name_resolution.c
```diff
@@ -1201,74 +1216,82 @@ pt_bind_scope (PARSER_CONTEXT * parser, PT_BIND_NAMES_ARG * bind_arg)
...
 		    }
 
-		  /* error from dblink(server, 'select ... from ...' */
-		  if (err < 0)
+		  if (table->info.dblink_table.cols == NULL)
 		    {
-		      return;
+		      if (pt_remake_dblink_select_list (parser, &spec->info.spec, rmt_tbl_cols) != NO_ERROR)
+			{
+			  return;
```
hyunikn: This line must be 'goto error_exit;' , instead of 'return' to free rmt_tbl_cols.

src/parser/name_resolution.c
```diff
@@ -5074,22 +5113,20 @@ pt_remake_dblink_select_list (PARSER_CONTEXT * parser, PT_SPEC_INFO * class_spec
...
+	  error = ER_DBLINK;
+	  goto error_exit;
 	}
 
       id_node->info.name.original = "c";
       if ((dblink_table->cols = pt_mk_attr_def_node (parser, id_node, NULL)) == NULL)
 	{
-	  return ER_FAILED;
+	  error = ER_DBLINK;
+	  goto error_exit;
```
hyunikn: It seems necessary to restore parser->custom_print before returning at line 5127.
An easy way to do this is to 'goto error_exit' instead of 'return NO_ERROR'. 
In this case, 'clean_up' or something similar can be a better name instead of the label 'error_exit' .

src/parser/name_resolution.c
beyondykk9: Ok, good point.

src/parser/name_resolution.c
```diff
@@ -5119,14 +5158,18 @@ pt_remake_dblink_select_list (PARSER_CONTEXT * parser, PT_SPEC_INFO * class_spec
...
 	    }
 	  attr_def_node = attr_def_node ? parser_append_node (tmp, attr_def_node) : tmp;
 	}
     }
 
   dblink_table->cols = attr_def_node;
-  return NO_ERROR;
+
+error_exit:
+  parser->custom_print &= ~PT_PRINT_SUPPRESS_FOR_DBLINK;
```
hyunikn, line 5170: Does the original value of parser->custom_print always have the bit PT_PRINT_SUPPRESS_FOR_DBLINK ?
Can we always delete the bit as this line does?
Storing the original value and restoring it at this line, which is a frequent pattern through out the code, seems more general way of handling  parser->custom_print.

src/parser/name_resolution.c
beyondykk9: Ok, good point.

src/parser/name_resolution.c
beyondykk9: printing the tree is called at various places through whole parser, compile, and optimize logic.
the custom_print should be kept the options to fit for purpose.

src/parser/parser_support.c
```diff
@@ -10769,42 +10769,46 @@ static PT_NODE *
...
+	pt_append_string (parser, derived_spec->info.dblink_table.remote_table_name,
+			  class_spec_info->entity_name->info.name.resolved);
+      derived_spec->info.dblink_table.remote_table_name =
+	pt_append_string (parser, derived_spec->info.dblink_table.remote_table_name, ".");
     }
+  derived_spec->info.dblink_table.remote_table_name =
+    pt_append_string (parser, derived_spec->info.dblink_table.remote_table_name,
+		      class_spec_info->entity_name->info.name.original);
+
+  parser->custom_print &= ~PT_PRINT_SUPPRESS_FOR_DBLINK;
```
hyunikn, line 10804: Can the original value of parser->custom_print have PT_PRINT_SUPPRESS_FOR_DBLINK bit?
If it can, deleting the bit does not always restore the original value.

beyondykk9: set and clear is always pair, so the flag is not set erroneously.

src/query/execute_statement.c
```diff
@@ -13687,20 +13692,21 @@ do_execute_insert (PARSER_CONTEXT * parser, PT_NODE * statement)
 
   CHECK_MODIFICATION_ERROR ();
 
-  if (statement->xasl_id == NULL)
+  /* for dblink: no need to check xal_is is NULL */
```
hyunikn: xal_is means xasl_id?

beyondykk9: ah ha, mis-spelled

hyunikn (approved):

src/parser/name_resolution.c
```diff
@@ -4836,7 +4868,8 @@ pt_dblink_table_fill_attr_def (PARSER_CONTEXT * parser, PT_NODE * attr_def_node,
   int is_default = 0;
 
   attr_def_node->data_type = NULL;
-  attr_def_node->type_enum = pt_type[attr->type_idx];
+  /* it needs to convert ext type to CCI_U_TYPE */
+  attr_def_node->type_enum = pt_type[(T_CCI_U_TYPE) CCI_GET_COLLECTION_DOMAIN (attr->type_idx)];
```
ctshim, line 4872: There is a dblink_get_basic_utype() function for a similar purpose. How about using this?
`attr_def_node->type_enum = dblink_get_basic_utype(attr->type_idx)`

src/parser/parse_tree.h
```diff
@@ -932,8 +932,8 @@ enum pt_custom_print
   PT_PRINT_SUPPRESS_SERVER_NAME = (0x1 << 25),
   /* suppress next_value to serial_next_value(...) or current_value to serial_current_value(...) */
   PT_PRINT_SUPPRESS_SERIAL_CONV = (0x1 << 26),
-  /* suppress target-name at 'DELETE target-name FROM' for dblink's other DBMS */
-  PT_PRINT_SUPPRESS_DELETE_TARGET = (0x1 << 27)
+  /* suppress print various generated functions including suppress delete targe for dblink */
+  PT_PRINT_SUPPRESS_FOR_DBLINK = (0x1 << 27),
```
ctshim, line 936: There is virtually no difference from PT_SUPPRESS_QUOTES. 
How about redefining only the naming for distinction?

`#define PT_PRINT_SUPPRESS_FOR_DBLINK   (PT_SUPPRESS_QUOTES)`

PR #4511, opened 2023-07-25
[CBRD-24872] [win] broker start ctrl-service wait for cas invocation

kisoo-han:
http://jira.cubrid.org/browse/CBRD-24872

**Description**
* in Windows, **broker_start** is executed by a windows service.
* Windows Service Control Manager (SCM) spawns the service process for the broker_startup
* The broker_start service request may end with a timeout event in the middle of creating a CAS process.
  (Even in this case, the creation of the cas process proceeds normally)

**Implementation**
* monitor TIMEOUT event and wait 200ms for normal termination of **broker_startup**
  (repeat it until the timeout event is not caught or retry count reached to 50)
* 40 is for slow Windows machine

### DIFF
--- a/src/win_tools/ctrlservice/ctrlservice.cpp
+++ b/src/win_tools/ctrlservice/ctrlservice.cpp
@@ -83,6 +83,7 @@ int
 _tmain (int argc, char *argv[])
 {
   bool rc;
+  int broker_start_loop_cnt_max = 40;
 
   // Install a Service if -i switch used
   if (argc == 2)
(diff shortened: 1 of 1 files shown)


airnet73 (approved):

src/win_tools/ctrlservice/ctrlservice.cpp
```diff
@@ -83,6 +83,7 @@ int
 _tmain (int argc, char *argv[])
 {
   bool rc;
+  int broker_start_loop_cnt_max = 40;
```
mhoh3963: How about setting it to maximum 10 seconds? That is, changing the value from 40 to 50.

kisoo-han: Okay, I'll change that.

mhoh3963 (approved):

PR #4513, opened 2023-07-26
[CBRD-24909] dblink related synonym processing

beyondykk9:
http://jira.cubrid.org/browse/CBRD-24909

The current system checks the user set in the synonym target when creating or altering a synonym. However, if the target is the dblink table of a remote server, user check cannot be performed. In this case, the user check must be replaced with the current user, and the target Also, the text entered by the user who created it must be reflected as it is.

CREATE SYNONYM rms1 for user_a.tbl1@srv1;
CREATE SYNONYM rms2 for tbl2@srv1;
ALTER SYNONYM rms1 for user_b.tbl1@srv1;
ALTER SYNONYM rms2 for tbl2@srv2;

The above DDL should be executed successfully.


src/parser/csql_grammar.y
```diff
@@ -3156,6 +3156,14 @@ create_stmt
 			      {
 				PT_ERROR (this_parser, node, "PUBLIC SYNONYM is not supported.");
 			      }
+
+			    if (node->info.synonym.target_name)
+			      {
+			        if (strchr (node->info.synonym.target_name->info.name.original, '@'))
+			          {
+				    node->info.synonym.dblinked = 1;
```
youngjinj, line 3164: How about setting 'node->info.synonym.dblinked = 1' in the following rule?
```
class_name_for_synonym
    ...
    | class_name_with_server_name
```

src/parser/parse_tree.h
```diff
@@ -3525,6 +3526,7 @@ struct pt_synonym_info
   PT_NODE *comment;		/* PT_VALUE */
   unsigned or_replace:1;	/* OR REPLACE clause for CREATE SYNONYM */
   unsigned if_exists:1;		/* IF EXISTS clause for DROP SYNONYM */
+  unsigned dblinked:1;		/* server name specified */
```
youngjinj: I wish there were some prefix rules.
```
is_ ...
if_ ...
has_ ...
with_ ...
...
```

src/parser/parser_support.c
```diff
@@ -10491,8 +10491,15 @@ pt_set_user_specified_name (PARSER_CONTEXT * parser, PT_NODE * node, void *arg,
...
-	    PT_SYNONYM_TARGET_OWNER_NAME (node) = pt_name (parser, target_owner_name);
+	    if (PT_SYNONYM_DBLINKED (node))
+	      {
+		PT_SYNONYM_TARGET_OWNER_NAME (node) = pt_name (parser, synonym_owner_name);
+	      }
+	    else
+	      {
+		assert (target_owner_name != NULL);
+		PT_SYNONYM_TARGET_OWNER_NAME (node) = pt_name (parser, target_owner_name);
+	      }
```
youngjinj: How about checking when setting a value for `target_owner_name`?
```
if (PT_SYNONYM_DBLINKED (node))
  {
    target_owner_name = synonym_owner_name;
  }
else
  {
    target_owner_name = pt_get_qualifier_name (parser, PT_SYNONYM_TARGET_NAME (node));
  }
```

src/query/execute_statement.c
```diff
@@ -17775,13 +17775,21 @@ do_alter_synonym (PARSER_CONTEXT * parser, PT_NODE * statement)
   else
     {
       /* PT_SYNONYM_TARGET_NAME (statement) != NULL */
+      if (PT_SYNONYM_DBLINKED (statement))
+	{
+	  sprintf (target_name_buf, "%s", (PT_NAME_ORIGINAL (PT_SYNONYM_TARGET_NAME (statement))));
+	  /* target_owner = synonym_owner */
+	  target_owner_obj = db_find_user (PT_NAME_ORIGINAL (PT_SYNONYM_OWNER_NAME (statement)));
```
youngjinj: In the `pt_set_user_specified_name` function, `target_owner_name` is set to `synonym_owner_name`.
So the branch to set `target_owner_obj` is unnecessary.
```
if (PT_SYNONYM_DBLINKED (node))
  {
    PT_SYNONYM_TARGET_OWNER_NAME (node) = pt_name (parser, synonym_owner_name);
  }
```

src/query/execute_statement.c
```diff
@@ -17775,13 +17775,21 @@ do_alter_synonym (PARSER_CONTEXT * parser, PT_NODE * statement)
   else
     {
       /* PT_SYNONYM_TARGET_NAME (statement) != NULL */
+      if (PT_SYNONYM_DBLINKED (statement))
+	{
+	  sprintf (target_name_buf, "%s", (PT_NAME_ORIGINAL (PT_SYNONYM_TARGET_NAME (statement))));
```
youngjinj: The reason for writing to target_name_buf is to change it to lower case.
If target_name doesn't need to be changed to lowercase, we can set the pointer directly to target_name without writing target_name_buf.

But doesn't the dblink's target_name really need to be changed to lower case?

src/parser/parser_support.c
youngjinj: Also need to change for 'case PT_ALTER_SYNONYM:'.

src/parser/csql_grammar.y
youngjinj: The use of '@' in PT_NAME is a special case.
How about setting a flag for PT_NAME as well?

src/parser/parse_tree.h
beyondykk9: Ok. I chose **is_dblinked**

src/parser/parser_support.c
beyondykk9: Ok, I changed the codes as you commented.

src/query/execute_statement.c
beyondykk9: but in case of not dblinked, it uses target_name_buf.
I'd like to keep the logic for readability.
because in case of "dblinked" the target_name_buf is not necceary but in vice-versa it's necessary.

src/query/execute_statement.c
beyondykk9: the target_obj is necessary for checking the owner in case of "not dblinked"

src/parser/csql_grammar.y
beyondykk9: then dblinked flag will be set at only this point. The flag would be not necessary.

youngjinj: Thank you.

src/parser/csql_grammar.y
```diff
@@ -3188,6 +3185,9 @@ class_name_for_synonym
 		    }
 		  cname->info.name.original = pt_append_string (this_parser, cname->info.name.original, "@");
 		  cname->info.name.original = pt_append_string (this_parser, cname->info.name.original, sname->info.name.original);
+
+		  PT_NAME_INFO_SET_FLAG (cname, PT_NAME_INFO_SERVER_SPECIFIED);
```
youngjinj: Most of the rules check if a node is NULL.
How about checking if cname and sname are NULL for consistency?
```
if (cname && sname)
  {
    ...
  }
```

src/parser/csql_grammar.y
```diff
@@ -4028,6 +4039,11 @@ alter_stmt
 			      {
 				PT_ERROR (this_parser, node, "PUBLIC SYNONYM is not supported.");
 			      }
+
+			    if (PT_NAME_INFO_IS_FLAGED(node->info.synonym.target_name, PT_NAME_INFO_SERVER_SPECIFIED))
```
youngjinj:
```suggestion
			    if (PT_NAME_INFO_IS_FLAGED (PT_SYNONYM_TARGET_NAME (node), PT_NAME_INFO_SERVER_SPECIFIED))
```

src/parser/csql_grammar.y
```diff
@@ -3156,6 +3156,11 @@ create_stmt
...
```
youngjinj:
```suggestion
			    if (PT_NAME_INFO_IS_FLAGED (PT_SYNONYM_TARGET_NAME (node), PT_NAME_INFO_SERVER_SPECIFIED))
```

src/parser/parser_support.c
```diff
@@ -10461,6 +10461,14 @@ pt_set_user_specified_name (PARSER_CONTEXT * parser, PT_NODE * node, void *arg,
...
 	    else
 	      {
+		if (PT_SYNONYM_DBLINKED (node))
+		  {
+		    target_owner_name = synonym_owner_name;
+		  }
+		else
+		  {
+		    target_owner_name = pt_get_qualifier_name (parser, PT_SYNONYM_TARGET_NAME (node));
+		  }
```
youngjinj, line 10471:
```suggestion
		else
		  {
		    target_owner_name = pt_get_qualifier_name (parser, PT_SYNONYM_TARGET_NAME (node));
		  }
```

youngjinj, line 10471: It is redundant with 10456 lines.

youngjinj: This code should replace line 10456.

src/parser/parser_support.c
```diff
@@ -10491,6 +10499,14 @@ pt_set_user_specified_name (PARSER_CONTEXT * parser, PT_NODE * node, void *arg,
...
 	else
 	  {
+	    if (PT_SYNONYM_DBLINKED (node))
+	      {
+		target_owner_name = synonym_owner_name;
+	      }
+	    else
+	      {
+		target_owner_name = pt_get_qualifier_name (parser, PT_SYNONYM_TARGET_NAME (node));
+	      }
```
youngjinj: This code should replace line 10494.

src/query/execute_statement.c
```diff
@@ -17775,13 +17774,20 @@ do_alter_synonym (PARSER_CONTEXT * parser, PT_NODE * statement)
   else
     {
       /* PT_SYNONYM_TARGET_NAME (statement) != NULL */
+      if (PT_SYNONYM_DBLINKED (statement))
```
youngjinj: I thought of a code like this.
```
int
do_alter_synonym (PARSER_CONTEXT * parser, PT_NODE * statement)
{
  ...
  const char *target_name = NULL;
  char target_name_buf[DB_MAX_IDENTIFIER_LENGTH] = { '\0' };
  ...
  if (PT_SYNONYM_DBLINKED (statement))
    {
      target_name = PT_NAME_ORIGINAL (PT_SYNONYM_TARGET_NAME (statement));
    }
  else
    {
      sm_user_specified_name (PT_NAME_ORIGINAL (PT_SYNONYM_TARGET_NAME (statement)), target_name_buf, DB_MAX_IDENTIFIER_LENGTH);
      target_name = target_name_buf;
    }

  /* target_owner */
  target_owner_obj = db_find_user (PT_NAME_ORIGINAL (PT_SYNONYM_TARGET_OWNER_NAME (statement)));
  if (target_owner_obj == NULL)
    {
      ASSERT_ERROR_AND_SET (error);
      return error;
      }
  ...
  error = do_alter_synonym_internal (synonym_name, target_name, target_owner_obj, comment, FALSE, PT_SYNONYM_DBLINKED (statement));
```

youngjinj: It was included in commit e23317e.

src/parser/parser_support.c
youngjinj: It was included in commit e23317e.

src/parser/parser_support.c
youngjinj: It was included in commit e23317e.

src/parser/csql_grammar.y
youngjinj: It was included in commit e23317e.

src/parser/csql_grammar.y
youngjinj: It was included in commit e23317e.

src/parser/csql_grammar.y
youngjinj: It was included in commit e23317e.

youngjinj (approved):

PR #4515, opened 2023-07-26
[LETS-778] Set the main conn even when uses_remote_storage () is false

wrlawodms:
http://jira.cubrid.org/browse/LETS-778

It occurs because the main connection is not set. Even when `uses_remote_storage ()` is false, there can be remote storages, aka PS. Therefore, we have to set the main connection.

PS. `uses_remote_storage()` and the `remote_storage` parameter seems confusing to me. It seems to mean whether there is remote storage or not, but what it really mean is we are using the local storage. I think it'd better to change it to "uses_local_storage_only()` and adapt all the code using it since it's opposite meaning.

### DIFF
--- a/src/server/tran_server.cpp
+++ b/src/server/tran_server.cpp
@@ -125,8 +125,12 @@ tran_server::boot (const char *db_name)
       return error_code;
     }
 
-  if (uses_remote_storage ())
+  if (m_page_server_conn_vec.empty () == false)
     {
+      /*
+       * At least one PS is given by the configuration.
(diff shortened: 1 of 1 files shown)


cristiarg (approved):
@wrlawodms I have started a test pipeline with this branch. Will keep you posted.

joohok (approved):

cristiarg:
@wrlawodms I think we can merge this one. There are still issues but not related to this.

wrlawodms:
@cristiarg Thank you for checking.

hornetmj (approved):

PR #4516, opened 2023-07-26

hornetmj:

Purpose
- fix the incremental restoration error when the incremental backup volumes donot have some of the volume headers

Implementation
- set the volume header at the level 0 restoration only

Remarks
- see the jira issue first


hornetmj:
Not acceptable fix.

PR #4517, opened 2023-07-27
[CBRD-24882]  Extracting index information that requires a lot of OVFP reading during vacuum

ctshim:
http://jira.cubrid.org/browse/CBRD-24882

* Manages information about indexes that require more oid overflow pages to be read than the threshold during vacuum operation. 
* Provide additional information via vacuumdb --dump 
* Add 2 system parameter
- vacuum_ovfp_check_duration 
- vacuum_ovfp_check_threshold


src/query/ovfp_threahold_monitor.hpp
```diff
@@ -0,0 +1,127 @@
...
+#define LOCK_ITEMS_SIZE    (VACUUM_MAX_WORKER_COUNT)
+  private:
+    std::mutex  m_ovfp_monitor_mutex;
+    int         m_lock_arr[LOCK_ITEMS_SIZE];
+
+  public:
+    ovfp_monitor_lock();
+    void lock (int lock_index, int owner_id);
+    void unlock (int lock_index, int owner_id);
+};
```
hgryoo:
```suggestion
  public:
    ovfp_monitor_lock();
    
    // Deleting the copy constructor and copy assignment operator
    ovfp_monitor_lock(const ovfp_monitor_lock&) = delete;
    ovfp_monitor_lock& operator=(const ovfp_monitor_lock&) = delete;

    // Deleting the move constructor and move assignment operator
    ovfp_monitor_lock(NonCopyableNonMovable&&) = delete;
    ovfp_monitor_lock& operator=(ovfp_monitor_lock&&) = delete;
    
    void lock (int lock_index, int owner_id);
    void unlock (int lock_index, int owner_id);
};
```
Non-copyable and non-movable. If the lock array is copied or moved, It could cause unexpected behavior.

hgryoo:
CUBRID has the monitor folder in src. It would be better to move ovfp_threshold_monitor.[cpp|hpp] to the folder with the following name.
```
monitor_ovfp_threshold_monitor.[cpp|hpp]
```

note: file name typo (ovfp_threahold_monitor -> ovfp_threshold_monitor)

src/query/ovfp_threahold_monitor.hpp
```diff
@@ -0,0 +1,127 @@
...
+  private:
+    ovfp_monitor_lock    m_ovfp_lock;
+    ovfp_threshold       m_ovfp_threshold[VACUUM_MAX_WORKER_COUNT];
+
+    UINT64 m_over_secs;
+    char   m_since_time[32];
+
+  public:
+    int   m_threshold_pages;
+
```
hgryoo:
```suggestion
    char   m_since_time[32];
    int m_threshold_pages;

  public:
    inline int get_threshold_page_cnt () const {
        return m_threshold_pages;
    }
```
Instead of accessing a member variable, I recommend inline getter method.

src/query/ovfp_threahold_monitor.hpp
```diff
@@ -0,0 +1,127 @@
...
+ *
+ */
+#ifndef _OVFP_THRESHOLD_MONITOR_H_
+#define _OVFP_THRESHOLD_MONITOR_H_
+
+#if !defined (SERVER_MODE)
+#error Belongs to server module
+#endif
+
+#if defined (SERVER_MODE)
```
hyunikn: This "#if defined" looks redundant with the lines 26 ~ 28.

src/query/vacuum.c
```diff
@@ -650,6 +651,9 @@ struct vacuum_dropped_files_rcv_data
 };
 
 bool vacuum_Is_booted = false;
+#if defined (SERVER_MODE)
+class ovfp_threshold_mgr g_ovfp_htreshold_mgr;
```
hyunikn: typo 'htreshold'

src/query/ovfp_threahold_monitor.cpp
```diff
@@ -0,0 +1,626 @@
...
+    }
+  m_head = m_free_head = NULL;
+}
+
+void
+ovfp_threshold::set_worker_idx (int idx, ovfp_monitor_lock *lock_mgr_p)
+{
+  assert (idx >= 0 && idx <= VACUUM_MAX_WORKER_COUNT);
+  m_lock_idx = idx;
+  if (m_lock_idx >= 0 && m_lock_idx < VACUUM_MAX_WORKER_COUNT)
```
hyunikn: Is it OK if this condition does not hold? Can we assert this condition at this line?

src/query/ovfp_threahold_monitor.hpp
ctshim: I agree, I will fix it. 
thank you

src/query/vacuum.c
ctshim: Oops, thank you.

ctshim: Oops, thank you.

src/query/ovfp_threahold_monitor.hpp
ctshim: Yes, that's right.
However, this is to maintain the format of separately classifying blocks that indicate error messages during build.
So, let's keep the status quo.
Thank you.

src/query/ovfp_threahold_monitor.hpp
ctshim: I've added the code based on comments.
However, due to the std::mutex originally included in this class, the phenomenon of concern is fundamentally blocked.

src/query/ovfp_threahold_monitor.cpp
ctshim: The assert check was done right above that.
Although the if statement is unnecessary because it is after the assert check,
Assert doesn't work in release, so it's added just in case.

If (idx == VACUUM_MAX_WORKER_COUNT), m_lock_ptr is not set.

hyunikn (approved):

hgryoo (approved):

beyondykk9 (approved):

PR #4535, opened 2023-07-28
[CBRD-24893] Gateway's DML bug fixed

airnet73:
http://jira.cubrid.org/browse/CBRD-24893

Purpose
* Add error messages for ODBC Driver's own errors
* Oracle does not support BIGINT, so BIGINT is changed to Numeric.
* change the binding method of double, float, decimal and numeric types in mariadb
* Modifying the number of rows in which insert, delete, update, and merge statements were performed

Implementation
N/A

Remarks
N/A


src/broker/cas_cgw.c
```diff
@@ -1227,6 +1252,17 @@ cgw_set_bindparam (T_CGW_HANDLE * handle, int bind_num, void *net_type, void *ne
       data_size = 0;
     }
 
+  // Oracle ODBC does not support the BIGINT type.
+  // So, change it to Numeric type.
+  if (curr_dbms_type == SUPPORTED_DBMS_ORACLE)
```
kisoo-han: what about remove inner-if by changing if condition

` if (curr_dbms_type == SUPPORTED_DBMS_ORACLE && type == CCI_U_TYPE_BIGINT)`

kisoo-han (approved):

src/broker/cas_cgw.c
airnet73: Thanks for the advice.

kisoo-han (approved):

PR #4536, opened 2023-07-31
[CBRD-24910] Fix the core dump caused by accessing a corrupted log record header

joohok:
http://jira.cubrid.org/browse/CBRD-24910

Purpose

`cdc_get_start_point_from_file ()` should return the LSA of log record which contains time value. 
But It was not returning the accurate address of the log. It was returning the end position of the log record instead. 

NOTE : 
When the end position of a log record coincides with the starting position of the next log record, it is possible to find the log record header of the next log record using the end position of the current log record. However, if the end position of the log record differs from the starting position of the next log record, it can lead to accessing a corrupted log record header

Crashed in `cdc_get_lsa_with_start_point`, `cdc_get_lsa_with_start_point ()` uses LSA returned from `cdc_get_start_point_from_file ()` and assumes that LSA is pointing to the log record header.

### DIFF
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -14471,6 +14471,7 @@ cdc_get_start_point_from_file (THREAD_ENTRY * thread_p, int arv_num, LOG_LSA * r
 
   LOG_LSA process_lsa = LSA_INITIALIZER;
   LOG_LSA forw_lsa = LSA_INITIALIZER;
+  LOG_LSA cur_log_lsa = LSA_INITIALIZER;
 
   LOG_RECORD_HEADER *log_rec_header;
   LOG_REC_DONETIME *donetime;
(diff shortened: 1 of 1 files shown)


src/transaction/log_manager.c
```diff
@@ -14577,10 +14580,9 @@ cdc_get_start_point_from_file (THREAD_ENTRY * thread_p, int arv_num, LOG_LSA * r
 	  LOG_READ_ADVANCE_WHEN_DOESNT_FIT (thread_p, sizeof (*donetime), &process_lsa, log_pgptr);
 	  donetime = (LOG_REC_DONETIME *) (log_pgptr->area + process_lsa.offset);
 
-	  LOG_READ_ADD_ALIGN (thread_p, sizeof (*donetime), &process_lsa, log_pgptr);
-	  LSA_COPY (ret_lsa, &process_lsa);
```
joohok, line 14581: `process_lsa` does not point to the starting address of the log record header.

src/transaction/log_manager.c
```diff
@@ -14682,8 +14685,6 @@ cdc_get_lsa_with_start_point (THREAD_ENTRY * thread_p, time_t * time, LOG_LSA *
 
 	      return NO_ERROR;
 	    }
-
-	  LOG_READ_ADD_ALIGN (thread_p, sizeof (*donetime), &process_lsa, log_page_p);
```
joohok, line 14686: unnecessary part

Rudeus (approved):

wrlawodms (approved):

hornetmj (approved):

PR #4537, opened 2023-07-31
[LETS-406] initialize log prior sender only when needed

cristiarg:
http://jira.cubrid.org/browse/LETS-406

Log prior sender is only initialized when needed - on ATS (which originates the log) and PS (which only relays the log from ATS to PTSs).

Implementation:
- the class usage is changed from a value member of `struct log_global` to a unique pointer
- both `log_global::m_prior_sender` and `log_global::m_prior_recver` have been made private and accessed via getters - `get_log_prior_sender` and `get_log_prior_receiver` (which existed before)
- `init_server_type`: initialize `log_global::m_prior_sender` for ATS and PS; will remain null for PTS as PTS does not dispatch log prior nodes any further
  - previously, on PTS, the function `log_Gl.get_log_prior_sender ().send_list` would be called
  - the log prior lists would be serialized and then thrown away as no log prior sinks are ever registered on a PTS
- `logpb_append_prior_lsa_list`: the dispatch of log prior linked lists is only done for Active Transaction Servers and Page Servers
- `xboot_shutdown_server`: it is needed to tear down connection infrastructure on ATS after the call to `log_final` (which flushes the log and, thus, needs to have the log prior sender alive); moved tearing down of connection infrastructure to peer servers on ATS after the call to `log_final` (1)
- adapted calls to use the getters
- adapted tests  following addition of an explicit destructor

With regard to (1), scalability tests have been run and execute successfully.

### DIFF
--- a/src/transaction/boot_sr.c
+++ b/src/transaction/boot_sr.c
@@ -3189,7 +3189,7 @@ xboot_shutdown_server (REFPTR (THREAD_ENTRY, thread_p), ER_FINAL_CODE is_er_fina
   vacuum_stop_master (thread_p);
 
 #if defined (SERVER_MODE)
-  if (get_server_type () == SERVER_TYPE_PAGE)
+  if (is_page_server ())
     {
       /*
        * The dependencies in PS shutdown: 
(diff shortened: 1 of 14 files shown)


hornetmj (approved):

src/transaction/boot_sr.c
```diff
@@ -3203,48 +3203,69 @@ xboot_shutdown_server (REFPTR (THREAD_ENTRY, thread_p), ER_FINAL_CODE is_er_fina
...
+#if defined (SA_MODE)
+  vacuum_sa_reflect_last_blockid (thread_p);
+#endif // SA_MODE
+  log_final (thread_p);
+
+#if defined(SERVER_MODE)
+  if (is_active_transaction_server ())
+    {
+      // for active transaction server, the log final needs to have the full connection
+      // infrastructure alive to fluhs the last of the log
```
wrlawodms, line 3263:
```suggestion
      // infrastructure alive to flush the last of the log
```

src/transaction/log_impl.h
```diff
@@ -714,29 +714,40 @@ struct log_global
...
+  std::atomic<bool> m_ps_lsa_up_to_date;
+  LOG_LSA m_ps_consensus_flushed_lsa;	// The quorum (number of the majority) of PS have done flushing log recrods until this.
+  // *INDENT-ON*
 
-  public:
-  log_global ();
-  ~log_global ();
+public:
+    log_global ();
+   ~log_global ();
```
wrlawodms:
```suggestion
  log_global ();
  ~log_global ();
```

src/transaction/log_global.c
```diff
@@ -141,15 +140,39 @@ log_global::wait_for_ps_flushed_lsa (const log_lsa &flush_lsa)
...
+{
+  assert (m_prior_sender == nullptr);
+  m_prior_sender = std::make_unique<cublog::prior_sender> ();
+}
+
+void
+log_global::finalize_log_prior_sender ()
+{
+  assert (m_prior_sender != nullptr);
+  assert (m_prior_sender->is_empty ());
```
wrlawodms: It says it's empty, but something may be added before the next line, doesn't it?

wrlawodms (commented):
So happy to avoid potential issues with this PR.

src/transaction/log_impl.h
cristiarg: I tried that. The `indent` too moves it back. Maybe because `indent` tool does not know dtor syntax as it only knows pure-C. I will except from indentation.

cristiarg:
@wrlawodms 
> So happy to avoid potential issues with this PR.

What do you mean?
In the meantime - while you were reviewing - I updated the description.

The rationale is that:
- I found out that this implementation is a "soft" prerequisite for http://jira.cubrid.org/browse/LETS-777 - in the sense that it makes that implementation easier
- and, in turn the implementation of LETS-777 is a prerequisite for the investigation in http://jira.cubrid.org/browse/LETS-773 where I want to add some delay to the replication in PTS vs the replication in PS and study the consequences in terms of performance
- and a lesser reason is that there is a PR open - https://github.com/CUBRID/cubrid/pull/3700 - since more than a year with the same thing. I actually copied partly from that one and took it further.

wrlawodms:
@cristiarg 
Sorry to confuse you. I mean that "I'm happy because this PR removes potential issues.". Without this PR, I might run into some troublesome issues such as incomplete log records while working on the multi-ps. It's just thank you message :).

src/transaction/log_global.c
cristiarg: Indeed. However it is a far-fetched scenario.

I adapted: moved the assert in the prior sender dtor; removed from here. And adapted unit tests.

joohok (approved):

src/transaction/log_prior_send.hpp
```diff
@@ -51,6 +53,8 @@ namespace cublog
       void add_sink (const sink_hook_t &fun);                     // add a hook for a new sink
       void remove_sink (const sink_hook_t &fun);                  // add a hook for a new sink
 
+      bool is_empty ();
```
wrlawodms: now it can be private.

wrlawodms (approved):

PR #4550, opened 2023-08-03
[LETS-780] Add details in "Quorum unsatisfied:.." msg for tests

wrlawodms:
http://jira.cubrid.org/browse/LETS-780

Add details in the log msg printe when the quorum is not met.
- In this case, the consensus_lsa is printed as NULL_LSA (-1|-1).

### DIFF
--- a/src/server/active_tran_server.cpp
+++ b/src/server/active_tran_server.cpp
@@ -77,12 +77,6 @@ active_tran_server::compute_consensus_lsa ()
 	}
     }
 
-  if (cur_node_cnt < quorum)
-    {
-      quorum_consenesus_er_log ("compute_consensus_lsa - Quorum unsatisfied: total node count = %d, curreunt node count = %d, quorum = %d\n",
-				total_node_cnt, cur_node_cnt, quorum);
-      return NULL_LSA;
-    }
   /*
    * Gather all PS'es saved_lsa and sort it in ascending order.
    * The <cur_node_count - quorum>'th element is the consensus LSA, upon which the majority (quorumn) of PS agrees.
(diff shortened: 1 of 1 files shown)


src/server/active_tran_server.cpp
```diff
@@ -93,17 +87,34 @@ active_tran_server::compute_consensus_lsa ()
...
       constexpr int BUF_SIZE = 1024;
       char msg_buf[BUF_SIZE];
       int n = 0;
+
+      n = snprintf (msg_buf, BUF_SIZE, "compute_consensus_lsa - ");
+
+      if (cur_node_cnt < quorum)
+	{
+	  n += snprintf (msg_buf + n, BUF_SIZE - n, "Quorum unsatisfied: ");
+	}
```
cristiarg: Nit:
```suggestion
      n += snprintf (msg_buf + n, BUF_SIZE - n, "Quorum %ssatisfied: ", ((cur_node_cnt < quorum) ? "un" : ""));
```

src/server/active_tran_server.cpp
```diff
@@ -93,17 +87,34 @@ active_tran_server::compute_consensus_lsa ()
...
+  LOG_LSA consensus_lsa;
+
+  if (cur_node_cnt < quorum)
+    {
+      consensus_lsa = NULL_LSA; // the quorum is not met.
+    }
+  else
+    {
+      consensus_lsa = collected_saved_lsa[cur_node_cnt - quorum];
+    }
```
cristiarg: Nit.
When the quorum is not met (a rather exception situation), the only extra useless operation is the `std::sort`.
So:
```suggestion
    LOG_LSA consensus_lsa;
    if (cur_node_cnt < quorum)
    {
      if (prm_get_bool_value (PRM_ID_ER_LOG_QUORUM_CONSENSUS))
        {
          // if the quorum is not met, only sort if we're about to log the details (see below); otherwise the sorting is useless   
          std::sort (collected_saved_lsa.begin (), collected_saved_lsa.end ());
        }
      consensus_lsa = NULL_LSA; // the quorum is not met.
    }
  else
    {
      // always do the sort if the quorum is met:
      std::sort (collected_saved_lsa.begin (), collected_saved_lsa.end ());
      consensus_lsa = collected_saved_lsa[cur_node_cnt - quorum];
    }
```
And delete the sort from above.

cristiarg (approved):

src/server/active_tran_server.cpp
wrlawodms: Good.

src/server/active_tran_server.cpp
wrlawodms: Thank you.

hornetmj (approved):

joohok (approved):

PR #4551, opened 2023-08-03
[CBRD-24842] Print the error message when the hostname lookup fails before the broker is initialized

eido5:
http://jira.cubrid.org/browse/CBRD-24842

**Description**

- in addition to #4497 ,
- fix below:

Print the error message when the hostname lookup fails before the broker is initialized.

**Remarks**

N/A

### DIFF
--- a/src/broker/broker_admin_pub.c
+++ b/src/broker/broker_admin_pub.c
@@ -74,6 +74,8 @@
 #include "chartype.h"
 #include "cubrid_getopt.h"
 #include "dbtype_def.h"
+#include "host_lookup.h"
+#include "system_parameter.h"
 
 #if defined(CAS_FOR_ORACLE) || defined(CAS_FOR_MYSQL)
 #define DB_EMPTY_SESSION        (0)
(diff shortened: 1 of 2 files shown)


src/broker/broker_admin_pub.c
```diff
@@ -324,9 +326,17 @@ admin_start_cmd (T_BROKER_INFO * br_info, int br_num, int master_shm_id, bool ac
     }
   chdir (envvar_bindir_file (path, BROKER_PATH_MAX, ""));
 
+  if (gethostname (hostname, sizeof (hostname)) < 0)
```
eido5, line 329: This is for the line 'hostname' of line 337.
If this line is not written, 'hostname' has no any chacters.

src/broker/broker_admin_pub.c
```diff
@@ -324,9 +326,17 @@ admin_start_cmd (T_BROKER_INFO * br_info, int br_num, int master_shm_id, bool ac
...
+    {
+      fprintf (stderr, "gethostname error\n");
```
kisoo-han: fprintf (stderr, "gethostname error: cannot get local hostname\n");

kisoo-han (approved):

src/broker/broker_admin_pub.c
eido5: I will apply this. Thank you!

PR #4552, opened 2023-08-03
[LETS-777] page server dispatches already packed log prior message without unpacking/repacking them

cristiarg:
http://jira.cubrid.org/browse/LETS-777

As of now, the Page Server processes by
- first unpacking them - as part of its own processing and, subsequent, replication 
- then, re-packing and dispatching the log prior lists again as part of flushing the log locally via the call path:
```
.. [log prior receiver worker thread] ..
prior_recver::loop_message_to_prior_info
  ==> m_prior_lsa_info.push_list
.. [log_wakeup_log_flush_daemon] ..
.. [log flush daemon thread] ..
  ==> log_flush_execute
  ==> logpb_flush_pages_direct
  ==> logpb_prior_lsa_append_all_list
  ==> logpb_append_prior_lsa_list
  ==> log_Gl.get_log_prior_sender ().send_list
```

### DIFF
--- a/msg/de_DE.utf8/utils.msg
+++ b/msg/de_DE.utf8/utils.msg
@@ -117,6 +117,8 @@ Vorhandene Admin-Utilitynamen:\n\
     paramdump\n\
     statdump\n\
     changemode\n\
+    copylogdb\n\
+    applylogdb\n\
     applyinfo\n\
     acldb\n\
     genlocale\n\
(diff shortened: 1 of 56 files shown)


cristiarg:
A feasible way to implement this optimization could not be achieved.

The few approaches that - almost - worked (and could have worked with some extra effort) are laborious and, in short, defeat the actual purpose of dispatching the log prior lists from the PS server to the registered PTSs faster and with less computation.

PR #4553, opened 2023-08-03

hornetmj:

See the jira issue.

### DIFF
--- a/src/storage/file_io.c
+++ b/src/storage/file_io.c
@@ -35,7 +35,6 @@
 #include <sys/stat.h>
 #include <assert.h>
 #include <signal.h>
-#include <algorithm>
 
 #if defined(WINDOWS)
 #include <io.h>
@@ -10394,6 +10393,17 @@ fileio_restore_volume (THREAD_ENTRY * thread_p, FILEIO_BACKUP_SESSION * session_
 		}
 
 	      fileio_dismount (thread_p, prev_vdes);
+
+	      if (unlinked_volinfo.count (volid))
+		{
+		  // The volume headers of both previous and current volumes are in the full or big incremental backup volumes.
+		  // Therefore, the link between the two volumes is naturally established during the restoration process.
+		  // So, there is no need to explicitly set it.
+		  unlinked_volinfo.erase (volid);
+
+		  _er_log_debug (ARG_FILE_LINE, "RESTOREDB: [UNSAVE UNLINK] [lv%d] volid=%d, vol=%s, prev_vol=%s\n",
+				 session_p->dbfile.level, volid, to_vol_label_p, prev_vol_label_p);
+		}
 	    }
 
 	  if (incremental_includes_volume_header == true)
@@ -10408,23 +10418,9 @@ fileio_restore_volume (THREAD_ENTRY * thread_p, FILEIO_BACKUP_SESSION * session_
 		      unlinked_volinfo[volid] =
 			std::make_pair (std::string (to_vol_label_p), std::string (prev_vol_label_p));
 
-
 		      _er_log_debug (ARG_FILE_LINE, "RESTOREDB: [SAVE UNLINK] [lv%d] volid=%d, vol=%s, prev_vol=%s\n",
 				     session_p->dbfile.level, volid, to_vol_label_p, prev_vol_label_p);
 		    }
-		}
-	    }
-	  else			/* full level */
-	    {
-	      if (unlinked_volinfo.count (volid))
-		{
-		  // The volume headers of both previous and current volumes are in the full level backup volume.
-		  // Therefore, the link between the two volumes is naturally established during the restoration process.
-		  // So, there is no need to explicitly set it.
-		  unlinked_volinfo.erase (volid);
-
-		  _er_log_debug (ARG_FILE_LINE, "RESTOREDB: [UNSAVE UNLINK] [lv%d] volid=%d, vol=%s, prev_vol=%s\n",
-				 session_p->dbfile.level, volid, to_vol_label_p, prev_vol_label_p);
 		}
 	    }
 
--- a/src/storage/file_io.c
+++ b/src/storage/file_io.c
@@ -35,6 +35,7 @@
 #include <sys/stat.h>
 #include <assert.h>
 #include <signal.h>
+#include <algorithm>
 
 #if defined(WINDOWS)
 #include <io.h>
@@ -10171,6 +10172,9 @@ exit_on_error:
 }
 
 #if !defined(CS_MODE)
+
+using namespace std;
+
 /*
  * fileio_restore_volume () - Restore a volume/file of given database
  *   return:
@@ -10185,7 +10189,8 @@ exit_on_error:
 int
 fileio_restore_volume (THREAD_ENTRY * thread_p, FILEIO_BACKUP_SESSION * session_p, char *to_vol_label_p,
 		       char *verbose_to_vol_label_p, char *prev_vol_label_p, FILEIO_RESTORE_PAGE_BITMAP * page_bitmap,
-		       bool is_remember_pages)
+		       bool is_remember_pages, bool & is_prev_vol_header_restored, vector < tuple < int, string,
+		       string >> &unlinked_vol_info)
 {
   int next_page_id = 0;
   INT64 total_nbytes = 0;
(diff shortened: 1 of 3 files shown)


src/storage/file_io.c
```diff
@@ -10171,6 +10172,9 @@ fileio_decompress_restore_volume (THREAD_ENTRY * thread_p, FILEIO_BACKUP_SESSION
 }
 
 #if !defined(CS_MODE)
+
+using namespace std;
+
```
wrlawodms:
```suggestion
```

src/storage/file_io.c
```diff
@@ -10393,6 +10398,63 @@ fileio_restore_volume (THREAD_ENTRY * thread_p, FILEIO_BACKUP_SESSION * session_
...
+	    int unlinked_volid = get<0> (unlinked_vol);
+
+	    if (unlinked_volid == volid)
+	      {
+		return true;
+	      }
+	    else
+	      {
+		return false;
+	      }
```
wrlawodms:
```suggestion
	    return unlinked_volid == volid;
```

src/storage/file_io.c
```diff
@@ -10393,6 +10398,63 @@ fileio_restore_volume (THREAD_ENTRY * thread_p, FILEIO_BACKUP_SESSION * session_
 
 	      fileio_dismount (thread_p, prev_vdes);
 	    }
+
+	  // *INDENT-OFF*
+	  auto find_unlinked_vol = [&volid] (const tuple <int, string, string> &unlinked_vol)
+	  {
+	    int unlinked_volid = get<0> (unlinked_vol);
```
wrlawodms:
```suggestion
	    const int unlinked_volid = get<0> (unlinked_vol);
```

src/storage/file_io.c
```diff
@@ -10393,6 +10398,63 @@ fileio_restore_volume (THREAD_ENTRY * thread_p, FILEIO_BACKUP_SESSION * session_
...
+	  // *INDENT-ON*
+
+	  if (incremental_includes_volume_header == true)
+	    {
+	      if (volid != LOG_DBFIRST_VOLID && is_prev_vol_header_restored == false)
+		{
+		  _er_log_debug (ARG_FILE_LINE, "RESTOREDB: [FOUND UNLINK] [lv%d] volid=%d, vol=%s, prev_vol=%s\n",
+				 session_p->dbfile.level, volid, to_vol_label_p, prev_vol_label_p);
+
+		  auto it_found = find_if (unlinked_volinfo.cbegin (), unlinked_volinfo.cend (), find_unlinked_vol);
```
wrlawodms: We usually use standard libraries with `std::` prefix to distinguish it instead of using `using namespace std;` above.
```suggestion
		  auto it_found = std::find_if (unlinked_volinfo.cbegin (), unlinked_volinfo.cend (), find_unlinked_vol);
```

src/transaction/log_page_buffer.c
```diff
@@ -8413,6 +8415,11 @@ logpb_restore (THREAD_ENTRY * thread_p, const char *db_fullname, const char *log
   REL_COMPATIBILITY compat;
   int dummy;
 
+  bool is_prev_volheader_restored = false;
+  // *INDENT-OFF*
+  FILEIO_UNLINKED_VOLINFO unlinked_volinfo;
+  // *INDENT-ON*
```
wrlawodms: it may not be needed?
```suggestion
  FILEIO_UNLINKED_VOLINFO unlinked_volinfo;
```

src/transaction/log_page_buffer.c
```diff
@@ -8714,6 +8721,8 @@ logpb_restore (THREAD_ENTRY * thread_p, const char *db_fullname, const char *log
 		case LOG_DBLOG_ARCHIVE_VOLID:
 		case LOG_DBTDE_KEYS_VOLID:
 
+		  is_prev_volheader_restored = false;
```
joohok, line 8722: `is_prev_volheader_restored` is always false. 
I think `is_prev_volheader_restored` need to be set true at the part where the previous volume header is restored.

src/storage/file_io.h
```diff
@@ -454,6 +454,10 @@ struct flush_stats
   unsigned int num_tokens;
 };
 
+// *INDENT-OFF*
+using FILEIO_UNLINKED_VOLINFO = std::vector <std::tuple <int, std::string, std::string>>;
```
wrlawodms:
```suggestion
using FILEIO_UNLINKED_VOLINFO_MAP = std::map <int, std::pair< std::string, std::string>>;
```
How about using the dictionary container, `std::map` or `std::unordered_map`, instead? 
in that 
1. It uses a unique key, volid. 
2. The way it is used seems what the dictionary is for so that the code gets simpler and clearer.

Please check the diff on my version with map: https://github.com/wrlawodms/cubrid/commit/52158b9589b77a770082cfb7e628bd7ca3576c0d
13 insertions(+), 34 deletions(-).

src/transaction/log_page_buffer.c
```diff
@@ -8854,6 +8863,39 @@ logpb_restore (THREAD_ENTRY * thread_p, const char *db_fullname, const char *log
       try_level = (FILEIO_BACKUP_LEVEL) (try_level - 1);
     }
 
+  // Incremental backup volumes often do not include volume header pages.
+  // Accordingly, it is necessary to set unlinked volumes after all volume header pages are restored.
+  for (int i = 0; i < unlinked_volinfo.size (); i++)
```
wrlawodms: It's recommended for a STL container to use for-range instead of accessing with an index. This line may generate a warning.
`for (const auto& volinfo :  unlinked_volinfo)` // volinfo := tuple<int,string,string>
or if you accept to use map I recommended above
`for (const auto &[volid, vol_names]:unlinked_volinfo)` //vol_names = pair<string, string>

src/storage/file_io.c
```diff
@@ -10393,6 +10398,63 @@ fileio_restore_volume (THREAD_ENTRY * thread_p, FILEIO_BACKUP_SESSION * session_
...
+				 session_p->dbfile.level, volid, to_vol_label_p, prev_vol_label_p);
+		}
+	    }
+
+	  is_prev_vol_header_restored = true;
+	}
+      else
+	{
+	  is_prev_vol_header_restored = false;
 	}
```
joohok, line 10432:
```suggestion
	}
```

These modified value is not used anywhere.

src/transaction/log_page_buffer.c
hornetmj: See the fileio_restore_volume(). The is_prev_volheader_restored is set in it when matching some condition.

src/transaction/log_page_buffer.c
```diff
@@ -8854,6 +8861,38 @@ logpb_restore (THREAD_ENTRY * thread_p, const char *db_fullname, const char *log
       try_level = (FILEIO_BACKUP_LEVEL) (try_level - 1);
     }
 
+  // Incremental backup volumes often do not include volume header pages.
+  // Accordingly, it is necessary to set unlinked volumes after all volume header pages are restored.
+  // *INDENT-OFF*
+  for (const auto &[volid, volnames]:unlinked_volinfo)
```
wrlawodms:
```suggestion
  for (const auto &[volid, volnames] : unlinked_volinfo)
```

src/storage/file_io.c
```diff
@@ -35,6 +35,7 @@
 #include <sys/stat.h>
 #include <assert.h>
 #include <signal.h>
+#include <algorithm>
```
wrlawodms:
```suggestion
```

src/transaction/log_page_buffer.c
joohok: Ah, this is in the while loop. I understand.

joohok (approved):

wrlawodms (approved):

src/storage/file_io.c
```diff
@@ -10392,7 +10393,42 @@ fileio_restore_volume (THREAD_ENTRY * thread_p, FILEIO_BACKUP_SESSION * session_
 		}
 
 	      fileio_dismount (thread_p, prev_vdes);
+
+	      if (unlinked_volinfo.count (volid))
```
hornetmj, line 10397: Leave the if statement for leaving the debugging log only in this condition.

PR #4559, opened 2023-08-08
[LETS-746] Initialize unsent_lsa conditionally to fix cores on PTS

wrlawodms:
http://jira.cubrid.org/browse/LETS-746

The cores occur because the prior sender is initialized only on ATS and PS.
- \+ Adapt a unit test

### DIFF
--- a/src/transaction/log_append.cpp
+++ b/src/transaction/log_append.cpp
@@ -185,7 +185,10 @@ LOG_RESET_APPEND_LSA (const LOG_LSA *lsa)
   log_Gl.hdr.append_lsa = *lsa;
   log_Gl.prior_info.prior_lsa = *lsa;
 #if defined (SERVER_MODE)
-  log_Gl.get_log_prior_sender ().reset_unsent_lsa (*lsa);
+  if (is_active_transaction_server () || is_page_server ())
+    {
+      log_Gl.get_log_prior_sender ().reset_unsent_lsa (*lsa);
+    }
 #endif /* SERVER_MODE */
   log_Gl.append.set_nxio_lsa (*lsa);
 }
--- a/unit_tests/log/test_main_prior_list_serialize.cpp
+++ b/unit_tests/log/test_main_prior_list_serialize.cpp
@@ -341,6 +341,12 @@ is_active_transaction_server ()
   return true;
 }
 
+bool
+is_page_server ()
+{
+  return false;
+}
+
 PGLENGTH
 db_log_page_size ()
 {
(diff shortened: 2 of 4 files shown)


hornetmj (approved):

src/transaction/log_append.cpp
```diff
@@ -185,7 +185,10 @@ LOG_RESET_APPEND_LSA (const LOG_LSA *lsa)
   log_Gl.hdr.append_lsa = *lsa;
   log_Gl.prior_info.prior_lsa = *lsa;
 #if defined (SERVER_MODE)
-  log_Gl.get_log_prior_sender ().reset_unsent_lsa (*lsa);
+  if (is_active_transaction_server () || is_page_server ())
```
joohok, line 188:
```suggestion
  if (!is_passive_transaction_server ())
```
Nit. I think it can be shortened

unit_tests/log/test_main_prior_list_serialize.cpp
```diff
@@ -341,6 +341,12 @@ is_active_transaction_server ()
   return true;
 }
 
+bool
+is_page_server ()
+{
+  return false;
+}
+
```
joohok, line 349:
```suggestion

bool
is_passive_transaction_server ()
{
  return false;
}
```
if above change is accepted

joohok (approved):

src/transaction/log_append.cpp
wrlawodms: I have considered this and `if log prior sender is initialized or not` as well, but I have decided to use this. What you suggested has the assumption that `NOT PTS -> ATS or PS`, which is actually true probably always, but I think it's better to avoid any assumption and be explicit if it doesn't cost a lot.

joohok: OK, I agree. Clearly stating a condition without inferring would likely make it easier to understand.

cristiarg: This change here mandates a corresponding assert in `prior_sender::reset_unsent_lsa`:
```
  assert (is_active_transaction_server () || is_page_server ());
```

cristiarg (approved):

src/transaction/log_append.cpp
cristiarg, line 188: This change here mandates a corresponding assert in `prior_sender::reset_unsent_lsa`:
```
  assert (is_active_transaction_server () || is_page_server ());
```

wrlawodms: It's been done and a unit test has been adapted.

PR #4561, opened 2023-08-08
[CBRD-24899] rev: check push predicate

beyondykk9:

We should also consider the arithmetic operator as like '+', '-', '*', '/', '%' to check predicate push terms.

### DIFF
--- a/src/parser/parser_support.c
+++ b/src/parser/parser_support.c
@@ -8462,6 +8462,30 @@ pt_convert_to_logical_expr (PARSER_CONTEXT * parser, PT_NODE * node, bool use_pa
 }
 
 /*
+ * pt_is_operator_arith() - returns TRUE if the operator has an arithmetic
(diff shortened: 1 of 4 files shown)


ctshim (approved):

src/parser/parser_support.c
```diff
@@ -8461,6 +8461,30 @@ pt_convert_to_logical_expr (PARSER_CONTEXT * parser, PT_NODE * node, bool use_pa
...
+ *
+ *   return: boolean
+ *   op(in): the operator
+ */
+bool
+pt_is_operator_arith (PT_OP_TYPE op)
+{
+  switch (op)
+    {
+    case PT_PLUS:
```
hyunikn, line 8477: As we discussed, allowing PLUS can cause a problem.

hgryoo (approved):

hyunikn (approved):

PR #4563, opened 2023-08-08
[CBRD-24929] Add identifier comparator class and refactor comparision of catalog names

hgryoo:
http://jira.cubrid.org/browse/CBRD-24929

- Add a class to provide a unified routine to compare identifiers
  - string_utility.hpp: For C++ STL's string comparison, lowercase_hash and lowercase_compare structs are added, string_set_ci_lower is defined
  - identifier_store.hpp: the constructor gets a set of identifier lists and provides is_exists () API.
- Refactor comparison routine related to system catalog names
  - schema_system_catalog.[cpp|hpp]: Use the new API of identifier_store for system catalogs
  - remove legacy duplicated routine (util_common.c, utility.h, load_server_loader.cpp...)


ctshim (approved):

src/object/schema_system_catalog.cpp
```diff
@@ -0,0 +1,112 @@
...
+    CTV_DB_SERVER_NAME,		// "db_server"
+    CTV_SYNONYM_NAME			// "db_synonym"
+  };
+
+  static const cubbase::identifier_store sm_catalog_names (sm_system_catalog_names, false);
+}
+
+bool sm_check_system_class_by_name (const std::string &name)
+{
+  return cubschema::sm_catalog_names.is_exists (name);
```
mhoh3963: The system class name can't be contained dot(.). 
So, before execute sm_catalog_names.is_exists(),  how about checking if there is "." and returning false if there is ?

hgryoo: Thank you for the review. In CUBRID's identifier writing rule, "." is not allowed in both cases whether a name is enclosed (by quotes) or not. 

- I've added a new static function, `check_identifier_is_valid`, which checks name is a valid identifier according to the name enclosed. (40c3c1e)
- In `check_identifier_condition()` in identifier_store, changed to use  `check_identifier_is_valid()` above.
- I don't think the function check_identifier_is_valid() is in right place. We need to implement a unified identifier writing rule checker needs somewhere. Before then, I'm going to leave it here.

mhoh3963 (approved):

src/object/schema_system_catalog.cpp
mhoh3963: OK, good !!

mhoh3963:
@hgryoo 
please check fails of test_sql, test_medium.

hgryoo:
> Member

System table names, such as _db_class, _db_attribute, do not follow the identifier's writing rules. Because of the checking identifier's validation, createdb failed in CI because of added logic at 40c3c1e.

hyunikn (approved):

beyondykk9 (approved):

PR #4566, opened 2023-08-08
[CBRD-24892] The restoredb with level 1 fails when some of the permanent volume header is not included (#4553)

hornetmj:

backport: #4553

### DIFF
--- a/src/transaction/log_page_buffer.c
+++ b/src/transaction/log_page_buffer.c
@@ -8350,6 +8350,8 @@ logpb_check_stop_at_time (FILEIO_BACKUP_SESSION * session, time_t stop_at, time_
   return NO_ERROR;
 }
 
+using namespace std;
+
 /*
  * logpb_restore - Restore volume from its backup
  *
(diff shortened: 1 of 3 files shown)


src/transaction/log_page_buffer.c
```diff
@@ -8350,6 +8350,8 @@ logpb_check_stop_at_time (FILEIO_BACKUP_SESSION * session, time_t stop_at, time_
   return NO_ERROR;
 }
 
+using namespace std;
+
```
wrlawodms, line 8354:
```suggestion
```
Is it needed?

wrlawodms (approved):
Others seem good to me.

src/transaction/log_page_buffer.c
hornetmj: No, it is not removed in the develop PR. It would be nice handle in the other PR cause the appveyor consume too much time. I will handle it soon.

wrlawodms: I don't think I fully understand what you mean.
> No, it is not removed in the develop PR

I see that it's also not removed in the develop branch. we missed it, unfortunately, right? I think we can remove this while doing backports for other branches, and deal with the develop branch in the following PR.

 >  It would be nice handle in the other PR cause the appveyor consume too much time. I will handle it soon.

I don't understand this.

hornetmj: To merging the PR, appveyor(windows build) need to pass. But, it comsume too much time because of waiting other PRs request.

PR #4569, opened 2023-08-08
[CBRD-24892] The restoredb with level 1 fails when some of the permanent volume header is not included (#4553)

hornetmj:

backport: #4553

### DIFF
--- a/src/transaction/log_page_buffer.c
+++ b/src/transaction/log_page_buffer.c
@@ -8449,14 +8449,15 @@ logpb_restore (THREAD_ENTRY * thread_p, const char *db_fullname, const char *log
   // Incremental backup volumes often do not include volume header pages.
   // Accordingly, it is necessary to set unlinked volumes after all volume header pages are restored.
   // *INDENT-OFF*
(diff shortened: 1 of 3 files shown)


src/transaction/log_page_buffer.c
```diff
@@ -8439,6 +8444,39 @@ logpb_restore (THREAD_ENTRY * thread_p, const char *db_fullname, const char *log
       try_level = (FILEIO_BACKUP_LEVEL) (try_level - 1);
     }
 
+  // Incremental backup volumes often do not include volume header pages.
+  // Accordingly, it is necessary to set unlinked volumes after all volume header pages are restored.
+  // *INDENT-OFF*
+  for (const auto &unlinked_itr : unlinked_volinfo)
```
wrlawodms, line 8450: Nit. it's not an iterator.
```suggestion
  for (const auto &volinfo : unlinked_volinfo)
```

wrlawodms (approved):

src/transaction/log_page_buffer.c
hornetmj: Thank you. But, I won't edit it to merge right away.

PR #4571, opened 2023-08-09
[CBRD-24899] rev2: check push predicate

beyondykk9:

The wrapped cast is pushed into dblink's predicate.


src/parser/type_checking.c
```diff
@@ -18929,7 +18929,16 @@ pt_check_dblink_related_expr (PARSER_CONTEXT * parser, PT_NODE * p, void *arg, i
 {
   bool *dblink_related = (bool *) arg;
 
-  if (p->node_type == PT_NAME && p->info.name.spec_id)
+  if (p->node_type == PT_EXPR)
```
hyunikn, line 18951: By adding this new condition, this function should be renamed to, for example, pt_check_to_skip_constant_folding.
The variable dblink_related should also be renamed to, for example, skip_constant_folding.

hyunikn (approved):

hgryoo (approved):

src/parser/type_checking.c
beyondykk9: I modified the name to pt_do_not_fold_dblink_related_cast, generally the "do_not_fold" prefix is used to mean "skip the folding constant term".

ctshim (approved):

PR #4580, opened 2023-08-11
[CBRD-24899] bug-fix: check push predicate

beyondykk9:


beyondykk9:
I will handle this issue at CBRD-24935

PR #4583, opened 2023-08-11
[CBRD-24404] Fix the hostname to be stored instead of alias

eido5:
http://jira.cubrid.org/browse/CBRD-24404

**Description**

in addition to #4380 and #3740 ,
$CUBRID/conf/cubrid_hosts.conf stores alias instead hostname, so we store hostname instead of alias.

**Remarks**

Overwritting PR.

### DIFF
--- a/src/connection/host_lookup.c
+++ b/src/connection/host_lookup.c
@@ -304,6 +304,11 @@ load_hosts_file ()
 	    }
 	  else
 	    {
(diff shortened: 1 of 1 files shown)


airnet73 (approved):

kisoo-han (approved):
It is also the responsibility of reviewers.

PR #4585, opened 2023-08-11
[CBRD-24867] regression: bug-fix, quoted-name for dblink

beyondykk9:

The quoted name should be compared with excluding quote when the column is collected from remote server by cci.


beyondykk9:
I will handle this issue at CBRD-24934.

PR #4587, opened 2023-08-15
[CBRD-24934] [Regression] extending dblink to DML: double quoted identifier

beyondykk9:
http://jira.cubrid.org/browse/CBRD-24934

To use reserved words as table or column names in other DBMS such as Oracle using extended table names in CUBRID, double quotes should be wrapped inside the braket like ["FROM"]. However, in the remote server, this identifier is being returned without double quotes. It can receive column information by sending a SELECT ["ALTER"], ["COUNT"], ["ADD"] FROM ["From"] query through cci_prepare. At this time, the column information becomes ALTER, COUNT, ADD without double quotes.

### DIFF
--- a/src/base/intl_support.c
+++ b/src/base/intl_support.c
@@ -2730,19 +2730,21 @@ intl_strcasecmp_utf8_one_cp (const ALPHABET_DATA * alphabet, unsigned char *str1
  * intl_identifier_casecmp_for_dblinke() - compares two identifiers strings
  *			       case insensitive excluding double quote for dblink
  *
- *   return: 0 if strings are equal, -1 if str1 < str2 , 1 if str1 > str2
- *   str1(in):
- *   str2(in):
+ *   return: 0 if dblink_col_name equals to remote_col_name
+ *   dblink_col_name(in):
+ *   remote_col_name(in):
  *
  * NOTE: this routine is the same as intl_identifier_casecmp
- *       the first argument str1 may start with double quote but the second one never
+ *       the first argument dblink_col_name may start with double quote
+ *       but the remote_col_name never
  */
 int
-intl_identifier_casecmp_for_dblink (const char *str1, const char *str2)
+intl_identifier_casecmp_for_dblink (const char *dblink_col_name, const char *remote_col_name)
 {
   int str1_size;
   int str2_size;
-  int offset;
+  char *str1 = (char *) dblink_col_name;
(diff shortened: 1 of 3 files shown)


hgryoo (approved):

src/base/intl_support.c
```diff
@@ -2726,6 +2726,36 @@ intl_strcasecmp_utf8_one_cp (const ALPHABET_DATA * alphabet, unsigned char *str1
...
+
+  assert (str1 != NULL);
+  assert (str2 != NULL);
+
+  str1_size = (*str1 == '\"') ? strlen (str1) - 2 : strlen (str1);
+  str2_size = strlen (str2);
+
+  offset = (*str1 == '\"') ? 1 : 0;
+
+  return intl_identifier_casecmp_w_size (lang_id (), (unsigned char *) str1 + offset, (unsigned char *) str2, str1_size,
```
ctshim: Is it correct to compare with casecmp?


```
drop table tbl;
create table tbl( "Id" int, "id" int, id int);
insert into tbl ("Id", "id", id) values (1,2,3);
select * from tbl;
```

If you execute the example query above in Oracle, the select result item will look like the following.

```
Id	id	ID
1	2	3
```

If it was enclosed in quotation marks, it seems that the comparison should be case sensitive.

beyondykk9: I know, but it should be a constraint as a specification of 11.3.

src/base/intl_support.c
```diff
@@ -2726,6 +2726,36 @@ intl_strcasecmp_utf8_one_cp (const ALPHABET_DATA * alphabet, unsigned char *str1
...
+ *
+ *   return: 0 if strings are equal, -1 if str1 < str2 , 1 if str1 > str2
+ *   str1(in):
+ *   str2(in):
+ *
+ * NOTE: this routine is the same as intl_identifier_casecmp
+ *       the first argument str1 may start with double quote but the second one never
+ */
+int
+intl_identifier_casecmp_for_dblink (const char *str1, const char *str2)
```
ctshim: Although explained in the NOTE, it is better to make a difference in the name of the argument argument.

What about dblink_col_name, remote_col_name?

beyondykk9: Ok. good point.

hyunikn (approved):

ctshim (approved):

PR #4588, opened 2023-08-16
[CBRD-24935] extending dblink to DML: function is pushed

beyondykk9:
http://jira.cubrid.org/browse/CBRD-24935

In dblink, predicate push should exclude functions. In the case of built-in functions such as "cast" and "str_to_date", if the argument consists only of constants, the function disappears by constant folding, and the function expression is replaced with a text field in the PT_VALUE node instead. do. While processing this in the predicate push logic, the corresponding function expression is being pushed when the PT_VALUE node is pushed.

### DIFF
--- a/src/parser/parse_tree_cl.c
+++ b/src/parser/parse_tree_cl.c
@@ -15940,9 +15940,9 @@ pt_print_value (PARSER_CONTEXT * parser, PT_NODE * p)
     case PT_TYPE_INTEGER:
     case PT_TYPE_BIGINT:
     case PT_TYPE_SMALLINT:
-      if (!(parser->custom_print & (PT_PRINT_SUPPRESS_FOR_DBLINK | PT_SUPPRESS_BIGINT_CAST)))
+      if (p->info.value.text != NULL)
 	{
-	  if ((p->info.value.text != NULL))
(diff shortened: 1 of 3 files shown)


src/parser/parse_tree_cl.c
```diff
@@ -15993,10 +15994,13 @@ pt_print_value (PARSER_CONTEXT * parser, PT_NODE * p)
     case PT_TYPE_DATETIME:
     case PT_TYPE_DATETIMETZ:
     case PT_TYPE_DATETIMELTZ:
-      if (p->info.value.text)
+      if (!(parser->custom_print & PT_PRINT_SUPPRESS_FOR_DBLINK))
 	{
-	  q = pt_append_nulstring (parser, q, p->info.value.text);
-	  break;
+	  if (p->info.value.text)
```
ctshim, line 15999: It's a good idea to check the "p->info.value.text" condition first.
The same goes for the code below.

beyondykk9: I think the print_custom comparison should come first. Because value.text compares with another comparison.

ctshim (approved):

hyunikn (approved):

hgryoo (approved):

PR #4590, opened 2023-08-16
[CBRD-24884] Implementing functions for network communication

Rudeus:
http://jira.cubrid.org/browse/CBRD-24884

Implementation of client-side network communication functions:

1. mmon_get_server/module/tran_info, mmon_get_module_info_summary
 - pack request parameter
 - send request buffer which include the parameter of server-side functions
 - unpack module/transaction information from buffer to MMON_MODULE/TRAN_INFO

2. Add memmon utility function code for verification


src/base/memory_monitor_sr.cpp
```diff
@@ -674,7 +674,7 @@ void mmon_aggregate_server_info (MMON_SERVER_INFO &info)
 void mmon_aggregate_module_info (int module_index, std::vector<MMON_MODULE_INFO> &info)
 {
   assert (mmon_Gl != nullptr);
-  assert (module_index < (int)MMON_MODULE_LAST && module_index >= 0);
+  assert (module_index <= (int)MMON_MODULE_LAST && module_index >= 0);
```
wrlawodms: Why is `module_index == MMON_MODULE_LSA` allowed? Could you describe the case?

src/communication/network_interface_cl.c
```diff
@@ -8124,6 +8124,285 @@ logtb_dump_trantable (FILE * outfp)
...
+      memcpy (server_info.name, temp_str, strlen (temp_str) + 1);
+      ptr = or_unpack_int64 (ptr, (int64_t *) & (server_info.total_mem_usage));
+    }
+  free_and_init (buffer);
+
+  return error;
+#else /* CS_MODE */
+  er_log_debug (ARG_FILE_LINE, "mmon_get_server_info: THIS IS ONLY a C/S function");
+  return ER_FAILED;
+#endif /* !CS_MODE */
```
wrlawodms: There seems a way to deal with the case in other utilities. Have you considered this way?
```
#else /* CS_MODE */
  /* Cannot run in standalone mode */
  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_NOT_IN_STANDALONE, 1, "checkpoint");

  return ER_NOT_IN_STANDALONE;
#endif /* !CS_MODE */
```

src/communication/network_interface_cl.c
```diff
@@ -8124,6 +8124,285 @@ logtb_dump_trantable (FILE * outfp)
...
+  if (req_error)
+    {
+      assert (er_errid () != NO_ERROR);
+      error = er_errid ();
+    }
+  else
+    {
+      ptr = reply;
+      ptr = or_unpack_int (ptr, &ival);
+      ptr = or_unpack_int (ptr, &ival);
```
wrlawodms: What does `ival` mean? Can we just ignore the previous packed values before error, which are overwritten? If it's the case, why they should be packed and sent?

src/communication/network_interface_cl.c
```diff
@@ -8124,6 +8124,285 @@ logtb_dump_trantable (FILE * outfp)
...
+	{
+	  module_count = 1;
+	}
+
+      module_info.resize (module_count);
+
+      ptr = buffer;
+
+      // *INDENT-OFF*
+      for (auto & m_info:module_info)
```
```suggestion
      for (auto &m_info : module_info)
```

src/communication/network_interface_sr.c
```diff
@@ -7564,6 +7579,10 @@ smmon_get_tran_info (THREAD_ENTRY * thread_p, unsigned int rid, char *request, i
   // Size of transaction id (OR_INT_SIZE)
   // and current memory usage (OR_INT64_SIZE)
   size += ((OR_INT_SIZE + OR_INT64_SIZE) * info.num_tran);
+  if (size != 0)
```
wrlawodms:
```suggestion
  if (size == 0)
```
Isn't it for the case where there is no trx?

src/executables/util_cs.c
```diff
@@ -4649,7 +4651,69 @@ memmon (UTIL_FUNCTION_ARG * arg)
...
+	}
+
+      // TODO: mmon_print_module_info (module_info);
+    }
+
+  if (!transaction && !module)
+    {
+      /* default case */
+      // *INDENT-OFF*
+      error_code = mmon_get_module_info_summary (std::min ((int) MMON_MODULE_LAST, 5), module_info);
```
```suggestion
      error_code = mmon_get_module_info_summary (std::min ((int) MMON_MODULE_LAST, DEFAULT_MODULE_INFO_CNT), module_info);
```
Nit. It's better to define a constant for the kind of thing to make it self-explanatory. `constexpr int DEFAULT_MODULE_INFO_CNT=5`, this const expression even doesn't cost at all.

src/executables/util_cs.c
```diff
@@ -4649,7 +4651,69 @@ memmon (UTIL_FUNCTION_ARG * arg)
...
+      // *INDENT-ON*
+      if (error_code != NO_ERROR)
+	{
+	  goto error_exit;
+	}
+
+      // TODO: mmon_print_module_info_summary (module_info);
+
+      transaction = true;
```
```suggestion
      does_print_transaction = true;
```
or whatever to distinguish a flag variable.

src/executables/util_cs.c
```diff
@@ -4649,7 +4651,69 @@ memmon (UTIL_FUNCTION_ARG * arg)
...
+  /* execute phase */
+  error_code = mmon_get_server_info (server_info);
+  if (error_code != NO_ERROR)
+    {
+      goto error_exit;
+    }
+
+  // TODO: mmon_print_server_info (server_info);
+
+  if (module_index >= 0)
```
wrlawodms: Can we pare it down like this? at your choice.
```suggestion
  if (module_index >= 0)
  {
  }
  else
  {
    if (!transaction)
    {
      // default case, set the tran count
    }
    else if (show_all)
    {
     //show_all case, set the tran count 
    }
    elese
    {
     assert (false); // a parameter combination not allowed for safety 
    }
    mmon_get_module_info_summary()
  }
  
```

src/base/memory_monitor_sr.cpp
Rudeus: Because client send module index start to 1(except all) but MMON_MODULE_ID start to 0. This occurs due to the design intention of avoiding the creation of a dummy. To align this discrepancy in indexes, 1 is subtracted from the module_index received from the client in memory_monitor::aggregate_module_info().

Rudeus: Because client send `module_index` start to 1(except all) but `MMON_MODULE_ID` start to 0. This occurs due to the design intention of avoiding the creation of a dummy. To align this discrepancy in indexes, 1 is subtracted from the `module_index` received from the client in `memory_monitor::aggregate_module_info()`.

src/communication/network_interface_cl.c
Rudeus: I agree. I change it.

src/communication/network_interface_cl.c
Rudeus: The first piece of information that is unpacked is the size information of the buffer, which is being transmitted through the css function when sending the buffer. Although it was considered unnecessary, I treated it as a legacy. However, upon further investigation, it seemed that it is not consistently used in all functions, so I decided to remove it.

+) However, upon removing it and testing, it was observed that the buffer containing the data was not being transmitted correctly. It appears that there is an internal procedure within the css function to check the size of the reply buffer. Therefore, removal is not feasible.

src/communication/network_interface_cl.c
Rudeus: Done.

src/communication/network_interface_sr.c
Rudeus: I miss it. Done.

src/executables/util_cs.c
Rudeus: Thanks. I change it.

src/executables/util_cs.c
Rudeus: I agree with removing code duplication. I change it.

src/communication/network_interface_cl.c
```diff
@@ -8124,6 +8124,285 @@ logtb_dump_trantable (FILE * outfp)
...
+    }
+
+  if (error == NO_ERROR)
+    {
+      module_info.resize (module_count);
+
+      ptr = buffer;
+
+      // *INDENT-OFF*
+      for (auto &m_info : module_info)
```
wrlawodms, line 8323:
```suggestion
      for (const auto &m_info : module_info)
```

src/communication/network_interface_cl.c
```diff
@@ -8124,6 +8124,285 @@ logtb_dump_trantable (FILE * outfp)
...
+
+  if (error == NO_ERROR)
+    {
+      ptr = buffer;
+      ptr = or_unpack_int (buffer, (int32_t *) & (tran_info.num_tran));
+
+      tran_info.tran_stat.resize (tran_info.num_tran);
+
+      // *INDENT-OFF*
+      for (auto &tran_stat : tran_info.tran_stat)
```
wrlawodms, line 8386:
```suggestion
      for (const auto &tran_stat : tran_info.tran_stat)
```
Let's make all const if possible.

src/executables/util_cs.c
```diff
@@ -4603,17 +4609,16 @@ memmon (UTIL_FUNCTION_ARG * arg)
 	}
     }
 
-  if (module)
+  if (print_module)
```
wrlawodms: It's not a flag, so I'm againt applying the same rule with `print_transaction`. I think just `module` is enough. What do you think?

src/base/memory_monitor_cl.hpp
```diff
@@ -24,10 +24,15 @@
 #ifndef _MEMORY_MONITOR_CL_HPP_
 #define _MEMORY_MONITOR_CL_HPP_
 
+#include <algorithm>
+
 #include "memory_monitor_common.h"
 
+constexpr int DEFAULT_OPTION_PRINT_CNT = 5;
```
wrlawodms: I'd put it in the `memmon` function as a local variable.

src/executables/util_cs.c
```diff
@@ -4649,9 +4652,71 @@ memmon (UTIL_FUNCTION_ARG * arg)
...
+	}
+
+      // TODO: mmon_print_module_info (module_info);
+    }
+  else
+    {
+      if (print_show_all)
+	{
+	  module_count = MMON_MODULE_LAST;
+	  print_transaction = true;
```
wrlawodms: The `print_transaction` is always true if `module_index < 0`. We can take it out of this branch.

src/base/memory_monitor_sr.cpp
wrlawodms: Ok, I understand. Hmm... to make the rule clearer, how about putting the subtraction in the client-side? I mean to have the server always assume that the module_index starts from 0.

src/communication/network_interface_cl.c
wrlawodms: Looking around other functions using `net_client_request2`, they use the first unpacked int, area_size, to decide to proceed to unpack variables. I don't really know about it, but I doubt we can just ignore it. Could you elaborate on it if we can ignore it?

src/executables/util_cs.c
```diff
@@ -4649,9 +4652,71 @@ memmon (UTIL_FUNCTION_ARG * arg)
...
+  /* execute phase */
+  error_code = mmon_get_server_info (server_info);
+  if (error_code != NO_ERROR)
+    {
+      goto error_exit;
+    }
+
+  // TODO: mmon_print_server_info (server_info);
+
+  if (module_index >= 0)
```
hornetmj: The module_index is initialized to MMON_MODULE_LAST. When the user executes the memon utility with default option, the mmon_get_module_info() is called unexpectedly.

src/executables/util_cs.c
```diff
@@ -4649,9 +4652,71 @@ memmon (UTIL_FUNCTION_ARG * arg)
...
+      // TODO: mmon_print_module_info (module_info);
+    }
+  else
+    {
+      if (print_show_all)
+	{
+	  module_count = MMON_MODULE_LAST;
+	  print_transaction = true;
+	}
+      else if (!print_transaction && !print_module)
```
hornetmj: Nit:

The print_transaction variable is set in the if block. However, it is used as a condition of the else if. It looks a bit awkward. I recommand you to set in advance all related condition variables like print_transaction and information variables like tran_count.

Additionally, how about introducing a flag variable like is_module_summary?
```
if (is_module_summary)
{
    mmon_get_module_info_summary()
}
else
{
    mmon_get_module_info()
}
```

src/communication/network_interface_cl.c
```diff
@@ -8124,6 +8124,281 @@ logtb_dump_trantable (FILE * outfp)
...
+  char *request = OR_ALIGNED_BUF_START (a_request);
+  OR_ALIGNED_BUF (OR_INT_SIZE + OR_INT_SIZE) a_reply;
+  char *reply = OR_ALIGNED_BUF_START (a_reply);
+  int req_error, dummy;
+  int error = NO_ERROR;
+
+  ptr = or_pack_int (request, tran_count);
+
+  req_error =
+    net_client_request2 (NET_SERVER_MMON_GET_MODULE_INFO_SUMMARY, request, OR_ALIGNED_BUF_SIZE (a_request),
```
```suggestion
    net_client_request2 (NET_SERVER_MMON_GET_TRAN_INFO, request, OR_ALIGNED_BUF_SIZE (a_request),
```

src/communication/network_interface_sr.c
```diff
@@ -7347,6 +7347,12 @@ smmon_get_module_info (THREAD_ENTRY * thread_p, unsigned int rid, char *request,
 
   if (module_index == 0)
     {
+      /* If there is no modules that registered in memory monitoring manager module,
```
hornetmj:
```suggestion
      /* If there is no modules that registered in memory monitoring manager,
```

hornetmj: MMON_MODULE_ID is already opened to the client. So, this cheking can be executed in the client-side.

src/communication/network_interface_cl.c
wrlawodms: oops, it's impossible. Sorry.

src/executables/util_cs.c
Rudeus: I believe it is better to use prefix `print_` with the main options entered by the user.

src/base/memory_monitor_cl.hpp
Rudeus: Done.

src/executables/util_cs.c
Rudeus: Done.

src/executables/util_cs.c
Rudeus: I add the flag `print_default` and set the flag at the end of input check phase.

src/communication/network_interface_cl.c
Rudeus: Done.

src/communication/network_interface_sr.c
Rudeus: You mean that moving this checking phase to the client-side is a better approach?

src/executables/util_cs.c
Rudeus: Done.

src/communication/network_interface_sr.c
hornetmj: Yes, I do. Do you have any other opinion?

src/executables/util_cs.c
```diff
@@ -4579,7 +4587,7 @@ memmon (UTIL_FUNCTION_ARG * arg)
       goto error_exit;
     }
```
hornetmj, line 4606: It would be better to check before or after checking options, not in between. (check_database_name ())

src/executables/util_cs.c
```diff
@@ -4624,14 +4631,12 @@ memmon (UTIL_FUNCTION_ARG * arg)
       if (module_index < 0 || module_index > MMON_MODULE_LAST)
```
```suggestion
      if (module_index < 0 || module_index >= MMON_MODULE_LAST)
```

src/executables/util_cs.c
```diff
@@ -4649,9 +4654,75 @@ memmon (UTIL_FUNCTION_ARG * arg)
...
+  /* execute phase */
+  error_code = mmon_get_server_info (server_info);
+  if (error_code != NO_ERROR)
+    {
+      goto error_exit;
+    }
+
+  // TODO: mmon_print_server_info (server_info);
+
+  if (module_index >= 0)
```
hornetmj: Nit:

Using print_module flag is better than module_index.

```
if (print_module)
{
    assert (module_index >= 0 && module_index < MMON_MODULE_LAST);
}
else
{

}
```

src/communication/network_interface_sr.c
hornetmj: I think it would be add assert for it instead of handling of this case. Because at least one moudle(heap) is added after terminating this project.

src/executables/util_cs.c
Rudeus: Done.

src/executables/util_cs.c
Rudeus: The `module_index` entered by the user has a value 1 greater than the enum value of the module registered in `MMON_MODULE_ID`, so having the same value as `MMON_MODULE_LAST` is valid.

src/executables/util_cs.c
Rudeus: I agree. Done.

src/base/memory_monitor_cl.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+/*
+ * memory_monitor_cl.cpp - implementation of memory monitoring manager client
+ */
+
+#include <cstring>
+
+#include "memory_monitor_cl.hpp"
+
+int mmon_convert_module_name_to_index (char *module_name)
```
wrlawodms:
```suggestion
int mmon_convert_module_name_to_index (const char *module_name)
```

src/communication/network_interface_cl.c
```diff
@@ -8124,6 +8124,281 @@ logtb_dump_trantable (FILE * outfp)
...
+    }
+  else
+    {
+      ptr = reply;
+      ptr = or_unpack_int (ptr, &dummy);
+      ptr = or_unpack_int (ptr, &error);
+    }
+
+  if (error == NO_ERROR)
+    {
```
wrlawodms, line 8161: Nit. I understand it works from your elaboration, but how about changing `dummy` to `area_size` like other network interfaces and adding an assertion, that claims your analysis?
```suggestion
    {
       assert (area_size > 0);
```

src/executables/util_cs.c
```diff
@@ -4547,39 +4547,47 @@ int
 memmon (UTIL_FUNCTION_ARG * arg)
 {
 #if defined(CS_MODE)
+  const int DEFAULT_OPTION_PRINT_CNT = 5;
```
```suggestion
  constexpr int DEFAULT_OPTION_PRINT_CNT = 5;
```

src/communication/network_interface_sr.c
```diff
@@ -7347,6 +7347,12 @@ smmon_get_module_info (THREAD_ENTRY * thread_p, unsigned int rid, char *request,
 
   if (module_index == 0)
     {
+      /* If there is no modules that registered in memory monitoring manager,
+       * just return */
+      if (MMON_MODULE_LAST == 0)
```
wrlawodms: What happens if `module_index != 0` and `MMON_MODULE_LAST == 0`? It seems that it could happen if we assume that `module_index ==0` and `MMON_MODULE_LSAT == 0` is possible.
```suggestion
      if (MMON_MODULE_LAST == 0)
```

src/communication/network_interface_cl.c
```diff
@@ -8124,6 +8124,281 @@ logtb_dump_trantable (FILE * outfp)
...
+      if (module_index == 0)
+	{
+	  module_count = MMON_MODULE_LAST;
```
wrlawodms, line 8219: Hmm, you assume the `modules_count` can be 0 (on server-side), than, we don't need to the following jobs if it's 0.

Or, I think we can just stop or return an error when the MMON_MODULE_LSA is 0 here, and add an assertion in the server-sizde code. What do you think?

src/communication/network_interface_sr.c
wrlawodms: depending on https://github.com/CUBRID/cubrid/pull/4590#discussion_r1297888097

src/communication/network_interface_cl.c
wrlawodms: No, I don't think it's needed. Forget this.

src/base/memory_monitor_cl.cpp
```diff
@@ -24,7 +24,7 @@
 
 #include "memory_monitor_cl.hpp"
 
-int mmon_convert_module_name_to_index (char *module_name)
+int mmon_convert_module_name_to_index (const char *module_name)
```
wrlawodms: it's declaration has to be changed as well :)

Rudeus: I'll check it.

src/communication/network_interface_cl.c
Rudeus: I prevent to request to server at `memmon()`. Please check it.

src/communication/network_interface_sr.c
Rudeus: I prevent to request to server at `memmon()`. Please check it.

src/executables/util_cs.c
Rudeus: Done.

wrlawodms (approved):

src/communication/network_interface_sr.c
Rudeus: I prevent to request to server at memmon(). Please check it.

src/executables/util_cs.c
hornetmj: I understand.

Nit:
To syncronize the module index entered by user and registered module index in the MMON_MODULE_ID, how about moving -all option to the last?

Rudeus: Regardless of -m all, I believe that starting indexing of specific modules from 1 in the user's perspective is correct for the consistency of cubrid utility(ex) diagdb).

hornetmj: I don't agree with you. Based on your opinion -all option should be -1 because diagdb uses -1 for all (-1 dump all). And the module index assigned by you has no specific information. Share your opinion.

Rudeus: Before the addition of the `-m all` option, the index of individual modules as seen by the user started from 1, based on the precedent of `diagdb`, which was already agreed. Additionally, in order to synchronize the indexing between the server and the user, there was a proposal to insert a dummy in `MMON_MODULE_ID`. However, due to concerns that inserting a dummy could lead to human errors, the decision was made to remove the dummy. The resulting index discrepancy was resolved by subtracting 1 from the index at server, a conclusion reached during the prior discussion.

Now, suggesting that the index of individual modules as viewed by the user should start from 0 is, in my opinion, a kind of specification change. I am not fully convinced or in agreement to make this change at this point.

Rudeus: Done.

hornetmj: We already know what you said. I just leave a comment to reduce mis-understanding caused by mis-matching between index inputed by user and index registered internally. And we agree with make two indexes matching off-line.

msg/en_US.utf8/utils.msg
```diff
@@ -1330,5 +1330,5 @@ valid options:\n\
                                   WARNING: This option has to be used with '-t(--transaction)' option.\n\
     -m, --module=TYPE             display detailed memory usage information of the specified TYPE.\n\
                                   TYPE of modules\n\
-                                  0 (or all)      all modules\n\
-                                  1 (or heap)     heap module\n
+                                  0 (or heap)     heap module\n\
+                                  1 (or all)      all modules\n
```
joohok: what if the new module is added ?
is the new module appended to the end ?  
```
0 (or heap)
1 (or all)
2 (or new-module) ...
```

Rudeus: `0 (or heap)
1 (or new-module)
2 (or all)`

wrlawodms: In my opinion, a fixed index should be for "all". I'd use -1 as the `diagdb` does. We should assume all options can be used in a script by a user. Then, the user has to revise the script every time a module is added.

wrlawodms (changes requested):

Rudeus:
https://github.com/CUBRID/cubrid/pull/4590#discussion_r1299450895

Considering that modifying the message file is necessary when adding a new module anyway, I don't think there would be a significant difference in the effort required. However, is there a compelling reason why the index for the all option needs to be fixed? If the index is set to -1, in contrast to `diagdb`, where entering a string still allows for selecting module, `memmon` would require additional handling for this purpose.

wrlawodms:
> [#4590 (comment)](https://github.com/CUBRID/cubrid/pull/4590#discussion_r1299450895)
> 
> Considering that modifying the message file is necessary when adding a new module anyway, I don't think there would be a significant difference in the effort required. However, is there a compelling reason why the index for the all option needs to be fixed? If the index is set to -1, in contrast to `diagdb`, where entering a string still allows for selecting module, `memmon` would require additional handling for this purpose.

"We should assume all options can be used in a script by a user. Then, the user has to revise the script every time a module is added.". That's the compelling reason for me. If another module is added so 2 is assigned to "all", the user runs into a problem and has to fix it. It's not about a developer modifying the message file, but a user using the utility.

Rudeus:
https://github.com/CUBRID/cubrid/pull/4590#issuecomment-1685428525

Ok. I understand it. What do you think of this comment? @hornetmj @joohok

hornetmj:
> [#4590 (comment)](https://github.com/CUBRID/cubrid/pull/4590#issuecomment-1685428525)
> 
> Ok. I understand it. What do you think of this comment? @hornetmj @joohok

Very good feedback @wrlawodms, @joohok. I agree with them.

joohok:
> [#4590 (comment)](https://github.com/CUBRID/cubrid/pull/4590#issuecomment-1685428525)
> 
> Ok. I understand it. What do you think of this comment? @hornetmj @joohok

Totally agree.

src/base/memory_monitor_sr.cpp
```diff
@@ -674,7 +672,7 @@ void mmon_aggregate_server_info (MMON_SERVER_INFO &info)
 void mmon_aggregate_module_info (int module_index, std::vector<MMON_MODULE_INFO> &info)
 {
   assert (mmon_Gl != nullptr);
-  assert (module_index < (int)MMON_MODULE_LAST && module_index >= 0);
+  assert (module_index < (int)MMON_MODULE_LAST && module_index >= -1);
```
joohok: Rather than using `-1` directly, how about using some meaningful constant value ? 
ex) MMON_MODULE_ALL ?

Rudeus: I believe that if there is a place where that meaning should be included, it would be `MMON_MODULE_ID`. However, I am uncertain whether placing that information here would be appropriate, so I opted to leave it as a raw value.

src/communication/network_interface_sr.c
```diff
@@ -7564,6 +7564,10 @@ smmon_get_tran_info (THREAD_ENTRY * thread_p, unsigned int rid, char *request, i
   // Size of transaction id (OR_INT_SIZE)
   // and current memory usage (OR_INT64_SIZE)
   size += ((OR_INT_SIZE + OR_INT64_SIZE) * info.num_tran);
+  if (size == 0)
```
wrlawodms: It won't be reached since `size >= OR_INT_SIZE` by the size of `num_tran`.

src/executables/util_admin.c
```diff
@@ -895,7 +895,7 @@ static GETOPT_LONG ua_Flashback_Option[] = {
 
 static UTIL_ARG_MAP ua_Memmon_Option_Map[] = {
   {OPTION_STRING_TABLE, {0}, {0}},
-  {MEMMON_MODULE_S, {ARG_STRING}, {0}},
+  {MEMMON_MODULE_S, {ARG_INTEGER}, {(void *) -2}},
```
wrlawodms, line 898:
```suggestion
  {MEMMON_MODULE_S, {ARG_INTEGER}, {(void *) -1}},
```
Could you tell me why it is -2?

src/base/memory_monitor_sr.cpp
wrlawodms: I think that it's not bad to put `MMON_MODULE_ALL` into the enum. At your choice.

src/executables/util_admin.c
Rudeus: because -1 is for "-m all" case. So we have to avoid default value with -1.

wrlawodms: Ah, ok. All is not the default.

src/communication/network_interface_sr.c
Rudeus: I got it. Thanks.

wrlawodms (approved):

src/base/memory_monitor_sr.cpp
Rudeus: I add it.

msg/en_US/utils.msg
```diff
@@ -1330,5 +1330,5 @@ valid options:\n\
                                   WARNING: This option has to be used with '-t(--transaction)' option.\n\
     -m, --module=TYPE             display detailed memory usage information of the specified TYPE.\n\
                                   TYPE of modules\n\
-                                  0 (or all)      all modules\n\
-                                  1 (or heap)     heap module\n
+                                  -1      all modules\n\
+                                   0      heap module\n
```
joohok:
```suggestion
```
how about adding heap module when memory monitoring for heap module is implemented? 

Since developers can see which files were modified when the heap module was added, developers looking to implement memory monitoring for other modules might find it easy to refer to

joohok (approved):

msg/en_US/utils.msg
Rudeus: You're right. I remove it.

src/executables/util_cs.c
```diff
@@ -4547,39 +4547,58 @@ int
...
     {
-      PRINT_AND_LOG_ERR_MSG (msgcat_message
-			     (MSGCAT_CATALOG_UTILS, MSGCAT_UTIL_SET_MEMMON, MEMMON_MSG_EXCLUSIVE_OPTION));
       goto error_exit;
     }
 
-  if (check_database_name (database_name))
+  /* input error check phase */
+
+  if (module_index >= MMON_MODULE_ALL && module_index < MMON_MODULE_LAST)
```
hornetmj, line 4584: Later, 0 module_index should be filtered out here.

src/executables/util_cs.c
```diff
@@ -4603,35 +4622,14 @@ memmon (UTIL_FUNCTION_ARG * arg)
 	}
     }
 
-  if (module)
+  if (module_index < MMON_MODULE_ALL || module_index >= MMON_MODULE_LAST)
```
hornetmj: Is it duplicated check at 4584 line? (if (module_index >= MMON_MODULE_ALL && module_index < MMON_MODULE_LAST))

It can combine to the one checking.

Rudeus: The reason is that I thought exclusive option error checking should precede module index error checking. I separated them to maintain the order, but do you think it's okay for the order to change when they are combined?

src/executables/util_cs.c
Rudeus: It's not clear and hard to understand the intention. Please provide a more detailed explanation.

hornetmj (approved):

PR #4592, opened 2023-08-17
[CBRD-24893] Get stmt type from dblink query.

airnet73:
http://jira.cubrid.org/browse/CBRD-24893

Purpose
The gateway does not have a query parser, so it is difficult to know the stmt type.
So, dblink server added stmt type to query comment.
The gateway can know the stmt type through comment of the query.

Implementation
N/A

Remarks
N/A


src/broker/cas_execute.c
```diff
@@ -11753,4 +11765,36 @@ ux_cgw_free_stmt (T_SRV_HANDLE * srv_handle)
...
+
+	      ut_trim (type_name);
+	      ut_tolower (type_name);
+
+	      stmt_type = get_stmt_type (type_name);
+	    }
+	}
+    }
+
+  if (stmt_type == CUBRID_STMT_NONE)
```
kisoo-han: get_stmt_type () returns **CUBRID_MAX_STMT_TYPE** if there is no string matching the input argument. Is it intended that it does not return CUBRID_STMT_NONE?

src/broker/cas_execute.c
```diff
@@ -11753,4 +11765,36 @@ ux_cgw_free_stmt (T_SRV_HANDLE * srv_handle)
 {
   cgw_free_stmt (srv_handle);
 }
+
+static char
+ux_cgw_get_stmt_type (char *stmt)
+{
+  char stmt_type = CUBRID_STMT_NONE;
+  const char *comment_start = strstr (stmt, "/*");
+  if (comment_start)
```
kisoo-han: If hints other than the STMT type hint could be sent to the gateway, hint parsing will have to be repeated. If STMT type hint is only one, its ok.

src/broker/cas_cgw.c
```diff
@@ -1770,6 +1770,12 @@ cgw_get_num_cols (SQLHSTMT hstmt, SQLSMALLINT * num_cols)
 
   SQL_CHK_ERR (hstmt, SQL_HANDLE_STMT, err_code = SQLNumResultCols (hstmt, num_cols));
 
+  if (*num_cols == 0)
+    {
+      SQL_CHK_ERR (hstmt, SQL_HANDLE_STMT, err_code = SQLExecute (hstmt));
+      SQL_CHK_ERR (hstmt, SQL_HANDLE_STMT, err_code = SQLNumResultCols (hstmt, num_cols));
+    }
+
   return (int) err_code;
```
kisoo-han, line 1779: Maybe you combined two different issues in one jira issue and a pull request. Next time, obey the rule that a jira issue should only cover one issue at a time.

kisoo-han (approved):

src/broker/cas_execute.c
airnet73: Exception handling will be added when CUBRID_MAX_STMT_TYPE is returned.

src/broker/cas_execute.c
airnet73: The current format has a hint at the beginning of sql, so only the first hint was searched. ( /* DBLINK SELECT */ select * from t1; )
I will consider the case of multiple hints.

src/broker/cas_cgw.c
airnet73: I will definitely separate them from next time.

kisoo-han (approved):

PR #4593, opened 2023-08-18
[LETS-747] Request to connect from PS to PS for the catch-up

wrlawodms:
http://jira.cubrid.org/browse/LETS-747

Add a new connection type, CONNECT_PAGE_TO_PAGE_SERVER.
Nothing to note. Please see the Jira issue.

### DIFF
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -504,14 +504,13 @@ page_server::set_passive_tran_server_connection (cubcomm::channel &&chn)
 void
 page_server::set_follower_page_server_connection (cubcomm::channel &&chn)
 {
-  assert (is_page_server ());
-
   chn.set_channel_name ("PS_PS_catchup_comm");
 
   assert (chn.is_connection_alive ());
   const auto channel_id = chn.get_channel_id ();
 
-  er_log_debug (ARG_FILE_LINE, "A page server connected to this page server. Channel id: %s.\n",
+  er_log_debug (ARG_FILE_LINE,
+		"A follower page server connected to this page server as a leader to catch up. Channel id: %s.\n",
 		channel_id.c_str ());
 
   // TODO add a connection_handler for this will be registered
@@ -553,6 +552,10 @@ page_server::connect_to_leader_page_server (std::string &&hostname, int32_t port
     {
       return ps_conn_error_lambda ();
     }
+
+  er_log_debug (ARG_FILE_LINE,
+		"This page server successfully connected to the leader page server to catch up. Channel id: %s.\n",
+		srv_chn.get_channel_id ().c_str ());
 
   return NO_ERROR;
 }
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -18,6 +18,7 @@
 
 #include "page_server.hpp"
 
+#include "communication_server_channel.hpp"
 #include "disk_manager.h"
 #include "error_manager.h"
 #include "log_impl.h"
@@ -492,6 +493,61 @@ page_server::set_passive_tran_server_connection (cubcomm::channel &&chn)
   }
 
   m_pts_mvcc_tracker.init_oldest_active_mvccid (channel_id);
+}
+
+void
+page_server::set_follower_page_server_connection (cubcomm::channel &&chn)
+{
+  assert (is_page_server ());
+
+  chn.set_channel_name ("PS_PS_catchup_comm");
+
+  assert (chn.is_connection_alive ());
+  const auto channel_id = chn.get_channel_id ();
+
+  er_log_debug (ARG_FILE_LINE, "A page server connected to this page server. Channel id: %s.\n",
+		channel_id.c_str ());
+
+  // TODO add a connection_handler for this will be registered
+}
+
+int
+page_server::connect_to_leader_page_server (cubcomm::node &&node)
+{
+  auto ps_conn_error_lambda = [&node] ()
+  {
+    er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_NET_PAGESERVER_CONNECTION, 1, node.get_host ().c_str ());
+    return ER_NET_PAGESERVER_CONNECTION;
+  };
+
+  // connect to leader page server
+  constexpr int CHANNEL_POLL_TIMEOUT = 1000;    // 1000 milliseconds = 1 second
+  cubcomm::server_channel srv_chn (m_server_name.c_str (), SERVER_TYPE_PAGE, CHANNEL_POLL_TIMEOUT);
+
+  srv_chn.set_channel_name ("PS_PS_catchup_comm");
+
+  auto comm_error_code = srv_chn.connect (node.get_host ().c_str (), node.get_port (),
+					  CMD_SERVER_SERVER_CONNECT);
+  if (comm_error_code != css_error_code::NO_ERRORS)
+    {
+      return ps_conn_error_lambda ();
+    }
+
+  constexpr auto conn_type = cubcomm::server_server::CONNECT_PAGE_TO_PAGE_SERVER;
+  if (srv_chn.send_int (static_cast<int> (conn_type)) != NO_ERRORS)
+    {
+      return ps_conn_error_lambda ();
+    }
+
+  int returned_code;
+  if (srv_chn.recv_int (returned_code) != css_error_code::NO_ERRORS)
+    {
+      return ps_conn_error_lambda ();
+    }
+  if (returned_code != static_cast<int> (conn_type))
+    {
+      return ps_conn_error_lambda ();
+    }
 }
 
 void
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -526,7 +526,6 @@ page_server::connect_to_leader_page_server (std::string &&hostname, int32_t port
     return ER_NET_PAGESERVER_CONNECTION;
   };
 
-  // connect to leader page server
   constexpr int CHANNEL_POLL_TIMEOUT = 1000;    // 1000 milliseconds = 1 second
   cubcomm::server_channel srv_chn (m_server_name.c_str (), SERVER_TYPE_PAGE, CHANNEL_POLL_TIMEOUT);
 
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -212,7 +212,7 @@ page_server::connection_handler::receive_start_catch_up (tran_server_conn_t::seq
     }
 
   // TODO: A thread will take the catch-up including establishing connection to avoid blocking ATS->PS reqeusts.
-  m_ps.connect_to_leader_page_server (std::move (host), port);
+  m_ps.connect_to_followee_page_server (std::move (host), port);
 }
 
 
@@ -511,14 +511,14 @@ page_server::set_follower_page_server_connection (cubcomm::channel &&chn)
   const auto channel_id = chn.get_channel_id ();
 
   er_log_debug (ARG_FILE_LINE,
-		"A follower page server connected to this page server as a leader to catch up. Channel id: %s.\n",
+		"A follower page server connected to this page server to catch up. Channel id: %s.\n",
 		channel_id.c_str ());
 
   // TODO Create a connection_handler for this.
 }
 
 int
-page_server::connect_to_leader_page_server (std::string &&hostname, int32_t port)
+page_server::connect_to_followee_page_server (std::string &&hostname, int32_t port)
 {
   auto ps_conn_error_lambda = [&hostname] ()
   {
@@ -558,7 +558,7 @@ page_server::connect_to_leader_page_server (std::string &&hostname, int32_t port
   // We will create a connection handler to keep this channel and handle requests for them.
 
   er_log_debug (ARG_FILE_LINE,
-		"This page server successfully connected to the leader page server to catch up. Channel id: %s.\n",
+		"This page server successfully connected to the followee page server to catch up. Channel id: %s.\n",
 		srv_chn.get_channel_id ().c_str ());
 
   return NO_ERROR;
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -553,6 +553,10 @@ page_server::connect_to_leader_page_server (std::string &&hostname, int32_t port
       return ps_conn_error_lambda ();
     }
 
+  // TODO
+  // For now, the srv_chn is destroyed here and it will close the connection.
+  // We will create a connection handler to keep this channel and handle requests for them.
+
   er_log_debug (ARG_FILE_LINE,
 		"This page server successfully connected to the leader page server to catch up. Channel id: %s.\n",
 		srv_chn.get_channel_id ().c_str ());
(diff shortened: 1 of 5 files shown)


src/server/page_server.cpp
```diff
@@ -205,6 +206,13 @@ page_server::connection_handler::receive_start_catch_up (tran_server_conn_t::seq
 
   er_log_debug (ARG_FILE_LINE, "receive_start_catch_up: hostname = %s, port = %d, LSA = (%lld|%d)\n", host.c_str (), port,
 		LSA_AS_ARGS (&catchup_lsa));
+  if (port == -1)
+    {
+      return; // TODO: It means that the ATS is booting up, it will be set properly after ATS recovery is implemented.
+    }
+
+  // TODO: A thread will take the catch-up including establishing connection to avoid blocking ATS->PS reqeusts.
+  m_ps.connect_to_leader_page_server (std::move (host), port);
```
hornetmj: Do every page servers which enter the cluster use this leader PS when catching up (in the completed design)? Or do you want use "leader ps" as a logical meaning?

wrlawodms: 1. Basically, it's a logical naming. More than one node can be leader.
2. We can guarantee that at least one node can be the leader. If there is only one reliable node, several page servers request the catch-up to that node. There is room to improve this with a catch-up policy. For example, 
    - While booting up the ATS, we can delay some catch-up to assign it to another node  which lesser falls behind so is expected get the catch-up done quickly.
    - While ruuing the ATS, there are at least majority of nodes are realiable, so we can spread the catch-up requests. In this case, we can make one node at most is assigned a PS.

hornetmj: The keyword 'leader' seems a bit unsuitable to me. Because it is usually thought to contain the meaning of being the only one of many. Accordingly, it can be felt that the PS that serves as the leader role among the PS clusters exists.

wrlawodms: I agree it can be confusing. I actually struggled to come up with a proper name. Could you recommend more suitable names?

joohok: I think error handling is required when `connect_to_leader_page_server ()` is failed.
it might need to respond to ATS with an error code. 

Would this be addressed in other issue ?

wrlawodms: yes, in [LETS-635](http://jira.cubrid.org/browse/LETS-635)
Stabilize Multi-PS and mitigate its restrictions.
```
Handle the exceptions during the catch-up process. ([LETS-632](http://jira.cubrid.org/browse/LETS-632))
- The target PS may truncate some of log pages requested.
- The target PS may be disconnected normally or abnormally.
- And any error during the process like corrupting log pages for some reason.
```

Errors or disconnection from the leader PS are detected in another issue in this epic, but it's going to be addressed in the above epic. I think, in any case, the follower PS re-request the catch-up information to the ATS.

joohok: OK! 
`The target PS may be disconnected normally or abnormally.` 
Then, this should include scenarios where the leader PS is shut-down before catch-up initiation or if the connection with the leader PS is disconnected during catch-up

hornetmj: How about supporting a general interface to connect to page server like connect_to_page_server() if there is no special action for catch up in the function?

wrlawodms: Right, it's described in http://jira.cubrid.org/browse/LETS-784, which will be handled in this epic.

wrlawodms: How about creating what it's needed and expanding it when more is needed? It's my approach and I believe that it simplifies the implementation. 
For example, I presume that requests for the catch-up are one-way from a follower ps to a leader ps, and the leader ps just reply. With this, we can simplify the disconnection process. When the follower ps wants to disconnect after it's done the catch-up, it can just disconnect.

src/server/page_server.hpp
```diff
@@ -213,6 +219,8 @@ class page_server
     responder_t &get_responder ();
 
   private: // members
+    std::string m_server_name;
```
cristiarg:
```suggestion
    const std::string m_server_name;
```

src/server/page_server.cpp
```diff
@@ -205,6 +206,13 @@ page_server::connection_handler::receive_start_catch_up (tran_server_conn_t::seq
 
   er_log_debug (ARG_FILE_LINE, "receive_start_catch_up: hostname = %s, port = %d, LSA = (%lld|%d)\n", host.c_str (), port,
 		LSA_AS_ARGS (&catchup_lsa));
+  if (port == -1)
```
cristiarg, line 209: Since this decision is taken on ATS, what is the reason to do a `active_tran_server::connection_handler::send_start_catch_up_request` -> `page_server::connection_handler::receive_start_catch_u` roundtrip to the PS just to stop here?
Or is there some ulterior motive?

src/server/page_server.cpp
```diff
@@ -494,6 +502,64 @@ page_server::set_passive_tran_server_connection (cubcomm::channel &&chn)
...
+    }
+  if (returned_code != static_cast<int> (conn_type))
+    {
+      return ps_conn_error_lambda ();
+    }
+
+  er_log_debug (ARG_FILE_LINE,
+		"This page server successfully connected to the leader page server to catch up. Channel id: %s.\n",
+		srv_chn.get_channel_id ().c_str ());
+
```
cristiarg, line 563: Nit.
Some TODO comment about what to do with the resulting `srv_chn`.
The `cubcomm::channel::~channel` will close the connection.

src/server/page_server.cpp
```diff
@@ -494,6 +502,64 @@ page_server::set_passive_tran_server_connection (cubcomm::channel &&chn)
   m_pts_mvcc_tracker.init_oldest_active_mvccid (channel_id);
 }
 
+void
+page_server::set_follower_page_server_connection (cubcomm::channel &&chn)
```
cristiarg, line 506: Nit.
There should be some fixed taxonomy here - `catchup` or `follower`. Or is `follower_page_server` the counterpart of `leader_page_server` ?
Just nitpicking. We can also decide later.

src/server/page_server.cpp
cristiarg: @wrlawodms
Remark.
Is it safe to assume that `page_server::connection_handler` class will also need to implement some state transition similar to `tran_server::connection_handler`?
Conceptual states:
1. `IDLE`
2. `CONNECTING` (ATS/PTS connecting to PS, maybe not needed)
3. `CATCHUP_CONNECTING` (PS needs to perfom catch-up and is connecting to the leader PS)
4. `CATCHUP_CATCHING_UP` (PS connected to leader PS and performing catching up)
5. `CATCHUP_DISCONNECTING` (PS done catching up and is disconnecting from leader PS, maybe not needed, jump straight to next one)
6. `CONNECTED` (PS online; either catch-up has been done or not needed)
7. `DISCONNECTING` (ATS/PTS disconnecting from PS, maybe not needed)

Transitions:
- 1 -> 2 -> 3 -> 4 -> 5 -> 6 -> 7 -> 1
- 1 -> 2 -> 6 -> 7 -> 1

hornetmj: Sorry, I catch your response too late.

I respect your approach. It is very good to me. However, we need to focus on the changes of the naming 'leader'. Do you have any good candidates for it? Or do you just use it? I can see that "leader" has a semantic issue, but I can't come up with a suitable alternative either.

src/server/page_server.cpp
wrlawodms: I don't think I understand you. The cachup process will be like this:
1. A new PS, follower, attaches its ATS
2. The ATS elects a leader PS, or it's been elected. The leader PS is reliable enough to catch up.
3. The ATS sends this information to the follower PS.
4. The follower PS requests the catch-up the leader PS based on the information. It happens in this function, receive_start_catch_up(). The catch-up process will be taken by another thread which will be initiated here to avoid blocking this PS's request handler for the ATS.
6. A connection between two PSs will be established.

Could you point out where you feels weird?

src/server/page_server.cpp
wrlawodms: Thank you for thinking thoroughly. In my opinion, we can define the state transition, but I can't figure out its necessity yet. If there are some operations only allowed in some states, I think that's the time to introduce the states. The `page_esrver::conneciton_handler` is even a transient object while the TS's is permanent. What do you think? Do you think we'd be better off introducing the states in advance?

src/server/page_server.cpp
wrlawodms: You're right. My taxonomy is:
- The action: catch-up
- The PS doing the catch-up: follower
- The PS which is the target to catch up with: leader.

However, other reviewers also point out it's quite confusing.. I should come up with other ones..

src/server/page_server.cpp
wrlawodms: > The cubcomm::channel::~channel will close the connection.

I'm aware of this.

> Some TODO comment about what to do with the resulting srv_chn.

Do you mean we should leave a comment here, right? OK.

src/server/page_server.cpp
cristiarg: It is sufficient for me.

src/server/page_server.cpp
cristiarg: No, I don't think it is needed right now. As you say, let's keep it simple for now.

I see the following structure moving forward - which is very much in line and a continuation of what is added by this PR:
- the one flavor currently existing `page_server::connection_handler` which is the passive one:
  - serves requests from xTSs
    - there is currently one instance per connected xTs
    - and this is very much permanent, not transient
  - this connection will be extended to respond to requests from the "follower" PS
    - no need to have yet another implementation on the "leader" side
    - for the connected follower PS a new - transient - instance of such a connector will be added by the leader PS and removed once the follower PS will disconnect (having performed the catch-up)
- a new flavor of connection handler - `catchup_connection_handler` which will be instantiated by the follower PS and will live as long as there is catchup to do
  - this new flavor is the one that is instantiated at about where we have now the line:
```
  // TODO: A thread will take the catch-up including establishing connection to avoid blocking ATS->PS reqeusts.
  m_ps.connect_to_leader_page_server (std::move (host), port);
```
  - and I see here a need for the states being introduced - the `page_server` object instance needs to "play" with this - conceptual entity - "catchup_connection_handler" asynchronously
  - the conceptual state of the `page_server` instance is suspended as long as the `catchup_connection_handler` has not finished it work in one way (success) or another (failure).

This is just a suggestion of how the structure might evolve. It may very well be that you have another - maybe clearer - picture.

src/server/page_server.cpp
cristiarg: Yes, it is more a circumstantial comment about the current state of the code - to put a comment about the intention.
For now, the server connection - via `srv_chn` - is instantiated and destroyed within this function.

src/server/page_server.cpp
cristiarg: In the light of what I described above, my first comment about the states and transitions is .. well .. out of date/incorrect/incomplete.

wrlawodms: We have a different picture for the connections. Lets me share mine, which is described other issues, but I understand it's hard to get together.

A page server will have two connection_handlers
- tran_server_connection_handler
  - it's only used to communicate with a transaction server.
- catchup_connection_handler
  - through which PSs communicate with each other. Both a follower and a leader use this connection handler. This will have both the request and its handler for the catch-up.

Let me depict them:
```

                PS1                                                                                                      
     +------------------------------+        +-------------------+                                                         
     |tran_server_connection_handler   <-->   connection_handler |                                                         
   --|catch_upconnection_handler    |        |                   |                                                         
   | +------------------------------+        |                   |                                                         
   | |tran_server_connection_handler|  <-->  |connection_handler |                                                         
   +-|catchup_connection_handler    |        +-------------------+                                                         
     +------------------------------+                ATS1                                                                  
                 PS2                                          
```                   

I can't imagine how we can handler requests of both TS and PS in one connection_handler. The internal server object, `request_sync_client_server`, define requests it can send and receive like `page_to_tran_request` or `tran_to_page_request`. We can add two `tran_to_page_request`, but I think it would get complicated. My plan is to define a new request type, `page_to_page_request` and create a connection_handler using it, which will be totally separate from the connection_handler with a TS.

And I think a page_server can communicate with a tran_server regardless of the state of its catchup_connection_handler.

wrlawodms: > and this is very much permanent, not transient

I think `permanent` is vague. To me, the connection_handler on tran_server always exists regardless of whether its internal connection is being established or not. But, it's not in the case of PS. In the PS, if a TS is disconnected, the conneciton_handler for it is released. It means there is no need for the IDLE state or some states.

Just to clarify.

src/server/page_server.cpp
wrlawodms: I totally agree. Thanks.

src/server/page_server.cpp
cristiarg: > I can't imagine how we can handler requests of both TS and PS in one connection_handler.

The `page_server::connection_handler` on a leader PS already has implemented the handler to server log pages - `tran_to_page_request::SEND_LOG_PAGE_FETCH`. It will probably need to implement some other handlers for extra requests as well.
So the existing connection_handler will be extended to stay on the "serving" side - from the point of view of the PS-PS catchup mechanism.

On the other hand `page_server::catchup_connection_handler` on a follower PS will do the heavy lifting of having to put in agreement whatever log pages it misses and the live prior log that it receives from ATS (here the approach might be different and involve the `unsent_lsa` that you recently added).
And this `catchup_connection_handler` will stay on the "request" side - from the point of view of the PS-PS catchup mechanism.

Having everything related to PS-PS catch-up mechanism bundled together in a separate `catchup_connection_handler` is certainly possible and doable as well. :)

src/server/page_server.cpp
cristiarg: I think, what I meant is the distinction between "cold start" and "catchup start" of PS.
- at cold start, the ATS can decide - in `active_tran_server::connection_handler::on_connecting ` - to skip the catchup step - of course PS connector has to support this
- at "hot/catchup start" the ATS will dispatch the catchup message in `active_tran_server::connection_handler::on_connecting`

But, please disregard this message as I don't want to confuse you even further. :)

wrlawodms: OK. Let's see when the cold-start comes in. :)

src/server/page_server.cpp
wrlawodms: > The page_server::connection_handler on a leader PS already has implemented the handler to server log pages - tran_to_page_request::SEND_LOG_PAGE_FETCH. It will probably need to implement some other handlers for extra requests as well.
So the existing connection_handler will be extended to stay on the "serving" side - from the point of view of the PS-PS catchup mechanism.

We need two internal `request_sync_client_server` in the page_server::connection_handler.
- cubcomm::request_sync_client_server<page_to_tran_request, tran_to_page_request, std::string>;
- cubcomm::request_sync_client_server<page_to_page_request, page_to_page_request, std::string>;

Do you mean this? Or, we can change tran_to_page_request -> xxx_to_page_request and combine the catchup_requests and tran_to_page_request, but it seems better to me to split them.

Besides, I guess we can simplify the serving side connection handler for the catchup by implementing only what's needed. I'll consider your approach too when implementing the connection_handler in the following issue. Let's see in the issue or PR.

wrlawodms: It seems better to use "followee" instead of "leader".

wrlawodms: or "catchup_server" and "catchup_client"?

hornetmj: Follower is used. So, "followee" would be good to me.

hornetmj (approved):

joohok (approved):

PR #4598, opened 2023-08-21
[LETS-783] Have waiters on send_receive notice the abnormal disconnection

wrlawodms:
http://jira.cubrid.org/browse/LETS-783

**Implementation**
- Push recv error handler from connection_handler into `request_server` and call it when an error occurs while receiving a message.
- The recv error handler the tran server uses is just to disconnect the page server.
- The `send_receive` no longer does disconnection.

**Unit tests**
- Most of the changes are about adding unit tests.
    - Detect errors on push_request
    - Detect errors on send_recv_request
- To do that, split the `request_sync_client_server` test case into three sections: the existing on and newly added tests.
- To do that, some functions for the `socket_direction`, the mock for an internal socket,  are added.
    - `disconnect_sender/receiver_socket_direction`: disconnect the socket direction.
    - `[un]freeze_receiver_socket_direction`: block to read a message to simulate a communication delay. Without this, we can't simulate the failure two internal actions, sending, and receiving, in the send_recv.
    - `does_receiver_socket_direction_have_message`: a helper to confirm a message is added.
- Add functions to see if there are requests left. It's only used in unit tests, but could be helpful for the future.

### DIFF
--- a/src/communication/request_client_server.hpp
+++ b/src/communication/request_client_server.hpp
@@ -161,6 +161,7 @@ namespace cubcomm
   {
     public:
       using server_request_handler = std::function<void (cubpacking::unpacker &upk)>;
+      using server_error_handler = std::function<void (css_error_code error)>;
       using server_request_id = MsgId;
 
       request_server () = delete;
@@ -176,6 +177,7 @@ namespace cubcomm
       void stop_thread ();	  // stop the thread
 
       void register_request_handler (MsgId msgid, const server_request_handler &handler);	  // register a handler
+      void register_error_handler (server_error_handler &&error_handler);	  // register an error handler
 
       inline const channel &get_channel () const;						  // get underlying channel
 
@@ -198,6 +200,7 @@ namespace cubcomm
       std::thread m_thread;				// thread that loops and handles requests
       bool m_shutdown = true;				// set to true when thread must stop
       request_handlers_container m_request_handlers;	// request handler map
+      server_error_handler m_error_handler;
       cubmem::extensible_block m_recv_extensible_block;
   };
 
@@ -265,6 +268,7 @@ namespace cubcomm
   template <typename MsgId>
   request_server<MsgId>::request_server (channel &&chn)
     : m_channel (std::move (chn))
+    , m_error_handler (nullptr)
     , m_recv_extensible_block { cubmem::CSTYLE_BLOCK_ALLOCATOR }
   {
     // arbitrary initial size; will be grown upon need
@@ -285,6 +289,13 @@ namespace cubcomm
       {
 	m_request_handlers[msgid] = handler;
       }
+  }
+
+  template <typename MsgId>
+  void request_server<MsgId>::register_error_handler (server_error_handler &&handler)
+  {
+    assert (m_error_handler == nullptr);
+    m_error_handler = std::move (handler);
   }
 
   template <typename MsgId>
@@ -323,16 +334,25 @@ namespace cubcomm
   void request_server<MsgId>::loop_handle_requests ()
   {
     size_t message_size = 0;
+    css_error_code error = NO_ERRORS;
+
     while (!m_shutdown)
       {
 	if (!has_request_on_channel ())
 	  {
 	    continue;
 	  }
-	if (receive_request_buffer (message_size) != NO_ERRORS)
+
+	error = receive_request_buffer (message_size);
+	if (error != NO_ERRORS)
 	  {
+	    if (m_error_handler != nullptr)
+	      {
+		m_error_handler (error);
+	      }
 	    break;
 	  }
+
 	handle_request (message_size);
       }
     er_log_thread_finished (this, &m_thread, m_thread.get_id ());
--- a/src/communication/request_sync_client_server.hpp
+++ b/src/communication/request_sync_client_server.hpp
@@ -74,6 +74,7 @@ namespace cubcomm
       void stop_response_broker ();
       void stop_incoming_communication_thread ();
       void stop_outgoing_communication_thread ();
+      void wait_until_all_requests_sent ();
 
       /* only used by unit tests
        */
@@ -233,6 +234,13 @@ namespace cubcomm
 
     // terminate receiving messages
     m_conn.reset (nullptr);
+  }
+
+  template <typename T_OUTGOING_MSG_ID, typename T_INCOMING_MSG_ID, typename T_PAYLOAD>
+  void
+  request_sync_client_server<T_OUTGOING_MSG_ID, T_INCOMING_MSG_ID, T_PAYLOAD>::wait_until_all_requests_sent ()
+  {
+    m_queue->wait_until_empty ();
   }
 
   template <typename T_OUTGOING_MSG_ID, typename T_INCOMING_MSG_ID, typename T_PAYLOAD>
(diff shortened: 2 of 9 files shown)


unit_tests/request_client_server/test_main.cpp
```diff
@@ -630,44 +647,104 @@ TEST_CASE ("Two request_sync_client_server communicate with each other", "")
...
   {
-    REPEAT_REQUEST_WITH_MSGID (env.push_request_on_scs_two, reqids_2_to_1::_1);
-  });
+    {
+      // Ensure no failure is detected.
+      env.push_request_on_scs_one (reqids_1_to_2::_1, 0);
+      env.wait_for_all_messages ();
 
-  for (auto &th : threads)
+      const auto [send_error, recv_error] = env.get_request_errors_on_scs_one ();
```
wrlawodms, line 696: We could put `REQUIRE`s into the `env` like `requires_errors (false, false)`, but it hides where a violation happens.

src/server/tran_server.hpp
```diff
@@ -178,7 +178,8 @@ class tran_server
       private:
 	void set_connection (cubcomm::channel &&chn);
 	// The default error handler for sending reqeust
```
hornetmj, line 180:
```suggestion
	// The default error handler for send/recv reqeust
```

hornetmj (approved):
LGTM

src/server/tran_server.hpp
wrlawodms: Actually, it's only for send (push_request). In the case of `send_recv`, the sending error handler is pushed per-request-based. See request_sync_client_server::send_recv(). An error handler is pushed with request.

src/communication/request_sync_client_server.hpp
```diff
@@ -230,6 +236,13 @@ namespace cubcomm
     m_conn.reset (nullptr);
   }
 
+  template <typename T_OUTGOING_MSG_ID, typename T_INCOMING_MSG_ID, typename T_PAYLOAD>
+  void
+  request_sync_client_server<T_OUTGOING_MSG_ID, T_INCOMING_MSG_ID, T_PAYLOAD>::wait_until_all_requests_sent ()
```
joohok, line 241: Where would the `wait_until_all_requests_sent ()` function be used? 
I'm asking because I only see it in unit tests

wrlawodms: yes, it is only used in the unit test. It would be made use of in the future, but I'm not sure.

unit_tests/request_client_server/comm_channel_mock.hpp
```diff
@@ -66,16 +66,25 @@ class mock_socket_direction
     void wait_for_all_messages ();              // wait until all messages are pulled and the message queue is empty
     void wait_until_message_count (size_t count);
 
+    void freeze ();                             // Block to read a message to simulate a communication delay
+    void unfreeze ();                           // Unblock to read a message to simulate a coomunication delay
```
joohok:
```suggestion
    void unfreeze ();                           // Unblock to read a message to simulate a communication delay
```

unit_tests/request_client_server/test_main.cpp
```diff
@@ -630,44 +647,104 @@ TEST_CASE ("Two request_sync_client_server communicate with each other", "")
...
+      env.send_recv_request_on_scs_one (reqids_1_to_2::_0, 0);
 
-  require_all_sent_requests_are_handled ();
+      const auto [send_error, recv_error] = env.get_request_errors_on_scs_one ();
+      REQUIRE (send_error == false);
+      REQUIRE (recv_error == false);
+    }
+    {
+      // Send a send_recv request from the scs1 to the scs2 while the internal socket is frozen.
+      // Then, disconnect the ineternal socket and confirm this disconnection is dectected on scs1.
```
joohok:
```suggestion
      // Then, disconnect the internal socket and confirm this disconnection is detected on scs1.
```

unit_tests/request_client_server/test_main.cpp
```diff
@@ -630,44 +647,104 @@ TEST_CASE ("Two request_sync_client_server communicate with each other", "")
...
+      const auto receiver_id = env.get_scs_one ().get_underlying_channel_id ();
+      freeze_receiver_socket_direction (receiver_id);
+
+      auto disconnecter = std::thread ([&] ()
+      {
+	// Wait until a request is pushed and disconnect the socket.
```
joohok:
```suggestion
	// Wait until a request is pushed and the socket is disconnected.
```

joohok (approved):

src/communication/request_client_server.hpp
```diff
@@ -161,6 +161,7 @@ namespace cubcomm
   {
     public:
       using server_request_handler = std::function<void (cubpacking::unpacker &upk)>;
+      using server_error_handler = std::function<void (css_error_code error)>;
```
cristiarg, line 164: this new `server_error_handler` (for receiving requests/messages) and the existing `send_queue_error_handler` (for sending messages/requests) should be moved somewhere together because one is the counterpart of the other;
```
  using send_error_callback_t = std::function<void (css_error_code, bool &)>;
  using recv_error_callback_t = std::function<void (css_error_code error)>;
```
and the usage of the `recv` one have the variables renamed with `recv`
and the usage of the `send` one have the variables renamed also

the idea being that, maybe in the future a better overall image surfaces for a subsequent unification/refactoring:

There are actually two levels of error handlers (callbacks)
1. the connection level error handlers which deal at the physical level - connection interrupted - these two which are discussed here

3. the logical level - error handlers which are needed on the `send_recv` side and which handle the actuall logical errors coming from the peer server (ie: the error that a TS receives from a PS when some request cannot be fulfilled)
    - currently, for each `send_recv` a separate instance of an error handler is created
    - it would be better to just tie an instance of an logical error handler to one `tran_to_page_request` message type - after all, at least now, there is a one-to-one correspondence between the message type and the follow-up handling of either error or response (there is http://jira.cubrid.org/browse/LETS-426 which describes this)

I added a comment on existing http://jira.cubrid.org/browse/LETS-426 that pings back here.

src/communication/request_sync_send_queue.hpp
```diff
@@ -296,6 +298,18 @@ namespace cubcomm
     assert (backbuffer.empty ());
   }
 
+  template <typename ReqClient, typename ReqPayload>
+  void
+  request_sync_send_queue<ReqClient, ReqPayload>::wait_until_empty ()
```
cristiarg, line 303: I think this function is incorrect as implemented here.
By looking at the implementation of `wait_not_empty_and_send_all` - specifically `m_request_queue.swap` - it does not guarantee that the return value is correct.
It would only work if there was a `m_shutdown` flag to instruct the `push` function to not accept requests anymore.
Or if you can guarantee that no more messages are posted to the send infrastructure.
Same remark for `request_sync_client_server::wait_until_all_requests_sent`.

cristiarg (commented):
@wrlawodms post-merge review; take everything with a grain of salt :)

I did not look into the unit tests.

src/communication/request_sync_send_queue.hpp
wrlawodms: Agree.. It's sufficient for the use case in a unit test, but you're right. At least it needs a comment for this.

PR #4603, opened 2023-08-23
[CBRD-24941] Fixing phase-0 PL/CSQL compiler small defects

hyunikn:
http://jira.cubrid.org/browse/CBRD-24941

fixing small defects such as

- Java code generation for floating point literal will use BigDecimal constructor with a string argument rather than a floating point literal argument, because the latter causes imprecise value due to the imprecise double type of the literal.
- Integer literal max value will be fixed to 2147483647 (it is now incorrect 2147483648).
- Cursor declaration and close and fetch statement will include "import java.sql.*; ", which was erroneously omitted.
- Integer remainder operator '%', which was mistakenly omitted, will be added.
- List of length zero, instead of Null, will be used for zero length arguments function calls : Null caused Null Pointer Exception for some cases
- Unused lexer definitions for floating point number literals will be removed
- Boolean and SYS_REFCURSOR types will be banned in parameter types and function return types as the specification says.


hgryoo (approved):
LGTM!

PR #4604, opened 2023-08-23
[CBRD-24921] Introducing new API for creating system catalogs

hgryoo:

- Extract routines related to creating system catalogs
   - schema_system_catalog_install.hpp has three exposed APIs to initialize and to install system catalogs
```
// schema_system_catalog_install.hpp
extern EXPORT_IMPORT void catcls_init (void);
extern EXPORT_IMPORT int catcls_install_class (void);
extern EXPORT_IMPORT int catcls_install_vclass (void);
```
- Separate the system catalog's definition and logic to create
  - See schema_system_catalog_definition.hpp and schema_system_catalog_builder.hpp respectively

Proposed API:

**AS-IS:**
```
  def = smt_edit_class_mop (class_mop, AU_ALTER);

  /* add attributes */
  error_code = smt_add_attribute (def, "class_of", CT_CLASS_NAME, NULL);
  if (error_code != NO_ERROR)
    {
      return error_code;
    }

  error_code = smt_add_attribute (def, "attr_name", "varchar(255)", NULL);
  if (error_code != NO_ERROR)
    {
      return error_code;
    }

...

  /* add constraints */
  error_code = db_add_constraint (class_mop, DB_CONSTRAINT_INDEX, NULL, index_col_names, 0);
  if (error_code != NO_ERROR)
    {
      return error_code;
    }

  if (locator_has_heap (class_mop) == NULL)
    {
      return er_errid ();
    }

  /* add authorizations */
  error_code = au_change_owner (class_mop, Au_dba_user);
  if (error_code != NO_ERROR)
    {
      return error_code;
    }
```

**TO-BE:**
System catalog's definition
```
system_catalog_definition catalog_definition (
// name
    "new_catalog",
// columns
    {
      {"class_of", OBJECT},
      {"unique_name", VARCHAR_255},
      {"new_name", VARCHAR_255},
      {"type", INTEGER},
      {"is_system_class", INTEGER},
      {"owner", AU_USER_CLASS_NAME},
    },
    {
      {DB_CONSTRAINT_INDEX, "i__db_new_unique_name", {"unique_name", nullptr}, false},
      {DB_CONSTRAINT_INDEX, "", {"new_name", "owner", nullptr}, false}
    },
// authorization
    {
      // owner, grants
      Au_dba_user, {}
    }
    );
```

Call schema_system_catalog_builder's API
```
// see catcls_install_class() and catcls_install_vclass() in schema_system_catalog_install.cpp
using catalog_builder = cubschema::system_catalog_builder;
MOP class_mop = catalog_builder::create_and_mark_system_class ("new_catalog");
error_code = catalog_builder::build_class (class_mop, catalog_definition);
```


src/object/schema_system_catalog_install.cpp
```diff
@@ -0,0 +1,5193 @@
...
+ *   class(IN) :
+ */
+static int
+boot_define_partition (MOP class_mop)
+{
+  SM_TEMPLATE *def;
+  int error_code = NO_ERROR;
+  const char *index_col_names[] = { "class_of", "pname", NULL };
+
+  def = smt_edit_class_mop (class_mop, AU_ALTER);
```
beyondykk9, line 1346: why not check if def is NULL like "boot_define_dual" fucntion

src/object/schema_system_catalog_install.cpp
```diff
@@ -0,0 +1,5193 @@
...
+ */
+static int
+boot_define_stored_procedure (MOP class_mop)
+{
+  SM_TEMPLATE *def;
+  char args_string[64];
+  int error_code = NO_ERROR;
+  const char *index_col_names[2] = { "sp_name", NULL };
+
+  def = smt_edit_class_mop (class_mop, AU_ALTER);
```
beyondykk9, line 1538: Why not check if def is NULL?

src/object/schema_system_catalog_install.cpp
```diff
@@ -0,0 +1,5193 @@
...
+ * Note:
+ *
+ */
+static int
+boot_add_charsets (MOP class_mop)
+{
+  int i;
+  int count_collations;
+
+  count_collations = lang_collation_count ();
```
beyondykk9, line 2332: it's not used in this function.

src/object/schema_system_catalog_install.cpp
```diff
@@ -0,0 +1,5193 @@
...
+
+  for (i = 0; i < num_cols; i++)
+    {
+      error_code = db_add_attribute (class_mop, columns[i].name.data (), columns[i].type.data (), NULL);
+      if (error_code != NO_ERROR)
+	{
+	  return error_code;
+	}
+    }
+
```
beyondykk9, line 3064: The stmt has 2048 length, but the query spec's string looks greater than 2048 length. Please check the length.

src/object/schema_system_catalog_install.cpp
hgryoo: I just moved the code from the legacy routines in boot_cl.c. I'm going to check in the new routine.

src/object/schema_system_catalog_install.cpp
hgryoo: ditto

src/object/schema_system_catalog_install.cpp
hgryoo: ditto

hyunikn (approved):

beyondykk9 (approved):

src/object/schema_system_catalog_definition.hpp
```diff
@@ -0,0 +1,99 @@
...
+  };
+
+  struct constraint
+  {
+    DB_CONSTRAINT_TYPE type;
+    std::string name;
+    std::vector<const char *> attribute_names;
+    bool is_class_attributes;
+
+    constraint (const DB_CONSTRAINT_TYPE &t, std::string_view n, const std::vector<const char *> &attrs,
```
ctshim, line 54: "name" is defined as type "std::string".
It is being received as a string_view type in the constructor.

hgryoo: No problem. It is copied like char*. Also this type can accept std::string type too.

ctshim (approved):

PR #4611, opened 2023-08-25
[LETS-782] Create connection handlers to catch up between PSes

wrlawodms:
http://jira.cubrid.org/browse/LETS-782

There are arguable implementation design choices. Please look into the Jira issue for these.:
- Extending the existing connection_handler to handle catchup requests vs. Creating a new separate dedicated connection_handler
- Bundling up follower and followee connection_handler or split them.

Some unrelated fixes:
- Use rvalue reference for page_server::connection_handler() and page_server::connection_handler::push_request().

Note:
- Disconnection will be addressed in the following issue.
- Defining requests will be addressed in the issues for them.
- Some descriptions and diagrams will be added when all issues related to these connection_handlers are reviewed and merged.
- `page_server::connection_handler` will be changed to `page_server::tran_server_connection_handler` after this PR is merged.

### DIFF
--- a/src/communication/request_sync_client_server.hpp
+++ b/src/communication/request_sync_client_server.hpp
@@ -157,7 +157,6 @@ namespace cubcomm
   , m_incoming_response_msgid { a_incoming_response_msgid }
   , m_response_broker { response_partition_count, NO_ERRORS, ERROR_ON_WRITE }
   {
-    assert (a_incoming_request_handlers.size () > 0);
     for (const auto &pair: a_incoming_request_handlers)
       {
 	assert (static_cast<bool> (pair.second));
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -49,7 +49,7 @@ page_server::~page_server ()
   m_async_disconnect_handler.terminate ();
 }
 
-page_server::connection_handler::connection_handler (cubcomm::channel &chn, transaction_server_type server_type,
+page_server::connection_handler::connection_handler (cubcomm::channel &&chn, transaction_server_type server_type,
     page_server &ps)
   : m_server_type { server_type }
   , m_connection_id { chn.get_channel_id () }
@@ -360,6 +360,30 @@ page_server::connection_handler::push_disconnection_request ()
   push_request (page_to_tran_request::SEND_DISCONNECT_REQUEST_MSG, std::string ());
 }
 
+page_server::follower_connection_handler::follower_connection_handler (cubcomm::channel &&chn, page_server &ps)
+  : m_ps { ps }
+{
+
+}
+
+void
+page_server::follower_connection_handler::push_request (page_to_tran_request id, std::string msg)
+{
+}
+
+int
+page_server::follower_connection_handler::send_receive (tran_to_page_request reqid, std::string &&payload_in,
+    std::string &payload_out)
+{
+  return NO_ERROR;
+}
+
+page_server::followee_connection_handler::followee_connection_handler (cubcomm::channel &&chn, page_server &ps)
+  : m_ps { ps }
+{
+
+}
+
 void page_server::pts_mvcc_tracker::init_oldest_active_mvccid (const std::string &pts_channel_id)
 {
   std::lock_guard<std::mutex> lockg { m_pts_oldest_active_mvccids_mtx };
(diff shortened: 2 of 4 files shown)


src/communication/request_sync_client_server.hpp
```diff
@@ -157,7 +157,6 @@ namespace cubcomm
   , m_incoming_response_msgid { a_incoming_response_msgid }
   , m_response_broker { response_partition_count, NO_ERRORS, ERROR_ON_WRITE }
   {
-    assert (a_incoming_request_handlers.size () > 0);
```
wrlawodms, line 160: Now, it's allowed. The follower does not have any request handler. It just requests.

src/server/tran_page_requests.hpp
```diff
@@ -56,5 +56,26 @@ enum class page_to_tran_request
...
+// requets from page server to page server to catchup
+enum class follower_to_followee_request
+{
+  // Reserve for responses
+  RESPOND,
+
+  SEND_DUMMY, // TODO for test. remove it.
+
+  // TODO SEND_DISCONNECT, /* response-less */
+  // TODO SEND_LOG_PAGES_FETCH, /* synchronously waiting for response */
```
hornetmj, line 68: Question about design approach. I think it is nature that the followee(as client) request to follower(as server) to get some log page to catch-up. However, the implementation seems to do the opposite. Is there a reason the follower is designed to send requests to the followee?

wrlawodms: I guess you got confused. The Followee is the target to be caught up with. It means it's like a server and serves log pages to the Follower.

hornetmj: OK. I resume the review.

joohok (approved):

hornetmj (approved):

src/server/page_server.hpp
```diff
@@ -209,6 +256,10 @@ class page_server
 	std::unordered_map<std::string, MVCCID> m_pts_oldest_active_mvccids;
 	std::mutex m_pts_oldest_active_mvccids_mtx;
     };
+  private:
```
cristiarg, line 259: Nit.
```suggestion
```
there is no need for `private` here; because, before and after it, there are still internal type definitions - see `private: // types` above

src/server/page_server.hpp
```diff
@@ -179,7 +179,54 @@ class page_server
...
+	void receive_dummy_request (follower_server_conn_t::sequenced_payload &&a_sp);  // TODO remove it
+
+	page_server &m_ps;
+	std::unique_ptr<follower_server_conn_t> m_conn;
+    };
+
+    class followee_connection_handler
+    {
+      public:
+	using followee_server_conn_t =
```
cristiarg, line 211: `followee_server_conn_t` can be private; same elsewhere

src/server/page_server.hpp
```diff
@@ -179,7 +179,54 @@ class page_server
...
+		cubcomm::request_sync_client_server<follower_to_followee_request, followee_to_follower_request, std::string>;
+
+	followee_connection_handler () = delete;
+	followee_connection_handler (cubcomm::channel &&chn, page_server &ps);
+
+	followee_connection_handler (const connection_handler &) = delete;
+	followee_connection_handler (connection_handler &&) = delete;
+
+	followee_connection_handler &operator= (const connection_handler &) = delete;
+	followee_connection_handler &operator= (connection_handler &&) = delete;
```
cristiarg, line 221: should have a dtor

src/server/page_server.cpp
```diff
@@ -368,6 +368,60 @@ page_server::connection_handler::push_disconnection_request ()
...
+}
+
+void
+page_server::followee_connection_handler::push_request (follower_to_followee_request reqid, std::string &&msg)
+{
+  m_conn->push (reqid, std::move (msg));
+}
+
+int
+page_server::followee_connection_handler::send_receive (follower_to_followee_request reqid, std::string &&payload_in,
```
cristiarg, line 419: nit;
maybe not justified to have this as a separate method; we'll extract one later if needed

cristiarg (commented):
@wrlawodms 
post-merge review; take everything with a grain of salt :)

src/server/page_server.hpp
wrlawodms: I thought that the default dtor is sufficient. All we need is done in the dtor of `m_conn`. Is there something we need to do explicitly in a dtor?

src/server/page_server.hpp
wrlawodms: Thanks. This will be fixed in the following PR.

src/server/page_server.hpp
wrlawodms: Thanks. This will be fixed in the following PR.

PR #4619, opened 2023-08-29
[CBRD-24973] The problem is that a private synonym can only be used by the owner, but can be used by other users who have authorization for the target

youngjinj:
http://jira.cubrid.org/browse/CBRD-24973

1. Add au_check_synonym_authorization()
   - Checks whether the current user is the owner of the synonym.
   - If the current user is not the owner of the synonym, the synonym must not be used.
   - Before calling sm_get_synonym_target_name (), au_check_synonym_authorization () must be checked.
      - pt_add_lock_class ()
      - sm_find_class_with_purpose ()
      - pt_convert_dblink_synonym ()
2. Removed do_find_synonym_by_query() as it was not being used.


youngjinj:
By Design

PR #4620, opened 2023-08-29
[CBRD-24949] CAS Protocol V12, send oracle_compat_number_behavior system parameter to the clients

airnet73:
http://jira.cubrid.org/browse/CBRD-24949

Purpose
* to eliminate user confusion, 2-way expression is unified with the oracle_compat_number_behavior system parameter.
* in order for the broker to deliver the value of the system parameters including oracle_compat_number_behavior to the client, the broker extends the broker info that the client receives using the RESERVED field.
* Change the CAS Protocol version from PROTOCOL_V11 to PROTOCOL_V12 to express the extension of the above protocol

Implementation
N/A

Remarks
N/A


kisoo-han (approved):
check code-style CI fails & build failure for **BROKER_INFO_RESERVED2** -> BROKER_INFO_SYSTEM_PARAM
cas_meta.c
cas_protocol.h

airnet73:
@kisoo-han 
Added V12 protocol in shard.

kisoo-han (approved):

PR #4627, opened 2023-08-30
[LETS-748] Add the request to get log pages to catch up with

wrlawodms:
http://jira.cubrid.org/browse/LETS-748

- `follower_responder_t`: is used to respond to the requests for the catch-up. I was not going to use this and make the catch-up communication sequential, but I've decided to use it. It is useful in that 
  - The follower can disconnect instantly while the catch-up is in progress.
  - The `thread::entry` is required to fetch log pages in followee, and workers in the `server_request_responder` has its entry.
  - I tried to integrate two `server_request_responder`s (for TS-PS and PS-PS), but I failed. Now, we have several connection_handlers and responders so we may need to refactor them. Welcome to any opinion on it. 
- Requests are sent in one direction (follower -> followee)
  - `request_log_pages()`: a follower requests multiple log pages for the catch up. 
  - DISCONNECT: TODO in the following PR.
  - `receive_log_pages_fetch()`: a follwee serves multiple log pages. It uses `serve_log_pages` internally.
- Now the log pages are requested in `receive_start_catch_up()`.
  - It will be moved in another dedicated thread in near future. We can implement some details then, but I've implemented it to test various cases to confirm the request works well. It's going to be revised more in the future. 
  - It allocates a huge buffer containing several pages (now 10 at most).
  - It requests pages in units of the buffer, ends up getting all the pages from `log_Gl.hdr.append_lsa.pageid` to `catchup_lsa.pageid`.
- I have confirmed that it can fetch multiple pages including ones in archives.

### DIFF
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -404,10 +404,6 @@ page_server::follower_connection_handler::follower_connection_handler (cubcomm::
   m_conn.reset (new follower_server_conn_t (std::move (chn),
   {
     {
-      follower_to_followee_request::SEND_DUMMY,
-      std::bind (&page_server::follower_connection_handler::receive_dummy_request, std::ref (*this), std::placeholders::_1)
-    },
-    {
       follower_to_followee_request::SEND_LOG_PAGES_FETCH,
       std::bind (&page_server::follower_connection_handler::receive_log_pages_fetch, std::ref (*this), std::placeholders::_1)
     },
@@ -801,9 +797,6 @@ page_server::connect_to_followee_page_server (std::string &&hostname, int32_t po
   er_log_debug (ARG_FILE_LINE,
 		"This page server successfully connected to the followee page server to catch up. Channel id: %s.\n",
 		srv_chn.get_channel_id ().c_str ());
-
-  // TODO remove it. Just a test to make sure the connection works.
-  m_followee_conn->push_request (follower_to_followee_request::SEND_DUMMY, std::string (""));
 
   return NO_ERROR;
 }
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -419,13 +419,6 @@ page_server::follower_connection_handler::follower_connection_handler (cubcomm::
 }
 
 void
-page_server::follower_connection_handler::receive_dummy_request (follower_server_conn_t::sequenced_payload
-    &&a_sp)
-{
-  er_log_debug (ARG_FILE_LINE, "A follower has requested a duumy request");
-}
-
-void
 page_server::follower_connection_handler::receive_log_pages_fetch (follower_server_conn_t::sequenced_payload &&a_sp)
 {
   auto log_serving_func = std::bind (&page_server::follower_connection_handler::serve_log_pages, std::ref (*this),
@@ -437,13 +430,13 @@ void
 page_server::follower_connection_handler::serve_log_pages (THREAD_ENTRY &, std::string &payload_in_out)
 {
   // Unpack the message data
-  cubpacking::unpacker unpacker {payload_in_out.c_str (), payload_in_out.size ()};
+  cubpacking::unpacker payload_unpacker { payload_in_out.c_str (), payload_in_out.size () };
   LOG_PAGEID start_pageid;
   int cnt;
 
   assert (payload_in_out.size () == (sizeof (LOG_PAGEID) + OR_INT_SIZE));
-  unpacker.unpack_bigint (start_pageid);
-  unpacker.unpack_int (cnt);
+  payload_unpacker.unpack_bigint (start_pageid);
+  payload_unpacker.unpack_int (cnt);
 
   const bool perform_logging = prm_get_bool_value (PRM_ID_ER_LOG_READ_LOG_PAGE);
 
@@ -534,19 +527,19 @@ page_server::followee_connection_handler::request_log_pages (LOG_PAGEID start_pa
 {
   int error_code = NO_ERROR;
   std::string response_message;
-  cubpacking::packer packer;
+  cubpacking::packer payload_packer;
   size_t size = 0;
 
   const bool perform_logging = prm_get_bool_value (PRM_ID_ER_LOG_READ_LOG_PAGE);
 
-  size += packer.get_packed_bigint_size (size); // start_pageid
-  size += packer.get_packed_int_size (size); // count
+  size += payload_packer.get_packed_bigint_size (size); // start_pageid
+  size += payload_packer.get_packed_int_size (size); // count
 
   std::unique_ptr < char[] > buffer (new char[size]);
-  packer.set_buffer (buffer.get (), size);
+  payload_packer.set_buffer (buffer.get (), size);
 
-  packer.pack_bigint (start_pageid);
-  packer.pack_int (count);
+  payload_packer.pack_bigint (start_pageid);
+  payload_packer.pack_int (count);
 
   if (perform_logging)
     {
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -395,6 +395,8 @@ page_server::follower_connection_handler::follower_connection_handler (cubcomm::
   RESPONSE_PARTITIONING_SIZE,
   nullptr)); // TODO handle abnormal disconnection.
 
+  m_ps.get_follower_responder ().register_connection (m_conn.get ());
+
   m_conn->start ();
 }
 
@@ -458,6 +460,19 @@ page_server::follower_connection_handler::push_async_response (F &&a_func,
   auto handler_func = std::bind (std::forward<F> (a_func), std::placeholders::_1, std::placeholders::_2,
 				 std::forward<Args> (args)...);
   m_ps.get_responder ().async_execute (std::ref (*m_conn), std::move (a_sp), std::move (handler_func));
+}
+
+page_server::follower_connection_handler::~follower_connection_handler ()
+{
+  // blocking call
+  // internally, this will also wait pending outgoing roundtrip (send-receive) requests
+  m_conn->stop_incoming_communication_thread ();
+
+  // blocking call
+  // wait async responder to finish processing in-flight incoming roundtrip requests
+  m_ps.get_follower_responder ().wait_connection_to_become_idle (m_conn.get ());
+
+  m_conn->stop_outgoing_communication_thread ();
 }
 
 page_server::followee_connection_handler::followee_connection_handler (cubcomm::channel &&chn, page_server &ps)
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -49,7 +49,7 @@ page_server::~page_server ()
   m_async_disconnect_handler.terminate ();
 }
 
-page_server::connection_handler::connection_handler (cubcomm::channel &chn, transaction_server_type server_type,
+page_server::connection_handler::connection_handler (cubcomm::channel &&chn, transaction_server_type server_type,
     page_server &ps)
   : m_server_type { server_type }
   , m_connection_id { chn.get_channel_id () }
@@ -360,6 +360,30 @@ page_server::connection_handler::push_disconnection_request ()
   push_request (page_to_tran_request::SEND_DISCONNECT_REQUEST_MSG, std::string ());
 }
 
+page_server::follower_connection_handler::follower_connection_handler (cubcomm::channel &&chn, page_server &ps)
+  : m_ps { ps }
+{
+
+}
+
+void
+page_server::follower_connection_handler::push_request (page_to_tran_request id, std::string msg)
+{
+}
+
+int
+page_server::follower_connection_handler::send_receive (tran_to_page_request reqid, std::string &&payload_in,
+    std::string &payload_out)
+{
+  return NO_ERROR;
+}
+
+page_server::followee_connection_handler::followee_connection_handler (cubcomm::channel &&chn, page_server &ps)
+  : m_ps { ps }
+{
+
+}
+
 void page_server::pts_mvcc_tracker::init_oldest_active_mvccid (const std::string &pts_channel_id)
 {
   std::lock_guard<std::mutex> lockg { m_pts_oldest_active_mvccids_mtx };
@@ -465,7 +489,7 @@ page_server::set_active_tran_server_connection (cubcomm::channel &&chn)
       disconnect_active_tran_server ();
     }
 
-  m_active_tran_server_conn.reset (new connection_handler (chn, transaction_server_type::ACTIVE, *this));
+  m_active_tran_server_conn.reset (new connection_handler (std::move (chn), transaction_server_type::ACTIVE, *this));
 }
 
 void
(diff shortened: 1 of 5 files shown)


src/server/server_request_responder.hpp
```diff
@@ -163,7 +163,8 @@ template<typename T_CONN>
 server_request_responder<T_CONN>::server_request_responder ()
   : m_threads_context_manager (std::make_unique<cubthread::system_worker_entry_manager> (TT_SYSTEM_WORKER))
 {
-  const auto THREAD_COUNT = std::thread::hardware_concurrency ();
+  const auto THREAD_COUNT = 4; // std::thread::hardware_concurrency ();
```
wrlawodms, line 166: Without this change, It caused failure to start server on my machine when I use two of server_request_responder. One is the existing one from PS to TS, and another is what I've added for PS->PS. However, even without adding the new one, it would cause the failure in a many-core machine. Please look into the issue I created : http://jira.cubrid.org/browse/LETS-788.

Welcome to opinions on this.

hornetmj (approved):

src/server/page_server.cpp
```diff
@@ -204,18 +204,79 @@ page_server::connection_handler::receive_start_catch_up (tran_server_conn_t::seq
...
+    const LOG_PAGEID end_pageid = catchup_lsa.pageid;
+    const size_t total_page_count = end_pageid - start_pageid + 1;
+
+    _er_log_debug (ARG_FILE_LINE, "[CATCH-UP] The catch-up pages range from %lld to %lld, count = %lld.\n",
+		   start_pageid, end_pageid, total_page_count);
+
+    const int page_cnt_in_buffer = (int) std::min (total_page_count, (size_t) MAX_PAGE_REQUEST_CNT_AT_ONCE);
+    const size_t buffer_size_in_total = page_cnt_in_buffer * LOG_PAGESIZE + MAX_ALIGNMENT;
+
+    auto log_pgbuf_buffer = std::unique_ptr<char []> { new char[buffer_size_in_total] };
```
joohok, line 240: I think we can get aligned buffer when using std::unique_ptr. memory allocated in the heap is typically aligned to 8 bytes

src/server/page_server.cpp
```diff
@@ -386,13 +447,78 @@ page_server::follower_connection_handler::follower_connection_handler (cubcomm::
...
+
+  // Pack the requested log pages
+  int error = NO_ERROR;
+  payload_in_out = { reinterpret_cast<const char *> (&error), sizeof (error) }; // Pack NO_ERROR assuming it'll succeed.
+  log_reader lr { LOG_CS_SAFE_READER };
+
+  for (int i = 0; i < cnt; i++)
+    {
+      const LOG_PAGEID log_pageid = start_pageid + i;
+      const log_lsa fetch_lsa { log_pageid, 0 };
```
```suggestion
      const log_lsa fetch_lsa { start_pageid + i, 0 };
```
`log_pageid` is only used for setting `fetch_lsa`, so I think we don't need to separate it.

src/server/page_server.cpp
wrlawodms: Oh, I didn't know that. Could you give a document or reference about that?

src/server/page_server.cpp
wrlawodms: You're right, thanks.

src/server/page_server.cpp
joohok: Sorry! I used to believe that when allocating heap memory using `malloc` or the `new` operator, memory addresses were aligned to 8-byte boundaries for optimization. However, I've learned that this behavior can vary depending on the compiler or platform, and the C++ language itself does not guarantee alignment. I apologize for any confusion.

ref : https://learn.microsoft.com/en-us/cpp/cpp/align-cpp?view=msvc-170

joohok (approved):

src/server/page_server.cpp
```diff
@@ -204,18 +204,80 @@ page_server::connection_handler::receive_start_catch_up (tran_server_conn_t::seq
...
-  m_ps.connect_to_followee_page_server (std::move (host), port);
+  // All the below here will be taken up.
+  {
+    m_ps.connect_to_followee_page_server (std::move (host), port);
+
+    assert (!log_Gl.hdr.append_lsa.is_null ());
+    assert (!catchup_lsa.is_null ());
+    assert (catchup_lsa >= log_Gl.hdr.append_lsa);
+
+    if (log_Gl.hdr.append_lsa == catchup_lsa)
```
cristiarg, line 223: @wrlawodms (post-merge comment)
Actually this decision can be done before connecting to the followee page server.

From an ATS's perspective there are 3 functionally different (2 structurally different start-up cases):
1. upon cold-start all PS's are at the same level (what is now assumed)
2. upon cold-start, a valid quorum of PS's are at the same level, but some of them PS's are behind with their log (think as if like ATS and all PSs would concurrently and non-deterministically be stopped at the same time)
  - in this case the ATS would instruct the left behind PS's to catch-up on a leader PS
  - this case is functionally different than 1. but can structurally be implemented the same as 1. (ie: 1. is a particular case of this)
3. hot-start of a PS into a running cluster

In light of this explanation, the `if (log_Gl.hdr.append_lsa == catchup_lsa)` condition above fits for situation 2) where the other up-to-date PSs that are equal to the leader PS are sent the catch-up command.

src/server/page_server.cpp
```diff
@@ -204,18 +204,80 @@ page_server::connection_handler::receive_start_catch_up (tran_server_conn_t::seq
...
+      {
+	_er_log_debug (ARG_FILE_LINE, "[CATCH-UP] There is nothing to catch up.\n");
+	return; // TODO the cold-start case. No need to catch up.
+      }
+
+    constexpr int MAX_PAGE_REQUEST_CNT_AT_ONCE = 10; // Arbitrarily chosen
+    const LOG_PAGEID start_pageid = log_Gl.hdr.append_lsa.pageid;
+    const LOG_PAGEID end_pageid = catchup_lsa.offset == 0 ? catchup_lsa.pageid - 1 : catchup_lsa.pageid;
```
cristiarg, line 231: @wrlawodms (post-merge comment)

`catchup_lsa.offset == 0`
please explain
is it because, if offset is 0 this means we also need the previous page to - possibly link wit the subsequent one?

src/server/page_server.cpp
cristiarg, line 223: Actually this decision can be done without connecting to the followee page server.

From an ATS's perspective there are 3 functionally different (2 structurally different start-up cases):
1. upon cold-start all PS's are at the same level (what is now assumed)
2. upon cold-start, a valid quorum of PS's are at the same level, but some of them PS's are behind with their log (think as if like ATS and all PSs would concurrently and non-deterministically be stopped at the same time)
  - in this case the ATS would instruct the left behind PS's to catch-up on a leader PS
  - this case is functionally different than 1. but can structurally be implemented the same as 1. (ie: 1. is a particular case of this)
3. hot-start of a PS into a running cluster

In light of this explanation, the `if (log_Gl.hdr.append_lsa == catchup_lsa)` condition above fits for situation 2) where the other up-to-date PSs that are equal to the leader PS are sent the catch-up command.

src/server/page_server.cpp
cristiarg, line 231: `catchup_lsa.offset == 0`
please explain
is it because, if offset is 0, this means we also need the previous page to - possibly link wit the subsequent one?
maybe, add a comment

src/server/tran_page_requests.hpp
```diff
@@ -56,19 +56,17 @@ enum class page_to_tran_request
   SEND_TO_PTS_LOG_PRIOR_LIST, /* response-less */
 };
 
-// requets from page server to page server to catchup
+// requests from page server to page server to catchup
 enum class follower_to_followee_request
```
cristiarg, line 60: Nit
`ps_follower_to_ps_followed_request`
also, my browser spell checker says that `followee` is incorrect and that `followed` is correct

src/server/tran_page_requests.hpp
```diff
@@ -56,19 +56,17 @@ enum class page_to_tran_request
...
-  SEND_DUMMY, // TODO for test. remove it.
-
   // TODO SEND_DISCONNECT, /* response-less */
-  // TODO SEND_LOG_PAGES_FETCH, /* synchronously waiting for response */
+  SEND_LOG_PAGES_FETCH, /* synchronously waiting for response */
 };
 
-// requets from page server to page server to catchup
+// requests from page server to page server to catchup
 enum class followee_to_follower_request
```
cristiarg, line 70: Nit
`ps_followed_to_ps_follower_request`

src/server/page_server.hpp
```diff
@@ -198,8 +198,12 @@ class page_server
 	follower_connection_handler &operator= (const connection_handler &) = delete;
 	follower_connection_handler &operator= (connection_handler &&) = delete;
 
+	~follower_connection_handler ();
```
cristiarg, line 201: Nit, I find these names quite confusing:
- `follower_connection_handler` -> `from_follower_connection_handler` (ie: connection from the followerr) or `as_followed_connection_handler` (ie: "acting as followed")
- `followee_connection_handler` -> `to_followed_connection_handler` (ie: connection to the followed) or 'as_follower_connection_handler` (ie: "acting as follower")

Whichever your prefer, or even some other naming scheme which is more explanatory than the current.

src/server/page_server.cpp
```diff
@@ -421,7 +547,91 @@ int
...
+      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_CONN_PAGE_SERVER_CANNOT_BE_REACHED, 0);
+      return ER_CONN_PAGE_SERVER_CANNOT_BE_REACHED;
+    }
+
+  return NO_ERROR;
+}
+
+int
+page_server::followee_connection_handler::request_log_pages (LOG_PAGEID start_pageid, int count,
+    std::vector<LOG_PAGE *> &log_pages_out)
```
cristiarg, line 562: `const std::vector<LOG_PAGE *> &log_pages_out`

src/server/page_server.cpp
```diff
@@ -204,18 +204,80 @@ page_server::connection_handler::receive_start_catch_up (tran_server_conn_t::seq
...
+    LOG_PAGEID request_start_pageid = start_pageid;
+    size_t remaining_page_cnt = total_page_count;
+    while (remaining_page_cnt > 0)
+      {
+	const int request_page_cnt = (int) std::min ((size_t) page_cnt_in_buffer, remaining_page_cnt);
+	// TODO request next log pages asynchronously with std::async
+	const int error_code = m_ps.m_followee_conn->request_log_pages (request_start_pageid, request_page_cnt, log_pgptr_vec);
+	if (error_code != NO_ERROR)
+	  {
+	    /* There are two kinds of erorrs either from client-side or server-side.
```
cristiarg, line 259: Nit. There are some typos in comments.

src/server/page_server.cpp
```diff
@@ -386,13 +448,77 @@ page_server::follower_connection_handler::follower_connection_handler (cubcomm::
...
+
+  if (perform_logging)
+    {
+      _er_log_debug (ARG_FILE_LINE,
+		     "[READ LOG] Sending log pages to the pager server (%s). Page Id from %lld to %lld, Error code: %d\n",
+		     m_conn->get_underlying_channel_id ().c_str (), start_pageid, start_pageid + cnt - 1, error);
+    }
+}
+
+page_server::follower_connection_handler::~follower_connection_handler ()
```
cristiarg, line 511: Instances of both `follower_connection_handler` and `followee_connection_handler` must be removed as soon as they are done.
Because underneath, they keep resources open: sockets, threads, ....

If there is no follow-up issue that deals with this, please add. If there is, my bad, please excuse. :)

src/server/page_server.cpp
```diff
@@ -386,13 +448,77 @@ page_server::follower_connection_handler::follower_connection_handler (cubcomm::
...
+      const log_lsa fetch_lsa { start_pageid + i, 0 };
+
+      assert (fetch_lsa.pageid != LOGPB_HEADER_PAGE_ID);
+
+      error = lr.set_lsa_and_fetch_page (fetch_lsa);
+
+      if (error != NO_ERROR)
+	{
+	  // All or Nothing. Abandon all appended pages and just set the error.
+	  payload_in_out = { reinterpret_cast<const char *> (&error), sizeof (error) };
```
cristiarg, line 494: Log error.

cristiarg (commented):
@wrlawodms post-merge review; take everything with a grain of salt :)

src/server/page_server.cpp
wrlawodms: > Actually this decision can be done without connecting to the followee page server.

Right. I noticed it after this's merged and  It has already been fixed in #4663. :)

> upon cold-start, a valid quorum of PS's are at the same level, but some of them PS's are behind with their log (think as if like ATS and all PSs would concurrently and non-deterministically be stopped at the same time)

Honestly, I don't think I fully understand what you're stating. In this case, the ATS will reboot with the quorum of PS, and they can be leader. A PS left behind will receive the catch-up request after the ATS boots up and do catch up with one of the PSes which attends the ATS's booting. The left-behind PS does "cold-start" and the PSs which attend the ATS' boot do the "hot-start"

src/server/page_server.cpp
wrlawodms: The `catchup_lsa` is the first LSA of log records the ATS will send. It means the followee PS may not have the log record. If the offset is 0, the followee may have not appended or allocated the page. In this case, the followee can't serve the page of `catchup_lsa.pageid`. 
 Basically, a follower has to pull log pages **UNTIL** `catchup_lsa`. The last page is catchup_lsa.pageid if the offset is not 0, and catchup_lsa.pageid -1 if the offset is 0.

src/server/tran_page_requests.hpp
wrlawodms: My checker does too, but followee is used quite broadly and described in Wikis. I guess it's used in Instagram..

src/server/tran_page_requests.hpp
wrlawodms: followee is not followed. It's the counter part of follower.

src/server/page_server.cpp
wrlawodms: Thank you. I'll fix them in the following PR.

src/server/page_server.cpp
wrlawodms: They will removed in follow-up issues. :)

src/server/page_server.cpp
wrlawodms: Thank you. I'll fix them in the following PR.

src/server/page_server.cpp
wrlawodms: Hmm.. I don't think we should set an error here since it's set under `lr.set_lsa_and_fetch_page`. All we need here is just throw the error to the client.

src/server/page_server.hpp
wrlawodms: I agree the naming is somewhat confusing and actually struggled to come up with a proper one. It seems that your suggestion is better. I'll create an issue to modify the name. In the issue, I'll modify the names to yours or something more explanatory.

PR #4628, opened 2023-08-30
[CBRD-24831] Type checking and code generation of CUBRID built-in functions

hyunikn:
http://jira.cubrid.org/browse/CBRD-24831

- All the names of built-in functions are registered in the 0 level of the symbol table stack
  - For a function call expression, if the function name matches one of the registered built-in function names, then the expression is deemed to be a built-in function call.
- For each supported type of PL/CSQL, "typical value string" is defined as follows:
  - string : cast(? as string)
  - numeric: cast(? as numeric)
  - short: cast(? as short)
  - int: cast(? as int)
  - bigint: cast(? as bigint)
  - float: cast(? as float)
  - double: cast(? as double)
  - date: cast(? as date)
  - time: cast(? as time)
  - datetime: cast(? as datetime)
  - timestamp: cast(? as timestamp)
  - boolean: X
  - sys_refcursor: X 
- For a built-in function call expression, a select statement text is made and sent to the server to get the type of the function call or an error
  - For example, a built-in function call expression "nvl2(3, 2.7, 'abc') " is converted to "select nvl2(cast(? as int), cast(? as numeric), cast(? as string))" and sent to the server
  - Type checking a built-in function call succeeds only if the result type from the server is one of the supported PL/CSQL types except for boolean and sys_refcursor.
- For a valid built-in function call expression, Java code is generated so that the select statement is executed in the server and the result is returned using the JDBC APIs.


hgryoo (approved):
LGTM!

hgryoo:
```
a built-in function call expression "nvl2(3, 2.7, 'abc') " is converted to "select nvl2(cast(? as int), cast(? as numeric), cast(? as string))"
```
This idea seems a really good way to know the result type of a built-in function. Let's try.

hyunikn:
> ```
> a built-in function call expression "nvl2(3, 2.7, 'abc') " is converted to "select nvl2(cast(? as int), cast(? as numeric), cast(? as string))"
> ```
> 
> This idea seems a really good way to know the result type of a built-in function. Let's try.

Thank you. As you might expect, it is to avoid constant folding while type checking which can result in omitting some necessary typing steps.

PR #4638, opened 2023-08-31
[CBRD-24946] field lookup failure when column name is of <table>.<column> form in the select list

hyunikn:
http://jira.cubrid.org/browse/CBRD-24946

* In such cases, we need to detach <table name alias>. from the beginning of the column name given by the server.

### DIFF
--- a/pl_engine/pl_server/src/main/java/com/cubrid/plcsql/compiler/ParseTreeConverter.java
+++ b/pl_engine/pl_server/src/main/java/com/cubrid/plcsql/compiler/ParseTreeConverter.java
@@ -2417,8 +2417,11 @@ public class ParseTreeConverter extends PcsParserBaseVisitor<AstNode> {
         assert colName != null;
 
         int dotIdx = colName.lastIndexOf(".");
-        if (dotIdx > 0 && dotIdx < colName.length() - 1 && ci.className != null && ci.className.length() > 0 &&
-            ci.attrName != null) {
+        if (dotIdx > 0
+                && dotIdx < colName.length() - 1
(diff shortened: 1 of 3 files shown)


pl_engine/pl_server/src/main/java/com/cubrid/plcsql/compiler/ParseTreeConverter.java
```diff
@@ -2412,6 +2412,28 @@ private String getSqlTypeNameFromCode(int sqlTypeCode) {
...
+        int dotIdx = colName.lastIndexOf(".");
+        if (dotIdx > 0
+                && dotIdx < colName.length() - 1
+                && ci.className != null
+                && ci.className.length() > 0
+                && ci.attrName != null) {
+
+            String afterDot = colName.substring(dotIdx + 1).toLowerCase();
+            String attrName = ci.attrName.toLowerCase();
+            if (afterDot.equals(attrName)) {
```
hgryoo:
```suggestion
            String afterDot = colName.substring(dotIdx + 1);
            if (afterDot.equalsIgnoreCase(ci.attrName)) {
```

equalsIgnoreCase is way faster because it doesn't copy objects.

hgryoo (approved):

pl_engine/pl_server/src/main/java/com/cubrid/plcsql/compiler/ParseTreeConverter.java
hyunikn: Thank you

PR #4646, opened 2023-09-04
[CBRD-24947] OUT parameters must be initialized to NULL

hyunikn:
http://jira.cubrid.org/browse/CBRD-24947


pl_engine/pl_server/src/main/java/com/cubrid/plcsql/builtin/DBMS_OUTPUT.java
```diff
@@ -82,7 +82,7 @@ public static void getLines(String[] line, int[] cnt) {
 
     public static void putLine(String line) {
         Context c = getContext();
-        c.getMessageBuffer().putLine(line);
+        c.getMessageBuffer().putLine(line == null ? "" : line);
```
hgryoo: By the putLine() specification, any data is not appended to the buffer if the line is null.

hgryoo (approved):

PR #4648, opened 2023-09-04
[CBRD-24623] Changed so that the file can be found when specifying the file path of schema_info file in the --schema-file-list option.

airnet73:
http://jira.cubrid.org/browse/CBRD-24623

Purpose
If you specify the path to demodb_schema_info in loaddb as shown below, the list of files in the demodb_schema_info file cannot be found. Modified so that it can be found in ./testdb/unload/ when looking for the file list.
--schema-file-list=./testdb/unload/demodb_schema_info

Implementation
N/A

Remarks
N/A

### DIFF
--- a/src/loaddb/load_db.c
+++ b/src/loaddb/load_db.c
@@ -1406,7 +1406,7 @@ ldr_check_file_list (std::string & file_name, int &num_files, int &error_code)
   char buffer[PATH_MAX] = { 0, };
   std::string read_file_name = "";
   std::string schema_info_fullpath = "";
-  size_t last_backslash = 0;
+  size_t seperator_pos = 0;
 #if defined(WINDOWS)
   size_t last_slash = 0;
 #endif
@@ -1470,11 +1470,11 @@ ldr_check_file_list (std::string & file_name, int &num_files, int &error_code)
 
       strcpy (schema_object_file->schema_file_name, buffer);
 
-      last_backslash = schema_info_fullpath.find_last_of (PATH_SEPARATOR);
+      seperator_pos = schema_info_fullpath.find_last_of (PATH_SEPARATOR);
 
-      if (last_backslash == std::string::npos)
+      if (seperator_pos == std::string::npos)
 	{
-	  last_backslash = 0;
+	  seperator_pos = 0;
 	}
 
 #if defined(WINDOWS)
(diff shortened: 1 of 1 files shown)


kisoo-han:
I have a question, if a user run loaddb in windows like
C> cubrid loaddb --schema-file-list=.C:/testdb/unload/demodb_schema_info testdb
Is it okay?

src/loaddb/load_db.c
```diff
@@ -1463,7 +1466,18 @@ ldr_check_file_list (std::string & file_name, int &num_files, int &error_code)
       num_files++;
 
       strcpy (schema_object_file->schema_file_name, buffer);
-      read_file_name = buffer;
+
+      last_slash = schema_info_fullpath.find_last_of (PATH_SEPARATOR);
```
kisoo-han: I have a question, if a user run loaddb in windows like
C> cubrid loaddb --schema-file-list=.C:/testdb/unload/demodb_schema_info testdb
Is it okay?

airnet73: Yes, no problem.

kisoo-han: I think you have check both '/' and '\' in Windows or use library for Linux and WIndows respectively.
```
dirname (path)
_splitpath_s(schema_file_list, drive, _MAX_DRIVE, dir, PATH_MAX, NULL, 0, NULL, 0) 
```

src/loaddb/load_db.c
```diff
@@ -1463,7 +1469,48 @@ ldr_check_file_list (std::string & file_name, int &num_files, int &error_code)
...
+	{
+	  schema_info_fullpath = schema_info_fullpath.substr (0, last_backslash + 1);
+	  read_file_name = schema_info_fullpath + buffer;
+	}
+      else
+	{
+	  read_file_name = buffer;
+	}
+#endif
+
```
kisoo-han, line 1503: Maybe modified code itself could working without problem. I propose minor change, just refer and do it as you like.
1. Change variable 'last_backslash' to 'seperator_pos'
2. change modified line#: 1472 ~ 1513 to the code below:
```
      seperator_pos = schema_info_fullpath.find_last_of (PATH_SEPARATOR);

      if (seperator_pos == std::string::npos)
        {
          seperator_pos = 0;
        }

#if defined(WINDOWS)
      last_slash = schema_info_fullpath.find_last_of ('/');

      if (last_slash == std::string::npos)
        {
          last_slash = 0;
        }

      if (last_slash > seperator_pos)
        {
          seperator_pos = last_slash;
        }
#endif

      if (seperator_pos > 0)
        {
          schema_info_fullpath = schema_info_fullpath.substr (0, seperator_pos + 1);
          read_file_name = schema_info_fullpath + buffer;
        }
      else
        {
          read_file_name = buffer;
        }

```

kisoo-han (approved):

src/loaddb/load_db.c
airnet73: I will modify it as per your suggestion.

PR #4658, opened 2023-09-06
[CBRD-24663] Outputs data values of MIDXKEY type in csql

ctshim:
http://jira.cubrid.org/browse/CBRD-24663

* Enables csql to output MIDXKEY type. 
  -Support the ";MIDXKEY ON|OFF" syntax with the csql tool command.
  -The added ";MIDXKEY ON|OFF" statement should not be exposed to ";Help".

* Supports casting of MIDXKEY type to varchar(char).


hyunikn (approved):

src/executables/csql_result_format.c
```diff
@@ -1938,6 +1958,27 @@ csql_db_value_as_string (DB_VALUE * value, int *length, bool plain_string, CSQL_
...
+	  {
+	    string_buffer sb;
+
+	    sb.clear ();
+	    db_sprint_value (value, sb);
+
+	    /*  Make sure it has the same format as the result of using cast().
+	     *  '{1, 'abc'}'  or  '{1, ''abc''}'  (depends on change_single_quote)
+	     */
+	    result =
```
beyondykk9, line 1973: doesn't it needs to check if it's null?

ctshim: What specifically do you think requires a NULL check?
There is no special NULL check anywhere else in this function with respect to the result variable.

beyondykk9 (approved):

hgryoo (approved):

PR #4662, opened 2023-09-07
[CBRD-24969] Remove semicolon after shard key

kisoo-han:
http://jira.cubrid.org/browse/CBRD-24969

**Description**
* remove additional semicolon after shard key.

**Remarks**
* **sp_make_int_sp_value_from_string** () expect integer number style shard key and it convert the string type key to bigint.
* In the case of a token with a shard key and followed by semicolon without space, the shard parser recognizes it as a shard key, including the semicolon. For example,

  [SHARD KEY USED IN Query] => [SHARD KEY TO CONVERT TO INTEGER]
  * /*+ shard_key */ 1 => '1' ==> parse_bigint (&new_val, "1", 10);
  * /*+ shard_key */ 1 ; => '1'
  * /*+ shard_key */ 1   ; => '1'
  * /*+ shard_key */ 1abc   ; => '1abc' ==> parse_bigint (&new_val, "1abc", 10); =====> raise error
  * /*+ shard_key */ **1;** => '**1;**' **==>** '1' ===>  parse_bigint (&new_val, "1", 10);
* Since the shard parser is a simple code based on string analysis, **if a semicolon is followed by the shard key, exclude the semicolon** and convert it to an integer.

### DIFF
--- a/src/broker/shard_parser.c
+++ b/src/broker/shard_parser.c
@@ -316,15 +316,14 @@ sp_make_int_sp_value_from_string (SP_VALUE * value_p, char *pos, int length)
 {
   int result = 0;
   char tmp = pos[length];
+  char *p;
 
-  if (length > 1 && pos[length - 1] == ';')
+  p = strchr (pos, ';');
+  if (p)
     {
-      pos[length - 1] = '\0';
+      *p = '\0';
     }
-  else
-    {
-      pos[length] = '\0';
-    }
+  pos[length] = '\0';
 
   result = parse_bigint (&value_p->integer, pos, 10);
   if (result != 0)
--- a/src/broker/shard_parser.c
+++ b/src/broker/shard_parser.c
@@ -316,7 +316,7 @@ sp_make_int_sp_value_from_string (SP_VALUE * value_p, char *pos, int length)
 {
   int result = 0;
   const int KEY_LENGTH = 128;
-  char shard_key [KEY_LENGTH];
+  char shard_key[KEY_LENGTH];
   char *p;
 
   snprintf (shard_key, KEY_LENGTH, "%s", pos);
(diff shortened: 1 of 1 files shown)


src/broker/shard_parser.c
```diff
@@ -318,7 +317,15 @@ sp_make_int_sp_value_from_string (SP_VALUE * value_p, char *pos, int length)
   int result = 0;
   char tmp = pos[length];
 
-  pos[length] = '\0';
+  if (length > 1 && pos[length - 1] == ';')
+    {
+      pos[length - 1] = '\0';
+    }
+  else
```
mhoh3963: How about checking which length < 0 for preventing error ?

src/broker/shard_parser.c
```diff
@@ -318,7 +317,15 @@ sp_make_int_sp_value_from_string (SP_VALUE * value_p, char *pos, int length)
   int result = 0;
   char tmp = pos[length];
 
-  pos[length] = '\0';
+  if (length > 1 && pos[length - 1] == ';')
```
mhoh3963: I think you need to check for cases where semicolons occur multiple times.
example)
/*+ shard_key */ 1;;

kisoo-han: You right, multiple semicolons in shard key should be cleared.
Thank you.

src/broker/shard_parser.c
kisoo-han: Thank you for your comment. I'll handle it together.

src/broker/shard_parser.c
```diff
@@ -317,8 +316,15 @@ sp_make_int_sp_value_from_string (SP_VALUE * value_p, char *pos, int length)
 {
   int result = 0;
   char tmp = pos[length];
+  char *p;
 
+  p = strchr (pos, ';');
+  if (p)
+    {
+      *p = '\0';
```
mhoh3963, line 326: I think that you need to save p's value and recovery it, if p's value is semicolon.

How about modifying as the following ?

char tmp;
char *p = strchr(pos, ';');

if (p)
{
    tmp = *p;
    *p = '\0';
} 
else
{ 
  tmp = pos[length];
   pos[length] = '\0';
}

......
......

if (p)
{
   *p = tmp;
}
else
{
    pos[length] = tmp;
}

kisoo-han: Sorry, I missed it. Thank you.

mhoh3963 (approved):

PR #4663, opened 2023-09-07
[LETS-785] Create a catch-up thread

wrlawodms:
http://jira.cubrid.org/browse/LETS-785

- Add a worker pool in `page_server` and create the catchup job as a task of it.
  - To append log pages, we need a `cubthread::entry`. For all I know, there are only two legal ways to allocate the entry for a thread: workerpool or daemon. The catchup job is a one-shot task, so I don't think it's a daemon's job.
  - The `page_server` creates a connection handler and the catch-up task, and launches it when it receives the catch-up start msg from its ATS. The task disconnects the connection handler when it's done.
  - The task checks if the connection_handler is alive whenever accessing it. If it's reset to `nullptr` (disconnected) by another, the task just finishes. It lets us destroy the connection_handler whenever we want regardless of whether the task is launched, not launched, or waiting to be launched. 
- It's done in parallel to fetch log pages and append pages.
- The disconnection of the `followee_connection_handler` happens (here and in the following PR)
  - when he task's done the catch-up job, here
  - when the follower PS stops (shutdown)
  - **NOT** when the followee disconnects. We now in the umbrella epic assume that the followee never dies or disconnects before the catchup is done.


src/server/page_server.cpp
```diff
@@ -30,13 +30,24 @@
...
 #include <functional>
 #include <thread>
 
+page_server::page_server (const char *db_name)
+  : m_server_name { db_name }
+{
+  const auto thread_count = 4; // Arbitrary chosen
+  const auto task_max_count = thread_count * 4;
+
+  m_workpool = cubthread::get_manager ()->create_worker_pool (thread_count, task_max_count, "page_server_workers",
```
wrlawodms: Introducing the worker pool, we have three worker pools in the `page_server`, one for the request responder for tran servers, one for the request responder for follower servers, and one for the catch-up.
I think it can be integrated into one and shared, but it's arguable. Please give me your opinions. It affects #4667. See https://github.com/CUBRID/cubrid/pull/4667#discussion_r1319572377.

src/thread/thread_manager.cpp
```diff
@@ -421,14 +421,16 @@ namespace cubthread
     std::size_t max_active_workers = NUM_NON_SYSTEM_TRANS;  // one per each connection
     std::size_t max_conn_workers = NUM_NON_SYSTEM_TRANS;    // one per each connection
     std::size_t max_vacuum_workers = prm_get_integer_value (PRM_ID_VACUUM_WORKER_COUNT);
+    std::size_t max_server_type_workers =
+	    4; // used in page_server or tran_server. TODO set depending on server type and configurable.
```
wrlawodms: #4667 addresses this, and here or there the number for the added worker pool has to be taken into consideration as well. After the discussion on https://github.com/CUBRID/cubrid/pull/4667#discussion_r1319572377.

src/server/page_server.cpp
```diff
@@ -30,13 +30,24 @@
 #include "page_buffer.h"
 #include "server_type.hpp"
 #include "system_parameter.h"
+#include "thread_entry_task.hpp"
```
```suggestion
#include "thread_manager.hpp"
```
To use `create_worker_pool ()`, thread_manager.hpp is required. Maybe it is already declared in some headers

src/server/page_server.cpp
```diff
@@ -30,13 +30,24 @@
...
 #include <cassert>
 #include <cstring>
 #include <functional>
 #include <thread>
 
+page_server::page_server (const char *db_name)
+  : m_server_name { db_name }
+{
+  const auto thread_count = 4; // Arbitrary chosen
+  const auto task_max_count = thread_count * 4;
```
joohok, line 45: I don't understand why we're multiplying thread_count by 4 
Could you explain it to me ?

src/server/page_server.cpp
joohok: I have a little knowledge about thread manager, so I have a question! 

It seems that one followee is responsible for handling requests from multiple followers. Depending on the situation, it's possible that there are more follower requests than the number of threads available in the worker pool. What happens when there are no more available threads in the worker pool? Does it wait until it can acquire thread resource to process follower requests, or does it mean it can't perform follower requests because it couldn't allocate thread?

If the situation leads to not being able to perform follower requests, would it be a good idea to prevent this by setting a maximum number of followers that can be processed at once using system parameters, for instance?

Additionally, would it be a good approach to use cubthread::get_manager()->is_pool_full() or cubthread::get_manager()->try_task() to inform followers whether their requests can be processed?

src/server/page_server.cpp
```diff
@@ -30,13 +30,24 @@
...
+
+  m_workpool = cubthread::get_manager ()->create_worker_pool (thread_count, task_max_count, "page_server_workers",
+	       nullptr, 1, false);
```
cristiarg: checkout:
```
```
I think we also need a context manager to specify thread type.

src/server/page_server.cpp
```diff
@@ -207,78 +221,30 @@ page_server::tran_server_connection_handler::receive_start_catch_up (tran_server
   unpacker.unpack_int (port);
   cublog::lsa_utils::unpack (unpacker, catchup_lsa);
 
-  _er_log_debug (ARG_FILE_LINE, "[CATCH-UP] Start catching up with the follower (%s:%d), until LSA = (%lld|%d)\n",
+  _er_log_debug (ARG_FILE_LINE,
+		 "[CATCH-UP] It's been requested to start catch-up with the follower (%s:%d), until LSA = (%lld|%d)\n",
```
Use `CATCH_UP` as it is one "word".
```suggestion
		 "[CATCH_UP] It's been requested to start catch-up with the follower (%s:%d), until LSA = (%lld|%d)\n",
```

src/server/page_server.cpp
```diff
@@ -944,6 +913,130 @@ page_server::disconnect_all_tran_servers ()
...
+    return error_code;
+  };
+
+  _er_log_debug (ARG_FILE_LINE, "[CATCH-UP] The catch-up starts with pages ranging from %lld to %lld, count = %lld.\n",
+		 start_pageid, end_pageid, total_page_count);
+
+  LOG_PAGEID request_start_pageid = start_pageid;
+  size_t remaining_page_cnt = total_page_count;
+  size_t request_page_cnt = std::min (log_pgptr_recv_vec.size (), remaining_page_cnt);
+  auto req_future = std::async (std::launch::async, request_pages_to_buffer, request_start_pageid, request_page_cnt);
```
cristiarg, line 996: I understand that `std::async` is to parallelize:
- requests to followee (producing log pages)
- and appending in follower (consume those)
Having `execute_catchup` run as an async task, while also making use of `std::async` which - behind the scene, creates another thread pool - somewhat defeats the purpose.

What I would do is:
- implement these two aspects as two proper separate implementations of `cubthread::task` - as the async responder does
- move the code in a separate file
- have these two tasks pass the received pages between them as done here
- when  the "produce" task is done, it signals this to the "consume" task
- when the "consume" task is done, it does the signalling

Another, less important aspect, the "swap" approach assumes that the "consume" is much faster than "produce" which is probably true 99.(9)% of the time.

src/server/page_server.cpp
wrlawodms: Honestly, I just dis as request `server_request_responder` does. The number of task doesn't matter if we don't use `try_task` mechanism, does it? Or what value do you think should be set?

src/server/page_server.cpp
wrlawodms: > Does it wait until it can acquire thread resource to process follower requests,

It's the case.

> If the situation leads to not being able to perform follower requests, would it be a good idea to prevent this by setting a maximum number of followers that can be processed at once using system parameters, for instance?
Additionally, would it be a good approach to use cubthread::get_manager()->is_pool_full() or cubthread::get_manager()->try_task() to inform followers whether their requests can be processed?

hmm. It can be a good idea, but some additional design should be followed. For example, if the follower is denied because a followee is too busy, it should attach to another followee.

On my mind, I have an another optimization approach when a followee gets a bottleneck, which would be introduced far later. When a cluster is running, the majority of PSes are alive, so we can assign at most one follower to a followee since the number of followee is larger than followers. And, when a cluster is lagging since less than the majority are alive, we have to assign more than one to a followee. In this case, it's no point to limit the number of followers (there is no other option but waiting).
 I think this way has more advantageous than just limiting the number of followers per a followee. In other words, there would be an optimal way in the system perspective. It means no need to pass on it to users.

Anyway, this is about an optimization and it's too premature to discuss it, I guess. Feel free to be against.

src/server/page_server.cpp
wrlawodms: done.

src/server/page_server.cpp
wrlawodms: done.

src/server/page_server.cpp
wrlawodms: > Having execute_catchup run as an async task, while also making use of std::async which - behind the scene, creates another thread pool - somewhat defeats the purpose.

I don't understand this. Do you mean this parallel approach is not worthy? Or, does it defeat the purpose to use `std::async`? I guess it's worth having two jobs parallel, though: appending 1\~10 pages into its local buffer, while pulling next pages, 11\~20, from remote. Are you against it? or just against using `std::async` instead of CUBRID's `task?

src/server/page_server.cpp
wrlawodms: I think you're right, but how about getting it done in http://jira.cubrid.org/browse/LETS-794? I guess we can use the existing one in the `server_request_responder`.

src/server/page_server.cpp
joohok: Ah, it seems to be used only in `try_task()`. I thought it was a meaningless value, but I wondered if there was any meaning in the multiplication. Thank you for your response

src/server/page_server.cpp
joohok: Oh, I see. Thank you for explaining. It might be a problem if it dies without acquiring the thread resource, but it should be fine if it waits until it gets the thread resource. The rest should really be discussed as part of the optimization phase, as you suggested.

joohok (approved):

src/server/page_server.cpp
cristiarg: The approach with 2 separate threads is definitely ok.
But the implementation is cumbersome with `std::async` because it mixes too many concepts together and it just does not look "clean" to my eyes.

It will certainly work.

However there can be a small performance bottleneck as the log page retrieval in the `std::async` is not completely independent of the actual `page_server::execute_catchup` execution - both threads have a "piece-meal" way of progressing. So, supposing that retrieval goes way faster than processing (very unlikely) the retrieval will be starved.

Take this comment as a "nit" - we can certainly change later.

cristiarg (approved):

src/server/page_server.cpp
wrlawodms: I've got your point. I've chosen the most handy one at that time, but yes, I relate to that it gets complicated with many concepts. I'll refactor it. Thank you.
I'll gather probable refactoring in http://jira.cubrid.org/browse/LETS-796, which I'll return after making progress.

PR #4667, opened 2023-09-07
[LETS-788] add server type thread count to the thread manager maximum thread count

cristiarg:
http://jira.cubrid.org/browse/LETS-788

- extract the page server's server request responder number of threads and tasks into a sysparam
- add the following maximum thread counts to the thread manager maximum thread count:
  - page server: parallel replication maximum thread count
  - page server: async server request responder maximum thread count


src/base/system_parameter.h
```diff
@@ -476,6 +476,8 @@ enum param_id
   PRM_ID_REPLICATION_COMPRESS_PAGES_TRANSFER,
   PRM_ID_SCAL_PERF_PTS_REPL_THREAD_IGNORE_UNFIX,
   PRM_ID_SCAL_PERF_PS_REPL_THREAD_IGNORE_UNFIX,
+  PRM_ID_SCAL_PERF_PS_REQ_RESPONDER_THREAD_COUNT,
+  PRM_ID_SCAL_PERF_PS_REQ_RESPONDER_TASK_COUNT,
```
cristiarg, line 480: The identifiers of these parameters might need to be changed later.
They have been added as a trial run during performance investigations. That's why they are cryptically called with `_SCAL_PERF_`.

src/server/server_request_responder.hpp
```diff
@@ -163,9 +164,13 @@ template<typename T_CONN>
 server_request_responder<T_CONN>::server_request_responder ()
   : m_threads_context_manager (std::make_unique<cubthread::system_worker_entry_manager> (TT_SYSTEM_WORKER))
 {
-  const auto THREAD_COUNT = 4; // std::thread::hardware_concurrency ();
-  // TODO configurable and register it to cubthread::manager (set_max_thread_count_from_config)
-  const auto TASK_MAX_COUNT = THREAD_COUNT * 4;
+  const int prm_thread_count = prm_get_integer_value (PRM_ID_SCAL_PERF_PS_REQ_RESPONDER_THREAD_COUNT);
```
wrlawodms: I've added another request_responder in the page server for the catchup connection_handler, `m_follower_responder`. Regarding this, we should either consider this to the thread number.
 Or it's better to both use one responder, which I tried, but failed.

wrlawodms: Besides this, I think I should add a work pool to `page_server`, for the catchup thread that I'm working on. This needs the `cubthread::entry` to append pages through the log manager. By my analysis, CUBRID only have two legal ways to get an entry, either with daemon or workepool. In my opinion, the catchup execution is a one-shot execution, so it's not a work for a daemon. 
All in all, we end up using three worker pools in the page_server. All of these have to be taken into considerations here. 
I doubt that it's a fair choice to introduce that many work pools, though. So, how about using just one work pool and share it with two request responders?
Welcome to any opinion. Actually, I'd like to have your opinion on the implementations I mentioned. Thank you.

src/base/system_parameter.c
```diff
@@ -2398,6 +2400,18 @@ static unsigned int prm_scal_perf_ps_repl_thread_ignore_unfix_flag = 0;
 static const bool prm_scal_perf_ps_repl_thread_ignore_unfix_default = false;
 bool PRM_SCAL_PERF_PS_REPL_THREAD_IGNORE_UNFIX_VALUE = prm_scal_perf_ps_repl_thread_ignore_unfix_default;
 
+static unsigned int prm_scal_perf_ps_req_responder_thread_count_flag = 0;
+static int prm_scal_perf_ps_req_responder_thread_count_default = 0;
+int PRM_SCAL_PERF_PS_REQ_RESPONDER_THREAD_COUNT_CURRENT_VALUE = 0;
+static int prm_scal_perf_ps_req_responder_thread_count_upper_value = 64;
+static int prm_scal_perf_ps_req_responder_thread_count_lower_value = 0;
```
wrlawodms: I think we can do it like the `prm_thread_core_count_default` does. 
```suggestion
static unsigned int prm_scal_perf_ps_req_responder_thread_count_flag = 0;
#if defined (SERVER_MODE)
static int prm_scal_perf_ps_req_responder_thread_count_default = (int) cubthread::system_core_count ();
#else
static int prm_scal_perf_ps_req_responder_thread_count_default = 1;
#endif
int PRM_SCAL_PERF_PS_REQ_RESPONDER_THREAD_COUNT_CURRENT_VALUE = 0;
static int prm_scal_perf_ps_req_responder_thread_count_upper_value = 64;
static int prm_scal_perf_ps_req_responder_thread_count_lower_value = 0;
```

With this, I think we can remove the setting default value in `get_maxim_extra_thread_count_by_server_type` and `server_request_responder()` like
```
   const size_t THREAD_COUNT =
 	  (size_t) ((prm_thread_count <= 0) ? std::thread::hardware_concurrency () : prm_thread_count);
   const size_t TASK_MAX_COUNT =
 	  (size_t) ((prm_task_count <= 0) ? (std::thread::hardware_concurrency () * 4) : prm_task_count);
```

src/base/server_type.cpp
```diff
@@ -271,6 +272,63 @@ bool is_tran_server_with_remote_storage ()
...
+	{
+	  // a safe bet
+	  replication_parallel_thread_count_max = std::thread::hardware_concurrency ();
+	}
+
+      // thread pool used by page server to asynchronously service requests from transaction servers
+      //
+      int server_request_responder_thread_count_min = -1;
+      int server_request_responder_thread_count_max = -1;
+      prm_error = sysprm_get_range (prm_get_name (PRM_ID_SCAL_PERF_PS_REQ_RESPONDER_THREAD_COUNT),
```
wrlawodms: Doesn't it retrieve the max value, 64? I think you'd like to use just `prm_get_integer_value (PRM_ID_SCAL_PERF_PS_REQ_RESPONDER_THREAD_COUNT)`, which is the number of threads used. Or, please correct me.

src/base/server_type.cpp
```diff
@@ -271,6 +272,63 @@ bool is_tran_server_with_remote_storage ()
...
+				    &server_request_responder_thread_count_min, &server_request_responder_thread_count_max);
+      if (prm_error != PRM_ERR_NO_ERROR)
+	{
+	  // a safe bet
+	  server_request_responder_thread_count_max = std::thread::hardware_concurrency ();
+	}
+
+      return replication_parallel_thread_count_max + server_request_responder_thread_count_max;
```
wrlawodms: Since we have two responder though it will be integrated into one in the end.
```suggestion
      return replication_parallel_thread_count_max + server_request_responder_thread_count_max * 2;
```

src/base/system_parameter.c

src/base/server_type.cpp
cristiarg: Correct. Fixed.

src/base/server_type.cpp
cristiarg: Done.

wrlawodms (approved):

PR #4675, opened 2023-09-11
[CBRD-24798] Changing the index structure using ALTER INDEX is treated as an error

ctshim:
http://jira.cubrid.org/browse/CBRD-24798

- We cannot modify the constituent column information of an index using the ALTER INDEX REBUILD statement, but the parser does not generate an error.
Fix it so that an explicit error occurs.

```
drop table if exists tbl;
create table tbl (id int, v int);
create index idx on tbl(v);
alter index idx on tbl (v, id) rebuild;  --   error !
```


hgryoo (approved):

beyondykk9 (approved):

hyunikn (approved):

hornetmj:
@ctshim This fix is related to the query parsing. So, my review is not necessary. Thanks.

PR #4687, opened 2023-09-14
[CBRD-24855] Memory Monitoring - Register heap file module

Rudeus:
http://jira.cubrid.org/browse/CBRD-24855

Register heap module
 - When base data structure is returned to its free-list, free-area, etc.,  subtract transaction memory usage.
 - Add mmon_tracking_tag in THREAD_ENTRY
 - Add memory tracking functions at base data structures
 - Add memory tracking functions at heap file module
   - will be upload per component

### DIFF
--- a/msg/en_US.utf8/utils.msg
+++ b/msg/en_US.utf8/utils.msg
@@ -1331,5 +1331,5 @@ valid options:\n\
     -m, --module=TYPE             display detailed memory usage information of the specified TYPE.\n\
                                   TYPE of modules\n\
                                   -1      all modules\n\
-                                   0      heap module\n\
-                                   1      common module\n
+                                   0      common module\n\
+                                   1      heap module\n
--- a/msg/en_US.utf8/utils.msg
+++ b/msg/en_US.utf8/utils.msg
@@ -1330,4 +1330,6 @@ valid options:\n\
                                   WARNING: This option has to be used with '-t(--transaction)' option.\n\
     -m, --module=TYPE             display detailed memory usage information of the specified TYPE.\n\
                                   TYPE of modules\n\
-                                  -1      all modules\n
+                                  -1      all modules\n\
+                                   0      heap module\n\
+                                   1      common module\n
--- a/src/base/area_alloc.c
+++ b/src/base/area_alloc.c
@@ -442,12 +442,12 @@ area_alloc (AREA * area)
   if (block == NULL)
     {
       pthread_mutex_unlock (&area->area_mutex);
+#ifdef SERVER_MODE
+      (void) mmon_set_tracking_tag (prev_tag);
+#endif
       /* error has been set */
       return NULL;
     }
-#ifdef SERVER_MODE
-  (void) mmon_set_tracking_tag (prev_tag);
-#endif
 
   /* alloc free entry from this new block */
   entry_idx = block->bitmap.get_entry ();
@@ -459,6 +459,9 @@ area_alloc (AREA * area)
       free_and_init (block);
 
       pthread_mutex_unlock (&area->area_mutex);
+#ifdef SERVER_MODE
+      (void) mmon_set_tracking_tag (prev_tag);
+#endif
       /* error has been set */
       return NULL;
     }
@@ -485,6 +488,10 @@ found:
 
   entry_ptr += AREA_PREFIX_SIZE;
 #endif /* !NDEBUG */
+
+#ifdef SERVER_MODE
+  (void) mmon_set_tracking_tag (prev_tag);
+#endif
 
   assert (entry_ptr < (block->data + area->block_size));
 
@@ -626,7 +633,7 @@ area_flush (AREA * area)
       next_blockset = blockset->next;
 
 #ifdef SERVER_MODE
-      mmon_sub_stat_with_tracking_tag (sizeof (AREA_BLOCK) * blockset->used_count + sizeof (AREA_BLOCKSET_LIST));
+      mmon_sub_stat_with_tracking_tag (sizeof (AREA_BLOCK) * blockset->used_count);
 #endif
       for (i = 0; i < blockset->used_count; i++)
 	{
--- a/src/base/area_alloc.c
+++ b/src/base/area_alloc.c
@@ -42,6 +42,8 @@
 
 #if !defined (SERVER_MODE)
 #include "work_space.h"
+#else /* defined (SERVER_MODE) */
+#include "memory_monitor_sr.hpp"
 #endif /* !defined (SERVER_MODE) */
 
 #if !defined (SERVER_MODE)
@@ -119,6 +121,9 @@ void
 area_final (void)
 {
   AREA *area, *next;
+#ifdef SERVER_MODE
+  MMON_STAT_ID prev_tag = mmon_set_tracking_tag (MMON_COMMON);
+#endif
 
   for (area = area_List, next = NULL; area != NULL; area = next)
     {
@@ -131,6 +136,9 @@ area_final (void)
   set_area_reset ();
 
   pthread_mutex_destroy (&area_List_lock);
+#ifdef SERVER_MODE
+  (void) mmon_set_tracking_tag (prev_tag);
+#endif
 }
 
 /*
@@ -249,6 +257,7 @@ area_destroy (AREA * area)
   AREA *a, *prev;
 #if defined(SERVER_MODE)
   int rv;
+  MMON_STAT_ID prev_tag = mmon_set_tracking_tag (MMON_COMMON);
 #endif /* SERVER_MODE */
 
   assert (area != NULL);
@@ -277,6 +286,9 @@ area_destroy (AREA * area)
   area_flush (area);
 
   free_and_init (area);
+#if defined(SERVER_MODE)
+  (void) mmon_set_tracking_tag (prev_tag);
+#endif
 }
 
 /*
@@ -308,6 +320,9 @@ area_alloc_block (AREA * area)
 
       return NULL;
     }
+#ifdef SERVER_MODE
+  mmon_add_stat_with_tracking_tag (total);
+#endif
 
   new_block->bitmap.init (LF_BITMAP_LIST_OF_CHUNKS, (int) area->alloc_count, LF_AREA_BITMAP_USAGE_RATIO);
   assert ((int) area->alloc_count == new_block->bitmap.entry_count);
@@ -365,6 +380,7 @@ area_alloc (AREA * area)
   char *entry_ptr;
 #if defined(SERVER_MODE)
   int rv;
+  MMON_STAT_ID prev_tag = mmon_set_tracking_tag (MMON_COMMON);
 #endif /* SERVER_MODE */
 #if !defined (NDEBUG)
   int *prefix;
@@ -429,6 +445,9 @@ area_alloc (AREA * area)
       /* error has been set */
       return NULL;
     }
+#ifdef SERVER_MODE
+  (void) mmon_set_tracking_tag (prev_tag);
+#endif
 
   /* alloc free entry from this new block */
   entry_idx = block->bitmap.get_entry ();
@@ -606,6 +625,9 @@ area_flush (AREA * area)
     {
       next_blockset = blockset->next;
 
+#ifdef SERVER_MODE
+      mmon_sub_stat_with_tracking_tag (sizeof (AREA_BLOCK) * blockset->used_count + sizeof (AREA_BLOCKSET_LIST));
+#endif
       for (i = 0; i < blockset->used_count; i++)
 	{
 	  block = blockset->items[i];
--- a/src/base/area_alloc.c
+++ b/src/base/area_alloc.c
@@ -129,6 +129,9 @@ area_final (void)
     {
       next = area->next;
       area_flush (area);
+#ifdef SERVER_MODE
+      mmon_sub_stat_with_tracking_tag (sizeof (AREA));
+#endif
       free_and_init (area);
     }
   area_List = NULL;
@@ -157,14 +160,21 @@ area_create (const char *name, size_t element_size, size_t alloc_count)
   size_t adjust;
 #if defined (SERVER_MODE)
   int rv;
+  MMON_STAT_ID prev_tag = mmon_set_tracking_tag (MMON_COMMON);
 #endif /* SERVER_MODE */
 
   area = (AREA *) malloc (sizeof (AREA));
   if (area == NULL)
     {
+#ifdef SERVER_MODE
+      (void) mmon_set_tracking_tag (prev_tag);
+#endif
       er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, sizeof (AREA));
       return NULL;
     }
+#ifdef SERVER_MODE
+  mmon_add_stat_with_tracking_tag (sizeof (AREA));
+#endif
   area->blockset_list = NULL;
 
   if (name == NULL)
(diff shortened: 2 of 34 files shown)


Rudeus:
The content provided so far is related to 'Add mmon_tracking_tag in THREAD_ENTRY' and 'Ready for adding memory tracking function like adding module_id, stat_id, etc.'.

Rudeus:
The content provided so far is related to 'Add mmon_tracking_tag in THREAD_ENTRY' and 'Ready for adding memory tracking function like adding module_id, stat_id, etc.'.

src/base/memory_monitor_sr.cpp
```diff
@@ -588,6 +599,15 @@ namespace cubperf
   {
     add_stat (src, - (size));
     add_stat (dest, size);
+
+    /* If move is occurred from other modules to MMON_COMMON,
+     * it means that memory is returned to common structures for keeping.
+     * ex) pool, free list, reserved, etc..
+     * In this case, transaction's memory usage will be subtracted. */
+    if (dest == MMON_COMMON)
```
joohok: The `if (src == MMON_COMMON)` condition will be addressed in following commit. 
it is discussed offline

Rudeus:
for tracking base data structures, add sub/move/resize functions for using `tracking_tag`.

Rudeus:
Add memory tracking functions for tracking memory usage of base data structures:
 - add memory tracking functions
 - add fake functions lock-free unit test

src/base/memory_monitor_sr.cpp
```diff
@@ -720,3 +731,54 @@ void mmon_aggregate_tran_info (int tran_count, MMON_TRAN_INFO &info)
 
   mmon_Gl->aggregate_tran_info (tran_count, info);
 }
+
+MMON_STAT_ID mmon_set_tracking_tag (MMON_STAT_ID new_tag)
+{
+  MMON_STAT_ID prev_tag;
+  THREAD_ENTRY *cur_thread_p;
+
+  cur_thread_p = thread_get_thread_entry_info ();
```
wrlawodms: What if `cur_thread_p` is `nullptr`?  Isn't a category needed for this like UNKNOWN? What's the design for this case?

src/base/memory_monitor_cl.cpp
```diff
@@ -152,7 +152,14 @@ mmon_print_tran_info (MMON_TRAN_INFO &tran_info)
 
   for (const auto &t_stat : tran_info.tran_stat)
     {
-      fprintf (stdout, "\t%14d | %17lu\n", t_stat.tranid, t_stat.cur_stat);
+      if (t_stat.cur_stat >= 0)
+	{
+	  fprintf (stdout, "\t%14d | %17lu\n", t_stat.tranid, t_stat.cur_stat);
+	}
+      else
+	{
```
wrlawodms, line 181: It's a non-self-describe code that needs a comment.

src/base/memory_monitor_sr.cpp
```diff
@@ -720,3 +731,54 @@ void mmon_aggregate_tran_info (int tran_count, MMON_TRAN_INFO &info)
...
+  mmon_add_stat (cur_thread_p, cur_thread_p->mmon_tracking_tag, size);
+}
+
+void mmon_sub_stat_with_tracking_tag (int64_t size)
+{
+  THREAD_ENTRY *cur_thread_p = thread_get_thread_entry_info ();
+  mmon_sub_stat (cur_thread_p, cur_thread_p->mmon_tracking_tag, size);
+}
+
+void mmon_move_stat_with_tracking_tag (int64_t size, bool tag_is_src)
```
wrlawodms: I think you'd better off dividing these into two `mmon_move_stat_to_common_from_tracking_tag` and `mmon_move_stat_to_tracking_tag_from_common`. Or, whatever it is,  the name should include that it is moved to `MMON_COMMON`.

src/base/object_representation_sr.c
```diff
@@ -1634,6 +1652,9 @@ or_install_btids_foreign_key_ref (DB_SEQ * fk_container, OR_INDEX * index)
 
       if (set_get_element_nocopy (fk_seq, 1, &val) != NO_ERROR)
 	{
+#ifdef SERVER_MODE
+	  mmon_sub_stat_with_tracking_tag (sizeof (OR_FOREIGN_KEY));
```
wrlawodms, line 1674: Why not subtract above two error cases, L1646 and L1638?

src/base/object_representation_sr.c
```diff
@@ -1894,13 +1964,20 @@ or_install_btids_class (OR_CLASSREP * rep, BTID * id, DB_SEQ * constraint_seq, i
       er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, sizeof (OR_ATTRIBUTE *) * att_cnt);
       return;
     }
+#ifdef SERVER_MODE
+  index->att_cnt = att_cnt;
```
wrlawodms, line 1986: Is it fixing a missing part? seems irrelevant to this issue.

wrlawodms (commented):
IN progress..

Rudeus:
Add memory tracking functions for heap_file module.

Rudeus:
finish local test

msg/en_US.utf8/utils.msg
```diff
@@ -1330,4 +1330,6 @@ valid options:\n\
                                   WARNING: This option has to be used with '-t(--transaction)' option.\n\
     -m, --module=TYPE             display detailed memory usage information of the specified TYPE.\n\
                                   TYPE of modules\n\
-                                  -1      all modules\n
+                                  -1      all modules\n\
+                                   0      heap module\n\
+                                   1      common module\n
```
wrlawodms: I'd put `common module` before `heap module`. If other modules come in, it will looks strange. It'd make `heap module` special.

src/base/area_alloc.c
```diff
@@ -149,14 +160,21 @@ area_create (const char *name, size_t element_size, size_t alloc_count)
   size_t adjust;
 #if defined (SERVER_MODE)
   int rv;
+  MMON_STAT_ID prev_tag = mmon_set_tracking_tag (MMON_COMMON);
```
wrlawodms, line 163: You can make `(void) mmon_set_tracking_tag (prev_tag);` called automatically and safely with an RAII object. It doesn't only look good, but it's also safe. Please consider this. 
Example:
```
class mmon_tag_tracker
{
        public:
        mmon_tag_tracker (int tag)
        {
                cur_tag = tag;
                prev_tag = mmon_set_tracking_tag (tag);
        }

        ~mmon_tag_tracker ()
        {
                int ttt =  mmon_set_tracking_tag (prev_tag);
                assert (origin_tag == ttt);
        }

        private:
                int prev_tag;
                int origin_tag; 
}

```
```suggestion
  mmon_tag_tracker tracker track (MMON_COMMON);
```

src/base/memory_monitor_sr.cpp
```diff
@@ -30,6 +30,11 @@
 #include "memory_monitor_sr.hpp"
 #include "log_impl.h"
 
+#if !defined (NDEBUG)
+thread_local char mmon_err_msg[DB_MAX_IDENTIFIER_LENGTH];
```
wrlawodms: Please elaborate on why this must be thread_local.

src/base/memory_monitor_sr.cpp
```diff
@@ -618,6 +667,61 @@ namespace cubperf
   {
     m_aggregater.get_transaction_info (tran_count, info);
   }
+
+  void memory_monitor::stat_dump_debug ()
```
wrlawodms: It's not about heap module. Please separate it to another issue.

src/base/memory_monitor_sr.cpp
```diff
@@ -720,3 +831,81 @@ void mmon_aggregate_tran_info (int tran_count, MMON_TRAN_INFO &info)
...
 }
+
+MMON_STAT_ID mmon_set_tracking_tag (MMON_STAT_ID new_tag)
+{
+  MMON_STAT_ID prev_tag = MMON_STAT_LAST;
+  THREAD_ENTRY *cur_thread_p;
+
+  if (mmon_Gl != nullptr)
+    {
+      cur_thread_p = thread_get_thread_entry_info ();
```
wrlawodms, line 757: What happens `cur_thread_p` is nullptr? It can be null though I'm not sure current usages run into this.

src/compat/db_json.cpp
```diff
@@ -856,6 +871,12 @@ JSON_VALIDATOR::JSON_VALIDATOR (const JSON_VALIDATOR &copy)
   else
     {
       m_schema_raw = strdup (copy.m_schema_raw);
+#ifdef SERVER_MODE
+      if (m_schema_raw != NULL)
```
wrlawodms, line 875: Do we need to consider this kind of exception while the existing code doesn't do it?

src/base/memory_monitor_sr.cpp
hornetmj: I agree with him. I would be nice to handle another PR or issue.

hornetmj (approved):

src/base/memory_monitor_cl.cpp
```diff
@@ -126,9 +131,15 @@ mmon_print_module_info_summary (uint64_t server_mem_usage, std::vector<MMON_MODU
 
   for (const auto &m_info : module_info)
     {
+      // TODO: It's a temporary measure for hiding MMON_OTHERS
+      if (!strcmp (m_info.name, "OTHERS"))
```
joohok, line 137: will this part be removed? 
I think It is unclear what the TODO is and when it should be done

src/base/memory_monitor_common.h
```diff
@@ -29,19 +29,19 @@
 
 typedef enum mmon_module_id
 {
-  /* TODO: this dummy modules will be changed when heap module is registered */
   MMON_MODULE_ALL = -1,
-  MMON_MODULE_DUMMY,
-  MMON_MODULE_LONG_DUMMY,
+  MMON_MODULE_HEAP,
+  MMON_MODULE_COMMON,
+  MMON_MODULE_OTHERS,
```
joohok:
```suggestion
  MMON_MODULE_COMMON,
  MMON_MODULE_HEAP,
  /* Add new modules before this */
  MMON_MODULE_OTHERS,
```

Maybe same opinion with above (in utils.msg)
And it would be better to leave comments on where to add new modules.

src/object/elo.c
```diff
@@ -288,6 +318,11 @@ elo_copy (DB_ELO * elo, DB_ELO * dest)
 		ret = er_errid ();
 		goto error_return;
 	      }
+#ifdef SERVER_MODE
+	    mmon_add_stat_with_tracking_tag (strlen (elo->locator) + 1);
+	    strcpy (out_uri, elo->locator);
```
joohok: why is this string copy needed?

joohok (commented):
not finished yet

msg/en_US.utf8/utils.msg
Rudeus: Done.

src/base/memory_monitor_sr.cpp
Rudeus: We do not assume that `thread_p == NULL`. Since `mmon_Gl` is initialized after the initialization of the `thread_manager`, it guarantees accessibility to all threads managed by the `thread_manager`. If another developer is creating threads without using `THREAD_ENTRY`, the issue is whether these threads should be incorporated into the thread_manager. Considering this should be a separate concern.

src/base/memory_monitor_cl.cpp
Rudeus: Done.

src/base/memory_monitor_sr.cpp
Rudeus: Done. I think that it is no use deviding MMON_COMMON use case, just the tag is used for source or destination.

src/base/object_representation_sr.c
Rudeus: Done.

src/base/object_representation_sr.c
Rudeus: It is for getting size of the structures when they are freed.

src/base/area_alloc.c
Rudeus: It can be added c-only code.

src/base/memory_monitor_sr.cpp
Rudeus: I devided it to another issue.

src/base/memory_monitor_sr.cpp
Rudeus: I devided it to another issue.

src/base/memory_monitor_sr.cpp
Rudeus: We do not assume that thread_p == NULL. Since mmon_Gl is initialized after the initialization of the thread_manager, it guarantees accessibility to all threads managed by the thread_manager. If another developer is creating threads without using THREAD_ENTRY, the issue is whether these threads should be incorporated into the thread_manager. Considering this should be a separate concern.

src/compat/db_json.cpp
Rudeus: In the reference of `strdup()`, it can return 0 when it fails memory allocation.

src/base/memory_monitor_cl.cpp
Rudeus: Done.

src/base/memory_monitor_common.h
Rudeus: Done.

src/object/elo.c
Rudeus: Because if the function goes to fail, it has to free allocated memories, but we don't know which case is hit in this circumstance. So I have to make recognize all string allocations at once. `out_uri` is a static char array.

src/base/area_alloc.c
wrlawodms: A c-complier can compile this code, can't it? It's just a declaration of a structure in the respective of c-compiler.

src/base/memory_monitor_cl.cpp
wrlawodms: I don't understand the comment. What makes it negative?

src/base/memory_monitor_sr.cpp
wrlawodms: OK, let's assume that only threads managed by the thread_manager use the mmon module.

src/base/object_representation_sr.c
wrlawodms: Why should `index->att_cnt` be set to "get size of the structure"?

src/compat/db_json.cpp
wrlawodms: I know it can return 0. What I meant is that if the `strdup` fails, the existing code doesn't seem to work. It means this code doesn't assume it can be 0. It's not for this code, but in general.
Or, is this allowed to be NULL? or just for safe?

wrlawodms (approved):

src/object/elo.c
```diff
@@ -338,11 +376,17 @@ elo_copy (DB_ELO * elo, DB_ELO * dest)
 error_return:
   if (locator != NULL)
     {
+#ifdef SERVER_MODE
+      mmon_add_stat_with_tracking_tag (strlen (out_uri) + 1);
```
joohok:
```suggestion
      mmon_sub_stat_with_tracking_tag (strlen (out_uri) + 1);
```

src/object/elo.c
joohok: If the length of the string is required when the it fails, then we don't need to copy the string I think. Instead we can store the length somewhere

src/object/elo.c
Rudeus: Done.

src/object/elo.c
Rudeus: I checked it and I agree that we don't need to use `out_uri` and strcpy. Done.

joohok (approved):

PR #4691, opened 2023-09-14
[LETS-784] Disconnect ps-to-ps connection

wrlawodms:
http://jira.cubrid.org/browse/LETS-784

- In the follower side: `disconnect_followee_page_server`
  - A follower can just send the disconnect msg and drop the connection_handler.
  - It's always done by another thread than the connection_handler's internal threads so that it doesn't need to be asynchronous. 
  - It disconnect when the server shuts down or it finishes the catch-up.
- In the followee side:  `disconnect_follower_server_async`
  - We can't deal with the case in which a follower disappears before the catch-up is done. It's TODO. So, A followee never disconnects itself, for now. 
  - In this issue, followee waits until all followers are done and disconnected instead of disconnecting when it wants to disconnect.
  - It has to be done by another thread to disconnect than internal threads because they can't join themselves and, in turn, can't destroy the connection handler. It's done using `std::async`. A `async` job can deal with a series of disconnections from multiple followers.
- I have tested this with the sleep_for(10s).
  - A follower finishes the catchup and disconnects.
  - A follower stops during the catchup
  - A followees tries to stop while two followers are doing the catch-up.


src/server/page_server.cpp
```diff
@@ -913,22 +978,93 @@ page_server::disconnect_all_tran_servers ()
...
+    {
+      return conn_uptr.get () == conn;
+    });
+
+    if (it == m_follower_conn_vec.cend ())
+      {
+	// It's already been disconnected or in progress by another thread.
+	return;
+      }
+    er_log_debug (ARG_FILE_LINE, "disconnect_follower_server_async 요청");
```
```suggestion
    er_log_debug (ARG_FILE_LINE, "disconnect_follower_server_async request");
```

src/server/page_server.cpp
```diff
@@ -913,22 +978,93 @@ page_server::disconnect_all_tran_servers ()
...
 	}
       m_followee_conn.reset (nullptr);
 
-      er_log_debug (ARG_FILE_LINE, "The followee page server has been disconnected.\n");
+      er_log_debug (ARG_FILE_LINE, "The followee page server has been disconnected. channel id: %s\n", channel_id.c_str ());
     }
 }
 
+void
+page_server::disconnect_follower_server_async (const follower_connection_handler *conn)
```
```suggestion
page_server::disconnect_follower_page_server_async (const follower_connection_handler *conn)
```

For consistency

hornetmj (approved):

src/server/page_server.cpp
wrlawodms: Removed. It was for debugging, which I missed to remove. Thanks.

src/server/page_server.cpp
wrlawodms: done.

src/server/page_server.cpp
```diff
@@ -913,22 +978,92 @@ page_server::disconnect_all_tran_servers ()
...
+    {
+      return conn_uptr.get () == conn;
+    });
+
+    if (it == m_follower_conn_vec.cend ())
+      {
+	// It's already been disconnected or in progress by another thread.
+	return;
+      }
+    disconnecting_conn_uptr = std::move (*it);
```
cristiarg: nit
cppreference says:
```suggestion
    disconnecting_conn_uptr.reset ((*it).release ());
```

src/server/page_server.hpp
```diff
@@ -300,7 +315,12 @@ class page_server
 
     followee_connection_handler_uptr_t m_followee_conn;
     std::vector<follower_connection_handler_uptr_t> m_follower_conn_vec;
+
     std::mutex m_followee_conn_mutex;
+    std::mutex m_follower_conn_vec_mutex;
+    std::condition_variable m_follower_conn_vec_cv;
+    std::future<void> m_follower_disc_future;
+    std::mutex m_follower_disc_mutex;
```
cristiarg, line 323: Nit.
It would be more reader friendly if the variables are grouped by the follower/followee category with some explanation. Not explanation for every variable. I mean explanation about the general mechanism in each case. With TODOs and current caveats for each one.

I have to admit that I struggle with the follower/followee taxonomy and it would be good to have some general explanation here as well:
```
  follower PS --------------------------> followee PS
[ the booting up PS                      [ the running PS that serves 
    that performs the catch-up           pages to the follower PS ]
    by requesting pages from
    the followee PS ]
```

src/server/page_server.cpp
```diff
@@ -913,22 +978,92 @@ page_server::disconnect_all_tran_servers ()
...
+      }
+
+    auto reset_func = [] (auto &&conn_uptr)
+    {
+      auto channel_id = conn_uptr->get_channel_id ();
+      conn_uptr.reset (nullptr);
+      er_log_debug (ARG_FILE_LINE, "The follower page server has been disconnected. channel id: %s\n", channel_id.c_str ());
+    };
+    m_follower_disc_future = std::async (std::launch::async, reset_func, std::move (disconnecting_conn_uptr));
+    assert (m_follower_disc_future.valid ());
```
cristiarg: add `assert (!m_follower_disc_future.valid ());` in the dtor ?

src/server/page_server.cpp
```diff
@@ -913,22 +978,92 @@ page_server::disconnect_all_tran_servers ()
...
 	}
       m_followee_conn.reset (nullptr);
 
-      er_log_debug (ARG_FILE_LINE, "The followee page server has been disconnected.\n");
+      er_log_debug (ARG_FILE_LINE, "The followee page server has been disconnected. channel id: %s\n", channel_id.c_str ());
     }
 }
 
+void
+page_server::disconnect_follower_page_server_async (const follower_connection_handler *conn)
```
cristiarg, line 1034: nit
```suggestion
page_server::disconnect_from_follower_page_server_async (const follower_connection_handler *conn)
```

src/server/page_server.cpp
```diff
@@ -913,22 +978,92 @@ page_server::disconnect_all_tran_servers ()
...
+
+  if (m_follower_disc_future.valid ())
+    {
+      m_follower_disc_future.get (); // wait until it's done if there is an on-going disconnection.
+    }
+  er_log_debug (ARG_FILE_LINE, "All follower connections have been disconnected.\n");
+}
+
 void
 page_server::disconnect_followee_page_server (const bool with_disc_msg)
```
cristiarg, line 1017: Nit
I think I have said it in a previous review. If so, please excuse.
I find more explicit the name `disconnect_from_followee_page_server`which - I think - coveys the message that it is the "follower" doing the disconnection.

src/server/page_server.cpp
```diff
@@ -913,22 +978,92 @@ page_server::disconnect_all_tran_servers ()
...
+    });
+
+    if (it == m_follower_conn_vec.cend ())
+      {
+	// It's already been disconnected or in progress by another thread.
+	return;
+      }
+    disconnecting_conn_uptr = std::move (*it);
+    m_follower_conn_vec.erase (it);
+    m_follower_conn_vec_cv.notify_one ();
```
I think there might be a hole here where the future is not "get" (and I am not sure whether it's a problem or not):
- imagine there is only one connection to a follower
- execution in `disconnect_follower_page_server_async` stack
  - erase from vector, and notify
- execution moves to the thread which executes `disconnect_all_follower_page_servers`
  - cv triggered, vector found to be empty
  - follow with checking whether there is a valid future - there isn't - because the one just erased has not been `std::async`'ed yet
  - code concludes that "All follower connections have been disconnected."
- execution moves back to this location
  - there is no previous future to get
  - `std::async` the destruction of the connection  handler
  - the future is not waited for

I think that it's more robust to move the `notify_one` after invoking `std::async` below.
And add an assert in the dtor that the future is not valid.

I am not sure though, it looks like it might happen although, in practice, it will be impossible to repro.

cristiarg: However, if we plan to get rid of the `std::async`'s and do everything with cubrid thread-pool and task, just leave it as is.

src/server/page_server.cpp
wrlawodms: Agreed. But, I think we should change other functions names as well like `disconnect_active_tran_server`. It's better to deal with all of them together in the TODO refactoring issue: http://jira.cubrid.org/browse/LETS-795. What do you think?

src/server/page_server.cpp
wrlawodms: > 1) Move assignment operator. Transfers ownership from r to *this as if by calling reset(r.release()) followed by assigning [get_deleter()](https://en.cppreference.com/w/cpp/memory/unique_ptr/get_deleter) from [std::forward](http://en.cppreference.com/w/cpp/utility/forward)<Deleter>(r.get_deleter()).
from https://en.cppreference.com/w/cpp/memory/unique_ptr/operator%3D

What I've found out is that the `move` is for this kind of job, including releasing ownership, transfer, and even moving deleter. And I saw some arguments that prefer move over release, and I relate to it. Could you give me the cppference that you mentioned?

cristiarg (approved):
Approved with only the "non nitpick" comment as mandatory. :)

src/server/page_server.cpp
wrlawodms: Thank you. You're right. Beyond the case you mentioned, It could happen even if we push back the `notify()`. If the `disconnect_all_follower_page_servers()` cuts in between two locks here and the vector is empty, It misses the async.

I think the thing is that we have to overlap the critical section. Then, a deadlock could happen, though. I've modified locks' placements. Please review it.

> However, if we plan to get rid of the std::async's and do everything with cubrid thread-pool and task, just leave it as is.

That's true, but I'd like to address it if it's not that complicated. :)

src/server/page_server.hpp
wrlawodms: I understand. Actually, there is a TODO comment about it above connection_handlers for the catch-up. I thought that the taxonomy could be changed along the way. It's quite fixed, though. Let me add the comment in the refactoring(rename) issue too. I'll link this in the issue to remind.

wrlawodms: It will be soon.

src/server/page_server.cpp
```diff
@@ -913,22 +980,95 @@ page_server::disconnect_all_tran_servers ()
...
+
+  if (it == m_follower_conn_vec.cend ())
+    {
+      // It's already been disconnected or in progress by another thread.
+      return;
+    }
+
+  auto disconnecting_conn_uptr = std::move (*it);
+  m_follower_conn_vec.erase (it);
+  m_follower_conn_vec_cv.notify_one ();
```
cristiarg, line 1052: Nit.
The problem I described in the previous review is not gone:
- a thread is blocked in `page_server::disconnect_all_follower_page_servers` waiting on the condition variable
- there "was" only one follower in the vector and after this `notify_one` the vector is empty
- after notify one the control moves and the function `page_server::disconnect_all_follower_page_servers` is executed completely
- control returns here and the `std::async` is dispatched for execution
- `page_server::~page_server` is called, future assert fails (I think)

The problem can be solved if the notification is only raised after dispatch the `std::async` below. Or by adding the `future....get ()` in the dtor.

As I said, this will never be the case. So, maybe just drop it. :)

src/server/page_server.cpp
```diff
@@ -913,22 +980,95 @@ page_server::disconnect_all_tran_servers ()
...
+
+  auto lockg_disc = std::lock_guard { m_follower_disc_mutex };
+  ulock_conn_vec.unlock ();
```
cristiarg, line 1055: Ingenious way to chain two critical sections. (I have never done this :) ).
But this leads be to believe that it is not necessary to have two mutexes and only one mutex suffices.

What do you think? Or maybe you have a better reason that I don't see.

The fact that the contention is reduced a bit by using the two mutexes has some merit if it were on a highly critical execution path. But this is not - catch-up is rarely occurring event.
So the code is needlessly complex given the real-life scenarios.

src/server/page_server.hpp
```diff
@@ -300,7 +315,12 @@ class page_server
 
     followee_connection_handler_uptr_t m_followee_conn;
     std::vector<follower_connection_handler_uptr_t> m_follower_conn_vec;
+
     std::mutex m_followee_conn_mutex;
+    std::mutex m_follower_conn_vec_mutex;
```
cristiarg, line 320: Nit.
`m_followee_conn_mutex` and `m_follower_conn_vec_mutex` are used mutually exclusive - AFAICT.
One PS cannot be at the same time a follower and a followee.
So, they can be just one - `m_follower_or_followee_depending_on_context_mutex` (a mostly silly suggestion :) ).

cristiarg (approved):

joohok (approved):

src/server/page_server.cpp
wrlawodms: The way I see it, it's resolved.
> after notify one the control moves and the function page_server::disconnect_all_follower_page_servers is executed completely

It's not the case, because `disconnect_all_follower_page_servers` can't finish. When checking if the vector is empty, it acquires the lock of `m_follower_conn_vec_mutex`, but it's blocked if it's not released in this function. And, if the lock is released, it acquires the lock of `m_follower_disc_mutex`, which is   held by `disconnect_follower_page_server_async`, so it's blocked. Note that the lock of `m_follower_conn_vec_mutex` is unlocked  means that `m_follower_disc_mutex` is held in `disconnect_follower_page_server_async`.

src/server/page_server.cpp
wrlawodms: In my view, it's necessary, not for reducing the critical section. Please see the [comment above](https://github.com/CUBRID/cubrid/pull/4691#discussion_r1357830592). What I tried first is using only one mutex, but it ended up a deadlock :(.

PR #4692, opened 2023-09-15
[CBRD-24830] Restoring original line number of PL/CSQL code from Java run-time exception

hyunikn:
http://jira.cubrid.org/browse/CBRD-24830

- Java code generation logic was moved from AST node classes to JavaCodeWriter calss.
- JavaCodeWriter class works in two phases
  - Visit phase (it follows the Visitor pattern) returns an object of CodeToResolve interface, which has 'resolve()' method
    - CodeToResolve objects are either a CodeTemplate or CodeTemplateList object
      - CodeTemplate contains a code template and substitution (filling template holes) specification
        - template holes are either a 'big' or 'small' holes. 
          - big holes are substituted by one or more lines of code
          - small holes are substituted by a single word in a line
      - CodeTemplateList: list of CodeTemplates
  - Resolve phase appends code lines one by one into a List<String> (codeLines) from the first line to the last, and appends 'code range markers' to a StringBuilder (codeRangeMarkers).
    - Every line of generated Java code amounts to a single AST node, in other words, no two AST nodes generate Java code in a single line.
    - Code Range Markers mark the beginning and end lines of an AST node in the generated Java code
    - The beginning markers keeps the original PL/CSQL line and column numbers of an AST node
    - Markers are not generated for AST nodes of literals because literals are not likely to cause an exception.
- Generated Java class contains code range markers string as a private static member, and uses it at run-time to map a Java line number to a pair (line # , column #) of the original PL/CSQL program


pl_engine/pl_server/src/main/java/com/cubrid/plcsql/predefined/PlcsqlRuntimeError.java
```diff
@@ -43,22 +43,20 @@ public int getLine() {
...
+        return this;
     }
 
     // --------------------------------------------
     // Private
     // --------------------------------------------
 
-    private int line;
-    private short column;
+    private int line = -1;; // -1: unknown yet
```
hgryoo:
```suggestion
    private int line = -1; // -1: unknown yet
```

hgryoo (approved):

PR #4695, opened 2023-09-15
[CBRD-24983] Fixed an issue where replication to the slave failed due to a change in the record type because applylogdb inserts space for MVCC_INSERT_ID into the record.

hornetmj:
http://jira.cubrid.org/browse/CBRD-24983

Purpose
- fix the failure of replication to the slave node when executing loaddb -C
- the reason is that the record type is changed from REC_HOME to REC_BIGONE
   - loaddb -C generated RVHF_INSERT log with BU lock (it means that non-mvcc type record is inserted)
   - applylogdb make a room for mvcc insert id(8 byte). So, the length of record is changed
   - the changed record length is over heap_Maxslotted_reclength. So, cub_server on slave node tries to change record header format to REC_BIGONE type (it means that additional space for the mvcc delete id and previous version lsa is needed)
   - however, it fails due to recdes.area_size is equals to recdes.length
   - in current mechnism, this works have to done by applylogdb. So, new function called la_make_room_for_mvcc_delid_and_prev_ver() is introduced. It is familiar with la_make_room_for_mvcc_insid()

Implementation
- introduce la_make_room_for_mvcc_delid_and_prev_ver() function
- additional bug fix
  - see la_make_room_for_mvcc_insid()
    - OR_INT_SIZE is changed to OR_MVCC_INSERT_ID_OFFSET to consider the chn (please check it together)

Remarks
- while performing the reported reproduction scenario, I discovered an issue where some tables were not replicated to the slave. This issue is currently being investigated, and a decision on how to handle it will be made after determining the cause
- some functions such as heap_get_class_num_objects_pages(), heap_has_instance(), etc have been moved to sort in the same order as network requests and arrange them in adjacent positions


src/transaction/log_applier.c
```diff
@@ -8145,6 +8194,10 @@ la_apply_log_file (const char *database_name, const char *log_path, const int ma
     }
 #endif /* UNSTABLE_TDE_FOR_REPLICATION_LOG */
 
+  (void) heap_get_maxslotted_reclength (la_Info.maxslotted_reclength);
+
```
wrlawodms, line 8199:
```suggestion
  error = heap_get_maxslotted_reclength (la_Info.maxslotted_reclength);
  if (error != NO_ERROR)
  {
      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_HA_LA_REPL_FILTER_GENERIC, 1,
		  "failed to get la_Info.maxslotted_reclength");
     return error;
  }

```

src/transaction/log_applier.c
```diff
@@ -8145,6 +8194,10 @@ la_apply_log_file (const char *database_name, const char *log_path, const int ma
...
+  er_log_debug (ARG_FILE_LINE, "la_Info.maxslotted_reclength=%d\n", la_Info.maxslotted_reclength);
+
```
wrlawodms, line 8201:
```suggestion
```

src/transaction/log_applier.c
```diff
@@ -4678,8 +4714,17 @@ la_get_recdes (LOG_LSA * lsa, LOG_PAGE * pgptr, RECDES * recdes, unsigned int *r
       OR_PUT_INT (recdes->data, repid_and_flag_bits);
 
       la_make_room_for_mvcc_insid (recdes);
-    }
 
+      if (recdes->length > la_Info.maxslotted_reclength)
```
wrlawodms, line 4719: With a short explanation.
```suggestion
      // It will be coverted to REC_BIGONE which contains all mvcc flags on. 
      if (recdes->length > la_Info.maxslotted_reclength)
```

src/transaction/log_applier.c
```diff
@@ -3686,7 +3689,40 @@ la_make_room_for_mvcc_insid (RECDES * recdes)
 
   assert (recdes->area_size >= recdes->length + OR_MVCCID_SIZE);
 
-  LA_MOVE_INSIDE_RECORD (recdes, OR_INT_SIZE + OR_MVCCID_SIZE, OR_INT_SIZE);
+  //LA_MOVE_INSIDE_RECORD (recdes, OR_INT_SIZE + OR_MVCCID_SIZE, OR_INT_SIZE);
+  LA_MOVE_INSIDE_RECORD (recdes, OR_MVCC_INSERT_ID_OFFSET + OR_MVCCID_SIZE, OR_MVCC_INSERT_ID_OFFSET);
```
wrlawodms:
```suggestion
  LA_MOVE_INSIDE_RECORD (recdes, OR_MVCC_INSERT_ID_OFFSET + OR_MVCC_INSERT_ID_SIZE, OR_MVCC_INSERT_ID_OFFSET);
```
1. OR_MVCC_INSERT_ID_SIZE seems more proper.
2. I think your change is correct. The existing code seems to move the record not from insert_mvcc_id, but from chn. I don't really know about `chn`, but it would've acted up. The only concern I have is somewhere else might have some fixes for this (I think it's not :)). Anyway, I agree with this change.
The record header format:
`RepID(3) | flags(1) | chn (4) | insert_mvcc_id | ...`

src/transaction/log_applier.c
```diff
@@ -3686,7 +3689,40 @@ la_make_room_for_mvcc_insid (RECDES * recdes)
...
+
+  repid_and_flag_bits = OR_GET_MVCC_REPID_AND_FLAG (recdes->data);
+  mvcc_flag = (char) ((repid_and_flag_bits >> OR_MVCC_FLAG_SHIFT_BITS) & OR_MVCC_FLAG_MASK);
+
+  assert ((mvcc_flag & OR_MVCC_FLAG_VALID_INSID) && (mvcc_flag & OR_MVCC_FLAG_VALID_DELID)
+	  && (mvcc_flag & OR_MVCC_FLAG_VALID_PREV_VERSION));
+
+  assert (recdes->area_size >= recdes->length + OR_MVCCID_SIZE + OR_MVCC_PREV_VERSION_LSA_SIZE);
+
+  LA_MOVE_INSIDE_RECORD (recdes, OR_MVCC_DELETE_ID_OFFSET (mvcc_flag) + OR_MVCCID_SIZE + OR_MVCC_PREV_VERSION_LSA_SIZE,
```
wrlawodms:
```suggestion
  LA_MOVE_INSIDE_RECORD (recdes, OR_MVCC_DELETE_ID_OFFSET (mvcc_flag) + OR_MVCC_DELETE_ID_SIZE + OR_MVCC_PREV_VERSION_LSA_SIZE,
```

src/transaction/log_applier.c
```diff
@@ -3686,7 +3689,40 @@ la_make_room_for_mvcc_insid (RECDES * recdes)
...
+
+  assert (recdes->type != REC_BIGONE);
+
+  repid_and_flag_bits = OR_GET_MVCC_REPID_AND_FLAG (recdes->data);
+  mvcc_flag = (char) ((repid_and_flag_bits >> OR_MVCC_FLAG_SHIFT_BITS) & OR_MVCC_FLAG_MASK);
+
+  assert ((mvcc_flag & OR_MVCC_FLAG_VALID_INSID) && (mvcc_flag & OR_MVCC_FLAG_VALID_DELID)
+	  && (mvcc_flag & OR_MVCC_FLAG_VALID_PREV_VERSION));
+
+  assert (recdes->area_size >= recdes->length + OR_MVCCID_SIZE + OR_MVCC_PREV_VERSION_LSA_SIZE);
```
wrlawodms:
```suggestion
  assert (recdes->area_size >= recdes->length + OR_MVCC_DELETE_ID_SIZE + OR_MVCC_PREV_VERSION_LSA_SIZE);
```

src/communication/network_interface_cl.c
```diff
@@ -1798,6 +1902,47 @@ heap_reclaim_addresses (const HFID * hfid)
...
+#else /* CS_MODE */
+  int success = ER_FAILED;
+
+  THREAD_ENTRY *thread_p = enter_server ();
+
+  success = xheap_get_maxslotted_reclength (maxslotted_reclength);
+
+  exit_server (*thread_p);
+
+  return success;
```
wrlawodms, line 1942: Does this interface need to support non-CS-mode?
```suggestion
```

wrlawodms (approved):
Please regard all comments as NIT :).

joohok (approved):

src/transaction/log_applier.c
hornetmj: Thanks to check second one.

src/communication/network_interface_cl.c
hornetmj: Yes, it doesn't need. I just add like other function in this file for the futher usages.

src/transaction/log_applier.c
hornetmj: The point of the error handling is good to me. However ER_HA_LA_REPL_FILTER_GENERIC is not a good choice. So, I would like to leave them as it is at now.

Rudeus (approved):

PR #4699, opened 2023-09-19
[CBRD-24907] Add system views to consolidate information from various database objects

hgryoo:
http://jira.cubrid.org/browse/CBRD-24907

The following views are added in this PR.
- DB_DBA_OBJECTS: describes all objects in the database. Only DBA user is able to access this view
- DB_USER_OBJECTS: describes all objects owned by the current user. This view does not display the OWNER column.
- DB_ALL_OBJECTS: describes all objects accessible to the current user.


hgryoo:
About `generated` column, the child tables of a partition need to be set 'Y'. But there is no such a information in the _db_class.
We need to discuss about it.

src/transaction/boot_cl.c
```diff
@@ -6720,6 +6723,255 @@ boot_define_view_synonym (void)
...
+       'N'                     AS [generated] \
+    FROM \
+       [db_trig] AS [t] \
+   UNION ALL  \
+    /* SERIAL */ \
+    SELECT \
+       'PUBLIC'   AS [owner], \
+       [s].[name] AS [object_name], \
+       'serial'   AS [object_type], \
+       'N'        AS [generated] \
```
beyondykk9, line 6770: db_serial can include generated attributes for "AUTO_INCREMENT", however, we cannot identify it.

src/transaction/boot_cl.c
```diff
@@ -6720,6 +6723,255 @@ boot_define_view_synonym (void)
...
+      return error_code;
+    }
+
+  error_code = au_change_owner (class_mop, Au_dba_user);
+  if (error_code != NO_ERROR)
+    {
+      return error_code;
+    }
+
+  error_code = au_grant (Au_public_user, class_mop, AU_SELECT, false);
```
hyunikn, line 6966: Every user will see the same query result from ALL_OBJECTS as that which DBA users see from DBA_OBJECTS, and DBA users will see the same result from ALL_OBJECTS and DBA_OBJECTS.
DBA_OBJECTS seems to be redundant. Isn't it?

src/transaction/boot_cl.c
hgryoo: IMO, along with partition tables, The auto_increment information also should be added in system catalog too.

src/transaction/boot_cl.c
ctshim: I also agree with hyunikn.

hgryoo: I agree with your opinion. I realize that our catalogs don't need such a 'dba_' system table. I'm going to leave only all_objects table and check current user is dba group in the WHERE claus of the query spec. Also, objects that the current user can be accessed will be found by OR predicate in the WHERE conditin like the other system views.

hyunikn (approved):

ctshim (approved):

PR #4701, opened 2023-09-20
[CBRD-24965] Typing fails for expressions of the form <date/time> +/- NULL and NULL +/- <date/time>

hyunikn:
http://jira.cubrid.org/browse/CBRD-24965


hgryoo:
LGTM.
I've tested the following too. It works well

```
select date'2023-10-10' + NULL;
select  date'2023-10-10' - NULL;

create or replace procedure poo as
    a date := date'2023-10-10' + NULL;
begin
    DBMS_OUTPUT.put_line (a);
end;

create or replace procedure poo2 as
    a date := date'2023-10-10' - NULL;
begin
    DBMS_OUTPUT.put_line (a);
end;

;server-output on
call poo (); // expected: no output
call poo2 (); // expected: no output
```

hgryoo (approved):

PR #4705, opened 2023-09-22
[CBRD-24921] Refactor system classes to use the new API

hgryoo:

On the top of #4604.

- Remove `schema_system_catalog_install.hpp`, Only include `schema_system_catalog.hpp` to use system catalog APIs
- `boot_add_collations()`, `boot_add_data_type()`, boot_add_charsets()` are moved into `schema_system_catalog_install.cpp`.
    - They are renamed to have `catclas_` prefix.
    - `util_sa.c` that using the `boot_add_collations()` to include `schema_system_catalog.hpp` instead of `boot_cl.h`
- Add `attribute_kind` enum class: Columns, Query specs, class methods are dealt with `class attribute`.  Depending on this `attribute_kind`, they are treated as different routines in the catalog builder. (for flexibility).
   - COLUMN, CLASS_METHOD are used for class
   - QUERY_SPEC is used for vclass

- Migrate remaining legacy routines creating system classes
   - Please see `schema_system_catalog_install.cpp` as separate view when you do code-review.
   - `_db_attribute, _db_domain, _db_method, _db_meth_sig, _db_meth_arg, _db_meth_file, _db_query_spec, _db_index, _db_index_key, _db_auth, _db_data_type, _db_stored_procedure, _db_stored_procedure_args, _db_partition, db_serial, db_ha_apply_info, _db_collation, _db_charset, dual, _db_server_, _db_synonym` are migrated.

Note that system vclasses (views) are refactored in the following PR.


hgryoo:
I found that APIs in `cnv.h` doesn't work and many of types are obsolete. That's why the NUMERIC default values are not properly created.

hyunikn (approved):

src/executables/util_sa.c
```diff
@@ -69,6 +68,7 @@
 #include "dbtype.h"
 #include "thread_manager.hpp"
 #include "log_volids.hpp"
+#include "schema_system_catalog.hpp"
```
beyondykk9, line 71: This header file looks empty. Is it necessary to include it?

beyondykk9 (approved):

src/object/schema_system_catalog_builder.cpp
```diff
@@ -111,26 +129,36 @@ namespace cubschema
...
+
 	int is_class_attribute = c.is_class_attributes ? 1 : 0;
-	if (type == DB_CONSTRAINT_INDEX)
+
+	switch (type)
+	  {
+	  case DB_CONSTRAINT_INDEX:
+	  case DB_CONSTRAINT_PRIMARY_KEY:
+	  case DB_CONSTRAINT_UNIQUE:
+	  case DB_CONSTRAINT_NOT_NULL:
```
ctshim, line 145: In the case of DB_CONSTRAINT_NOT_NULL, there was a check using do_check_rows_for_null() by calling the db_constrain_non_null() function, but is this no longer necessary?

ctshim, line 145: Are the following three cases not covered here?
- DB_CONSTRAINT_REVERSE_UNIQUE
- DB_CONSTRAINT_REVERSE_INDEX
- DB_CONSTRAINT_FOREIGN_KEY

hgryoo: Because this routine is creating a new system table, IMO, there is no reason to check that rows in the table has null. Because, in this routine, every constraint is added in empty system table, do_check_rows_for_null () performs `SELECT count(*) FROM tbl WHERE att IS NULL` unnecessarily.

hgryoo: Yes. If these cases are used in the future, they should be added. Since they have never been used until now, I didn't implement them.

ctshim (approved):

PR #4714, opened 2023-09-26
[CBRD-25008] A host variable in a SELECT list causes a semantic error

hyunikn:
http://jira.cubrid.org/browse/CBRD-25008

- If a column of a select list has DB_VARIABLE (14) type, then give it OBJECT type except for the case when it is a single host variable, in which case give it the type of the host variable.


hgryoo:
Please check test_medium failure. Maybe it would be fixed after giving it a try to merge with HEAD of feature/plcsql in upstream.

hgryoo (approved):

hyunikn:
> Please check test_medium failure. Maybe it would be fixed after giving it a try to merge with HEAD of feature/plcsql in upstream.

The error message 'Error: Could not find or load main class com.navercorp.cubridqa.ctp.IniCommand' of test_medium and test_sql seems related to CTP, not to this update. Let me see further.

PR #4719, opened 2023-10-04
[CBRD-25006] addition and subtraction on TIMESTAMP values produce a wrong value

hyunikn:
http://jira.cubrid.org/browse/CBRD-25006

- The initial plan to solve this problem was to use another Java type for Cubrid TIMESTAMP type and still use Java method overloading.
- However, the actual implementation is to give different names to methods implementing addition and subtraction of TIMESTAMP values, and still use java.sql.Timestamp type for Cubrid TIMESTAMP type.

### DIFF
--- a/pl_engine/pl_server/src/main/java/com/cubrid/plcsql/compiler/ast/ExprBetween.java
+++ b/pl_engine/pl_server/src/main/java/com/cubrid/plcsql/compiler/ast/ExprBetween.java
@@ -44,6 +44,12 @@ public class ExprBetween extends Expr {
     public final Expr lowerBound;
     public final Expr upperBound;
 
+    public boolean forTimestampParam;
+
+    public void setForTimestampParam() {
+        forTimestampParam = true;
+    }
+
     public ExprBetween(ParserRuleContext ctx, Expr target, Expr lowerBound, Expr upperBound) {
         super(ctx);
 
--- a/pl_engine/pl_server/src/main/java/com/cubrid/plcsql/predefined/sp/SpLib.java
+++ b/pl_engine/pl_server/src/main/java/com/cubrid/plcsql/predefined/sp/SpLib.java
@@ -746,13 +746,23 @@ public class SpLib {
         throw new PROGRAM_ERROR(); // unreachable
     }
     public static Boolean opNullSafeEqTimestamp(Timestamp l, Timestamp r) {
-        if (l == null || r == null) {
-            return null;
-        }
-        assert l.getNanos() == 0;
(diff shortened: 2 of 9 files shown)


pl_engine/pl_server/src/main/java/com/cubrid/plcsql/compiler/ast/ExprBetween.java
```diff
@@ -44,6 +44,12 @@ public <R> R accept(AstVisitor<R> visitor) {
     public final Expr lowerBound;
     public final Expr upperBound;
 
+    public boolean forTimestampParam;
```
hgryoo, line 47:
```suggestion
    public boolean forTimestampParam = false;
```

hyunikn: Default value of boolean values is false. So, the above two lines have the same behavior.

pl_engine/pl_server/src/main/java/com/cubrid/plcsql/predefined/sp/SpLib.java
```diff
@@ -664,6 +664,16 @@ public static Boolean opEq(ZonedDateTime l, ZonedDateTime r) {
         throw new PROGRAM_ERROR(); // unreachable
     }
 
+    public static Boolean opEqTimestamp(Timestamp l, Timestamp r) {
+        if (l == null || r == null) {
+            return null;
+        }
```
hgryoo, line 670: It will be checked in commonOpEq.

pl_engine/pl_server/src/main/java/com/cubrid/plcsql/predefined/sp/SpLib.java
```diff
@@ -813,6 +843,16 @@ public static Boolean opNeq(ZonedDateTime l, ZonedDateTime r) {
         throw new PROGRAM_ERROR(); // unreachable
     }
 
+    public static Boolean opNeqTimestamp(Timestamp l, Timestamp r) {
+        if (l == null || r == null) {
+            return null;
+        }
```
hgryoo, line 849: ditto

hgryoo (approved):

pl_engine/pl_server/src/main/java/com/cubrid/plcsql/predefined/sp/SpLib.java
hyunikn: OK. Thank you.

PR #4720, opened 2023-10-04
[LETS-750] [LETS-781] Append received pages during catch-up

wrlawodms:
http://jira.cubrid.org/browse/LETS-750

http://jira.cubrid.org/browse/LETS-781

- Call `logpb_catchup_append_page()` for each received page and call `logpb_catchup_finish()` at the end.
  - `logpb_catchup_append_page()`: appends a page. This creates a new page except for the first page. This sets `prev_lsa` and `append_lsa` to valid ones. They don't need to be precise. It's enough to set them to existing LSA and advance them constantly when flushing or archiving happens. It is just that log records after them are ignored.
  - `logpb_catchup_finish()`: searches log pages and sets the `prev_lsa` and `append_lsa` to proper ones to accept new log records.
- `logpb_catchup_start_append()`, `logpb_catchup_end_append()`: set the partial appending status properly before and after appending a page.
  - The thing is that it can happen to flush log pages and archive while appending log pages. We must set `prev_lsa`, `append_lsa`, and `log_Pb.partial_append.status` properly for them to work well. It's described in the comment for the functions about the status.
  - Basically, the `log_Pb.partial_append.status` during the catch-up has the same meaning as the partial appending of a big log record since it happens in the catchup also when a big log record ranges several pages. The `log_Pb.partial_append.status` transition in the catch-up.:
```
┌──► ┌─►_SUCCESS
│    │      │
│    │ (*)  │ start appending a page
│    │      ▼
│    └──_IN_PROGRESS
│           │
│           │ It happens to flush log pages during the catch-up.
│           │ so EOF is appended at the prev_lsa.
│           ▼
│       _PARTIAL_FLUSHED_END_OF_LOG
│           │
│           │ It's done to append log pages containing an incomplete log record.
│           │ If there is no incomplete log record, just a page has been appended.
│           ▼
│       _PARTIAL_ENDED
│           │
│           │ flush log pages to remove the EOF.
│           │ It appends the EOF at the append_lsa instead.
│           ▼
│       _PARTIAL_FLUSHED_ORIGINAL
│           │
│           │ Remove the EOF at the append_lsa on the memory buffer.
└───────────┘ This will be flushed in the next time.

(*): It's done to append log pages containing an incomplete log record without flushing in the middle.

```

Note:
- Two issues are done together. I was going to split them, but it turns out that most of what's for LETS-781 is also needed for LETS-750. Mainly, the change of the partial appending status is needed for LETS-750 as well because of flushing while adding a page.
- It's beyond the scope to describe all behavior of the partial append. We can have a call to share how they behave and more detailed design discussion.

### DIFF
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -655,6 +655,7 @@ void page_server::pts_mvcc_tracker::update_oldest_active_mvccid (const std::stri
   er_log_debug (ARG_FILE_LINE, ss.str ().c_str ());
 #endif
 }
+
 void page_server::pts_mvcc_tracker::delete_oldest_active_mvccid (const std::string &pts_channel_id)
 {
   std::lock_guard<std::mutex> lockg { m_pts_oldest_active_mvccids_mtx };
@@ -995,6 +996,12 @@ page_server::execute_catchup (cubthread::entry &entry, const LOG_LSA catchup_lsa
   size_t request_page_cnt = std::min (log_pgptr_recv_vec.size (), remaining_page_cnt);
   auto req_future = std::async (std::launch::async, request_pages_to_buffer, request_start_pageid, request_page_cnt);
   int error = NO_ERROR;
+
+  LOG_CS_ENTER (&entry);
+
+  logpb_flush_pages_direct (&entry);
+  logpb_flush_header (&entry);
+
   while (remaining_page_cnt > 0)
     {
       error = req_future.get (); // waits until the previous requests are replied
@@ -1014,15 +1021,22 @@ page_server::execute_catchup (cubthread::entry &entry, const LOG_LSA catchup_lsa
 	  req_future = std::async (std::launch::async, request_pages_to_buffer, request_start_pageid, request_page_cnt);
 	}
       // TODO append pages in log_pgptr_vec to the log buffer while pulling next pages.
+      for (size_t i = 0; i < request_page_cnt; i++)
+	{
+	  logpb_append_catchup_page (&entry, log_pgptr_vec[i]);
+	}
     }
 
   if (error == NO_ERROR)
     {
       assert (remaining_page_cnt == 0);
+
+      logpb_end_catchup (&entry, catchup_lsa);
+
       _er_log_debug (ARG_FILE_LINE, "[CATCH_UP] The catch-up is completed, ranging from %lld to %lld, count = %lld.\n",
 		     start_pageid, end_pageid, total_page_count);
       // TODO: send CATCHUP_DONE_MSG to the ATS
-      // TODO: start appneding log prior nodes from the ATS.
+      // TODO: start appending log prior nodes from the ATS.
 
       constexpr bool with_disc_msg = true;
       disconnect_followee_page_server (with_disc_msg);
@@ -1035,6 +1049,8 @@ page_server::execute_catchup (cubthread::entry &entry, const LOG_LSA catchup_lsa
       constexpr bool with_disc_msg = false;
       disconnect_followee_page_server (with_disc_msg);
     }
+
+  LOG_CS_EXIT (&entry);
 }
 
 bool
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -1025,7 +1025,7 @@ page_server::execute_catchup (cubthread::entry &entry, const LOG_LSA catchup_lsa
       // TODO append pages in log_pgptr_vec to the log buffer while pulling next pages.
       for (size_t i = 0; i < receive_page_cnt; i++)
 	{
-	  logpb_append_catchup_page (&entry, log_pgptr_vec[i]);
+	  logpb_catchup_append_page (&entry, log_pgptr_vec[i]);
 	}
     }
 
@@ -1033,7 +1033,7 @@ page_server::execute_catchup (cubthread::entry &entry, const LOG_LSA catchup_lsa
     {
       assert (remaining_page_cnt == 0);
 
-      logpb_finish_catchup (&entry, catchup_lsa);
+      logpb_catchup_finish (&entry, catchup_lsa);
 
       _er_log_debug (ARG_FILE_LINE, "[CATCH_UP] The catch-up is completed, ranging from %lld to %lld, count = %lld.\n",
 		     start_pageid, end_pageid, total_page_count);
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -1176,7 +1176,7 @@ page_server::execute_catchup (cubthread::entry &entry, const LOG_LSA catchup_lsa
 	}
     }
 
-  bool with_disc_msg;
+  bool with_disc_msg = true;
   if (error == NO_ERROR)
     {
       assert (remaining_page_cnt == 0);
@@ -1195,8 +1195,7 @@ page_server::execute_catchup (cubthread::entry &entry, const LOG_LSA catchup_lsa
 
       push_request_to_active_tran_server (page_to_tran_request::SEND_CATCHUP_COMPLETE, std::string());
 
-      constexpr bool with_disc_msg = true;
-      disconnect_followee_page_server (with_disc_msg);
+      with_disc_msg = true;
     }
   else
     {
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -998,6 +998,10 @@ page_server::execute_catchup (cubthread::entry &entry, const LOG_LSA catchup_lsa
   int error = NO_ERROR;
 
   LOG_CS_ENTER (&entry);
+  auto unlock_log_cs_on_exit = scope_exit {[&entry] ()
+  {
+    LOG_CS_EXIT (&entry);
+  }};
 
   while (remaining_page_cnt > 0)
     {
@@ -1052,8 +1056,6 @@ page_server::execute_catchup (cubthread::entry &entry, const LOG_LSA catchup_lsa
     }
 
   disconnect_followee_page_server (with_disc_msg);
-
-  LOG_CS_EXIT (&entry);
 }
 
 bool
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -1034,7 +1034,7 @@ page_server::execute_catchup (cubthread::entry &entry, const LOG_LSA catchup_lsa
       error = logpb_catchup_finish (&entry, catchup_lsa);
       if (error != NO_ERROR)
 	{
-	  // not reachable.
+	  // unacceptable.
 	  assert_release (false);
 	}
 
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -994,6 +994,7 @@ page_server::execute_catchup (cubthread::entry &entry, const LOG_LSA catchup_lsa
   LOG_PAGEID request_start_pageid = start_pageid;
   size_t remaining_page_cnt = total_page_count;
   size_t request_page_cnt = std::min (log_pgptr_recv_vec.size (), remaining_page_cnt);
+  size_t receive_page_cnt = 0;
   auto req_future = std::async (std::launch::async, request_pages_to_buffer, request_start_pageid, request_page_cnt);
   int error = NO_ERROR;
 
@@ -1010,8 +1011,9 @@ page_server::execute_catchup (cubthread::entry &entry, const LOG_LSA catchup_lsa
 	  break;
 	}
 
-      remaining_page_cnt -= request_page_cnt;
-      request_start_pageid += request_page_cnt;
+      receive_page_cnt = request_page_cnt;
+      remaining_page_cnt -= receive_page_cnt;
+      request_start_pageid += receive_page_cnt;
 
       log_pgptr_vec.swap (log_pgptr_recv_vec);
 
@@ -1021,7 +1023,7 @@ page_server::execute_catchup (cubthread::entry &entry, const LOG_LSA catchup_lsa
 	  req_future = std::async (std::launch::async, request_pages_to_buffer, request_start_pageid, request_page_cnt);
 	}
       // TODO append pages in log_pgptr_vec to the log buffer while pulling next pages.
-      for (size_t i = 0; i < request_page_cnt; i++)
+      for (size_t i = 0; i < receive_page_cnt; i++)
 	{
 	  logpb_append_catchup_page (&entry, log_pgptr_vec[i]);
 	}
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -1022,7 +1022,7 @@ page_server::execute_catchup (cubthread::entry &entry, const LOG_LSA catchup_lsa
 	  request_page_cnt = std::min (log_pgptr_recv_vec.size (), remaining_page_cnt);
 	  req_future = std::async (std::launch::async, request_pages_to_buffer, request_start_pageid, request_page_cnt);
 	}
-      // TODO append pages in log_pgptr_vec to the log buffer while pulling next pages.
+
       for (size_t i = 0; i < receive_page_cnt; i++)
 	{
 	  logpb_catchup_append_page (&entry, log_pgptr_vec[i]);
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -998,7 +998,6 @@ page_server::execute_catchup (cubthread::entry &entry, const LOG_LSA catchup_lsa
   int error = NO_ERROR;
 
   LOG_CS_ENTER (&entry);
-  logpb_flush_pages_direct (&entry);
 
   while (remaining_page_cnt > 0)
     {
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -1026,28 +1026,33 @@ page_server::execute_catchup (cubthread::entry &entry, const LOG_LSA catchup_lsa
 	}
     }
 
+  bool with_disc_msg;
   if (error == NO_ERROR)
     {
       assert (remaining_page_cnt == 0);
 
-      logpb_catchup_finish (&entry, catchup_lsa);
+      error = logpb_catchup_finish (&entry, catchup_lsa);
+      if (error != NO_ERROR)
+	{
+	  // not reachable.
+	  assert_release (false);
+	}
 
       _er_log_debug (ARG_FILE_LINE, "[CATCH_UP] The catch-up is completed, ranging from %lld to %lld, count = %lld.\n",
 		     start_pageid, end_pageid, total_page_count);
       // TODO: send CATCHUP_DONE_MSG to the ATS
       // TODO: start appending log prior nodes from the ATS.
-
-      constexpr bool with_disc_msg = true;
-      disconnect_followee_page_server (with_disc_msg);
+      with_disc_msg = true;
     }
   else
     {
       assert (remaining_page_cnt > 0);
       _er_log_debug (ARG_FILE_LINE, "[CATCH_UP] The catch-up stops with the error: %d. remainder: %lld total = %lld.\n",
 		     error, remaining_page_cnt, total_page_count);
-      constexpr bool with_disc_msg = false;
-      disconnect_followee_page_server (with_disc_msg);
+      with_disc_msg = false;
     }
+
+  disconnect_followee_page_server (with_disc_msg);
 
   LOG_CS_EXIT (&entry);
 }
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -1049,6 +1049,13 @@ page_server::execute_catchup (cubthread::entry &entry, const LOG_LSA catchup_lsa
     }
   else
     {
+      /*
+       * TODO
+       * In this case, we are going to re-initiate the catch-up with another PS.
+       * Now, we asuume it doesn't happen.
+       */
+      assert (false);
+
       assert (remaining_page_cnt > 0);
       _er_log_debug (ARG_FILE_LINE, "[CATCH_UP] The catch-up stops with the error: %d. remainder: %lld total = %lld.\n",
 		     error, remaining_page_cnt, total_page_count);
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -962,8 +962,8 @@ page_server::execute_catchup (cubthread::entry &entry, const LOG_LSA catchup_lsa
   const LOG_PAGEID end_pageid = catchup_lsa.offset == 0 ? catchup_lsa.pageid - 1 : catchup_lsa.pageid;
   const size_t total_page_count = end_pageid - start_pageid + 1;
 
-  // Request pages to the followee
-  auto request_pages_to_buffer = [this, &log_pgptr_recv_vec] (LOG_PAGEID start_pageid, size_t request_page_cnt) -> int
+  // A functor to request pages to the followee
+  auto request_pages_to_buffer = [this, &log_pgptr_recv_vec] (LOG_PAGEID start_pageid, size_t request_page_cnt)
   {
     auto lockg = std::lock_guard <std::mutex> (m_followee_conn_mutex);
     if (m_followee_conn == nullptr)
@@ -994,14 +994,11 @@ page_server::execute_catchup (cubthread::entry &entry, const LOG_LSA catchup_lsa
   LOG_PAGEID request_start_pageid = start_pageid;
   size_t remaining_page_cnt = total_page_count;
   size_t request_page_cnt = std::min (log_pgptr_recv_vec.size (), remaining_page_cnt);
-  size_t receive_page_cnt = 0;
   auto req_future = std::async (std::launch::async, request_pages_to_buffer, request_start_pageid, request_page_cnt);
   int error = NO_ERROR;
 
   LOG_CS_ENTER (&entry);
-
   logpb_flush_pages_direct (&entry);
-  logpb_flush_header (&entry);
 
   while (remaining_page_cnt > 0)
     {
@@ -1011,7 +1008,7 @@ page_server::execute_catchup (cubthread::entry &entry, const LOG_LSA catchup_lsa
 	  break;
 	}
 
-      receive_page_cnt = request_page_cnt;
+      auto receive_page_cnt = request_page_cnt;
       remaining_page_cnt -= receive_page_cnt;
       request_start_pageid += receive_page_cnt;
 
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -1033,7 +1033,7 @@ page_server::execute_catchup (cubthread::entry &entry, const LOG_LSA catchup_lsa
     {
       assert (remaining_page_cnt == 0);
 
-      logpb_end_catchup (&entry, catchup_lsa);
+      logpb_finish_catchup (&entry, catchup_lsa);
 
       _er_log_debug (ARG_FILE_LINE, "[CATCH_UP] The catch-up is completed, ranging from %lld to %lld, count = %lld.\n",
 		     start_pageid, end_pageid, total_page_count);
--- a/src/transaction/log_page_buffer.c
+++ b/src/transaction/log_page_buffer.c
@@ -3416,12 +3416,12 @@ logpb_catchup_append_page (THREAD_ENTRY * thread_p, const LOG_PAGE * const log_p
  *   
  *       ┌──► ┌─►_SUCCESS
  *       │    │      │
- *       │    │ (*)  │ start appending a page
+ *       │    │ (*)  │ Start appending a page
  *       │    │      ▼
  *       │    └──_IN_PROGRESS
  *       │           │
  *       │           │ It happens to flush log pages during the catch-up.
- *       │           │ so EOF is appended at the prev_lsa.
+ *       │           │ So, EOF is appended at the prev_lsa.
  *       │           ▼
  *       │       _PARTIAL_FLUSHED_END_OF_LOG
  *       │           │
@@ -3430,7 +3430,7 @@ logpb_catchup_append_page (THREAD_ENTRY * thread_p, const LOG_PAGE * const log_p
  *       │           ▼
  *       │       _PARTIAL_ENDED
  *       │           │
- *       │           │ flush log pages to remove the EOF.
+ *       │           │ Flush log pages to remove the EOF.
  *       │           │ It appends the EOF at the append_lsa instead.
  *       │           ▼
  *       │       _PARTIAL_FLUSHED_ORIGINAL
--- a/src/transaction/log_page_buffer.c
+++ b/src/transaction/log_page_buffer.c
@@ -3583,7 +3583,7 @@ logpb_catchup_finish (THREAD_ENTRY * thread_p, const LOG_LSA catchup_lsa)
 
       nav_lsa = log_rec->forw_lsa;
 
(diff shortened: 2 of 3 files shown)


src/server/page_server.cpp
```diff
@@ -995,6 +996,10 @@ page_server::execute_catchup (cubthread::entry &entry, const LOG_LSA catchup_lsa
   size_t request_page_cnt = std::min (log_pgptr_recv_vec.size (), remaining_page_cnt);
   auto req_future = std::async (std::launch::async, request_pages_to_buffer, request_start_pageid, request_page_cnt);
   int error = NO_ERROR;
+
+  LOG_CS_ENTER (&entry);
+  logpb_flush_pages_direct (&entry);
```
hornetmj: What is the purpose of calling logpb_flush_pages_direct() at this point?

src/transaction/log_page_buffer.c
```diff
@@ -3361,6 +3364,250 @@ logpb_write_toflush_pages_to_archive (THREAD_ENTRY * thread_p)
...
+}
+
+/*
+ * logpb_catchup_start_append - Start appending a page during catch-up.
+ *      LOGPB_APPENDREC_SUCCESS: 
+ *        A normal case. The previous appending has been successfully.
+ *      LOGPB_APPENDREC_IN_PROGRESS:
+ *        The previous appended page doesn't have a start of a record. It only constains a middle part 
+ *        or the last part of a record. In this case, the prev_lsa has not been advanced, 
+ *        and the EOF will be appended when flushing log pages halfway through appending a page.
```
hornetmj:
```suggestion
 *      LOGPB_APPENDREC_IN_PROGRESS:
 *        The previous appended page doesn't have a start of a log record. It only constains a middle part 
 *        or the last part of a log record. In this case, the prev_lsa has not been advanced, 
 *        and the EOF will be appended when flushing log pages halfway through appending a page.
```

src/transaction/log_page_buffer.c
```diff
@@ -3361,6 +3364,250 @@ logpb_write_toflush_pages_to_archive (THREAD_ENTRY * thread_p)
...
+ *      LOGPB_APPENDREC_PARTIAL_FLUSHED_END_OF_LOG:
+ *        Log pages are flushed during appending a page including only a partial record, and the temporary EOF
```
hornetmj:
```suggestion
 *        Log pages are flushed during appending a page including only a partial log record, and the temporary EOF 
```

src/transaction/log_page_buffer.c
```diff
@@ -3361,6 +3364,250 @@ logpb_write_toflush_pages_to_archive (THREAD_ENTRY * thread_p)
...
+ * logpb_catchup_start_append - Start appending a page during catch-up.
+ *      LOGPB_APPENDREC_SUCCESS: 
+ *        A normal case. The previous appending has been successfully.
+ *      LOGPB_APPENDREC_IN_PROGRESS:
+ *        The previous appended page doesn't have a start of a record. It only constains a middle part 
+ *        or the last part of a record. In this case, the prev_lsa has not been advanced, 
+ *        and the EOF will be appended when flushing log pages halfway through appending a page.
+ *      LOGPB_APPENDREC_PARTIAL_FLUSHED_END_OF_LOG:
+ *        Log pages are flushed during appending a page including only a partial record, and the temporary EOF 
+ *        has been appended at the prev_lsa. We will remove this at logpb_catchup_end_append() when the partial record completes.
```
hornetmj:
```suggestion
 *        has been appended at the prev_lsa. We will remove this at logpb_catchup_end_append() when the partial log record completes.
```

src/transaction/log_page_buffer.c
```diff
@@ -3361,6 +3364,250 @@ logpb_write_toflush_pages_to_archive (THREAD_ENTRY * thread_p)
...
+ *      The status transition:
+ *   
+ *       ┌──► ┌─►_SUCCESS
+ *       │    │      │
+ *       │    │ (*)  │ start appending a page
+ *       │    │      ▼
+ *       │    └──_IN_PROGRESS
+ *       │           │
+ *       │           │ It happens to flush log pages during the catch-up.
+ *       │           │ so EOF is appended at the prev_lsa.
```
hornetmj:
```suggestion
 *       │           │ So, EOF is appended at the prev_lsa.
```

src/transaction/log_page_buffer.c
```diff
@@ -3361,6 +3364,250 @@ logpb_write_toflush_pages_to_archive (THREAD_ENTRY * thread_p)
...
+ * (*): It's done to append log pages containing an incomplete log record without flushing in the middle.
+ * 
+ *      See logpb_catchup_end_append() and logpb_flush_all_append_pages() as well for the complete picture.
+ *     
+ *   log_pgptr(in): a log page pointer to append
+ */
+static void
+logpb_catchup_start_append (THREAD_ENTRY * thread_p)
+{
+  assert (LOG_CS_OWN_WRITE_MODE (thread_p));
```
hornetmj, line 3452: It is already checked in the logpb_catchup_append_page().

src/transaction/log_page_buffer.c
```diff
@@ -3361,6 +3364,250 @@ logpb_write_toflush_pages_to_archive (THREAD_ENTRY * thread_p)
...
+      {
+	LOG_RECORD_HEADER *origin_append_log_record = LOG_GET_LOG_RECORD_HEADER (log_pgptr, &log_Gl.hdr.append_lsa);
+
+	// The temporary EOF log record at the prev_lsa has to be removed.
+	log_Pb.partial_append.status = LOGPB_APPENDREC_PARTIAL_ENDED;
+	logpb_flush_all_append_pages (thread_p);
+	assert (log_Pb.partial_append.status == LOGPB_APPENDREC_PARTIAL_FLUSHED_ORIGINAL);
+
+	// A new EOF has been appended at the append_lsa in logpb_flush_all_append_pages. Remove it as well.
+	LOG_RECORD_HEADER *append_log_rec = LOG_GET_LOG_RECORD_HEADER (log_Gl.append.log_pgptr, &log_Gl.hdr.append_lsa);
```
hornetmj: memcpy (log_Gl.append.log_pgptr, log_pgptr, LOG_PAGESIZE) is executed before calling this function. So, I think that log_Gl.append.log_pgptr has same data with log_pgptr.

At 3516 line, origin_append_log_record is copied into append_log_rec. This seems like a meaningness behavior.

In my opinion, it would be good to first conduct an off-line review focusing on the partial append mechanism.

src/transaction/log_page_buffer.c
```diff
@@ -3361,6 +3364,250 @@ logpb_write_toflush_pages_to_archive (THREAD_ENTRY * thread_p)
...
+	    }
+	}
+
+      // If offset is missing, it is because we archive an incomplete log record.
+      if (nav_lsa.offset == NULL_OFFSET)
+	{
+	  nav_lsa.offset = log_pgptr->hdr.offset;
+	  if (nav_lsa.offset == NULL_OFFSET)
+	    {
+	      // There is nothing in thie page. Here is a part of incomplete log record.
```
hornetmj:
```suggestion
	      // There is nothing in this page. Here is a part of incomplete log record.
```

src/transaction/log_page_buffer.c
```diff
@@ -3361,6 +3364,250 @@ logpb_write_toflush_pages_to_archive (THREAD_ENTRY * thread_p)
...
+	    }
+	}
+
+      prev_lsa = nav_lsa;
+
+      LOG_RECORD_HEADER *log_rec = LOG_GET_LOG_RECORD_HEADER (log_pgptr, &prev_lsa);
+
+      nav_lsa = log_rec->forw_lsa;
+
+      if (nav_lsa.is_null () && logpb_is_page_in_archive (nav_lsa.pageid))
```
hornetmj: if nav_lsa == NULL, then logpb_is_page_in_archive() always returns true.

So, I don't understand the purpose of this line.

hornetmj: What is the meaning of the prefix nav_?

src/transaction/log_page_buffer.c
wrlawodms: done.

src/transaction/log_page_buffer.c
wrlawodms: I did it intentionally as other does like `logpb_prior_lsa_append_all_list(`) -> `logpb_append_prior_lsa_list()`. I think it's worth it in case the `logpb_catchup_append_page` is changed or this function is used in others.

src/server/page_server.cpp
wrlawodms: I don't think it's necessary. Removed.

src/transaction/log_page_buffer.c
wrlawodms: In `logpb_flush_all_append_pages` with the status of `LOGPB_APPENDREC_PARTIAL_ENDED`, an EOF is appended at the `log_Gl.hdr.append_lsa`. It is mainly to remove the previous EOF temporarily appended at the `prev_lsa`, but we always need `EOF`, so it's then at the `append_lsa`. We should remove it in the memory buffer, and it will overwrite the EOF when it's flushed.

src/transaction/log_page_buffer.c
wrlawodms: > What is the meaning of the prefix nav_?

I guess it stands for `navigation`. I get it from other traversing code in recovery.

wrlawodms: > So, I don't understand the purpose of this line.

Fixed. It must've been 
```
if (nav_lsa.is_null () && logpb_is_page_in_archive (prev_lsa.pageid))
```

I follow other traversing codes in which it assumes that an incomplete log record can be archived and the forward lsa is NULL_LSA in this case. See `log_recovery_find_first_postpone()`. 

However, I'm not sure about this case because what I see is that an incomplete log record header is not archived. I follow this, though, with "better safe than sorry".

hornetmj (approved):
LGTM

src/server/page_server.cpp
```diff
@@ -1014,27 +1019,41 @@ page_server::execute_catchup (cubthread::entry &entry, const LOG_LSA catchup_lsa
 	  req_future = std::async (std::launch::async, request_pages_to_buffer, request_start_pageid, request_page_cnt);
 	}
       // TODO append pages in log_pgptr_vec to the log buffer while pulling next pages.
```
```suggestion
```

src/transaction/log_page_buffer.c
```diff
@@ -3361,6 +3364,250 @@ logpb_write_toflush_pages_to_archive (THREAD_ENTRY * thread_p)
...
+
+  log_Gl.append.prev_lsa = prev_lsa;
+  log_Gl.hdr.append_lsa = catchup_lsa;
+  log_Gl.get_log_prior_sender ().reset_unsent_lsa (catchup_lsa);
+  {
+    auto lockg = std::lock_guard { log_Gl.prior_info.prior_lsa_mutex };
+    log_Gl.prior_info.prev_lsa = prev_lsa;
+    log_Gl.prior_info.prior_lsa = catchup_lsa;
+  }
+
```
joohok, line 3605: It seems like it would be better to flush the log header after the catch-up is finished (header information has been updated).  `logpb_flush_header ()`

src/transaction/log_page_buffer.c
```diff
@@ -3361,6 +3364,250 @@ logpb_write_toflush_pages_to_archive (THREAD_ENTRY * thread_p)
...
+	{
+	  nav_lsa.pageid = prev_lsa.pageid + 1;
+	  nav_lsa.offset = NULL_OFFSET;
+	}
+    }
+
+  assert_release (nav_lsa == catchup_lsa);
+
+  log_Gl.append.prev_lsa = prev_lsa;
+  log_Gl.hdr.append_lsa = catchup_lsa;
```
joohok, line 3598: At the recovery phase, `eof_lsa` may be used rarely, but it is also used in emergency mode (log_manager.c:1426, log_initialize_internal()). Therefore, I think it would be good to update it with the accurate `append_lsa` value. Or, could it be updated elsewhere?
```suggestion
  log_Gl.hdr.append_lsa = catchup_lsa;
  log_Gl.hdr.eof_lsa = log_Gl.hdr.append_lsa;
```

src/server/page_server.cpp
```diff
@@ -995,6 +996,9 @@ page_server::execute_catchup (cubthread::entry &entry, const LOG_LSA catchup_lsa
   size_t request_page_cnt = std::min (log_pgptr_recv_vec.size (), remaining_page_cnt);
   auto req_future = std::async (std::launch::async, request_pages_to_buffer, request_start_pageid, request_page_cnt);
   int error = NO_ERROR;
+
+  LOG_CS_ENTER (&entry);
```
joohok, line 1147: I may not remember it well... 
but what does this critical section protect against?

wrlawodms: Honestly, I don't know exact lock scope of this critical section, but it's acquired when modifying some variables I modify in this operation and calling some function I used.
For example, ` log_Gl.hdr.append_lsa` and `logpb_flush_all_append_pages`. 

Additionally, I considered reducing the critical section by holding the locking only while appending, not receiving from the followee. But, I doubt it works because the state of the log_manager is not concluded while appending pages. For example, before the `logpb_catchup_finish()` which is the last call of the catch-up process, the `append_lsa` points to a middle of log records. If another thread flushes pages at this point, it will append EOF log record in the middle of log. It would happen if we release the lock halfway through the process.

src/transaction/log_page_buffer.c
wrlawodms: Could you tell me why we should keep the `eof_lsa` updated to the `append_lsa` here? What I understand is that `eof_lsa` is the location where the EOF log record is appended while flushing log pages. This will be updated while flushing, and here the `append_lsa` doesn't point to the EOL log record.

src/transaction/log_page_buffer.c
wrlawodms: Could you elaborate on why we should flush the header page here?

I considered it, but I can't find the necessity I have to flush the header page after appending pages. It could be safer, but I'm not sure it's necessary. Similarly, appending a log record may add a page and change the variables changed here, but it does neither flush the header page. So, I concluded that I can leave it to the log manager to flush the header page. That's my approach, I'd love it if you correct what I missed.

src/server/page_server.cpp
joohok: I apologize for the unclear question. I agree that LOG_CS_ENTER() is not very clear and appears to protect too many objects. Every time changes are made to the log buffer, log volume, or log_Gl, it enters a critical section.

I initially thought that replication wouldn't proceed while PS was catching up. I believed that only catch-up threads had access to the log buffer, log volume, and log_Gl. Therefore, I questioned the need to enter the critical section. However, if replication also occurs simultaneously, it seems appropriate to enter the critical section through LOG_CS_ENTER(). This is because replication involves accessing the log buffer/volume and obtaining pages while it proceeds.

src/transaction/log_page_buffer.c
joohok: As discussed yesterday, I had previously considered associating catch-up with recovery. In other words, I had thought about the scenario where if PS were to die after completing catch-up without performing a flush, catch-up would need to be done again. However, it seems that this should be viewed from an optimization perspective

src/transaction/log_page_buffer.c
joohok: I think I briefly got confused about this. I'm sorry

joohok (approved):

src/server/page_server.cpp
wrlawodms: No need to apologize. Actually, it brings up an issue I missed. It seemed that the replicator works well along with the catch-up mechanism in parallel without any modification. And, it did when I tested, but there are some cases it can't. The replicator conditionally acquires the lock such as fetching a log page from disk. In this case, the replicator gets stuck and waits for the catch-up to be done. I have to address this issue to make them work in parallel.

Regarding when to deal with this, this issue was made earlier as a TODO: http://jira.cubrid.org/browse/LETS-753. I was going to just close this, but it still has points. Let me address the problem in the issue.

src/transaction/log_page_buffer.c
```diff
@@ -3361,6 +3364,250 @@ logpb_write_toflush_pages_to_archive (THREAD_ENTRY * thread_p)
...
+ *
+ *   log_pgptr(in): a log page pointer appended
+ */
+static void
+logpb_catchup_end_append (THREAD_ENTRY * thread_p, const LOG_PAGE * const log_pgptr)
+{
+  assert (LOG_CS_OWN_WRITE_MODE (thread_p));
+
+  const bool have_only_partial_record_in_page = log_pgptr->hdr.offset == NULL_LOG_OFFSET;
+
```
cristiarg:
```suggestion
```

src/transaction/log_page_buffer.c
```diff
@@ -3361,6 +3364,250 @@ logpb_write_toflush_pages_to_archive (THREAD_ENTRY * thread_p)
...
+{
+  assert (LOG_CS_OWN_WRITE_MODE (thread_p));
+  assert (log_Gl.append.log_pgptr != NULL);
+
+  logpb_catchup_start_append (thread_p);
+
+  if (log_Gl.append.log_pgptr->hdr.logical_pageid + 1 == log_pgptr->hdr.logical_pageid)
+    {
+      // The normal case. Only exception is the first page. 
+      log_Gl.append.prev_lsa = log_Gl.hdr.append_lsa;
```
cristiarg, line 3384: Nit.
Does it have any value to add an assert in each `if` block as to the possible `log_Pb.partial_append.status` values? Or, if simpler, what values cannot be present in each case...

src/transaction/log_page_buffer.c
```diff
@@ -3361,6 +3364,250 @@ logpb_write_toflush_pages_to_archive (THREAD_ENTRY * thread_p)
...
+      logpb_next_append_page (thread_p, LOG_SET_DIRTY);
+    }
+  else if (log_Gl.append.log_pgptr->hdr.logical_pageid == log_pgptr->hdr.logical_pageid)
+    {
+      // For the first page, just overwrite the page.
+      // DON'T change meta data like log_Gl.append.prev_lsa. They don't go backward
+    }
+  else
+    {
+      logpb_fatal_error (thread_p, true, ARG_FILE_LINE, "logpb_catchup_append_page");
```
cristiarg, line 3396: Could you add some explanation for this branch?

My idea is this is a future development where the "this" PS tries to do cathc-up while it is actually ahead of the starting "catch-up LSA" and must discard some log pages.

src/server/page_server.cpp
```diff
@@ -1014,27 +1019,41 @@ page_server::execute_catchup (cubthread::entry &entry, const LOG_LSA catchup_lsa
...
+      with_disc_msg = true;
     }
   else
     {
       assert (remaining_page_cnt > 0);
       _er_log_debug (ARG_FILE_LINE, "[CATCH_UP] The catch-up stops with the error: %d. remainder: %lld total = %lld.\n",
 		     error, remaining_page_cnt, total_page_count);
-      constexpr bool with_disc_msg = false;
-      disconnect_followee_page_server (with_disc_msg);
+      with_disc_msg = false;
```
cristiarg, line 1212: What happens in this case?
Can "this" PS recover?
Or is "this" PS left in hanging state?
I suppose this should be a `logpb_fatal_error` ?

src/transaction/log_page_buffer.c
```diff
@@ -3361,6 +3364,250 @@ logpb_write_toflush_pages_to_archive (THREAD_ENTRY * thread_p)
...
+  assert (log_Gl.hdr.append_lsa.pageid == log_pgptr->hdr.logical_pageid);
+  log_Gl.hdr.append_lsa.offset = log_pgptr->hdr.offset;
+
+  LOG_RECORD_HEADER *log_rec = LOG_GET_LOG_RECORD_HEADER (log_pgptr, &log_Gl.hdr.append_lsa);
+  log_Gl.append.prev_lsa = log_rec->back_lsa;
+
+  switch (log_Pb.partial_append.status)
+    {
+    case LOGPB_APPENDREC_IN_PROGRESS:
+      /* success, fall through */
```
cristiarg: Nit.
```suggestion
      /* success */
```
In my understanding of c/c++ "fall-through" is where a `case` label does not have a `break`.

https://en.cppreference.com/w/cpp/language/attributes/fallthrough

src/transaction/log_page_buffer.c
```diff
@@ -3361,6 +3364,250 @@ logpb_write_toflush_pages_to_archive (THREAD_ENTRY * thread_p)
...
+  if (have_only_partial_record_in_page)
+    {
+      // Do nothing. It will be finished up at a following call.
+      return;
+    }
+
+  assert (log_Gl.hdr.append_lsa.pageid == log_pgptr->hdr.logical_pageid);
+  log_Gl.hdr.append_lsa.offset = log_pgptr->hdr.offset;
+
+  LOG_RECORD_HEADER *log_rec = LOG_GET_LOG_RECORD_HEADER (log_pgptr, &log_Gl.hdr.append_lsa);
```
cristiarg:
```suggestion
  const LOG_RECORD_HEADER *const log_rec = LOG_GET_LOG_RECORD_HEADER (log_pgptr, &log_Gl.hdr.append_lsa);
```

src/transaction/log_page_buffer.c
```diff
@@ -3361,6 +3364,250 @@ logpb_write_toflush_pages_to_archive (THREAD_ENTRY * thread_p)
...
+  log_Gl.append.prev_lsa = log_rec->back_lsa;
+
+  switch (log_Pb.partial_append.status)
+    {
+    case LOGPB_APPENDREC_IN_PROGRESS:
+      /* success, fall through */
+      break;
+    case LOGPB_APPENDREC_PARTIAL_FLUSHED_END_OF_LOG:
+      {
+	LOG_RECORD_HEADER *origin_append_log_record = LOG_GET_LOG_RECORD_HEADER (log_pgptr, &log_Gl.hdr.append_lsa);
```
cristiarg:
```suggestion
	const LOG_RECORD_HEADER *const origin_append_log_record = LOG_GET_LOG_RECORD_HEADER (log_pgptr, &log_Gl.hdr.append_lsa);
```
It has the same value as `const LOG_RECORD_HEADER *const log_rec` so you could actually do
```suggestion
	const LOG_RECORD_HEADER *const origin_append_log_record = log_rec;
```

src/transaction/log_page_buffer.c
```diff
@@ -3361,6 +3364,250 @@ logpb_write_toflush_pages_to_archive (THREAD_ENTRY * thread_p)
...
+
+	// The temporary EOF log record at the prev_lsa has to be removed.
+	log_Pb.partial_append.status = LOGPB_APPENDREC_PARTIAL_ENDED;
+	logpb_flush_all_append_pages (thread_p);
+	assert (log_Pb.partial_append.status == LOGPB_APPENDREC_PARTIAL_FLUSHED_ORIGINAL);
+
+	// A new EOF has been appended at the append_lsa in logpb_flush_all_append_pages. Remove it as well.
+	LOG_RECORD_HEADER *append_log_rec = LOG_GET_LOG_RECORD_HEADER (log_Gl.append.log_pgptr, &log_Gl.hdr.append_lsa);
```
cristiarg:
```suggestion
	LOG_RECORD_HEADER *const append_log_rec = LOG_GET_LOG_RECORD_HEADER (log_Gl.append.log_pgptr, &log_Gl.hdr.append_lsa);
```

src/server/page_server.cpp
cristiarg, line 1147: Could use a `scope_exit` here:
```
  LOG_CS_ENTER (thread_p);
  // *INDENT-OFF*
  scope_exit<std::function<void (void)>> unlock_log_cs_on_exit ([thread_p] ()
  {
    LOG_CS_EXIT (thread_p);
  });
  // *INDENT-ON*

```

src/transaction/log_page_buffer.c
```diff
@@ -3361,6 +3364,250 @@ logpb_write_toflush_pages_to_archive (THREAD_ENTRY * thread_p)
...
+      nav_lsa = log_rec->forw_lsa;
+
+      if (nav_lsa.is_null () && logpb_is_page_in_archive (prev_lsa.pageid))
+	{
+	  nav_lsa.pageid = prev_lsa.pageid + 1;
+	  nav_lsa.offset = NULL_OFFSET;
+	}
+    }
+
+  assert_release (nav_lsa == catchup_lsa);
```
cristiarg, line 3595: and?
```suggestion
  assert_release (prev_lsa < nav_lsa);
  assert_release (nav_lsa == catchup_lsa);
```

src/transaction/log_page_buffer.c
```diff
@@ -3361,6 +3364,250 @@ logpb_write_toflush_pages_to_archive (THREAD_ENTRY * thread_p)
...
+    default:
+      /* invalid state */
+      assert_release (false);
+    }
+  log_Pb.partial_append.status = LOGPB_APPENDREC_SUCCESS;
+}
+
+/*
+ * logpb_catchup_finish - finish the catch-up job. 
+ *      This sets the prev_lsa and append_lsa to the latest ones traversing log pages from known log record.
```
cristiarg, line 3532: Unfortunately I don't fully understand what this function does. Could you expand the comment a bit please?

From what I understood, it traverses the log records of either:
- the last added log page started from the first complete log record in that page
- or the last log record when that last log record is a big one spanning multiple log pages

and
 - retains the LSA of the last log record as the `log_Gl.append.prev_lsa`
 - while `log_Gl.hdr.append_lsa`/`catchup_lsa` is the LSA of the next log record (still not present) to be added

Is this correct?

I think you could have avoided this extra function with the price of passing a `log_catchup_context` pointer struct to the `logpb_catchup_append_page` and `logpb_catchup_end_append` functions and extracting the value from there.

Not necessary to change the code in this direction, but I am wondering if it is possible.

cristiarg (approved):

src/server/page_server.cpp
wrlawodms: Thanks: [5245eaa](https://github.com/CUBRID/cubrid/pull/4720/commits/5245eaa59a164c4f3abcbb69c88d422cfec45770)

src/server/page_server.cpp
wrlawodms: In the end, it is going to request the catchup from where it fails with another connection. 

Regarding recovery, it always can recover to the log record flushed. If all log records have been flushed before the crash, it's supposed to do recovery with the log records and start catch-up from there. It's TODO, now we assume no error happens during the catchup. It'll be addressed when the catch-up re-request comes in.

For now, I should add an assert and a TODO comment.

src/transaction/log_page_buffer.c
wrlawodms: Good Idea. In the case of the first page, only `LOGPB_APPENDREC_IN_PROGRESS` is allowed. In the other case, it can be all cases in `logpb_catchup_start_append()` and it has an assert inside for invalid states.

src/transaction/log_page_buffer.c
wrlawodms: Discarding log records beyond the Durable LSA is a separate issue, and it is a prerequisite task for catching up. Here, in the future, it's done to discard, and it appends pages one by one from the durable LSA.

About the conditions, let's assume the page id of the first page is 5.
1. We start fetching and appending pages from the page 5. It is the first case here represented as
`log_Gl.append.log_pgptr->hdr.logical_pageid == log_pgptr->hdr.logical_pageid`. 
2. After copying the received page, the `log_Gl.append.log_pgptr` is unchanged. 
3. And, then the next page is page 6. "This" ps allocates a new page and copies the received page 6 to it. From here, it always hits the condition above, `(log_Gl.append.log_pgptr->hdr.logical_pageid + 1 == log_pgptr->hdr.logical_pageid)`.

src/transaction/log_page_buffer.c
wrlawodms: Absolutely right,,

src/transaction/log_page_buffer.c
wrlawodms: hmm,  I don't think it's needed. It means that the forw_lsa of a log record points to itself. And this seems that is should be blocked not here, but where it's generated.

src/transaction/log_page_buffer.c
wrlawodms: Done. I agree. I just took the base of this code from `logpb_end_append()`.

src/transaction/log_page_buffer.c
wrlawodms: > the last added log page started from the first complete log record in that page
or the last log record when that last log record is a big one spanning multiple log pages

Correct.

> retains the LSA of the last log record as the log_Gl.append.prev_lsa
while log_Gl.hdr.append_lsa/catchup_lsa is the LSA of the next log record (still not present) to be added

I don't understand your words. Probably, we have a misunderstanding. Let me clarify what this function does (my intention).
After appending a page, we don't have precise append_lsa and prev_lsa. Our goal here is to make them have appropriate ones. Basically, they have values as the following which are set in `logpb_catchup_end_append()`:
- append_lsa: log_pgptr->hdr.offset, which is the location we can guarantee there is a log record. 
- prev_lsa: the prev_lsa of the log record append_lsa points to.

This is what has to be kept while appending pages, but it's not what they really mean. The `prev_lsa` is supposed to point to the last log record and the `append_lsa` is supposed to point to the next one, which has not been appended yet. So, here at the end of the whole catch-up process, they are set as they're supposed to be.

How: we start from the most advanced log record we can know, which the `prev_lsa` points to. The `append_lsa` is not reliable because if the last page has no `log_pgptr->hdr.offset`, it just points to the start of the page. And we just traverse log records until we met the last log record.

We can make them every time after appending a page, but we don't need to. So, I do it at the end only once.

src/transaction/log_page_buffer.c
cristiarg: I understood the other if/else scopes.
My comment was specifically about this `logpb_fatal_error`.
The way it crashes the application is both drastic and laconic - with just the `logpb_catchup_append_page` message which does not say much.

But also my "Could you add some explanation for this branch?" was cryptic because I meant to ask "Could you add some __comment__ for this branch?":
- commenting why the fatal error
- maybe changing te logged message to something more meaningful _ "... not yet implemented", "not possible to occur" ...
- or a TODO for the future (if the case)

src/transaction/log_page_buffer.c
cristiarg: > The prev_lsa is supposed to point to the last log record and the append_lsa is supposed to point to the next one, which has not been appended yet. So, here at the end of the whole catch-up process, they are set as they're supposed to be.

Ok. I knew [at least a part of] that.

> We can make them every time after appending a page, but we don't need to. So, I do it at the end only once.

Yes, it also crossed my mind that my suggestion to use a `log_catchup_context` pointer ends up with doing this investigation many times for each log page added.

cristiarg (approved):

src/transaction/log_page_buffer.c
wrlawodms: Oh, I see.

PR #4721, opened 2023-10-05
[LETS-749] Set aside prior nodes from ATS during the catch-up

wrlawodms:
http://jira.cubrid.org/browse/LETS-749

Start the prior_recv thread when the catch-up is done. Until it's started, the prior nodes from ATS are kept in its internal queue, just not being consumed.

Tested on #4720.

### DIFF
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -224,10 +224,14 @@ page_server::tran_server_connection_handler::receive_start_catch_up (tran_server
   _er_log_debug (ARG_FILE_LINE,
 		 "[CATCH_UP] It's been requested to start catch-up with the follower (%s:%d), until LSA = (%lld|%d)\n",
 		 host.c_str (), port,LSA_AS_ARGS (&catchup_lsa));
+
   if (port == -1)
     {
       // TODO: It means that the ATS is booting up.
       // it will be set properly after ATS recovery is implemented.
+
+      // TODO reply the catchup_done msg.
+      log_Gl.get_log_prior_receiver ().start_thread ();
       return;
     }
 
@@ -238,11 +242,12 @@ page_server::tran_server_connection_handler::receive_start_catch_up (tran_server
   if (log_Gl.hdr.append_lsa == catchup_lsa)
     {
       _er_log_debug (ARG_FILE_LINE, "[CATCH_UP] There is nothing to catch up.\n");
+      log_Gl.get_log_prior_receiver ().start_thread ();
       return; // TODO the cold-start case. No need to catch up. Just send a catchup_done msg to the ATS
     }
 
-  // Establish a connection with the PS to catch up with, and start the cathup asynchronously.
-  // The connection will be destoryed at the end of the catch-up.
+  // Establish a connection with the PS to catch up with, and start the catch-up asynchronously.
+  // The connection will be destroyed at the end of the catch-up.
   m_ps.connect_to_followee_page_server (std::move (host), port);
   m_ps.start_catchup (catchup_lsa);
 }
@@ -1023,6 +1028,8 @@ page_server::execute_catchup (cubthread::entry &entry, const LOG_LSA catchup_lsa
 		     start_pageid, end_pageid, total_page_count);
       // TODO: send CATCHUP_DONE_MSG to the ATS
       // TODO: start appneding log prior nodes from the ATS.
+
+      log_Gl.get_log_prior_receiver ().start_thread ();
 
       constexpr bool with_disc_msg = true;
       disconnect_followee_page_server (with_disc_msg);
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -1027,7 +1027,6 @@ page_server::execute_catchup (cubthread::entry &entry, const LOG_LSA catchup_lsa
       _er_log_debug (ARG_FILE_LINE, "[CATCH_UP] The catch-up is completed, ranging from %lld to %lld, count = %lld.\n",
 		     start_pageid, end_pageid, total_page_count);
       // TODO: send CATCHUP_DONE_MSG to the ATS
-      // TODO: start appneding log prior nodes from the ATS.
 
       log_Gl.get_log_prior_receiver ().start_thread ();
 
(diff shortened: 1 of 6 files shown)


wrlawodms:
The internal thread in `prior_recv` must stop only if it stats. I'll re-request review after it's done.

wrlawodms:
> The internal thread in prior_recv must stop only if it stats. I'll re-request review after it's done.

Done.

hornetmj (approved):

src/server/page_server.cpp
```diff
@@ -224,10 +224,14 @@ page_server::tran_server_connection_handler::receive_start_catch_up (tran_server
   _er_log_debug (ARG_FILE_LINE,
 		 "[CATCH_UP] It's been requested to start catch-up with the follower (%s:%d), until LSA = (%lld|%d)\n",
 		 host.c_str (), port,LSA_AS_ARGS (&catchup_lsa));
+
   if (port == -1)
     {
       // TODO: It means that the ATS is booting up.
       // it will be set properly after ATS recovery is implemented.
+
+      // TODO reply the catchup_done msg.
```
joohok: It appears that ATS is still booting, and there seems to be no information available for the main connection. In this situation, it doesn't seem advisable to request the prior node since catch-up hasn't finished. It appears that a catch-up retry is necessary. What do you think ?

Oh.. it can be addressed when ATS recovery is covered.

joohok (approved):

src/transaction/log_prior_recv.cpp
```diff
@@ -28,12 +28,14 @@ namespace cublog
   prior_recver::prior_recver (log_prior_lsa_info &prior_lsa_info)
     : m_prior_lsa_info (prior_lsa_info)
   {
-    start_thread ();
```
cristiarg, line 31: `prior_recver` is also used by PTS.
I am not mistaken, removing this from here breaks PTS since for PTS this thread is not started anywhere.
- the - now public - `prior_recver::start_thread` function must be called in `server_type` for PTS
- or, feel free to create a new `log_global::initialize_and_start_log_prior_receiver` that does both and call that one for PTS in `server_type`

src/transaction/log_prior_recv.hpp
```diff
@@ -53,13 +53,13 @@ namespace cublog
       prior_recver &operator = (prior_recver &&) = delete;
 
       void push_message (std::string &&str);                  // push message from prior_sender into message queue
+      void start_thread ();                                   // run loop_message_to_prior_info in a thread
```
cristiarg, line 56: Nit
Add a remark to the class comment that this class implementation is not thread-safe because it does not need to be.

src/server/page_server.cpp
```diff
@@ -224,10 +224,14 @@ page_server::tran_server_connection_handler::receive_start_catch_up (tran_server
...
+      log_Gl.get_log_prior_receiver ().start_thread ();
```
cristiarg, line 235: Nit.
I think that we must first start the thread and then reply with the catch-up done message - to avoid race conditions:
- PS returns the message
- ATS starts pushing already cached messages
- PS log prior receiver thread is not started

It practice it is probably impossible for this to happen.

wrlawodms: It's not the case because the `start_thread()` just starts consuming the log records and all log records from ATS may be already accumulated in the receiver's queue. So, the order doesn't count.

Regarding this, it could be more readable to change the name of the function from `start_thread()` to `start_consumer_thread()` or `start_appending_thread()`. What do you think?

src/transaction/log_prior_recv.cpp
wrlawodms: Definitely. I should've considered it.

Sorry, I've tended to miss PTS's basic functionality. To prevent it, I'll create and conduct a small test sets for it like starting/stopping servers and a basic replication behavior.

wrlawodms: done: [8bfe418](https://github.com/CUBRID/cubrid/pull/4721/commits/8bfe418493ce3d8d226b6f1359ca3ca7ae4b4fa0)

src/transaction/log_global.c
```diff
@@ -159,6 +159,19 @@ log_global::initialize_log_prior_receiver ()
 {
   assert (m_prior_recver == nullptr);
   m_prior_recver = std::make_unique<cublog::prior_recver> (prior_info);
+  
+  if (is_passive_transaction_server ())
```
cristiarg: I cannot say I like this approach :) .
This introduces the `is_..._server` conditional fork in yet another of the legacy files whereas there is already a fork in `server_type.cpp` file which directly calls this code.

If you do not prefer adding a new function ... then it's a one-liner in `init_server_type`:
```
...
	{
	  assert (pts_Gl == nullptr);

	  // passive tran server also needs (a) prior receiver(s) to receive log from page server(s)
	  // the mechanism is that of a pub sub; a PTS subscribes to receive log from one page server (for now)
	  log_Gl.initialize_log_prior_receiver ();
	  log_Gl.get_log_prior_receiver ().start_thread ();

	  pts_Gl = new passive_tran_server ();
	  ts_Gl.reset (pts_Gl);
	}
...
```

And for the corresponding PS block in `server_type` add a comment explaining it is done after the catch-up.

cristiarg (approved):
Approved with a strong suggestion.

src/server/page_server.cpp
cristiarg: > It's not the case because the start_thread() just starts consuming the log records and all log records from ATS may be already accumulated in the receiver's queue. So, the order doesn't count.

It is my understanding that the PS does not receive the log prior messages until the catch-up is done. But I think I'm wrong because in `active_tran_server::connection_handler::on_connecting` the log prior sink si registered before the catch-up process. So, my bad.

`log_Gl.get_log_prior_receiver ().start_thread ();` is enough for me, as it is already topological: prior received only contains a single thread.

cristiarg:
Build failed, but I think it is spurious:
```
/home/unit_tests/log_repl/test_main_log_repl_atomic_helper.cpp:20:10: fatal error: catch2/catch.hpp: No such file or directory
 #include "catch2/catch.hpp"
          ^~~~~~~~~~~~~~~~~~
compilation terminated.
gmake[2]: *** [unit_tests/log_repl/CMakeFiles/test_log_repl_atomic_helper.dir/build.make:121: unit_tests/log_repl/CMakeFiles/test_log_repl_atomic_helper.dir/test_main_log_repl_atomic_helper.cpp.o] Error 1
```

src/transaction/log_global.c
wrlawodms: Oh, you're right. I didn't consider the call path. Thank you.

PR #4722, opened 2023-10-05
[LETS-751] Notify to the ATS that the catch-up is completed

wrlawodms:
http://jira.cubrid.org/browse/LETS-751

### DIFF
--- a/src/server/active_tran_server.cpp
+++ b/src/server/active_tran_server.cpp
@@ -166,8 +166,11 @@ active_tran_server::connection_handler::get_request_handlers ()
 
   auto saved_lsa_handler = std::bind (&active_tran_server::connection_handler::receive_saved_lsa, this,
 				      std::placeholders::_1);
+  auto catchup_complete_handler = std::bind (&active_tran_server::connection_handler::receive_catchup_complete, this,
+				  std::placeholders::_1);
 
   handlers_map.insert (std::make_pair (page_to_tran_request::SEND_SAVED_LSA, saved_lsa_handler));
+  handlers_map.insert (std::make_pair (page_to_tran_request::SEND_CATCHUP_COMPLETE, catchup_complete_handler));
 
   return handlers_map;
 }
@@ -190,6 +193,13 @@ active_tran_server::connection_handler::receive_saved_lsa (page_server_conn_t::s
 
   quorum_consenesus_er_log ("Received saved LSA = %lld|%d from %s.\n", LSA_AS_ARGS (&saved_lsa),
 			    get_channel_id ().c_str ());
+}
+
+void
+active_tran_server::connection_handler::receive_catchup_complete (page_server_conn_t::sequenced_payload &&a_sp)
+{
+  _er_log_debug (ARG_FILE_LINE, "[CATCH_UP] the catchup has been completed. channel id: %s\n",
+		 get_channel_id ().c_str ());
 }
 
 void
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -228,6 +228,7 @@ page_server::tran_server_connection_handler::receive_start_catch_up (tran_server
     {
       // TODO: It means that the ATS is booting up.
       // it will be set properly after ATS recovery is implemented.
+      m_ps.push_request_to_active_tran_server (page_to_tran_request::SEND_CATCHUP_COMPLETE, std::string());
       return;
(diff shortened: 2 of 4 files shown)


src/server/page_server.cpp
```diff
@@ -237,7 +238,10 @@ page_server::tran_server_connection_handler::receive_start_catch_up (tran_server
 
   if (log_Gl.hdr.append_lsa == catchup_lsa)
     {
+
```
hornetmj:
```suggestion
```

src/server/page_server.cpp
```diff
@@ -237,7 +238,10 @@ page_server::tran_server_connection_handler::receive_start_catch_up (tran_server
...
       _er_log_debug (ARG_FILE_LINE, "[CATCH_UP] There is nothing to catch up.\n");
+
+      m_ps.push_request_to_active_tran_server (page_to_tran_request::SEND_CATCHUP_COMPLETE, std::string());
       return; // TODO the cold-start case. No need to catch up. Just send a catchup_done msg to the ATS
```
hornetmj:
```suggestion
      return;
```

hornetmj (approved):

src/server/page_server.cpp
```diff
@@ -1021,9 +1025,10 @@ page_server::execute_catchup (cubthread::entry &entry, const LOG_LSA catchup_lsa
       assert (remaining_page_cnt == 0);
       _er_log_debug (ARG_FILE_LINE, "[CATCH_UP] The catch-up is completed, ranging from %lld to %lld, count = %lld.\n",
 		     start_pageid, end_pageid, total_page_count);
-      // TODO: send CATCHUP_DONE_MSG to the ATS
       // TODO: start appneding log prior nodes from the ATS.
 
+      push_request_to_active_tran_server (page_to_tran_request::SEND_CATCHUP_COMPLETE, std::string());
```
joohok, line 1169: Even in a situation where an error occurs during the catch-up process and it seems like the catch-up will stop,
wouldn't it be better to inform ATS of both the completion of the catch-up and the situation where it stops ?

wrlawodms: Do you mean that where the catch-up fails for some reason? It will be addressed in a TODO issue when the way to re-request comes in. And about where to start the catch-up again, it doesn't need to be stored or notified to the ATS since the ATS just sends the latest catchup_lsa and the PS starts the catch_up from its local `append_lsa`, which is the point the previous catch-up stops.

src/server/active_tran_server.cpp
```diff
@@ -192,6 +195,13 @@ active_tran_server::connection_handler::receive_saved_lsa (page_server_conn_t::s
 			    get_channel_id ().c_str ());
 }
 
+void
+active_tran_server::connection_handler::receive_catchup_complete (page_server_conn_t::sequenced_payload &&a_sp)
+{
+  _er_log_debug (ARG_FILE_LINE, "[CATCH_UP] the catchup has been completed. channel id: %s\n",
```
cristiarg, line 201: The part
```
  {
    auto lockg_state = std::lock_guard<std::shared_mutex> { m_state_mtx };
    assert (m_state == state::CONNECTING);

    m_state = state::CONNECTED;
  }
```
from at the end of `tran_server::connection_handler::connect` should be moved here.

I saw that it is counter-argumented in http://jira.cubrid.org/browse/LETS-751?focusedCommentId=4764448&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-4764448 - but I see no problem to just move it here.

src/server/page_server.cpp
joohok: OK, I got it

joohok (approved):

src/server/active_tran_server.cpp
wrlawodms: This is being done in the following issue that I'm working on, http://jira.cubrid.org/browse/LETS-752. This state change bring up some issues to address.
1. Get it divergent in the case of PTS.
2. Change the meaning of the callback function,`on_connecting()`, which meant to be called before the transition to CONNECTED, but it triggers to change the state itself by sending the catchup-msg. I'm thinking of naming or semantic of this. `on_disconnecting()` also might be changed along for the naming or semantic consistency.

cristiarg (approved):

PR #4725, opened 2023-10-12
[CBRD-25032] Add a dump function and stat logging in debug mode

Rudeus:
http://jira.cubrid.org/browse/CBRD-25032

 - Implement dump functions of memory_monitor class on the server
 - Add debug log when add_stat() function adds stat.

### DIFF
--- a/src/base/memory_monitor_sr.cpp
+++ b/src/base/memory_monitor_sr.cpp
@@ -182,7 +182,9 @@ namespace cubperf
       void aggregate_module_info (int module_index, std::vector<MMON_MODULE_INFO> &info) const;
       void aggregate_module_info_summary (std::vector<MMON_MODULE_INFO> &info) const;
       void aggregate_tran_info (int tran_count, MMON_TRAN_INFO &info) const;
+#if !defined (NDEBUG)
       void stat_dump_debug () const;
+#endif
 
     private:
       inline int get_module_idx (MMON_STAT_ID stat_id) const;
@@ -645,6 +647,7 @@ namespace cubperf
     m_aggregater.get_transaction_info (tran_count, info);
   }
 
+#if !defined (NDEBUG)
   void memory_monitor::stat_dump_debug () const
   {
     MMON_SERVER_INFO server_info;
@@ -716,6 +719,7 @@ namespace cubperf
 
     er_log_debug (ARG_FILE_LINE, result.c_str());
   }
+#endif
 } // namespace cubperf
 
 /* APIs */
--- a/src/base/memory_monitor_sr.cpp
+++ b/src/base/memory_monitor_sr.cpp
@@ -30,6 +30,11 @@
 #include "memory_monitor_sr.hpp"
 #include "log_impl.h"
 
+#if !defined (NDEBUG)
+thread_local char mmon_err_msg[DB_MAX_IDENTIFIER_LENGTH];
+thread_local int mmon_err_msg_size = 0;
+#endif
+
 namespace cubperf
 {
   class memory_monitor;
@@ -177,6 +182,7 @@ namespace cubperf
       void aggregate_module_info (int module_index, std::vector<MMON_MODULE_INFO> &info) const;
       void aggregate_module_info_summary (std::vector<MMON_MODULE_INFO> &info) const;
       void aggregate_tran_info (int tran_count, MMON_TRAN_INFO &info) const;
+      void stat_dump_debug () const;
 
     private:
       inline int get_module_idx (MMON_STAT_ID stat_id) const;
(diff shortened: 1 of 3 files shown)


src/base/memory_monitor_sr.cpp
```diff
@@ -177,6 +182,7 @@ namespace cubperf
       void aggregate_module_info (int module_index, std::vector<MMON_MODULE_INFO> &info) const;
       void aggregate_module_info_summary (std::vector<MMON_MODULE_INFO> &info) const;
       void aggregate_tran_info (int tran_count, MMON_TRAN_INFO &info) const;
+      void stat_dump_debug () const;
```
hornetmj, line 186: wrap around to #if !defined (NDEBUG)

src/base/memory_monitor_sr.cpp
```diff
@@ -618,6 +644,78 @@ namespace cubperf
   {
     m_aggregater.get_transaction_info (tran_count, info);
   }
+
+  void memory_monitor::stat_dump_debug () const
```
hornetmj, line 663: wrap around to #if !defined (NDEBUG)

hornetmj (approved):

wrlawodms (commented):
How about adding a system parameter for this like `er_log_mmon_debug`, which enables to print. Without it, this is always printed. The debug msg here seems too long to be printed always.

wrlawodms:
> How about adding a system parameter for this like er_log_mmon_debug, which enables to print. Without it, this is always printed. The debug msg here seems too long to be printed always. 

Plus, in terms of purpose, this message is not for general inquiry, but for some specific test cases, isn't it?

src/base/memory_monitor_sr.cpp
Rudeus: Done.

src/base/memory_monitor_sr.cpp
Rudeus: Done.

Rudeus:
https://github.com/CUBRID/cubrid/pull/4725#pullrequestreview-1675528395

Rudeus:
https://github.com/CUBRID/cubrid/pull/4725#pullrequestreview-1675528395
@wrlawodms 
The dump log is generated if server satisfies all two conditions: "when Memory Monitoring option is enabled" and "when the application is built in debug mode". Furthermore, it's only printed once when the server is shut down. I believe that the situation where both of these conditions are met is unlikely to occur while the server is actively used by real users. As you mentioned, the dump function is designed to be removed after all modules are registered with the Memory Monitoring Manager. Therefore, I think it's best not to create additional system parameters.

wrlawodms:
> Furthermore, it's only printed once when the server is shut down. I believe that the situation where both of these conditions are met is unlikely to occur while the server is actively used by real users.

`[MMON-ADD]` msg is printed frequently. It would make it hard to investigate the log by developers. Do I get it wrong?

> As you mentioned, the dump function is designed to be removed after all modules are registered with the Memory Monitoring Manager

Will the functions themselves be removed? I thought that the code will be removed on #4678. If it will be removed, what's the purpose of this PR? No offense, I just don't undestarnd the purpose.

Rudeus:
> `[MMON-ADD]` msg is printed frequently. It would make it hard to investigate the log by developers. Do I get it wrong?

You're right. That description is about dump function. Nevertheless, running the server with the Memory Monitoring option enabled and built in debug mode is a situation that can be considered to require further analysis.

> Will the functions themselves be removed? I thought that the code will be removed on #4678. If it will be removed, what's the purpose of this PR? No offense, I just don't undestarnd the purpose.

The removal of the function will be carried out through a separate additional PR. The reason this function should be retained until then is that it's necessary to verify whether the registration process for any module with the Memory Monitoring Manager is successful. Even after registering the heap module in this project, there are still pending registrations for additional modules like the lock module and transaction manager module. Therefore, it's not appropriate to question "What's the purpose of this PR?" that it will be removed in the future.

wrlawodms:
> Even after registering the heap module in this project, there are still pending registrations for additional modules like the lock module and transaction manager module. Therefore, it's not appropriate to question "What's the purpose of this PR?" that it will be removed in the future.

OK, then it's not going to be removed in the current phase. I thought it's not worth the discussion if it will be removed soon.
Then, I am still insisting to add the system parameter not to make the debug logging cluttered. The dump at the end could be accepted if it's only printed when the mmon is enabled, but I think the frequent `[MMON-ADD]` is not acceptable.

joohok (approved):

wrlawodms (commented):
just to change the review status.

Rudeus:
Add system parameter and add condition in printing logs and dump

wrlawodms (approved):

PR #4727, opened 2023-10-16
[CBRD-25035] extend query cache to CTE

beyondykk9:
http://jira.cubrid.org/browse/CBRD-25035

Among the methods for expanding the result cache, we conducted a POC that applied the result cache to the CTE sub-query, and it was confirmed that performance was improved by referencing the result through the cached sub-query.

1) Use hints in CTE queries.
2) Applying the /*+ QUERY_CACHE */ hint to a CTE query changes from applying one result cache to one query to using the result cache for subqueries of the CTE as well.
3) When applying a CTE sub-query to the result-cache, the query is not auto committed internally even if the current setting is auto commit on because the main query including CTE sub-query should be executed as one transaction.
4) If an error occurs when executing a CTE sub-query from result-cache, an error is generated in the same way as before.


src/query/xasl_to_stream.c
```diff
@@ -5866,6 +5963,14 @@ xts_sizeof_xasl_node (const XASL_NODE * xasl)
       size += xts_sizeof_access_spec_type (access_spec);
     }
 
+  size += OR_INT_SIZE * 8;	/* cte_xasl_id = sha1(5) + flag(1) + time(2)... */
+  size += OR_INT_SIZE;		/* for CTE host variable count */
+
+  for (i = 0; i < xasl->cte_host_var_count; i++)
+    {
+      size += OR_INT_SIZE;
+    }
```
ctshim:
```suggestion
   size += (OR_INT_SIZE * xasl->cte_host_var_count);
```

src/query/xasl_to_stream.c
```diff
@@ -1454,6 +1454,81 @@ xts_save_val_list (const VAL_LIST * val_list)
...
+  else
+    {
+      buf_p = (char *) malloc (size);
+      if (buf_p == NULL)
+	{
+	  xts_Xasl_errcode = ER_OUT_OF_VIRTUAL_MEMORY;
+	  return ER_FAILED;
+	}
+
+      is_buf_alloced = true;
```
ctshim, line 1519: What is the purpose of the is_buf_alloced variable?

src/query/xasl_to_stream.c
```diff
@@ -1454,6 +1454,81 @@ xts_save_val_list (const VAL_LIST * val_list)
...
+    }
+
+  buf = xts_process_cte_xasl_id (buf_p, cte_xasl_id);
+  memcpy (&xts_Stream_buffer[offset], buf_p, size);
+
```
ctshim, line 1531: You need to free buf_p.
Also, Is there any guarantee that the space in xts_Stream_buffer[offset] is large enough?

src/query/xasl_to_stream.c
beyondykk9: I fixed to free the memory when the is_buf_alloced is true.

src/query/xasl_to_stream.c
beyondykk9: The code is resemble with xts_process_cte_list_id.

src/query/xasl_to_stream.c
beyondykk9: I fixed as your comment.

src/query/xasl_to_stream.c
```diff
@@ -5866,6 +5975,11 @@ xts_sizeof_xasl_node (const XASL_NODE * xasl)
       size += xts_sizeof_access_spec_type (access_spec);
     }
 
+  size += OR_INT_SIZE * 8;	/* cte_xasl_id = sha1(5) + flag(1) + time(2)... */
```
hgryoo: In case of cte_xasl_id is NULL?
```suggestion
if (xasl->cte_xasl_id == NULL)
    {
  size += OR_INT_SIZE;
    }
else
    {
  size += OR_INT_SIZE * 8;	/* cte_xasl_id = sha1(5) + flag(1) + time(2)... */
    }
```

beyondykk9: I fixed the size for "cte_xasl_id" as like handling of list_id.

src/query/stream_to_xasl.c
```diff
@@ -2336,6 +2395,29 @@ stx_build_cache_attrinfo (char *ptr)
...
+static char *
+stx_build_cte_xasl_id (THREAD_ENTRY * thread_p, char *ptr, XASL_ID * cte_xasl_id)
+{
+  int i;
+
+  for (i = 0; i < 5; i++)
+    {
+      cte_xasl_id->sha1.h[i] = OR_GET_INT (ptr);
+      ptr += OR_INT_SIZE;
+    }
```
ctshim: Please use the prepared macro

 OR_GET_SHA1 (ptr, &(cte_xasl_id->sha1));
 ptr += OR_SHA1_SIZE;
 
There are also functions that can be used instead of macros, as shown below.
or_pack_sha1(), or_unpack_sha1()

src/query/xasl_to_stream.c
```diff
@@ -1454,6 +1454,94 @@ xts_save_val_list (const VAL_LIST * val_list)
...
+static char *
+xts_process_cte_xasl_id (char *ptr, const XASL_ID * cte_xasl_id)
+{
+  int i;
+
+  for (i = 0; i < 5; i++)
+    {
+      OR_PUT_INT (ptr, cte_xasl_id->sha1.h[i]);
+      ptr += OR_INT_SIZE;
+    }
```
ctshim: Please use the prepared macro

  OR_PUT_SHA1 (ptr, &(cte_xasl_id->sha1));
  ptr += OR_SHA1_SIZE;
  
There are also functions that can be used instead of macros, as shown below.
or_pack_sha1(), or_unpack_sha1()

beyondykk9: OK, good point.
I fixed as your comment.

src/query/stream_to_xasl.c
beyondykk9: OK, good point.
I fixed as your comment.

hyunikn:
Is the following result (got from commit 1b15641) correct? It looks different from the case in the issue CBRD-25035.
-=-=-=
csql> with cte0 as (select /*+ query_cache */ count(*) from public.game where host_year > 2000),
csql> cte1 as (select /*+ query_cache */ count(*) from public.game where nation_code = 'KOR')
csql> select /*+ recompile */ * from cte0 union select * from cte1;

=== <Result of SELECT Command in Line 3> ===

              count(*)
======================
                   316
                  1948

2 rows selected. (0.052305 sec) Committed. (0.000016 sec) 

csql> ;info qcache
LIST_CACHE {
  n_hts 0
  n_entries 0  n_pages 0
  lookup_counter 0
  hit_counter 0
  miss_counter 0
  full_counter 0
}

Committed.
csql>

src/query/stream_to_xasl.c
```diff
@@ -1385,6 +1386,40 @@ stx_restore_OID_array (THREAD_ENTRY * thread_p, char *ptr, int nelements)
   return oid_array;
 }
 
+static XASL_ID *
+stx_restore_cte_xasl_id (THREAD_ENTRY * thread_p, char *ptr)
+{
+  int i, offset;
+  XASL_ID *cte_xasl_id;
+  XASL_UNPACK_INFO *xasl_unpack_info = get_xasl_unpack_info_ptr (thread_p);
```
hyunikn: This local variable is not used in this function.

hyunikn (approved):

ctshim (approved):

src/query/stream_to_xasl.c
beyondykk9: Good point. Thanks. I fixed it.

src/query/execute_statement.c
```diff
@@ -14504,6 +14560,122 @@ do_prepare_session_statement (PARSER_CONTEXT * parser, PT_NODE * statement)
...
+  cte_list = statement->info.query.with->info.with_clause.cte_definition_list;
+
+  CACHE_TIME_RESET (&clt_cache_time);
+
+  while (cte_list)
+    {
+      stmt = cte_list->info.cte.non_recursive_part;
+      if (stmt && (stmt->info.query.hint & PT_HINT_QUERY_CACHE))
+	{
+	  host_variables = (DB_VALUE *) parser_alloc (parser, sizeof (DB_VALUE) * stmt->cte_host_var_count);
```
hgryoo: In parser_support.c, host_variable array is allocated by malloc/realloc.

hgryoo (approved):

src/query/execute_statement.c
beyondykk9: This is related to query parsing, so I keep the logic using parser_alloc.

hgryoo: It is about lifecycle of an allocated memory by parser_alloc (). When and where the `cte_context.host_variables` is freed?

hgryoo: I couldn't find a free function corresponding to parser_alloc(). Maybe the allocated memory will be freed depends on other object's lifecycle.

beyondykk9: The parser alloced memory is free when the parser end by parser_free_parser function.

PR #4729, opened 2023-10-17
[CBRD-25005] Semantic API returns error when a field lookup expression is used in Static SQL

hgryoo:
http://jira.cubrid.org/browse/CBRD-25005

### DIFF
--- a/pl_engine/pl_server/src/main/java/com/cubrid/plcsql/compiler/ParseTreeConverter.java
+++ b/pl_engine/pl_server/src/main/java/com/cubrid/plcsql/compiler/ParseTreeConverter.java
@@ -2370,15 +2370,18 @@ public class ParseTreeConverter extends PcsParserBaseVisitor<AstNode> {
                     String[] split = hostExpr.split("\\.");
                     if (split.length == 1) {
 
-                        ExprId id = visitNonFuncIdentifier(hostExpr, ctx); // s408: undeclared id ...
+                        ExprId id =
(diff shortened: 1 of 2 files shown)


hgryoo:
@hyunikn 
By 28d9279, the error is changed as following.
```
create or replace procedure poo as
    s varchar;
begin
    for r in (select coll_id id from db_collation) loop
       select coll_name into s from db_collation where coll_id = r.id;
       dbms_output.put_line('s=' || s);
    end loop;
end;
```
```
Stored procedure compile error: undeclared id R.ID
```

It seems that the label of the host variable is now properly sent to the PL/CSQL compiler. The rest can be handled by the compiler.

hyunikn:
> It seems that the label of the host variable is now properly sent to the PL/CSQL compiler. The rest can be handled by the compiler.

Yes. Let me add some updates.

hyunikn (approved):

hyunikn:
@hgryoo I pushed an update. Please take a look.

hgryoo:
```
;server-output on

create or replace procedure poo as
    s varchar;
begin
    for r in (select coll_id id from db_collation) loop
       select coll_name into s from db_collation where coll_id = r.id;
       dbms_output.put_line('s=' || s);
    end loop;
end;

call poo();
```

Result
```
s=iso88591_bin
s=utf8_bin
s=iso88591_en_cs
s=iso88591_en_ci
s=utf8_en_cs
s=utf8_en_ci
s=utf8_tr_cs
s=utf8_ko_cs
s=euckr_bin
s=binary
```

LGTM.

PR #4731, opened 2023-10-17
[LETS-752]  Set the state of connection to CONNECTED after its catch-up

wrlawodms:
http://jira.cubrid.org/browse/LETS-752

- Change the state of the connection_handler of a PS from CONNECTING to CONNECTED when its catchup is done.
- on_connecting() -> transition_to_connected()
  - It was intended as a callback which is called between the transition from CONNECTING to CONNECTED. 
  - However, now it's responsible for the state transition as well.
  - I wanted to change the on_disconnecting() as well for consistency, but I failed, so I just leave it as it is. So, now they are not the same kind of functions.
- In `tran_server::boot()`, having established all connections with PSs, it may not set the main connection since no connection has not yet CONNECTED, but CONNECTING. So, it waits for now. It will be removed in the future. It's described in the comment.
- I've tested this with manual `sleep_for()` while processing the catch-up msg on PS.


src/server/tran_server.hpp
```diff
@@ -175,6 +153,28 @@ class tran_server
...
+	  , m_node { std::move (node) }
+	{ }
+
+	virtual request_handlers_map_t get_request_handlers ();
+	void push_request_regardless_of_state (tran_to_page_request reqid, std::string &&payload);
+
+	// Do the server-type-specific jobs and transition state from CONNECTING to CONNECTED.
+	// The m_state MUST be changed to CONNECTED either synchronously or asynchronously.
+	virtual void transition_to_connected () = 0;
+	// Do the rver-type-specific jobs before disconnected. DISCONNECTING -> (*) -> IDLE
```
hornetmj:
```suggestion
	// Do the server-type-specific jobs before disconnected. DISCONNECTING -> (*) -> IDLE
```

hornetmj: Has 'server-type-specific' the special meaning? Or is it just general meaning?

src/server/tran_server.cpp
```diff
@@ -125,17 +125,44 @@ tran_server::boot (const char *db_name)
...
   if (m_page_server_conn_vec.empty () == false)
     {
-      /*
-       * At least one PS is given by the configuration.
-       * Even if uses_remote_storage () == false, the remote storage can exist.
-       */
-      error_code = reset_main_connection ();
-      if (error_code != NO_ERROR)
+      const auto start_time = std::chrono::steady_clock::now ();
+      while (true)
```
hornetmj, line 136:
```suggestion

      while (true)
```

src/server/tran_server.cpp
```diff
@@ -125,17 +125,44 @@ tran_server::boot (const char *db_name)
...
+	      * TODO: Remove this and just make sure reset_main_connection doesn't fail
+	      *       when the ATS recovery with handshakes with multiple page servers comes in.
+	      *
+	      * For now, the main connection may not be able to be set for a while. It can be after one it set to CONNECTED state.
+	      * When the handshakes comes in, it's guaranteed that at least one connection is completely CONNECTED here
+	      * and it will be the main connection. Until then, we just waits here until a conenction is ready.
+	      */
+	      const auto current_time = std::chrono::steady_clock::now ();
+	      const auto duration = std::chrono::duration_cast<std::chrono::seconds> (current_time - start_time).count();
+	      if (duration > 30) // timeout: 30 seconds
```
hornetmj, line 156: Does it make sense to wait at this point? Since m_ps_connector.start(); is called outside the while loop, waiting doesn't seem to work that you exepected.

wrlawodms: It doesn't wait for the `m_ps_connector` to connect. All connections have already been established in `init_page_server_hosts` above. However, the state of them could be "CONNECTING", which can't be selected as the main connection. It will be set to CONNECTED when a PS reply that its catchup is done.

src/server/tran_server.hpp
wrlawodms: The`server-type-specific' means that jobs depend on the type, pts or ats.

src/server/tran_server.cpp
hornetmj: I thought a case that all connect() call returns the ER_NET_PAGESERVER_CONNECTION error.

src/server/tran_server.hpp
hornetmj: OK. Fix the typo. Or let me know the meaning of 'rver-type-specific' too.

hornetmj:
@wrlawodms You request review again to me while our discussion is in progress. I think we need to complete our previous discussion first.

src/server/tran_server.hpp
wrlawodms: lol. It's just a typo. I'll fix it.

src/server/tran_server.cpp
wrlawodms: You mean all `connect()` in `init_page_server_hosts()` above, right? In this case, the error is returned below the call of the function. This loop is not reached.

hornetmj: Oh. I got it.

hornetmj (approved):

src/server/tran_server.cpp
```diff
@@ -125,17 +125,44 @@ tran_server::boot (const char *db_name)
...
+	  else
+	    {
+	      /* TEMPORARY waiting and timeout
+	      *
+	      * TODO: Remove this and just make sure reset_main_connection doesn't fail
+	      *       when the ATS recovery with handshakes with multiple page servers comes in.
+	      *
+	      * For now, the main connection may not be able to be set for a while. It can be after one it set to CONNECTED state.
+	      * When the handshakes comes in, it's guaranteed that at least one connection is completely CONNECTED here
+	      * and it will be the main connection. Until then, we just waits here until a conenction is ready.
```
joohok:
```suggestion
	      * and it will be the main connection. Until then, we just wait here until a conenction is ready.
```

joohok (approved):

cristiarg (approved):

PR #4734, opened 2023-10-18
[LETS-817] Make it clear and robust the DISCONNECTING state of PS.

wrlawodms:
http://jira.cubrid.org/browse/LETS-817

- The main modification is to reset m_conn and to change the state at once with two locks held.
- Revise an assertion `request_sync_client_server::push`. It doesn't need to check the liveness of the receive thread, but it should check the sender thread.
- Correct some typos.

### DIFF
--- a/src/server/tran_server.cpp
+++ b/src/server/tran_server.cpp
@@ -555,13 +555,13 @@ tran_server::connection_handler::disconnect_async (bool with_disc_msg)
 
     /*
      * Stop incoming communication and wake up threads waiting for a response and tell them it won't be served.
-     * This has to be out of the lock of m_conn_mtx because a waitor may hold the lock.
+     * This has to be out of the lock of m_conn_mtx because waitors are holding the lock. (send_receive())
      * m_conn is not nullptr here since it's only set to nullptr here below.
      */
     m_conn->stop_incoming_communication_thread ();
 
-    auto lockg_state = std::lock_guard<std::shared_mutex> { m_state_mtx };
-    auto ulock_conn = std::unique_lock<std::shared_mutex> { m_conn_mtx };
+    auto lockg_state = std::lock_guard { m_state_mtx };
+    auto lockg_conn = std::lock_guard { m_conn_mtx };
     const std::string channel_id = get_channel_id ();
 
     assert (m_state == state::DISCONNECTING);
--- a/src/server/tran_server.cpp
+++ b/src/server/tran_server.cpp
@@ -554,23 +554,15 @@ tran_server::connection_handler::disconnect_async (bool with_disc_msg)
     on_disconnecting (); // server-type specific jobs before disconnecting.
 
     /*
-     * Wake up threads waiting for a response and tell them it won't be served.
+     * Stop incoming communication and wake up threads waiting for a response and tell them it won't be served.
      * This has to be out of the lock of m_conn_mtx because a waitor may hold the lock.
      * m_conn is not nullptr here since it's only set to nullptr here below.
      */
-    m_conn->stop_response_broker ();
-
-    auto ulock_conn = std::unique_lock<std::shared_mutex> { m_conn_mtx };
-    const std::string channel_id = get_channel_id ();
-
-    // This may call the function (disconnect_async) while digesting all requests since there might be a disconnection msg.
-    // In that, the lock of m_state_mtx will be acquired again, leading to a deadlock. We have to clean incoming requests first.
     m_conn->stop_incoming_communication_thread ();
 
-    ulock_conn.unlock (); // To keep the order of locks: m_state_mtx->m_conn_mtx
-
     auto lockg_state = std::lock_guard<std::shared_mutex> { m_state_mtx };
-    ulock_conn.lock ();
+    auto ulock_conn = std::unique_lock<std::shared_mutex> { m_conn_mtx };
+    const std::string channel_id = get_channel_id ();
 
     assert (m_state == state::DISCONNECTING);
 
--- a/src/server/tran_server.cpp
+++ b/src/server/tran_server.cpp
@@ -551,34 +551,35 @@ tran_server::connection_handler::disconnect_async (bool with_disc_msg)
 
   m_disconn_future = std::async (std::launch::async, [this, with_disc_msg]
   {
-    on_disconnecting (); // server-type specific jobs before disconnecting.
-    // m_conn is not nullptr since it's only set to nullptr here below once
-
     m_conn->stop_response_broker (); // wake up threads waiting for a response and tell them it won't be served.
     auto ulock_conn = std::unique_lock<std::shared_mutex> { m_conn_mtx };
     const std::string channel_id = get_channel_id ();
+
+    // This may call this function (disconnect_async) while digesting all requests.
+    // In that, the lock of m_state_mtx will be acquired again, leading to a deadlock. We have to clean incoming requests first.
+    m_conn->stop_incoming_communication_thread ();
+
+    ulock_conn.unlock ();
+
+    on_disconnecting (); // server-type specific jobs before disconnecting.
+
+    auto lockg_state = std::lock_guard<std::shared_mutex> { m_state_mtx };
+    ulock_conn.lock ();
+
+    assert (m_state == state::DISCONNECTING);
+
     if (with_disc_msg)
       {
 	const int payload = static_cast<int> (m_ts.m_conn_type);
 	std::string msg (reinterpret_cast<const char *> (&payload), sizeof (payload));
 	m_conn->push (tran_to_page_request::SEND_DISCONNECT_MSG, std::move (msg));
-	// After sending SEND_DISCONNECT_MSG, the page server may release all resources releated to this connection.
+	// After sending SEND_DISCONNECT_MSG, the page server may release all resources related to this connection.
+	// So, it has to be the last msg.
       }
 
-    // stop_incoming_communication_thread() has to be done explicitly before m_conn.reset () to avoid a request handler or an error handler accesses nullptr of m_conn.
-    m_conn->stop_incoming_communication_thread ();
-    m_conn->stop_outgoing_communication_thread ();
     m_conn.reset (nullptr);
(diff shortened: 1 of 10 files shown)


src/server/tran_server.cpp
```diff
@@ -551,34 +551,35 @@ tran_server::connection_handler::disconnect_async (bool with_disc_msg)
 
   m_disconn_future = std::async (std::launch::async, [this, with_disc_msg]
   {
-    on_disconnecting (); // server-type specific jobs before disconnecting.
-    // m_conn is not nullptr since it's only set to nullptr here below once
-
     m_conn->stop_response_broker (); // wake up threads waiting for a response and tell them it won't be served.
```
joohok: I think it can be removed too, because this function is also called inside the `m_conn->stop_incoming_communication_thread ()`.

src/server/tran_server.cpp
```diff
@@ -551,34 +551,35 @@ tran_server::connection_handler::disconnect_async (bool with_disc_msg)
...
     auto ulock_conn = std::unique_lock<std::shared_mutex> { m_conn_mtx };
     const std::string channel_id = get_channel_id ();
+
+    // This may call this function (disconnect_async) while digesting all requests.
+    // In that, the lock of m_state_mtx will be acquired again, leading to a deadlock. We have to clean incoming requests first.
+    m_conn->stop_incoming_communication_thread ();
+
+    ulock_conn.unlock ();
+
+    on_disconnecting (); // server-type specific jobs before disconnecting.
```
joohok: Could you tell me why this function was moved here? If it wasn't moved, it seems like there would be no need to unlock the `ulock_conn`

src/server/tran_server.cpp
wrlawodms: You're right. It can be much simpler. Thank you.

src/server/tran_server.cpp
wrlawodms: I don't see the necessity. It can be much simpler with the other change. Thanks.

joohok (approved):

src/server/tran_server.cpp
```diff
@@ -552,33 +552,32 @@ tran_server::connection_handler::disconnect_async (bool with_disc_msg)
...
-    m_conn->stop_response_broker (); // wake up threads waiting for a response and tell them it won't be served.
+    /*
+     * Stop incoming communication and wake up threads waiting for a response and tell them it won't be served.
+     * This has to be out of the lock of m_conn_mtx because a waitor may hold the lock.
+     * m_conn is not nullptr here since it's only set to nullptr here below.
+     */
+    m_conn->stop_incoming_communication_thread ();
+
+    auto lockg_state = std::lock_guard<std::shared_mutex> { m_state_mtx };
     auto ulock_conn = std::unique_lock<std::shared_mutex> { m_conn_mtx };
```
hornetmj: ulock_conn.unlock () is removed. So, how about use std::lock_guard(m_conn_mtx) or std::scoped_lock(m_state_mtx, m_conn_mtx).

src/server/tran_server.cpp
```diff
@@ -552,33 +552,32 @@ tran_server::connection_handler::disconnect_async (bool with_disc_msg)
   m_disconn_future = std::async (std::launch::async, [this, with_disc_msg]
   {
     on_disconnecting (); // server-type specific jobs before disconnecting.
-    // m_conn is not nullptr since it's only set to nullptr here below once
 
-    m_conn->stop_response_broker (); // wake up threads waiting for a response and tell them it won't be served.
+    /*
+     * Stop incoming communication and wake up threads waiting for a response and tell them it won't be served.
+     * This has to be out of the lock of m_conn_mtx because a waitor may hold the lock.
```
hornetmj: Where is the case?

wrlawodms: in `tran_server::connection_handler::send_receive()`. I'll add it in comment.

src/server/tran_server.cpp
wrlawodms: Good. Done.

hornetmj (approved):
LGTM

src/server/tran_server.cpp
```diff
@@ -552,33 +552,32 @@ tran_server::connection_handler::disconnect_async (bool with_disc_msg)
   m_disconn_future = std::async (std::launch::async, [this, with_disc_msg]
   {
     on_disconnecting (); // server-type specific jobs before disconnecting.
-    // m_conn is not nullptr since it's only set to nullptr here below once
 
-    m_conn->stop_response_broker (); // wake up threads waiting for a response and tell them it won't be served.
-    auto ulock_conn = std::unique_lock<std::shared_mutex> { m_conn_mtx };
+    /*
+     * Stop incoming communication and wake up threads waiting for a response and tell them it won't be served.
+     * This has to be out of the lock of m_conn_mtx because waitors are holding the lock. (send_receive())
```
```suggestion
     * This has to be out of the lock of m_conn_mtx because waiters are holding the lock. (send_receive())
```

src/server/tran_server.cpp
```diff
@@ -552,33 +552,32 @@ tran_server::connection_handler::disconnect_async (bool with_disc_msg)
   m_disconn_future = std::async (std::launch::async, [this, with_disc_msg]
   {
     on_disconnecting (); // server-type specific jobs before disconnecting.
-    // m_conn is not nullptr since it's only set to nullptr here below once
 
-    m_conn->stop_response_broker (); // wake up threads waiting for a response and tell them it won't be served.
-    auto ulock_conn = std::unique_lock<std::shared_mutex> { m_conn_mtx };
+    /*
```
cristiarg, line 556: I as understand, previously, `stop_response_broker` was being called twice
- once from here - deleted
- once from `stop_incoming_communication_thread` - remaining

Thus, `response_broker<T_PAYLOAD, T_ERROR>::bucket::notify_terminate_and_wait` has a section
```
      if (m_terminate)
	{
	  return; // it has been already terminated.
	}
```
Is this still needed? Or can be an assert `!m_terminate` ?

src/server/tran_server.cpp
```diff
@@ -552,33 +552,32 @@ tran_server::connection_handler::disconnect_async (bool with_disc_msg)
...
+     * Stop incoming communication and wake up threads waiting for a response and tell them it won't be served.
+     * This has to be out of the lock of m_conn_mtx because waitors are holding the lock. (send_receive())
+     * m_conn is not nullptr here since it's only set to nullptr here below.
```
cristiarg: Nit
Continuation for the explanation:
```
... and if function `tran_server::connection_handler::disconnect_async` happens to be called more than once for a connection handler, it will be guarded by the state checks at the beginning of the function (see above).
```
Correct?

src/server/tran_server.cpp
```diff
@@ -552,33 +552,32 @@ tran_server::connection_handler::disconnect_async (bool with_disc_msg)
...
 	std::string msg (reinterpret_cast<const char *> (&payload), sizeof (payload));
 	m_conn->push (tran_to_page_request::SEND_DISCONNECT_MSG, std::move (msg));
-	// After sending SEND_DISCONNECT_MSG, the page server may release all resources releated to this connection.
+	// After sending SEND_DISCONNECT_MSG, the page server may release all resources related to this connection.
+	// So, it has to be the last msg.
       }
 
-    // stop_incoming_communication_thread() has to be done explicitly before m_conn.reset () to avoid a request handler or an error handler accesses nullptr of m_conn.
-    m_conn->stop_incoming_communication_thread ();
-    m_conn->stop_outgoing_communication_thread ();
```
cristiarg, line 570: So, for ATS and PTS, the outgoing communication thread will be implicitly stopped in the `request_queue_autosend<ReqQueue>::~request_queue_autosend`
For PS, it is still specifically called at the end of the connection handler dtor.

All in all, I think `request_queue_autosend<ReqQueue>::~request_queue_autosend` could be removed altogether from everywhere. When it was added, I think it was added only as a matter of precaution to protect against race conditions.

cristiarg (approved):

src/server/tran_server.cpp
wrlawodms: I think we can make it `stop_response_broker` private and only called in `stop_incoming_communication_thread`. It's unreasonable to keep incoming thread when the response broker doesn't work. Who delivers the response? I'll create a refactoring issue for this.

src/server/tran_server.cpp
wrlawodms: > request_queue_autosend<ReqQueue>::~request_queue_autosend could be removed altogether from everywhere.

Do you happen to mean `stop_outgoing_communication_thread`? Or do you mean we should call `stop_outgoing_communication_thread` explicitly everywhere?

src/server/tran_server.cpp
wrlawodms: Yes. And revised with this.

src/server/tran_server.cpp
wrlawodms: On second thought, it does have a meaning in that a connection_handler can receive requests from outside, but can't do `send_recv`. What do you think?

src/server/tran_server.cpp
cristiarg: Sorry, my previous message is confusing even for me now. :(

I meant `request_sync_client_server::stop_outgoing_communication_thread` could be removed altogether from everywhere.
And then the queue autosend will be stopped via: `request_sync_client_server` dtor -> `request_queue_autosend` dtor -> `request_queue_autosend::stop_thread`.

The goal of removing it is to keep the same approach on the xTS and PS side.

wrlawodms: No problem. I think we can remove it and make it consistent, but I think it'd be better to postpone the refactoring later. It's the fair of the `stop_incoming_communication_thread`, and allow us a kind of "read-only" connection_handler. I don't know it will be needed. But, I think we can refactor it later when the connection_handler's are more mature and stabilized. Or, do you think it is better to refactor as many as possible we can and it's kind of "good way"? No sarcasm at all, I just want to listen to your insight.

PR #4735, opened 2023-10-18
[CBRD-25037] Simplify MVCC check in heap scan

xmilex-git:
http://jira.cubrid.org/browse/CBRD-25037

- Heap scan performance in CUBRID 11.4 has declined by 56.8% when compared to CUBRID 9.3 due to operations performed to check MVCC id. Two improvements were suggested:

1. Eliminating unnecessary calls to pgbuf_replace_watcher() function in heap_next_internal() by using old_page_watcher element to cache old page watcher.

2. Optimizing the heap_next_internal() function to reduce the number of accesses to fetch the address of a single record from three times to once under specific conditions, enhancing efficiency.


hornetmj:
@xmilex-git Three improvements were made at the same time. Can I know what the contribution of each improvement item is?

hornetmj:
@xmilex-git I add @wrlawodms as a reviewer becuase he has a deep understanding about page_buffer and heap_file.

src/base/object_representation_sr.c
```diff
@@ -4283,6 +4305,20 @@ or_mvcc_get_header (RECDES * record, MVCC_REC_HEADER * mvcc_header)
...
+  mvcc_header->chn = NULL_CHN;
+  mvcc_header->mvcc_ins_id = MVCCID_NULL;
+  mvcc_header->mvcc_del_id = MVCCID_NULL;
+
+  if (rc != NO_ERROR)
+    {
+      goto exit_on_error;
+    }
+
+  return NO_ERROR;
```
wrlawodms: The `rc != NO_ERROR)` seems to be always true because or_underflow() returns ER_TF_BUFFER_UNDERFLOW. Then, we can just move this `underflow_detected` section above exit_on_error without this.
```suggestion 
 exit_on_error:
   return (rc == NO_ERROR && (rc = er_errid ()) == NO_ERROR) ? ER_FAILED : rc;
```

src/base/object_representation_sr.c
```diff
@@ -4242,38 +4242,60 @@ or_mvcc_get_header (RECDES * record, MVCC_REC_HEADER * mvcc_header)
...
+  int flag;
 
   assert (record != NULL && record->data != NULL && record->length >= OR_MVCC_REP_SIZE && mvcc_header != NULL);
 
   or_init (&buf, record->data, record->length);
 
-  repid_and_flag_bits = or_mvcc_get_repid_and_flags (&buf, &rc);
-  if (rc != NO_ERROR)
+  header_size = OR_HEADER_SIZE (buf.buffer);
+  if ((buf.ptr + header_size) > buf.endptr)
```
wrlawodms: I can't say that I like the approach of the modification to `or_mvcc_get_header`. Yes, it reduces the number of the condition checks from at most 5 to 1, but it harms readability. In most cases, the check would be 3 or fewer. If it has a big impact (or is reasonable at least) on performance, of course, we should optimize it, but I doubt it does. Do you think it's worth it at the cost of readability?

Regarding reducing function calls, it probably does not reduce it because the functions are small and inline.

Please share your opinion.

src/storage/heap_file.c
```diff
@@ -7177,6 +7184,10 @@ heap_scancache_quick_end (THREAD_ENTRY * thread_p, HEAP_SCANCACHE * scan_cache)
 	    {
 	      pgbuf_ordered_unfix (thread_p, &scan_cache->page_watcher);
 	    }
+	  if (scan_cache->old_page_watcher.pgptr != NULL)
+	    {
+	      pgbuf_ordered_unfix (thread_p, &scan_cache->old_page_watcher);
+	    }
```
wrlawodms: it won't happen. It's always unfixed on heap_next_internal().
```suggestion
	  assert (scan_cache->old_page_watcher.pgptr == NULL);
```

src/storage/heap_file.c
```diff
@@ -7858,9 +7867,6 @@ heap_next_internal (THREAD_ENTRY * thread_p, const HFID * hfid, OID * class_oid,
       class_oid = &scan_cache->node.class_oid;
     }
 
-  PGBUF_INIT_WATCHER (&curr_page_watcher, PGBUF_ORDERED_HEAP_NORMAL, hfid);
-  PGBUF_INIT_WATCHER (&old_page_watcher, PGBUF_ORDERED_HEAP_NORMAL, hfid);
```
wrlawodms: I think it's good idea to make use of `scan_cache->page_watcher`instead of `curr_page_watcher`. It removes the redundant copies to and from the local variable. I don't see any side effect from this. :)

However, I can't see why `old_page_watcher` also needs to be in the scan_cache. It seems used locally in this function and it does not reduce the call of `pgbuf_replace_watcher`. Could you tell me the reason?

src/base/object_representation_sr.c
wrlawodms: Forget this, this was on the revert code.

src/base/object_representation_sr.c
wrlawodms: Forget this, this was on the revert code.

xmilex-git: I also think that this modification would slow down the performance, so I reverted the change and committed it.

src/storage/heap_file.c
xmilex-git: Including old_page_watcher in scan_cache has the advantage of slightly improving the overhead caused by PGBUF_INIT_WATCHER. This was the reason for including it.

wrlawodms: I don't think it can be the reason unless the initialization includes heavy loads. All local variables here can be located in the scan_cache to reduce initialization cost, but we don't. Because it's used only locally. The `scan_cache` is for caching data beyond the call of this function. Let's keep local things local. Or, does it cost that much?

xmilex-git: Based on the performance testing, it has been confirmed that there is about a 1~2% overhead. Since it's not a heavy load, it seems more appropriate to manage it with local variables. I will change it to local variables and commit.

src/storage/heap_file.c
```diff
@@ -8110,15 +8101,11 @@ heap_next_internal (THREAD_ENTRY * thread_p, const HFID * hfid, OID * class_oid,
       pgbuf_ordered_unfix (thread_p, &old_page_watcher);
     }
 
-  if (curr_page_watcher.pgptr != NULL)
+  if (scan_cache->page_watcher.pgptr != NULL)
     {
       if (!scan_cache->cache_last_fix_page)
```
```suggestion
  if (scan_cache->page_watcher.pgptr != NULL && scan_cache->cache_last_fix_page == false)
    {
```

src/storage/heap_file.c
```diff
@@ -24863,11 +24851,48 @@ heap_get_visible_version (THREAD_ENTRY * thread_p, const OID * oid, OID * class_
...
+    {
+      MVCC_REC_HEADER mvcc_header = MVCC_REC_HEADER_INITIALIZER;
+      if (or_mvcc_get_header (forward_recdes, &mvcc_header) != NO_ERROR)
+	{
+	  /* Unexpected. */
+	  assert (false);
+	}
+      if (scan_cache != NULL && scan_cache->mvcc_snapshot != NULL
+	  && scan_cache->mvcc_snapshot->snapshot_fnc != NULL && !mvcc_is_mvcc_disabled_class (class_oid)
+	  && scan_cache->mvcc_snapshot->snapshot_fnc (thread_p, &mvcc_header,
```
hornetmj: I think that if you use 'scan_cache->mvcc_snapshot) == SNAPSHOT_SATISFIED' condition, then you can remove the redundant code.

src/storage/heap_file.c
```diff
@@ -24863,11 +24851,48 @@ heap_get_visible_version (THREAD_ENTRY * thread_p, const OID * oid, OID * class_
...
 SCAN_CODE
 heap_scan_get_visible_version (THREAD_ENTRY * thread_p, const OID * oid, OID * class_oid, RECDES * recdes,
-			       HEAP_SCANCACHE * scan_cache, int ispeeking, int old_chn)
+			       RECDES * forward_recdes, HEAP_SCANCACHE * scan_cache, int ispeeking, int old_chn)
 {
   SCAN_CODE scan = S_SUCCESS;
   HEAP_GET_CONTEXT context;
 
+  /*
+   * Before fetching a record, check forward_recdes to see if the record type is REC_HOME,
```
hornetmj: This should be done in heap_get_visible_version_internal(), but this part was redundantly written to improve performance. Therefore, I think the need for this code block will be well shared if you clearly state that this is a part of the work you did to improve performance.

src/storage/heap_file.c
xmilex-git: I think it is good idea. Using this, I simplified this if statement and committed it.

src/storage/heap_file.c
xmilex-git: I agree with your review. So I have added comments regarding the fact that this process should be done in heap_get_visible_version_internal(), and that it's additional code for performance improvement.

src/storage/heap_file.c
hornetmj: Nit:

1. assert (scan_cache != NULL); iso checking scan_cache != NULL; See the heap_next_internal().
2. assert (recdes != NULL); or checking it at the first if stmt.

hornetmj (approved):

src/storage/heap_file.c
```diff
@@ -24863,11 +24850,45 @@ heap_get_visible_version (THREAD_ENTRY * thread_p, const OID * oid, OID * class_
...
+      if (scan_cache != NULL && scan_cache->mvcc_snapshot != NULL
+	  && scan_cache->mvcc_snapshot->snapshot_fnc != NULL && !mvcc_is_mvcc_disabled_class (class_oid)
+	  && scan_cache->mvcc_snapshot->snapshot_fnc (thread_p, &mvcc_header,
+						      scan_cache->mvcc_snapshot) == SNAPSHOT_SATISFIED
+	  && recdes != NULL)
+	{
+	  *recdes = *forward_recdes;
+	  return scan;
+	}
+      /*fall through to get visible version... */
```
```suggestion
      /* fall through to get visible version... */
```

src/storage/heap_file.c
wrlawodms: This change seems to change the logic. In the previous version, if `scan_cache == NULL`, `scan_cache->mvcc_snapshot == NULL`, `scan_cache->mvcc_snapshot->snapshot_fnc`, or `mvcc_is_mvcc_disabled_class`, it skips the visibility check. I thought that it was what you intended.  Now, it does the optimization only when the snapshot is satisfied. For example, this optimization worked for db_serial by `mvcc_is_mvcc_disabled_class()`, but it's not with the current version.

To me, it should be like
```
// as it was before and as in heap_get_visible_version_internal.
const need_check_visibility = scan_cache != NULL && scan_cache->mvcc_snapshot != NULL
 	  && scan_cache->mvcc_snapshot->snapshot_fnc != NULL && !mvcc_is_mvcc_disabled_class (class_oid)
 	  && scan_cache->mvcc_snapshot->snapshot_fnc (thread_p, &mvcc_header, scan_cache->mvcc_snapshot) != SNAPSHOT_SATISFIED; 

if (!need_check_visibility)
{
  *recdes = *forward_recdes;
  return scan;
}
/* fall through.. *//
} 

```

src/storage/heap_file.c
```diff
@@ -24863,11 +24850,45 @@ heap_get_visible_version (THREAD_ENTRY * thread_p, const OID * oid, OID * class_
...
+      if (or_mvcc_get_header (forward_recdes, &mvcc_header) != NO_ERROR)
+	{
+	  /* Unexpected. */
+	  assert (false);
+	}
+      if (scan_cache != NULL && scan_cache->mvcc_snapshot != NULL
+	  && scan_cache->mvcc_snapshot->snapshot_fnc != NULL && !mvcc_is_mvcc_disabled_class (class_oid)
+	  && scan_cache->mvcc_snapshot->snapshot_fnc (thread_p, &mvcc_header,
+						      scan_cache->mvcc_snapshot) == SNAPSHOT_SATISFIED
+	  && recdes != NULL)
```
wrlawodms: Can it be NULL?

xmilex-git: In heap_get_visible_version_internal (),

>   if (context->recdes_p != NULL)
>     {
>       scan = heap_get_record_data_when_all_ready (thread_p, context);
>     }

The part I referred to is the if statement in heap_get_visible_version_internal() that checks if recdes_p is null.

xmilex-git: It seems like it would be okay to check with assert(recdes != NULL); and remove this condition. I will make the modifications and commit.

src/storage/heap_file.c
xmilex-git: @wrlawodms 
This is a very good code modification suggestion. You implemented what I was trying to do. Thank you. I have applied the modification and committed it.

src/storage/heap_file.c
```diff
@@ -24863,11 +24850,51 @@ heap_get_visible_version (THREAD_ENTRY * thread_p, const OID * oid, OID * class_
...
+    {
+      MVCC_REC_HEADER mvcc_header = MVCC_REC_HEADER_INITIALIZER;
+
+      assert (scan_cache != NULL);
+      assert (recdes != NULL);
+
+      if (or_mvcc_get_header (forward_recdes, &mvcc_header) != NO_ERROR)
+	{
+	  /* Unexpected. */
+	  assert (false);
```
shparkcubrid, line 24881:
```suggestion
	  assert (false);
	  return S_ERROR;
```

shparkcubrid (approved):

src/storage/heap_file.c
```diff
@@ -8110,16 +8101,11 @@ heap_next_internal (THREAD_ENTRY * thread_p, const HFID * hfid, OID * class_oid,
...
-	{
-	  pgbuf_ordered_unfix (thread_p, &curr_page_watcher);
-	}
-      else
-	{
-	  pgbuf_replace_watcher (thread_p, &curr_page_watcher, &scan_cache->page_watcher);
-	}
+      {
+	pgbuf_ordered_unfix (thread_p, &scan_cache->page_watcher);
+      }
```
```suggestion
	pgbuf_ordered_unfix (thread_p, &scan_cache->page_watcher);
```

src/storage/heap_file.c
```diff
@@ -24863,11 +24850,52 @@ heap_get_visible_version (THREAD_ENTRY * thread_p, const OID * oid, OID * class_
 */
 SCAN_CODE
 heap_scan_get_visible_version (THREAD_ENTRY * thread_p, const OID * oid, OID * class_oid, RECDES * recdes,
-			       HEAP_SCANCACHE * scan_cache, int ispeeking, int old_chn)
+			       RECDES * forward_recdes, HEAP_SCANCACHE * scan_cache, int ispeeking, int old_chn)
```
wrlawodms: Nit. 
I'd name it in the view of this function. At your choice.
```suggestion
			       RECDES * peeked_recdes, HEAP_SCANCACHE * scan_cache, int ispeeking, int old_chn)
```

wrlawodms (approved):
Looks good to me. :)

src/storage/heap_file.c
```diff
@@ -24863,11 +24848,52 @@ heap_get_visible_version (THREAD_ENTRY * thread_p, const OID * oid, OID * class_
...
+   * performance closer to the heap scan performance of CUBRID before the introduction of MVCC. 
+   * Following is the explanation for the code below.
+   * Before fetching a record, check peeked_recdes to see if the record type is REC_HOME,
+   * and it's being PEEKed (meaning there's no need to COPY the record data to a new space). 
+   * In this case, we can use peeked_recdes as the record without executing the
+   * heap_get_visible_version_internal() function. If the conditions above are not met,
+   * or the mvcc_snapshot does not satisfy, then carry out the necessary steps through 
+   * the heap_get_visible_version_internal() function.
+   */
+  if (peeked_recdes->type == REC_HOME && ispeeking == PEEK)
```
youngjinj, line 24870: I think peeked_recdes should be checked to see if it is null.
Since heap_scan_get_visible_version is only called from heap_next_internal, the following code may be sufficient.
```
assert (peeked_recdes != NULL);
```

youngjinj (approved):

src/storage/heap_file.c
xmilex-git: This is a good suggestion. I will modify it and commit.

PR #4736, opened 2023-10-19
[LETS-808] Leave a log record when the transaction acquire SCH_M_LOCK in ATS

joohok:
http://jira.cubrid.org/browse/LETS-808

Purpose

* Define a new log record type for replicating the lock that DDL statement acquires
* Leave a new log record when SCH_M_LOCK is acquired when DDL is executed
  * excluded cases 
    * lock the object to reserve the classname
    * lock the object to check if it exists in the server or not
    * lock the object in SA_MODE like when it create system classes during `createdb` 

Remark
* all DDL operations, except CREATE TABLE/VIEW, acquire locks using `locator_fetch()`
  * Hence, I had no choice but to determine whether SCH_M_LOCK is acquired within `locator_fetch()`
* For CREATE TABLE/VIEW, it acquire SCH_M_LOCK through `heap_get_insert_location_with_lock()`
* when not in SERVER_MODE, the path for acquiring locks is also taken during the execution of createdb, which is why #if defined (SERVER_MODE) was used.

### DIFF
--- a/src/storage/heap_file.c
+++ b/src/storage/heap_file.c
@@ -20433,16 +20433,20 @@ heap_get_insert_location_with_lock (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONT
       lk_result = lock_object (thread_p, &context->res_oid, &context->class_oid, lock, LK_COND_LOCK);
       if (lk_result == LK_GRANTED)
 	{
-	  /* if lock == SCH_M_LOCK, then it is the class creation */
 	  /* successfully locked! */
+#if defined (SERVER_MODE)
 	  if (lock == SCH_M_LOCK && OID_IS_ROOTOID (&context->class_oid))
 	    {
+	      /* if lock == SCH_M_LOCK, then it is the class creation */
 	      /* Schema modification lock is acquired to create Class or VClass */
+	      assert (!OID_ISNULL (&context->res_oid));
+
 	      if (is_active_transaction_server ())
 		{
 		  log_append_schema_modification_lock (thread_p, &context->res_oid);
 		}
 	    }
+#endif
 
 	  return NO_ERROR;
 	}
--- a/src/storage/heap_file.c
+++ b/src/storage/heap_file.c
@@ -20437,8 +20437,6 @@ heap_get_insert_location_with_lock (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONT
 #if defined (SERVER_MODE)
 	  if (lock == SCH_M_LOCK && OID_IS_ROOTOID (&context->class_oid))
 	    {
-	      /* if lock == SCH_M_LOCK, then it is the class creation */
-	      /* Schema modification lock is acquired to create Class or VClass */
 	      assert (!OID_ISNULL (&context->res_oid));
 
 	      if (is_active_transaction_server ())
--- a/src/storage/heap_file.c
+++ b/src/storage/heap_file.c
@@ -20433,7 +20433,17 @@ heap_get_insert_location_with_lock (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONT
       lk_result = lock_object (thread_p, &context->res_oid, &context->class_oid, lock, LK_COND_LOCK);
       if (lk_result == LK_GRANTED)
 	{
+	  /* if lock == SCH_M_LOCK, then it is the class creation */
 	  /* successfully locked! */
+	  if (lock == SCH_M_LOCK && OID_IS_ROOTOID (&context->class_oid))
+	    {
+	      /* Schema modification lock is acquired to create Class or VClass */
+	      if (is_active_transaction_server ())
+		{
+		  log_append_schema_modification_lock (thread_p, &context->res_oid);
+		}
+	    }
+
 	  return NO_ERROR;
 	}
       else if (lk_result != LK_NOTGRANTED_DUE_TIMEOUT)
--- a/src/storage/heap_file.c
+++ b/src/storage/heap_file.c
@@ -20438,11 +20438,9 @@ heap_get_insert_location_with_lock (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONT
 	  if (lock == SCH_M_LOCK && OID_IS_ROOTOID (&context->class_oid))
 	    {
 	      assert (!OID_ISNULL (&context->res_oid));
+	      assert (is_active_transaction_server ());
 
-	      if (is_active_transaction_server ())
-		{
-		  log_append_schema_modification_lock (thread_p, &context->res_oid);
-		}
+	      log_append_schema_modification_lock (thread_p, &context->res_oid);
 	    }
 #endif
 
--- a/src/transaction/locator_sr.c
+++ b/src/transaction/locator_sr.c
@@ -12885,6 +12885,15 @@ locator_lock_and_get_object_internal (THREAD_ENTRY * thread_p, HEAP_GET_CONTEXT 
   else
     {
       lock_acquired = true;
+
+      if (OID_IS_ROOTOID (context->class_oid_p) && lock_mode == SCH_M_LOCK)
+	{
+	  /* Lock is acquired to modify the class record (DDL is executed) */
+	  if (is_active_transaction_server ())
+	    {
+	      log_append_schema_modification_lock (thread_p, context->oid_p);
+	    }
+	}
     }
 
   assert (OID_IS_ROOTOID (context->class_oid_p) || lock_mode == S_LOCK || lock_mode == X_LOCK);
--- a/src/transaction/locator_sr.c
+++ b/src/transaction/locator_sr.c
@@ -2542,6 +2542,19 @@ xlocator_fetch (THREAD_ENTRY * thread_p, OID * oid, int chn, LOCK lock,
 		   * rules. This object must be also locked. */
 		  object_locked = true;
 		}
+
+#if defined (SERVER_MODE)
+	      if (lock == SCH_M_LOCK)
+		{
+		  assert (OID_IS_ROOTOID (class_oid));
+		  assert (!OID_ISNULL (p_oid));
+
+		  if (is_active_transaction_server ())
+		    {
+		      log_append_schema_modification_lock (thread_p, p_oid);
+		    }
+		}
+#endif
 	    }
 	  break;
 	}
@@ -12886,15 +12899,6 @@ locator_lock_and_get_object_internal (THREAD_ENTRY * thread_p, HEAP_GET_CONTEXT 
   else
     {
       lock_acquired = true;
-
-      if (OID_IS_ROOTOID (context->class_oid_p) && lock_mode == SCH_M_LOCK)
-	{
-	  /* Lock is acquired to modify the class record (DDL is executed) */
-	  if (is_active_transaction_server ())
-	    {
-	      log_append_schema_modification_lock (thread_p, context->oid_p);
-	    }
-	}
     }
 
   assert (OID_IS_ROOTOID (context->class_oid_p) || lock_mode == S_LOCK || lock_mode == X_LOCK);
--- a/src/transaction/locator_sr.c
+++ b/src/transaction/locator_sr.c
@@ -59,6 +59,7 @@
 #endif /* ENABLE_SYSTEMTAP */
 #include "record_descriptor.hpp"
 #include "slotted_page.h"
+#include "server_type.hpp"
 #include "xasl_cache.h"
 #include "xasl_predicate.hpp"
 #include "thread_manager.hpp"	// for thread_get_thread_entry_info
--- a/src/transaction/locator_sr.c
+++ b/src/transaction/locator_sr.c
@@ -2544,15 +2544,12 @@ xlocator_fetch (THREAD_ENTRY * thread_p, OID * oid, int chn, LOCK lock,
 		}
 
 #if defined (SERVER_MODE)
-	      if (lock == SCH_M_LOCK)
+	      if (lock == SCH_M_LOCK && is_active_transaction_server ())
 		{
 		  assert (OID_IS_ROOTOID (class_oid));
 		  assert (!OID_ISNULL (p_oid));
 
-		  if (is_active_transaction_server ())
-		    {
-		      log_append_schema_modification_lock (thread_p, p_oid);
-		    }
+		  log_append_schema_modification_lock (thread_p, p_oid);
 		}
 #endif
 	    }
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -3739,12 +3739,23 @@ log_append_schema_modification_lock (THREAD_ENTRY * thread_p, const OID * classo
 {
   assert (!OID_ISNULL (classoid));
 
-  LOG_TDES *tdes = LOG_FIND_CURRENT_TDES (thread_p);
-  assert (tdes != nullptr);
+  int tran_index = LOG_FIND_THREAD_TRAN_INDEX (thread_p);
+  LOG_TDES *tdes = LOG_FIND_TDES (tran_index);
+  if (tdes == nullptr)
+    {
+      assert (false);
+      er_set (ER_FATAL_ERROR_SEVERITY, ARG_FILE_LINE, ER_LOG_UNKNOWN_TRANINDEX, 1, tran_index);
+      return;
+    }
+
 
   LOG_PRIOR_NODE *node =
     prior_lsa_alloc_and_copy_data (thread_p, LOG_SCHEMA_MODIFICATION_LOCK, RV_NOT_DEFINED, NULL, 0, NULL, 0, NULL);
-  assert (node != nullptr);
+  if (node == nullptr)
+    {
+      assert (false);
+      return;
+    }
 
   auto record = (LOG_REC_SCHEMA_MODIFICATION_LOCK *) node->data_header;
   COPY_OID (&record->classoid, classoid);
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -7291,7 +7291,7 @@ log_dump_record_schema_modification_lock (THREAD_ENTRY * thread_p, FILE * out_fp
   /* Get the DATA HEADER */
   LOG_READ_ADVANCE_WHEN_DOESNT_FIT (thread_p, sizeof (*log_rec), log_lsa, log_page_p);
   log_rec = ((LOG_REC_SCHEMA_MODIFICATION_LOCK *) ((char *) log_page_p->area + log_lsa->offset));
-  fprintf (out_fp, "OID = %d|%d|%d \n", OID_AS_ARGS (&log_rec->classoid));
+  fprintf (out_fp, ", CLASSOID = %d|%d|%d\n", OID_AS_ARGS (&log_rec->classoid));
 
   return log_page_p;
 }
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -7291,7 +7291,7 @@ log_dump_record_schema_modification_lock (THREAD_ENTRY * thread_p, FILE * out_fp
   /* Get the DATA HEADER */
   LOG_READ_ADVANCE_WHEN_DOESNT_FIT (thread_p, sizeof (*log_rec), log_lsa, log_page_p);
   log_rec = ((LOG_REC_SCHEMA_MODIFICATION_LOCK *) ((char *) log_page_p->area + log_lsa->offset));
-  fprintf (out_fp, ", CLASSOID = %d|%d|%d\n", OID_AS_ARGS (&log_rec->classoid));
+  fprintf (out_fp, ", classoid = %d|%d|%d\n", OID_AS_ARGS (&log_rec->classoid));
 
   return log_page_p;
 }
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -3748,7 +3748,6 @@ log_append_schema_modification_lock (THREAD_ENTRY * thread_p, const OID * classo
       return;
     }
 
-
   LOG_PRIOR_NODE *node =
     prior_lsa_alloc_and_copy_data (thread_p, LOG_SCHEMA_MODIFICATION_LOCK, RV_NOT_DEFINED, NULL, 0, NULL, 0, NULL);
   if (node == nullptr)
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -3735,9 +3735,9 @@ log_append_assigned_mvccid (THREAD_ENTRY * thread_p, MVCCID mvccid)
 }
 
 void
-log_append_schema_modification_lock (THREAD_ENTRY * thread_p, const OID classoid)
+log_append_schema_modification_lock (THREAD_ENTRY * thread_p, const OID * classoid)
 {
-  assert (!OID_ISNULL (&classoid));
+  assert (!OID_ISNULL (classoid));
 
   LOG_TDES *tdes = LOG_FIND_CURRENT_TDES (thread_p);
   assert (tdes != nullptr);
@@ -3747,7 +3747,7 @@ log_append_schema_modification_lock (THREAD_ENTRY * thread_p, const OID classoid
   assert (node != nullptr);
 
   auto record = (LOG_REC_SCHEMA_MODIFICATION_LOCK *) node->data_header;
-  COPY_OID (&record->classoid, &classoid);
+  COPY_OID (&record->classoid, classoid);
 
   (void) prior_lsa_next_record (thread_p, node, tdes);
 }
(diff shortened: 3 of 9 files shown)


src/transaction/log_replication_atomic.cpp
```diff
@@ -216,6 +217,28 @@ namespace cublog
...
+	    const LOG_REC_SCHEMA_MODIFICATION_LOCK log_rec =
+		    m_redo_context.m_reader.reinterpret_copy_and_add_align<LOG_REC_SCHEMA_MODIFICATION_LOCK> ();
+
+	    /* TODO:
+	     * All these debug logging part will be removed.
+	     * Lock will be acquired for the class that log_rec.classoid indicates
+	     */
+	    int error = heap_get_class_name (&thread_entry, &log_rec.classoid, &classname);
+	    _er_log_debug (ARG_FILE_LINE,"[REPL LOCK] Schema modification lock is aquired on %s (OID = %d|%d|%d)\n",
+			   OID_AS_ARGS (&log_rec.classoid), error != NO_ERROR ? "null" : classname);
```
```suggestion
			   error != NO_ERROR ? "null" : classname, OID_AS_ARGS (&log_rec.classoid));
```

src/storage/heap_file.c
```diff
@@ -20434,6 +20434,20 @@ heap_get_insert_location_with_lock (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONT
       if (lk_result == LK_GRANTED)
 	{
 	  /* successfully locked! */
+#if defined (SERVER_MODE)
+	  if (lock == SCH_M_LOCK && OID_IS_ROOTOID (&context->class_oid))
+	    {
+	      /* if lock == SCH_M_LOCK, then it is the class creation */
+	      /* Schema modification lock is acquired to create Class or VClass */
+	      assert (!OID_ISNULL (&context->res_oid));
```
hornetmj, line 20440: Nit:

I understand your intention however, I think it can remove when I read the previous few lines.

hornetmj (approved):
LGTM.

src/transaction/log_replication_atomic.cpp
joohok: Oh.. Thanks

src/storage/heap_file.c
joohok: OK, I remove it

src/storage/heap_file.c
```diff
@@ -20434,6 +20434,18 @@ heap_get_insert_location_with_lock (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONT
       if (lk_result == LK_GRANTED)
 	{
 	  /* successfully locked! */
+#if defined (SERVER_MODE)
+	  if (lock == SCH_M_LOCK && OID_IS_ROOTOID (&context->class_oid))
+	    {
+	      assert (!OID_ISNULL (&context->res_oid));
+
+	      if (is_active_transaction_server ())
```
wrlawodms: Can any other server types reach here? Do they acquire a lock? If it's not, it's enough with `assert (is_active_transaction_server())`. If it's not, please share the scenario.

src/transaction/log_manager.c
```diff
@@ -7363,6 +7399,10 @@ log_dump_record (THREAD_ENTRY * thread_p, FILE * out_fp, LOG_RECTYPE record_type
       log_page_p = log_dump_record_supplemental_info (thread_p, out_fp, log_lsa, log_page_p);
       break;
 
+    case LOG_SCHEMA_MODIFICATION_LOCK:
+      log_page_p = log_dump_record_schema_modification_lock (thread_p, out_fp, log_lsa, log_page_p);
```
wrlawodms, line 7413: This has to be tested. I'm mentioning since this is not on the A/C list.

src/transaction/log_replication_atomic.cpp
```diff
@@ -216,6 +217,28 @@ namespace cublog
...
+	    m_redo_context.m_reader.advance_when_does_not_fit (sizeof (LOG_REC_SCHEMA_MODIFICATION_LOCK));
+	    const LOG_REC_SCHEMA_MODIFICATION_LOCK log_rec =
+		    m_redo_context.m_reader.reinterpret_copy_and_add_align<LOG_REC_SCHEMA_MODIFICATION_LOCK> ();
+
+	    /* TODO:
+	     * All these debug logging part will be removed.
+	     * Lock will be acquired for the class that log_rec.classoid indicates
+	     */
+	    int error = heap_get_class_name (&thread_entry, &log_rec.classoid, &classname);
+	    _er_log_debug (ARG_FILE_LINE,"[REPL LOCK] Schema modification lock is aquired on %s (OID = %d|%d|%d)\n",
```
```suggestion
	    _er_log_debug (ARG_FILE_LINE,"[REPL_LOCK] Schema modification lock is aquired on %s (OID = %d|%d|%d)\n",
```
To make it easy for text-search. It was recommended by Cristi. :)

src/transaction/log_replication_atomic.cpp
```diff
@@ -216,6 +217,28 @@ namespace cublog
...
+
+	    m_redo_context.m_reader.advance_when_does_not_fit (sizeof (LOG_REC_SCHEMA_MODIFICATION_LOCK));
+	    const LOG_REC_SCHEMA_MODIFICATION_LOCK log_rec =
+		    m_redo_context.m_reader.reinterpret_copy_and_add_align<LOG_REC_SCHEMA_MODIFICATION_LOCK> ();
+
+	    /* TODO:
+	     * All these debug logging part will be removed.
+	     * Lock will be acquired for the class that log_rec.classoid indicates
+	     */
+	    int error = heap_get_class_name (&thread_entry, &log_rec.classoid, &classname);
```
wrlawodms, line 232: Can we guarantee that we can always get the classname here? I'm asking because I remember that I encountered a tricky case back when I treated the TRUNCATE improvement. When creating a table, it seems that sch_m_lock is acquired but the class record is not yet determined and it's determined later. 

In my quick analysis, when creating a table:

sch_m_lock is acquired at:
do_create_entity()->dbt_create_class()->dbt_reserve_name()->locator_reserve_class_name()->xlocator_reserve_class_names()-xlocator_reserve_class_name()->lock_object()

the classed seems to be assigned later at:
do_create_entity()->dbt_finish_class()->sm_finish_class()->update_class()

wrlawodms: In this case, we probably even can't believe the oid. We don't leave a log record for every sch_m_lock?

wrlawodms (commented):
One of my huge comment is about an excluded case.
Please elaborate on the excluded cases. Why are they excluded?
```
lock the object to reserve the classname
lock the object to check if it exists in the server or not
lock the object in SA_MODE like when it create system classes during createdb
```
Can we skip the lock even though ATS acquired the lock and did something?

wrlawodms:
a lock_object() on oid_Root_class_oid with SCH_M_LOCK. Please explain why it doesn't need the log record.
- xlocator_delete_class_name() -> lock_object(), seems similar when reserving class name.

joohok:
> One of my huge comment is about an excluded case. Please elaborate on the excluded cases. Why are they excluded?
> 
> ```
> lock the object to reserve the classname
> lock the object to check if it exists in the server or not
> lock the object in SA_MODE like when it create system classes during createdb
> ```
> 
> Can we skip the lock even though ATS acquired the lock and did something?

First of all, thank you for the detailed review. I believe there is no need to log every time a lock is acquired in ATS. 
Currently, I intend to log only when two conditions are met: 
1) the lock mode is SCH_M_LOCK, and 
2) the instance belongs to the Root class.

However, during the point where we reserve the classname, these two conditions are satisfied, but it tries to acquire a lock on an invalid instance. it acquire the lock on an instance with an OID of (-2, -2, -2), which is merely done to prevent any other transaction in ATS from having the same name. Other transactions will also attempt to acquire a lock on OID (-2, -2, -2) to reserve the class name. This lock is unnecessary in PTS because PTS cannot perform DDL directly.

Furthermore, locks are sometimes acquired to determine whether a class exists, but as soon as it is acquired, the lock is released. Therefore, this is not necessary in PTS either. The fact that class changes are logged means that the class exists.

Finally, locks are attempted in SA mode, but locks acquired in SA mode are not needed, so I have excluded logging for them.

joohok:
> a lock_object() on oid_Root_class_oid with SCH_M_LOCK. Please explain why it doesn't need the log record.
> 
> * xlocator_delete_class_name() -> lock_object(), seems similar when reserving class name.

This function is executed within the DROP statement. When the DROP statement is executed, 
1. it first acquires a lock on the instance (table) through locator_fetch(), 
2. then removes the heap, 
3. deletes the name(in xlocator_delete_class_name()) , and 
4. finally releases the lock
 
in other words, at the point where xlocator_delete_class_name() attempts to acquire the lock, the lock is already held.
Therefore, I thought there is no need to log the same action twice. I'm not sure it is advisable to log redundantly in this case. 

Since PTS also uses the same lock manager mechanism, there shouldn't be any functional issues.

src/transaction/log_manager.c
joohok: OK, I will add the scenario to the AC list

src/transaction/log_replication_atomic.cpp
joohok: You're absolutely right. During the CREATE TABLE operation, we only have information about where the class record will be inserted (OID), and we don't know the name at that point. The class name information will only be available at the commit point. In this case, only the OID information will be logged, 
and whether this OID corresponds to the actual class OID or not can only be determined by examining the debuggig message from other DDL statements on the same table or by leaving one more debugging message at the commit point.

In the xlocator_reserve_class_names() function, it attempt to obtain SCH_M_LOCK using an invalid OID (-2, -2, -2). Therefore, in this case, we do not log the lock

src/transaction/log_replication_atomic.cpp
joohok: Oh.. Thanks!

src/storage/heap_file.c
joohok: OK, I will change it. I thought PS can reach here, but it doesn't.

wrlawodms:
> However, during the point where we reserve the classname, these two conditions are satisfied, but it tries to acquire a lock on an invalid instance. it acquire the lock on an instance with an OID of (-2, -2, -2), which is merely done to prevent any other transaction in ATS from having the same name. Other transactions will also attempt to acquire a lock on OID (-2, -2, -2) to reserve the class name. This lock is unnecessary in PTS because PTS cannot perform DDL directly.

It makes sense. One question: When is the lock on the special OID (-2,-2,-2) released? Does it follow the general rule? I mean, is it released when the transaction that holds the lock ends?

> Finally, locks are attempted in SA mode, but locks acquired in SA mode are not needed, so I have excluded logging for them.

I think it's needed even if locks are held on SA mode because trxs on PTS would want to access. Or, do you mean it's due to SA mode will be removed on LETS?

> in other words, at the point where xlocator_delete_class_name() attempts to acquire the lock, the lock is already held.

From what you said, I think it means we don't need to hold the lock on ATS because it's redundant. Is it correct?

src/transaction/log_replication_atomic.cpp
```diff
@@ -216,6 +217,28 @@ namespace cublog
...
 	  }
+	  case LOG_SCHEMA_MODIFICATION_LOCK:
+	  {
+	    char *classname = NULL;
+
+	    m_redo_context.m_reader.advance_when_does_not_fit (sizeof (LOG_REC_SCHEMA_MODIFICATION_LOCK));
+	    const LOG_REC_SCHEMA_MODIFICATION_LOCK log_rec =
+		    m_redo_context.m_reader.reinterpret_copy_and_add_align<LOG_REC_SCHEMA_MODIFICATION_LOCK> ();
+
+	    /* TODO:
```
wrlawodms, line 228: Is going to be removed at the end of the epic? Otherwise, it should be wrapped with the DEBUG definition.

joohok: Right, this part will be removed as we progress with the epic. It was added to demonstrate replication on the PTS and to test if it satisfies the AC

joohok:
> It makes sense. One question: When is the lock on the special OID (-2,-2,-2) released? Does it follow the general rule? I mean, is it released when the transaction that holds the lock ends?

right, it is released when the transaction is committed or aborted.
 
> I think it's needed even if locks are held on SA mode because trxs on PTS would want to access. Or, do you mean it's due to SA mode will be removed on LETS?

I thought that there wouldn't be a need for a lock in SA mode. This was because I believed that in SA mode, multiple transactions couldn't run concurrently. In the current code, when it's in SA_MODE, lock_object() assumes it always successfully acquires the lock and returns it to the caller because there will be no concurrent transaction. 

Additionally, as you mentioned, there hasn't been a specification yet on whether transactions performed by ATS in SA mode will be immediately replicated to PTS in LETS. Because of this uncertainty, I attempted to avoid logging in SA mode.

> From what you said, I think it means we don't need to hold the lock on ATS because it's redundant. Is it correct?

Oh yes. Since transaction in ATS already held the SCH_M_LOCK lock, it does not acquire a lock again when it tries to acquire a same lock in xlocator_delete_class_name().

wrlawodms:
> right, it is release when the transaction is committed or aborted.

It means only one transaction can reserve a class name at a time. This means, in turn, only one transaction can create a table at a time and others wait for the transaction to be completed. It doesn't make sense, does it? What am I missing?

> I thought that there wouldn't be a need for a lock in SA mode. This was because I believed that in SA mode, multiple transactions couldn't run concurrently. In the current code, when it's in SA_MODE, lock_object() assumes it always successfully acquires the lock and returns it to the caller because there will be no concurrent transaction.

I still think It's arguable, but it isn't worth it if we don't consider SA on LETS.

> Oh yes. Since transaction in ATS already held the SCH_M_LOCK lock, it does not acquire a lock again when it tries to acquire a same lock in xlocator_delete_class_name().

If it's certain, we should make a maintenance issue on CBRD to remove it.

joohok:
> It means only one transaction can reserve a class name at a time. This means, in turn, only one transaction can create a table at a time and others wait for the transaction to be completed. It doesn't make sense, does it? What am I missing?

I think my short explanation may have caused some confusion.  Sorry! 
I will provide a more detailed explanation of the process of reserving a classname:

1. Allocate an entry in the locator_Mht_classnames.
2. Assign a classname and a dummy OID to the allocated entry.
3. Obtain an SCH_M_LOCK for the dummy OID.
The dummy OID is assigned as (-2, -2, **-2 - N**) each time an entry is allocated. In other words, when reserving a classname with a different name, OID (-2, -2, -3) is assigned to the entry.

Subsequently, other transactions check if the classname is in the cache and, if it is, they attempt to acquire a lock using the OID stored in the cache entry (in order to wait for the completion of a transaction that is already reserving it).

Due to the described mechanism, multiple CREATE TABLE statements can be executed simultaneously

> I still think It's arguable, but it isn't worth it if we don't consider SA on LETS.

I agree. As I know we don't consider SA_MODE on LETS. @hornetmj  is this right ?

joohok:
> a lock_object() on oid_Root_class_oid with SCH_M_LOCK. Please explain why it doesn't need the log record.
> 
> * xlocator_delete_class_name() -> lock_object(), seems similar when reserving class name.

I'm truly sorry. I completely confused it with another function T_T..... The part where lock_object() is mentioned in the function is used to wait for another transaction that is dropping the object. If that transaction aborts, it initiates the process of removing the classname again. If the transaction commits, it throws an error.

joohok:
> I still think It's arguable, but it isn't worth it if we don't consider SA on LETS.

As @hornetmj  said, SA_MODE will be excluded on LETS.

src/transaction/locator_sr.c
```diff
@@ -2541,6 +2542,19 @@ xlocator_fetch (THREAD_ENTRY * thread_p, OID * oid, int chn, LOCK lock,
...
+
+#if defined (SERVER_MODE)
+	      if (lock == SCH_M_LOCK)
+		{
+		  assert (OID_IS_ROOTOID (class_oid));
+		  assert (!OID_ISNULL (p_oid));
+
+		  if (is_active_transaction_server ())
+		    {
+		      log_append_schema_modification_lock (thread_p, p_oid);
```
cristiarg: Between the moment that the `SCH_M_LOCK` on the `p_oid` is obtained and the moment that a first log record for the transaction is actually added to the log, there could be:
  - a large pause in time - let's say, in a worst case scenario
  - failed verification and conditions which would cause the actual modifying log records to not be added (eg: the DDL modifying transaction to fail to validate for some reason) - resulting in a false negative on the PTS side

These are both unlikely scenarios but they can occur.
The first bullet above has the potential to stall a lot of PTS client transactions that could - otherwise - to their job just fine. And as far as I see, there would be no race condition as to the consistency of the log because, once the object is locked on ATS, no other ATS transaction can change it
Isn't it better to just collect the information as to what objects have been locked in the transaction descriptor - similar to `tx_transient_class_registry m_modified_classes` - but more with the semantics of `tx_transient_class_registry m_locked_classes` - and only add the schema-modification-lock log record upon actually adding the first log record of that transaction - eg: around `prior_lsa_next_record_internal`?

src/transaction/locator_sr.c
```diff
@@ -2541,6 +2542,19 @@ xlocator_fetch (THREAD_ENTRY * thread_p, OID * oid, int chn, LOCK lock,
...
 		  object_locked = true;
 		}
+
+#if defined (SERVER_MODE)
+	      if (lock == SCH_M_LOCK)
+		{
+		  assert (OID_IS_ROOTOID (class_oid));
+		  assert (!OID_ISNULL (p_oid));
+
+		  if (is_active_transaction_server ())
```
cristiarg:
```suggestion
	      if (lock == SCH_M_LOCK && if (is_active_transaction_server ()))
		{
		  assert (OID_IS_ROOTOID (class_oid));
		  assert (!OID_ISNULL (p_oid));
		  ....
```

src/transaction/log_manager.c
```diff
@@ -7260,6 +7282,20 @@ log_dump_record_supplemental_info (THREAD_ENTRY * thread_p, FILE * out_fp, LOG_L
...
+static LOG_PAGE *
+log_dump_record_schema_modification_lock (THREAD_ENTRY * thread_p, FILE * out_fp, LOG_LSA * log_lsa,
+					  LOG_PAGE * log_page_p)
+{
+  LOG_REC_SCHEMA_MODIFICATION_LOCK *log_rec;
+
+  /* Get the DATA HEADER */
+  LOG_READ_ADVANCE_WHEN_DOESNT_FIT (thread_p, sizeof (*log_rec), log_lsa, log_page_p);
+  log_rec = ((LOG_REC_SCHEMA_MODIFICATION_LOCK *) ((char *) log_page_p->area + log_lsa->offset));
+  fprintf (out_fp, ", CLASSOID = %d|%d|%d\n", OID_AS_ARGS (&log_rec->classoid));
```
cristiarg: Nit.
From experience, there's way too much uppercase in the log dump :)
```suggestion
  fprintf (out_fp, ", classoid = %d|%d|%d\n", OID_AS_ARGS (&log_rec->classoid));
```

Some values are good to be uppoercase:
```
LSA = 105818|1248, Forw log = 105818|1368, Backw log = 105818|1176,
     Trid =  -3, Prev tran logrec = 105818|776
     Type = LOG_SYSOP_START_POSTPONE,
     lastparent_lsa = -1|-1, prv_topresult_lsa = -1|-1, type = LOG_SYSOP_END_COMMIT 
     posp_lsa = 105818|776 
```
but not the labels

src/transaction/log_manager.c
```diff
@@ -3730,6 +3734,24 @@ log_append_assigned_mvccid (THREAD_ENTRY * thread_p, MVCCID mvccid)
   (void) prior_lsa_next_record (thread_p, node, tdes);
 }
 
+void
+log_append_schema_modification_lock (THREAD_ENTRY * thread_p, const OID * classoid)
+{
+  assert (!OID_ISNULL (classoid));
+
+  LOG_TDES *tdes = LOG_FIND_CURRENT_TDES (thread_p);
+  assert (tdes != nullptr);
```
cristiarg: In old `log_append_..` functions, these cases are treated with a return.
In case this happens in release builds, the server would just crash below in the `prior_lsa_next_record` where `tdes` is expected to have a value.
I think the guideline is that "the server should never crash; it can abort some transactions, but it should never crash".

Share your opinion please.

src/transaction/locator_sr.c
cristiarg: @wrlawodms @joohok 
I think some of the aspects discussed by you above regarding to [possible] multiple logging of schema modification log records might be avoided with a lazy approach as described above. Can also be left out as a possible improvement.

src/transaction/log_manager.c
joohok: The server should never crash, but that doesn't mean that changes made by transactions should go un-logged. I firmly believe that logs must always be preserved. However, in the current cubrid implementation, failing to append logs does not result in transaction abortion. In other words, even if there are changes, logs are not preserved. If there were a mechanism to abort a transaction when log appending fails, error codes would have been returned instead of using asserts.

The server should never crash, but ultimately, if logs are not recorded, it could potentially become the root cause of issues. Therefore, I think it seems advisable to proceed with the assumption that logs are always preserved during implementation. Afterward, a more comprehensive revision may be necessary

What do you think?

src/transaction/log_manager.c
joohok: OK, I agree.

src/transaction/locator_sr.c
joohok: It seems like this could be an optimization point. I think I understand the concerns and the proposed solution.

If waiting for class modifications takes time after a lock for a class is obtained through `locator_fetch()`, PTS might have no choice but to wait. 
However, I believe that logging actions should occur where they take place. 
I think the approach of appending schema modification lock log within the log appending function seems unconventional.

There might also be a way to manage this from the PTS replicator's perspective. Instead of locking immediately when meeting a schema modification lock log, it could wait until one or more logs for that transaction have been replicated.

It appears that there is room for optimization regarding the gap between locking and actual changes within the PTS. I'd like to pursue this during the optimization phase. What are your thoughts on how to proceed?
Thanks.

src/transaction/locator_sr.c
joohok: OK, I changed

src/transaction/locator_sr.c
cristiarg: > There might also be a way to manage this from the PTS replicator's perspective. Instead of locking immediately when meeting a schema modification lock log, it could wait until one or more logs for that transaction have been replicated.

True.
The implementation of lazy-application of log records on the PTS side would be to the same end effect - not stalling PTS client read-only transactions - and would even be more localized - only in the bookkeeping & acquiring locks implementation.

src/transaction/log_manager.c
cristiarg: I think that the `log_append_...` functions failing should return some error code that should only affect the requesting transaction (or, if it's a system transaction, should be dealt-with with some retry mechanism or else ...).
But it is out of scope for us.
I personally would return as the older functions do.
Assert has no value in production and would just crash the server.
Up to you. :)

joohok: Alright, I understand. I will make the modification to return in case the pointer value is null, just like in `log_append_redo_crumbs()`.

src/transaction/log_manager.c
```diff
@@ -3730,6 +3734,35 @@ log_append_assigned_mvccid (THREAD_ENTRY * thread_p, MVCCID mvccid)
   (void) prior_lsa_next_record (thread_p, node, tdes);
 }
 
+void
+log_append_schema_modification_lock (THREAD_ENTRY * thread_p, const OID * classoid)
+{
+  assert (!OID_ISNULL (classoid));
+
+  int tran_index = LOG_FIND_THREAD_TRAN_INDEX (thread_p);
+  LOG_TDES *tdes = LOG_FIND_TDES (tran_index);
```
wrlawodms, line 3743: Nit.
```suggestion
  LOG_TDES *tdes = LOG_FIND_CURRENT_TDES (thread_p);
```

src/transaction/log_manager.c
```diff
@@ -3730,6 +3734,35 @@ log_append_assigned_mvccid (THREAD_ENTRY * thread_p, MVCCID mvccid)
...
+  if (tdes == nullptr)
+    {
+      assert (false);
+      er_set (ER_FATAL_ERROR_SEVERITY, ARG_FILE_LINE, ER_LOG_UNKNOWN_TRANINDEX, 1, tran_index);
+      return;
+    }
+
+
```
```suggestion
```

wrlawodms (approved):
It seems that you've missed to check the new log record type on `- log_2pc_recovery_analysis_record`.

src/transaction/log_manager.c
joohok: I declared `tran_index` for use in the error message below

hornetmj:
@joohok Please merge to proceed with the next task.

PR #4744, opened 2023-10-30
[CBRD-25036] improve range scan in btree

ctshim:
http://jira.cubrid.org/browse/CBRD-25036

is_cur_key_decompressed in btree_scan
-  If true, cur_key is a complete key,  otherwise it must be combined with C_page_info.compress_key.

Variables of the BTREE_PAGE_PREFIX_INFO structure.
- reader_type:                  Type of use
- vpid:                               vpid of the current page; 
                                         If the VPID is the same, it means that reading occurred within the same page as the information contained in the current structure.
- leaf_lsa:                            LOG_LSA information of the current page
                                           Even if the VPID is the same, if the LOG_LSA information is different, it means that there has been a change in the page.
- is_midxkey:                      Whether the index key is MIDXKEY
- use_comparing:               Whether to use the read key for later comparison
- satisfied_range_in_page: If true, it guarantees that all keys on the current page satisfy the range condition.
- n_compress_size:            Size of common prefix 
- compress_key:                 DB_VALUE to store common prefix
- clear_compress_key:

IMPROVE_RANGE_SCAN_IN_BTREE definintion 
- IMPROVE_RANGE_SCAN_IN_BTREE is used to determine whether to activate the added function. This is maintained for performance comparison and will be organized after performance measurement.

Etc.
- Change pgbuf_page_has_changed() to an inline function


src/storage/btree.h
```diff
@@ -161,6 +161,56 @@ typedef enum bts_key_status BTS_KEY_STATUS;
...
+#if defined(IMPROVE_RANGE_SCAN_IN_BTREE)
+
+typedef enum
+{
+  en_reader_none = 0,
+  en_reader_range_scan,
+  en_reader_stat,
+  en_reader_capacity,
+  en_reader_type_max
+} READER_TYPE;
```
beyondykk9: By coding convention, enum's members are defined upper case only.

ctshim: OK, I will fix it.
Thank you.

src/storage/btree.h
```diff
@@ -161,6 +161,62 @@ typedef enum bts_key_status BTS_KEY_STATUS;
...
+       LSA_SET_NULL(&((pg_prefix)->leaf_lsa));               \
+       (pg_prefix)->n_compress_size = COMMON_PREFIX_UNKNOWN; \
+    }                                                        \
+} while(0)
+
+#else
+#define BTREE_PAGE_PREFIX_INFO  void*
+#define INIT_BTREE_PAGE_PREFIX_INFO(pg_prefix, type)
+#define RESET_BTREE_PAGE_PREFIX_INFO(pg_prefix)
+#endif
```
hornetmj, line 221:
```suggestion
#endif /* IMPROVE_RANGE_SCAN_IN_BTREE */
```

src/storage/btree.h
```diff
@@ -161,6 +161,62 @@ typedef enum bts_key_status BTS_KEY_STATUS;
...
+
 struct btree_scan
```
hornetmj, line 224: Move 'typedef struct btree_scan BTREE_SCAN; /* BTS */' to in front of struct btree_scan.

src/storage/btree.h
```diff
@@ -161,6 +161,62 @@ typedef enum bts_key_status BTS_KEY_STATUS;
...
+#if defined(IMPROVE_RANGE_SCAN_IN_BTREE)
+typedef enum
+{
+  READER_TYPE_NONE = 0,
+  READER_TYPE_RANGE_SCAN,
+  READER_TYPE_STAT,
+  READER_TYPE_CAPACITY,
+  READER_TYPE_MAX
+} READER_TYPE;
+typedef struct
```
hornetmj:
```suggestion

typedef struct btree_page_prefix_info BTREE_PAGE_PREFIX_INFO;
struct btree_page_prefix_info
```

src/storage/btree.h
```diff
@@ -192,6 +248,13 @@ struct btree_scan
   DB_VALUE cur_key;		/* current key value */
   bool clear_cur_key;		/* clear flag for current key value */
 
+  //---------------------------------------------------------------------------------------------   
+  /* IMPROVE_RANGE_SCAN_IN_BTREE */
+  bool is_cur_key_decompressed;	/* If true, cur_key is a complete key.
+				 * Otherwise it must be combined with C_page_info.compress_key. */
+  BTREE_PAGE_PREFIX_INFO C_page_info;
```
hornetmj, line 260:
```suggestion
  BTREE_PAGE_PREFIX_INFO cur_page_info;
```

src/storage/btree.h
```diff
@@ -161,6 +161,62 @@ typedef enum bts_key_status BTS_KEY_STATUS;
...
+
+#if defined(IMPROVE_RANGE_SCAN_IN_BTREE)
+typedef enum
+{
+  READER_TYPE_NONE = 0,
+  READER_TYPE_RANGE_SCAN,
+  READER_TYPE_STAT,
+  READER_TYPE_CAPACITY,
+  READER_TYPE_MAX
+} READER_TYPE;
```
hornetmj: How about adding BTREE_ prefix?

src/storage/btree.c
```diff
@@ -1730,6 +1730,16 @@ static int btree_rv_save_keyval_for_undo_two_objects (BTID_INT * btid, DB_VALUE
...
 
+#if defined(IMPROVE_RANGE_SCAN_IN_BTREE)
+static void btree_init_page_prefix_info (BTREE_SCAN * bt_scan, bool is_midxkey, bool is_deduplicate,
+					 READER_TYPE reader_type);
+static int btree_read_record_in_leafpage (THREAD_ENTRY * thread_p, PAGE_PTR pgptr, int copy_key, BTREE_SCAN * bts);
+
+static void btree_check_within_range_in_page (THREAD_ENTRY * thread_p, PAGE_PTR pgptr, BTREE_SCAN * bts);
+static void btree_make_complete_key_including_prefix (BTREE_SCAN * bts, int compress_size, DB_VALUE * compress_key);
+static void btree_check_complete_decompressed_key (BTREE_SCAN * bts);
+#endif
```
hornetmj, line 1741:
```suggestion
#endif /* IMPROVE_RANGE_SCAN_IN_BTREE */
```

src/storage/btree.c
```diff
@@ -4287,6 +4285,282 @@ btree_read_record (THREAD_ENTRY * thread_p, BTID_INT * btid, PAGE_PTR pgptr, REC
...
+static void
+btree_init_page_prefix_info (BTREE_SCAN * bts, bool is_midxkey, bool is_deduplicate, READER_TYPE reader_type)
+{
+  assert (bts != NULL);
+  assert (reader_type == READER_TYPE_RANGE_SCAN || reader_type == READER_TYPE_STAT
+	  || reader_type == READER_TYPE_CAPACITY);
+
+  INIT_BTREE_PAGE_PREFIX_INFO (&bts->C_page_info, reader_type);
+
+  bts->C_page_info.is_midxkey = is_midxkey;
```
hornetmj, line 4409: How about use a local variable like BTREE_PAGE_PREFIX_INFO *c_page_info = &bts->C_page_info? If c_ means current, then I prefer to use cur_page_info.

src/storage/btree.c
```diff
@@ -4287,6 +4285,282 @@ btree_read_record (THREAD_ENTRY * thread_p, BTID_INT * btid, PAGE_PTR pgptr, REC
...
+  bts->C_page_info.satisfied_range_in_page = false;
+  assert (bts->C_page_info.use_comparing == true);
```
hornetmj:
```suggestion
```

src/storage/btree.h
```diff
@@ -161,6 +161,62 @@ typedef enum bts_key_status BTS_KEY_STATUS;
...
+
+  bool is_midxkey;
+  bool use_comparing;		// key compare  
+  bool satisfied_range_in_page;	// If true, it guarantees that all keys on the current page satisfy the range condition.  
+} BTREE_PAGE_PREFIX_INFO;
+
+#define INIT_BTREE_PAGE_PREFIX_INFO(pg_prefix, type)   do {  \
+    assert((pg_prefix) != NULL);                             \
+    (pg_prefix)->reader_type = (type);                       \
+    (pg_prefix)->use_comparing = true;                       \
```
hornetmj, line 195: IMHO, the default value of use_comparing should be false. And if the condition is satisfied, it should be changed to true. This allows us to only track cases where use_comparing==true.

src/storage/btree.c
```diff
@@ -4287,6 +4285,282 @@ btree_read_record (THREAD_ENTRY * thread_p, BTID_INT * btid, PAGE_PTR pgptr, REC
...
+    {
+      return NO_ERROR;
+    }
+
+  if (page_info->reader_type == READER_TYPE_RANGE_SCAN)
+    {
+      assert (pgptr == bts->C_page && VPID_EQ (pgbuf_get_vpid_ptr (pgptr), &bts->C_vpid));
+      if (VPID_EQ (&(bts->C_vpid), &page_info->vpid) && LSA_EQ (&page_info->leaf_lsa, pgbuf_get_lsa (pgptr)))
+	{			// same page         
+	  assert (page_info->n_compress_size != COMMON_PREFIX_UNKNOWN);
```
hornetmj, line 4322:
```suggestion
```

src/storage/btree.h
```diff
@@ -161,6 +161,62 @@ typedef enum bts_key_status BTS_KEY_STATUS;
...
+  VPID vpid;
+  LOG_LSA leaf_lsa;
+
+  int n_compress_size;		// size of common prefix key parts.
+  DB_VALUE compress_key;	// common prefix key part 
+  bool clear_compress_key;
+
+  bool is_midxkey;
+  bool use_comparing;		// key compare  
+  bool satisfied_range_in_page;	// If true, it guarantees that all keys on the current page satisfy the range condition.
```
hornetmj, line 189: I think that the satisfied_range_in_page optimization is not related to the caching of the common prefix. So, it would good to me move it to BTREE_SCAN.

src/storage/btree.c
```diff
@@ -4287,6 +4285,282 @@ btree_read_record (THREAD_ENTRY * thread_p, BTID_INT * btid, PAGE_PTR pgptr, REC
...
+	  bts->is_cur_key_decompressed = true;
+	  return n_prefix;
+	}
+    }
+
+  return NO_ERROR;
+}
+
+static void
+btree_init_page_prefix_info (BTREE_SCAN * bts, bool is_midxkey, bool is_deduplicate, READER_TYPE reader_type)
```
hornetmj, line 4401: I think if you support both btree_init_page_prefix_info() function and INIT_BTREE_PAGE_PREFIX_INFO macro, then they make someone confused because of the name(...init...). I suggest 2 option for you.
1. remove INIT_BTREE_PAGE_PREFIX_INFO macro and btree_init_page_prefix_info take over it's initialization job. And bts->C_page_info.reader_type is just set to READER_TYPE_NONE in the BTREE_INIT_SCAN macro.
2. rename btree_init_page_prefix_info() to reduce the confussion.

I prefer the first way.

src/storage/btree.c
```diff
@@ -4287,6 +4285,282 @@ btree_read_record (THREAD_ENTRY * thread_p, BTID_INT * btid, PAGE_PTR pgptr, REC
...
+  assert (bts != NULL);
+  assert (reader_type == READER_TYPE_RANGE_SCAN || reader_type == READER_TYPE_STAT
+	  || reader_type == READER_TYPE_CAPACITY);
+
+  INIT_BTREE_PAGE_PREFIX_INFO (&bts->C_page_info, reader_type);
+
+  bts->C_page_info.is_midxkey = is_midxkey;
+  bts->C_page_info.satisfied_range_in_page = false;
+  assert (bts->C_page_info.use_comparing == true);
+  if (reader_type == READER_TYPE_RANGE_SCAN)
```
hornetmj, line 4412: When I see the following code in the btree_read_record_in_leafpage(),
```
4312   if (btree_leaf_is_flaged (&bts->key_record, BTREE_LEAF_RECORD_OVERFLOW_KEY)
4313       || btree_leaf_is_flaged (&bts->key_record, BTREE_LEAF_RECORD_FENCE) || !page_info->is_midxkey)
```

it is enough to set to bts->C_page_info.use_comparing = false and return before 4412 line if it is not a midxkey. It is more understandable at this moment.

src/storage/btree.c
```diff
@@ -4287,6 +4285,282 @@ btree_read_record (THREAD_ENTRY * thread_p, BTID_INT * btid, PAGE_PTR pgptr, REC
...
+    {
+      if (bts->key_range.range != GE_LE)
+	{
+	  bts->C_page_info.use_comparing = (bool) (bts->key_range.upper_key != NULL);
```
hornetmj, line 4416: When I check the following condition,
```
{range = GT_INF, lower_key = 0x7fc128047648, upper_key = 0x0, num_index_term = 1}
```
the bts->C_page_info.use_comparing is set to false. I think it should be set to true.

I expect that the btree_apply_key_range_and_filter() function will operate normally based on the following conditions, but it would be better to handle it consistently.
```
if (bts->key_range.upper_key == NULL || satisfied_range_all)
```

hornetmj (commented):
I think the two optimizations can be separated from each other. It doesn't seem appropriate to put everything in a structure (BTREE_PAGE_PREFIX_INFO). In particular, I think the satisfied_range_in_page member variable is not related to 'PAGE_PREFIX' scope. Can you separate the two optimizations into different PRs or sub-tasks and conduct a review? It's not easy to me to review because the two optimizations are mixed together.

src/storage/btree.h
ctshim: PAGE_PREFIX satisfied_range_in_page can only make sense if you can maintain relevant information.
I could divide it into leading issues and lagging issues, but for now I don't want to divide the issues into separate issues.

src/storage/btree.h
ctshim: OK, I will fix it.

src/storage/btree.h
ctshim: OK, Thank you.

src/storage/btree.h
ctshim: There are variables that were previously used such as C_vpid and C_page.
I think it makes sense to start with a capital "C_" because it is associated with these variables.

I'll keep this name unless there are other important reasons.

src/storage/btree.h
ctshim: Thank you, I agree.

src/storage/btree.c
ctshim: Thank you.

src/storage/btree.c
ctshim: OK, Thank you.

src/storage/btree.h
ctshim: satisfied_range_in_page is meaningless if you can't determine whether it's the same page or not.
I will leave this variable in its current location.

src/storage/btree.c
ctshim: I will reflect your opinion when finally applying IMPROVE_RANGE_SCAN_IN_BTREE define by default without declaration.
Thank you.

src/storage/btree.c
ctshim: Thank you,
I will reflect your opinion.

src/storage/btree.c
ctshim: Thank you.

ctshim:
It was decided to proceed by dividing it into several sub-tasks.
I will close this PR and create a new one to request it.

PR #4745, opened 2023-10-30
[LETS-809] Acquire and release the locks in PTS for ddl replication

joohok:
http://jira.cubrid.org/browse/LETS-809

Purpose

During the process of applying DDL changes, the replicator and read transactions cannot simultaneously access the same object. Therefore, Replicator acquire SCH_M_LOCK on the class when LOG_SCHEMA_MODIFICATION_LOCK is replicated in PTS. 
 Furthermore, since it's possible to perform multiple DDL operations within a single transaction, the replicator needs to store locks taken per transaction.

tested on #4736

### DIFF
--- a/src/transaction/lock_manager.c
+++ b/src/transaction/lock_manager.c
@@ -6631,18 +6631,15 @@ lock_unlock_object_and_cleanup (THREAD_ENTRY * thread_p, const OID * oid, const 
 #if !defined (SERVER_MODE)
   return;
 #else /* !SERVER_MODE */
-  int tran_index;
-  LK_TRAN_LOCK *tran_lock;
-
-  tran_index = LOG_FIND_THREAD_TRAN_INDEX (thread_p);
-  tran_lock = &lk_Gl.tran_lock_table[tran_index];
+  const int tran_index = LOG_FIND_THREAD_TRAN_INDEX (thread_p);
+  LK_TRAN_LOCK *tran_lock = &lk_Gl.tran_lock_table[tran_index];
 
   assert (tran_lock->non2pl_list == NULL);
 
   lock_unlock_object_lock_internal (thread_p, oid, class_oid, lock, true, false);
 
   /* TODO:
-   * We have to determine if deadlock information should be cleared
+   * We have to decide if deadlock information should be cleared
    * (like lock_clear_deadlock_victim () in lock_unlock_all ())
    * this will be handled when handling the deadlock for replicator
    */
--- a/src/transaction/lock_manager.c
+++ b/src/transaction/lock_manager.c
@@ -6632,7 +6632,7 @@ lock_unlock_object_and_cleanup (THREAD_ENTRY * thread_p, const OID * oid, const 
   return;
 #else /* !SERVER_MODE */
   const int tran_index = LOG_FIND_THREAD_TRAN_INDEX (thread_p);
-  LK_TRAN_LOCK *tran_lock = &lk_Gl.tran_lock_table[tran_index];
+  LK_TRAN_LOCK *const tran_lock = &lk_Gl.tran_lock_table[tran_index];
 
   assert (tran_lock->non2pl_list == NULL);
 
--- a/src/transaction/lock_manager.c
+++ b/src/transaction/lock_manager.c
@@ -6615,6 +6615,41 @@ lock_unlock_object_donot_move_to_non2pl (THREAD_ENTRY * thread_p, const OID * oi
 }
 
 /*
+ * lock_unlock_object_and_cleaup - Unlock an object lock on the specified object,
+ *                                 release lock entry and do not move to non2pl list
+ *                                 Used by replicator for PTS (atomic_replicator)
+ *   return:
+ *   thread_p(in):
+ *   oid(in):  Identifier of instance to unlock from
+ *   class_oid(in): Identifier of the class of the instance
+ *   lock(in): Lock to release
+ *
+ */
+void
+lock_unlock_object_and_cleanup (THREAD_ENTRY * thread_p, const OID * oid, const OID * class_oid, LOCK lock)
+{
+#if !defined (SERVER_MODE)
+  return;
+#else /* !SERVER_MODE */
+  int tran_index;
+  LK_TRAN_LOCK *tran_lock;
+
+  tran_index = LOG_FIND_THREAD_TRAN_INDEX (thread_p);
+  tran_lock = &lk_Gl.tran_lock_table[tran_index];
+
+  assert (tran_lock->non2pl_list == NULL);
+
+  lock_unlock_object_lock_internal (thread_p, oid, class_oid, lock, true, false);
+
+  /* TODO:
+   * We have to determine if deadlock information should be cleared
+   * (like lock_clear_deadlock_victim () in lock_unlock_all ())
+   * this will be handled when handling the deadlock for replicator
+   */
+#endif
+}
+
+/*
  * lock_unlock_object - Unlock an object according to transaction isolation level
  *
  * return: nothing..
--- a/src/transaction/log_replication_atomic.cpp
+++ b/src/transaction/log_replication_atomic.cpp
@@ -21,6 +21,7 @@
 #include "log_replication_jobs.hpp"
 
 #include "log_recovery_redo_parallel.hpp"
+#include "heap_file.h"
 
 namespace cublog
 {
@@ -214,6 +215,20 @@ namespace cublog
 	      {
 		m_replicator_mvccid->new_assigned_mvccid (header.trid, log_rec.mvccid);
 	      }
+	    break;
+	  }
+	  case LOG_SCHEMA_MODIFICATION_LOCK:
+	  {
+	    char *classname;
+
+	    m_redo_context.m_reader.advance_when_does_not_fit (sizeof (LOG_REC_SCHEMA_MODIFICATION_LOCK));
+	    const LOG_REC_SCHEMA_MODIFICATION_LOCK log_rec =
+		    m_redo_context.m_reader.reinterpret_copy_and_add_align<LOG_REC_SCHEMA_MODIFICATION_LOCK> ();
+
+	    int error = heap_get_class_name (&thread_entry, &log_rec.classoid, &classname);
+	    _er_log_debug (ARG_FILE_LINE,"[REPL LOCK] OID = %d|%d|%d, classname = %s\n", OID_AS_ARGS (&log_rec.classoid),
+			   classname);
+	    free_and_init (classname);
 	    break;
 	  }
 	  default:
--- a/src/transaction/log_replication_atomic.cpp
+++ b/src/transaction/log_replication_atomic.cpp
@@ -16,7 +16,6 @@
  *
  */
 
-#include "lock_manager.h"
 #include "log_replication.cpp.hpp"
 #include "log_replication_atomic.hpp"
 #include "log_replication_jobs.hpp"
@@ -401,6 +400,17 @@ namespace cublog
       }
 
     m_locked_classes.emplace (trid, *classoid);
-  }
 
+#if !defined (NDEBUG)
+    /* there will be no duplicated object in the map */
+    auto [begin, end] = m_locked_classes.equal_range (trid);
+    for (auto it = begin; it != end; it++)
+      {
+	if (std::count (begin, end, *it) > 1)
+	  {
+	    assert (false);
+	  }
+      }
+#endif
+  }
 }
--- a/src/transaction/log_replication_atomic.cpp
+++ b/src/transaction/log_replication_atomic.cpp
@@ -234,24 +234,12 @@ namespace cublog
 	  }
 	  case LOG_SCHEMA_MODIFICATION_LOCK:
 	  {
-	    char *classname = NULL;
-
 	    m_redo_context.m_reader.advance_when_does_not_fit (sizeof (LOG_REC_SCHEMA_MODIFICATION_LOCK));
 	    const LOG_REC_SCHEMA_MODIFICATION_LOCK log_rec =
 		    m_redo_context.m_reader.reinterpret_copy_and_add_align<LOG_REC_SCHEMA_MODIFICATION_LOCK> ();
 
-	    /* TODO:
-	     * All these debug logging part will be removed.
-	     * Lock will be acquired for the class that log_rec.classoid indicates
-	     */
-	    int error = heap_get_class_name (&thread_entry, &log_rec.classoid, &classname);
-	    _er_log_debug (ARG_FILE_LINE,"[REPL_LOCK] Schema modification lock is aquired on %s (OID = %d|%d|%d)\n",
-			   error != NO_ERROR ? "null" : classname, OID_AS_ARGS (&log_rec.classoid));
+	    acquire_lock (thread_entry, header.trid, &log_rec.classoid);
 
-	    if (classname != NULL)
-	      {
-		free_and_init (classname);
-	      }
 	    break;
 	  }
 	  default:
@@ -392,5 +380,16 @@ namespace cublog
       {
 	lock_unlock_object (&thread_entry, & (it->second), oid_Root_class_oid, SCH_M_LOCK, true);
       }
+
+    m_locked_objects.erase (trid);
   }
+
+  void
+  atomic_replicator::acquire_lock (cubthread::entry &thread_entry, const TRANID trid, const OID *classoid)
+  {
+    lock_object (&thread_entry, classoid, oid_Root_class_oid, SCH_M_LOCK, LK_UNCOND_LOCK);
+
+    m_locked_objects.emplace (trid, *classoid);
+  }
+
 }
--- a/src/transaction/log_replication_atomic.cpp
+++ b/src/transaction/log_replication_atomic.cpp
@@ -231,7 +231,7 @@ namespace cublog
 	     */
 	    int error = heap_get_class_name (&thread_entry, &log_rec.classoid, &classname);
 	    _er_log_debug (ARG_FILE_LINE,"[REPL LOCK] Schema modification lock is aquired on %s (OID = %d|%d|%d)\n",
-			   OID_AS_ARGS (&log_rec.classoid), error != NO_ERROR ? "null" : classname);
+			   error != NO_ERROR ? "null" : classname, OID_AS_ARGS (&log_rec.classoid));
 
 	    if (classname != NULL)
 	      {
--- a/src/transaction/log_replication_atomic.cpp
+++ b/src/transaction/log_replication_atomic.cpp
@@ -16,6 +16,7 @@
  *
  */
 
+#include "lock_manager.h"
 #include "log_replication.cpp.hpp"
 #include "log_replication_atomic.hpp"
 #include "log_replication_jobs.hpp"
@@ -377,7 +378,7 @@ namespace cublog
     auto [begin, end] = m_locked_classes.equal_range (trid);
     for (auto it = begin; it != end; it++)
       {
-	lock_unlock_object (&thread_entry, & (it->second), oid_Root_class_oid, SCH_M_LOCK, true);
+	lock_unlock_object_and_cleanup (&thread_entry, & (it->second), oid_Root_class_oid, SCH_M_LOCK);
       }
 
     m_locked_classes.erase (trid);
--- a/src/transaction/log_replication_atomic.cpp
+++ b/src/transaction/log_replication_atomic.cpp
@@ -20,9 +20,9 @@
 #include "log_replication_atomic.hpp"
 #include "log_replication_jobs.hpp"
 
-#include "oid.h"
-#include "log_recovery_redo_parallel.hpp"
 #include "heap_file.h"
+#include "log_recovery_redo_parallel.hpp"
+#include "oid.h"
 
 namespace cublog
 {
@@ -391,7 +391,16 @@ namespace cublog
   {
     assert (!OID_ISNULL (classoid) && !OID_ISTEMP (classoid));
 
-    lock_object (&thread_entry, classoid, oid_Root_class_oid, SCH_M_LOCK, LK_UNCOND_LOCK);
+    /* TODO:
+     * If a PTS read transaction holds a lock for an extended period without releasing it for the same class,
+     * the replicator could wait too long to acquire the lock.
+     * In such cases, there might be an introduction of a mechanism to abort read transaction that holds lock.
+     */
+
+    if (lock_object (&thread_entry, classoid, oid_Root_class_oid, SCH_M_LOCK, LK_UNCOND_LOCK) != LK_GRANTED)
+      {
+	assert_release (false);
+      }
 
     m_locked_classes.emplace (trid, *classoid);
   }
--- a/src/transaction/log_replication_atomic.cpp
+++ b/src/transaction/log_replication_atomic.cpp
@@ -43,7 +43,6 @@ namespace cublog
      * and, thus, does not need to be left in consistent state; thus, no check as to the consistent
      * termination state for atomic replication is needed
      */
-    assert (m_locked_classes.empty ());
   }
 
   void
--- a/src/transaction/log_replication_atomic.cpp
+++ b/src/transaction/log_replication_atomic.cpp
@@ -121,10 +121,9 @@ namespace cublog
 		m_replicator_mvccid->complete_mvcc (header.trid, replicator_mvcc::COMMITTED);
 	      }
 
-	    if (m_locked_classes.count (header.trid))
+	    if (m_locked_classes.count (header.trid) > 0)
 	      {
-		/* unlock the objects that has been locked for DDL replication */
-		release_locks_by_tranid (thread_entry, header.trid);
+		release_all_locks_for_ddl (thread_entry, header.trid);
 	      }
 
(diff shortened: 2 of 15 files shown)


src/transaction/log_replication_atomic.hpp
```diff
@@ -98,6 +101,11 @@ namespace cublog
 
       log_lsa m_lowest_unapplied_lsa;
       mutable std::mutex m_lowest_unapplied_lsa_mutex;
+
+      /* Store the locked objects for DDL replication.
+       * Since multiple DDL operation can be executed within single trnasaction,
```
joohok:
```suggestion
       * Since multiple DDL operations can be executed within single transaction,
```

src/transaction/log_replication_atomic.cpp
```diff
@@ -367,4 +370,26 @@ namespace cublog
 	  }
       }
   }
+
+  void
+  atomic_replicator::release_locks_by_tranid (cubthread::entry &thread_entry, const TRANID trid)
+  {
+    assert (m_locked_objects.count (trid) > 0);
+
+    for (auto it = m_locked_objects.lower_bound (trid); it != m_locked_objects.upper_bound (trid); it++)
```
wrlawodms: Nit. It would be more verbose but seems clearer to me.
```suggestion
    auto [begin, end] = m_locked_objects.equal_range (trid);
    for (auto it = begin; it != end; it++)
```

src/transaction/log_replication_atomic.cpp
```diff
@@ -367,4 +370,26 @@ namespace cublog
...
+      {
+	lock_unlock_object (&thread_entry, & (it->second), oid_Root_class_oid, SCH_M_LOCK, true);
```
wrlawodms: Q: In the case of the Rootclass or non-mvcc class, is the last argument, `force`, is `true? Please tell me why shortly.

src/transaction/log_replication_atomic.cpp
```diff
@@ -367,4 +370,26 @@ namespace cublog
...
+      }
+
+    m_locked_objects.erase (trid);
+  }
+
+  void
+  atomic_replicator::acquire_lock (cubthread::entry &thread_entry, const TRANID trid, const OID *classoid)
+  {
+    lock_object (&thread_entry, classoid, oid_Root_class_oid, SCH_M_LOCK, LK_UNCOND_LOCK);
```
wrlawodms: I doubt we can skip the error checking here. If it fails, the isolation will be broken. 
If it's supposed to succeed, I think we should use assert_release (), and if not, we have to handle it in a way.

wrlawodms: And, the "kick-out user transactions" would come in here. We need a TODO comment for that.

src/transaction/log_replication_atomic.hpp
```diff
@@ -98,6 +101,11 @@ namespace cublog
 
       log_lsa m_lowest_unapplied_lsa;
       mutable std::mutex m_lowest_unapplied_lsa_mutex;
+
+      /* Store the locked objects for DDL replication.
+       * Since multiple DDL operations can be executed within single transaction,
+       * more than one objects can be mapped to one transaction */
+      std::multimap <TRANID, OID> m_locked_objects;
```
wrlawodms: Let's make it clear it's classes.
```suggestion
      std::multimap <TRANID, OID> m_locked_classes; 
```

src/transaction/log_replication_atomic.cpp
```diff
@@ -367,4 +370,26 @@ namespace cublog
...
+      {
+	lock_unlock_object (&thread_entry, & (it->second), oid_Root_class_oid, SCH_M_LOCK, true);
+      }
+
+    m_locked_objects.erase (trid);
+  }
+
+  void
+  atomic_replicator::acquire_lock (cubthread::entry &thread_entry, const TRANID trid, const OID *classoid)
+  {
```
wrlawodms, line 388:
```suggestion
  {
   assert (!OID_ISNULL  (classoid) && !OID_ISTEMP(classoid));
```

src/transaction/log_replication_atomic.hpp
wrlawodms: How about adding an assert `m_locked_classes.empty();` in the dtor, `~atomic_replicator()`?

src/transaction/log_replication_atomic.cpp
joohok: it looks good, I meant to use this when studying C++17, but I forgot. Thanks

src/transaction/log_replication_atomic.cpp
joohok: 1. Should we unlock regardless of the isolation level? --> force = true,
2. Should we consider the isolation level when unlocking? --> force = false

However, when modifying records in the root class or non-MVCC classes, unlocking should be done regardless of the isolation level, so the `force` value is set to true.

------------
Hmmm.
It seems like we should always use 'force=true' for records in the root class or non-MVCC classes. However, internally in the 'lock_unlock_object()' function, when the 'lock_mode' is a shared lock, it appears to be implemented as if 'force' could be set to false. For READ_COMMITTED, it unlocks shared locks, but for other isolation levels, it doesn't perform unlocking. In the case of the root class, it should lock regardless of the isolation level. This is something that needs to be raised as an issue

src/transaction/log_replication_atomic.cpp
joohok: Thanks, I agree

src/transaction/log_replication_atomic.hpp
joohok: OK, thanks.

joohok: Agree! 
I should also consider cases where ATS performs DDL but dies before committing or aborting. Locks might remain in place. Perhaps ATS will create an abort log during recovery, but until then, it may continue to hold locks. Thanks.

src/transaction/log_replication_atomic.cpp
joohok: I strongly agree. While working on the replicator code, I've realized that error handling is challenging. Even if an error occurs during the process of applying logs, it continues to move forward. 
I've been contemplating whether to proceed as the existing code or not, but I did follow the existing code..; 

Perhaps cases where `lock_object()` fails are related to not being able to allocate memory internally. In such situations, would it be a good idea to implement a mechanism for retrying the lock?

src/transaction/log_replication_atomic.hpp
wrlawodms: Oh, you're right. Then, `assert (m_locked_classes.empty())` can't be true always. Beyond that, transactions blocked by DDL by the replicator on PTS wait forever. Have we considered this? The transactions definitely are bound to be aborted.

wrlawodms: but, it's also true that we can't release DDL locks since it's not reverted. In this case, are all transactions wanting to access a table that is locked by the replicator just stuck until a new ATS arises?

src/transaction/log_replication_atomic.cpp
```diff
@@ -42,6 +43,7 @@ namespace cublog
      * and, thus, does not need to be left in consistent state; thus, no check as to the consistent
      * termination state for atomic replication is needed
      */
+    assert (m_locked_classes.empty ());
```
wrlawodms: As we discussed, it can't be true, if a DDL transaction on ATS is killed before completing

src/transaction/log_replication_atomic.hpp
joohok: It seems like @hornetmj  had a similar question before. At that time, the answer was to wait for ATS to recover and then release the lock. Also, it was considered risky to release the lock while the DDL performed by ATS had not been completely applied to PTS. 
I can't think of any alternatives other than PTS transactions waiting..

src/transaction/log_replication_atomic.cpp
joohok: Yes, I will remove it

src/transaction/log_replication_atomic.cpp
joohok: The code assumes that functions calling 'lock_object()' suppose to obtain the lock, and in case it fails to acquire the lock, transactions are aborted. However, it's a bit perplexing since there's no separate error handling in other functionalities. 

Currently, 'lock_object()' fails only when there are issues like memory allocation problems internally, and repeatedly attempting 'lock_object()' in such cases may not yield success. It might be a good idea to use `assert_release()` for now and explore options for error handling in the `replicator`. Determining how the replicator should behave in the event of an error during execution would likely make error handling much smoother.

src/transaction/log_replication_atomic.hpp
wrlawodms: It seems not a simple issue, but I think we should address it. The infinite wait even can be cascaded to other transactions. If a Trx1 is blocked by a DDL transaction by ATS which will be never completed until an ATS is reconnected, other transactions blocked by the Trx1 are all blocked, and so on.

I assume we have an assumption that PTS can run even without ATS, but it seems to break the assumption easily. At least, we should have an idea about how to deal with this. For example, if a replicator is blocked too long (timeout) with a DDL, we can make disallow to access the table locked with sch_m_lock by the replicator. In other words, if a transaction tries to access the table, just return an error. It, at least, allows transactions which don't access the table to be run.

There probably are better idea. We should define this problem and come up with a way to handle this case. It would be best to prevent this problem.

joohok: Since PTS transactions are read-only, they are unlikely to impact each other. It seems like we only need to be concerned about transactions blocked by the replicator. Given that this is not a simple issue, I've come up with a few ideas.

- When a DDL start log arrives, it seems feasible to temporarily pause the replicator until replicator knows whether the DDL has been committed or aborted.
ex: 
  1)  DDL start (LOG_SCHEMA_MODIFICATION_LOCK)
  2) Pause replicator
  3) DDL transaction complete (LOG_COMMIT, LOG_ABORT)
  4) Resume replicator.

In this case, a LOCK mechanism would likely be necessary, but it merely delays the replicator's application.

- Alternatively, it might be possible to check both the connection status with ATS and lock mode, then report errors each time a read transaction in PTS access the object that is being locked by DDL.

I think we should create a separate issue to address this problem in more detail.

wrlawodms: > Since PTS transactions are read-only, they are unlikely to impact each other.

Oh, you're right. It's way less serious than I thought. We can just create an issue for this and move on for now.

src/transaction/log_replication_atomic.cpp
```diff
@@ -20,8 +20,9 @@
 #include "log_replication_atomic.hpp"
 #include "log_replication_jobs.hpp"
 
-#include "log_recovery_redo_parallel.hpp"
 #include "heap_file.h"
+#include "log_recovery_redo_parallel.hpp"
```
wrlawodms, line 24: why does it come down?

src/transaction/log_replication_atomic.cpp
wrlawodms: OK, we should always make it call the `lock_internal_perform_unlock_object` as the `lock_unlock_all()` does when commit.

Another question: 
I assume we unlock the locks as it does when commit. In this regard, I see some differences and I'd like to know it's allowed.

Here: 
`lock_internal_perform_unlock_object (thread_p, entry_ptr, false, true);`
In the `lock_unlock_all()`:
`lock_internal_perform_unlock_object (thread_p, entry_ptr, true, false);`

in the `lock_unlock_all()`,
- /* remove non2pl locks */ part
- lock_clear_deadlock_victim

It could be a silly question. I just don't know. If you've already considered this, just answer it shortly please.

src/transaction/log_replication_atomic.cpp
wrlawodms: How about other return values? It has 6 other return values. Are they not allowed at all?

wrlawodms: > Determining how the replicator should behave in the event of an error during execution would likely make error handling much smoother.

I think it's where @cristiarg should kick in a help. It's related to both the replicator and lock conflict.

src/transaction/log_replication_atomic.cpp
joohok: Wow, thank you so much... I had overlooked that. 
If we only perform `lock_unlock_object`, the entries managed by the lock manager could have been left uncleared. Additionally, the non2pl entries would have kept accumulating. 
It seems like I should make a separate function to perform the unlock.

src/transaction/log_replication_atomic.cpp
joohok: > How about other return values? It has 6 other return values. Are they not allowed at all?

Yes they are not allowed.
Except for LK_GRANTED, all the other return values are LK_NOT_GRANTED_* or currently unused.

src/transaction/log_replication_atomic.cpp
joohok: Due to sorting alphabetically like the other files, it has been modified

src/transaction/log_replication_atomic.cpp
joohok: Summary
1. Cleaning up the deadlock-related information seems a bit risky. The Replicator is registered with the lock manager as a single transaction, so it might affect other DDL statements. This issue appears to be something that needs to be addressed in LETS-787 (which deals with deadlock)
2. it may not be necessary to clean up the Non2pl list.
3. I will go ahead and implement the call to `lock_internal_perform_unlock_object(thread_p, entry_ptr, true, false)` as you suggested.

Detail
1. I wasn't sure if we could clear the deadlock in this PR.  
The Replicator should not become a victim when a deadlock occurs.  If we clear the deadlock information for the replicator, there might be issues with DDL transactions still in progress. 
 For example, 
  if in ATS, trx1 and trx2 are executing DDL simultaneously,  PTS can potentially trigger a deadlock while applying the changes made by trx1 and trx2.  
  Clearing deadlock information when trx1 commits may still affect trx2, which is still in progress,  because in PTS, the replicator is registered with the lock manager as a single transaction. 
  This part might need to be addressed in the lock manager extension (LETS-787).

2. The Non2pl list accumulates when a transaction releases a lock when it is not the end of the transaction. 
 As previously discussed, transactions sometimes temporarily use the lock manager to check the existence of a class and immediately release a lock, such as in xlocator_find_class_oid(). 
  These locks don't follow the 2PL rule, so they are stored in the Non2pl list and need to be cleaned up at commit time.   
  
  However, Non2pl entries are only created through `lock_unlock_object()`, and the replicator does not explicitly call `lock_unlock_object()`.  Therefore, there's no need to clean it up. When we use locator functions like xlocator_find_class_oid() in the future, we can clean up the Non2pl list as needed.
 
3. I will create a separate function to call `lock_internal_perform_unlock_object(thread_p, entry_ptr, true, false)` 
    - Since it's the commit or abort point, the third argument (release_flag) should be set to true, to clean up the lock entry. 
    - The replicator doesn't need to stack entries in this Non2pl list, so the last argument (move_to_non2pl) should be set to false.

src/transaction/log_replication_atomic.cpp
```diff
@@ -119,6 +120,13 @@ namespace cublog
 	      {
 		m_replicator_mvccid->complete_mvcc (header.trid, replicator_mvcc::COMMITTED);
 	      }
+
+	    if (m_locked_classes.count (header.trid))
```
cristiarg:
```suggestion
	    if (m_locked_classes.count (header.trid) > 0)
```

src/transaction/log_replication_atomic.cpp
```diff
@@ -148,6 +156,13 @@ namespace cublog
 	      {
 		m_replicator_mvccid->complete_mvcc (header.trid, replicator_mvcc::ABORTED);
 	      }
+
+	    if (m_locked_classes.count (header.trid))
```
cristiarg:
```suggestion
	    if (m_locked_classes.count (header.trid) > 0)
```

src/transaction/log_replication_atomic.cpp
```diff
@@ -367,4 +370,38 @@ namespace cublog
...
+  {
+    assert (!OID_ISNULL (classoid) && !OID_ISTEMP (classoid));
+
+    /* TODO:
+     * If a PTS read transaction holds a lock for an extended period without releasing it for the same class,
+     * the replicator could wait too long to acquire the lock.
+     * In such cases, there might be an introduction of a mechanism to abort read transaction that holds lock.
+     */
+
+    if (lock_object (&thread_entry, classoid, oid_Root_class_oid, SCH_M_LOCK, LK_UNCOND_LOCK) != LK_GRANTED)
```
cristiarg, line 397: Isn't it possible for the `LOG_SCHEMA_MODIFICATION_LOCK` to be logged multiple times by the same ATS transaction?
I think we discussed this somewhere.

I think we should introduce a mechanism to at least war against this.
Without such a mechanism:
- the `lock_object` would likely succeed because repeated lock requests are issued by the same transaction (the replication transaction) and lock manager is built to support this
- and, correspondingly, the `lock_unlock_object` call afterwards will attempt to unlock multiple times the same object - I also think this will work
- but the multimap below will contain more than one entry with the same `<trid, classoid>`.

Do we want to check that multiple logging of `LOG_SCHEMA_MODIFICATION_LOCK` on the ATS side by the same transaction for the same oid is prevented?
If so, maybe the best container would be a `std::map<TRANID, std::set<OID>>`.
What is your opinion?

src/transaction/log_replication_atomic.hpp
```diff
@@ -90,6 +90,9 @@ namespace cublog
       void set_lowest_unapplied_lsa ();
       void replicate_sysop_start_postpone (cubthread::entry &thread_entry, const LOG_RECORD_HEADER &rec_header);
 
+      void release_locks_by_tranid (cubthread::entry &thread_entry, const TRANID trid);
+      void acquire_lock (cubthread::entry &thread_entry, const TRANID trid, const OID *classoid);
```
hornetmj: Nit:
How about clearly conveying the purpose through the function name rather than comments?

acquire_lock_for_ddl()
release_all_lock_for_ddl()

hornetmj (approved):

src/transaction/log_replication_atomic.cpp
joohok: I think I understand your concern. If we only look at the server-side lock mechanism, your point is very valid. 

However, CAS caches lock information in the workspace, so if an SCH_M_LOCK is already acquired for a class, it won't request lock_object from the server (locator_can_skip_fetch_from_server()). Therefore, it seems there won't be multiple logs for a single classOID, and that's why I used multimap. Thanks to your input, I had a chance to double-check.

src/transaction/log_replication_atomic.hpp
joohok: OK, I will. Thanks

src/transaction/log_replication_atomic.cpp
hornetmj: Nit:

In debug mode or limited to system parameters, why not add some logic to check the validity of the class_oid passed to LOG_SCHEMA_MODIFICATION_LOCK or the presence of duplicate std::pairs in std::multimap <TRANID, OID> m_locked_classes?

It depends on you.

joohok: > Nit:
> 
> In debug mode or limited to system parameters, why not add some logic to check the validity of the class_oid passed to LOG_SCHEMA_MODIFICATION_LOCK or the presence of duplicate std::pairs in std::multimap <TRANID, OID> m_locked_classes?
> 
> It depends on you.

I will verify whether duplicate pairs exist. However, I intend to address the validation of classOID when dealing with DDL-related caches. This is because the validation check for OIDs will be included in the process of storing and discarding the cache

wrlawodms: >  it seems there won't be multiple logs for a single classOID,

Then, how about going with an assert for this?

src/transaction/log_replication_atomic.cpp
wrlawodms: Then, doesn't we need to move up the `heap_file.h`? There are other headers starting with `L`.

joohok: It appears that the `log_replication_*` header files are kept separately, while the rest of the header files are declared below. Therefore, I have sorted the lower section alphabetically without altering the paragraph above.

In `log_checkpoint_info.cpp` and other places, it seems that directly related header files are declared alphabetically first, and the rest of the header files are declared alphabetically in the following section.

src/transaction/log_replication_atomic.cpp
joohok: I have set it to check only in debug mode. If duplicate values are found, I have added assert(false)

src/transaction/lock_manager.c
```diff
@@ -6614,6 +6614,41 @@ lock_unlock_object_donot_move_to_non2pl (THREAD_ENTRY * thread_p, const OID * oi
...
+lock_unlock_object_and_cleanup (THREAD_ENTRY * thread_p, const OID * oid, const OID * class_oid, LOCK lock)
+{
+#if !defined (SERVER_MODE)
+  return;
+#else /* !SERVER_MODE */
+  int tran_index;
+  LK_TRAN_LOCK *tran_lock;
+
+  tran_index = LOG_FIND_THREAD_TRAN_INDEX (thread_p);
+  tran_lock = &lk_Gl.tran_lock_table[tran_index];
```
wrlawodms: Nit.
```suggestion
  const int tran_index = LOG_FIND_THREAD_TRAN_INDEX (thread_p);
  LK_TRAN_LOCK *tran_lock = &lk_Gl.tran_lock_table[tran_index];
```

src/transaction/lock_manager.c
```diff
@@ -6614,6 +6614,41 @@ lock_unlock_object_donot_move_to_non2pl (THREAD_ENTRY * thread_p, const OID * oi
...
+
+  assert (tran_lock->non2pl_list == NULL);
+
+  lock_unlock_object_lock_internal (thread_p, oid, class_oid, lock, true, false);
+
+  /* TODO:
+   * We have to determine if deadlock information should be cleared
```
wrlawodms:
```suggestion
   * We have to decide if deadlock information should be cleared
```

src/transaction/log_replication_atomic.cpp
```diff
@@ -367,4 +368,49 @@ namespace cublog
...
+    /* there will be no duplicated object in the map */
+    auto [begin, end] = m_locked_classes.equal_range (trid);
+    for (auto it = begin; it != end; it++)
+      {
+	if (std::count (begin, end, *it) > 1)
+	  {
+	    assert (false);
+	  }
+      }
+#endif
```
wrlawodms: I'd check the duplication only for <trid, *classoid>. It seems to check all classoid for a trid.
```suggestion
#if !defined (NDEBUG)
    /* there will be no duplicated object in the map */
    auto [begin, end] = m_locked_classes.equal_range (trid);
    for (auto it = begin; it != end; it++) // or count_if()?
      {
          assert (it.second != *classoid);
      }
#endif
    m_locked_classes.emplace (trid, *classoid);
```

src/transaction/log_replication_atomic.hpp
```diff
@@ -98,6 +101,11 @@ namespace cublog
 
       log_lsa m_lowest_unapplied_lsa;
       mutable std::mutex m_lowest_unapplied_lsa_mutex;
+
+      /* Store the locked objects for DDL replication.
```
wrlawodms: Nit.
```suggestion
      /* Keep track of the locked class objects for DDL replication.
```

wrlawodms (approved):

src/transaction/log_replication_atomic.cpp
joohok: Oh..?! this is much simpler, thanks

src/transaction/log_replication_atomic.hpp
joohok: Done!

src/transaction/lock_manager.c
joohok: Done!

src/transaction/lock_manager.c
joohok: done! thanks

src/transaction/lock_manager.c
```diff
@@ -6614,6 +6614,38 @@ lock_unlock_object_donot_move_to_non2pl (THREAD_ENTRY * thread_p, const OID * oi
...
+ *
+ */
+void
+lock_unlock_object_and_cleanup (THREAD_ENTRY * thread_p, const OID * oid, const OID * class_oid, LOCK lock)
+{
+#if !defined (SERVER_MODE)
+  return;
+#else /* !SERVER_MODE */
+  const int tran_index = LOG_FIND_THREAD_TRAN_INDEX (thread_p);
+  LK_TRAN_LOCK *tran_lock = &lk_Gl.tran_lock_table[tran_index];
```
cristiarg: Nit
```suggestion
  LK_TRAN_LOCK *const tran_lock = &lk_Gl.tran_lock_table[tran_index];
```

cristiarg (approved):

src/transaction/lock_manager.c
joohok: Thanks.

PR #4753, opened 2023-10-31
[LETS-749] Pause the prior log receiver when a new ATS connects

wrlawodms:
http://jira.cubrid.org/browse/LETS-749

**The previous approach**: Start the prior receiver thread when the catch-up is done.  
**The previous issue**: It leads to duplicated thread initializations over ats re-connecting.

**The current approach**: 
- When ATS connects, it digests all existing prior nodes and pauses the internal thread.
  - We have some choices here. We can abandon the remainder and we can pause the thread when the previous ATS disconnects. 
  - Abandon the remainder? : It would be safer and reduce how much to catch up with to digest existing ones.
  - Pause when the previous disconnects? : I'd like to do it as lazily as possible.
- When a PS has done catching up, it resumes the thread, which starts appending log prior nodes that have been kept.

Add a unit test for pause/resume
- Refactor existing tests to make use of internal functions.
  - Take out `wait_until_count_and_reset` and `require_prior_list_match`.
  - Adapt `require_prior_list_match` to accept paused prior receiver scenarios. If any node has not arrived at the dest_prior_node after flushing, it means a receiver must've been paused. 

I'll make modifications to the Jira issue after this review.

### DIFF
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -779,6 +779,9 @@ page_server::set_active_tran_server_connection (cubcomm::channel &&chn)
       disconnect_active_tran_server ();
     }
 
+  // Pause the log prior receiver to process log records from the enw ATS. It just keeps them until the catch-up is compeleted.
+  log_Gl.get_log_prior_receiver ().wait_until_empty_and_pause ();
+
   m_active_tran_server_conn.reset (new tran_server_connection_handler (std::move (chn), transaction_server_type::ACTIVE,
 				   *this));
 }
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -1215,6 +1215,7 @@ page_server::execute_catchup (cubthread::entry &entry, const LOG_LSA catchup_lsa
 void
 page_server::complete_catchup ()
 {
+  log_Gl.get_log_prior_receiver ().resume ();
   push_request_to_active_tran_server (page_to_tran_request::SEND_CATCHUP_COMPLETE, std::string());
 }
 
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -230,7 +230,7 @@ page_server::tran_server_connection_handler::receive_start_catch_up (tran_server
     {
       // TODO: It means that the ATS is booting up.
       // it will be set properly after ATS recovery is implemented.
-      m_ps.push_request_to_active_tran_server (page_to_tran_request::SEND_CATCHUP_COMPLETE, std::string());
+      m_ps.complete_catchup ();
       return;
     }
 
@@ -242,7 +242,7 @@ page_server::tran_server_connection_handler::receive_start_catch_up (tran_server
     {
       _er_log_debug (ARG_FILE_LINE, "[CATCH_UP] There is nothing to catch up.\n");
 
-      m_ps.push_request_to_active_tran_server (page_to_tran_request::SEND_CATCHUP_COMPLETE, std::string());
+      m_ps.complete_catchup ();
       return;
     }
 
@@ -1189,9 +1189,8 @@ page_server::execute_catchup (cubthread::entry &entry, const LOG_LSA catchup_lsa
 
       _er_log_debug (ARG_FILE_LINE, "[CATCH_UP] The catch-up is completed, ranging from %lld to %lld, count = %lld.\n",
 		     start_pageid, end_pageid, total_page_count);
-      // TODO: start appneding log prior nodes from the ATS.
 
-      push_request_to_active_tran_server (page_to_tran_request::SEND_CATCHUP_COMPLETE, std::string());
+      complete_catchup ();
 
       with_disc_msg = true;
     }
@@ -1211,6 +1210,12 @@ page_server::execute_catchup (cubthread::entry &entry, const LOG_LSA catchup_lsa
     }
 
   disconnect_followee_page_server (with_disc_msg);
+}
+
+void
+page_server::complete_catchup ()
+{
+  push_request_to_active_tran_server (page_to_tran_request::SEND_CATCHUP_COMPLETE, std::string());
 }
 
 bool
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -779,7 +779,7 @@ page_server::set_active_tran_server_connection (cubcomm::channel &&chn)
       disconnect_active_tran_server ();
     }
 
-  // Pause the log prior receiver to process log records from the enw ATS. It just keeps them until the catch-up is compeleted.
+  // Pause the log prior receiver to process log records from the new ATS. It just keeps them until the catch-up is completed.
   log_Gl.get_log_prior_receiver ().wait_until_empty_and_pause ();
 
   m_active_tran_server_conn.reset (new tran_server_connection_handler (std::move (chn), transaction_server_type::ACTIVE,
--- a/src/transaction/log_prior_recv.cpp
+++ b/src/transaction/log_prior_recv.cpp
@@ -52,6 +52,7 @@ namespace cublog
   {
     assert (m_thread.get_id () == std::thread::id ()); // make sure that it's not been started.
     m_shutdown = false;
+    m_paused = false;
     m_thread = std::thread (&prior_recver::loop_message_to_prior_info, std::ref (*this));
   }
 
@@ -71,11 +72,32 @@ namespace cublog
   {
     std::unique_lock<std::mutex> ulock (m_messages_mutex);
 
+    assert (m_paused == false);
+    assert (m_shutdown == false);
+
     constexpr auto sleep_30ms = std::chrono::milliseconds (30);
     while (!m_messages_consume_cv.wait_for (ulock, sleep_30ms, [this]
     {
       return !m_messages.empty();
       }));
+
+    m_paused = true;
+  }
+
+  void
+  prior_recver::resume ()
+  {
+    std::unique_lock<std::mutex> ulock (m_messages_mutex);
+
+    assert (m_paused == true);
+    assert (m_shutdown == false);
+
+    m_paused = false;
+
+    if (!m_messages.empty ())
+      {
+	m_messages_push_cv.notify_one ();
+      }
   }
 
   void
@@ -88,7 +110,7 @@ namespace cublog
       {
 	m_messages_push_cv.wait (ulock, [this]
 	{
-	  return m_shutdown || !m_messages.empty ();
+	  return m_shutdown || (!m_messages.empty () && !m_paused);
 	});
 
 	if (m_shutdown)
--- a/src/transaction/log_prior_recv.cpp
+++ b/src/transaction/log_prior_recv.cpp
@@ -157,9 +157,9 @@ namespace cublog
 	    backbuffer.pop ();
 	  }
 
-	ulock.lock ();
-
 	m_messages_consume_cv.notify_one ();
+
+	ulock.lock ();
       }
   }
 }
--- a/src/transaction/log_prior_recv.cpp
+++ b/src/transaction/log_prior_recv.cpp
@@ -44,7 +44,7 @@ namespace cublog
     std::unique_lock<std::mutex> ulock (m_messages_mutex);
     m_messages.push (std::move (str));
     ulock.unlock ();
-    m_messages_condvar.notify_one ();
+    m_messages_push_cv.notify_one ();
   }
 
   void
@@ -61,7 +61,7 @@ namespace cublog
     std::unique_lock<std::mutex> ulock (m_messages_mutex);
     m_shutdown = true;
     ulock.unlock ();
-    m_messages_condvar.notify_one ();
+    m_messages_push_cv.notify_one ();
 
     m_thread.join ();
   }
@@ -74,7 +74,7 @@ namespace cublog
     std::unique_lock<std::mutex> ulock (m_messages_mutex);
     while (true)
       {
-	m_messages_condvar.wait (ulock, [this]
+	m_messages_push_cv.wait (ulock, [this]
 	{
 	  return m_shutdown || !m_messages.empty ();
 	});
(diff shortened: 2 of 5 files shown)


unit_tests/log/test_main_prior_sendrecv.cpp
```diff
@@ -237,21 +308,32 @@ test_env::require_prior_list_match () const
...
 	  source_node = source_node->next;
 	  dest_node = dest_node->next;
 	}
-      REQUIRE (dest_node == nullptr);
-      REQUIRE (m_source_prior_info.prev_lsa == dest_prior_info->prev_lsa);
-      REQUIRE (m_source_nodes_tail->start_lsa == dest_prior_info->prior_list_tail->start_lsa);
+
+      if (source_node != nullptr)
+	{
+	  // some from source_node haven't been arrrived.
```
wrlawodms:
```suggestion
	  // some from source_node haven't been arrived.
```

src/server/page_server.cpp
```diff
@@ -779,6 +779,9 @@ page_server::set_active_tran_server_connection (cubcomm::channel &&chn)
       disconnect_active_tran_server ();
     }
 
+  // Pause the log prior receiver to process log records from the enw ATS. It just keeps them until the catch-up is compeleted.
```
wrlawodms:
```suggestion
  // Pause the log prior receiver to process log records from the new ATS. It just keeps them until the catch-up is completed.
```

src/transaction/log_prior_recv.cpp
```diff
@@ -58,11 +59,44 @@ namespace cublog
...
     m_shutdown = true;
     ulock.unlock ();
-    m_messages_condvar.notify_one ();
+    m_messages_push_cv.notify_one ();
 
     m_thread.join ();
   }
 
+  void
+  prior_recver::wait_until_empty_and_pause ()
```
wrlawodms, line 69: I assume that once this function is called, no log records are added from the previous ATS. It's called between the new ATS connection and the previous one. If it's not met, we'll see it happens when appending log records because log records from two are not consecutive.

hornetmj:
> Pause when the previous disconnects? : I'd like to do it as lazily as possible.

What is the advantage of this approach compared to pausing immediately after disconnect?

wrlawodms:
@hornetmj 
> What is the advantage of this approach compared to pausing immediately after disconnect?

It has some advantages in terms of performance. Basically, it doesn't need to be done when disconnecting.
1. Let's say there are some log prior nodes left to consume when disconnecting.  The later we make sure all log nodes are consumes,  the lesser we have to wait to be consumed. 
2. If we make sure all nodes are consumed before disconnecting, the disconnection takes more time.

joohok:
Is the catch-up in progress, and has the ATS disconnected, requiring a reconnection?
Could a scenario like the following occur?

First, let's illustrate the logs in catching-up PS :
Previous Conn:   | Catch-up     | Prior nodes upto LSA 20|
Next Conn  (1):   | Catch-up     | Empty space     | Prior nodes received from LSA 19|
Next Conn  (2):   | Catch-up     | Empty space                           | Prior nodes received from LSA 22|

Let's assume that ATS re-starts and re-connects with PS. 
Even though we received logs up to `LSA 20` from the previous connection, in the next connection, 
it's possible that ATS may send logs earlier than that or later than that. 
If, in the next connection, ATS send logs starting from `LSA 19`, handling duplicate logs by simply ignoring them should work fine. However, if in the next connection, ATS send logs starting from `LSA 22`, it could potentially cause issues. The reason I believe the latter scenario could occur is that a PS in the midst of catching up doesn't participate in the quorum.
(I'm not sure that case 1 could occur)
For the latter case, it can be solved by executing catch-up process again, I think;

wrlawodms:
@joohok 

> Next Conn (1): | Catch-up | Empty space | Prior nodes received from LSA 19|

It can happen. This is going to be covered later when we introduce the Durable LSA. If a new ATS sends past LSA, it means that the Durable LSA is earlier or equal to the LSA, and the PS must've thrown away all the log records between the Durable LSA and the LSA.

>Next Conn (2): | Catch-up | Empty space | Prior nodes received from LSA 22| .. it could potentially cause issues..

hmm. I don't get your point. The way I see it, it's what's the catch-up mechanism for. In other words, log records between 20 and 22 are fetched from a peer PS. I guess I'm missing something about what you're talking about. Could you elaborate on this?

joohok:
> > Next Conn (2): | Catch-up | Empty space | Prior nodes received from LSA 22| .. it could potentially cause issues..
> 
> hmm. I don't get your point. The way I see it, it's what's the catch-up mechanism for. In other words, log records between 20 and 22 are fetched from a peer PS. I guess I'm missing something about what you're talking about. Could you elaborate on this?

I expected that catch-up will proceed only up to the `catchup_lsa` received from the previous connection.

1. PS will receive logs up to catchup_lsa (LSA 10) from a peer PS
2. then, In the previous connection, it received prior nodes from ATS up to LSA 20.
3. In the new connection, it receives prior nodes starting from LSA 22.

In this scenario, I thought that LSA 21 would be missing. 

Could the catch-up process proceed beyond catchup_lsa?

wrlawodms:
@joohok 

> Could the catch-up process proceed beyond catchup_lsa?

When a new ATS is connected, a new catch-up is triggered with a new catch-up LSA, which is the LSA of the first log record the new ATS will send.

src/transaction/log_prior_recv.cpp
```diff
@@ -58,11 +59,50 @@ namespace cublog
...
+      return m_messages.empty();
+      }));
+
+    m_paused = true;
+  }
+
+  void
+  prior_recver::resume ()
+  {
+    std::unique_lock<std::mutex> ulock (m_messages_mutex);
```
joohok: does `m_message_mutex` also protect the `m_paused` variable?  
Maybe we can remove `ulock ` or unlock it before notify_one ().

joohok (approved):

src/transaction/log_prior_recv.hpp
```diff
@@ -66,10 +69,12 @@ namespace cublog
 
       message_container m_messages;                           // internal queue of messages
       std::mutex m_messages_mutex;                            // protect access on message queue
-      std::condition_variable m_messages_condvar;             // notify internal thread of new messages
+      std::condition_variable m_messages_push_cv;             // notify internal thread of new messages
+      std::condition_variable m_messages_consume_cv;          // notify a set of message has been consumed
 
       std::thread m_thread;                                   // internal thread
       bool m_shutdown = false;                                // true to stop internal thread
+      bool m_paused = false;                                  // true to pause consuming messages
```
cristiarg: Now we have running / paused (idle) / shutdown states - which all seem to be mutually exclusive.
According to the "rule of three" (DRY principle), it makes sense to refactor into an enum.

With an enum you can also introduce an intermediary state `pausing` that will help in `wait_until_empty_and_pause` in combination with a check in `prior_recver::push_message` to prevent "pausing starvation" (although, as you say in comment https://github.com/CUBRID/cubrid/pull/4753/files#r1377288157 , this situation is not occuring).

src/transaction/log_prior_recv.cpp
```diff
@@ -58,11 +59,50 @@ namespace cublog
...
+
+    assert (m_paused == true);
+    assert (m_shutdown == false);
+
+    m_paused = false;
+
+    if (!m_messages.empty ())
```
cristiarg: Nit.
`notify_one` from within the lock?
I would enclose the messages mutex in a scope; in the locked scope save a bool flag whether to notify or not; notify if needed outside the scope:
```
  bool notify_one = false;
  { // locked scope
    std::unique_lock<std::mutex> ulock (m_messages_mutex);
    // ...
    notify_one = !m_messages.empty ();
    // ...
  }
  if (notify_one)
    {
      // notify_one
    }
```

Same remark/question for `m_messages_consume_cv.notify_one ();` below.

cristiarg:
Executed tests on this branch. No regressions other than the known ones.
```
==================================== PRINT SUMMARY ====================================
Test category: scalability
Total cases: 94
Passed cases: 85
Failed cases: 1
Timeout cases: 0
Cores found: 0
Skipped cases: 8
Total execution time: 20527 s
```

hornetmj (approved):

src/transaction/log_prior_recv.cpp
wrlawodms: yep. And `m_messages.empty ()` also has to be in the scope. But, we can put the `notify_one()` as @cristiarg  suggested.

src/transaction/log_prior_recv.cpp
wrlawodms: Thank you for the pointer. Good idea.

src/transaction/log_prior_recv.hpp
wrlawodms: I've introduced the state concept. With this abstraction, I have even caught  a race condition that I missed between shutdown and wait_and_pause. Thank you.

wrlawodms: Re-requesting review to all since it's made changes.

cristiarg: Just curious what was the race condition?

src/transaction/log_prior_recv.cpp
```diff
@@ -28,6 +28,7 @@ namespace cublog
   prior_recver::prior_recver (log_prior_lsa_info &prior_lsa_info)
     : m_prior_lsa_info (prior_lsa_info)
   {
+    m_state = state::RUN;
```
cristiarg, line 31: this should be in `start_thread`
or, better, remove `start_thread` altogether because it is private anyway

src/transaction/log_prior_recv.cpp
```diff
@@ -39,44 +40,101 @@ namespace cublog
...
+    if (notify_one)
+      {
+	m_messages_push_cv.notify_one ();
+      }
+
+    if (prm_get_bool_value (PRM_ID_ER_LOG_PRIOR_TRANSFER))
+      {
+	_er_log_debug (ARG_FILE_LINE,
+		       "[LOG_PRIOR_TRANSFER] Resume prior_recver: the number of kept messages while paused = %lu.\n", m_messages.size ());
+      }
```
cristiarg: the logging needs to be before `notify_one` and within the same scope - guarded by the if
otherwise race condition can happen - see `m_messages.swap (backbuffer);` below - and the logging will actually be `0`

even better, change the `notify_one` bool to `size_t message_count` and you have both:
- the correct value for the `if (message_count > 0)` condition
- and accessing the `m_messages.size ()` under the lock as is proper

the whole function:
```
  size_t message_count = 0u;
  {
    // lock
    // ...
    message_count = m_messages.size();
  }
  if (message_count >0)
    {
      // log...
      // notify ...
    }
```

Personally I would remove the `if (message_count >0)` condition and notify and log anyway; because the function is rarely used.

src/transaction/log_prior_recv.cpp
```diff
@@ -39,44 +40,101 @@ namespace cublog
...
+    if (m_state == state::SHUTDOWN)
+      {
+	return;
+      }
+
+    assert (m_state == state::RUN);
+    m_state = state::PAUSING;
+
+    constexpr auto sleep_30ms = std::chrono::milliseconds (30);
+    while (!m_messages_consume_cv.wait_for (ulock, sleep_30ms, [this]
```
cristiarg, line 82: There might still be a race condition here:
- if the condition is checked just after the `loop_message_to_prior_info` thread executes
  ```
  m_messages.swap (backbuffer);
  ulock.unlock ();
  ```
  lines below at ~142-~143, it will pass;
- the state would immediately transition to `PAUSED` which the consumption thread executing `loop_message_to_prior_info` would still be deserializing and pushing log prior messages - a possibly lengthty process

Not sure if this would be a problem or not - and if it is, I don't know how to solve it. 
If this race condition can occur as I described above, you could at least describe it in a comment and we'll see afterwards.

src/transaction/log_prior_recv.cpp
```diff
@@ -104,6 +162,8 @@ namespace cublog
 	  }
 
 	ulock.lock ();
+
+	m_messages_consume_cv.notify_one ();
```
cristiarg, line 162: This notification is still inside the protected critical section; swap with the `ulock.lock()` above?

cristiarg (approved):

src/transaction/log_prior_recv.cpp
wrlawodms: I'd put this in the function, but I thought it should be in the ctor because the state::RUN in the initial one. I'll remove the `start_thread` then.

src/transaction/log_prior_recv.cpp
wrlawodms: You're right. I follow both suggestions.

src/transaction/log_prior_recv.cpp
wrlawodms: I missed. done.

src/transaction/log_prior_recv.cpp
wrlawodms: It really happens. What I wanted to prevent by the "wait" part of `wait_until_empty_and_pause` is that no more log records from a previous ATS is appended to the log buffer. It would happen even if we waits until all message in the `back buffer` is consumed because the `log_Flush_daemon` is responsible for this. 

I'll handle this in another separate issue. Thank you very much.

src/transaction/log_prior_recv.hpp
wrlawodms: In `wait_until_empty_and_pause()`, it did not consider the shutdown of the `prior_recver`. It could wait until the queue is empty, which is not met if it's shutdown.

src/transaction/log_prior_recv.cpp
wrlawodms: http://jira.cubrid.org/browse/LETS-850

joohok (approved):

PR #4756, opened 2023-11-02
[CBRD-25048] Core dumped in pgbuf_replace_watcher_debug at src/storage/page_buffer.c:12574

xmilex-git:
http://jira.cubrid.org/browse/CBRD-25048

This issue arises due to the modifications induced by the CBRD-25037 issue. The reason for the error is that the processing within the `heap_next_internal ()` differs between when the `select_record_info` hint is used and when it is not used. Therefore, when the `select_record_info` hint is used, modifications have been made to transfer the value to the local page_watcher variable as before the change, and then execute `heap_get_record_info ()`.

### DIFF
--- a/src/storage/heap_file.c
+++ b/src/storage/heap_file.c
@@ -8043,7 +8043,6 @@ heap_next_internal (THREAD_ENTRY * thread_p, const HFID * hfid, OID * class_oid,
       /* A record was found */
       if (get_rec_info)
 	{
-
 	  PGBUF_INIT_WATCHER (&rec_info_page_watcher, PGBUF_ORDERED_HEAP_NORMAL, hfid);
 	  pgbuf_replace_watcher (thread_p, &scan_cache->page_watcher, &rec_info_page_watcher);
 	  scan =
--- a/src/storage/heap_file.c
+++ b/src/storage/heap_file.c
@@ -7831,6 +7831,7 @@ heap_next_internal (THREAD_ENTRY * thread_p, const HFID * hfid, OID * class_oid,
   int get_rec_info = cache_recordinfo != NULL;
   bool is_null_recdata;
   PGBUF_WATCHER old_page_watcher;
+  PGBUF_WATCHER rec_info_page_watcher;
 
   assert (scan_cache != NULL);
 
@@ -8042,8 +8043,11 @@ heap_next_internal (THREAD_ENTRY * thread_p, const HFID * hfid, OID * class_oid,
       /* A record was found */
       if (get_rec_info)
 	{
+
+                 PGBUF_INIT_WATCHER (&rec_info_page_watcher, PGBUF_ORDERED_HEAP_NORMAL, hfid);
+                 pgbuf_replace_watcher (thread_p, &scan_cache->page_watcher, &rec_info_page_watcher);
 	  scan =
-	    heap_get_record_info (thread_p, oid, recdes, forward_recdes, &scan_cache->page_watcher, scan_cache,
+	    heap_get_record_info (thread_p, oid, recdes, forward_recdes, &rec_info_page_watcher, scan_cache,
 				  ispeeking, cache_recordinfo);
 	}
       else
--- a/src/storage/heap_file.c
+++ b/src/storage/heap_file.c
@@ -8044,8 +8044,8 @@ heap_next_internal (THREAD_ENTRY * thread_p, const HFID * hfid, OID * class_oid,
       if (get_rec_info)
 	{
 
-                 PGBUF_INIT_WATCHER (&rec_info_page_watcher, PGBUF_ORDERED_HEAP_NORMAL, hfid);
-                 pgbuf_replace_watcher (thread_p, &scan_cache->page_watcher, &rec_info_page_watcher);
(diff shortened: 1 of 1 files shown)


src/storage/heap_file.c
```diff
@@ -8042,8 +8043,11 @@ heap_next_internal (THREAD_ENTRY * thread_p, const HFID * hfid, OID * class_oid,
       /* A record was found */
       if (get_rec_info)
 	{
+
+	  PGBUF_INIT_WATCHER (&rec_info_page_watcher, PGBUF_ORDERED_HEAP_NORMAL, hfid);
+	  pgbuf_replace_watcher (thread_p, &scan_cache->page_watcher, &rec_info_page_watcher);
 	  scan =
-	    heap_get_record_info (thread_p, oid, recdes, forward_recdes, &scan_cache->page_watcher, scan_cache,
+	    heap_get_record_info (thread_p, oid, recdes, forward_recdes, &rec_info_page_watcher, scan_cache,
```
shparkcubrid, line 8049: Is it okay if there is no routine to assign 'scan_cache->page_watcher' from 'rec_info_page_watcher'?

xmilex-git: The `heap_get_record_info()` function includes a process where a page that is fixed to the page_watcher is unfixed, or it is replaced with the scan_cache's page_watcher.

wrlawodms (commented):
Why don't you adapt the `heap_get_record_info()` function as you do in `heap_next_internal()`? 
- If you think the previous job is worth it, the same probably goes for this.
- We should keep the same semantic through the process. You changed it from "Does locally and keep it in cache if needed" to "Does with the cache". It should be the same through the process.

Otherwise, please share any issue why you can't apply it into the `heap_get_record_info()`.

xmilex-git:
> Why don't you adapt the `heap_get_record_info()` function as you do in `heap_next_internal()`?
> 
> * If you think the previous job is worth it, the same probably goes for this.
> * We should keep the same semantic through the process. You changed it from "Does locally and keep it in cache if needed" to "Does with the cache". It should be the same through the process.
> 
> Otherwise, please share any issue why you can't apply it into the `heap_get_record_info()`.

The reasons for not applying the same approach to the `heap_get_record_info()` function are as follows:

1. The `heap_get_record_info` function is used when utilizing the `select_record_info` hint, and it was determined that there is no need to enhance performance when this hint is used.

2. Inside the `heap_get_record_info()` function, different operations are performed based on the `record_type`, and during these operations, there is a process where pages get `unfixed` or `replaced`, which was resolved by using local variables.

If we are to proceed with additional work based on the logical reason of using `local variables` versus `scan_cache`, I believe it would be proper to refactor and reimplement after redefining the concept. This implementation was chosen as a quick response to the error in issue.

wrlawodms:
> The heap_get_record_info function is used when utilizing the select_record_info hint, and it was determined that there is no need to enhance performance when this hint is used.

I agree that it's way minor one. However, I'm also not sure the optimization worth it by removing `pgbuf_replace_watcher` in the `heap_next_internal` which doesn't seem to be a bottleneck. If it causes this kind of different semantic for each procedure so that it harms readability and maintainability, I think we should reconsider this optimization. 

Of course, if the optimization have an impact in general or most use cases, it's worth it to harm some maintainability. <- Is this the case?

xmilex-git:
> Of course, if the optimization have an impact in general or most use cases, it's worth it to harm some maintainability. <- Is this the case?

Yes, that's correct. According to the results from my callgrind profiling, the `pgbuf_replace_watcher` accounted for more than 28% of the total execution time of the `heap_next()` function during heap scan queries. Therefore, I suggested a solution to use `pgbuf_replace_watcher` less, which resulted in a performance improvement of about 4-5%.

shparkcubrid (approved):

wrlawodms:
@xmilex-git Great. I assumed that a bottleneck in this function would be the fix(latch) contention (it'd be common on OLTP db) or IO occurred while fetching page. That operations are way more expensive than just cpu operations like `PGBUF_INIT_WATCHER`. How do you get that `4-5%` improvement? It's improved in general? or Isolated transaction without any I/O?

src/storage/heap_file.c
```diff
@@ -8042,8 +8043,11 @@ heap_next_internal (THREAD_ENTRY * thread_p, const HFID * hfid, OID * class_oid,
       /* A record was found */
       if (get_rec_info)
 	{
+
```
```suggestion
```

xmilex-git:
> @xmilex-git Great. I assumed that a bottleneck in this function would be the fix(latch) contention (it'd be common on OLTP db) or IO occurred while fetching page. That operations are way more expensive than just cpu operations like `PGBUF_INIT_WATCHER`. How do you get that `4-5%` improvement? It's improved in general? or Isolated transaction without any I/O?

It's isolated transaction without any I/O. In this PR, I have made changes to reduce the overhead of the heap scan process itself, rather than the slowdown caused by I/O or page fixes.

wrlawodms (approved):
`It's isolated transaction without any I/O.`
Honestly, I still doubt it's worth improving with this assumption at the cost of harming maintenance. But, yeah, it may work. I don't want to block it keeping arguing. Let's move forward and see it!

hornetmj (approved):

youngjinj (approved):

PR #4757, opened 2023-11-02
[CBRD-25049] Core dumped in mvcc_is_mvcc_disabled_class at src/transaction/mvcc.c:618

xmilex-git:
http://jira.cubrid.org/browse/CBRD-25049

This issue is due to the changes made by the CBRD-25037 issue. The error occurred because the handling of the case where `class_oid`, used as an argument in `mvcc_is_mvcc_disabled_class()`, is NULL before executing `mvcc_is_mvcc_disabled_class()` was not handled. Therefore, modifications have been made to handle this case.

### DIFF
--- a/src/storage/heap_file.c
+++ b/src/storage/heap_file.c
@@ -24883,9 +24883,10 @@ heap_scan_get_visible_version (THREAD_ENTRY * thread_p, const OID * oid, OID * c
 	}
 
       const bool need_check_visibility = scan_cache->mvcc_snapshot != NULL
-	&& scan_cache->mvcc_snapshot->snapshot_fnc != NULL && !mvcc_is_mvcc_disabled_class (class_oid)
-	&& scan_cache->mvcc_snapshot->snapshot_fnc (thread_p, &mvcc_header,
-						    scan_cache->mvcc_snapshot) != SNAPSHOT_SATISFIED;
+	&& scan_cache->mvcc_snapshot->snapshot_fnc != NULL && class_oid != NULL
+	&& !mvcc_is_mvcc_disabled_class (class_oid)
+	&& scan_cache->mvcc_snapshot->snapshot_fnc (thread_p, &mvcc_header, scan_cache->mvcc_snapshot) !=
+	SNAPSHOT_SATISFIED;
 
       if (!need_check_visibility)
 	{


shparkcubrid (approved):

src/storage/heap_file.c
```diff
@@ -24883,9 +24883,10 @@ heap_scan_get_visible_version (THREAD_ENTRY * thread_p, const OID * oid, OID * c
 	}
 
       const bool need_check_visibility = scan_cache->mvcc_snapshot != NULL
-	&& scan_cache->mvcc_snapshot->snapshot_fnc != NULL && !mvcc_is_mvcc_disabled_class (class_oid)
-	&& scan_cache->mvcc_snapshot->snapshot_fnc (thread_p, &mvcc_header,
-						    scan_cache->mvcc_snapshot) != SNAPSHOT_SATISFIED;
+	&& scan_cache->mvcc_snapshot->snapshot_fnc != NULL && class_oid != NULL
```
wrlawodms, line 24886: Is this optimization still effective with this additional condition? If it's not, we should take out this part from the `heap_get_visible_version_internal`-> `heap_prepare_get_context`. I hate it, though.
```
 7442   /* Output class_oid if necessary. */
 7443   if (context->class_oid_p != NULL && OID_ISNULL (context->class_oid_p)
 7444       && heap_get_class_oid_from_page (thread_p, context->home_page_watcher.pgptr, context->class_oid_p) != NO_ERROR)
 7445     {
 7446       /* Unexpected. */
 7447       assert_release (false);
 7448       goto error;
 7449     }
```

xmilex-git: Even if there is a condition checking if `class_oid` is NULL, this solution is effective. Moreover, moving this part inside `heap_prepare_get_context` has a performance overhead. This is because there are two uses of the `replace_watcher` function when using the `context` data type, and profiling shows that this takes up more than 25% of the time in `heap_next()`.

wrlawodms: Good that we don't need to do that for this optimization. :)

wrlawodms (approved):

youngjinj (approved):

youngjinj (approved):

hornetmj (approved):

PR #4759, opened 2023-11-03
[LETS-842] Use the sysprm, checkpoint_interval, for checkpoint_tran as well

wrlawodms:
http://jira.cubrid.org/browse/LETS-842

I don't think we need a new system parameter to set the interval of `log_checkpoint_trantable_daemon`. We can make use of the existing one for `log_checkpoint_daemon`. Each of two daemons are for its own checkpoint mechanism of each server types, ATS and PS.

With this, the system parameter, `checkpoint_interval` means
- on ATS: the interval of `log_checkpoint_trantable_daemon`
- on PS:   the interval of `log_checkpoint_daemon`.

### DIFF
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -10885,8 +10885,7 @@ log_checkpoint_trantable_daemon_init ()
 
   if (is_tran_server_with_remote_storage ())
     {
-      constexpr auto loop_time = std::chrono::seconds (60);
-      cubthread::looper looper (loop_time);
+      cubthread::looper looper (log_get_checkpoint_interval);
       cubthread::entry_callable_task * daemon_task =
           new cubthread::entry_callable_task (log_checkpoint_trantable_execute);
       log_Checkpoint_trantable_daemon =


hornetmj (approved):

cristiarg (approved):

cristiarg:
@wrlawodms 

I gave my review without giving much thought.

But, on second thought, I think they should be 2 separate parameters:
- first, `checkpoint_interval` is currently default at 360 seconds (6 minutes); the cost of the PS checkpoint is "high" (as far as I know) as it effectively means flushing a potentially large number of pages on the disk
- the transaction table checkpoint on the ATS side is a cheap operation (in terms of I/O) as it only takes an "in memory" snapshot  of the transaction table and creates a log record out of it
- it will also be hard for a DBA to discern as to `checkpoint_interval` on the ATS means something and on PS node it means something else (even if documented appropriately)

"my 2 cents"

wrlawodms:
@cristiarg 

> But, on second thought, I think they should be 2 separate parameters:

It makes much sense. I think that they are conceptually same so that it would make DBA even harder to understand two separate parameters. However, yes, as you said, these two operations have way different cost, that I missed. I'll introduce new one: `checkpoint_trantable_interval`.

hornetmj:
@wrlawodms @cristiarg In my opinion, handling of http://jira.cubrid.org/browse/LETS-682 should be considered before introducing new system parameters. Therefore, it would not be too late to decide whether to separate system parameters after processing LETS-682.

wrlawodms:
@hornetmj The way I see it, considering that issue supports, @cristiarg 's opinion makes more sense because it makes the cost difference greater. Or do you meant to postpone to create a system parameter for a feature until the feature gets more solid?

hornetmj:
@wrlawodms I believe that the checkpoints performed in ATS may disappear depending on LETS-682 processing. Therefore, I think we can decide whether to reflect cristian's opinion after LETS-682 is processed.

wrlawodms:
@hornetmj  I'm not against it to proceed separate system parameters. I like the way, honestly. However, the issue doesn't remove the checkpoint on ATS, but just reduce the jobs of the checkpoint from using metafile + log record to log record.

wrlawodms:
So can I create an issue to separate and go with merging this?

hornetmj:
@wrlawodms I am not against it either. However, I have a slightly different opinion. From the user's perspective, if TS and PS need to set checkpoints using different system parameters, the user first needs to understand why they need to do so. And I believe this also has a similar problem with the data_buffer_size system parameter, etc. Anyway, I think the reason for distinguishing between system parameters is because of the default values. I prefer to provide a single system parameter and change the default values ​​for each transaction server type. (I don't know if it's possible) Also, I still think there will be something consideration of this issue when LETS-682 processing.

If you don't agree with my opinion, this is a good time to introduce new system parameter for ts checkpoint. It depends on you.

wrlawodms:
@hornetmj  
It's arguable. I think both have their point.
It also has some small issues to introduce new system parameter like naming, unit, and consistency with the existing one. What I need now is just that it should be configurable. So, I'd like to go with this minimum change and revisit later it gets more important to decide this. I'm quite sure that we will have a chance to organize all system parameters for LETS. Let's see then.  I think it's way later.

PR #4762, opened 2023-11-03
[CBRD-24941] miscellaneous code refinement and enhancement

hyunikn:
http://jira.cubrid.org/browse/CBRD-24941

- add syntax for comment to CREATE PROCEDRE/FUNCTION
  - But, comments are ignored by the Java compiler
- hardcode the default session timezone +09:00 and default locale en_US to print date-time values.
- log times taken by each compilation steps
- naming convention : use only plc  to indicate PL/CSQL and remove pcs
- check if a loop label is already declared
- error messages
- expressions in USING clause cannot be a CURSOR
- host expressions in Static SQLs cannot be a BOOLEAN, CURSOR, or SYS_REFCURSOR


src/jsp/jsp_cl.c
```diff
@@ -594,7 +594,7 @@ jsp_create_stored_procedure (PARSER_CONTEXT * parser, PT_NODE * statement)
   if (lang == SP_LANG_PLCSQL)
     {
       std::string pl_code (statement->sql_user_text, statement->sql_user_text_len);
-      err = plcsql_transfer_file (pl_code, false, compile_info);
+      err = plcsql_transfer_file (pl_code, true, compile_info);
```
hgryoo: The verbose mode is only for debugging.
```suggestion
      err = plcsql_transfer_file (pl_code, false, compile_info);
```

hyunikn: It can be benefitial to keep verbose mode for a few initial release versions of PL/CSQL and collect information from the customer sites obtained from the verbose mode logs.
Currently, the additional jobs done in the verbose mode are measuring times taken by compilations steps and logging them, whose overhead is expected to be small.

hyunikn: Updated as suggested.

hgryoo (approved):

PR #4765, opened 2023-11-06
[CBRD-25053] [Regression] Core dumped in tp_is_domain_cached at src/object/object_domain.c:2276

beyondykk9:
http://jira.cubrid.org/browse/CBRD-25053

The charset could be different for dblink. The charset should be checked together with collation.

### DIFF
--- a/src/object/object_domain.c
+++ b/src/object/object_domain.c
@@ -2245,6 +2245,7 @@ tp_is_domain_cached (TP_DOMAIN * dlist, TP_DOMAIN * transient, TP_MATCH exact, T
 
 	      match = ((domain->precision == transient->precision) && (domain->collation_id == transient->collation_id)
 		       && (domain->is_desc == transient->is_desc)
+		       && (domain->codeset == transient->codeset)
 		       && (domain->collation_flag == transient->collation_flag));
 	    }
 	  else if (exact == TP_STR_MATCH)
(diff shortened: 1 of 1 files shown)


hgryoo (approved):

src/object/object_domain.c
```diff
@@ -2245,6 +2245,7 @@ tp_is_domain_cached (TP_DOMAIN * dlist, TP_DOMAIN * transient, TP_MATCH exact, T
 
 	      match = ((domain->precision == transient->precision) && (domain->collation_id == transient->collation_id)
 		       && (domain->is_desc == transient->is_desc)
+		       && (domain->codeset == transient->codeset)
```
ctshim, line 2247: Is it only a problem if DB_TYPE_VARCHAR?
In the case of DB_TYPE_CHAR, is there no need to check the code set?

ctshim: Due to the added code, the assert statement at line 2279 loses its meaning.
assert (domain->codeset == transient->codeset);

src/object/object_domain.c
```diff
@@ -2683,7 +2686,8 @@ tp_domain_find_charbit (DB_TYPE type, int codeset, int collation_id, unsigned ch
 		{
 		  break;	/* found */
 		}
-	      else if (dom->collation_id == collation_id && dom->collation_flag == collation_flag)
+	      else if (dom->collation_id == collation_id && dom->collation_flag == collation_flag
+		       && dom->codeset == codeset)
```
ctshim, line 2694: I have the same question as the one I asked above.

src/object/object_domain.c
beyondykk9: OK, good point. I fixed it for CHAR and NCHAR.

src/object/object_domain.c
beyondykk9: OK, good point. I fixed it for CHAR and NCHAR.

hyunikn (approved):

ctshim (approved):

PR #4766, opened 2023-11-06
[CBRD-25054] Numeric type column inserted by using rownum does not show error even it's outranged.

beyondykk9:
http://jira.cubrid.org/browse/CBRD-25054

numeric_db_value_coerce_to_num doesn't handle overflow case for BIGINT type.


src/object/object_domain.h
```diff
@@ -517,6 +519,7 @@ extern "C"
   extern int tp_enumeration_to_varchar (const DB_VALUE * src, DB_VALUE * result);
   extern int tp_domain_status_er_set (TP_DOMAIN_STATUS status, const char *file_name, const int line_no,
 				      const DB_VALUE * src, const TP_DOMAIN * domain);
+  extern int tp_get_precision (DB_VALUE * value);
```
ctshim: There seems to be no need to declare it as extern.

src/object/object_domain.c
```diff
@@ -11475,6 +11475,77 @@ tp_domain_references_objects (const TP_DOMAIN * dom)
...
+
+  static INT64 max_value[19];	/* max size of a big integer is 19 */
+  static bool init_bigint_value = false;
+
+  if (!init_bigint_value)
+    {
+      int i;
+
+      max_value[0] = 1;
+      for (i = 1; i < 18; i++)
```
ctshim: Shouldn't it be "i < 19"?

beyondykk9: mistake.

src/object/object_domain.h
beyondykk9: Ok, fix it.

hyunikn (approved):

ctshim (approved):

hornetmj (approved):

hgryoo (approved):

PR #4767, opened 2023-11-07
[LETS-812] Discard the ddl related caches

joohok:
http://jira.cubrid.org/browse/LETS-812

Purpose

Caches that have been modified during DDL operation should be discarded, since the caches can be used by read transactions in PTS.

### DIFF
--- a/src/transaction/log_replication_atomic.cpp
+++ b/src/transaction/log_replication_atomic.cpp
@@ -161,7 +161,6 @@ namespace cublog
 
 	    if (m_locked_classes.count (header.trid) > 0)
 	      {
-		discard_caches_for_ddl (thread_entry, header.trid);
 		release_all_locks_for_ddl (thread_entry, header.trid);
 	      }
 
--- a/src/transaction/log_replication_atomic.cpp
+++ b/src/transaction/log_replication_atomic.cpp
@@ -434,7 +434,7 @@ namespace cublog
       {
 	const OID classoid = it->second;
 
-	heap_classrepr_decache (&thread_entry, &classoid);
+	(void) heap_classrepr_decache (&thread_entry, &classoid);
 	xcache_remove_by_oid (&thread_entry, &classoid);
 	partition_decache_class (&thread_entry, &classoid);
       }
--- a/src/transaction/log_replication_atomic.cpp
+++ b/src/transaction/log_replication_atomic.cpp
@@ -21,8 +21,10 @@
 #include "log_replication_jobs.hpp"
 
 #include "heap_file.h"
+#include "locator_sr.h"
 #include "log_recovery_redo_parallel.hpp"
 #include "oid.h"
+#include "xasl_cache.h"
 
 namespace cublog
 {
@@ -123,6 +125,7 @@ namespace cublog
 
 	    if (m_locked_classes.count (header.trid) > 0)
 	      {
+		discard_caches_for_ddl (thread_entry, header.trid);
 		release_all_locks_for_ddl (thread_entry, header.trid);
 	      }
 
@@ -158,6 +161,7 @@ namespace cublog
 
 	    if (m_locked_classes.count (header.trid) > 0)
 	      {
+		discard_caches_for_ddl (thread_entry, header.trid);
 		release_all_locks_for_ddl (thread_entry, header.trid);
 	      }
 
@@ -412,5 +416,27 @@ namespace cublog
 #endif
 
     m_locked_classes.emplace (trid, *classoid);
+  }
+
+  void
+  atomic_replicator::discard_caches_for_ddl (cubthread::entry &thread_entry, const TRANID trid)
+  {
+    assert (m_locked_classes.count (trid) > 0);
+
+    /* TODO:
+     * This is to update locator_Mht_classnames which contains class names,
+     * and it will be replaced with a function to update specific class name only if needed.
+     */
+    locator_initialize (&thread_entry);
+
+    auto [begin, end] = m_locked_classes.equal_range (trid);
+    for (auto it = begin; it != end; it++)
+      {
+	const OID classoid = it->second;
+
+	heap_classrepr_decache (&thread_entry, &classoid);
+	xcache_remove_by_oid (&thread_entry, &classoid);
+	partition_decache_class (&thread_entry, &classoid);
+      }
   }
 }
--- a/src/transaction/log_replication_atomic.cpp
+++ b/src/transaction/log_replication_atomic.cpp
@@ -431,9 +431,10 @@ namespace cublog
     auto [begin, end] = m_locked_classes.equal_range (trid);
     for (auto it = begin; it != end; it++)
       {
-	const OID classoid = it->second;
+	auto classoid = it->second;
 
 	(void) heap_classrepr_decache (&thread_entry, &classoid);
+	(void) heap_delete_hfid_from_cache (&thread_entry, &classoid);
 	xcache_remove_by_oid (&thread_entry, &classoid);
 	partition_decache_class (&thread_entry, &classoid);
       }
--- a/src/query/partition.c
+++ b/src/query/partition.c
@@ -2339,6 +2339,9 @@ reload_from_cache:
       /* Cache the loaded info. If the call below is successful, pinfo will be returned holding the cached information */
       partition_cache_pruning_context (pinfo, &already_exists);
 
+      /* cache is not modified during replication in PTS */
+      assert (thread_p->type != TT_REPLICATION_PTS);
+
       /* Multiple thread can reach here synchronously. In this case redo the action of load from cache. */
       if (already_exists)
 	{
--- a/src/query/xasl_cache.c
+++ b/src/query/xasl_cache.c
@@ -1369,6 +1369,9 @@ xcache_insert (THREAD_ENTRY * thread_p, const compile_context * context, XASL_ST
   assert (stream != NULL);
   assert (stream->buffer != NULL || !context->recompile_xasl);
 
+  /* cache can not be modified during replication in PTS */
+  assert (thread_p->type != TT_REPLICATION_PTS);
+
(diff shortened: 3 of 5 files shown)


wrlawodms (approved):

src/transaction/log_replication_atomic.cpp
```diff
@@ -123,6 +125,7 @@ namespace cublog
 
 	    if (m_locked_classes.count (header.trid) > 0)
 	      {
+		discard_caches_for_ddl (thread_entry, header.trid);
```
cristiarg, line 128: This remark has nothing to do with this PR. But with the previous one. :)

In `release_all_locks_for_ddl`:
```
    auto [begin, end] = m_locked_classes.equal_range (trid);
...
    m_locked_classes.erase (trid); // <------ this can be m_locked_classes.erase (begin, end);
```

src/transaction/log_replication_atomic.cpp
```diff
@@ -413,4 +417,26 @@ namespace cublog
...
+     * and it will be replaced with a function to update specific class name only if needed.
+     */
+    locator_initialize (&thread_entry);
+
+    auto [begin, end] = m_locked_classes.equal_range (trid);
+    for (auto it = begin; it != end; it++)
+      {
+	const OID classoid = it->second;
+
+	heap_classrepr_decache (&thread_entry, &classoid);
```
```suggestion
	(void) heap_classrepr_decache (&thread_entry, &classoid);
```

hornetmj (approved):

src/transaction/log_replication_atomic.cpp
joohok: Done!

src/transaction/log_replication_atomic.cpp
```diff
@@ -413,4 +417,26 @@ namespace cublog
...
+	xcache_remove_by_oid (&thread_entry, &classoid);
+	partition_decache_class (&thread_entry, &classoid);
```
cristiarg, line 439: Not sure what is the contention on these critical sections on PTS, but - does it make sense to create dedicated functions that take a list of OIDs and decache all of them?

For `heap_classrepr_decache` - critical section `CSECT_HEAP_CHNGUESS` is entered/exited
In the case of `xcache_remove_by_oid`, it's a bit more complicated - a completely new implementation for `xcache_invalidate_entries` is needed that takes a list of OIDs, collects the XASL id's for those that have to deal with the respective OIDs; and then removes the XASL caches
For `partition_decache_class` - critical section `CSECT_PARTITION_CACHE` is entered/exited

For the first and the last, it is relatively easy. For the second one it looks like more work.

Keep in mind that this is done "inline" by the PTS single threaded replication - and we do not know whether we'll be able to ever parallelize this replication.

I am thinking that, for example this de-cache-ing of class OIDs, could be "thrown out" on an async task. Does this make sense? Would it violate consistency on PTS (think - worst case scenario)?

The `locator_initialize` fucntion is even more costly - by retrieving pages from the PS to re-fill the class cache.

src/transaction/log_replication_atomic.cpp
```diff
@@ -158,6 +161,7 @@ namespace cublog
 
 	    if (m_locked_classes.count (header.trid) > 0)
 	      {
+		discard_caches_for_ddl (thread_entry, header.trid);
```
cristiarg: Honest question.
In case of an ATS transaction abort, the ATS will add undo log records for every "redo" log record added by the transaction up to that point.
Thus, after the transaction rollback, the database schema remains unchanged - from a logical point of view.
Is it the same from a physical point of view as well?
Or do some of the OIDs for the class definitions change following the application of the undo log records?
Because, if anything changes, then it makes sense to decache in the same way is done for commit.

Thus, does it make sense to discard the caches at this point?

Thank you.

joohok: I initially thought that the process of replicating DDL in PTS would modify the cache, so I considered discarding the cache safely even in case of a rollback. 
However, in ATS, modifying the cache is part of the logical changes during DDL execution, while in PTS, it only applies physical changes, so there's no need to modify the cache. Thanks! I'll test it further and incorporate the changes accordingly.

src/transaction/log_replication_atomic.cpp
joohok: I think there's no need to discard `heap_Guesschn` which uses `CSECT_HEAP_CHNGUESS` within `heap_classrepr_decache ()`. Therefore, I'll only call `heap_classsrepr_decache_guessed_last()`. Since `heap_Guesschn` gets updated with every read transaction, discarding it is unnecessary.

`xcache_remove_by_oid()` internally clears cache entries related to the provided OID. Transactions trying to use the xcache are already excluded from contention by SCH_M_LOCK, so I think additional optimization may not be necessary.
Or if needed, we can make a function `xcache_entry_is_related_to_oids()` similar to `xcache_entry_is_related_to_oid ()`

If we create a function like `partition_decache_classes()`, it should be possible to replace the `partition_decache_class()` function

The use of `locator_initialize()` is temporary, and it will likely be replaced by another function in upcoming issues. It's important that the de-cache process is completed before unlocking. Therefore, asynchronous operation seems risky. If the unlock happens before the cache is fully cleared, a waiting transaction, blocked by the lock, might read incorrect cache values.

src/transaction/log_replication_atomic.cpp
joohok: It appears that the behaviors of `m_locked_classes.erase (trid);` and `m_locked_classes.erase (begin, end);` are identical. Is there any advantage to changing them ? I'm not sure which one is better

src/transaction/log_replication_atomic.cpp
wrlawodms: @joohok Didn't you say that a cache (or some caches?) is updated in a redo function while replicating?

wrlawodms: And is there no chance a cache is updated while replicating? like
1. an entry for a class is de-cached due to a policy like LRU.
2. a new entry for the class is cached based on the changed physical image of the class while replicating.

If this is true, we should drop the cache after aborting. Isn't it possible?

joohok: @wrlawodms  
Yes, only the cache related to HFID (heap_Hfid_table) is updated during replication. 

1. I haven't considered it from an LRU perspective, but it seems that de-caching should not be a concern. The main goal is to prevent incomplete information from being cached during the replication process. 
Additionally, there should be some caching during the replication process for de-caching by LRU to take place.

2. Cache is modified only by the logical modifications. The sequence involves 1) a logical change request, 2) utilizing, modifying, and storing cache information, 3) followed by a physical image change. However, since the replicator only modifies the physical image, there are no cache changes except for HFID caches.

wrlawodms: 1. Is not the hfid_cache a target to decache on commit or abort by the replicator?
2. Would it be safer to add assert()s in cache-update functions that it must not be a replicator transaction or thread? Can it be identified?

joohok: 1. yes, because it is already de-cached by replicator in (vacuum_rv_notify_dropped_file ())
---------------------------
BUT .. I should also de-cache the hfid_table here, because I didn't consider the TRUNCATE case. It will have same OID but different HFID, so I have to discard the cache. If it is not de-cached, read transaction can refer to old heap file (heap file is changed by TRUNCATE TABLE statement).   
2. maybe we can identify by the thread type (=TT_REPLICATION_PTS), and I will add an assert for that

src/transaction/log_replication_atomic.cpp
cristiarg: > Transactions trying to use the xcache are already excluded from contention by SCH_M_LOCK, so I think additional optimization may not be necessary.

Indeed there would be no contention because of the incompatible locks. However, my concern is with the performance and the time it takes for the single replication thread itself.

> Or if needed, we can make a function xcache_entry_is_related_to_oids() similar to xcache_entry_is_related_to_oid ()

Yes, that is what I was thinking - make it so that there as few `CSECT` acquisitions and releases as possible.

> If we create a function like partition_decache_classes(), it should be possible to replace the partition_decache_class() function

:)

> Therefore, asynchronous operation seems risky. If the unlock happens before the cache is fully cleared, a waiting transaction, blocked by the lock, might read incorrect cache values.

Yes, you are right. My bad for not thinking about this when I wrote the comment.
That's why it is important for the implementation of the de-cache in the replicator to be as optimal as possible - to not stall the replication too much.
Even if the DDL operations will be __very rare__ compared with the DML operations.

src/transaction/log_replication_atomic.cpp
cristiarg: > an entry for a class is de-cached due to a policy like LRU.

Where is this mechanism implemented? You mean `heap_classrepr_cache::LRU_list` ?

> a new entry for the class is cached based on the changed physical image of the class while replicating.

That for sure will lead to inconsistent cache - the caching logic might even not be able to conclude the caching because heap pages that contain the class information are still scrambled.

This can be mitigated by putting the DDL replication log record sequences under the umbrella of "atomic sequences" (thus the affected heap pages will be kept locked for writing until the image is consistent). And the caching mechanism will not be queued to access these pages to reconstruct the cache.

wrlawodms: > Where is this mechanism implemented? You mean heap_classrepr_cache::LRU_list ?

Oh, no. LRU is a hypothetical example. If this kind of policy exists, the cache could be inconsistent. So, I just asked.

cristiarg (approved):
Reviewed. We will need tests for the `LOG_ABORT` scenario.

src/transaction/log_replication_atomic.cpp
joohok: I agree. I will add optimization issue for this. 
Since it looks like there will be quite a few changes, I'll create a new issue and handle it in parallel!

PR #4771, opened 2023-11-09
[LETS-813] Update the classname cache only if needed

joohok:
http://jira.cubrid.org/browse/LETS-813

Purpose
When starting DDL:
1. Check if there is a classname associated with the classOID.
2. If there is a classname, temporarily keep it (`m_classname_map`)
3. Upon DDL commit, check if the classname has changed.
   - If the classname was not present but is now, it's a CREATE statement.
   - If the classname was present but is now missing, it's a DROP statement.
   - If the classname has changed, it's a RENAME statement.
4. Update locator_Mht_classnames accordingly for each case.

NOTE: 
Modified the data structure of `m_locked_classes` from `std::multimap <TRANID, OID>` to `std::unordered_map <TRANID, std::unordered_set <OID>>`.
because : 
- Previous assumption
   - Even if multiple DDLs are performed for the same class within a single transaction, 
   only one LOG_SCHEMA_MODIFICATION_LOCK log should be generated. 
   - Because the <classname, oid, lock mode> is cached in CAS (workspace), and client would use the cached information. 
- Actual test result
   - Cached lock in CAS (workspace) is de-cached when semantic check error occur
      - example
      Rename table tbl to tbl2;  -- lock info is cached in workspace
      Rename table tbl to tbl2;  -- lock info is decached from workspace
      Rename table tbl2 to tbl;  -- To cache the lock information, it request lock info from server, and leave a LOG_SCHEMA_MODIFICATION_LOCK again.
- A cleaner solution will be implemented in the following issue

### DIFF
--- a/src/transaction/locator_sr.c
+++ b/src/transaction/locator_sr.c
@@ -13971,18 +13971,6 @@ locator_put_classname_entry (THREAD_ENTRY * thread_p, const char *classname, con
 
   LOCATOR_CLASSNAME_ENTRY *entry;
 
-  if (csect_enter (thread_p, CSECT_LOCATOR_SR_CLASSNAME_TABLE, INF_WAIT) != NO_ERROR)
-    {
-      assert (false);
-      return ER_FAILED;
-    }
-  // *INDENT-OFF*
-  scope_exit <std::function <void (void)>> classname_csect_exit ([&thread_p] ()
-      {
-        csect_exit (thread_p, CSECT_LOCATOR_SR_CLASSNAME_TABLE);
-      });
-  // *INDENT-ON*
-
   entry = (LOCATOR_CLASSNAME_ENTRY *) malloc (sizeof (*entry));
   if (entry == nullptr)
     {
@@ -13991,6 +13979,22 @@ locator_put_classname_entry (THREAD_ENTRY * thread_p, const char *classname, con
     }
 
   locator_initialize_classname_entry (entry, classname, classoid);
+
+  if (csect_enter (thread_p, CSECT_LOCATOR_SR_CLASSNAME_TABLE, INF_WAIT) != NO_ERROR)
+    {
+      assert (false);
+      return ER_FAILED;
+    }
+
+  // *INDENT-OFF*
+  auto classname_csect_exit = scope_exit ([&thread_p] ()
+      {
+        csect_exit (thread_p, CSECT_LOCATOR_SR_CLASSNAME_TABLE);
+      });
+  // *INDENT-ON*
+
+  assert (locator_is_exist_class_name_entry (thread_p, entry));
+  assert (mht_get (locator_Mht_classnames, classname) == nullptr);
 
   (void) mht_put (locator_Mht_classnames, entry->e_name, entry);
 
@@ -14013,18 +14017,9 @@ locator_remove_classname_entry (THREAD_ENTRY * thread_p, const char *classname)
       assert (false);
       return ER_FAILED;
     }
-
-  if (csect_enter (thread_p, CSECT_CT_OID_TABLE, INF_WAIT) != NO_ERROR)
-    {
-      assert (false);
-      csect_exit (thread_p, CSECT_LOCATOR_SR_CLASSNAME_TABLE);
-      return ER_FAILED;
-    }
-
   // *INDENT-OFF*
-  scope_exit <std::function <void (void)>> classname_csect_exit ([&thread_p, &classname] ()
+  auto classname_csect_exit = scope_exit ([&thread_p] ()
       {
-        csect_exit (thread_p, CSECT_CT_OID_TABLE);
         csect_exit (thread_p, CSECT_LOCATOR_SR_CLASSNAME_TABLE);
       });
   // *INDENT-ON*
@@ -14035,6 +14030,20 @@ locator_remove_classname_entry (THREAD_ENTRY * thread_p, const char *classname)
       assert (false);
       return ER_FAILED;
     }
+
+  if (csect_enter (thread_p, CSECT_CT_OID_TABLE, INF_WAIT) != NO_ERROR)
+    {
+      assert (false);
+      csect_exit (thread_p, CSECT_LOCATOR_SR_CLASSNAME_TABLE);
+      return ER_FAILED;
+    }
+
+  // *INDENT-OFF*
+  auto catalog_csect_exit = scope_exit ([&thread_p, &classname] ()
+      {
+        csect_exit (thread_p, CSECT_CT_OID_TABLE);
+      });
+  // *INDENT-ON*
 
   (void) locator_force_drop_class_name_entry (entry->e_name, entry, NULL);
 
@@ -14063,18 +14072,9 @@ locator_update_classname_entry (THREAD_ENTRY * thread_p, const char *old_classna
       return ER_FAILED;
     }
 
-  if (csect_enter (thread_p, CSECT_CT_OID_TABLE, INF_WAIT) != NO_ERROR)
-    {
-      assert (false);
-      csect_exit (thread_p, CSECT_LOCATOR_SR_CLASSNAME_TABLE);
-
-      return ER_FAILED;
-    }
-
   // *INDENT-OFF*
-  scope_exit <std::function <void (void)>> classname_csect_exit ([&thread_p] ()
+  auto classname_csect_exit = scope_exit ([&thread_p] ()
       {
-        csect_exit (thread_p, CSECT_CT_OID_TABLE);
         csect_exit (thread_p, CSECT_LOCATOR_SR_CLASSNAME_TABLE);
       });
   // *INDENT-ON*
@@ -14093,7 +14093,23 @@ locator_update_classname_entry (THREAD_ENTRY * thread_p, const char *old_classna
 
   locator_initialize_classname_entry (new_entry, new_classname, &old_entry->e_current.oid);
 
-  (void) mht_put (locator_Mht_classnames, new_entry->e_name, new_entry);
+  assert (locator_is_exist_class_name_entry (thread_p, new_entry));
+  assert (mht_get (locator_Mht_classnames, new_classname) == nullptr);
+
+  (void) mht_put (locator_Mht_classnames, new_classname, new_entry);
+
+  if (csect_enter (thread_p, CSECT_CT_OID_TABLE, INF_WAIT) != NO_ERROR)
+    {
+      assert (false);
+      return ER_FAILED;
+    }
+
+  // *INDENT-OFF*
+  auto catalog_csect_exit = scope_exit ([&thread_p, &old_classname] ()
+      {
+        csect_exit (thread_p, CSECT_CT_OID_TABLE);
+      });
+  // *INDENT-ON*
 
   /* drop the existing entry with key (old_classname) */
   (void) locator_force_drop_class_name_entry (old_entry->e_name, old_entry, NULL);
--- a/src/transaction/locator_sr.c
+++ b/src/transaction/locator_sr.c
@@ -14034,7 +14034,6 @@ locator_remove_classname_entry (THREAD_ENTRY * thread_p, const char *classname)
   if (csect_enter (thread_p, CSECT_CT_OID_TABLE, INF_WAIT) != NO_ERROR)
     {
       assert (false);
-      csect_exit (thread_p, CSECT_LOCATOR_SR_CLASSNAME_TABLE);
       return ER_FAILED;
     }
 
--- a/src/transaction/locator_sr.c
+++ b/src/transaction/locator_sr.c
@@ -13934,15 +13934,7 @@ locator_put_classname_entry (THREAD_ENTRY * thread_p, const char *classname, con
 
   if (csect_enter (thread_p, CSECT_LOCATOR_SR_CLASSNAME_TABLE, INF_WAIT) != NO_ERROR)
     {
-      /* Some kind of failure. We must notify the error to the caller. */
       assert (false);
-      return;
-    }
-
-  if (csect_enter (thread_p, CSECT_CT_OID_TABLE, INF_WAIT) != NO_ERROR)
-    {
-      assert (false);
-      csect_exit (thread_p, CSECT_LOCATOR_SR_CLASSNAME_TABLE);
       return;
     }
 
@@ -13971,7 +13963,6 @@ locator_put_classname_entry (THREAD_ENTRY * thread_p, const char *classname, con
   (void) mht_put (locator_Mht_classnames, entry->e_name, entry);
 
   csect_exit (thread_p, CSECT_LOCATOR_SR_CLASSNAME_TABLE);
-  csect_exit (thread_p, CSECT_CT_OID_TABLE);
 }
 
 void
@@ -13981,7 +13972,6 @@ locator_remove_classname_entry (THREAD_ENTRY * thread_p, const char *classname)
 
   if (csect_enter (thread_p, CSECT_LOCATOR_SR_CLASSNAME_TABLE, INF_WAIT) != NO_ERROR)
     {
-      /* Some kind of failure. We must notify the error to the caller. */
       assert (false);
       return;
     }
@@ -13999,6 +13989,32 @@ locator_remove_classname_entry (THREAD_ENTRY * thread_p, const char *classname)
 
   csect_exit (thread_p, CSECT_LOCATOR_SR_CLASSNAME_TABLE);
   csect_exit (thread_p, CSECT_CT_OID_TABLE);
+}
+
+void
+locator_update_classname_entry (THREAD_ENTRY * thread_p, const char *classname)
+{
+  LOCATOR_CLASSNAME_ENTRY *entry = NULL;
+
+  if (csect_enter (thread_p, CSECT_LOCATOR_SR_CLASSNAME_TABLE, INF_WAIT) != NO_ERROR)
+    {
+      assert (false);
+      return;
+    }
+
+  entry = (LOCATOR_CLASSNAME_ENTRY *) mht_get (locator_Mht_classnames, classname);
+ 
+  assert (entry != NULL);
+  assert (entry->e_name != NULL);
+  free_and_init (entry->e_name);
+
+  entry->e_tran_index = NULL_TRAN_INDEX;
+  entry->e_name = strdup ((char *) classname);
+  entry->e_tran_index = NULL_TRAN_INDEX;
+  entry->e_current.action = LC_CLASSNAME_EXIST;
+  LSA_SET_NULL (&entry->e_current.savep_lsa);
+
+  csect_exit (thread_p, CSECT_LOCATOR_SR_CLASSNAME_TABLE);
 }
 
 bool
--- a/src/transaction/locator_sr.c
+++ b/src/transaction/locator_sr.c
@@ -13925,6 +13925,19 @@ locator_multi_insert_force (THREAD_ENTRY * thread_p, HFID * hfid, OID * class_oi
   return error_code;
 }
 
+bool
+has_errors_filtered_for_insert (std::vector<int> error_filter_array)
+{
+  if (std::find (error_filter_array.begin(), error_filter_array.end(), ER_BTREE_UNIQUE_FAILED)
+                 != error_filter_array.end ())
+  {
+    return true;
+  }
+
+  return false;
+}
+// *INDENT-ON*
+
 /* locator_put_classname_entry () - Put a new entry in the locator_Mht_classnames, used in PTS replicator (atomic_replicator)
  *
  * return	     : void
@@ -13933,7 +13946,7 @@ locator_multi_insert_force (THREAD_ENTRY * thread_p, HFID * hfid, OID * class_oi
  * classoid (in)     : class oid of the classname
  */
 void
-locator_put_classname_entry (THREAD_ENTRY * thread_p, const char *classname, const OID *classoid)
+locator_put_classname_entry (THREAD_ENTRY * thread_p, const char *classname, const OID * classoid)
 {
   assert (classname != NULL);
   assert (strlen (classname) < DB_MAX_IDENTIFIER_LENGTH);
@@ -13949,29 +13962,29 @@ locator_put_classname_entry (THREAD_ENTRY * thread_p, const char *classname, con
   scope_exit <std::function <void ()>> classname_csect_exit ([&thread_p] () {csect_exit (thread_p, CSECT_LOCATOR_SR_CLASSNAME_TABLE); });
   // *INDENT-ON*
 
-entry = ((LOCATOR_CLASSNAME_ENTRY *) malloc (sizeof (*entry)));
-if (entry == NULL)
-  {
-    er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, sizeof (*entry));
-    return;
-  }
+  entry = ((LOCATOR_CLASSNAME_ENTRY *) malloc (sizeof (*entry)));
+  if (entry == NULL)
+    {
+      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, sizeof (*entry));
+      return;
+    }
 
-entry->e_name = strdup ((char *) classname);
-if (entry->e_name == NULL)
-  {
-    free_and_init (entry);
-    er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, (size_t) (strlen (classname) + 1));
-    return;
-  }
+  entry->e_name = strdup ((char *) classname);
+  if (entry->e_name == NULL)
+    {
+      free_and_init (entry);
+      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, (size_t) (strlen (classname) + 1));
+      return;
+    }
 
-entry->e_tran_index = NULL_TRAN_INDEX;
+  entry->e_tran_index = NULL_TRAN_INDEX;
 
-entry->e_current.action = LC_CLASSNAME_EXIST;
-COPY_OID (&entry->e_current.oid, classoid);
-LSA_SET_NULL (&entry->e_current.savep_lsa);
-entry->e_current.prev = NULL;
+  entry->e_current.action = LC_CLASSNAME_EXIST;
+  COPY_OID (&entry->e_current.oid, classoid);
+  LSA_SET_NULL (&entry->e_current.savep_lsa);
+  entry->e_current.prev = NULL;
 
-(void) mht_put (locator_Mht_classnames, entry->e_name, entry);
+  (void) mht_put (locator_Mht_classnames, entry->e_name, entry);
 }
 
 /* locator_remove_classname_entry () - Remove an entry from the locator_Mht_classnames, used in PTS replicator (atomic_replicator)
@@ -14073,20 +14086,6 @@ locator_update_classname_entry (THREAD_ENTRY * thread_p, const char *old_classna
 
   csect_exit (thread_p, CSECT_CT_OID_TABLE);
 }
-
-bool
-has_errors_filtered_for_insert (std::vector < int >error_filter_array)
-{
-  if (std::find (error_filter_array.begin (), error_filter_array.end (), ER_BTREE_UNIQUE_FAILED)
-      != error_filter_array.end ())
-    {
-      return true;
-    }
-
-  return false;
-}
-
-// *INDENT-ON*
 
 void
 xsynonym_remove_xasl_by_oid (THREAD_ENTRY * thread_p, OID * oidp)
--- a/src/transaction/locator_sr.c
+++ b/src/transaction/locator_sr.c
@@ -238,8 +238,8 @@ static DB_LOGICAL locator_mvcc_reev_cond_and_assignment (THREAD_ENTRY * thread_p
 							 MVCC_REC_HEADER * mvcc_header_p,
 							 const OID * curr_row_version_oid_p, RECDES * recdes);
 
-static inline int locator_initialize_classname_entry (LOCATOR_CLASSNAME_ENTRY * entry, const char *classname,
-						      const OID * class_oid);
+static inline void locator_initialize_classname_entry (LOCATOR_CLASSNAME_ENTRY * entry, const char *classname,
+						       const OID * class_oid);
 
 /*
  * locator_initialize () - Initialize the locator on the server
@@ -13942,22 +13942,18 @@ has_errors_filtered_for_insert (std::vector<int> error_filter_array)
 // *INDENT-ON*
 
 
-static inline int
+static inline void
 locator_initialize_classname_entry (LOCATOR_CLASSNAME_ENTRY * entry, const char *classname, const OID * classoid)
 {
-  entry->e_name = strdup (classname);
-  if (entry->e_name == nullptr)
-    {
-      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, sizeof (classname));
-      return ER_OUT_OF_VIRTUAL_MEMORY;
-    }
+  assert (classname != nullptr);
+
+  /* Memory for classname is allocated in atomic_replicator::update_classname_cache_for_ddl () */
+  entry->e_name = (char *) classname;
   entry->e_tran_index = NULL_TRAN_INDEX;
   entry->e_current.action = LC_CLASSNAME_EXIST;
   COPY_OID (&entry->e_current.oid, classoid);
   LSA_SET_NULL (&entry->e_current.savep_lsa);
   entry->e_current.prev = nullptr;
-
-  return NO_ERROR;
 }
 
 /* locator_put_classname_entry () - Put a new entry in the locator_Mht_classnames, used in PTS replicator (atomic_replicator)
@@ -13994,12 +13990,7 @@ locator_put_classname_entry (THREAD_ENTRY * thread_p, const char *classname, con
       return ER_OUT_OF_VIRTUAL_MEMORY;
     }
 
-  if (locator_initialize_classname_entry (entry, classname, classoid) != NO_ERROR)
-    {
-      free_and_init (entry->e_name);
-      free_and_init (entry);
-      return ER_FAILED;
-    }
+  locator_initialize_classname_entry (entry, classname, classoid);
 
   (void) mht_put (locator_Mht_classnames, entry->e_name, entry);
 
@@ -14100,10 +14091,7 @@ locator_update_classname_entry (THREAD_ENTRY * thread_p, const char *old_classna
       return ER_OUT_OF_VIRTUAL_MEMORY;
     }
 
-  if (locator_initialize_classname_entry (new_entry, new_classname, &old_entry->e_current.oid) != NO_ERROR)
-    {
-      return ER_FAILED;
-    }
+  locator_initialize_classname_entry (new_entry, new_classname, &old_entry->e_current.oid);
 
   (void) mht_put (locator_Mht_classnames, new_entry->e_name, new_entry);
 
--- a/src/transaction/locator_sr.c
+++ b/src/transaction/locator_sr.c
@@ -13949,7 +13949,7 @@ locator_initialize_classname_entry (LOCATOR_CLASSNAME_ENTRY * entry, const char 
   if (entry->e_name == nullptr)
     {
       er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, sizeof (classname));
-      return ER_FAILED;
+      return ER_OUT_OF_VIRTUAL_MEMORY;
     }
   entry->e_tran_index = NULL_TRAN_INDEX;
   entry->e_current.action = LC_CLASSNAME_EXIST;
--- a/src/transaction/locator_sr.c
+++ b/src/transaction/locator_sr.c
@@ -238,8 +238,8 @@ static DB_LOGICAL locator_mvcc_reev_cond_and_assignment (THREAD_ENTRY * thread_p
 							 MVCC_REC_HEADER * mvcc_header_p,
 							 const OID * curr_row_version_oid_p, RECDES * recdes);
 
-static inline void locator_initialize_classname_entry (LOCATOR_CLASSNAME_ENTRY * entry, const char *classname,
-						       const OID * class_oid);
+static inline int locator_initialize_classname_entry (LOCATOR_CLASSNAME_ENTRY * entry, const char *classname,
+						      const OID * class_oid);
 
 /*
  * locator_initialize () - Initialize the locator on the server
@@ -318,20 +318,11 @@ locator_initialize (THREAD_ENTRY * thread_p)
 	  goto error;
 	}
 
-      entry->e_name = strdup ((char *) classname);
-      if (entry->e_name == NULL)
+      if (locator_initialize_classname_entry (entry, classname, &class_oid) != NO_ERROR)
 	{
 	  free_and_init (entry);
-	  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, (size_t) (strlen (classname) + 1));
 	  goto error;
 	}
-
-      entry->e_tran_index = NULL_TRAN_INDEX;
-
-      entry->e_current.action = LC_CLASSNAME_EXIST;
-      COPY_OID (&entry->e_current.oid, &class_oid);
-      LSA_SET_NULL (&entry->e_current.savep_lsa);
-      entry->e_current.prev = NULL;
 
       assert (locator_is_exist_class_name_entry (thread_p, entry));
 
@@ -13942,18 +13933,26 @@ has_errors_filtered_for_insert (std::vector<int> error_filter_array)
 // *INDENT-ON*
 
 
-static inline void
+static inline int
 locator_initialize_classname_entry (LOCATOR_CLASSNAME_ENTRY * entry, const char *classname, const OID * classoid)
 {
   assert (classname != nullptr);
 
   /* Memory for classname is allocated in atomic_replicator::update_classname_cache_for_ddl () */
-  entry->e_name = (char *) classname;
+  entry->e_name = strdup (classname);
+  if (entry->e_name == nullptr)
+    {
+      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, strlen (classname));
+      return ER_OUT_OF_VIRTUAL_MEMORY;
+    }
+
   entry->e_tran_index = NULL_TRAN_INDEX;
   entry->e_current.action = LC_CLASSNAME_EXIST;
   COPY_OID (&entry->e_current.oid, classoid);
   LSA_SET_NULL (&entry->e_current.savep_lsa);
   entry->e_current.prev = nullptr;
+
+  return NO_ERROR;
 }
 
 /* locator_put_classname_entry () - Put a new entry in the locator_Mht_classnames, used in PTS replicator (atomic_replicator)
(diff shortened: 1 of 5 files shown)


src/transaction/locator_sr.c
```diff
@@ -13937,6 +13938,155 @@ has_errors_filtered_for_insert (std::vector<int> error_filter_array)
...
+  if (csect_enter (thread_p, CSECT_LOCATOR_SR_CLASSNAME_TABLE, INF_WAIT) != NO_ERROR)
+    {
+      assert (false);
+      return;
+    }
+  // *INDENT-OFF*
+  scope_exit <std::function <void ()>> classname_csect_exit ([&thread_p] () {csect_exit (thread_p, CSECT_LOCATOR_SR_CLASSNAME_TABLE); });
+  // *INDENT-ON*
+
+  entry = ((LOCATOR_CLASSNAME_ENTRY *) malloc (sizeof (*entry)));
```
hornetmj:
```suggestion
  entry = (LOCATOR_CLASSNAME_ENTRY *) malloc (sizeof (*entry));
```

src/transaction/locator_sr.c
```diff
@@ -13937,6 +13938,155 @@ has_errors_filtered_for_insert (std::vector<int> error_filter_array)
...
+    }
+
+  if (csect_enter (thread_p, CSECT_CT_OID_TABLE, INF_WAIT) != NO_ERROR)
+    {
+      assert (false);
+      csect_exit (thread_p, CSECT_LOCATOR_SR_CLASSNAME_TABLE);
+      return;
+    }
+
+  entry = (LOCATOR_CLASSNAME_ENTRY *) mht_get (locator_Mht_classnames, classname);
```
hornetmj: The returted entry need to check or add assert for it.

src/transaction/locator_sr.c
```diff
@@ -13937,6 +13938,155 @@ has_errors_filtered_for_insert (std::vector<int> error_filter_array)
...
+
+  // *INDENT-OFF*
+  scope_exit <std::function <void ()>> classname_csect_exit ([&thread_p] () {csect_exit (thread_p, CSECT_LOCATOR_SR_CLASSNAME_TABLE); });
+  // *INDENT-ON*
+
+  old_entry = (LOCATOR_CLASSNAME_ENTRY *) mht_get (locator_Mht_classnames, old_classname);
+  assert (old_entry != NULL);
+
+  /* create new entry with new key (new_classname) */
+  new_entry = ((LOCATOR_CLASSNAME_ENTRY *) malloc (sizeof (*new_entry)));
```
hornetmj:
```suggestion
  new_entry = (LOCATOR_CLASSNAME_ENTRY *) malloc (sizeof (*new_entry));
```

src/transaction/log_replication_atomic.hpp
```diff
@@ -107,6 +110,7 @@ namespace cublog
        * Since multiple DDL operations can be executed within single transaction,
        * more than one classes can be mapped to one transaction */
       std::multimap <TRANID, OID> m_locked_classes;
+      std::unordered_map <OID, std::string> m_classname_map;
```
hornetmj, line 116: It looks like problems may arise if multiple ddl statements are performed on the same table in one transaction. Do you also have a verification scenario for this case?

src/transaction/locator_sr.c
joohok: OK! done.

src/transaction/log_replication_atomic.hpp
joohok: There is no problem I found. 
Only the names of the class 
1) at the time of lock acquisition (m_classname_map) and 
2) at the time of commit are needed. 

Even if multiple `LOG_SCHEMA_MODIFICATION_LOCK`s are replicated, the name stored in m_classname_map is from the moment of the first lock acquisition, so it doesn't matter.

I have tested various DDL operations within a single transaction and confirmed that there are no problems.
- CREATE - RENAME - DROP
- RENAME - RENAME - RENAME
- RENAME - RENAME - DROP
- CREATE - RENAME - RENAME

src/transaction/log_replication_atomic.cpp
```diff
@@ -402,37 +402,72 @@ namespace cublog
...
-#endif
+    m_locked_classes[trid].emplace (*classoid);
+  }
+
+  void
+  atomic_replicator::bookkeep_classname (cubthread::entry &thread_entry, const OID *classoid)
+  {
+    assert (!OID_ISNULL (classoid) && !OID_ISTEMP (classoid));
+    char *classname = NULL;
+
```
```suggestion
    char *classname = NULL;
    
    assert (!OID_ISNULL (classoid) && !OID_ISTEMP (classoid));

```

src/transaction/log_replication_atomic.cpp
```diff
@@ -402,37 +402,72 @@ namespace cublog
...
+    (void) heap_get_class_name (&thread_entry, classoid, &classname);
+    if (classname != NULL)
+      {
+	m_classname_map.emplace (*classoid, classname);
```
hornetmj: I guess it needs to guarantee that a entry with classoid key in the map doesn't exist. How about adding checking logic in debug build?

src/transaction/log_replication_atomic.cpp
```diff
@@ -402,37 +402,72 @@ namespace cublog
...
+	free_and_init (classname);
+      }
+    else
+      {
+	/* classname == NULL */
+	if (m_classname_map.find (*classoid) != m_classname_map.end ())
+	  {
+	    /* If the classname is removed after DDL, then it is DROP TABLE/VIEW case */
+	    locator_remove_classname_entry (&thread_entry, m_classname_map[*classoid].c_str ());
+	  }
```
```suggestion
	/* If the classname is removed after DDL, then it is DROP TABLE/VIEW case */
	if (m_classname_map.find (*classoid) != m_classname_map.end ())
	  {	    
	    locator_remove_classname_entry (&thread_entry, m_classname_map[*classoid].c_str ());
	  }
```

hornetmj (approved):
LGTM.

src/transaction/log_replication_atomic.cpp
joohok: Since multiple LOG_SCHEMA_MODIFICATION_LOCKs can be replicated, an entry with classOID key can exist in the map. 

I think it is OK because `emplace` won't be performed if there is already a value for the same key (classoid)

src/transaction/log_replication_atomic.cpp
joohok: Done!

src/transaction/log_replication_atomic.cpp
joohok: Done!

src/transaction/locator_sr.c
```diff
@@ -13937,6 +13938,162 @@ has_errors_filtered_for_insert (std::vector<int> error_filter_array)
...
+ *
+ * return	     : void
+ * thread_p (in)     : thread entry
+ * classname (in)    : classname to be inserted
+ * classoid (in)     : class oid of the classname
+ */
+void
+locator_put_classname_entry (THREAD_ENTRY * thread_p, const char *classname, const OID * classoid)
+{
+  assert (classname != NULL);
```
cristiarg: Nit: use `nullptr` for all new code.
```suggestion
  assert (classname != nullptr);
```

src/transaction/locator_sr.c
```diff
@@ -13937,6 +13938,162 @@ has_errors_filtered_for_insert (std::vector<int> error_filter_array)
...
+
+  LOCATOR_CLASSNAME_ENTRY *entry;
+
+  if (csect_enter (thread_p, CSECT_LOCATOR_SR_CLASSNAME_TABLE, INF_WAIT) != NO_ERROR)
+    {
+      assert (false);
+      return;
+    }
+  // *INDENT-OFF*
+  scope_exit <std::function <void ()>> classname_csect_exit ([&thread_p] () {csect_exit (thread_p, CSECT_LOCATOR_SR_CLASSNAME_TABLE); });
```
cristiarg: Nit, easier to read
```suggestion
  scope_exit <std::function <void (void)>> classname_csect_exit ([&thread_p] ()
    {
      csect_exit (thread_p, CSECT_LOCATOR_SR_CLASSNAME_TABLE);
    });
```

src/transaction/locator_sr.c
```diff
@@ -13937,6 +13938,162 @@ has_errors_filtered_for_insert (std::vector<int> error_filter_array)
...
+  // *INDENT-ON*
+
+  entry = (LOCATOR_CLASSNAME_ENTRY *) malloc (sizeof (*entry));
+  if (entry == NULL)
+    {
+      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, sizeof (*entry));
+      return;
+    }
+
+  entry->e_name = strdup ((char *) classname);
```
cristiarg: In this particular case you could just take over the `classname` pointer. Because in the calling code it is freed afterwards any. But I suppose it is clearer semantics to just follow suit with all the surrounding code which duplicates strings.

src/transaction/locator_sr.c
```diff
@@ -13937,6 +13938,162 @@ has_errors_filtered_for_insert (std::vector<int> error_filter_array)
...
+      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, (size_t) (strlen (classname) + 1));
+      return;
+    }
+
+  entry->e_tran_index = NULL_TRAN_INDEX;
+
+  entry->e_current.action = LC_CLASSNAME_EXIST;
+  COPY_OID (&entry->e_current.oid, classoid);
+  LSA_SET_NULL (&entry->e_current.savep_lsa);
+  entry->e_current.prev = NULL;
```
cristiarg: An initialization macro/inline function is appropriate for `LOCATOR_CLASSNAME_ACTION`.

src/transaction/locator_sr.c
```diff
@@ -13937,6 +13938,162 @@ has_errors_filtered_for_insert (std::vector<int> error_filter_array)
...
 
+/* locator_put_classname_entry () - Put a new entry in the locator_Mht_classnames, used in PTS replicator (atomic_replicator)
+ *
+ * return	     : void
+ * thread_p (in)     : thread entry
+ * classname (in)    : classname to be inserted
+ * classoid (in)     : class oid of the classname
+ */
+void
+locator_put_classname_entry (THREAD_ENTRY * thread_p, const char *classname, const OID * classoid)
```
cristiarg, line 13965: What if these new functions fail?
Or must they never fail?
I think we need to return the error and - at least - be able to debug log the error in the PTS replicator logic.

src/transaction/locator_sr.c
```diff
@@ -13937,6 +13938,162 @@ has_errors_filtered_for_insert (std::vector<int> error_filter_array)
...
+      assert (false);
+      csect_exit (thread_p, CSECT_CT_OID_TABLE);
+      csect_exit (thread_p, CSECT_LOCATOR_SR_CLASSNAME_TABLE);
+      return;
+    }
+
+  (void) locator_force_drop_class_name_entry (entry->e_name, entry, NULL);
+
+  csect_exit (thread_p, CSECT_LOCATOR_SR_CLASSNAME_TABLE);
+  csect_exit (thread_p, CSECT_CT_OID_TABLE);
```
cristiarg: Isn't the  order here the other way around from how the critical sections where acquired?
- first free `CSECT_CT_OID_TABLE`
- then free `CSECT_LOCATOR_SR_CLASSNAME_TABLE`

And I think it is cleaner to use `scope_exit` everywhere. Or not use it at all. :)

src/transaction/locator_sr.c
```diff
@@ -13937,6 +13938,162 @@ has_errors_filtered_for_insert (std::vector<int> error_filter_array)
...
+  LOCATOR_CLASSNAME_ENTRY *new_entry = NULL;
+
+  if (csect_enter (thread_p, CSECT_LOCATOR_SR_CLASSNAME_TABLE, INF_WAIT) != NO_ERROR)
+    {
+      assert (false);
+      return;
+    }
+
+  // *INDENT-OFF*
+  scope_exit <std::function <void ()>> classname_csect_exit ([&thread_p] () {csect_exit (thread_p, CSECT_LOCATOR_SR_CLASSNAME_TABLE); });
```
cristiarg: Nit. Alignment.

src/transaction/locator_sr.c
```diff
@@ -13937,6 +13938,162 @@ has_errors_filtered_for_insert (std::vector<int> error_filter_array)
...
+  // *INDENT-ON*
+
+  old_entry = (LOCATOR_CLASSNAME_ENTRY *) mht_get (locator_Mht_classnames, old_classname);
```
cristiarg: Nit:
```suggestion
  LOCATOR_CLASSNAME_ENTRY *const old_entry = (LOCATOR_CLASSNAME_ENTRY *) mht_get (locator_Mht_classnames, old_classname);
```

src/transaction/locator_sr.c
```diff
@@ -13937,6 +13938,162 @@ has_errors_filtered_for_insert (std::vector<int> error_filter_array)
...
+      free_and_init (new_entry);
+      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, (size_t) (strlen (new_classname) + 1));
+      return;
+    }
+
+  new_entry->e_tran_index = NULL_TRAN_INDEX;
+  new_entry->e_current.action = LC_CLASSNAME_EXIST;
+  COPY_OID (&new_entry->e_current.oid, &old_entry->e_current.oid);
+  LSA_SET_NULL (&new_entry->e_current.savep_lsa);
+  new_entry->e_current.prev = NULL;
```
cristiarg: Nit. initializer macro/inline function for `LOCATOR_CLASSNAME_ACTION`.

src/transaction/log_replication_atomic.cpp
```diff
@@ -402,37 +402,72 @@ namespace cublog
...
-#endif
+    m_locked_classes[trid].emplace (*classoid);
+  }
+
+  void
+  atomic_replicator::bookkeep_classname (cubthread::entry &thread_entry, const OID *classoid)
+  {
+    char *classname = NULL;
+
+    assert (!OID_ISNULL (classoid) && !OID_ISTEMP (classoid));
```
cristiarg, line 416: Question. Is `assert !OID_IS_ROOTOID` relevant here?

src/transaction/log_replication_atomic.cpp
```diff
@@ -402,37 +402,72 @@ namespace cublog
...
 
-    m_locked_classes.emplace (trid, *classoid);
+    (void) heap_get_class_name (&thread_entry, classoid, &classname);
+    if (classname != NULL)
+      {
+	m_classname_map.emplace (*classoid, classname);
```
cristiarg: I think the `char *` can be moved into an `std::string`:
```suggestion
	m_classname_map.emplace (*classoid, std::move (std::string { classname } ));
```
.. and no need to `free_and_init` below.

src/transaction/log_replication_atomic.hpp
```diff
@@ -90,6 +90,9 @@ namespace cublog
       void set_lowest_unapplied_lsa ();
       void replicate_sysop_start_postpone (cubthread::entry &thread_entry, const LOG_RECORD_HEADER &rec_header);
 
+      void bookkeep_classname (cubthread::entry &thread_entry, const OID *classoid);
```
cristiarg: I would move all DDL related code to a separate helper.
What do you think? Maybe in a separate PR?

src/transaction/log_replication_atomic.cpp
```diff
@@ -402,37 +402,72 @@ namespace cublog
...
+
+  void
+  atomic_replicator::update_classname_cache_for_ddl (cubthread::entry &thread_entry, const OID *classoid)
+  {
+    /* update locator_Mht_classnames only if needed */
+    char *classname = NULL;
+    (void) heap_get_class_name (&thread_entry, classoid, &classname);
+    if (classname != NULL)
+      {
+	if (m_classname_map.find (*classoid) == m_classname_map.end ())
```
cristiarg: Call `find`, store result it in an iterator variable. Reuse below instead of `m_classname_map[*classoid]`.
Every access to the `find`/`operator[]` actually does a search in the hash-map underneath.

src/transaction/log_replication_atomic.cpp
```diff
@@ -402,37 +402,72 @@ namespace cublog
...
+
+	free_and_init (classname);
+      }
+    else
+      {
+	/* If the classname is removed after DDL, then it is DROP TABLE/VIEW case */
+	if (m_classname_map.find (*classoid) != m_classname_map.end ())
+	  {
+	    locator_remove_classname_entry (&thread_entry, m_classname_map[*classoid].c_str ());
+	  }
```
cristiarg, line 490: else here, explaining that the class name did not change?

src/transaction/log_replication_atomic.cpp
```diff
@@ -402,37 +402,72 @@ namespace cublog
...
+      {
+	/* if classname == NULL, then it is CREATE TABLE/VIEW case */
+      }
+  }
+
+  void
+  atomic_replicator::update_classname_cache_for_ddl (cubthread::entry &thread_entry, const OID *classoid)
+  {
+    /* update locator_Mht_classnames only if needed */
+    char *classname = NULL;
```
cristiarg: Just to make it clear for me - to make sure I understand:
- the class name obtained from `heap_get_class_name` here is the actual class name obtained from the heap page and is the actual result __after__ replicating the DDL-related log records
- the class name in `m_classname_map` is the class as obtained from `heap_get_class_name` __before__ replicating the DDL-related log records - in `bookkeep_classname` above.

And based on this, below it is decided if it is a CREATE/DROP/UPDATE statement.

Question:
Are we sure that - for `CREATE` statements - the class name does not appear somewhere in the log records chain as as a result of multiple `LOG_SCHEMA_MODIFICATION_LOCK` for a `CREATE ...` statement? Or is this possibility covered by-design by the `locator_update_classname_entry` function.

In other words:
Is it possible for a `CREATE` statement to not be detected as such because the __before__ (in  `m_classname_map`) and __after__ (directly from the heap) classnames contain the same name?

src/transaction/log_replication_atomic.hpp
joohok: When creating this PR, I did consider creating a helper for DDL-related jobs. I think it is better to minimize coupling for now and, when finishing the epic, consolidate the DDL-related code as a refactoring.

src/transaction/log_replication_atomic.cpp
joohok: Thanks..! I found a hole.. I will find out some general solution for this
possible scenario : 

CREATE TABLE tbl ( a int);
ALTER TABLE tbl add column b int;
ALTER TABLE tbl add column b int; -- [semantic check error], insert "tbl" into `m_classname_map` 
COMMIT; -- before and after can contain the same name..

cristiarg: One idea would be to only take into account the first occurence of the `LOG_SCHEMA_MODIFICATION_LOCK` (as, in your example - resulting from the first `ALTER TABLE` statement).
Upon a second time, the fact that the `OID` points to a valid class name is not taken into account.

As in, for a transaction occuring in the log record chain:
1. `LOG_SCHEMA_MODIFICATION_LOCK` encountered, class name not found - record this information and the fact that we're dealing with a `CREATE`
2. `LOG_SCHEMA_MODIFICATION_LOCK` encountered again, do not check anymore for the class name
3. at `COMMIT` - the logic should work and properly detect a `CREATE` statement

This is more of a "Captain Obvious" remark, as I am sure you have already a solution for this. :)

src/transaction/locator_sr.c
joohok: Done!

src/transaction/locator_sr.c
joohok: Done!

src/transaction/locator_sr.c
joohok: I will use `scope_exit`

src/transaction/locator_sr.c
joohok: Done!

src/transaction/log_replication_atomic.cpp
joohok: I've pondered if there's a cleaner way. In truth, logging just one `LOG_SCHEMA_MODIFICATION_LOCK` in the ATS would be the cleanest option. because logging and replicating is also the cost. However, for the sake of clarity, I want to address it within the PTS first. Your suggested approach seems to be the neatest if we aim to resolve it within the PTS. Thanks a lot!

src/transaction/log_replication_atomic.cpp
joohok: OK, done

src/transaction/log_replication_atomic.cpp
joohok: I will add some comment.

src/transaction/log_replication_atomic.cpp
joohok: Thanks

src/transaction/locator_sr.c
joohok: I return the error, and add debug logging

src/transaction/log_replication_atomic.cpp
joohok: Since we already check whether it's a root class or not when logging, it seems unnecessary to assert it here.

src/transaction/locator_sr.c
joohok: Done.

src/transaction/locator_sr.c
joohok: Done.

src/transaction/locator_sr.c
joohok: Done.

src/transaction/locator_sr.c
```diff
@@ -13937,6 +13941,166 @@ has_errors_filtered_for_insert (std::vector<int> error_filter_array)
...
 // *INDENT-ON*
 
+
+static inline void
+locator_initialize_classname_entry (LOCATOR_CLASSNAME_ENTRY * entry, const char *classname, const OID * classoid)
+{
+  assert (classname != nullptr);
+
+  /* Memory for classname is allocated in atomic_replicator::update_classname_cache_for_ddl () */
+  entry->e_name = (char *) classname;
```
wrlawodms: Shouldn't it be copied?
```suggestion
      entry->e_name = strdup ((char *) classname);
      if (entry->e_name == NULL)
```

src/transaction/locator_sr.c
```diff
@@ -13937,6 +13941,166 @@ has_errors_filtered_for_insert (std::vector<int> error_filter_array)
...
+
+  LOCATOR_CLASSNAME_ENTRY *entry;
+
+  if (csect_enter (thread_p, CSECT_LOCATOR_SR_CLASSNAME_TABLE, INF_WAIT) != NO_ERROR)
+    {
+      assert (false);
+      return ER_FAILED;
+    }
+  // *INDENT-OFF*
+  scope_exit <std::function <void (void)>> classname_csect_exit ([&thread_p] ()
```
wrlawodms: Nit. I think it can be simplified with the _template_agrument_deduction_.
```suggestion
  auto classname_csect_exit = scope_exit {([&thread_p] ()
```

src/transaction/locator_sr.c
```diff
@@ -13937,6 +13941,166 @@ has_errors_filtered_for_insert (std::vector<int> error_filter_array)
...
+
+  entry = (LOCATOR_CLASSNAME_ENTRY *) malloc (sizeof (*entry));
+  if (entry == nullptr)
+    {
+      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, sizeof (*entry));
+      return ER_OUT_OF_VIRTUAL_MEMORY;
+    }
+
+  locator_initialize_classname_entry (entry, classname, classoid);
+
```
wrlawodms, line 14017:
```suggestion
assert (locator_is_exist_class_name_entry (thread_p, entry));
```

src/transaction/locator_sr.c
```diff
@@ -13937,6 +13941,166 @@ has_errors_filtered_for_insert (std::vector<int> error_filter_array)
...
+/* locator_put_classname_entry () - Put a new entry in the locator_Mht_classnames, used in PTS replicator (atomic_replicator)
+ *
+ * return	     : error code
+ * thread_p (in)     : thread entry
+ * classname (in)    : classname to be inserted
+ * classoid (in)     : class oid of the classname
+ */
+int
+locator_put_classname_entry (THREAD_ENTRY * thread_p, const char *classname, const OID * classoid)
```
wrlawodms, line 13965: Is this function not supposed to be called with the same class name? Can we make an assert for that?

src/transaction/locator_sr.c
```diff
@@ -13937,6 +13941,166 @@ has_errors_filtered_for_insert (std::vector<int> error_filter_array)
...
+  if (csect_enter (thread_p, CSECT_LOCATOR_SR_CLASSNAME_TABLE, INF_WAIT) != NO_ERROR)
+    {
+      assert (false);
+      return ER_FAILED;
+    }
+
+  if (csect_enter (thread_p, CSECT_CT_OID_TABLE, INF_WAIT) != NO_ERROR)
+    {
+      assert (false);
+      csect_exit (thread_p, CSECT_LOCATOR_SR_CLASSNAME_TABLE);
```
wrlawodms: How about this?

```
csect_enter (thread_p, CSECT_LOCATOR_SR_CLASSNAME_TABLE);
init scope_exit for `CSECT_LOCATOR_SR_CLASSNAME_TABLE`.
csect_enter (thread_p, CSECT_CT_OID_TABLE, INF_WAIT);
init scope_exit for `CSECT_CT_OID_TABLE`.
```
```suggestion
```

src/transaction/locator_sr.c
```diff
@@ -13937,6 +13941,166 @@ has_errors_filtered_for_insert (std::vector<int> error_filter_array)
...
+  // *INDENT-ON*
+
+  entry = (LOCATOR_CLASSNAME_ENTRY *) malloc (sizeof (*entry));
+  if (entry == nullptr)
+    {
+      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, sizeof (*entry));
+      return ER_OUT_OF_VIRTUAL_MEMORY;
+    }
+
+  locator_initialize_classname_entry (entry, classname, classoid);
```
wrlawodms: It seems that this initialization of an entry can be out of the latch, `CSECT_LOCATOR_SR_CLASSNAME_TABLE`. What do you think?

src/transaction/locator_sr.c
```diff
@@ -13937,6 +13941,166 @@ has_errors_filtered_for_insert (std::vector<int> error_filter_array)
 }
 // *INDENT-ON*
 
+
+static inline void
+locator_initialize_classname_entry (LOCATOR_CLASSNAME_ENTRY * entry, const char *classname, const OID * classoid)
```
wrlawodms, line 13937: Can't this be used in `locator_initialize` as well?

wrlawodms, line 13937: I think we can merge the "alloc" and this "initialization" since they are always together, and this function doesn't return anything. `locator_alloc_classname_entry`

src/transaction/locator_sr.c
```diff
@@ -13937,6 +13941,166 @@ has_errors_filtered_for_insert (std::vector<int> error_filter_array)
...
+  new_entry = (LOCATOR_CLASSNAME_ENTRY *) malloc (sizeof (*new_entry));
+  if (new_entry == nullptr)
+    {
+      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, sizeof (*new_entry));
+      return ER_OUT_OF_VIRTUAL_MEMORY;
+    }
+
+  locator_initialize_classname_entry (new_entry, new_classname, &old_entry->e_current.oid);
+
+  (void) mht_put (locator_Mht_classnames, new_entry->e_name, new_entry);
```
wrlawodms, line 14137: Can we add an assert that the entry doesn't exist?

src/transaction/log_replication_atomic.cpp
```diff
@@ -239,8 +239,15 @@ namespace cublog
 	    const LOG_REC_SCHEMA_MODIFICATION_LOCK log_rec =
 		    m_redo_context.m_reader.reinterpret_copy_and_add_align<LOG_REC_SCHEMA_MODIFICATION_LOCK> ();
 
+	    // check if m_locked_classes[header.trid] has log_rec.classoid
```
wrlawodms: Nit. I think it's redundant and even reveals details unnecessarily.
```suggestion
```

src/transaction/log_replication_atomic.cpp
```diff
@@ -402,41 +408,122 @@ namespace cublog
...
 
-    m_locked_classes.emplace (trid, *classoid);
+  void
+  atomic_replicator::bookkeep_classname (cubthread::entry &thread_entry, const OID *classoid)
+  {
+    char *classname = nullptr;
+
+    assert (!OID_ISNULL (classoid) && !OID_ISTEMP (classoid));
+
+    (void) heap_get_class_name (&thread_entry, classoid, &classname);
```
wrlawodms: No need to check the returned error?

src/transaction/log_replication_atomic.cpp
```diff
@@ -402,41 +408,122 @@ namespace cublog
...
+  {
+    if (m_locked_classes.count (trid) > 0)
+      {
+	if (m_locked_classes[trid].count (*classoid) > 0)
+	  {
+	    return true;
+	  }
+      }
+
+    return false;
```
```suggestion
    return m_locked_classes.count (trid) > 0 && m_locked_classes[trid].count (*classoid);
```

src/transaction/locator_sr.c
wrlawodms: On second thought, it seems better to split it. Sorry, forget this.

src/transaction/locator_sr.c
joohok: I used `strdup` at initial commit. However, since the classname was also allocated in the caller function, I attempted to use an already allocated heap memory to reduce the number of memory allocations and free. However, I am a bit concerned about the readability

src/transaction/locator_sr.c
joohok: I thought I block that situation by using `atomic_replicator::is_already_locked_for_ddl`, 
but I will add an assert

src/transaction/locator_sr.c
joohok: Thanks!!

src/transaction/locator_sr.c
joohok: Agree, I will move the start point of critical section after this line.

src/transaction/log_replication_atomic.cpp
joohok: Thanks

src/transaction/locator_sr.c
joohok: Yes, it can be used in `locator_initialize ()` as well if `entry->e_name` is copied by `strdup ()` as you suggested, 
but I didn't replace the code in `locator_initialize()` because it is not related to this issue.

src/transaction/locator_sr.c
joohok: I will add this assertion

src/transaction/locator_sr.c
joohok: OK, it looks cleaner

src/transaction/locator_sr.c
joohok: Yes! I will add

src/transaction/log_replication_atomic.cpp
joohok: OK, I will remove it

src/transaction/log_replication_atomic.cpp
joohok: hmm.. I can add an assert for the error code.
when CREATE TABLE is replicated, 
`classname == nullptr` and error code will be set to `ER_HEAP_UNKNOWN_OBJECT`.

How about adding an assertion when 
if classname == nullptr, then error ==  ER_HEAP_UNKNOWN_OBJECT
else classname != nullptr, then error ==  ER_HEAP_UNKNOWN_OBJECT

then we can filter out other failure cases (e.g ER_FAILED)

Rudeus (approved):

src/transaction/locator_sr.c
wrlawodms: I'd use `strdup` not only for readability, but also the fact that a caller should know the passed string may be released somewhere. It's "moved" actually, and this should be warned by users of this function. I doubt it's worth it, so I'd just use `strdup`.

src/transaction/locator_sr.c
wrlawodms: Why can't we touch the existing code, even though we take the snippet as a function? It seems to me that it's related. We don't change any logic.

src/transaction/locator_sr.c
```diff
@@ -13937,6 +13941,181 @@ has_errors_filtered_for_insert (std::vector<int> error_filter_array)
...
+ * old_classname (in) : old classname to be removed
+ * new_classname (in) : new classname to be inserted
+ */
+int
+locator_update_classname_entry (THREAD_ENTRY * thread_p, const char *old_classname, const char *new_classname)
+{
+  assert (old_classname != nullptr);
+  assert (new_classname != nullptr);
+
+  LOCATOR_CLASSNAME_ENTRY *new_entry = nullptr;
```
wrlawodms: Nit. 
Let's keep consistent: declare a variable at the beginning or with its allocation:

`LOCATOR_CLASSNAME_ENTRY *new_entry = nullptr;`
`LOCATOR_CLASSNAME_ENTRY *const old_entry =
     (LOCATOR_CLASSNAME_ENTRY *) mht_get (locator_Mht_classnames, old_classname);`

src/transaction/locator_sr.c
```diff
@@ -13937,6 +13941,181 @@ has_errors_filtered_for_insert (std::vector<int> error_filter_array)
...
+
+  /* create new entry with new key (new_classname) */
+  new_entry = (LOCATOR_CLASSNAME_ENTRY *) malloc (sizeof (*new_entry));
+  if (new_entry == nullptr)
+    {
+      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, sizeof (*new_entry));
+      return ER_OUT_OF_VIRTUAL_MEMORY;
+    }
+
+  locator_initialize_classname_entry (new_entry, new_classname, &old_entry->e_current.oid);
```
wrlawodms: It can be located out of the lock as the `locator_put_classname_entry` does.

src/transaction/locator_sr.c
```diff
@@ -13937,6 +13941,181 @@ has_errors_filtered_for_insert (std::vector<int> error_filter_array)
...
+  (void) mht_put (locator_Mht_classnames, new_classname, new_entry);
+
+  if (csect_enter (thread_p, CSECT_CT_OID_TABLE, INF_WAIT) != NO_ERROR)
+    {
+      assert (false);
+      return ER_FAILED;
+    }
+
+  // *INDENT-OFF*
+  auto catalog_csect_exit = scope_exit ([&thread_p, &old_classname] ()
```
wrlawodms, line 14147: This seems because the `catcls_remove_entry` is called internally, but we don't put any a this kind of entry in the `locator_put_classname_entry`. Is it OK? De we need to consider `catcls_Class_oid_to_oid_hash_table` or not?

src/transaction/log_replication_atomic.cpp
```diff
@@ -402,41 +407,114 @@ namespace cublog
...
+  {
+    char *classname = nullptr;
 
-    m_locked_classes.emplace (trid, *classoid);
+    assert (!OID_ISNULL (classoid) && !OID_ISTEMP (classoid));
+
+    (void) heap_get_class_name (&thread_entry, classoid, &classname);
+    if (classname != nullptr)
+      {
+	m_classname_map.emplace (*classoid, std::move (std::string (classname)));
```
```suggestion
	m_classname_map.emplace (*classoid, std::string (classname));
```
I don't think we need to cast it to R-ref since it's already R-ref. Please correct me if I'm wrong.

src/transaction/log_replication_atomic.cpp
```diff
@@ -402,41 +407,114 @@ namespace cublog
...
-      assert (false);
-    }
-#endif
+    m_locked_classes[trid].emplace (*classoid);
+  }
+
+  void
+  atomic_replicator::bookkeep_classname (cubthread::entry &thread_entry, const OID *classoid)
+  {
+    char *classname = nullptr;
```
wrlawodms, line 418: The `heap_get_class_name` says: The classname space must be released by the caller.
and std::string() copies it.

src/transaction/log_replication_atomic.cpp
wrlawodms: OK. I think this way is enough, 
```
if (error != NO_ERROR)
{
   assert (error == ER_HEAP_UNKNOWN_OBJECT);
}
```

wrlawodms: How about branching based on the error code, not by the `classname`? It seems better since the return value means "what's happened" and intuitive. like
```
if (error == NO_ERROR)
{
 assert (class_name != nullptr);
  m_classname_map.emplace (*classoid, std::move (std::string (classname)));
} 
else if (error == ER_HEAP_UNKNOWN_OBJECT)
{
  assert (class_name == nullptr);
  /* no class record means it's creating the class */
}
else 
{
  assert (false);
}
```

wrlawodms: The same goes for the `update_classname_cache_for_ddl`.

src/transaction/log_replication_atomic.cpp
```diff
@@ -402,41 +407,114 @@ namespace cublog
...
+	    if (it->second != classname)
+	      {
+		error_code = locator_update_classname_entry (&thread_entry, it->second.c_str(), classname);
+		if (error_code != NO_ERROR)
+		  {
+		    _er_log_debug (ARG_FILE_LINE, "[REPL_DDL] Failed to update entry for (%s) to (%s) in locator_Mht_classnames\n",
+				   it->second.c_str (), classname);
+		  }
+	      }
+
```
```suggestion
```

src/transaction/log_replication_atomic.cpp
```diff
@@ -402,41 +407,114 @@ namespace cublog
...
+	  }
+      }
+    else
+      {
+	/* if classname == nullptr, then it is DROP TABLE/VIEW case */
```
```suggestion
	/* if classname == nullptr, the class (table/view) has been dropped during the transaction. */
```
The transaction include not only DROP, but other DDLs.

src/transaction/locator_sr.c
joohok: OK, I think it make sense. I will replace the existing code with this function.

src/transaction/locator_sr.c
joohok: OK, I agree. I will change it

src/transaction/locator_sr.c
joohok: I initialized `old_entry` after entering the critical section because it accessed `locator_Mht_classnames` using `mht_get()`. I will make the modifications as consistently as possible

src/transaction/locator_sr.c
joohok: `old_entry` is used to initialize the entry, so it in the critical section.

src/transaction/locator_sr.c
joohok: You are right. The critical section is used because of `catcls_remove_entry()`. However, `catcls_Class_oid_to_oid_hash_table` is only used in ATS. It is used only when creating or updating classes, and the results of these operations are logged and replicated to PTS.

Even if there is no entry in catcls_Class_oid_to_oid_hash_table and fails to remove the entry, it always returns NO_ERROR in `catcls_remove_entry ()`, so there was no issue from this removal.

src/transaction/log_replication_atomic.cpp
joohok: OK, I will release the memory for classname

src/transaction/log_replication_atomic.cpp
joohok: I agree, I will modify as you suggest

src/transaction/log_replication_atomic.cpp
joohok: Oh, yes. Thanks.

src/transaction/log_replication_atomic.cpp
joohok: I agree, and I changed it.

src/transaction/log_replication_atomic.cpp
joohok: Done, thanks.

src/transaction/locator_sr.c
```diff
@@ -13937,6 +13932,231 @@ has_errors_filtered_for_insert (std::vector<int> error_filter_array)
 }
 // *INDENT-ON*
 
+
+static inline int
+locator_initialize_classname_entry (LOCATOR_CLASSNAME_ENTRY * entry, const char *classname, const OID * classoid)
+{
+  assert (classname != nullptr);
+
+  /* Memory for classname is allocated in atomic_replicator::update_classname_cache_for_ddl () */
```
wrlawodms:
```suggestion
```

src/transaction/locator_sr.h
```diff
@@ -139,6 +139,13 @@ extern int locator_multi_insert_force (THREAD_ENTRY * thread_p, HFID * hfid, OID
 				       HEAP_SCANCACHE * scan_cache, int *force_count, int pruning_type,
 				       PRUNING_CONTEXT * pcontext, FUNC_PRED_UNPACK_INFO * func_preds,
 				       UPDATE_INPLACE_STYLE force_in_place, bool dont_check_fk);
+// *INDENT-ON*
+extern int locator_put_classname_entry (THREAD_ENTRY * thread_p, const char *classname, const OID * class_oid);
+extern int locator_remove_classname_entry (THREAD_ENTRY * thread_p, const char *classname);
+extern int locator_update_classname_entry (THREAD_ENTRY * thread_p, const char *old_classname,
+					   const char *new_classname);
+
+// *INDENT-OFF*
```
wrlawodms:
```suggestion
extern int locator_put_classname_entry (THREAD_ENTRY * thread_p, const char *classname, const OID * class_oid);
extern int locator_remove_classname_entry (THREAD_ENTRY * thread_p, const char *classname);
extern int locator_update_classname_entry (THREAD_ENTRY * thread_p, const char *old_classname,
					   const char *new_classname);
```

src/transaction/log_replication_atomic.cpp
```diff
@@ -239,8 +239,14 @@ namespace cublog
 	    const LOG_REC_SCHEMA_MODIFICATION_LOCK log_rec =
 		    m_redo_context.m_reader.reinterpret_copy_and_add_align<LOG_REC_SCHEMA_MODIFICATION_LOCK> ();
 
+	    if (is_already_locked_for_ddl (header.trid, &log_rec.classoid))
```
```suggestion
	    if (is_locked_for_ddl (header.trid, &log_rec.classoid))
```
 I don't think "already" is needed. At your choice.

src/transaction/log_replication_atomic.cpp
```diff
@@ -239,8 +239,14 @@ namespace cublog
...
+	      {
+		break;
+	      }
+
 	    acquire_lock_for_ddl (thread_entry, header.trid, &log_rec.classoid);
 
+	    bookkeep_classname (thread_entry, &log_rec.classoid);
```
wrlawodms: Nit. 
It needs a more specific name like
- `bookkeep_classname_for_ddl`,
- `bookkeep_classname_for_cache` or something describing its purpose.

wrlawodms (approved):

src/transaction/locator_sr.c
joohok: Oh.. thanks..

src/transaction/locator_sr.h
joohok: this was because above and below functions are using c++ standard library, but I will move the newly defined function names below.

src/transaction/log_replication_atomic.cpp
joohok: OK, I will change the name

src/transaction/log_replication_atomic.cpp
joohok: bookkeep_classname_for_ddl is better, I will change

PR #4772, opened 2023-11-09
[CBRD-25063] Save original view query

beyondykk9:
http://jira.cubrid.org/browse/CBRD-25063

In 11.3, when a view is created in a public account and unloaded, the view information includes public account even though user information was not provided when creating the view.
Unloading with a specific user can be used to transfer only the specific user's schema/table to another location. And that user can change. Therefore, it must be saved as the original select statement used by the user when creating it. I understand that other DBs are also stored in sentence form when created.
It can also create a view by referencing another schema. This should be included because it is not yours, but there is no need to include it. Of course, if it did that when creating it, it has no choice, so it has to leave it as is.


src/transaction/boot_cl.c
```diff
@@ -4706,6 +4706,8 @@ boot_define_view_vclass (void)
 	  "[%s] AS [c] "
 	"WHERE "
 	  "[q].[class_of] = [c] "
+          "AND (([q].[class_of].[is_system_class] = 1 and [q].[spec] > 'S')"
+          "  OR ([q].[class_of].[is_system_class] = 0 and [q].[spec] < 'S'))"
```
ctshim, line 4710: It would be nice to have a comment on the reason for dividing based on 'S'.

src/query/execute_schema.c
```diff
@@ -379,6 +379,8 @@ static int do_alter_index_status (PARSER_CONTEXT * parser, const PT_NODE * state
 
 int ib_thread_count = 0;
 
+#define MAX_QUERY_STRING_LENGTH 4096
```
ctshim, line 382: How about calling it MAX_VIEW_QUERY_STRING_LENGTH.
I think the meaning may be misunderstood.

In the code below, before copying with strncpy(), you need to check whether the maximum length is exceeded and handle an error.

src/parser/csql_lexer.l
```diff
@@ -65,6 +65,10 @@ int str_identifier = '\0';
 #define IS_BLANK_CHAR(c) \
                     ((c) == ' ' || (c) == '\t' || (c) == '\r' || (c) == '\n')
 
+/* for extracting view's query spec in loaddb */
+char stream_buffer[4096];
```
ctshim, line 69: It would be better to declare it using MAX_QUERY_STRING_LENGTH.

src/parser/parse_tree_cl.c
```diff
@@ -1573,6 +1578,9 @@ fgetin (PARSER_CONTEXT * p)
 
   c = fgetc (p->file);
 
+  /* a statement buffer to extract view's query spec in loaddb */
+  stream_buffer[stream_ptr++] = (char) c;
```
ctshim, line 1582: You must check that the maximum buffer size is not exceeded. A core dump may occur if there is an error in the input data or an unusually long query string.
For example, let's assume the following situation.

CREATE or replace VIEW xview AS SELECT /* long long long comment */ * FROM tbl WHERE v1 > 0 /* comment */ and v2 > 0 /* comment */;

ctshim:
Looking at the operation of the function below, it appears that the rewritten query and the annotated original query are connected in the form of ordered pairs.
`smt_drop_query_spec(), smt_change_query_spec()`

This must always be guaranteed.
If so, I think we can simplify the inner loop a bit more.

src/object/schema_template.c
```diff
@@ -4066,12 +4066,18 @@ int
 smt_drop_query_spec (SM_TEMPLATE * def, const int index)
 {
   int error = NO_ERROR;
-  SM_QUERY_SPEC *file, *prev, *found;
+  SM_QUERY_SPEC *file, *prev, *found, *user_query_spec;
   int i;
   char indexname[20];
 
-  for (file = def->query_spec, prev = NULL, found = NULL, i = 1; file != NULL && found == NULL; file = file->next, i++)
+  for (file = def->query_spec, prev = NULL, found = NULL, i = 1; file != NULL && found == NULL; file = file->next)
```
ctshim, line 4073: I think it can be simplified as below.

```
for (file = def->query_spec, prev = NULL, found = NULL, i = 1; file != NULL;  i++)
    {
      if (index == i)
	{
	  found = file;
          break;
	}
       assert(file->next != NULL && file->next->specification[0] == '-')
       prev = file->next;
       file = file->next->next;
    }
 assert(found == NULL ||  (found->next != NULL && found->next->specification[0] == '-'));
```

src/object/schema_template.c
```diff
@@ -4090,13 +4098,24 @@ smt_drop_query_spec (SM_TEMPLATE * def, const int index)
     }
   else
     {
+      user_query_spec = found->next;
       if (prev == NULL)
 	{
 	  def->query_spec = found->next;
+	  if (user_query_spec && user_query_spec->specification[0] == '-')
```
ctshim, line 4105: This if statement must always be true.
How about replacing it with assert?

PR #4775, opened 2023-11-10
[CBRD-25060] The problem of not selecting a better index which is superset of primary key.

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-25060

Rule-based optimization works when the cost difference is within 20%. The rule-based logic has been modified as follows.

- single table : Delete routine that unconditionally prefers unique index.
- join table : Select superset index by comparing terms of each inner index.


src/optimizer/query_planner.c
```diff
@@ -3562,6 +3562,23 @@ qo_plan_cmp (QO_PLAN * a, QO_PLAN * b)
...
     }
 
+  /* check for superset of index */
+  if (a->plan_type == QO_PLANTYPE_JOIN && QO_IS_NL_JOIN (a) && qo_is_iscan (a->plan_un.join.inner) &&
+      b->plan_type == QO_PLANTYPE_JOIN && QO_IS_NL_JOIN (b) && qo_is_iscan (b->plan_un.join.inner))
+    {
+      temp_res = qo_plan_iscan_terms_cmp (a->plan_un.join.inner, b->plan_un.join.inner);
+      if (temp_res == PLAN_COMP_LT)
+	{
+	  QO_PLAN_CMP_CHECK_COST (af + aa, bf + ba);
```
xmilex-git, line 3572: The QO_PLAN_CMP_CHECK_COST() macro appears to be defined without any content. What is the purpose of having this code?

shparkcubrid: I think It is legacy code for checking cost difference when rule base is applied.

xmilex-git (approved):

youngjinj (approved):

PR #4776, opened 2023-11-10
[LETS-814] Modify the classname cache status when DROP TABLE/VIEW is replicated

joohok:
http://jira.cubrid.org/browse/LETS-814

Purpose

The status of classname cache should be changed when DROP table is executed. 
when DROP table is executed and not committed, 
1. the class record in Root class is deleted by RVHF_DELETE log
2. In order to get class information, read transaction in PTS get classOID from classname cache, if status is not LC_CLASSNAME_DELETED
3. class record (indicated by classOID) does not exist and assertion fails in `locator_is_exist_class_name_entry`

Implementation
1. change the status when RVHF_DELETE on record in root class
2. revert the status when DROP statement is rolled back (RVHF_COMPENSATE_WITH_INSERT)

### DIFF
--- a/src/storage/heap_file.c
+++ b/src/storage/heap_file.c
@@ -15817,37 +15817,6 @@ heap_rv_redo_insert (THREAD_ENTRY * thread_p, const LOG_RCV * rcv)
       recdes.area_size = recdes.length = *(INT16 *) recdes.data;
       recdes.data = NULL;
     }
-  else
-    {
-      /* Classname cache (locator_Mht_classnames) status should be reverted when
-       * DROP TABLE statement is rolled back (RVHF_COMPENSATE_WITH_INSERT).
-       * The cache status is modified at heap_rv_redo_delete() (RVHF_DELETE).
-       * When CREATE TABLE is executed, recdes.type is always REC_ASSIGN_ADDRESS,
-       * and we need to filter out the CREATE TABLE case.
-       */
-
-      if (is_passive_transaction_server ())
-	{
-	  assert (thread_p->type == TT_REPLICATION_PTS);
-
-	  const bool is_deleted = spage_is_record_deleted (rcv->pgptr, slotid);
-	  if (is_deleted)
-	    {
-	      OID classoid;
-	      heap_get_class_oid_from_page (thread_p, rcv->pgptr, &classoid);
-	      if (OID_EQ (&classoid, oid_Root_class_oid))
-		{
-		  VPID vpid;
-		  OID oid;
-
-		  (void) pgbuf_get_vpid (rcv->pgptr, &vpid);
-		  SET_OID (&oid, vpid.volid, vpid.pageid, slotid);
-
-		  locator_restore_classname_entry_if_exists (thread_p, &oid);
-		}
-	    }
-	}
-    }
 
   sp_success = spage_insert_for_recovery (thread_p, rcv->pgptr, slotid, &recdes);
   pgbuf_set_dirty (thread_p, rcv->pgptr, DONT_FREE);
@@ -15861,6 +15830,34 @@ heap_rv_redo_insert (THREAD_ENTRY * thread_p, const LOG_RCV * rcv)
 	}
       assert (er_errid () != NO_ERROR);
       return er_errid ();
+    }
+
+  if (is_passive_transaction_server ())
+    {
+      if (recdes.type != REC_ASSIGN_ADDRESS)
+	{
+	  /* Classname cache (locator_Mht_classnames) status should be reverted when
+	   * DROP TABLE statement is rolled back (RVHF_COMPENSATE_WITH_INSERT).
+	   * The cache status is modified at heap_rv_redo_delete() (RVHF_DELETE).
+	   * When CREATE TABLE is executed, recdes.type is always REC_ASSIGN_ADDRESS,
+	   * and we need to filter out the CREATE TABLE case.
+	   */
+
+	  assert (thread_p->type == TT_REPLICATION_PTS);
+
+	  OID classoid;
+	  heap_get_class_oid_from_page (thread_p, rcv->pgptr, &classoid);
+	  if (OID_EQ (&classoid, oid_Root_class_oid))
+	    {
+	      VPID vpid;
+	      OID oid;
+
+	      (void) pgbuf_get_vpid (rcv->pgptr, &vpid);
+	      SET_OID (&oid, vpid.volid, vpid.pageid, slotid);
+
+	      locator_change_classname_entry_status_if_exists (thread_p, &oid, LC_CLASSNAME_EXIST);
+	    }
+	}
     }
 
   return NO_ERROR;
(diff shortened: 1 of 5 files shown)


src/storage/heap_file.c
```diff
@@ -15817,6 +15817,37 @@ heap_rv_redo_insert (THREAD_ENTRY * thread_p, const LOG_RCV * rcv)
...
+		  VPID vpid;
+		  OID oid;
+
+		  (void) pgbuf_get_vpid (rcv->pgptr, &vpid);
+		  SET_OID (&oid, vpid.volid, vpid.pageid, slotid);
+
+		  locator_restore_classname_entry_if_exists (thread_p, &oid);
+		}
+	    }
+	}
```
hornetmj: If this logic is executed after spage_insert_for_recovery(), I expect to be able to combine locator_mark_classname_entry_deleted() and locator_restore_classname_entry_if_exists(). For example, locator_change_classname_entry_status(LC_CLASSNAME_DELETED or LC_CLASSNAME_EXIST, ...).

src/storage/slotted_page.c
```diff
@@ -4011,6 +4011,27 @@ spage_get_record_type (PAGE_PTR page_p, PGSLOTID slot_id)
   return slot_p->record_type;
 }
 
+bool
+spage_is_record_deleted (PAGE_PTR page_p, PGSLOTID slot_id)
```
hornetmj: Nit:

Too general function name. However, it cannot use in the reuse-oid table.

hornetmj (approved):

joohok:
I will handle the exception and request a review again.

Exception to be handled:
RENAME TABLE tbl to tbl3;
DROP TABLE tbl3;

Proceeding with DROP while the cache of the renamed class is still not updated.

src/storage/slotted_page.c
joohok: this function is removed

src/storage/heap_file.c
joohok: I apply your suggestion

src/storage/heap_file.c
```diff
@@ -15832,6 +15832,34 @@ heap_rv_redo_insert (THREAD_ENTRY * thread_p, const LOG_RCV * rcv)
       return er_errid ();
     }
 
+  if (is_passive_transaction_server ())
+    {
+      if (recdes.type != REC_ASSIGN_ADDRESS)
+	{
+	  /* Classname cache (locator_Mht_classnames) status should be reverted when
+	   * DROP TABLE statement is rolled back (RVHF_COMPENSATE_WITH_INSERT).
```
hornetmj, line 15840: heap_rv_redo_insert function is also used in replication(redo) of RVHF_INSERT and RVHF_INSERT_NEWHOME. I think they don't need to update the status of classname entry.

src/storage/heap_file.c
```diff
@@ -15832,6 +15832,34 @@ heap_rv_redo_insert (THREAD_ENTRY * thread_p, const LOG_RCV * rcv)
       return er_errid ();
     }
 
+  if (is_passive_transaction_server ())
+    {
+      if (recdes.type != REC_ASSIGN_ADDRESS)
+	{
+	  /* Classname cache (locator_Mht_classnames) status should be reverted when
```
hornetmj, line 15839: It would be good to me if the comment also explain why the status of classname entry should be changed.

src/transaction/locator_sr.c
```diff
@@ -13937,6 +13937,76 @@ has_errors_filtered_for_insert (std::vector<int> error_filter_array)
...
+	   * DROP TABLE tbl2;           -- looking for "tbl2" in the cache, but not found.
+	   *                               But we need to change the status of "tbl" entry
+	   *                               because READ transaction in PTS could access DELETED "tbl" object
+	   *                               using cached entry which causes failure.
+	   * COMMIT;
+	   */
+
+	  HENTRY_PTR hentry = NULL;
+	  HENTRY_PTR next = NULL;
+	  for (hentry = locator_Mht_classnames->act_head; hentry != NULL; hentry = next)
```
hornetmj, line 13987: This may seem like a rather expensive operation. If you agree, please leave a TODO comment for improvements.

src/transaction/locator_sr.c
```diff
@@ -13937,6 +13937,76 @@ has_errors_filtered_for_insert (std::vector<int> error_filter_array)
...
+	   * DROP TABLE tbl;            -- looking for "tbl" in the locator_Mht_classnames above, but not found.
+	   *                               We don't need to change the status of "tbl" entry,
+	   *                               it will be changed at LOG_COMMIT time
+	   * COMMIT;
+	   *
+	   * 2) RENAME
+	   * RENAME TABLE tbl to tbl2;  -- "tbl2" is not cached, but only "tbl" is cached
+	   * DROP TABLE tbl2;           -- looking for "tbl2" in the cache, but not found.
+	   *                               But we need to change the status of "tbl" entry
+	   *                               because READ transaction in PTS could access DELETED "tbl" object
```
hornetmj, line 14001: If someone don't have background knowledge, it may be difficult to understand. Even though the replicator is holding SCH_M_LOCK, it seems to say that the read transaction is accessible.

src/storage/heap_file.c
joohok: I added a condition to update the cache only for operations related to records of the root class. Operations on the root class only occur when there is a CREATE TABLE or when a DROP TABLE is rolled back

src/storage/heap_file.c
joohok: Stating that it reverts due to rollback seemed sufficient. Are you interested in a description of what happens if a revert is not performed?

src/transaction/locator_sr.c
joohok: OK, I think it needs more description. 

The classname cache can be accessed regardless of whether a lock is acquired. In a read transaction, access to the cache is made to obtain the class OID corresponding to the classname. If the cache status for that class is LC_CLASSNAME_EXIST, a check is performed on the actual heap to see if a record exists using the class OID. However, since it has already been deleted, access is not possible, leading to a core dump at this point.

src/storage/heap_file.c
hornetmj: I think if you add a description why the cache status should be changed in front of the locator_change_classname_entry_status_if_exists definition, then you can cover all of my request about comment modification. I expect you to describe it based on the DROP statement and the your reply as below: 

> The classname cache can be accessed regardless of whether a lock is acquired. In a read transaction, access to the cache is made to obtain the class OID corresponding to the classname. If the cache status for that class is LC_CLASSNAME_EXIST, a check is performed on the actual heap to see if a record exists using the class OID

src/storage/heap_file.c
hornetmj: Do you mean that 'if (OID_EQ (&classoid, oid_Root_class_oid))' can filter out all the cases of RVHF_INSERT and RVHF_INSERT_NEWHOME?

src/storage/heap_file.c
joohok: OK, I will

src/storage/heap_file.c
joohok: While I initially thought there would be no REC_NEWHOME record insertion for root classes, as they are not MVCC tables, it seems that we should handle REC_NEWHOME as an exception. This is because the record size can change due to ALTER statements.

There are likely two scenarios where new records might be inserted into root classes:
1. REC_ASSIGN_ADDRESS: This appears when CREATE TABLE statements.
2. REC_NEWHOME: This appears when the record size changes during ALTER TABLE.

For other record types, such as REC_HOME, REC_BIGONE, REC_RELOCATION, we can see those types during the execution of the UNDO operation for DROP statements. It's not certain whether a root class's record can become REC_BIGONE.

PR #4777, opened 2023-11-10
[LETS-794]  Integrate work pools into one in page_server

wrlawodms:
http://jira.cubrid.org/browse/LETS-794

- Share the `m_worker_pool` with server request responders in the `page_server`.
- There are two server request responders in the `page_server` and the number of threads by `PRM_ID_SCAL_PERF_PS_REQ_RESPONDER_THREAD_COUNT` is for both of them. 
  - It can get separate later if needed. Probably, the maximum of tran_server for the one for tran servers and the maximum of followers for another for followers.
- It's out of scope to define a `entry_task` for extra async jobs in page_server like the catch-up.
- Adapt a unit test.

### DIFF
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -1304,3 +1304,15 @@ page_server::finalize_request_responder ()
   m_tran_server_responder.reset (nullptr);
   m_follower_responder.reset (nullptr);
 }
+
+const size_t
+page_server::get_internal_thread_count ()
+{
+  // thread pool used by page server to perform parallel replication
+  const int replication_parallel_thread_count = prm_get_integer_value (PRM_ID_REPLICATION_PARALLEL_COUNT);
+
+  // Refer to page_server::page_server () to see threads
+  const size_t worker_count = m_worker_pool->get_max_count ();
+
+  return replication_parallel_thread_count + worker_count;
+}
--- a/src/server/page_server.cpp
+++ b/src/server/page_server.cpp
@@ -45,7 +45,7 @@ page_server::page_server (const char *db_name)
   size_t thread_count = 4; // Arbitrary chosen
   size_t task_max_count = thread_count * 4;
 
-  // two async responder: m_tran_server_responder, m_follower_responder
+  // For two async responders: m_tran_server_responder, m_follower_responder
   const size_t responder_thread_cnt = (size_t) prm_get_integer_value (PRM_ID_SCAL_PERF_PS_REQ_RESPONDER_THREAD_COUNT);
   const size_t responder_max_task_cnt = (size_t) prm_get_integer_value (PRM_ID_SCAL_PERF_PS_REQ_RESPONDER_TASK_COUNT);
 
@@ -67,6 +67,9 @@ page_server::~page_server ()
   assert (m_passive_tran_server_conn.size () == 0);
 
   m_async_disconnect_handler.terminate ();
+
+  assert (m_tran_server_responder == nullptr);
+  assert (m_follower_responder == nullptr);
 
   cubthread::get_manager ()->destroy_worker_pool (m_worker_pool);
   assert (m_worker_pool == nullptr);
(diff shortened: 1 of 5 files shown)


hornetmj (approved):
LGTM.

unit_tests/server/test_server_request_responder_main.cpp
```diff
@@ -115,22 +115,33 @@ struct test_env
...
+  cubthread::entry_workpool *m_worker_pool;
+  std::unique_ptr<server_request_responder<test_conn>> m_rrh;
 
   test_env ()
   {
+    const size_t responder_thread_cnt = (size_t) prm_get_integer_value (PRM_ID_SCAL_PERF_PS_REQ_RESPONDER_THREAD_COUNT);
+    const size_t responder_max_task_cnt = (size_t) prm_get_integer_value (PRM_ID_SCAL_PERF_PS_REQ_RESPONDER_TASK_COUNT);
+
+    m_worker_pool = cubthread::get_manager ()->create_worker_pool (responder_thread_cnt, responder_max_task_cnt,
+		    "test_async_responder_workers", NULL, 1, false);
```
joohok, line 127: Q. Is it okay not to consider 'm_worker_context_manager' in unit tests? 
The implementation includes it, but I'm asking because it's not present in the unit tests

joohok (approved):

unit_tests/server/test_server_request_responder_main.cpp
wrlawodms: I don't think it's needed. The context manager determines the `tran_index` and thread type of entries (workers). It may affect the work each cubrid thread does, but does not affect the `server_request_responder` itself. This unit test is for `server_request_responder`.

Rudeus (approved):

src/server/page_server.hpp
```diff
@@ -323,7 +323,15 @@ class page_server
     std::future<void> m_follower_disc_future;
     std::mutex m_follower_disc_mutex;
 
-    cubthread::entry_workpool *m_worker_pool; // a worker_pool to take some background jobs that needs a thread entry
+    /*
+     * A worker_pool to take some asynchronous jobs that need a thread entry.
+     * It's used by two server_request_responders for tran servers and follower servers,
+     * and some extra jobs like the catch-up.
+     *
+     * Note that server_request_responders have to be destroyed earlier since they ensure all jobs are done in their dtors.
```
cristiarg: This comment makes half sense to me. Suggestion:
```suggestion
     * Note that both tran_server_responder and follower_responder hold references to the worker pool.
     * Thus, they must be destroyed prior to the worker pool - to avoid dangling references.
     * The responders ensure that all jobs are waited for completion in their dtors.
```

src/server/page_server.cpp
```diff
@@ -40,12 +40,20 @@
...
+  // For two async responders: m_tran_server_responder, m_follower_responder
+  const size_t responder_thread_cnt = (size_t) prm_get_integer_value (PRM_ID_SCAL_PERF_PS_REQ_RESPONDER_THREAD_COUNT);
+  const size_t responder_max_task_cnt = (size_t) prm_get_integer_value (PRM_ID_SCAL_PERF_PS_REQ_RESPONDER_TASK_COUNT);
+
+  thread_count += responder_thread_cnt;
+  task_max_count +=  responder_max_task_cnt;
 
   m_worker_pool = cubthread::get_manager ()->create_worker_pool (thread_count, task_max_count, "page_server_workers",
-		  nullptr, 1, false);
+		  &m_worker_context_manager, 1, false);
```
cristiarg, line 56: This is a pointer in page_server. Then passed as a referece via the ctor. Then passed here further as a pointer.
Maybe it should be a pointer everywhere.

cristiarg (approved):

src/server/page_server.cpp
wrlawodms: OK. I've changed it to a const pointer.

PR #4778, opened 2023-11-13
[CBRD-24758] Remove useless condition in if statement in function heap_stats_sync_bestspace

hyahong:
http://jira.cubrid.org/browse/CBRD-24758

Purpose
```
static int
heap_stats_sync_bestspace (...)
{
        int min_freespace;

        ...
        min_freespace = heap_stats_get_min_freespace (heap_hdr);

        ...
        if (free_space >= min_freespace && free_space > HEAP_DROP_FREE_SPACE)
```
the function, heap_stats_get_min_freespace returns a integer value less than or equal to HEAP_DROP_FREE_SPACE.
therefore, the condition `free_space >= min_freespace` is not needed since `free_space > HEAP_DROP_FREE_SPACE` contains the condition `free_space >= min_freespace`.

Implementation
N/A

Remarks
N/A

### DIFF
--- a/src/storage/heap_file.c
+++ b/src/storage/heap_file.c
@@ -3903,7 +3903,8 @@ heap_stats_sync_bestspace (THREAD_ENTRY * thread_p, const HFID * hfid, HEAP_HDR_
 
 	  free_space = spage_max_space_for_new_record (thread_p, pg_watcher.pgptr);
 
-	  if (free_space > HEAP_DROP_FREE_SPACE)
+    /* TODO: if the value returned by heap_stats_get_min_freespace (...) changes, this condition should be checked. */
+	  if (/* free_space >= heap_stats_get_min_freespace (heap_hdr) && */ free_space > HEAP_DROP_FREE_SPACE)
 	    {
 	      if (prm_get_integer_value (PRM_ID_HF_MAX_BESTSPACE_ENTRIES) > 0)
 		{
--- a/src/storage/heap_file.c
+++ b/src/storage/heap_file.c
@@ -3903,8 +3903,8 @@ heap_stats_sync_bestspace (THREAD_ENTRY * thread_p, const HFID * hfid, HEAP_HDR_
 
 	  free_space = spage_max_space_for_new_record (thread_p, pg_watcher.pgptr);
 
-    /* TODO: if the value returned by heap_stats_get_min_freespace (...) changes, this condition should be checked. */
-	  if (/* free_space >= heap_stats_get_min_freespace (heap_hdr) && */ free_space > HEAP_DROP_FREE_SPACE)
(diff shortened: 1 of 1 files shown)


swi0110:
@hyahong 
Please refer to this.
Currently, the CI's test env has a little problem.
So all CI tests fail.
I will share it when resolve the problem.

swi0110:
Fixed a problem.
and I re-run the CI test.

hornetmj (approved):

hyahong:
@swi0110 I checked your comment. Thank you for the information!

src/storage/heap_file.c
```diff
@@ -3906,7 +3903,7 @@ heap_stats_sync_bestspace (THREAD_ENTRY * thread_p, const HFID * hfid, HEAP_HDR_
 
 	  free_space = spage_max_space_for_new_record (thread_p, pg_watcher.pgptr);
 
-	  if (free_space >= min_freespace && free_space > HEAP_DROP_FREE_SPACE)
```
wrlawodms, line 3909: This has a point in that it's open to the change of `heap_stats_get_min_freespace`. If it's taken into consideration, then I'll follow.

wrlawodms (approved):

src/storage/heap_file.c
hyahong: I think it's just function which returns the minimum space. so, I don't have a good idea how to change `heap_stats_get_min_freespace`.
If you have any good ideas, could you share them?

wrlawodms: I didn't mean that we should make a change to the `heap_stats_get_min_freespace`. What I meant is the condition is not "useless" and has a point.

The `heap_stats_sync_bestspace` doesn't know how the minimum space for statistics is calculated by the `heap_stats_get_min_freespace` function. It just considers both the minimum space and `HEAP_DROP_FREE_SPACE`. With your change, it doesn't consider "the minimum space". 
Looking into the function, it can't be larger than `HEAP_DROP_FREE_SPACE` for now, but the upper function doesn't need to know and should not know this. In the existing code, if the way we calculate the minimum space is changed, the `heap_stats_sync_bestspace` takes this into consideration, but it will not with your change. That's what I meant by `it's open to the change of heap_stats_get_min_freespace`.

wrlawodms: To justify the condition as "useless", I think we should ensure that "the minimum space for statistics" doesn't need to be taken into consideration in the view of the `heap_stats_sync_bestspace`. It might be true, but I don't know.

hyahong: I've certainly overlooked that aspect! I'll comment as opinions are being organized, but it might take a little while to check.

hyahong: @wrlawodms 
I have been asked to quickly address these heap refactoring issues. I will revisit these issues following the analysis of the heap manager.

PR #4781, opened 2023-11-13
[CBRD-25062] Fix the assert condition in function heap_scan_pb_lock_and_fetch

hyahong:
http://jira.cubrid.org/browse/CBRD-25062

Purpose
```
if (scan_cache->page_latch == NULL_LOCK)
{
	page_lock = NULL_LOCK;
}
else
{
	assert (scan_cache->page_latch >= NULL_LOCK);
	assert (lock >= NULL_LOCK);
	page_lock = lock_Conv[scan_cache->page_latch][lock];
	assert (page_lock != NA_LOCK);
}
```
assert (scan_cache->page_latch >= NULL_LOCK) should not contains a condition about `scan_cache->page_latch == NULL_LOCK`

and, `lock` can be only S_LOCK or X_LOCK since all the callers of this function gives S_LOCK or X_LOCK as parameter `lock`.  so assert (lock >= NULL_LOCK) can be changed assert (lock == S_LOCK || lock == X_LOCK) and moved to beginning of this function.

Implementation
N/A

Remarks
N/A

### DIFF
--- a/src/storage/heap_file.c
+++ b/src/storage/heap_file.c
@@ -1154,7 +1154,7 @@ heap_stats_del_bestspace_by_vpid (THREAD_ENTRY * thread_p, VPID * vpid)
   (void) heap_stats_entry_free (thread_p, ent, NULL);
   ent = NULL;
 
-  heap_Bestspace->num_stats_entries -= 1;
+  heap_Bestspace->num_stats_entries--;
 
 end:
   assert (mht_count (heap_Bestspace->vpid_ht) == mht_count (heap_Bestspace->hfid_ht));
@@ -1233,6 +1233,8 @@ heap_scan_pb_lock_and_fetch_debug (THREAD_ENTRY * thread_p, const VPID * vpid_pt
   LOCK page_lock;
   PGBUF_LATCH_MODE page_latch_mode;
 
+  assert (lock == S_LOCK || lock == X_LOCK);
+
   if (scan_cache != NULL)
     {
       if (scan_cache->page_latch == NULL_LOCK)
@@ -1241,8 +1243,7 @@ heap_scan_pb_lock_and_fetch_debug (THREAD_ENTRY * thread_p, const VPID * vpid_pt
 	}
       else
 	{
-	  assert (scan_cache->page_latch >= NULL_LOCK);
-	  assert (lock >= NULL_LOCK);
+	  assert (scan_cache->page_latch > NULL_LOCK);
 	  page_lock = lock_Conv[scan_cache->page_latch][lock];
 	  assert (page_lock != NA_LOCK);
 	}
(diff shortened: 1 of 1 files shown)


hornetmj (approved):

src/storage/heap_file.c
```diff
@@ -1233,6 +1233,8 @@ heap_scan_pb_lock_and_fetch_debug (THREAD_ENTRY * thread_p, const VPID * vpid_pt
   LOCK page_lock;
   PGBUF_LATCH_MODE page_latch_mode;
 
+  assert (lock == S_LOCK || lock == X_LOCK);
```
wrlawodms, line 1237: >  lock can be only S_LOCK or X_LOCK since all the callers of this function gives S_LOCK or X_LOCK as parameter lock.

This pre-condition could be true, but I don't think this can be the reason. If others are not used doesn't mean that it's not allowed. It could be that there is just no use case. We need another proper reason to go with this.

hornetmj: @wrlawodms That's a good comment. However, verification of your request is beyond the scope of this issue. So, we have two options. One is to not edit the part you mentioned, and the other is to leave a comment with relevant information in the edited part. I think it would be a good idea to leave information through the second option(track it as a comment) and then check whether only S_LOCK and X_LOCK are allowed in later.

wrlawodms: Do we need to change the pre-condition check even without verification? It's trivial, but I don't see any benefit of this change.

wrlawodms: I mean, is this worth it though it even needs a comment saying "we're not sure" for this?

hornetmj: This issue is not one that accomplishes any meaningful work. The purpose is to reduce minor errors by removing unnecessary code and removing unnecessary conditions. Comments will be written in TODO: format and will simply confirm whether LOCKs other than S_LOCK and X_LOCK are allowed.

wrlawodms: >  The purpose is to reduce minor errors by removing unnecessary code and removing unnecessary conditions

I think it's meaningful, but we don't know if the assertion is unnecessary.
If you don't want to debate, you can go ignoring my comment. :)  If you want to keep the debate, I still don't know how this is a work for the purpose you said.

hornetmj: I always feel our discussion is worth. However this part is too minor and we have to use our time more productive way. So, I choose my opinion and leave a comment for this to track the issue that this function is only allowed by S_LOCK and X_LOCK. Thanks : )

wrlawodms: Yes, that's what I mean:)

hornetmj (approved):

PR #4789, opened 2023-11-16
[CBRD-25082] If the leaf nodes are compressed, store the common prefix in BTREE_NODE_HEADER

youngjinj:
http://jira.cubrid.org/browse/CBRD-25082

### In TO-BE, the common prefix is stored in BTREE_NODE_HEADER.
- We are calculating the common prefix every time we scan the index. The common prefix does not change unless the node is split or merged, so calculating it every time is inefficient. To calculate the common prefix, we need to check whether both the lower fence key and the upper fence key exist in the leaf node and compare the two keys to check how many columns there are in common.
- The purpose is to skip the common prefix when performing a binary search in compressed leaf nodes.
- In TO-BE, the common prefix is stored in BTREE_NODE_HEADER. This way, we can know the common prefix right away without doing inefficient calculations.

### The two functions below are separated by purpose.
```
btree_node_common_prefix, btree_node_is_compressed
  -> btree_node_calculate_common_prefix, btree_node_get_common_prefix
```

### DIFF
--- a/src/storage/btree.c
+++ b/src/storage/btree.c
@@ -12723,15 +12723,18 @@ btree_node_get_common_prefix (THREAD_ENTRY * thread_p, BTID_INT * btid, PAGE_PTR
     {
       is_compressed = false;
     }
-#endif /* !NDEBUG */
 
   if (header->common_prefix > 0 && header->common_prefix < btid->key_type->precision)
     {
       assert (is_compressed);
+    }
+#endif /* !NDEBUG */
+
+  if (header->common_prefix > 0 && header->common_prefix < btid->key_type->precision)
+    {
       return header->common_prefix;
     }
 
-  assert (!is_compressed);
   return 0;
 }
 
--- a/src/storage/btree.c
+++ b/src/storage/btree.c
@@ -12739,7 +12739,7 @@ btree_node_get_common_prefix (THREAD_ENTRY * thread_p, BTID_INT * btid, PAGE_PTR
  * btree_node_calculate_common_prefix () - Checks whether the given page can be compressed
  *                                         and returns the common prefix if it can.
  *
- *   return: Common prefix. If 0, the page was not compressed.
+ *   return: Common prefix. If 0, compression is not possible.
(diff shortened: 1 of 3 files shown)


xmilex-git (approved):
The function `btree_node_is_compressed()` has been renamed to `btree_node_get_common_prefix()`. How about adding a comment that if the return value of `btree_node_get_common_prefix()` is 0, it means that the node is not compressed? I suggest this considering future development where it may be necessary to check whether a node is compressed or not.

youngjinj:
> The function `btree_node_is_compressed()` has been renamed to `btree_node_get_common_prefix()`. How about adding a comment that if the return value of `btree_node_get_common_prefix()` is 0, it means that the node is not compressed? I suggest this considering future development where it may be necessary to check whether a node is compressed or not.

Thank you. I will add comments to help other developers understand.

shparkcubrid (approved):

src/storage/btree.c
```diff
@@ -12689,57 +12769,92 @@ btree_node_common_prefix (THREAD_ENTRY * thread_p, BTID_INT * btid, PAGE_PTR pag
...
+  if (header == NULL)
+    {
+      assert (false);
+      goto not_compressed;
+    }
+
+  if (header->node_level > 1)
+    {
+      /* Compression is only possible at leaf nodes. */
+      goto not_compressed;
```
hornetmj, line 12794: Nit:

The code in this part is identical to the part of btree_node_get_common_prefix(). How about writing it as a separate function and using it?

hornetmj (approved):

src/storage/btree.c
youngjinj: Thank you.
As you suggested, I thought about whether it would be better to make this common code into separate functions.

I left these codes so that there would be no difference with AS-IS.
I wanted to minimize the code that needed verification in this PR.

I think you should have checked this code before calling this function.
If I had to change anything further here, I think it would be to make sure that these codes are checked before this function is called.
I don't think there's a need to make it a separate function.

PR #4794, opened 2023-11-20
[CBRD-24914] Core dumped in pt_ordbynum_to_key_limit_multiple_ranges at src/parser/xasl_generation.c:23489

xmilex-git:
http://jira.cubrid.org/browse/CBRD-24914

When using `MRO optimization`, we fixed an issue where a condition created by rewrite that is `always true/false` was recognized as a data filter and asserted. 

- The following are the revised contents: 

1. We modified the `qo_reduce_equality_terms()` function to remove expressions found to be `always true` before the function ends. 

3. In `qo_check_terms_for_multiple_range_opt()`, if a term is identified as `PT_VALUE`, it is treated as `always false` since the `always true` terms have been removed, thus preventing `MRO optimization`.


src/optimizer/query_rewrite.c
```diff
@@ -1844,6 +1851,89 @@ qo_reduce_equality_terms (PARSER_CONTEXT * parser, PT_NODE * node, PT_NODE ** wh
...
+		    {
+		      cut_off = true;
+		    }
+		}
+	    }
+	}
+      else
+	{
+	  dnf_prev = NULL;
+	  while ((dnf_node = ((dnf_prev) ? dnf_prev->or_next : expr)))
```
youngjinj: Is there a scenario to test when the optimizer is trying to MRO when DNF has an always true condition?

xmilex-git: I attempted to use the following scenario as a test scenario for the DNF node, but failed.
```sql
select 
    /*+ recompile ordered */ t.c2 
from 
    t0 inner join t1 as t on t.c2 = 5 or t0.c1 = t.c1 and t0.c1 = 1 
where 
    t.c1 = 1 order by t.c2 limit 1;
```

In this case, the following query statement is written:

```sql
select 
    /*+ ORDERED */ t.c2 
from 
    t0 t0 inner join t1 t on ((t.c2= ?:2 ) or (t0.c1= ?:3 )) and ((t.c2= ?:0 ) or (t0.c1= ?:1 )) 
where 
    t.c1= ?:4 order by 1 for orderby_num()<= ?:5
```
Despite mentioning that `t0.c1` is a constant, I am encountering difficulties as the terms inside `((t.c2= ?:2 ) or (t0.c1= ?:3 ))` are not being changed.

shparkcubrid: How about excluding DNF from deletion? I don’t think it’s a target of the reduce_equality_term.

xmilex-git: @shparkcubrid I think it's good idea. I'll modify and commit.

xmilex-git: This example helps.
```sql
select /*+ recompile ordered */ t.c2
from t0 inner join t1 as t on (t.c2 = 1 or t0.c1 = t.c1)
where t.c1 = 1 and t0.c1 = 1
order by t.c2
limit 1;
```

src/optimizer/query_rewrite.c
```diff
@@ -1844,6 +1851,63 @@ qo_reduce_equality_terms (PARSER_CONTEXT * parser, PT_NODE * node, PT_NODE ** wh
...
+      if (expr->or_next == NULL)
+	{
+	  if (opd1 && opd2 && op == PT_EQ && opd1->node_type == PT_VALUE && opd2->node_type == PT_VALUE)
+	    {
+	      domain = pt_xasl_node_to_domain (parser, expr);
+	      qualifier = expr->info.expr.qualifier;
+	      dbv1 = pt_value_to_db (parser, opd1);
+	      dbv2 = pt_value_to_db (parser, opd2);
+	      opd3 = expr->info.expr.arg3;
+	      dbv3 = pt_value_to_db (parser, opd3);
```
youngjinj: PT_EQ does not use arg3.
So, I think it is okay to pass dbv3 and opd3 as NULL here.
```
pt_evaluate_db_value_expr (parser, expr, expr->info.expr.op, dbv1, dbv2, NULL, &dbval_res, domain, opd1, opd2, NULL, qualifier)
```

If necessary for the future, how about changing it to the following?
- We have a problem calling the pt_value_to_db function without checking `opd3->node_type == PT_VALUE`.
```suggestion
	      dbv3 = (opd3 != NULL && opd3->node_type == PT_VALUE) ? pt_value_to_db (parser, opd3) : NULL;
```

src/optimizer/query_rewrite.c
```diff
@@ -1844,6 +1851,63 @@ qo_reduce_equality_terms (PARSER_CONTEXT * parser, PT_NODE * node, PT_NODE ** wh
...
+	  else
+	    {
+	      *orgp = expr->next;
+	    }
+	  expr->next = NULL;
+	  parser_free_tree (parser, expr);
+	}
+      else
+	{
+	  expr_prev = (expr_prev) ? expr_prev->next : expr;
```
youngjinj: Why should we check expr_prev?
What problems will occur if we do the following?
```suggestion
	  expr_prev = expr;
```

xmilex-git: The initial value of expr_prev is set to NULL. In this scenario, expr is pointing to the first element in the list. It seems likely that an error will occur when first entering the while loop. This implementation method was referenced from the where_type function.

src/optimizer/plan_generation.c
```diff
@@ -3975,7 +3975,22 @@ qo_check_terms_for_multiple_range_opt (QO_PLAN * plan, int first_sort_col_idx, b
...
+	  /*
+	   * We decided not to support MRO (Multiple Row Optimization) if the return value
+	   * of the bitset_iterate() function is checked and found to be -1. 
+	   * This is related to the CBRD-24914 issue, which involves a core dump occurring
+	   * when term->pt_expr is converted to PT_VALUE resulting in an always true/false condition.
+	   * Therefore, in the process of reducing the equality term, some parts of the 
+	   * always true condition have been removed. In cases of an always false condition, 
+	   * or an always true condition that has not been removed, this implementation does not support MRO optimization.
+	   */
+	  return NO_ERROR;
```
youngjinj:
```suggestion
	  return NO_ERROR;	/* give up */
```

src/optimizer/query_rewrite.c
```diff
@@ -1844,6 +1850,61 @@ qo_reduce_equality_terms (PARSER_CONTEXT * parser, PT_NODE * node, PT_NODE ** wh
...
+	      dbv1 = pt_value_to_db (parser, opd1);
+	      dbv2 = pt_value_to_db (parser, opd2);
+	      if (pt_evaluate_db_value_expr
+		  (parser, expr, expr->info.expr.op, dbv1, dbv2, NULL, &dbval_res, domain, opd1, opd2, NULL, qualifier))
+		{
+		  if (DB_VALUE_TYPE (&dbval_res) == DB_TYPE_INTEGER && db_get_int (&dbval_res) == 1)
+		    {
+		      cut_off = true;
+		    }
+		}
```
youngjinj: This is only for PT_EQ, but I think calling the pt_evaluate_db_value_expr function is excessive.
```suggestion
	      dbv1 = pt_value_to_db (parser, opd1);
	      dbv2 = pt_value_to_db (parser, opd2);
	      if (db_value_compare (dbv1, dbv2) == DB_EQ)
		{
		  cut_off = true;
		}
```

src/optimizer/query_rewrite.c
```diff
@@ -1360,6 +1360,12 @@ qo_reduce_equality_terms (PARSER_CONTEXT * parser, PT_NODE * node, PT_NODE ** wh
   PT_NODE *save_where_next;
   bool copy_arg2;
   PT_NODE *dt1, *dt2;
+  bool cut_off;
+  PT_NODE *expr_prev = NULL;
+  PT_NODE *opd1, *opd2;
+  DB_VALUE *dbv1, *dbv2, dbval_res;
+  PT_MISC_TYPE qualifier = (PT_MISC_TYPE) 0;
+  TP_DOMAIN *domain;
```
youngjinj: There are variables that can be removed.
```suggestion
  DB_VALUE *dbv1, *dbv2;
```

src/optimizer/plan_generation.c
```diff
@@ -3975,7 +3975,22 @@ qo_check_terms_for_multiple_range_opt (QO_PLAN * plan, int first_sort_col_idx, b
 	  return NO_ERROR;
 	}
 
-      for (s = bitset_iterate (&(termp->segments), &iter_s); s != -1; s = bitset_next_member (&iter_s))
+      s = bitset_iterate (&(termp->segments), &iter_s);
```
shparkcubrid: How about 'if (bitset_is_empty (&(termp->segments))' ?

xmilex-git: I think it's good idea, I'll modify it and commit.

shparkcubrid (approved):

youngjinj (approved):

src/optimizer/plan_generation.c
```diff
@@ -3975,6 +3975,21 @@ qo_check_terms_for_multiple_range_opt (QO_PLAN * plan, int first_sort_col_idx, b
 	  return NO_ERROR;
 	}
 
+
+      if (bitset_is_empty (&(termp->segments)))
+	{
+	  /*
+	   * We decided not to support MRO (Multiple Row Optimization) if the return value
+	   * of the bitset_iterate() function is checked and found to be -1.
```
youngjinj: Please sync comments and code.

PR #4795, opened 2023-11-21
[LETS-806] Fix the core when the RVPGBUF_DEALLOC log record is replicated to PTS

joohok:
http://jira.cubrid.org/browse/LETS-806

Purpose

When applying RVPGBUF_DEALLOC in the PTS replicator, the redo function already invalidates and unfixes the corresponding page. 
 However, because the replicator is unaware of whether the page has been unfixed, it attempts to unfix it again. The goal is to resolve the crash occurring during this process.
If the page has already been unfixed, it is necessary to invalidate `rcv.pgptr` used by the replicator.

### DIFF
--- a/src/transaction/log_recovery_redo.hpp
+++ b/src/transaction/log_recovery_redo.hpp
@@ -16,8 +16,8 @@
  *
  */
 
-#ifndef _LOC_RECOVERY_REDO_HPP_
-#define _LOC_RECOVERY_REDO_HPP_
+#ifndef _LOG_RECOVERY_REDO_HPP_
+#define _LOG_RECOVERY_REDO_HPP_
 
 #include "log_compress.h"
 #include "log_lsa.hpp"
@@ -703,4 +703,4 @@ log_rv_redo_rec_info<T>::log_rv_redo_rec_info (const log_lsa lsa, LOG_RECTYPE ty
 
 }
 
-#endif // _LOC_RECOVERY_REDO_HPP_
+#endif // _LOG_RECOVERY_REDO_HPP_
--- a/src/transaction/log_recovery_redo.hpp
+++ b/src/transaction/log_recovery_redo.hpp
@@ -645,14 +645,14 @@ void log_rv_redo_record_sync_apply (THREAD_ENTRY *thread_p, log_rv_redo_context 
 			 VPID_AS_ARGS (&rcv_vpid), (int) rcv.offset, (int) rcv.length);
     }
 
-  if (is_passive_transaction_server() && log_data.rcvindex == RVPGBUF_DEALLOC)
+  if (log_data.rcvindex == RVPGBUF_DEALLOC && is_passive_transaction_server())
     {
       /* TODO:
        * redofunc for RVPGBUF_DEALLOC (pgbuf_rv_dealloc_redo ()) invalidates the page buffer in PTS,
        * but it does not set page lsa (pgbuf_set_lsa ()) before invalidation.
        * We need to set page lsa when redoing RVPGBUF_DEALLOC log, as what ATS does (see the caller of LOG_NEED_TO_SET_LSA ()).
        */
-      assert (rcv.pgptr != nullptr && pgbuf_get_fix_count (rcv.pgptr) <= 0);
+      assert (rcv.pgptr != nullptr && pgbuf_get_fix_count (rcv.pgptr) == 0);
       rcv.pgptr = nullptr;
     }
 
--- a/src/transaction/log_recovery_redo.hpp
+++ b/src/transaction/log_recovery_redo.hpp
@@ -647,7 +647,11 @@ void log_rv_redo_record_sync_apply (THREAD_ENTRY *thread_p, log_rv_redo_context 
 
   if (is_passive_transaction_server() && log_data.rcvindex == RVPGBUF_DEALLOC)
     {
-      // rcv pgptr is already unfixed and invalidated by the deallocation (pgbuf_rv_dealloc_redo ())
+      /* TODO:
+       * redofunc for RVPGBUF_DEALLOC (pgbuf_rv_dealloc_redo ()) invalidates the page buffer in PTS,
+       * but it does not set page lsa (pgbuf_set_lsa ()) before invalidation.
+       * We need to set page lsa when redoing RVPGBUF_DEALLOC log, as what ATS does (see the caller of LOG_NEED_TO_SET_LSA ()).
+       */
       assert (rcv.pgptr != nullptr && pgbuf_get_fix_count (rcv.pgptr) <= 0);
       rcv.pgptr = nullptr;
     }
--- a/src/transaction/log_recovery_redo.hpp
+++ b/src/transaction/log_recovery_redo.hpp
@@ -647,7 +647,7 @@ void log_rv_redo_record_sync_apply (THREAD_ENTRY *thread_p, log_rv_redo_context 
 
   if (is_passive_transaction_server() && log_data.rcvindex == RVPGBUF_DEALLOC)
     {
-      // rcv.pgptr is invalidated and unfixed by pgbuf_rv_dealloc_redo ()
+      // rcv pgptr is already unfixed and invalidated by the deallocation (pgbuf_rv_dealloc_redo ())
       rcv.pgptr = nullptr;
     }
 
--- a/src/transaction/log_recovery_redo.hpp
+++ b/src/transaction/log_recovery_redo.hpp
@@ -26,6 +26,7 @@
 #include "log_recovery.h"
 #include "page_buffer.h"
 #include "perf_monitor_trackers.hpp"
+#include "server_type.hpp"
 #include "scope_exit.hpp"
 #include "system_parameter.h"
 #include "type_helper.hpp"
@@ -642,6 +643,12 @@ void log_rv_redo_record_sync_apply (THREAD_ENTRY *thread_p, log_rv_redo_context 
 			 "rcv = {mvccid=%llu, vpid=(%d, %d), offset = %d, data_length = %d}",
 			 LSA_AS_ARGS (&record_info.m_start_lsa), (long long int) rcv.mvcc_id,
 			 VPID_AS_ARGS (&rcv_vpid), (int) rcv.offset, (int) rcv.length);
+    }
+
+  if (is_passive_transaction_server() && log_data.rcvindex == RVPGBUF_DEALLOC)
+    {
+      // rcv.pgptr is invalidated and unfixed by pgbuf_rv_dealloc_redo ()
+      rcv.pgptr = nullptr;
     }
 
   if (rcv.pgptr != nullptr)
--- a/src/transaction/log_recovery_redo.hpp
+++ b/src/transaction/log_recovery_redo.hpp
@@ -648,6 +648,7 @@ void log_rv_redo_record_sync_apply (THREAD_ENTRY *thread_p, log_rv_redo_context 
   if (is_passive_transaction_server() && log_data.rcvindex == RVPGBUF_DEALLOC)
     {
       // rcv pgptr is already unfixed and invalidated by the deallocation (pgbuf_rv_dealloc_redo ())
+      assert (rcv.pgptr != nullptr && pgbuf_get_fix_count (rcv.pgptr) <= 0);
       rcv.pgptr = nullptr;
     }
 
--- a/unit_tests/log_repl/test_main_log_repl_atomic_helper.cpp
+++ b/unit_tests/log_repl/test_main_log_repl_atomic_helper.cpp
@@ -1118,11 +1118,13 @@ log_rv_check_redo_is_needed (const PAGE_PTR & /*pgptr*/, const LOG_LSA & /*rcv_l
 bool
 is_passive_transaction_server()
 {
+  assert_release (false);
   return false;
 }
 
 int
 pgbuf_get_fix_count (PAGE_PTR /*pgptr*/)
 {
+  assert_release (false);
   return 0;
 }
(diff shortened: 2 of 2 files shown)


src/transaction/log_recovery_redo.hpp
```diff
@@ -644,6 +645,13 @@ void log_rv_redo_record_sync_apply (THREAD_ENTRY *thread_p, log_rv_redo_context
 			 VPID_AS_ARGS (&rcv_vpid), (int) rcv.offset, (int) rcv.length);
     }
 
+  if (is_passive_transaction_server() && log_data.rcvindex == RVPGBUF_DEALLOC)
```
hornetmj: How about process this in pgbuf_rv_dealloc_redo()?

hornetmj (approved):

src/transaction/log_recovery_redo.hpp
joohok: In pgbuf_rv_dealloc_redo (), we can not modify the rcv.pgptr value because the function receives that value as a constant. 
And the format of the recovery functions is predetermined, so it cannot be changed. This is because it is a member of the `rvfun`

src/transaction/log_recovery_redo.hpp
```diff
@@ -644,6 +645,13 @@ void log_rv_redo_record_sync_apply (THREAD_ENTRY *thread_p, log_rv_redo_context
...
+    {
+      // rcv pgptr is already unfixed and invalidated by the deallocation (pgbuf_rv_dealloc_redo ())
+      assert (rcv.pgptr != nullptr && pgbuf_get_fix_count (rcv.pgptr) <= 0);
+      rcv.pgptr = nullptr;
```
wrlawodms, line 656: This prevents a caller, `log_rv_redo_record_sync`, from unfixing it again, but another call path for atomic sequence seems to still unfix it.

(1) atomic_log_sequence::apply_and_unfix
(2) -> atomic_log_entry::apply_log_redo()
(3) -> atomic_log_entry::apply_log_by_type<LOG_REC_UNDOREDO>
(4) -> log_rv_redo_record_sync_apply()

In (3), this `rcv` is just a local variable, and the UNFIX procedure is done with another pgptr variable in (1)->page_ptr_bookkeeping.unfix_page. The nullified rcv.pgptr here doesn't affect the pgptr variable which is used to unfix in `page_ptr_bookkeeping.unfix_page`.

wrlawodms: I presume an atomic sequence can include any log record even though now there is no case, or it's not been spotted.

joohok: As far as I remember, there didn't seem to be any issues with the RVPGBUF_DEALLOC in the atomic sequence. Upon rechecking, it appears that in `page_ptr_bookkeeping.unfix_page` -> `pgbuf_unfix_or_ordered_unfix()`, if the recovery index is `RVPGBUF_DEALLOC`, it skips the unfix. This function is specifically implemented to unfix pages in the atomic sequence.

The issue arose from the fact that `RVPGBUF_DEALLOC` logs, which are not part of the atomic sequence, were causing problems when replicated to the PTS

wrlawodms: Oh, I missed it. You're right.

src/transaction/log_recovery_redo.hpp
```diff
@@ -644,6 +645,13 @@ void log_rv_redo_record_sync_apply (THREAD_ENTRY *thread_p, log_rv_redo_context
...
+    }
+
   if (rcv.pgptr != nullptr)
     {
       pgbuf_set_lsa (thread_p, rcv.pgptr, &record_info.m_start_lsa);
```
wrlawodms, line 661: This is out of the scope of this issue, but I think we should advance the page lsa before unfixing the page to redo it correctly. If it's the case, it needs an issue for it.

wrlawodms: With a TODO comment.

joohok: I was looking for reasons to set the page LSA in PTS. However, not only are the reasons important, but it also seems odd that when the same page has the same LSA applied, ATS, PS, and PTS have different Page LSAs. I think it's inappropriate to set the page LSA for the same page_ptr after invalidating the page. According to your opinion, I think it's necessary to leave comments and create an issue for resolution.

@cristiarg  What is your opinion?

joohok: I added http://jira.cubrid.org/browse/LETS-879

wrlawodms (approved):
Approved, but this PR needs @cristiarg's review.

Rudeus (approved):

src/transaction/log_recovery_redo.hpp
```diff
@@ -644,6 +645,17 @@ void log_rv_redo_record_sync_apply (THREAD_ENTRY *thread_p, log_rv_redo_context
 			 VPID_AS_ARGS (&rcv_vpid), (int) rcv.offset, (int) rcv.length);
     }
 
+  if (is_passive_transaction_server() && log_data.rcvindex == RVPGBUF_DEALLOC)
+    {
+      /* TODO:
+       * redofunc for RVPGBUF_DEALLOC (pgbuf_rv_dealloc_redo ()) invalidates the page buffer in PTS,
+       * but it does not set page lsa (pgbuf_set_lsa ()) before invalidation.
```
cristiarg, line 652: The sequence of lines in `pgbuf_rv_dealloc_redo`:

```
pgbuf_rv_dealloc_redo (THREAD_ENTRY * thread_p, const LOG_RCV * rcv)
{
  ...
#if defined(SERVER_MODE)
  if (is_passive_transaction_server ())
    {
      ...
      const int error_code = pgbuf_invalidate (thread_p, rcv->pgptr);
      ...
    }
#endif
   ...
}
```

has been added as part of fixes for http://jira.cubrid.org/browse/LETS-558 and http://jira.cubrid.org/browse/LETS-511 .
See PR #3960 .

Can we revisit that fix with the knowledge of this situation?

> The issue arose from the fact that RVPGBUF_DEALLOC logs, which are not part of the atomic sequence, were causing problems when replicated to the PTS

That fix was aimed to ensure that a page was not mistakenly reused by the PTS while having already gone through the de-allocate - re-allocate sequence.

If there is another way to solve this de-allocate - re-allocate sequence without going through the `pgbuf_invalidate` I think we should revisit the fix done in `pgbuf_rv_dealloc_redo`.

Does a call to `pgbuf_set_lsa` solve the problem?
I think **not** because the set LSA would still be smaller than the required LSA needed upon a subsequent access on PTS.

By looking at `pgbuf_dealloc_page`, it also goes a long way to invalidate the page and make sure that if it is reused, it is correctly reused (ie: re-allocated).
Specifically, I am looking at the comment:
```
  /* how it works: page is "deallocated" by resetting its type to PAGE_UNKNOWN. also prepare bcb for victimization.
   *
   * note: the bcb used to be invalidated. but that means flushing page to disk and waiting for IO write. that may be
   *       too slow. if we add the bcb to the bottom of a lru list, it will be eventually flushed by flush thread and
   *       victimized. */
```

Maybe `pgbuf_rv_dealloc_redo` can do the same on PTS. Instead of just calling `pgbuf_invalidate`.
Can we extract a chunk out of `pgbuf_dealloc_page` (ie: the part after the call to `PGBUF_BCB_LOCK`) and reuse it in `pgbuf_rv_dealloc_redo`?

The reason I ask this is because there are already a lot of "exceptions".
In particular I would want to avoid `if (is_passive_transaction_server()` as much as possible.
With this fix it looks like fixing a workaround (that in #3960 ) with yet another workaround.

src/transaction/log_recovery_redo.hpp
```diff
@@ -644,6 +645,17 @@ void log_rv_redo_record_sync_apply (THREAD_ENTRY *thread_p, log_rv_redo_context
...
+       * We need to set page lsa when redoing RVPGBUF_DEALLOC log, as what ATS does (see the caller of LOG_NEED_TO_SET_LSA ()).
+       */
+      assert (rcv.pgptr != nullptr && pgbuf_get_fix_count (rcv.pgptr) <= 0);
```
cristiarg: Why `pgbuf_get_fix_count (rcv.pgptr) <= 0` and not `pgbuf_get_fix_count (rcv.pgptr) == 0` ?

src/transaction/log_recovery_redo.hpp
```diff
@@ -644,6 +645,17 @@ void log_rv_redo_record_sync_apply (THREAD_ENTRY *thread_p, log_rv_redo_context
 			 VPID_AS_ARGS (&rcv_vpid), (int) rcv.offset, (int) rcv.length);
     }
 
+  if (is_passive_transaction_server() && log_data.rcvindex == RVPGBUF_DEALLOC)
```
cristiarg: I think 
```suggestion
  if (log_data.rcvindex == RVPGBUF_DEALLOC && is_passive_transaction_server())
```
is much more likely to make use of short-circuit evaluation.

src/transaction/log_recovery_redo.hpp
cristiarg: Aside from uniform handling, and considering that - on PTS - the de-allocated page is not reused, is there another reason to update the LSA?

unit_tests/log_repl/test_main_log_repl_atomic_helper.cpp
```diff
@@ -1114,3 +1114,15 @@ log_rv_check_redo_is_needed (const PAGE_PTR & /*pgptr*/, const LOG_LSA & /*rcv_l
   assert_release (false);
   return false;
 }
+
+bool
+is_passive_transaction_server()
+{
+  return false;
```
cristiarg, line 1122: I think
```suggestion
  assert (false);
  return false;
```

same below ...

src/transaction/log_recovery_redo.hpp
joohok: I get code snippet from `pgbuf_is_valid_page_ptr ()`, but I think `pgbuf_get_fix_count (rcv.pgptr) == 0` is correct because it set fix count to 0 in `pgbuf_invalidate ()`

unit_tests/log_repl/test_main_log_repl_atomic_helper.cpp
joohok: OK,  I will add assertion

src/transaction/log_recovery_redo.hpp
joohok: It seems most sensible to revisit the issue and make overall revisions. I agree with your opinion. Continuously inserting exception code makes it challenging to maintain the code context.

It would be neat if we could extract a portion of the code for use in `pgbuf_rv_dealloc`. 
However, I'm currently fixing bugs to run tests for DDL replication. Revisiting the issue, analyzing, implementing, and testing might become a substantial task in this situation.

Creating a refactoring issue, detailing the implementation intent, and specifying the codes to be removed might be a good approach I think. We can then address it in a PTS replicator refactoring phase.

What do you think?

src/transaction/log_recovery_redo.hpp
joohok: To be honest, I'm not quite sure about other reasons. It doesn't address the deallocate-reallocate situation, and after the corresponding page buffer is invalidated, when PTS read the same page, it obtains page from PS, so it seems the page LSA won't be used in PTS. However, I aimed to eliminate potential errors by making it behave similarly to ATS and PS.

wrlawodms: >  it obtains page from PS, so it seems the page LSA won't be used in PTS

When determining the LSA for fetching a page (`pgbuffer_pool.highest_evicted_lsa`), the last evicted page's stuff is used. The PTS expects the requested PS to apply redo log records up to that point. Let's imagine the last evicted page is the deallocated page by the replicator in `pgvbuf_rv_dealloc_redo`. It makes a change to the page, but the newly fetched page may not have the modification since the corresponding page_lsa is not updated. 

Can't it happen?

joohok: > When determining the LSA for fetching a page (pgbuffer_pool.highest_evicted_lsa), the last evicted page's stuff is used

As far as I know, PTS uses `highest_process_lsa` instead of `highest_evicted_lsa`. 
Therefore, when the PTS requests a page, the PS is expected to return a page that includes the changes up to `RVPGBUF_DEALLOC`

code snippet : 
```
in pgbuf_read_page_from_file_or_page_server ()
...
 8206       const LOG_LSA target_repl_lsa = is_passive_transaction_server ()?
 8207         get_passive_tran_server_ptr ()->get_highest_processed_lsa () : pgbuf_Pool.get_highest_evicted_lsa ();
...
```

wrlawodms: Oh, you're correct. I was confused.

unit_tests/log_repl/test_main_log_repl_atomic_helper.cpp
cristiarg: Also make sure that the unit test passes.

src/transaction/log_recovery_redo.hpp
cristiarg: I think that the problem on PTS is a little more complicated - as described in http://jira.cubrid.org/browse/LETS-523 .

I agree that it is out of scope for this issue. 

I pinged-back from http://jira.cubrid.org/browse/LETS-523 to these related issues - both this one and LETS-558 - for a proper analysis when the time comes.

Thank you.

cristiarg (approved):

src/transaction/log_recovery_redo.hpp
joohok: Thank you, I will add more description in the issue

PR #4798, opened 2023-11-21
[CBRD-25089] Add 'LEADING' sql hint

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-25089

The LEADING hint instructs the optimizer to use the specified set of tables as the prefix in the execution plan. This hint is more versatile than the ORDERED hint.
The LEADING hint is ignored if the tables specified cannot be joined first in the order specified because of dependencies in the join graph.
If you specify two or more LEADING hints, then only the first one is activated and all of them except the first are ignored.
If you specify the ORDERED hint, all LEADING hint is ignored.


src/parser/parse_tree_cl.c
```diff
@@ -14080,10 +14086,16 @@ pt_print_select (PARSER_CONTEXT * parser, PT_NODE * p)
 	  if (p->info.query.q.select.hint & PT_HINT_ORDERED)
 	    {
 	      /* force join left-to-right */
-	      q = pt_append_nulstring (parser, q, "ORDERED");
-	      if (p->info.query.q.select.ordered)
+	      q = pt_append_nulstring (parser, q, "ORDERED ");
+	    }
+
+	  if (p->info.query.q.select.hint & PT_HINT_LEADING)
```
youngjinj: If PT_HINT_ORDERED and PT_HINT_LEADING are printed together, it is difficult to know which hint was used.
```suggestion
	  else if (p->info.query.q.select.hint & PT_HINT_LEADING)
```
Please see the example.
```
drop table if exists t1, t2, t3, t4;

create table t1 (c1 int);
create table t2 (c1 int);
create table t3 (c1 int, c2 int);
create table t4 (c1 int, c2 int);

insert into t1 values (1);
insert into t2 values (1);
insert into t3 values (1, 2);
insert into t4 values (1, 2);

select /*+ recompile ordered leading(t4,t1) */ count (*) from t1, t2, t3, t4 where t1.c1 = t2.c1 and t2.c1 = t3.c1 and t3.c2 = t4.c2;

/*
...
Query stmt:

select /*+ ORDERED LEADING(t4, t1) */ count(*) from t1 t1, t2 t2, t3 t3, t4 t4 where t1.c1=t2.c1 and t2.c1=t3.c1 and t3.c2=t4.c2
*/
```

src/parser/parse_tree_cl.c
```diff
@@ -14080,10 +14086,16 @@ pt_print_select (PARSER_CONTEXT * parser, PT_NODE * p)
...
+	    {
+	      /* force join left-to-right */
+	      q = pt_append_nulstring (parser, q, "LEADING");
```
youngjinj, line 14094: If `p->info.query.q.select.leading` is NULL, I think it is better not to output `/*+ LEADING */` either.
Or would it be better to change it to PT_HINT_ORDERED internally?

src/parser/parser_support.c
```diff
@@ -8094,7 +8094,7 @@ pt_split_delete_stmt (PARSER_CONTEXT * parser, PT_NODE * delete_stmt)
     }
 
   /* if we have hints that refers globally to the join statement then we skip the split */
-  if ((delete_stmt->info.delete_.hint & PT_HINT_ORDERED && delete_stmt->info.delete_.ordered_hint == NULL)
+  if ((delete_stmt->info.delete_.hint & PT_HINT_ORDERED)
```
youngjinj, line 8097: I wonder if the results below are as intended.
```
drop table if exists t1, t2, t3, t4;

create table t1 (c1 int);
create table t2 (c1 int);
create table t3 (c1 int, c2 int);
create table t4 (c1 int, c2 int);
```
```
delete /*+ recompile leading(t1, t2) */     t1, t2, t4 from t1, t2, t3, t4 where t1.c1 = t2.c1 and t2.c1 = t3.c1;
  -> select /*+ LEADING(t1, t2) */ t2,class t2,t1,class t1 from t1 t1, t2 t2, t3 t3 where t1.c1=t2.c1 and t2.c1=t3.c1
  -> select /*+ LEADING */ t4,class t4 from t4 t4

delete /*+ recompile leading(t1, t2, t4) */ t1, t2, t4 from t1, t2, t3, t4 where t1.c1 = t2.c1 and t2.c1 = t3.c1;
  -> [Warning: HINT ignored]
     select t4,class t4,t2,class t2,t1,class t1 from t1 t1, t2 t2, t3 t3, t4 t4 where t1.c1=t2.c1 and t2.c1=t3.c1

delete /*+ recompile ordered */             t1, t2, t4 from t1, t2, t3, t4 where t1.c1 = t2.c1 and t2.c1 = t3.c1;
  -> select /*+ ORDERED */ t4,class t4,t2,class t2,t1,class t1 from t1 t1, t2 t2, t3 t3, t4 t4 where t1.c1=t2.c1 and t2.c1=t3.c1
```

src/parser/parse_tree_cl.c
shparkcubrid: I agree. I'll fix it.

src/parser/parse_tree_cl.c
shparkcubrid: I agree. I'll fix it.

src/parser/parser_support.c
youngjinj: I think it would be better to ignore t4 mentioned in the hint in the second query and run it like the first query.

shparkcubrid: I think it is correct to ignore the impossible join order. 
The t4 is classified into a different partition and a plan is created separately because it does not have a join term.
There seems to be no problem with adjusting join order.

youngjinj: Since the hint was ignored, I think it is correct that a separate query should be executed for t4 in this case as well.
- Current PR
```
delete /*+ recompile leading(t1, t2, t4) */ t1, t2, t4 from t1, t2, t3, t4 where t1.c1 = t2.c1 and t2.c1 = t3.c1;
  -> [Warning: HINT ignored]
     select t4,class t4,t2,class t2,t1,class t1 from t1 t1, t2 t2, t3 t3, t4 t4 where t1.c1=t2.c1 and t2.c1=t3.c1
```
- My opinion
```
delete /*+ recompile leading(t1, t2, t4) */ t1, t2, t4 from t1, t2, t3, t4 where t1.c1 = t2.c1 and t2.c1 = t3.c1;
  -> [Warning: HINT ignored]
     select t2,class t2,t1,class t1 from t1 t1, t2 t2, t3 t3 where t1.c1=t2.c1 and t2.c1=t3.c1
  -> [Warning: HINT ignored]
     select t4,class t4 from t4 t4
```
I will approve this PR because it is not an important issue.

youngjinj (approved):

xmilex-git (approved):

src/parser/scanner_support.c
```diff
@@ -418,6 +412,27 @@ pt_get_hint (const char *text, PT_HINT hint_table[], PT_NODE * node)
...
+	      node->info.delete_.leading_hint = hint_table[i].arg_list;
+	      hint_table[i].arg_list = NULL;
+	    }
+	  else if (node->node_type == PT_UPDATE)
+	    {
+	      node->info.update.hint = (PT_HINT_ENUM) (node->info.update.hint | hint_table[i].hint);
+	      node->info.update.leading_hint = hint_table[i].arg_list;
+	      hint_table[i].arg_list = NULL;
+	    }
+	  break;
```
beyondykk9, line 434: no necessary for PT_MERGE?

beyondykk9 (approved):

src/parser/scanner_support.c
shparkcubrid: PT_MERGE will be added if requested in the future.

PR #4806, opened 2023-11-23
[CBRD-25074] The symbol in the LIMIT clause is not bypassed and an error is returned

hyunikn:
http://jira.cubrid.org/browse/CBRD-25074

- add the identifier case to the limit_factor non-terminal of the grammar in order to allow a PL/CSQL host variable in the LIMIT clause.
- for the identifier case, return PT_HOST_VAR node if it is parsing PL/CSQL Static SQL, otherwise set an error
- for the parser to decide it is parsing a PL/CSQL Static SQL or not, a parameter named 'flags' is added to db_open_buffer(), and db_open_buffer_local() functions
  - currently, the flag is set to a non-zero value only when db_open_buffer() is called from method_query_handler.c, 
  - and the only flag is PARSER_FOR_PLCSQL_STATIC_SQL.

UPDATED (2023-11-28):
- The new flags parameter of db_open_buffer() and db_open_buffer_local() was deleted and instead a global variable g_open_buffer_control_flags was added and is used for the same purpose.


hgryoo:
semantic checks should be considered.
If a column name is specified as follow, it should return proper error.
```
select a from t1 ORDER BY a LIMIT a;

csql: ../src/parser/parser_support.c:9524: int pt_get_query_limit_from_limit(PARSER_CONTEXT*, PT_NODE*, DB_VALUE*): Assertion `limit->node_type == PT_VALUE || limit->node_type == PT_HOST_VAR || limit->node_type == PT_EXPR' failed.
```

src/parser/csql_grammar.y
```diff
@@ -15769,7 +15769,13 @@ limit_factor
                         PARSER_SAVE_ERR_CONTEXT ($$, @$.buffer_pos)
 
                 DBG_PRINT}}
+        | identifier    /* for PL/CSQL Static SQL only */
```
ctshim: Is '.' allowed in variable names in PL/CSQL?
Otherwise, use identifier_without_dot

Also, you should consider that identifiers such as `aaa` and [aaa] are also allowed.

hyunikn:
> semantic checks should be considered
Good point. Thank you.

src/parser/csql_grammar.y
hyunikn: > Otherwise, use identifier_without_dot
Good point. Thank you.

> Also, you should consider that identifiers such as `aaa` and [aaa] are also allowed.
Identifiers can be one of the forms "...", [...], `...` as in CUBRID SQL.

hgryoo:
I suggest adding the flags in `DB_SESSION`. Because the Query parser performs in cubridcs, it is running as single thread. maybe we can set the flags by checking the `DB_SESSION` is created in query_handler::prepare(), which is called in SP's query API module.

hyunikn:
> I suggest adding the flags in `DB_SESSION`. Because the Query parser performs in cubridcs, it is running as single thread. maybe we can set the flags by checking the `DB_SESSION` is created in query_handler::prepare(), which is called in SP's query API module.

Currently the flag is used only by parser. Do you mean a future use of it?

kisoo-han (approved):

hgryoo (approved):

ctshim (approved):

beyondykk9 (approved):

airnet73 (approved):

PR #4807, opened 2023-11-24
[CBRD-25079] occurs to core when if set 'export CUBRID_MSG_LANG="" ' before cubrid server start

beyondykk9:
http://jira.cubrid.org/browse/CBRD-25079

Setting er_set before the server thread has yet started cannot properly handle errors, and leaving them in the log is not easy, so it would be best to exclude the er_set part.


hyunikn (approved):

src/base/language_support.c
```diff
@@ -1099,8 +1099,7 @@ set_msg_lang_from_env (void)
       if (status != NO_ERROR)
 	{
 	  sprintf (err_msg, "invalid value '%s' for CUBRID_MSG_LANG", lang_Msg_loc_name);
-	  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_LOC_INIT, 1, err_msg);
-	  strcpy (lang_Msg_loc_name, LANG_NAME_DEFAULT);
+	  strcpy (lang_Msg_loc_name, LANG_NAME_ENGLISH "." LANG_CHARSET_UTF8);
```
hyunikn, line 1102: What about printing the error message to the console before the thread is ready.

hgryoo:
Without er_set(), the error message is not written. If this situation happens because of typo or any reason, It would be hard for users to know how to fix.
Also, there is no way for users to know the databases sets `LANG_NAME_ENGLISH "." LANG_CHARSET_UTF8` internally. A warning message should be shown if we go to the spec of PR's change.

src/base/language_support.c
beyondykk9: I tried printing a message to the console as your comment, but the server booting ended abnormally afterwards.

ctshim (approved):

hgryoo (approved):

PR #4809, opened 2023-11-26
[CBRD-25098] Problem where inst_num() is inappropriately changed to order_by_num() when a subquery containing an ORDER BY is view merged.

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-25098

fix the case below so that it does not walk into the subquery. This prevents unnecessary exchanges.
```
inst_num() -> order_by_num()
order_by_num() -> inst_num()
order_by_num() -> group_by_num()
```

### DIFF
--- a/src/parser/parse_tree_cl.c
+++ b/src/parser/parse_tree_cl.c
@@ -138,6 +138,8 @@ static void strcat_with_realloc (PT_STRING_BLOCK * sb, const char *tail);
 static PT_NODE *pt_lambda_check_reduce_eq (PARSER_CONTEXT * parser, PT_NODE * tree_or_name, void *void_arg,
 					   int *continue_walk);
 static PT_NODE *pt_lambda_node (PARSER_CONTEXT * parser, PT_NODE * tree_or_name, void *void_arg, int *continue_walk);
+static PT_NODE *pt_lambda_node_pre (PARSER_CONTEXT * parser, PT_NODE * tree_or_name, void *void_arg,
+				    int *continue_walk);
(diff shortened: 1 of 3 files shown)


src/parser/parse_tree_cl.c
```diff
@@ -571,6 +573,34 @@ pt_lambda_check_reduce_eq (PARSER_CONTEXT * parser, PT_NODE * tree_or_name, void
...
+ */
+static PT_NODE *
+pt_lambda_node_pre (PARSER_CONTEXT * parser, PT_NODE * tree_or_name, void *void_arg, int *continue_walk)
+{
+  if (!tree_or_name)
+    {
+      return tree_or_name;
+    }
+
+  if (PT_IS_QUERY_NODE_TYPE (tree_or_name->node_type))
```
xmilex-git, line 592: Is there a reason for creating a new function instead of implementing a case for `PT_IS_QUERY_NODE_TYPE` within the `pt_lambda_check_reduce_eq()` function?

The corresponding case is as follows : 

```
#define PT_IS_QUERY_NODE_TYPE(x) \
    (  (x) == PT_SELECT     || (x) == PT_UNION \
    || (x) == PT_DIFFERENCE || (x) == PT_INTERSECTION)
```

xmilex-git (approved):

src/parser/parse_tree_cl.c
shparkcubrid: pt_lambda_with_arg() is added in cases where the pt_lambda_check_reduce_eq() is not needed.

beyondykk9 (approved):

youngjinj (approved):

PR #4811, opened 2023-11-28
[LETS-873] Temporary solution for PTS read transactions to wait when accessing a DROPping table

joohok:
http://jira.cubrid.org/browse/LETS-873

Purpose
`locator_is_exist_class_name_entry()` is called during PTS read transactions checking if a table exists in both the cache and heap. The assertion failure occurs in ATS when the cache state is LC_CLASSNAME_EXIST but the table is not in the heap. In the context of PTS, since the cache state doesn't guarantee the presence of a heap record, this assertion is unnecessary. However, this is a temporary workaround, and in the future, when implementing logic to update the cache state, this change will be reverted

### DIFF
--- a/src/transaction/locator_sr.c
+++ b/src/transaction/locator_sr.c
@@ -12606,7 +12606,8 @@ locator_is_exist_class_name_entry (THREAD_ENTRY * thread_p, LOCATOR_CLASSNAME_EN
       assert (entry->e_tran_index == NULL_TRAN_INDEX);
 
       assert (!OID_ISNULL (&entry->e_current.oid));
-      assert (heap_does_exist (thread_p, oid_Root_class_oid, &entry->e_current.oid));
+      assert (is_passive_transaction_server ()
+	      || heap_does_exist (thread_p, oid_Root_class_oid, &entry->e_current.oid));
 
       assert (LSA_ISNULL (&entry->e_current.savep_lsa));
       assert (entry->e_current.prev == NULL);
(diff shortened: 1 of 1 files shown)


hornetmj (approved):

wrlawodms (approved):

Rudeus (approved):

src/transaction/locator_sr.c
```diff
@@ -12606,7 +12606,17 @@ locator_is_exist_class_name_entry (THREAD_ENTRY * thread_p, LOCATOR_CLASSNAME_EN
       assert (entry->e_tran_index == NULL_TRAN_INDEX);
 
       assert (!OID_ISNULL (&entry->e_current.oid));
-      assert (heap_does_exist (thread_p, oid_Root_class_oid, &entry->e_current.oid));
+      /* TODO:
```
cristiarg, line 12609: @joohok 
Please allow me to ask: how do we plan to track these 'TODO' scaterred around the code?

I have to admit that I also added a bunch of them, but they are mostly in new code files.

Is it reasonable to ask to track isolated TODO's in legacy code - such as this one by explicit Jira issues? At least to have them all in Jira and someone/sometime will see them there :)

Thanks.

PS. My comment is triggered by this remark in the comment:
> However, since the underlying problem stems from not properly updating the state of the cache, this assertion should be reverted when incorporating logic to update the cache status in PTS.

joohok: This PR is a temporary fix as we couldn't determine the design direction for LETS-814.
LETS-814 will soon address this TODO. However, the TODO is not explicitly mentioned in the LETS-814 issue. 
I should document it explicitly in LETS-814 for now.

I agree with your suggestion to create and track an issue for any TODOs that arise. Especially if TODOs in legacy code are not tracked, it can pose risks

Thank you

PR #4813, opened 2023-11-28
[CBRD-24921] Minor refactoring: Split initializing definitions to separate file

hgryoo:

- Split defining query spec from schema_system_catalog_install.cpp to schema_system_catalog_install_query_spec.cpp
- Add a new common header, schema_system_catalog_install.hpp
- clear catalog information after building catalogs are succed.
  - catcls_init (): Init clist and vclist according to the catalog's definition
  - catcls_build_class(), catcls_build_vclass(): build system catalogs and then clear defintion structrues.


hyunikn (approved):

src/object/schema_system_catalog_install.cpp
```diff
@@ -452,25 +500,25 @@ namespace cubschema
...
   {
 
-    static system_catalog_definition sm_define_method (
-	    // name
-	    CT_METHOD_NAME,
-	    // columns
+    return system_catalog_definition (
+		   // name
+		   CT_METHOD_NAME,
+		   // columns
```
ctshim, line 515: It would be a good idea to adjust the indent for the first argument of system_catalog_definition(). It's indented too much.
t would look better if you aligned it with the comment line " // constraints " at the bottom.

ctshim (approved):

beyondykk9 (approved):

PR #4814, opened 2023-11-28
[LETS-810] Acquire a lock for serial operation in PTS

joohok:
http://jira.cubrid.org/browse/LETS-810

Purpose
Made the log record type and log data, as well as the locking functions on the PTS side, more generic, as it requires handling locks for the SERIAL

- Changed LOG_SCHEMA_MODIFICATION_LOCK to LOG_REPL_DDL_LOCK_INFO.
  - Previously, only the OID (classoid) of the locked object was recorded in the log, but this issue, it record the oid, classoid, and lock modes all together
- Leave the LOG_REPL_DDL_LOCK_INFO when the record in DB_SERIAL is fetched or inserted
  - `xlocator_fectch ()`, `heap_get_insert_location_with_lock ()`
- X_LOCK will be acquired on the record in DB_SERIAL

### DIFF
--- a/src/storage/heap_file.c
+++ b/src/storage/heap_file.c
@@ -20668,7 +20668,7 @@ heap_get_insert_location_with_lock (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONT
 	      assert (!OID_ISNULL (&context->res_oid));
 	      assert (is_active_transaction_server ());
 
-	      log_append_locked_object (thread_p, &context->class_oid, &context->res_oid, lock);
+	      log_append_repl_ddl_lock_info (thread_p, &context->class_oid, &context->res_oid, lock);
 	    }
 #endif
 
--- a/src/storage/heap_file.c
+++ b/src/storage/heap_file.c
@@ -20657,8 +20657,9 @@ heap_get_insert_location_with_lock (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONT
 	{
 	  /* successfully locked! */
 #if defined (SERVER_MODE)
-	  if ((lock == SCH_M_LOCK && OID_IS_ROOTOID (&context->class_oid))
-	      || (lock == X_LOCK && OID_EQ (&context->class_oid, oid_Serial_class_oid)))
+	  const bool is_lock_for_class = OID_IS_ROOTOID (&context->class_oid) && lock == SCH_M_LOCK;
+	  const bool is_lock_for_serial = OID_EQ (&context->class_oid, oid_Serial_class_oid) && lock == X_LOCK;
+	  if (is_lock_for_class || is_lock_for_serial)
 	    {
 	      assert (!OID_ISNULL (&context->res_oid));
 	      assert (is_active_transaction_server ());
--- a/src/storage/heap_file.c
+++ b/src/storage/heap_file.c
@@ -20662,7 +20662,7 @@ heap_get_insert_location_with_lock (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONT
 	   * when a serial is created (class_oid == oid_Serial_class_oid).
 	   * The creation of a serial implies the insertion of a record into db_serial*/
 	  const bool is_lock_for_class = OID_IS_ROOTOID (&context->class_oid) && lock == SCH_M_LOCK;
-	  const bool is_lock_for_serial = OID_EQ (&context->class_oid, oid_Serial_class_oid) && lock == X_LOCK;
+	  const bool is_lock_for_serial = oid_is_serial (&context->class_oid) && lock == X_LOCK;
 	  if (is_lock_for_class || is_lock_for_serial)
 	    {
 	      assert (!OID_ISNULL (&context->res_oid));
--- a/src/storage/heap_file.c
+++ b/src/storage/heap_file.c
@@ -20657,12 +20657,13 @@ heap_get_insert_location_with_lock (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONT
 	{
 	  /* successfully locked! */
 #if defined (SERVER_MODE)
-	  if (lock == SCH_M_LOCK && OID_IS_ROOTOID (&context->class_oid))
+	  if ((lock == SCH_M_LOCK && OID_IS_ROOTOID (&context->class_oid))
+	      || (lock == X_LOCK && OID_EQ (&context->class_oid, oid_Serial_class_oid)))
 	    {
 	      assert (!OID_ISNULL (&context->res_oid));
 	      assert (is_active_transaction_server ());
 
-	      log_append_schema_modification_lock (thread_p, &context->res_oid);
+	      log_append_locked_object (thread_p, &context->class_oid, &context->res_oid, lock);
 	    }
 #endif
 
--- a/src/storage/heap_file.c
+++ b/src/storage/heap_file.c
@@ -20657,6 +20657,10 @@ heap_get_insert_location_with_lock (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONT
 	{
 	  /* successfully locked! */
 #if defined (SERVER_MODE)
+	  /* lock information is logged to replicate DDL operations on PTS.
+	   * Logging here occurs when a class is created (class_oid == oid_Root_class_oid) or
+	   * when a serial is created (class_oid == oid_Serial_class_oid).
+	   * The creation of a serial implies the insertion of a record into db_serial*/
 	  const bool is_lock_for_class = OID_IS_ROOTOID (&context->class_oid) && lock == SCH_M_LOCK;
 	  const bool is_lock_for_serial = OID_EQ (&context->class_oid, oid_Serial_class_oid) && lock == X_LOCK;
 	  if (is_lock_for_class || is_lock_for_serial)
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -3738,6 +3738,8 @@ void
 log_append_locked_object (THREAD_ENTRY * thread_p, const OID * classoid, const OID * oid, const LOCK lock)
 {
   assert (!OID_ISNULL (classoid));
+  assert (!OID_IS_ROOTOID (classoid) || lock == SCH_M_LOCK);	// if classoid is rootoid, lock must be SCH_M_LOCK
+  assert (!OID_EQ (classoid, oid_Serial_class_oid) || lock == X_LOCK);	// if classoid is serial class oid, lock must be X_LOCK
 
   int tran_index = LOG_FIND_THREAD_TRAN_INDEX (thread_p);
   LOG_TDES *tdes = LOG_FIND_TDES (tran_index);
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -3758,7 +3758,7 @@ log_append_locked_object (THREAD_ENTRY * thread_p, const OID * classoid, const O
       return;
     }
 
-  auto record = (LOG_REC_LOCKED_OBJECT *) node->data_header;
+  auto *const record = (LOG_REC_LOCKED_OBJECT *) node->data_header;
   COPY_OID (&record->oid, oid);
   COPY_OID (&record->classoid, classoid);
   record->lock_mode = lock;
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -294,8 +294,8 @@ static LOG_PAGE *log_dump_record_assigned_mvccid (THREAD_ENTRY * thread_p, FILE 
 						  LOG_PAGE * log_page_p);
 static LOG_PAGE *log_dump_record_supplemental_info (THREAD_ENTRY * thread_p, FILE * out_fp, LOG_LSA * log_lsa,
 						    LOG_PAGE * log_page_p);
-static LOG_PAGE *log_dump_record_locked_object (THREAD_ENTRY * thread_p, FILE * out_fp, LOG_LSA * log_lsa,
-						LOG_PAGE * log_page_p);
+static LOG_PAGE *log_dump_record_repl_ddl_lock_info (THREAD_ENTRY * thread_p, FILE * out_fp, LOG_LSA * log_lsa,
+						     LOG_PAGE * log_page_p);
 static LOG_PAGE *log_dump_record (THREAD_ENTRY * thread_p, FILE * out_fp, LOG_RECTYPE record_type, LOG_LSA * lsa_p,
 				  LOG_PAGE * log_page_p, LOG_ZIP * log_zip_p);
 static void log_rollback_record (THREAD_ENTRY * thread_p, LOG_LSA * log_lsa, LOG_PAGE * log_page_p,
@@ -511,8 +511,8 @@ log_to_string (LOG_RECTYPE type)
       return "LOG_DUMMY_GENERIC";
     case LOG_SUPPLEMENTAL_INFO:
       return "LOG_SUPPLEMENTAL_INFO";
-    case LOG_LOCKED_OBJECT:
-      return "LOG_LOCKED_OBJECT";
+    case LOG_REPL_DDL_LOCK_INFO:
+      return "LOG_REPL_DDL_LOCK_INFO";
     case LOG_SMALLER_LOGREC_TYPE:
     case LOG_LARGER_LOGREC_TYPE:
       break;
@@ -3735,11 +3735,12 @@ log_append_assigned_mvccid (THREAD_ENTRY * thread_p, MVCCID mvccid)
 }
 
 void
-log_append_locked_object (THREAD_ENTRY * thread_p, const OID * classoid, const OID * oid, const LOCK lock)
+log_append_repl_ddl_lock_info (THREAD_ENTRY * thread_p, const OID * classoid, const OID * oid, const LOCK lock)
 {
-  assert (!OID_ISNULL (classoid));
+  assert (!OID_ISNULL (classoid) && !OID_ISNULL (oid));
   assert (!OID_IS_ROOTOID (classoid) || lock == SCH_M_LOCK);	// if classoid is rootoid, lock must be SCH_M_LOCK
-  assert (!OID_EQ (classoid, oid_Serial_class_oid) || lock == X_LOCK);	// if classoid is serial class oid, lock must be X_LOCK
+  assert (!oid_is_serial (classoid) || lock == X_LOCK);	// if classoid is serial class oid, lock must be X_LOCK
+  assert ((OID_IS_ROOTOID (classoid) && lock == SCH_M_LOCK) || (oid_is_serial (classoid) && lock == X_LOCK));
 
   int tran_index = LOG_FIND_THREAD_TRAN_INDEX (thread_p);
   LOG_TDES *tdes = LOG_FIND_TDES (tran_index);
@@ -3751,14 +3752,14 @@ log_append_locked_object (THREAD_ENTRY * thread_p, const OID * classoid, const O
     }
 
   LOG_PRIOR_NODE *node =
-    prior_lsa_alloc_and_copy_data (thread_p, LOG_LOCKED_OBJECT, RV_NOT_DEFINED, NULL, 0, NULL, 0, NULL);
+    prior_lsa_alloc_and_copy_data (thread_p, LOG_REPL_DDL_LOCK_INFO, RV_NOT_DEFINED, NULL, 0, NULL, 0, NULL);
   if (node == nullptr)
     {
       assert (false);
       return;
     }
 
-  auto *const record = (LOG_REC_LOCKED_OBJECT *) node->data_header;
+  auto *const record = (LOG_REC_REPL_DDL_LOCK_INFO *) node->data_header;
   COPY_OID (&record->oid, oid);
   COPY_OID (&record->classoid, classoid);
   record->lock_mode = lock;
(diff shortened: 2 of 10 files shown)


src/transaction/log_record.hpp
```diff
@@ -152,11 +152,11 @@ enum log_rectype
                                  * it contains transaction user info, DDL statement, undo lsa, redo lsa for DML,
                                  * or undo images that never retrieved from the log. */
 
-  LOG_SCHEMA_MODIFICATION_LOCK, /* Log when SCH_M_LOCK is acquired, and this is for PTS replication to know
-                                 * when to acquire a SCH_M_LOCK for replicating DDL modification, and
-                                 * which table to be locked.
-                                 * PTS needs to block the read transactions which try to access the same class
+  LOG_LOCKED_OBJECT,            /* Log when the lock is acquired for DDL operation, and this is for PTS replicator
```
wrlawodms: I think the name should include more specific context like 
- `LOG_LOCK_NON_MVCC_OBJECT`
- or `LOG_LOCK_OBJECT_ON_PTS`.

The first one describes when the record is created and the second one is the purpose of this lock. I prefer the second. Or your suggestion.

src/transaction/log_replication_atomic.cpp
```diff
@@ -123,7 +124,7 @@ namespace cublog
 		m_replicator_mvccid->complete_mvcc (header.trid, replicator_mvcc::COMMITTED);
 	      }
 
-	    if (m_locked_classes.count (header.trid) > 0)
+	    if (m_locked_objects.count (header.trid) > 0)
```
wrlawodms: Does a change to the _db_serial trigger to invalidate all those caches?

src/transaction/log_manager.c
```diff
@@ -7296,11 +7298,11 @@ static LOG_PAGE *
 log_dump_record_schema_modification_lock (THREAD_ENTRY * thread_p, FILE * out_fp, LOG_LSA * log_lsa,
```
wrlawodms: It should be changed as well.

src/transaction/log_replication_atomic.cpp
```diff
@@ -233,20 +234,20 @@ namespace cublog
...
+	  case LOG_LOCKED_OBJECT:
 	  {
-	    m_redo_context.m_reader.advance_when_does_not_fit (sizeof (LOG_REC_SCHEMA_MODIFICATION_LOCK));
-	    const LOG_REC_SCHEMA_MODIFICATION_LOCK log_rec =
-		    m_redo_context.m_reader.reinterpret_copy_and_add_align<LOG_REC_SCHEMA_MODIFICATION_LOCK> ();
+	    m_redo_context.m_reader.advance_when_does_not_fit (sizeof (LOG_REC_LOCKED_OBJECT));
+	    const LOG_REC_LOCKED_OBJECT log_rec =
+		    m_redo_context.m_reader.reinterpret_copy_and_add_align<LOG_REC_LOCKED_OBJECT> ();
 
 	    if (is_locked_for_ddl (header.trid, &log_rec.classoid))
```
```suggestion
	    if (is_locked_for_ddl (header.trid, &log_rec.oid))
```
Am I correct? it seems to check Rootclass and db_serial.. Then I don't know how it passed your tests. Sorry if I'm missing something.

src/transaction/log_replication_atomic.cpp
```diff
@@ -381,42 +382,53 @@ namespace cublog
   void
   atomic_replicator::release_all_locks_for_ddl (cubthread::entry &thread_entry, const TRANID trid)
```
wrlawodms, line 402: Now it's not only for DDL. A row in db_serial can be modified by not only DDL but also a DML.
```
CREATE TABLE athlete_idx( code INT, name VARCHAR(40) );
CREATE SERIAL order_no START WITH 10000 INCREMENT BY 2 MAXVALUE 20000;
INSERT INTO athlete_idx VALUES (order_no.NEXT_VALUE, 'Park');
INSERT INTO athlete_idx VALUES (order_no.NEXT_VALUE, 'Kim');
INSERT INTO athlete_idx VALUES (order_no.NEXT_VALUE, 'Choo');
INSERT INTO athlete_idx VALUES (order_no.CURRENT_VALUE, 'Lee');
```

wrlawodms (commented):
Some might be silly questions..

src/transaction/log_manager.c
joohok: Thanks, I missed it

src/transaction/log_record.hpp
joohok: I prefer `LOG_LOCK_OBJECT_ON_PTS`, because we are not logging for all the non-mvcc objects.

src/transaction/log_replication_atomic.cpp
joohok: In `discard_caches_for_ddl ()`, it checks if it related to class record . `(OID_EQ (&log_rec.classoid, oid_Root_class_oid))`. So, changes to the _db_serial does not invalidate caches

src/transaction/log_replication_atomic.cpp
joohok: !!! Oh... thanks..

it's possible that the lock record was applied multiple times. While there may not be any functional issues, the `m_locked_object` data structure could potentially contain multiple instances of the same lock information. It seems I overlooked this while requesting and releasing the same lock multiple times. Since there don't seem to be any functional problems, I'll fix it. Thank you.

src/transaction/log_replication_atomic.cpp
joohok: Sorry for not clearly conveying the issue. Currently, I'd like to focus solely on DDL statements for serials. Since serials do not acquire a separate lock during DML operations, separate handling is required

src/transaction/log_record.hpp
joohok: If I were to change this name, it would require quite a bit of modification. While I prefer `LOG_LOCK_OBJECT_ON_PTS`, I'd like to know other reviewers' opinions and then consider modifications.

joohok (commented):
comments were pending state..

src/transaction/log_manager.c
```diff
@@ -3735,7 +3735,7 @@ log_append_assigned_mvccid (THREAD_ENTRY * thread_p, MVCCID mvccid)
 }
 
 void
-log_append_schema_modification_lock (THREAD_ENTRY * thread_p, const OID * classoid)
+log_append_locked_object (THREAD_ENTRY * thread_p, const OID * classoid, const OID * oid, const LOCK lock)
 {
   assert (!OID_ISNULL (classoid));
```
```suggestion
  assert (!OID_ISNULL (classoid));
  assert (classoid != rootclass || lock == SCH_M_LOCK); // rootclass -> SCH_M_LOCK
  assert (classoid != db_serial || lock == X_LOCK);  // db_serial -> X_LOCK
```

src/storage/heap_file.c
```diff
@@ -20657,12 +20657,14 @@ heap_get_insert_location_with_lock (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONT
 	{
 	  /* successfully locked! */
 #if defined (SERVER_MODE)
-	  if (lock == SCH_M_LOCK && OID_IS_ROOTOID (&context->class_oid))
+	  const bool is_lock_for_class = OID_IS_ROOTOID (&context->class_oid) && lock == SCH_M_LOCK;
```
wrlawodms, line 20664: I think we need a descriptive comment here. It was only for class, and the code itself describes it. However, now we take the serial class into account and this code doesn't describe why. 
```suggestion
          // leaving a lock information for DDL. The insertion to the db_serial means a new serial is created with the DDL like create serial ..
	  const bool is_lock_for_class = OID_IS_ROOTOID (&context->class_oid) && lock == SCH_M_LOCK;
```

wrlawodms: The same goes for others leaving the log record as well.

src/transaction/log_replication_atomic.cpp
wrlawodms: For readability, I suggest two options:
1. `m_locked_objects` -> `m_locked_objects_for_ddl`. It's simple, but honestly I don't like the big `if` in `discard_caches_for_ddl`. But, OK.
2. Separating the map for each db_serial and Rootclass, and do what it's needed to do explicitly.
```
if (m_locked_class.count () > 0)
{
  discard_caches_for_ddl (thread_entry, header.trid);
}
if (m_locked_class.count () + m_locked_serial.count() > 0)
{
 release_all_locks_for_ddl (thread_entry, header.trid);
}
```
3. or add a comment.

src/transaction/log_replication_atomic.cpp
```diff
@@ -381,42 +382,53 @@ namespace cublog
...
-  atomic_replicator::acquire_lock_for_ddl (cubthread::entry &thread_entry, const TRANID trid, const OID *classoid)
+  atomic_replicator::acquire_lock_for_ddl (cubthread::entry &thread_entry, const TRANID trid,
+      const LOG_REC_LOCKED_OBJECT &log_rec)
   {
-    assert (!OID_ISNULL (classoid) && !OID_ISTEMP (classoid));
+    assert (!OID_ISNULL (&log_rec.classoid) && !OID_ISTEMP (&log_rec.classoid));
+    assert (!OID_ISNULL (&log_rec.oid) && !OID_ISTEMP (&log_rec.oid));
 
     /* TODO:
      * If a PTS read transaction holds a lock for an extended period without releasing it for the same class,
```
```suggestion
     * If a PTS read transaction holds a lock for an extended period without releasing it for the same class or the same object in the db_serial,
```

src/transaction/log_replication_atomic.cpp
```diff
@@ -381,42 +382,53 @@ namespace cublog
...
+  atomic_replicator::bookkeep_classname_for_ddl (cubthread::entry &thread_entry, const LOG_REC_LOCKED_OBJECT &log_rec)
   {
-    assert (!OID_ISNULL (classoid) && !OID_ISTEMP (classoid));
+    assert (!OID_ISNULL (&log_rec.oid) && !OID_ISTEMP (&log_rec.oid));
+
+    if (!OID_EQ (&log_rec.classoid, oid_Root_class_oid))
+      {
+	assert (OID_EQ (&log_rec.classoid, oid_Serial_class_oid));
+	return;
+      }
```
wrlawodms: I think it's better to do this filtering in the caller. Why should we even call this function for the db_serial class? it's not only unnecessary, but harms readability.

wrlawodms (commented):
Q. Is there a way to guarantee that the X_LOCK on an object in the `db_serial` goes through the paths in this PR? at least with an assert? I think you must know what I'm concerned. It'd be very hard to catch that something gets wrong.

src/transaction/log_record.hpp
hornetmj: How about LOG_LOCK_INFO_FOR_DDL_REPL?

hornetmj (approved):

src/storage/heap_file.c
joohok: OK, I will leave a comment

src/transaction/log_manager.c
joohok: thanks

src/transaction/log_replication_atomic.cpp
joohok: Separating the map will be clear, I will do as second one.

src/transaction/log_replication_atomic.cpp
joohok: Thanks

src/transaction/log_replication_atomic.cpp
joohok: OK, it will be removed

src/transaction/log_record.hpp
joohok: Hi! 
Hi! Thank you for your opinions. I tried to reference the logs used exclusively in PTS, but it wasn't helpful.
- Replication logs (e.g., LOG_START_ATOMIC_REPL, LOG_END_ATOMIC_REPL)
- Logs conveying information for PTS (e.g., LOG_ASSIGNED_MVCCID)

Therefore, based on the suggestions you provided, I aimed to incorporate details about what information the logs carry and what actions they perform. 
- lock information obtained during DDL execution and 
- convey the meaning of replication.

 ** `LOG_REPL_DDL_LOCK_INFO` **
Is it acceptable to proceed in this manner?

src/storage/heap_file.c
```diff
@@ -20657,12 +20657,18 @@ heap_get_insert_location_with_lock (THREAD_ENTRY * thread_p, HEAP_OPERATION_CONT
 	{
 	  /* successfully locked! */
 #if defined (SERVER_MODE)
-	  if (lock == SCH_M_LOCK && OID_IS_ROOTOID (&context->class_oid))
+	  /* lock information is logged to replicate DDL operations on PTS.
+	   * Logging here occurs when a class is created (class_oid == oid_Root_class_oid) or
+	   * when a serial is created (class_oid == oid_Serial_class_oid).
+	   * The creation of a serial implies the insertion of a record into db_serial*/
+	  const bool is_lock_for_class = OID_IS_ROOTOID (&context->class_oid) && lock == SCH_M_LOCK;
+	  const bool is_lock_for_serial = OID_EQ (&context->class_oid, oid_Serial_class_oid) && lock == X_LOCK;
```
cristiarg: Nit
`OID_EQ (&context->class_oid, oid_Serial_class_oid)` -> `OID_IS_SERIAL_OID...`

src/transaction/log_manager.c
```diff
@@ -3749,15 +3751,17 @@ log_append_schema_modification_lock (THREAD_ENTRY * thread_p, const OID * classo
...
-    prior_lsa_alloc_and_copy_data (thread_p, LOG_SCHEMA_MODIFICATION_LOCK, RV_NOT_DEFINED, NULL, 0, NULL, 0, NULL);
+    prior_lsa_alloc_and_copy_data (thread_p, LOG_LOCKED_OBJECT, RV_NOT_DEFINED, NULL, 0, NULL, 0, NULL);
   if (node == nullptr)
     {
       assert (false);
       return;
     }
 
-  auto record = (LOG_REC_SCHEMA_MODIFICATION_LOCK *) node->data_header;
+  auto record = (LOG_REC_LOCKED_OBJECT *) node->data_header;
```
cristiarg: Nit:
```suggestion
  LOG_REC_LOCKED_OBJECT *const record = (LOG_REC_LOCKED_OBJECT *) node->data_header;
```
Not sure if `auto *const record` is an option. I forgot these details.

src/transaction/log_record.hpp
cristiarg: `LOG_REPL_DDL_LOCK_INFO` is fine by me;
It would be good if all the other have `_REPL_` in their name as well to convey the info that it's about replication.

`LOG_ASSIGNED_MVCCID` should also be renamed to `LOG_REPL_ASSIGNED_MVCCID` at some point

src/transaction/log_replication_atomic.cpp
```diff
@@ -126,6 +127,10 @@ namespace cublog
 	    if (m_locked_classes.count (header.trid) > 0)
 	      {
 		discard_caches_for_ddl (thread_entry, header.trid);
+	      }
```
cristiarg: Nit.

What is the point of these checks done here? Can't the checks be moved to within the function. Nothing against, but I don't see the purpose of this verbosity and I am curious.

src/transaction/log_replication_atomic.cpp
```diff
@@ -126,6 +127,10 @@ namespace cublog
...
+
+	    if (m_locked_classes.count (header.trid) + m_locked_serials.count (header.trid) > 0)
```
cristiarg: Nit.
Same here? What's the point of checking?

src/transaction/log_replication_atomic.cpp
```diff
@@ -159,7 +164,7 @@ namespace cublog
 		m_replicator_mvccid->complete_mvcc (header.trid, replicator_mvcc::ABORTED);
 	      }
 
-	    if (m_locked_classes.count (header.trid) > 0)
+	    if (m_locked_classes.count (header.trid) + m_locked_serials.count (header.trid) > 0)
```
cristiarg: Same here ...

src/transaction/log_manager.c
```diff
@@ -3735,9 +3735,11 @@ log_append_assigned_mvccid (THREAD_ENTRY * thread_p, MVCCID mvccid)
 }
 
 void
-log_append_schema_modification_lock (THREAD_ENTRY * thread_p, const OID * classoid)
+log_append_locked_object (THREAD_ENTRY * thread_p, const OID * classoid, const OID * oid, const LOCK lock)
 {
   assert (!OID_ISNULL (classoid));
+  assert (!OID_IS_ROOTOID (classoid) || lock == SCH_M_LOCK);	// if classoid is rootoid, lock must be SCH_M_LOCK
+  assert (!OID_EQ (classoid, oid_Serial_class_oid) || lock == X_LOCK);	// if classoid is serial class oid, lock must be X_LOCK
```
cristiarg: Are these 2 asserts equivalent to the conditions used to call this function?
```
		  const bool is_lock_for_class = OID_IS_ROOTOID (class_oid) && lock == SCH_M_LOCK;
		  const bool is_lock_for_serial = OID_EQ (class_oid, oid_Serial_class_oid) && lock == X_LOCK;
		  if (is_lock_for_class || is_lock_for_serial)
```

As they are written, each one can be valid in the negative form - when the class oid is neither root class nor serial.
As the function is called, at least one of them has to be valid in the affirmative form.

src/transaction/log_replication_atomic.cpp
```diff
@@ -233,20 +238,24 @@ namespace cublog
...
 	      }
 
-	    acquire_lock_for_ddl (thread_entry, header.trid, &log_rec.classoid);
+	    acquire_lock_for_ddl (thread_entry, header.trid, log_rec, is_class);
 
-	    bookkeep_classname_for_ddl (thread_entry, &log_rec.classoid);
+	    if (is_class)
+	      {
+		bookkeep_classname_for_ddl (thread_entry, &log_rec.oid);
+	      }
```
cristiarg: Nit.
Same here, all these checks could be moved inside the function.

src/transaction/log_replication_atomic.cpp
```diff
@@ -507,21 +530,34 @@ namespace cublog
...
+	partition_decache_class (&thread_entry, &log_rec.oid);
       }
   }
 
   bool
-  atomic_replicator::is_locked_for_ddl (const TRANID trid, const OID *classoid)
+  atomic_replicator::is_locked_for_ddl (const TRANID trid, const OID *oid, const bool is_class)
   {
-    return m_locked_classes.count (trid) > 0 && m_locked_classes[trid].count (*classoid) > 0;
+    const auto locked_objects = is_class ? m_locked_classes : m_locked_serials;
```
cristiarg:
```suggestion
    const auto& locked_objects = is_class ? m_locked_classes : m_locked_serials;
```
Otherwise, isn't it a copy of the container?

Also, `const` function.

src/transaction/log_manager.c
joohok: Do you mean that additional assertion is required ? 
```
assert (is_lock_for_class || is_lock_for_serial); 
```

src/transaction/log_replication_atomic.cpp
joohok: DDL has a lower execution rate than DML, and I thought that calling DDL functions for every commit/abort log would incur costs. There are stack creations/destructions, changes in register values, etc. If the caller knows the conditions for calling the function, I wanted to selectively call the function based on these conditions. If the specified conditions are not suitable for the context, are not understandable, or are too verbose, would it be better to place them inside?

src/transaction/log_replication_atomic.cpp
joohok: Oh.. thanks

src/transaction/log_manager.c
```diff
@@ -3735,9 +3735,11 @@ log_append_assigned_mvccid (THREAD_ENTRY * thread_p, MVCCID mvccid)
 }
 
 void
-log_append_schema_modification_lock (THREAD_ENTRY * thread_p, const OID * classoid)
+log_append_locked_object (THREAD_ENTRY * thread_p, const OID * classoid, const OID * oid, const LOCK lock)
 {
   assert (!OID_ISNULL (classoid));
```
```suggestion
  assert (!OID_ISNULL (classoid) && !OID_ISNULL (oid));
```

src/storage/heap_file.c
joohok: I used `oid_is_serial ()`. Done!

wrlawodms (approved):

src/transaction/log_manager.c
cristiarg: Yes, something analogue to that.

src/transaction/log_replication_atomic.cpp
```diff
@@ -507,21 +531,34 @@ namespace cublog
   void
   atomic_replicator::discard_caches_for_ddl (cubthread::entry &thread_entry, const TRANID trid)
```
cristiarg, line 542: This can also be a `const` function

src/transaction/log_replication_atomic.cpp
cristiarg: Your logic makes sense. I would have done the same to optimize a bit.

But then, it makes more sense to reuse the result of `m_locked_classes.count (header.trid)` instead of calling it repeatedly (called at ~127 and then again at ~132).
The compiler might be able to optimize that in release compilation, but I am not sure since there is some side effect inbetween.

Better still, in `atomic_replicator::discard_caches_for_ddl` an equal range is extracted for `trid` that can also be used to infer the result of `count (header.trid)` - `count()` value is not actually needed, what is needed is "is there at least one element in the range".

In general whatever you do in a function as processing far outweighs the cost of calling that function.
Compare current implementation, for each LOG_COMMIT/LOG_ABORT:
- call to `m_locked_classes.count`
- a second call to `m_locked_classes.count` (which could be avoided) and a call to `m_locked_serials.count`

versus a possible:
- call to `atomic_replicator::discard_caches_for_ddl`
- call to `equal_range`, find if the range contains anything by iterating and breaking it (as far as I know, there is no `std::...` algorithm to answer the question "do I have at least one element in this range of iterators"
- based o that, call, or not `discard_caches_for_ddl` with the equal range you already have
- and something similar for `m_locked_serials` ..
- I think that, all in all would result in a more optimized execution, but it's just nit-picking :)

The cheapest thing is to reuse the result of `m_locked_classes.count` here and - probably - for abort as well.

cristiarg (approved):
Approved :)

src/transaction/log_replication_atomic.cpp
joohok: > versus a possible:

call to atomic_replicator::discard_caches_for_ddl
call to equal_range, find if the range contains anything by iterating and breaking it (as far as I know, there is no std::... algorithm to answer the question "do I have at least one element in this range of iterators"
based o that, call, or not discard_caches_for_ddl with the equal range you already have
and something similar for m_locked_serials ..
I think that, all in all would result in a more optimized execution, but it's just nit-picking :)

I agree with you, there seem to be areas where optimization can be applied. 
For now, I will reuse the `m_locked_classes.count`

PR #4820, opened 2023-11-30
[CBRD-25107] [Regression] Core dumped in stats_get_ndv_by_query at src/storage/statistics_cl.c:474

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-25107

The problem occurs because the number of oids is greater than the actual oids. so I fix oid_cnt to check ignore_class when performing update_statistics in loaddb and add routine to Init class_attr_ndv.

### DIFF
--- a/src/communication/network_interface_cl.c
+++ b/src/communication/network_interface_cl.c
@@ -5786,7 +5786,7 @@ stats_update_statistics (MOP classop, int with_fullscan)
   char *reply;
   char *ptr;
   int request_size;
-  CLASS_ATTR_NDV class_attr_ndv;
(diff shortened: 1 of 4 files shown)


xmilex-git (approved):
Why is NDV initialized to 0 when it cannot be 0 and must have a value of 1 or more? Is it because the implementation adds 1 each time it finds a different value?

youngjinj (approved):

shparkcubrid:
> Why is NDV initialized to 0 when it cannot be 0 and must have a value of 1 or more? Is it because the implementation adds 1 each time it finds a different value?

It is 0 if statistics has not been updated, and in this case, default selectivity is set.

PR #4821, opened 2023-11-30
[LETS-852] Fix the core dump in PTS when rollback the CREATE table statement 

joohok:
http://jira.cubrid.org/browse/LETS-852

After executing CREATE TABLE in ATS and then rollback, a core dump occurred in PTS. 

PTS replicator keeps track of logs that need to be applied atomically, such as sysops. (atomic_replication_helper)
When replicating logs like LOG_SYSOP_END, it checks if it can apply the logs within the sysop scope (sysop start ~ sysop end) and then purge the  bookkept logs.

However, the issue arose because the handling for `LOG_SYSOP_END_LOGICAL_COMPENSATE` type was missing, leading to the failure to purge the bookkept logs.

To address this problem, I added a case within the atomic_log_sequence to accommodate the `LOG_SYSOP_END_LOGICAL_COMPENSATE` log type.

Log sequence found when `CREATE TABLE` is rollbacked :  
```
LOG_SYSOP_ATOMIC_START
RVFL_PARTSECT_DEALLOC
RVFL_FHEAD_DEALLOC
RVPGBUF_DEALLOC
LOG_SYSOP_END_LOGICAL_COMPENSATE
```

The condition for LOG_SYSOP_END_LOGICAL_COMPENSATE is different from LOG_SYSOP_END_LOGICAL_UNDO because the last parent LSA for LOG_SYSOP_END_LOGICAL_COMPENSATE could not be NULL.

### DIFF
--- a/src/transaction/atomic_replication_helper.cpp
+++ b/src/transaction/atomic_replication_helper.cpp
@@ -763,7 +763,7 @@ namespace cublog
 	    const atomic_log_entry &last_but_one_entry = *last_but_one_entry_it;
 
 	    if (!LSA_ISNULL (&last_entry.m_sysop_end_last_parent_lsa) &&
-		(last_but_one_entry.m_lsa > last_entry.m_sysop_end_last_parent_lsa))
+		(last_but_one_entry.m_lsa >= last_entry.m_sysop_end_last_parent_lsa))
 	      {
 		assert (last_but_one_entry.m_rectype == LOG_SYSOP_ATOMIC_START);
 
--- a/src/transaction/atomic_replication_helper.cpp
+++ b/src/transaction/atomic_replication_helper.cpp
@@ -713,9 +713,10 @@ namespace cublog
 	      }
 	  }
 	// part of scenario (3)
-	// atomic replication sequence within a postpone sequence
+	// atomic replication sequence within a postpone sequence or logical compensation sequence
 	else if (LOG_SYSOP_END == last_entry.m_rectype
-		 && LOG_SYSOP_END_LOGICAL_RUN_POSTPONE == last_entry.m_sysop_end_type)
+		 && (LOG_SYSOP_END_LOGICAL_RUN_POSTPONE == last_entry.m_sysop_end_type
+		     || LOG_SYSOP_END_LOGICAL_COMPENSATE == last_entry.m_sysop_end_type))
 	  {
 	    const atomic_log_entry_vector_type::const_iterator last_but_one_entry_it = (last_entry_it - 1);
 	    const atomic_log_entry &last_but_one_entry = *last_but_one_entry_it;
--- a/src/transaction/atomic_replication_helper.cpp
+++ b/src/transaction/atomic_replication_helper.cpp
@@ -756,6 +756,26 @@ namespace cublog
 		  }
 	      }
 	  }
+	// scenario (6)
+	else if (LOG_SYSOP_END == last_entry.m_rectype
+		 && LOG_SYSOP_END_LOGICAL_COMPENSATE == last_entry.m_sysop_end_type)
+	  {
+	    const atomic_log_entry_vector_type::const_iterator last_but_one_entry_it = (last_entry_it - 1);
+	    const atomic_log_entry &last_but_one_entry = *last_but_one_entry_it;
+
+	    if (!LSA_ISNULL (&last_entry.m_sysop_end_last_parent_lsa) &&
+		(last_but_one_entry.m_lsa > last_entry.m_sysop_end_last_parent_lsa))
+	      {
+		assert (last_but_one_entry.m_rectype == LOG_SYSOP_ATOMIC_START);
+
+		m_log_vec.erase (last_but_one_entry_it, m_log_vec.cend ());
+
+		if (prm_get_bool_value (PRM_ID_ER_LOG_PTS_ATOMIC_REPL_DEBUG))
+		  {
+		    dump ("sequence::can_purge(10) after erase");
+		  }
+	      }
+	  }
 	// scenario (1)
 	else if (LOG_END_ATOMIC_REPL == last_entry.m_rectype)
 	  {
@@ -781,7 +801,7 @@ namespace cublog
 
 		    if (prm_get_bool_value (PRM_ID_ER_LOG_PTS_ATOMIC_REPL_DEBUG))
 		      {
-			dump ("sequence::can_purge(10) after erase");
+			dump ("sequence::can_purge(11) after erase");
 		      }
 
 		    break;
@@ -791,7 +811,7 @@ namespace cublog
 		  {
 		    if (prm_get_bool_value (PRM_ID_ER_LOG_PTS_ATOMIC_REPL_DEBUG))
 		      {
-			dump ("sequence::can_purge(11) error");
+			dump ("sequence::can_purge(12) error");
 		      }
 
 		    assert_release ("inconsistent atomic log sequence found" == nullptr);
--- a/src/transaction/atomic_replication_helper.cpp
+++ b/src/transaction/atomic_replication_helper.cpp
@@ -713,10 +713,9 @@ namespace cublog
 	      }
 	  }
 	// part of scenario (3)
-	// atomic replication sequence within a postpone sequence or logical compensation sequence
+	// atomic replication sequence within a postpone sequence
 	else if (LOG_SYSOP_END == last_entry.m_rectype
-		 && (LOG_SYSOP_END_LOGICAL_RUN_POSTPONE == last_entry.m_sysop_end_type
-		     || LOG_SYSOP_END_LOGICAL_COMPENSATE == last_entry.m_sysop_end_type))
+		 && LOG_SYSOP_END_LOGICAL_RUN_POSTPONE == last_entry.m_sysop_end_type)
 	  {
 	    const atomic_log_entry_vector_type::const_iterator last_but_one_entry_it = (last_entry_it - 1);
 	    const atomic_log_entry &last_but_one_entry = *last_but_one_entry_it;
--- a/src/transaction/atomic_replication_helper.hpp
+++ b/src/transaction/atomic_replication_helper.hpp
@@ -143,6 +143,16 @@ namespace cublog
    *     |                          RVHF_SET_PREV_VERSION_LSA)
    *     |
    *     \--LOG_END_ATOMIC_REPL
+   *
+   *  (6)
+   *    standalone sysop atomic sequences with sysop end type LOG_SYSOP_END_LOGICAL_COMPENSATE
+   *    occurs when CREATE TABLE statement is aborted
+   *
+   *    LOG_SYSOP_ATOMIC_START
+   *     |
+   *     |  .. redo records .. (eg: RVFL_PARTSECT_DEALLOC, RVFL_FHEAD_DEALLOC, RVPGBUF_DEALLOC)
+   *     |
+   *     \--LOG_SYSOP_END with LOG_SYSOP_END_LOGICAL_COMPENSATE
    */
   class atomic_replication_helper
   {
--- a/unit_tests/log_repl/test_main_log_repl_atomic_helper.cpp
+++ b/unit_tests/log_repl/test_main_log_repl_atomic_helper.cpp
@@ -501,6 +501,43 @@ TEST_CASE ("LOG_SYSOP_ATOMIC_START/LOG_SYSOP_START_POSTPONE/LOG_SYSOP_END - 2 Tr
   test_spec.check_after_execution (trid2, atomic_helper);
 }
 
+TEST_CASE ("LOG_SYSOP_ATOMIC_START/LOG_SYSOP_END-LOG_SYSOP_END_LOGICAL_COMPENSATE", "")
+{
+  // logging snippet:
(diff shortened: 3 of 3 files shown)


src/transaction/atomic_replication_helper.cpp
```diff
@@ -713,9 +713,10 @@ namespace cublog
 	      }
 	  }
 	// part of scenario (3)
-	// atomic replication sequence within a postpone sequence
+	// atomic replication sequence within a postpone sequence or logical compensation sequence
 	else if (LOG_SYSOP_END == last_entry.m_rectype
-		 && LOG_SYSOP_END_LOGICAL_RUN_POSTPONE == last_entry.m_sysop_end_type)
+		 && (LOG_SYSOP_END_LOGICAL_RUN_POSTPONE == last_entry.m_sysop_end_type
+		     || LOG_SYSOP_END_LOGICAL_COMPENSATE == last_entry.m_sysop_end_type))
```
wrlawodms: To be honest, I'm having hard time to understand this function. Could you explain some?
1. Why `initial_log_vec_size` must be lesser than 4 if there is no change?
```
 802         // TODO: idle stop condition should normally be removed
 803         if (initial_log_vec_size == m_log_vec.size ())
 804           {
 805             if (initial_log_vec_size >= 4)
 806               {
 807                 if (prm_get_bool_value (PRM_ID_ER_LOG_PTS_ATOMIC_REPL_DEBUG))
 808                   {
 809                     dump ("sequence::can_purge - too many log entries");
 810                   }
 811
 812                 assert (false);
 813               }
 814             break;
 815           }
```
2. Why can we deal with `LOG_SYSOP_END_LOGICAL_COMPENSATE` in the same way with the`LOG_SYSOP_END_LOGICAL_RUN_POSTPONE`?

wrlawodms (commented):
1. Have you considered adding the missing case in the `test_main_log_repl_atomic_helper.cpp` and the comment on `atomic_replication_helper`?
2. Other unhandled sysop end type like `LOG_SYSOP_END_LOGICAL_MVCC_UNDO` is handled when we run into a case using it?

src/transaction/atomic_replication_helper.cpp
wrlawodms: If you think it's hard to explain or the background is too huge, I think I can skip this and just approve it leaving the review to @cristiarg who is the writer of this code.

joohok:
> 1. Have you considered adding the missing case in the `test_main_log_repl_atomic_helper.cpp` and the comment on `atomic_replication_helper`?
> 2. Other unhandled sysop end type like `LOG_SYSOP_END_LOGICAL_MVCC_UNDO` is handled when we run into a case using it?

@wrlawodms 
1. I didn't think about adding cases to 'test_main_log_repl_atomic_helper.cpp.' Thank you, I will add. And The comments doesn't consider specifying up to the type of 'LOG_SYSOP_END' in the atomic replication helper, so I didn't think about it.
I think we can add some explanations when branching with if statements. What do you think about adding more comments?

2. Right, for now, that's the plan. Alternatively, we could address it after this issue, in http://jira.cubrid.org/browse/LETS-541.

src/transaction/atomic_replication_helper.cpp
joohok: I agree...

1. As far as I understand, currently there are no cases where more than 4 control logs are stacked. That's why the threshold value is set to 4. For example, situations like `LOG_SYSOP_ATOMIC_START`,  `LOG_SYSOP_ATOMIC_START`, `LOG_SYSOP_ATOMIC_START`, `LOG_SYSOP_ATOMIC_START` .. where 4 sysops are nested have been deemed non-existent.
We need @cristiarg opinion.

2. Both 'LOG_SYSOP_END_LOGICAL_COMPENSATE' and 'LOG_SYSOP_END_LOGICAL_RUN_POSTPONE' have `last parent_lsa`. `Last parent lsa` is the LSA value of the log just before the sysop start. Therefore, the action of erasing logs from `last parent lsa +1` to `sysop end` is the same for both logs. 
For 'LOG_SYSOP_END_LOGICAL_UNDO,' we classified it separately because the `last_parent_lsa` value can be null."

wrlawodms:
> Right, for now, that's the plan. Alternatively, we could address it after this issue, in http://jira.cubrid.org/browse/LETS-541.

Oh, OK. We already have the issue.

> And The comments doesn't consider specifying up to the type of 'LOG_SYSOP_END' in the atomic replication helper, so I didn't think about it.
I think we can add some explanations when branching with if statements. What do you think about adding more comments?

Other sysop end types are already explained like below. I think we should add the case in the comment.
```
116    *    LOG_START_ATOMIC_REPL
117    *     |   |
118    *     |   |  .. redo records .. (eg: RVDK_RESERVE_SECTORS, RVPGBUF_NEW_PAGE, RVFL_EXTDATA_ADD)
119    *     |   |
120    *     |   \--LOG_SYSOP_END with LOG_SYSOP_END_LOGICAL_UNDO
121    *     |           (with valid lastparent_lsa)
122    *     |   |
123    *     |   |  .. redo records .. (eg: RVHF_STATS[+],
124    *     |   |
125    *     |   \--LOG_SYSOP_END with LOG_SYSOP_END_COMMIT
126    *     |           (with valid lastparent_lsa)
...
```

joohok:
> > Right, for now, that's the plan. Alternatively, we could address it after this issue, in http://jira.cubrid.org/browse/LETS-541.
> 
> Oh, OK. We already have the issue.
> 
> > And The comments doesn't consider specifying up to the type of 'LOG_SYSOP_END' in the atomic replication helper, so I didn't think about it.
> > I think we can add some explanations when branching with if statements. What do you think about adding more comments?
> 
> Other sysop end types are already explained like below. I think we should add the case in the comment.
> 
> ```
> 116    *    LOG_START_ATOMIC_REPL
> 117    *     |   |
> 118    *     |   |  .. redo records .. (eg: RVDK_RESERVE_SECTORS, RVPGBUF_NEW_PAGE, RVFL_EXTDATA_ADD)
> 119    *     |   |
> 120    *     |   \--LOG_SYSOP_END with LOG_SYSOP_END_LOGICAL_UNDO
> 121    *     |           (with valid lastparent_lsa)
> 122    *     |   |
> 123    *     |   |  .. redo records .. (eg: RVHF_STATS[+],
> 124    *     |   |
> 125    *     |   \--LOG_SYSOP_END with LOG_SYSOP_END_COMMIT
> 126    *     |           (with valid lastparent_lsa)
> ...
> ```

Ah..! OK, Thanks. I will add the case

src/transaction/atomic_replication_helper.cpp
hornetmj: > Last parent lsa is the LSA value of the log just before the sysop start

Based on the your comment, Can 'last_but_one_entry.m_lsa >= last_entry.m_sysop_end_last_parent_lsa' condition change to 'last_but_one_entry.m_lsa > last_entry.m_sysop_end_last_parent_lsa' in the below?

hornetmj (approved):

src/transaction/atomic_replication_helper.cpp
joohok: @hornetmj 
Yes, that's correct. In fact, those two values cannot be equal. Based on the current implementation, the LSA of the log before the LOG_SYSOP_ATOMIC_START log is the last parent LSA.  
* last_but_one_entry is LOG_SYSOP_ATOMIC_START
* log_sysop_start_atomic(): set the last parent LSA and then leave the LOG_SYSOP_ATOMIC_START log

In the current atomic replication helper code and related unit tests, it is assumed that the LSA of LOG_SYSOP_ATOMIC_START and the last parent LSA can be the same.

There should not be any issues in terms of functionality because the LOG_SYSOP_ATOMIC_START log only provides timing information.

However, the intent seems important. 

If we want LOG_SYSOP_ATOMIC_START to be equal to the last parent LSA, 
we'll need to change the sequence of logging for LOG_SYSOP_ATOMIC_START. (logging first and set last parent lsa)

If we want to handle LOG_SYSOP_ATOMIC_START > last parent LSA, 
we'll need to modify the atomic replication code and unit tests. 
This is because the unit tests set LOG_SYSOP_ATOMIC_START == last parent LSA.

@cristiarg What are your thoughts on this?

cristiarg: @wrlawodms 
> If you think it's hard to explain or the background is too huge, I think I can skip this and just approve it leaving the review to @cristiarg who is the writer of this code.

It is a purely empirical thing. In practice I found that if there are more than 4 control log remaining in the "stack", it's a sign of alarm and it needs attention.

I say "stack" because you can look at this mechanism implemented by the `atomic_replication_helper` and the nested `atomic_log_sequence` as a sort of convoluted [Reverese Polish Notation](https://en.wikipedia.org/wiki/Reverse_Polish_notation) processing machine:
- there are "control" (aka "operator") log records: these are used to clean up the stack
- and "redo" (aka "operands") log records: there are just pushed onto the stack and cleaned with the help of the control log records

Once the stack contains too many "operators" it's a sign that something might be wrong (`>= 4`).

I am not sure that this is the best design for this matter (if you remember I spent a huge amount of time arriving at this design). This mechanism might be improved/refactored by a developer with an expert understanding of the mechanics of the transactional log.

cristiarg: > As far as I understand, currently there are no cases where more than 4 control logs are stacked. That's why the threshold value is set to 4. For example, situations like LOG_SYSOP_ATOMIC_START, LOG_SYSOP_ATOMIC_START, LOG_SYSOP_ATOMIC_START, LOG_SYSOP_ATOMIC_START .. where 4 sysops are nested have been deemed non-existent.
We need @cristiarg opinion.

I have no better explanation that this one . :)

cristiarg (approved):

Rudeus (approved):

src/transaction/atomic_replication_helper.cpp
wrlawodms: > If we want LOG_SYSOP_ATOMIC_START to be equal to the last parent LSA,
we'll need to change the sequence of logging for LOG_SYSOP_ATOMIC_START. (logging first and set last parent lsa)
> If we want to handle LOG_SYSOP_ATOMIC_START > last parent LSA,
we'll need to modify the atomic replication code and unit tests.
This is because the unit tests set LOG_SYSOP_ATOMIC_START == last parent LSA.
> @cristiarg What are your thoughts on this?

Is it resolved? Which are we going to change? the unit test or the order of operations?

joohok: Not resolved yet. I will ask him again

cristiarg: @hornetmj  @joohok @wrlawodms 

> Based on the your comment, Can 'last_but_one_entry.m_lsa >= last_entry.m_sysop_end_last_parent_lsa' condition change to 'last_but_one_entry.m_lsa > last_entry.m_sysop_end_last_parent_lsa' in the below?

I understand your concern but I cannot remember why it's `>=` instead of `>`.
All the relevant scenarios have been documented in the header comment. Unfortunately I did not also add log dump snippets - which would have been helpful to discern as to where the `>=` is actually `==`.

Maybe it has to do with repeated occurences of `LOG_SYSOP_END` - see for example scenario (4) in the comment - `LOG_SYSOP_ATOMIC_START` - `LOG_SYSOP_END` - `LOG_SYSOP_END` - and see the comment about the optimization. In this case:
- `last_but_one_entry` will be a `LOG_SYSOP_END`
- and `last_entry` will also be a `LOG_SYSOP_END`
- and it might be that in this case the `>=` is actually `==`

IOW:
- `>` might be true when last but one is `LOG_SYSOP_ATOMIC_START`
- but `==` when last but one is a previous `LOG_SYSOP_END`

I have nothing against changing the code towards the stricter `>`. And if it crashes, we'll find out.
The problem is that if it crashes - for example - in the tests, I am not sure it will be easy to repro.

And this leads me to suggest that we should add some systematic testing for this atomic replication mechanism - at some point in the future.

All these scenarios implemented in the atomic replication helper have been done by reverse engineering the dumped transactional log.
Sometimes I tried to go a step forward and identify the source in code where this log is produced.
I even thought of extending the transactional log records with some array fields to log the `__FILE__` and  `__LINE__` where they are logged such as to complete the circle.
It is/was a rudimentary approach but I could not do better.

joohok: In the scenario you described, I understand the situation if the sysop is in a consecutive state.
```
LOG_START_ATOMIC_REPL
    LOG_SYSOP_ATOMIC_START
     .. redo records ..
    LOG_SYSOP_END (with LOG_SYSOP_END_LOGICAL_UNDO )
    (sysop start)
     .. redo records ..
    LOG_SYSOP_END (with LOG_SYSOP_END_COMMIT)
LOG_END_ATOMIC_REPL
```

In fact, the question was about which option would be better among the following choices:
1) Should we change the logging method for LOG_SYSOP_ATOMIC_START to align with the logic checking the atomic log sequence?
   --> Modification needed in log_sysop_start_atomic()
2) or should we adjust the logic checking the atomic log sequence to fit the current logging method?
   --> Unit test modification needed

However, even if we choose 1 or 2, it seems to go beyond the scope of this issue. For now, I suggest wrapping up by just adding case following the existing implementation and creating a new issue to address this part.

If we only add case, I plan to add unit tests and conclude the issue.

wrlawodms: http://jira.cubrid.org/browse/LETS-889

PR #4824, opened 2023-12-04
[CBRD-25104] [Regression] query cost is changed in query plan

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-25104

There is an issue where the total number of rows in the partition table are added per partition. the total number of rows is used without adding.
There is an issue where the selectivity is set to 1 when the varchar is over 4000. it is fixed to be set to 1 only when NDV is 0.

### DIFF
--- a/src/storage/statistics_cl.c
+++ b/src/storage/statistics_cl.c
@@ -507,7 +507,8 @@ stats_get_ndv_by_query (const MOP class_mop, CLASS_ATTR_NDV * class_attr_ndv, FI
 	{
 	  goto end;
 	}
-      class_attr_ndv->attr_ndv[i].ndv = MAX (DB_GET_BIGINT (&value), 1);
+      /* if NDV is 0, it is all null values. */
+      class_attr_ndv->attr_ndv[i].ndv = DB_GET_BIGINT (&value) == 0 ? 1 : DB_GET_BIGINT (&value);
     }
 
   /* get count(*) */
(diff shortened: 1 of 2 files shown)


youngjinj (approved):

xmilex-git (commented):
I was looking at the Repro of the JIRA issue, and it showed that the cost was 3, but when I tried to reproduce it myself, the cost came out as 8. Additionally, the cardinality was reported as 13. Is this because the table is partitioned? Is it okay?
```sql
csql> select /*+ RECOMPILE */ * from foo where id=1;
seg[0]: [0]
seg[1]: id[0] (f)
seg[2]: a[0] (f)
seg[3]: b[0] (f)
Join graph nodes:
node[0]: dba.foo dba.foo(4/10) (sargs 0) (loc 0)
Join graph terms:
term[0]: [dba.foo].id=1 (sel 0.333333) (sarg term) (not-join eligible) (indexable id[0]) (loc 0)

Query plan:

iscan
    class: foo node[0]
    index: i_foo_id term[0]
    cost:  8 card 1

Query stmt:

select foo.id, foo.a, foo.b from foo foo where foo.id= ?:0 


           id            a  b                   
================================================
            1           -1  'one'               
            1           -2  'two'               

2 rows selected. (0.034438 sec) Committed. (0.000055 sec) 

```

xmilex-git (approved):

PR #4825, opened 2023-12-04
[CBRD-25097] PL/CSQL phase-0 internal release

hgryoo:
http://jira.cubrid.org/browse/CBRD-25097

This PR includes PL/CSQL project's phase-0 implementations (feature/plcsql).
After this PR is merged in the develop branch, the internal release is considered completed once QA testing is completed for the features that must be satisfied in phase-0.

The features are as follows:
- **PL/CSQL program compiler**: The PL/CSQL compiler parses the program according to the PL/CSQL grammar defined in the issue and then includes a module that communicates with the DB server for information needed in CUBRID SQL parts such as Static SQL.
- **Support PL/CSQL in DDL (CREATE PROCEDURE/FUNCTION) statement**: In CREATE PROCEDURE/FUNCTION statement, stored routine can be written between BEGIN ... END. The stored routine body is sent to the PL/CSQL compiler and it is registered if compiling succeed.
- **Support TCL (COMMIT/ROLLBACK) in stored routine**: COMMIT and ROLLBACK have been ignored in Java SP. In this change, they are supported if pl_transaction_control parameter is on, which is introduced for backward compatibility.
- **Support DBMS_OUTPUT built-in functions**: Yet package is not added in CUBRID. But, DBMS_OUTPUT is used frequently to debug or log when users write procedure/function.


hornetmj (approved):
There are 'FIXME' comment where I have to review. So, I just approve with expectation you will change them soon.

hyunikn (approved):

hgryoo:
> There are 'FIXME' comment where I have to review. So, I just approve with expectation you will change them soon.

Ok. Let me explain about the FIXME part briefly. 
Previously, if any transactional object is not changed in a transaction, TRANID was not updated. 
I've changed this behavior; TRANID is always accumulated at the end of a transaction.
CUBRID's stored routine runs using java class file. Because the file is managed out of transaction, CUBRID's stored routine is not consistent within a transaction.
One of examples, 
```
CALL sp ();
-- java class changed by loadjava
CALL sp ();
-- java class changed by loadjava
COMMIT;
-- java class changed by loadjava
CALL sp ();
-- session ends
```
Stored routine's body is updated Only Java SP server's class loader loads the updated file.

IMO, When a user changes a java class file, the corresponding stored routine should be changed too when transaction ends.
Even though a user changes a java class file and any table data is not updated, TRANID is not updated. So the CUBRID's stored routine module can not know the transaction ends.

In this context, there were two options for updated stored routine:
* every loadjava, updated stored routine is seen.
* every session ends, the next session can see the updated stored routine

Hmm.. I don't think the both two options are good.

As a quick solution:
To catch transaction ends even though any data is not updated, I've changed the 'FIXME' part to demonstrate the following.
* every transaction ends, updated stored routine can be seen.

I thought accumulating TRANID doesn't have a significant impact to our system.
I would like to discuss about the topic when PL/CSQL phase-1 begins.
Thank you

beyondykk9 (approved):

shparkcubrid (approved):

kisoo-han (approved):

PR #4826, opened 2023-12-04
[CBRD-24921] New API for creating system catalogs

hgryoo:

This PR introduces the new API for creating system catalogs. The current routines that define system tables and views in boot_cl.c share many common parts. 

Proposed API:

**AS-IS:**
defined in `boot_cl.c`
```
  def = smt_edit_class_mop (class_mop, AU_ALTER);

  /* add attributes */
  error_code = smt_add_attribute (def, "class_of", CT_CLASS_NAME, NULL);
  if (error_code != NO_ERROR)
    {
      return error_code;
    }

  error_code = smt_add_attribute (def, "attr_name", "varchar(255)", NULL);
  if (error_code != NO_ERROR)
    {
      return error_code;
    }

...

  /* add constraints */
  error_code = db_add_constraint (class_mop, DB_CONSTRAINT_INDEX, NULL, index_col_names, 0);
  if (error_code != NO_ERROR)
    {
      return error_code;
    }

  if (locator_has_heap (class_mop) == NULL)
    {
      return er_errid ();
    }

  /* add authorizations */
  error_code = au_change_owner (class_mop, Au_dba_user);
  if (error_code != NO_ERROR)
    {
      return error_code;
    }
```

**TO-BE:**

Extract routines related to creating system catalogs
- schema_system_catalog.hpp has two exposed APIs to initialize and to install system catalogs

```
// schema_system_catalog.hpp
extern EXPORT_IMPORT void catcls_init (void);
extern EXPORT_IMPORT int catcls_install (void);
```

Define system catalog's definition (see `schema_system_catalog_install.cpp`)
```
system_catalog_definition catalog_definition (
// name
    "new_catalog",
// columns
    {
      {"class_of", OBJECT},
      {"unique_name", VARCHAR_255},
      {"new_name", VARCHAR_255},
      {"type", INTEGER},
      {"is_system_class", INTEGER},
      {"owner", AU_USER_CLASS_NAME},
    },
    {
      {DB_CONSTRAINT_INDEX, "i__db_new_unique_name", {"unique_name", nullptr}, false},
      {DB_CONSTRAINT_INDEX, "", {"new_name", "owner", nullptr}, false}
    },
// authorization
    {
      // owner, grants
      Au_dba_user, {}
    }
    );
```

The common routine to create a system catalog according to the `system_catalog_definition` above are defined in `schema_system_catalog_builder.cpp`. see `catcls_install ()` in `schema_system_catalog_install.cpp`
```
using catalog_builder = cubschema::system_catalog_builder;
MOP class_mop = catalog_builder::create_and_mark_system_class ("new_catalog");
error_code = catalog_builder::build_class (class_mop, catalog_definition);
```


hgryoo:
I'm going to give a presentation soon about the proposed API in this PR for all reviewers

hyunikn (approved):

src/object/schema_system_catalog_install.cpp
```diff
@@ -0,0 +1,1972 @@
...
+  const size_t num_classes = clist.size ();
+  const size_t num_vclasses = vclist.size ();
+  std::vector<MOP> class_mop (num_classes, nullptr);
+  int save;
+  size_t i;
+  AU_DISABLE (save);
+
+  using catalog_builder = cubschema::system_catalog_builder;
+
+  for (i = 0; i < num_classes; i++)
```
mhoh3963, line 271: How about merging create_and_mark_system_class() and build_class() into a single for loop like code (295-314)?

hgryoo: If you see the next loops below, create_and_mark_system_class () should be called in an iteration when a system view is built. If this order is not followed, assertion occurs.

src/object/schema_system_catalog_install.cpp
```diff
@@ -0,0 +1,1972 @@
...
+	{
+	  error_code = er_errid ();
+	}
+
+
+      if (error_code != NO_ERROR)
+	{
+	  goto end;
+	}
+    }
```
hgryoo, line 314: @mhoh3963 
here is the loop for vclasses

beyondykk9 (approved):

kisoo-han (approved):
Good Work!

hgryoo:
The SQL test regressions are unstable cases caused because queries retrieving the catalog name doesn't have ORDER BY.

shparkcubrid (approved):

src/object/schema_system_catalog.hpp
```diff
@@ -28,6 +28,21 @@
 
 #include "porting.h"
 
+extern EXPORT_IMPORT void catcls_init (void);
+extern EXPORT_IMPORT int catcls_install (void);
+
+// test class_name is of system (class/vclass)s for legacy routine
```
hornetmj:
```suggestion
```

src/object/schema_system_catalog.cpp
```diff
@@ -79,7 +81,10 @@ namespace cubschema
 
     /* currently, not implemented */
     CT_RESOLUTION_NAME,		// "_db_resolution"
+  };
```
hornetmj:
```suggestion
    CT_RESOLUTION_NAME		// "_db_resolution"
  };
```

src/object/schema_system_catalog.cpp
```diff
@@ -105,11 +110,24 @@ namespace cubschema
...
+  static const identifier_store sm_catalog_vclass_names (sm_system_vclass_names, false);
 }
 
 bool sm_check_system_class_by_name (const std::string_view name)
 {
   // TODO: bool is_enclosed = identifier_store::is_enclosed (name);
-  return identifier_store::check_identifier_is_valid (name, false) && cubschema::sm_catalog_names.is_exists (name);
+  return identifier_store::check_identifier_is_valid (name, false)
+	 && (cubschema::sm_catalog_class_names.is_exists (name)
+	     || cubschema::sm_catalog_vclass_names.is_exists (name));
```
hornetmj: Nit:

How about use sm_is_system_class() and sm_is_system_vclass()?

src/object/schema_system_catalog_builder.cpp
```diff
@@ -0,0 +1,270 @@
...
+ *  limitations under the License.
+ *
+ */
+
+#include "schema_system_catalog_builder.hpp"
+
+// #include "dbi.h" /* db_create_class () */
+#include "db.h"
+#include "dbtype.h"
+// #include "dbtype_function.h" /* DB_IS_NULL () */
```
hornetmj:
```suggestion
#include "db.h"
#include "dbtype.h"
```

src/object/schema_system_catalog_definition.hpp
```diff
@@ -0,0 +1,111 @@
...
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+
+
+/*
+ * schema_system_catalog_builder.hpp
```
hornetmj:
```suggestion
 * schema_system_catalog_definition.hpp
```

src/object/schema_system_catalog_definition.hpp
```diff
@@ -0,0 +1,111 @@
...
+
+#include <string_view>
+#include <vector>
+#include <optional>
+
+#include "work_space.h" // struct db_object
+
+namespace cubschema
+{
+  using nullable_string = std::optional<std::string>;
```
hornetmj:
```suggestion
```

hornetmj: Seems un-used type.

src/object/schema_system_catalog_definition.hpp
```diff
@@ -0,0 +1,111 @@
...
+  struct attribute
+  {
+    using default_value_init_type = std::function<int (DB_VALUE *)>;
+
+    attribute_kind kind;
+    std::string name;
+    std::string type;
+    default_value_init_type dvalue_func;
+
+    attribute (const std::string &n, const std::string &t); // attribute_kind::COLUMN
```
hornetmj:
```suggestion
    attribute (const std::string &name, const std::string &type); // attribute_kind::COLUMN
```

hornetmj: NIT:

Plus,

How about using default argument?
```
attribute (const std::string &name, const std::string &type, const attribute_kind kind = attribute_kind::COLUMN);
attribute (const std::string &name, const std::string &type, default_value_init_type dval_f, const attribute_kind kind = attribute_kind::COLUMN);
```

src/object/schema_system_catalog_definition.hpp
```diff
@@ -0,0 +1,111 @@
...
+
+  struct authorization
+  {
+    struct db_object *owner;
+    std::vector<grant> grants;
+
+    authorization (struct db_object *&o, const std::vector<grant> &gs);
+  };
+
+  struct system_catalog_definition
```
hornetmj, line 91: Is there a reason to use struct rather than class?

src/object/schema_system_catalog_definition.cpp
```diff
@@ -0,0 +1,98 @@
...
+    , dvalue_func {dval_f}
+  {
+  }
+
+  constraint::constraint (const DB_CONSTRAINT_TYPE &t, const std::string_view n,
+			  const std::vector<const char *> &attrs, bool is_class_attr)
+    : type {t}, name {n.begin(), n.end ()}, attribute_names {attrs}, is_class_attributes {is_class_attr}
+  {
+    //
+  }
```
hornetmj, line 72:
```suggestion
  { }
```

src/object/schema_system_catalog_constants.h
```diff
@@ -0,0 +1,94 @@
...
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+
+
+/*
+ * schema_system_catalog_constants.h - declare systen catalog name constants
```
hornetmj:
```suggestion
 * schema_system_catalog_constants.h - declare system catalog name constants
```

src/object/schema_system_catalog_builder.hpp
```diff
@@ -0,0 +1,67 @@
...
+
+#ifndef _SCHEMA_SYSTEM_CATALOG_BUILDER_HPP_
+#define _SCHEMA_SYSTEM_CATALOG_BUILDER_HPP_
+
+#include "schema_system_catalog.hpp"
+#include "schema_system_catalog_definition.hpp"
+
+namespace cubschema
+{
+  class system_catalog_builder;
```
hornetmj:
```suggestion
```

src/object/schema_system_catalog_builder.cpp
```diff
@@ -0,0 +1,270 @@
...
+  system_catalog_builder::create_and_mark_system_class (const std::string_view name)
+  {
+    if (name.empty ())
+      {
+	// should not happen
+	assert (false);
+	return nullptr;
+      }
+
+    // class_mop
```
hornetmj:
```suggestion
```

src/object/schema_system_catalog_builder.cpp
```diff
@@ -0,0 +1,270 @@
...
+      {
+	class_mop = db_create_class (name.data ());
+	if (class_mop != nullptr)
+	  {
+	    sm_mark_system_class (class_mop, 1);
+	  }
+      }
+    else if (sm_is_system_vclass (name))
+      {
+	class_mop = db_create_vclass (name.data ());
```
hornetmj, line 52: Don't the vclass also need to mark system class?

I just check using 'select class_name, is_system_class from _db_class' query to identify the vclass has 1 value of is_system_class column. And I see they have 1 value.

-- It would be good to check code but, I didn't that.

src/object/schema_system_catalog_builder.cpp
```diff
@@ -0,0 +1,270 @@
...
+      }
+    else
+      {
+	assert (false);
+      }
+
+    assert (er_errid () == NO_ERROR);
```
hornetmj: Some of callstack like db_create_class() -> locator_reserve_class_name() can set error. So, this condition of assert is not always true. I thought you intend to handle error at the caller function. With this, I think this assert is invalid.

src/object/schema_system_catalog_builder.cpp
```diff
@@ -0,0 +1,270 @@
...
+
+	    if (attr.dvalue_func)
+	      {
+		DB_VALUE default_value;
+		db_make_null (&default_value);
+		attr.dvalue_func (&default_value);
+		error_code = smt_set_attribute_default (def, name, 0, (DB_VALUE *) &default_value, NULL);
+	      }
+	  }
+	  break;
```
hornetmj, line 101:
```suggestion
	    break;
```

src/object/schema_system_catalog_builder.cpp
```diff
@@ -0,0 +1,270 @@
...
+	  {
+	  case DB_CONSTRAINT_INDEX:
+	  case DB_CONSTRAINT_PRIMARY_KEY:
+	  case DB_CONSTRAINT_UNIQUE:
+	  case DB_CONSTRAINT_NOT_NULL:
+	  {
+	    error_code = db_add_constraint (class_mop, type, name, (const char **) c.attribute_names.data(),
+					    is_class_attribute);
+	  }
+	  break;
```
hornetmj, line 147:
```suggestion
	    break;
```

src/object/schema_system_catalog_builder.cpp
```diff
@@ -0,0 +1,270 @@
...
+  }
+
+  int
+  system_catalog_builder::build_class (const MOP class_mop, const system_catalog_definition &catalog_def)
+  {
+    int error_code = NO_ERROR;
+
+    if (class_mop == nullptr)
+      {
+	// should not happen
```
hornetmj, line 69:
```suggestion
```

hornetmj: assert(false) has 'should not happen' information.

src/object/schema_system_catalog_builder.cpp
```diff
@@ -0,0 +1,270 @@
...
+      {
+	assert (g.target_user != nullptr);
+	error_code = au_grant (g.target_user, class_mop, g.auth, g.with_grant_option);
+	if (error_code != NO_ERROR)
+	  {
+	    return error_code;
+	  }
+      }
+
+    return error_code; // NO_ERROR
```
hornetmj:
```suggestion
    return error_code;
```

src/object/schema_system_catalog_install.hpp
```diff
@@ -0,0 +1,107 @@
...
+ */
+
+#ifndef _SCHEMA_SYSTEM_CATALOG_INSTALL_HPP_
+#define _SCHEMA_SYSTEM_CATALOG_INSTALL_HPP_
+
+namespace cubschema
+{
+  // forward definitions
+  struct system_catalog_definition;
+  struct catcls_function;
```
hornetmj:
```suggestion
```

src/object/schema_system_catalog_install.cpp
```diff
@@ -0,0 +1,1972 @@
...
+    "TIMESTAMP", "DATE", "MONETARY", NULL /* VARIABLE */, NULL /* SUB */,
+    NULL /* POINTER */, NULL /* ERROR */, "SHORT", NULL /* VOBJ */,
+    NULL /* OID */,
+    NULL /* VALUE */, "NUMERIC", "BIT", "VARBIT", "CHAR",
+    "NCHAR", "VARNCHAR", NULL /* RESULTSET */, NULL /* MIDXKEY */,
+    NULL /* TABLE */,
+    "BIGINT", "DATETIME",
+    "BLOB", "CLOB", "ENUM",
+    "TIMESTAMPTZ", "TIMESTAMPLTZ", "DATETIMETZ", "DATETIMELTZ",
+    "JSON"
```
hornetmj, line 63: NIT:

For readability, how about the following style?
```
"INTEGER",
"FLOAT",
"DOUBLE",
...
```

src/object/schema_system_catalog_install.cpp
hornetmj: Please, explain more? Does the following code have any problem?

```
  for (i = 0; i < num_classes; i++)
      class_mop[i] = catalog_builder::create_and_mark_system_class (clist[i].name);
      if (class_mop[i] == nullptr)
	{
	  assert (false);
	  error_code = er_errid ();
	  goto end;
	}
 
      // new routine
      error_code = catalog_builder::build_class (class_mop[i], clist[i].definition);
      if (error_code != NO_ERROR)
	{
	  assert (false);
	  error_code = er_errid ();
	  goto end;
	}
    }
```

hgryoo: When I build with the above code, `error_code = catalog_builder::build_class (class_mop[i], clist[i].definition);` returns -244 and assert occurs in `cubrid createdb`

src/object/schema_system_catalog_install.cpp
hgryoo: This array is a local variable only used within this function. I prefer to keep code lines shorter, as long as it doesn't significantly impact readability. :-)

src/object/schema_system_catalog_definition.hpp
hgryoo: Because there is no operation, only used to represent the catalog's data structure.

mhoh3963 (approved):
good job.

src/object/schema_system_catalog_builder.cpp
hgryoo: After db_init() -- here catalogs are created, By sm_mark_system_classes(), all resident classes are marked as system class. Because of this, When I try to remove `sm_mark_system_class (class_mop, 1);` at line 47 above, It works.

However, as this routine has been migrated from legacy code, my emphasis is on introducing an improved API. I intend to move forward by addressing the removal of redundant functions in a new issue.

src/object/schema_system_catalog_builder.cpp
hgryoo: Ok. I think I've added it for debugging in the middle of working it. Any failure from the above functions will return NULL class_mop, I believe that this situation will be handled anyway.

hornetmj: `Any failure from the above functions will return NULL class_mop` doesn't guarantee always if some one change the some of above functions.

src/object/schema_system_catalog_install.cpp
hornetmj: There remain questions about the cause of the problem, but since this is a comment from a readability perspective, there are no further comments. It is your choice to reflect our opinion through further analysis or leave it as is.

hornetmj (approved):

src/object/schema_system_catalog_builder.cpp
```diff
@@ -0,0 +1,266 @@
...
+    else if (sm_is_system_vclass (name))
+      {
+	class_mop = db_create_vclass (name.data ());
+      }
+    else
+      {
+	assert (false);
+      }
+
+    assert (er_errid () == NO_ERROR);
```
hgryoo:
```suggestion
```

src/object/schema_system_catalog_install.cpp
kangmin5505: @hornetmj 
It has been a long time, so you may not remember everything clearly, but I wanted to leave a note in case someone else reviews this in the future.

If the code is written as described, the issue arises because it attempts to add an attribute to a class that has not yet been defined. For example, when defining `_db_class`, the `inst_attrs` attribute is of type `sequence of _db_attribute`, but `_db_attribute` has not yet been defined, which results in an error.

I hope this will be helpful.

PR #4827, opened 2023-12-05
[LETS-883] Cleanup lock resources used by PTS replicator

joohok:
http://jira.cubrid.org/browse/LETS-883

purpose
Since the current PTS replicator also uses a lock manager, it is necessary to clean up any remaining locks upon replicator termination or when all the locks have been released

Use `lock_unlock_all ()` to clean-up resources  
- Release of IX_LOCK for the Root class.
   - When performing SCH_M_LOCK on the class record, an intention lock is acquired for the root class.
   - This is currently not being cleaned up.
   - When there is no more ddl statements to replicate, then replicator call `lock_unlock_all ()`
- Release of locks during PTS shutdown when DDL is in progress.
  - Lock resources obtained for DDL execution need to be cleaned up.
  - call `lock_unlock_all ()` in dtor of atomic_replicator

### DIFF
--- a/src/transaction/log_replication_atomic.cpp
+++ b/src/transaction/log_replication_atomic.cpp
@@ -47,7 +47,7 @@ namespace cublog
      * termination state for atomic replication is needed
      */
 
-    cleanup_lock_resources ();
+    cleanup_lock_resources_for_ddl ();
   }
 
   void
@@ -575,7 +575,7 @@ namespace cublog
   }
 
   void
-  atomic_replicator::cleanup_lock_resources ()
+  atomic_replicator::cleanup_lock_resources_for_ddl ()
   {
     cubthread::entry &thread_entry = cubthread::get_entry ();
     lock_unlock_all (&thread_entry);
--- a/src/transaction/log_replication_atomic.cpp
+++ b/src/transaction/log_replication_atomic.cpp
@@ -45,6 +45,8 @@ namespace cublog
      * and, thus, does not need to be left in consistent state; thus, no check as to the consistent
      * termination state for atomic replication is needed
      */
+
+    cleanup_lock_resources ();
   }
 
   void
@@ -383,9 +385,17 @@ namespace cublog
   {
     assert (m_locked_classes.count (trid) > 0);
 
-    for (auto classoid : m_locked_classes[trid])
+    if (m_locked_classes.size() == 1)
       {
-	lock_unlock_object_and_cleanup (&thread_entry, &classoid, oid_Root_class_oid, SCH_M_LOCK);
+	// If there are no more DDL statements to replicate, the intention locks should also be removed.
+	lock_unlock_all (&thread_entry);
+      }
+    else
+      {
+	for (auto classoid : m_locked_classes[trid])
+	  {
+	    lock_unlock_object_and_cleanup (&thread_entry, &classoid, oid_Root_class_oid, SCH_M_LOCK);
+	  }
       }
 
     m_locked_classes.erase (trid);
@@ -523,5 +533,12 @@ namespace cublog
   atomic_replicator::is_locked_for_ddl (const TRANID trid, const OID *classoid)
   {
     return m_locked_classes.count (trid) > 0 && m_locked_classes[trid].count (*classoid) > 0;
+  }
+
+  void
+  atomic_replicator::cleanup_lock_resources ()
+  {
+    cubthread::entry &thread_entry = cubthread::get_entry ();
+    lock_unlock_all (&thread_entry);
   }
 }
--- a/src/transaction/log_replication_atomic.cpp
+++ b/src/transaction/log_replication_atomic.cpp
@@ -46,7 +46,7 @@ namespace cublog
      * termination state for atomic replication is needed
      */
 
-    cleanup_lock_resources ();
+    cleanup_lock_resources_for_ddl ();
   }
 
   void
@@ -385,17 +385,9 @@ namespace cublog
   {
     assert (m_locked_classes.count (trid) > 0);
 
-    if (m_locked_classes.size() == 1)
+    for (auto classoid : m_locked_classes[trid])
       {
-	// If there are no more DDL statements to replicate, the intention locks should also be removed.
-	lock_unlock_all (&thread_entry);
-      }
-    else
-      {
-	for (auto classoid : m_locked_classes[trid])
-	  {
-	    lock_unlock_object_and_cleanup (&thread_entry, &classoid, oid_Root_class_oid, SCH_M_LOCK);
-	  }
+	lock_unlock_object_and_cleanup (&thread_entry, &classoid, oid_Root_class_oid, SCH_M_LOCK);
       }
 
     m_locked_classes.erase (trid);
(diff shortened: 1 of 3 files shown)


src/transaction/log_replication_atomic.cpp
```diff
@@ -383,9 +385,17 @@ namespace cublog
   {
     assert (m_locked_classes.count (trid) > 0);
 
-    for (auto classoid : m_locked_classes[trid])
+    if (m_locked_classes.size() == 1)
       {
-	lock_unlock_object_and_cleanup (&thread_entry, &classoid, oid_Root_class_oid, SCH_M_LOCK);
+	// If there are no more DDL statements to replicate, the intention locks should also be removed.
```
wrlawodms: Do we really need this? Does it harm anything? The replicator is the only owner of the IX_LOCK on Rootclass in PTS. It may even help to reduce the cost of acquiring the next lock acquisition since it can skip acquiring the IX_LOCK.
If it's released strictly when a transaction ends, I think it has a point, but we even don't.

src/transaction/log_replication_atomic.cpp
```diff
@@ -524,4 +534,11 @@ namespace cublog
   {
     return m_locked_classes.count (trid) > 0 && m_locked_classes[trid].count (*classoid) > 0;
   }
+
+  void
+  atomic_replicator::cleanup_lock_resources ()
```
```suggestion
  atomic_replicator::cleanup_lock_resources_for_ddl ()
```

src/transaction/log_replication_atomic.cpp
```diff
@@ -45,6 +45,8 @@ namespace cublog
      * and, thus, does not need to be left in consistent state; thus, no check as to the consistent
      * termination state for atomic replication is needed
      */
+
+    cleanup_lock_resources ();
```
hornetmj: It should be called before calling lock_finalize(). When I check it, it is already called like that. However, add comment to leave information for others. The method of leaving comments mimics existing code, and it would be nice if there was a clearer way.
```
xboot_shutdown_server()
{
    ...
    pts_ptr->finish_replication_during_shutdown (*thread_p); // stop receiving log before log_final()
```

src/transaction/log_replication_atomic.cpp
hornetmj: low probability; However, I guess that the read transaction with repeatable read isolation level or higher is blocked to escalation from IS to S on the Rootclass. I will look into the related code more to check the situation I tell can occur.

hornetmj (approved):

src/transaction/log_replication_atomic.cpp
```diff
@@ -383,9 +385,17 @@ namespace cublog
   {
     assert (m_locked_classes.count (trid) > 0);
 
-    for (auto classoid : m_locked_classes[trid])
+    if (m_locked_classes.size() == 1)
```
cristiarg: Can you please explain this?
If this check does not take `trid` into account, how is it ensure that the single entry in the `m_locked_classes` is from this very `trid`? And we not are mistakenly unlocking resources from another ATS transaction executing concurrently with this `trid`.

Why not a more conservative approach: if we only want to unlock the `root_class_hold` just extract a function that does that out of `lock_unlock_all` and call that; this would also ensure that other possible mishandlings of lock manager are caught by lock manager verifications.

And if @wrlawodms comment below is valid, then only call this function (that unlocks the root class) in debug mode and leave it out for release mode.

joohok: > Can you please explain this?
If this check does not take trid into account, how is it ensure that the single entry in the m_locked_classes is from this very trid? And we not are mistakenly unlocking resources from another ATS transaction executing concurrently with this trid.

`std::unordered_map <TRANID, std::unordered_set <OID>>  m_locked_classes `
When using the `size()` function to check if the number of stored TRANIDs in the map is one, I attempted to invoke `lock_unlock_all()`. `release_all_locks_for_ddl()` is called when the transaction associated with a TRANID is completed (commit or abort). If that transaction happens to be the last DDL transaction being executed by the replicator, it is considered safe to invoke `lock_unlock_all()`.

> Why not a more conservative approach: if we only want to unlock the root_class_hold just extract a function that does that out of lock_unlock_all and call that; this would also ensure that other possible mishandlings of lock manager are caught by lock manager verifications.

OK, 
to catch potential misuses of the lock manager module within the replicator using lock manager verifications or to prevent unintended functionalities that replicator may not desire in lock_unlock_all(), it seems advisable to extract a function that unlocks root class holds.Thanks.

> And if @wrlawodms comment below is valid, then only call this function (that unlocks the root class) in debug mode and leave it out for release mode.

I don't think I understand your comments.. Is there a reason to unlock in debug mode?

src/transaction/log_replication_atomic.cpp
joohok: OK, I will. thanks

src/transaction/log_replication_atomic.cpp
joohok: @wrlawodms 
In PTS, the transaction acquiring IX_LOCK on the root class is only the replicator. 
Therefore, I believe, as you mentioned, there's no need to explicitly release IX_LOCK. 
Additionally, since replicator is not attempting to acquire IX_LOCK again, it would be more efficient not to release it.

However, 
the fact that replicator, which holds no locks at all, acquires IX_LOCK on the root class struck me as conceptually odd. In my understanding, having IX_LOCK on the root class implies holding X_LOCK on one or more entities (class or instance) beneath it. Hence, I hadn't considered it from your perspective.

> If it's released strictly when a transaction ends, I think it has a point, but we even don't.

but I agree with this, so I think it is better to follow your suggestion

@hornetmj 
It's worth noting that in the function `lock_internal_perform_lock_object()`, there is no consideration for lock escalation when the record's class is the root class.
```
3318   if (class_oid != NULL && !OID_IS_ROOTOID (class_oid))
3319     {
3320       /* instance lock request */
3321
3322       /* do lock escalation if it is needed and check if an implicit lock has been acquired. */
3323       ret_val = lock_escalate_if_needed (thread_p, class_entry, tran_index);
```

src/transaction/log_replication_atomic.cpp
joohok: I will add that

src/transaction/log_replication_atomic.cpp
wrlawodms: > the fact that replicator, which holds no locks at all, acquires IX_LOCK on the root class struck me as conceptually odd. In my understanding, having IX_LOCK on the root class implies holding X_LOCK on one or more entities (class or instance) beneath it. Hence, I hadn't considered it from your perspective.

Sorry to flip-flop, but I think it makes more sense than mine conceptually. As a lock owner, the replicator releases the IX_LOCK when it has no X_LOCK on a class. I didn't consider it this way. Although it is different from things on ATS (multiple IX_LOCK for each trx -> one IX_LOCK for the replicator), I think it has its point and am convinced. I'll leave it to your choice.

wrlawodms (approved):

Rudeus (approved):

PR #4828, opened 2023-12-05
[CBRD-25113] Add newlines to the diagdb output.

hyahong:
http://jira.cubrid.org/browse/CBRD-25113

Purpose
There is a part where a newline is missing in diagdb output.

Implementation
N/A

Remarks
N/A

### DIFF
--- a/src/storage/file_manager.c
+++ b/src/storage/file_manager.c
@@ -1421,6 +1421,9 @@ file_header_dump (THREAD_ENTRY * thread_p, const FILE_HEADER * fhead, FILE * fp)
 STATIC_INLINE void
 file_header_dump_descriptor (THREAD_ENTRY * thread_p, const FILE_HEADER * fhead, FILE * fp)
 {
+  BTID btid;
+  char *index_name = NULL;
+
   switch (fhead->type)
     {
     case FILE_HEAP:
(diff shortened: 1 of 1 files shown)


hornetmj (approved):

src/storage/file_manager.c
```diff
@@ -1468,6 +1468,7 @@ file_header_dump_descriptor (THREAD_ENTRY * thread_p, const FILE_HEADER * fhead,
     case FILE_DROPPED_FILES:
     case FILE_VACUUM_DATA:
     default:
+      fprintf (fp, "\n");
```
wrlawodms, line 1472: Hmm.. it seems that we need this also for the`FILE_BTREE` case when the `heap_get_indexinfo_of_btid()` returns an error.

wrlawodms (approved):

src/storage/file_manager.c
hyahong: Oh, I see. I missed a part. I will make the correction.

joohok (approved):

src/storage/file_manager.c
```diff
@@ -1443,10 +1443,11 @@ file_header_dump_descriptor (THREAD_ENTRY * thread_p, const FILE_HEADER * fhead,
 					&index_name, NULL) == NO_ERROR)
 	  {
 	    file_print_name_of_class (thread_p, fp, &fhead->descriptor.btree.class_oid);
-	    fprintf (fp, ", %s, ATTRID: %5d \n", index_name != NULL ? index_name : "*UNKNOWN-INDEX*",
+	    fprintf (fp, ", %s, ATTRID: %5d ", index_name != NULL ? index_name : "*UNKNOWN-INDEX*",
 		     fhead->descriptor.btree.attr_id);
 	  }
       }
+      fprintf (fp, "\n");
```
hornetmj, line 1450: Please update jira description and Expected/Actual result.

hornetmj (approved):

hornetmj:
@hyahong rebase your code to fix the ci test failures.

Rudeus (approved):

wrlawodms (approved):

src/storage/file_manager.c
wrlawodms:
```suggestion
	  }
      fprintf (fp, "\n");
      }
```

wrlawodms (approved):

src/storage/file_manager.c
hyahong: I think placing line breaks outside of the scope { ... } would be more intuitive since they are present in every case. However, does it seem better to put them inside the scope?

hyahong:
@hornetmj I performed a rebase

src/storage/file_manager.c
wrlawodms: It's a trivial one. So, I'm not that against your choice. That's my "approval" means.

If you want to know my view, the way I see it, the curly-braces are for the whole case and the `break` also belongs to the case. So, the `break` should be in the braces as well.

It's as things in https://google.github.io/styleguide/cppguide.html#Switch_Statements, although it's not about our case. Just see the formats.

wrlawodms (approved):

PR #4831, opened 2023-12-07
[CBRD-25070] Fix segfault when SP is called on SA mode

hgryoo:
http://jira.cubrid.org/browse/CBRD-25070

To get session_id, session_get_session_id () should be used.


beyondykk9 (approved):

hgryoo:
FYI, @hyunikn.

PR #4833, opened 2023-12-07
[CBRD-25067] ORDER BY causes row to be omitted with index and case expression

xmilex-git:
http://jira.cubrid.org/browse/CBRD-25067

When using order by, an error occurred where null values did not appear in the results even though expressions capable of null input existed as terms in the query, and an index scan was used. 

The revised code is a function that checks for the presence of nullable key expressions before creating order by plans. 
- It has been modified to not use index scan in order by or group by plans if there are nullable terms in the key expressions, by using the function qo_check_nullable_expr() to check for nullable terms in key_expr.

### DIFF
--- a/src/optimizer/query_planner.c
+++ b/src/optimizer/query_planner.c
@@ -9970,9 +9970,7 @@ qo_search_isnull_key_expr (PARSER_CONTEXT * parser, PT_NODE * tree, void *arg, i
 	  int nullable_terms = 0;
 	  qo_check_nullable_expr (parser, tree, &nullable_terms, NULL);
 	  /* this expr contains the key segment */
-	  if (tree->info.expr.op == PT_IS_NULL || tree->info.expr.op == PT_IS_NOT_NULL
-	      || tree->info.expr.op == PT_IFNULL || tree->info.expr.op == PT_NULLSAFE_EQ || (is_orderby
-											     && nullable_terms >= 1))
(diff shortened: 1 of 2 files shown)


src/optimizer/query_planner.c
```diff
@@ -9963,9 +9963,11 @@ qo_search_isnull_key_expr (PARSER_CONTEXT * parser, PT_NODE * tree, void *arg, i
       /* now check if the key segment is in there */
       if (bitset_intersects (&expr_segments, &key_segment))
 	{
+	  int nullable_terms = 0;
+	  qo_check_nullable_expr (parser, tree, &nullable_terms, NULL);
 	  /* this expr contains the key segment */
 	  if (tree->info.expr.op == PT_IS_NULL || tree->info.expr.op == PT_IS_NOT_NULL
-	      || tree->info.expr.op == PT_IFNULL || tree->info.expr.op == PT_NULLSAFE_EQ)
+	      || tree->info.expr.op == PT_IFNULL || tree->info.expr.op == PT_NULLSAFE_EQ || nullable_terms >= 1)
```
shparkcubrid: PT_IS_NOT_NULL is strange. How about fixing the qo_check_nullable_expr () function a bit and checking with just this?

shparkcubrid (approved):

youngjinj (approved):

PR #4836, opened 2023-12-11
[CBRD-25122] [Regression] When performing the pr_midxkey_add_prefix function, explicitly set nullmap

youngjinj:
http://jira.cubrid.org/browse/CBRD-24988
http://jira.cubrid.org/browse/CBRD-25122

When performing the pr_midxkey_add_prefix function, explicitly set nullmap.
In addition, critical typos were corrected.

### DIFF
--- a/src/object/object_primitive.c
+++ b/src/object/object_primitive.c
@@ -9056,6 +9056,7 @@ pr_midxkey_add_prefix (DB_VALUE * result, DB_VALUE * prefix, DB_VALUE * postfix,
 #if !defined(NDEBUG)
   for (i = 0; i < n_prefix; i++)
     {
+      /* Compressed columns must be null. */
       assert (or_multi_is_null (midx_postfix->buf, i));
     }
 #endif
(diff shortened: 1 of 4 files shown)


src/object/object_primitive.c
```diff
@@ -9056,15 +9056,18 @@ pr_midxkey_add_prefix (DB_VALUE * result, DB_VALUE * prefix, DB_VALUE * postfix,
 #if !defined(NDEBUG)
   for (i = 0; i < n_prefix; i++)
     {
+      /* Compressed columns must be null. */
       assert (or_multi_is_null (midx_postfix->buf, i));
     }
 #endif
 
+  i = n_prefix;
```
shparkcubrid, line 9064: How about meaningful name instead of 'i'

shparkcubrid (approved):

beyondykk9 (approved):

src/object/object_primitive.c
youngjinj: This is the index of the column currently being accessed.
`col_index` can be used instead of `i`.
But the function is not long. Isn't it okay to use `i`?

xmilex-git (approved):

PR #4837, opened 2023-12-12
[CBRD-25069] Implement memory monitor with memmon utility with MVP model

Rudeus:
http://jira.cubrid.org/browse/CBRD-25069

Implementation of memory monitoring manager with memmon utility with MVP model

- implementation of wrapper functions of allocation functions
 - new / delete
 - malloc / calloc / realloc / strdup / free
 - other allocation functions using in CUBRID (ex. db_private_alloc, ...)
- implementation of memory monitor
- implementation of memmon utility


Rudeus:
License checks and errors occurring in cppcheck are not due to this PR, but are errors stemming from the existing code. They can be disregarded and overlooked for the time being.

And the target of this PR is only linux/unix, not windows.

PR #4840, opened 2023-12-15
[LETS-469] Add a separate request for log HDR page

wrlawodms:
http://jira.cubrid.org/browse/LETS-469

~~Put the data page size and the log page size into the boot info which is gotten during TS's boot. This allows us things like ```cubrid createdb --log-page-size=4096 --db-page-size=8192 $DBNAME en_US```.~~

-- https://github.com/CUBRID/cubrid/pull/4840#issuecomment-1876751517 -->

The log page size can be different while booting up because it's not set properly yet on ATS. So, in this PR, I'm creating a new request for the log header page which accepts the length of the page. 
- Add functions for the log page header corresponding to `SEND_LOG_PAGE_FETCH, logpb_respond_fetch_log_page_request,  logpb_request_log_page_from_page_server`.
  - `SEND_LOG_HDR_PAGE_FETCH_ON_BOOT`, logpb_respond_fetch_log_hdr_page_on_boot_request, logpb_request_log_hdr_page_on_boot_from_page_server`
- Use them when requesting the hdr page on boot iso `SEND_LOG_PAGE_FETCH`.

I considered handling two cases in one function, `logpb_request_log_page_from_page_server` especially because they have the same error handling, but I've decided to separate them in that
- they use different requests
- it'd better to be explicit
- It could get divergent for each request.

Welcome to any feedback on this.

### DIFF
--- a/src/transaction/log_page_buffer.c
+++ b/src/transaction/log_page_buffer.c
@@ -2204,6 +2204,38 @@ logpb_respond_fetch_log_page_request (THREAD_ENTRY &thread_r, std::string &paylo
       payload_in_out.append (reinterpret_cast<const char *> (lr.get_page ()), LOG_PAGESIZE);
     }
 }
+
+void
+logpb_respond_fetch_log_hdr_page_request (THREAD_ENTRY &thread_r, std::string &payload_in_out)
+{
+  assert (is_page_server ());
+
+  log_lsa fetch_lsa { LOGPB_HEADER_PAGE_ID, 0 };
+  log_reader lr { LOG_CS_SAFE_READER };
+
+  // Make sure log page header is updated
+  logpb_force_flush_header_and_pages (&thread_r);
+
+  int error = lr.set_lsa_and_fetch_page (fetch_lsa);
+
+  // Response message
+  if (prm_get_bool_value (PRM_ID_ER_LOG_READ_LOG_PAGE))
+    {
+      _er_log_debug (ARG_FILE_LINE,
+		     "[READ LOG] Sending log page to Active Tran Server. Page ID: %lld Error code: %ld\n",
+		      LOGPB_HEADER_PAGE_ID, error);
+    }
+
+  // pack error first
+  payload_in_out = { reinterpret_cast<const char *> (&error), sizeof (error) };
+
+  if (error == NO_ERROR)
+    {
+      // pack page data too
+      payload_in_out.append (LOG_PAGESIZE, sizeof (LOG_PAGESIZE));
+      payload_in_out.append (reinterpret_cast<const char *> (lr.get_page ()), LOG_PAGESIZE);
+    }
+}
 // *INDENT-ON*
 #endif // SERVER_MODE
 
--- a/src/transaction/log_page_buffer.c
+++ b/src/transaction/log_page_buffer.c
@@ -376,6 +376,7 @@ static int logpb_append_prior_lsa_list (THREAD_ENTRY * thread_p, LOG_PRIOR_NODE 
 static int logpb_copy_page (THREAD_ENTRY * thread_p, LOG_PAGEID pageid, LOG_CS_ACCESS_MODE access_mode,
 			    LOG_PAGE * log_pgptr);
 static int logpb_request_log_page_from_page_server (LOG_PAGEID log_pageid, LOG_PAGE * log_pgptr);
+static int logpb_request_log_hdr_page_from_page_server (LOG_PAGE * log_pgptr);
 
 static void logpb_fatal_error_internal (THREAD_ENTRY * thread_p, bool log_exit, bool need_flush, const char *file_name,
 					const int lineno, const char *fmt, va_list ap);
@@ -1608,7 +1609,7 @@ logpb_fetch_header_from_page_server (LOG_HEADER * hdr, LOG_PAGE * log_pgptr)
 {
   assert (is_tran_server_with_remote_storage ());
 
-  int err = logpb_request_log_page_from_page_server (LOGPB_HEADER_PAGE_ID, log_pgptr);
+  int err = logpb_request_log_hdr_page_from_page_server (log_pgptr);
   if (err != NO_ERROR)
     {
       ASSERT_ERROR ();
@@ -2205,13 +2206,91 @@ logpb_respond_fetch_log_page_request (THREAD_ENTRY &thread_r, std::string &paylo
     }
 }
 
+static int
+logpb_request_log_hdr_page_from_page_server (LOG_PAGE * log_pgptr)
+{
+  // *INDENT-OFF*
+  const bool perform_logging = prm_get_bool_value (PRM_ID_ER_LOG_READ_LOG_PAGE);
+  if (perform_logging)
+    {
+      _er_log_debug (ARG_FILE_LINE, "[READ LOG] Sent request for log to Page Server. Page ID: %lld \n", LOGPB_HEADER_PAGE_ID);
+    }
+  std::string response_message;
+  int error_code = ts_Gl->send_receive (tran_to_page_request::SEND_LOG_HDR_PAGE_FETCH,
+                                           "", response_message);
+  // there are two layers of errors to he handled here:
+  //  - client side communication to page server error
+  //  - page server side errors
+
+  // client side communication to page server error
+  if (error_code != NO_ERROR)
+    {
+      ASSERT_ERROR ();
+      if (perform_logging)
+        {
+	  _er_log_debug (ARG_FILE_LINE, "[READ LOG] Received error log page message from Page Server. Error code: %d\n", error_code);
+        }
+
+      // TODO handling the case such as shutdown
+      assert_release (error_code != ER_CONN_NO_PAGE_SERVER_AVAILABLE);
+      return error_code;
+    }
+
+  assert (response_message.size () > 0);
+  const char *message_ptr = response_message.c_str ();
+  std::memcpy (&error_code, message_ptr, sizeof (error_code));
+  message_ptr += sizeof (error_code);
+
+  if (error_code == NO_ERROR)
+    {
+      PGLENGTH log_page_size = 0;
+      std::memcpy (&log_page_size, message_ptr, sizeof (LOG_PAGESIZE));
+
+      assert_release (log_page_size >= IO_MIN_PAGE_SIZE
+          && log_page_size <= IO_MAX_PAGE_SIZE
+          && IS_POWER_OF_2 (log_page_size));
+  
+      std::memcpy (log_pgptr, message_ptr, log_page_size);
+      message_ptr += log_page_size;
+
+      assert (LOGPB_HEADER_PAGE_ID == log_pgptr->hdr.logical_pageid);
+
+      if (perform_logging)
+	{
+	  _er_log_debug (ARG_FILE_LINE, "[READ LOG] Received log page message from Page Server. Page ID: %lld\n",
+			 log_pgptr->hdr.logical_pageid);
+	}
+    }
+  else
+    {
+      if (error_code == ER_LOG_PAGE_CORRUPTED)
+	{
+	  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_LOG_PAGE_CORRUPTED, 1, LOGPB_HEADER_PAGE_ID);
+	}
+      else
+	{
+	  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_GENERIC_ERROR, 0);
+	}
+      if (perform_logging)
+	{
+	  _er_log_debug (ARG_FILE_LINE, "[READ LOG] Received error log page message from Page Server. Error code: %d\n", error_code);
+	}
+    }
+  return error_code;
+  // *INDENT-ON*
+}
+
 void
-logpb_respond_fetch_log_hdr_page_request (THREAD_ENTRY &thread_r, std::string &payload_in_out)
+logpb_respond_fetch_log_hdr_page_request (THREAD_ENTRY & thread_r, std::string & payload_in_out)
 {
   assert (is_page_server ());
 
-  log_lsa fetch_lsa { LOGPB_HEADER_PAGE_ID, 0 };
-  log_reader lr { LOG_CS_SAFE_READER };
+  log_lsa fetch_lsa
+  {
+  LOGPB_HEADER_PAGE_ID, 0};
+  log_reader lr
+  {
+  LOG_CS_SAFE_READER};
 
   // Make sure log page header is updated
   logpb_force_flush_header_and_pages (&thread_r);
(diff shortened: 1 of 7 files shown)


joohok (commented):
It is understood that `log_get_io_page_size()` reads the active log header information and then determines the `IO page size`. It seems redundant to set the IO page size during booting to support this function (for setting the IO page size), resulting in repetitive tasks. While it seems beneficial to eliminate this redundancy, I don't have ideas..

One option is to add an exception to the assertion, but can not recommend(e.g., 
`assert(log_pageid == LOGPB_HEADER_PAGE_ID || 
message_ptr == (response_message.c_str() + response_message.size()));)`,

cristiarg:
> It is understood that log_get_io_page_size() reads the active log header information and then determines the IO page size. It seems redundant to set the IO page size during booting to support this function (for setting the IO page size), resulting in repetitive tasks. While it seems beneficial to eliminate this redundancy, I don't have ideas..

In `boot_restart_server`:
- `init_server_type` - which calls `boot` for transaction servers
- [[ we need to introduce retrieval of active log header page here, on a "transaction server with remote storage" ]]
- `log_get_io_page_size` - first use
- `pgbuf_initialize`
- the rest of the modules, also `log_initialize`

Retrieving the boot info happens before the log header page - `LOGPB_HEADER_PAGE_ID` - is done.
Can we force the retrieval of page `LOGPB_HEADER_PAGE_ID` right after the server type is initialized?

Side note - it does help that these are macros instead of inline functions and, thus cannot be made to guard ahead-of-init access. But maybe not actually needed.
```
#define LOG_PAGESIZE            db_Log_page_size
#define IO_PAGESIZE             db_Io_page_size
#define DB_PAGESIZE             db_User_page_size
```

wrlawodms:
@cristiarg 
> Can we force the retrieval of page LOGPB_HEADER_PAGE_ID right after the server type is initialized?

I don't understand the point. I guess I'm missing something. Does it remove the redundant fetch that jooho mentioned? The fetch will be done in the `log_get_io_page_size` and I don't catch the point to move up the fetch. Could you elaborate on it a little bit more?

wrlawodms:
@joohok 

> It is understood that log_get_io_page_size() reads the active log header information and then determines the IO page size. It seems redundant to set the IO page size during booting to support this function (for setting the IO page size), resulting in repetitive tasks. While it seems beneficial to eliminate this redundancy, I don't have ideas..

I agree that it's redundant for the page size. I thought about skipping the `log_get_io_page_size()`, but it actually does more than just setting the page size, so I decided to just keep it. 

> One option is to add an exception to the assertion, but can not recommend(e.g.,
assert(log_pageid == LOGPB_HEADER_PAGE_ID ||  message_ptr == (response_message.c_str() + response_message.size()));),

It could be a way, but I don't want to let `logpb_request_log_page_from_page_server` regard them differently.
Welcome any other ideas..

joohok (approved):
To eliminate redundancy, it seems necessary to retrieve active log header information from the page server and set log_Gl.hdr before calling `log_get_io_page_size()`. However, it seems awkward that this log_Gl value is initialized before `log_initialize()`. Therefore, the current approach seems to be the best

cristiarg:
> To eliminate redundancy, it seems necessary to retrieve active log header information from the page server and set log_Gl.hdr before calling log_get_io_page_size(). However, it seems awkward that this log_Gl value is initialized before log_initialize().

@joohok 
Can you please elaborate?

cristiarg:
@wrlawodms 

> I don't understand the point. I guess I'm missing something. Does it remove the redundant fetch that jooho mentioned? The fetch will be done in the log_get_io_page_size and I don't catch the point to move up the fetch. Could you elaborate on it a little bit more?

I was also not clear on this. Sorry.

Explanation:
Expanding further from [what I undestand from] @joohok comments, I think a solution to eliminating this redundancy is to:
- customize/extend the `logpb_fetch_header_from_page_server` implementation to not crash no matter what the size of the log page size is on the database opened by the page server
- this means - maybe - to create a new specific request into the page server - separate from reusing `SEND_LOG_PAGE_FETCH` request - i.e. `SEND_LOG_HEADER_PAGE_FETCH`
- `SEND_LOG_PAGE_FETCH` will only remain to be used by all the rest of the log pages requests (we can even add an assert in `logpb_respond_fetch_log_page_request` that the log page id is not `LOGPB_HEADER_PAGE_ID`)

This might be needed because, aside from log page size and IO page size, I see that the function `logpb_find_header_parameters` - `logpb_fill_header_parameters` also deals with compatibility number and charset (which is probably of lesser importance but it would be good to deal with that as well).

Other things:
- maybe better to close http://jira.cubrid.org/browse/LETS-469 and re-open http://jira.cubrid.org/browse/LETS-831 (which is more detailed);
- or, at least, mention both in the PR message
- have you also tested the scenario for PTS?

wrlawodms:
@cristiarg  About creating a dedicated request for the page volume header page, my concern is that it is the case only while booting up. After booting, the size of log pages including the header page is the same and other codes assume it. Do we need to create a dedicated request for the only case and assume that the page size can be different? I think it makes sense it's retrieved with the BOOTUP msg once during boot sequence. Furthermore, it would make even more sense that it requests the log hdr page expecting it to have a proper size.

My suggestion is
- Accept the redundancy as it is, or
- Add a new request to retrieve `log_header` instead of `log hdr page`. Only the `log_header` is used for the function where this problem occurs. In this case, I'll change the `logpb_fetch_header_from_page_server` only to return `hdr` and use the new request.

What do you think? @joohok @cristiarg 

> have you also tested the scenario for PTS?

Yes

> maybe better to close http://jira.cubrid.org/browse/LETS-469 and re-open http://jira.cubrid.org/browse/LETS-831 (which is more detailed);

I think the link between them would be enough. It seems the same to me. I'll leave it to @hornetmj 's choice.

cristiarg:
@wrlawodms 

> my concern is that it is the case only while booting up.

That's precisely the reason I suggested this.
The function will be implemented to be able to retrieve an entire log page regardless of the size (eg: will contain information about the result of the call - error code - size of the retrieved page, and the log header page size itself).

My suggestion is motivated by:
- the fact that it would already be inline with the current boot structure and does not incur any changes (it just fixes the "weak-link" of `logpb_fetch_header_from_page_server` not working with when the default 16k log page size does not match the actual page size of the database).
- I see no harm in creating a specific request just for the log header page - it has zero-performance cost 
- tries to keep clutter/spaghetti to a minimum - the booting up of monolithic/ATS/PTS/PS flavors from the same code-base already has some exceptions (see `boot_restart_server` - `is_passive_transaction_server`/`log_initialize_passive_tran_server` - `!is_passive_transaction_server` / `log_initialize` )
  - with this new change the way it is, we add yet another step to this exception

I approved. :)

cristiarg (approved):

wrlawodms:
@cristiarg OK. I'll follow yours. I'm convinced by `tries to keep clutter/spaghetti to a minimum `. Please re-review it once I'm done.

src/transaction/log_page_buffer.c
```diff
@@ -2204,6 +2191,112 @@ logpb_respond_fetch_log_page_request (THREAD_ENTRY &thread_r, std::string &paylo
...
+  assert (response_message.size () > 0);
+  const char *message_ptr = response_message.c_str ();
+  std::memcpy (&error_code, message_ptr, sizeof (error_code));
+  message_ptr += sizeof (error_code);
+
+  if (error_code == NO_ERROR)
+    {
+      PGLENGTH log_page_size = 0;
+      std::memcpy (&log_page_size, message_ptr, sizeof (LOG_PAGESIZE));
+      message_ptr += sizeof (LOG_PAGESIZE);
```
```suggestion
      std::memcpy (&log_page_size, message_ptr, sizeof (PGLENGTH));
      message_ptr += sizeof (PGLENGTH);
```

joohok (approved):

src/transaction/log_page_buffer.c
```diff
@@ -2203,8 +2193,118 @@ logpb_respond_fetch_log_page_request (THREAD_ENTRY &thread_r, std::string &paylo
       // pack page data too
       payload_in_out.append (reinterpret_cast<const char *> (lr.get_page ()), LOG_PAGESIZE);
     }
+  // *INDENT-ON*
+}
+
+static int
+logpb_request_log_hdr_page_from_page_server (LOG_PAGE * log_pgptr)
```
cristiarg: A small header comment for the function describing the rationale.

src/transaction/log_page_buffer.c
cristiarg: or, at least, `sizeof (log_page_size)`.

src/transaction/log_page_buffer.c
```diff
@@ -2339,27 +2439,39 @@ logpb_read_page_from_file_or_page_server (THREAD_ENTRY * thread_p, LOG_PAGEID pa
   const SERVER_TYPE server_type = get_server_type ();
   if (server_type == SERVER_TYPE_TRANSACTION && ts_Gl->is_page_server_connected ())
     {
+      // *INDENT-OFF*
+      auto page_request_func = [pageid](LOG_PAGE* log_pgptr) -> int
```
cristiarg: I do not understand why this indirection is still needed.

Do requests still come for `LOGPB_HEADER_PAGE_ID` via `logpb_read_page_from_file_or_page_server` function ?

There are still 2 calls to
```
    logpb_read_page_from_file_or_page_server (thread_p, LOGPB_HEADER_PAGE_ID, LOG_CS_FORCE_USE, temp_pgptr);
```
in the function `log_create_internal` which I have added some time back.
The calls are guarded by `#if defined(CUBRID_DEBUG)` - thus, not executed by default.
I now think that those are incorrect and should be replaced back to `logpb_read_page_from_file` because the function `log_create_internal` should not be called on a non-PS?
I did not debug to check this.

Rudeus (approved):

src/transaction/log_page_buffer.c
```diff
@@ -2203,8 +2193,118 @@ logpb_respond_fetch_log_page_request (THREAD_ENTRY &thread_r, std::string &paylo
...
+  // *INDENT-ON*
+}
+
+static int
+logpb_request_log_hdr_page_from_page_server (LOG_PAGE * log_pgptr)
+{
+  const bool perform_logging = prm_get_bool_value (PRM_ID_ER_LOG_READ_LOG_PAGE);
+  if (perform_logging)
+    {
+      _er_log_debug (ARG_FILE_LINE, "[READ LOG] Sent request for log to Page Server. Page ID: %lld \n",
```
hornetmj, line 2215: Nit:

Sent request for log to Page Server. Page ID: %lld \n
-> Sent request for log header page to Page Server \n

src/transaction/log_page_buffer.c
```diff
@@ -2203,8 +2193,118 @@ logpb_respond_fetch_log_page_request (THREAD_ENTRY &thread_r, std::string &paylo
...
+		     LOGPB_HEADER_PAGE_ID);
+    }
+  std::string response_message;
+  int error_code = ts_Gl->send_receive (tran_to_page_request::SEND_LOG_HDR_PAGE_FETCH,
+					"", response_message);
+  // there are two layers of errors to he handled here:
```
hornetmj:
```suggestion
  // there are two layers of errors to be handled here:
```

src/transaction/log_page_buffer.c
```diff
@@ -2203,8 +2193,118 @@ logpb_respond_fetch_log_page_request (THREAD_ENTRY &thread_r, std::string &paylo
...
+    }
+  else
+    {
+      if (error_code == ER_LOG_PAGE_CORRUPTED)
+	{
+	  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_LOG_PAGE_CORRUPTED, 1, LOGPB_HEADER_PAGE_ID);
+	}
+      else
+	{
+	  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_GENERIC_ERROR, 0);
```
hornetmj, line 2273: In this case, it would be good if we can get error code in the error log.

src/transaction/log_page_buffer.c
```diff
@@ -2203,8 +2193,118 @@ logpb_respond_fetch_log_page_request (THREAD_ENTRY &thread_r, std::string &paylo
...
+	{
+	  _er_log_debug (ARG_FILE_LINE, "[READ LOG] Received error log page message from Page Server. Error code: %d\n",
+			 error_code);
+	}
+    }
+  return error_code;
+}
+
+void
+logpb_respond_fetch_log_hdr_page_request (THREAD_ENTRY & thread_r, std::string & payload_in_out)
```
hornetmj: Nit:

How about using thread_p iso thread_r for code consistency?

hornetmj (approved):

src/transaction/log_page_buffer.c
cristiarg: I agree with the consistency remark. I am to "blame" since I wrote the initial `logpb_respond_fetch_log_page_request` sister function.

Question:
I know that the THREAD_ENTRY was added relatively recently - sometime around 2015-16 maybe? - and I want to know if it is safe to assume that, when `SERVER_MODE` is active, the `THREAD_ENTRY` is also assumed to be present no matter what/when/where. I know that there are few functions where there is still this sequence
```
  if (thread_p == NULL)
    {
      thread_p = thread_get_thread_entry_info ();
    }
```
but only ~80 occurrences among the thousands of functions that cubrid has defined (both module-public and module-private).

wrlawodms: > and I want to know if it is safe to assume that, when SERVER_MODE is active, the THREAD_ENTRY is also assumed to be present to matter what/when/where

My understanding is that `thread_get_thread_entry_info` returns not-null thread entry only when the thread calling it has its own entry from `cubthread::manager`. That is stored in the thread's local variable. This works only when it's a daemon or worker thread of a worker pool spawned by the `cubthread::manager`. So, if a thread is created just by `pthread` or `std::thread` bypassing the manager, it doesn't have a valid entry. Does this answer you?

> bot only ~80 occurrences among the thousands of functions that cubrid has defined

has defined what?

cristiarg: > bot only ~80 occurrences among the thousands of functions that cubrid has defined

just that; in Romanian, when translated ad-litteram, it makes perfect sense :)

Trying to rephrase:
but there are only ~80 occurrences among the thousands of functions that cubrid codebase is composed of

src/transaction/log_page_buffer.c
wrlawodms: I run into some cases the HDR log page is retrieved with this function. For example, 
`logpb_fetch_header-> logpb_fetch_header_with_buffer -> ... -> logpb_read_page_from_file_or_page_server`.
 And, I just fixed this function to handle the case naively. However, yes, if a HDR page is requested it should come through `logpb_hdr_read_page_from_file_or_page_server`, which is for the case. I should've, but I missed.

Having analyzed further, there are two cases for getting the HDR log page on the 'monolithic' server:
- On Running: `logpb_fetch_header_with_buffer`, which gets it through the general log page read interface, `logpb_fetch_page`.
- On Recovery: `logpb_fetch_header_from_file`

With the modification in this PR, the `logpb_hdr_read_page_from_file_or_page_server` uses `logpb_fetch_header_from_file`. It can't be used for the `logpb_fetch_header` case. I should come up with a new idea taking this into account.

The `logpb_fetch_header_with_buffer` is used in a few not general cases, but I think the HDR page should be handled like a normal log page when the recovery is done. I'm not sure, though. I'll get back with a new solution.

src/transaction/log_page_buffer.c
cristiarg: > My understanding is that thread_get_thread_entry_info returns not-null thread entry only when the thread calling it has its own entry from cubthread::manager. That is stored in the thread's local variable. This works only when it's a daemon or worker thread of a worker pool spawned by the cubthread::manager. So, if a thread is created just by pthread or std::thread bypassing the manager, it doesn't have a valid entry. Does this answer you?

Yes. Somehow I should have already know this - after all, it is in the code. :)
So, this leads me to think that using the reference instead of the pointer has no drawbacks in this particular situation.

Side note, the passing of the thread entry as argument down the callstack has performance as its main benefit because retrieving a `thread_local` variable in each place it is needed is much more costly than accessing a stack variable (as it boils down to when passed as an argument).

wrlawodms:
I've made `SEND_LOG_HDR_PAGE_FETCH` and related ones used only on boot and added a suffix like `SEND_LOG_HDR_PAGE_FETCH_ON_BOOT`. 

The added msg and stuff are only used on boot as the `logpb_fetch_header_from_file_or_page_server` is originally supposed to be. Otherwise, the general request, `logpb_read_page_from_file_or_page_server`, is used. This is just to follow existing mechanism and add a request for the log hdr page on boot case.

`logpb_request_log_hdr_page_on_boot_from_page_server` can be used not only on boot, but on run. However, there is no point using this since it costs more (length).

PR #4842, opened 2023-12-18
[CBRD-25128] In STRING_CAT '||' operation, incorrect handling of null when setting oracle_style_empty_string

beyondykk9:
http://jira.cubrid.org/browse/CBRD-25128

When oracle_style_empty_string is set, empty string '' is treated as null. And NULL is treated as NULL, but STRING_CAT operations such as '||' must be treated as empty string ''. Because ORACLE has NULL and || This is because the operation result is not NULL but an empty string.

### DIFF
--- a/src/executables/csql_result.c
+++ b/src/executables/csql_result.c
@@ -485,8 +485,6 @@ display_empty_result (int stmt_type, int line_no)
   return;
 }
 
-#define IS_STRING_TYPE(type)     (type == DB_TYPE_STRING || \
-                                  type == DB_TYPE_CHAR)
 /*
  * get_current_result() - get the attribute values of the current result
  *   return: pointer newly allocated value array. On error, NULL.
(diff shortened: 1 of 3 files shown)


src/executables/csql_result.c
```diff
@@ -559,7 +561,8 @@ get_current_result (int **lengths, const CUR_RESULT_INFO * result_info, const CS
       assert (value_type == DB_TYPE_NULL
 	      /* UNKNOWN, maybe host variable */
 	      || result_info->attr_types[i] == DB_TYPE_NULL || result_info->attr_types[i] == DB_TYPE_VARIABLE
-	      || value_type == result_info->attr_types[i]);
+	      || value_type == result_info->attr_types[i]
+	      || (IS_STRING_TYPE (value_type) && IS_STRING_TYPE (result_info->attr_types[i])));
```
hyunikn: Does this assertion fail without addition of this line?

beyondykk9: By changing the return from null type node to string type, assert should also reflect this.

src/executables/csql_result.c
```diff
@@ -485,6 +485,8 @@ display_empty_result (int stmt_type, int line_no)
   return;
 }
 
+#define IS_STRING_TYPE(type)     (type == DB_TYPE_STRING || \
+                                  type == DB_TYPE_CHAR)
```
ctshim: Is there a reason to define it separately rather than using PT_IS_STRING_TYPE()?
I am concerned that I am missing processing for the VARCHAR type.

beyondykk9: Good point. I fixed to to use TP_IS_CHAR_TYPE defined at "object_domain.h".

hyunikn (approved):

ctshim (approved):

PR #4848, opened 2023-12-28
[CBRD-25139] When using oracle style empty string, the result of replace is null.

beyondykk9:
http://jira.cubrid.org/browse/CBRD-25139

When setting oracle style empty string=y, the result is null when using syntax such as replace(?,'-','').

CREATE TABLE tbl(i VARCHAR(100));
INSERT INTO tbl SELECT '2013-01-01' FROM dual;
SELECT  REPLACE(i,'-','') FROM tbl;

Expected Result
```
  replace(i, '-', '')
======================
  '20130101' 
```

Actual Result:
```
  replace(i, '-', '')
======================
  NULL
```

### DIFF
--- a/src/parser/type_checking.c
+++ b/src/parser/type_checking.c
@@ -18253,19 +18253,22 @@ pt_fold_const_expr (PARSER_CONTEXT * parser, PT_NODE * expr, void *arg)
       if (domain
 	  && (QSTR_IS_ANY_CHAR_OR_BIT (TP_DOMAIN_TYPE (domain)) || type1 == PT_TYPE_NULL || type2 == PT_TYPE_NULL))
 	{
-	  if (opd3 && opd3->node_type == PT_VALUE && type3 == PT_TYPE_NULL)	/* REPLACE */
+	  if (opd3)		/* REPLACE */
 	    {
-	      /* need to replace the NULL with empty string */
-	      int err =
-		db_string_make_empty_typed_string (&opd3->info.value.db_value, DB_TYPE_STRING, DB_DEFAULT_PRECISION,
-						   TP_DOMAIN_CODESET (domain), TP_DOMAIN_COLLATION (domain));
-	      if (err != NO_ERROR)
+	      if (opd3->node_type == PT_VALUE && type3 == PT_TYPE_NULL)
 		{
-		  has_error = true;
-		}
-	      else
-		{
-		  result = expr;
+		  /* need to replace the NULL with empty string */
+		  int err =
+		    db_string_make_empty_typed_string (&opd3->info.value.db_value, DB_TYPE_STRING, DB_DEFAULT_PRECISION,
+						       TP_DOMAIN_CODESET (domain), TP_DOMAIN_COLLATION (domain));
+		  if (err != NO_ERROR)
+		    {
+		      has_error = true;
+		    }
+		  else
+		    {
+		      result = expr;
+		    }
 		}
 	    }
 	  else if (opd1 && opd1->node_type == PT_VALUE && type1 == PT_TYPE_NULL)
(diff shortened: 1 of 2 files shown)


src/parser/type_checking.c
```diff
@@ -18237,17 +18237,31 @@ pt_fold_const_expr (PARSER_CONTEXT * parser, PT_NODE * expr, void *arg)
...
       if (domain
 	  && (QSTR_IS_ANY_CHAR_OR_BIT (TP_DOMAIN_TYPE (domain)) || type1 == PT_TYPE_NULL || type2 == PT_TYPE_NULL))
 	{
-	  if (opd1 && opd1->node_type == PT_VALUE && type1 == PT_TYPE_NULL)
+	  if (opd3 && opd3->node_type == PT_VALUE && type3 == PT_TYPE_NULL)	/* REPLACE */
+	    {
+	      result = parser_copy_tree (parser, opd1);
+	      if (result == NULL)
+		{
+		  has_error = true;
```
hyunikn: no 'goto end;' after this line?

beyondykk9: Good point.

src/parser/type_checking.c
```diff
@@ -18237,24 +18237,37 @@ pt_fold_const_expr (PARSER_CONTEXT * parser, PT_NODE * expr, void *arg)
...
+	}
+
       /* use the caching variant of this function ! */
       domain = pt_xasl_node_to_domain (parser, expr);
 
       if (domain
 	  && (QSTR_IS_ANY_CHAR_OR_BIT (TP_DOMAIN_TYPE (domain)) || type1 == PT_TYPE_NULL || type2 == PT_TYPE_NULL))
 	{
-	  if (opd1 && opd1->node_type == PT_VALUE && type1 == PT_TYPE_NULL)
+	  if (opd3 && opd3->node_type == PT_VALUE && type3 == PT_TYPE_NULL)	/* REPLACE */
```
hyunikn: Adding this conditional block results in 
```
csql> select replace('abc', 'a', null);

   replace('abc', 'a', null)
======================
  'abc'
```
while the current result is 
```
csql> select replace('abc', 'a', null);

   replace('abc', 'a', null)
======================
  NULL                
```
Note the Oracle's result of 
```
SELECT replace('abc', 'a', null) FROM dual;
```
is 'bc'.

beyondykk9: Good point, I fixed it.

src/parser/type_checking.c
```diff
@@ -18237,24 +18237,44 @@ pt_fold_const_expr (PARSER_CONTEXT * parser, PT_NODE * expr, void *arg)
...
+	      if (err != NO_ERROR)
+		{
+		  has_error = true;
+		}
+	      else
+		{
+		  result = expr;
+		}
+	    }
+	  else if (opd1 && opd1->node_type == PT_VALUE && type1 == PT_TYPE_NULL)
```
hyunikn, line 18274: "select replace(null, 'a', 'A');" results in a wrong value 'a' due to this conditional block.
Separate handling of REPLACE seems necessary.

src/query/fetch.c
```diff
@@ -1551,7 +1551,8 @@ fetch_peek_arith (THREAD_ENTRY * thread_p, REGU_VARIABLE * regu_var, val_descr *
 	      goto error;
 	    }
 	}
-      if (DB_IS_NULL (peek_left) || (peek_third && DB_IS_NULL (peek_third)))
+      if (!prm_get_bool_value (PRM_ID_ORACLE_STYLE_EMPTY_STRING)
+	  && (DB_IS_NULL (peek_left) || (peek_third && DB_IS_NULL (peek_third))))
```
hyunikn:
```suggestion
	  && (DB_IS_NULL (peek_left) || DB_IS_NULL (peek_right) || (peek_third && DB_IS_NULL (peek_third))))
```
just for readability.

src/parser/type_checking.c
beyondykk9: Good point, I fixed it.

ctshim (approved):

hyunikn (approved):

PR #4855, opened 2024-01-04
[LETS-863] add system parameter and use it to instantiate transaction descriptor on PTS

cristiarg:
http://jira.cubrid.org/browse/LETS-863

Add a new system parameter - `pts_log_replication_withheld_lock_timeout` in units of time - and use it to instantiate the transaction descriptor corresponding to the system transaction on passive transaction server.

On a Passive Transaction Server, the system transaction descriptor (at index `0`) is used exclusively (as of now) by the replication thread to identify itself for various actions within the engine.
One areas is the lock manager (for now, we only have only one thread because the log replication on a Passive Transaction Server is single threaded) will use the system.
The lock manager will interrogate the transaction descriptor to decide how to act in case of:
- lock deadlock - eg: `lock_select_deadlock_victim`
- lock timeout - eg: `lock_check_timeout_expired_and_count_suspended_mapfunc`

The patch has two aspects:
- the values `tdes->has_deadlock_priority` and `tdes->num_log_records_written` are used by the deadlock selection logic in `lock_select_deadlock_victim`
- the value `tdes->wait_msecs` is initialized with the value of the newly introduced `pts_log_replication_withheld_lock_timeout` system parameter will be used by the logic in subsequent http://jira.cubrid.org/browse/LETS-864 .


joohok (approved):

wrlawodms (commented):
When the time's up, are other client lock holders kicked out? I'm asking to remind.

wrlawodms (approved):

cristiarg:
> When the time's up, are other client lock holders kicked out? I'm asking to remind.

That is the final intention.
But it is not part of this patch. It will be part of the modifications added with http://jira.cubrid.org/browse/LETS-864 .

I have added more details to the PR description. Let me know if it answers your question.

wrlawodms:
@cristiarg 
> I have added more details to the PR description. Let me know if it answers your question.

More than enough. Thanks.

PR #4857, opened 2024-01-05
[LETS-849] Prevent PTS transactions from viewing DDL changes before they are committed

joohok:
http://jira.cubrid.org/browse/LETS-849

Describe here


joohok:
Move to the new repository

PR #4861, opened 2024-01-08
[CBRD-25145] Problem of removing expr from sort spec of analytic function when statement pooling

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-25145

When creating xasl with a rewrite query, the expression pointed to by sort_spec changed to PT_VALUE cannot be found.
In to-be, the previous expression is backed up in prev_expr of sort_spec.

### DIFF
--- a/src/parser/parse_tree.h
+++ b/src/parser/parse_tree.h
@@ -3017,6 +3017,7 @@ struct pt_killstmt_info
 struct pt_sort_spec_info
 {
   PT_NODE *expr;		/* PT_EXPR, PT_VALUE, PT_NAME */
+  PT_NODE *prev_expr;		/* PT_EXPR, PT_VALUE, PT_NAME */
   QFILE_TUPLE_VALUE_POSITION pos_descr;	/* Value position descriptor */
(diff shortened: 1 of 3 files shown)


src/parser/parse_tree.h
```diff
@@ -3017,6 +3017,7 @@ struct pt_killstmt_info
 struct pt_sort_spec_info
 {
   PT_NODE *expr;		/* PT_EXPR, PT_VALUE, PT_NAME */
+  PT_NODE *prev_expr;		/* PT_EXPR, PT_VALUE, PT_NAME */
```
youngjinj, line 3020: In pt_init_sort_spec, prev_expr initialization is required.

youngjinj: I checked that memset is being called in parser_init_node.

shparkcubrid: It is already initialized in parser_init_node().

youngjinj:
[CUBRIDSUS-16864](http://jira.cubrid.org/browse/CUBRIDSUS-16864) has already been merged, so close this PR.

PR #4866, opened 2024-01-10
[LETS-864] Implement PTS log replication prioritization and client read-only transaction unilaterally abort

cristiarg:
http://jira.cubrid.org/browse/LETS-864

Introduce a mechanism that give priority to the log replication thread (aka. system transaction) on a Passive Transaction Server when lock collisions occur between the log replication thread and user read-only transactions threads.

The patch is made up of few physically separated but functionally bound parts:
1. a new lock manager infrastructure to keep track of OID's locked by the log replication transaction/thread:
  - the `lock_log_replication_withheld_..` structures and functions
2. lock manager implementation for the log replication thread prioritization in case of class lock conflicts with user transactions:
  - lock manager `lock_object` and `lock_internal_perform_lock_object` functions
3. logic to unilaterally abort read-only transactions that hold conflicting locks with the log replication transaction:
  - lock manger: `logtb_is_passive_transaction_server_log_replication_transaction`
  - log manager: `log_abort_read_only_on_pts`, `log_abort_local_read_only_on_pts`
4. logic to check if a new request of a client transaction occurs after the transaction has been unilaterally aborted
  - in `net_server_request`

Explanation 1) :
- the mechanism implemented by `lock_log_replication_withheld_..` is in essence an anti-starvation mechanism that gives priority to the log replication transaction
- the need for this is that the locking performed by the log replication is not atomic within the internal implementation of the lock manager:
  - when requesting a schema modification lock for a class OID in `lock_object` and the object is already locked by a user transaction, the implementation in `lock_internal_perform_lock_object` will return `LK_NOTGRANTED_DUE_TIMEOUT` - because it is already assumed and implied that the replication transaction does not have infinite lock wait - see implementation for http://jira.cubrid.org/browse/LETS-863
  - but, while in `lock_internal_perform_lock_object` the replication transaction will already register its intention to lock the class OID in the `lock_log_replication_withheld_..` mechanism
  - once in `lock_object` again, the replication transaction will abort all user transactions for the class OID - in `lock_interrupt_and_free_locks_for_all_holders`
  - and will goto and retry to acquire the lock for the class OID; this time succeeding to
  - once the lock for the class OID is obtained, the intention is removed with `lock_log_replication_withheld_clear_all`

The Jira issue contains comments where various test scenarios and the expected result are detailed:
- `CREATE/ALTER TABLE` and one single PTS read-only transaction
- `CREATE/ALTER VIEW` and one single PTS read-only transaction
- `CREATE/ALTER INDEX` and one single PTS read-only transaction
- `ALTER TABLE` and 2 PTS user read-only transactions that simulate newly occurring transaction being forbidden to acquire locks for which the replication transaction expressed "intent" to lock
  - this is accomplished with the `lock_log_replication_withheld_..` mechanism


src/thread/thread_entry.hpp
```diff
@@ -304,9 +312,14 @@ namespace cubthread
 
       void return_lock_free_transaction_entries (void);
 
+    private:
+      friend void ::thread_lock_entry (::cubthread::entry *thread_p);
```
cristiarg, line 316: These changes helped during investigation for this PR in the following way: I wanted to make sure that the `lock`, `unlock` function are never called directly. Thus, I made them private and allowed `::thread_lock_entry` and `::thread_unlock_entry` as friend functions to be able to call them.
Are unrelated to the rest of the PR and I am willing to split them to a separate PR if so advised by reviewers.

src/transaction/lock_manager.c
```diff
@@ -2921,7 +2941,7 @@ lock_grant_blocked_waiter_partial (THREAD_ENTRY * thread_p, LK_RES * res_ptr, LK
  *
  */
 static bool
-lock_check_escalate (THREAD_ENTRY * thread_p, LK_ENTRY * class_entry, LK_TRAN_LOCK * tran_lock)
+lock_check_escalate (THREAD_ENTRY * thread_p, const LK_ENTRY * class_entry, const LK_TRAN_LOCK * tran_lock)
```
cristiarg, line 2941: These `const` additions are also not needed for the proper functioning of the test and I am willing to remove them.
Investigating the CUBRID codebase is hard giving that const correctness is not made use of and the vast majority of the pointer arguments of the functions which could be `pointer to const` are not.

src/transaction/lock_manager.c
```diff
@@ -3231,12 +3251,12 @@ lock_internal_hold_lock_object_instant (THREAD_ENTRY * thread_p, int tran_index,
...
 static int
-lock_internal_perform_lock_object (THREAD_ENTRY * thread_p, int tran_index, const OID * oid, const OID * class_oid,
-				   LOCK lock, int wait_msecs, LK_ENTRY ** entry_addr_ptr, LK_ENTRY * class_entry)
+lock_internal_perform_lock_object (THREAD_ENTRY * thread_p, const int tran_index, const OID * const oid,
+				   const OID * const class_oid, const LOCK lock, const int wait_msecs,
+				   LK_ENTRY ** entry_addr_ptr, LK_ENTRY * class_entry)
 {
   LF_TRAN_ENTRY *t_entry_ent = thread_get_tran_entry (thread_p, THREAD_TS_OBJ_LOCK_ENT);
   LK_RES_KEY search_key;
-  TRAN_ISOLATION isolation;
```
cristiarg, line 3239: This was not used at all.

cristiarg:
manually moved to new CUBRID organization

PR #4869, opened 2024-01-11
[CBRD-24046] Result cache for correlated scalar subquery

xmilex-git:
http://jira.cubrid.org/browse/CBRD-24046

A correlated subquery is executed for each row of the main query to get a value. If there are many duplicated inputs, the same query is repeatedly executed, which may cause a decrease in performance. We eliminate the cause of the slowdown through the result cache.


src/query/query_dump.c
```diff
@@ -3274,6 +3275,13 @@ qdump_print_stats_text (FILE * fp, xasl_node * xasl_p, int indent)
   qdump_print_stats_text (fp, xasl_p->scan_ptr, indent);
   qdump_print_stats_text (fp, xasl_p->connect_by_ptr, indent);
 
+  if (xasl_p->sq_cache_flag & SQ_CACHE_INITIALIZED_FLAG && !(xasl_p->sq_cache_flag & SQ_CACHE_NOT_CACHING_FLAG))
```
shparkcubrid, line 3278: How about using XASL.flag and XASL_IS_FLAGED(x, f).

src/query/stream_to_xasl.c
```diff
@@ -242,6 +242,7 @@ stx_map_stream_to_xasl (THREAD_ENTRY * thread_p, xasl_node ** xasl_tree, bool us
   xasl->class_oid_list = NULL;
   xasl->class_locks = NULL;
   xasl->tcard_list = NULL;
+  xasl->sq_cache_flag = 0;
```
shparkcubrid: Isn't initialization redundant?

xmilex-git: Upon checking, it was confirmed that sq_cache_flag is properly initialized even without initialization in that part. Therefore, I will revert the changes in that section back to their original state.

src/query/subquery_cache.c
```diff
@@ -0,0 +1,749 @@
...
+    {
+      regu_var->xasl->sq_cache_flag |= SQ_CACHE_NOT_CACHING_FLAG;
+      return false;
+    }
+  if (!(xasl->sq_cache_flag & SQ_CACHE_INITIALIZED_FLAG))
+    {
+      sq_cache_initialize (xasl);
+    }
+
+  ret = (sq_val *) mht_get (xasl->sq_cache_ht, key);
```
youngjinj: It is meaningless to run `mht_get` immediately after `sq_cache_initialize`.

src/query/subquery_cache.c
```diff
@@ -0,0 +1,749 @@
...
+	    {
+	      regu_var->xasl->status = XASL_SUCCESS;
+	      return true;
+	    }
+	  EXECUTE_REGU_VARIABLE_XASL (thread_p, regu_var, (val_descr *) vd);
+	  if (CHECK_REGU_VARIABLE_XASL_STATUS (regu_var) != XASL_SUCCESS)
+	    {
+	      return false;
+	    }
+	  sq_put (regu_var->xasl, regu_var);
```
youngjinj, line 727: xasl was executed because there were no cached values.
But should we check once more whether the same key exists by calling `mht_put_if_not_exists` on `sq_put`?

src/query/subquery_cache.c
```diff
@@ -0,0 +1,749 @@
...
+	      if (sq_walk_xasl_check_not_caching (regu_var->xasl))
+		{
+		  regu_var->xasl->sq_cache_flag |= SQ_CACHE_NOT_CACHING_FLAG;
+		}
+	      regu_var->xasl->sq_cache_flag |= SQ_CACHE_NOT_CACHING_CHECKED_FLAG;
+	    }
+	}
+    }
+
+  EXECUTE_REGU_VARIABLE_XASL (thread_p, regu_var, (val_descr *) vd);
```
youngjinj: I think it is better for readability to separate execute_regu_variable_xasl_with_sq_cache and EXECUTE_REGU_VARIABLE_XASL.
How about keeping EXECUTE_REGU_VARIABLE_XASL in its original location?

src/query/subquery_cache.c
xmilex-git: I revised it to return immediately without performing mht_get.

src/query/subquery_cache.c
xmilex-git: I used the execute_regu_variable_xasl_with_sq_cache() instead of the EXECUTE_REGU_VARIABLE_XASL() for two reasons. 

1. In the part of the fetch_peek_dbval()'s content that consists of a switch case statement, when it's a constant regu var, this macro is executed. However, placing the implementation of execute_regu_variable_xasl_with_sq_cache() there would make the implementation too bulky. 

2. In the future, this approach is not only intended for correlated scalar subqueries in the select clause but will also be applied to cases following an EXISTS. This is to facilitate easier implementation for those scenarios. 

Could I ask for your opinion on how to improve this for readability?

youngjinj: I feel that among the functions in `src/query/subquery_cache.c`, the `execute_regu_variable_xasl_with_sq_cache` function is something different.
I think only functions for managing the result cache of subqueries should be in `src/query/subquery_cache.c`.
It is strange that the function in `src/query/subquery_cache.c` directly calls the `qexec_execute_mainblock` function.

src/query/xasl.h
```diff
@@ -1034,6 +1034,14 @@ struct xasl_node
   int cte_host_var_count;
   int *cte_host_var_index;
 
+  MHT_TABLE *sq_cache_ht;
+  int sq_cache_flag;
+  size_t sq_cache_hit;
+  size_t sq_cache_miss;
+  size_t sq_cache_size;
+  size_t sq_cache_size_max;
+  size_t sq_cache_miss_max;
```
youngjinj: It is not good for XASL_NODE to have statistical information directly.
Please see the following:
```
struct xasl_node
{
  ...
#if defined (SERVER_MODE) || defined (SA_MODE)
  ORDERBY_STATS orderby_stats;
  GROUPBY_STATS groupby_stats;
  XASL_STATS xasl_stats;
  ...
#endif
  ...
};
```

src/query/subquery_cache.c
```diff
@@ -0,0 +1,754 @@
...
+ *
+ * This function clears all cache entries associated with a given XASL node. It does so by iterating over the hash table and
+ * freeing all keys and values. This function is typically called when resetting or destroying the cache for a XASL node.
+ */
+void
+sq_cache_drop_all (xasl_node * xasl)
+{
+  if (xasl->sq_cache_flag & SQ_CACHE_INITIALIZED_FLAG)
+    {
+      if (xasl->sq_cache_ht != NULL)
```
youngjinj: Is it necessary to check `xasl->sq_cache_flag & SQ_CACHE_INITIALIZED_FLAG`?
Is it not enough to check `xasl->sq_cache_ht != NULL`?

src/query/subquery_cache.c
xmilex-git: I defined this function because initializing the **sq_cache**, determining if the query is suitable for caching, managing the flags, and directly executing the `qexec_execute_mainblock()` function if it's not suitable, are considered the most crucial roles in managing the **subquery_cache**. 

Wouldn't it be odd if the mentioned **sq_cache-related** operations were defined in the `fetch_peek_dbval()` and other subquery handling sections?

src/query/subquery_cache.c
xmilex-git: I will proceed to eliminate one of the two preds.

src/query/subquery_cache.c
youngjinj: I don't think it's strange to call the sq_cache function before calling the qexec_execute_mainblock function in fetch_peek_dbval.
If qexec_execute_mainblock was executed, there was no result cache in sq_cache, and we call the sq_cache function again to cache the result.

src/query/subquery_cache.c
```diff
@@ -0,0 +1,754 @@
...
+
+/*
+ * sq_cache_drop_all () - Clears all cache entries for a given XASL node.
+ *   xasl(in): The XASL node for which all cache entries will be cleared.
+ *
+ * This function clears all cache entries associated with a given XASL node. It does so by iterating over the hash table and
+ * freeing all keys and values. This function is typically called when resetting or destroying the cache for a XASL node.
+ */
+void
+sq_cache_drop_all (xasl_node * xasl)
```
youngjinj: How about merging the code of the `sq_cache_drop_all` function with the `sq_cache_destroy` function?

src/query/xasl.h
xmilex-git: Okay, I fixed it.

src/query/subquery_cache.c
xmilex-git: Okay, I merged it.

src/query/subquery_cache.c
xmilex-git: Is it okay to modify _fetch_peek_dbval()_?

```C
    case TYPE_CONSTANT:	/* fetch constant-column value */
      /* is not constant */
      REGU_VARIABLE_SET_FLAG (regu_var, REGU_VARIABLE_FETCH_NOT_CONST);
      assert (!REGU_VARIABLE_IS_FLAGED (regu_var, REGU_VARIABLE_FETCH_ALL_CONST));

      /* execute linked query */
      if (regu_var->xasl && !(regu_var->xasl->sq_cache_flag & SQ_CACHE_NOT_CACHING_FLAG)
	  && (regu_var->xasl->status == XASL_CLEARED || regu_var->xasl->status == XASL_INITIALIZED))
	{

	  if (regu_var->xasl->sq_cache_flag & SQ_CACHE_ENABLED_FLAG)
	    {
	      if (sq_get (regu_var->xasl, regu_var))
		{
		  regu_var->xasl->status = XASL_SUCCESS;
		}
	      else
		{
		  EXECUTE_REGU_VARIABLE_XASL (thread_p, regu_var, (val_descr *) vd);

		  sq_put (regu_var->xasl, regu_var);
		}
	    }
	  else if (regu_var->xasl->sq_cache_flag == 0)
	    {
	      regu_var->xasl->sq_cache_flag |= SQ_CACHE_ENABLED_FLAG;
	      if (!(regu_var->xasl->sq_cache_flag & SQ_CACHE_NOT_CACHING_CHECKED_FLAG))
		{
		  if (sq_walk_xasl_check_not_caching (regu_var->xasl))
		    {
		      regu_var->xasl->sq_cache_flag |= SQ_CACHE_NOT_CACHING_FLAG;
		    }
		  regu_var->xasl->sq_cache_flag |= SQ_CACHE_NOT_CACHING_CHECKED_FLAG;
		}
	    }
	}

      EXECUTE_REGU_VARIABLE_XASL (thread_p, regu_var, (val_descr *) vd);
      if (CHECK_REGU_VARIABLE_XASL_STATUS (regu_var) != XASL_SUCCESS)
	{
	  goto exit_on_error;
	}
      *peek_dbval = regu_var->value.dbvalptr;
      break;
```

youngjinj: I wish the if statement was simple.
```c++
case TYPE_CONSTANT: /* fetch constant-column value */
  /* is not constant */
  REGU_VARIABLE_SET_FLAG (regu_var, REGU_VARIABLE_FETCH_NOT_CONST);
  assert (!REGU_VARIABLE_IS_FLAGED (regu_var, REGU_VARIABLE_FETCH_ALL_CONST));
  
  if (sq_check_enable (...) == ENABLE) /* pseudocode */
    {
      if (sq_find_cache (...) == NOT_FOUND) /* pseudocode */
        {
          /* execute linked query */
          EXECUTE_REGU_VARIABLE_XASL (thread_p, regu_var, vd);
          if (CHECK_REGU_VARIABLE_XASL_STATUS (regu_var) != XASL_SUCCESS)
            {
              goto exit_on_error;
            }
        
           /* pseudocode */
          if (sq_cache (...) != SUCCESS)
            {
              sq_disable_cache (...);
            }
        }
      else
        {
          /* FOUND */
          regu_var->xasl->status = XASL_SUCCESS;
        }
    }
  else
    {
      /* execute linked query */
      EXECUTE_REGU_VARIABLE_XASL (thread_p, regu_var, vd);
      if (CHECK_REGU_VARIABLE_XASL_STATUS (regu_var) != XASL_SUCCESS)
        {
          goto exit_on_error;
        }
    }
        
  *peek_dbval = regu_var->value.dbvalptr;
  break;
```

xmilex-git: Understood. I will apply it after implementation.

youngjinj: I think it would be good to create the `EXECUTE_REGU_VARIABLE_XASL_WITH_SQ_CACHE` macro function for this code.
However, the location must be `xasl.h`, not `sq_cache`.

src/query/query_executor.c
```diff
@@ -2248,6 +2249,10 @@ qexec_clear_xasl (THREAD_ENTRY * thread_p, xasl_node * xasl, bool is_final)
       pg_cnt += qexec_clear_agg_orderby_const_list (thread_p, xasl, is_final);
     }
 
+  if (xasl->sq_cache_flag != 0)
```
youngjinj: Why should we check if `sq_cache_flag` is 0 instead of checking if `sq_cache_ht` is NULL?
```c++
if (xasl->sq_cache_flag != 0)
  {
    assert (xasl->sq_cache_ht != NULL);
  }
```
I think it would be safer to check if sq_cache_ht is NULL.

For example:
```c++
/* destroy the classname cache */
if (Classname_cache != NULL)
  {
    mht_destroy (Classname_cache);
    Classname_cache = NULL;
  }
```

xmilex-git: It doesn't matter which of the two check. Since checking if sq_cache_ht is NULL reduces the likelihood of a memory leak, I will make that adjustment.

src/query/subquery_cache.h
```diff
@@ -0,0 +1,46 @@
...
+
+#define SQ_TYPE_XASL 0
+#define SQ_TYPE_PRED 1
+#define SQ_TYPE_REGU_VAR 2
+#define SQ_TYPE_DBVAL 3
+
+#define SQ_CACHE_ENABLED_FLAG 0x1
+#define SQ_CACHE_INITIALIZED_FLAG 0x2
+#define SQ_CACHE_NOT_CACHING_CHECKED_FLAG 0x4
+#define SQ_CACHE_NOT_CACHING_FLAG 0x8
```
youngjinj: How about using `(XASL_NODE *)->flag`?
```c++
/* src/query/xasl.h */

#define XASL_LINK_TO_REGU_VARIABLE      0x01    /* is linked to regu variable ? */
...
#define XASL_INCLUDES_TDE_CLASS       0x10000   /* is any tde class related */

#define XASL_IS_FLAGED(x, f)        (((x)->flag & (int) (f)) != 0)
#define XASL_SET_FLAG(x, f)         (x)->flag |= (int) (f)
#define XASL_CLEAR_FLAG(x, f)       (x)->flag &= (int) ~(f)
```

src/query/subquery_cache.c
xmilex-git: If implemented according to the above pseudocode, I think the readability of the code will be high.

src/query/subquery_cache.h
xmilex-git: I removed sq_cache_flag and add XASL->flag.

src/query/xasl.h
```diff
@@ -923,6 +924,15 @@ struct xasl_stat
   UINT64 fetch_time;
 };
 
+struct sq_stat
+{
+  UINT64 sq_cache_hit;
+  UINT64 sq_cache_miss;
+  UINT64 sq_cache_size;
+  UINT64 sq_cache_size_max;
+  UINT64 sq_cache_miss_max;
```
youngjinj: These three are not related to statistical information.
How about this?
```c++
struct xasl_node
{
  ...
#if defined (SERVER_MODE) || defined (SA_MODE)
  struct sq_cache
  {
    MHT_TABLE *ht;
    UINT64 max_size; /* UINT64 ? */
    UINT64 size
  
    int flag; /* I wonder if it is necessary. */
  
    struct
      {
      UINT64 hit;
      UINT64 miss;
    } stats;
  
    UINT64 max_miss; /* I wonder if it is necessary. */
  };
#endif
  ...
};
```

xmilex-git: Okay, I'll modify it.

src/query/subquery_cache.c
youngjinj: I think the `XASL_SQ_CACHE_INITIALIZED` flag is unnecessary.
In other functions, `XASL_SQ_CACHE_INITIALIZED` is checked instead of null checking for `sq_cache.ht`.

src/query/xasl.h
youngjinj: This is an example.
`sq_cache` must be created as a separate struct and typedefed.

In xasl_node, consider how `sq_cache` exists as a pointer variable and is dynamically allocated when needed.

src/query/subquery_cache.c
```diff
@@ -0,0 +1,724 @@
...
+
+  if (xasl->sq_cache.size_max < xasl->sq_cache.size + new_entry_size)
+    {
+      XASL_SET_FLAG (xasl, XASL_SQ_CACHE_NOT_CACHING);
+      sq_free_key (key);
+      sq_free_val (val);
+      return ER_FAILED;
+    }
+
+  ret = mht_put_if_not_exists (xasl->sq_cache.ht, key, val);
```
youngjinj: I think it will be a problem if collision occurs.
It seems that in cases where only the keys are the same, different cached results may be returned.

I think we need to create another hash function that connects the entry to next, or check cmp_func before returning.

It is also related to the `sq_get` function.

xmilex-git: When executing a correlated scalar subquery, we should assume that caching is not possible if the variables used as keys are the same but the results differ. 

Moreover, within this _mht_put_if_not_exists()_ function, if the key is the same, it does not execute the input, and in this case, if a different val is returned for ret (ret!=val), it will return ER_FAILED and will not perform caching. 

Is there still a risk of collision in this case?

src/query/subquery_cache.c
```diff
@@ -0,0 +1,724 @@
...
+ */
+int
+sq_check_enable (xasl_node * xasl)
+{
+  if (!(xasl) || XASL_IS_FLAGED (xasl, XASL_SQ_CACHE_NOT_CACHING)
+      || !(xasl->status == XASL_CLEARED || xasl->status == XASL_INITIALIZED))
+    {
+      return FALSE;
+    }
+  if (XASL_IS_FLAGED (xasl, XASL_SQ_CACHE_ENABLED))
```
shparkcubrid: Is XASL_SQ_CACHE_ENABLED necessary?
I think it can be checked with XASL_SQ_CACHE_NOT_CACHING_CHECKED and XASL_SQ_CACHE_NOT_CACHING.

src/query/subquery_cache.c
```diff
@@ -0,0 +1,724 @@
...
+#include "db_set_function.h"
+
+#include "subquery_cache.h"
+
+typedef union _sq_regu_value
+{
+  /* fields used by both XASL interpreter and regulator */
+  DB_VALUE *dbvalptr;		/* for constant values */
+  QFILE_SORTED_LIST_ID *srlist_id;	/* sorted list identifier for subquery results */
+} sq_regu_value;
```
shparkcubrid: Please write in capital letters as shown below.
```
typedef struct analytic_state ANALYTIC_STATE;
struct analytic_state
{
```

src/query/subquery_cache.c
```diff
@@ -0,0 +1,724 @@
...
+
+  p = xasl->spec_list;
+
+  if (p && !p->where_key && !p->where_pred && !p->where_range)
+    {
+      /* if there's no predicate in xasl, not caching */
+      return NULL;
+    }
+
+  keyp = (sq_key *) malloc (sizeof (sq_key));
```
shparkcubrid: Please use db_private_alloc ().

src/query/subquery_cache.c
```diff
@@ -0,0 +1,724 @@
...
+sq_key *
+sq_make_key (xasl_node * xasl)
+{
+  sq_key *keyp;
+  int cnt = 0;
+  ACCESS_SPEC_TYPE *p;
+
+  p = xasl->spec_list;
+
+  if (p && !p->where_key && !p->where_pred && !p->where_range)
```
shparkcubrid: move it to sq_walk_xasl_check_not_caching()

src/query/subquery_cache.c
```diff
@@ -0,0 +1,724 @@
...
+
+  if (!p)
+    {
+      return 0;
+    }
+
+  switch (type)
+    {
+    case SQ_TYPE_XASL:
+      if (1)
```
shparkcubrid: Why did you write 'if (1)'?

src/query/subquery_cache.c
```diff
@@ -0,0 +1,724 @@
...
+	    {
+	      cnt += sq_walk_xasl_and_add_val_to_set (src->pe.m_pred.lhs, SQ_TYPE_PRED, key);
+	      cnt += sq_walk_xasl_and_add_val_to_set (src->pe.m_pred.rhs, SQ_TYPE_PRED, key);
+	    }
+	  else if (src->type == T_EVAL_TERM)
+	    {
+	      COMP_EVAL_TERM t = src->pe.m_eval_term.et.et_comp;
+	      cnt += sq_walk_xasl_and_add_val_to_set (t.lhs, SQ_TYPE_REGU_VAR, key);
+	      cnt += sq_walk_xasl_and_add_val_to_set (t.rhs, SQ_TYPE_REGU_VAR, key);
+	    }
```
shparkcubrid: Please handle 'else'.

src/query/subquery_cache.c
```diff
@@ -0,0 +1,724 @@
...
+
+    case SQ_TYPE_REGU_VAR:
+      if (1)
+	{
+	  REGU_VARIABLE *src = (REGU_VARIABLE *) p;
+	  if (src->type == TYPE_CONSTANT)
+	    {
+	      cnt += sq_walk_xasl_and_add_val_to_set (src->value.dbvalptr, SQ_TYPE_DBVAL, key);
+	    }
+	  if (src->type == TYPE_DBVAL)
```
shparkcubrid: if ==> else if

src/query/subquery_cache.c
```diff
@@ -0,0 +1,724 @@
...
+	    {
+	      cnt += sq_walk_xasl_and_add_val_to_set (&src->value.dbval, SQ_TYPE_DBVAL, key);
+	    }
+	  else if (src->type == TYPE_INARITH)
+	    {
+	      cnt += sq_walk_xasl_and_add_val_to_set (src->value.arithptr->leftptr, SQ_TYPE_REGU_VAR, key);
+	      cnt += sq_walk_xasl_and_add_val_to_set (src->value.arithptr->rightptr, SQ_TYPE_REGU_VAR, key);
+	    }
```
shparkcubrid: Please handle 'else'.

src/query/subquery_cache.c
shparkcubrid: In this case, is it a real constant? ( col1 = 3)
Since it has the same value, please review whether it is appropriate to exclude them from the key.

src/query/xasl.h
```diff
@@ -1073,6 +1076,18 @@ struct xasl_node
...
   XASL_STATS xasl_stats;
+  struct sq_cache
+  {
+    MHT_TABLE *ht;
+    UINT64 size_max;
+    UINT64 size;
+    UINT64 miss_max;
+    struct
+    {
+      UINT64 hit;
```
shparkcubrid: I think INT is enough.

src/query/xasl.h
```diff
@@ -1073,6 +1076,18 @@ struct xasl_node
   ORDERBY_STATS orderby_stats;
   GROUPBY_STATS groupby_stats;
   XASL_STATS xasl_stats;
+  struct sq_cache
+  {
+    MHT_TABLE *ht;
+    UINT64 size_max;
+    UINT64 size;
+    UINT64 miss_max;
```
shparkcubrid: I think there may be some variables that aren't needed. Because XASL is used a lot, we need to make it as lightweight as possible.

src/query/subquery_cache.c
```diff
@@ -0,0 +1,724 @@
...
+  if (key == NULL)
+    {
+      XASL_SET_FLAG (xasl, XASL_SQ_CACHE_NOT_CACHING);
+      return false;
+    }
+  if (!XASL_IS_FLAGED (xasl, XASL_SQ_CACHE_INITIALIZED))
+    {
+      sq_cache_initialize (xasl);
+      xasl->sq_cache.stats.miss++;
+      sq_free_key (key);
```
shparkcubrid: This can be removed if checked before sq_make_key().
```
if (XASL_IS_FLAGED (xasl, XASL_SQ_CACHE_INITIALIZED))
    {  .... }
else
    { ....

File: src/query/subquery_cache.c [hunk 24]
```diff
@@ -0,0 +1,724 @@
...
+static void sq_free_key (sq_key * key);
+static void sq_free_val (sq_val * val);
+static void sq_unpack_val (sq_val * val, REGU_VARIABLE * retp);
+
+static unsigned int sq_hash_func (const void *key, unsigned int ht_size);
+static int sq_cmp_func (const void *key1, const void *key2);
+static int sq_rem_func (const void *key, void *data, void *args);
+
+static int sq_walk_xasl_check_not_caching (xasl_node * xasl);
+static int sq_walk_xasl_and_add_val_to_set (void *p, int type, sq_key * pred_set);
```
shparkcubrid: sq_walk_xasl_check_not_caching ()
sq_walk_xasl_and_add_val_to_set ()
You need to move it to XASL generation in next step.

src/query/subquery_cache.c
shparkcubrid: The key can be used as a variable without dynamic allocation.

src/query/subquery_cache.c
```diff
@@ -0,0 +1,724 @@
...
+ */
+int
+sq_put (xasl_node * xasl, REGU_VARIABLE * regu_var)
+{
+  sq_key *key;
+  sq_val *val;
+  const void *ret;
+  UINT64 new_entry_size = 0;
+
+  key = sq_make_key (xasl);
```
shparkcubrid: The key generated by sq_get() is reusable.

src/query/subquery_cache.c
```diff
@@ -0,0 +1,724 @@
...
+
+  if (key == NULL)
+    {
+      XASL_SET_FLAG (xasl, XASL_SQ_CACHE_NOT_CACHING);
+      return ER_FAILED;
+    }
+
+  val = sq_make_val (regu_var);
+
+  if (!XASL_IS_FLAGED (xasl, XASL_SQ_CACHE_INITIALIZED))
```
shparkcubrid: Is this check necessary?

src/query/subquery_cache.c
```diff
@@ -0,0 +1,724 @@
...
+
+  ret->t = val->type;
+
+  switch (ret->t)
+    {
+    case TYPE_CONSTANT:
+      ret->val.dbvalptr = db_value_copy (val->value.dbvalptr);
+      break;
+
+    case TYPE_LIST_ID:
```
shparkcubrid, line 113: What is this case?

src/query/subquery_cache.c
```diff
@@ -0,0 +1,724 @@
...
+      ret->val.srlist_id = (QFILE_SORTED_LIST_ID *) malloc (sizeof (QFILE_SORTED_LIST_ID));
+      ret->val.srlist_id->list_id = qfile_clone_list_id (val->value.srlist_id->list_id, true);
+      ret->val.srlist_id->sorted = val->value.srlist_id->sorted;
+      break;
+
+    default:
+      assert (0);
```
shparkcubrid: Error handling is required.

src/query/subquery_cache.c
```diff
@@ -0,0 +1,724 @@
...
+    }
+
+  val = sq_make_val (regu_var);
+
+  if (!XASL_IS_FLAGED (xasl, XASL_SQ_CACHE_INITIALIZED))
+    {
+      sq_cache_initialize (xasl);
+    }
+
+  new_entry_size += (UINT64) or_db_value_size (key->pred_set) + sizeof (sq_key);
```
shparkcubrid:
```suggestion
  xasl->sq_cache.size += (UINT64) or_db_value_size (key->pred_set) + sizeof (sq_key);
```

src/query/stream_to_xasl.c
```diff
@@ -1700,6 +1700,9 @@ stx_build_xasl_node (THREAD_ENTRY * thread_p, char *ptr, XASL_NODE * xasl)
 
   /* initialize xasl status */
   xasl->status = XASL_INITIALIZED;
+  XASL_CLEAR_FLAG (xasl,
```
shparkcubrid: Since it is already initialized, this code seems unnecessary.

src/query/subquery_cache.c
```diff
@@ -0,0 +1,724 @@
...
+	  retp->value.srlist_id->sorted = v->val.srlist_id->sorted;
+	}
+      else
+	{
+	  assert (0);
+	  retp->value.srlist_id->list_id = qfile_clone_list_id (v->val.srlist_id->list_id, true);
+	}
+      break;
+
+    default:
```
shparkcubrid, line 210: Error handling required

src/query/subquery_cache.c
```diff
@@ -0,0 +1,724 @@
...
+  sq_key *k = (sq_key *) key;
+  int i;
+  unsigned int h = 0;
+  DB_SET *set = db_get_set (k->pred_set);
+  DB_VALUE v;
+
+  for (i = 0; i < db_set_size (set); i++)
+    {
+      db_set_get (set, i, &v);
+      h += mht_valhash (&v, ht_size);
```
shparkcubrid: How about XOR '^'

src/query/subquery_cache.c
```diff
@@ -0,0 +1,724 @@
...
+      return 0;
+    }
+
+  for (i = 0; i < sz1; i++)
+    {
+      db_set_get (set1, i, &v1);
+      db_set_get (set2, i, &v2);
+      if (!mht_compare_dbvalues_are_equal (&v1, &v2))
+	{
+	  pr_clear_value (&v1);
```
shparkcubrid: Is there any way to not call pr_clear_value()?

src/query/subquery_cache.c
xmilex-git: I have added uppercase typedefs for each struct.

src/query/subquery_cache.c
xmilex-git: I replaced malloc() with db_private_alloc(), and the content concerning key allocation only once will be revised after changes related to xasl_generation.

src/query/subquery_cache.c
xmilex-git: I have applied it.

src/query/subquery_cache.c
xmilex-git: To improve the readability of the implementation of local variables, I have added code blocks.

src/query/subquery_cache.c
xmilex-git: Upon review, since DBVAL is not a constant, it is correct to exclude it, so I have omitted that.

src/query/xasl.h
xmilex-git: I have changed it to int.

src/query/xasl.h
xmilex-git: I will delete miss_max and use it as a value proportional to size_max.

src/query/subquery_cache.c
xmilex-git: This is code that will be used in the future. I added a comment stating that it will be implemented later and deleted the code (related to EXISTS implementation).

src/query/stream_to_xasl.c
xmilex-git: I have deleted the code.

src/query/subquery_cache.c
xmilex-git: I also think that XOR would be faster. I will incorporate the changes.

src/query/subquery_cache.c
xmilex-git: The purpose of pre-calculating new_entry_size was to prevent sq_cache->size from exceeding size_max. If we allow for a slight exceedance in notation, such a change would be feasible. Is this approach more appropriate?

src/parser/xasl_generation.c
```diff
@@ -13960,6 +13966,27 @@ pt_to_outlist (PARSER_CONTEXT * parser, PT_NODE * node_list, SELUPD_LIST ** selu
 		      regu->value.dbvalptr = value_list->val;
 		      /* move to next db_value holder */
 		      value_list = value_list->next;
+		      if (node->info.query.q.select.hint & PT_HINT_NO_SUBQUERY_CACHE)
```
shparkcubrid: How about checking it in pt_check_corr_subquery_hash_result_cache()

src/parser/xasl_generation.c
```diff
@@ -13960,6 +13966,27 @@ pt_to_outlist (PARSER_CONTEXT * parser, PT_NODE * node_list, SELUPD_LIST ** selu
...
+			{
+			  /* it means SUBQUERY RESULT won't be cached. */
+			}
+		      else
+			{
+			  if (pt_check_corr_subquery_hash_result_cache (xasl))
+			    {
+			      bool cachable = true;
+			      parser_walk_tree (parser, node, NULL, NULL, pt_check_corr_subquery_not_cachable_expr,
```
shparkcubrid: How about checking it in pt_check_corr_subquery_hash_result_cache()?

shparkcubrid: Since pt_check_corr_subquery_hash_result_cache() is called recursively, it would be a good idea to create a new function.

src/parser/xasl_generation.c
xmilex-git: If we apply it as you mentioned, since the function is called recursively, if there is a hint in any query lower than itself (whether it is aptr or dptr), it will prevent itself from using the hash result cache. Is it okay to apply it this way?

src/parser/xasl_generation.c
xmilex-git: The `pt_check_corr_subquery_hash_result_cache` function recursively searches the created XASL structure, while `pt_check_corr_subquery_not_cachable_expr` recursively searches the created PT_NODE structure. Combining these two would likely complicate the implementation.

src/parser/xasl_generation.c
```diff
@@ -27187,3 +27214,238 @@ pt_to_instnum_pred (PARSER_CONTEXT * parser, XASL_NODE * xasl, PT_NODE * pred)
...
+{
+  int n_elements;
+  DB_VALUE **sq_key_struct;
+
+  n_elements = pt_make_sq_cache_key_struct (NULL, (void *) xasl, SQ_TYPE_XASL);
+  sq_key_struct = (DB_VALUE **) pt_alloc_packing_buf (n_elements * sizeof (DB_VALUE *));
+  memset ((void *) sq_key_struct, 0, n_elements * sizeof (DB_VALUE *));
+  pt_make_sq_cache_key_struct (sq_key_struct, (void *) xasl, SQ_TYPE_XASL);
+
+  if (n_elements <= 0)
```
shparkcubrid: It would be better if it was under the sentence below
```
27287 n_elements = pt_make_sq_cache_key_struct (NULL, (void *) xasl, SQ_TYPE_XASL);
```

src/parser/xasl_generation.c
shparkcubrid: Since pt_check_corr_subquery_hash_result_cache() is called recursively, it would be a good idea to create a new function.

src/parser/xasl_generation.c
```diff
@@ -27187,3 +27214,238 @@ pt_to_instnum_pred (PARSER_CONTEXT * parser, XASL_NODE * xasl, PT_NODE * pred)
...
+	  cnt += pt_make_sq_cache_key_struct (key_struct, (void *) pred_src->pe.m_pred.lhs, SQ_TYPE_PRED);
+	  cnt += pt_make_sq_cache_key_struct (key_struct, (void *) pred_src->pe.m_pred.rhs, SQ_TYPE_PRED);
+	}
+      else if (pred_src->type == T_EVAL_TERM)
+	{
+	  COMP_EVAL_TERM t = pred_src->pe.m_eval_term.et.et_comp;
+	  cnt += pt_make_sq_cache_key_struct (key_struct, (void *) t.lhs, SQ_TYPE_REGU_VAR);
+	  cnt += pt_make_sq_cache_key_struct (key_struct, (void *) t.rhs, SQ_TYPE_REGU_VAR);
+	}
+      else
```
shparkcubrid, line 27433: Is it okay not to consider the 'T_NOT_TERM' type? 
What is T_NOT_TERM for?

src/parser/xasl_generation.c
xmilex-git: I have understood the intent behind your review and applied it to the code. Please check and confirm.

src/parser/xasl_generation.c
xmilex-git: I have understood the intent behind your review and applied it to the code. Please check and confirm.

src/parser/xasl_generation.c
xmilex-git: I have applied it to the code. Please check and confirm.

src/parser/xasl_generation.c
```diff
@@ -27187,3 +27214,238 @@ pt_to_instnum_pred (PARSER_CONTEXT * parser, XASL_NODE * xasl, PT_NODE * pred)
...
+	      cnt += pt_make_sq_cache_key_struct (key_struct, (void *) scan_ptr, SQ_TYPE_XASL);
+	    }
+	}
+      if (xasl_src->aptr_list)
+	{
+	  for (aptr = xasl_src->aptr_list; aptr != NULL; aptr = aptr->next)
+	    {
+	      cnt += pt_make_sq_cache_key_struct (key_struct, (void *) aptr, SQ_TYPE_XASL);
+	    }
+	}
```
shparkcubrid, line 27374: how about for 'xasl->after_join_pred', 'xasl->if_pred'.
```
cnt += pt_make_sq_cache_key_struct (key_struct, (void *) xasl->after_join_pred, SQ_TYPE_PRED);
cnt += pt_make_sq_cache_key_struct (key_struct, (void *) xasl->if_pred, SQ_TYPE_PRED);
```

src/parser/xasl_generation.c
```diff
@@ -27187,3 +27214,238 @@ pt_to_instnum_pred (PARSER_CONTEXT * parser, XASL_NODE * xasl, PT_NODE * pred)
...
+	{
+	  cnt += pt_make_sq_cache_key_struct (key_struct, (void *) regu_src->value.dbvalptr, SQ_TYPE_DBVAL);
+	}
+      else if (regu_src->type == TYPE_INARITH)
+	{
+	  cnt += pt_make_sq_cache_key_struct (key_struct, (void *) regu_src->value.arithptr->leftptr, SQ_TYPE_REGU_VAR);
+	  cnt +=
+	    pt_make_sq_cache_key_struct (key_struct, (void *) regu_src->value.arithptr->rightptr, SQ_TYPE_REGU_VAR);
+	}
+      else
```
shparkcubrid: Can types such as TYPE_OUTARITH and TYPE_FUNC not be considered when generating keys?
Missing keys may produce incorrect results. If there are cases that cannot be considered, the subquery cache should be disabled.

src/parser/xasl_generation.c
xmilex-git: T_NOT_TERM is used in the following cases. 
`and NOT a.c3 + b.c1 = 2`
In most cases, when there is an '=' after a NOT, it is rewritten as NEQ, so T_NOT_TERM is not generated, which is why it seems there are no errors in testcases. However, it is also correct to include it to prevent any potential errors.

src/query/xasl.h
```diff
@@ -949,6 +952,22 @@ struct partition_spec_node
...
 #endif /* defined (SERVER_MODE) || defined (SA_MODE) */
 
+struct sq_cache
+{
+  DB_VALUE **sq_key_struct;
+  int n_elements;
+#if defined (SERVER_MODE) || defined (SA_MODE)
+  MHT_TABLE *ht;
+  UINT64 size_max;
+  UINT64 size;
```
shparkcubrid: I think unsigned int will be enough.

src/parser/xasl_generation.c
```diff
@@ -27187,3 +27198,287 @@ pt_to_instnum_pred (PARSER_CONTEXT * parser, XASL_NODE * xasl, PT_NODE * pred)
...
+	{
+	  i = 0;
+	  if (key_struct[i])
+	    {
+	      while (key_struct[i] != NULL)
+		{
+		  i++;
+		}
+	    }
+	  key_struct[i] = (DB_VALUE *) p;
```
shparkcubrid: How about allocating dynamic memory here without calling the function twice? 
For example, you can use a linked list like db_value->next and allocate only one db_value size.

xmilex-git: In the execution phase, fixed length is required, and having a fixed length in the implementation related to stx and xts enhances both complexity and readability of the implementation. Therefore, I will implement it as a fixed-size array.

src/base/system_parameter.c
```diff
@@ -737,6 +737,8 @@ static const char sysprm_ha_conf_file_name[] = "cubrid_ha.conf";
 #define PRM_NAME_DEDUPLICATE_KEY_LEVEL     "deduplicate_key_level"
 #define PRM_NAME_PRINT_INDEX_DETAIL        "print_index_detail"
 
+#define PRM_NAME_MAX_SUBQUERY_CACHE_SIZE    "subquery_cache_size"
```
youngjinj: Please use the same name.

How about this?
```suggestion
#define PRM_NAME_SUBQUERY_CACHE_MAX_SIZE    "subquery_cache_max_size"
```

xmilex-git: I'll name it "max_subquery_cache_size"

src/parser/xasl_generation.c
xmilex-git: I added code that handles regu_type. please check and confirm.

src/query/stream_to_xasl.c
```diff
@@ -1685,6 +1686,7 @@ stx_build_xasl_node (THREAD_ENTRY * thread_p, char *ptr, XASL_NODE * xasl)
 {
   int offset;
   int tmp, i;
+  UINTPTR tmp_ptr;
```
youngjinj: Not used.

src/query/xasl_to_stream.c
```diff
@@ -3234,6 +3246,23 @@ xts_process_xasl_node (char *ptr, const XASL_NODE * xasl)
...
+      ptr = or_pack_int (ptr, xasl->sq_cache->n_elements);
+    }
+  else
+    {
+      ptr = or_pack_int (ptr, 0);
+    }
+
+  offset = xts_save_sq_cache (xasl->sq_cache);
+
+  if (offset == ER_FAILED)
```
youngjinj:
```suggestion
  offset = xts_save_sq_cache (xasl->sq_cache);
  if (offset == ER_FAILED)
```

src/query/xasl_to_stream.c
```diff
@@ -6085,7 +6114,9 @@ xts_sizeof_xasl_node (const XASL_NODE * xasl)
 
   size += (OR_INT_SIZE		/* iscan_oid_order */
 	   + PTR_SIZE		/* query_alias */
-	   + PTR_SIZE);		/* next */
+	   + PTR_SIZE		/* next */
+	   + PTR_SIZE		/* sq_cache_n_elements */
```
youngjinj: OR_INT_SIZE ?

src/query/stream_to_xasl.c
xmilex-git: Okay. I modified it.

src/query/xasl_to_stream.c
xmilex-git: Okay. I modified it.

src/parser/xasl_generation.c
xmilex-git: I have applied it to the code. Please check and confirm.

src/parser/xasl_generation.c
xmilex-git: I have applied it to the code. Please check and confirm.

src/query/stream_to_xasl.c
```diff
@@ -2413,6 +2418,26 @@ stx_build_cte_xasl_id (THREAD_ENTRY * thread_p, char *ptr, XASL_ID * cte_xasl_id
   return ptr;
 }
 
+static SQ_CACHE *
+stx_restore_sq_cache (THREAD_ENTRY * thread_p, char *ptr, int n_elements)
```
youngjinj: I think the `stx_build_` prefix should be used.
Functions using the 'stx_restore_' prefix call the stx_get_struct_visited_ptr and stx_mark_struct_visited functions.

And I think we should get sq_cache->n_elements here.
For this, ptr should be added to the parameters.

src/query/stream_to_xasl.c
```diff
@@ -2413,6 +2418,26 @@ stx_build_cte_xasl_id (THREAD_ENTRY * thread_p, char *ptr, XASL_ID * cte_xasl_id
...
+{
+  SQ_CACHE *sq_cache = NULL;
+
+  if (n_elements > 0)
+    {
+      sq_cache = (SQ_CACHE *) stx_alloc_struct (thread_p, sizeof (SQ_CACHE));
+      sq_cache->sq_key_struct = stx_restore_db_value_array_extra (thread_p, ptr, n_elements, n_elements);
```
youngjinj: Exception handling is required.

src/query/subquery_cache.c
```diff
@@ -0,0 +1,513 @@
...
+      new_entry_size += (UINT64) or_db_value_size (val->val.dbvalptr) + sizeof (SQ_VAL);
+      break;
+
+    default:
+      break;
+    }
+
+  if (xasl->sq_cache->size_max < xasl->sq_cache->size + new_entry_size)
+    {
+      XASL_CLEAR_FLAG (xasl, XASL_SQ_CACHE);
```
youngjinj: If the information set when generating xasl is changed during execution, it must be restored at the end.

Here, the `XASL_SQ_CACHE` flag must not be cleared.
If the flag is cleared to disable the subquery cache, the flag must be set again at the end.

Please refer to the reproducible scenario.
Do not use recompile hint.
```sql
drop table if exists t1, t2;
create table t1 (c1 int, c2 int);
create table t2 (c1 int, c2 int);

insert into t1 values (1, 1), (2, 2);
insert into t2 values (1, 10), (2, 20);

set trace on;

select t1.c2, (select t2.c2 from t2 where t2.c1 = t1.c1) from t1;
show trace;
/* SUBQUERY_CACHE (hit: 0, miss: 2, size: 96 Bytes, status: enabled) */

select t1.c2, (select t2.c2 from t2 where t2.c1 = t1.c1) from t1;
show trace;
/* nothing */
```

The xasl stored in xcache_entry->cache_clones is reused.

src/query/query_executor.c
```diff
@@ -2248,6 +2249,11 @@ qexec_clear_xasl (THREAD_ENTRY * thread_p, xasl_node * xasl, bool is_final)
       pg_cnt += qexec_clear_agg_orderby_const_list (thread_p, xasl, is_final);
     }
 
+  if (xasl->sq_cache != NULL)
+    {
+      sq_cache_destroy (xasl);
```
youngjinj: The xasl stored in xcache_entry->cache_clones is reused.
When generating xasl, information set in sq_cache must not be cleared.

src/parser/xasl_generation.c
```diff
@@ -27187,3 +27198,457 @@ pt_to_instnum_pred (PARSER_CONTEXT * parser, XASL_NODE * xasl, PT_NODE * pred)
...
+int
+pt_check_corr_subquery_hash_result_cache (PARSER_CONTEXT * parser, PT_NODE * node, XASL_NODE * xasl)
+{
+  bool cachable;
+  if (node->info.query.q.select.hint & PT_HINT_NO_SUBQUERY_CACHE)
+    {
+      /* it means SUBQUERY RESULT won't be cached. */
+      return false;
+    }
+  else
```
shparkcubrid: NIT : I think it's better for readability not to use 'else'.
```
if (node->info.query.q.select.hint & PT_HINT_NO_SUBQUERY_CACHE)
    {
      /* it means SUBQUERY RESULT won't be cached. */
      return false;
    }
if (!pt_recursive_check_corr_subquery_hash_result_cache (xasl))
    {
      return false;
    }
...
return true;
```

src/query/stream_to_xasl.c
xmilex-git: I've applied your review. Please confirm.

src/query/stream_to_xasl.c
xmilex-git: I've applied your review. Please confirm.

src/parser/xasl_generation.c
```diff
@@ -27187,3 +27198,457 @@ pt_to_instnum_pred (PARSER_CONTEXT * parser, XASL_NODE * xasl, PT_NODE * pred)
...
+  if (xasl)
+    {
+      XASL_NODE *scan_ptr, *aptr;
+      bool ret;
+
+      if (!xasl->spec_list)
+	{
+	  return false;
+	}
+      else if (!xasl->spec_list->where_key && !xasl->spec_list->where_pred && !xasl->spec_list->where_range)
```
shparkcubrid: if_pred or after_join_term is missing. and This check can be done with the number of keys in pt_make_sq_cache_key_struct ().

src/parser/xasl_generation.c
```diff
@@ -27187,3 +27198,457 @@ pt_to_instnum_pred (PARSER_CONTEXT * parser, XASL_NODE * xasl, PT_NODE * pred)
...
+	{
+	  return false;
+	}
+      if (xasl->dptr_list)
```
shparkcubrid: check for bptr_list and fptr_list is needed.

src/parser/xasl_generation.c
```diff
@@ -27187,3 +27198,457 @@ pt_to_instnum_pred (PARSER_CONTEXT * parser, XASL_NODE * xasl, PT_NODE * pred)
...
+	{
+	  return false;
+	}
+      if (!XASL_IS_FLAGED (xasl, XASL_LINK_TO_REGU_VARIABLE) && !(xasl->type == SCAN_PROC))
```
shparkcubrid: This case is for what?. Please add comment.

src/parser/xasl_generation.c
shparkcubrid: Please modify limitation in JIRA and add test cases.

src/query/subquery_cache.c
xmilex-git: I've handled that error. Please confirm.

xmilex-git: (I added `bool enabled` in sq_cache struct)

src/query/query_executor.c
xmilex-git: I've handled that error. Please confirm.

src/parser/xasl_generation.c
xmilex-git: I've applied your review. Please confirm.

src/query/stream_to_xasl.c
```diff
@@ -2413,6 +2419,40 @@ stx_build_cte_xasl_id (THREAD_ENTRY * thread_p, char *ptr, XASL_ID * cte_xasl_id
...
+
+  if (n_elements > 0)
+    {
+      new_sq_cache = (SQ_CACHE *) stx_alloc_struct (thread_p, sizeof (SQ_CACHE));
+      if (new_sq_cache == NULL)
+	{
+	  stx_set_xasl_errcode (thread_p, ER_OUT_OF_VIRTUAL_MEMORY);
+	  return NULL;
+	}
+      new_sq_cache->sq_key_struct =
```
youngjinj: Exception handling is required.

src/query/xasl_to_stream.c
```diff
@@ -3234,6 +3246,22 @@ xts_process_xasl_node (char *ptr, const XASL_NODE * xasl)
...
+      ptr = or_pack_int (ptr, 0);
+    }
+
+  offset = xts_save_sq_cache (xasl->sq_cache);
+  if (offset == ER_FAILED)
+    {
+      return NULL;
+    }
+  ptr = or_pack_int (ptr, offset);
+
```
youngjinj: Please collect related codes in `xts_save_sq_cache`.
Refer to `stx_build_sq_cache`.

src/parser/xasl_generation.c
xmilex-git: This case is to check whether the query connected to `regu_var`. However, if we only check for that, when there is a **join query** connected to regu_var, that query will also be recursively checked, resulting in a situation where the entire query cannot be cached. Therefore, `SCAN_PROC` has been added as an exception case.

src/query/stream_to_xasl.c
```diff
@@ -2413,6 +2419,40 @@ stx_build_cte_xasl_id (THREAD_ENTRY * thread_p, char *ptr, XASL_ID * cte_xasl_id
...
+	stx_restore_db_value_array_extra (thread_p, &xasl_unpack_info->packed_xasl[offset], n_elements, n_elements);
+      new_sq_cache->n_elements = n_elements;
+      new_sq_cache->ht = NULL;
+      new_sq_cache->size_max = 0;
+      new_sq_cache->size = 0;
+      new_sq_cache->enabled = false;
+      new_sq_cache->stats.hit = 0;
+      new_sq_cache->stats.miss = 0;
```
youngjinj: How about calling memset after `stx_alloc_struct`?

src/parser/xasl_generation.c
xmilex-git: This case is to check whether the query connected to regu_var. However, if we only check for that, when there is a join query connected to regu_var, that query will also be recursively checked, resulting in a situation where the entire query cannot be cached. Therefore, SCAN_PROC has been added as an exception case.

src/parser/xasl_generation.c
xmilex-git: I've move predicate check to pt_make_sq_cache_key_struct (). Please confirm.

src/query/query_dump.c
```diff
@@ -3274,6 +3275,21 @@ qdump_print_stats_text (FILE * fp, xasl_node * xasl_p, int indent)
   qdump_print_stats_text (fp, xasl_p->scan_ptr, indent);
   qdump_print_stats_text (fp, xasl_p->connect_by_ptr, indent);
 
+  if (xasl_p->sq_cache && XASL_IS_FLAGED (xasl_p, XASL_SQ_CACHE))
+    {
+      fprintf (fp, "%*c", indent, ' ');
+      if (xasl_p->sq_cache->enabled)
+	{
+	  fprintf (fp, "SUBQUERY_CACHE (hit: %d, miss: %d, size: %lu Bytes, status: enabled)\n",
```
youngjinj: The words "hit," "miss," and "status" are printed in lowercase letters. However, the word "Bytes" is not printed in lowercase.
How about using "B" instead of the word "Bytes" or printing it in lowercase?

xmilex-git: Okay.

src/parser/xasl_generation.c
xmilex-git: This case is to check whether the query connected to regu_var. However, if we only check for that, when there is a join query connected to regu_var, that query will also be recursively checked, resulting in a situation where the entire query cannot be cached. Therefore, SCAN_PROC has been added as an exception case.

src/parser/xasl_generation.c
xmilex-git: I added bptr_list and fptr_list to that place. Please confirm.

src/query/stream_to_xasl.c
xmilex-git: I've re-designed that function. Please confirm.

src/query/xasl_to_stream.c
xmilex-git: I've re-designed that function. Please confirm.

src/query/stream_to_xasl.c
xmilex-git: I've applied it. Please confirm.

src/query/xasl.h
```diff
@@ -495,6 +497,7 @@ struct cte_proc_node
 #define XASL_NEED_SINGLE_TUPLE_SCAN   0x8000	/* for exists operation */
 #define XASL_INCLUDES_TDE_CLASS	      0x10000	/* is any tde class related */
 #define XASL_SAMPLING_SCAN	      0x20000	/* is sampling scan */
+#define XASL_SQ_CACHE	      0x40000
```
youngjinj: How about this?
```suggestion
#define XASL_USES_SQ_CACHE	      0x40000   /* subquery uses result cache */
```

src/query/stream_to_xasl.c
```diff
@@ -2413,6 +2419,40 @@ stx_build_cte_xasl_id (THREAD_ENTRY * thread_p, char *ptr, XASL_ID * cte_xasl_id
...
+	  return NULL;
+	}
+      new_sq_cache->sq_key_struct =
+	stx_restore_db_value_array_extra (thread_p, &xasl_unpack_info->packed_xasl[offset], n_elements, n_elements);
+      if (new_sq_cache->sq_key_struct == NULL)
+	{
+	  stx_set_xasl_errcode (thread_p, ER_OUT_OF_VIRTUAL_MEMORY);
+	  return NULL;
+	}
+      memset (new_sq_cache, 0, sizeof (SQ_CACHE));
```
youngjinj:
```suggestion
      memset (new_sq_cache, 0, sizeof (SQ_CACHE));
      new_sq_cache->sq_key_struct =
	stx_restore_db_value_array_extra (thread_p, &xasl_unpack_info->packed_xasl[offset], n_elements, n_elements);
      if (new_sq_cache->sq_key_struct == NULL)
	{
	  stx_set_xasl_errcode (thread_p, ER_OUT_OF_VIRTUAL_MEMORY);
	  return NULL;
	}
```

src/query/query_dump.c
```diff
@@ -3274,6 +3293,21 @@ qdump_print_stats_text (FILE * fp, xasl_node * xasl_p, int indent)
   qdump_print_stats_text (fp, xasl_p->scan_ptr, indent);
   qdump_print_stats_text (fp, xasl_p->connect_by_ptr, indent);
 
+  if (xasl_p->sq_cache && XASL_IS_FLAGED (xasl_p, XASL_SQ_CACHE))
+    {
+      fprintf (fp, "%*c", indent, ' ');
+      if (xasl_p->sq_cache->enabled)
+	{
+	  fprintf (fp, "SUBQUERY_CACHE (hit: %d, miss: %d, size: %lu B, status: enabled)\n",
```
youngjinj: I think it is better to have no space before the letter "B".

src/query/stream_to_xasl.c
xmilex-git: I edited that. My mistakes.

src/query/query_dump.c
xmilex-git: Okay.

src/query/xasl.h
xmilex-git: Okay.

xmilex-git: how about this? 

> #define XASL_CAN_USE_SQ_CACHE	      0x40000   /* subquery can use sq_cache */

src/query/subquery_cache.c
```diff
@@ -0,0 +1,449 @@
...
+ *
+ * Allocates and initializes a new SQ_VAL structure based on the given REGU_VARIABLE. The function handles
+ * different types of REGU_VARIABLE (e.g., TYPE_CONSTANT, TYPE_LIST_ID) appropriately by copying or cloning
+ * the necessary data. It returns a pointer to the newly allocated and initialized SQ_VAL structure.
+ */
+SQ_VAL *
+sq_make_val (THREAD_ENTRY * thread_p, REGU_VARIABLE * val)
+{
+  SQ_VAL *ret;
+  ret = (SQ_VAL *) db_private_alloc (NULL, sizeof (SQ_VAL));
```
youngjinj:
```suggestion
  ret = (SQ_VAL *) db_private_alloc (thread_p, sizeof (SQ_VAL));
```

And `thread_p` is also required in `sq_free_key`.

src/query/subquery_cache.h
```diff
@@ -0,0 +1,70 @@
...
+{
+  /* fields used by both XASL interpreter and regulator */
+  DB_VALUE *dbvalptr;		/* for constant values */
+  QFILE_SORTED_LIST_ID *srlist_id;	/* sorted list identifier for subquery results */
+};
+
+struct sq_key
+{
+  DB_VALUE **dbv_array;
+  int n_elements;
```
youngjinj, line 87: How is this used?
How about checking `SQ_CACHE` when needed?

Is there a case where the number of DB_VALUE lists used as a hash key and the number of DB_VALUE lists searched for are different?

xmilex-git: There is no case where the lengths are different. However, since the cmp function used to look up keys within the hash table cannot check the sq_cache, n_elements is necessary. I will remove the routine that compares lengths.

src/query/subquery_cache.c
xmilex-git: Okay, I'll add `thread_p` to all `db_private_alloc` in `subquery_cache.c`

src/query/subquery_cache.c
```diff
@@ -0,0 +1,449 @@
...
+ * This function attempts to retrieve a value from the cache for a given XASL node. It generates a key based on the XASL node's
+ * structure and looks up the cache for a matching value. If a cached value is found, it is unpacked into the specified regu
+ * variable, and the function returns True. Otherwise, the function updates cache miss counters and returns False.
+ */
+bool
+sq_get (THREAD_ENTRY * thread_p, SQ_KEY * key, XASL_NODE * xasl, REGU_VARIABLE * regu_var)
+{
+  SQ_VAL *ret;
+
+  if (XASL_IS_FLAGED (xasl, XASL_SQ_CACHE) && xasl->sq_cache->ht)
```
youngjinj: I think it is better not to check the flag.

xmilex-git: Okay.

src/query/subquery_cache.h
youngjinj: Yes, I understand.

src/query/query_dump.c
```diff
@@ -3007,6 +3009,23 @@ qdump_print_stats_json (xasl_node * xasl_p, json_t * parent)
...
+      json_object_set_new (sq_cache, "size", json_integer (xasl_p->sq_cache->size));
+      if (xasl_p->sq_cache->enabled)
+	{
+	  json_object_set_new (sq_cache, "status", json_string ("enabled"));
+	}
+      else
+	{
+	  json_object_set_new (sq_cache, "status", json_string ("disabled"));
+	}
+      json_object_set_new (proc, "sq_cache", sq_cache);
```
youngjinj: How about this?
```suggestion
      json_object_set_new (proc, "SUBQUERY_CACHE", sq_cache);
```

src/query/query_dump.c
```diff
@@ -3274,6 +3293,21 @@ qdump_print_stats_text (FILE * fp, xasl_node * xasl_p, int indent)
   qdump_print_stats_text (fp, xasl_p->scan_ptr, indent);
   qdump_print_stats_text (fp, xasl_p->connect_by_ptr, indent);
 
+  if (xasl_p->sq_cache && XASL_IS_FLAGED (xasl_p, XASL_USES_SQ_CACHE))
+    {
+      fprintf (fp, "%*c", indent, ' ');
+      if (xasl_p->sq_cache->enabled)
+	{
+	  fprintf (fp, "SUBQUERY_CACHE (hit: %d, miss: %d, size: %luB, status: enabled)\n",
```
youngjinj: When I tested again, the output of the "B" character made it less readable.
How about explaining in the manual that the unit is bytes and not printing the letter "B"?

There is no unit output in JSON format output.
There is no unit output in time output.

xmilex-git: Okay.

src/query/query_dump.c
xmilex-git: Okay.

src/query/query_executor.c
```diff
@@ -2248,6 +2249,11 @@ qexec_clear_xasl (THREAD_ENTRY * thread_p, xasl_node * xasl, bool is_final)
       pg_cnt += qexec_clear_agg_orderby_const_list (thread_p, xasl, is_final);
     }
 
+  if (xasl->sq_cache != NULL)
```
youngjinj, line 2252: Have you tested whether the `sq_cache_destroy` function should not be executed depending on `is_final`?

xmilex-git: We can adjust the extent of clearing based on `is_final`. However, currently, there doesn't seem to be any part inside the `sq_cache_destroy` that needs to be cleared dynamically based on whether `is_final` is true or not.

src/query/xasl.h
```diff
@@ -949,6 +952,23 @@ struct partition_spec_node
 };
 #endif /* defined (SERVER_MODE) || defined (SA_MODE) */
 
+struct sq_cache
```
youngjinj: I think it is better to have it defined in the `src/query/subquery_cache.h` file.

src/query/xasl.h
```diff
@@ -949,6 +952,23 @@ struct partition_spec_node
...
+{
+  DB_VALUE **sq_key_struct;
+  int n_elements;
```
youngjinj: How about declaring it as `SQ_KEY` type?

src/query/subquery_cache.c
```diff
@@ -0,0 +1,444 @@
...
+ * sq_cache_initialize () - Initializes the cache for a given XASL node.
+ *   return: NO_ERROR if successful, ER_FAILED otherwise.
+ *   xasl(in/out): The XASL node for which the cache is being initialized.
+ *
+ * This function creates a hash table for caching the results of the XASL node. It sets up initial values for cache hit and miss
+ * counters and marks the cache as initialized. The function returns NO_ERROR upon successful initialization, or ER_FAILED if the
+ * hash table could not be created.
+ */
+int
+sq_cache_initialize (THREAD_ENTRY * thread_p, XASL_NODE * xasl)
```
youngjinj: How about only receiving `SQ_CACHE` as a parameter?

src/query/subquery_cache.c
```diff
@@ -0,0 +1,444 @@
...
+			xasl, xasl->sq_cache->stats.hit, xasl->sq_cache->stats.miss, xasl->sq_cache->size);
+	  mht_clear (xasl->sq_cache->ht, sq_rem_func, NULL);
+	  mht_destroy (xasl->sq_cache->ht);
+	  xasl->sq_cache->ht = NULL;
+	}
+      xasl->sq_cache->size_max = 0;
+      xasl->sq_cache->size = 0;
+      xasl->sq_cache->enabled = false;
+      xasl->sq_cache->stats.hit = 0;
+      xasl->sq_cache->stats.miss = 0;
```
youngjinj: How about initializing it in `sq_cache_initialize` except for `enabled`?
I don't think there is a need to initialize it in destroy.

How about receiving `SQ_CACHE` instead of `XASL_NODE`?

src/query/subquery_cache.c
```diff
@@ -0,0 +1,444 @@
...
+ *   xasl(in): The XASL node for which the cache is being destroyed.
+ *
+ * This function destroys the cache associated with a given XASL node. It clears all cache entries and then destroys the hash
+ * table itself. It also resets cache-related flags and counters for the XASL node. This function is called when a XASL node is
+ * no longer needed or before it is deallocated.
+ */
+void
+sq_cache_destroy (XASL_NODE * xasl)
+{
+  int i;
```
youngjinj: Not used.

src/query/subquery_cache.c
```diff
@@ -0,0 +1,444 @@
...
+ */
+void
+sq_free_key (SQ_KEY * key)
+{
+  int i;
+  for (i = 0; i < key->n_elements; i++)
+    {
+      pr_free_ext_value (key->dbv_array[i]);
+    }
+  db_private_free_and_init (NULL, key->dbv_array);
```
youngjinj: The pointer to `THREAD_ENTRY` must not be NULL.

If it is NULL, does it find and use the current thread?

src/query/subquery_cache.c
xmilex-git: In `fetch.c`, it is handled as follows: 

> !(regu_var->xasl->sq_cache->ht && !regu_var->xasl->sq_cache->enabled)

Therefore, it seems appropriate to initialize the corresponding XASL before it is executed again in `destroy`. Receiving `SQ_CACHE` is a good suggestion.

src/query/subquery_cache.c
xmilex-git: I've applied it.

src/query/subquery_cache.c
xmilex-git: I've removed it

src/query/subquery_cache.c
xmilex-git: It finds current thread, I think.

src/query/xasl.h
xmilex-git: I've applied it.

src/query/xasl.h
xmilex-git: I've applied it.

src/query/subquery_cache.c
```diff
@@ -0,0 +1,444 @@
...
+  SQ_VAL *val;
+  const void *ret;
+  UINT64 new_entry_size = 0;
+  int i;
+
+  val = sq_make_val (thread_p, regu_var);
+
+  if (!xasl->sq_cache->ht)
+    {
+      sq_cache_initialize (thread_p, xasl);
```
shparkcubrid: Error handling required. And it seems better to change the location as shown below.
```
qexec_start_mainblock_iterations  <== sq_cache_initialize 
qexec_end_mainblock_iterations  <== sq_cache_destroy
```

xmilex-git: If `destroy` is added to `end_mainblock_iteration`, the issue of statistical information not being aggregated occurs. I will consider adding `init` to `start_mainblock_iteration`.

src/query/fetch.c
```diff
@@ -3871,12 +3872,52 @@ fetch_peek_dbval (THREAD_ENTRY * thread_p, REGU_VARIABLE * regu_var, val_descr *
...
 
-      /* execute linked query */
-      EXECUTE_REGU_VARIABLE_XASL (thread_p, regu_var, vd);
-      if (CHECK_REGU_VARIABLE_XASL_STATUS (regu_var) != XASL_SUCCESS)
+      if (regu_var->xasl && XASL_IS_FLAGED (regu_var->xasl, XASL_USES_SQ_CACHE)
+	  && !(regu_var->xasl->sq_cache->ht && !regu_var->xasl->sq_cache->enabled))
 	{
-	  goto exit_on_error;
+	  SQ_KEY *key;
+	  if ((key = sq_make_key (thread_p, regu_var->xasl)) == NULL)
```
shparkcubrid: I think the key can be dynamically assigned only when calling sq_put().

src/query/xasl.h
```diff
@@ -147,6 +148,7 @@ using ANALYTIC_EVAL_TYPE = cubxasl::analytic_eval_type;
 using PRED_EXPR = cubxasl::pred_expr;
 // *INDENT-ON*
 
+
```
youngjinj:
```suggestion
```

src/query/xasl.h
```diff
@@ -949,6 +952,7 @@ struct partition_spec_node
 };
 #endif /* defined (SERVER_MODE) || defined (SA_MODE) */
 
+
```
youngjinj:
```suggestion
```

src/query/subquery_cache.h
```diff
@@ -0,0 +1,87 @@
...
+extern bool sq_get (THREAD_ENTRY * thread_p, SQ_KEY * key, xasl_node * xasl, REGU_VARIABLE * regu_var);
+extern void sq_cache_destroy (SQ_CACHE * sq_cache);
+extern SQ_KEY *sq_make_key (THREAD_ENTRY * thread_p, xasl_node * xasl);
+extern void sq_free_key (SQ_KEY * key);
+
+#define SQ_CACHE_MIN_HIT_RATIO 9	/* it means 90% */
+
+struct sq_cache
+{
+  SQ_KEY *sq_key_struct;
```
youngjinj, line 53: How about renaming the variable to `key`?

src/query/subquery_cache.h
```diff
@@ -0,0 +1,87 @@
...
+struct sq_key
+{
+  DB_VALUE **dbv_array;
+  int n_elements;
+};
+
+struct sq_val
+{
+  SQ_REGU_VALUE val;
+  REGU_DATATYPE t;
```
youngjinj: How about this?
```suggestion
  REGU_DATATYPE type;
```

src/query/fetch.c
xmilex-git: I applied that on code

src/query/subquery_cache.h
xmilex-git: Okay.

src/query/subquery_cache.h
xmilex-git: how about rename it to `orig_key`. Is it okay?

src/parser/xasl_generation.c
```diff
@@ -27187,3 +27196,405 @@ pt_to_instnum_pred (PARSER_CONTEXT * parser, XASL_NODE * xasl, PT_NODE * pred)
...
+ * and other flags within the structure.
+ *
+ * return : bool - Returns true if the XASL node and its subqueries satisfy the conditions for caching, false otherwise.
+ * parser (in): PARSER_CONTEXT* - Pointer to the parser context, provides contextual information necessary for processing the node.
+ * node (in): PT_NODE* - Pointer to the query node being checked. This node can represent various query elements such as expressions.
+ * xasl (in) : XASL_NODE* - Pointer to the XASL node being checked for caching compatibility.
+ */
+
+int
+pt_check_corr_subquery_hash_result_cache (PARSER_CONTEXT * parser, PT_NODE * node, XASL_NODE * xasl)
```
youngjinj: How about separating the `pt_prepare_sq_cache` function from the `pt_check_corr_subquery_hash_result_cache` function?
Initialization is being done in a function for the purpose of checking.

shparkcubrid (approved):

Hamkua (approved):

beyondykk9 (approved):

youngjinj (approved):

PR #4873, opened 2024-01-12
[CBRD-25158] improve read key btree range scan

ctshim:
http://jira.cubrid.org/browse/CBRD-25158

* improve read key btree range scan
   - The common_prefix_key once read is maintained and used.
   - To extract db_value from the midxkey of the index, cur_key and common_prefix_key are used together.


src/storage/btree.c
```diff
@@ -4280,13 +4302,171 @@ btree_read_record (THREAD_ENTRY * thread_p, BTID_INT * btid, PAGE_PTR pgptr, REC
...
+
+  assert (pgptr == bts->C_page && VPID_EQ (pgbuf_get_vpid_ptr (pgptr), &bts->C_vpid));
+
+  if (bts->common_prefix_size >= 0)
+    {
+      assert (VPID_EQ (&(bts->C_vpid), &bts->cur_common_prefix_page_vpid));
+#if defined(CHECK_VERIFY_COMMON_PREFIX_PAGE_INFO)
+      assert (LSA_EQ (&bts->cur_common_prefix_page_lsa, pgbuf_get_lsa (pgptr)));
+      assert (bts->common_prefix_size == btree_node_get_common_prefix (thread_p, &bts->btid_int, pgptr));
+#endif
```
youngjinj: How about calling `check_validate` ?

src/storage/btree.h
```diff
@@ -192,6 +210,19 @@ struct btree_scan
...
+  //---------------------------------------------------------------------------------------------   
+  /* IMPROVE_RANGE_SCAN_IN_BTREE */
+#if defined(CHECK_VERIFY_COMMON_PREFIX_PAGE_INFO)
+  VPID cur_common_prefix_page_vpid;
+  LOG_LSA cur_common_prefix_page_lsa;
+#endif
+  int common_prefix_size;	// size of common prefix key parts.
+  DB_VALUE common_prefix_key;	// common prefix key part 
+  bool clear_common_prefix_key;	// 
+  bool is_cur_key_compressed;	/* If false, cur_key is a complete key.
```
youngjinj, line 218: I think it can be differentiated by common_prefix_size, but is is_cur_key_compressed needed?
- If prefix_size is -1, unknown.
- If prefix_size is 0, complete key.
- If prefix_size is greater than 0, there is a prefix key.

There seems to be no problem if prefix_size is changed to 0 when pr_midxkey_add_prefix is called.

src/storage/btree.c
```diff
@@ -4280,13 +4302,171 @@ btree_read_record (THREAD_ENTRY * thread_p, BTID_INT * btid, PAGE_PTR pgptr, REC
...
+	  assert_release (false);
 	  return n_prefix;
 	}
     }
 
   return NO_ERROR;
 }
 
+static void
+btree_make_complete_key_including_prefix (BTREE_SCAN * bts, DB_VALUE * common_prefix_key, int common_prefix_size)
```
youngjinj, line 4421: How about passing only the necessary ones like pr_midxkey_add_prefix?

youngjinj, line 4421: This function is no different from pr_midxkey_add_prefix and is not called much.
So I think it is unnecessary.
How about using `pr_midxkey_add_prefix` ?

src/storage/btree.c
```diff
@@ -6177,6 +6357,9 @@ void
 btree_scan_clear_key (BTREE_SCAN * btree_scan)
 {
   btree_clear_key_value (&btree_scan->clear_cur_key, &btree_scan->cur_key);
+#if defined(IMPROVE_RANGE_SCAN_IN_BTREE)
+  btree_clear_common_prefix_page_info (btree_scan);
```
youngjinj: I think the `btree_clear_common_prefix_page_info` function is unnecessary.
This is to clear `bts->common_prefix_key`, but one more step is needed to know this.

src/storage/btree.c
```diff
@@ -14695,6 +14875,7 @@ btree_find_lower_bound_leaf (THREAD_ENTRY * thread_p, BTREE_SCAN * bts, BTREE_ST
       goto exit_on_error;
     }
 
+  COMMON_PREFIX_PAGE_SIZE_RESET (bts);
```
youngjinj: Is there any reason to do it here?
Is initialization of common prefix information necessary to find the lower bound leaf?

src/storage/btree.c
```diff
@@ -4280,13 +4302,171 @@ btree_read_record (THREAD_ENTRY * thread_p, BTID_INT * btid, PAGE_PTR pgptr, REC
...
+}
+
+/*
+ * btree_check_decompress_key () - 
+ *               This is a function to convert the two parts, cur_key and common_prefix_key, into a complete key by combining them with cur_key.
+ *   return:
+ *   bts(in):
+ */
+static void
+btree_check_decompress_key (BTREE_SCAN * bts)
```
youngjinj, line 4451: What is the purpose of this function, check or make?

src/storage/btree.c
```diff
@@ -15775,6 +15956,10 @@ btree_prepare_bts (THREAD_ENTRY * thread_p, BTREE_SCAN * bts, BTID * btid, INDX_
   /* Set other arguments specific to scan type. */
   bts->bts_other = bts_other;
 
+#if defined(IMPROVE_RANGE_SCAN_IN_BTREE)
+  btree_init_common_prefix_page_info (bts);
```
youngjinj: I think the btree_init_common_prefix_page_info function is unnecessary.
```suggestion
  COMMON_PREFIX_PAGE_SIZE_RESET (bts);
  btree_clear_key_value (&bts->clear_common_prefix_key, &bts->common_prefix_key);
  //btree_init_temp_key_value (&bts->clear_common_prefix_key, &bts->common_prefix_key); 
```

src/storage/btree.c
```diff
@@ -16151,6 +16338,7 @@ btree_find_next_index_record_holding_current_helper (THREAD_ENTRY * thread_p, BT
 	      goto exit_on_error;
 	    }
 
+	  COMMON_PREFIX_PAGE_SIZE_RESET (bts);
```
youngjinj: Should this code be placed right after `pgbuf_fix` ?

How about adding a function that wraps `pgbuf_fix` and `pgbuf_unfix` to perform `COMMON_PREFIX_PAGE_SIZE_RESET` ?

src/storage/btree.c
```diff
@@ -16226,13 +16412,35 @@ btree_apply_key_range_and_filter (THREAD_ENTRY * thread_p, BTREE_SCAN * bts, boo
...
+      *is_key_range_satisfied = true;
     }
   else
     {
-      c = btree_compare_key (bts->key_range.upper_key, &bts->cur_key, bts->btid_int.key_type, 1, 1, NULL);
+#if defined(IMPROVE_RANGE_SCAN_IN_BTREE)
+      if (bts->is_cur_key_compressed)
+	{
+	  int start_col = 0;
+	  if (start_col < bts->common_prefix_size)
```
youngjinj: Is this wrong?
```suggestion
	  if (start_col <= bts->common_prefix_size)
```

src/storage/btree.c
```diff
@@ -16212,12 +16400,10 @@ btree_find_next_index_record_holding_current_helper (THREAD_ENTRY * thread_p, BT
  */
 static int
 btree_apply_key_range_and_filter (THREAD_ENTRY * thread_p, BTREE_SCAN * bts, bool is_iss, bool * is_key_range_satisfied,
-				  bool * is_key_filter_satisfied, bool need_to_check_null)
+				  bool * is_key_filter_satisfied)
 {
   int c;			/* comparison result */
```
youngjinj:
```suggestion
  DB_VALUE_COMPARE_RESULT c;	/* comparison result */
```

src/storage/btree.c
```diff
@@ -16683,12 +16955,20 @@ btree_get_next_key_info (THREAD_ENTRY * thread_p, BTID * btid, BTREE_SCAN * bts,
...
+  else
+    {
+      /* No current page fixed. */
+      LSA_SET_NULL (&bts->cur_leaf_lsa);
+    }
 
   if (bts->O_page != NULL)
     {
-      pgbuf_unfix_and_init (thread_p, bts->C_page);
+      pgbuf_unfix_and_init (thread_p, bts->O_page);
```
youngjinj, line 16971: 😱

src/storage/btree.c
```diff
@@ -19381,12 +19661,38 @@ btree_range_opt_check_add_index_key (THREAD_ENTRY * thread_p, BTREE_SCAN * bts,
...
+  if (bts->is_cur_key_compressed)
     {
-      error = pr_midxkey_get_element_nocopy (new_mkey, multi_range_opt->sort_att_idx[i], &new_key_value[i], NULL, NULL);
-      if (error != NO_ERROR)
+      DB_MIDXKEY *prefix_mkey = NULL;
+      int idx;
+      assert (bts->common_prefix_size > 0);
+
+      check_validate (bts);
+      prefix_mkey = db_get_midxkey (&(bts->common_prefix_key));
```
youngjinj: How about using `&(bts->common_prefix_key)` directly?

src/storage/btree.c
```diff
@@ -4280,13 +4302,171 @@ btree_read_record (THREAD_ENTRY * thread_p, BTID_INT * btid, PAGE_PTR pgptr, REC
...
+#endif
+    }
+}
+#else
+#define check_validate(bts) ;
+#endif
+
+#if defined(IMPROVE_RANGE_SCAN_IN_BTREE)
+static int
+btree_read_record_in_leafpage (THREAD_ENTRY * thread_p, PAGE_PTR pgptr, int copy_key, BTREE_SCAN * bts)
```
youngjinj, line 4336: The meaning of the names of the functions `btree_read_record` and `btree_read_record_without_decompression` is clear.
However, the `btree_read_record_in_leafpage` function does not.

How about adding the code from the `btree_read_record_in_leafpage` function to the `btree_read_record` function?

src/storage/btree.c
youngjinj: How about adding a function that wraps pgbuf_fix and pgbuf_unfix to perform COMMON_PREFIX_PAGE_SIZE_RESET?

src/storage/btree.c
ctshim: Currently, btree_read_record_in_leafpage() is called only from btree_range_scan_read_record(). So you could put your code directly in btree_range_scan_read_record().

However, since it needs to be called in the [CBRD-25159] issue that will be added later, we will keep a separate name.
btree_read_record() reads keys from all nodes of B-Tree, but this function is named to mean that it only reads keys from leaf nodes.

Please recommend a good name.

src/storage/btree.c
ctshim: Check items are different.
I will keep it as is.  
Thank you.

src/storage/btree.c
ctshim: Are you asking whether it can be implemented by calling pr_midxkey_add_prefix() directly in btree_check_decompress_key()?
It is possible if the question was asked in that sense.
However, there are plans to call this function directly in other issues in the future.

src/storage/btree.c
ctshim: Check it and make it if necessary.
The name is not appropriate.

How about btree_make_after_checking_completed_key() or  btree_decompress_after_checking_key()?

src/storage/btree.c
ctshim: Are you saying you want to change it to something like a macro?

src/storage/btree.c
ctshim: When bts->C_page changes, COMMON_PREFIX_PAGE_SIZE_RESET() is called to discard the current information.

src/storage/btree.c
ctshim: That makes sense.
But I'll keep it like this.

This is because in other issues, there is a need to perform this content without calling btree_prepare_bts(), and it seems that it can be recognized intuitively by leaving it out as a function.
Even if it is a function call, it will not affect performance.

Thank you.

src/storage/btree.c
ctshim: I wanted to do that too.
But it wasn't simple.

src/storage/btree.c
ctshim: This is a discussion outside of this issue, but I think it was intentionally set to int.
Except checking DB_UNK separately
You don't need to worry about DB_SUBSET, DB_NE, DB_SUPERSET, etc. Could it be that the original developer was intentionally trying to think only of negative and positive numbers?

I think you need to think about the code below.
   `c = -c;`

src/storage/btree.c
ctshim: Even if bts->common_prefix_size=0, it means that there is no information in common_prefix_key.
Thank you.

src/storage/btree.c
ctshim: This is not what we want to cover in this issue, but since it was such an obvious error, we fixed it together.

There may have been a memory leak in the past, but since it is only called when using the **SELECT_KEY_INFO** hint, it is unlikely to occur often on actual sites.

ctshim: I think it would be good to hear others opinion.

src/storage/btree.c
youngjinj: It seems to be the same because it satisfies the following.
```
assert (pgptr == bts->C_page && VPID_EQ (pgbuf_get_vpid_ptr (pgptr), &bts->C_vpid));
```
Compared to `check_validate`, there is one more line below.
```
assert (bts->common_prefix_size == btree_node_get_common_prefix (thread_p, &bts->btid_int, pgptr));
```
I think this line should be included in `check_validate`.

src/storage/btree.h
ctshim: is_cur_key_compressed indicates the state of cur_key, and common_prefix_size is a value for the state of the page.
That is, regardless of common_prefix_size, cur_key may be compressed or uncompressed.

src/storage/btree.c
youngjinj: If we do not create a separate struct inside the BTS in the future, it would be better to clear it directly from `btree_scan_clear_key`.

Are we planning to create a separate struct within `BTS` in the future?

I don't think it needs to be implemented as a macro or function.

src/storage/btree.c
youngjinj: I think the functions for check and make purposes should be separate.

src/storage/btree.c
ctshim: <del> I will not cover this matter here, but will create a separate issue to handle it. </del>
The variable is not used, so no actual lick occurs.

src/storage/btree.c
youngjinj: Because I think it affects readability.

Variables inside `BTS` are initialized directly here.
I think there is no need to differentiate and initialize only the variables inside the `btree_init_common_prefix_page_info` function.

src/storage/btree.c
youngjinj: Wouldn't it be simple to use it selectively only within the `btree` functions?
Even if it is only used in currently changed code, it would be good.

src/storage/btree.c
youngjinj: For backport purposes, it would be better to resolve this issue in another issue.
I am shocked that there was a typo like this until now.

src/storage/btree.c
youngjinj: I agree that that may have been what the previous developer intended.
But I think the below expression in `DB_VALUE_COMPARE_RESULT` is also correct.
```
c = -c
```

src/storage/btree.h
youngjinj: When cur_key changes to the uncompressed state, if common_prefix_size is also changed to 0, isn't is_cur_key_compressed unnecessary?

youngjinj: I understood that the life cycle of common_prefix_size and is_cur_key_compressed changing is different.
If common_prefix_size must remain the same within the same page even if cur_key changes, is_cur_key_compressed is needed.

src/storage/btree.c
youngjinj: The purpose is to express my opinion, so it is okay to end without changes.

src/storage/btree.c
youngjinj: The purpose is to express my opinion, so it is okay to end without changes.

src/storage/btree.c
youngjinj: The purpose is to express my opinion, so it is okay to end without changes.

src/storage/btree.c
youngjinj: The purpose is to express my opinion, so it is okay to end without changes.

src/storage/btree.c
youngjinj: The purpose is to express my opinion, so it is okay to end without changes.

src/storage/btree.c
youngjinj: It was explained that this was to avoid repeated calls within loop statements.

src/storage/btree.c
youngjinj: The `btree_read_record` function reads records without distinguishing between leaf nodes and non-leaf nodes.
However, it is ambiguous that there is an additional `btree_read_record_in_leafpage` function to read records only from leaf nodes.
There is no ambiguity if the `btree_read_record` function calls the `btree_read_record_in_leafpage` function.

The purpose is to express my opinion, so it is okay to end without changes.

src/storage/btree.c
youngjinj: The functions `btree_make_complete_key_including_prefix`, `btree_check_decompress_key`, and `pr_midxkey_add_prefix` have the same purpose.
However, there is no difference in the results of code execution.

The `pr_midxkey_add_prefix` function has different layers.
So my opinion was that the functions `btree_make_complete_key_including_prefix` and `btree_check_decompress_key` should be reduced to one.

I understood that the purpose of the call was different.

The purpose is to express my opinion, so it is okay to end without changes.

src/storage/btree.c
youngjinj: I think it is better to call `COMMON_PREFIX_PAGE_SIZE_RESET` after `bts->C_page = NULL; (line: 16318)`.

And I think the name of the `COMMON_PREFIX_PAGE_SIZE_RESET` macro function should mean initializing meta information about `C_page`. Isn’t that what this macro function does?

src/storage/btree.c
```diff
@@ -24630,8 +24989,15 @@ btree_range_scan_resume (THREAD_ENTRY * thread_p, BTREE_SCAN * bts)
 
   assert (bts->force_restart_from_root || !VPID_ISNULL (&bts->C_vpid));
   assert (bts->C_page == NULL);
-  assert (!DB_IS_NULL (&bts->cur_key));
   assert (!BTS_IS_INDEX_ILS (bts));
+#if defined(IMPROVE_RANGE_SCAN_IN_BTREE)
+  // TODO: 
+  btree_check_decompress_key (bts);
```
youngjinj: What is the purpose of checking for decompression here?

src/storage/btree.c
```diff
@@ -24922,6 +25300,7 @@ btree_range_scan_advance_over_filtered_keys (THREAD_ENTRY * thread_p, BTREE_SCAN
 		  ASSERT_ERROR_AND_SET (error_code);
 		  return error_code;
 		}
+	      COMMON_PREFIX_PAGE_SIZE_RESET (bts);
 	      /* Advance to next node. */
 	      pgbuf_unfix (thread_p, bts->C_page);
```
youngjinj: The order below seems more appropriate for reviewing code.
```suggestion
	      /* Advance to next node. */
	      pgbuf_unfix (thread_p, bts->C_page);
	      COMMON_PREFIX_PAGE_SIZE_RESET (bts);
```

src/storage/btree.c
```diff
@@ -25038,6 +25417,7 @@ btree_range_scan_descending_fix_prev_leaf (THREAD_ENTRY * thread_p, BTREE_SCAN *
       pgbuf_unfix_and_init (thread_p, bts->C_page);
       bts->C_page = prev_leaf;
       VPID_COPY (&bts->C_vpid, &prev_leaf_vpid);
+      COMMON_PREFIX_PAGE_SIZE_RESET (bts);
```
youngjinj: The order below seems more appropriate for reviewing code.
```suggestion
      pgbuf_unfix_and_init (thread_p, bts->C_page);
      COMMON_PREFIX_PAGE_SIZE_RESET (bts);
      bts->C_page = prev_leaf;
      VPID_COPY (&bts->C_vpid, &prev_leaf_vpid);
```

src/storage/btree.c
```diff
@@ -25090,6 +25470,12 @@ btree_range_scan_descending_fix_prev_leaf (THREAD_ENTRY * thread_p, BTREE_SCAN *
       return error_code;
     }
 
+  COMMON_PREFIX_PAGE_SIZE_RESET (bts);
```
youngjinj: What is the purpose of calling the `COMMON_PREFIX_PAGE_SIZE_RESET` macro function?
1. The purpose of initializing existing information after `pgbuf_unfix`.
2. The purpose of initializing existing information after `pgbuf_fix`.
3. Or is there code that needs to be added in the future?

src/storage/btree.c
youngjinj: I understood that COMMON_PREFIX_PAGE_SIZE_RESET had to be called because bts->C_page was changed.

src/storage/btree.c
youngjinj: I understood it to be necessary for cases `1.` and `2.`.
If only `C_page` is changed without calling `pgbuf_unfix`, it becomes case `2.`.

src/storage/btree.c
youngjinj: The purpose is to express my opinion, so it is okay to end without changes.

src/storage/btree.c
youngjinj: The purpose is to express my opinion, so it is okay to end without changes.

src/storage/btree.c
youngjinj: The purpose is to express my opinion, so it is okay to end without changes.

src/storage/btree.c
youngjinj: I thought wrong.

src/storage/btree.c
youngjinj: When `C_page` cannot be reused, we have to find the location of the previous key again and start scanning.
To find the location of the previous key, only `&bts->cur_key` is passed to the `btree_locate_key` function, so it is necessary to add `common_prefix_key` to `cur_key`.

youngjinj (approved):

src/storage/btree.c
ctshim: Overall, I will maintain the current appearance for now.
After reviewing your feedback, some changes were necessary.
Thank you.

src/storage/btree.c
ctshim: I'll keep it as is for now and review it one more time in [CBRD-25159]
Thank you.

src/storage/btree.c
ctshim: Oops,
c = -c ==> error: invalid conversion from ‘int’ to ‘DB_VALUE_COMPARE_RESULT’ [-fpermissive]
c = (DB_VALUE_COMPARE_RESULT)(-(int)c);  ==> Success!

I think it was declared as int because a cast operation is needed.
I will keep it as is.

src/storage/btree.c
ctshim: This will be covered again in [CBRD-25159].
I will review it again then.
Thank you

beyondykk9 (approved):

hgryoo:
Could you remove the blocks the else part of #if defined(IMPROVE_RANGE_SCAN_IN_BTREE)? I can test it previous version with using thg [btree_range_scan_dev](https://github.com/CUBRID/cubrid/tree/feature/btree_range_scan_dev)'s commit id.

hyunikn (approved):

hgryoo (approved):

shparkcubrid (approved):

PR #4874, opened 2024-01-12
[CBRD-25153] Shorten code range markers.

hyunikn:
http://jira.cubrid.org/browse/CBRD-25153

- omit code range markers for AST nodes that cannot be a innermost one that raises a runtime exception
    - DeclRoutine
    - DeclParamIn
    - DeclParamOut
    - Some cases of DeclVar
    - Some cases of DeclConst
    - DeclCursor
    - DeclException
    - ExprCase
    - ExprCond
    - StmtAssign without not-null check
    - StmtBasicLoop
    - StmtBlock
    - StmtExit
    - StmtContinue
    - Some cases of StmtIf 
    - StmtLocalProcCall
    - StmtNull
    - StmtReturn
    - StmtWhileLoop
    - Body
    - ExHandler
    - CondExpr
    - CondStmt
    - cursor open statement's arguments setting code
    - out parameter's variable arguments
- let the code range markers of coercions be those of argument expressions to reduce the markers

### DIFF
--- a/pl_engine/pl_server/src/main/java/com/cubrid/plcsql/compiler/visitor/JavaCodeWriter.java
+++ b/pl_engine/pl_server/src/main/java/com/cubrid/plcsql/compiler/visitor/JavaCodeWriter.java
@@ -1727,9 +1727,7 @@ public class JavaCodeWriter extends AstVisitor<JavaCodeWriter.CodeToResolve> {
 
         return Misc.isEmpty(node.args)
                 ? new CodeTemplate(
-                        "StmtLocalProcCall",
-                        Misc.UNKNOWN_LINE_COLUMN,
-                        block + node.name + "();")
+                        "StmtLocalProcCall", Misc.UNKNOWN_LINE_COLUMN, block + node.name + "();")
                 : new CodeTemplate(
                         "StmtLocalProcCall",
                         Misc.UNKNOWN_LINE_COLUMN,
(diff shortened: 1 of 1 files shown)


ctshim (approved):

pl_engine/pl_server/src/main/java/com/cubrid/plcsql/compiler/visitor/JavaCodeWriter.java
```diff
@@ -2578,19 +2583,18 @@ private static class CodeTemplate implements CodeToResolve {
...
             if (plcsqlLine < 0 && plcsqlColumn < 0) {
-                this.plcsqlLineColumn = null; // do not mark code range in this case
+                // OK
             } else {
                 assert plcsqlLine > 0 && plcsqlColumn > 0
                         : String.format(
                                 "%s - line and column numbers of code templates must be positive integers: (%d, %d)",
                                 astNode, plcsqlLine, plcsqlColumn);
-                this.plcsqlLineColumn = String.format("%d,%d", plcsqlLine, plcsqlColumn);
             }
```
hgryoo: If this block is only for assert, the if-statement seems redundant.

hgryoo (approved):

beyondykk9 (approved):

pl_engine/pl_server/src/main/java/com/cubrid/plcsql/compiler/visitor/JavaCodeWriter.java
hyunikn: It needs to be simplified. Thank you.

PR #4881, opened 2024-01-16
[CBRD-25164] Fix error in covered index query results using function indexes

ctshim:
http://jira.cubrid.org/browse/CBRD-25164

* Fix error in covered index query results using function indexes
 - This problem occurs because the value is not properly obtained during the key filtering process.
 
 * test query
```
drop table if exists tx;
create table tx (id int, v1 int, v2 int, v3 int);
insert into tx values(1,1,10,100), (2,2,20,200);

create index idx on tx(v1,  ln(v2), v3);
select v3 from tx where v1 = 1 and v3 = 100;

drop index idx on tx;
create index idx on tx(v1,  ln(v1), v3);
select v3 from tx where v1 = 1 and v3 = 100;

drop index idx on tx;
create index idx on tx(v1,  ln(v1), v3, v2);
select v2 from tx where v1 = 1 and v3 = 100;

drop index idx on tx;
create index idx on tx(ln(v1), v2, v3);
select v2 from tx where ln(v1) > 0 and v3 = 100;
```

### DIFF
--- a/src/query/query_evaluator.c
+++ b/src/query/query_evaluator.c
@@ -2824,6 +2824,8 @@ eval_key_filter (THREAD_ENTRY * thread_p, DB_VALUE * value, FILTER_INFO * filter
     {
       if (DB_VALUE_TYPE (value) == DB_TYPE_MIDXKEY)
 	{
+	  int func_idx_col_id = filterp->func_idx_col_id;
+
 	  midxkey = db_get_midxkey (value);
 
 	  if (filterp->btree_num_attrs <= 0 || !filterp->btree_attr_ids || !midxkey)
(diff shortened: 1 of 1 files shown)


src/query/query_evaluator.c
```diff
@@ -2824,6 +2824,8 @@ eval_key_filter (THREAD_ENTRY * thread_p, DB_VALUE * value, FILTER_INFO * filter
     {
       if (DB_VALUE_TYPE (value) == DB_TYPE_MIDXKEY)
 	{
+	  int func_idx_col_id = filterp->func_idx_col_id;
```
beyondykk9, line 2827: no necessary to check filterp is null ?

beyondykk9 (approved):

hgryoo (approved):

hyunikn (approved):

PR #4884, opened 2024-01-18
[CBRD-25170] Mistaking NULL return of a built-in function as the default value of the return type

hyunikn:
http://jira.cubrid.org/browse/CBRD-25170

- added wasNull check after every ResultSet.getXXX() call
    - built-in function call
    - field lookup
    - serial value lookup
    - into clause of cursor fetch
    - into clause of Static/Dynamic SQL SELECT

### DIFF
--- a/pl_engine/pl_server/src/main/java/com/cubrid/plcsql/compiler/visitor/JavaCodeWriter.java
+++ b/pl_engine/pl_server/src/main/java/com/cubrid/plcsql/compiler/visitor/JavaCodeWriter.java
@@ -1162,9 +1162,7 @@ public class JavaCodeWriter extends AstVisitor<JavaCodeWriter.CodeToResolve> {
             Coercion c = node.coercions.get(i);
             String idCode = id.javaCode();
             ret.add(String.format("%s = %s;", idCode, c.javaCode(resultStr)));
-            ret.add(
-                    String.format(
-                            "if (%s != null && rs.wasNull()) { %s = null; }", idCode, idCode));
+            ret.add(String.format("if (%1$s != null && rs.wasNull()) { %1$s = null; }", idCode));
 
             i++;
         }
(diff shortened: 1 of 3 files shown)


pl_engine/pl_server/src/main/java/com/cubrid/plcsql/compiler/visitor/JavaCodeWriter.java
```diff
@@ -1157,7 +1160,11 @@ private static String[] getSetIntoVarsCode(StmtCursorFetch node) {
             }
 
             Coercion c = node.coercions.get(i);
-            ret.add(String.format("%s = %s;", id.javaCode(), c.javaCode(resultStr)));
+            String idCode = id.javaCode();
+            ret.add(String.format("%s = %s;", idCode, c.javaCode(resultStr)));
+            ret.add(
+                    String.format(
+                            "if (%s != null && rs.wasNull()) { %s = null; }", idCode, idCode));
```
hgryoo: Nit
```suggestion
                    String.format(
                            "if (%1$s != null && rs.wasNull()) { %1$s = null; }", idCode));
```

pl_engine/pl_server/src/main/java/com/cubrid/plcsql/compiler/visitor/JavaCodeWriter.java
```diff
@@ -1319,14 +1326,17 @@ private String[] getSetResultsCode(StmtSql node, List<ExprId> intoVarList) {
...
 
             Coercion c = node.coercions.get(i);
-            boolean checkNotNull = (id.decl instanceof DeclVar) && ((DeclVar) id.decl).notNull;
-            if (checkNotNull) {
+            String idCode = id.javaCode();
+            ret.add(String.format("%s = %s;", idCode, c.javaCode(resultStr)));
+            ret.add(
+                    String.format(
+                            "if (%s != null && r%%'LEVEL'%%.wasNull()) { %s = null; }",
+                            idCode, idCode));
```
hgryoo: Nit
```suggestion
                    String.format(
                            "if (%1$s != null && r%%'LEVEL'%%.wasNull()) { %1$s = null; }",
                            idCode));
```

hgryoo (approved):

ctshim (approved):

beyondykk9 (approved):

pl_engine/pl_server/src/main/java/com/cubrid/plcsql/compiler/visitor/JavaCodeWriter.java
hyunikn: Good point, Thank you

PR #4889, opened 2024-01-24
[CBRD-25178] core dump occurs when a false query is created due to constant folding.

beyondykk9:
http://jira.cubrid.org/browse/CBRD-25178

If a false query (null query) is created in a CTE query due to constant folding, the xasl of the non-recursive part may become null, causing a problem of referring to a null point.


hgryoo:
I wonder whether the same fix is required for the `recursive_part_xasl`.

hgryoo (approved):

hyunikn (approved):

ctshim (approved):

PR #4893, opened 2024-01-25
[CBRD-25180] prevent temp volume growth, when performing prepare() -> multiple execute()

mhoh3963:
http://jira.cubrid.org/browse/CBRD-25180

The results (temporary volume) is dangled, when preforming execute() multiple time after preparing a sql. This problem does not always occur, but it occurs when there are more than two pages of results and you do not explicitly clear the results after committing. It caused after applying server's auto commit (CBRD-20159, CBRD-22015).

-  modify to perform to close deferred end_queries, excluding the results performed this time at stran_server_auto_commit_or_abort(), if end_query_allowed is false.

### DIFF
--- a/src/communication/network_interface_sr.c
+++ b/src/communication/network_interface_sr.c
@@ -225,6 +225,7 @@ stran_server_auto_commit_or_abort (THREAD_ENTRY * thread_p, unsigned int rid, QU
 				   TRAN_STATE * tran_state, bool * should_conn_reset)
 {
   int error_code, all_error_code, i;
+  int num_query_ids = n_query_ids;
 
   assert (tran_state != NULL && should_conn_reset != NULL && end_query_allowed != NULL);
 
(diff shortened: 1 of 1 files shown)


beyondykk9:
**sqmgr_execute_query** calls **stran_server_auto_commit_or_abort** and **stran_server_auto_commit_or_abort** checks **end_query_allowed** for executing **xqmgr_end_query**.

I suggest that the sqmgr_execute_query logic is modified as follows:
```
stran_server_auto_commit_or_abort (
    thread_p, rid, p_net_Deferred_end_queries, 
    (end_query_allowed) ? n_query_ids : n_query_ids - 1, 
    tran_abort, has_updated, &end_query_allowd, &tran_state, &should_conn_reset
);
```

The **stran_server_auto_commit_or_abort** should be also modified as follows:
```
if (*end_query_allowed == false)
  {
    if (prm_get_bool_value (PRM_ID_DEBUG_AUTOCOMMIT))
      {
         _er_log_debug (ARG_FILE_LINE, "stran_server_auto_commit_or_abort: active transaction.\n");
      }
    /* go through for executing end_query */
  }
```

mhoh3963:
@beyondykk9 
This fix may cause dangling results if xqmgr_end_query fails in stran_server_auto_commit_or_abort().
So, I close this PR to use another fixing way.

PR #4894, opened 2024-01-25
[CBRD-24474] The problem that the allocated memory is not freed even when the lock is released

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-24474

The maximum number of locks maintained is limited by 'LOCK_ESCALATION' system parameter. default is 100,000. If the number of allocated locks exceeds the set value, locks are immediately freed upon retirement.


xmilex-git (approved):

src/base/lock_free.c
```diff
@@ -893,6 +893,20 @@ lf_freelist_retire (LF_TRAN_ENTRY * tran_entry, LF_FREELIST * freelist, void *en
 	}
     }
 
+  /* free entry if alloc_cnt is greater than max_alloc_cnt */
+  if (freelist->alloc_cnt > edesc->max_alloc_cnt)
```
beyondykk9, line 897: it looks correct "freelist->alloc_cnt >= edesc->max_alloc_cnt" ?

shparkcubrid: It would be better to handle the problem of overflow occurring when freelist->alloc_cnt exceeds max int in another issue. '>' is correct because it is free for locks exceeding 'lock escalation'.

beyondykk9 (approved):

src/transaction/lock_manager.c
```diff
@@ -1113,6 +1115,7 @@ lock_initialize_object_hash_table (void)
 
   const int block_count = 2;
   const int block_size = (int) MAX ((lk_Gl.max_obj_locks * LK_RES_RATIO) / block_count, 1);
+  lk_Obj_lock_res_desc.max_alloc_cnt = prm_get_integer_value (PRM_ID_LK_ESCALATION_AT);
```
youngjinj, line 1118: I agree to this change.
But I think it is better to have `max_alloc_cnt` in `LF_FREELIST`.
```
/* Add argument: max_alloc_cnt */
lockfree_hashmap<Key, T>::init ()

  /* Add argument: max_alloc_cnt */
  -> init_as_old ()

    /* Add argument: max_alloc_cnt */
    -> m_old_hash.init ()

      /* Add argument: max_alloc_cnt */
      -> lf_freelist_init ()

        freelist->max_alloc_cnt = max_alloc_cnt;
```

youngjinj (approved):

src/transaction/lock_manager.c
shparkcubrid: Why? The max_alloc already exists in the descriptor. If it's in free_list, it's a redundant.

youngjinj: My opinion is that it should not be in the `descriptor` but in the `freelist`.

src/base/lock_free.c
```diff
@@ -893,6 +893,20 @@ lf_freelist_retire (LF_TRAN_ENTRY * tran_entry, LF_FREELIST * freelist, void *en
...
+    {
+      edesc->f_free (entry);
```
youngjinj, line 899: To return the freed memory to the OS without keeping it in the heap, `malloc_trim (0);` must be called.
However, because it is called once per LOCK, the performance of `lf_freelist_retire` decreases.
If the performance of `lf_freelist_retire` decreases, the performance of COMMIT also decreases.
```
/* lock count: 1,000,000 */

/* edesc->f_free (entry); */
csql> commit;
Execute OK. (9.542932 sec)

/* edesc->f_free (entry); -> malloc_trim (0); */
csql> commit;
Execute OK. (126.872019 sec)
```

shparkcubrid: This issue is intended to defend against cases where OOM has occurred in the past. To prevent OOM, I think it's enough for the OS to recognize the memory as reusable.

src/transaction/lock_manager.c
shparkcubrid: I don't think so.

hornetmj:
@shparkcubrid @joohok conducts the code review on my behalf.

joohok (approved):

PR #4896, opened 2024-01-26
[RND-2068] for ctp test

beyondykk9:
http://jira.cubrid.org/browse/CBRD-XXXX

This is a temporary for CTP


beyondykk9:
This was a temporary PR. I close the PR.

PR #4904, opened 2024-01-29
[CBRD-25200] Add system parameter 'memory_monitoring'

Rudeus:
http://jira.cubrid.org/browse/CBRD-25200

Add system parameter 'memory_monitoring'.

Using in configuration file
memory_monitoring = (y/n)

+) minor change : add cscope.files in .gitignore


conf/cubrid.conf
```diff
@@ -76,4 +76,4 @@ log_max_archives=0
 # for deduplicate (-1 ~ 14, -1 forces explicit statements to also be ignored.)
 deduplicate_key_level=-1
 print_index_detail=no
-
+memory_monitoring=n
```
hornetmj:
```suggestion
memory_monitoring=no
```

hornetmj: For consistency with print_index_detail=`no`

hornetmj (approved):

conf/cubrid.conf
wrlawodms: Is it a spec to set it NO explicitly though it's by default? If it's not, I think we should remove this.

src/base/system_parameter.c
```diff
@@ -737,6 +737,8 @@ static const char sysprm_ha_conf_file_name[] = "cubrid_ha.conf";
 
 #define PRM_NAME_ORACLE_STYLE_DIVIDE "oracle_style_divide"
 
+#define PRM_NAME_MEMORY_MONITORING "memory_monitoring"
```
wrlawodms: How about this to say it a flag.
```suggestion
#define PRM_NAME_MEMORY_MONITORING "enable_memory_monitoring"
```

wrlawodms (approved):
Approval with some opinions.

src/base/system_parameter.c
hornetmj: > Is it a spec to set it NO explicitly though it's by default? If it's not, I think we should remove this.

He is right. I think it would be good to romove it.

conf/cubrid.conf
Rudeus: Done. I remove it.

src/base/system_parameter.c
Rudeus: Done. I remove it. And I think your suggestion "enable_memory_monitoring" is better than "memory_monitoring". I change it.

conf/cubrid.conf
```diff
@@ -76,4 +76,3 @@ log_max_archives=0
 # for deduplicate (-1 ~ 14, -1 forces explicit statements to also be ignored.)
 deduplicate_key_level=-1
 print_index_detail=no
-
```
joohok, line 79: unrelated change

joohok (approved):

PR #4905, opened 2024-01-29
[CBRD-25201] Add 'memory_monitor' class with implementing internal functions of add/sub_stat()

Rudeus:
http://jira.cubrid.org/browse/CBRD-25201

Add 'memory_monitor' class with implementing internal function of add/sub_stat()
 - Create memory_monitor class
  - add m_server_name for registering server name to memory_monitor when cub_server starts.
 - Implement internal functions 'make_tag_name()'.


src/base/memory_monitor_sr.hpp
```diff
@@ -0,0 +1,46 @@
...
+{
+  class memory_monitor
+  {
+    public:
+      memory_monitor (const char *server_name);
+      ~memory_monitor () {}
+
+    private:
+      std::string make_tag_name (const char *file, const int line);
+      int generate_checksum (int tag_id, uint64_t size);
```
```suggestion
      static std::string make_tag_name (const char *file, const int line);
      static int generate_checksum (int tag_id, uint64_t size);
```

src/base/memory_monitor_sr.cpp
```diff
@@ -0,0 +1,63 @@
...
+    if (pos != std::string::npos)
+      {
+	filecopy = filecopy.substr (pos);
+      }
+
+    ret = filecopy + ':' + std::to_string (line);
+    return ret;
+  }
+
+  int memory_monitor::generate_checksum (int tag_id, uint64_t size)
```
wrlawodms: How about using a "magic string" like "MMON" instead of hashing like MD5? Do we need to map tad_id + size to a distinguishable value?

src/base/memory_monitor_sr.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+    private:
+      std::string m_server_name;
```
wrlawodms, line 47: Nit. how about introducing this when it's used?

src/base/memory_monitor_sr.cpp
```diff
@@ -0,0 +1,63 @@
...
+    // Find the last occurrence of "src" in the path
+    size_t pos = filecopy.rfind ("src");
+
+    if (pos != std::string::npos)
+      {
+	filecopy = filecopy.substr (pos);
+      }
+
+    ret = filecopy + ':' + std::to_string (line);
+    return ret;
```
```suggestion
    return filecopy + ':' + std::to_string (line);
```

src/base/memory_monitor_sr.cpp
```diff
@@ -0,0 +1,63 @@
...
+  memory_monitor::memory_monitor (const char *server_name)
+    : m_server_name {server_name} {}
+
+  std::string memory_monitor::make_tag_name (const char *file, const int line)
+  {
+    std::string filecopy (file);
+    std::string ret;
+
+    // Find the last occurrence of "src" in the path
+    size_t pos = filecopy.rfind ("src");
```
wrlawodms: What if the name of a directory of file includes `src`? If it's not the case for now, it make "src" as a special string that we can't use for files.

I know it's a kind of "requirement", but do we really need to includes all the path starting with the `src`? I think the file name is good enough. Please share your opinion on this.

wrlawodms:
Please elaborate on the A/C on the Jira issue to help us know what kind of tests this PS has been passed.

src/base/memory_monitor_sr.cpp
Rudeus: "rfind(str)" method finds the last matching str that the function meets. So I think this cannot be a matter.

src/base/memory_monitor_sr.cpp
Rudeus: I add `ret.reserve()` to reduce string copy overhead. So I think it is better to use `ret`.

src/base/memory_monitor_sr.cpp
Rudeus: I agree about your opinion. Replacing the hash function with a "MAGIC STRING" reduces the overhead of using the hash function while still fundamentally fulfilling its intended purpose. Therefore, adding the "MAGIC STRING" seems sufficient to be done in the next PR, implementation of the add/sub_stat() functions. Hence, the generate_checksum() function will be removed.

src/base/memory_monitor_sr.hpp

src/base/memory_monitor_sr.hpp
Rudeus: OK. I think it's enough to add the description at PR description.

src/base/memory_monitor_sr.cpp
wrlawodms: Does `ret.reserve()` work for that? From what I know, 
`ret = filecopy + ':' + std::to_string (line);` abandons the reserved space and just copy the RHS.

src/base/memory_monitor_sr.cpp
wrlawodms: What if 
- `src/a/module_a_src.c`
-  `src/a/a_src_b/.c`
-  `src/a/a_descriptor.c`,, oh, we have `record_descriptor.cpp`. Does it work with this?

Rudeus: After considering it, I have decided to maintain the existing method. However, to address the scenario you mentioned in your comment, I changed "/src/" to the target string of rfind. Additionally, in the path of the 3rd-party code, "/src/" may exist. To internally verify this, when building in debug mode, I compare the positions of the leftmost and rightmost occurrences of "/src/" to ensure that only one "/src/" exists. This decision was made after discussion with @hornetmj.

src/base/memory_monitor_sr.cpp
Rudeus: Oh I find it. I change it.

src/base/memory_monitor_sr.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+namespace cubmem
+{
+  memory_monitor::memory_monitor (const char *server_name)
+    : m_server_name {server_name} {}
+
+  std::string memory_monitor::make_tag_name (const char *file, const int line)
+  {
+    std::string filecopy (file);
+    std::string target ("/src/");
```
joohok, line 37: Is the memory monitoring module considering only the Linux environment? 
I'm asking because I'm not sure;

src/base/memory_monitor_sr.cpp
wrlawodms: Honestly, I'm still against this in that I doubt this job is worth it with the cost by many string copies and comparisons. This is a function called every time a memory is allocated, isn't it?

Anyway, that's just my concern and I'll follow your decision. We can revisit if it really matters :).

src/base/memory_monitor_sr.cpp
Rudeus: Yes.

wrlawodms (approved):

joohok (approved):

src/base/memory_monitor_sr.cpp
joohok: OK, then we don't need to consider the delimiter used in Windows.

src/base/memory_monitor_sr.hpp
```diff
@@ -0,0 +1,51 @@
...
+      ~memory_monitor () {}
+
+      memory_monitor (const memory_monitor &) = delete;
+      memory_monitor (memory_monitor &&) = delete;
+
+      memory_monitor &operator = (const memory_monitor &) = delete;
+      memory_monitor &operator = (memory_monitor &&) = delete;
+
+    private:
+      static std::string make_tag_name (const char *file, const int line);
```
hornetmj, line 44: Should make_tag_name() be static?

Rudeus: It means that the function may not change member variable.

hornetmj: I think you just add const keyword like `std::string make_tag_name (const char *file, const int line) const`. It meet your requirement.

src/base/memory_monitor_sr.cpp
hornetmj: > with the cost by many string copies and comparisons.

I aggree with this comment. One of option I think is to use c style approach using char * and strrstr(). Let's see the result of performance test and visit again if it is necessary.

src/base/memory_monitor_sr.hpp
wrlawodms: In my opinion, static is more proper than const for this function. This function doesn't depend on any member variable. The `const` qualifier is for a method to depend on member variable, but not to modify it.

wrlawodms: Actually, this function is not a member function(dependent on an instance), but a class function.

hornetmj: In my point of view, static keyword allow us it can call without creating object. However this function is always called after creating memory_monitor object and it doesn't need to call without object. What is your opinion in this point?

Rudeus: I change it static to const.

wrlawodms: As you said, this method may not be called if there is no instance, but it doesn't need any instance. I don't see an advantage to avoid calling it without an instance.

wrlawodms: hmm.. to me, the `static` keyword is exactly for this kind of method, and `const` is for another case.

hornetmj: Both opinion from you and me guarantee through private: and static keyword. I didn't consider private: keyword. So, I follow your suggestion.

hornetmj (approved):

PR #4908, opened 2024-01-29
[CBRD-25207] count(*) value is incorrect after no-logging option loaddb

beyondykk9:
http://jira.cubrid.org/browse/CBRD-25207

After loaddb with no-logging option, the result of "SELECT count(*)" query always comes out as 0.

### DIFF
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -3825,7 +3825,7 @@ log_sysop_commit_internal (THREAD_ENTRY * thread_p, LOG_REC_SYSOP_END * log_reco
     }
 
   if ((LSA_ISNULL (&tdes->tail_lsa) || LSA_LE (&tdes->tail_lsa, LOG_TDES_LAST_SYSOP_PARENT_LSA (tdes)))
-      && log_record->type == LOG_SYSOP_END_COMMIT)
+      && (log_record->type == LOG_SYSOP_END_COMMIT || log_No_logging))
     {
       /* No change. */
       assert (LSA_ISNULL (&LOG_TDES_LAST_SYSOP (tdes)->posp_lsa));
--- a/src/transaction/log_manager.c
+++ b/src/transaction/log_manager.c
@@ -1018,6 +1018,18 @@ log_set_no_logging (void)
 }
 
 /*
+ * This function is for an external interface
+ * to process statistical information of the B-tree index header.
+ * In a no logging environment, the statistical information of the B-tree index header
+ * should not depend on the updated log. This is a function to check this.
+ */
+bool
+log_is_no_logging (void)
+{
+  return log_No_logging;
+}
+
+/*
  * log_initialize - Initialize the log manager
  *
  * return: nothing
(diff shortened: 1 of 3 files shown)


src/transaction/log_tran_table.c
```diff
@@ -5042,7 +5042,7 @@ logtb_reflect_global_unique_stats_to_btree (THREAD_ENTRY * thread_p)
        stats = (GLOBAL_UNIQUE_STATS *) lf_hash_iterate (&it))
     {
       /* reflect only if some changes were logged */
-      if (!LSA_ISNULL (&stats->last_log_lsa))
+      if (log_is_no_logging () || !LSA_ISNULL (&stats->last_log_lsa))
```
hgryoo, line 5045: I wonder `btree_reflect_global_unique_statistics()` works well even though `stats->last_log_lsa` is NULL. I don't have enough background of this routine. Could you please explain in detail?

beyondykk9: In **logtb_update_global_unique_stats_by_delta**, the **last_log_lsa** of stats is copying the lsa of the transaction by LSA_COPY (&stats->last_log_lsa, &tdes->tail_lsa).
This is to check with log's lsa that stats have changed.
However, in the case of no-logging, it is not possible to know whether changes have been made through lsa, so statistical information must always be reflected.

src/transaction/log_manager.c
```diff
@@ -3824,7 +3836,8 @@ log_sysop_commit_internal (THREAD_ENTRY * thread_p, LOG_REC_SYSOP_END * log_reco
        * we don't actually allow empty logical system operation because it might hide a logic flaw. however, there are
        * unusual cases when a logical operation does not really require logging (see RVPGBUF_FLUSH_PAGE). if you create
        * such a case, you should add a dummy log record to trick this assert. */
-      assert (!LSA_ISNULL (&tdes->tail_lsa) && LSA_GT (&tdes->tail_lsa, LOG_TDES_LAST_SYSOP_PARENT_LSA (tdes)));
+      assert (log_No_logging || (!LSA_ISNULL (&tdes->tail_lsa)
```
wrlawodms: I doubt it is correct and safe.. It must be the last option to drop or change the existing assertion. The assert makes much sense to me. In this `else`, a logical undo log record (or other sysop_end) will be appended below, but if there is no redo log records have been generated for the logical operation, it means that there is an logical UNDO for no physical REDO modifications. The UNDO assumes that there are proper physical changes.

Analyzing related-code, most of time, it would be not a problem because during `log_No_logging`, if a crash occurs, it just stops the entire RECOVERY or ROLLBACK with a fatal error. However, it's wrong logically, so it has a potential error. An example would be the `log_sysop_abort`, which seems try to undo even during `log_No_logging` when a sysop encounters an internal error.

The way I see it, it does more make sense to change the `if` state for no log record case above
` if ((LSA_ISNULL (&tdes->tail_lsa) || LSA_LE (&tdes->tail_lsa, LOG_TDES_LAST_SYSOP_PARENT_LSA (tdes)))
      && log_record->type == LOG_SYSOP_END_COMMIT)`
->
` if ((LSA_ISNULL (&tdes->tail_lsa) || LSA_LE (&tdes->tail_lsa, LOG_TDES_LAST_SYSOP_PARENT_LSA (tdes)))
      && (log_record->type == LOG_SYSOP_END_COMMIT || log_No_logging))`

However, it needs some additional tests for the change. Please share your opinion on it.

wrlawodms (changes requested):

src/transaction/log_manager.c
beyondykk9: I don't know much about recovery and log management.
All I know is that this function is supposed to not return an error and assert if something goes wrong.
Therefore, assert is only made in case a critical error occurs.
Your opinion is to use an if statement instead of assert, but ultimately else should be handled as assert(false).

wrlawodms: > Your opinion is to use an if statement instead of assert, but ultimately else should be handled as assert(false).

I don't understand what you mean. What I meant is even if `log_No_logging` is true, the existing assert should be met. The issue case will be go through the `if` case.

beyondykk9: I misunderstood your comment.
I now understand that it is not an assert but a higher level if statement. I aggree with you.

wrlawodms: This assert should be reverted as well.

beyondykk9: I modified it as your comment.

wrlawodms (approved):
The change in the `log_sysop_commit_internal` seems hard to test for every case. Let's see the regressions tests :)

hgryoo (approved):

ctshim (approved):

hornetmj:
@beyondykk9 @wrlawodms's review replaces mine.

hornetmj (approved):

PR #4910, opened 2024-01-30
[CBRD-25206] Change the statistical information of the B-tree index root header to 64 bits.

beyondykk9:
http://jira.cubrid.org/browse/CBRD-25206

Regarding the count(*) optimization using the statistical information of the B-tree index header introduced from version 11.2, it is necessary to unify num_nulls, num_oids, and num_keys, which were treated as 32 bits + 10 bits, to 64 bits in consideration of backward compatibility.
Since the disk image of version 11.4 (fig-cake) is not compatible with previous versions, it is necessary to process it in 64-bit from this version to simplify the logic and improve performance.

### DIFF
--- a/src/storage/btree_load.h
+++ b/src/storage/btree_load.h
@@ -207,9 +207,9 @@ typedef struct btree_root_header BTREE_ROOT_HEADER;
 struct btree_root_header
 {
   BTREE_NODE_HEADER node;
-  long long num_oids;		/* Number of OIDs stored in the Btree */
-  long long num_nulls;		/* Number of NULLs (they aren't stored) */
-  long long num_keys;		/* Number of unique keys in the Btree */
+  INT64 num_oids;		/* Number of OIDs stored in the Btree */
+  INT64 num_nulls;		/* Number of NULLs (they aren't stored) */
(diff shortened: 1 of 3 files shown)


ctshim (approved):

src/storage/btree_load.h
```diff
@@ -207,21 +207,13 @@ typedef struct btree_root_header BTREE_ROOT_HEADER;
 struct btree_root_header
 {
   BTREE_NODE_HEADER node;
-  int num_oids;			/* Number of OIDs stored in the Btree */
-  int num_nulls;		/* Number of NULLs (they aren't stored) */
-  int num_keys;			/* Number of unique keys in the Btree */
+  long long num_oids;		/* Number of OIDs stored in the Btree */
+  long long num_nulls;		/* Number of NULLs (they aren't stored) */
+  long long num_keys;		/* Number of unique keys in the Btree */
```
hgryoo: How about using `UINT64`?

beyondykk9: When these values are not unique keys, they may be set to -1.

hgryoo: then you can use `INT64`.

hgryoo (approved):

PR #4915, opened 2024-01-30
[CBRD-25180] Occur temp volume growth, when performing prepare() -> multiple execute().

mhoh3963:
http://jira.cubrid.org/browse/CBRD-25180

**Description**
The results (temporary volume) is dangled, when preforming execute() multiple time after preparing a sql. This problem does not always occur, but it occurs if all of the following conditions are met.
 - transaction mode is auto commit mode.
 - the resultset is forward and it is not read until the end.
 - the resultset explicitly doesn't close the results after execute() and fetch().
 - the result's page of execute() are more than two pages.

**Resolution**
-  modify the code to be net_Deferred_end_queries count = 0 only if end_query_result returned from server is NO_ERROR, instead of unconditionally changing net_Deferred_end_queries count = 0.

### DIFF
--- a/src/communication/network_interface_cl.c
+++ b/src/communication/network_interface_cl.c
@@ -7222,6 +7222,7 @@ qmgr_execute_query (const XASL_ID * xasl_id, QUERY_ID * query_idp, int dbval_cnt
       if (IS_QUERY_EXECUTE_WITH_COMMIT (flag))
 	{
 	  ptr = or_unpack_int (ptr, &end_query_result);
+	  /* If all net_Deferred_end_queries are closed on the server, the value of end_query_result is NO_ERROR. */
 	  if (end_query_result == NO_ERROR)
 	    {
 	      net_Deferred_end_queries_count = 0;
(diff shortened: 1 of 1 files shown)


src/communication/network_interface_cl.c
```diff
@@ -7223,6 +7222,10 @@ qmgr_execute_query (const XASL_ID * xasl_id, QUERY_ID * query_idp, int dbval_cnt
       if (IS_QUERY_EXECUTE_WITH_COMMIT (flag))
 	{
 	  ptr = or_unpack_int (ptr, &end_query_result);
+	  if (end_query_result == NO_ERROR)
+	    {
+	      net_Deferred_end_queries_count = 0;
```
beyondykk9, line 7227: NO_ERROR is already 0. It looks meaningless.

src/communication/network_interface_cl.c
```diff
@@ -7223,6 +7222,10 @@ qmgr_execute_query (const XASL_ID * xasl_id, QUERY_ID * query_idp, int dbval_cnt
       if (IS_QUERY_EXECUTE_WITH_COMMIT (flag))
 	{
 	  ptr = or_unpack_int (ptr, &end_query_result);
+	  if (end_query_result == NO_ERROR)
```
beyondykk9, line 7227: It would be better to add a comment to explain the logic.

mhoh3963: I added comment by your opinion, how about it ?

beyondykk9: Good.

beyondykk9 (approved):

beyondykk9 (approved):

PR #4917, opened 2024-01-30
[CBRD-25208] Set java.io.tmpdir according to the CUBRID_TMP in Java SP server

hgryoo:
http://jira.cubrid.org/browse/CBRD-25208

### DIFF
--- a/src/jsp/com/cubrid/jsp/Server.java
+++ b/src/jsp/com/cubrid/jsp/Server.java
@@ -52,6 +52,7 @@ public class Server {
     private static String spPath;
     private static String rootPath;
     private static String udsPath;
+    private static String tmpPath;
 
     private static List<String> jvmArguments = null;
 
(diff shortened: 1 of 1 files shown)


src/jsp/com/cubrid/jsp/Server.java
```diff
@@ -86,6 +87,12 @@ private Server(
             loggingThread = new LoggingThread(logFilePath);
             loggingThread.start();
 
+            tmpPath = System.getProperty("CUBRID_TMP");
+            if (tmpPath == null) {
+                tmpPath = rootPath + File.separator + "tmp";
```
mhoh3963, line 92: If CUBRID_TMP is null, is tmpPath $CUBRID/tmp ?
How about changing default tmpPath to $CUBRID/var/CUBRID_SOCK ?

hgryoo: java.io.tmpdir path is not only for socket file, but Any file that a user create by createTmpFile() API in Java SP.

beyondykk9 (approved):

ctshim (approved):

src/jsp/com/cubrid/jsp/Server.java
mhoh3963: OK, i got it.

mhoh3963 (approved):

PR #4918, opened 2024-01-30
[CBRD-25202] Implement 'get_alloc_size()' function with API

Rudeus:
http://jira.cubrid.org/browse/CBRD-25202

 - Implement 'get_alloc_size()' function
 - Implement 'mmon_get_alloc_size()' API
 - Implement 'mmon_is_mem_tracked()' API
 - add magic number using magic string "MMON"

### DIFF
--- a/src/base/memory_monitor_sr.cpp
+++ b/src/base/memory_monitor_sr.cpp
@@ -41,12 +41,11 @@ struct mmon_metainfo
 
 namespace cubmem
 {
-  static const std::string magic_string ("MMON");
   memory_monitor *mmon_Gl = nullptr;
 
   memory_monitor::memory_monitor (const char *server_name)
     : m_server_name {server_name},
-      m_magic_number {*reinterpret_cast <const int *> (magic_string.c_str())}
+      m_magic_number {*reinterpret_cast <const int *> ("MMON")}
   {}
 
   size_t memory_monitor::get_allocated_size (const char *ptr)
--- a/src/base/memory_monitor_sr.cpp
+++ b/src/base/memory_monitor_sr.cpp
@@ -27,10 +27,6 @@
 
 #include "memory_monitor_sr.hpp"
 
-#ifndef HAVE_USR_INCLUDE_MALLOC_H
-#define HAVE_USR_INCLUDE_MALLOC_H
-#endif
-
 typedef struct mmon_metainfo MMON_METAINFO;
 struct mmon_metainfo
 {
--- a/src/base/memory_monitor_sr.cpp
+++ b/src/base/memory_monitor_sr.cpp
@@ -45,14 +45,17 @@ namespace cubmem
   memory_monitor *mmon_Gl = nullptr;
 
   memory_monitor::memory_monitor (const char *server_name)
-    : m_server_name {server_name}
-  {
-    memcpy (&m_magic_number, magic_string.c_str (), sizeof (this->m_magic_number));
-  }
+    : m_server_name {server_name},
+      m_magic_number {*reinterpret_cast <const int *> (magic_string.c_str())}
+  {}
 
   size_t memory_monitor::get_allocated_size (const char *ptr)
   {
+#if defined(WINDOWS)
+    size_t allocated_size = _msize ((void *)ptr);
+#else
     size_t allocated_size = malloc_usable_size ((void *)ptr);
+#endif // WINDOWS
 
     if (allocated_size <= MMON_ALLOC_META_SIZE)
       {
@@ -74,7 +77,11 @@ namespace cubmem
   std::string memory_monitor::make_tag_name (const char *file, const int line)
   {
     std::string filecopy (file);
+#if defined(WINDOWS)
+    std::string target ("\\src\\");
+#else
     std::string target ("/src/");
+#endif // WINDOWS
 
     // Find the last occurrence of "src" in the path
     size_t pos = filecopy.rfind (target);
--- a/src/base/memory_monitor_sr.cpp
+++ b/src/base/memory_monitor_sr.cpp
@@ -34,7 +34,7 @@
 typedef struct mmon_metainfo MMON_METAINFO;
 struct mmon_metainfo
 {
-  uint64_t alloc_size;
+  uint64_t allocated_size;
   int tag_id;
   int magic_number;
 };
@@ -50,14 +50,14 @@ namespace cubmem
     memcpy (&m_magic_number, magic_string.c_str (), sizeof (int));
   }
 
-  size_t memory_monitor::get_alloc_size (const char *ptr)
+  size_t memory_monitor::get_allocated_size (const char *ptr)
   {
-    size_t alloc_size = malloc_usable_size ((void *)ptr);
-    const char *meta_ptr = ptr + alloc_size - MMON_ALLOC_META_SIZE;
+    size_t allocated_size = malloc_usable_size ((void *)ptr);
+    const char *meta_ptr = ptr + allocated_size - MMON_ALLOC_META_SIZE;
 
-    if (alloc_size <= MMON_ALLOC_META_SIZE)
+    if (allocated_size <= MMON_ALLOC_META_SIZE)
       {
-	return alloc_size;
+	return allocated_size;
       }
 
     assert (meta_ptr > ptr);
@@ -65,10 +65,10 @@ namespace cubmem
 
     if (metainfo->magic_number == m_magic_number)
       {
-	alloc_size = (size_t) metainfo->alloc_size - MMON_ALLOC_META_SIZE;
+	allocated_size = (size_t) metainfo->allocated_size - MMON_ALLOC_META_SIZE;
       }
 
-    return alloc_size;
+    return allocated_size;
   }
 
   std::string memory_monitor::make_tag_name (const char *file, const int line)
@@ -94,16 +94,16 @@ namespace cubmem
 
 using namespace cubmem;
 
-bool mmon_is_mem_tracked ()
+bool mmon_is_mem_traced ()
 {
   return (mmon_Gl != nullptr);
 }
 
-size_t mmon_get_alloc_size (char *ptr)
+size_t mmon_get_allocated_size (char *ptr)
 {
-  if (mmon_is_mem_tracked ())
+  if (mmon_is_mem_traced ())
     {
-      return mmon_Gl->get_alloc_size (ptr);
+      return mmon_Gl->get_allocated_size (ptr);
     }
   // unreachable
   return 0;
--- a/src/base/memory_monitor_sr.cpp
+++ b/src/base/memory_monitor_sr.cpp
@@ -47,20 +47,20 @@ namespace cubmem
   memory_monitor::memory_monitor (const char *server_name)
     : m_server_name {server_name}
   {
-    memcpy (&m_magic_number, magic_string.c_str (), sizeof (int));
+    memcpy (&m_magic_number, magic_string.c_str (), sizeof (this->m_magic_number));
   }
 
   size_t memory_monitor::get_allocated_size (const char *ptr)
   {
     size_t allocated_size = malloc_usable_size ((void *)ptr);
-    const char *meta_ptr = ptr + allocated_size - MMON_ALLOC_META_SIZE;
 
     if (allocated_size <= MMON_ALLOC_META_SIZE)
       {
 	return allocated_size;
       }
 
-    assert (meta_ptr > ptr);
+    const char *meta_ptr = ptr + allocated_size - MMON_ALLOC_META_SIZE;
+
     const MMON_METAINFO *metainfo = (const MMON_METAINFO *) meta_ptr;
 
     if (metainfo->magic_number == m_magic_number)
@@ -94,17 +94,18 @@ namespace cubmem
 
 using namespace cubmem;
 
-bool mmon_is_mem_traced ()
+bool mmon_is_memory_monitor_enabled ()
 {
   return (mmon_Gl != nullptr);
 }
 
 size_t mmon_get_allocated_size (char *ptr)
 {
-  if (mmon_is_mem_traced ())
+  if (mmon_is_memory_monitor_enabled ())
     {
       return mmon_Gl->get_allocated_size (ptr);
     }
   // unreachable
+  assert (false);
   return 0;
 }
--- a/src/base/memory_monitor_sr.cpp
+++ b/src/base/memory_monitor_sr.cpp
@@ -20,18 +20,58 @@
  * memory_monitor_sr.cpp - Implementation of memory monitor module
  */
 
+#include <malloc.h>
+#include <cassert>
 #include <cstring>
 #include <algorithm>
-#include <cassert>
 
+#include "openssl/md5.h"
 #include "memory_monitor_sr.hpp"
+
+#ifndef HAVE_USR_INCLUDE_MALLOC_H
+#define HAVE_USR_INCLUDE_MALLOC_H
+#endif
+
+typedef struct mmon_metainfo
+{
+  uint64_t size;
+  int tag_id;
+  int magic_number;
+} MMON_METAINFO;
 
 namespace cubmem
 {
-  memory_monitor::memory_monitor (const char *server_name)
-    : m_server_name {server_name} {}
+  memory_monitor *mmon_Gl = nullptr;
 
-  static std::string memory_monitor::make_tag_name (const char *file, const int line)
+  memory_monitor::memory_monitor (const char *server_name)
+    : m_server_name {server_name}
+  {
+    std::string magic_string ("MMON");
+    memcpy (&magic, magic_string.c_str (), sizeof (int));
+  }
+
+  size_t memory_monitor::get_alloc_size (char *ptr)
+  {
+    size_t ret;
+    size_t alloc_size = malloc_usable_size (ptr);
+    char *meta_ptr = ptr + alloc_size - MMON_ALLOC_META_SIZE;
+    MMON_METAINFO metainfo;
+
+    memcpy (&metainfo, meta_ptr, MMON_ALLOC_META_SIZE);
+
+    if (metainfo.magic_number == magic)
+      {
+	ret = (size_t) metainfo.size - MMON_ALLOC_META_SIZE;
+      }
+    else
+      {
+	ret = alloc_size;
+      }
+
+    return ret;
+  }
+
+  std::string memory_monitor::make_tag_name (const char *file, const int line)
   {
     std::string filecopy (file);
     std::string target ("/src/");
@@ -50,4 +90,21 @@ namespace cubmem
 
     return filecopy + ':' + std::to_string (line);
   }
+}
+
+using namespace cubmem;
+
+bool mmon_is_mem_tracked ()
+{
+  return (mmon_Gl != nullptr);
+}
+
+size_t mmon_get_alloc_size (char *ptr)
+{
+  if (mmon_is_mem_tracked ())
+    {
+      return mmon_Gl->get_alloc_size (ptr);
+    }
+  // unreachable
+  return 0;
 }
--- a/src/base/memory_monitor_sr.cpp
+++ b/src/base/memory_monitor_sr.cpp
@@ -31,31 +31,31 @@
 #define HAVE_USR_INCLUDE_MALLOC_H
 #endif
 
-typedef struct mmon_metainfo
+typedef struct mmon_metainfo MMON_METAINFO;
+struct mmon_metainfo
 {
-  uint64_t size;
+  uint64_t alloc_size;
   int tag_id;
   int magic_number;
-} MMON_METAINFO;
+};
 
 namespace cubmem
 {
+  static const std::string magic_string ("MMON");
   memory_monitor *mmon_Gl = nullptr;
 
   memory_monitor::memory_monitor (const char *server_name)
     : m_server_name {server_name}
   {
-    std::string magic_string ("MMON");
-    memcpy (&magic, magic_string.c_str (), sizeof (int));
+    memcpy (&m_magic_number, magic_string.c_str (), sizeof (int));
   }
 
   size_t memory_monitor::get_alloc_size (const char *ptr)
   {
-    size_t ret;
     size_t alloc_size = malloc_usable_size ((void *)ptr);
     const char *meta_ptr = ptr + alloc_size - MMON_ALLOC_META_SIZE;
 
-    if (alloc_size < MMON_ALLOC_META_SIZE)
+    if (alloc_size <= MMON_ALLOC_META_SIZE)
       {
 	return alloc_size;
       }
@@ -63,16 +63,12 @@ namespace cubmem
     assert (meta_ptr > ptr);
     const MMON_METAINFO *metainfo = (const MMON_METAINFO *) meta_ptr;
 
-    if (metainfo->magic_number == magic)
+    if (metainfo->magic_number == m_magic_number)
       {
-	ret = (size_t) metainfo->size - MMON_ALLOC_META_SIZE;
-      }
-    else
-      {
-	ret = alloc_size;
+	alloc_size = (size_t) metainfo->alloc_size - MMON_ALLOC_META_SIZE;
       }
 
-    return ret;
+    return alloc_size;
   }
 
   std::string memory_monitor::make_tag_name (const char *file, const int line)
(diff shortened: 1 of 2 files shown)


src/base/memory_monitor_sr.hpp
```diff
@@ -24,8 +24,15 @@
...
 #define _MEMORY_MONITOR_SR_HPP_
 
+#include <stdint.h>
 #include <string>
 
+// IMPORTANT!!
+// This meta size is related with allocation byte align
+// Don't adjust it freely
+// 8 byte size + 4 byte tag + 4 byte magicnumber
+#define MMON_ALLOC_META_SIZE 16
```
joohok: If this is used only for memory_monitor, 
then how about declare this constant in the class ? 
`static constexpr int alloc_meta_size = 16; `?

Rudeus: It will be used allocation wrapping functions, like `cub_alloc()`.

src/base/memory_monitor_sr.cpp
```diff
@@ -51,3 +90,20 @@ namespace cubmem
...
+}
+
+size_t mmon_get_alloc_size (char *ptr)
+{
+  if (mmon_is_mem_tracked ())
+    {
+      return mmon_Gl->get_alloc_size (ptr);
+    }
+  // unreachable
+  return 0;
```
wrlawodms: If this is unreachable
```suggestion
  assert (mmon_is_mem_tracked ()); // unreachable
  
  return mmon_Gl->get_alloc_size (ptr);
```

src/base/memory_monitor_sr.hpp
```diff
@@ -40,12 +47,18 @@ namespace cubmem
...
 
+    public:
+      size_t get_alloc_size (char *ptr);
+
     private:
       static std::string make_tag_name (const char *file, const int line);
 
     private:
       std::string m_server_name;
+      int magic;
```
wrlawodms: How about using char*? I'm suggesting to make it const.
```suggestion
      static constexpr char* magic = "MMON";
      static constexpr int magic_size = 4;
```

src/base/memory_monitor_sr.hpp
wrlawodms: Then, how about `static size_t memory_monitor::get_alloc_meta_size()`?
The magic is in the `memory_monitor` and other attributes are probably dependent on the class, aren't they? This should be in at least `cubmem` namespace as well.

src/base/memory_monitor_sr.cpp
```diff
@@ -20,18 +20,57 @@
...
   memory_monitor::memory_monitor (const char *server_name)
-    : m_server_name {server_name} {}
+    : m_server_name {server_name}
+  {
+    std::string magic_string ("MMON");
+    memcpy (&magic, magic_string.c_str (), sizeof (int));
+  }
 
-  static std::string memory_monitor::make_tag_name (const char *file, const int line)
+  size_t memory_monitor::get_alloc_size (char *ptr)
```
wrlawodms: It seems to be a `static` function. What do you think? Will some meta info be referred to like the tag map or something in the class?

```suggestion
  size_t memory_monitor::get_alloc_size (const char *ptr)
```

src/base/memory_monitor_sr.cpp
```diff
@@ -20,18 +20,57 @@
...
+  {
+    size_t ret;
+    size_t alloc_size = malloc_usable_size (ptr);
+    char *meta_ptr = ptr + alloc_size - MMON_ALLOC_META_SIZE;
```
```suggestion
    const char * meta_ptr = ptr + alloc_size - MMON_ALLOC_META_SIZE;
    assert (meta_ptr > ptr);
```

src/base/memory_monitor_sr.cpp
```diff
@@ -20,18 +20,57 @@
...
+    MMON_METAINFO metainfo;
+
+    memcpy (&metainfo, meta_ptr, MMON_ALLOC_META_SIZE);
```
wrlawodms: I think we can remove the copy.
```suggestion
    const MMON_METAINFO * metainfo = (const MMON_METAINFO*) meta_ptr;
```

src/base/memory_monitor_sr.cpp
Rudeus: get_alloc_size() checks `magic` member variable for checking scope.

src/base/memory_monitor_sr.cpp
Rudeus: done.

src/base/memory_monitor_sr.cpp
Rudeus: `assert` is not called in the release build, following the approach you suggested may lead to a SIGSEGV. Is this intentional, given that it is unreachable?

src/base/memory_monitor_sr.hpp
Rudeus: When memory_monitoring option is turned on, cub_alloc will do allocation like this:
malloc (size + MMON_ALLOC_META_SIZE);
I think your suggestion will make all the allocations `malloc (size + mmon_get_alloc_meta_size())`.

src/base/memory_monitor_sr.hpp
Rudeus: I think magic string makes a lot of calls `strcmp()` for checking about the memory is in the memory_monitor scope. So I decided to make it integer.

src/base/memory_monitor_sr.cpp

src/base/memory_monitor_sr.hpp
```diff
@@ -40,12 +47,18 @@ namespace cubmem
       memory_monitor &operator = (const memory_monitor &) = delete;
       memory_monitor &operator = (memory_monitor &&) = delete;
 
+    public:
+      size_t get_alloc_size (const char *ptr);
```
hornetmj:
```suggestion
      size_t get_alloced_size (const char *ptr);
```

src/base/memory_monitor_sr.cpp
```diff
@@ -20,18 +20,62 @@
...
+#include <cassert>
 #include <cstring>
 #include <algorithm>
-#include <cassert>
 
 #include "memory_monitor_sr.hpp"
 
+#ifndef HAVE_USR_INCLUDE_MALLOC_H
+#define HAVE_USR_INCLUDE_MALLOC_H
+#endif
```
hornetmj: What is the purpose?

src/base/memory_monitor_sr.cpp
```diff
@@ -20,18 +20,62 @@
...
+
+typedef struct mmon_metainfo
+{
+  uint64_t size;
+  int tag_id;
+  int magic_number;
+} MMON_METAINFO;
```
hornetmj: I recommand this style for the readability.

```suggestion
typedef struct mmon_metainfo MMON_METAINFO;
typedef struct mmon_metainfo
{
  uint64_t size;
  int tag_id;
  int magic_number;
};
```

src/base/memory_monitor_sr.hpp
```diff
@@ -40,12 +47,18 @@ namespace cubmem
...
+
     private:
       static std::string make_tag_name (const char *file, const int line);
 
     private:
       std::string m_server_name;
+      int magic;
```
hornetmj:
```suggestion
      int m_magic_number;
```

src/base/memory_monitor_sr.cpp
```diff
@@ -20,18 +20,62 @@
...
 namespace cubmem
 {
+  memory_monitor *mmon_Gl = nullptr;
+
   memory_monitor::memory_monitor (const char *server_name)
-    : m_server_name {server_name} {}
+    : m_server_name {server_name}
+  {
+    std::string magic_string ("MMON");
+    memcpy (&magic, magic_string.c_str (), sizeof (int));
```
```suggestion
    memcpy (&magic, magic_string.c_str (), sizeof (this->magic));
```

src/base/memory_monitor_sr.cpp
```diff
@@ -20,18 +20,62 @@
...
+
 namespace cubmem
 {
+  memory_monitor *mmon_Gl = nullptr;
+
   memory_monitor::memory_monitor (const char *server_name)
-    : m_server_name {server_name} {}
+    : m_server_name {server_name}
+  {
+    std::string magic_string ("MMON");
```
hornetmj: "MMON" should be define somewhere rather then using character string directly.

src/base/memory_monitor_sr.cpp
```diff
@@ -20,18 +20,62 @@
...
 
 #include "memory_monitor_sr.hpp"
 
+#ifndef HAVE_USR_INCLUDE_MALLOC_H
+#define HAVE_USR_INCLUDE_MALLOC_H
+#endif
+
+typedef struct mmon_metainfo
+{
+  uint64_t size;
```
hornetmj: NIT:

`alloced_size` is more helpful when I read the code.
```suggestion
  uint64_t alloced_size;
```

src/base/memory_monitor_sr.cpp
```diff
@@ -20,18 +20,62 @@
...
+    memcpy (&magic, magic_string.c_str (), sizeof (int));
+  }
+
+  size_t memory_monitor::get_alloc_size (const char *ptr)
+  {
+    size_t ret;
+    size_t alloc_size = malloc_usable_size ((void *)ptr);
+    const char *meta_ptr = ptr + alloc_size - MMON_ALLOC_META_SIZE;
+
+    if (alloc_size < MMON_ALLOC_META_SIZE)
```
```suggestion
    if (alloc_size <= MMON_ALLOC_META_SIZE)
```

src/base/memory_monitor_sr.cpp
```diff
@@ -20,18 +20,62 @@
...
-    : m_server_name {server_name} {}
+    : m_server_name {server_name}
+  {
+    std::string magic_string ("MMON");
+    memcpy (&magic, magic_string.c_str (), sizeof (int));
+  }
+
+  size_t memory_monitor::get_alloc_size (const char *ptr)
+  {
+    size_t ret;
```
hornetmj: How about reuse `alloc_size` local variable? The `ret` is not good readability.

src/base/memory_monitor_sr.cpp
```diff
@@ -51,3 +95,20 @@ namespace cubmem
     return filecopy + ':' + std::to_string (line);
   }
 }
+
+using namespace cubmem;
+
+bool mmon_is_mem_tracked ()
```
hornetmj: NIT:

```suggestion
bool mmon_is_tracked_mem ()
```

src/base/memory_monitor_sr.cpp
Rudeus: `memory_monitor_sr.cpp` has to include `malloc.h` because of `malloc_usable_size()`. But when user includes `malloc.h`, user has to define `HAVE_USR_INCLUDE_MALLOC_H`.

src/base/memory_monitor_sr.cpp
hornetmj: What is the meaning of `tag_` in this context? How about `mem_id` or `tracked_id` if there is no special meaning of `tag_`?

src/base/memory_monitor_sr.cpp
Rudeus: I think `alloc_size` or `allocated_size` is better because "alloc" is an abbreviation for "allocate", so "alloced" is not appropriate. (I think it should be "alloc-ed.")

src/base/memory_monitor_sr.cpp
Rudeus: I moved it as `static const`.

src/base/memory_monitor_sr.cpp
Rudeus: I think `sizeof(int)` is more intuitive than `sizeof(this->magic)`. I want to know why are you think your suggestion is better.

src/base/memory_monitor_sr.cpp
Rudeus: I understand it. Done.

src/base/memory_monitor_sr.cpp

src/base/memory_monitor_sr.hpp
Rudeus: I think this suggestion is inappropriate because `alloced` can make misunderstanding of word. I think `get_alloc_size` is enough but if changed, I think `get_allocated_size` would be appropriate.

src/base/memory_monitor_sr.hpp

src/base/memory_monitor_sr.hpp
hornetmj: We usually use `alloced` to show something already allocated. You can easily find this if you execute `grep -r 'alloced' *` in the cubrid src directory. I think `alloced` is enough to show the meaning and I prefer it. but, if you want to use `allocated`, then that is also good to me.

Rudeus: I change it to `allocated`.

src/base/memory_monitor_sr.cpp
Rudeus: I change it to `allocated`.

src/base/memory_monitor_sr.cpp
Rudeus: This function check whether memory monitoring option is on or not. But I think your suggestion could get meaning like "This memory is tracked (about the memory pointer)" rather than "cubrid memory monitoring option is turned on". What about `mmon_is_mem_traced()`?

src/base/memory_monitor_sr.cpp
hornetmj: Both sizeof(variable) and sizeof(type) are acceptable, but in my opinion, sizeof(variable) is more commonly used for the following reasons:
1. For arrays, you must use sizeof(variable) to get the size of the array. Even if it is not an array, sizeof(variable) should be used to ensure consistent code. (It may be an exception to templates, but historically arrays were introduced first.)
2. If the type of the variable changes when using sizeof(type), finding the location where sizeof() was used is more difficult than when using sizeof(variable).

src/base/memory_monitor_sr.cpp
hornetmj: If you just want to show the meaning `cubrid memory monitoring option is turned on`, then how about `mmon_is_memory_monitoring_enabled`?

src/base/memory_monitor_sr.hpp
wrlawodms: It seems you are probably concerned about the overhead of function calls..

src/base/memory_monitor_sr.cpp
```diff
@@ -20,18 +20,58 @@
...
+
+  size_t memory_monitor::get_allocated_size (const char *ptr)
+  {
+    size_t allocated_size = malloc_usable_size ((void *)ptr);
+    const char *meta_ptr = ptr + allocated_size - MMON_ALLOC_META_SIZE;
+
+    if (allocated_size <= MMON_ALLOC_META_SIZE)
+      {
+	return allocated_size;
+      }
```
```suggestion
    size_t allocated_size = malloc_usable_size ((void *)ptr);
    if (allocated_size <= MMON_ALLOC_META_SIZE)
      {
	return allocated_size;
      }
      
    const char *meta_ptr = ptr + allocated_size - MMON_ALLOC_META_SIZE;
```

src/base/memory_monitor_sr.cpp
```diff
@@ -20,18 +20,58 @@
...
 
-  static std::string memory_monitor::make_tag_name (const char *file, const int line)
+    assert (meta_ptr > ptr);
```
wrlawodms: This can be removed with the `if-return` above.
```suggestion
```

src/base/memory_monitor_sr.hpp
```diff
@@ -40,12 +47,18 @@ namespace cubmem
       memory_monitor &operator = (const memory_monitor &) = delete;
       memory_monitor &operator = (memory_monitor &&) = delete;
 
+    public:
+      size_t get_allocated_size (const char *ptr);
```
wrlawodms, line 52: How is this used? I can't find stuff on the Jira issue or PR. To review the correctness. Does it just return the size of "user space" in the allocated space? When is it used?

src/base/memory_monitor_sr.cpp
wrlawodms: OK. What happens it returns 0? what does it mean? How is is used?

src/base/memory_monitor_sr.hpp
wrlawodms: Fair enough

src/base/memory_monitor_sr.cpp
wrlawodms: Is the value of the `magic` different for each instance? It seems to be static as well.

src/base/memory_monitor_sr.hpp
wrlawodms: Please remind me if it's described in the meeting. Better to describe it in the Jira issue. Please give me a slack. For reviewers, it's hard to remember every single thing.

src/base/memory_monitor_sr.cpp
```diff
@@ -20,18 +20,58 @@
...
 namespace cubmem
 {
+  static const std::string magic_string ("MMON");
+  memory_monitor *mmon_Gl = nullptr;
+
   memory_monitor::memory_monitor (const char *server_name)
-    : m_server_name {server_name} {}
+    : m_server_name {server_name}
+  {
+    memcpy (&m_magic_number, magic_string.c_str (), sizeof (int));
```
wrlawodms: How about just setting `magic` to a random integer like `12345678` if you want it to be an integer? Then, it can be `static` and `constexpr` easily without this dynamic copy. Or, tell me the point of the human-readable magic string "MMON" if I'm missing.

src/base/memory_monitor_sr.hpp
wrlawodms: Ah.. probably it's used to subtract the memory usage, that occurred to me.

src/base/memory_monitor_sr.hpp
joohok: What I thought was, to easily understand that this constant is associated with the memory monitoring module, it would be better to declare it statically within the class. It facilitates namespace management as well.

Rudeus: Oh I understand it. I change it to `static constexpr cubmem::MMON_ALLOC_META_SIZE`.

src/base/memory_monitor_sr.cpp
Rudeus: This function is called only when the 'memory_monitoring' option is enabled in cub_realloc(), so this function cannot be called in situations where the memory_monitor is NULL. Therefore, if this function returns 0, it indicates a serious bug. While I initially considered handling this as an error, if your suggestion implies that a SIGSEGV should occur for such a serious bug, I believe a more intuitive approach is needed. The action of calling `return mmon_Gl->get_allocated_size()` is not intuitive because it would cause a SIGSEGV, but it requires understanding the context of the function's operation. I agree with adding an assert.

src/base/memory_monitor_sr.cpp
Rudeus: In my knowledge base, `static` function means "This member function doesn't use member variable.". But I think `magic` (is changed `m_magic_number`) should be a member variable. What do you think?

src/base/memory_monitor_sr.cpp
Rudeus: http://jira.cubrid.org/browse/CBRD-25196
I intended to show the relation tag_name(like "base/heap_file.c:496") <-> tag_id <-> memory usage.

src/base/memory_monitor_sr.cpp
Rudeus: OK. I understand it. Done.

src/base/memory_monitor_sr.cpp
Rudeus: I change it to `mmon_is_memory_monitor_enabled`.

src/base/memory_monitor_sr.cpp
Rudeus: I think this is a remnant of the C-based implementation.. Thanks. I change it.

src/base/memory_monitor_sr.cpp

src/base/memory_monitor_sr.hpp
Rudeus: I write the purpose of this function since the first time of issue opening..
"This function is used to getting user-requested size of memory for wrapping function of realloc."

src/base/memory_monitor_sr.cpp
Rudeus: I wish the magic number wasn't so easy, and it would be nice if it indicated where it's being used. Since this operation occurs only once when the memory_monitor is initially created, I don't think it's a major issue. If the way of setting this magic number is significantly flawed in terms of implementation, functionality, or performance, I'd appreciate an explanation. Otherwise, I don't feel strongly about changing it.

src/base/memory_monitor_sr.cpp
wrlawodms: I see. I agree that this feature should not cause a SIGSEGV on an error. To achieve this, we should handle when it returns 0 by the caller. Let's see when the caller code comes in :).

src/base/memory_monitor_sr.hpp
wrlawodms: I wondered what happens if it returns 0, so I wanted to know when it is used, but I couldn't find stuff about this. Let's see in a following PR.

src/base/memory_monitor_sr.cpp
wrlawodms: The way I see it, the `magic` is an instance-dependent variable. I think that should be `static`, and this function should get static along.

src/base/memory_monitor_sr.cpp
wrlawodms: I don't think the performance is a thing here. I think I haven't described my intention. From https://github.com/CUBRID/cubrid/pull/4918#discussion_r1472299166 to here, what I'm trying to do is, I insist is, the `magic` or `m_magic_number` should be "constant". This must not be changed through the entire time, so I'd like to prevent it at all. And, the way I see it, it could be `static` as well.

In the previous comment, I agree that it would be costly if it's string. So, this time, I'm suggesting a way to keep it an integer.

> I wish the magic number wasn't so easy, and it would be nice if it indicated where it's being used.

Agreed. If I were you, I'd want to do so. But, it doesn't seem necessary if we can resolve the issue I described.

wrlawodms:
Please fix the `build-windows` error.

src/base/memory_monitor_sr.cpp
Rudeus: I agree about the magic number should be "constant". But I don't think the magic number should be "static". I think "static" is optional property of the magic number.
Then, how about making `m_magic_number` just a "constant" variable?
`const int m_magic_number = *reinterpret_cast<const int*> (magic_string.c_str ())`

Rudeus:
> Please fix the `build-windows` error.

Done.

wrlawodms (approved):

src/base/memory_monitor_sr.cpp
```diff
@@ -20,21 +20,67 @@
...
-    : m_server_name {server_name} {}
+    : m_server_name {server_name},
+      m_magic_number {*reinterpret_cast <const int *> ("MMON")}
+  {}
 
-  static std::string memory_monitor::make_tag_name (const char *file, const int line)
+  size_t memory_monitor::get_allocated_size (const char *ptr)
+  {
+#if defined(WINDOWS)
+    size_t allocated_size = _msize ((void *)ptr);
```
joohok, line 50: I heard that memory monitoring does not support WINDOWS platform. 
Has this constraint been lifted ?

joohok (approved):

src/base/memory_monitor_sr.cpp
Rudeus: This change is for success window build, not for support WINDOWS. This topic seems to require a discussion on safety aspects possibly after all implementations are completed.

src/base/memory_monitor_sr.cpp
hornetmj: I heard from you offline this is needed to compile later. So, I request you add this when you need to this to complie and discuss it. Because I think this macro only releated to re-define `struct mallinfo` when I read the manual.

Rudeus: I removed the define. Let's talk about it later issue :)

src/base/memory_monitor_sr.cpp
hornetmj: Rather than handling each case, why not exclude all in Windows?

hornetmj (approved):

PR #4921, opened 2024-01-31
[CBRD-24871] Add AUTHID Pragma syntax for stored procedure/function

hgryoo:
http://jira.cubrid.org/browse/CBRD-24871

Add the following sytnax in CUBRID SQL parser and PL/CSQL parser
```
[AUTHID [DEFINER|OWNER] | AUTHID [CALLER|CURRENT_USER]]
```


pl_engine/pl_server/src/main/antlr/PlcLexer.g4
```diff
@@ -45,13 +45,15 @@ TRUNCATE:                     T R U N C A T E   { staticSqlParenMatch++; checkFi
...
 AS:                           A S ;
+AUTHID:                       A U T H I D ;
 AUTONOMOUS_TRANSACTION:       A U T O N O M O U S '_' T R A N S A C T I O N ;
 BEGIN:                        B E G I N ;
 BETWEEN:                      B E T W E E N ;
 BIGINT:                       B I G I N T ;
 BOOLEAN:                      B O O L E A N ;
 BY:                           B Y ;
 CASE:                         C A S E ;
+CALLER:                       C A L L E R;
```
hyunikn: Oracle syntax only allows AUTHID CURRENT_USER or AUTHID DEFINER.
CALLER and OWNER seems unnecessary.

pl_engine/pl_server/src/main/antlr/PlcParser.g4
```diff
@@ -40,7 +40,7 @@ create_routine
 
 routine_definition
     : (PROCEDURE | FUNCTION) identifier ( (LPAREN parameter_list RPAREN)? | LPAREN RPAREN ) (RETURN type_spec)?
-      (IS | AS) (LANGUAGE PLCSQL)? seq_of_declare_specs? body (SEMICOLON)?
+      (authid_spec)? (IS | AS) (LANGUAGE PLCSQL)? seq_of_declare_specs? body (SEMICOLON)?
```
hyunikn, line 43: Because the routine_definition non-terminal is shared by local procedure/functions, we need to check if the authid_spec is present in a local procedure/function definition and report an error if present.
Note that a similar check is done for LANGUAGE PLCSQL clause in visitRoutine_definition() in ParseTreeConverter.java.

pl_engine/pl_server/src/main/java/com/cubrid/plcsql/compiler/ParseTreeConverter.java
```diff
@@ -1172,6 +1172,14 @@ public DeclRoutine visitRoutine_definition(Routine_definitionContext ctx) {
                     "illegal keywords LANGUAGE PLCSQL for a local procedure/function");
         }
 
+        if (ctx.authid_spec() != null && symbolStack.getCurrentScope().level > 1) {
+            int[] lineColumn = Misc.getLineColumnOf(ctx);
+            throw new SyntaxError(
+                    lineColumn[0],
+                    lineColumn[1],
+                    "illegal keywords AUTHID for a local procedure/function");
```
hyunikn: keywords --> keyword

hyunikn (approved):

pl_engine/pl_server/src/main/antlr/PlcLexer.g4
hgryoo: While Oracle does not allow the use of CALLER and OWNER, incorporating support for these keywords would be beneficial for users seeking a better understanding of this feature. This is particularly relevant as the term 'OWNER' is extensively used in various contexts within CUBRID, whereas 'DEFINER' is not.

src/parser/csql_grammar.y
```diff
@@ -1100,6 +1101,7 @@ static char *g_plcsql_text;
 %token ATTACH
 %token ATTRIBUTE
 %token AVG
+%token AUTHID
```
ctshim: Do  you need to register as a keyword?
If so, please refer to the keywords[] variable in the keyword.c file.
If it can be used as an identifier, please refer to the identifier definition in csql_grammar.y.

Please also review the tokens added below.

hgryoo: As Oracle, they don't have to be reserved as an identifier. Thank you for the review!

beyondykk9 (approved):

ctshim (approved):

PR #4926, opened 2024-02-01
[CBRD-25176] Failure when creating fk of deduplicate attribute

ctshim:
http://jira.cubrid.org/browse/CBRD-25176

* Failed to create fk index with deduplicate attribute after changing the structure of MIDXKEY

This is to fix a malfunction issue that was affected by changes in the MIDXKEY structure in CBRD-24988.
AS-IS:
     After cloning MIDXKEY, use a trick to compare by reducing the number of columns by only one.
TO-BE:
     Create and use a new key normally.

### DIFF
--- a/src/storage/btree.c
+++ b/src/storage/btree.c
@@ -5951,7 +5951,7 @@ btree_remake_foreign_key_with_PK (THREAD_ENTRY * thread_p, BTID * btid, DB_VALUE
   DB_MIDXKEY midxkey;
   TP_DOMAIN *tp_dom = NULL;
   TP_DOMAIN *setdomain_ptr = NULL;
-  DB_VALUE new_key_dbvals_ary[8];
+  DB_VALUE new_key_dbvals_array[8];
   DB_VALUE *new_key_dbvals = NULL;
   DB_VALUE *dbvals_ptr = NULL;
   ATTR_ID last_attrid;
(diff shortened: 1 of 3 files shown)


src/storage/btree.c
```diff
@@ -6066,6 +6081,73 @@ btree_remake_foreign_key_with_PK (THREAD_ENTRY * thread_p, BTID * btid, DB_VALUE
...
+    }
+  for (i = 0; i < pk_column_cnt; i++)
+    {
+      db_make_null (&(dbvals_ptr[i]));
+    }
+
+  for (i = 0; i < pk_column_cnt; i++)
+    {
+      pr_midxkey_get_element_nocopy (&(fk_key->data.midxkey), i, &dbvals_ptr[i], NULL, NULL);
+    }
```
hgryoo:
```suggestion
  for (i = 0; i < pk_column_cnt; i++)
    {
      db_make_null (&dbvals_ptr[i]);
      pr_midxkey_get_element_nocopy (&(fk_key->data.midxkey), i, &dbvals_ptr[i], NULL, NULL);
    }
```

hgryoo (approved):

src/storage/btree.c
ctshim: Thank you.

src/storage/btree.c
```diff
@@ -5951,6 +5951,7 @@ btree_remake_foreign_key_with_PK (THREAD_ENTRY * thread_p, BTID * btid, DB_VALUE
   DB_MIDXKEY midxkey;
   TP_DOMAIN *tp_dom = NULL;
   TP_DOMAIN *setdomain_ptr = NULL;
+  DB_VALUE new_key_dbvals_ary[8];
```
beyondykk9:
```suggestion
  DB_VALUE new_key_dbval_list[8];
```
"ary" is an unfamiliar postfix.

beyondykk9 (approved):

hyunikn (approved):

src/storage/btree.c
ctshim: I will modify it with `new_key_dbvals_array`.

PR #4927, opened 2024-02-01
[CBRD-25075] Add a column for AUTHID in _db_stored_procedure catalog

hgryoo:
http://jira.cubrid.org/browse/CBRD-25075

Add directive column for AUTHID

### DIFF
--- a/src/object/schema_system_catalog_install_query_spec.cpp
+++ b/src/object/schema_system_catalog_install_query_spec.cpp
@@ -1172,6 +1172,7 @@ sm_define_view_stored_procedure_spec (void)
 	    "END AS [return_type], "
 	  "[sp].[arg_count] AS [arg_count], "
 	  "CASE [sp].[lang] WHEN 0 THEN 'PLCSQL' WHEN 1 THEN 'JAVA' ELSE 'UNKNOWN' END AS [lang], "
+          "CASE [sp].[directive] & 1 WHEN 0 THEN 'OWNER' ELSE 'CALLER' END AS [authid], "
 	  "[sp].[target] AS [target], "
 	  "CAST ([sp].[owner].[name] AS VARCHAR(255)) AS [owner], " /* string -> varchar(255) */
 	  "[sp].[comment] AS [comment] "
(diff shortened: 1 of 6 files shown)


src/object/schema_system_catalog_install_query_spec.cpp
```diff
@@ -1172,6 +1172,7 @@ sm_define_view_stored_procedure_spec (void)
 	    "END AS [return_type], "
 	  "[sp].[arg_count] AS [arg_count], "
 	  "CASE [sp].[lang] WHEN 0 THEN 'PLCSQL' WHEN 1 THEN 'JAVA' ELSE 'UNKNOWN' END AS [lang], "
+          "CASE [sp].[directive] & 1 WHEN 0 THEN 'OWNER' ELSE 'CALLER' END AS [authid], "
```
hyunikn: Considering Oracle compatibility, DEFINER and CURRENT_USER seem more suitable.

hgryoo: Good point. thank you

src/object/schema_system_catalog_install_query_spec.cpp
```diff
@@ -1172,6 +1172,7 @@ sm_define_view_stored_procedure_spec (void)
 	    "END AS [return_type], "
 	  "[sp].[arg_count] AS [arg_count], "
 	  "CASE [sp].[lang] WHEN 0 THEN 'PLCSQL' WHEN 1 THEN 'JAVA' ELSE 'UNKNOWN' END AS [lang], "
+          "CASE [sp].[directive] & 1 WHEN 0 THEN 'DEFINER' ELSE 'CURRENT_USER' END AS [authid], "
```
ctshim, line 1175: If _directive != 0_, is it judged as 'DEFINER'?
It may be clearer to use the equal operation instead of the bit operator, such as "[sp].[directive] = 1".

hgryoo: Sorry for not to mention about this.
The directvie column will be used for other pragma attributes. such as:
- Autonomous transaction
- function result cache
- any other hints

hyunikn (approved):

src/object/schema_system_catalog_install_query_spec.cpp
beyondykk9: Do you mean it requires a separate bit allocation for the directive as below?

autonomous transaction = 2
function result cache = 4
any other hints =8

hgryoo: Exactly

beyondykk9 (approved):

src/object/schema_system_catalog_install_query_spec.cpp
ctshim: OK, I see.

ctshim (approved):

PR #4932, opened 2024-02-02
[CBRD-25043] refactor functions related to isspace, tolower, and toupper

ctshim:
http://jira.cubrid.org/browse/CBRD-25043

* Refactor the functions in the following list to inline.
  -  char_islower, char_isupper, char_isalpha, char_isdigit, char_isalnum, 
  -  char_isspace, char_iseol, char_isxdigit, char_tolower, char_toupper, 
  - char_isupper_iso8859, char_islower_iso8859, char_tolower_iso8859, char_toupper_iso8859
*  Each function can immediately check and obtain values from a pre-prepared array.


src/base/chartype.c
```diff
@@ -27,188 +27,240 @@
...
+      str[len] = '\0';
+    }
+
+  return (str);
 }
 
-/* Specialized for ISO 8859-1 */
+//=============================================================================
+#if 0
+// build mapper and prop table
```
beyondykk9, line 203: "#if 0" is intensive for the future?

src/base/intl_support.c
```diff
@@ -66,9 +66,8 @@
               && ((unsigned char) ch <= (unsigned char) 0xfe) )
 #endif
 
-#define CHAR_BYTE_TO_LOWER(c) ((c) + ('a' - 'A'))
-
-#define CHAR_BYTE_TO_UPPER(c) ((c) - ('a' - 'A'))
+//#define CHAR_BYTE_TO_LOWER(c) ((c) + ('a' - 'A'))
```
beyondykk9: remove the codes if it's not necessary.

src/broker/broker_util.c
```diff
@@ -102,7 +102,7 @@ ut_access_log (int as_index, struct timeval *start_time, char error_flag, int er
     clt_appl = (char *) "-";
   for (p = clt_appl; *p; p++)
     {
-      if (*p == ' ' || *p == '\t' || *p == '\n' || *p == '\r')
+      if (char_isspace2 (*p))
```
beyondykk9, line 105: why not char_isspace ?

hyunikn (approved):

hgryoo (approved):

kisoo-han (approved):

src/base/intl_support.c
ctshim: OK, Thank you.

src/broker/broker_util.c
ctshim: I wrote a brief description in chartype.h.

```
/* In some codes, only four items (' ', '\t', '\n', '\r') were checked.
 * In order to maintain its original form, it was given a different name. */
#define char_isspace2  char_isspace
```

src/base/chartype.c
ctshim: This content has nothing to do with actual practice.
However, this is code to create contents such as char_upper_mapper, char_properties, etc.
This is for reference when there is a need to change the structure of the content in the future.

I will reinforce the comment.

beyondykk9 (approved):

PR #4933, opened 2024-02-02
[CBRD-25212] remove warning messages from compilation

beyondykk9:
http://jira.cubrid.org/browse/CBRD-25212

We need remove the warning messages from compilation (CUBRID engine build). By doing this, we can prevent potential problems in advance.

The target source code is as follows.

src/base/
src/compat/
src/connection/
src/communication/
src/heaplayers/
src/thread/
src/object/
src/session/
src/jsp/
src/method/
src/parser/
src/query/
src/xasl/

### DIFF
--- a/src/base/ddl_log.c
+++ b/src/base/ddl_log.c
@@ -622,7 +622,7 @@ logddl_make_copy_filename (T_APP_NAME app_name, const char *file_full_path, char
   const char *name_tmp = NULL;
   int retval = 0;
 
-  if (file_full_path == NULL || copy_filename == NULL || buf_size < 0)
+  if (file_full_path == NULL || copy_filename == NULL)
     {
       return -1;
     }
@@ -658,7 +658,7 @@ logddl_make_copy_dir (T_APP_NAME app_name, char *copy_filename, char *copy_fullp
   const char *env_root = NULL;
   int retval = 0;
 
-  if (copy_filename == NULL || copy_fullpath == NULL || buf_size < 0)
+  if (copy_filename == NULL || copy_fullpath == NULL)
     {
       return -1;
     }
@@ -883,7 +883,7 @@ logddl_write_tran_str (const char *fmt, ...)
 	  len = DDL_LOG_BUFFER_SIZE;
 	}
 
-      if (len < 0 || fwrite (msg, sizeof (char), len, fp) != len)
+      if (len < 0 || fwrite (msg, sizeof (char), len, fp) != (size_t) len)
 	{
 	  goto write_error;
 	}
--- a/src/base/ini_parser.c
+++ b/src/base/ini_parser.c
@@ -488,7 +488,7 @@ static INI_LINE_STATUS
 ini_parse_line (char *input_line, char *section, char *key, char *value)
 {
   INI_LINE_STATUS status;
-  char line[INI_BUFSIZ + 1];
+  char line[INI_BUFSIZ + 4];
   int len;
 
   strcpy (line, ini_str_trim (input_line));
@@ -515,7 +515,7 @@ ini_parse_line (char *input_line, char *section, char *key, char *value)
       leading_char = section[0];
       if (leading_char == '@' || leading_char == '%')
 	{
-	  sprintf (section, "%c%s", leading_char, ini_str_trim (section + 1));
+	  snprintf (section, INI_BUFSIZ + 2, "%c%s", leading_char, ini_str_trim (section + 1));
 	}
 
       if (leading_char != '@')
--- a/src/base/object_representation.h
+++ b/src/base/object_representation.h
@@ -2684,6 +2684,8 @@ or_get_oid (OR_BUF * buf, OID * oid)
 {
   ASSERT_ALIGN (buf->ptr, INT_ALIGNMENT);
 
+  OID_SET_NULL (oid);
+
   if ((buf->ptr + OR_OID_SIZE) > buf->endptr)
     {
       return or_underflow (buf);
--- a/src/executables/compactdb_common.c
+++ b/src/executables/compactdb_common.c
@@ -255,7 +255,7 @@ get_class_mops_from_file (const char *input_filename, MOP ** class_list, int *nu
 	  goto end;
 	}
 
-      strncpy (class_names[i], buffer, len);
+      strncpy (class_names[i], buffer, len + 1);
       class_names[i][len] = 0;
     }
 
--- a/src/executables/csql.c
+++ b/src/executables/csql.c
@@ -162,7 +162,7 @@ char csql_Scratch_text[SCRATCH_TEXT_LEN];
 int csql_Error_code = NO_ERROR;
 
 static char csql_Prompt[100];
-static char csql_Prompt_offline[100];
+static char csql_Prompt_offline[101];
 static char csql_Name[100];
 
 /*
--- a/src/method/method_struct_query.cpp
+++ b/src/method/method_struct_query.cpp
@@ -493,7 +493,7 @@ namespace cubmethod
     fprintf (stdout, "tuple_count: %d\n", tuple_count);
     fprintf (stdout, "ins_oid (%d, %d, %d)\n", ins_oid.pageid, ins_oid.slotid, ins_oid.volid);
     fprintf (stdout, "include_oid: %d\n", include_oid);
-    fprintf (stdout, "query_id: %lld\n", query_id);
+    fprintf (stdout, "query_id: %lu\n", query_id);
   }
 
   void
--- a/src/method/method_struct_query.cpp
+++ b/src/method/method_struct_query.cpp
@@ -493,7 +493,7 @@ namespace cubmethod
     fprintf (stdout, "tuple_count: %d\n", tuple_count);
     fprintf (stdout, "ins_oid (%d, %d, %d)\n", ins_oid.pageid, ins_oid.slotid, ins_oid.volid);
     fprintf (stdout, "include_oid: %d\n", include_oid);
-    fprintf (stdout, "query_id: %lu\n", query_id);
+    fprintf (stdout, "query_id: %ud\n", (unsigned int)query_id);
   }
 
   void
--- a/src/object/object_domain.c
+++ b/src/object/object_domain.c
@@ -11518,7 +11518,7 @@ tp_value_auto_cast_with_precision_check (const DB_VALUE * src, DB_VALUE * dest, 
       /* if the numeric's precision is 19 or more, then it can get the bigint enough */
       if (desired_domain->type->id == DB_TYPE_NUMERIC && desired_domain->precision < 19)
 	{
-	  INT64 bigint;
+	  INT64 bigint = 0;
 
 	  assert (desired_domain->precision >= 0);
 
--- a/src/object/object_primitive.c
+++ b/src/object/object_primitive.c
@@ -3568,7 +3568,7 @@ static int
 mr_data_readval_utime (OR_BUF * buf, DB_VALUE * value, TP_DOMAIN * domain, int size, bool copy, char *copy_buf,
 		       int copy_buf_len)
 {
-  DB_UTIME utm;
+  DB_UTIME utm = 0;
   int rc = NO_ERROR;
 
   if (value == NULL)
@@ -3591,7 +3591,7 @@ static int
 mr_data_readval_timestampltz (OR_BUF * buf, DB_VALUE * value, TP_DOMAIN * domain, int size, bool copy, char *copy_buf,
 			      int copy_buf_len)
 {
-  DB_UTIME utm;
+  DB_UTIME utm = 0;
   int rc = NO_ERROR;
 
   if (value == NULL)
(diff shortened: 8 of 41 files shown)


src/base/ddl_log.c
```diff
@@ -622,7 +622,7 @@ logddl_make_copy_filename (T_APP_NAME app_name, const char *file_full_path, char
   const char *name_tmp = NULL;
   int retval = 0;
 
-  if (file_full_path == NULL || copy_filename == NULL || buf_size < 0)
+  if (file_full_path == NULL || copy_filename == NULL)
```
ctshim, line 625: I think it would be better to remove the buf_size argument and specify PATH_MAX directly.

src/base/ddl_log.c
```diff
@@ -658,7 +658,7 @@ logddl_make_copy_dir (T_APP_NAME app_name, char *copy_filename, char *copy_fullp
   const char *env_root = NULL;
   int retval = 0;
 
-  if (copy_filename == NULL || copy_fullpath == NULL || buf_size < 0)
+  if (copy_filename == NULL || copy_fullpath == NULL)
```
ctshim, line 661: I think it would be better to remove the buf_size argument and specify PATH_MAX directly.

src/parser/view_transform.c
```diff
@@ -3514,6 +3514,8 @@ pt_check_pushable_subquery_select_list (PARSER_CONTEXT * parser, PT_NODE * query
 		    cinfo.method_found = true;
 		    break;
 		  }
+		parser_walk_leaves (parser, list, pt_check_pushable, &cinfo, NULL, NULL);
+		break;
```
ctshim: This fix doesn't seem to be related to the warning message.

src/query/execute_schema.c
```diff
@@ -10631,13 +10631,12 @@ do_change_att_schema_only (PARSER_CONTEXT * parser, DB_CTMPL * ctemplate, PT_NOD
 	case PT_TYPE_SMALLINT:
 	  break;
 
-	case PT_TYPE_NUMERIC:
-	  if (attribute->data_type->info.data_type.dec_precision == 0)
+	default:
+	  if (attribute->type_enum == PT_TYPE_NUMERIC && attribute->data_type->info.data_type.dec_precision == 0)
```
ctshim:
```suggestion
       case PT_TYPE_NUMERIC:
	  if (attribute->data_type->info.data_type.dec_precision == 0)
	    {
	       break;
	    } 
	  [[fallthrough]];
	    
	default:	  
```

src/parser/type_checking.c
```diff
@@ -7809,12 +7809,15 @@ pt_fold_constants_pre (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int *
...
+    case PT_EXPR:
+      if (node->node_type == PT_FUNCTION)
 	{
-	  // we want to test full execution of sub-tree; don't fold it!
-	  *continue_walk = PT_LIST_WALK;
+	  if (node->info.function.function_type == F_BENCHMARK)
+	    {
+	      // we want to test full execution of sub-tree; don't fold it!
+	      *continue_walk = PT_LIST_WALK;
+	    }
```
ctshim: if (node->info.function.function_type == F_BENCHMARK)
    {
	  // we want to test full execution of sub-tree; don't fold it!
	  *continue_walk = PT_LIST_WALK;
     }
    [[fallthrough]];

    case PT_EXPR:

src/query/execute_schema.c
```diff
@@ -11207,6 +11206,10 @@ build_attr_change_map (PARSER_CONTEXT * parser, DB_CTMPL * ctemplate, PT_NODE *
 	  break;
 	case PT_CONSTRAIN_NULL:
 	  attr_chg_properties->p[P_NOT_NULL] = ATT_CHG_PROPERTY_LOST;
+	  constr_att_list = cnstr->info.constraint.un.not_null.attr;
+	  chg_prop_idx = P_NOT_NULL;
+	  save_pt_costraint = true;
+	  break;
```
ctshim: How about handling it like below?

constr_att_list = NULL;
[[fallthrough]];

src/query/execute_schema.c
beyondykk9: I modified as your opinion.

src/parser/view_transform.c
beyondykk9: I resolve it by [[fallthrough]].

src/query/execute_schema.c
beyondykk9: I modified it as your opinion.

src/base/ddl_log.c
airnet73: I think the buf_size parameter should be kept. The reason is that the size of the string can change.

src/base/ddl_log.c
airnet73: I think the buf_size parameter should be kept. The reason is that the size of the string can change.

shparkcubrid (approved):

src/executables/csql.c
```diff
@@ -162,7 +162,7 @@ char csql_Scratch_text[SCRATCH_TEXT_LEN];
 int csql_Error_code = NO_ERROR;
 
 static char csql_Prompt[100];
-static char csql_Prompt_offline[100];
+static char csql_Prompt_offline[101];
```
airnet73, line 165: I think "snprintf (csql_Prompt_offline, sizeof (csql_Prompt_offline) - 1, "!%s", csql_Prompt);" is better than "csql_Prompt_offline[101]". (line 2772)

airnet73 (approved):

src/base/object_representation.h
```diff
@@ -2684,6 +2684,8 @@ or_get_oid (OR_BUF * buf, OID * oid)
 {
   ASSERT_ALIGN (buf->ptr, INT_ALIGNMENT);
 
+  OID_SET_NULL (oid);
```
hyunikn, line 2687: This should be moved to the then part of the if statement below because setting the oid fields is done in else part.

src/object/object_domain.c
```diff
@@ -11518,7 +11518,7 @@ tp_value_auto_cast_with_precision_check (const DB_VALUE * src, DB_VALUE * dest,
       /* if the numeric's precision is 19 or more, then it can get the bigint enough */
       if (desired_domain->type->id == DB_TYPE_NUMERIC && desired_domain->precision < 19)
 	{
-	  INT64 bigint;
+	  INT64 bigint = 0;
```
hyunikn, line 11521: This should be moved to the default branch of the switch statement below because this is overwritten in other branches.

src/object/object_primitive.c
```diff
@@ -3568,7 +3568,7 @@ static int
 mr_data_readval_utime (OR_BUF * buf, DB_VALUE * value, TP_DOMAIN * domain, int size, bool copy, char *copy_buf,
 		       int copy_buf_len)
 {
-  DB_UTIME utm;
+  DB_UTIME utm = 0;
```
hyunikn, line 3571: This definition should be moved to the else branch of the if statement below because utm is used only in that branch.

src/object/object_primitive.c
```diff
@@ -3591,7 +3591,7 @@ static int
 mr_data_readval_timestampltz (OR_BUF * buf, DB_VALUE * value, TP_DOMAIN * domain, int size, bool copy, char *copy_buf,
 			      int copy_buf_len)
 {
-  DB_UTIME utm;
+  DB_UTIME utm = 0;
```
hyunikn, line 3594: This definition should be moved to the else branch of the if statement below because utm is used only in that branch.

src/parser/parser_support.c
```diff
@@ -11607,7 +11607,7 @@ pt_convert_dblink_dml_query (PARSER_CONTEXT * parser, PT_NODE * node,
   PT_NODE *list = NULL;		/* for insert select list */
   PT_NODE *spec, *into_spec = NULL, *upd_spec = NULL, *server;
 
-  PARSER_VARCHAR *comment, *dml;
+  PARSER_VARCHAR *comment = NULL, *dml;
```
hyunikn, line 11610: This NULL assignment should be moved to the default branch of the switch statement at line 11726 because in other branches the assignment is overwritten.

src/query/execute_statement.c
```diff
@@ -15348,12 +15348,12 @@ do_supplemental_statement (PARSER_CONTEXT * parser, PT_NODE * statement, RESERVE
 
 	    if (cls_info[i]->objtype == CDC_TABLE)
 	      {
-		strncpy (drop_stmt, drop_prefix, strlen (drop_prefix));
+		strncpy (drop_stmt, drop_prefix, strlen (drop_prefix) + 1);
```
hyunikn, line 15351: strncpy(dst, src, strlen(src) + 1) is equivalent to strcpy(dst, src) and can be simplified to the latter.

src/base/ini_parser.c
```diff
@@ -515,7 +515,7 @@ ini_parse_line (char *input_line, char *section, char *key, char *value)
       leading_char = section[0];
       if (leading_char == '@' || leading_char == '%')
 	{
-	  sprintf (section, "%c%s", leading_char, ini_str_trim (section + 1));
+	  snprintf (section, INI_BUFSIZ + 2, "%c%s", leading_char, ini_str_trim (section + 1));
```
ctshim, line 518: The size of the section is (INI_BUFSIZ + 1).

src/executables/compactdb_common.c
```diff
@@ -255,7 +255,7 @@ get_class_mops_from_file (const char *input_filename, MOP ** class_list, int *nu
 	  goto end;
 	}
 
-      strncpy (class_names[i], buffer, len);
+      strncpy (class_names[i], buffer, len + 1);
```
ctshim, line 258: Is there a reason to copy with a size of "len+1"?
If you decide to modify it anyway, I recommend doing it as follows.

memcpy (class_names[i], buffer, len);

src/method/method_struct_query.cpp
```diff
@@ -493,7 +493,7 @@ namespace cubmethod
     fprintf (stdout, "tuple_count: %d\n", tuple_count);
     fprintf (stdout, "ins_oid (%d, %d, %d)\n", ins_oid.pageid, ins_oid.slotid, ins_oid.volid);
     fprintf (stdout, "include_oid: %d\n", include_oid);
-    fprintf (stdout, "query_id: %lld\n", query_id);
+    fprintf (stdout, "query_id: %ud\n", (unsigned int)query_id);
```
ctshim, line 496: query_id is of type UINTPTR.
This is defined differently depending on define.
Therefore, if you do it like below, you will be able to see the warning again even if the type changes in the future.

fprintf (stdout, "query_id: %ud\n", query_id);

src/query/execute_statement.c
ctshim: I habitually use strncpy inside this function, but it would be better to replace it with strcpy.
If we go a little further,
A more desirable solution is possible by changing the order of the process of finding the length after copying.


```
strncpy (drop_stmt, drop_prefix, strlen (drop_prefix) + 1);
drop_copied_length = strlen (drop_prefix);

```
Replace the form above as below (there is a small gain in performance)
```
drop_copied_length = strlen (drop_prefix);
memcpy (drop_stmt, drop_prefix, len + 1);
```

src/storage/btree_load.c
```diff
@@ -1149,8 +1148,9 @@ xbtree_load_index (THREAD_ENTRY * thread_p, BTID * btid, const char *bt_name, TP
       pr_clear_value (&load_args->current_key);
 
       BTID_SET_NULL (btid);
-      if (xbtree_add_index (thread_p, btid, key_type, &class_oids[0], attr_ids[0], unique_pk, sort_args->n_oids,
-			    sort_args->n_nulls, load_args->n_keys, btid_int.deduplicate_key_idx) == NULL)
+      if (xbtree_add_index (thread_p, btid, key_type, &class_oids[0], attr_ids[0], unique_pk, (INT64) sort_args->n_oids,
```
ctshim, line 1151: Maybe I should change the member variable of sort_args from int to INT64?

src/base/object_representation.h
beyondykk9: Ok, I aggree.

src/executables/compactdb_common.c
hyunikn: +1 is necessary because class_names[i] must end with a terminating zero. Isn't it?
Note that len is strlen(buffer).

src/method/method_struct_query.cpp
hyunikn: It must be %u rather than %ud.
Note that the following code prints '123d'
```
#include <stdio.h>

int
main() {
    printf("%ud\n", 123);
}
```

src/executables/compactdb_common.c
ctshim: @hyunikn
In the code, '\0' is filled with the substitution statement immediately below.
`class_names[i][len] = 0;`

hornetmj (approved):

beyondykk9:
타 이슈에서 진행이 완료되어서, 클로즈 합니다.

PR #4934, opened 2024-02-05
[CUBRID-24687] query result is strange, when there is hidden columns in the select list.

mhoh3963:
http://jira.cubrid.org/browse/CBRD-24687

**Description**
The query result is strange, when there is hidden columns in ther select list. This is the case when sp is used in the order by clause, as in the following query. (ex : SELECT col1, col2) FROM tbl1 ORDER BY intTest(col1))

**Resolution**
- fix to use query_type instead of col_cnt at db_query_column_count().

### DIFF
--- a/src/broker/cas_execute.c
+++ b/src/broker/cas_execute.c
@@ -4914,7 +4914,7 @@ cur_tuple (T_QUERY_RESULT * q_result, int max_col_size, char sensitive_flag, DB_
   char *null_type_column = q_result->null_type_column;
   int err_code;
 
-  ncols = q_result->num_column;
+  ncols = db_query_column_count (result);
   for (i = 0; i < ncols; i++)
     {
       if (sensitive_flag == TRUE && col_update_info[i].updatable == TRUE)
--- a/src/broker/cas_execute.c
+++ b/src/broker/cas_execute.c
@@ -4914,7 +4914,7 @@ cur_tuple (T_QUERY_RESULT * q_result, int max_col_size, char sensitive_flag, DB_
   char *null_type_column = q_result->null_type_column;
   int err_code;
 
-  ncols = db_query_column_count (result);
+  ncols = q_result->num_column;
   for (i = 0; i < ncols; i++)
     {
       if (sensitive_flag == TRUE && col_update_info[i].updatable == TRUE)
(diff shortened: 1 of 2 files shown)


kisoo-han (approved):

src/compat/db_query.c
```diff
@@ -3164,7 +3167,12 @@ db_query_column_count (DB_QUERY_RESULT * result)
       return -1;
     }
 
-  return (DB_OID_INCLUDED (result)) ? (result->col_cnt - 1) : result->col_cnt;
+  for (t = result->query_type; t != NULL; t = db_query_format_next (t), num_attrs++)
```
beyondykk9:
```suggestion
  for (t = result->query_type; t != NULL; t = db_query_format_next (t))
     {
       num_attrs++;
     }
```

src/compat/db_query.c
```diff
@@ -3164,7 +3167,12 @@ db_query_column_count (DB_QUERY_RESULT * result)
...
+    {
+      ;
+    }
+
+  return (num_attrs);
```
beyondykk9:
```suggestion
  return num_attrs;
```

src/compat/db_query.c
```diff
@@ -3150,6 +3150,9 @@ db_query_tuple_count (DB_QUERY_RESULT * result)
 int
 db_query_column_count (DB_QUERY_RESULT * result)
 {
+  DB_QUERY_TYPE *t;
+  int num_attrs = 0;
```
beyondykk9: the variable "num_cols" would be better than "num_attrs"

src/broker/cas_execute.c
```diff
@@ -4914,7 +4914,7 @@ cur_tuple (T_QUERY_RESULT * q_result, int max_col_size, char sensitive_flag, DB_
   char *null_type_column = q_result->null_type_column;
   int err_code;
 
-  ncols = db_query_column_count (result);
+  ncols = q_result->num_column;
```
beyondykk9: it would be better add a comment to explain the logic

src/compat/db_query.c
mhoh3963: Thank you, I accept.

src/compat/db_query.c
mhoh3963: Thank you, I accept.

src/compat/db_query.c
mhoh3963: Thank you, I accept.

src/broker/cas_execute.c
mhoh3963: This code was occurred a problem when executing the call and evaluation statement. So I will revert it.

beyondykk9 (approved):

PR #4938, opened 2024-02-06
[CBRD-25208] Set java.io.tmpdir according to the CUBRID_TMP in Java SP server (#4917)

hgryoo:
http://jira.cubrid.org/browse/CBRD-25208

Set java.io.tmpdir according to the CUBRID_TMP. If CUBRID_TMP is set, it is utilized as java.io.tmpdir, otherwise $CUBRID/tmp is used.

backport of #4917


beyondykk9 (approved):

beyondykk9:
please, check windows build fail.

PR #4940, opened 2024-02-06
[CBRD-25203] Implement 'memory_monitor::add/sub_stat()' with API

Rudeus:
http://jira.cubrid.org/browse/CBRD-25203

This PR implements `add/sub_stat()` with API and adds member variables into `memory_monitor` class for the functions
 - add member variables at memory_monitor class
 - implement `add_stat()` and `sub_stat()` functions

### DIFF
--- a/src/base/memory_monitor_sr.cpp
+++ b/src/base/memory_monitor_sr.cpp
@@ -131,16 +131,13 @@ namespace cubmem
 
   void memory_monitor::sub_stat (char *ptr)
   {
+#if defined(WINDOWS)
     size_t allocated_size = 0;
-
-    if (ptr == NULL)
-      {
-	return;
-      }
-
-#if !defined(WINDOWS)
-    allocated_size = malloc_usable_size ((void *)ptr);
+#else
+    size_t allocated_size = malloc_usable_size ((void *)ptr);
 #endif // !WINDOWS
+
+    assert (ptr != NULL);
 
     if (allocated_size >= MMON_ALLOC_META_SIZE)
       {
--- a/src/base/memory_monitor_sr.cpp
+++ b/src/base/memory_monitor_sr.cpp
@@ -47,7 +47,7 @@ namespace cubmem
   size_t memory_monitor::get_allocated_size (const char *ptr)
   {
 #if defined(WINDOWS)
-    size_t allocated_size = _msize ((void *)ptr);
+    size_t allocated_size = 0;
 #else
     size_t allocated_size = malloc_usable_size ((void *)ptr);
 #endif // WINDOWS
@@ -73,7 +73,7 @@ namespace cubmem
   {
     std::string filecopy (file);
 #if defined(WINDOWS)
-    std::string target ("\\src\\");
+    std::string target (""); // not supported
 #else
     std::string target ("/src/");
 #endif // WINDOWS
@@ -83,7 +83,7 @@ namespace cubmem
 #if !defined (NDEBUG)
     size_t lpos = filecopy.find (target);
     assert (pos == lpos);
-#endif // NDEBUG
+#endif // !NDEBUG
 
     if (pos != std::string::npos)
       {
@@ -117,9 +117,8 @@ namespace cubmem
       {
 	metainfo.tag_id = m_tag_map.size ();
 	// tag_id is start with 0
-	std::pair <std::string, int> tag_map_entry (tag_name, metainfo.tag_id);
-	m_tag_map.insert (tag_map_entry);
-	m_stat_map.insert (std::make_pair (metainfo.tag_id, metainfo.allocated_size));
+	m_tag_map.emplace (tag_name, metainfo.tag_id);
+	m_stat_map.emplace (metainfo.tag_id, metainfo.allocated_size);
       }
     tag_map_lock.unlock ();
 
@@ -132,18 +131,16 @@ namespace cubmem
 
   void memory_monitor::sub_stat (char *ptr)
   {
-    size_t allocated_size;
+    size_t allocated_size = 0;
 
     if (ptr == NULL)
       {
 	return;
       }
 
-#if defined(WINDOWS)
-    allocated_size = _msize ((void *)ptr);
-#else
+#if !defined(WINDOWS)
     allocated_size = malloc_usable_size ((void *)ptr);
-#endif // WINDOWS
+#endif // !WINDOWS
 
     if (allocated_size >= MMON_ALLOC_META_SIZE)
       {
--- a/src/base/memory_monitor_sr.cpp
+++ b/src/base/memory_monitor_sr.cpp
@@ -92,6 +92,77 @@ namespace cubmem
 
     return filecopy + ':' + std::to_string (line);
   }
+
+  void memory_monitor::add_stat (char *ptr, const size_t size, const char *file, const int line)
+  {
+    std::string tag_name;
+    MMON_METAINFO metainfo;
+
+    // size should not be 0 because of MMON_ALLOC_META_SIZE
+    assert (size > 0);
+
+    metainfo.allocated_size = (uint64_t) size;
+    m_total_mem_usage += metainfo.allocated_size;
+
+    tag_name = make_tag_name (file, line);
+
+    std::unique_lock <std::mutex> tag_map_lock (m_tag_map_mutex);
+    auto search = m_tag_map.find (tag_name);
+    if (search != m_tag_map.end ())
+      {
+	metainfo.tag_id = search->second;
+	m_stat_map[metainfo.tag_id] += metainfo.allocated_size;
+      }
+    else
+      {
+	metainfo.tag_id = m_tag_map.size ();
+	// tag_id is start with 0
+	std::pair <std::string, int> tag_map_entry (tag_name, metainfo.tag_id);
+	m_tag_map.insert (tag_map_entry);
+	m_stat_map.insert (std::make_pair (metainfo.tag_id, metainfo.allocated_size));
+      }
+    tag_map_lock.unlock ();
+
+    // put meta info into the allocated chunk
+    char *meta_ptr = ptr + metainfo.allocated_size - MMON_ALLOC_META_SIZE;
+    metainfo.magic_number = m_magic_number;
+    memcpy (meta_ptr, &metainfo, MMON_ALLOC_META_SIZE);
+    m_meta_alloc_count++;
+  }
+
+  void memory_monitor::sub_stat (char *ptr)
+  {
+    size_t allocated_size;
+
+    if (ptr == NULL)
+      {
+	return;
+      }
+
+#if defined(WINDOWS)
+    allocated_size = _msize ((void *)ptr);
+#else
+    allocated_size = malloc_usable_size ((void *)ptr);
+#endif // WINDOWS
+
+    if (allocated_size >= MMON_ALLOC_META_SIZE)
+      {
+	char *meta_ptr = ptr + allocated_size - MMON_ALLOC_META_SIZE;
+	const MMON_METAINFO *metainfo = (const MMON_METAINFO *) meta_ptr;
+
+	if (metainfo->magic_number == m_magic_number)
+	  {
+	    assert ((metainfo->tag_id >= 0 && metainfo->tag_id <= m_stat_map.size()));
+	    assert (m_stat_map[metainfo->tag_id] >= metainfo->allocated_size);
+
+	    m_total_mem_usage -= metainfo->allocated_size;
+	    m_stat_map[metainfo->tag_id] -= metainfo->allocated_size;
+
+	    memset (meta_ptr, 0, MMON_ALLOC_META_SIZE);
+	    m_meta_alloc_count--;
+	  }
+      }
+  }
 }
 
 using namespace cubmem;
@@ -110,4 +181,20 @@ size_t mmon_get_allocated_size (char *ptr)
   // unreachable
   assert (false);
   return 0;
+}
+
+void mmon_add_stat (char *ptr, const size_t size, const char *file, const int line)
+{
+  if (mmon_is_memory_monitor_enabled ())
+    {
+      mmon_Gl->add_stat (ptr, size, file, line);
+    }
+}
+
+void mmon_sub_stat (char *ptr)
+{
+  if (mmon_is_memory_monitor_enabled ())
+    {
+      mmon_Gl->sub_stat (ptr);
+    }
 }
--- a/src/base/memory_monitor_sr.cpp
+++ b/src/base/memory_monitor_sr.cpp
@@ -41,37 +41,10 @@ namespace cubmem
 
   memory_monitor::memory_monitor (const char *server_name)
     : m_server_name {server_name},
-      m_magic_number {*reinterpret_cast <const int *> ("MMON")}
-  {
-    m_total_mem_usage = 0;
-    m_meta_alloc_count = 0;
-  }
-
-  size_t memory_monitor::get_allocated_size (const char *ptr)
-  {
-#if defined(WINDOWS)
-    size_t allocated_size = 0;
-    assert (false);
-#else
-    size_t allocated_size = malloc_usable_size ((void *)ptr);
-#endif // !WINDOWS
-
-    if (allocated_size <= MMON_ALLOC_META_SIZE)
-      {
-	return allocated_size;
-      }
-
-    const char *meta_ptr = ptr + allocated_size - MMON_ALLOC_META_SIZE;
-
-    const MMON_METAINFO *metainfo = (const MMON_METAINFO *) meta_ptr;
-
-    if (metainfo->magic_number == m_magic_number)
-      {
-	allocated_size = (size_t) metainfo->allocated_size - MMON_ALLOC_META_SIZE;
-      }
-
-    return allocated_size;
-  }
+      m_magic_number {*reinterpret_cast <const int *> ("MMON")},
+      m_total_mem_usage {0},
+      m_meta_alloc_count {0}
+  {}
 
   std::string memory_monitor::make_tag_name (const char *file, const int line)
   {
@@ -98,6 +71,37 @@ namespace cubmem
     return filecopy + ':' + std::to_string (line);
   }
 
+  char *memory_monitor::get_meta_ptr (char *ptr, size_t size)
+  {
+    return ptr + size - MMON_ALLOC_META_SIZE;
+  }
+
+  size_t memory_monitor::get_allocated_size (char *ptr)
+  {
+#if defined(WINDOWS)
+    size_t allocated_size = 0;
+    assert (false);
+#else
+    size_t allocated_size = malloc_usable_size ((void *)ptr);
+#endif // !WINDOWS
+
+    if (allocated_size <= MMON_ALLOC_META_SIZE)
+      {
+	return allocated_size;
+      }
+
+    char *meta_ptr = get_meta_ptr (ptr, allocated_size);
+
+    const MMON_METAINFO *metainfo = (const MMON_METAINFO *) meta_ptr;
+
+    if (metainfo->magic_number == m_magic_number)
+      {
+	allocated_size = (size_t) metainfo->allocated_size - MMON_ALLOC_META_SIZE;
+      }
+
+    return allocated_size;
+  }
+
   void memory_monitor::add_stat (char *ptr, const size_t size, const char *file, const int line)
   {
     std::string tag_name;
@@ -121,14 +125,14 @@ namespace cubmem
     else
       {
 	metainfo.tag_id = m_tag_map.size ();
-	// tag_id is start with 0
+	// tag_id starts with 0
 	m_tag_map.emplace (tag_name, metainfo.tag_id);
 	m_stat_map.emplace (metainfo.tag_id, metainfo.allocated_size);
       }
     tag_map_lock.unlock ();
 
     // put meta info into the allocated chunk
-    char *meta_ptr = ptr + metainfo.allocated_size - MMON_ALLOC_META_SIZE;
+    char *meta_ptr = get_meta_ptr (ptr, metainfo.allocated_size);
     metainfo.magic_number = m_magic_number;
     memcpy (meta_ptr, &metainfo, MMON_ALLOC_META_SIZE);
     m_meta_alloc_count++;
@@ -147,7 +151,7 @@ namespace cubmem
 
     if (allocated_size >= MMON_ALLOC_META_SIZE)
       {
-	char *meta_ptr = ptr + allocated_size - MMON_ALLOC_META_SIZE;
+	char *meta_ptr = get_meta_ptr (ptr, allocated_size);
 	const MMON_METAINFO *metainfo = (const MMON_METAINFO *) meta_ptr;
 
 	if (metainfo->magic_number == m_magic_number)
--- a/src/base/memory_monitor_sr.cpp
+++ b/src/base/memory_monitor_sr.cpp
@@ -42,7 +42,10 @@ namespace cubmem
   memory_monitor::memory_monitor (const char *server_name)
     : m_server_name {server_name},
       m_magic_number {*reinterpret_cast <const int *> ("MMON")}
-  {}
+  {
+    m_total_mem_usage = 0;
+    m_meta_alloc_count = 0;
+  }
 
   size_t memory_monitor::get_allocated_size (const char *ptr)
   {
@@ -74,6 +77,7 @@ namespace cubmem
     std::string filecopy (file);
 #if defined(WINDOWS)
     std::string target (""); // not supported
+    assert (false);
 #else
     std::string target ("/src/");
 #endif // WINDOWS
@@ -154,6 +158,7 @@ namespace cubmem
 
 	    memset (meta_ptr, 0, MMON_ALLOC_META_SIZE);
 	    m_meta_alloc_count--;
+	    assert (m_meta_alloc_count >= 0);
 	  }
       }
   }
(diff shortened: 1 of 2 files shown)


src/base/memory_monitor_sr.cpp
```diff
@@ -92,6 +92,77 @@ namespace cubmem
 
     return filecopy + ':' + std::to_string (line);
   }
+
+  void memory_monitor::add_stat (char *ptr, const size_t size, const char *file, const int line)
```
joohok, line 105:
```suggestion
  void
  memory_monitor::add_stat (char *ptr, const size_t size, const char *file, const int line)
```

Hmm, the indentation for most C++ codes in Cubrid is configured like this.
It can make too much changes..

src/base/memory_monitor_sr.cpp
```diff
@@ -92,6 +92,77 @@ namespace cubmem
...
+	metainfo.tag_id = search->second;
+	m_stat_map[metainfo.tag_id] += metainfo.allocated_size;
+      }
+    else
+      {
+	metainfo.tag_id = m_tag_map.size ();
+	// tag_id is start with 0
+	std::pair <std::string, int> tag_map_entry (tag_name, metainfo.tag_id);
+	m_tag_map.insert (tag_map_entry);
+	m_stat_map.insert (std::make_pair (metainfo.tag_id, metainfo.allocated_size));
```
```suggestion
	m_tag_map.emplace (tag_name, metainfo.tag_id);
	m_stat_map.emplace (metainfo.tag_id, metainfo.allocated_size);
```

src/base/memory_monitor_sr.cpp
```diff
@@ -92,6 +92,77 @@ namespace cubmem
...
+	    assert ((metainfo->tag_id >= 0 && metainfo->tag_id <= m_stat_map.size()));
+	    assert (m_stat_map[metainfo->tag_id] >= metainfo->allocated_size);
+
+	    m_total_mem_usage -= metainfo->allocated_size;
+	    m_stat_map[metainfo->tag_id] -= metainfo->allocated_size;
+
+	    memset (meta_ptr, 0, MMON_ALLOC_META_SIZE);
+	    m_meta_alloc_count--;
+	  }
+      }
```
joohok, line 170: what about `else`? (allocated_size < MMON_ALLOC_META_SIZE)

src/base/memory_monitor_sr.cpp
```diff
@@ -92,6 +92,77 @@ namespace cubmem
...
+    if (ptr == NULL)
+      {
+	return;
+      }
+
+#if defined(WINDOWS)
+    allocated_size = _msize ((void *)ptr);
+#else
+    allocated_size = malloc_usable_size ((void *)ptr);
+#endif // WINDOWS
```
```suggestion
#if !defined(WINDOWS)
    allocated_size = malloc_usable_size ((void *)ptr);
#endif // !WINDOWS
```

how about defining the allocate_size only if it does not run on WINDOWS ?

src/base/memory_monitor_sr.cpp
```diff
@@ -92,6 +92,77 @@ namespace cubmem
...
+    metainfo.magic_number = m_magic_number;
+    memcpy (meta_ptr, &metainfo, MMON_ALLOC_META_SIZE);
+    m_meta_alloc_count++;
+  }
+
+  void memory_monitor::sub_stat (char *ptr)
+  {
+    size_t allocated_size;
+
+    if (ptr == NULL)
```
joohok: It seems that `ptr` should not be retrieved as a null value, and the same goes for `add_stat`. 
I think filtering should either be done at the caller or at the caller's caller level. What are your thoughts on this?

src/base/memory_monitor_sr.cpp
Rudeus: I'll change it with future refactoring issue :) Thanks.

src/base/memory_monitor_sr.cpp

src/base/memory_monitor_sr.cpp
Rudeus: `else` case means that the memory allocation is occurred out-of-scope of `memory_monitor`. So there is nothing to do something.

src/base/memory_monitor_sr.cpp
Rudeus: I change the code for Windows build with doing nothing.

src/base/memory_monitor_sr.cpp
```diff
@@ -73,7 +73,7 @@ namespace cubmem
   {
     std::string filecopy (file);
 #if defined(WINDOWS)
-    std::string target ("\\src\\");
+    std::string target (""); // not supported
```
wrlawodms, line 53:
```suggestion
    std::string target (""); // not supported
    assert_release (false);
```
To prevent an undefined behavior. Or, just assert(false); if it's not allowed asset_release() for this feature(MMON).

src/base/memory_monitor_sr.cpp
Rudeus: You're right. I fix it.

src/base/memory_monitor_sr.cpp
```diff
@@ -92,6 +92,74 @@ namespace cubmem
...
+    std::unique_lock <std::mutex> tag_map_lock (m_tag_map_mutex);
+    auto search = m_tag_map.find (tag_name);
+    if (search != m_tag_map.end ())
+      {
+	metainfo.tag_id = search->second;
+	m_stat_map[metainfo.tag_id] += metainfo.allocated_size;
+      }
+    else
+      {
+	metainfo.tag_id = m_tag_map.size ();
```
wrlawodms, line 127: Q. Is the size of `m_tag_map` ever-increasing?
If it's not the case, there can be a duplicated id for different allocations.

src/base/memory_monitor_sr.cpp
```diff
@@ -92,6 +92,74 @@ namespace cubmem
...
+	// tag_id is start with 0
+	m_tag_map.emplace (tag_name, metainfo.tag_id);
+	m_stat_map.emplace (metainfo.tag_id, metainfo.allocated_size);
```
wrlawodms, line 130: What if there is already an entry which has the same `tag_id`? The same allocation on the same location can occur multiple time.

src/base/memory_monitor_sr.cpp
```diff
@@ -92,6 +92,74 @@ namespace cubmem
...
+	char *meta_ptr = ptr + allocated_size - MMON_ALLOC_META_SIZE;
+	const MMON_METAINFO *metainfo = (const MMON_METAINFO *) meta_ptr;
+
+	if (metainfo->magic_number == m_magic_number)
+	  {
+	    assert ((metainfo->tag_id >= 0 && metainfo->tag_id <= m_stat_map.size()));
+	    assert (m_stat_map[metainfo->tag_id] >= metainfo->allocated_size);
+
+	    m_total_mem_usage -= metainfo->allocated_size;
+	    m_stat_map[metainfo->tag_id] -= metainfo->allocated_size;
```
wrlawodms, line 164: Isn't the entry removed even if it gets 0?

src/base/memory_monitor_sr.cpp
```diff
@@ -92,6 +92,74 @@ namespace cubmem
...
+
+	    memset (meta_ptr, 0, MMON_ALLOC_META_SIZE);
+	    m_meta_alloc_count--;
```
wrlawodms, line 167:
```suggestion
	    m_meta_alloc_count--;
	    assert (m_meta_alloc_count >= 0);
```

src/base/memory_monitor_sr.hpp
```diff
@@ -50,16 +52,25 @@ namespace cubmem
...
+      void add_stat (char *ptr, const size_t size, const char *file, const int line);
+      void sub_stat (char *ptr);
 
     private:
       static std::string make_tag_name (const char *file, const int line);
 
     private:
       std::string m_server_name;
+      mutable std::mutex m_tag_map_mutex;
+      std::unordered_map <std::string, int> m_tag_map;              // tag name <-> tag id
```
```suggestion
      std::unordered_map <std::string, int> m_tag_map;              // tag name -> tag id
```

src/base/memory_monitor_sr.hpp
```diff
@@ -50,16 +52,25 @@ namespace cubmem
...
+      std::unordered_map <int, std::atomic <uint64_t>> m_stat_map;  // tag id <-> memory usage
+      std::atomic <uint64_t> m_total_mem_usage;
+      int m_meta_alloc_count;                                       // for checking occupancy of memory used by metainfo space
```
wrlawodms: You should initialize the added member variables in ctor.

src/base/memory_monitor_sr.cpp
Rudeus: I add assert(false). Maybe this function will not be called because I am not going to create `mmon_Gl` if cub_server is started in Windows despite `enable_memory_monitoring=yes`.

src/base/memory_monitor_sr.cpp
Rudeus: The entry of `m_tag_map` will not be deleted, so the size of `m_tag_map` will be ever-increasing.

src/base/memory_monitor_sr.cpp
Rudeus: The entry of m_tag_map will not be deleted, so the size of m_tag_map will be ever-increasing.

src/base/memory_monitor_sr.cpp
Rudeus: No. The entry will no be removed even if it gets 0.

src/base/memory_monitor_sr.cpp
Rudeus: Done.

src/base/memory_monitor_sr.hpp
Rudeus: I change it.

src/base/memory_monitor_sr.hpp
Rudeus: Oh I missed it. Done.

src/base/memory_monitor_sr.cpp
```diff
@@ -42,12 +42,15 @@ namespace cubmem
...
+  {
+    m_total_mem_usage = 0;
+    m_meta_alloc_count = 0;
+  }
 
   size_t memory_monitor::get_allocated_size (const char *ptr)
   {
 #if defined(WINDOWS)
-    size_t allocated_size = _msize ((void *)ptr);
+    size_t allocated_size = 0;
```
joohok: how about adding an assertion here too?

src/base/memory_monitor_sr.cpp
```diff
@@ -73,7 +76,8 @@ namespace cubmem
   {
     std::string filecopy (file);
 #if defined(WINDOWS)
-    std::string target ("\\src\\");
+    std::string target (""); // not supported
+    assert (false);
 #else
     std::string target ("/src/");
 #endif // WINDOWS
```
```suggestion
#endif // !WINDOWS
```

src/base/memory_monitor_sr.cpp
wrlawodms: This was about the `m_stat_map` here, not `m_tag_map`. 
Anyway, I guess now I understand this and why you don't remove an entry `m_stat_map` though it's possible. It's the way to keep the implementation simple.

joohok (approved):

src/base/memory_monitor_sr.cpp

src/base/memory_monitor_sr.cpp

src/base/memory_monitor_sr.cpp
```diff
@@ -92,6 +96,72 @@ namespace cubmem
...
+    auto search = m_tag_map.find (tag_name);
+    if (search != m_tag_map.end ())
+      {
+	metainfo.tag_id = search->second;
+	m_stat_map[metainfo.tag_id] += metainfo.allocated_size;
+      }
+    else
+      {
+	metainfo.tag_id = m_tag_map.size ();
+	// tag_id is start with 0
```
```suggestion
	// tag_id starts with 0
```

src/base/memory_monitor_sr.cpp
wrlawodms: Ok, please add a comment about that on the declaration of the container.

src/base/memory_monitor_sr.cpp
```diff
@@ -42,12 +42,15 @@ namespace cubmem
   memory_monitor::memory_monitor (const char *server_name)
     : m_server_name {server_name},
       m_magic_number {*reinterpret_cast <const int *> ("MMON")}
-  {}
+  {
+    m_total_mem_usage = 0;
+    m_meta_alloc_count = 0;
```
wrlawodms: Nit.

It's better to use one way to initialize variables if possible. And, it seems better to me to use the initializer list.
```
  memory_monitor::memory_monitor (const char *server_name)
     : m_server_name {server_name},
       m_magic_number {*reinterpret_cast <const int *> ("MMON")},
       m_total_mem_usage { 0 }, 
      m_meta_alloc_count { 0 }
```

src/base/memory_monitor_sr.hpp
```diff
@@ -50,16 +52,25 @@ namespace cubmem
...
     private:
       static std::string make_tag_name (const char *file, const int line);
 
     private:
       std::string m_server_name;
+      mutable std::mutex m_tag_map_mutex;
+      std::unordered_map <std::string, int> m_tag_map;              // key: tag name, value: tag id
+      std::unordered_map <int, std::atomic <uint64_t>> m_stat_map;  // key: tag id, value: memory usage
+      std::atomic <uint64_t> m_total_mem_usage;
+      int m_meta_alloc_count;                                       // for checking occupancy of memory used by metainfo space
```
```suggestion
     std::atomic<int> m_meta_alloc_count;                                       // for checking occupancy of memory used by metainfo space
```

src/base/memory_monitor_sr.cpp
```diff
@@ -92,6 +96,72 @@ namespace cubmem
...
+      {
+	metainfo.tag_id = m_tag_map.size ();
+	// tag_id is start with 0
+	m_tag_map.emplace (tag_name, metainfo.tag_id);
+	m_stat_map.emplace (metainfo.tag_id, metainfo.allocated_size);
+      }
+    tag_map_lock.unlock ();
+
+    // put meta info into the allocated chunk
+    char *meta_ptr = ptr + metainfo.allocated_size - MMON_ALLOC_META_SIZE;
```
wrlawodms: I'd extract this as a static private function like `get_meta_ptr (ptr, user_size)` because the same calculation is used across functions and must be the same. It's safer.

src/base/memory_monitor_sr.hpp
wrlawodms: How about adding `assert (m_meta_alloc_count == 0); assert (m_total_mem_usage == 0);` in its dtor? It helps us catching potential missing points.

wrlawodms:
In one of following PRs, some elaboration for the `memory_monitor` should be added on the class definition: what's it for and how. Especially some unintuitive part like "why do we need the magic number although it's turned on/off for entire lifetime" or "The header should be included at the last location and why". I'm leaving this comment here because I can't find an issue about this.

src/base/memory_monitor_sr.cpp
Rudeus: I agree. Done.

src/base/memory_monitor_sr.cpp

src/base/memory_monitor_sr.cpp
Rudeus: Done.

src/base/memory_monitor_sr.hpp
Rudeus: In my knowledge, ++ and -- guarantee atomic even if it's not std::atomic<int> but I change it because your suggestion makes it clearer.

src/base/memory_monitor_sr.cpp
Rudeus: I agree. I add it.

src/base/memory_monitor_sr.hpp
Rudeus: I believe your suggestion is valid, but as shown in the test results using the previous PoC version, there is a memory leak issue with the current CUBRID. Therefore, adding such an assertion always would lead SIGABORT, which seems unfavorable.

Rudeus:
> In one of following PRs, some elaboration for the `memory_monitor` should be added on the class definition: what's it for and how. Especially some unintuitive part like "why do we need the magic number although it's turned on/off for entire lifetime" or "The header should be included at the last location and why". I'm leaving this comment here because I can't find an issue about this.

I have a question because I want to clearly understand your comment. Is it OK if the description added with comment in code? Or JIRA issue? You mean the description must be into just "PR description"?

src/base/memory_monitor_sr.hpp
wrlawodms: oh, I see. Is that gonna be resolved?

wrlawodms:
> I have a question because I want to clearly understand your comment. Is it OK if the description added with comment in code? Or JIRA issue? You mean the description must be into just "PR description"?

No need to describe that in the PR or Jira issue. It's about the code comment :)

wrlawodms (approved):

src/base/memory_monitor_sr.hpp
Rudeus: Memory leak issue is another thing that has no relation with memory monitoring. So I think It's gonna be resolved at sm_maintenance.

Rudeus:
> > I have a question because I want to clearly understand your comment. Is it OK if the description added with comment in code? Or JIRA issue? You mean the description must be into just "PR description"?
> 
> No need to describe that in the PR or Jira issue. It's about the code comment :)

I will add some comments for understanding in this PR :) Thank you for your feedback.
And "The header should be included at the last location and why" case is for wrapping functions of C++ objects (new/delete). So I will mention about that in future issue!

src/base/memory_monitor_sr.cpp
```diff
@@ -111,3 +187,19 @@ size_t mmon_get_allocated_size (char *ptr)
   assert (false);
   return 0;
 }
+
+void mmon_add_stat (char *ptr, const size_t size, const char *file, const int line)
+{
+  if (mmon_is_memory_monitor_enabled ())
```
hornetmj: It is expected to check this in the caller like cub_alloc(). So, it doesn't need to check here.

src/base/memory_monitor_sr.cpp
```diff
@@ -41,23 +41,56 @@ namespace cubmem
...
+#endif // !WINDOWS
+
+    // Find the last occurrence of "src" in the path
+    size_t pos = filecopy.rfind (target);
+#if !defined (NDEBUG)
+    size_t lpos = filecopy.find (target);
+    assert (pos == lpos);
+#endif // !NDEBUG
+
+    if (pos != std::string::npos)
```
hornetmj, line 66: Check pos == std::string::npos in DEBUG mode because it is also un-expected case.

```
if !defined (NDEBUG)
else
{
    assert (0);
}
#endif
```

src/base/memory_monitor_sr.cpp
```diff
@@ -69,28 +102,71 @@ namespace cubmem
...
+
+    // size should not be 0 because of MMON_ALLOC_META_SIZE
+    assert (size > 0);
+
+    metainfo.allocated_size = (uint64_t) size;
+    m_total_mem_usage += metainfo.allocated_size;
+
+    tag_name = make_tag_name (file, line);
+
+    std::unique_lock <std::mutex> tag_map_lock (m_tag_map_mutex);
```
hornetmj: NIT:

m_tag_map_mutex protects both `m_tag_map` and `m_stat_map` but, the name of it means that it only cosider `m_tag_map`.

src/base/memory_monitor_sr.cpp
```diff
@@ -41,23 +41,56 @@ namespace cubmem
...
+
+    if (pos != std::string::npos)
+      {
+	filecopy = filecopy.substr (pos + target.length ());
+      }
+
+    return filecopy + ':' + std::to_string (line);
+  }
+
+  char *memory_monitor::get_meta_ptr (char *ptr, size_t size)
```
hornetmj: NIT:

I think that pos(ition) is better than ptr.
```suggestion
  char *memory_monitor::get_meta_pos (char *ptr, size_t size)
```

src/base/memory_monitor_sr.cpp
```diff
@@ -69,28 +102,71 @@ namespace cubmem
...
+	// tag_id starts with 0
+	m_tag_map.emplace (tag_name, metainfo.tag_id);
+	m_stat_map.emplace (metainfo.tag_id, metainfo.allocated_size);
+      }
+    tag_map_lock.unlock ();
+
+    // put meta info into the allocated chunk
+    char *meta_ptr = get_meta_ptr (ptr, metainfo.allocated_size);
+    metainfo.magic_number = m_magic_number;
+    memcpy (meta_ptr, &metainfo, MMON_ALLOC_META_SIZE);
```
hornetmj: NIT:

`MMON_METAINFO_SIZE` is more consistent than `MMON_ALLOC_META_SIZE`.

src/base/memory_monitor_sr.cpp
hornetmj: And `meta` and `metainfo` is mixed. It would be good to use one of them for consistency.

src/base/memory_monitor_sr.cpp
```diff
@@ -69,28 +102,71 @@ namespace cubmem
...
       {
-	filecopy = filecopy.substr (pos + target.length ());
-      }
+	char *meta_ptr = get_meta_ptr (ptr, allocated_size);
+	const MMON_METAINFO *metainfo = (const MMON_METAINFO *) meta_ptr;
 
-    return filecopy + ':' + std::to_string (line);
+	if (metainfo->magic_number == m_magic_number)
+	  {
+	    assert ((metainfo->tag_id >= 0 && metainfo->tag_id <= m_stat_map.size()));
```
```suggestion
	    assert ((metainfo->tag_id >= 0 && metainfo->tag_id < m_stat_map.size()));
```

src/base/memory_monitor_sr.cpp
```diff
@@ -69,28 +102,71 @@ namespace cubmem
...
-    // Find the last occurrence of "src" in the path
-    size_t pos = filecopy.rfind (target);
-#if !defined (NDEBUG)
-    size_t lpos = filecopy.find (target);
-    assert (pos == lpos);
-#endif // NDEBUG
+    assert (ptr != NULL);
 
-    if (pos != std::string::npos)
+    if (allocated_size >= MMON_ALLOC_META_SIZE)
```
```suggestion
    if (allocated_size > MMON_ALLOC_META_SIZE)
```

hornetmj: NIT:

To remove nested if statement.

```
if (allocated_size > MMON_ALLOC_META_SIZE)
{
    return;
}

...
```

src/base/memory_monitor_sr.cpp
```diff
@@ -69,28 +102,71 @@ namespace cubmem
...
+	  {
+	    assert ((metainfo->tag_id >= 0 && metainfo->tag_id <= m_stat_map.size()));
+	    assert (m_stat_map[metainfo->tag_id] >= metainfo->allocated_size);
+
+	    m_total_mem_usage -= metainfo->allocated_size;
+	    m_stat_map[metainfo->tag_id] -= metainfo->allocated_size;
+
+	    memset (meta_ptr, 0, MMON_ALLOC_META_SIZE);
+	    m_meta_alloc_count--;
+	    assert (m_meta_alloc_count >= 0);
```
hornetmj, line 168:
```suggestion
	    assert (m_total_mem_usage >= 0);
	    assert (m_meta_alloc_count >= 0);
```

src/base/memory_monitor_sr.cpp
Rudeus: You're right. I fix it.

src/base/memory_monitor_sr.cpp
Rudeus: `make_tag_name()` can get two types of parameter `file` because pre-defined macro `__FILE__` can give two types of return.
1. /home/rudii/cubrid/src/.../filename.cpp
2. filename.cpp

So case 1 touches `if (pos != std::string::npos)` to make substring `.../filename.cpp` and case 2 passes the condition. In case 2, the function returns just `filename.cpp:line`.

src/base/memory_monitor_sr.cpp
Rudeus: I was already aware of the question you raised, but I hadn't thought of a suitable name to express integrated information. What do you think about renaming it to `m_map_mutex`?

src/base/memory_monitor_sr.cpp
Rudeus: You're right. Done.

src/base/memory_monitor_sr.cpp

src/base/memory_monitor_sr.cpp

src/base/memory_monitor_sr.cpp
Rudeus: Even when malloc is requested to allocate memory of size 0, it returns a valid pointer instead of nullptr. Additionally, it is possible to free this allocated memory. Therefore, the situation where `allocated_size == MMON_METAINFO_SIZE` should also be considered.

src/base/memory_monitor_sr.cpp
Rudeus: I think `assert(m_stat_map[metainfo->tag_id] >= metainfo->allocated_size)` can guarantee `m_total_mem_usage >= 0` after sub_stat but we can check another abnormal case if we add this assertion. Done.

src/base/memory_monitor_sr.cpp
hornetmj: Good to me.

hornetmj (approved):

PR #4949, opened 2024-02-15
[CBRD-25204] Implement aggregate function of memory monitor

Rudeus:
http://jira.cubrid.org/browse/CBRD-25204

 - Add structure MMON_SERVER_INFO at the new file `memory_monitor_common.hpp` for using both saving the information of `memory_monitor` class and receiving the information at the client-side
 - Implement aggregation method `aggregate_server_info()` in memory_monitor class
 - Implement aggregation API `mmon_aggregate_server_info()`
 - Implement debug API `finalize_dump()` for checking all memory allocations freed

### DIFF
--- a/src/base/memory_monitor_common.hpp
+++ b/src/base/memory_monitor_common.hpp
@@ -1,0 +1,43 @@
+/*
+ *
+ * Copyright 2016 CUBRID Corporation
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+
+/*
+ * memory_monitor_common.hpp - Declaration of structures which is using
+ *                             both server and client
+ */
+
+#ifndef _MEMORY_MONITOR_COMMON_HPP_
+#define _MEMORY_MONITOR_COMMON_HPP_
+
+#include <stdint.h>
+#include <string>
+#include <vector>
+#include <algorithm>
+
+#define MMON_MAX_SERVER_NAME_LENGTH 255
+
+typedef struct mmon_server_info MMON_SERVER_INFO;
+struct mmon_server_info
+{
+  char server_name[MMON_MAX_SERVER_NAME_LENGTH];
+  uint64_t total_mem_usage;
+  uint64_t mmon_metainfo_mem_usage;
+  int num_stat;
+  std::vector<std::pair<std::string, uint64_t>> stat_info;
+};
+#endif // _MEMORY_MONITOR_COMMON_HPP_
--- a/src/base/memory_monitor_common.hpp
+++ b/src/base/memory_monitor_common.hpp
@@ -29,15 +29,15 @@
 #include <vector>
 #include <algorithm>
 
-#define MMON_MAX_SERVER_NAME_LENGTH 255
+#define MMON_MAX_SERVER_NAME_LENGTH 32
 
 typedef struct mmon_server_info MMON_SERVER_INFO;
 struct mmon_server_info
 {
   char server_name[MMON_MAX_SERVER_NAME_LENGTH];
   uint64_t total_mem_usage;
-  uint64_t mmon_metainfo_mem_usage;
+  uint64_t total_metainfo_mem_usage;
   int num_stat;
-  std::vector<std::pair<std::string, uint64_t>> stat_info;
+  std::vector<std::pair<std::string, uint64_t>> stat_info; // <stat name, memory usage>
 };
 #endif // _MEMORY_MONITOR_COMMON_HPP_
--- a/src/base/memory_monitor_sr.cpp
+++ b/src/base/memory_monitor_sr.cpp
@@ -31,7 +31,7 @@ typedef struct mmon_metainfo MMON_METAINFO;
 struct mmon_metainfo
 {
   uint64_t allocated_size;
-  int tag_id;
+  int stat_id;
   int magic_number;
 };
 
@@ -46,7 +46,7 @@ namespace cubmem
       m_meta_alloc_count {0}
   {}
 
-  std::string memory_monitor::make_tag_name (const char *file, const int line)
+  std::string memory_monitor::make_stat_name (const char *file, const int line)
   {
     std::string filecopy (file);
 #if defined(WINDOWS)
@@ -104,7 +104,7 @@ namespace cubmem
 
   void memory_monitor::add_stat (char *ptr, const size_t size, const char *file, const int line)
   {
-    std::string tag_name;
+    std::string stat_name;
     MMON_METAINFO metainfo;
 
     // size should not be 0 because of MMON_METAINFO_SIZE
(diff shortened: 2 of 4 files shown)


src/base/memory_monitor_common.hpp
```diff
@@ -0,0 +1,43 @@
...
+#define MMON_MAX_SERVER_NAME_LENGTH 255
+
+typedef struct mmon_server_info MMON_SERVER_INFO;
+struct mmon_server_info
+{
+  char server_name[MMON_MAX_SERVER_NAME_LENGTH];
+  uint64_t total_mem_usage;
+  uint64_t mmon_metainfo_mem_usage;
+  int num_stat;
+  std::vector<std::pair<std::string, uint64_t>> stat_info;
```
wrlawodms: Nit. a simple comment for the pair; not sure about the name :)
```suggestion
  std::vector<std::pair<std::string, uint64_t>> stat_info;  // <stat_name, mem_usage>
```

src/base/memory_monitor_sr.cpp
```diff
@@ -169,6 +169,62 @@ namespace cubmem
...
+  {
+    strncpy (server_info.server_name, m_server_name.c_str (), m_server_name.size () + 1);
+    server_info.total_mem_usage = m_total_mem_usage.load ();
+    server_info.mmon_metainfo_mem_usage = m_meta_alloc_count * MMON_METAINFO_SIZE;
+    server_info.num_stat = m_tag_map.size ();
+
+    for (auto it = m_tag_map.begin (); it != m_tag_map.end (); ++it)
+      {
+	server_info.stat_info.emplace_back (it->first, m_stat_map[it->second].load ());
+      }
```
wrlawodms: Nit. modern c++ way

c++17: more readable and simpler: 
```suggestion
    for (const auto& [stat_name, mem_usage] : m_tag_map)
      {
	server_info.stat_info.emplace_back (stat_name, mem_usage);
      }
```

src/base/memory_monitor_sr.cpp
```diff
@@ -169,6 +169,62 @@ namespace cubmem
...
+
+    const auto &comp = [] (const auto &stat_pair1, const auto &stat_pair2)
+    {
+      return stat_pair1.second > stat_pair2.second;
```
wrlawodms, line 192: Nit.
```suggestion
      return stat_pair1.second > stat_pair2.second; // Sort in descending order by memory usage
```

src/base/memory_monitor_common.hpp
```diff
@@ -0,0 +1,43 @@
...
+#include <string>
+#include <vector>
+#include <algorithm>
+
+#define MMON_MAX_SERVER_NAME_LENGTH 255
+
+typedef struct mmon_server_info MMON_SERVER_INFO;
+struct mmon_server_info
+{
+  char server_name[MMON_MAX_SERVER_NAME_LENGTH];
```
wrlawodms, line 37: I think an existing macro for the db name length should be used instead.
I'm looking around and `DB_MAX_IDENTIFIER_LENGTH` seems right for this.
Another alternative is `PATH_MAX` or `NAME_MAX` that is used more generally.
@joohok Do you know more proper macro?

src/base/memory_monitor_sr.cpp
```diff
@@ -169,6 +169,62 @@ namespace cubmem
...
+    };
+    std::sort (server_info.stat_info.begin (), server_info.stat_info.end (), comp);
+  }
+
+  void memory_monitor::finalize_dump ()
+  {
+    double mem_usage_ratio = 0.0;
+    FILE *outfile_fp = fopen ("finalize_dump.txt", "w+");
```
wrlawodms, line 200: I don't know how it's used. If it's used only for debugging, I think it's OK. If it's used by clients through a utility, the path should be given by them, not hardcoded one.

src/base/memory_monitor_sr.cpp
```diff
@@ -169,6 +169,62 @@ namespace cubmem
...
+    fprintf (outfile_fp, "====================cubrid memmon====================\n");
+    fprintf (outfile_fp, "Server Name: %s\n", server_info.server_name);
+    fprintf (outfile_fp, "Total Memory Usage(KB): %lu\n\n", MMON_CONVERT_TO_KB_SIZE (server_info.total_mem_usage));
+    fprintf (outfile_fp, "Total Metainfo Memory Usage(KB): %lu\n\n",
+	     MMON_CONVERT_TO_KB_SIZE (server_info.mmon_metainfo_mem_usage));
+    fprintf (outfile_fp, "-----------------------------------------------------\n");
+
+    fprintf (outfile_fp, "\t%-100s | %17s(%s)\n", "File Name", "Memory Usage", "Ratio");
+
+    for (const auto &s_info : server_info.stat_info)
```
wrlawodms: Nit.
I'm not sure it works for vector of pairs, but it does probably:
```suggestion
    for (const auto& [stat_name, mem_usage] : server_info.stat_info)
```

src/base/memory_monitor_sr.cpp
```diff
@@ -169,6 +169,62 @@ namespace cubmem
...
+      }
+
+    const auto &comp = [] (const auto &stat_pair1, const auto &stat_pair2)
+    {
+      return stat_pair1.second > stat_pair2.second;
+    };
+    std::sort (server_info.stat_info.begin (), server_info.stat_info.end (), comp);
+  }
+
+  void memory_monitor::finalize_dump ()
```
wrlawodms, line 197: How about just `dump`? It can be used not only when finalizing.
```suggestion
  void memory_monitor::dump ()
```

wrlawodms (approved):
All of them are NIT :)

src/base/memory_monitor_common.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+#ifndef _MEMORY_MONITOR_COMMON_HPP_
+#define _MEMORY_MONITOR_COMMON_HPP_
+
+#include <stdint.h>
+#include <string>
+#include <vector>
+#include <algorithm>
+
+#define MMON_MAX_SERVER_NAME_LENGTH 255
```
hornetmj: NIT:

The length of dbname cannot over 17 character.

src/base/memory_monitor_common.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+typedef struct mmon_server_info MMON_SERVER_INFO;
+struct mmon_server_info
+{
+  char server_name[MMON_MAX_SERVER_NAME_LENGTH];
+  uint64_t total_mem_usage;
+  uint64_t mmon_metainfo_mem_usage;
```
hornetmj:
```suggestion
  uint64_t total_metainfo_mem_usage;
```

src/base/memory_monitor_common.hpp
hornetmj: NIT:

How about matching keywords added for memory monitoring, such as `tag` and `stat`? For example, if you change tag and tag ID to stat_name and stat ID, the tag keyword is no longer needed, and you can provide uniformity throughout the code. I have made many comments about the keywords and variable/function names you used but, the purpose is to provide consistency throughout the code, and through this, to improve code understanding in the future. Please think about this and consider it in all future PRs.

src/base/memory_monitor_sr.cpp
hornetmj, line 197: Comment on if this function is used only debugging purpose and remove this after completing implementation if unnecessary.

If it is, then we don't need to review this.

src/base/memory_monitor_sr.cpp
hornetmj: This is comment for the @wrlawodms's suggestion.

mem_usage have to get from the m_stat_map not m_tag_map.

src/base/memory_monitor_sr.cpp
```diff
@@ -169,6 +169,62 @@ namespace cubmem
 	  }
       }
   }
+
+  void memory_monitor::aggregate_server_info (MMON_SERVER_INFO &server_info)
+  {
+    strncpy (server_info.server_name, m_server_name.c_str (), m_server_name.size () + 1);
+    server_info.total_mem_usage = m_total_mem_usage.load ();
+    server_info.mmon_metainfo_mem_usage = m_meta_alloc_count * MMON_METAINFO_SIZE;
+    server_info.num_stat = m_tag_map.size ();
```
hornetmj: m_map_mutex protection is needed to block race condition with mmon_add_stat with new tag.

src/base/memory_monitor_common.hpp
Rudeus: The file exists at the lowest level of the build dependency hierarchy, making it burdensome to include other files to use the defined value.

src/base/memory_monitor_common.hpp
Rudeus: Done.

src/base/memory_monitor_sr.cpp

src/base/memory_monitor_sr.cpp
Rudeus: Done.

src/base/memory_monitor_sr.cpp
Rudeus: It's only for debugging at finalization of the `memory_monitor` class.

src/base/memory_monitor_sr.cpp
Rudeus: It's only for servers with debug mode at the finalization of the `memory_monitor` class. I think it's sufficient because in other cases, we can check it through `memmon` utility.

src/base/memory_monitor_common.hpp
Rudeus: Done.

src/base/memory_monitor_common.hpp
Rudeus: Oh I don't know about that.  I will shrink the size to 32 (for align).

src/base/memory_monitor_common.hpp
Rudeus: I'll think about it. Thanks.

src/base/memory_monitor_sr.cpp
Rudeus: I will leave it because we can check about memory leak when we build with debug mode.

src/base/memory_monitor_sr.cpp
Rudeus: I agree. I change `mutex` to `shared_mutex` for using it to `shared/unique_lock`.

src/base/memory_monitor_sr.cpp
```diff
@@ -113,23 +113,23 @@ namespace cubmem
...
     m_total_mem_usage += metainfo.allocated_size;
 
-    tag_name = make_tag_name (file, line);
+    stat_name = make_stat_name (file, line);
 
-    std::unique_lock <std::mutex> map_lock (m_map_mutex);
-    auto search = m_tag_map.find (tag_name);
-    if (search != m_tag_map.end ())
+    std::unique_lock <std::shared_mutex> stat_map_write_lock (m_stat_map_mutex);
+    auto search = m_stat_name_map.find (stat_name);
```
```suggestion
    const auto search = m_stat_name_map.find (stat_name);
```

if there is no intention to modify `search`

joohok (approved):

src/base/memory_monitor_sr.cpp

hornetmj (approved):

PR #4950, opened 2024-02-15
[CBRD-25228] Query with CTE sub-query, query-cache does not work.

beyondykk9:
http://jira.cubrid.org/browse/CBRD-25228

In the case of queries that have CTE queries as sub-queries, query-cache is not being processed.
This is because do_prepare_select is simply checking the CTE query and calling do_prepare_cte.
In other words, the existence of CTE queries, including sub-queries, must be checked, and do_prepare_cte must be called for all CTE sub-queries.

It's a kind of bug.


ctshim (approved):

hyunikn (approved):

hgryoo:
Please check my comment in JIRA.

beyondykk9:
> Please check my comment in JIRA.

I wrote the repro query incorrectly. I have uploaded the modified code to JIRA.
Please check again.

hgryoo (approved):

PR #4952, opened 2024-02-15
[CBRD-25227] Add a new system catalog for code of stored routines

hgryoo:
http://jira.cubrid.org/browse/CBRD-25227

- Add `_db_stored_procedure_code` to store PL/CSQL source and object code (.class or .jar).
- Add adding and dropping routine for `_db_stored_procedure_code`
  - NOTE: Calling the routines are commented out for future use.


src/jsp/jsp_cl.cpp
```diff
@@ -1102,6 +1157,72 @@ drop_stored_procedure (const char *name, SP_TYPE_ENUM expected_type)
...
+      goto error;
+    }
+
+  err = db_get (code_mop, SP_ATTR_IS_SYSTEM_GENERATED, &generated_val);
+  if (err != NO_ERROR)
+    {
+      goto error;
+    }
+
+  if (!DB_IS_NULL (&generated_val) && 1 == db_get_int (&generated_val))
```
ctshim, line 1208: trivial;
 I think it would be better to use the defined value rather than 1.

ctshim (approved):

src/jsp/jsp_cl.cpp
hgryoo: In system catalog, Boolean column type does not exists, so it is represented as Integer type with 1 and 0 values. I think we need a definition such as `BOOLEAN_TRUE (1)` and `BOLLEAN_FALSE (0)` in system_catalog module. IMO, it is not scope of this PR, I'd like to add the definition in another system catalog refactoring issue.
Thank you for helpful refactoring tip.

src/jsp/jsp_cl.cpp
```diff
@@ -1063,6 +1099,25 @@ drop_stored_procedure (const char *name, SP_TYPE_ENUM expected_type)
...
 
+  // delete _db_stored_procedure_code
+  err = db_get (sp_mop, SP_ATTR_TARGET, &target_val);
+  if (err != NO_ERROR)
+    {
+      goto error;
+    }
+
+  target = db_get_string (&target_val);
+  class_name = get_class_name (target);
```
beyondykk9: class_name, it's not used currently. Is it left to use in the future (commentd as TODO) ?

src/jsp/jsp_cl.cpp
```diff
@@ -1063,6 +1099,25 @@ drop_stored_procedure (const char *name, SP_TYPE_ENUM expected_type)
...
+
+  // TODO: CBRD-24552
+  /*
```
hgryoo:
```suggestion
  // TODO: CBRD-24552
  /*
    class_name = get_class_name (target);

```

src/jsp/jsp_cl.cpp
hgryoo: I've moved the line inside of the TODO comment block.

beyondykk9 (approved):

hyunikn (approved):

PR #4957, opened 2024-02-19
[CBRD-25147] Add SQL syntax to add users as members of the group

jongmin-won:
http://jira.cubrid.org/browse/CBRD-25147

To add a member to the current group, you need to execute 'CREATE USER...GROUPS/MEMBERS' and 'CALL add_member()'

The two methods have the following inconvenience:
1) An already existing user must execute the 'DROP USER … → CREATE USER ... GROUPS' syntax
2) Performing the CALL add_member() method allows adding a group to an already existing user. However, the results of the method calls in the HA environment are out of sync and must be done separately by the master and slave

Therefore, add SQL syntax to add users as members of the group can alleviate the inconvenience


src/query/execute_schema.c
```diff
@@ -2248,6 +2250,71 @@ do_alter_user (const PARSER_CONTEXT * parser, const PT_NODE * statement)
 	}
     }
 
+  /* member */
+  node = statement->info.alter_user.members;
+  member_name = (node && IS_NAME (node)) ? GET_NAME (node) : NULL;
+  if (member_name != NULL)
+    {
+      if (alter_user_code == PT_ADD_MEMBERS && (ws_is_same_object (user, Au_user) || au_is_dba_group_member (Au_user)))
```
youngjinj: How about checking `ws_is_same_object` and `au_is_dba_group_member` before `alter_user_code`?

src/parser/csql_grammar.y
```diff
@@ -3582,11 +3582,11 @@ alter_stmt
...
-	  opt_password
-	  opt_comment_spec
+	| ALTER				/* 1 */
+	  USER				/* 2 */
+	  identifier	        	/* 3 */
+	  opt_password  	        /* 4 */
+	  opt_comment_spec              /* 5 */
 		{{ DBG_TRACE_GRAMMAR(alter_stmt, | ALTER USER identifier opt_password opt_comment_spec);
 
 			PT_NODE *node = parser_new_node (this_parser, PT_ALTER_USER);
```
youngjinj, line 3592: How about initializing `info.alter_user.code` to prevent unintended behavior?

src/parser/parse_tree_cl.c
```diff
@@ -6425,6 +6426,25 @@ pt_print_alter_user (PARSER_CONTEXT * parser, PT_NODE * p)
...
+      if (p->info.alter_user.members != NULL)
+	{
+	  r1 = pt_print_bytes (parser, p->info.alter_user.members);
+	  b = pt_append_nulstring (parser, b, " add members ");
+	  b = pt_append_varchar (parser, b, r1);
+	}
+    }
+  else if (p->info.alter_user.code == PT_DROP_MEMBERS)
+    {
+      if (p->info.alter_user.members != NULL)
```
youngjinj: How about checking `info.alter_user.members` before `info.alter_user.code`?

src/query/execute_schema.c
jongmin-won: We've fixed it, please confirm

src/parser/parse_tree_cl.c
jongmin-won: We've fixed it, please confirm

src/query/execute_schema.c
youngjinj: I thought of separating the if statements checking `ws_is_same_object` and `au_is_dba_group_member`.

```
if (!ws_is_same_object (user, Au_user) && !au_is_dba_group_member (Au_user))
  {
    /* give up */
  }

switch (alter_user_code)
  {
    case PT_ADD_MEMBERS:
      do
        {
          ...
        }
      while (member_name != NULL)
      break;
    case PT_DROP_MEMBERS:
      do
        {
          ...
        }
      while (member_name != NULL)
      break;
    default:
      /* error */
      break;
  }
```

jongmin-won: The code you reviewed looks more readable, so we've fixed it.

src/parser/parse_tree_cl.c
```diff
@@ -6425,6 +6426,22 @@ pt_print_alter_user (PARSER_CONTEXT * parser, PT_NODE * p)
...
+	  r1 = pt_print_bytes (parser, p->info.alter_user.members);
+	  b = pt_append_nulstring (parser, b, " add members ");
+	  b = pt_append_varchar (parser, b, r1);
+	}
+      else if (p->info.alter_user.code == PT_DROP_MEMBERS)
+	{
+	  r1 = pt_print_bytes (parser, p->info.alter_user.members);
+	  b = pt_append_nulstring (parser, b, " drop members ");
+	  b = pt_append_varchar (parser, b, r1);
+	}
```
beyondykk9, line 6442: missing "else" clause ?
How about assert(false) for debugging ?

src/query/execute_schema.c
```diff
@@ -2248,6 +2250,78 @@ do_alter_user (const PARSER_CONTEXT * parser, const PT_NODE * statement)
...
+		{
+		  goto end;
+		}
+
+	      node = node->next;
+	      member_name = (node && IS_NAME (node)) ? GET_NAME (node) : NULL;
+	    }
+	  while (member_name != NULL);
+	  break;
+	default:
```
beyondykk9, line 2317: How about assert (false) for debuging.

src/parser/parse_tree_cl.c
jongmin-won: I didn't think of adding the "else" clause because I thought that the flag value in the code would only get PT_ADD_MEMBERS and PT_DROP_MEMBERS, and I didn't consider the other cases.

Following the feedback from the code review, I added assert(false) for debugging purposes.

Thank you for checking.

src/query/execute_schema.c
jongmin-won: I didn't consider adding assert(false) for debugging.

Following the feedback from the code review, I added assert(false) for debugging purposes.

Thank you for checking.

beyondykk9 (approved):

youngjinj (approved):

shparkcubrid (approved):

PR #4958, opened 2024-02-19
[CBRD-25064] loaddb terminates with exit code 0 even though an error occurred

mhoh3963:
http://jira.cubrid.org/browse/CBRD-25064

- modify the status value is changed to 3 in the ldr_sa_load(), when an error occurs

### DIFF
--- a/src/loaddb/load_sa_loader.cpp
+++ b/src/loaddb/load_sa_loader.cpp
@@ -6444,7 +6444,6 @@ ldr_sa_load (load_args *args, int *status, bool *interrupted)
 						 LOADDB_MSG_LAST_COMMITTED_LINE), lastcommit);
 		}
 	      *interrupted = true;
-	      *status = 3;
 	    }
 	  else
 	    {
@@ -6521,6 +6520,11 @@ ldr_sa_load (load_args *args, int *status, bool *interrupted)
 		}
 	    }
 	}
+    }
+
+  if (errors > 0)
+    {
+      *status = 3;
     }
 
   ldr_final ();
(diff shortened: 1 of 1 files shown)


src/loaddb/load_sa_loader.cpp
```diff
@@ -6444,7 +6444,6 @@ ldr_sa_load (load_args *args, int *status, bool *interrupted)
 						 LOADDB_MSG_LAST_COMMITTED_LINE), lastcommit);
 		}
 	      *interrupted = true;
-	      *status = 3;
```
airnet73, line 6447: The previous code terminates with exit code 3 when an interrupt occurs.
However, in the changed code, if an interrupt occurs and no error occurs, it terminates with exit code 0. The behavior is different from the previous code. What do you think?

mhoh3963: If interrupted is true or status != 0 after ldr_sa_loader() is called, it goes down, so there doesn't seem to be a problem in that case.

[refer to code]
...
 ldr_sa_load (&args, &status, &interrupted);
 if (interrupted || status != 0)
        {
          // failed
          db_end_session ();
          db_shutdown ();
          goto error_return;
        }

mhoh3963: Thank you.
As you commented, if an interrupt occurs and there are no errors, it may exit with 0.
So I will change my code according to your comments.

airnet73 (approved):

kisoo-han (approved):

PR #4961, opened 2024-02-20
[CBRD-25205] Start/end 'memory_monitor' when server booting/shutdown

Rudeus:
http://jira.cubrid.org/browse/CBRD-25205

In this PR,
 - add mmon_initialize() and mmon_finalize()
   - mmon_initialize() creates global memory_monitor class `mmon_Gl` with nothrow. If it fails to create, return `ER_OUT_OF_VIRTUAL_MEMORY`
   - mmon_finalize() deletes `mmon_Gl` if it exists.
   - call `mmon_Gl->finalize_dump()` for checking all memories are freed when we build debug mode.
 - call mmon_initialize() at boot_restart_server() and mmon_finalize() at net_server_start()

### DIFF
--- a/src/base/memory_monitor_sr.cpp
+++ b/src/base/memory_monitor_sr.cpp
@@ -246,6 +246,7 @@ int mmon_initialize (const char *server_name)
   int error = NO_ERROR;
 
   assert (mmon_Gl == nullptr);
+  assert (server_name != nullptr);
 
   if (prm_get_bool_value (PRM_ID_ENABLE_MEMORY_MONITORING))
     {
--- a/src/base/memory_monitor_sr.cpp
+++ b/src/base/memory_monitor_sr.cpp
@@ -25,6 +25,8 @@
 #include <cstring>
 #include <algorithm>
 
+#include "error_manager.h"
+#include "system_parameter.h"
 #include "memory_monitor_sr.hpp"
 
 typedef struct mmon_metainfo MMON_METAINFO;
@@ -237,6 +239,38 @@ using namespace cubmem;
 bool mmon_is_memory_monitor_enabled ()
 {
   return (mmon_Gl != nullptr);
+}
+
+int mmon_initialize (const char *server_name)
+{
+  int error = NO_ERROR;
+
+  assert (mmon_Gl == nullptr);
+  assert_release (server_name != NULL);
+
+  if (prm_get_bool_value (PRM_ID_ENABLE_MEMORY_MONITORING))
+    {
+      mmon_Gl = new (std::nothrow) memory_monitor (server_name);
+      if (mmon_Gl == nullptr)
+	{
+	  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, sizeof (memory_monitor));
+	  error = ER_OUT_OF_VIRTUAL_MEMORY;
+	  return error;
+	}
+    }
+  return error;
+}
+
+void mmon_finalize ()
+{
+  if (mmon_is_memory_monitor_enabled ())
+    {
+#if !defined (NDEBUG)
+      mmon_Gl->finalize_dump ();
+#endif
+      delete mmon_Gl;
+      mmon_Gl = nullptr;
+    }
 }
 
 size_t mmon_get_allocated_size (char *ptr)
--- a/src/base/memory_monitor_sr.cpp
+++ b/src/base/memory_monitor_sr.cpp
@@ -246,7 +246,6 @@ int mmon_initialize (const char *server_name)
   int error = NO_ERROR;
 
   assert (mmon_Gl == nullptr);
-  assert_release (server_name != NULL);
 
   if (prm_get_bool_value (PRM_ID_ENABLE_MEMORY_MONITORING))
     {
--- a/src/communication/network_sr.c
+++ b/src/communication/network_sr.c
@@ -60,6 +60,7 @@
 #include "thread_manager.hpp"
 #include "session.h"
 #include "network_request_def.hpp"
+#include "memory_monitor_sr.hpp"
 
 static void net_server_init (void);
 static int net_server_request (THREAD_ENTRY * thread_p, unsigned int rid, int request, int size, char *buffer);
@@ -1178,6 +1179,7 @@ net_server_start (const char *server_name)
 
   cubthread::finalize ();
   cubthread::internal_tasks_worker_pool::finalize ();
+  mmon_finalize ();
   er_final (ER_ALL_FINAL);
   csect_finalize_static_critical_sections ();
   (void) sync_finalize_sync_stats ();
--- a/src/communication/network_sr.c
+++ b/src/communication/network_sr.c
@@ -1132,7 +1132,12 @@ net_server_start (const char *server_name)
       status = -1;
       goto end;
     }
-
+#if defined(SERVER_MODE)
+  if (mmon_initialize (server_name) != NO_ERROR)
+    {
+      goto end;
+    }
+#endif
   net_server_init ();
   css_initialize_server_interfaces (net_server_request, net_server_conn_down);
 
--- a/src/communication/network_sr.c
+++ b/src/communication/network_sr.c
@@ -1132,12 +1132,14 @@ net_server_start (const char *server_name)
       status = -1;
       goto end;
     }
-#if defined(SERVER_MODE)
+
   if (mmon_initialize (server_name) != NO_ERROR)
     {
+      PRINT_AND_LOG_ERR_MSG ("Failed to initialize memory_monitor\n");
+      status = -1;
       goto end;
     }
-#endif
+
   net_server_init ();
   css_initialize_server_interfaces (net_server_request, net_server_conn_down);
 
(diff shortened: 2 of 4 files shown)


src/base/memory_monitor_sr.cpp
```diff
@@ -239,6 +241,38 @@ bool mmon_is_memory_monitor_enabled ()
...
+int mmon_initialize (const char *server_name)
+{
+  int error = NO_ERROR;
+
+  assert (mmon_Gl == nullptr);
+  assert_release (server_name != NULL);
+
+  if (prm_get_bool_value (PRM_ID_ENABLE_MEMORY_MONITORING))
+    {
+      mmon_Gl = new (std::nothrow) memory_monitor (server_name);
```
wrlawodms, line 253: Q. Why is the `nothrow` needed?

src/communication/network_sr.c
```diff
@@ -1178,6 +1179,7 @@ net_server_start (const char *server_name)
 
   cubthread::finalize ();
   cubthread::internal_tasks_worker_pool::finalize ();
+  mmon_finalize ();
```
wrlawodms, line 1189: The counter part of `boot_restart_server` is the `xboot_shutdown_server`. The way I see it, it should be there. If you already know the function, please tell me why this is the place for this call.

src/base/memory_monitor_sr.cpp
Rudeus: To prevent abnormal termination of the program due to `bad_alloc` when memory allocation fails in memory_monitor because of memory shortage, and to incorporate it into the error handling process of CUBRID.

src/communication/network_sr.c
Rudeus: I acknowledges that `xboot_shutdown_server()` corresponds to `boot_restart_server()`, it's impossible to precisely align the invocation of `mmon_finalize()` due to different termination orders of various modules of error handling section in `boot_restart_server()` and `xboot_shutdown_server()`, which differ from their initialization order. Although differing initialization and termination orders don't pose a problem for shutting down the modules, the discrepancy in order raises concerns regarding memory management and tracing in `memory_monitor`'s shutdown, as it cannot guarantee the release of all traced memory. Therefore, we positioned the termination of `memory_monitor` at the very end of the boot-shutdown sequence to ensure that CUBRID releases all allocated memory that `memory_monitor` traced with the basis of all modules within the tracing scope have been terminated.

src/base/memory_monitor_sr.cpp
```diff
@@ -239,6 +241,38 @@ bool mmon_is_memory_monitor_enabled ()
   return (mmon_Gl != nullptr);
 }
 
+int mmon_initialize (const char *server_name)
+{
+  int error = NO_ERROR;
+
+  assert (mmon_Gl == nullptr);
+  assert_release (server_name != NULL);
```
```suggestion
  assert_release (server_name != nullptr);
```

src/base/memory_monitor_sr.cpp
wrlawodms: I've got it. I didn't know this. Thank you.

Probably, we should use this way for other `new` as well at least from now on. The following kind of code doesn't work properly but just causes abnormal termination. Do I get it correct?
```
1218   /* Initialize the log block data buffer */
1219   /* *INDENT-OFF* */
1220   vacuum_Block_data_buffer = new lockfree::circular_queue<vacuum_data_entry> (VACUUM_BLOCK_DATA_BUFFER_CAPACITY);
1221   /* *INDENT-ON* */
1222   if (vacuum_Block_data_buffer == NULL)
1223     {
1224       goto error;
1225     }
```

src/communication/network_sr.c
wrlawodms: Understood. Then, how about putting the `mmon_initialize()` in here, `net_server_start`? The following location could be an option. The parameter and error service seem ready here.
```
1129   if (er_init (NULL, prm_get_integer_value (PRM_ID_ER_EXIT_ASK)) != NO_ERROR)
1130     {
1131       PRINT_AND_LOG_ERR_MSG ("Failed to initialize error manager\n");
1132       status = -1;
1133       goto end;
1134     }
1135.  // =================================== HERE ==============================
1136   net_server_init ();
1137   css_initialize_server_interfaces (net_server_request, net_server_conn_down);
1138
```
My intention is basically to initialize and finalize in the same layer if possible.

src/base/memory_monitor_sr.cpp
hornetmj: When server_name is NULL, server just leave a notification error. After that, I think constructing memory_monitor is failed due to `m_server_name {server_name}`. In this case the `OOM` error message is not a good choice.

src/transaction/boot_sr.c
```diff
@@ -2264,6 +2265,13 @@ boot_restart_server (THREAD_ENTRY * thread_p, bool print_restart, const char *db
 
   event_log_init (db_name);
 
+#if defined(SERVER_MODE)
+  if (mmon_initialize (db_name) != NO_ERROR)
+    {
+      goto error;
```
hornetmj: Don't you need to add `mmon_finalize ();` after error: label?

src/base/memory_monitor_sr.cpp
Rudeus: I got it. I'll fix it.

src/base/memory_monitor_sr.cpp
Rudeus: If `memory_monitor` is merged into the develop branch, then there is no need to use `std::nothrow` for new operators within the scope of `memory_monitor`. This is because if allocation fails, the return will proceed similarly to nothrow. However, the usage presented under the assumption of generally using new is incorrect.

src/communication/network_sr.c
Rudeus: I try to locate the init function that you mentioned when I implement the PoC of this project, but it failed. :(
While not all the details of why it failed are retained over time, I recall there was something lacking in initializing the `memory_monitor`. If you wish to receive the analysis findings, I will reanalyze and share them.

src/transaction/boot_sr.c
Rudeus: There is no need to add the function in error handling because `mmon_finalize()` is called after the function (boot_restart_server()) exits.

src/communication/network_sr.c
hornetmj: ~~`boot_restart_server` is also called when `restoredb` is excuted. In this case, mmon doesn't need to 'initialize' and 'finalize' even if system parameter turn on.~~ @wrlawodms's suggestion sounds like very interesting information.

NOTE:
@Rudeus remind me `boot_restart_sever` is not called because of #if defined(SERVER_MODE) ~ #endif macro.

src/base/memory_monitor_sr.cpp
hornetmj: This comment was made because it was judged that the code was written to mean that mmon initialization is allowed even if server_name is nullptr in release mode. Therefore, it would be better to leave `assert (server_name != nullptr)`.

hornetmj (approved):

src/communication/network_sr.c
wrlawodms: ＞ I recall there was something lacking in initializing the memory_monitor. If you wish to receive the analysis findings, I will reanalyze and share them.

Yeah, I'd love it if you share it. What do you mean by "lacking" by the way? Did you run into a failure? or some memstat was missed?

Rudeus: Boot fail occurred. "lacking" means it is not enough to initialize memory_monitor at that time.

src/base/memory_monitor_sr.cpp
joohok: I understand that `nothrow` was likely used in this function to return an error code, enabling the caller to handle errors. Therefore, shouldn't this `nothrow` be retained even when merged into the develop branch?

joohok (approved):

src/communication/network_sr.c
Rudeus: After applying the subsequent PR, there remains a possibility of issues arising, but as of now, it's deemed problem-free, so the function call has been moved to `net_server_start()`. This can be revisited for discussion if problems arise in the future.

src/communication/network_sr.c
```diff
@@ -1131,7 +1132,12 @@ net_server_start (const char *server_name)
       status = -1;
       goto end;
     }
-
+#if defined(SERVER_MODE)
+  if (mmon_initialize (server_name) != NO_ERROR)
+    {
+      goto end;
+    }
+#endif
```
wrlawodms:
```suggestion
  if (mmon_initialize (server_name) != NO_ERROR)
    {
      PRINT_AND_LOG_ERR_MSG ("Failed to initialize memont monitor.\n");
      status = -1;
      goto end;
    }
```
1. I don't think `SERVER_MODE` macro is needed here. Otherwise, correct me.
2. Follow the error handling in this function. 
3. If you want to keep running even though mmon is unavailable for some reason, remove `goto end;` and just print a warning error. (I'm not sure this is required for the mmon)

wrlawodms (changes requested):

src/communication/network_sr.c
wrlawodms: > I don't think SERVER_MODE macro is needed here. Otherwise, correct me.

Among cs,sa,cubrid/CMakelist.txt, only cubrid has the `${COMMUNICATION_DIR}/network_sr.c`.

Rudeus: You're right. I modify it.

wrlawodms (approved):

PR #4971, opened 2024-02-22
[CBRD-25224] Implement wrapping functions of allocation functions

Rudeus:
http://jira.cubrid.org/browse/CBRD-25224

Implement wrappings functions of allocation functions and overload C++ new/delete for tracking memory:
 - add `cub_alloc()` / `cub_calloc()` / `cub_realloc()` / `cub_strcpy()` / `cub_free()` for wrapping allocation functions
 - add `get_allocated_size()` for copying data of old memory chunk in `cub_realloc()`
 - add definitions of malloc/calloc/realloc/strdup/free for changing the functions to wrapping functions
 - add overloaded `new` / `delete` to track memory
 - add a definition of `new` for changing original new to overloaded new

### DIFF
--- a/src/base/memory_cppwrapper.hpp
+++ b/src/base/memory_cppwrapper.hpp
@@ -28,14 +28,12 @@
 #ifdef SERVER_MODE
 inline void *operator new (size_t size, const char *file, const int line)
 {
-  void *p = cub_alloc (size, file, line);
-  return p;
+  return cub_alloc (size, file, line);
 }
 
 inline void *operator new[] (size_t size, const char *file, const int line)
 {
-  void *p = cub_alloc (size, file, line);
-  return p;
+  return cub_alloc (size, file, line);
 }
 
 inline void operator delete (void *ptr) noexcept
--- a/src/base/memory_cppwrapper.hpp
+++ b/src/base/memory_cppwrapper.hpp
@@ -26,6 +26,10 @@
 #include "memory_cwrapper.h"
 
 #ifdef SERVER_MODE
+// TODO: The usage of operator new encompasses various additional methods beyond basic usage.
+// However, as CUBRID does not currently utilize such additional methods, they are not overloaded.
+// It has been decided that overloading will be undertaken should any issues arise from
+// the discovery of the utilization of these additional methods.
 inline void *operator new (size_t size, const char *file, const int line)
 {
   return cub_alloc (size, file, line);
--- a/src/base/memory_cppwrapper.hpp
+++ b/src/base/memory_cppwrapper.hpp
@@ -1,0 +1,54 @@
+/*
+ *
+ * Copyright 2016 CUBRID Corporation
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+
+/*
+ * memory_cppwrapper.hpp - Overloading operator new, delete
+ */
+
+#ifndef _MEMORY_CPPWRAPPER_HPP_
+#define _MEMORY_CPPWRAPPER_HPP_
+
+#include "memory_cwrapper.h"
+
+#ifdef SERVER_MODE
+inline void *operator new (size_t size, const char *file, const int line)
+{
+  void *p = cub_alloc (size, file, line);
+  return p;
+}
+
+inline void *operator new[] (size_t size, const char *file, const int line)
+{
+  void *p = cub_alloc (size, file, line);
+  return p;
+}
+
+inline void operator delete (void *ptr) noexcept
+{
+  cub_free (ptr);
+}
+
+inline void operator delete (void *ptr, size_t sz) noexcept
+{
+  cub_free (ptr);
+}
+
+#define new new(__FILE__, __LINE__)
+#endif // SERVER_MODE
+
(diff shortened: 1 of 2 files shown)


src/base/memory_cppwrapper.hpp
```diff
@@ -0,0 +1,54 @@
...
+#ifndef _MEMORY_CPPWRAPPER_HPP_
+#define _MEMORY_CPPWRAPPER_HPP_
+
+#include "memory_cwrapper.h"
+
+#ifdef SERVER_MODE
+inline void *operator new (size_t size, const char *file, const int line)
+{
+  void *p = cub_alloc (size, file, line);
+  return p;
```
wrlawodms: Nit.
```suggestion
  return cub_alloc (size, file, line);
```

src/base/memory_cppwrapper.hpp
```diff
@@ -0,0 +1,54 @@
...
+}
+
+inline void *operator new[] (size_t size, const char *file, const int line)
```
wrlawodms, line 38: Doesn't it need its counterpart `delete[]`?
`void operator delete[]( void* ptr )`

src/base/memory_cwrapper.h
```diff
@@ -0,0 +1,152 @@
...
+    }
+}
+
+inline void
+cub_free (void *ptr)
+{
+  if (mmon_is_memory_monitor_enabled () && ptr != NULL)
+    {
+      mmon_sub_stat ((char *) ptr);
+      assert (malloc_usable_size (ptr) != 0);
```
wrlawodms, line 53: It seems to be a precondition, not postcondition. Then, it should be above `mmon_sub_stat`. Furthermore, I think it should be in the `mmon_sub_stat` if it's the case.

src/base/memory_cwrapper.h
```diff
@@ -0,0 +1,152 @@
...
+  char *ret = NULL;
+
+  p = cub_alloc (strlen (str) + 1, file, line);
+  if (p != NULL)
+    {
+      ret = (char *) p;
+      memcpy (ret, str, strlen (str) + 1);
+    }
+
+  return ret;
```
wrlawodms: Nit.
```suggestion
  void *p = NULL;

  p = cub_alloc (strlen (str) + 1, file, line);
  if (p != NULL)
    {
      memcpy (p, str, strlen (str) + 1);
    }

  return (char *) p;
```

wrlawodms (approved):

src/base/memory_cwrapper.h
```diff
@@ -0,0 +1,152 @@
...
+  if (mmon_is_memory_monitor_enabled ())
+    {
+      p = malloc (size + cubmem::MMON_METAINFO_SIZE);
+      if (p != NULL)
+	{
+	  mmon_add_stat ((char *) p, malloc_usable_size (p), file, line);
+
+	  if (ptr != NULL)
+	    {
+	      memcpy (p, ptr, get_allocated_size (ptr));
```
hornetmj: Don't you need to consider size(old) > size(new)?

hornetmj: NIT:

It would be more readable if you use prefix or suffix like old_ and new_ or _old and _new at the variable name.

src/base/memory_cwrapper.h
```diff
@@ -0,0 +1,152 @@
...
+  return p;
+}
+
+inline char *
+cub_strdup (const char *str, const char *file, const int line)
+{
+  void *p = NULL;
+  char *ret = NULL;
+
+  p = cub_alloc (strlen (str) + 1, file, line);
```
hornetmj, line 143: You can remove one of local variable if you use casting operation like (char *) cub_alloc().

src/base/memory_cppwrapper.hpp
```diff
@@ -0,0 +1,54 @@
...
+  void *p = cub_alloc (size, file, line);
+  return p;
+}
+
+inline void operator delete (void *ptr) noexcept
+{
+  cub_free (ptr);
+}
+
+inline void operator delete (void *ptr, size_t sz) noexcept
```
hornetmj, line 48: Whar for?

hornetmj (approved):

src/base/memory_cppwrapper.hpp
Rudeus: Done.

src/base/memory_cppwrapper.hpp
Rudeus: `delete[]` calls `delete`, so we don't need to care about `delete[]`.

src/base/memory_cwrapper.h
Rudeus: I think that it has to move before calling `mmon_sub_stat()`.

src/base/memory_cwrapper.h
Rudeus: You're right. I fix it.

src/base/memory_cwrapper.h
Rudeus: I add conditions what `realloc()` has to do. Thanks :)

src/base/memory_cwrapper.h
Rudeus: Done.

src/base/memory_cppwrapper.hpp
Rudeus: According to cppreference,
`delete (ptr, sz)` Called instead of `delete (ptr)` if a user-defined replacement is provided, except that it's unspecified whether `delete (ptr)` or `delete (ptr, sz)` is called when deleting objects of incomplete type and arrays of non-class and trivially-destructible class types. A memory allocator can use the given size to be more efficient.
 The standard library implementations are identical to `delete (ptr)`.
 
 When testing PoC version, some delete cases goes to `delete (ptr, sz)` so I added it.

joohok (commented):
Q. 
according to https://en.cppreference.com/w/cpp/memory/new/operator_new 

1. Will both new and delete supporting various argument values be implemented? 
2. Also, will both new[] and delete[] be supported? Currently, it seems that only new[] is supported. Does new[] internally call new ?

joohok (approved):

PR #4997, opened 2024-02-28
[CBRD-25225] Include headers of wrapping functions into the cubrid files

Rudeus:
http://jira.cubrid.org/browse/CBRD-25225

Include headers 'memory_cwrapper.h' and 'memory_cppwrapper.hpp' into CUBRID source code :
 - These headers should be included at the end of including phase. This is because if they are included in the middle of including phase, they will affect allocations of glibc and will make build failure.
 - Especially, `memory_cppwrapper.hpp` can influence all new/delete occurrences within glibc, and caution must be exercised due to the user-defined `#define new new(__FILE__, __LINE__)` line within that file, which can cause incorrect usage for all overloaded new functions used internally by glibc, potentially leading to build failures.
 - For the same reason, `memory_cppwrapper.hpp` cannot be included in .h/.hpp files.

 **IMPORTANT** 
This PR may fail in the license check as it involves files not covered by the CUBRID license. And this PR may also fail in the cppcheck because 3rdparty malloc(malloc2.8.3) fails to satisfy cpp checking rules.  However, this is an expected outcome.


src/thread/thread_looper.cpp
```diff
@@ -36,7 +37,7 @@ namespace cubthread
   static const cubperf::statset_definition Looper_statistics =
   {
     cubperf::stat_definition (STAT_LOOPER_SLEEP_COUNT_AND_TIME, cubperf::stat_definition::COUNTER_AND_TIMER,
-			      "looper_sleep_count", "looper_sleep_time"),
+                              "looper_sleep_count", "looper_sleep_time"),
```
Rudeus:
```suggestion
			      "looper_sleep_count", "looper_sleep_time"),
```

src/base/memory_cwrapper.h
```diff
@@ -27,11 +27,28 @@
...
+#if defined(__SVR4)
+// *INDENT-OFF*
+extern "C" size_t malloc_usable_size (void *);
+// *INDENT-ON*
+#elif defined(__APPLE__)
+#include <malloc/malloc.h>
+
+#ifndef HAVE_USR_INCLUDE_MALLOC_H
+#define HAVE_USR_INCLUDE_MALLOC_H
+#endif
```
joohok, line 54: Do we support the OSes other than linux or windows?

src/base/memory_cwrapper.h
```diff
@@ -88,7 +107,9 @@ cub_calloc (size_t num, size_t size, const char *file, const int line)
       if (p != NULL)
 	{
 	  memset (p, 0, num * size + cubmem::MMON_METAINFO_SIZE);
+#if !defined(WINDOWS)
```
joohok, line 121: how about making `mmon_is_memory_monitor_enabled ()` to return false when it is running on Windows? 

Because, on Windows, it does not collect the stat.

joohok: Ah, this is for windows build.

src/compat/db_json.cpp
```diff
@@ -870,7 +871,23 @@ JSON_VALIDATOR &JSON_VALIDATOR::operator= (const JSON_VALIDATOR &copy)
   if (this != &copy)
     {
       this->~JSON_VALIDATOR ();
-      new (this) JSON_VALIDATOR (copy);
+      if (copy.m_document == NULL)
```
joohok, line 874: Q. why is the `JSON_VALIDATOR::JSON_VALIDATOR (const JSON_VALIDATOR &copy)` not called directly? 
Is it because of overloading with `new` that the `strdup` inside the constructor is being ignored ?

src/heaplayers/heaps/top/mallocheap.h
```diff
@@ -27,7 +27,9 @@ extern "C" size_t malloc_usable_size (void *);
 #elif defined(__linux__)
 #include <malloc.h>
 #else
-extern "C" size_t malloc_usable_size (void *) throw ();
+extern "C" size_t
+malloc_usable_size (void *)
+throw ();
```
joohok:
```suggestion
extern "C" size_t malloc_usable_size (void *) throw ();
```
Modified during indentation?

joohok: If you refer this macros and use it in src/base/memory_cwrapper.h, then I think this line also need to added in src/base/memory_cwrapper.h.

hornetmj:
Question:
When including the `memory_cppwrapper.h` or `memory_cwrapper.h` file, should the developer check whether new, delete, etc. are used? You may have considered providing only one header file called like `memory_wrapper.h`. What are the pros and cons?
```
src/base/error_manager.c:  er_Singleton_context_p = new context ();
```

Rudeus:
> Question: When including the `memory_cppwrapper.h` or `memory_cwrapper.h` file, should the developer check whether new, delete, etc. are used? You may have considered providing only one header file called like `memory_wrapper.h`. What are the pros and cons?
> 
> ```
> src/base/error_manager.c:  er_Singleton_context_p = new context ();
> ```

@hornetmj 
Since `memory_cppwrapper.hpp` includes `memory_cwrapper.h`, including `memory_cppwrapper.hpp` alone at the target file covers both C and C++ cases. However, due to constraints inherent in `memory_cppwrapper.hpp` (e.g., it can only be included in .c/.cpp files and must be included last of the including phase), its generality is restricted. Therefore, it is unavoidable to include it only in files using C++ operators such as new/delete.

src/base/memory_cwrapper.h
Rudeus: We do not support WINDOWS. But I think we have to support unix-based OSes and linux.

src/compat/db_json.cpp
Rudeus: It's because `new (this)` makes build error.

src/heaplayers/heaps/top/mallocheap.h
Rudeus: The indentation result of git-hook can't make that. We have to add *INDENT-OFF* to make the function definition at one line in .h file(in .hpp or .cpp files, it doesn't make matter. So, what do you think? (one line vs. three lines)

src/thread/thread_looper.cpp
```diff
@@ -36,7 +37,7 @@ namespace cubthread
   static const cubperf::statset_definition Looper_statistics =
   {
     cubperf::stat_definition (STAT_LOOPER_SLEEP_COUNT_AND_TIME, cubperf::stat_definition::COUNTER_AND_TIMER,
-			      "looper_sleep_count", "looper_sleep_time"),
+    "looper_sleep_count", "looper_sleep_time"),
```
Rudeus:
```suggestion
			      "looper_sleep_count", "looper_sleep_time"),
```

Rudeus:
@hornetmj @joohok I integrate the including header to `memory_wrapper.hpp` except `memory_alloc.h` because `memory_wrapper.hpp` may not be included into header files.

+) re-interpret `placement new` to some code lines and remove it for `#define new new(__FILE__, __LINE__)`.

src/transaction/boot_sr.c
```diff
@@ -3330,6 +3330,7 @@ xboot_register_client (THREAD_ENTRY * thread_p, BOOT_CLIENT_CREDENTIAL * client_
 #if defined(ENABLE_SYSTEMTAP) && defined(SERVER_MODE)
   CUBRID_CONN_START (thread_p->conn_entry->client_id, client_credential->get_db_user ());
 #endif /* ENABLE_SYSTEMTAP */
+#include "memory_cwrapper.h"
```
hornetmj: intend?

src/base/memory_alloc.h
```diff
@@ -45,6 +45,7 @@
 #include <memory>
 #include <functional>
 #endif
+#include "memory_cwrapper.h"
```
hornetmj, line 54: How about leave comment why you include `_cwrapper` iso `_wrapper` with a caution 'Don't change it to _wrapper'?

src/transaction/boot_sr.c
Rudeus: No. I remove it.

src/base/memory_alloc.h
Rudeus: I add a caution comment.

src/compat/db_json.cpp
joohok: While the 'new (this)' syntax may not be widely used in CUBRID, it could still be applicable in the future. How can we promote awareness of its potential usage? leave it as restriction?

src/heaplayers/heaps/top/mallocheap.h
joohok: In other cpp and h files, it was written in a single line, so I'm wondering if there might be a mistake. For consistency, I think it is good to follow others (one line)

src/base/memory_wrapper.hpp
joohok, line 1: header macro, and description comment also required to be modified

src/base/memory_cwrapper.h
joohok: OK

joohok (commented):
For maintenance purposes or for future additions of files, 
how about adding guidelines to the wrapper header file? (memory_cwrapper.h, memory_wrapper.hpp)

It would be beneficial to include instructions such as when to use 'cwrapper,' the typical placement of the wrapper header at the end of header list, and so on.

Additionally, how can we ensure that these header is added to all files using malloc, free, new, and delete?

src/compat/db_json.cpp
Rudeus: To alleviate the constraint, the added `#define new new(FILE, LINE)` to allow for automatic substitution with overloaded new could be removed, relieving the burden on users (developers). However, considering it currently impractical to remove and add it to every code line, we opted to retain the constraint for now.

src/heaplayers/heaps/top/mallocheap.h
Rudeus: OK. I change it.

src/base/memory_wrapper.hpp
Rudeus: Done.

Rudeus:
> For maintenance purposes or for future additions of files, how about adding guidelines to the wrapper header file? (memory_cwrapper.h, memory_wrapper.hpp)
> 
> It would be beneficial to include instructions such as when to use 'cwrapper,' the typical placement of the wrapper header at the end of header list, and so on.
> 
> Additionally, how can we ensure that these header is added to all files using malloc, free, new, and delete?

I add the description(I think it can be the guideline) at the header files.

hornetmj (approved):

joohok (approved):

PR #5008, opened 2024-03-04
[CBRD-25243] case-insensitive hostname for CUBRID userhosts

kisoo-han:
http://jira.cubrid.org/browse/CBRD-25243

**Description**
* case-insensitive hostname service for CUBRID userhosts, like that of GNU
* '**Cent7**' and '**cent7**' will be recognized as same host names.

**Implementation**
* to simplify modifications
  * we will store **{host name converted to uppercase, ipAddress}** in unordered_map
  * hostname will be converted to uppercase, and the search will proceed in unordered_Map.

**Remarks**
* We also neeed to make the function **are_hostnames_equal** () to case-insensitive
* check HA startup for the following configuration (with GNU host lookup, **use_user_hosts=no** in cubrid.conf)
  * hostname: '**dev1**' in lowercase, /etc/hosts entry, uppercase '**DEV1**'
```
   $ cat /etc/hosts
   192.168.2.34 DEV1
   192.168.2.38 DEV2
   $ hostname
    dev1

```
* this issue will be handled in [[CBRD-25245](http://jira.cubrid.org/browse/CBRD-25245)] first, and then will progress this issue

### DIFF
--- a/src/executables/util_common.c
+++ b/src/executables/util_common.c
@@ -545,7 +545,7 @@ are_hostnames_equal (const char *hostname_a, const char *hostname_b)
   const char *a;
   const char *b;
 
-  for (a = hostname_a, b = hostname_b; *a && *b && (toupper (*a) == toupper (*b)); ++a, ++b)
+  for (a = hostname_a, b = hostname_b; *a && *b && (*a == *b); ++a, ++b)
     ;
 
   if (*a == '\0' && *b != '\0')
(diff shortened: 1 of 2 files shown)


src/executables/util_common.c
mhoh3963, line 1: @kisoo-han 
You would be better to create a separate issue and fix it, because the function is not related to this issue.

mhoh3963 (approved):

airnet73 (approved):

beyondykk9 (approved):

PR #5017, opened 2024-03-07
[CBRD-25248] Server crash for Java SPs running a SELECT statement with 'WHERE 0 <> 0' clause

hgryoo:
http://jira.cubrid.org/browse/CBRD-25248

For false query such as "SELECT * FROM db_class WHERE 0 <> 0", dummy query id is set.
I've changed to set the query_id as NULL_QUERY_ID if a query is not executed (`tuple_count <= 0`).

### DIFF
--- a/src/method/method_invoke_group.cpp
+++ b/src/method/method_invoke_group.cpp
@@ -343,6 +343,12 @@ namespace cubmethod
   query_cursor *
   method_invoke_group::create_cursor (QUERY_ID query_id, bool oid_included)
   {
+    if (query_id == NULL_QUERY_ID || query_id >= SHRT_MAX)
+      {
+	// false query e.g) SELECT * FROM db_class WHERE 0 <> 0
+	return nullptr;
+      }
+
     m_cursor_set.insert (query_id);
     return m_rctx->create_cursor (m_thread_p, query_id, oid_included);
   }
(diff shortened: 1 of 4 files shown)


src/method/method_invoke_group.cpp
```diff
@@ -343,6 +343,12 @@ namespace cubmethod
   query_cursor *
   method_invoke_group::create_cursor (QUERY_ID query_id, bool oid_included)
   {
+    if (query_id == NULL_QUERY_ID || query_id >= SHRT_MAX)
```
beyondykk9, line 346: what's the reason for SHRT_MAX is used to check validation ?

hyunikn (approved):

src/method/method_invoke_java.cpp
```diff
@@ -523,6 +520,7 @@ namespace cubmethod
 	  }
 	i++;
       }
+
```
beyondykk9: usless

src/method/method_invoke_group.cpp
hgryoo: There is the same validation in query manager.

src/method/method_invoke_java.cpp
hgryoo:
```suggestion
```

beyondykk9 (approved):

ctshim (approved):

PR #5018, opened 2024-03-08
[CBRD-25249] Fix not to resolve names of static SQL without considering user variables in compiling PL/CSQL

hgryoo:
http://jira.cubrid.org/browse/CBRD-25249

When compiling Static SQL, the query parser interprets PL's variable as a SQL's user variable. In PL/CSQL compilation, only static information should be considered, without regarding to runtime values.

### DIFF
--- a/src/parser/name_resolution.c
+++ b/src/parser/name_resolution.c
@@ -941,9 +941,6 @@ pt_bind_name_or_path_in_scope (PARSER_CONTEXT * parser, PT_BIND_NAMES_ARG * bind
       /* If pt_name in group by/ having, maybe it's alias. We will try to resolve it later. */
       if (!is_pt_name_in_group_having (in_node))
 	{
-	  /* it may be a naked parameter or a path expression anchored by a naked parameter. Try and resolve it as
-	   * such. */
-	  node = pt_bind_parameter_path (parser, in_node);
 
 	  if (node == NULL && parser->flag.is_parsing_static_sql == 1)
 	    {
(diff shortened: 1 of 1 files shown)


hyunikn (approved):

src/parser/name_resolution.c
```diff
@@ -941,9 +941,6 @@ pt_bind_name_or_path_in_scope (PARSER_CONTEXT * parser, PT_BIND_NAMES_ARG * bind
       /* If pt_name in group by/ having, maybe it's alias. We will try to resolve it later. */
       if (!is_pt_name_in_group_having (in_node))
 	{
-	  /* it may be a naked parameter or a path expression anchored by a naked parameter. Try and resolve it as
-	   * such. */
-	  node = pt_bind_parameter_path (parser, in_node);
 
 	  if (node == NULL && parser->flag.is_parsing_static_sql == 1)
```
ctshim: If the code at line 946 is removed, the `node == NULL` condition at line 948 will always be true.
This is because the condition for entering this block is when node is NULL.

ctshim (approved):

beyondykk9 (approved):

src/parser/name_resolution.c
hgryoo: I see. The condition seems redundant. I'm going to change as follow. Thank you.
```suggestion
	  if (parser->flag.is_parsing_static_sql == 1)
```

PR #5030, opened 2024-03-12
[CBRD-25193] [bugfix] TO_NUMBER('1234567', '9999999') call results in a compile time error

hgryoo:
http://jira.cubrid.org/browse/CBRD-25193

the arg2's type is checked in query parser. This information should be checked in type checking.


hgryoo:
will be fixed at #5060

PR #5031, opened 2024-03-13
[CBRD-25199] Make Unit-test of memory monitor module

Rudeus:
http://jira.cubrid.org/browse/CBRD-25199

Unit tests for memory_monitor class:
 - Devide `memory_monitor_sr.cpp` into two files with class methods of memory_monitor(`memory_monitor_sr.cpp`) and APIs(`memory_monitor_api.cpp`) because of minimize dependency of unit-test
   - ex) mmon_initialize() need system_parameter.c, error_manager.c and error_manager.c needs something.c and so on... But the test don't need mmon_initialize() that sever uses.
   - modify `cubrid/CMakelists.txt`
   - modify `util/CMakelists.txt` for `unittest_lf`
 - Modify `CMakelists.txt` of unit_tests directory to add unit-test directory for memory_monitor and add `CATCH2` external project
 - Add a new unit test with main source file(`test_memory_monitor_main.cpp`) and its `CMakelists.txt`
 - Modify `build.sh` for  make a executable file of the unit test of memory_monitor class

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -33,7 +33,7 @@ build_generator="make"
 source_dir=`pwd`
 default_java_dir="/usr/lib/jvm/java"
 java_dir=""
-configure_options=""
+configure_options="-DUNIT_TEST_MEMORY_MONITOR=ON"
 # default build_dir = "$source_dir/build_${build_target}_${build_mode}"
 build_dir=""
 prefix_dir=""
--- a/build.sh
+++ b/build.sh
@@ -33,7 +33,9 @@ build_generator="make"
 source_dir=`pwd`
 default_java_dir="/usr/lib/jvm/java"
 java_dir=""
-configure_options="-DUNIT_TEST_MEMORY_MONITOR=ON"
+configure_options=""
+# if you turn on the unit test of memory monitor
+# configure_options="-DUNIT_TEST_MEMORY_MONITOR=ON" or "-DUNIT_TESTS=ON"
 # default build_dir = "$source_dir/build_${build_target}_${build_mode}"
 build_dir=""
 prefix_dir=""
--- a/src/base/memory_monitor_sr.cpp
+++ b/src/base/memory_monitor_sr.cpp
@@ -69,6 +69,18 @@ namespace cubmem
     std::string target ("/src/");
 #endif // !WINDOWS
 
+    // TODO: To minimize the search cost and prevent unnecessary paths
+    //       from being included in the tag_name, we are cutting the string based on
+    //       the rightmost "/src/" found. However, in this case, when the allocation
+    //       occurs on the same 'line', "/src/test.c" and "/src/thirdparty/src/test.c"
+    //       get the same tag_name, making it impossible to distinguish memory alloc-
+    //       ations from two different files. However, such exceptional situations
+    //       are very rare, as memory allocations must occur on the same line number.
+    //       Handling these exceptions would increase the overall cost of this function.
+    //       Moreover, currently, although these exceptional situations theoretically exist,
+    //       they have not been encountered in developer tests.
+    //       Therefore, it has been decided to address them if discovered in the future.
+
     // Find the last occurrence of "src" in the path
     size_t pos = filecopy.rfind (target);
 #if !defined (NDEBUG)
--- a/src/base/memory_monitor_sr.cpp
+++ b/src/base/memory_monitor_sr.cpp
@@ -40,8 +40,6 @@ extern "C" size_t malloc_usable_size (void *);
 #endif
 #endif
 
-#include "error_manager.h"
-#include "system_parameter.h"
 #include "memory_monitor_sr.hpp"
 
 typedef struct mmon_metainfo MMON_METAINFO;
@@ -54,8 +52,6 @@ struct mmon_metainfo
 
 namespace cubmem
 {
-  memory_monitor *mmon_Gl = nullptr;
-
   memory_monitor::memory_monitor (const char *server_name)
     : m_server_name {server_name},
       m_magic_number {*reinterpret_cast <const int *> ("MMON")},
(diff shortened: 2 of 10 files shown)


joohok:
When is the `get_allocated_size ()` called?

build.sh
```diff
@@ -33,7 +33,7 @@ build_generator="make"
 source_dir=`pwd`
 default_java_dir="/usr/lib/jvm/java"
 java_dir=""
-configure_options=""
+configure_options="-DUNIT_TEST_MEMORY_MONITOR=ON"
```
joohok: will this be removed when this PR is merged?

src/base/memory_monitor_sr.cpp
```diff
@@ -73,6 +69,18 @@ namespace cubmem
...
+    //       from being included in the tag_name, we are cutting the string based on
+    //       the rightmost "/src/" found. However, in this case, when the allocation
+    //       occurs on the same 'line', "/src/test.c" and "/src/thirdparty/src/test.c"
+    //       get the same tag_name, making it impossible to distinguish memory alloc-
+    //       ations from two different files. However, such exceptional situations
+    //       are very rare, as memory allocations must occur on the same line number.
+    //       Handling these exceptions would increase the overall cost of this function.
+    //       Moreover, currently, although these exceptional situations theoretically exist,
+    //       they have not been encountered in developer tests.
+    //       Therefore, it has been decided to address them if discovered in the future.
```
joohok, line 82: Hmm, since this is an implementation part that leaves room for potential malfunctions rather than optimization, I think it would be better to create issue as a backlog to track easily rather than leaving it as a TODO comment. What do you think?

unit_tests/memory_monitor/test_memory_monitor_main.cpp
```diff
@@ -0,0 +1,297 @@
+/*
+ * Copyright 2008 Search Solution Corporation
```
joohok:
```suggestion
```

unit_tests/memory_monitor/CMakeLists.txt
```diff
@@ -0,0 +1,49 @@
...
+project(test_memory_monitor)
+
+#
+# test_memory_monitor
+#
+
+set(TEST_MEMORY_MONITOR_SOURCES
+  ${BASE_DIR}/memory_monitor_sr.cpp
+  test_memory_monitor_main.cpp
+  )
```
joohok, line 27: don't we need to set headers here? 

```
set(TEST_MEMORY_MONITOR_HEADERS
  $(BASE_DIR)/memory_monitor_sr.hpp
     )
```

unit_tests/memory_monitor/test_memory_monitor_main.cpp
```diff
@@ -0,0 +1,297 @@
...
+#include "catch2/catch.hpp"
+
+#include <cstdlib>
+#include <malloc.h>
+#include <thread>
+
+#include "memory_monitor_sr.hpp"
+
+// For multithread test
+static const int num_threads = 100;
```
joohok:
```suggestion
constexpr int num_threads = 100;
```
If there is no duplicated definition for `num_threads`, I think we can omit `static` keyword

unit_tests/memory_monitor/CMakeLists.txt
joohok: AH, you include this header in unit test file.

unit_tests/memory_monitor/test_memory_monitor_main.cpp
```diff
@@ -0,0 +1,297 @@
...
+ * they must consider the entire test suite.
+ */
+
+// Test add_stat()
+//
+// Test add_stat() with single thread when different file paths
+// that will be made same stat name are given
+//
+// Expected Result: stat name "add_test.c:100" with value "allocated_size + allocated_size_2"
+TEST_CASE ("Test mmon_add_stat", "")
```
joohok, line 122: I think one more case is required.

This scenario seems like an abnormal case. How about adding the normal operation case first? 

and then when adding the abnormal case, additional explanation may be needed.
For the abnormal case explanation, I think it is enough to summarize it as "an issue where the module fails to distinguish different path names".

unit_tests/memory_monitor/test_memory_monitor_main.cpp
```diff
@@ -0,0 +1,297 @@
...
+  std::vector <std::thread> threads;
+  size_t allocated_size[num_threads];
+  size_t total_allocated_size = 0;
+  int ret;
+
+  mmon_aggregate_server_info (test_server_info);
+  ret = find_test_stat (test_server_info, "base/add_test_multithread.c:100");
+  REQUIRE (ret == -1);
+  test_server_info.stat_info.clear();
+  test_mem_in_the_scope_multithread = (char **) malloc (sizeof (char *) * num_threads);
```
joohok, line 213: I'm not sure, but do we have to free these allocated memory in unit test?

build.sh
Rudeus: Is this right to remove the unit-test option? Please give your opinion.

src/base/memory_monitor_sr.cpp
Rudeus: I think TODO comment is better than backlog issue because
 - That's rare case so the error-case scenario will be appear very far-future. In that case, I think comment is more better to track than issue.
 - I already decide to leave comment through offline meeting with @hornetmj. We think that's enough.

unit_tests/memory_monitor/CMakeLists.txt
Rudeus: The header is automatically added in `unit_tests/CMakeLists.txt`.

unit_tests/memory_monitor/test_memory_monitor_main.cpp
Rudeus: I don't understand your intention of the question. Can you explain more details?

unit_tests/memory_monitor/test_memory_monitor_main.cpp
Rudeus: I agree.

unit_tests/CMakeLists.txt
```diff
@@ -82,3 +99,7 @@ if (UNIT_TESTS OR UNIT_TEST_MONITOR)
   message("    monitor")
   add_subdirectory(monitor)
 endif(UNIT_TESTS OR UNIT_TEST_MONITOR)
+if (UNIT_TESTS OR UNIT_TEST_MEMORY_MONITOR)
```
hornetmj, line 103:
```suggestion

if (UNIT_TESTS OR UNIT_TEST_MEMORY_MONITOR)
```

build.sh
joohok: It seems more appropriate to use a different approach rather than incorporating it into the default options (`configure_options`) of this script. 
For instance, creating a separate option specifically for unit testing in the build script, 
(e.g. `$sh build.sh -u memory_monitor ..`)

or providing a separate option only when unit tests are needed would be more suitable.
(e.g. `$sh build.sh -c -DUNIT_TESTS=ON`)  

`$sh build.sh -c -DUNIT_TESTS=ON` is used when compiling a LETS code in CI

src/base/memory_monitor_sr.cpp
joohok: In fact, I'm worried and skeptical.

1. Will we be able to recognize when a problem occurs?
2. if there's a potential issue in the future, when will it be addressed?
3. can the person creating a new directory be aware of this matter?

When it comes to maintenance, we often forget most of the context. 
Even if we don't create an issue, it would be good to write constraints in the epic

unit_tests/memory_monitor/CMakeLists.txt
joohok: OK, I see

unit_tests/memory_monitor/test_memory_monitor_main.cpp
```diff
@@ -0,0 +1,345 @@
...
+  ret = find_test_stat (test_server_info, "base/add_test.c:100");
+  REQUIRE (ret != -1);
+  REQUIRE (test_server_info.stat_info[ret].second == allocated_size_2);
+}
+
+// Test add_stat() w/ abnormal case
+//
+// Test add_stat() with single thread when different file paths
+// that will be made same stat name are given (abnormal case)
+// There is a potential issue scenario that can't distinguish different allocation. (TODO)
```
joohok: How about write a function name ? 

```suggestion
// There is a potential issue scenario that can't distinguish different allocation. (TODO in memory_monitor::make_stat_name ())
```

unit_tests/memory_monitor/test_memory_monitor_main.cpp
joohok: Ah, this allocated memory is `free`d in other function. 

I thought `malloc`ed memory should be freed but I can not find the part that free the memory within this function.

joohok (approved):
Nit comments

unit_tests/memory_monitor/test_memory_monitor_main.cpp
Rudeus: They are global variable so they free their memory at the last test (sub_stat w/ multithreads).

build.sh
Rudeus: I understand. Thanks.

src/base/memory_monitor_sr.cpp
Rudeus: I added it  as limitation part in http://jira.cubrid.org/browse/CBRD-25201.

hornetmj (approved):

PR #5032, opened 2024-03-13
[CBRD-25259] 'CTRL C' does not work while executing query in SA mode

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-25259

init tm_Tran_check_interrupt to true.
thread_p->check_interrupt for SERVER_MODE is initialized to true. tm_Tran_check_interrupt is same variable as thread_p->check_interrupt. it is for SA_MODE. I think initializing it to false was just a mistake.


youngjinj (approved):

hornetmj:
@shparkcubrid Instead of changing the default value, how about setting it to true where necessary? SM code that needs to prevent `interrupt` is given explicit handling, as follows:
```
(void) logtb_set_check_interrupt (thread_p, false);
disk_add_volume ();
logtb_set_check_interrupt (thread_p, check_interrupt);
```
However, if this processing is missing somewhere, there is a possibility that changing the default value may cause problems.  Therefore, I would like to ask if it is possible to turn this value on/off before starting the query execution and after it is completed.

shparkcubrid:
> @shparkcubrid Instead of changing the default value, how about setting it to true where necessary? SM code that needs to prevent `interrupt` is given explicit handling, as follows:
> 
> ```
> (void) logtb_set_check_interrupt (thread_p, false);
> disk_add_volume ();
> logtb_set_check_interrupt (thread_p, check_interrupt);
> ```
> 
> However, if this processing is missing somewhere, there is a possibility that changing the default value may cause problems. Therefore, I would like to ask if it is possible to turn this value on/off before starting the query execution and after it is completed.


> tm_Tran_check_interrupt

thread_p->check_interrupt for SERVER_MODE is initialized to true. tm_Tran_check_interrupt is same variable as thread_p->check_interrupt. it is for SA_MODE.
I think initializing it to false was just a mistake.

xmilex-git (approved):

hornetmj (approved):

PR #5038, opened 2024-03-15
[CBRD-25241] Add SQL syntax to change the owner of the serial

jongmin-won:
http://jira.cubrid.org/browse/CBRD-25241

To change the owner of the current serial, the 'CALLchange_serial_owner()' method must be performed, and the results of this method are successfully synchronized in the HA environment.

However, for user convenience, it is necessary to add an SQL statement to change the owner of serial.


src/query/execute_statement.c
```diff
@@ -2846,6 +2860,28 @@ do_alter_serial (PARSER_CONTEXT * parser, PT_NODE * statement)
 	  goto end;
 	}
       pr_clear_value (&value);
+      opt_serial_option_list_flag = true;
+    }
+
+  /* owner to */
+  serial_owner = statement->info.serial.owner_name;
+  if (serial_owner != NULL)
+    {
```
hgryoo: Given that the existence of the `owner_name` node is guaranteed during the syntax check, this condition doesn't have to be checked. It would be more suitable to add `assert(serial_owner != NULL)` instead of the if-statement.

hgryoo (approved):

src/query/execute_statement.c
jongmin-won: Thank you for the code review.

As you mentioned, the existence of the owner_name node is guaranteed by performing a syntax check in advance.

The reason for adding the if-statement is as follows.

In the ALTER SERIAL syntax, both options below are performing the do_alter_serial() function.
Option 1) opt_serial_option_list(INCREMENT BY, START WITH, MINVALUE, MAXVALUE, CYCLE, CACHE)
Option 2) OWNER TO

Therefore, in the do_alter_serial() function, an if statement was added to check whether the OWNER TO option is used.

beyondykk9 (approved):

src/query/execute_statement.c
hgryoo: It seems that the behavior can be controlled by `opt_serial_option_list_flag` that you introduced.

hyunikn (approved):

ctshim (approved):

src/query/execute_statement.c
jongmin-won: I revised it by referring to the offline review. Please confirm.

1) Added semantic check to PT_ALTER_SERIAL
  - A semantic check was performed along with the existing syntax check, but a semantic check was added to operate according to meaning in each session.

2) We have modified the duplicate inspection part.
  - Fixed a part in the execution session that checks for the same owner_name.

msg/de_DE.utf8/cubrid.msg
```diff
@@ -1930,6 +1930,7 @@ $set 8 MSGCAT_SET_PARSER_SEMANTIC
 314 Synonym "%1$s" does not exist.
 315 Rename cannot change owner.
 316 Rename cannot be changed to the same name.
+317 DBA and members of DBA group can perform %1$s.
```
hyunikn:
```suggestion
317 Only DBA and DBA group members can perform %1$s.
```

src/parser/csql_grammar.y
```diff
@@ -3757,21 +3757,51 @@ alter_stmt
...
+			        node->info.serial.code = PT_SERIAL_COMMENT;
+			      }
 			  }
 
 		DBG_PRINT}}
+	| ALTER				/* 1 */
+	  SERIAL			/* 2 */
+	  serial_name			/* 3 */
+	  OWNER TO identifier		/* 4, 5, 6 */
+	  opt_comment_spec		/* 7 */
```
hyunikn: What about removing opt_comment_spec?
Serial comments can be altered with the above form of ALTER SERIAL which changes serial options.

jongmin-won: Thank you for the code review.

If opt_comment_spec is removed from that syntax, OWNER TO and COMMENT cannot be used together as shown below
Therefore, OWNER TO .. opt_comment_spec is required.

ex)
ALTER SERIAL [serial_name] OWNER TO [user_name] COMMENT 'comment txt';

hyunikn: Yes. I suggest not to use COMMENT block in the new ALTER SERIAL statement. 
It would be OK because comments can be altered in the old ALTER SERIAL statement.

src/query/execute_statement.c
```diff
@@ -2874,7 +2945,7 @@ do_alter_serial (PARSER_CONTEXT * parser, PT_NODE * statement)
   obj_tmpl = NULL;
 
 end:
-  if (!OID_ISNULL (&serial_obj_id))
+  if (!OID_ISNULL (&serial_obj_id) && alter_serial_code == PT_SERIAL_OPTION)
```
hyunikn: [Q] What does serial_decache() function do? Please give a description.

src/parser/semantic_check.c
```diff
@@ -9928,6 +9930,121 @@ pt_check_kill (PARSER_CONTEXT * parser, PT_NODE * node)
...
+      cyclic = node->info.serial.cyclic;
+      no_max = node->info.serial.no_max;
+      no_min = node->info.serial.no_min;
+      no_cyclic = node->info.serial.no_cyclic;
+      cached_num_val = node->info.serial.cached_num_val;
+      no_cache = node->info.serial.no_cache;
+
+      if (!start_val && !increment_val && !max_val && !min_val
+	  && cyclic == 0 && no_max == 0 && no_min == 0
+	  && no_cyclic == 0 && !cached_num_val && no_cache == 0 && node->info.serial.comment == NULL)
```
ctshim, line 10003: Can `cycle `and `no_cycle `be 0 at the same time?

jongmin-won: This condition checks if there is no option when executing the old ALTER SERIAL statement.

As shown in the example below, if there is no option after ALTER SERIAL, cycle and no_cycle become 0 at the same time.

ex)
CREATE SERIAL cubrid_seq;
ALTER SERIAL cubrid_seq;

src/query/execute_statement.c
jongmin-won: The serial_decache() function only serves to free the cache values in the serial_cache_pool.
When the user calls nextval, it updates the serial_cache_pool and fetches the serial value in the cache_pool.

ex) When executing nextval after decache, current_val increases by the cache value.

`CREATE SERIAL cubrid_seq cache 5;`
`SELECT cubrid_seq.nextval FROM dual;`
> serial_current_value(cubrid_seq)
======================
 1  

`SELECT name, current_val, cached_num FROM db_serial WHERE unique_name='dba.cubrid_seq';`
>   name          |        current_val      |       cached_num
=========================================================
  'cubrid_seq'    |      5                       |        5

/* Use serial_decache() */
`ALTER SERIAL cubrid_seq MAXVALUE 30000;`
`SELECT name, current_val, cached_num FROM db_serial WHERE unique_name='dba.cubrid_seq';`

>   name       |      current_val       |     cached_num
=========================================================
  'cubrid_seq'      |        5               |                5

/* Update serial_cache_pool and get the serial value in cache_pool */
`SELECT cubrid_seq.nextval FROM dual;`
>   serial_next_value(cubrid_seq, 1)
======================
   6   

`SELECT name, current_val, cached_num FROM db_serial WHERE unique_name='dba.cubrid_seq';`
>   name            |      current_val     |       cached_num
=========================================================
  'cubrid_seq'      |    10                    |          5

jongmin-won: After serial_decache(), executing nextval calls the xserial_get_next_value_internal() function to update the serial_cache_pool.

> src/query/serial.c:xserial_get_next_value()

```
  ...
  /* used serial cache */
  entry = (SERIAL_CACHE_ENTRY *) mht_get (serial_Cache_pool.ht, oid_p);
  if (entry != NULL)
  {
	/* 
	 * If serail_cache_pool exists
	 * Get the serial value from cache_pool.
	 */
	...
	ret = serial_get_next_cached_value (thread_p, entry, num_alloc);
  }
  else
  {
	/* 
	 * If serail_cache_pool does not exist
	 * After creating serail_cache_pool, get the serial value from cache_pool.
	 */
	...
	ret = xserial_get_next_value_internal (thread_p, result_num, oid_p, num_alloc);
  }
  ...
```

msg/de_DE.utf8/cubrid.msg
jongmin-won: We've fixed it, please confirm

When viewed on the Files changed page, the error messages for ko_KR.euckr and ko_KR.utf8 appear to have broken text.
Therefore, please check the link below

[ko_KR_euckr](https://github.com/CUBRID/cubrid/blob/0086df5f611ed097d28e1f30b6e4744f1d6e6851/msg/ko_KR.euckr/cubrid.msg#L1933)
[ko_KR.utf8](https://github.com/CUBRID/cubrid/blob/0086df5f611ed097d28e1f30b6e4744f1d6e6851/msg/ko_KR.utf8/cubrid.msg#L1933)

ctshim (approved):

hgryoo:
1) In the following example, a user can not know `change_serial_owner` is used internally at the ALTER SERIAL statement. IMO, proper error message should be shown.

```
CREATE USER test_user1; 
CREATE SERIAL cubrid_seq;

CALL login ('public', '') ON CLASS db_user;

ALTER SERIAL dba.cubrid_seq OWNER TO test_user1;

/* 
ERROR: before ' ; '
DBA, and members of DBA group can perform change_serial_owner.
*/
```

2) About the seiral decach logic that you described in the comment of JIRA issue, It seems not related to the adding SQL syntax. Please add a new issue that changing serial decache behavior. If adding SQL syntax and changing the behavior should be resolved together, please change the title of the JIRA issue and the PR.

src/parser/parse_tree.h
```diff
@@ -1444,7 +1444,9 @@ typedef enum
   PT_CHANGE_INDEX_COMMENT,
   PT_CHANGE_INDEX_STATUS,
   PT_ADD_MEMBERS,		/* alter user type */
-  PT_DROP_MEMBERS
+  PT_DROP_MEMBERS,
+  PT_SERIAL_OPTION,		/* alter serial type */
+  PT_SERIAL_COMMENT
```
hgryoo: PT_SERIAL_COMMENT seems redundant. the comment column could be changed in the other codes (SERIAL_OPTION, CHANGE_OWNER). The behavior of corresponding code should be independent.

src/parser/csql_grammar.y
jongmin-won: The new SERIAL syntax has been removed, and OWNER has been added to the old SERIAL syntax.
please confirm

src/parser/parse_tree.h
jongmin-won: We've fixed it, please confirm.

src/query/execute_statement.c
jongmin-won: The Serial Decache logic is not related to adding SQL syntax, so we will create and address new issues.

msg/de_DE.utf8/cubrid.msg
jongmin-won: The error message 'change_serial_owner' is output in %1.
This was unclear to the user, so we modified it to be clearer as shown below. 

AS-IS
Only DBA and DBA group members can perform change_serial_owner.

TO-BE
Only DBA and DBA group members can change the SERIAL OWNER.

hgryoo (approved):

src/parser/parse_tree.h
jongmin-won: **Although it has been approved, I think that PT_SERIAL_COMMENT needs to be added again.
Because, when only the comment is executed the following two problems occur.**

**1. When only the comment is executed, it is necessary to check if the PT_SERIAL_OPTION(INCREMENT BY, START WITH, MINVALUE, MAXVALUE, CYCLE, CACHE) values exist.**
* As your already mentioned, the COMMENT can be changed in PT_SERIAL_OPTION and PT_CHANGE_OWNER.
* However, when only the comment is executed, we first check if the value of SERIAL_OPTION exists, and then modify the comment.

**2. When only the comment is executed, decache logic is executed.**

Decache should be separated from SQL syntax addition, but since it shares the same logic, so I write the content.

* Decache logic should only be executed if PT_SERIAL_OPTION.
* However, even when only the comment is executed, the decache is executed. This is because the code value of the comment is PT_SERIAL_OPTION.
* Therefore, when executing only comments, the code must have a different value in the grammar to prevent decache logic from being performed.

We added PT_SERIAL_COMMENT for the same reason as above, and also removed duplicate code to execute comments in /src/query/execute_statement.c.

hyunikn (approved):

src/parser/parse_tree.h
jongmin-won: Now that I think about it, I didn't need to add PT_SERIAL_COMMENT again.

I concluded that it was correct to remove PT_SERIAL_COMMENT because when only changing the COMMENT, it doesn't have any effect inside the switch statement and is modified externally, so I decided to remove it.

We plan to add the decache flag to the grammar part so that the decache logic can be checked in advance.

PR #5039, opened 2024-03-15
[CBRD-25254] initialize the lexer lineno on JDBC connection too

hyunikn:
http://jira.cubrid.org/browse/CBRD-25254

Lerxer lineno initialization code is added in ux_prepare() so that the line number is correct on JDBC connection too.

### DIFF
--- a/src/broker/cas_execute.c
+++ b/src/broker/cas_execute.c
@@ -916,6 +916,7 @@ ux_prepare (char *sql_stmt, int flag, char auto_commit_mode, T_NET_BUF * net_buf
       goto prepare_result_set;
     }
 
+  db_init_lexer_lineno();
   session = db_open_buffer (sql_stmt);
   if (!session)
     {
--- a/src/broker/cas_execute.c
+++ b/src/broker/cas_execute.c
@@ -880,6 +880,7 @@ ux_prepare (char *sql_stmt, int flag, char auto_commit_mode, T_NET_BUF * net_buf
 	  goto prepare_error;
 	}
 
+      db_init_lexer_lineno ();
       session = db_open_buffer (tmp);
       if (!session)
 	{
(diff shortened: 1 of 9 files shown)


kisoo-han (approved):

hgryoo (approved):

src/broker/cas_execute.c
```diff
@@ -916,6 +916,7 @@ ux_prepare (char *sql_stmt, int flag, char auto_commit_mode, T_NET_BUF * net_buf
       goto prepare_result_set;
     }
 
+  db_init_lexer_lineno ();
```
beyondykk9, line 920: why not direct call **csql_yyset_lineno**?

hyunikn: csql_yyset_lineno() cannot be referenced in cas_execute.c without including parser.h, and 
I'm not sure including parser.h in cas_execute.c is structurally OK or not.

hgryoo:
Please check the following error on Windows
```
cas_execute.obj : error LNK2019: unresolved external symbol db_init_lexer_lineno referenced in function "int __cdecl ux_prepare(char *,int,char,struct t_net_buf *,struct t_req_info *,unsigned int)" (?ux_prepare@@YAHPEADHDPEAUt_net_buf@@PEAUt_req_info@@I@Z) 
```

src/broker/cas_execute.c
ctshim: db_open_buffer() is used in several places, including on line 883.
I think there is a need to review other areas other than the added 919 line.

hyunikn: good point.

hyunikn:
I added more db_init_lexer_lineno() calls to other places which needs the initialization.
Please check them too.

beyondykk9 (approved):

PR #5045, opened 2024-03-21
[CBRD-25268] Refactor authenticate: move cache structures to a new file (AU_CLASS_CACHE, AU_USER_CACHE)

hgryoo:
http://jira.cubrid.org/browse/CBRD-25268

- Move authenticate cache related code to a new file
- Use common struct to remove duplicated code
  - AU_TYPE_SET
  - AU_TYPE_SET_LEN
  - AU_OBJECT_CLASS_NAME


src/object/authenticate.c
```diff
@@ -4384,43 +3867,40 @@ check_grant_option (MOP classop, SM_CLASS * sm_class, DB_AUTH type)
...
 	  return er_errid ();
 	}
-      cache_bits = cache->data[Au_cache_index];
+      cache_bits = get_cache_bits (sm_class);
     }
 
   /* build the complete bit mask */
   mask = type | (type << AU_GRANT_SHIFT);
-  if ((cache_bits & mask) != mask)
+  if ((*cache_bits & mask) != mask)
```
beyondykk9, line 3909: no need to check NULL of cache_bits?

src/object/authenticate.c
```diff
@@ -4107,7 +3592,7 @@ update_cache (MOP classop, SM_CLASS * sm_class, AU_CLASS_CACHE * cache)
 
   need_pop_er_stack = true;
 
-  bits = &cache->data[Au_cache_index];
+  bits = get_cache_bits (sm_class);
 
   /* initialize the cache bits */
   *bits = AU_NO_AUTHORIZATION;
```
beyondykk9, line 3603: no need to check NULL of bits?

src/object/authenticate.c
```diff
@@ -6420,37 +5900,31 @@ check_authorization (MOP classobj, SM_CLASS * sm_class, DB_AUTH type)
...
+	  if (*bits == AU_CACHE_INVALID)
 	    {
 	      /* update the cache and try again */
 	      error = update_cache (classobj, sm_class, cache);
 	      if (error == NO_ERROR)
 		{
-		  bits = cache->data[Au_cache_index];
-		  if ((bits & type) != type)
+		  bits = get_cache_bits (sm_class);
+		  if ((*bits & type) != type)
```
beyondykk9, line 5930: no need to check NULL of bits?

hgryoo:
@beyondykk9 In commit fac7df6, I implemented null checking for the `get_cache_bits()` function. It's expected that the cache_bits are obtained from the authentication cache the first time `get_cache_bits()` is called. This update introduces an assertion check for the subsequent calls to `get_cache_bits()`. If, during these calls, the cache bits are found to be NULL, this is considered an unexpected behavior.

beyondykk9 (approved):

hyunikn (approved):

ctshim (approved):

src/object/authenticate.c
```diff
@@ -4107,7 +3592,12 @@ update_cache (MOP classop, SM_CLASS * sm_class, AU_CLASS_CACHE * cache)
 
   need_pop_er_stack = true;
 
-  bits = &cache->data[Au_cache_index];
+  bits = get_cache_bits (sm_class);
```
hyunikn, line 3595: By this update, the third argument,  AU_CLASS_CACHE * cache, of update_cache() can be removed because it is unused in the function.
Note that he two call sites of update_cache(), one from check_authorization() and the other from au_get_class_privilege(), send uninitialized garbage values to this cache argument.

hgryoo: Thank you. I've removed the redundant arg at 6820282.

PR #5051, opened 2024-03-22
[CBRD-25264] include SP body ending label to the PL/CSQL text

hyunikn:
http://jira.cubrid.org/browse/CBRD-25264

commit e3e4e5f is the actual update and the previous one is a refactoring.

### DIFF
--- a/src/parser/csql_grammar.y
+++ b/src/parser/csql_grammar.y
@@ -12486,17 +12486,26 @@ opt_lang_plcsql
         ;
 
 pl_language_spec
-	: opt_lang_plcsql plcsql_text
+	: opt_lang_plcsql plcsql_text opt_identifier
 		{{ DBG_TRACE_GRAMMAR(pl_language_spec, : opt_lang_plcsql plcsql_text);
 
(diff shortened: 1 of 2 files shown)


src/parser/csql_grammar.y
```diff
@@ -12480,37 +12480,32 @@ is_or_as
...
-		{{ DBG_TRACE_GRAMMAR(pl_language_spec, | LANGAUGE PLCSQL );
-
-			PT_NODE *node = parser_new_node (this_parser, PT_SP_BODY);
-
-			if (node)
-			  {
                             assert(g_plcsql_text != NULL);
 
+                            if ($3) {
+                                len = $2 + strlen($3->info.name.original);
```
ctshim, line 12501: Shouldn't we add 1 for space characters?

hyunikn: The lexer rule is updated this time to include one or more space characters after the END keyword.

hgryoo (approved):

beyondykk9 (approved):

PR #5052, opened 2024-03-22
[CBRD-25276] fix several cost formulas derived from join order benchmark

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-25276

fix several cost formula
- Change the minimum number of tables for performing partial search from 4 to 8.
- Choose a better NDV between heap and index.
- add heap_access cpu cost for index scan.


src/optimizer/query_planner.c
```diff
@@ -8391,7 +8393,7 @@ qo_clean_planner (QO_PLANNER * planner)
  * -------------------------------------------
  * Tables joined | Tables considered at a time
  * --------------+----------------------------
- *  4..25        | 4
+ *  8..25        | 4
```
shparkcubrid:
```suggestion
 *  8..25        | 8
```

youngjinj (approved):

xmilex-git (approved):

PR #5072, opened 2024-03-29
[CBRD-25286] CTE result cache at PREPARE query and DDL/DML query

beyondykk9:
http://jira.cubrid.org/browse/CBRD-25286

1) When using the result-cache of a CTE query through the PREPARE statement, additional session information transmitted to the server is required.
2) When using the WITH CTE ... SELECT statement in DDL/DML, logic different from the process of handling the SELECT query is required.

For example, the below SQL statment should be executed successfully in csql on "demodb"

```
prepare st from
'
with ctea as (select /*+ query_cache */ code, name from public.nation where code like ''A%''),
     cteb as (select /*+ query_cache */ code, name from public.nation where code like ''B%''),
     ctec as (select /*+ query_cache */ code, name from public.nation where code like ''C%''),
     cted as (select /*+ query_cache */ code, name from public.nation where code like ''D%''),
     ctee as (select /*+ query_cache */ code, name from public.nation where code like ''E%'')
select host_year, name from public.game g, ctea where ctea.code = g.nation_code and host_year > ?
union
select host_year, name from public.game g, cteb where cteb.code = g.nation_code and host_year > ?
';

execute st using 1996, 1996;
```


shparkcubrid (approved):

src/query/query_executor.c
```diff
@@ -16070,11 +16070,17 @@ qexec_execute_cte (THREAD_ENTRY * thread_p, XASL_NODE * xasl, XASL_STATE * xasl_
 
 	  if (ent == NULL || list_cache_entry_p == NULL || list_id == NULL)
```
ctshim: trivial,
We can get the same result by just checking `list_id`.
```suggestion
	  if (list_id == NULL)
```

src/compat/db_query.c
```diff
@@ -600,6 +633,38 @@ db_unpack_prepare_info (DB_PREPARE_INFO * info, char *buffer)
...
+	{
+	  er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1,
+		  cte_num * sizeof (DB_PREPARE_CTE_INFO));
+	  goto error;
+	}
+      for (q = 0; q < cte_num; q++)
+	{
+	  OR_UNPACK_XASL_ID (ptr, &info->cte_info[q].cte_xasl_id);
+	  ptr = or_unpack_int (ptr, &info->cte_info[q].cte_host_var_count);
+	  info->cte_info[q].cte_host_var_index = (int *) malloc (info->cte_info[q].cte_host_var_count * sizeof (int));
```
ctshim: Isn't there any possibility that `cte_host_var_count` = 0?

src/compat/db_vdb.c
```diff
@@ -2308,6 +2364,86 @@ char_array_to_name_list (PARSER_CONTEXT * parser, char **names, int length)
...
+    {
+      er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1,
+	      sizeof (int) * cte_query->cte_host_var_count);
+      goto err_exit;
+    }
+
+  for (i = 0; i < cte_query->cte_host_var_count; i++)
+    {
+      cte_info[q].cte_host_var_index[i] = cte_query->cte_host_var_index[i];
+    }
```
ctshim:
```suggestion
  if (cte_query->cte_host_var_count > 0)
    {
      memcpy(cte_info[q].cte_host_var_index, cte_query->cte_host_var_index, cte_query->cte_host_var_count * sizeof(int));
    }
```

src/parser/xasl_generation.c
```diff
@@ -21596,7 +21596,8 @@ pt_to_update_xasl (PARSER_CONTEXT * parser, PT_NODE * statement, PT_NODE ** non_
     }
 
 cleanup:
-  if (aptr_statement != NULL)
+  /* postphoned freeing for CTE query execution */
+  if (aptr_statement != NULL && !pt_is_allowed_result_cache ())
```
ctshim, line 21600: If `aptr_statement != NULL` and `(!pt_is_allowed_result_cache ()) == false`, where do you free `aptr_statement`?

src/query/execute_statement.c
```diff
@@ -14532,118 +14491,155 @@ do_prepare_session_statement (PARSER_CONTEXT * parser, PT_NODE * statement)
...
+    }
 
-	  cte_context = *parser;
-	  cte_context.dbval_cnt = 0;
-	  cte_context.host_var_count = cte_context.auto_param_count = 0;
+  /* to ininitialize empty index */
+  for (i = 0; i < var_count; i++)
+    {
+      cte_context.cte_host_var_index[i] = -1;
+    }
```
ctshim:
```suggestion
  memset(cte_context.cte_host_var_index, 0xFF,  var_count * sizeof(int));
```

src/query/execute_statement.c
```diff
@@ -14532,118 +14491,155 @@ do_prepare_session_statement (PARSER_CONTEXT * parser, PT_NODE * statement)
...
+
+  cte_context = *parser;
+
+  cte_context.host_var_count = 0;
+  var_count = parser->host_var_count + parser->auto_param_count;
 
-  for (cte_def_list = statement; cte_def_list; cte_def_list = cte_def_list->next)
+  stmt->info.query.flag.cte_query_cached = 1;
+
+  cte_context.cte_host_var_index = (int *) parser_alloc (parser, var_count * sizeof (int));
```
ctshim: Is there any possibility that `var_count` = 0?

src/compat/db_query.c
beyondykk9: good point. thanks

src/query/query_executor.c
beyondykk9: good point. thanks.

src/compat/db_vdb.c
beyondykk9: good point. thanks.

src/query/execute_statement.c
beyondykk9: good point. thanks.

src/query/execute_statement.c
beyondykk9: good point. thanks.

src/parser/xasl_generation.c
beyondykk9: maybe, it would be free at the end of compiling automatically

src/query/execute_statement.c
```diff
@@ -14217,93 +14248,24 @@ static PT_NODE *
 pt_cte_host_vars_index (PARSER_CONTEXT * parser, PT_NODE * node, void *arg, int *continue_walk)
 {
   PARSER_CONTEXT *query = (PARSER_CONTEXT *) arg;
```
ctshim: In the modified code, `query `is no longer used.

src/query/execute_statement.c
```diff
@@ -14532,118 +14491,155 @@ do_prepare_session_statement (PARSER_CONTEXT * parser, PT_NODE * statement)
...
+	  return ER_OUT_OF_VIRTUAL_MEMORY;
+	}
+      /* to ininitialize empty index */
+      memset ((void *) cte_context.cte_host_var_index, -1, var_count * sizeof (int));
+    }
 
-	  cte_context = *parser;
-	  cte_context.dbval_cnt = 0;
-	  cte_context.host_var_count = cte_context.auto_param_count = 0;
+  parser_walk_tree (&cte_context, stmt, pt_cte_host_vars_index, parser, NULL, NULL);
```
ctshim: `pt_cte_host_vars_index() `has been changed so that the argument is no longer needed.
```suggestion
  parser_walk_tree (&cte_context, stmt, pt_cte_host_vars_index, NULL, NULL, NULL);
```

src/query/execute_statement.c
```diff
@@ -14532,118 +14491,155 @@ do_prepare_session_statement (PARSER_CONTEXT * parser, PT_NODE * statement)
...
 {
-  int error = NO_ERROR;
-  PT_NODE *cte_statement;
-  PT_NODE *cte_def_list;
+  int err = NO_ERROR;
   PARSER_CONTEXT cte_context;
-  int var_count;
+  int i, var_count;
+
+  cte_context = *parser;
```
ctshim, line 14502: Rather than referencing an existing `parser`, a new `PARSER_CONTEXT `object is being created.
Shouldn't we clean up the memory allocated to this object?

src/query/execute_statement.c
beyondykk9: good point, thanks.

src/query/execute_statement.c
beyondykk9: It's just initialization by existing parser context.

src/query/execute_statement.c
beyondykk9: good point. thanks.

beyondykk9: good point. thanks.

src/query/execute_statement.c
beyondykk9: Since most of the context in the parser is not used and COMPILE_CONTEXT in the parser is newly allocated and used, there does not seem to be a problem.
And since the CTE query is first prepared and cached in xasl before the main query is processed, there seems to be no problem with the main query processing thereafter.

src/compat/db_vdb.c
```diff
@@ -2308,6 +2364,87 @@ char_array_to_name_list (PARSER_CONTEXT * parser, char **names, int length)
   return list;
 }
 
+static DB_PREPARE_CTE_INFO *
+set_prepare_cte_info (PT_NODE * cte_query, DB_PREPARE_CTE_INFO * cte_info, int cte_num_query)
+{
+  int i, q = cte_num_query;
+  int alloc_size = sizeof (DB_PREPARE_CTE_INFO) * (cte_num_query / 4 + 1) * 4;
```
hyunikn, line 2371: alloc_size is used only the in the following if statement block

hyunikn (approved):

src/compat/db_vdb.c
beyondykk9: Ok. good point.

ctshim (approved):

hgryoo (approved):

PR #5079, opened 2024-04-03
[CBRD-25290] _db_auth's rows are not deleted when dropping an user

hgryoo:
http://jira.cubrid.org/browse/CBRD-25290

- Add au_delete_auth_of_dropping_user ()
   - delete all records that grantee is deleted user


ctshim (approved):

src/object/authenticate.c
```diff
@@ -1591,6 +1593,77 @@ au_delete_auth_of_dropping_table (const char *class_name)
   return error;
 }
 
+/*
+ * au_delete_auth_of_dropping_table - delete _db_auth records refers to the given grantee user.
```
hyunikn: table --> user

hyunikn (approved):

src/object/authenticate.c
hgryoo:
```suggestion
 * au_delete_auth_of_dropping_user - delete _db_auth records refers to the given grantee user.
```

hgryoo: Thank you

beyondykk9 (approved):

PR #5083, opened 2024-04-04
[CBRD-25289] When using the --split-schema-files option in unloaddb, the unique index must be separated from the dbname_schema_class file.

airnet73:
http://jira.cubrid.org/browse/CBRD-25289

Purpose
It is necessary to separate the Unique Index from the dbname_schema_class file.
Unloaddb can use the --split-schema-files option to separate Unique indexes, and the separated Unique indexes are recorded in dbname_schema_uk, and dbname_schema_uk must be recorded in dbname_schema_info.
Also, dbname_schema_info containing dbname_schema_uk must be able to be loaded from loaddb.

Implementation
- dbname_schema_info file
dbname_schema_user
dbname_schema_class
dbname_schema_vclass
dbname_schema_synonym
dbname_schema_serial
dbname_schema_procedure
dbname_schema_server
dbname_schema_pk
dbname_schema_fk
**dbname_schema_uk**
dbname_schema_grant
dbname_schema_vclass_query_spec

Remarks
N/A

### DIFF
--- a/src/executables/unload_schema.c
+++ b/src/executables/unload_schema.c
@@ -2400,7 +2400,10 @@ emit_instance_attributes (extract_context & ctxt, print_output & output_ctx, DB_
 
   if (reverse_unique_flag)
     {
-      emit_reverse_unique_def (ctxt, output_ctx, class_);
+      if (split_schema_files == false)
+	{
+	  emit_reverse_unique_def (ctxt, output_ctx, class_);
+	}
     }
 
   return true;
@@ -5514,6 +5517,7 @@ emit_unique_key (extract_context & ctxt, print_output & output_ctx, DB_OBJLIST *
 {
   DB_OBJLIST *cl = NULL;
   int is_vclass = 0;
+  int reverse_unique_flag = 0;
   const char *class_type = NULL;
   const char *name = NULL;
   int is_partitioned = 0;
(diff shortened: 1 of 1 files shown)


mhoh3963:
@airnet73 
It seems that when using --split-schema-file, the reverse unique index should also be written to the schema_uk file.
please check emit_reverse_unique_def().

src/executables/unload_schema.c
```diff
@@ -5437,6 +5509,58 @@ emit_grant (extract_context & ctxt, print_output & output_ctx, DB_OBJLIST * clas
...
+  const char *name = NULL;
+  int is_partitioned = 0;
+  int unique_flag = 0;
+  DB_ATTRIBUTE *attribute_list = NULL;
+  DB_ATTRIBUTE *a = NULL;
+
+  for (cl = ctxt.classes; cl != NULL; cl = cl->next)
+    {
+      is_vclass = db_is_vclass (cl->op);
+      class_type = (is_vclass > 0) ? "VCLASS" : "CLASS";
```
mhoh3963: The vclass can't have unique index, so how about skipping if the class is vclass ?

airnet73: I will skip VCLASS.

kisoo-han (approved):

src/executables/unload_schema.c
airnet73: fixed it

mhoh3963 (approved):

PR #5100, opened 2024-04-09
[CBRD-25268] Refactor authenticate: extract method related code

hgryoo:
http://jira.cubrid.org/browse/CBRD-25268

Static method links are not part of `authenticate` module. I've extracted `XXX_method()` functions to a new file (`db_method_static.cpp`) and make it only visible to `method` module.

note that chaning owner routines are refactored not to use XXX_method() directly.


hgryoo:
@shparkcubrid qo_set_cost() method is a backdoor function for QA testing. By this PR's change, the function is just moved to a new file from query_planner.c. The `qo_set_cost()` function is not called anywhere except the `method` module.

hyunikn (approved):

src/query/execute_statement.c
```diff
@@ -6688,19 +6688,17 @@ do_alter_trigger (PARSER_CONTEXT * parser, PT_NODE * statement)
...
-		  db_make_string (&user_val, trigger_owner_name);
-
-		  au_change_trigger_owner_method (t->op, &returnval, &trigger_name_val, &user_val);
-
-		  pr_clear_value (&trigger_name_val);
-		  pr_clear_value (&user_val);
+		  owner_mop = au_find_user (trigger_owner_name);
+		  if (owner_mop == NULL)
+		    {
+		      ASSERT_ERROR_AND_SET (error);
```
ctshim, line 6694: When is the `error` variable covered here set?

ctshim (approved):

beyondykk9 (approved):

shparkcubrid:
I disagree with fixing the source code a lot in this way without adding features. This can be confusing when checking source code history. I think that since each developer has a different code style, it is impossible to know which one is better. 
I approve of this PR, but please take my opinion into consideration later.

shparkcubrid (approved):

hgryoo:
> I disagree with fixing the source code a lot in this way without adding features. This can be confusing when checking source code history. I think that since each developer has a different code style, it is impossible to know which one is better. I approve of this PR, but please take my opinion into consideration later.

  I understand your cautious approach towards refactoring modules that are interconnected with numerous other modules.
  However, I believe you are aware that CUBRID's code has aged without proper refactoring, which has degraded its readability and efficiency. I believe that having multiple separate modules implemented in a single file is worse in terms of complexity than having them implemented separately by each developer with different coding styles and contents.
  Furthermore, there is a principle in code refactoring: "Never refactor and implement new features at the same time." Doing both simultaneously can lead to decrease quality and increase risk of bugs. Therefore, it is generally recommended to focus on refactoring and improving the code separately from adding new features.
  For the refactoring of modules that are complexly connected with others, I will take your opinion into careful consideration. Thank you for your advice.

PR #5104, opened 2024-04-09
[CBRD-25298] trim class and method names of the target method

hyunikn:
http://jira.cubrid.org/browse/CBRD-25298

. trim the class name and the method name before search for the method by reflection
. use familiar type names of parameter types in the error messages of NoSuchMethod

### DIFF
--- a/src/jsp/com/cubrid/jsp/TargetMethod.java
+++ b/src/jsp/com/cubrid/jsp/TargetMethod.java
@@ -38,7 +38,10 @@ import java.math.BigDecimal;
 import java.sql.Date;
 import java.sql.Time;
 import java.sql.Timestamp;
+import java.util.Arrays;
 import java.util.HashMap;
+import java.util.List;
+import java.util.stream.Collectors;
 
 public class TargetMethod {
     private String className;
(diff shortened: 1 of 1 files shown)


src/jsp/com/cubrid/jsp/TargetMethod.java
```diff
@@ -201,14 +204,26 @@ private static void initdescriptorMap() {
...
-            throw new ClassNotFoundException (className);
+            throw new ClassNotFoundException(className);
+        }
+        try {
+            return c.getMethod(methodName, argsTypes);
+        } catch (NoSuchMethodException e) {
+            List<String> argTypeNameList =
+                    Arrays.stream(argsTypes)
+                            .map(cl -> cl.getTypeName())
+                            .collect(Collectors.toList());
```
hgryoo: trivial: please note that target JDK version of CUBRID 10.2 and 11.0 is 6. Arrays.stream() doesn't support on JDK 6. some additional works needs.

hgryoo (approved):

src/jsp/com/cubrid/jsp/TargetMethod.java
hyunikn: OK. Thank you.

ctshim (approved):

PR #5132, opened 2024-04-13
[CBRD-25304] Remove the obsolete code line

joohok:
http://jira.cubrid.org/browse/CBRD-25304

Purpose

The system parameters(PRM_ID_MUTEX_BUSY_WAITING_CNT) have become obsolete, but the related code has not been removed


src/storage/page_buffer.c
```diff
@@ -6913,11 +6910,6 @@ pgbuf_search_hash_chain (THREAD_ENTRY * thread_p, PGBUF_BUFFER_HASH * hash_ancho
 		  goto two_phase;
 		}
 
-	      if (loop_cnt++ < mbw_cnt)
```
hornetmj, line 6916: The following things can be romoved:
- int loop_cnt;
- mutex_lock:
- mutex_lock2:

hornetmj (approved):

src/storage/page_buffer.c
joohok: OK, I removed it

PR #5133, opened 2024-04-17
[CBRD-25297] building Java type import list at code generation stage

hyunikn:
http://jira.cubrid.org/browse/CBRD-25297

. building Java type import list in earlier stages of compilation has been the cause of omitting required Java type imports by mistake. It should be near the Java code generated.
. a small update irrelevant to this issue is applied: semantic error "not supported yet" on autonomous transaction pragma.

### DIFF
--- a/pl_engine/pl_server/src/main/java/com/cubrid/plcsql/compiler/visitor/JavaCodeWriter.java
+++ b/pl_engine/pl_server/src/main/java/com/cubrid/plcsql/compiler/visitor/JavaCodeWriter.java
@@ -46,12 +46,12 @@ public class JavaCodeWriter extends AstVisitor<JavaCodeWriter.CodeToResolve> {
 
     private Set<String> javaTypesUsed = new HashSet<>();
 
-    private void addToImports(String fullJavaType) {
+    private void addToImports(String fullJavaType) {}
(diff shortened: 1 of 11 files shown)


pl_engine/pl_server/src/main/java/com/cubrid/plcsql/compiler/visitor/JavaCodeWriter.java
```diff
@@ -44,11 +44,28 @@
 
 public class JavaCodeWriter extends AstVisitor<JavaCodeWriter.CodeToResolve> {
 
+    private Set<String> javaTypesUsed = new HashSet<>();
+
+    private void addToImports(String fullJavaType) {}
```
beyondykk9: It's just dummy method? It looks better to remove the method if it would be used any more.

hyunikn: It is unused code and should be removed.
Thank you.

ctshim (approved):

beyondykk9 (approved):

hgryoo (approved):

PR #5146, opened 2024-04-25
[CBRD-25313] Get HFID from class-name using caches and dump with HFID

YeunjunLee:
http://jira.cubrid.org/browse/CBRD-25313

Retrieve the HFID of a class from class-name and dump the class heap file

- Convert `class-name` to `class oid` with `locator_Mht_classnames`cache  using `xlocator_find_class_oid` function.
- Convert `class oid` to `HFID` with `heap_Hfid_table` cache using `heap_hfid_cache_get` function.
- Dump corresponding heap file by retrieved `HFID` using `heap_dump` function.

### DIFF
--- a/src/storage/heap_file.c
+++ b/src/storage/heap_file.c
@@ -14534,10 +14534,11 @@ heap_dump_heap_file (THREAD_ENTRY * thread_p, FILE * fp, bool dump_records, cons
   error_code = heap_hfid_cache_get (thread_p, &class_oid, &hfid, NULL, NULL);
   if (error_code != NO_ERROR)
     {
+      ASSERT_ERROR ();
       return;
     }
-  heap_dump (thread_p, fp, &hfid, dump_records);
 
+  heap_dump (thread_p, fp, &hfid, dump_records);
 }
 #endif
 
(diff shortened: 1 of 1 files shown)


src/storage/heap_file.c
```diff
@@ -14520,7 +14520,24 @@ heap_dump (THREAD_ENTRY * thread_p, FILE * fp, HFID * hfid, bool dump_records)
...
-  /* TODO: Fetch HFID of class which corresponds to class_name, and dump heap file by HFID. Will be handled at CBRD-25313 and CBRD-25314. */
+  int error_code = NO_ERROR;
+  OID class_oid;
+  LC_FIND_CLASSNAME status;
+  HFID hfid;
+
+  status = xlocator_find_class_oid (thread_p, class_name, &class_oid, S_LOCK);
+  if (status != LC_CLASSNAME_EXIST)
+    {
+      return;
```
joohok, line 14531: We need to handle the exceptional cases. 
1. We can define error code (`error_code.h`), set error code here (`er_set`), and then return the error code to the caller
2. Or, we can write the error message into the file (at fp)

hornetmj (approved):

src/storage/heap_file.c
YeunjunLee: When the class which has `class_name` doesn't exists, it was decided to print any message (Only print Head and Tail message of dump). And it was due to coherence with other utility functions.

However, for error_code for `heap_hfid_cache_get` function, we can call assertion to check HFID exists.

joohok: OK, I agree

joohok (approved):

Rudeus (approved):

PR #5150, opened 2024-04-26
[CBRD-25310] Add conditions in memory_monitor::sub_stat() for preventing noise error

Rudeus:
http://jira.cubrid.org/browse/CBRD-25068

Fix regression test error:
 - Add conditions in memory_monitor::sub_stat() for preventing noise error in regression test

### DIFF
--- a/src/base/memory_monitor_sr.hpp
+++ b/src/base/memory_monitor_sr.hpp
@@ -176,11 +176,10 @@ retry:
       {
 	MMON_METAINFO *metainfo = (MMON_METAINFO *) get_metainfo_pos (ptr, allocated_size);
 
-	if (metainfo->magic_number == m_magic_number)
+	if (metainfo->magic_number == m_magic_number && metainfo->allocated_size == allocated_size
+	    && metainfo->stat_id < MMON_MAP_RESERVE_SIZE && metainfo->stat_id >= 0)
 	  {
-	    assert (metainfo->stat_id >= 0 && metainfo->stat_id < MMON_MAP_RESERVE_SIZE);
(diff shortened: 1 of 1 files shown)


src/base/memory_monitor_sr.hpp
```diff
@@ -176,11 +176,10 @@ namespace cubmem
       {
 	MMON_METAINFO *metainfo = (MMON_METAINFO *) get_metainfo_pos (ptr, allocated_size);
 
-	if (metainfo->magic_number == m_magic_number)
+	if (metainfo->magic_number == m_magic_number && metainfo->allocated_size == allocated_size
+	    && metainfo->stat_id < MMON_MAP_RESERVE_SIZE && metainfo->stat_id >= 0)
```
joohok, line 180: Nit. 
I can't understand what this condition is for just by looking at the code. How about giving some meaning to the conditions ? 
```suggestion
        const bool meaningful_name = metainfo->magic_number == m_magic_number 
        && metainfo->allocated_size == allocated_size
        && metainfo->stat_id < MMON_MAP_RESERVE_SIZE
        && metainfo->stat_id >= 0
        
	if (meaningful_name)
```

joohok (approved):

src/base/memory_monitor_sr.hpp
hornetmj: Same opinion with @joohok.

Basically, this two conditions, allocated_size >= MMON_METAINFO_SIZE and metainfo->magic_number == m_magic_number, guarantee that this memory is traced by mmon. The reason why tests with different conditions are needed must be explained in comments.

PR #5153, opened 2024-04-28
[CBRD-25307] Remove meaningless comments from heap_create_internal ( ... ) header.

hyahong:
http://jira.cubrid.org/browse/CBRD-25307

Purpose
the parameter `exp_npgs` was removed from old commit, so the comments of parameter `exp_npgs` is meaningless now.

Implementation
N/A

Remarks
N/A

### DIFF
--- a/src/storage/heap_file.c
+++ b/src/storage/heap_file.c
@@ -5235,7 +5235,6 @@ heap_manager_finalize (void)
  *   hfid(in/out): Object heap file identifier.
  *                 All fields in the identifier are set, except the volume
  *                 identifier which should have already been set by the caller.
- *   exp_npgs(in): Expected number of pages
  *   class_oid(in): OID of the class for which the heap will be created.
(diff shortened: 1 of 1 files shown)


src/storage/heap_file.c
```diff
@@ -5235,7 +5235,6 @@ heap_manager_finalize (void)
  *   hfid(in/out): Object heap file identifier.
  *                 All fields in the identifier are set, except the volume
  *                 identifier which should have already been set by the caller.
- *   exp_npgs(in): Expected number of pages
```
hornetmj, line 5238: How about updating the return type together? The description said it it HFID * but it is int.

hornetmj: `estimated from the number of expected pages.`

The comment should be changed together. However it is hard to you at this moment. So, think about it and share your opinion.

hyahong: I think it would be better to deal with it all at once. I will reply once i have analyzed this function.

PR #5162, opened 2024-04-30
[CBRD-25312] Print usage message when class-name is given for non heap_dump mode

YeunjunLee:

Originally, when class-name is given, `diagdb` prints usage if dump mode is `DIAGDUMP_ALL`.
For correct behavior, modify to print usage if dump mode is not `DIAGDUMP_HEAP`.


joohok:
please update your feature branch and then push again.
Because there are un-related changes included in this PR.

PR #5166, opened 2024-04-30
[CBRD-25329] truncate milli-seconds when converting DATETIME to TIMESTAMP in server side JDBC

hyunikn:
http://jira.cubrid.org/browse/CBRD-25329


ctshim (approved):

beyondykk9 (approved):

hyunikn:
I added a commit (c64142e) to suppress milli-seconds truncation of arguments to DATETIME type parameters and another commit (4a69274) that properly uses toDatetime() instead of incorrect toTimestamp() in some places.

hgryoo (approved):

PR #5181, opened 2024-05-08
[CBRD-24902] Store PL/CSQL source code in _db_stored_procedure_code catalog

hgryoo:
http://jira.cubrid.org/browse/CBRD-24902

- call sp_add_stored_procedure_code() if registering stored routine is PL/CSQL.
  - `create_time` is generated in sp_add_stored_procedure().
  - `stype`, `otype` and `ocode` have dummy value now. their value will be set properly in the next PR.
- In drop_stored_procedure (), To delete a row of PL/CSQL's code, drop_stored_procedure_code () is called.
- To fix assertion in btree_unique(), {"name"}'s constraints is changed from DB_CONSTRAINT_UNIQUE to DB_CONSTRAINT_NOT_NULL.

### DIFF
--- a/src/jsp/jsp_cl.cpp
+++ b/src/jsp/jsp_cl.cpp
@@ -828,7 +828,7 @@ jsp_create_stored_procedure (PARSER_CONTEXT *parser, PT_NODE *statement)
       code_info.stype = 0;
       code_info.scode = pl_code;
       code_info.otype = 0;
-      code_info.ocode = "dummy"; // TODO
+      code_info.ocode = "dummy"; // TODO: CBRD-24552
       code_info.owner = Au_user; // current user
 
       err = sp_add_stored_procedure_code (code_info);
--- a/src/jsp/jsp_cl.cpp
+++ b/src/jsp/jsp_cl.cpp
@@ -33,6 +33,11 @@
 
 #include <vector>
 #include <functional>
+#include <chrono>
+#include <ctime>
+#include <iomanip>
+#include <chrono>
+#include <sstream>
 
 #include "authenticate.h"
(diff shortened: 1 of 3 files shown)


src/jsp/jsp_cl.cpp
```diff
@@ -33,6 +33,11 @@
 
 #include <vector>
 #include <functional>
+#include <chrono>
+#include <ctime>
+#include <iomanip>
+#include <chrono>
```
ctshim: double include^^

ctshim (approved):

src/jsp/jsp_cl.cpp
hgryoo: Thank you :-)

```suggestion
```

src/jsp/jsp_cl.cpp
```diff
@@ -1100,23 +1130,31 @@ drop_stored_procedure (const char *name, SP_TYPE_ENUM expected_type)
...
-      goto error;
+      err = db_get (sp_mop, SP_ATTR_TARGET, &target_val);
+      if (err != NO_ERROR)
+	{
+	  goto error;
+	}
+
+      target = db_get_string (&target_val);
+      class_name = get_class_name (std::string (target));
+
```
beyondykk9, line 1151: no need to check null data for target and class_name ?

hgryoo: it is guaranteed that the target is not NULL. It is checked in syntax checking of PT_CREATE_STORED_PROCEDURE parser node.

hyunikn (approved):

beyondykk9 (approved):

PR #5183, opened 2024-05-09
[CBRD-25347] Fix tracking holes of memory monitoring

Rudeus:
http://jira.cubrid.org/browse/CBRD-25347

add free_and_init macro using cub_free() in memory_alloc.h


Rudeus:
This PR will be closed as it has been decided to maintain the existing code, unless any further issues arise from it in the future.

PR #5185, opened 2024-05-09
[CBRD-25347] Fix tracking holes of memory monitoring

Rudeus:
http://jira.cubrid.org/browse/CBRD-25347

Add behavior when the ptr argument of cub_realloc is set to NULL

### DIFF
--- a/src/base/memory_cwrapper.h
+++ b/src/base/memory_cwrapper.h
@@ -135,6 +135,11 @@ cub_realloc (void *ptr, size_t size, const char *file, const int line)
 
   if (mmon_is_memory_monitor_enabled ())
     {
+      if (ptr == NULL)
+	{
+	  return cub_alloc (size, file, line);
+	}
+
       if (size == 0)
 	{
 	  cub_free (ptr);
--- a/src/base/memory_cwrapper.h
+++ b/src/base/memory_cwrapper.h
@@ -135,11 +135,13 @@ cub_realloc (void *ptr, size_t size, const char *file, const int line)
 
   if (mmon_is_memory_monitor_enabled ())
     {
+      /* Realloc spec.: Realloc should work same as malloc if ptr == NULL */
       if (ptr == NULL)
 	{
 	  return cub_alloc (size, file, line);
 	}
 
+      /* Realloc spec.: If input size is zero, just free ptr */
       if (size == 0)
 	{
 	  cub_free (ptr);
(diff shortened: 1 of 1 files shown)


src/base/memory_cwrapper.h
```diff
@@ -145,13 +152,11 @@ cub_realloc (void *ptr, size_t size, const char *file, const int line)
...
-		  size_t old_size = get_allocated_size (ptr);
-		  size_t copy_size = old_size < size ? old_size : size;
-		  memcpy (new_ptr, ptr, copy_size);
-		  cub_free (ptr);
-		}
+
+              size_t old_size = get_allocated_size (ptr);
+              size_t copy_size = old_size < size ? old_size : size;
+              memcpy (new_ptr, ptr, copy_size);
+              cub_free (ptr);
```
Rudeus:
```suggestion
             size_t old_size = get_allocated_size (ptr);
             size_t copy_size = old_size < size ? old_size : size;
             memcpy (new_ptr, ptr, copy_size);
             cub_free (ptr);
```

joohok (approved):

src/base/memory_cwrapper.h
```diff
@@ -145,13 +152,11 @@ cub_realloc (void *ptr, size_t size, const char *file, const int line)
...
-		  size_t old_size = get_allocated_size (ptr);
-		  size_t copy_size = old_size < size ? old_size : size;
-		  memcpy (new_ptr, ptr, copy_size);
-		  cub_free (ptr);
-		}
+
+             size_t old_size = get_allocated_size (ptr);
+             size_t copy_size = old_size < size ? old_size : size;
+             memcpy (new_ptr, ptr, copy_size);
+             cub_free (ptr);
```
Rudeus:
```suggestion
	      size_t old_size = get_allocated_size (ptr);
	      size_t copy_size = old_size < size ? old_size : size;
	      memcpy (new_ptr, ptr, copy_size);
	      cub_free (ptr);
```

src/base/memory_cwrapper.h
```diff
@@ -135,6 +135,13 @@ cub_realloc (void *ptr, size_t size, const char *file, const int line)
...
   if (mmon_is_memory_monitor_enabled ())
     {
+      /* Realloc spec.: Realloc should work same as malloc if ptr == NULL */
+      if (ptr == NULL)
+	{
+	  return cub_alloc (size, file, line);
+	}
+
+      /* Realloc spec.: If input size is zero, just free ptr */
       if (size == 0)
```
hornetmj, line 145: About size == 0 case, the [manual](https://man7.org/linux/man-pages/man3/realloc.3p.html) define as the following:

> If size is 0, either:
> *  A null pointer shall be returned and, if ptr is not a null
>    pointer, [errno](https://man7.org/linux/man-pages/man3/errno.3.html) shall be set to an implementation-defined
>    value.
> *  A pointer to the allocated space shall be returned, and the
>    memory object pointed to by ptr shall be freed. The
>    application shall ensure that the pointer is not used to
>    access an object.

We must follow one or the other behavior. What specification did you refer to for size==0 case?

Rudeus: I referred to MSDN spec.
But according to [this page](https://wariua.github.io/man-pages-ko/realloc%283%29/), glibc spec is same as MSDN spec. I tested the case that `after = realloc(before, 0);` It returns Null pointer with `before` pointer freed.

hornetmj (approved):

PR #5195, opened 2024-05-13
[CBRD-25257] Include headers of wrapping functions into "all" cubrid files

Rudeus:
http://jira.cubrid.org/browse/CBRD-25257

Fix new workflow with adding false include memory_wrapper.hpp in the files not using it.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -243,6 +243,12 @@ jobs:
               fi
             fi
 
+            filename=$(basename $f)
+            check_server_file=$(grep $filename cubrid/CMakeLists.txt)
+            if [ "$check_server_file" == "" ]; then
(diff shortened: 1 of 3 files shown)


.github/workflows/check.yml
```diff
@@ -243,6 +243,12 @@ jobs:
               fi
             fi
 
+            filename=$(basename $f)
```
hornetmj, line 246: This check is more efficent just after file type checking like .c or .cpp.

Rudeus: This behavior modifies the check results for files not present in cubrid/CMakeLists.txt, converting the check result from FAIL to PASS. If we move it as you suggested, it will create a larger condition statement structure wrapping the check section. Are you suggesting this behavior to skip the check altogether?

PR #5197, opened 2024-05-14
[CBRD-25358] error handling when using for update clause for system tables such as db_root, dual, etc.

beyondykk9:
http://jira.cubrid.org/browse/CBRD-25358

No errors occur when using the "for update" clause for system tables such as db_root and dual. Semantic error handling is required.

`select /*+ recompile */ * from db_root for update; `

expected result:
ERROR: update is not autorized on db_root.

### DIFF
--- a/src/parser/name_resolution.c
+++ b/src/parser/name_resolution.c
@@ -5588,7 +5588,16 @@ pt_resolve_correlation (PARSER_CONTEXT * parser, PT_NODE * in_node, PT_NODE * sc
 	  return NULL;
 	}
 
-      corr_name = pt_name (parser, "");
+      if (PT_NAME_INFO_IS_FLAGED (in_node, PT_NAME_FOR_UPDATE))
+	{
+	  corr_name = pt_name (parser, in_node->info.name.original);
+	  PT_NAME_INFO_SET_FLAG (corr_name, PT_NAME_FOR_UPDATE);
+	}
+      else
+	{
+	  corr_name = pt_name (parser, "");
+	}
+
       PT_NODE_COPY_NUMBER_OUTERLINK (corr_name, in_node);
       in_node->next = NULL;
       in_node->or_next = NULL;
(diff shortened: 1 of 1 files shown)


src/parser/name_resolution.c
```diff
@@ -9198,6 +9216,17 @@ pt_resolve_names (PARSER_CONTEXT * parser, PT_NODE * statement, SEMANTIC_CHK_INF
...
+	      for (entity = spec->info.spec.flat_entity_list; entity; entity = entity->next)
+		{
+		  db = entity->info.name.db_object;
+		  if (db_is_system_class (db) && (db_check_authorization (db, DB_AUTH_UPDATE) != NO_ERROR))
+		    {
+		      PT_ERRORmf2 (parser, entity, MSGCAT_SET_PARSER_RUNTIME, MSGCAT_RUNTIME_IS_NOT_AUTHORIZED_ON,
+				   "UPDATE", entity->info.name.original);
+		      return NULL;
+		    }
+		  spec->info.spec.flag = (PT_SPEC_FLAG) (spec->info.spec.flag | PT_SPEC_FLAG_FOR_UPDATE_CLAUSE);
```
hyunikn, line 9228: This is done already at line 9218.

src/parser/name_resolution.c
```diff
@@ -9187,6 +9194,17 @@ pt_resolve_names (PARSER_CONTEXT * parser, PT_NODE * statement, SEMANTIC_CHK_INF
 			      node->info.name.original);
 		  return NULL;
 		}
+
+	      for (entity = spec->info.spec.flat_entity_list; entity; entity = entity->next)
+		{
+		  db = entity->info.name.db_object;
+		  if (db_is_system_class (db) && (db_check_authorization (db, DB_AUTH_UPDATE) != NO_ERROR))
```
ctshim, line 9201: There is the same code on line 9222.
Do both conditions have to be met? Isn't this an OR condition?

beyondykk9: This routine is for "FOR UPDATE OF class-name-list" and the below routine is for "FOR UPDATE" without "OF" clause.

src/parser/name_resolution.c
beyondykk9: This routine is for "FOR UPDATE" without "OF" clause and the above routine is for "FOR UPDATE OF class-name-list" clause.

hgryoo (approved):

hyunikn (approved):

jongmin-won (approved):

ctshim (approved):

PR #5201, opened 2024-05-14
[CBRD-25360] An issue where the serial value is not unloaded when auto_increment exists in the table.

airnet73:
http://jira.cubrid.org/browse/CBRD-25360

Purpose
If auto_increment exists in the table, the corresponding serial value is not unloaded when unloading, and an error occurs when using the DB after reconstructing the DB using loaddb for data without a serial value.

Implementation
- Add the code below to the schema file
ALTER SERIAL <table_name>_ai_<column_name> START WITH <current_value>;

Remarks


kisoo-han (approved):
It seems that there is extra code for the functionality required, but it's ok.
LGTM!

beyondykk9 (approved):

PR #5202, opened 2024-05-15
[CBRD-25359] uhosts version of freeaddrinfo () to release memory explicitly

kisoo-han:
http://jira.cubrid.org/browse/CBRD-25359

**Description**
* because **uhosts** alloc memory using malloc () and free using freeaddrinfo (), CUBRID memory utility cannot catch memory release point inside freeaddrinfo (3) glibc library. 
* we will introduce new function, **freeaddrinfo_uhost** (), and in case uhosts enabled, we will free memory allocated by getaddrinfo () externally using free (3).

### DIFF
--- a/src/connection/host_lookup.c
+++ b/src/connection/host_lookup.c
@@ -64,8 +64,6 @@
           }                     \
         } while (0)
 
-extern void KSHAN (const char *, ...);
-
 typedef enum
 {
   HOSTNAME_TO_IPADDR = 0,
--- a/src/connection/host_lookup.c
+++ b/src/connection/host_lookup.c
@@ -56,6 +56,9 @@
 #define MAX_NUM_IPADDR_PER_HOST      (1)
 
 #define NUM_DIGIT(VAL)              (size_t)(log10 (VAL) + 1)
+
+#if !defined (MALLOC)
+#define MALLOC(SIZE)            malloc(SIZE)
 #define FREE_MEM(PTR)           \
         do {                    \
           if (PTR) {            \
@@ -63,6 +66,7 @@
             PTR = 0;            \
           }                     \
         } while (0)
+#endif
 
 typedef enum
 {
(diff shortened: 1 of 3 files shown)


src/connection/host_lookup.c
```diff
@@ -686,6 +686,22 @@ getaddrinfo_uhost (char *node, char *service, struct addrinfo *hints, struct add
...
+freeaddrinfo_uhost (struct addrinfo *res)
+{
+  if (prm_get_bool_value (PRM_ID_USE_USER_HOSTS) == USE_GLIBC_HOSTS)
+    {
+      return (freeaddrinfo (res));
+    }
+
+  if (res)
+    {
+      FREE_MEM (res->ai_canonname);
```
hornetmj, line 703: It is good to me to replace malloc to MALLOC() macro if you want to use FREE_MEM() macro. Or how about use free()?

kisoo-han: I think it's a good idea to use MALLOC and FREE_MEM as a pair in point of memory managements.
fixed.

Rudeus:
Since this issue occurs in a situation where memory monitoring project is applied, I believe it is appropriate to change the target branch for merging from the `develop` branch to the `feature/memory_monitoring_ver2_merge` branch.

hornetmj:
> Since this issue occurs in a situation where memory monitoring project is applied, I believe it is appropriate to change the target branch for merging from the `develop` branch to the `feature/memory_monitoring_ver2_merge` branch.

@kisoo-han To explain further, we are performing regression testing in {{memory_monitoring_ver2_merge}} branch. We requested this because we thought it would be better if the changes made in this PR were verified together. I feel like the request was made without sufficient reason, so I would like to explain further. thank you for the quick response. Also, I apologize for the sudden change request.

kisoo-han:
This issue will be moved to #5211  due to base branch problem.

PR #5206, opened 2024-05-17
[CBRD-25366] Display execution plan in plandump utility and add function to specific plan cache deletion.

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-25366

Display execution plan in plandump utility and add function to specific plan cache deletion.
```
# cubrid plandump demodb
...
Entries:
  XASL_ID = {
              sha1 = { 6290f2b1 44f088e5 d37d6f0f c8303155 9ef2096a },
                  time_stored = 1715925895 sec, 58124 usec
            }
...
    sql plan text =
 Sequential scan(public.game dba.game)
...

# cubrid plandump -s '6290f2b1 44f088e5 d37d6f0f c8303155 9ef2096a' demodb
...
/* delete specific plan cache and display current plan caches. */
...
```


xmilex-git (approved):

src/executables/utility.h
```diff
@@ -1241,6 +1241,8 @@ typedef struct _ha_config
 #define PLANDUMP_DROP_L                         "drop"
 #define PLANDUMP_OUTPUT_FILE_S		        'o'
 #define PLANDUMP_OUTPUT_FILE_L                  "output-file"
+#define PLANDUMP_SHA1_S		        	's'
+#define PLANDUMP_SHA1_L                  	"sha1"
```
youngjinj, line 1245: The '-s' option can also be used to view the plan for a given SHA1.
How about changing the option name?

I think it is better to use '--drop-sha1' without the short option name if we think it will not be used often.

youngjinj (approved):

src/executables/utility.h
shparkcubrid: Good point. But I don't want to do it because I have to edit the entire message.

Hamkua (approved):

beyondykk9 (approved):

PR #5217, opened 2024-05-20
[CBRD-25348] Add alloc/free counter with memory peak check

Rudeus:
http://jira.cubrid.org/browse/CBRD-25348

 - Add alloc/free counter
 - Add memory peak check
 - Add output function which is called when server shutdown (`mmon_finalize ()`)

### DIFF
--- a/src/base/memory_monitor_sr.cpp
+++ b/src/base/memory_monitor_sr.cpp
@@ -25,6 +25,7 @@
 #include <cstring>
 #include <algorithm>
 
+#include "environment_variable.h"
 #include "memory_monitor_sr.hpp"
 
 namespace cubmem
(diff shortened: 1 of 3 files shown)


src/base/memory_monitor_sr.cpp
```diff
@@ -251,5 +259,68 @@ namespace cubmem
...
+    fprintf (outfile_fp, "\t%-50s | %10s | %10s | %10s\n", "File Name", "Alloc Cnt", "Free Cnt", "Peak Mem");
+
+    int cnt = 0;
+    for (const auto &[stat_name, mem_usage] : server_info.stat_info)
+      {
+	if (server_info.total_mem_usage != 0)
+	  {
+	    mem_usage_ratio = mem_usage / (double) server_info.total_mem_usage;
+	    mem_usage_ratio *= 100;
+	  }
```
```suggestion
```

src/base/memory_monitor_sr.hpp
```diff
@@ -37,7 +37,7 @@
  * 2: debug tracking error
  * 3: debug all
  */
-#define MMON_DEBUG_LEVEL 0
+#define MMON_DEBUG_LEVEL 1
```
hornetmj, line 40: Revert to 0 after getting result of our performace test.

hornetmj (approved):

src/base/memory_monitor_sr.cpp
```diff
@@ -251,5 +259,63 @@ namespace cubmem
     fflush (outfile_fp);
     fclose (outfile_fp);
   }
+
+#if (MMON_DEBUG_LEVEL == 1) || (MMON_DEBUG_LEVEL == 3)
+  void memory_monitor::print_debug_result ()
+  {
+    double mem_usage_ratio = 0.0;
+    FILE *outfile_fp = fopen ("mmon_debug_result.txt", "w+");
```
joohok: Is this intended to be used temporarily? If not, wouldn't it be better to place this file in $CUBRID/log/server/?

joohok: We need to handle the case where the outfile_fp value is not properly obtained due to an fopen error. For example, it could be set to stderr.

src/base/memory_monitor_sr.hpp
joohok: Can we set the default to 0 and change it at compile time using the -D option?

Rudeus:
https://github.com/CUBRID/cubrid/pull/5217#discussion_r1607507804
Currently, we have not found an appropriate solution.

src/base/memory_monitor_sr.cpp
Rudeus: done.

joohok (approved):

PR #5218, opened 2024-05-20
[CBRD-25369] add new broker parameter NET_BUF_SIZE

kisoo-han:
http://jira.cubrid.org/browse/CBRD-25369

**Description**
* extend cubrid_broker.conf for NET_BUF_SIZE
  * 16K (default)
  * 32K
  * 48K
  * 64K

### DIFF
--- a/src/broker/cas_net_buf.h
+++ b/src/broker/cas_net_buf.h
@@ -69,7 +69,7 @@
 
 #define NET_BUF_KBYTE                   1024
 #define SHARD_NET_BUF_SIZE              (512 * NET_BUF_KBYTE)
-#define NET_BUF_SIZE                    (16 * NET_BUF_KBYTE)
+#define NET_BUF_SIZE                    (get_net_buf_size ())
 #define NET_BUF_EXTRA_SIZE              (64 * NET_BUF_KBYTE)
 #define NET_BUF_ALLOC_SIZE              (NET_BUF_SIZE + NET_BUF_EXTRA_SIZE)
 #define SHARD_NET_BUF_ALLOC_SIZE        (SHARD_NET_BUF_SIZE + NET_BUF_EXTRA_SIZE)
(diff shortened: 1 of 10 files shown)


src/broker/cas_net_buf.h
```diff
@@ -88,6 +88,8 @@
 #define NET_BUF_TYPE_SIZE(net_buf) (DOES_CLIENT_UNDERSTAND_THE_PROTOCOL \
 ((net_buf)->client_version, PROTOCOL_V7) ? 2 * NET_SIZE_BYTE : NET_SIZE_BYTE)
 
+#define INT_DEFAULE_NET_BUF_SIZE        (16 * 1024)
```
mhoh3963: How about changing NET_BUF_KBYTE to 1024?

kisoo-han: For the readability, it is good to change 1024 to NET_BUF_KBYTE.
Changed.

src/broker/shard_proxy.c
```diff
@@ -42,6 +42,7 @@
 /* SHARD SHM */
 int appl_server_shm_id = -1;
 T_SHM_APPL_SERVER *shm_as_p = NULL;
+T_SHM_APPL_SERVER *shm_appl = NULL;
```
mhoh3963, line 45: Why was shm_appl created additionally instead of using shm_as_p?
In your modified code, the shm_appl variable is only used in the set_net_buf_size() function.

kisoo-han: cas & cascgw uses **shm_appl** for shared memory reference, it is defined in cas.c.
I try to change/set netbuf size in car_net_buf.c using shared memory value **shm_appl->net_buf_size**,
it will be set in broker_config ().

However, shard proxy defined **shm_as_p** instead of **shm_appl**, moreover, it do not uses NET_BUF_SIZE.
If shm_appl is not defined, a **syntax error occurs when compiling shard proxy**.
If net_buf_init () or set_net_buf_size () in shard proxy, value of shm_appl will be NULL and **_net_buf_size** will be set to 16K.

mhoh3963: OK, I got it.

mhoh3963 (approved):

airnet73 (approved):

PR #5220, opened 2024-05-21
[CBRD-25007] bugfix in Java code generation of long string literals

hyunikn:
http://jira.cubrid.org/browse/CBRD-25007

(1) change the string literal cut length from 65535 to 16383
(2) keep the original (unescaped) version of the string literal until Java code generation

### DIFF
--- a/pl_engine/pl_server/src/main/java/com/cubrid/plcsql/compiler/ast/ExprStr.java
+++ b/pl_engine/pl_server/src/main/java/com/cubrid/plcsql/compiler/ast/ExprStr.java
@@ -81,5 +81,4 @@ public class ExprStr extends Expr {
     //   65534: maximum byte length of a Java string literal,
     //   4: maximum byte length of a character in UTF-8
     private static final int STR_LITERAL_CUT_LEN = 16383;
-
 }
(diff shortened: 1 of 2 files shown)


pl_engine/pl_server/src/main/java/com/cubrid/plcsql/compiler/ast/ExprStr.java
```diff
@@ -51,29 +52,23 @@ public ExprStr(ParserRuleContext ctx, String val) {
     public String javaCode() {
 
         int len = val.length();
-        if (len <= MAX_STR_LITERAL_LEN) {
-            return "(\"" + val + "\")";
+        if (len <= STR_LITERAL_CUT_LEN) {
```
ctshim, line 55: Shouldn’t we find the length after escape processing?
`if( StringEscapeUtils.escapeJava(val).length() <= STR_LITERAL_CUT_LEN)`

hyunikn: Length must be of the original (unescaped) string. For example, unescaped "\t" rather than escaped "\\" + "t".

hgryoo (approved):

beyondykk9 (approved):

jongmin-won (approved):

ctshim (approved):

PR #5228, opened 2024-05-24
[CBRD-25377] Enable the 'monitor_waiting_thread' feature regardless of whether the EnableThreadMonitoring macro is defined

hornetmj:
http://jira.cubrid.org/browse/CBRD-25377

Purpose
- During SERVER_MODE builds, define the EnableThreadMonitoring macro to enable the 'monitor_waiting_thread' functionality
- Through error logs, we can determine how much time a thread spent on tasks such as file I/O or acquiring locks.

Implementation
- Add EnableThreadMonitoring definition in cubrid/CMakefiles.txt
  - Removing #ifdef EnableThreadMonitoring ~ #endif from the code is possible, but this approach was chosen to clearly distinguish the related code segments.

Remarks
- @mhoh3963
  - Could we also support SA_MODE?
  - The er_log_debug() function is used for logging. However, I suggest using _er_log_debug() with a prefix such as '[FILEIO]' or introducing a new error function, like monitor_waiting_thread_log(), for this purpose in this area. I prefer simply using _er_log_debug().

### DIFF
--- a/cubrid/CMakeLists.txt
+++ b/cubrid/CMakeLists.txt
@@ -521,7 +521,7 @@ if(WIN32)
   set_target_properties(cubrid PROPERTIES OUTPUT_NAME libcubrid)
 endif(WIN32)
 
-target_compile_definitions(cubrid PRIVATE SERVER_MODE CUBRID_EXPORTING ${COMMON_DEFS})
+target_compile_definitions(cubrid PRIVATE SERVER_MODE EnableThreadMonitoring CUBRID_EXPORTING ${COMMON_DEFS})
 if(NOT USE_CUBRID_ENV)
   target_compile_definitions(cubrid PRIVATE ${DIR_DEFS})
 endif(NOT USE_CUBRID_ENV)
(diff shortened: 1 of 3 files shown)


mhoh3963 (approved):

mhoh3963:
@hornetmj 
I don't think it's need to support it in SA_MODE and please  change it to _er_log_debug() as you suggested.

hornetmj:
> @hornetmj I don't think it's need to support it in SA_MODE and please change it to _er_log_debug() as you suggested.

I have modified it to '_er_log_debug()' and used '[MONITOR_WAITING_THREAD]' as the prefix. We can quickly gather the necessary information using the grep command as follows.
```
$ grep -r 'MONITOR_WAITING_THREAD' * | grep 'lock_'
server/demodb_20240524_1814.err:[MONITOR_WAITING_THREAD] lock_object:      9.689406
```

joohok (approved):

hornetmj:
@mhoh3963, @joohok As I mentioned in the description of jira, it need to check for any performance issues after the merge. If performance issues are detected, I will proceed with a revert.

PR #5233, opened 2024-05-27
[CBRD-25380] Add finalize option of memory_monitor class

Rudeus:
http://jira.cubrid.org/browse/CBRD-25380

Add new option --disable-force in `memmon` utility
 - disable-force option finalize `class memory_monitor mmon_Gl` forcibly
 - Change return type of `mmon_aggregate_server_info()` void to int
   - If mmon_Gl is exist, call mmon_Gl->aggregate_server_info() and return 0
   - If mmon_Gl is not exist, return -1
   - If `cubrid memmon demodb` is called when memory monitor is terminated using --disable-force, print nothing


src/communication/network_interface_sr.c
```diff
@@ -11324,3 +11329,36 @@ smmon_get_server_info (THREAD_ENTRY * thread_p, unsigned int rid, char *request,
...
+ */
+void
+smmon_finalize_force (THREAD_ENTRY * thread_p, unsigned int rid, char *request, int reqlen)
+{
+  char *ptr;
+  OR_ALIGNED_BUF (OR_INT_SIZE + OR_INT_SIZE) a_reply;
+  char *reply = OR_ALIGNED_BUF_START (a_reply);
+  int error = NO_ERROR;
+#if !defined(WINDOWS)
+  mmon_finalize ();
```
joohok: What should be done if other thread accesses `mmon_Gl` during finalization? If a particular thread once checked that `mmon_Gl` is not nullptr and performs operations, a crash might occur.

Rudeus: How about changing logic of mmon_finalize() like this?

temp_mmon_Gl = mmon_Gl;
mmon_Gl = nullptr;
// sleep a second
delete temp_mmon_Gl;

joohok:
Some Questions.
1. After forcefully disabling the memory monitoring feature, I think an appropriate error message should be displayed to the user if users try to use the `memmon` utility.
2. Since memory monitoring is disabled, should there still be communication between the utility and the server every time the `memmon` utility is executed?

src/base/memory_monitor_api.cpp
```diff
@@ -74,8 +74,16 @@ size_t mmon_get_allocated_size (char *ptr)
...
 }
 
-void mmon_aggregate_server_info (MMON_SERVER_INFO &server_info)
+int mmon_aggregate_server_info (MMON_SERVER_INFO &server_info)
 {
-  mmon_Gl->aggregate_server_info (server_info);
+  if (mmon_is_memory_monitor_enabled ())
+    {
+      mmon_Gl->aggregate_server_info (server_info);
+      return 0;
```
joohok: I think error code should be set when error occurred.

src/executables/util_cs.c
```diff
@@ -4603,6 +4605,17 @@ memmon (UTIL_FUNCTION_ARG * arg)
       goto error_exit;
     }
 
+  if (finalize_force)
```
hornetmj: We don't need to support -o and --disable-force together.

src/communication/network_interface_cl.c
```diff
@@ -11345,3 +11345,46 @@ mmon_get_server_info (MMON_SERVER_INFO & server_info)
   return ER_NOT_IN_STANDALONE;
 #endif /* !CS_MODE */
 }
+
+/*
+ * mmon_finalize_force - request to server to finalize memory_monitor forcely
+ *
+ * return : cubrid error
+ *
+ *   server_info(in/out): save memory usage information of the server
```
```suggestion
```

src/communication/network_interface_sr.c
hornetmj: > How about changing logic of mmon_finalize() like this?
> 
> temp_mmon_Gl = mmon_Gl; mmon_Gl = nullptr; // sleep a second delete temp_mmon_Gl;

It looks un-safe. The mmon_Gl is not volatile type. So, some thread use cached value after mmon_Gl is destroyed. And I think this approach can not guarantee that there are no threads in the momery monitor context.

src/base/memory_monitor_api.cpp
```diff
@@ -74,8 +74,16 @@ size_t mmon_get_allocated_size (char *ptr)
   return mmon_Gl->get_allocated_size (ptr);
 }
 
-void mmon_aggregate_server_info (MMON_SERVER_INFO &server_info)
+int mmon_aggregate_server_info (MMON_SERVER_INFO &server_info)
 {
-  mmon_Gl->aggregate_server_info (server_info);
+  if (mmon_is_memory_monitor_enabled ())
```
hornetmj: Comment for this.

src/communication/network_interface_sr.c
hornetmj: It would be nice to logging when cub_server receives/handles this request. So, user(DBA) can find/check this log in the error log.

src/base/memory_monitor_api.cpp
```diff
@@ -50,20 +50,22 @@ int mmon_initialize (const char *server_name)
...
-  if (mmon_is_memory_monitor_enabled ())
+  if (mmon_Gl != nullptr)
     {
 #if !defined (NDEBUG)
       mmon_Gl->finalize_dump ();
 #endif
 #if (MMON_DEBUG_LEVEL == 1) || (MMON_DEBUG_LEVEL == 3)
       mmon_Gl->print_debug_result ();
 #endif
+      mmon_disabled = true;
```
joohok:
```suggestion
```
I think this is superfluous

src/communication/network_interface_sr.c
```diff
@@ -11246,7 +11246,14 @@ smmon_get_server_info (THREAD_ENTRY * thread_p, unsigned int rid, char *request,
 #if !defined(WINDOWS)
   MMON_SERVER_INFO server_info;
 
-  mmon_aggregate_server_info (server_info);
+  if (mmon_is_memory_monitor_enabled ())
```
joohok, line 11249: Do we need to check if mmon is enabled?

src/executables/util_cs.c
```diff
@@ -4603,6 +4605,17 @@ memmon (UTIL_FUNCTION_ARG * arg)
...
+    {
+      error_code = mmon_disable_force ();
+      if (error_code != NO_ERROR)
+	{
```
joohok, line 4620: I think we need to print error message here

src/executables/util_cs.c
```diff
@@ -4603,6 +4605,17 @@ memmon (UTIL_FUNCTION_ARG * arg)
...
+	  goto error_exit;
+	}
+      fprintf (stdout, msgcat_message (MSGCAT_CATALOG_UTILS, MSGCAT_UTIL_SET_MEMMON, MEMMON_MSG_DISABLE_SUCCESS));
+      goto success_exit;
+    }
+
   /* execute phase */
   error_code = mmon_get_server_info (server_info);
   if (error_code != NO_ERROR)
```
joohok, line 4638: I think user need to know which error has occurred during executing utility.

src/communication/network_interface_sr.c
Rudeus: Because we have to check this to prevent to print outdated information.

src/executables/util_cs.c
Rudeus: Done.

src/executables/util_cs.c
Rudeus: Currently, mmon_disable_force() does not fail.

src/communication/network_interface_sr.c
```diff
@@ -11246,7 +11246,15 @@ smmon_get_server_info (THREAD_ENTRY * thread_p, unsigned int rid, char *request,
...
+    {
+      mmon_aggregate_server_info (server_info);
+    }
+  else
+    {
+      error = ER_FAILED;
```
joohok, line 11256: I think we need to leave a error log.

Wherever an error code is set, a log should be left to indicate why the error occurred. By seeing the log, one should be able to infer both the location and reason of the error. We use an interface called `er_set()` to set error codes and log messages for this purpose.

Honestly, I hope to have clear error codes and messages for specific situations. If this would result in too much work, we should look for ways to reduce the workload. 
(e.g, use `_er_log_debug` or something new function like `logpb_fatal_error` which don't require new error codes)

src/executables/util_cs.c
joohok: I understand that the server returns errors when running in a Windows environment. If disable-force is executed but fails because the server is running on Windows, an error message should be displayed.

src/executables/util_cs.c
joohok: I think we need to distinguish whether the memory usage cannot be displayed because the server is running on Windows (not on Linux?), or if it is due to disable-force.

msg/en_US.utf8/utils.msg
```diff
@@ -1319,9 +1319,13 @@ $set 59 MSGCAT_UTIL_SET_MEMMON
 1 Please set the system parameter 'enable_memory_monitoring' in 'cubrid.conf' and restart the database server.\n
 2 memmon cannot run as standalone mode.\n
 3 Fail to open output file with filename %1$s.\n
+4 Server memory monitor is disabled successfully.\n
```
joohok:
```suggestion
4 Memory monitoring on the server is disabled successfully.\n
```

msg/en_US.utf8/utils.msg
```diff
@@ -1319,9 +1319,13 @@ $set 59 MSGCAT_UTIL_SET_MEMMON
...
+5 disable-force cannot be used with other options.\n
+6 Memory monitor is disabled. You can use it again after restarting the server.\n
```
joohok:
```suggestion
6 Memory monitoring on the server has been disabled. You can use it again after restarting the server.\n
```

src/executables/util_cs.c
Rudeus: I understand. I'll do it.

src/executables/util_cs.c
Rudeus: Done.

src/communication/network_interface_sr.c
Rudeus: I add an error log using `_er_log_debug`. Thanks.

Rudeus: I add an error log using `_er_log_debug`. Thanks.

joohok (approved):

src/base/memory_monitor_api.cpp
```diff
@@ -50,13 +50,14 @@ int mmon_initialize (const char *server_name)
 	  error = ER_OUT_OF_VIRTUAL_MEMORY;
 	  return error;
 	}
+      mmon_disabled = false;
```
hornetmj, line 53: How about updating assert condition like assert (mmon_Gl == nullptr && mmon_disabled == true)?

hornetmj (approved):

src/base/memory_monitor_api.cpp
Rudeus: Done.

PR #5234, opened 2024-05-28
[CBRD-24658] Name binding error for SP in CTE and Aggregate statements

beyondykk9:
http://jira.cubrid.org/browse/CBRD-24658

When using javasp function in sql statements including UNION, the result is different compared to not using javasp function.


hgryoo (approved):

hgryoo:
LGTM!

hyunikn (approved):

ctshim (approved):

PR #5238, opened 2024-05-29
[CBRD-25312] Add error message for class_name which does not exists

YeunjunLee:

- Change short option of `class_name` from -c to -n
- Add error message for `class_name` which does not exists in database.

Both modification is for coherence with optimizedb utility. Optimizedb utility uses -n flag for `class_name`, and if the class_name does not exists, It prints `Unknown class "class_name".`

### DIFF
--- a/src/executables/util_sa.c
+++ b/src/executables/util_sa.c
@@ -1574,7 +1574,7 @@ diagdb (UTIL_FUNCTION_ARG * arg)
       goto print_diag_usage;
     }
 
-  if (diag == DIAGDUMP_ALL && class_name != NULL)
+  if (diag != DIAGDUMP_HEAP && class_name != NULL)
     {
       goto print_diag_usage;
     }
--- a/src/executables/util_sa.c
+++ b/src/executables/util_sa.c
@@ -1737,7 +1737,7 @@ diagdb (UTIL_FUNCTION_ARG * arg)
 	    {
 	      if (utility_check_class_name (class_name) != NO_ERROR)
 		{
-      db_shutdown ();
+		  db_shutdown ();
 		  goto error_exit;
 		}
 	    }
@@ -1750,7 +1750,7 @@ diagdb (UTIL_FUNCTION_ARG * arg)
 					 (MSGCAT_CATALOG_UTILS, MSGCAT_UTIL_SET_DIAGDB, DIAGDB_MSG_UNKNOWN_CLASS),
 					 class_name);
 		}
-        db_shutdown ();
+	      db_shutdown ();
 	      goto error_exit;
 	    }
 	}
--- a/src/executables/util_sa.c
+++ b/src/executables/util_sa.c
@@ -1738,7 +1738,7 @@ diagdb (UTIL_FUNCTION_ARG * arg)
 		{
 		  goto error_exit;
 		}
-      }
+	    }
 	  heap_dump_heap_file (thread_p, outfp, dump_records, class_name);
 	}
     }
--- a/src/executables/util_sa.c
+++ b/src/executables/util_sa.c
@@ -1536,6 +1536,7 @@ diagdb (UTIL_FUNCTION_ARG * arg)
   bool is_emergency = false;
   DIAGDUMP_TYPE diag;
   THREAD_ENTRY *thread_p;
+  int error_code = NO_ERROR;
 
   db_name = utility_get_option_string_value (arg_map, OPTION_STRING_TABLE, 0);
   if (db_name == NULL)
@@ -1739,11 +1740,16 @@ diagdb (UTIL_FUNCTION_ARG * arg)
 		  goto error_exit;
 		}
 	    }
-	  if (heap_dump_heap_file (thread_p, outfp, dump_records, class_name) == ER_LC_UNKNOWN_CLASSNAME)
+	  error_code = heap_dump_heap_file (thread_p, outfp, dump_records, class_name);
+	  if (error_code != NO_ERROR)
 	    {
-	      PRINT_AND_LOG_ERR_MSG (msgcat_message
-				     (MSGCAT_CATALOG_UTILS, MSGCAT_UTIL_SET_DIAGDB, DIAGDB_MSG_UNKNOWN_CLASS),
-				     class_name);
+	      if (error_code == ER_LC_UNKNOWN_CLASSNAME)
+		{
+		  PRINT_AND_LOG_ERR_MSG (msgcat_message
+					 (MSGCAT_CATALOG_UTILS, MSGCAT_UTIL_SET_DIAGDB, DIAGDB_MSG_UNKNOWN_CLASS),
+					 class_name);
+		}
+	      goto error_exit;
 	    }
 	}
     }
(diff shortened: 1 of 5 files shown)


src/storage/heap_file.c
```diff
@@ -14530,9 +14531,13 @@ heap_dump_heap_file (THREAD_ENTRY * thread_p, FILE * fp, bool dump_records, cons
   status = xlocator_find_class_oid (thread_p, class_name, &class_oid, S_LOCK);
   if (status != LC_CLASSNAME_EXIST)
     {
+      fprintf (stderr, msgcat_message (MSGCAT_CATALOG_UTILS, MSGCAT_UTIL_SET_DIAGDB, DIAGDB_MSG_UNKNOWN_CLASS),
```
joohok: I think it makes sense for utility error messages to be handled within the utility itself. 
This function simply return an error code, and the utility could use that error code to output the error message.
The message should not only be printed to stderr but also logged in the utility error file. What do you think?

YeunjunLee: I checked whether optimizedb writes a log in the utility error file when a given class-name does not exist. I found that optimizedb does not log in this situation. I could be mistaken, but if optimizedb indeed behaves this way, I believe diagdb should also avoid logging in the utility error file for consistency.

joohok: Please, see why and how we use `PRINT_AND_LOG_ERR_MSG` function in `diagdb`. 
There is no need to follow the implementation of optimizedb. It is just logging the necessary situations. 

And the server and utilities should be separated and display error messages accordingly.

src/executables/util_sa.c
```diff
@@ -1739,7 +1739,12 @@ diagdb (UTIL_FUNCTION_ARG * arg)
 		  goto error_exit;
 		}
 	    }
-	  heap_dump_heap_file (thread_p, outfp, dump_records, class_name);
+	  if (heap_dump_heap_file (thread_p, outfp, dump_records, class_name) == ER_LC_UNKNOWN_CLASSNAME)
+	    {
+	      PRINT_AND_LOG_ERR_MSG (msgcat_message
+				     (MSGCAT_CATALOG_UTILS, MSGCAT_UTIL_SET_DIAGDB, DIAGDB_MSG_UNKNOWN_CLASS),
+				     class_name);
+	    }
```
joohok:
```suggestion
          error_code = heap_dump_heap_file ();
	  if (error_code != NO_ERROR)
	    {
	       if (error_code == ER_LC_UNKNOWN_CLASSNAME) // or using switch statement
                 {	      
	             PRINT_AND_LOG_ERR_MSG ();
	         }
	         
	       goto error_exit;	    
	     }
```

I think there should be more error cases which are not handled now.
How about make additional variable to get error code ?

YeunjunLee: Other error codes have been decided not to be handled in this issue. However, I believe this change could improve handling of other error cases in the future.

src/executables/util_sa.c
```diff
@@ -1739,11 +1740,16 @@ diagdb (UTIL_FUNCTION_ARG * arg)
...
-	      PRINT_AND_LOG_ERR_MSG (msgcat_message
-				     (MSGCAT_CATALOG_UTILS, MSGCAT_UTIL_SET_DIAGDB, DIAGDB_MSG_UNKNOWN_CLASS),
-				     class_name);
+	      if (error_code == ER_LC_UNKNOWN_CLASSNAME)
+		{
+		  PRINT_AND_LOG_ERR_MSG (msgcat_message
+					 (MSGCAT_CATALOG_UTILS, MSGCAT_UTIL_SET_DIAGDB, DIAGDB_MSG_UNKNOWN_CLASS),
+					 class_name);
+		}
+	      goto error_exit;
```
joohok, line 1754: Ah, one thing I missed here is that we should call `db_shutdown ()` ;; 
We can call `db_shutdown ()` by removing this `goto` statement or just call the `db_shutdown ()` above the `goto` label

YeunjunLee:
```suggestion     
	      db_shutdown ();
	      goto error_exit;  
```

I think `goto` statement in above (line 1740) also needs `db_shutdown ()`.

src/storage/heap_file.c
```diff
@@ -14530,23 +14531,24 @@ heap_dump_heap_file (THREAD_ENTRY * thread_p, FILE * fp, bool dump_records, cons
...
     }
 
+  fprintf (fp, "\n*** DUMP HEAP OF %s ***\n", class_name);
+
   error_code = heap_hfid_cache_get (thread_p, &class_oid, &hfid, NULL, NULL);
   if (error_code != NO_ERROR)
     {
-      assert (false);
-      return;
+      return error_code;
```
```suggestion
      assert (false);
      return error_code;
```
Since it should never happen, it is good to leave the assert(false) as it is. Let's use this to check for issues in debug build mode.

src/storage/heap_file.c
```diff
@@ -14530,23 +14531,24 @@ heap_dump_heap_file (THREAD_ENTRY * thread_p, FILE * fp, bool dump_records, cons
...
     }
 
   heap_dump (thread_p, fp, &hfid, dump_records);
 
   error_code = heap_get_class_partitions (thread_p, &class_oid, &parts, &parts_count);
   if (error_code != NO_ERROR)
     {
-      assert (false);
-      return;
+      return error_code;
```
```suggestion
      assert (false);
      return error_code;
```

src/storage/heap_file.c
```diff
@@ -14517,7 +14518,7 @@ heap_dump (THREAD_ENTRY * thread_p, FILE * fp, HFID * hfid, bool dump_records)
  * dump_records (in) : true to dump records
  * class_name (in)   : name of class to dump
  */
-void
+int
```
hornetmj, line 14520: Add a comment explaining `* return :`.

src/storage/heap_file.c
```diff
@@ -14530,23 +14531,24 @@ heap_dump_heap_file (THREAD_ENTRY * thread_p, FILE * fp, bool dump_records, cons
   status = xlocator_find_class_oid (thread_p, class_name, &class_oid, S_LOCK);
   if (status != LC_CLASSNAME_EXIST)
     {
-      return;
+      er_set (ER_WARNING_SEVERITY, ARG_FILE_LINE, ER_LC_UNKNOWN_CLASSNAME, 1, class_name);
```
hornetmj: It is not an issue if a non-existent table name is entered by the user. Therefore, there seems to be no need to explicitly perform error setting (er_set()).

src/storage/heap_file.c
```diff
@@ -46,6 +46,7 @@
 #include "btree.h"
 #include "btree_unique.hpp"
 #include "schema_system_catalog_constants.h"	/* for CT_SERIAL_NAME */
+#include "utility.h"
```
joohok:
```suggestion
```

joohok (approved):

src/storage/heap_file.c
YeunjunLee: Done.

src/storage/heap_file.c
```diff
@@ -14530,14 +14530,16 @@ heap_dump_heap_file (THREAD_ENTRY * thread_p, FILE * fp, bool dump_records, cons
   status = xlocator_find_class_oid (thread_p, class_name, &class_oid, S_LOCK);
   if (status != LC_CLASSNAME_EXIST)
     {
-      return;
+      return ER_LC_UNKNOWN_CLASSNAME;
```
hornetmj: NIT:

How about using a goto statement to consistently return the error_code?

YeunjunLee: Done.

hornetmj (approved):

PR #5244, opened 2024-06-04
[CBRD-25396] When using GROUP BY and ORDER BY simultaneously within a subquery, the NO_MERGE hint does not function correctly.

Hamkua:
http://jira.cubrid.org/browse/CBRD-25396

When GROUP BY and ORDER BY are used in a subquery, the queries are separated and recombined in the rewriter.

The existing subquery is modified to include an additional subquery in the FROM clause, during which the original hints are moved to the newly added subquery.

The GROUP BY clause moves inside the newly added subquery, allowing the hints to be applied. However, the ORDER BY clause remains outside the newly added subquery, so the hints cannot be applied.

The user writing the query cannot know that the subquery they wrote has been split and recombined, and they will perceive that the hints are not being applied, requiring modification.

It is not necessary to modify all hints to ensure they do not move inside the newly added subquery. Only the hints related to Query Caching and Merge need to be adjusted to remain outside.


shparkcubrid (approved):

xmilex-git (approved):

src/parser/view_transform.c
```diff
@@ -4811,8 +4811,8 @@ mq_rewrite_aggregate_as_derived (PARSER_CONTEXT * parser, PT_NODE * agg_sel)
     }
 
   /* move hint, from, where, group_by, using_index part over */
-  derived->info.query.q.select.hint = agg_sel->info.query.q.select.hint;
-  agg_sel->info.query.q.select.hint = PT_HINT_NONE;
+  derived->info.query.q.select.hint = agg_sel->info.query.q.select.hint & (~PT_HINT_NO_MERGE);
+  agg_sel->info.query.q.select.hint &= PT_HINT_NO_MERGE;
```
shparkcubrid: Please check whether PT_HINT_QUERY_CACHE should be reflected.

shparkcubrid: 아래 이슈에서 PT_HINT_QUERY_CACHE도 복사하도록 수정되었습니다. 해당 힌트도 같이 복사되도록 수정해주세요.

https://github.com/CUBRID/cubrid/pull/4973
```
-587,6 +587,7 @@ pt_get_hint (const char *text, PT_HINT hint_table[], PT_NODE * node)
		{
		  node->info.query.flag.do_cache = 1;
		}
	      node->info.query.q.select.hint = (PT_HINT_ENUM) (node->info.query.q.select.hint | hint_table[i].hint);
	    }
	  break;
```

youngjinj (approved):

src/parser/view_transform.c
Hamkua: 해당 이슈 [#4973](https://github.com/CUBRID/cubrid/pull/4973)의 파싱 과정에서
QUERY_CACHE 힌트를 기존의 `info.query.hint` 뿐만 아니라`info.query.q.select.hint`에도 표기하는 것을 확인했으며, 해당 힌트도 같이 복사되도록 수정하겠습니다.

PR #5245, opened 2024-06-04
[CBRD-25311] Dump a specific class by taking class-name argument

YeunjunLee:
http://jira.cubrid.org/browse/CBRD-25311

- Support the functionality to specify only a specific class instead of dumping the entire heap files.

Note: This functionality only gets single class_name in present, and usage message is not supported yet. Multi class_name and usage message will be supported in further issue.


joohok:
@YeunjunLee 
I think we can close this

PR #5246, opened 2024-06-04
[CBRD-25312] Add error message when class_name do not exists in database

YeunjunLee:

- Add error message when class_name do not exists in database
- Change short option of class name from -c to -n


src/executables/util_sa.c
```diff
@@ -1736,11 +1737,22 @@ diagdb (UTIL_FUNCTION_ARG * arg)
 	    {
 	      if (utility_check_class_name (class_name) != NO_ERROR)
 		{
+		  db_shutdown ();
```
hornetmj: How about considering the following approach? With this, the next developer won't have to worry about calling db_shutdown().

```
bool need_db_shutdown = false;

...
db_restart ();
need_db_shutdown = true;

...

error_exit:

if (need_db_shutdown == true)
{
    db_shutdown ();
}
```

src/executables/util_sa.c
```diff
@@ -1736,11 +1738,22 @@ diagdb (UTIL_FUNCTION_ARG * arg)
 	    {
 	      if (utility_check_class_name (class_name) != NO_ERROR)
 		{
+		  need_db_shutdown = true;
```
hornetmj: After a successful call of db_restart(), if branching to error_exit:, it is necessary to always call db_shutdown(). Therefore, it is more appropriate to set once need_db_shutdown to true immediately after the successful invocation of db_restart().

```
  if (db_restart (arg->command_name, TRUE, db_name) != NO_ERROR)
    {
      PRINT_AND_LOG_ERR_MSG ("%s\n", db_error_string (3));
      goto error_exit;
    }

    need_db_shutdown = true;
```

YeunjunLee: I understand. I changed it.

hornetmj (approved):

PR #5254, opened 2024-06-13
[CBRD-25419] If you use the -i and --input-class-only options together in unloaddb, unrelated alter serial statements will be output.

airnet73:
http://jira.cubrid.org/browse/CBRD-25419

Purpose
When extracting schema information using the -i and --input-class-only options together in unloaddb, the schema of the table specified in the input file and the related ALTER SERIAL statement are included and extracted.

Implementation
N/A

Remarks
N/A

### DIFF
--- a/src/executables/unload_schema.c
+++ b/src/executables/unload_schema.c
@@ -1001,7 +1001,7 @@ emit_class_alter_serial (extract_context & ctxt, print_output & output_ctx)
     {
       do
 	{
-	  for (i = 0; i < SERIAL_VALUE_INDEX_MAX; i++)
+	  for (i = 0; i < ALTER_SERIAL_VALUE_INDEX_MAX; i++)
 	    {
 	      error = db_query_get_tuple_value (query_result, i, &values[i]);
 	      if (error != NO_ERROR)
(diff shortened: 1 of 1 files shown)


src/executables/unload_schema.c
```diff
@@ -988,7 +1012,7 @@ emit_class_alter_serial (extract_context & ctxt, print_output & output_ctx)
 	      /* Validation of the result value */
```
kisoo-han, line 1012: This switch is in the for-loop on line 1004.
`1004 for (i = 0; i < SERIAL_VALUE_INDEX_MAX; i++)`

Shouldn't this for-loop termination condition be changed ? (I'm not sure)
`1004 for (i = 0; i < ALTER_SERIAL_VALUE_INDEX_MAX; i++)`

airnet73: Change the size of the for-loop and values[] array.
SERIAL_VALUE_INDEX_MAX -> ALTER_SERIAL_VALUE_INDEX_MAX
Thanks for the advice.

mhoh3963 (approved):

kisoo-han (approved):

PR #5260, opened 2024-06-17
[CBRD-25426] If the current_val of serial is the same as max_val, an error occurs during unloaddb/loaddb.

airnet73:
http://jira.cubrid.org/browse/CBRD-25426

Purpose
If you create SERIAL in ascending/descending order and increase current_val to max_val and then perform unload/load, two errors occur.
* In the case of ascending Serial, an error occurs when performing unloaddb.
* In the case of descending Serial, no error occurs in unladdb. However, an error occurs when performing loaddb.

Implementation

Remarks

### DIFF
--- a/src/executables/unload_schema.c
+++ b/src/executables/unload_schema.c
@@ -824,58 +824,6 @@ export_serial (extract_context & ctxt, print_output & output_ctx)
 		}
 	    }
 
-	  if (db_get_int (&values[SERIAL_STARTED]) == 1)
-	    {
-	      /* Calculate next value of serial */
-	      db_make_null (&diff_value);
-	      error = numeric_db_value_sub (&values[SERIAL_MAX_VAL], &values[SERIAL_CURRENT_VAL], &diff_value);
-	      if (error == ER_IT_DATA_OVERFLOW)
-		{
-		  // max - curr might be flooded.
-		  diff_value = values[SERIAL_MAX_VAL];
-		  er_clear ();
-		}
-	      else if (error != NO_ERROR)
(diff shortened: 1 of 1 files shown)


src/executables/unload_schema.c
```diff
@@ -897,6 +845,11 @@ export_serial (extract_context & ctxt, print_output & output_ctx)
 	    }
 	  output_ctx (";\n");
 
+	  if (db_get_int (&values[SERIAL_STARTED]) == 1)
+	    {
+	      output_ctx ("SELECT %s%s%s.NEXT_VALUE;\n ", PRINT_IDENTIFIER (db_get_string (&values[SERIAL_NAME])));
```
kisoo-han, line 850: What about use of '**NEXTVAL**' instead of '**NEXT_VALUE**'

kisoo-han (approved):
I think this will work, however, shouldn't we think about whether there is a method other than SELECT .. nextval?

src/executables/unload_schema.c
airnet73: I know NEXT_VALUE and NEXTVAL mean the same thing. Is there a reason why you recommend NEXTVAL?
For reference, I used NEXT_VALUE when extracting SERIAL generated by AUTO_INCREMENT.

kisoo-han: Definitely, 'next_val' could working, but I think 'nextval' is more common term for next value for SERIAL/SEQUENCE. Its your choice.

mhoh3963 (approved):

PR #5382, opened 2024-08-05
[CBRD-25498] Pass SP's owner to PL Compiler when creating PL/CSQL

hgryoo:
http://jira.cubrid.org/browse/CBRD-25498

To handle CREATE PROCEDURE with [user name].[sp_name] properly, the owner name should be passed to PL compiler to know which user is going to have to own the procedure.

### DIFF
--- a/src/executables/plcsql.cpp
+++ b/src/executables/plcsql.cpp
@@ -49,7 +49,7 @@ static std::string input_string;
 static bool verbose = false;
 
 // output
-static PLCSQL_COMPILE_INFO compile_info;
+static PLCSQL_COMPILE_RESPONSE compile_response;
 
 #define DB_PLCSQL_AS_ARGS(arg) \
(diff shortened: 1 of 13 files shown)


jongmin-won (approved):

hyunikn (approved):

src/executables/plcsql.cpp
```diff
@@ -49,7 +49,7 @@ static std::string input_string;
 static bool verbose = false;
 
 // output
-static PLCSQL_COMPILE_INFO compile_info;
+static PLCSQL_COMPILE_RESPONSE compile_response;
```
hyunikn, line 52: (질문) 이 executable 을 유지할 필요가 있을까요.

hgryoo: 필요하면 검토 후 별도의 이슈에서 제거 할 예정입니다.

beyondykk9 (approved):

PR #5573, opened 2024-10-23
[CBRD-25563] Set bypass mask for SHOW GRANTS statement

hgryoo:
http://jira.cubrid.org/browse/CBRD-25563

In SHOW Statement, table's auth does not have to be checked because the show statement's query is system generated.
Through pt_add_table_name_to_from_list(), the bypass mask is set. however the show statement's query is generated through parser_parse_XX API, the mask is not set.

In this change, through parser_walk_tree (), auth_bypass_mask is set to avoid checking system table's auth in SHOW GRANTS.

Repro
```
CREATE FUNCTION test1() RETURN int as
language java name 'SpTest3.typetestint0() return int';
 
CREATE USER t1;
GRANT EXECUTE ON PROCEDURE test1 TO t1;
 
call login('t1','') on class db_user;
SELECT * FROM db_auth WHERE grantee_name = 'T1';
SHOW GRANTS FOR T1;
```

Expected
```
  Grants for T1
======================
  'GRANT EXECUTE ON dba.test1 TO T1'
```

Actual
```
In line 2, column 244,
ERROR: SELECT is not authorized on _db_auth.
 
In line 2, column 536,
ERROR: SELECT is not authorized on _db_auth.

```


jongmin-won (approved):

src/parser/parser_support.c
```diff
@@ -7769,6 +7769,32 @@ pt_make_query_show_grants_curr_usr (PARSER_CONTEXT * parser)
...
+  DB_AUTH *auth_bypass;
+
+  assert (arg != NULL);
+
+  if (node == NULL)
+    {
+      return node;
+    }
+
+  switch (node->node_type)
```
ctshim: PT_SPEC인 경우만 처리하는데 switch문으로 구성하는 이유가 있나요?
이런 경우 if문이 더 효율적이지 않나 싶습니다.

ctshim (approved):

hyunikn (approved):

beyondykk9 (approved):

vimkim (approved):

PR #5668, opened 2024-12-04
[CBRD-25154] Character set consideration of DB server's charset

hgryoo:
http://jira.cubrid.org/browse/CBRD-25154

- Set Java's default Charset when bootraping PL server
- When compiling PL/CSQL, the source code and object code are stored with system charset
- When communticating strings between DB server and PL server, packer/unpacker considers codeset of the DB server.


beyondykk9 (approved):

hyunikn:
ServerConfig.java 에 라이센스 문구가 없습니다.

hyunikn (approved):

PR #5780, opened 2025-01-08
[CBRD-25801] Problems that occur when a hidden column is generated due to ORDER BY clause in a subquery

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-25801

In CBRD-24181, when optimizing the select list of a subquery, a column in ORDER BY is added. At that time, a hidden column property is added, and the query that has already been rewritten in qo_rewrite_hidden_col_as_derived() is unnessarary rewritten twice, causing an issue. Fix to remove the hidden property if the existing select list does not have it.

### DIFF
--- a/src/parser/view_transform.c
+++ b/src/parser/view_transform.c
@@ -1473,6 +1473,7 @@ mq_remove_select_list_for_inline_view (PARSER_CONTEXT * parser, PT_NODE * statem
   PT_NODE *query_spec_columns, *tmp_query, *save_order_by, *save_select_list;
   PT_NODE *attributes, *attr, *as_attr_list;
   PT_NODE *col, *new_select_list, *spec, *pred, *subquery;
+  bool is_unset_hidden_col;
 
   assert (PT_IS_SELECT (statement));
   if (derived_spec == NULL || !PT_SPEC_IS_DERIVED (derived_spec))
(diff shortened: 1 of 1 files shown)


src/parser/view_transform.c
```diff
@@ -1586,6 +1587,27 @@ mq_remove_select_list_for_inline_view (PARSER_CONTEXT * parser, PT_NODE * statem
 	}
     }
 
+  /* check whether to unset hidden col. If no hidden column in original, remove hidden settings */
+  is_unset_hidden_col = true;
+  for (col = subquery->info.query.q.select.list; col; col = col->next)
+    {
+      if (col->flag.is_hidden_column)
+	{
+	  is_unset_hidden_col = false;
```
youngjinj, line 1596: `false`도 문제 없지만 다른 코드에서는 1 또는 0을 설정하고 있습니다.
```suggestion
	  is_unset_hidden_col = 0;
```

src/parser/view_transform.c
```diff
@@ -1586,6 +1587,27 @@ mq_remove_select_list_for_inline_view (PARSER_CONTEXT * parser, PT_NODE * statem
...
+	  break;
+	}
+    }
+  if (is_unset_hidden_col)
+    {
+      for (col = tmp_query->info.query.q.select.list; col; col = col->next)
+	{
+	  if (col->flag.is_hidden_column)
+	    {
+	      col->flag.is_hidden_column = false;
```
youngjinj: `false`도 문제 없지만 다른 코드에서는 1 또는 0을 설정하고 있습니다.
```suggestion
	      col->flag.is_hidden_column = 0;
```

youngjinj (approved):

PR #5826, opened 2025-01-20
[CBRD-25658] fix incorrect PT_DEFAULT_TYPE

mhoh3963:
http://jira.cubrid.org/browse/CBRD-25658

- Fix an error where incorrect PT_DEFAULT_TYPE was returned due to a string length comparison error.


mhoh3963:
다른 PR과 동일하게 중복해결하여 close 합니다.

PR #5833, opened 2025-01-23
[CBRD-25848] Fix an issue where the loadjava tool would not output an error message when there was no class file

ctshim:
http://jira.cubrid.org/browse/CBRD-25848

*  Fix an issue where the loadjava tool would not output an error message when there was no class file
* Check file extensions (.class, .jar)

```
$ loadjava demodb xxx
 loadjava fail: 'xxx' is not exist.
$
$ loadjava demodb SpTest.java
loadjava fail: The extension name of 'SpTest.java' is invalid.
$
```

### DIFF
--- a/src/executables/loadjava.cpp
+++ b/src/executables/loadjava.cpp
@@ -183,6 +183,7 @@ copy_file (const fs::path &java_dir_path)
       fs::path src_path = fs::path (Src_class);
       if (fs::exists (src_path) == false)
 	{
+	  fprintf (stderr, "loadjava fail: '%s' is not exist.\n", src_path.c_str ());
 	  return ER_FAILED;
 	}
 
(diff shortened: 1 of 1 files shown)


hgryoo (approved):

beyondykk9 (approved):

src/executables/loadjava.cpp
```diff
@@ -183,6 +183,14 @@ copy_file (const fs::path &java_dir_path)
       fs::path src_path = fs::path (Src_class);
       if (fs::exists (src_path) == false)
 	{
+	  fprintf (stderr, "loadjava fail: '%s' is not exist.\n", src_path.c_str ());
```
hyunikn: is not exist --> does not exisit

hyunikn (approved):

src/executables/loadjava.cpp
ctshim: Thank you.

PR #5928, opened 2025-02-24
[CBRD-25457] backport: Executing the “AUTO_INCREMENT” and “DEFAULT” attributes separately using the “ALTER TABLE MODIFY” syntax on the same column resulted in an error of two attributes co-existing in that column (#5364)

beyondykk9:
http://jira.cubrid.org/browse/CBRD-25457

### Purpose

The following error occurs when the 'AUTO_INCREMENT' and 'DEFAULT' properties are used at the same time when executing the 'CREATE TABLE' and 'ALTER COLUMN' statements. ERROR: SHARED, DEFAULT and AUTO_INCREMENT cannot be defined with each other.

However, when using the 'ALTER TABLE .. MODIFY' statement to execute 'AUTO_INCREMENT' and 'DEFAULT' attributes separately, the two attributes can coexist on the same column.

Therefore, we modified it so that only the last-applied 'AUTO_INCREMENT' and 'DEFAULT' attribute is applied.

### Implementation

Backport

### Remarks

11.3 Patch 3에 적용합니다.


jongmin-won:
test_sql의 실패 1건 TC의 답안지(sql/_13_issues/_14_1h/cases/bug_bts_12663.sql) 수정이 필요합니다.

답안지 수정한 PR : https://github.com/CUBRID/cubrid-testcases/pull/1800

jongmin-won (approved):

PR #4365, opened 2023-05-16
[CBRD-24795] When using a function under <= and >= conditions, the plan generator does not perform a range scan

youngjinj:
http://jira.cubrid.org/browse/CBRD-24795

Make the function expression used in the function index the target in the process of rewriting the query below.
1. qo_converse_sarg_terms
   - convert terms of the form 'const op attr' to 'attr op const'
   ```
   (-) attribute1 operation (-) attribute2 → attribute1 converse_operation attribute2
   [ x ] (-) attribute1 operation attribute2 → attribute1 converse_operation (-) attribute2
   (-) attribute1 operation constant → attribute1 converse_operation (-) constant
   constant operation (-) attribute1 → (-) constant converse_operation attribute1
   
   /* swap */
   attribute2 #1 operation attribute1 #3 → attribute1 #3 converse_operation attribute2 #1
   constant operation attribute1 → attribute1 converse_operation constant
   ```
2. qo_reduce_comp_pair_terms (qo_search_comp_pair_term)
   - reduce a pair of comparison terms into one BETWEEN term
   ```
   attribute operation const1 AND attribute operation const2
    → attribute BETWEEN const1 PT_BETWEEN_operation> const2
   ```
3. qo_convert_to_range (qo_convert_to_range_helper, qo_merge_range_helper)
   - convert comparison terms to RANGE
   ```
   attribute1 operation constant1 OR attribute1 operation constant2
   → attribute1 RANGE (constant1 operation gt_inf max OR min inf_lt constant2)
   
   attribute1 BETWEEN constant1 gt_lt constant2
   → attribute1 RANGE (constant1 gt_lt constant2)
   
   attribute1 IN (constant1, constant2, constant3);
   → attribute1 RANGE (constant1 = OR constant2 = OR constant3 =)
   
   /* merge */
   (attribute1 operation constant1 OR attribute1 operation constant3) AND (attribute1 operation constant2 OR attribute1 operation constant3)
   → (attribute1 operation constant1) AND (attribute1 operation constant2 OR attribute1 operation constant3)
   ```
4. qo_apply_range_intersection (qo_apply_range_intersection_helper)
   - narrow search range by applying range intersection
   ```
   /* merge */
   attribute1 RANGE (constant1 operation constant2) AND attribute1 RANGE (constant3 = OR constant4 = OR constant5 =)
   → attribute1 RANGE (constant3 = OR constant4 = OR constant5 =)
   ```
5. qo_do_auto_parameterize
[...]

PR #5937, opened 2025-02-24
[CBRD-24795] When using a function under <= and >= conditions, the plan generator does not perform a range scan (#4365, #4830)

youngjinj:
http://jira.cubrid.org/browse/CBRD-24795
http://jira.cubrid.org/browse/CBRD-25116

backport #4365, #4830


youngjinj:
`CBRD-24795`가 `10.2 Patch 16` 대상에서 제외되어 close 합니다.

PR #6246, opened 2025-06-12
[CBRD-26109] fix coredump that could occur when cas receives sigterm and clears.

mhoh3963:
http://jira.cubrid.org/browse/CBRD-26109

- When a signal handler is executed during boot_client_all_finalize() execution and the the function is executed, a coredump may occur.
- So, modify this function to ignore the signal handler at the start and restore it at the end.


beyondykk9 (approved):

hornetmj (approved):
꼼꼼히 리뷰 못하고 승인합니다.

PR #6314, opened 2025-07-14
[CBRD-26111] modify to output exactly the same as the prev version, when using --for-cm

mhoh3963:
http://jira.cubrid.org/browse/CBRD-26111

When used --for-cm,
 - print all parameters
 - print the parameters of the client part, when used -b together

### DIFF
--- a/src/executables/util_cs.c
+++ b/src/executables/util_cs.c
@@ -2341,29 +2341,45 @@ paramdump (UTIL_FUNCTION_ARG * arg)
       goto error_exit;
     }
 
-#if defined(SA_MODE)
   if (for_cm_flag)
     {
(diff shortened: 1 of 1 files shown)


mhoh3963:
cubrid manager (cubrid admin) 호환성을 맞추기 위해 --for-cm 인 경우에 예전 방식과 동일하게 출력하게 수정

kisoo-han (approved):
This fix produces output identical to previous versions of paramdump with respect to CM, and works without issues with CM/CA.
Thank you for your efforts!

ctshim (approved):

beyondykk9 (approved):

PR #6357, opened 2025-08-05
[CBRD-26228] Remove GCC 9 warnings in csql.c and cas_runner.c

hgryoo:
http://jira.cubrid.org/browse/CBRD-26228

### Purpose

- Remove warnings on GCC 9

### Implementation

- Declare the l variable as volatile to suppress the warning caused by setjmp in csql.c.
- Move fprintf(... %s... , info_str) before FREE_MEM(info_str) to avoid use-after-free and handle nulls.

### Remarks

N/A


beyondykk9:
sql fail 체크가 필요합니다.

beyondykk9 (approved):

PR #6516, opened 2025-10-01
[CBRD-26313] View-merging issue even when there is a query condition other than rownum

shparkcubrid:
http://jira.cubrid.org/browse/CBRD-26313

View-merging is prohibited if there is a query condition other than rownum.


shparkcubrid:
http://jira.cubrid.org/browse/CBRD-26313

View-merging is prohibited if there is a query condition other than rownum.

PR #6678, opened 2025-12-01
[CBRD-24584] Improved to minimize the impact of query execution speed when setting sql_trace_slow (#4002)

beyondykk9:
<http://jira.cubrid.org/browse/CBRD-24584>

### Purpose

When setting the sql_trace_slow value in cubrid.conf to a positive number other than -1, the query performance is highly affected compared to 9.x.

The reason for the increased impact on query performance compared to 9.x is that statdump information is calculated before starting a query and after finishing query execution.

Therefore, it is necessary to minimize the impact on query performance by improving it to calculate only the information necessary for setting sql_trace_slow.

### Implementation

backport from 11.3

### Remarks

also need to backport 11.0, and 10.2

### DIFF
--- a/src/base/perf_monitor.c
+++ b/src/base/perf_monitor.c
@@ -983,6 +983,30 @@ perfmon_server_get_stats (THREAD_ENTRY * thread_p)
 }
 
 /*
+ *   xperfmon_server_copy_stats_for - Copy recorded server statistics for trace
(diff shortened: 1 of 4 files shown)


beyondykk9:
close the PR, because the issue do not need to backport to 11.2 patch

PR #6732, opened 2025-12-24
Cubrid histogram

sohee-dgist:
<http://jira.cubrid.org/browse/CBRD-XXXX>

### Purpose

N/A

### Implementation

N/A

### Remarks

N/A


sohee-dgist:
@cursor review

PR #6765, opened 2026-01-12
[CUBVEC-151] Fix recall degradation on nytimes-256-angular dataset

hgryoo:
http://jira.cubrid.org/browse/CUBVEC-151

### Purpose

Recall on the nytimes-256-angular dataset was severely degraded, in some cases approaching zero. The primary root cause was graph disconnection at layer 0.
- Due to the data distribution characteristics of the nytimes-256-angular dataset, the graph constructed at layer 0 could become partially disconnected.
- Since layer 0 serves as the final navigation layer for ANN search, this disconnection critically limited reachability and caused search to terminate in isolated subgraphs, resulting in near-zero recall.

### Implementation

Improve layer-0 graph connectivity (primary fix)
- The maximum number of neighbors allowed at layer 0 is increased to 2× the previous limit.
- This significantly reduces the risk of graph disconnection and improves navigability for skewed or clustered data distributions.

Enforce consistent normalization for COSINE metric (supporting fix)
- Vectors are normalized before being stored in the index.
- User-provided query vectors are normalized at query time to ensure correct cosine similarity semantics.

### Remarks

N/A

### DIFF
--- a/src/storage/index_hnsw/hnsw_algo.hpp
+++ b/src/storage/index_hnsw/hnsw_algo.hpp
@@ -398,13 +398,9 @@ namespace cubhnsw
     // TODO: now, connectivity_base is not considered.
     // std::size_t connecitvity_max = m_connectivity;
     std::size_t top_limit = std::max (connectivity_max, m_expansion);
-    if (!top.reserve (top_limit))
+    if (!top.reserve (top_limit) || !next.reserve (top_limit))
       {
-	assert (false);
-	return {ER_FAILED, OID_INITIALIZER};
-      }
-    if (!next.reserve (m_expansion))
-      {
+	er_set (ER_ERROR_SEVERITY, ARG_FILE_LINE, ER_OUT_OF_VIRTUAL_MEMORY, 1, top_limit * sizeof (candidate_t<Traits>));
 	assert (false);
 	return {ER_FAILED, OID_INITIALIZER};
       }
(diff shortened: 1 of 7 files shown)


YeunjunLee (approved):

hyahong (approved):

hornetmj (approved):

youngjun9072 (approved):

H2SU (approved):

vimkim (approved):

src/storage/index_hnsw/hnsw_algo.hpp
```diff
@@ -496,23 +511,16 @@ namespace cubhnsw
       }
 
     algo_context_t<Traits> context;
-    context.m_thread_p = thread_get_thread_entry_info();
+    context.m_thread_p = thread_p;
     context.clear_candidates();
```
vimkim, line 447: OOS 프로젝트에서도 thread_get_thread_entry_info() 를 사용하고 있어서 생긴 순수 궁금증입니다.

함수를 thread_p 로 교체하신 이유가 궁금합니다. 혹시 문제가 발생했을까요?

InChiJun (approved):

src/storage/index_hnsw/hnsw_algo.hpp
hgryoo: 다른 PR에 의해 파라미터로 전달해주고 있어서, 함수 호출 대신 바로 사용하는 것으로 변경했습니다. 큰 이유는 없어요.

PR #6768, opened 2026-01-13
[CUBVEC-154] Add thread_entry parameter to hnsw_index interfaces

hgryoo:
http://jira.cubrid.org/browse/CUBVEC-154

### Purpose

when data is loaded in parallel using the loaddb -C option, concurrent data insertion into a vector index is not properly handled. In this scenario, multiple loader threads concurrently access a shared THREAD_ENTRY member variable inside the index object, which leads to race conditions and ultimately causes a core dump.

This PR enables safe parallel insertion of vector data into vector indexes created via loaddb -C.

### Implementation

- The cubhnsw::hnsw_index and cubhnsw::hnsw_usearch_ng class, which represents the HNSW index object context and concrete implementation respectively, no longer stores THREAD_ENTRY as a member variable.
- Instead, THREAD_ENTRY is passed explicitly as a parameter to each public interface function.
- This change ensures that each concurrent loader thread operates with its own THREAD_ENTRY, eliminating unsafe shared state and preventing crashes during parallel data loading.

### Remarks

N/A

### DIFF
--- a/src/storage/index_hnsw/hnsw.cpp
+++ b/src/storage/index_hnsw/hnsw.cpp
@@ -87,7 +87,7 @@ class hnsw_index_manager
     hnsw_index *get_index (const BTID *btid) const;
     int delete_index (const BTID *btid);
 
-    void print_index_info (const BTID *btid);
+    void print_index_info (THREAD_ENTRY *thread_p, const BTID *btid);
 
     // index management on disk
     int save_index (THREAD_ENTRY *thread_p, hnsw_index *index);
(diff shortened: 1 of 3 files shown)


YeunjunLee (approved):

hyahong (approved):

hornetmj:
@codex 한글로 리뷰해줘.

hornetmj (approved):

youngjun9072 (approved):

PR #6936, opened 2026-03-24
[CBRD-26598] [CMS] remove forcing insertion of 3 keys, _DBNAME/_DBID/_DBPASSWD

kisoo-han:
<http://jira.cubrid.org/browse/CBRD-26598>

### Purpose
* Removes the function to forcibly insert the three keys, **_DBNAME**/**_DBID**/**_DBPASSWD**.
* They will be changed to **dbname**, **dbuser**, and **dbpassword**, respectively.
for example, 
* currently, the **_classinfo_** CMS API requires only the following two keys, and replaces **_DBID** and **_DBPASSWD** with automatically inserted information for the user/password,
{dbname, dbstatus}
* afterwards, users must explicitly enter dbuser and dbpasswd as shown below:
{dbname, **dbuser**, **dbpassword**, dbstatus}

### Implementation

N/A

### Remarks
* check if the request has a key, dbuser, dbpassword, this is because we want distinguish 
 * dbpassword is present and value is null 
 * dbpassword is abesent

### DIFF
--- a/src/cm_common/cm_dep.h
+++ b/src/cm_common/cm_dep.h
@@ -269,6 +269,8 @@ extern "C"
   int make_temp_filename (char *tempfile, const char *prefix, int size);
   int make_temp_filepath (char *tempfile, char *tempdir, char *prefix, int task_code, int size);
 
+  bool key_not_exist (nvplist *req, const char *key);
+
 #ifdef __cplusplus
 }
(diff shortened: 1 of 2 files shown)


kisoo-han:
Hash code update planned for a separate PR.

PR #7044, opened 2026-04-14
[CBRD-26528] change to the criteria for restarting the CAS process

mhoh3963:
http://jira.cubrid.org/browse/CBRD-26528

* change in CAS process memory measurement : from virtual memory size to resident and swap memory size


mhoh3963:
이미 review 및 테스트가 완료된 이슈여서, 강제 머지합니다.

PR #7048, opened 2026-04-15
[CBRD-26710] Fix CAS coredump in authenticate_context and reset cache depth

mhoh3963:
http://jira.cubrid.org/browse/CBRD-26710

- Prevent a coredump in authenticate_context::set_user() by adding a null check for the value returned by au_get_user_name(). (Fixes crash when passing nullptr to std::string in make_user_cache)
- Fix a potential memory (class_cache->data) growth issue where cache_depth was not reset in reset_user_cache()


mhoh3963:
@greptile 리뷰해줘요.

hgryoo (approved):

PR #7081, opened 2026-04-21
[CBRD-26718] fix assertion failure in perfmon_add_at_offset()

mhoh3963:
http://jira.cubrid.org/browse/CBRD-26718

 - prevent assertion in perfmon_add_at_offset() by checking for valid transaction indices.
 - This addresses a race condition where a transaction ends before its interrupted state is processed.


mhoh3963:
@greptile 리뷰해줘요.

beyondykk9 (approved):

PR #7106, opened 2026-04-24
[CBRD-26745] flush deferred query handlers on net callback error path

hgryoo:
http://jira.cubrid.org/browse/CBRD-26745

### Purpose

                                                                                               
  CAS crashes with mspace_free abort in db_close_session_local while walking parser->host_variables. Stack always ends at callback_handler::free_deferred_query_handler → ~query_handler → close_and_free_session.                                                           
                                                                                                                                                                                                                                                                             
  net_client_request_method_callback (and _with_callback) only flush the process-wide callback_handler::m_deferred_query_free_handler on the normal loop exit path. The return set_server_error(...) early-returns on socket receive errors skip the flush, leaving a query_handler * with a freed DB_SESSION in the singleton queue. The next client request on the same CAS (even a plain SELECT 1) reaches the flush at end-of-RPC and deletes the stale handler → double free on host_variables payload → heaplayers abort.                  
                                                                                                                                                                                                                                                                             
  A second variant of the same root cause shows up at CAS shutdown: query_handler entries cached in m_query_handlers (and the deferred queue) outlive ws_final(), so their dtor's db_close_session_local() walks an already-destroyed workspace heap.

  Triggers in the wild: cub_pl_server / cub_server restart or crash, HA failover, JDBC queryTimeout, Statement.cancel, TCP idle timeout on LB/NAT, container overlay flap, etc.    

### Implementation

RPC early-return path (src/communication/network_cl.c)
- Replace the single end-of-function flush in net_client_request_with_callback / net_client_request_method_callback with an RAII deferred_flush_guard declared at function entry. Its dtor runs callback_handler::free_deferred_query_handler() on every return path (socket  errors, allocation failures, normal exit) when !tran_is_in_libcas().

Fast-path guard (src/method/method_callback.{hpp,cpp})
- Add callback_handler::has_deferred_query_handler() so the RAII guard skips the libcas-depth check and the flush call when the deferred queue is empty (the common case).

CAS shutdown drain (src/method/method_callback.{hpp,cpp}, src/transaction/boot_cl.c)
- Add callback_handler::clear_all_query_handlers() which drains both m_deferred_query_free_handler and the live m_query_handlers cache (and clears m_sql_handler_map / m_qid_handler_map). Expose it via C wrapper method_callback_final() and call it from boot_client_all_finalize() immediately before ws_final(), so no query_handler outlives the workspace heap.                                                                                                                                                                 


### Remarks

A fault injection via env `CBRD26745_INJECT_EVERY=N` (not in this PR) forces the early-return path every N-th `pl_call`.:
  - Before fix: 8/8 CAS coredump, stack matches the ticket                                                                 
  - After fix: 0 coredump on the same workload

### DIFF
--- a/src/communication/network_cl.c
+++ b/src/communication/network_cl.c
@@ -77,13 +77,13 @@ namespace
   {
     ~deferred_flush_guard ()
     {
-      cubmethod::callback_handler *h = cubmethod::get_callback_handler ();
+      cubmethod::callback_handler * h = cubmethod::get_callback_handler ();
       /* fast path: skip the libcas-depth check and the flush call entirely
          when no PL callback pushed anything to the queue. */
       if (h->has_deferred_query_handler () && !tran_is_in_libcas ())
-        {
-          h->free_deferred_query_handler ();
-        }
+	{
+	  h->free_deferred_query_handler ();
+	}
     }
   };
 }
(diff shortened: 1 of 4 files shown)


beyondykk9:
code-style 다시 체크해주세요.

src/communication/network_cl.c
```diff
@@ -1888,6 +1910,23 @@ net_client_request_method_callback (int request, char *argbuf, int argsize, char
 
 		css_queue_receive_data_buffer (rc, replydata, replydata_size);
 		error = css_receive_data_from_server (rc, &reply, &size);
+		/* CBRD-26745 fault injection: every Nth pl_call force the early
+		   return path so a stale query_handler is left in the deferred
+		   queue. Set env CBRD26745_INJECT_EVERY=N to enable. */
+		{
+		  const char *_inj_env = getenv ("CBRD26745_INJECT_EVERY");
```
mhoh3963: greptile이 리뷰한 것처럼 injection 테스트는 DEBUG mode에서 동작하게 하는 것이 좋겠네요.

hgryoo: reproduction을 위해 임시로 작성한 코드입니다. 원인으로 파악되면 제거할 예정입니다.

beyondykk9 (approved):

hornetmj (approved):

PR #7123, opened 2026-04-28
Cbrd histogram update

soheejung-cs:
<http://jira.cubrid.org/browse/CBRD-XXXX>

### Purpose

N/A

### Implementation

N/A

### Remarks

N/A


soheejung-cs:
@claude 리뷰

PR #7309, opened 2026-06-16
[CBRD-26074] PL/CSQL table valued function

hyunikn:
<http://jira.cubrid.org/browse/CBRD-26074>

### Purpose

N/A

### Implementation

N/A

### Remarks

N/A


hyunikn:
@greptile 리뷰해 주세요.

